PR #33399, opened 2024-11-18
Un-Hardcode water evaporation & mopping behavior

Zap527:
## About the PR
Made evaporation and mopping logic more generic.

## Why / Balance
Water = "water"
On a more serious note, this is needed for downstreams, and hardcoding behavior with magic strings is not good.

## Technical details
Added two new parameters to reagents, to determine their evaporation speed and their absorbance status.
Removed hardcoded water evaporation and mopping behavior and made it depend on the reagent's evaporation speed/absorbence parameter.

## Requirements

## Breaking changes
`EvaporationReagents` was removed in favor of `GetEvaporatingReagents(solution)` and `GetAbsorbentReagents(solution)`

**Changelog**
Not player facing


dvir001:
@Zap527 Can you please make holy water also work the same as water in this PR?

Zap527:
I could, but I feel like this might be something which would be better suited for a follow-up PR. I don't think currently ice and holy water evaporate? Ideally I don't change game behavior with this.

Zap527:
@SlamBamActionman You said I should request a review when done, so here it is!

SlamBamActionman (approved):

Schrodinger71 (approved):
it looks good but my approval won't do anything.. t-t

Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
```diff
@@ -1,20 +1,19 @@
 using System.Collections.Frozen;
 using System.Linq;
+using Content.Shared.FixedPoint;
```
Quantum-cross, line 3: not sure how much people care about dumb things like this but you can make this using statement back to where it was do avoid diff bloat

Quantum-cross (commented):
Looks good, does this mean that we could theoretically set space cleaner evaporation rate higher and clean with that, causing puddles to evaporate faster?

Quantum-cross (approved):

Zap527:
> Looks good, does this mean that we could theoretically set space cleaner evaporation rate higher and clean with that, causing puddles to evaporate faster?

Yeah, if you make it so cleaning leaves space cleaner puddles behind.

PR #32986, opened 2024-10-24
Reduce network burden of the hunger system

Centronias:
## About the PR
Replace `HungerComponent`'s `CurrentHunger` with `HungerSystem.GetHunger`, which calculates the current hunger value from an initial value, a time when that initial value was set, and a constant decay rate.

See also #33403 

## Why / Balance
#28070

## Technical details
* `HungerComponent` no longer has `CurrentHunger`, and instead has `LastAuthoritativeHungerValue` and `LastAuthoritativeHungerChangeTime`
* Replaced all usages of `HungerComponent.CurrentHunger` with `HungerSystem.GetHunger`, which calculates the current hunger value by predicting the decay since the last authoritative update:
  * `LastAuthoritativeHungerValue - (IGameTiming.CurrTime - LastAuthoritativeHungerChangeTime) * ActualDecayRate`
* `HungerSystem.Update` now only looks to update the current hunger threshold (because there's no value to update)
* Added `HungerSystem.SetAuthoritativeHungerValue`, which sets `LastAuthoritativeHungerValue` and `LastAuthoritativeHungerChangeTime` and dirties the component, causing those values to get pushed to the client, resetting the "current" calculation's basepoint.
* `HungerSystem.SetHunger` and anything that modifies the current threshold call `HungerSystem.SetAuthoritativeHungerValue` to ensure any unpredictable changes are replicated to the client.

## Media
*Before:*
(With non-game-logic-altering modifications made to raise an event on `HungerComponent` replication + popup creation when getting that event)

https://github.com/user-attachments/assets/f1559b39-e362-4091-8652-d84a84155338

*After:*
(With non-game-logic-altering modifications made to raise an event on `HungerComponent` replication + popup creation when getting that event; and create a popup on every `HungerSystem` update tick)

https://github.com/user-attachments/assets/9409118c-c159-4b9a-826f-f6412932a8a8

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
* `HungerComponent`'s `CurrentHunger` no longer exists and must be replaced by a call to `HungerSystem.GetHunger`

**Changelog**
My understanding is that this is a totally under-the-hood change, so no CL.

Issue #28070, opened 2024-05-16
Hunger + thirst comp states need optimising


metalgearsloth:
Right now it does the stupid 1 second update rate which spams data out.

Instead it should just send starting value, last changed time + change rate and client can infer the current value from this data and save thousands of states from going out.

PR #33403, opened 2024-11-18
Reduce network burden of thirst system

Centronias:
## About the PR
Replace `ThirstComponent`'s `CurrentThirst` with `ThirstSystem.GetThirst`, which calculates the current thirst value from an initial value, a time when that initial value was set, and a constant decay rate.

See also #32986 

## Why / Balance
#28070 

## Technical details
* `ThirstComponent` no longer has `CurrentThirst`, and instead has `LastAuthoritativeThirstValue` and `LastAuthoritativeThirstChangeTime`
* Replaced all usages of `ThirstComponent.CurrentThirst` with `ThirstSystem.GetThirst`, which calculates the current thirst value by predicting the decay since the last authoritative update:
  * `LastAuthoritativeThirstValue - (IGameTiming.CurrTime - LastAuthoritativeThirstChangeTime) * ActualDecayRate`
* `ThirstSystem.Update` now only looks to update the current thirst threshold (because there's no value to update)
* Added `ThirstSystem.SetAuthoritativeThirstValue`, which sets `LastAuthoritativeThirstValue` and `LastAuthoritativeThirstChangeTime` and dirties the component, causing those values to get pushed to the client, resetting the "current" calculation's basepoint.
* `ThirstSystem.SetThirst` and anything that modifies the current threshold call `ThirstSystem.SetAuthoritativeThirstValue` to ensure any unpredictable changes are replicated to the client.

## Media
(With non-game-logic-altering modifications made to raise an event on `ThirstComponent` replication + popup creation when getting that event)

https://github.com/user-attachments/assets/2e6d1125-1ddd-4bbb-8a5b-f828bd32cbfe

## Requirements

## Breaking changes
* `ThirstComponent`'s `CurrentThirst` is gone. It needs to be replaced by:
  * `ThirstSystem.GetThirst` in C#
  * `lastAuthoritativeThirstValue` in YAML (eg. for initial thirst values on prototypes)

**Changelog**
Technical implementation changes, no CL.


Content.Shared/Nutrition/Components/ThirstComponent.cs
```diff
@@ -26,45 +26,56 @@ public sealed partial class ThirstComponent : Component
...
     public ThirstThreshold LastThirstThreshold;
 
-    [ViewVariables(VVAccess.ReadWrite)]
-    [DataField("startingThirst")]
+    /// <summary>
+    /// The thirst value as authoritatively set by the server as of <see cref="LastAuthoritativeThirstChangeTime"/>.
+    /// This value should be updated relatively infrequently. To get the current thirst, which changes with each update,
+    /// use <see cref="ThirstSystem.GetThirst"/>.
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadOnly)]
```
MilonPL, line 34: Declare attributes on the same line, remove `ViewVariables`

Content.Shared/Nutrition/Components/ThirstComponent.cs
```diff
@@ -26,45 +26,56 @@ public sealed partial class ThirstComponent : Component
...
+    [AutoNetworkedField]
+    public float LastAuthoritativeThirstValue = -1.0f;
+
+    /// <summary>
+    /// The time at which <see cref="LastAuthoritativeThirstValue"/> was last updated.
+    /// </summary>
+    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer)), ViewVariables(VVAccess.ReadWrite)]
```
MilonPL, line 41: AutoPausedField, remove ViewVariables

Centronias: Can you explain why the `ViewVariables` annotations should be removed, please?

Tayrtahn: The DataField attribute automatically gives ReadWrite view variables access, so the ViewVariables attribute is redundant here. 

This wasn't always the case, so you'll see it being added manually in older code, but it's the standard going forward.

Centronias: Ahh, makes sense. Somebody ought to put that in the doc of `DataField`, and update the RA inspection...

Content.Shared/Nutrition/Components/ThirstComponent.cs
Centronias: In this case, `ViewVariables` is necessary because it explicitly prevents this field from being edited.

Centronias:
Closing in favor of https://github.com/space-wizards/space-station-14/pull/34166

Issue #31363, opened 2024-08-23
There is no responsiveness clicking on vending machine buttons


UbaserB:
## Description

I don't know if it was like this before, but it feels weird not having the button not feel like you're clicking on it (almost as if there's a delay or something)

lzk228:
real, just was playing and noticed this

metalgearsloth:
Working on this.

PR #33412, opened 2024-11-19
Predict vending machine UI

metalgearsloth:
At least the sound + amount. Not the entity itself.

https://github.com/user-attachments/assets/02934846-0743-4d15-a6fb-ffb8f822718f

Not sure on the button still working while disabled but seems like a listcontainer bug. Not sure if we should get a listcontainer with the correct styles either so we don't have to do it manually? I think cargo + one other thing uses this too

Resolves https://github.com/space-wizards/space-station-14/issues/31363

Breaking change:
- speedLimiter datafield changed to canShoot for vendingmachinecomponent
- Namespace changes for vendingmachine related activities.

:cl:
- tweak: Predict vending machine UI and have the buttons reflect whether the UI is useable.


Content.Shared/VendingMachines/SharedVendingMachineSystem.cs
```diff
@@ -1,37 +1,302 @@
...
+        if (!Resolve(uid, ref vendComponent))
+            return false;
+
+        if (!TryComp<AccessReaderComponent>(uid, out var accessReader))
+            return true;
+
+        if (_accessReader.IsAllowed(sender, uid, accessReader) || HasComp<EmaggedComponent>(uid))
+            return true;
+
+        Popup.PopupClient(Loc.GetString("vending-machine-component-try-eject-access-denied"), uid);
```
slarticodefast, line 163:
```suggestion
        Popup.PopupEntity(Loc.GetString("vending-machine-component-try-eject-access-denied"), uid);
```
The popup currently does not show up. For `PopupClient` the second parameter is the recipient, not the entity the popup is shown above. Maybe `PopupPredicted` works here as well and should be used instead, not sure.

slarticodefast (changes requested):
Found some bugs while testing:
- The last item in the inventory list seems to be always missing
- In some cases you cannot scroll to the bottom of the list without resizing the window
grafik
grafik

Content.Client/Advertise/Systems/SpeakOnUIClosedSystem.cs
```diff
@@ -0,0 +1,8 @@
+using Content.Shared.Advertise.Systems;
+
+namespace Content.Client.Advertise.Systems;
+
+public sealed class SpeakOnUIClosedSystem : SharedSpeakOnUIClosedSystem
```
MilonPL, line 5: Just put a semicolon at the end of this line, doesn't need a body declared.

Content.Client/VendingMachines/UI/VendingMachineMenu.xaml.cs
```diff
@@ -9,6 +10,7 @@
 using Content.Client.UserInterface.Controls;
 using Content.Shared.IdentityManagement;
 using Robust.Client.Graphics;
+using Robust.Shared.Timing;
```
MilonPL, line 13: Unused.

Content.Client/VendingMachines/UI/VendingMachineMenu.xaml.cs
```diff
@@ -68,18 +77,23 @@ private void GenerateButton(ListData data, ListContainerButton button)
             if (data is not VendorItemsListData { ItemProtoID: var protoID, ItemText: var text })
```
MilonPL, line 77: Might want to rename it to `protoId` if you're touching this.

Content.Client/VendingMachines/UI/VendingMachineMenu.xaml.cs
```diff
@@ -119,18 +136,50 @@ public void Populate(List<VendingMachineInventoryEntry> inventory)
...
         }
 
+        /// <summary>
+        /// Updates text entries for vending data in place without modifying the list controls.
+        /// </summary>
+        public void UpdateAmounts(List<VendingMachineInventoryEntry> cachedInventory, bool enabled)
+        {
+            _enabled = enabled;
+
+            foreach (var (proto, uid) in _dummies)
```
MilonPL, line 162: `uid` is never being used.

Content.Client/VendingMachines/UI/VendingMachineMenu.xaml.cs
```diff
@@ -139,4 +188,7 @@ private void SetSizeAfterUpdate(int longestEntryLength, int contentCount)
     }
 }
 
-public record VendorItemsListData(EntProtoId ItemProtoID, string ItemText, int ItemIndex) : ListData;
+public record VendorItemsListData(EntProtoId ItemProtoID, int ItemIndex) : ListData
```
MilonPL, line 191: Since we're not using a file scoped namespace in this file (the curse of old code), this will not be namespaced properly. Might want to move it up.

Content.Client/UserInterface/Controls/ListContainer.cs
```diff
@@ -382,8 +382,9 @@ public sealed class ListContainerButton : ContainerButton, IEntityControl
     public readonly int Index;
     // public PanelContainer Background;
 
-    public ListContainerButton(ListData data, int index)
+    public ListContainerButton(ListData data, int index) : base()
```
MilonPL, line 385: Is there any reason to call `base()` here?

Content.Client/VendingMachines/VendingMachineSystem.cs
```diff
@@ -1,29 +1,82 @@
...
         base.Initialize();
 
         SubscribeLocalEvent<VendingMachineComponent, AppearanceChangeEvent>(OnAppearanceChange);
         SubscribeLocalEvent<VendingMachineComponent, AnimationCompletedEvent>(OnAnimationCompleted);
-        SubscribeLocalEvent<VendingMachineComponent, AfterAutoHandleStateEvent>(OnVendingAfterState);
+        SubscribeLocalEvent<VendingMachineComponent, ComponentHandleState>(OnVendingHandleState);
     }
 
-    private void OnVendingAfterState(EntityUid uid, VendingMachineComponent component, ref AfterAutoHandleStateEvent args)
+    private void OnVendingHandleState(EntityUid uid, VendingMachineComponent component, ref ComponentHandleState args)
```
MilonPL, line 23: Can this use `Entity<T>` instead?

Content.Shared/VendingMachines/SharedVendingMachineSystem.cs
```diff
@@ -1,37 +1,302 @@
...
         SubscribeLocalEvent<VendingMachineComponent, MapInitEvent>(OnMapInit);
         SubscribeLocalEvent<VendingMachineRestockComponent, AfterInteractEvent>(OnAfterInteract);
+
+        Subs.BuiEvents<VendingMachineComponent>(VendingMachineUiKey.Key, subs =>
+        {
+            subs.Event<VendingMachineEjectMessage>(OnInventoryEjectMessage);
+        });
+    }
+
+    private void OnVendingGetState(EntityUid uid, VendingMachineComponent component, ref ComponentGetState args)
```
MilonPL, line 52: This doesn't have any overrides so should be fine to just use `Entity<T>`.

Content.Shared/VendingMachines/SharedVendingMachineSystem.cs
```diff
@@ -1,37 +1,302 @@
...
+                if (curTime > comp.DispenseOnHitEnd)
+                {
+                    comp.DispenseOnHitEnd = null;
+                    Dirty(uid, comp);
+                }
+            }
+        }
+    }
+
+    private void OnInventoryEjectMessage(EntityUid uid, VendingMachineComponent component, VendingMachineEjectMessage args)
```
MilonPL, line 128: Above.

Content.Shared/VendingMachines/VendingMachineComponent.cs
```diff
@@ -22,31 +24,45 @@ public sealed partial class VendingMachineComponent : Component
...
+        public bool Ejecting => EjectEnd != null;
+
+        [ViewVariables]
+        public bool Denying => DenyEnd != null;
+
+        [ViewVariables]
+        public bool DispenseOnHitCoolingDown => DispenseOnHitEnd != null;
+
+        [DataField]
+        public TimeSpan? EjectEnd;
```
MilonPL, line 59: These are runtime-modified, is there a reason to not using a TimeOffsetSerializer?

MilonPL (commented):
Mostly covered by the other review, minor nitpicks. 
Edit: add the PublicAPI attribute to public methods!

Content.Shared/VendingMachines/SharedVendingMachineSystem.cs
```diff
@@ -1,37 +1,302 @@
...
+        if (!Resolve(entity.Owner, ref entity.Comp))
+            return;
+
+        if (entity.Comp.Denying)
+            return;
+
+        entity.Comp.DenyEnd = Timing.CurTime + entity.Comp.DenyDelay;
+        Audio.PlayPredicted(entity.Comp.SoundDeny, entity.Owner, user, AudioParams.Default.WithVolume(-2f));
+        TryUpdateVisualState(entity);
+        Dirty(entity);
```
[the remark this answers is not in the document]
metalgearsloth: Dirty just flags the component for PVS which runs entirely out of band after all systems have run (via servergamestatemanager or via clientgamestatemanager due to funny prediction reasons).

Content.Shared/VendingMachines/VendingMachineComponent.cs
```diff
@@ -254,4 +273,22 @@ public sealed partial class VendingMachineSelfDispenseEvent : InstantActionEvent
     {
 
     };
+
+    [Serializable, NetSerializable]
+    public sealed class VendingMachineComponentState : ComponentState
```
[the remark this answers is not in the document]
metalgearsloth: Dictionaries aren't copied properly so when the client re-runs states it mutates the componentstate and causes bad times (I have an engine PR but have been lazy on finishing it).

slarticodefast:
Still can't grab the last item without resizing the window. This will need to be fixed first.
example

PR #31939, opened 2024-09-08
Make fire leave burnt decals on the tiles

Winkarst-cpu:
## About the PR
Made fire leave burnt decals on the tiles.

## Why / Balance
Resolves #28303.

## Technical details
Code has comments.

## Media

https://github.com/user-attachments/assets/f0b4b06d-72dd-43b5-8245-66a178592724

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: Now fire leaves burn marks on the tiles that were affected by it.

PR #33422, opened 2024-11-20
Removes burnt tiles

Flareguy:
## About the PR
Removes burnt tiles & burnt plating from the game. They have been migrated to their damaged variants.

## Why / Balance
#31939 added tile burning behavior, but it uses decals instead of these tiles. There's no point in having duplicate versions of these that don't actually reflect how these tiles look when burnt by fire.

The decal versions are arguably very slightly subpar, but it certainly isn't by much, and the burn marks don't really rely on the shape of the tile the way the damaged ones do currently. The burnt look present on them could therefore be removed from the tiles and turned in to a decal if desired.

## Requirements

## Breaking changes
Burnt tiles & burnt plating have been removed and migrated to their damaged counterparts. Mappers should check their map if they used burnt tiles extensively and replace them with the appropriate burn mark decals. If you've just used them as another kind of damaged tile, though, you'll probably be fine.

**Changelog**
no cl no fun


lzk228:
so, now i can't clean burned tiles with space cleaner?

TheShuEd (approved):

PR #33429, opened 2024-11-20
Replace glass_knock.ogg

Shaddap1:
## About the PR
Replace glass_knock.ogg with a less clipped version New version has peak compression, reduced high-end frequencies, reverb moved to low-end, and a crossfade. 

## Why / Balance
The original glass_knock.ogg file was recorded poorly and clipped.  This produces an unpleasant clicking sound that most people will never notice or care about.

The result is a reduced clicking sound to due to clipping, reduced clicking when looping the file, and a less aggressive, somewhat muddier sound. Knocking will likely sound somewhat less frantic.

## Technical details
replaced glass_knock.ogg with a newer version

## Media
I cannot attach .ogg files directly in discord so here is a link to a discord message with both versions: https://discord.com/channels/310555209753690112/770682801607278632/1308749760798130226

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- tweak: replaced window knocking sound with a less clipped, more subtle version


Aeshus (changes requested):
Make sure to update the attribution!

Shaddap1:
> Make sure to update the attribution!

I have no idea what this means T_T

Aeshus:
> I have no idea what this means T_T

There should be an entry in [/Resources/Audio/Effects/attributions.yml](https://github.com/space-wizards/space-station-14/blob/master/Resources/Audio/Effects/attributions.yml) that lists the copyright information for the sound effect. The spec is [here](https://docs.spacestation14.com/en/specifications/robust-generic-attribution.html) if you're wondering. 

I am mildly concerned that I can't seem to find the original attribution for `glass_knock.ogg`, but there's no time like the present!

Shaddap1:
> > I have no idea what this means T_T
> 
> There should be an entry in [/Resources/Audio/Effects/attributions.yml](https://github.com/space-wizards/space-station-14/blob/master/Resources/Audio/Effects/attributions.yml) that lists the copyright information for the sound effect. The spec is [here](https://docs.spacestation14.com/en/specifications/robust-generic-attribution.html) if you're wondering.
> 
> I am mildly concerned that I can't seem to find the original attribution for `glass_knock.ogg` (it's _probably_ from cmss13, as many of the other glass sound effects originated from there, though you should probably find the providence on your own), but there's no time like the present!

Thanks.  I did find the same file in cmss13, but their github doesnt have an attribution for it.  I really have no idea how I would track down the original and first use of the file.  Should I just go ahead with a cmss13 attribution?

Aeshus:
> Thanks. I did find the same file in cmss13, but their github doesnt have an attribution for it. I really have no idea how I would track down the original and first use of the file. Should I just go ahead with a cmss13 attribution?

Yep! Cite it as originally coming from cmss13, and then cite yourself for editing/modifying it.

Example:
https://github.com/space-wizards/space-station-14/blob/cb246f5d7d79ba8c8b7a250a063abf17f0e6d969/Resources/Audio/Items/attributions.yml#L51-L59

metalgearsloth (changes requested):
> Thanks. I did find the same file in cmss13, but their github doesnt have an attribution for it. I really have no idea how I would track down the original and first use of the file. Should I just go ahead with a cmss13 attribution?

Last I heard a significant number of cm13 files were illegally obtained from AvP or sometimes Alien Isolation so this will need triple checking.

Aeshus:
> Last I heard a significant number of cm13 files were illegally obtained from AvP or sometimes Alien Isolation so this will need triple checking.

Sounds like we have a bigger situation on our hands...

Issue #33316, opened 2024-11-14
Until a player switches hands, names won't be shown for items in their inactive hand


MossyGreySlope:
## Description
After connecting to a server, if a player inserts any item into their non-active hand it will not have it's name shown. Instead, the sprite will be shown as normal, but the text will read "*No held item*". Additionally, if the player starts with an item in their inactive hand (probably after disconnecting), it's name will also be hidden.
The player can also remove the item from their off-hand (take it with the active hand) and then insert a different item and the hand will still read "*No held item*". This behavior will continue until the player switches hands, and then the UI will work as normal.

**Reproduction**
- Connect.
- Pick up any item.
- Switch hands.
- Disconnect.
- Reconnect.
- Look at the hands UI.

**Videos**

https://github.com/user-attachments/assets/05aeb58e-4695-4a33-afe4-5b19193638ea

https://github.com/user-attachments/assets/07190ffa-2580-451d-88d1-4f8227db7d80

https://github.com/user-attachments/assets/1ac9d2d7-4170-4391-90c6-45b958d85c64

PR #33431, opened 2024-11-20
Fix dev server crashes caused by putting pills in drinks, depsiting spesos, making burgers

MossyGreySlope:
## About the PR
Several interactions that delete the in-hand item didn't mark the triggering even as handled, leading to the debug server crashing. This fixes that issue for adding pills to a solution, depositing spesos into the cargo console, and adding ingredients to a custom burger. There's probably other cases of this bug that still need to be fixed.

This PR is basically just https://github.com/space-wizards/space-station-14/issues/33316 applied to another three systems.

## Technical details
- Combined the `TrySpike` and `OnInteractUsing` functions in `SolutionSpikerSystem`, since `OnInteractUsing` was a trivial wrapper and `TrySpike` wasn't used anywhere else.
- Added `Handled = true` to `SolutionSpikerSystem.OnInteractUsing`, though only when the spiking item is actauly removed. There aren't any cases in the game where that item *isn't* deleted, so I'm not 100% sure what the correct behavior would be in that case.
- Added `Handled = true` to `CargoSystem.OnInteractUsing`.
- Assigned the return value of `TryAddFoodElement` to `Handled` in `SharedFoodSequenceSystem.OnInteractUsing`. 

## Media
burger
lobbying
spike

## Requirements

**Changelog**
No changelog - minor bugfixes.

### DIFF
--- a/Content.Server/Cargo/Systems/CargoSystem.Orders.cs
+++ b/Content.Server/Cargo/Systems/CargoSystem.Orders.cs
@@ -62,6 +62,7 @@ namespace Content.Server.Cargo.Systems
             _audio.PlayPvs(component.ConfirmSound, uid);
             UpdateBankAccount(stationUid.Value, bank, (int) price);
             QueueDel(args.Used);
+            args.Handled = true;
         }
 
         private void OnInit(EntityUid uid, CargoOrderConsoleComponent orderConsole, ComponentInit args)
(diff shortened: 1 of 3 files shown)


slarticodefast:
Two of these got already fixed in other PRs, but the burger one was still missing.
I went ahead and fixed the merge conflict for you.

Tayrtahn (approved):

PR #33443, opened 2024-11-21
Remove steel sheet hull fixing

SaphireLattice:
## About the PR
Changes the steel sheets to not be able to place down Plating (base hull tiles)

## Why / Balance
It's unintuitive that it's wasteful, there's not really any systems to NOT make it wasteful, and it's a UX trap. No this is not "noob trap", no. It's an UX issue because if hundreds of people do it over and over again without noticing anything wrong, well, something fails to properly note that there's an issue with what they are doing. Aka user experience issue. Aka not a problem with the user but the design.

Also could be a shitty doafter action that makes you wonder if something's wrong

Personally: I hate this PR and I think we need to fix the part where people avoid the crafting and building menus. Yes, building menus have an issue too. Otherwise people wouldn't be building walls with RCD all the time. "Ran out of materials" might be a common reason, but I suspect a lot of people just never *discover* the damn menu until someone yells at them - which sprinkles an already awful UX sandwich with potential community interaction issues. Because who the hell would be happy to be told that the action they thought was fine and intuitive is actually "bad", or in some cases that can be directed at the player.

So yeah. This PR kinda sucks. I wish I could just fix the UX. The amount of work for that is enormous. For now we can at least axe this whole "feature" to avoid people fighting over "being wasteful"/etc. Because with current system there's **ZERO** discovery on player side that this perfectly quick action they can do is, in fact, not optimal. No messaging at all. No feedback.

Axe the feature, don't blame the people for using it.

## Technical details
YAML

## Media
Imagine a video of an engineer failing to use steel sheets on a lattice here

## Requirements

**Changelog**
:cl:
- remove: Steel sheets no longer have unintuitively wasteful plating placement that can be done with steel tiles instead

### DIFF
--- a/Content.IntegrationTests/Tests/Tiles/TileConstructionTests.cs
+++ b/Content.IntegrationTests/Tests/Tiles/TileConstructionTests.cs
@@ -82,7 +82,7 @@ public sealed class TileConstructionTests : InteractionTest
         AssertGridCount(1);
 
         // Lattice -> Plating
-        await InteractUsing(Steel);
+        await InteractUsing(FloorItem);
         Assert.That(Hands.ActiveHandEntity, Is.Null);
         await AssertTile(Plating);
         AssertGridCount(1);
--- a/Resources/Prototypes/Entities/Objects/Materials/Sheets/metal.yml
+++ b/Resources/Prototypes/Entities/Objects/Materials/Sheets/metal.yml
@@ -56,9 +56,6 @@
   - type: Item
     heldPrefix: steel
   - type: Appearance
-  - type: FloorTile
-    outputs:
-    - Plating
   - type: Extractable
     grindableSolutionName: steel
   - type: SolutionContainerManager


a-person5660:
this has never been or seemed unintuitive to me or anyone i know, i dont understand why this needs to go.

SixplyDev:
Why would this be axed at all? It's not unintuitive and is actually useful for stuff. I don't understand why this PR was posted in the first place if you hate it either?

metalgearsloth:
> Why would this be axed at all?

> i dont understand why this needs to go.

> It's unintuitive that it's wasteful, there's not really any systems to NOT make it wasteful, and it's a UX trap

Wasting 4x as much materials isn't really intuitive.

Merrokitsune:
We could have a timer, so not only the player knows what they’re doing before placing a tile, but also let’s another person know what the player is doing because a timer loading bar usually pops up whenever a person is placing a tile.

SaphireLattice:
Systems to make this less of a UX trap could and should be implemented. But right now it's better for it to go until such is in place.

.his is not a feature anyone actually uses, except occasional emergency... where you end up wasting 4x the material with no warning, and are likely to run out of material overall. And even in an emergency, you either need to run away quickly, or you have enough time to craft the tiles.

Alternative to this PR in my mind: proper signaling of this not being a "preferred" way to create hull. Something like a do-after that takes a solid while - a bit shorter than panicked menu flailing but definitely NOT instant. Metal and other materials used for construction giving players the construction window filtered by what uses it. Categories and generally better UX of construction menu

LankLTE:
I think needing to craft steel tiles to put down new hulls is also pretty unintuitive. For a long time I didn’t even know you could do that because laying tile with steel makes sense & I assume floor tile was.. floor tiles that you put over the hull.

Ilya246:
just put floor building into the guidebook

Callmore:
yknow i originally was annoyed by this pr but i read and thought about it a bit

yeah no it's completely unreasonable that you can waste metal sheets like that with no warning, plus it's not intuitive to put a raw material onto the floor to create a flooring, as the floor rod pr showed

only suggestion if it's not already done is maybe a text popup saying to use floor tiles instead of sheets if you attempt to use sheets

hyphenationc:
The crafting system using radial menus sounds like *hell*

I agree a client popup could be useful for this, though how it'd be added is... unclear.

SaphireLattice:
> I don't really like the concept of a do-after because it:
> 
>     * Invalidates the speediness tradeoff of using steel versus tile. Engineers sometimes use steel to patch holes quickly to stop in-progress spacing rather than waste time using hull tile.
> 
>     * Still does not properly communicate to the player that there is a better way of doing it. A player can still continue with the do-after and think "well this is just how it is". There is zero intuitive connection between floor tiles and being able to use them for hull plating.
> 
> 
> The steel to hull plating recipe should probably be removed. The descriptions of all of the tiling should also have a "can be used to fill in lattices with hull plating" appended to them.
> 
> I will note the build interactions for steel rods and steel tile in my engineering guidebook update when I get around to that.

Yeah, pretty much same thoughts and concerns. Guess it's just better to remove it? Even though it's a neat interaction that you can use in a rush, it's too much of a UX issue to keep around. And rebalancing everything around it would be difficult. You would need to change every metal sheet related thing to be 4x expensive.

qwertyyoppit:
what if you get 3 steel tiles back after using steel for a hull? no waste and it should tell the player that a single tile is the cost for a hull

hyphenationc:
> what if you get 3 steel tiles back after using steel for a hull? no waste and it should tell the player that a single tile is the cost for a hull

This was brought up at some point... don't remember where. I think the conclusion was it's more annoying to implement and didn't fix the core of the issue, which itself is in progress.

PR #31625, opened 2024-08-29
Fix sprite layering upon buckling/unbuckling

themias:
## About the PR

Fixes #31583 

## Technical details

We already have logic for updating the draw depth of a character upon rotating in a chair. This adds that same logic when buckling into or out of a chair.

This way, unbuckling from a chair resets the draw depth, and buckling into a north-facing chair will lower the draw depth appropriately.

## Media

https://github.com/user-attachments/assets/fdf6be00-779c-403d-8528-f8cc95c9fe48

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl:
- fix: Fixed players sometimes rendering under tables

PR #33447, opened 2024-11-21
Use screen local north for buckling/unbuckling layering, plus #31625

SaphireLattice:
## About the PR
It's #31625 but with screen rotation rather than just local

## Why / Balance
Allows people to hide under tables, kinda sorta

## Technical details
It honestly need engine to support rotation specific draw depth modifiers, or layer specific or whatever the heck. Point is: this is jank and it will not properly handle any sort of rotation updating. Rotating the camera means all the current draw depths are wrong. If you look at another grid rotating, it will also be wrong after a bit. At least it will just get reset when people unbuckle now.

## Requirements

**Changelog**

:cl: SaphireLattice, themias
- fix: Chair buckle draw depth is slightly less broken


lzk228:
looks like your cl is broken

beck-thompson (approved):
Tested and make sense! If something goes horribly wrong we can always revert :laughing:

PR #33460, opened 2024-11-22
Make going/being critical more noticeable

### LINKED ISSUE #31199: There should be some way to differentiate between critical and suffocating while bucked.
## Description

It's impossible to tell unless you drag them off of the chair. This lead to everyone around me believing that i was just afk, even my very own pAI.

**Screenshots**

image

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Resolves #31199.

## Media

https://github.com/user-attachments/assets/d32647ce-bc5c-4487-a397-bca67dc616bb

## Requirements

## Breaking changes

- DeathgaspSystem was renamed into MobsSystem and was moved into Systems folder.
- DeathgaspComponent was moved into Components folder.
- CritMobActionsSystem.cs was moved into Systems folder.

**Changelog**

:cl:
- add: Now mobs send an emote upon getting into the crit state. Also, you can tell they are in crit by examining them.


Resources/Locale/en-US/unconscious.ftl
lzk228, line 1: i think this file should be in mob states folder or something

lzk228 (approved):
looks fine, as i can tell

keronshb (changes requested):
The main issue with this is that it should never be this obvious. It ruins the idea of a fake death/feigning death. This is what a health analyzer is also for.

Content.Server/Mobs/Systems/MobsSystem.cs
```diff
@@ -0,0 +1,68 @@
...
+        if (!Resolve(uid, ref component, false))
+            return false;
+
+        if (HasComp<MutedComponent>(uid))
+            return false;
+
+        _chat.TryEmoteWithChat(uid, component.Prototype, ignoreActionBlocker: true);
+
+        return true;
+    }
```
Aeshus, line 51: I'm pretty sure it's the style now to merge the component and the EntityUid into one parameter (https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#public-api-method-signature).

As you're already making a breaking change by renaming the system, it's probably fine to change the method signatures to be more up to date with code-base conventions.

```suggestion
    public bool Deathgasp(Entity<DeathgaspComponent?> ent)
    {
        if (!Resolve(ent, ref ent.Comp, false))
            return false;

        if (HasComp<MutedComponent>(ent))
            return false;

        _chat.TryEmoteWithChat(ent, ent.Comp.Prototype, ignoreActionBlocker: true);

        return true;
    }
```

You'll also need to change the calls to `Deathgasp((uid, component))` if you resolve it before passing it in.

(Note that I'm not a maintainer, and I'm not 100% sure what's with the transition in style.)

Winkarst-cpu: Tbh I seen both approaches all over the codebase, so, idk.

Aeshus: From my understanding, the old approach is "on the way out."

The problem is that it's so entrenched that it'll take years at the current pace for all of them to be updated... (Plus it causes a public API change, which we generally want to avoid.)

Winkarst-cpu:
> The main issue with this is that it should never be this obvious. It ruins the idea of a fake death/feigning death. This is what a health analyzer is also for.

Can make it be shown only if player is in a detail examine range.

Aeshus:
> > The main issue with this is that it should never be this obvious. It ruins the idea of a fake death/feigning death. This is what a health analyzer is also for.
> 
> Can make it be shown only if player is in a detail examine range.

Can't you already see if they're actually dead by just inspecting them (it will lack the death information)?

Winkarst-cpu:
> Can't you already see if they're actually dead by just inspecting them (it will lack the death information)?

This is true too. If player is dead than you will see his rot level.

Resources/Locale/en-US/chat/emotes.ftl
```diff
@@ -62,3 +63,4 @@ chat-emote-msg-cathisses = hisses!
 chat-emote-msg-monkeyscreeches = screeches!
 chat-emote-msg-yawn = yawns.
 chat-emote-msg-snore = snores.
+chat-emote-msg-unconscious = staggered, then collapsed, unconscious...
```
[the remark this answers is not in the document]
lzk228: you meant in past simple?

HerCoyote23:
Note that popups are almost always in present tense, while this is in past tense.

So I think it'd be better if the text was: 
"staggers, then collapses, unconscious..."
rather than
"staggered, then collapsed, unconscious..."

LaCumbiaDelCoronavirus:
Does this have a 'cooldown'? Rapidly going into and out of crit (for example sometimes using a medipen while going into crit can cause that) can cause alot of spam if there isn't one (you can see this with tracking implants and security comms).

NoElkaTheGod:
> The main issue with this is that it should never be this obvious. It ruins the idea of a fake death/feigning death. This is what a health analyzer is also for.

But it is already this obvious. You can see if someone is dead by just examining them.

Pumkin69:
> The main issue with this is that it should never be this obvious. It ruins the idea of a fake death/feigning death. This is what a health analyzer is also for.

maybe when we actually have faking death or feigning death or crit but we already have ways of telling if people are actually dead or not even without a health analyzer so why wouldn't we have a way to tell if they are crit or not without a health analyzer?

slarticodefast:
Hey, I brought this up for maintainer discussion and we came to the conclusion that this should simply be solved by an examination text, like when being asleep or dead.
grafik

PR #33468, opened 2024-11-22
Fix Disposals Ejections

### LINKED ISSUE #31459: Items exiting disposals do not exit in the direction that the pipe is facing, if grid is rotated
## Description
It appears that the orientation of a disposal pipe during placement dictates the way that an item shot through it will exit. For instance, if I place a disposal unit and pipe facing north towards 0 degrees, and rotate the grid 90 degrees, any item shot out will exit towards 0 degrees, not 90.

**Reproduction**
On any shuttle with a disposal unit that exits to space, shoot something out of it and record the direction. Rotate the shuttle, and fire something else out it. The direction the items traveled should be the same. 

**Screenshots**
I'll attach some screenshots tomorrow

**Additional context**

sleepyyapril:
## About the PR

Disposals had two issues:
1. It did not respect the direction of a bend if it was the last piece in the disposals pipe.
2. The ejection would use the transform of the map instead of the one from the grid and throw into the wrong direction.

## Why / Balance

Closes #31459, probably.

## Technical details

DisposableSystem.cs's ExitDisposals now uses CurrentDirection to eject/throw people, unless CurrentDirection is equal to Direction.Invalid in which case it'll use PreviousDirection.
It also uses the grid's xform instead of the parent's, to better handle from-grid ejections rather than from-map.

## Media

Old behavior:
https://github.com/user-attachments/assets/bc819d34-cc27-41a9-85b4-02af36476a2a

New behavior:
https://github.com/user-attachments/assets/cadd65a2-1913-4be3-ae2f-a7f166da64bc
https://github.com/user-attachments/assets/6f42ead3-fd6d-4e1c-9ccd-3a6ff21a92a0

## Requirements

### DIFF
--- a/Content.Server/Disposal/Unit/EntitySystems/DisposableSystem.cs
+++ b/Content.Server/Disposal/Unit/EntitySystems/DisposableSystem.cs
@@ -137,10 +137,10 @@ namespace Content.Server.Disposal.Unit.EntitySystems
                     _xformSystem.AttachToGridOrMap(entity, xform);
                     var direction = holder.CurrentDirection == Direction.Invalid ? holder.PreviousDirection : holder.CurrentDirection;
 
-                    if (direction != Direction.Invalid && _xformQuery.TryGetComponent(gridUid, out var parentXform))
+                    if (direction != Direction.Invalid && _xformQuery.TryGetComponent(gridUid, out var gridXform))
                     {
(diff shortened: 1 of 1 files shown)


metalgearsloth (approved):

chromiumboy:
Looks good, tested in-game fine

PR #33477, opened 2024-11-23
Don't show drag-drop outline if climbing

metalgearsloth:
It won't actually do anything.

:cl:
- tweak: Drag-drop outline no longer shows the vaulting outlines if you're vaulting.


slarticodefast (approved):
Works as intended.

PR #33047, opened 2024-10-29
Simplify separated screen top menu

metalgearsloth:
After:
image

Okay so the problem is we try to fit a heap of buttons, with images, into a tiny space, which leads to the controls either spilling over the edge or being cramped. This is even worse if you're an admin and have even more buttons.. Realistically you don't need all of the vertical chat space on separated screen and we can just make this a gridcontainer instead.

:cl:
- tweak: Adjusted the top menu on the separated game screen. The buttons will now form multiple rows and no longer overflow into the viewport.

PR #33482, opened 2024-11-23
Fix Separated HUD bugs and set limits to keep UI pretty

### LINKED ISSUE #25494: Separated HUD Layout remembers your chat width, but does not start with it
## Description
If using oldchat, the chat tab will always start at the default width, only updating to your saved setting the first time an Alert changes

**Reproduction**
Switch to Separated, adjust the chat pane to some very noticeably oversized width
Restart game
Observe that Chat Tab is normal sized
Walk out into space
Observe that Chat Tab resizes the moment the pressure alert appears

https://github.com/space-wizards/space-station-14/assets/35878406/6c46a5a4-e0b8-48d3-813e-91ccac3c081d

SpaceManiac:
## About the PR
Sets limits on the widths of the left/right halves of the Separated HUD that avoid letterboxing and UI overlaps, and fixes some related bugs. It's still imperfect, in particular not taking into account non-integer viewport scales, but it should be an improvement over status quo.

## Technical details
1. Remember the selected width. Fixes #25494.
2. Fix top menu buttons overlapping each other at smaller chat widths.
3. Dynamically assign a min/max for the separated split based on several factors:
    - Minimum at 15-tile-wide viewport at current integer scale (480, 960, 1440)
    - Minimum at 1060 pixels, below which inventory UI elements begin to overlap
    - Soft maximum at 21-tile-wide viewport at current integer scale (672, 1344, 2016)
    - Soft maximum at screen width minus actual width of the chat
    - Hard maximum at screen width minus 372 pixels (width of the chat for non-admins)
4. Fix SplitContainer's clickable and visual widths desyncing after a UI scale settings change.

See also https://github.com/space-wizards/RobustToolbox/pull/5529

## Media
Before video:

https://github.com/user-attachments/assets/c05741d3-231a-490f-80f8-f8a8a8d38aa9

After video:

https://github.com/user-attachments/assets/1c764d40-9e84-4abc-8c31-36d0c1e99151

## Requirements

**Changelog**
:cl:
- tweak: Fix some visual bugs with the Separated HUD, and set limits that make letterboxing less likely.


Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
```diff
@@ -36,10 +60,53 @@ private void ResizeActionContainer()
...
+        var minFromChat = Width - _chatMinWidth;
+
+        var min = Math.Min(Math.Max(minFromVp, minFromUi), minFromChat);
+        var max = Math.Max(maxFromVp, min);
+
+        // SplitContainer doesn't respect MaxSize, so set a MinSize on the chat instead.
+        ViewportContainer.MinWidth = min;
+        SeparatedChatPanel.MinWidth = Math.Max(_chatMinWidth, Width - max);
+
+        if (_deferredSplitFraction is float fraction)
```
Aeshus, line 88: We generally avoid directly typing things if we don't have to, and this is functionally equivalent.

```suggestion
        if (_deferredSplitFraction is { } fraction)
```

Content.Client/UserInterface/Systems/MenuBar/Widgets/GameTopMenuBar.xaml
```diff
@@ -20,6 +20,7 @@
         BoundKey = "{x:Static ic:EngineKeyFunctions.EscapeMenu}"
         ToolTip="{Loc 'game-hud-open-escape-menu-button-tooltip'}"
         MinSize="70 64"
+        SizeFlagsStretchRatio="70"
```
Aeshus, line 23: SizeFlagStretchRatio just adds up all of them and then distributes it based on the ratio. Thus, we can simplify 70:42 to 5:3 and it should be the same.
 
(You'll need to fix the rest of them yourself)

```suggestion
        SizeFlagsStretchRatio="5"
```

Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
```diff
@@ -36,10 +60,53 @@ private void ResizeActionContainer()
         Actions.ActionsContainer.MaxGridWidth = ViewportContainer.Size.X - indent;
     }
 
+    private void RecalculateViewportDesiredSize()
+    {
+        // When the main window is reszied, calculate a new min/max for the
```
Aeshus, line 65:
```suggestion
        // When the main window is resized, calculate a new min/max for the
```

Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
```diff
@@ -36,10 +60,53 @@ private void ResizeActionContainer()
...
+        var uiScale = _cfg.GetCVar(CVars.DisplayUIScale);
+        if (uiScale == 0f)
+            uiScale = _uiManager.DefaultUIScale;
+
+        // CurrentRenderScale would be more correct, but it gets updated after us.
+        var gameScale = MainViewport.Viewport.FixedRenderScale;
+        var minFromVp = EyeManager.PixelsPerMeter * gameScale * _cfg.GetCVar(CCVars.ViewportMinimumWidth) / uiScale;
+        var maxFromVp = EyeManager.PixelsPerMeter * gameScale * _cfg.GetCVar(CCVars.ViewportMaximumWidth) / uiScale;
+        // Inventory bar also supplies a minWidth such that body+shoes don't overlap PDA.
+        var minFromUi = 1060;
```
Aeshus, line 77: Can you tell us exactly this comes from? 

If you're gonna define constants, it would probably be better to actually make them such and document exactly where they come from (e.g. manually collected).

Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
```diff
@@ -1,14 +1,25 @@
...
 using Robust.Client.UserInterface.XAML;
+using Robust.Shared;
+using Robust.Shared.Configuration;
 
 namespace Content.Client.UserInterface.Screens;
 
 [GenerateTypedNameReferences]
 public sealed partial class SeparatedChatGameScreen : InGameScreen
 {
+    private readonly IConfigurationManager _cfg = IoCManager.Resolve<IConfigurationManager>();
```
Aeshus, line 17: Use the `[Dependency]` attribute instead of directly resolving.

```suggestion
    [Dependency] private readonly IConfigurationManager _cfg = default!;
```

Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
```diff
@@ -1,14 +1,25 @@
...
+    [Dependency] private readonly IUserInterfaceManager _uiManager = default!;
+
+    private readonly float _chatMinWidth;
+    private float? _deferredSplitFraction = null;
+
     public SeparatedChatGameScreen()
     {
         RobustXamlLoader.Load(this);
```
Aeshus, line 25: Add this line to actually resolve things with a `[Dependency]` attribute.

```suggestion
        RobustXamlLoader.Load(this);
        IoCManager.InjectDependencies(this); 
```

Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
```diff
@@ -1,14 +1,25 @@
...
 namespace Content.Client.UserInterface.Screens;
 
 [GenerateTypedNameReferences]
 public sealed partial class SeparatedChatGameScreen : InGameScreen
 {
+    private readonly IConfigurationManager _cfg = IoCManager.Resolve<IConfigurationManager>();
+    [Dependency] private readonly IUserInterfaceManager _uiManager = default!;
+
+    private readonly float _chatMinWidth;
+    private float? _deferredSplitFraction = null;
```
Aeshus, line 21: Nullable values default to null, so you don't need to explicitly set it.

```suggestion
    private float? _deferredSplitFraction;
```

Aeshus (changes requested):
Thanks for the PR!

When I first tested this version, it forced itself to 45% of my screen and wouldn't let me move it.

image

After messing around in settings for a while, clicking random buttons, it seems to have fixed itself for full-screen. I could now resize it fine and it wasn't 45% of my screen.

image

Though, after the full-screen got fixed, my floating version broke as the chat now overlapped the actions and the chat was going off of the screen.

image

I suspect there's an issue with the viewport scaling & how it interacts with the settings, as when I toggle between the settings, it occasionally breaks and then fixes itself for different resolutions.

Content.Client/UserInterface/Systems/MenuBar/Widgets/GameTopMenuBar.xaml
SpaceManiac: I think it's more obvious where the numbers come from in this case if they're left equal to the X MinSize.

Aeshus: Yeah that's fine.

SpaceManiac:
> when I toggle between the settings, it occasionally breaks and then fixes itself for different resolutions.

I spent some more time analyzing how the settings interact and how have a scale calculation that should work more reliably.

The only caveat is that it ignores "Vertical Viewport Fitting" (always acts as if it's on), but I don't think there's a good way around this, because doing otherwise would introduce a circular dependency.

SpaceManiac:
Conflict resolution pending #33576 settling

SpaceManiac:
Revert of #33047 has deconflicted this PR

Errant-4:
When I first tried it, it worked as expected. But when I switched to the Default layout and then back, it went to a squashed size and I was no longer able to resize it.
After restarting the game, it went but to normal and I was unable to reproduce the error. Sounds like the same/similar issue to what Aeshus was describing

SpaceManiac:
Unfortunately I'm not able to reproduce the problem

Aeshus (commented):
Are you still interested in working on this?

I can walk through causing the bug this week if you're interested.

SpaceManiac:
Yes, if the overall direction of the change still works for you, I'm willing to figure out this detail.

Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
```diff
@@ -1,17 +1,36 @@
...
 {
+    [Dependency] private readonly IConfigurationManager _cfg = default!;
+    [Dependency] private readonly IUserInterfaceManager _uiManager = default!;
+
+    // The minimum width of InventoryGui + HotbarGui combined.
+    // Used to keep the PDA from overlapping the shoes slot.
+    private const int INVENTORY_MINIMUM_WIDTH =
+        (125 /* width */ - 2 /* negative margin */) * 2 /* ItemStatusPanel */
+        + 68 /* width */ * 10 /* ItemSlotButton, including inventory */
+        + 67 /* width */ * 2 /* hands */;
```
metalgearsloth, line 25: These values are bound to fall out of line so you can't just rely on magic numbers like this.

Issue #20561, opened 2023-09-28
Webhook to relay watchlist connections


Chief-Engineer:
When someone with an active watch connects, relay:
- the server name,
- player name,
- and watchlist message.

Bonus points for:
- Including the time that the watchlist expires in discord's relative time format
- Delaying reporting by up to a configurable amount of time and grouping all the reports for that time period to reduce the chance of hitting a discord rate limit

PR #33483, opened 2024-11-23
Add Discord webhook on watchlist connection

Palladinium:
## About the PR
This adds a Discord webhook, configurable via CVars, that sends a notification whenever a player with a watchlist note connects to the server. Closes https://github.com/space-wizards/space-station-14/issues/20561.

## Why / Balance
See linked issue.

## Technical details
I added a new `WatchlistWebhookSystem` which listens for `NetManager.Connected` events, and checks if the player who just connected has any watchlists, and if they do, relays a message to Discord through a webhook.

The webhook system automatically batches all messages that would be sent within a certain time period into a single message, to alleviate spam and/or rate limits from a player repeatedly connecting and disconnecting. This is by default 5 seconds, and can be changed with the `discord.watchlist_connection_buffer_time` CVar.

## Media
20241123_20h46m42s_grim

## Requirements

## Breaking changes
None

**Changelog**
:cl:
ADMIN:
- add: Added Discord relay notifications when a watchlisted player connects.


Content.Server/Administration/WatchlistWebhookSystem.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Linq;
+using Timer = Robust.Shared.Timing.Timer;
+
+namespace Content.Server.Administration;
+
+/// <summary>
+///     This system sends a Discord webhook notification whenever a player with an active
+///     watchlist joins the server.
+/// </summary>
+public sealed class WatchlistWebhookSystem : EntitySystem
```
PJB3005, line 18: This is an out-of-simulation system, it should not use an entity system and should instead use a regular IoC manager.

Content.Server/Administration/WatchlistWebhookSystem.cs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private async void OnConnected(object? sender, NetChannelArgs e)
+    {
+        var watchlists = await _adminNotes.GetActiveWatchlists(e.Channel.UserId);
+
+        if (watchlists.Count == 0)
+            return;
+
+        var session = _playerManager.GetSessionByChannel(e.Channel);
```
PJB3005, line 52: Since you need the `IPlayerSession` here, it's better if you listen for `IPlayerManager.PlayerStatusChanged` instead.

Content.Server/Administration/WatchlistWebhookSystem.cs
```diff
@@ -0,0 +1,142 @@
...
+        watchlistConnections.Add(new WatchlistConnection(session.Name, watchlists));
+
+        var bufferTime = _cfg.GetCVar(CCVars.DiscordWatchlistConnectionBufferTime);
+
+        if (bufferTime > 0f)
+        {
+            if (!buffering)
+            {
+                Timer.Spawn((int)(bufferTime * 1000f), OnBufferTimeElapsed);
```
PJB3005, line 61: You cannot use a timer here as those use simulation timing. This means, for example, that a player connecting to a paused server would not have a watchlist alert raised immediately.

Just have a manual update loop & timer check called from `EntryPoint`.

Resources/Locale/en-US/discord/watchlist-connections.ftl
```diff
@@ -0,0 +1,17 @@
...
+
+discord-watchlist-connection-no-expiry = (permanent)
+discord-watchlist-connection-expiry = (expires <t:{$expiry}:R>)
+
+discord-watchlist-connection-message = with message "{$message}"
+
+discord-watchlist-connection-more = (and { $watchlists ->
+        [one] {$watchlists} other watchlist
+        *[other] {$watchlists} other watchlists
+    })
```
PJB3005, line 17: Concatenating messages like this is not the correct way to use Fluent, this is impossible to localize to most languages. You should try to keep these messages kept-together as much as possible.

PJB3005 (changes requested):

Palladinium:
I'm still not entirely sure I get Fluent, but I _think_ this is better now.

Resources/Locale/en-US/discord/watchlist-connections.ftl
```diff
@@ -0,0 +1,16 @@
...
+discord-watchlist-connection-entry = - {$playerName} with message "{$message}"
+discord-watchlist-connection-entry-expires = - {$playerName} with message "{$message}" (expires <t:{$expiry}:R>)
+discord-watchlist-connection-entry-more = - {$playerName} with message "{$message}" and { $otherWatchlists ->
+        [one] {$otherWatchlists} other watchlist
+        *[other] {$otherWatchlists} other watchlists
+    }
+discord-watchlist-connection-entry-expires-more = - {$playerName} with message "{$message}" (expires <t:{$expiry}:R>) and { $otherWatchlists ->
+        [one] {$otherWatchlists} other watchlist
+        *[other] {$otherWatchlists} other watchlists
+    }
```
PJB3005, line 16: You could still make this all one message and just do `[zero]` matching for the expiry and watchlist count, probably.

Content.Server/Administration/Managers/WatchlistWebhookManager.cs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private async void SendDiscordMessage()
+    {
+        try
+        {
+            if (_webhookIdentifier == null)
+                return;
+
+            var message = Loc.GetString("discord-watchlist-connection-header",
```
PJB3005, line 98: Use a `StringBuilder` for this.

Content.Server/Administration/Managers/WatchlistWebhookManager.cs
```diff
@@ -0,0 +1,166 @@
...
+    [Dependency] private readonly IConfigurationManager _cfg = default!;
+    [Dependency] private readonly DiscordWebhook _discord = default!;
+    [Dependency] private readonly IPlayerManager _playerManager = default!;
+
+    private ISawmill _sawmill = default!;
+
+    private WebhookIdentifier? _webhookIdentifier;
+    private List<WatchlistConnection> watchlistConnections = new();
+    private TimeSpan _bufferTime;
+    private RStopwatch? _bufferingStopwatch;
```
PJB3005, line 33: Compare with `IGameTiming.RealTime` instead of using your own stopwatch.

Content.Server/Administration/Managers/WatchlistWebhookManager.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        var bufferTimeSeconds = _cfg.GetCVar(CCVars.DiscordWatchlistConnectionBufferTime);
+        _bufferTime = TimeSpan.FromSeconds(bufferTimeSeconds);
+
+        var webhook = _cfg.GetCVar(CCVars.DiscordWatchlistConnectionWebhook);
+        if (!string.IsNullOrWhiteSpace(webhook))
+        {
+            _discord.GetWebhook(webhook, data => _webhookIdentifier = data.ToIdentifier());
+            _playerManager.PlayerStatusChanged += OnPlayerStatusChanged;
+        }
```
PJB3005, line 47: Didn't you have this using `Subs.CVar()` before? Any reason you stopped doing that?

Palladinium: I couldn't figure out the right way to get `Subs` outside of an `EntitySystem` - the only place I could find in the codebase where `new Submissions()` is called is inside the base `EntitySystem` class, and I wasn't sure how to use it elsewhere, so I opted for less extra complexity. I presume the difference is that `Subs` will re-evaluate the passed closure if the CVars are changed after startup? It probably doesn't matter too much here since the webhook CVars are `SERVER_ONLY`.

Resources/Locale/en-US/discord/watchlist-connections.ftl
Palladinium: `[zero]` wouldn't match `0` for some reason, maybe because English doesn't have a "zero" CLDR plural category. `[0]` works though.

Content.Server/Administration/Managers/WatchlistWebhookManager.cs
PJB3005: You need to use `IConfigurationManager.OnValueChanged`.

Content.Server/Entry/EntryPoint.cs
```diff
@@ -160,6 +163,7 @@ public override void Update(ModUpdateLevel level, FrameEventArgs frameEventArgs)
                 {
                     _euiManager.SendUpdates();
                     _voteManager.Update();
+                    _watchlistWebhookManager.Update();
```
PJB3005, line 166: This needs to be in `FramePostEngine`, not `PostEngine`.

PJB3005 (changes requested):
One minor nit then this PR is good.

Content.Server/Entry/EntryPoint.cs
Palladinium: I'm not sure I understand this. The doc comment for `FramePostEngine` says
>  This update is called after the main state manager on render frames, thus only applies to the client.
`PostEngine` is the same, without the "only applies to the client" part.
Since this is a server-only system, wouldn't that be the opposite of what we want?

Palladinium: I'll still push the change through, I'm mostly curious to understand how this works

PJB3005: The doc comment is incorrect, it still happens on the server, but outside simulation timing (which we want).

PJB3005 (approved):

VasilisThePikachu (approved):

PR #33496, opened 2024-11-23
Names camera routers

IamVelcroboy:
## About the PR
Small QoL to be able to tell which router is which.

## Why / Balance
They look identical and unless the mapper names them, there is nothing to communicate to the player which is which. 

## Technical details
n/a

## Media
image
image

## Requirements

## Breaking changes
n/a

**Changelog**
n/a

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Machines/surveillance_camera_routers.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/surveillance_camera_routers.yml
@@ -48,6 +48,7 @@
 - type: entity
   parent: SurveillanceCameraRouterBase
   id: SurveillanceCameraRouterEngineering
+  name: engineering camera router
   suffix: Engineering
   components:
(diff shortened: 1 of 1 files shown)


slarticodefast (approved):
good stuff

IamVelcroboy:
Test fail is fake, yeah? Def didn't touch nukie code 🌝

UbaserB:
something doesn't feel right about this, telecomms aren't named either but camera routers are also.. uhhh.. yeah.... it should probably be consistent in one way or another but IDK

IamVelcroboy:
> something doesn't feel right about this, telecomms aren't named either but camera routers are also.. uhhh.. yeah.... it should probably be consistent in one way or another but IDK

Telecomms are tricky because you can add/remove keys from them. So, the naming might not be consistent. Plus you can inspect them to see what boards are loaded. You can't do that with the routers. It's just a guess.

PR #33499, opened 2024-11-23
Make emag events check for handled

### LINKED ISSUE #6783: Emag events should check for handled
Forgot to submit my review so only recycler checks for it atm.

Also stuff like this:
image

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Resolves #6783.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Silicons/Laws/SiliconLawSystem.cs
+++ b/Content.Server/Silicons/Laws/SiliconLawSystem.cs
@@ -154,6 +154,8 @@ public sealed class SiliconLawSystem : SharedSiliconLawSystem
 
     private void OnEmagLawsAdded(EntityUid uid, SiliconLawProviderComponent component, ref GotEmaggedEvent args)
     {
+        if (args.Handled)
+            return;
 
         if (component.Lawset == null)
             component.Lawset = GetLawset(component.Laws);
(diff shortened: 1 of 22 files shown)


Content.Server/Silicons/Laws/SiliconLawSystem.cs
```diff
@@ -154,6 +154,8 @@ private void OnIonStormLaws(EntityUid uid, SiliconLawProviderComponent component
 
     private void OnEmagLawsAdded(EntityUid uid, SiliconLawProviderComponent component, ref GotEmaggedEvent args)
     {
+        if (args.Handled)
+            return;
```
slarticodefast, line 158: Since the other subscription below marks it as handled, wouldn't it mean that this one is skipped sometimes if it happens afterwards? Both are important, but I'm not sure why we need two different ones in the first place. They can likely be combined.

slarticodefast (changes requested):
Even though the issue says otherwise, I'm not sure if adding `if (args.Handled) return;` to emag subscriptions is something that should be done. In cases where we have a composed device with multiple emag interactions, this would cause only one of them to be triggered randomly depending on the order the subscriptions are handled. In my opinion since you are emagging the device itself and not some specific parts of it, all effects should activate with only using a single charge. Having to use it multiple times feels unintuitive.

Content.Server/Silicons/Laws/SiliconLawSystem.cs
Winkarst-cpu: Because one on below is an override of SharedSiliconLawSystem's method.

slarticodefast:
I asked sloth, and the original reason for the issue seems to be fixed by now. Adding the handled check would require airlocks to be emagged twice, once for the access reader and once for bolting them open. Similar problems would be indroduced for other devices. Therefore I will be closing this PR as not needed.
grafik

PR #32104, opened 2024-09-12
Rework the HoS's Energy Shotgun (Varying energy consumption depending on fire-mode + re-adds a toned down self recharge.)

BramvanZijp:
## About the PR
This PR changes the HoS's Energy shotgun in 2 main ways:
1. The capacity has been set to 1200, and the consumption per shot has been set to 120 for disabler (10 shots), 150 for wide spread (8 shots) and 200 for narrow spread (6 shots)

2. The self recharge feature has been readded, it takes 50 seconds to fully recharge, BUT the weapon does not self recharge if it has been used within the last 10 seconds, meaning you need to fully stop using it for 60 seconds to recharge it. 

## Why / Balance
Alright, here we go, time for me to justify readding the self recharge.
The energy shotgun in an ideal world is a weapon that while not being too overpowered, like what it was when it had the original self recharge, is still a substantial item which is worth to carry for the HoS and gives them a unique ability.
While currently the item is not in that OP state anymore, it is now so weak that most HoS's leave it in favor of their WT550, either stuffing it in their locker or handing it down to the warden or another security officer, which defeats the purpose of it being the HoS's high value steal objective item.

This PR aims to change that by re-adding the self recharge without being as OP as it was.
The fact that you need to stop using the weapon for it to be able to recharge means that it still has substantial downtime, so while it has infinite ammo on paper, in practice you will need to either take cover for a whole minute or switch to another weapon every time your charges run dry if you are in a sustained firefight such as in a nukie round.

Furthermore, narrow spread has always been better then wide as your standard, but with it getting 2 extra shots, it should be more worth using now in close range, and disabler may also see more use with its fire mode now having 10 charges.

These changes were discussed in the #design-discussion channel in the SS14 discord, and the general consensus was that it is worth trying this out to see how it is, and then from there we can better decide if it needs to be tweaked or reverted, since the in-game implications of these changes can only be estimated so much without seeing it in action in real rounds.

## Technical details
I'll start off by saying if the code needs to be cleaned up before this is merged, I get that.
The BatterySelfRechargerComponent now has 3 extra fields, all of which have defaults to avoid messing up other items that use it.

autoRechargePause is a bool which defaults to false, and as such by default this does not change how self recharge works unless set to true.
autoRechargePauseTime, a float defaulted to 0, decides how long the auto recharge should be on cooldown for after a charge is used from the battery.
[...]

PR #33500, opened 2024-11-23
BatteryWeaponPowerCell tweaks

Kirus59:
## About the PR

`BaseWeaponPowerCell` used information about what to shoot from the battery, which made it impossible to use different fire modes (from `BatteryWeaponFireModeComponent`) on this weapon (however, it was possible on the battery itself 😃)

And now, if the entity hasn't `BatteryComponent` then the energy will be used from the battery (Of course, if it has `PowerCellSlotComponent`).
And also, what the weapon shoots (energy consumption for this and other `BatteryAmmoProvider` parameters) will not depend on the battery.

## Requirements

## Breaking changes

no cl, no fun


Kirus59:
There is an issue with the `AmmoCounter` on the client, it does not update when the user insert/remove the battery into/from the weapon.
To solve this I need `UpdateClientAmmoEvent` from https://github.com/space-wizards/space-station-14/pull/32104

Can I add this event before merging that PR?

TheShuEd:
Blessed PR
Now we can have move mariant of shuttle laser cannons

BramvanZijp:
@Kirus59 https://github.com/space-wizards/space-station-14/pull/32104 has been merged.

Content.Server/Weapons/Ranged/Systems/GunSystem.Battery.cs
```diff
@@ -36,12 +40,28 @@ private void OnBatteryChargeChange(EntityUid uid, BatteryAmmoProviderComponent c
         UpdateShots(uid, component, args.Charge, args.MaxCharge);
     }
 
+    private void OnPowerCellChanged(EntityUid uid, BatteryAmmoProviderComponent component, ref PowerCellChangedEvent args)
```
MilonPL, line 43: `Entity<BatteryAmmoProviderComponent> ent, ref PowerCellChangedEvent args`

Content.Server/Weapons/Ranged/Systems/GunSystem.Battery.cs
```diff
@@ -104,7 +130,9 @@ private void OnBatteryDamageExamine(EntityUid uid, BatteryAmmoProviderComponent
 
     protected override void TakeCharge(EntityUid uid, BatteryAmmoProviderComponent component)
```
MilonPL, line 131: `Entity<BatteryAmmoProviderComponent>` pattern

Content.Server/Weapons/Ranged/Systems/GunSystem.cs
```diff
@@ -35,6 +36,7 @@ public sealed partial class GunSystem : SharedGunSystem
     [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly StaminaSystem _stamina = default!;
     [Dependency] private readonly SharedContainerSystem _container = default!;
+    [Dependency] private readonly PowerCellSystem _powerCellSystem = default!;
```
MilonPL, line 39: Just `_powerCell`

Content.Server/Weapons/Ranged/Systems/GunSystem.Battery.cs
Kirus59: This pattern does not support the use of inherited from `BatteryAmmoProviderComponent` classes in `SubscribeLocalEvent`

image

MilonPL: Oh, it's used for both the Projectile and Hitscan components, I somehow missed that in the first review. In this case, you would want to use a slightly different pattern:
```cs
private void OnBatteryStartup<T>(Entity<T> ent, ref ComponentStartup args) 
    where T : BatteryAmmoProviderComponent
{
    UpdateShots((ent, ent.Comp));
}
```
and so on for every other method, just switching the event

You should be able to then update both of the `UpdateShots` methods to take `Entity<BatteryAmmoProviderComponent>` rather than Uid and Comp. Let me know if you need any help with it.

Content.Server/Weapons/Ranged/Systems/GunSystem.Battery.cs
```diff
@@ -36,12 +40,33 @@ private void OnBatteryChargeChange(EntityUid uid, BatteryAmmoProviderComponent c
...
+        else if (TryComp<PowerCellSlotComponent>(uid, out var powerCellSlot) &&
+            _powerCell.TryGetBatteryFromSlot(uid, out var powerCell, powerCellSlot))
+        {
+            currentCharge = powerCell.CurrentCharge;
+            maxCharge = powerCell.MaxCharge;
+        }
 
-        UpdateShots(uid, component, battery.CurrentCharge, battery.MaxCharge);
+        UpdateShots(uid, component, currentCharge, maxCharge);
     }
```
metalgearsloth, line 70: Isn't the whole point of chargechanged to handle this? As is this is turning into a bit of a mess trying to handle 2 codepaths and if we start having to do this everywhere it will soon become unmaintainable.

metalgearsloth (dismissed):

Content.Server/Weapons/Ranged/Systems/GunSystem.Battery.cs
Kirus59: done

KaiserMaus:
This PR move projectile/Hitscan from battery provider to weapon components?

BramvanZijp:
Also, it will mean removing the damage indicators from power cells, which felt odd having there since thats only relevant for the svallin laser gun and machine gun.

Kirus59:
> Also, it will mean removing the damage indicators from power cells, which felt odd having there since thats only relevant for the svallin laser gun and machine gun.

Why should there be a damage indicator for a specific type of weapon on the power cell at all?
In the future, there may be more such "svallins" and each of them may have its own damage, energy consumption, etc.
And I think that standard power cells shouldn't have parameters only for a specific type of weapon, but should serve as a source of energy for many different types.

PR #33504, opened 2024-11-24
Grant elevated door access to Janitor

dylanstrategie:
## About the PR

This would significantly increase the access level of the Janitor role on doors only by enabling Janitor access to a large selection of doors for various departments, allowing him into most departments but keeping him out of Command and Security-level areas and some specifically very dangerous areas like Atmospherics or Chemistry.

## Why / Balance

Since this is a pretty significant change in terms of role balance, I will split this as to why I did this code wise, and why I think this is viable regardless gameplay-wise:

**Code-wise**, it's obvious this is a bit of a brute-force method to allow the Janitor the ability to enter and clean a lot of departments, and gives him elevated access even to areas that are a bit risky like Science's anomaly containment chamber or the Engineering engine and power substations.

However, there is no way to further refine what rooms he has access to without a **major access and mapping refactor**. For example Engineering is a single universal access category, which doesn't hugely make sense where you're thinking of allowing other roles to enter more benign areas like Engineering lobby, to clean, rescue an injured person or arrest a perp.

Some other areas, Security notably, have far more detailed access levels which means the Janitor would have a tightly controlled access level, using what was already set up for the Lawyer so he can't open cells, go into Security equipment rooms or other risky areas.

**Gameplay-wise**, I think this makes sense insofar as what a Janitor is supposed to be. Wherever it is in real life or in tropes, a Janitor by definition has access to a large amount of areas, many containing sensitive equipment or information, because he's the person who is cleaning them. This is currently completely impossible and the Janitor has the same access level as a Passenger minus his own janitorial office.

I believe this hugely detracts from how the job works because having to beg departmental workers to be let into areas to clean them, especially since the station ends up junked up at roundstart, is frankly hugely frustrating, especially otherwise benign areas like Medbay or Cargo. It also puts an unreasonable burden on departments to clean their own departments when we have a dedicated job for cleaning.

While this elevated access could obviously be used for ill-intent, I also believe this isn't crazy roleplay and gameplay wise. Janitors clearly stick out, and departments tend to be territorial and would likely keep an eye on them. This would also make this job special compared to literally any civilian who scavenged a mop and bucket from maintenance by allowing them to be -very- independent and only requiring departmental supervision to access dangerous areas (as current access allows, sadly Engineering and Research only have a single, universal access level)

I will also note on a miscellaneous note that this kind of elevated access makes sense for multiple other roles, notably Paramedics (to rescue injured crew), Security (for arrests and investigations) and Engineering (for repairs). However to keep the PR atomic I am only addressing Janitors here, which I think are the highest priority. Paramedics also don't have their own access type.

## Technical details

Janitor access can now interact with a much wider array of airlocks, here is a cleaned-up list of affected airlocks. **This access is for airlocks and windoors only, not machinery, buttons, lockers or other access checks**:

Service/Civilian: Bar, Kitchen, Hydroponics, Theatre, Lawyer, Chapel. All mixed access between those.
Cargo: Cargo, Salvage
Medbay: Medical, Morgue
Research: Research
Engineering: Engineering
Security: Brig, Lawyer/Brig

He is excluded from all Command airlocks, external, Chemistry (to avoid mopping up water-reactive materials), Atmospherics (Atmospherics gear like pumps are not ID-locked, so too dangerous) and Security (to not give access to cells or equipment)

## Requirements

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/Electronics/door_access.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/Electronics/door_access.yml
@@ -30,7 +30,7 @@
   suffix: Bar, Locked
   components:
   - type: AccessReader
-    access: [["Bar"]]
+    access: [["Bar"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -38,7 +38,7 @@
   suffix: Bar, Locked
   components:
   - type: AccessReader
-    access: [["Bar"], ["Kitchen"]]
+    access: [["Bar"], ["Kitchen"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -46,7 +46,7 @@
   suffix: Hydroponics, Locked
   components:
   - type: AccessReader
-    access: [["Hydroponics"]]
+    access: [["Hydroponics"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -54,7 +54,7 @@
   suffix: Chapel, Locked
   components:
   - type: AccessReader
-    access: [["Chapel"]]
+    access: [["Chapel"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -62,7 +62,7 @@
   suffix: Theatre, Locked
   components:
   - type: AccessReader
-    access: [["Theatre"]]
+    access: [["Theatre"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -70,7 +70,7 @@
   suffix: Kitchen, Locked
   components:
   - type: AccessReader
-    access: [["Kitchen"]]
+    access: [["Kitchen"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -78,7 +78,7 @@
   suffix: Kitchen/Hydroponics, Locked
   components:
   - type: AccessReader
-    access: [["Kitchen"], ["Hydroponics"]]
+    access: [["Kitchen"], ["Hydroponics"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -94,7 +94,7 @@
   suffix: Lawyer, Locked
   components:
   - type: AccessReader
-    access: [["Lawyer"]]
+    access: [["Lawyer"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -119,7 +119,7 @@
   suffix: Salvage, Locked
   components:
   - type: AccessReader
-    access: [["Salvage"]]
+    access: [["Salvage"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -127,7 +127,7 @@
   suffix: Cargo, Locked
   components:
   - type: AccessReader
-    access: [["Cargo"]]
+    access: [["Cargo"], ["Janitor"]]
 
 # Engineering
 - type: entity
@@ -152,7 +152,7 @@
   suffix: Engineering, Locked
   components:
   - type: AccessReader
-    access: [["Engineering"]]
+    access: [["Engineering"], ["Janitor"]]
 
 # Science
 - type: entity
@@ -161,7 +161,7 @@
   suffix: Morgue, Locked
   components:
   - type: AccessReader
-    access: [["Medical"], ["Detective"], ["Chapel"]]
+    access: [["Medical"], ["Detective"], ["Chapel"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
@@ -177,7 +177,7 @@
   suffix: Medical/Science, Locked
   components:
   - type: AccessReader
-    access: [["Research"], ["Medical"]]
+    access: [["Research"], ["Medical"], ["Janitor"]]
 
 - type: entity
   parent: DoorElectronics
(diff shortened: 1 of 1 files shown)


SpaceRox1244:
Extra access makes sense for janitor as a maintenance role because theyre equipped for cleaning the places they go. It is a very humble role with a humble set of equipment. On the contrary, engineers and security are not humble roles with humble equipment. Their entrance needs are very topical, meanwhile janitorial upkeep is very passive. It is the difference between arguing for access for an entire team of capable people and arguing for access for a single or so independent-agent(s).

SpaceRox1244:
This change will be very positive for player culture. Currently, there is a taboo about being present in a department that you are not wearing the colors of. People can be very high strung about even letting you into the workspace to do something they do not deem absolutely necessary. There is a fixation with only doing what is absolutely necessary, and in many cases this relates to workplace upkeep. Lights will go unreplaced because they dont ruin visibility, trash will pile on tables because it has no mechanical consequence, people will not properly secure job equipment in lockers because they already deem the department a totally locked-down restricted area. A simple service job having access to these workspaces makes the whole station feel more communal and less territorial. I think this ID access change, or any change that granted janitor more movement freedom, would be very good for player trust and cooperation.

SpaceRox1244:
In my opinion though, I think some of the finer accesses included in the PR could be pruned, particularly the ones that relate to other "independent-agent" roles, like chefs and bartenders. These accesses would relate to their small personal workspace in which they are closely present and working uninterrupted.

Additionally, I do not personally think that any amount of security access is warranted, but I guess it's worth seeing how others weigh in on that.

beck-thompson:
You really should add a change log for this! Big change

dylanstrategie:
> You really should add a change log for this! Big change

Sorry, I forgot the changelog starts in comment tag, it was written up

deltanedas:
access levels wyci
command access 1 go into the meeting room but no announcements
sec access 1 for old Brig, sec access 2 for old Security, sec access 3 for old Armory
med 1 for lobby/treatment room but not equipment
engi 1 for lobby thats it

Djungelskog2:
Why not just give them the jani keyring from 13? Solves the issue without making them massive antag targets and still makes the access require some amount of permission and cooperation from other players.

Djungelskog2:
> Why not just give them the jani keyring from 13? Solves the issue without making them massive antag targets and still makes the access require some amount of permission and cooperation from other players.

I should elaborate, it's an item that heads can use to give partial AA for their department specifically. Whenever you use it, it has a ~5-15 second doafter that then opens the airlock of that department you used it on, the access only lasts about 5-15 mins after you are given access by that head.

You only need to ask one person for access once without needing to take any trips to the hop. It's nerfed/with enough precautions that it doesn't make janitors a particular target for antags or make things too easy for antag janitors either.

Pumkin69:
yeah making the new player role a massive antag target isn't really amazing but i see the vision

deltanedas:
give jani 50 hours requirement then to keep noobs safe from diet AA stealers :trollface:

Snorkcom:
It is better to add a notification to the service channel on the cleaning light switch (with a delay for the next notification).
Also display the sender and department in the chat

SpaceRox1244:
Id argue janitor's new only-door access isnt special enough to make janitors a massive murder-and-rob target for an antag, especially if my suggested extraneous accesses were pruned. Antags come already loaded with a bunch of options for getting into a place that doesnt involve robbing someone to take their access item. Even that strategy can require some TC spending, if you need a weapon or an agent ID.

This concern is valid but it is based on a current game where it is unprecedented that someone non-departmental could have intended access to your space. One of the expected benefits of this PR is that people will not be so antsy about other peoples presence alone and will understand that they need to secure their workplace better.

SpaceRox1244:
> > Why not just give them the jani keyring from 13? Solves the issue without making them massive antag targets and still makes the access require some amount of permission and cooperation from other players.
> 
> I should elaborate, it's an item that heads can use to give partial AA for their department specifically. Whenever you use it, it has a ~5-15 second doafter that then opens the airlock of that department you used it on, the access only lasts about 5-15 mins after you are given access by that head.
> 
> You only need to ask one person for access once without needing to take any trips to the hop. It's nerfed/with enough precautions that it doesn't make janitors a particular target for antags or make things too easy for antag janitors either.

This suggestion is a good solution to the excessive babysitting that a janitor needs, because repeatedly interacting with someone just to use their door access is not really a valuable roleplay interaction. I will say though that it still leaves the decision with the department, which isnt going to address what I hope this change or some change like it would, which is departments being territorial. Ive had paranoid heads refuse entrance with a clear mess in sight because it wasnt *absolutely gameplay necessary*. Id just generally like people to have less direct authority over whether or not someone else gets to play their part of the game.

SpaceRox1244:
If youd ask me for some other idea for a janitor access item, id suggest a keyring as well, but different. To me, itd start a do-after to open permitted doors and do something like play a sound for beginning and ending the interaction. That way it slows down how quickly a janitor can move through a department and announces to the nearby crew that they are moving through it, because the concerns about antags having the ID kind of access is fair, with how unannounced and immediate it is, but the freedom would still be good.

slarticodefast:
Hey, I brought this up for a quick maintainer discussion and we decided not to merge this.
- It makes the janitor a kill target for the soft AA
- I creates a lot of potential for raiders to simply walk into any department, especially engineering
- You can just talk to people or the AI to let you in

If you want to make it more convenient for the janitor, porting SS13's janitor keyring would be a better approach in my opinion since it still requires approval from a department head and is only temporary.
grafik

PR #33517, opened 2024-11-24
Buffing slugs and replacing beanbags from the Bulldog bundle

Thatonestomf:
## About the PR

Buffed slugs significantly by upping their pierce damage to 40 and adding 15 structural on top of that, making it a much closer sidegrade to regular shells as well as making it more suitable for breaking fortifications. Also replaced the beanbag shells from the bulldog bundle with slugs and added slugs to the ammo bundle.
## Why / Balance

Slugs were horrible before, with only 28 pierce to their name, no one in their right mind would ever pick it over shells or another weapon. this change makes them far more suitable for combat, trading in raw damage and ease of use, for damage at range.
It also made little sense why a 20 tc firearm would bother having a less than lethal ammo type, so i got rid of them. These slugs should fill that gap nicely. 
## Technical details

Just some YML stuff, nothing much
## Media

image
image
image
image

## Requirements

- [x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

None
**Changelog**

:cl:
- add: Added slugs to both the bulldog and ammo bundle.
- tweak: Slugs are alot more powerful than they were before, now dealing 40 pierce and 15 structural
- tweak: Basic shells now deal structural damage


Boaz1111:
This PR should probably be atomized.

Thatonestomf:
6424c238c2fb4
On god?

ruddygreat:
the slug damage change should definitely be moved into another PR
it's also _probably_ an overcorrection, it makes slugs basically equal to normal shells at ranges < 4 tiles & outright better than normal shells at ranges > 4 tiles, assuming you actually hit the shot

deltanedas:
> outright better than normal shells at ranges > 4 tiles

it should, otherwise you have no reason to use a slug if buckshot does the same damage at range

ruddygreat:
> it should, otherwise you have no reason to use a slug if buckshot does the same damage at range

oh yeah no this is fine, my (admittedly poorly worded) point is more that it's performance is identical at close ranges where you'd "expect" the buckshot to be better

Thatonestomf:
If all pellets land, Buckshot deals 60 pierce and you can hit all 6 pellets for about 2.5 tiles away. At mid range they deal about the same damage as slugs, but since it's all pierce deal more bleed. At longer ranges you can still pepper targets easily.

Slugs however are all or nothing, you either deal massive damage or none. At longer distances you can deal damage that the spread on Buckshot could never allow. And up close you still miss out on damage Buckshot could reliably deal anyways. 

Before the trade off in damage for accuracy wasn't even close to worth it. But now it can really compete.

Thatonestomf:
image
this look good?

KaiserMaus:
2 shells in second
100 hp per second on no armor human...
maybe add some Structural type damage  and lower damage to ~40?

Thatonestomf:
normal shells deal 60 peirce if all pellets hit, for a dps of 120
removing the 2 hit would make it alot weaker

Plykiya:
my hero...

Pumkin69:
i fucking love shotguns soooo much i cant wait to roll entire crews with a bulldog as commander good lord somebody stop me

SlamBamActionman:
I think this change is far too powerful of a buff for the slugs. It's 28 -> 50, and while I agree that slugs are at the moment a fairly undesirable option, this would make them correct in *any* scenario.

The tradeoff for shotguns is that you get insane burst damage at short range. Going purely by damage, shotguns consistently rank for highest DPS in the game due to this. But this is only permissible due to the range requirement.

Slugs turn shotguns into a long-range weapon with no accuracy penalty. While the shotgun *does* still suffer from low mag size and wielding requirements, it no longer has the range issue. A 50 damage slug shotgun has 5 higher DPS over the Lecter and near twice as much as the Inspector. Considering how multi-use this would make the shotgun, I think 50 is overdoing the buff. I'd land somewhere at around 37.

37 would bring it to a moderately higher DPS than the Inspector (74 vs 52.5) which should offset the bulkiness of the weapon while still allowing the Bulldog to dish out damage. You won't have the same insane damage as a C-20r, but instead hover around a WT-550 in effectiveness against a Sec Hardsuit (bulldog w/ 37 slug: 2.00s time to crit; wt550: 1.92s; c-20r: 1.32s).

The structural damage feels unnecessary and kind of counterintuitive for how a slug would work. If anything normal shotgun shells should deal minor structural since the damage being spread out over multiple shells means most objects with resistances (like crates) ignore shotgun blasts. Since slugs deal all damage in one projectile they can already blast through.

Thatonestomf:
Sorry for the delay of like... 1 month i was busy doing nothing.
But ive taken your advice and am doing some tests, expect a new commit in a bit

Thatonestomf:
dang, its not committing for some reason

superjj18:
3 hits to kill someone with more than 0% pierce resistance seems fair.  A unarmored zombie for instance should not be surviving two or more slugs.

Thatonestomf:
Fixed now, if anyone has any more suggestions on balance feel free to tell me so that this can finally get merged
yes im aware the delay is mostly my fault

semensponge:
I already did a PR to replace the beanbag with a normal mag of buckshot. #34945

I do like the changes to slugs though and think you could atomize and push that through. It might even be worth considering adding an additional mag of slugs to the Bulldog bundle because it has the same overall damage as the C20 but costs more. An additional mag could justify the cost.

EmoGarbage404 (approved):

ArtisticRoomba (approved):
Discussed on disc. Good balance changes for slugs, they're pretty underutilized.

Issue #12971, opened 2022-12-11
Mech feedback


metalgearsloth:
- It has its own version of damageable when it should be using damageable (Integrity).
- The footstep sounds are irregular (and seem different per client?)
- It has its own version of batterycomponent (Energy).
- Clamps grab the entity before the do_after is finished.
- It uses component references.

metalgearsloth:
Also added to this is https://github.com/space-wizards/space-station-14/pull/14646 mech tools should just be like regular tools but emo said he'd fix it later.

metalgearsloth:
- Should be using ContainerFill.

Also see https://github.com/space-wizards/space-station-14/issues/14864

SlamBamActionman:
This issue would benefit from being split up into multiple issues, to make it easier for contributors to process them.

PR #33520, opened 2024-11-24
Make mech use BatteryComponent of the battery it uses 

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
See #12971.

## Requirements

## Breaking changes

**Changelog**


Winkarst-cpu:
@beck-thompson, it's not a bugfix, but a cleanup since MechComponent was doing its own battery thing instead of using an existing system.

Content.Server/Mech/Systems/MechSystem.cs
```diff
@@ -321,28 +327,24 @@ public override void BreakMech(EntityUid uid, MechComponent? component = null)
...
 
-    public override bool TryChangeEnergy(EntityUid uid, FixedPoint2 delta, MechComponent? component = null)
+    /// <summary>
+    /// Attempts to change the amount of energy in the mech.
+    /// </summary>
+    /// <param name="uid">The mech itself</param>
+    /// <param name="delta">The change in energy</param>
+    /// <param name="component"></param>
+    /// <returns>If the energy was successfully changed.</returns>
+    public bool TryChangeEnergy(EntityUid uid, FixedPoint2 delta, MechComponent? component = null)
```
MilonPL, line 337: If you're going to change this, I think you might as well use `Entity<T?>`

Content.Shared/Mech/SharedMech.cs
```diff
@@ -60,3 +60,16 @@ public sealed partial class MechOpenUiEvent : InstantActionEvent
...
 {
 }
+
+[Serializable, NetSerializable]
+public sealed class MechEnergyMessage : BoundUserInterfaceMessage
+{
+    public float Charge { get; }
+    public float MaxEnergy { get; }
+
+    public MechEnergyMessage(float charge, float maxEnergy)
```
MilonPL, line 70: Might want to make this a primary constructor?

Content.Server/Mech/Systems/MechSystem.cs
```diff
@@ -311,6 +314,10 @@ public override void UpdateUserInterface(EntityUid uid, MechComponent? component
             EquipmentStates = ev.States
         };
         _ui.SetUiState(uid, MechUiKey.Key, state);
+
+        TryComp<BatteryComponent>(component.BatterySlot.ContainedEntity, out var batteryComp);
+
+        _ui.ServerSendUiMessage(uid, MechUiKey.Key, new MechEnergyMessage(batteryComp?.CurrentCharge ?? 0f, batteryComp?.MaxCharge ?? 0f));
```
Centronias, line 320: This is the only time `MechEnergyMessage` is instantiated, so why not make the constructor more particular to the usage?
eg.
```csharp
new MechEnergyMessage(batteryComp)
```
and then move the null handling into its implementation?

Content.Shared/Mech/SharedMech.cs
```diff
@@ -60,3 +60,10 @@ public sealed partial class MechOpenUiEvent : InstantActionEvent
 public sealed partial class MechEjectPilotEvent : InstantActionEvent
 {
 }
+
+[Serializable, NetSerializable]
+public sealed class MechEnergyMessage(float charge, float maxEnergy) : BoundUserInterfaceMessage
+{
+    public float Charge { get; } = charge;
+    public float MaxEnergy { get; } = maxEnergy;
+}
```
Centronias, line 69: Why not
```suggestion
public sealed class MechEnergyMessage(float charge, float maxEnergy) : BoundUserInterfaceMessage
{
    public readonly float Charge = charge;
    public readonly float MaxEnergy = maxEnergy;
}
```
?
This is less of a suggestion and more of a "I don't know what's better, so I'd appreciate insight" question.

Winkarst-cpu: Will cause CS0106.

Centronias: I believe that CS0106 would be raised if it were
```csharp
readonly float Charge { get; } = charge;
```
, because that's a property with a synthetic getter _method_.
I'm suggesting a normal field that's readonly:
```csharp
readonly float Charge = charge;
```

Did you not remove the `{ get; }`, or am I misunderstanding?

Winkarst-cpu: Done.

PR #33536, opened 2024-11-25
Airlock fixes round two

metalgearsloth:
I forgot what was even in this PR but anyway:
- Moved wire panel opening to shared to fix client mispredicting opens.
- Removed clicking airlocks to hold them open as it's weird, unintuitive, and doesn't work properly.
- Fix emag mispredict
- Some general deduplication of code.

:cl:
- fix: Fix emag audio mispredicting when used on airlocks.
- remove: Removed clicking airlocks to hold them open as it didn't work properly.
- fix: Fix wire panel UI open not being predicted and opening even if you didn't have an appropriate tool.


Content.Shared/Doors/Systems/SharedDoorSystem.cs
```diff
@@ -138,10 +140,12 @@ private void OnAttemptEmag(EntityUid uid, DoorComponent door, ref OnAttemptEmagE
         }
     }
 
-    private void OnEmagged(EntityUid uid, DoorComponent door, ref GotEmaggedEvent args)
+    private void OnDoorEmagged(EntityUid uid, DoorComponent door, ref GotEmaggedEvent args)
     {
-        if (!SetState(uid, DoorState.Emagging, door))
+        if (!SetState(uid, DoorState.Opening, door))
```
slarticodefast, line 145:
```suggestion
        if (!SetState(uid, DoorState.Emagging, door))
```
Emagging the door uses other timings for the state changes than opening it normally, so you should keep this.

slarticodefast (changes requested):
Works as intended but you changed the door opening timing for the emag.

Also I found another problem while testing (which existed before this PR):
- Open VV and open an airlock's doorcomponent, set it to refresh
- open the airlock and bolt it open using a door remote
- the NextStateChange variable keeps incrementing every second in attempt to close the door, even though it is bolted

Since unbolting already resets the NextStateChange, that datafield should be set to null when being bolted to keep the door from updating.

Content.Shared/Doors/Systems/SharedAirlockSystem.cs
```diff
@@ -22,11 +24,46 @@ public override void Initialize()
...
+            Dirty(ent);
+        }
+    }
+
+    private void OnAirlockInteractUsing(Entity<AirlockComponent> ent, ref InteractUsingEvent args)
+    {
+        if (args.Handled || !_wiresSystem.CanWireAct(args.Used))
+            return;
+
+        args.Handled = true;
```
SlamBamActionman, line 54: This early Handled makes multitools no longer able to device link with doors, even those with closed wire panels. 

Move it down to inside the `panel.Open` if statement.

SlamBamActionman (changes requested):
Other than the multitool bug, code looks good. Def makes it feel nicer that I can finally close doors with E while holding a screwdriver. 

PR description should be made to follow the PR template. Since this changes the event `DoorBoltChangedEvent`, please include that as a Breaking Change.

VerinSenpai:
this happen to fix the issue where an emagged door will mispredict declining an open when someone with normally no access to a door attempts to open it?

PR #33553, opened 2024-11-25
Mind Swap Fix

ActiveMammmoth:
## About the PR

Stores use mind entityuid instead of body entityuid.
## Why / Balance

Originally a part of my Mind Swap PR, I was asked to split this off as a separate PR by Keron. This is necessary to make Mind Swap function correctly; without this the grimoire would be owned by the person who the wizard swaps with, allowing them to not only purchase spells but even refund the spells from the original wizard. Effectively Mind Swap would be just giving up your wiz role to someone else entirely, which obviously is not the intention.
## Breaking Changes
StoreComponent AccountOwner now uses the EntityUid associated with the mind instead of the body.


slarticodefast (dismissed):
You will need to adjust all the store conditions for this.
For example `BuyerAntagCondition` which gets the mind from the buyer entity, which is now the mind itself.
grafik
Or `BuyerSpeciesCondition` which now checks for `HumanoidAppearanceComponent` on the mind entity instead of the player.

Content.Server/Store/Systems/StoreSystem.cs
```diff
@@ -69,10 +70,13 @@ private void OnStoreOpenAttempt(EntityUid uid, StoreComponent component, Activat
         if (!component.OwnerOnly)
             return;
 
-        component.AccountOwner ??= args.User;
+        if (!_mind.TryGetMind(args.User, out var mind, out var mindComp))
+            return;
+
+        component.AccountOwner ??= new Entity<MindComponent?>(mind, mindComp);
```
keronshb, line 76:
```suggestion
        component.AccountOwner ??= (mind, mindComp);
```

You dont need to make a new Entity, you can just make a tuple.
Might need another maint to sanity check me on this.

keronshb (changes requested):
One potential change that I need another maint to sanity check for me. Ran tests in game with wizard and revenant, ran into no issues.

Content.Server/Store/Systems/StoreSystem.cs
metalgearsloth: The tuple gets cast to entity anyway so entity is probably better.

Content.Shared/Store/Components/StoreComponent.cs
```diff
@@ -41,7 +42,7 @@ public sealed partial class StoreComponent : Component
     /// regardless of who activated it. I.E. role specific items for uplinks.
     /// </summary>
     [ViewVariables(VVAccess.ReadWrite)]
-    public EntityUid? AccountOwner = null;
+    public Entity<MindComponent?>? AccountOwner = null;
```
metalgearsloth, line 45: This looks like it should be a datafield and in turn an entityuid, I'm not really sure why it isn't a datafield.

keronshb (approved):
lg2m

Content.Shared/Store/ListingCondition.cs
```diff
@@ -1,3 +1,4 @@
+using Content.Shared.Mind;
 using JetBrains.Annotations;
 using Robust.Shared.Serialization;
```
ScarKy0, line 3: Accidental import?

Content.Shared/Store/Components/StoreComponent.cs
```diff
@@ -1,4 +1,5 @@
 using Content.Shared.FixedPoint;
+using Content.Shared.Mind;
 using Robust.Shared.Audio;
```
ScarKy0, line 3: Accidental import?

Content.Server/Traitor/Uplink/UplinkSystem.cs
```diff
@@ -61,19 +63,23 @@ public bool AddUplink(
     /// </summary>
     private void SetUplink(EntityUid user, EntityUid uplink, FixedPoint2 balance, bool giveDiscounts)
     {
+        if (!_mind.TryGetMind(user, out var mind, out _))
+            mind = user; // Nothing without a mind should be setting up an uplink but discount integration test does so ¯\_(ツ)_/¯
+
```
ScarKy0, line 68: The integration test should be updated instead, this is just weird looking code.

Content.Server/Store/Systems/StoreSystem.Listings.cs
```diff
@@ -1,4 +1,5 @@
 using System.Diagnostics.CodeAnalysis;
+using Content.Shared.Mind;
 using Content.Shared.Store;
```
ScarKy0, line 3: Accidental import?

Content.Server/Store/Conditions/BuyerDepartmentCondition.cs
```diff
@@ -30,14 +30,12 @@ public override bool Condition(ListingConditionArgs args)
...
 
-        // this is for things like surplus crate
-        if (!minds.TryGetMind(args.Buyer, out var mindId, out _))
-            return true;
+        if (!ent.TryGetComponent<MindComponent>(args.Buyer, out var _))
+            return true; // inanimate objects don't have minds
 
         var jobs = ent.System<SharedJobSystem>();
-        jobs.MindTryGetJob(mindId, out var job);
+        jobs.MindTryGetJob(args.Buyer, out var job);
```
ScarKy0, line 39: Get rid of ent and just use TryComp<T>, way cleaner that way and I don't think theres a differance

Content.Server/Store/Conditions/BuyerAntagCondition.cs
```diff
@@ -27,13 +27,12 @@ public sealed partial class BuyerAntagCondition : ListingCondition
...
-        var minds = ent.System<SharedMindSystem>();
 
-        if (!minds.TryGetMind(args.Buyer, out var mindId, out var mind))
-            return true;
+        if (!ent.HasComponent<MindComponent>(args.Buyer))
+            return true; // inanimate objects don't have minds
 
         var roleSystem = ent.System<SharedRoleSystem>();
-        var roles = roleSystem.MindGetAllRoleInfo(mindId);
+        var roles = roleSystem.MindGetAllRoleInfo(args.Buyer);
```
ScarKy0, line 35: Get rid of ent and just do HasComp<T>, cleaner but im unsure if theres a differance here

Content.Server/Store/Conditions/BuyerJobCondition.cs
```diff
@@ -27,14 +27,12 @@ public sealed partial class BuyerJobCondition : ListingCondition
...
 
-        // this is for things like surplus crate
-        if (!minds.TryGetMind(args.Buyer, out var mindId, out _))
-            return true;
+        if (!ent.TryGetComponent<MindComponent>(args.Buyer, out var _))
+            return true; // inanimate objects don't have minds
 
         var jobs = ent.System<SharedJobSystem>();
-        jobs.MindTryGetJob(mindId, out var job);
+        jobs.MindTryGetJob(args.Buyer, out var job);
```
ScarKy0, line 35: No ent, do TryComp<T>

Content.Server/Store/Conditions/BuyerSpeciesCondition.cs
```diff
@@ -27,7 +28,10 @@ public override bool Condition(ListingConditionArgs args)
     {
         var ent = args.EntityManager;
 
-        if (!ent.TryGetComponent<HumanoidAppearanceComponent>(args.Buyer, out var appearance))
+        if (!ent.TryGetComponent<MindComponent>(args.Buyer, out var mind))
+            return true; // needed to obtain body entityuid to check for humanoid appearance
+
+        if (!ent.TryGetComponent<HumanoidAppearanceComponent>(mind.OwnedEntity, out var appearance))
```
ScarKy0, line 34: TryComp<T>

ScarKy0 (changes requested):
Not fully familiar with this code but get reviewed.

slarticodefast:
@ScarKy0 The shorthands `TryComp` and `GetComp` are only available inside entity systems.

ScarKy0:
> @ScarKy0 The shorthands TryComp and HasComp are only available inside entity systems.

Very good to know for the future, thanks

ScarKy0 (approved):
Wizard, just a week away

PR #33574, opened 2024-11-26
Make gun wielding automatic

### LINKED ISSUE #18819: Wielding should use UseDelay
Like a 1s cooldown or whatever.

### LINKED ISSUE #28220: Separate the wielding and bolting hotkeys
## Description
The biggest issue with the recent gun changes is that both wielding and bolting are bound to the same key, it would be much better to have it on a different key OR make it so you have to press one more button to wield the gun.
How it currently is:
Wield = Alt+Z
Bolt = Alt+Z

What I'm suggesting is:
Wield = Ctrl+Alt+Z
Bolt = Alt+Z

Why i suggest it being ctrl+alt+z and not another key entirely is it more thematic for it to also be bound to a Z key interaction since it's the "use while in hand key" and having a third option for uses can open up other items to have it as well, and for the case with heavy guns it makes since it takes a moment to wield a gun and move all your fingers off of the wasd keys and press the keys below them. still making it a slightly hard maneuver to pull off while in combat but doable when given a moment to take cover

### LINKED ISSUE #33537: Wielding should be automatic
## Description
Wielding items, such as guns, should automatically occur when the player has two hands free. More or less, wielding should be turned in to "the state an item goes in to when both of your hands are free."

With all of the wielding QOL changes (you are no longer blocked by wielding something in most instances,) there's basically no reason for this to not be a feature now. What I said above about wielding being a default state is basically already true, except that you have to push z to initiate it. It's a fairly unnecessary and obnoxious piece of hand management that adds nothing to the game.
In addition to this, wielding prevents other interactions that use Z from being on the same item. This means that gun bolting cannot be added to stuff like the Kammerer or bolt-actions until this is fixed.

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Resolves #33537, Resolves #28220. Resolves #18819 (not a use anymore).

## Technical details
New behavior inspired by proposal by @slarticodefast).

## Media

https://github.com/user-attachments/assets/5b2cc4bb-59a2-4577-8ada-f44d90b9abac

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Now guns are wielded and unwielded automatically.

### DIFF
--- a/Content.Shared/Hands/EntitySystems/SharedHandsSystem.cs
+++ b/Content.Shared/Hands/EntitySystems/SharedHandsSystem.cs
@@ -314,6 +314,18 @@ public abstract partial class SharedHandsSystem
         return hands.Hands.TryGetValue(handId, out hand);
     }
 
+    public int CountFreeHands(Entity<HandsComponent> hands)
+    {
+        var free = 0;
+        foreach (var hand in hands.Comp.Hands.Values)
+        {
+            if (hand.IsEmpty)
+                free++;
+        }
+
+        return free;
+    }
+
     public int CountFreeableHands(Entity<HandsComponent> hands)
     {
         var freeable = 0;
--- a/Content.Shared/Wieldable/WieldableDoAfterEvent.cs
+++ b/Content.Shared/Wieldable/WieldableDoAfterEvent.cs
@@ -1,9 +1,0 @@
-﻿using Content.Shared.DoAfter;
-using Robust.Shared.Serialization;
-
-namespace Content.Shared.Wieldable;
-
-[Serializable, NetSerializable]
-public sealed partial class WieldableDoAfterEvent : SimpleDoAfterEvent
-{
-}
--- a/Content.Shared/Wieldable/WieldableSystem.cs
+++ b/Content.Shared/Wieldable/WieldableSystem.cs
@@ -40,7 +40,6 @@ public sealed class WieldableSystem : EntitySystem
     {
         base.Initialize();
 
-        SubscribeLocalEvent<WieldableComponent, UseInHandEvent>(OnUseInHand, before: [typeof(SharedGunSystem)]);
         SubscribeLocalEvent<WieldableComponent, ItemUnwieldedEvent>(OnItemUnwielded);
         SubscribeLocalEvent<WieldableComponent, GotUnequippedHandEvent>(OnItemLeaveHand);
         SubscribeLocalEvent<WieldableComponent, VirtualItemDeletedEvent>(OnVirtualItemDeleted);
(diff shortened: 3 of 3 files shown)


thebadman4662:
Could there be a way added to unwield it by pressing X and swapping to wieldhand? At least to avoid getting shot because sec sees you wielding a gun ready to shoot at all times. Right now if someone wields weapon most people expect you to be expecting combat right now or shortly after either because for example you will magdump them in face.

slarticodefast:
> Right now if someone wields weapon most people expect you to be expecting combat right now or shortly after either because for example you will magdump them in face.

The roleplay aspect of this might be an important point. Manual wielding signals getting ready for combat to those around you and could be used as a threat to better visualize escalation in fights.

thebadman4662:
Perhaps autowield should be in harm mode, while without it we would have just like its now? That way you could have best of both worlds kind of? You still kinda need to juggle harm mode to interact with UI even in combat but it would be something at least.

SlimmSlamm:
This is a really bad change. It's not really needed at all. This change will only cause more harm than good.

Hanzdegloker:
This feels unnecessary, its one button press that takes a second to do, plus what if I am just trying to pickup up a gun to put it away or move it, It will seem like I am about to start firing. Plus the RP aspect of being able to wield a gun in order to intimidate or signal your intentions is muddled with this.

keronshb:
I do not agree that wielding should be automatic, it's a single button press.
What if I wanted to pick a gun up but not wield it? This is a major issue.

nkokic:
In defense of this PR:
This is a great quality of life PR and removes gun input ambiguity.
Guns are heavy! It is alright to hold them with both hands at all times from realism perspective. 
Some items do not have wield options, while they look they do until you activate them in hand.
Pistols and SMG's like WT550 are NOT wieldable, while they looks so.
This solves ambiguity very well. Had an idea to implement it myself, but I am glad someone thought of it way before.

To comment on hostility escalation:
Picking up a gun already is a hostile escalation. In real life situation, you should not even attempt picking up a weapon in front of an officer. 
In case trying to secure a weapon for greater cause, you can take effort to unwield it by switching hands to give a nearby officer a sign you really won't use it. It really is a good way to show where your main hand is. Unless you have no free hands, then it is ambiguous.
Some sprites of wielded guns don't look too hostile anyway. (Sprites with barel facing sideways while wielding)

Not sure if people noticed or was video additionally uploaded, but from video it is clear than when switching hands to wieldable item while both hands are busy, you will not wield until you drop the other item which might be more input than previously.

What if I don't want to wield it? I would just switch hands.
What if I want to shoot it with one hand? I would just have an item in another hand.
Last part could be imagined with supporting a gun on a shield or gripping something small together with gun grip.
The gun stability still could be implemented as supporting on shield could be less practical and gripping 2 objects at same time with one hand speaks for itself.

Also need to ask this: Doesn't this PR also affect wieldable melee weapons?
If so:
For mops it might be pretty okayish. It is common to wield mop with both hands from perspective in real life.
Bats, axes and spears. They never give a friendly sight to begin with. Keep it in offhand if you really wish to turn down hostility.
Crusher melee weapons have different attack behavior when wielded. Have both hands busy if you want to mine with crushers. (might not be convenient for new players, but they may stumble on that feature when attempting to mine with both busy hands or just told by player).

Summary: This is a good PR. Player input and item action ambiguity is bad and this PR solves it.

Pumkin69:
i think this is a good PR that doesn't change the way the game is fundamentally played and reduces confusion. As long as you can swap hands to un-wield it is perfectly fine.

keronshb:
We discussed this and overall voted not to merge the PR.
Most of us prefer to have manual wielding.

PR #33582, opened 2024-11-26
Hotfix the randomly occurring DeleteAllThenGhost test failures

Errant-4:
## Technical details
During the test, mind entities are created along with mind roles, but then the mind roles are deleted manually rather than through the helper function. If a mind role gets deleted before its mind entity then this caused the mind entities to retain references to mind roles that no longer exist - when a certain code tries to check for these mind roles, things break. Now any references to a mind role are forcibly removed from it's mind entity when the MindRoleComponent shuts down for any reason. Even if the mind role entity would remain, it would be useless without this component so better to just write it off, as far as mind roles go


Errant-4:
> > If a mind role gets deleted before its mind entity then this caused the mind entities to retain references to mind roles that no longer exist
> 
> cant this just get handled by ComponentShutdown of MindRoleComponent

That is in fact what happens.

metalgearsloth (approved):

PR #33584, opened 2024-11-26
fix a station event weighting bug

IProduceWidgets:
## About the PR

Due to how the weighted event selector works, FindEvent(..), only integer weights resulted in the event ever getting selected.

## Why / Balance

bugfix

## Technical details

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: StationEvents with very low weights will now actually appear.

### DIFF
--- a/Content.Server/StationEvents/Components/StationEventComponent.cs
+++ b/Content.Server/StationEvents/Components/StationEventComponent.cs
@@ -9,14 +9,14 @@ namespace Content.Server.StationEvents.Components;
 [RegisterComponent, AutoGenerateComponentPause]
 public sealed partial class StationEventComponent : Component
 {
-    public const int WeightVeryLow = 1;
-    public const int WeightLow = 5;
-    public const int WeightNormal = 10;
-    public const int WeightHigh = 15;
-    public const int WeightVeryHigh = 20;
+    public const float WeightVeryLow = 0.0f;
+    public const float WeightLow = 5.0f;
+    public const float WeightNormal = 10.0f;
+    public const float WeightHigh = 15.0f;
+    public const float WeightVeryHigh = 20.0f;
 
     [DataField]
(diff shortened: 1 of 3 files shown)


IProduceWidgets:
We *could* support floats, but the solution here is kinda elegant, so Id rather just restrict to int in the component so people dont make the mistake.

slarticodefast:
Your changelog is formatted wrong.

IProduceWidgets:
> Your changelog is formatted wrong.

50% of the time it works every time.

slarticodefast:
Why does it cast the weight to an integer anyways?
Weights are already floats.
grafik
And it looks like it FindEvent could be easily changed to use floats.

IProduceWidgets:
> Why does it cast the weight to an integer anyways? Weights are already floats. grafik And it looks like it FindEvent could be easily changed to use floats.

I assume its to avoid a random float and dealing with the bounds.
Consider 5 events with weights 1, 1, .2, .8, 1, and 3 random draws of 2, 3, and 4.

Draw a selects the 2nd event
Draw b selects the 4th event
Draw c selects the 5th event

With integer random draws its not possible to get all the events.

slarticodefast:
Couldn't the random integer just be changed to a random float from 0 to sumOfWeights?
Or is my math wrong here?

IProduceWidgets:
> Couldn't the random integer just be changed to a random float from 0 to sumOfWeights? Or is my math wrong here?

No it can, just a stylistic difference I suppose.
I went ahead and kept floats using NextFloat()

slarticodefast (approved):

slarticodefast:
Good catch with the bug. We would likely have never noticed and it looks like it affects some other weights as well, for example the Urist swarm.
Thanks for the fix!

chromiumboy:
The math makes sense, thanks for the fix

PR #33585, opened 2024-11-26
Standardize vials

Zap527:
## About the PR
Vials now behave like bottles.

## Why / Balance
Vials previously behaved almost like bottles, but with a lot of weird minor differences such as not fitting in medical belts, not breaking when thrown, etc. 
This has been ironed out, allowing vials to be inserted where you feel like they really should be able to be inserted. Also adjusted the vial health to be in line with the bottle's.

## Technical details
Made `BaseChemistryEmptyVial` inherit from `BaseChemistryEmptyBottle`

## Requirements

## Breaking changes
The vial's solution was renamed from `beaker` (???) to `drink`. Vials filled with reagents will need to adjust this.

**Changelog**
:cl: Jajsha
- fix: Vials now fit in medical belts and chemsitry shelves.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/chemistry-vials.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/chemistry-vials.yml
@@ -2,20 +2,17 @@
 
 - type: entity
   name: vial
-  parent: BaseItem
+  parent: BaseChemistryEmptyBottle
   id: BaseChemistryEmptyVial
   description: A small vial.
   components:
   - type: Tag
     tags:
     - Trash
+    - Bottle
     - CentrifugeCompatible
-  - type: PhysicalComposition
-    materialComposition:
-      Glass: 25
   - type: FitsInDispenser
-    solution: beaker
-  - type: SpaceGarbage
+    solution: drink
   - type: Sprite
     sprite: Objects/Specific/Chemistry/vial.rsi
     layers:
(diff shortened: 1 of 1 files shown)


Everturning:
vials were intentionally meant to not be broken when thrown

Zap527:
I feel like vials and bottles should behave similarily, for consistancy's sake. Vials would still break when thrown, it just takes two throws (and the first doesn't even give any feedback that the vial was damaged). 

I can bring back the fact that they dont break immediately if it's crucial to the identity of the vial, but in my opinion, all glass chemistry instruments should be brittle and break when thrown.

thebadman4662:
> Vials previously behaved almost like bottles, but with a lot of weird minor differences such as not fitting in medical belts
?????
They were fitting my chembags, medibelts and pill canisters just fine ever since they got added?

Zap527:
They did go in chembags, but they didn't have the `Bottle` tag, so didn't pass the whitelist for the medical belt. 
image
image

They *did* fit in pill bottles while that was a thing, because pill bottles just checked for item size. But it's been removed now.

thebadman4662:
Hmm, now that I think about it maybe they went into my medbelt because I kept them in pill canisters. 😅 

Maybe thats even the reason why I started using canisters and I just forgot about it.

SlamBamActionman:
Hello, we've had a maintainer discussion and come to the consensus of closing this PR for the following reasons:

Vials being another flavor of bottle doesn't provide interesting gameplay, and instead they should be used in a medicine-production situation. Bottles, being portable in the medical belt, favor paramedics and doctors on the go. Vials are instead intended to be used by chemists and non-medical consumers (e.g. botany). This makes sense from the perspective that test tubes are not very portable, and if you search for real-life examples of portable medicinal containers there's a strong preference towards bottles. 

We'd rather see vials be given more unique interactions, like how they have the centrifuge-compatible tag, than being moved towards bottle behavior. They shouldn't be a downgrade, but a sidegrade that favors chemists over paramedics.

SlamBamActionman:
On a personal note, I do think there would be value in making vials fit inside of chemistry lockers at the very least, since those are stationary chemistry equipment. But it shouldn't be done by giving the vials a bottle tag in that case.

PR #33589, opened 2024-11-26
Makes MultipleToolComponent able to have different speed modifiers per mode

SpaceRox1244:
## About the PR

Adds optional speed modifiers to tools with multiple modes.

## Why / Balance

This PR was previously adding this optional field so that it could give screwdrivers very bad prying abilities as an alternative toggled mode. The feedback to this change led to the PR being specifically about the change to the component, because why not.

## Requirements

## Breaking changes

**Changelog**

No CL

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Tools/tools.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/tools.yml
@@ -78,19 +78,6 @@
       - Screwing
     useSound:
       collection: Screwdriver
-  - type: MultipleTool
-    statusShowBehavior: true
-    entries:
-      - behavior: Screwing
-        useSound:
-          collection: Screwdriver
-      - behavior: Prying
-        speedModifier: 0.4
-        useSound:
-          path: /Audio/Items/crowbar.ogg
-  - type: Prying
-    enabled: false
-  - type: ToolTileCompatible
   - type: RandomSprite
     available:
       - enum.DamageStateVisualLayers.Base:
--- a/Resources/Prototypes/Entities/Objects/Tools/tools.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/tools.yml
@@ -78,6 +78,19 @@
       - Screwing
     useSound:
       collection: Screwdriver
+  - type: MultipleTool
+    statusShowBehavior: true
+    entries:
+      - behavior: Screwing
+        useSound:
+          collection: Screwdriver
+      - behavior: Prying
+        speedModifier: 0.4
+        useSound:
+          path: /Audio/Items/crowbar.ogg
+  - type: Prying
+    enabled: false
+  - type: ToolTileCompatible
   - type: RandomSprite
     available:
       - enum.DamageStateVisualLayers.Base:
@@ -252,6 +265,7 @@
     statusShowBehavior: true
     entries:
       - behavior: Screwing
+        speedModifier: 1.5
         sprite:
           sprite: Objects/Tools/drill.rsi
           state: drill_screw
@@ -260,6 +274,7 @@
         changeSound:
           path: /Audio/Items/change_drill.ogg
       - behavior: Anchoring
+        speedModifier: 1.5
         sprite:
           sprite: Objects/Tools/drill.rsi
           state: drill_bolt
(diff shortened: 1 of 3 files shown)


Cojoke-dot:
This would get annoying very fast I think... trying to close panels and instead prying machines/doors open when they should instead be screwed. A lot of building functions work where prying and screwing are the steps right before and after, like how you pried the components from the lathe when normally the screwdriver would complete the machine.

slarticodefast:
Uff, this will probably break a *lot* of interactions where you can both pry and screw.

deathride58:
image

Gonna have to leave our vote for closure here. The tools are all designed to serve specific purposes, and giving overlapping functionality to existing tools would lead to an unnecessary level of redundancy among the roster. Doafter timer modifiers are not sufficient to balance it out, as the ability for an item to function as a given tool is far more valuable than how fast it can be done.

Resources/Prototypes/Entities/Objects/Tools/tools.yml
```diff
@@ -78,6 +78,19 @@
...
+  - type: MultipleTool
+    statusShowBehavior: true
+    entries:
+      - behavior: Screwing
+        useSound:
+          collection: Screwdriver
+      - behavior: Prying
+        speedModifier: 0.4
+        useSound:
+          path: /Audio/Items/crowbar.ogg
```
nikthechampiongr, line 90: This does not actually alter the speed of prying stuff like doors. The Prying component has its own modifier

hyphenationc:
I gotta agree with Deathride here. Partially because redundancy in the roster, and partially because *your screwdriver would not survive this with a vacuum-sealed, motor-operated airlock*

Resources/Prototypes/Entities/Objects/Tools/tools.yml
SpaceRox1244: I tested it and it does seem to affect the speed of prying doors. Before I added it, screwdrivers and crowbar pried doors at the same speed.

SpaceRox1244:
I'll respond more completely when I can but I absolutely see what yall are saying about prying open airlocks and all of that can be charged. I wanted to include it to see reception because it's easier to pull back than push for more and all that.

SpaceRox1244:
Alrighty so I'm seeing the reception and I think I'm gonna close this one up here. It was worth a shot to explore the concept. The motivation was primarily following "realism", where I could see a screwdriver being used by someone as some crude prying object. Unfortunately there is something a little bit uncomfortable about how when you apply the similar inspection to the other tools, you can't really come up with similar improvisational applications.

SpaceRox1244:
Thank you all for your feedbacks and take care :0

SpaceRox1244:
Or actually, lemme just make the PR about the tool speed. It's a helpful little addition. Will reopen shortly

Resources/Prototypes/Entities/Objects/Tools/tools.yml
SpaceRox1244: Oh shoot im so sorry. I didnt even push the change that adds the speed modifier to the actual PryingComponent.

SpaceRox1244:
It's anyone else's choice if this wants to be merged now or not. Admittedly it already creates a tiny bit more yaml noise for two current items for an option that isnt taken advantage of yet. Just wanted to salvage what I could.

slarticodefast:
The functionality with the extra datafield for the tool speed is ok to have I guess, even if we don't really need it maybe someone else can.
But for the included balance changes I will need a better reason than "why not" or remove them before this can be merged.

SpaceRox1244:
> The functionality with the extra datafield for the tool speed is ok to have I guess, even if we don't really need it maybe someone else can. But for the included balance changes I will need a better reason than "why not" or remove them before this can be merged.

I don't think this PR currently includes any balance changes. That being said, if it's just an 'I guess' and salvaging the PR isn't really that helpful, then I'm not too enthusiastic about adding it and I'd be ok with closing this up. This was easy enough to implement that it'd probably be better to just let someone when they need it in the future. 🤷

SpaceRox1244:
Thanks for the review btw

PR #33594, opened 2024-11-27
Fix space ambient music

slarticodefast:
## About the PR
Space ambient music wasn't playing anymore when you were floating in space.
I also reduced the range for engineering ambient music a little. It checks for nearby devices with engineering access as a condition and has higher priority than maintenance ambient music. But since there is almost always a substation door or firelock nearby in maints, you could never hear it.

## Why / Balance
Bugfix

## Technical details
The rule checking if you have a grid in range was not inverting the result correctly.

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- fix: Fixed space ambient music not playing.

### DIFF
--- a/Content.Shared/Random/Rules/GridInRange.cs
+++ b/Content.Shared/Random/Rules/GridInRange.cs
@@ -34,6 +34,6 @@ public sealed partial class GridInRangeRule : RulesRule
             return !Inverted;
         }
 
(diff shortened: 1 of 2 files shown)


Golinth:
Thank you for this, I’ve been meaning to make a bug report because I’ve been missing the music

PR #27111, opened 2024-04-19
cham projector fixes/rewrite

deltanedas:
## About the PR
- disguise is now just a disguise so if you dont have a hardsuit you can take pressure damage
  fixes #27099
- fixed ciggie visualiser so you can disguise as burned and lit ciggies :trollface:
- cant disguise as things in containers
- finally made it not use polymorph
- fixes #27130

## Why / Balance
:bug: :hammer:

disguising as something in your pocket is cringe like !pch on prop hunt, randomly turn into something that cant be there

## Technical details
- an aghost using projector no longer throws
- instead of polymorph, it makes your player invisible and spawns a disguise parented to it.
  being invisible disables verb menu and click attacks but shooting wide swing spacing radiation etc still damage you
- damage revealing is now just based on taking X damage while disguised, instead of a polymorph entity going crit

## Media
popup when trying to disguise as something in your pocket/bag/whatever

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes
no longer uses polymorph, the disguise entity is now just for visuals

**Changelog**
no cl no fun

PR #33602, opened 2024-11-27
Fix BuckleSystem always marking InteractHandEvent as Handled

### LINKED ISSUE #33263: Pests like mice, mothroaches and cockroaches can't be picked up by clicking on them.
## Description

It does work if you right click and select pick up but not if you just click on them. Also the trashbag picks them up.

**Reproduction**

- spawn mouse
- click on mouse
- the mouse should remain on the floor

- right click mouse
- select pick up
- mouse is now in your hand

**Screenshots**

**Additional context**

MilenVolf:
## About the PR
The issue has popped up only recently. Due to #27111 being merged, SharedChameleonSystem required to work with InteractHandEvent before SharedItemSystem, so SharedBuckleSystem.Interaction was working with that event before SharedItemSystem. Because of SharedBuckleSystem.Interaction always marking InteractHandEvent as Handled, it was impossible to pickup the mob.

## Why / Balance
Fixes #33263

## Technical details
This PR adds check for a successful unbuckling before marking InteractHandEvent as Handled

## Media

https://github.com/user-attachments/assets/e497f9ca-3743-46f3-99aa-6cf7c4b5009b
## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Mouses, mothroaches, cockroaches and etc can be picked up once again.

### DIFF
--- a/Content.Shared/Buckle/SharedBuckleSystem.Interaction.cs
+++ b/Content.Shared/Buckle/SharedBuckleSystem.Interaction.cs
@@ -114,14 +114,10 @@ public abstract partial class SharedBuckleSystem
         if (args.Handled)
             return;
 
-        var canUnbuckle = false;
-
         if (ent.Comp.BuckledTo != null)
             args.Handled = TryUnbuckle(ent!, args.User, popup: true);
 
         // TODO BUCKLE add out bool for whether a pop-up was generated or not.
-        if (canUnbuckle)
-            args.Handled = true;
     }
 
     private void AddStrapVerbs(EntityUid uid, StrapComponent component, GetVerbsEvent<InteractionVerb> args)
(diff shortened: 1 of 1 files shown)


Content.Shared/Buckle/SharedBuckleSystem.Interaction.cs
```diff
@@ -114,11 +114,14 @@ private void OnBuckleInteractHand(Entity<BuckleComponent> ent, ref InteractHandE
         if (args.Handled)
             return;
 
+        var canUnbuckle = false;
+
         if (ent.Comp.BuckledTo != null)
-            TryUnbuckle(ent!, args.User, popup: true);
+            canUnbuckle = TryUnbuckle(ent!, args.User, popup: true);
```
slarticodefast, line 120:
```suggestion
            args. Handled = TryUnbuckle(ent!, args.User, popup: true);
```
just do it like this and remove canUnbuckle

slarticodefast (changes requested):

slarticodefast (approved):
Tested and works as intended. Thanks for fixing it!

SaphireLattice (approved):
Unconditional "handled" is bad, yeah

Works fine, does not seem to break things from *very* brief testing

PR #33610, opened 2024-11-27
Refactor audio system to send collection IDs over the network

sowelipililimute:
## About the PR
This updates SS14 to compile with the changes made in https://github.com/space-wizards/RobustToolbox/pull/5540.

## Why / Balance
This is important groundwork for future features such as captioning, as a caption and other data can be associated with the collection prototype instead of passing extra data everywhere with the sound.

## Technical details
- Updated RobustToolbox submodule (see https://github.com/space-wizards/RobustToolbox/pull/5540)
- Updated miscellaneous files to use the new `RobustSoundSpecifier` types as required for the compile to work

## Requirements

## Breaking changes
- The `AudioSystem`s now deal in `ResolvedSoundSpecifiers` rather than `string`s as the "raw" path format. An implicit conversion from `string` to `ResolvedPathSpecifier` was added to ease the breaking nature of the changes.

Requires https://github.com/space-wizards/RobustToolbox/pull/5540

### DIFF
--- a/Content.Client/Audio/ClientGlobalSoundSystem.cs
+++ b/Content.Client/Audio/ClientGlobalSoundSystem.cs
@@ -66,7 +66,7 @@ public sealed class ClientGlobalSoundSystem : SharedGlobalSoundSystem
     {
         if(!_adminAudioEnabled) return;
 
-        var stream = _audio.PlayGlobal(soundEvent.Filename, Filter.Local(), false, soundEvent.AudioParams);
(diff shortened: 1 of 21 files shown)


RobustToolbox
Aeshus, line 1: When making a PR that requires an RT (RobustToolbox) change, you don't commit the RT submodule change.

https://docs.spacestation14.com/en/general-development/tips/prs-with-engine-changes.html

This is because it can cause merge conflict issues, and if your change accepted, the maintainers will handle ensuring that both get merged at the same time (as they have different release schedules/systems).

Instead, you should just have this in your PR description:

```
requires [pr]
```

PR #33613, opened 2024-11-27
Remove Mop bucket from janitorial supplies

NotSoDana:
## About the PR

Removed mop bucket from cargo janitorial supplies crate

## Why / Balance

This bucket is a real headache for cargo and janitors, cause when you take it out of crate, you can't put it back in because of its collision. As a result, it remains deadweight that cant be disposed or putten in crate, but only broken, which is very annoying. Its absence will not cause any problems, cause there is no difference from an ordinary bucket or cart, except perhaps in its inconvenience.

## Technical details

yaml

## Requirements

## Breaking changes

**Changelog**

:cl: Dezzzix
- remove: Removed mop bucket from janitorial supplies crate!

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Crates/service.yml
+++ b/Resources/Prototypes/Catalog/Fills/Crates/service.yml
@@ -7,7 +7,6 @@
   - type: StorageFill
     contents:
       - id: MopItem
-      - id: MopBucket
       - id: Bucket
         amount: 3
       - id: WetFloorSign


lzk228:
it deserves a cl
also you might add another way obtaining the mop bucket, maybe just another crate

SlimmSlamm:
you should just make it so it can go in a crate.

Hanzdegloker:
I pondered this once, make the jani bucket able to be folded and unfolded, collapsible, like a roller bed. Then it could be in the crate collapsed and just be unfolded. Course make it spill everything in it, if applicable, when collapsed and yeah. somethin somethin 2x2 in inventory yeah, if you wanna know how its still water tight, its space magic.

IProduceWidgets:
Just order the bucket as if it were a crate.

slarticodefast:
> Just order the bucket as if it were a crate.

Yeah, completely removing it from cargo would mean you cannot obtain any new ones. Making it a separate order could work, similar to the janicart.

superjj18:
Add a screwdriver mechanic to just break it down into plastic?

TheKilk:
Why can't we change it to an object instead of a structure, prohibit putting it in a backpack to maintain balance with a simple bucket, and keep its filled version as a structure so that it can't be hidden in a crate?

southbridge-fur:
I feel like folks are forgetting there already is a cargo order for the nice janitor trolley. Only difference is you can't put a shark plushie in it.

slarticodefast:
Closing in favor of https://github.com/space-wizards/space-station-14/pull/34586

PR #33616, opened 2024-11-28
Stop mobs spawning inside the off-limits trade station room

IProduceWidgets:
## About the PR

Adds a marker to prevent events from spawning mobs near them.

image

## Why / Balance

Some places shouldn't get random mobs, its not fun.

## Requirements

## Breaking changes

**Changelog**

no cl no fun


IProduceWidgets:
> Could we add the markers to the rest of the ATS too
> 
> Not that I don't want slimes or spiders to spawn there, the main problem is the dragon can sometimes spawn there and get stuck.

Dragon spawns in space, so this wont help with that.

Content.Server/StationEvents/Events/RandomSpawnRule.cs
```diff
@@ -1,19 +1,32 @@
...
 {
+    [Dependency] private readonly EntityLookupSystem _lookup = default!;
+
     protected override void Started(EntityUid uid, RandomSpawnRuleComponent comp, GameRuleComponent gameRule, GameRuleStartedEvent args)
     {
         base.Started(uid, comp, gameRule, args);
 
-        if (TryFindRandomTile(out _, out _, out _, out var coords))
+        for (var i = 0; i < 10; i++) // infinite loop prevention
         {
```
metalgearsloth, line 18: Please don't do arbitrary loop checks unless the actual solution is unbounded (e.g. pathfinding).

metalgearsloth (changes requested):
This is just areas, just make them only spawn on station grid instead.

IProduceWidgets:
> This is just areas, just make them only spawn on station grid instead.

They already do that, the ATS is a stationmember because cargo / bank account. I 'want to exclude all members what if multi-part station?

Content.Server/StationEvents/Events/RandomSpawnRule.cs
IProduceWidgets: Have a better idea? try 10 is already what TryFindRandomTile does.

IProduceWidgets: I could stick this check into TryFindRandomTile, which I did think about, but it felt inelegant when that method gets used for other non-mob stuff a lot.

IProduceWidgets: The only truly bounded way I can think to do this is enumerating all the entities with the prevention component, getting their coords, and then outerjoining that with a dict of all tiles or something.

IProduceWidgets: > The only truly bounded way I can think to do this is enumerating all the entities with the prevention component, getting their coords, and then outerjoining that with a dict of all tiles or something.

I sat with this for a few hours looking at how to do it. I could probably do it with a snowflake method that only works for this specific component. Doing it in a more generic method for other components would mean I have to do something batty like reflection over EntityQueryEnumerator which I'm going to go out on a limb and say is a non-starter.

deltanedas:
change station grids to be actual parts of the station and grids it owns, have bank etc check owned grids and have events spawn on actual grid(s)

TheShuEd:
@metalgearsloth ?

IProduceWidgets:
Ahhh! Sorry I'm being slow. A wild life appeared and kicked my ass.

TheShuEd:
Given that this PR has been in a “Awaiting Changes” state for over 3 months without any activity, I'm closing it as derelict. If you would like to continue, please write and the PR will be reopened.

PR #33619, opened 2024-11-28
Randomized maints rooms

Spessmann:
## About the PR

I used the roomfill vgroid exped gen thing to make some maints rooms.
Mostly made as a test, ~~more rooms of are planned to be added~~ edit: I added more. now we force jr mappers to make even more for us :trollface: 
- It is not mapped anywhere yet
- rotation is randomized (note: because the orientation is randomized i needed to use spawners so the items would appear at the correct orientation. Also tileoverlay decals dont work properly with rotation so they are not used)
- they are not wired, nor have lights in order to reduce conflicts with maps
## Why / Balance

Randomized maints adds more variety

## Media

image
image
image

## Requirements

**Changelog**

no cl


Thatonestomf:
Damn, this is really gonna change the maints experience.

Spessmann:
something something i need to add the prototype i made somewhere? I don't really know where

muburu:
1 step closer to full procgen stations :godo:

beck-thompson:
Just FYI the test fail appears to be real!

TheShuEd:
You need to add Maints tag prototype
Use github search for
```- type: Tag```

Ko4ergaPunk:
What about wires?

Spessmann:
> What about wires?

doesn't include wires or atmos, each maints may or may not choose to have them plus you cant really know where its gonna be (to avoid stacking wires) so its better just to have the mapper manually add it if they wish

Spessmann:
To fix the orientation issues (the orientation/rotation of each room is random when it is spawned) I'm going to replace the mapped items with spawners/crates/lockers

SaphireLattice:
As I understand, that spawns a room centered at a specific marker right?

It probably would be good to use two corner markers that are linked together somehow (same multitool linking even), so that rooms of different sizes are possible

Spessmann:
> As I understand, that spawns a room centered at a specific marker right?
> 
> 
> 
> It probably would be good to use two corner markers that are linked together somehow (same multitool linking even), so that rooms of different sizes are possible

No, it spawns the bottom left corner of the room at the marker

Rooms of different sizes are possible

ps3moira:
is it only 5x5? no varying size?

ps3moira:
also zone code when 👀

Spessmann:
consistency is key

Ratyyy:
how did you make spawners like this?
image

Emisse (approved):

Boaz1111 (approved):
Looks good to me, only one issue with extra tiles spawning when the room are exposed to space on the left side, but that's unrelated to this PR, and happens with the already implemented random rooms too. Thanks for the PR!

PR #33630, opened 2024-11-28
Improved French accent (th sound)

ObaniGemini:
## About the PR

Enhance 'ze `th` part of 'ze French accent

## Why / Balance

2 reasons:
1.  The french accent couldn't have uppercase `'z` characters. While it looks good on the start of sentences, have fully uppercase messages looking like **'zIS IS 'zE WAY** was imo a problem.

2. Most french people use `'s` or `'f` when `th` preceeds a consonna.

Since it varies a lot depending on the people who use it and knowing that `'f` may be confusing (*'free* for *three* sounds confusing), I put `'s` as a general rule. Also, french people tend to force the `'S` consonna, so I doubled it randomly.

It's still imperfect tho, since french people would for example not say *zink* but *sink* for the word *think*, but it's still closer to reality imo.

I'm still not sure if the character `'` should be before the `s` either, but I feel like it makes it clearer that it's an accent artifact.

## Technical details

The RegEx replace for `th` was replaced by a RegEx matching, following the German Accent example.

The scary part of this code may be the `offset` variable which is there to ensure we offset the matching indexes when doubling the `'s`.

The code should be easily readable and testable by just logging to a server and testing the edgecases (which I think I already covered in my own tests).

## Media

image
image
image

## Requirements

## Breaking changes
No breaking changes

### DIFF
--- a/Content.Server/Speech/EntitySystems/FrenchAccentSystem.cs
+++ b/Content.Server/Speech/EntitySystems/FrenchAccentSystem.cs
@@ -1,5 +1,6 @@
 using Content.Server.Speech.Components;
 using System.Text.RegularExpressions;
+using Robust.Shared.Random;
 
 namespace Content.Server.Speech.EntitySystems;
 
@@ -8,6 +9,7 @@ namespace Content.Server.Speech.EntitySystems;
 /// </summary>
 public sealed class FrenchAccentSystem : EntitySystem
 {
+    [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly ReplacementAccentSystem _replacement = default!;
 
     private static readonly Regex RegexTh = new(@"th", RegexOptions.IgnoreCase);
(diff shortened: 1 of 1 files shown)


ObaniGemini:
Should I remove the double consonna part?
I think it probably would interfere with other language modifiers too much (and removing it would simplify the code greatly, especially on what looks like critical points)

ObaniGemini:
I simplified the code by removing the double consonna

BackeTako:
Omg yes I've been meaning to fix this, its been so annoying.

Content.Server/Speech/EntitySystems/FrenchAccentSystem.cs
```diff
@@ -27,14 +27,28 @@ public string Accentuate(string message, FrenchAccentComponent component)
...
+        
+        // replaces th with 'z or 's depending on the case
+        foreach (Match match in RegexTh.Matches(msg))
+        {
+            var uppercase = msg.Substring(match.Index, 2).Contains("TH");
+            var Z = uppercase ? "Z" : "z";
+            var S = uppercase ? "S" : "s";
+            var idxLetter = match.Index + 2;
+
+            // If th is alone, just do 'zis
```
Centronias, line 44:
```suggestion
            // If th is alone, just do 'z
```
is what this is doing, right? The extra `is` in the comment threw me off.

Centronias (commented):
Not that it's something you can properly code, but the difference between
`think` -> `'sink` and `this` -> `'zis` is [voicing](https://en.wikipedia.org/wiki/Voice_(phonetics)). In english, generally a `th` sound followed by a consonant will be unvoiced, which a person with a french accent would approximate with an `s`, which is also unvoiced.

So in a magical world where the game has infinitely more knowledge of english phonotactics, that'd be the implementation...

Content.Server/Speech/EntitySystems/FrenchAccentSystem.cs
ObaniGemini: maybe I shouldn't put jokes in my comments. Yes, that's what it does

tutoumi:
sooooooo, and now ?

ObaniGemini:
no idea, been waiting for something but nobody really reviewed this

EmoGarbage404 (approved):

beck-thompson (approved):
seems fine

PR #33638, opened 2024-11-29
Remove grasshopper from the panic bunker message

VasilisThePikachu:
Can be retargetted to stable if preferred, which will then be a hotfix


VasilisThePikachu:
Alerting head game admin @nikthechampiongr

nikthechampiongr (approved):

PR #32653, opened 2024-10-05
Feature/make radial menu great again

Fildrance:
## About the PR

Radial menu was great addition to UI but it requires precision as it is now. To bring up better UX it needs to be usable without too much precision so it can be used effectively in panic situations. This problem was already solved by wheel menu in most modern games - use sectors of circle to pick options and not icons themselves.

## Why / Balance

Robust wheel UI!

## Technical details

RadialMenu now uses 'special' button as context button in center - RadialMenuContextualCentralTextureButton. It considers space inside inner circle of parent RadialContainer half radius, and ALL outer space outside twice of its radius - as itself.
RadialContainer now puts angle sector data in each child  RadialMenuTextureButton.
RadialMenuTextureButton now can draw bagle segments!

## Media

[new-radia-menu.webm](https://github.com/user-attachments/assets/addb3dd8-45f8-45fe-8dce-770123c5adab)

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

removed RadialMenuButton - was not used.
now RadialMenuTextureButton requires to be inside RadialContainer to work properly

**Changelog**

:cl: Fildrance, SaphireLattice, eoineoineoin
- tweak: Radial menu UIs have been updated for readability.

PR #33640, opened 2024-11-29
Make RCD and emote radial menus open centered

Plykiya:
## About the PR

Makes the radial menu for the RCD and emote menu open at the center of your screen where your character is rather than wherever your mouse happens to be

## Why / Balance

Kind of an opinionated change but this feels a bit better. When you attempt to open radial menus at the moment, what happens is that the menu doesn't open immediately but instead gets "moved" around with your mouse for 100 or 200 ms, but it's enough to be annoying that you can't go straight for the option that you want. You have to wait for the menu to stop moving and then you can finally click on it. If the menu opens centered on your screen, you don't have that kind of delayed experience.

## Technical details

Deleted the code that caused the radial menus to open where your mouse location is

## Media

https://github.com/user-attachments/assets/4fe33b6c-37fb-4789-bf85-bb3fa5e80795

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The radial menu for emoting and RCDs now opens centered on your character.

### DIFF
--- a/Content.Client/RCD/RCDMenuBoundUserInterface.cs
+++ b/Content.Client/RCD/RCDMenuBoundUserInterface.cs
@@ -28,10 +28,6 @@ public sealed class RCDMenuBoundUserInterface : BoundUserInterface
         _menu = this.CreateWindow<RCDMenu>();
         _menu.SetEntity(Owner);
         _menu.SendRCDSystemMessageAction += SendRCDSystemMessage;
-
-        // Open the menu, centered on the mouse
-        var vpSize = _displayManager.ScreenSize;
-        _menu.OpenCenteredAt(_inputManager.MouseScreenPosition.Position / vpSize);
     }
 
     public void SendRCDSystemMessage(ProtoId<RCDPrototype> protoId)
--- a/Content.Client/UserInterface/Systems/Emotes/EmotesUIController.cs
+++ b/Content.Client/UserInterface/Systems/Emotes/EmotesUIController.cs
@@ -28,7 +28,7 @@ public sealed class EmotesUIController : UIController, IOnStateChanged<GameplayS
     {
         CommandBinds.Builder
             .Bind(ContentKeyFunctions.OpenEmotesMenu,
-                InputCmdHandler.FromDelegate(_ => ToggleEmotesMenu(false)))
+                InputCmdHandler.FromDelegate(_ => ToggleEmotesMenu()))
             .Register<EmotesUIController>();
     }
 
@@ -37,7 +37,7 @@ public sealed class EmotesUIController : UIController, IOnStateChanged<GameplayS
         CommandBinds.Unregister<EmotesUIController>();
     }
 
-    private void ToggleEmotesMenu(bool centered)
+    private void ToggleEmotesMenu()
     {
         if (_menu == null)
         {
@@ -50,16 +50,7 @@ public sealed class EmotesUIController : UIController, IOnStateChanged<GameplayS
             if (EmotesButton != null)
                 EmotesButton.SetClickPressed(true);
 
-            if (centered)
-            {
-                _menu.OpenCentered();
-            }
-            else
-            {
-                // Open the menu, centered on the mouse
-                var vpSize = _displayManager.ScreenSize;
-                _menu.OpenCenteredAt(_inputManager.MouseScreenPosition.Position / vpSize);
-            }
+            _menu.OpenCentered();
         }
         else
         {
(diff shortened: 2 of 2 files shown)


metalgearsloth (approved):

slarticodefast:
Wouldn't this be bad when we merge #32653?
Think of how other games with radial menus do it. You open it and move the mouse into the direction of the option you want to choose.

Hanzdegloker:
Can it be a setting to either open at the mouse or the center of the screen? I love it opening at the mouse as is right now since it means minimal mouse movement. Though I can see why some might want it at the center.

Errant-4 (commented):
It does not appear on your character in Separated UI mode, because it's not in the center of the screen. It feels kind of weird, because conceptually the player probably considers their character "the center", despite actual screen geometry

image

Plykiya:
> It does not appear on your character in Separated UI mode, because it's not in the center of the screen. It feels kind of weird, because conceptually the player probably considers their character "the center", despite actual screen geometry
> 
> image

right....

Errant-4:
> right... would you happen to know any way to get the width of the "separated" part of the UI?

ui.separated_chat_size

Note that there is an ancient bug which might be relevant here. when the game loads, the separated chat is always drawn with the default width despite remembering where you set it. When any alert icon gets updated for the first time, it will jump to the saved width. I never got around to fixing that one but it will matter here because the initial "apparent" chat width will not match what's in the cvar

chromiumboy:
I'll just say that when I made the original RCD PR, a staff member told me explicitly to make the menu open at the position of the cursor

ArtisticRoomba:
IMO I don't really like this change for a few reasons:
- Having the RCD select be at the center of the screen instead of your cursor on interact means that you now have to make an unnecessary cursor movement towards the center of the display. It is faster and more intuitive for the RCD to have the selection menu start at the cursor. When I start the selection, my cursor is at the center of all of the options, equidistant from all of them. If it was placed at the center of the screen, I would lose all of that muscle memory, and have to make a longer movement. This would just make the RCD feel sluggish to use and upset people that have gotten used to the current one.
- Okay, so why don't we make this for just the emote wheel? Well, it introduces discontinuity. Having the wheel system pop up at the position of the cursor for some, but not for others, is a huge break for consistency and wouldn't feel great.
- When radial menus are further expanded to cover interaction with more items, I expect the radial menu to center around the item (where my cursor basically is) to intuitively highlight *exactly* what I'm interacting with. This ties into the aforementioned consistency problem, if we were to make this behave for some items but not others. For the RCD, this centers around the area I'm trying to construct/deconstruct.

It would be like having the right click verb menu open on the center of your screen or on the side instead of at the point of your cursor. It causes an unnecessary movement, breaks a player expectation, and just feels a bit odd.

If it was done right now, I think just the emote wheel can be fine placed at the center of your character, and have this be the only discontinuity when it comes to emote wheels not at the cursor. After all, your emotes center around your character. Although, some players might explicitly bind their emote wheel to their mouse or keyboard, so when it brings it up, it's at the center of their cursor, allowing them to emote faster. Stuff like this might be good to have as an option in the settings. For example:

- Place interaction wheels at the center of your (character, cursor)
- Place the emote wheel at the center of your (character, cursor)

If these were options, I think it would be a good change for those that want it.

ilovehans10:
I'm a big fan of radial menus, and the whole point of them is to be as ergonomic as possible by [putting all menu option the same distance from the cursor](https://en.wikipedia.org/wiki/Pie_menu#Comparison_with_other_interaction_techniques). For an example of a radial menu in a program that doesn't lock the mouse to the center of the screen, Autodesk Inventor used to have a nice one, and quite a few other AutoCAD programs also do last I checked.

If the issue is the menu moving, then the simple answer here is to make the menu not do that. :godo: /hj

And for the love of god, please please please at least make this an option like suggested by ArtisticRoomba because I, for one, would still appreciate both radial menus opening at the mouse, even if it is the jankier option.

Errant-4 (commented):
I do agree that this should be optional

Fildrance:
> every game I can think of that does it also forces your mouse to be centered on screen at all times, but besides that, I can't remember the specific reason Sloth said as for why the menu doesn't stay in place and why it's hard to fix

Also remember that any click outside of new radial menu gonna close menu, and it appearing not on your mouse gonna make it easier to mess up.

PJB3005 (changes requested):
> what happens is that the menu doesn't open immediately but instead gets "moved" around with your mouse for 100 or 200 ms, but it's enough to be annoying that you can't go straight for the option that you want.

This is a prediction bug that should be fixed properly, not by working around it like this.

PR #33641, opened 2024-11-29
Hide emotes performed out of view

Plykiya:
## About the PR

Makes emotes that are performed out of visible view of the recipient not show up in chat

## Why / Balance

You should not be able to know that "Commander Sigma" is pointing at a door or motioning their allies to follow them when you can't even see them to know who they are or what they're doing. This is one of the biggest sources of meta-information that people abuse because the game offers it to them and not because they're intending to abuse it.

## Technical details

It more or less works the same way whisper code does since it's taken from whisper code. The only difference is that you don't see any obfuscated message, you either see it or you don't since that's the nature of actions performed by the character.

## Media

https://github.com/user-attachments/assets/d0e011d3-13c7-4e03-b054-b5df8df66dcc

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: You can no longer see emotes in chat that are performed out of your view.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -595,7 +595,29 @@ public sealed partial class ChatSystem : SharedChatSystem
 
         if (checkEmote)
             TryEmoteChatInput(source, action);
-        SendInVoiceRange(ChatChannel.Emotes, action, wrappedMessage, source, range, author);
+
+        // Taken more or less from whispering
+        // Get all people who could receive the message in normal speaking range
+        foreach (var (session, data) in GetRecipients(source, VoiceRange))
+        {
+            EntityUid listener;
+
+            if (session.AttachedEntity is not { Valid: true } playerEntity)
+                continue;
+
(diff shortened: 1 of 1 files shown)


ScarKy0:
Does this also fix "Commander Quebec wields their c-20" when in lockers/crates? :godo:

Plykiya:
> Does this also fix "Commander Quebec wields their c-20" when in lockers/crates? :godo:

no, because that's not an emote

Hrosts:
Can you still hear the sounds emotes make?

Plykiya:
> Can you still hear the sounds emotes make?

yes

SlamBamActionman (approved):
While this touches chatcode, I think it adds enough value and utility to the game that I can accept it bypassing the restriction. Since it seems to use the center of the entity for the check you can get some odd situations where you can see parts of the character but still not see the emote, but it's kind of an edgecase. 

It is worth noting that this will remove the ability to do stuff like "* bangs on the door" but I can accept that; if it ends up impacting player expression there is always an option to revert.

Plykiya:
Is there anything left for me to do regarding this change?

Issue #18968, opened 2023-08-10
Examines "go through" the Strip window


Errant-4:
## Description
The Stripping window gets overridden by objects underneath it for purposes of Examine
Clicking on any item in the Strip window works correctly, but Examine gives station objects priority over the interface, so often you can only examine them by awkwardly clicking around in the inventory box or maneuvering the stripping window over a sparser area

This is an old issue but I found no open ticket for it

## Replication
Open strip interface
Move one of the window's slots over a large clutter of items
Try to examine the slot's item

**Screenshots**
![image](https://github.com/space-wizards/space-station-14/assets/35878406/a9622217-a546-444c-855b-2d0df0a627f3)

Pangogie:
Fixed at some point.

lzk228:
no it is not
![image](https://github.com/space-wizards/space-station-14/assets/124214523/1bebd90a-f16f-40d1-96fa-8d5bbb96e29b)
i have rebinded examine button and it happens EVERY time when i NOT shift+leftclicking

PR #33642, opened 2024-11-29
Fix for inspecting entities in hand

Plykiya:
## About the PR

You can now inspect entities in your hand without it attempting to inspect the entity beneath the hand-slot in the world

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/27221

You weren't able to inspect the entity in your hand if there was any inspectable entity underneath where you clicked

## Technical details

Handle the click arg on the hand slot instead of letting the click continue on to do whatever other things

## Media

Before:

https://github.com/user-attachments/assets/4018baad-1c5c-463d-b257-e8e527eff5c3

After:

https://github.com/user-attachments/assets/09d4a78b-9b18-4a96-9dc8-56298dc6da4a

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- fix: You can now inspect the entity in your hand instead of the entity that happened to be underneath your hand.

PR #33644, opened 2024-11-29
Fix for inspecting entities in the stripping window

Plykiya:
## About the PR

You can now inspect entities in the stripping window without it attempting to inspect the entity underneath the window

## Why / Balance

It was broken

## Technical details

Same fix from https://github.com/space-wizards/space-station-14/pull/33642 applied to this, just handle the event so it doesn't try to continue examining other stuff

## Media

https://github.com/user-attachments/assets/23ceb642-4e3f-4d7b-a178-6432e29f6df5

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: You can now inspect entities in the stripping window instead of the entity that happened to be underneath the window.

### DIFF
--- a/Content.Client/Inventory/StrippableBoundUserInterface.cs
+++ b/Content.Client/Inventory/StrippableBoundUserInterface.cs
@@ -191,9 +191,15 @@ namespace Content.Client.Inventory
                 return;
 
             if (ev.Function == ContentKeyFunctions.ExamineEntity)
+            {
                 _examine.DoExamine(slot.Entity.Value);
+                ev.Handle();
+            }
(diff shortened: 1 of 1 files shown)


Errant-4:
fixes #18968

Errant-4 (approved):
Thank you

slarticodefast:
I still want to find the underlying cause of why this only happens with certain keybinds. But I guess it works for now.

Errant-4:
The default uses a key AND a click. I guess that somehow makes a difference from all the custom binds we all use where it's just a single key/button

PR #33645, opened 2024-11-29
Hide wielding popup for entities not in same container as wielder

Plykiya:
## About the PR

Hides the popup text that others see when wielding and unwielding guns if the gun is wielded from inside a container and the other player is not in the container as well

## Why / Balance

People tend to fidget with wielding and unwielding their guns while in crates/lockers waiting for an ambush and players will use this meta information.

## Technical details

Cleanup:
- Adds summary and comments to WieldingSystem
- Rearranges some events listeners such as moving the GetMeleeDamageEvent near the GunRequiresWieldToMelee comp, updated every function to use the `Entity<T>` format
- Deletes the WieldableComponent listener for the ItemUnwieldedEvent and merges it with TryUnwield

Actual changes:
- In TryWield() and TryUnwield(): If the wielder is in a container, the wielding and unwielding popup is hidden from players that aren't in the same or parent container as the wielder

This kind of non-meta storage excluding popup is probably common enough that it deserves its own helper method or something...

## Media

https://github.com/user-attachments/assets/7e259d7c-f0ae-42cf-975c-d8345c8e562e

## Requirements

## Breaking changes

Changed CanWield, TryWield, and TryUnwield from public to private access
All WieldableSystem code updated to use `Entity<T>` instead of Entity and Component

**Changelog**

:cl:
- tweak: You no longer see the popup text for wielding a gun if the person wielding is in a container and you are not in the same container.

### DIFF
--- a/Content.Shared/Wieldable/WieldableSystem.cs
+++ b/Content.Shared/Wieldable/WieldableSystem.cs
@@ -18,7 +18,9 @@ using Content.Shared.Weapons.Ranged.Events;
 using Content.Shared.Weapons.Ranged.Systems;
 using Content.Shared.Wieldable.Components;
 using Robust.Shared.Audio.Systems;
+using Robust.Shared.Containers;
 using Robust.Shared.Network;
+using Robust.Shared.Player;
 using Robust.Shared.Timing;
 
 namespace Content.Shared.Wieldable;
@@ -35,6 +37,7 @@ public sealed class WieldableSystem : EntitySystem
     [Dependency] private readonly SharedGunSystem _gun = default!;
     [Dependency] private readonly IGameTiming _timing = default!;
     [Dependency] private readonly INetManager _netManager = default!;
+    [Dependency] private readonly SharedContainerSystem _container = default!;
 
     public override void Initialize()
     {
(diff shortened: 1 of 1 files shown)


Plykiya:
not my test fail

Plykiya:
strangely it seems to duplicate the unwielding message now? going to figure out why

Plykiya:
it unwields the gun once when you actually unwield it and then again when the virtual item is deleted, despite there being a check for the wield status... I'm so confused.

Plykiya:
I really don't understand how there is supposedly a difference in the networking of a PopupPredicted and a PopupEntity when they both just use RaiseNetworkEvent...

Plykiya:
I mean, functionally it works, it just makes an existing bug more apparent (unwield message is duplicated) that I don't quite know how to fix.

The problem is that two ItemUnwieldedEvents are thrown, once through the swap/verb/drop methods, and then again whenever the VirtualItem is deleted. I can't set the event arg to handled because they're occurring through two different event handlers. I can't just delete the VirtualItemDeletedEvent handling because it's the only thing covering the situation where the user presses the drop button on the VirtualItem in their offhand.

Theoretically it's checking that component.Wielded is true before unwielding, but because both events happen at the same time and are networked, it just doesn't update in time?

Content.Shared/Wieldable/WieldableSystem.cs
```diff
@@ -276,9 +288,7 @@ public bool TryUnwield(EntityUid used, WieldableComponent component, EntityUid u
         if (ev.Cancelled)
             return false;
 
-        component.Wielded = false;
         var targEv = new ItemUnwieldedEvent(user);
-
         RaiseLocalEvent(used, targEv);
         return true;
     }
```
metalgearsloth, line 294: ... Is this subscribing to its own event? Might as well just roll it into it and save the performance.

Plykiya: sure thing

Plykiya:
so, confusingly, I added a Logger.Debug() message right before the PopupEntity() line to see if it was actually getting sent multiple times

every client and the server all agree that there is only one TryUnwield call and only one PopupEntity call, but somehow the popup gets duplicated for other clients

Issue #32562, opened 2024-09-30
Keyup and walk key will break pulling


Plykiya:
## Description

Be walking in one direction, get pulled by someone, once you lift up on the movement key it will break the pull.

Similarly, you can stop someone pulling you by spamming the shift (walk) key non-stop.

**Reproduction**

**Screenshots**

**Additional context**

PR #33648, opened 2024-11-29
Don't break pull on walk key or keyup movement

Plykiya:
## About the PR

Pulling no longer stops if the person being pulled activates their walk key or if their movement key is lifted

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/32562

It's not intuitive that pressing the walk key counts as a movement input to break pulling, and choosing not to move by lifting your hand off your movement keys should also not count as a movement for the sake of breaking a pull.

## Media

Holding left key, controlling the mob, lifting up on the key, then spamming shift.

https://github.com/user-attachments/assets/604fc2ca-c6a3-455b-b71d-8293d047d335

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Pulling will no longer be broken if the person being pulled toggles their walk key or lifts up from a movement key.

### DIFF
--- a/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
+++ b/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
@@ -248,7 +248,7 @@ public sealed class PullingSystem : EntitySystem
         if (!_blocker.CanMove(entity))
             return;
 
-        if (args.OldMovement == MoveButtons.None)
+        if (args.OldMovement != MoveButtons.None)
             return;
 
         if (args.OldMovement == MoveButtons.Walk)
--- a/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
+++ b/Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
@@ -248,6 +248,12 @@ public sealed class PullingSystem : EntitySystem
         if (!_blocker.CanMove(entity))
             return;
 
+        if (args.OldMovement == MoveButtons.None)
+            return;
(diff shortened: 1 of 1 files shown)


Plykiya:
I don't actually know if it 100% solves the keyup cases but it's at least better than it was before

ScarKy0:
Jeez youre on a bugfixing spree

Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
```diff
@@ -248,6 +248,12 @@ private void OnPullableMoveInput(EntityUid uid, PullableComponent component, ref
         if (!_blocker.CanMove(entity))
             return;
 
+        if (args.OldMovement == MoveButtons.None)
```
slarticodefast, line 251:
```suggestion
        if (args.OldMovement != MoveButtons.None)
```
Isn't this the wrong way around?
I tested it and now keydown won't break the pull.
Keydown means switching from none to a direction.

slarticodefast (changes requested):

slarticodefast:
Gave it another test.
That fixed the keydown problem, but now the walk key cancels it again on keydown...
You might need some more complex logic for the guard statement since they are bit flags.

Plykiya:
> Gave it another test.
> That fixed the keydown problem, but now the walk key cancels it again on keydown...
> You might need some more complex logic for the guard statement since they are bit flags.

Tragic.

PR #33660, opened 2024-11-29
Juiceable slimeballs

luizwritescode:
## About the PR

Adds extractable component to slimeball prototype.

Now slimeballs can be juiced for 10u slime instead of just grinded for nutriment.

I'm sure there are more items that could be juiced but slimeball always confused me.
## Why / Balance

Slimeball is a child of FoodMeat prototype. You can get protein and nutriment from it but, weirdly, no slime.

This allows to extract slime from the item without removing the original recipe, in turn adding one more blender interaction to the game.
## Technical details

N/A

## Requirements

## Breaking changes

None

**Changelog**

Such a small change does it really need cl entry?

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/meat.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/meat.yml
@@ -678,6 +678,11 @@
     tags:
     - Raw
     - Meat
+  - type: Extractable
+    juiceSolution:
+      reagents:
+      - ReagentId: Slime
+        Quantity: 10
   - type: Sprite
     state: slime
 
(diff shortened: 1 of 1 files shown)


slarticodefast (approved):
Looks good to me 👍

thebadman4662:
Always assumed slimeballs are not juiceable and in ball form in first place because they are condensed and basically dry nutriments(266% nutrition compared to slime) while all that slime otherwise is still nutritious but spills and satiates thirst due to water content you can get out in electrolyze.

luizwritescode:
> Always assumed slimeballs are not juiceable and in ball form in first place because they are condensed and basically dry nutriments

Slimeballs are for slimes what meat is for humans so having nutriment makes sense, but a 'slime'ball not having slime is a bit weird. In practice all this PR really does is allow kitchen to make more slime-based food.

chromiumboy:
Sure, why not?

PR #33661, opened 2024-11-30
Makes kukri not use combat knife's storage sprite

SpaceRox1244:
## About the PR

Makes kukri not use the combat knife's grid-inv storage sprite because of an inheritance oversight.

## Media

kukri

## Requirements

**Changelog**

No CL

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/knife.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/knife.yml
@@ -143,6 +143,7 @@
         Slash: 15
   - type: Item
     sprite: Objects/Weapons/Melee/kukri_knife.rsi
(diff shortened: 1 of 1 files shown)


slarticodefast (approved):
Ideally this item would have its own storedSprite to fix the rotation, but this is better than using that of the combat knife.
Thanks for fixing it.

SpaceRox1244:
Yeah I agree its just that I dont want to sprite the kukri right now hehe

RedBookcase:
Glad i'm not the only one that noticed this. I swear to god though, the Kukri used to be a 1x1x1 sized item (like a pistol) and it had a curved storage sprite. Maybe i'm going crazy though.

chromiumboy:
Thanks for the fix

PR #33665, opened 2024-11-30
Zombies can see Initial Infected

psykana:
## About the PR
Zombies can now see initial infected.

## Why / Balance
Many times I've seen Initial Infected's sabotage plans get ruined by another II who's starting to bite people 5 minutes into the match, making the former fight both the living and the agressive turned.
This should improve gameplay as Initial Infected and zombies in general, as they will cooperate better now.

This is kind of controversial in terms of lore, but both options can be justified:
- On one hand, zombies are completely braindead and just attack all living.
- On the other hand, Initial Infected are Syndicate bioterrorists who have/keep the virus since the beginning on the match, and zombies can sense it somehow, just like they can sense each other.

## Technical details
Only prototype change.

## Media
image

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- add: Zombies can now see initial infected

### DIFF
--- a/Resources/Locale/en-US/zombies/zombie.ftl
+++ b/Resources/Locale/en-US/zombies/zombie.ftl
@@ -1,9 +1,9 @@
 zombie-transform = {CAPITALIZE(THE($target))} turned into a zombie!
-zombie-infection-greeting = You have become a zombie. Your goal is to seek out the living and to try to infect them.  Work together with the other zombies to overtake the station.
+zombie-infection-greeting = You have become a zombie. Your goal is to seek out the living and to try to infect them.  Work together with the other zombies and remaining initial infected to overtake the station.
 
 zombie-generic = zombie
 zombie-name-prefix = zombified {$baseName}
 zombie-role-desc =  A malevolent creature of the dead.
-zombie-role-rules = You are an antagonist. Search out the living and bite them in order to infect them and turn them into zombies. Work together with the other zombies to overtake the station.
+zombie-role-rules = You are an antagonist. Search out the living and bite them in order to infect them and turn them into zombies. Work together with the other zombies and remaining initial infected to overtake the station.
 
 zombie-permadeath = This time, you're dead for real.
--- a/Resources/Prototypes/StatusIcon/faction.yml
+++ b/Resources/Prototypes/StatusIcon/faction.yml
@@ -17,6 +17,7 @@
     components:
     - ShowAntagIcons
     - InitialInfected
+    - Zombie
   icon:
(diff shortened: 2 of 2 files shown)


slarticodefast (approved):
I assume it has been this way before because zombies are supposed to be mindless and attack everyone living in sight. But it practice I agree with you and I had similar situations happen before. When two initial infected try to coordinate to turn and attack at the same time losing the icon makes it hard to keep track of your former ally. And in uncoordinated cases it can ruin the round for the initial infected when they get randomly attacked by another zombie. Hiding among the living and striking at a good opportunity makes the roundflow more interesting in my opinion.

2013HORSEMEATSCANDAL:
I'd justify it by saying that the II are already infected, their symptoms just aren't noticeable to humans.

chromiumboy:
I agree with the change personally, but it probably should get approval from the admin team first

UbaserB:
>On the other hand, Initial Infected are Syndicate bioterrorists who have/keep the virus since the beginning on the match, and zombies can sense it somehow, just like they can sense each other.

Might be cool to let zombies also sense who is already infected and not be allowed to attack those people

thebadman4662:
> > Might be cool to let zombies also sense who is already infected and not be allowed to attack those people
> 
> Allowing zombies to know who's infected would be cool, but just straight up not allowing them to attack someone infected would be bad. Infections take quite awhile to actually do anything, which is good, since it gives survivors a chance to get a cure, so I don't think it should be changed. If I just straight up could not fight the person shooting me because they already got infected I would be pretty mad. There's also cases with you just getting mauled by a horde and you shouldn't be able to just run out alive after one of the bites trigger an infection and suddenly you're immune to zombies.

Pretty sure by infected they meant IIs antags, not crew with zombie DoT.

slarticodefast:
Hey, this was brought up for internal discussion and for rule clarity reasons we have one small change request before this can be merged:
Could you update the character info text for zombies so they know not to attack intial infected?
grafik
Something like "Work together with the other zombies and initial infected to overtake the station" should do.
See `zombie.ftl` for that translation string.

lzk228:
then maybe make zombies not to bite IIs?

slarticodefast (approved):
Thanks!

psykana:
Hi @slarticodefast, what's stopping this from getting merged?

lzk228:
> Hi @slarticodefast, what's stopping this from getting merged?

double approval requirement

slarticodefast:
The code has been approved by me and the admins gave the conceptual ok. Acoording to maintainer policy another maintainer will have to double check the PR before merging. I can try and annoy a few people :)

Issue #33386, opened 2024-11-18
Missing Inhand Sprites MegaIssue™


TiniestShark:
## Description
There's many items in the game currently lacking any form of Inhand sprites and just appear invisible when held. While not exactly a high priority, fixing and adding inhands to many of these will help with general immersion and visibility.

This is by no means entirely comprehensive (I am one human being) and if there's any I missed, please let me know. Some items were grouped together that could be solved with a generic inhand for their particular category. (I'm also not including items there are already open or already merged PRs for)

I'll update this as PRs are added, Food/Drinks/Plants will be their own issue at some point (Just assume its all of them)

**Item Groups:**

- [ ] [Bedsheets](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Objects/Misc/bedsheets.rsi)
- [ ] [Books (Bible & Necronomicon excluded)](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Objects/Misc/books.rsi)
- [ ] [Borg Pieces](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Objects/Specific/Robotics/cyborg_parts.rsi)
- [ ] [Borg Modules](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Objects/Misc/module.rsi)
- [ ] [CD Disks](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Objects/Misc/cd.rsi)
- [ ] [Computer Boards](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Objects/Misc/module.rsi)
- [ ] [Crew Medals](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Clothing/Neck/Medals)
- [ ] [Door Remotes](https://github.com/space-wizards/space-station-14/tree/31d5a668669cf14558e25c6607a446aac0927b5d/Resources/Textures/Objects/Devices/door_remote.rsi)
[...]

PR #33676, opened 2024-12-01
some robotics inhands

muburu:
## About the PR
adds inhands for the following robotical stuff

- all cyborg modules
- generic borg parts (does not include other chassis as borgimorph removed the ability to build them)
- proximity sensor

## Why / Balance
inhand good
resolves some of https://github.com/space-wizards/space-station-14/issues/33386

## Technical details
the modules use a greyscale base which is recolored in the YML using inhandvisuals

the science modules are no longer considered generic modules (new BaseBorgModuleScience) but are still usable by the generic chassis ~~sciborgs when~~

if needed i can make the science modules generic again but the YML for the colored modules is a bit beefy so it's done to save space and allow more science modules to be added in the future
## Media
**moduols**
modulehands
**borgiparts**
borgparthand

- i know some modules have differently colored parts on the end but i'm not making another layer for that as the YML is already beefy enough

## Requirements

**Changelog**
what's a changelog


muburu:
i also have positronic brain and mmi inhands completely finished in terms of spriting but i can't think of a way to have their different states shown in the inhand as well so i just left them out of this pr
active-inhand-left
searching-inhand-left
mmi-brain-inhand-left
mmi-inhand-left

Resources/Textures/Objects/Specific/Robotics/cyborg_parts.rsi/meta.json
```diff
@@ -11,20 +11,52 @@
       "name": "borg_l_arm"
     },
     {
-      "name": "borg_r_arm"
+        "name": "borg_r_arm"
```
beck-thompson, line 14: Index is wrong

beck-thompson (approved):

Tayrtahn (approved):

Issue #23110, opened 2023-12-28
deprecated commands in UIs


Tunguso4ka:
## Description
Some UI buttons have deprecated commands.

in Admin UI, Adminbus "Delete Singularities" linked to the non existing command "deleteewc"

in the same place "Load blueprints" kills debug client, i think that this is my dev env problem
`Unhandled exception. System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')`

in Sandbox UI "Link machines" linked to "signallink"

**Reproduction**
Click buttons

**Screenshots**
![зображення](https://github.com/space-wizards/space-station-14/assets/71643624/317b1898-7d48-4ae1-9d81-d6ac571cd53b)
![зображення](https://github.com/space-wizards/space-station-14/assets/71643624/91f0da1f-a61f-407c-b5db-67bb1c9303a0)

**Additional context**

PR #4323, opened 2021-07-21
machine linking refactor to ecs

PaulRitter:
what it says on the tin

did not test the rangecheck since imma go play payday now

PR #33684, opened 2024-12-02
Organize the Sandbox Panel window

SpaceManiac:
## About the PR
This window was a bit disorganized and had buttons that were misspelled or did nothing. Now it's less so.

## Technical details
- Remember the Sandbox Panel's position like other windows, instead of forcibly centering it on open
- Group buttons into logical sections
- Rename "Shows Spawns" to "Show Spawns"
- Remove "Link Machines" button that has done nothing for 3 years.
  - Related issue #23110
  - Related PR #4323
- Unset ToggleMode on Suicide button

## Media
| Before | After |
|-|-|
| image | image |

## Requirements

### DIFF
--- a/Content.Client/UserInterface/Systems/Sandbox/SandboxUIController.cs
+++ b/Content.Client/UserInterface/Systems/Sandbox/SandboxUIController.cs
@@ -1,4 +1,5 @@
-﻿using Content.Client.Administration.Managers;
+﻿using System.Numerics;
+using Content.Client.Administration.Managers;
 using Content.Client.Gameplay;
 using Content.Client.Markers;
 using Content.Client.Sandbox;
@@ -7,9 +8,7 @@ using Content.Client.UserInterface.Controls;
 using Content.Client.UserInterface.Systems.DecalPlacer;
 using Content.Client.UserInterface.Systems.Sandbox.Windows;
 using Content.Shared.Input;
-using Content.Shared.Silicons.StationAi;
 using JetBrains.Annotations;
-using Robust.Client.Console;
 using Robust.Client.Debugging;
 using Robust.Client.Graphics;
 using Robust.Client.Input;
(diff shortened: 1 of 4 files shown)


Content.Client/UserInterface/Systems/Sandbox/SandboxUIController.cs
```diff
@@ -164,7 +166,7 @@ public void OnStateExited(GameplayState state)
     {
         if (_window != null)
         {
-            _window.Dispose();
+            _window.Close();
```
slarticodefast, line 169: I tested everything and it works as intended but this is the only part I got no clue about if it is correct, and neither has sloth. I gave PJB a ping to figure this out.
The rest of the changes look good to me.

PJB3005: It's a bunch of default buttons, this is fine.

Content.Client/UserInterface/Systems/Sandbox/SandboxUIController.cs
```diff
@@ -109,9 +108,13 @@ public void LoadButton()
 
     private void EnsureWindow()
     {
-        if(_window is { Disposed: false })
+        if (_window is { Disposed: false })
             return;
         _window = UIManager.CreateWindow<SandboxWindow>();
+        // Pre-center the window without forcing it to the center every time.
+        _window.OpenCentered();
+        _window.Close();
```
PJB3005, line 116: Hm, definitely feels like a proper native way to do this would be nice. This is very much a hack.

PJB3005 (approved):

slarticodefast (approved):

Content.Client/UserInterface/Systems/Sandbox/SandboxUIController.cs
SpaceManiac: There's `SetAnchoredPreset` but it attaches the top-left of the window to the anchor point instead of e.g. center-to-center or bottom-to-bottom.

PJB3005: I was thinking more like a "Open centered or last position if this isn't the first time" function in the engine, but it's not the biggest problem in the world.

PR #33691, opened 2024-12-02
Fix sinks and toilets not draining

Partmedia:
## About the PR
Per the system comments, AutoDrain is designed to automatically move puddles into the drain (like a floor drain). Drains without AutoDrain are still supposed to gradually empty the buffer, but not remove puddles (like sinks and toilets).

However, a logic error in the original implementation causes drains with AutoDrain set to false to simply not work. Hence sinks never emptied.

## Why / Balance
Fixes sinks draining.

## Media
https://github.com/user-attachments/assets/972a188a-a7ab-4ce0-8464-7716cae2fc23

## Requirements

## Breaking changes
None

**Changelog**
:cl: notafet
- fix: Fix sinks and toilets not draining.

### DIFF
--- a/Content.Server/Fluids/EntitySystems/DrainSystem.cs
+++ b/Content.Server/Fluids/EntitySystems/DrainSystem.cs
@@ -134,13 +134,6 @@ public sealed class DrainSystem : SharedDrainSystem
             }
             drain.Accumulator -= drain.DrainFrequency;
 
-            // Disable ambient sound from emptying manually
-            if (!drain.AutoDrain)
-            {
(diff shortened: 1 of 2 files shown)


slarticodefast (changes requested):
Tested and works as intended. Thanks for the fix!
One small request: Could you update the documentation for the `AutoDrain` datafield in the `DrainComponent` accordingly? It seems like thi behaviour was originally intended, but your version makes more sense to me.
grafik

Feel free to self merge after that.

slarticodefast (approved):

Issue #33279, opened 2024-11-12
Project Zero Warnings


PJB3005:
## Description
The game produces a ton of compiler warnings when built right now:

![image](https://github.com/user-attachments/assets/05947e30-d0d9-49d8-b0f1-21d7bffd7b2f)

This has gradually piled up over years of development to RT and SS14, primarily due to deprecation of old APIs. 

Incrementally cleaning up the code to remove these warnings is something any contributor can do in small batches. Contributions appreciated!

For MOTIVATIONAL REASONS, if want to track our progress, we have a [public Grafana dashboard](https://grafana.spacestation14.com/public-dashboards/5b5d6f6e091745b69aedd15fa8fd301a) that tracks the repo's total warning count over time!

thetolbean:
Can we assume then that maintainers are going to be more focused on preventing PR's from introducing more warnings?

PJB3005:
> Can we assume then that maintainers are going to be more focused on preventing PR's from introducing more warnings?

My current stance is that PRs adding warnings isn't big enough of a deal to warrant the effort to implement an automatic CI enforcement system like that. Best to just fix all the warnings right now.

benev0:
I have a branch open for updating the static logger to sawmill. It will probably be in later this week

Tayrtahn:
For those who are so inclined, you can run the following shell command to produce a list in your home directory of every warning, sorted by type (including grouping uses of the same obsolete method together).

`dotnet clean; dotnet build | grep warning | sed -r 's/(.*): (warning .*):(.*)/\2 :\3 : \1/g' | sort > ~/warnings.log`

This can be helpful to find a method that needs cleanup - once you figure out how to update it in one instance, it's generally pretty easy to fix other instances.

Tayrtahn:
Here's another one:
`dotnet clean; dotnet build | grep warning | sed -r 's/(.*): (warning .*):(.*)/\1/g' | sort | uniq | sed -r 's/(.*)\(.*/\1/g' | uniq -c | sort -r`
This will print a list of all files that have warnings in them, including the number of warnings in each file, sorted from most to least, and removing any duplicates that might occur due to the way the different assemblies are compiled. I'm not going to pretend that this is well written, but it does seem to work!

kin98:
[...]

PR #33705, opened 2024-12-03
update static logger to sawmill

benev0:
for #33279

### DIFF
--- a/Content.Server/Preferences/Managers/ServerPreferencesManager.cs
+++ b/Content.Server/Preferences/Managers/ServerPreferencesManager.cs
@@ -42,7 +42,7 @@ namespace Content.Server.Preferences.Managers
             _netManager.RegisterNetMessage<MsgSelectCharacter>(HandleSelectCharacterMessage);
             _netManager.RegisterNetMessage<MsgUpdateCharacter>(HandleUpdateCharacterMessage);
             _netManager.RegisterNetMessage<MsgDeleteCharacter>(HandleDeleteCharacterMessage);
-            _sawmill = _logManager.GetSawmill("SawmillName");
+            _sawmill = _logManager.GetSawmill(SawmillName);
         }
 
         private async void HandleSelectCharacterMessage(MsgSelectCharacter message)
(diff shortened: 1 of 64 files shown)


CrafterKolyan:
I am not an expert on the whole infrastructure stuff and not sure how all of this `IOCManager` works to be honest and neither the maintainer of the project. But shouldn't all of these loggers be:
```
private static readonly
```
so that we wouldn't need to create a new logger on every object instantiation?

I guess maintainers may want to comment if it makes sense to make all of them `private static readonly`

CrafterKolyan:
Ah, I checked the implementation of `LogManager` and saw that it actually caches loggers by name which makes sense to be honest, so strictly speaking we are not allocating a new object, though I guess every object that contains logging will have extra 3 fields. I guess we can definitely move `string` one to be `static` to optimize at least that. Not sure about the other ones, as I guess we pay the cost of 16 bytes per logged object in memory to make the logging testable?

In the end I guess none of this may matter at this point as I saw other places where `static` optimization may be made but I guess `IOCManager` complicates the actual ability to use `static` variables.

Content.Server/Preferences/Managers/ServerPreferencesManager.cs
```diff
@@ -41,7 +42,7 @@ public void Init()
             _netManager.RegisterNetMessage<MsgSelectCharacter>(HandleSelectCharacterMessage);
             _netManager.RegisterNetMessage<MsgUpdateCharacter>(HandleUpdateCharacterMessage);
             _netManager.RegisterNetMessage<MsgDeleteCharacter>(HandleDeleteCharacterMessage);
-            _sawmill = _log.GetSawmill("prefs");
+            _sawmill = _logManager.GetSawmill("SawmillName");
```
CrafterKolyan, line 45: I think this should be a variable `SawmillName` rather than a string `"SawmillName"`

benev0:
forgot to post this:
Due to nearly identical code behaving differently within different regions, I have used multiple patterns. This is roughly where I targeted the patterns.
- static + (log on critical fail only)
- full injection (managers)
- lazy singleton (multiple log calls on the same object)

There was a lot of trouble with null deref segfaults due to these patterns being

Content.Server/Preferences/Managers/ServerPreferencesManager.cs
benev0: yes

PR #33706, opened 2024-12-03
T1 Firearm Manufacturing

OneZerooo0:
## About the PR

Adds a new T1 technology to the Arsenal science discipline, which costs 8,000 points and enables fabrication of MK-58 and Inspector security sidearms to the security techfab, or to an emagged protolathe. MK-58 costs 15 steel, and five plastic, while the Inspector costs 10 steel, 5 plastic, 5 wood.

## Why / Balance

Currently, MK-58s are only found through roundstart spawns or cargo's pistols crate, and the detective's Inspector cannot be recovered outside of roundstart. This technology will allow recovery of Detective's signature weapon, an alternative sidearm to the MK-58 later into the round, and an additional way to acquire more MK-58s outside of cargo. (Of course, this requires materials which can be more easily gained from cargo than a complete pistol crate, or from salvaging maints, and the willingness of Science to actually _research_ the technology.) The point is a new source for Security's faithful sidearm and a way to get more revolvers.

## Technical details

Slightly changed MK-58's description to match the fact that science can make them now
Added recipes for mk-58 and inspector
Added mk-58 and inspector recipes to security techfab and emagged protolathe 
Updated technologies.ftl with name for tech
Added the actual technology into arsenal.yml

## Media

image
image
image

## Requirements

## Breaking changes

Shouldn't be any, just YML changes

**Changelog**

:cl: Onezero0
- add: Added T1 Firearm Manufacturing, which enables fabrication of MK-58 and Inspectors for Security!

### DIFF
--- a/Resources/Locale/en-US/research/technologies.ftl
+++ b/Resources/Locale/en-US/research/technologies.ftl
@@ -37,7 +37,7 @@ research-technology-portable-microfusion-weaponry = Portable Microfusion Weaponr
 research-technology-experimental-battery-ammo = Experimental Battery Ammo
 research-technology-basic-shuttle-armament = Shuttle basic armament
 research-technology-advanced-shuttle-weapon = Advanced shuttle weapons
-research-technology-firearms-manufacturing = Firearms Manufacturing
+research-technology-firearm-manufacturing = Firearm Manufacturing
 
 research-technology-basic-robotics = Basic Robotics
 research-technology-basic-anomalous-research = Basic Anomalous Research
--- a/Resources/Locale/en-US/research/technologies.ftl
+++ b/Resources/Locale/en-US/research/technologies.ftl
@@ -37,6 +37,7 @@ research-technology-portable-microfusion-weaponry = Portable Microfusion Weaponr
 research-technology-experimental-battery-ammo = Experimental Battery Ammo
 research-technology-basic-shuttle-armament = Shuttle basic armament
 research-technology-advanced-shuttle-weapon = Advanced shuttle weapons
+research-technology-firearms-manufacturing = Firearms Manufacturing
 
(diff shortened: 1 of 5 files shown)


kosticia:
In my opinion now we have pretty interesting choice between ways of refill of weapons. Cargo - for order letal weapons that uses ammo. R&D - for creating more situational but not less useful weapons such as lasers, grenades and shields. Adding MK in secfab ruins this balance.

OneZerooo0:
> In my opinion now we have pretty interesting choice between ways of refill of weapons. Cargo - for order letal weapons that uses ammo. R&D - for creating more situational but not less useful weapons such as lasers, grenades and shields. Adding MK in secfab ruins this balance.

Not necessarily, because both roads lead to cargo. The way Security actually makes science’s weapons is through materials (gained from cargo one way or another during SOP, save for scrapping maints, which is much less effective.) Thus letting security make mk-58s through fabrication just lets science help sec get new sidearms more easily by asking cargo for materials, rather than asking for a specific weapons box repeatedly for (sometimes) half a shift. 
(Don’t forget that cargo can also purchase the laser crate, which draws a nice parallel with this tech)

baddiePRO:
personally I think the inspector should be roundstart only, and let det keep their unique gun

OneZerooo0:
Ah, true, this seems to be heading in the Svalinn direction - I will leave the PR open but I doubt it will be merged, considering these arguments. Thanks for the feedback!

ps3moira:
I think shifting crews guns to be MORE kinetic instead of lasers makes it harder for syndicate and nukies to counter and creates a balance issue

MilonPL:
After internal discussion we've decided to close this PR, as we would like ballistic weapons to remain obtainable only through cargo, rather than security being able to manufacture them.

PR #33710, opened 2024-12-04
Adds Store on Collide and Wand of the Locker

keronshb:
## About the PR

Adds Wand of the Locker for Wizards! Which is a wand that shoots out cursed lockers that will store anyone it _collides_ with
Which requires the Store on Collide Component! This component will store entities into storage when collided.

## Why / Balance

Wand of Locker is part of the wizard arsenal and it's a fun goofy weapon. More of an annoyance than something super dangerous. Could allow for some fun combinations in the future!

Store on Collide will be nice not only for the locker projectiles but it can open possibilities for things like ore boxes (dragging boxes over ore for easy storage).

## Technical details

Store on Collide will store any entities that collides with it via Collide Event.
Options to lock and to disable were added for further control.

## Media

https://github.com/user-attachments/assets/6c644221-5a52-4dba-ac8e-9afc5111566a

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Adds Wand of the Locker - a wand that shoots out cursed lockers, that will store anyone they touch!


Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
```diff
@@ -344,6 +344,14 @@ public bool TryCloseStorage(EntityUid target)
         return true;
     }
 
+    public bool IsOpen(EntityUid target, SharedEntityStorageComponent? component = null)
```
slarticodefast, line 347: Can't you just do a `TryComp` and check `component.Open` where you need it?

Resources/Prototypes/Access/wizard.yml
```diff
@@ -0,0 +1,3 @@
+﻿- type: accessLevel
```
slarticodefast, line 1: Add this to the `AccessConfiguratorUniversal` so admins and mappers can set it.
An ID card for the wizard might be useful as well, but that does not has to be part of this PR.

Content.Shared/Storage/EntitySystems/StoreOnCollideSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        TryLockStorage(ent);
+    }
+
+    private void TryLockStorage(Entity<StoreOnCollideComponent> ent)
+    {
+        var storageEnt = ent.Owner;
+        var comp = ent.Comp;
+
+        if (comp.LockOnCollide && !_lock.IsLocked(storageEnt))
```
slarticodefast, line 61: It might make more sense to move the `LockOnCollide` check into `OnCollide` and call `TryLockStorage` from there so that the functions are more independent and reusable instead of one big chain.

Content.Shared/Storage/EntitySystems/StoreOnCollideSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    private void TryStoreTarget(Entity<StoreOnCollideComponent> ent, EntityUid target)
+    {
+        var storageEnt = ent.Owner;
+        var comp = ent.Comp;
+
+        if (_netMan.IsClient || _gameTiming.ApplyingState)
+            return;
+
+        if (ent.Comp.Disabled || storageEnt == target || Transform(target).Anchored || _storage.IsOpen(storageEnt) || _whitelist.IsWhitelistFail(comp.Whitelist, target))
```
slarticodefast, line 48: Wouldn't it be better for the locker to be open when shot and closing it when stopping? It is a little weird that is has to be closed to be able to store something, especially if this component would be reused elsewhere.

slarticodefast (commented):
A few things I noticed while testing:
- You might want to layer the paper layer below the magic lock effect:
grafik
- Welding the locker has the layers in the wrong order as well:
grafik
- The locker has no collisions with players once it stops. You can simply walk through it.
grafik
- The locker has no mass (drag it around and i will slide). Not sure if this is intended, might make it easier to drag around this way though.
- Should

lzk228:
is there any point in the cl if the players can't get the wand yet?

keronshb:
> is there any point in the cl if the players can't get the wand yet?

The other alternative is a huge changelog when wizard is out, which I do not want to do.

Content.Shared/Storage/EntitySystems/StoreOnCollideSystem.cs
keronshb: The issue with that is people being able to leave the locker immediately while it's in transit. It needs to stay closed and arguably locked while flying or else it's kinda useless.

Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
keronshb: The issue is that in shared I can't do a TryComp for SharedStorageComp, I get an error for it. But when it gets resolved like this, it works fine.

keronshb:
> A few things I noticed while testing:
> 
>     * You might want to layer the paper layer below the magic lock effect:
>       grafik
> 
>     * Welding the locker has the layers in the wrong order as well:
>       grafik
> 
>     * The locker has no collisions with players once it stops. You can simply walk through it.
>       grafik
> 
>     * The locker has no mass (drag it around and i will slide). Not sure if this is intended, might make it easier to drag around this way though.

That's due to the shooter being excluded from colliding with the locker, since it's a projectile. If it collided with the shooter when shot, you can see the issue there.

Content.Shared/Storage/EntitySystems/StoreOnCollideSystem.cs
slarticodefast: Good point!

slarticodefast (approved):
Looks good to me. Let's get this into the next release and see how much chaos the admins will cause with it :)

Issue #31734, opened 2024-09-01
airlock wire ui is kinda broken


lzk228:
![Screenshot_20240902_010529](https://github.com/user-attachments/assets/8aeea8ab-40a0-4799-b0a5-a55532ad4e78)

slarticodefast:
Happening to me as well. Same with vending machines.
I assume the addtion of the new AIA light for the AI access did not move the wire graphics downwards, so it is now overlapping.
![grafik](https://github.com/user-attachments/assets/cbf289a6-1c50-49f5-aec6-4be05e650427)

PR #33714, opened 2024-12-04
Fix layout on wires UI

PJB3005:
Layout would break for machines with >6 lights because the column count was hardcoded. Uncap the UI width and fix the rows count instead.

Lights with less than 4 characters of text weren't aligned right, now they are.

Before:
image

After:
image

:cl:
- fix: Fixed broken layout on some wire hacking UIs.


slarticodefast:
Fixes #31734

PR #33715, opened 2024-12-04
Added several variables to make ClumsyComponent more modular for developers.

widgetbeck:
## About the PR

Added nine new variables to ClumsyComponent and its system to allow developers to specify that certain clumsy events should not be applied to a given entity, and to allow developers to have custom per-entity "failed" popup message.

## Why / Balance

ClumsyComponent does a lot of different things, and sometimes you only want it to do a few of those things. For example, I was developing a species that is incapable of using guns on my server - I added these variables so that I could disable the hypospray, defibrilator, and table bonk events on those entities. 
In short, these changes allow developers to find more creative uses for this component, without the hassle that would be separating each event into its own system. 

## Technical details

Very simple. Added a check to each event that, if that event's boolean is false, skips the event, and changed the hardcoded strings in ClumsySystem to public LocIds that are read from ClumsyComponent.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/Clumsy/ClumsyComponent.cs
+++ b/Content.Shared/Clumsy/ClumsyComponent.cs
@@ -56,7 +56,7 @@ public sealed partial class ClumsyComponent : Component
     /// <summary>
     ///     Noise to play after failing to shoot a gun. Boom!
     /// </summary>
-    [DataField, AutoNetworkedField]
+    [DataField]
     public SoundSpecifier GunShootFailSound = new SoundPathSpecifier("/Audio/Weapons/Guns/Gunshots/bang.ogg");
 
     /// <summary>
--- a/Content.Shared/Clumsy/ClumsyComponent.cs
+++ b/Content.Shared/Clumsy/ClumsyComponent.cs
@@ -56,48 +56,48 @@ public sealed partial class ClumsyComponent : Component
     /// <summary>
     ///     Noise to play after failing to shoot a gun. Boom!
     /// </summary>
-    [DataField]
+    [DataField, AutoNetworkedField]
     public SoundSpecifier GunShootFailSound = new SoundPathSpecifier("/Audio/Weapons/Guns/Gunshots/bang.ogg");
-	
-	/// <summary>
-	///		Whether or not to apply Clumsy to hyposprays.
-	/// </summary>
-	[DataField, AutoNetworkedField]
-	public bool ClumsyHypo = true;
-	
-	/// <summary>
-	///		Whether or not to apply Clumsy to defibs.
-	/// </summary>
-	[DataField, AutoNetworkedField]
-	public bool ClumsyDefib = true;
-	
-	/// <summary>
-	///		Whether or not to apply Clumsy to guns.
-	/// </summary>
-	[DataField, AutoNetworkedField]
-	public bool ClumsyGuns = true;
-	
-	/// <summary>
-	///		Whether or not to apply Clumsy to vaulting.
-	/// </summary>
-	[DataField, AutoNetworkedField]
-	public bool ClumsyVaulting = true;
-	
-	/// <summary>
-	///		Lets you define a new "failed" message for each event.
-	/// </summary>
-	[DataField]
-	public LocId HypoFailedMessage = "hypospray-component-inject-self-clumsy-message";
-	
-	[DataField]
-	public LocId GunFailedMessage = "gun-clumsy";
-	
-	[DataField]
-	public LocId VaulingFailedMessageSelf = "bonkable-success-message-user";
-	
-	[DataField]
-	public LocId VaulingFailedMessageOthers = "bonkable-success-message-others";
-	
-	[DataField]
-	public LocId VaulingFailedMessageForced = "forced-bonkable-success-message";
+
+    /// <summary>
+    ///      Whether or not to apply Clumsy to hyposprays.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool ClumsyHypo = true;
+
+    /// <summary>
+    ///      Whether or not to apply Clumsy to defibs.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool ClumsyDefib = true;
+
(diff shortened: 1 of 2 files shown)


Content.Shared/Clumsy/ClumsySystem.cs
```diff
@@ -40,6 +40,10 @@ private void BeforeHyposprayEvent(Entity<ClumsyComponent> ent, ref SelfBeforeHyp
         // Clumsy people sometimes inject themselves! Apparently syringes are clumsy proof...
         if (!_random.Prob(ent.Comp.ClumsyDefaultCheck))
             return;
+		
+		// imp. checks if ClumsyHypo is false, if so, skips.
+		if (!ent.Comp.ClumsyHypo)
```
slarticodefast, line 45: Put this above the randomization so we don't have to do that in case it's disabled. Same for the other new bools.

Content.Shared/Clumsy/ClumsyComponent.cs
```diff
@@ -58,4 +58,46 @@ public sealed partial class ClumsyComponent : Component
...
+	///		Imp. Whether or not to apply Clumsy to vaulting.
+	/// </summary>
+	[DataField]
+	public bool ClumsyVaulting = true;
+	
+	/// <summary>
+	///		Imp. Lets you define a new "failed" message for each event.
+	/// </summary>
+	[DataField]
+	public string HypoFailedMessage = "hypospray-component-inject-self-clumsy-message";
```
slarticodefast, line 90: Use a `LocId` instead of a string. Same below.

Content.Shared/Clumsy/ClumsyComponent.cs
```diff
@@ -58,4 +58,46 @@ public sealed partial class ClumsyComponent : Component
     /// </summary>
     [DataField]
     public SoundSpecifier GunShootFailSound = new SoundPathSpecifier("/Audio/Weapons/Guns/Gunshots/bang.ogg");
+	
+	/// <summary>
+	///		Imp. Whether or not to apply Clumsy to hyposprays.
```
slarticodefast, line 63: What does the "Imp." mean?

slarticodefast (changes requested):

Content.Shared/Clumsy/ClumsyComponent.cs
widgetbeck: The name of our private server. Had that in there to mark changes that are exclusive to our server, so that in the event of a future merge conflict it's easier to see what changes are ours. I can remove those from the upstream PR if you want.

Content.Shared/Clumsy/ClumsySystem.cs
widgetbeck: done.

Content.Shared/Clumsy/ClumsyComponent.cs
widgetbeck: I removed it, just in case.

Content.Shared/Clumsy/ClumsyComponent.cs
```diff
@@ -58,4 +58,46 @@ public sealed partial class ClumsyComponent : Component
     /// </summary>
     [DataField]
     public SoundSpecifier GunShootFailSound = new SoundPathSpecifier("/Audio/Weapons/Guns/Gunshots/bang.ogg");
+	
+	/// <summary>
+	///		Whether or not to apply Clumsy to hyposprays.
+	/// </summary>
+	[DataField]
```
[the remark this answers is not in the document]
widgetbeck: Done, no problem.

slarticodefast (changes requested):
I noticed another thing, please use spaces instead of tabs everywhere:
grafik
grafik
I reccomend using an IDE instead of a text editor. That way you will be corrected if you do things conflicting with our code conventions defined in the .editorconfig file.

widgetbeck:
Ahh, I see. I'll fix that. Out of curiosity, is there a reason that this matters functionally? I've been using tabs on all my custom content for our server.

slarticodefast:
Consistency reasons mostly, functionally all whitespace should be the same afaik. If you would be using an IDE the tabs would get automatically converted into spaces, so it wouldn't even be possible to mix both. But if you do it in a texteditor and open it in an IDE afterwards the IDE will scream in pain and throw yellow squiggly lines at you :)

Content.Shared/Clumsy/ClumsyComponent.cs
```diff
@@ -56,6 +56,48 @@ public sealed partial class ClumsyComponent : Component
     /// <summary>
     ///     Noise to play after failing to shoot a gun. Boom!
     /// </summary>
-    [DataField]
+    [DataField, AutoNetworkedField]
```
slarticodefast, line 59:
```suggestion
    [DataField]
```
This one doesn't really need to be networked, the other sounds aren't either. So keep it as it was.

beck-thompson:
us becks like the clumsy system I guess 😆!

chromiumboy:
Honestly, I feel like this approach of having a separate bool for every possible clumsy event is something that could get out of hand. If someone wants to add clumsiness to a specific tool in the future, for example, that's another bool to add to the list

I'd prefer a system in which clumsy people have a list of clumsiness rules (i.e. prototypes), which get checked against when a potentially clumsy event occurs. If a match is found, the event occurs; any required data regarding the event (e.g., the message to display to the player) can just be stored in the rule prototype

slarticodefast:
> I'd prefer a system in which clumsy people have a list of clumsiness rules (i.e. prototypes), which get checked against when a potentially clumsy event occurs. If a match is found, the event occurs; any required data regarding the event (e.g., the message to display to the player) can just be stored in the rule prototype

I think that would probably make things unnecessarily complicated in this case. Another approach would be to split the component up into a separate one for each effect, so that you could use composition to select the effects you want. But that again is not really needed in upstream at the moment.

VasilisThePikachu (approved):

PR #33717, opened 2024-12-04
Predicted gas pumps

metalgearsloth:
I wanted to try out atmos and first thing I found.

The original code is oh so very cursed.

Changed it slightly and made it use a float spinbox too.

image

## Breaking changes
- comp-gas-pump-ui-needs-anchor = ui-needs-anchor
- namespace on the 2 atmos events moved to shared, AtmosDeviceDisabled + Enabled.
- namespace on GasPressurePumpComponent.

:cl:
- fix: Fix gas pumps being capped to atmos max pressure and not their own max pressure.
- tweak: Predict gas pump UI and you no longer need to be in details range to examine it.


Content.Shared/Atmos/Components/GasPressurePumpComponent.cs
```diff
@@ -0,0 +1,25 @@
...
+    public bool Enabled = true;
+
+    [DataField("inlet")]
+    public string InletName = "inlet";
+
+    [DataField("outlet")]
+    public string OutletName = "outlet";
+
+    [DataField, AutoNetworkedField]
+    public float TargetPressure = Atmospherics.OneAtmosphere;
```
Partmedia, line 18: Does this leak the target pressure setpoint over the network when the gas pump is out of examine range but still inside PVS? This doesn't seem to me like an important issue, but we should probably be aware of what variables we are sending to the client. Perhaps we don't need the out of range examine check anymore?

Content.Shared/Atmos/EntitySystems/SharedGasPressurePumpSystem.cs
```diff
@@ -0,0 +1,97 @@
...
+using Content.Shared.UserInterface;
+
+namespace Content.Shared.Atmos.EntitySystems;
+
+public abstract class SharedGasPressurePumpSystem : EntitySystem
+{
+    [Dependency] private   readonly ISharedAdminLogManager _adminLogger = default!;
+    [Dependency] protected readonly SharedAppearanceSystem Appearance = default!;
+    [Dependency] private   readonly SharedPowerReceiverSystem _receiver = default!;
+    [Dependency] protected readonly SharedUserInterfaceSystem UserInterfaceSystem = default!;
```
Partmedia, line 21: Huh, when did we start aligning variable modifiers like this?

metalgearsloth: I do it but I think it's just a me thing.

Partmedia (approved):
Looks good to me.

PR #33718, opened 2024-12-04
Added sort function to Chemmaster

DmitriTheDemon:
## About the PR
Sorted ChemMaster output window. I was asked to push this change upstream from a fork as others might desire it.

## Why / Balance
Quality of life change to allow finding reagents easier in ChemMaster

## Technical details
Added orderby to for each loop to sort contents

## Media
Space Wizard Sorted ChemMaster

## Requirements

## Breaking changes
N/A

**Changelog**

:cl:
- tweak: Sorted ChemMaster reagent list

### DIFF
--- a/Content.Client/Chemistry/UI/ChemMasterWindow.xaml.cs
+++ b/Content.Client/Chemistry/UI/ChemMasterWindow.xaml.cs
@@ -182,7 +182,7 @@ namespace Content.Client.Chemistry.UI
             };
             bufferHBox.AddChild(bufferVol);
 
-            foreach (var (reagent, quantity) in state.BufferReagents)
+            foreach (var (reagent, quantity) in state.BufferReagents.OrderBy(x => x.Reagent.Prototype))
             {
                 // Try to get the prototype for the given reagent. This gives us its name.
                 _prototypeManager.TryIndex(reagent.Prototype, out ReagentPrototype? proto);


dvir001:
Before anyone say "But Frontier AGPL" he made the PR, he is making it MIT.

Content.Client/Chemistry/UI/ChemMasterWindow.xaml.cs
```diff
@@ -182,7 +182,7 @@ private void UpdatePanelInfo(ChemMasterBoundUserInterfaceState state)
             };
             bufferHBox.AddChild(bufferVol);
 
-            foreach (var (reagent, quantity) in state.BufferReagents)
+            foreach (var (reagent, quantity) in state.BufferReagents.OrderBy(x => x.Reagent.Prototype))
```
0x6273, line 185: Shouldn't this sort by localized name?

7-62x54r:
Maybe make it an _option_ and not forced? I like to organize master by hand, it's not that hard...

0tito:
Id rather have an option to manually move the chems, making them automatically sorted will only make it more annoying

lexaSvarshik:
Make it an option, because sometimes it's easier to use it as is, without sorting.
So. it will be like "Press that button to see chem sorted", "Press another button and return to original order"

DmitriTheDemon:
It would seem this change is not desired upstream feel free to implement at a later point if that opinion changes.

Issue #33720, opened 2024-12-04
"Vox Boxes" should be removed from all maps immediately


PJB3005:
These have outright racist connotations. If I had known that some maps are as bad as they are I would have flipped shit about this months ago.

Do not peanut post on this issue, I am going to be zero tolerance on any punishments handed out from this fiasco.

Edit: people are finding this thread without being aware of the surrounding context. See here: https://discord.com/channels/310555209753690112/900426319433728030/1313935602252513312

Nox38:
where are voxes supposed to eat?

PJB3005:
> where are voxes supposed to eat?

We are discussing possible alternative manners to deal with this.

Ghost-Prince:
Do you mean that the nitrogen only rooms, to be moved to more communal rooms (nearer to kitchens and bars) so that they can have their mask's down an eat without getting poisoned by oxygen in regular atmosphere?

PJB3005:
No, remove them entirely

> Do you mean that the nitrogen only rooms, to be moved to more communal rooms (nearer to kitchens and bars) so that they can have their mask's down an eat without getting poisoned by oxygen in regular atmosphere?

Warpzoned:
> No, remove them entirely.

Both conceptually and thematically? Even though I partially understand a re-branding due to their less-than-adequate connotations, why remove N2 rooms as a whole? They're a great accessibility feature instead of having to micro-manage your tank and food.

PJB3005:
> > No, remove them entirely.
> 
> Both conceptually and thematically? Even though I partially understand a re-branding due to their less-than-adequate connotations, why remove N2 rooms as a whole? They're a great accessibility feature instead of having to micro-manage your tank and food.

Yes. Solutions that avoid the micro-management you described are better.

Djungelskog2:
Not in the current conversation occuring on discord but *please* go down the route of having box masks let people eat through them. Clown and mime vox already have this ability and it gives vox an extremely niche "upside" I'm their custom gear. It would also just be way simpler and more fun for the players.

Golinth:
I'm going to post what I already have in the discord, as it was largely ignored. 

[...]

PR #33722, opened 2024-12-04
Give vox health regen for poison below 20 damage

PJB3005:
This enables Vox to go about 30 seconds in oxygen and have the damage naturally go away. It's also just a nice species plus for them.

This is done in context of https://github.com/space-wizards/space-station-14/issues/33720

:cl:
- tweak: Vox now have natural poison regeneration, allowing them to go for ~30 seconds in oxygen and heal the damage away.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/vox.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/vox.yml
@@ -32,6 +32,18 @@
   - type: Damageable
     damageContainer: Biological
     damageModifierSet: Vox
+  - type: PassiveDamage
+    # Augment normal health regen to be able to tank some Poison damage
+    # This allows Vox to take their mask off temporarily to eat something without needing a trip to medbay afterwards.
+    allowedStates:
+    - Alive
+    damageCap: 20
+    damage:
+      types:
+        Heat: -0.07
+        Poison: -0.2
+      groups:
+        Brute: -0.07
   - type: DamageVisuals
     damageOverlayGroups:
       Brute:


Errant-4 (approved):
Approximately 36s of breathing standard airmix before the damage becomes "irreversible"
Health recovers in up to 1m40s

Total time to crit went from ~~2m~~ 2m12s to 2m20s, that seems negligible.

At first glance this does not seem like it would upset anything else. For a hotfix, I would say let's go. Possibly it can even stay until vox gets an actual design review?

slarticodefast (approved):
Not sure if this needs to be a hotfix considering the next release is right around the corner, but in case you want to hotfix it, here is the third approval.

Errant-4:
Oh I did not mean "literal" hotfix

Errant-4:
I will merge this in about 12 hours if there aren't any compelling objections, and then we can discuss it during the regular staging review on sunday, before it gets to stable

Djungelskog2:
I really think the better solution is to just give vox masks that they can eat/breathe through. Eating and drinking shouldnt be a challenge that vox have to face and its just a chore. Giving vox NOTHING to combat this issue can arguably feel MORE spacist, just like how we had those crappy nitrogen lockers with a paper sign taped on. myself and other admins have already expressed our thoughts in the discord thread and I believe its not doing anything to solve the current issues of implications of specism, eating/drinking being a chore to play with, and vox already gunning for the theatre masks roundstart because they have that functionality (and no, the solution isnt to nerf the theatre masks).

PJB3005:
I do not necessarily disagree that allowing vox to eat through masks is a bad solution. I just think that no matter what we do this PR is nice to have regardless and doesn't have much future risk associated with it.

IProduceWidgets:
Made #33731 as a potential alternative

PJB3005:
> Why not make them heal all poison damage passively as their gimmick, they're basically vultures that exhale ammonia so it wouldn't be too farfetched imo

Would've been more than a 10 line YAML change.

Errant-4:
Additionally, that would be way more than just a limited QOL feature that mostly just addresses a particular species-specific interaction issue, and should need to be backed up by an actual approved design doc

Errant-4:
For the record, I have taken more accurate, repeated measurements of the crit time on live and with the PR. My new numbers are now validated to be accurate to within* 2 seconds 
(there is a delay of variable length before the first oxygen breath, depending on when in the 4-second breath-cycle the mask is taken off)

The difference in time to crit from full health is just 7.7 seconds. Small enough to be effectively irrelevant

PR #33731, opened 2024-12-05
Coopt vox box for universal healthcare

IProduceWidgets:
## About the PR

Vox now get their own special survival boxes that contain charcoal pills to help them alleviate accidental poisoning and encourage interacting with the chemistry department. Also they no longer receive an extra respirator mask.

## Why / Balance

DSzFWYDVMAA7Ay8

## Media

image
image
(and similarly for the different roles / departments etc...)

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Vox now have charcoal pills in their vox box survival kit.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Boxes/emergency.yml
+++ b/Resources/Prototypes/Catalog/Fills/Boxes/emergency.yml
@@ -1,3 +1,5 @@
+# Standrad
+
 - type: entity
   parent: BoxCardboard
   id: BoxSurvival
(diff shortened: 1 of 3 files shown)


MissKay1994:
You will need to make sure if this is accepted that this works with loneops, ninjas, and ghost role ships

IProduceWidgets:
It does sans the ghost role ships, they actually don't get any extra survival gear at the moment. Oopsie (but needs to be fixed in another PR, doesn't make sense here)

Errant-4:
I dont know if it's just me, but I would much rather have a spare mask than a flare

lzk228:
so, removing vox boxes because they are racist, but giving medicine to the specific race instead?
adding roundstart medicine feels too powergamey for me

deltanedas:
roundstart medicine called epinephrine and tranexamic acid:

SlimmSlamm:
> so, removing vox boxes because they are racist, but giving medicine to the specific race instead? adding roundstart medicine feels too powergamey for me

Power gaming isn't a thing on SS14

Errant-4:
I think the code is fine, though I still strongly think that it should be the flare that gets replaced. Having a backup mask might _save your life_ one day, whereas having a flare is... perhaps somewhat useful, occasionally, but I mean you can always just use your PDA light

IProduceWidgets:
I'm not too attached to it, but I decided to leave the flare because its got other (admittedly minor) uses as part of the standard kit like lighting fires and dealing with long or unexpected power outages.

thebadman4662:
Charcoal is potent liquid/powder bloodstream specialized medicine that also costs good bit of carbon to make.

Dylo pills would probably be better if they are to get any at all.

Errant-4:
Thank you for the contribution, but we have decided that with vox poison regen having worked out pretty well, this change is not necessary. Particularly since it would require removing something else from the survival box. 
If crewmembers want charcoal pills as a convenience item, they can ask chemistry.

PR #33742, opened 2024-12-06
SpiderSystem refactor and rename 

iaada:
## About the PR

Refactor of the component and systems that give spiders the ability to spawn webs, and renamed them to a less generic name. Changes include:

- Webs no longer spawn on webs. (bugfix)
- Webs no longer spawn in space.
- Play a sound when webs spawn.
- More yaml control in where webs spawn (still somewhat limited).

## Why / Balance

The only balance change is that spiderwebs no longer spawn on other spiderwebs. This is a bugfix though, as the code to do this already existed and just didn't work.

## Technical details

`SpiderSystem`, `SharedSpiderSystem` and `SpiderComponent` renamed to `WebPlacerComponent`, `SharedWebPlacerSystem`, and `WebPlacerSystem`.
`SpiderWebObjectComponent` and `IgnoreSpiderWebComponent` deleted. Both of these existed to be found by a whitelist in yaml, and have been replaced with tags.
`SpiderWebVisuals` enumerator deleted. This was used alongside `SpiderWebObjectComponent` to give webs a random appearance, and has been replaced with `RandomSpriteComponent` and a generic enum.
Minor yaml changes. Changed the renamed component, replaced the deleted components. Clown spiders have a honk horn sound.

## Media

None.

## Requirements

## Breaking changes

`SpiderComponent` renamed to `WebPlacerComponent`.
`SharedSpiderSystem` renamed to `SharedWebPlacerSystem`.
`SpiderSystem` renamed to `WebPlacerSystem`.
Namespace `Content.Shared.Spider` changed to `Content.Shared.WebPlacer`.
Namespace `Content.Server.Spider` changed to `Content.Server.WebPlacer`.
`SpiderWebObjectComponent` removed. Replaced with `Tag: SpiderWeb`.
`IgnoreSpiderWebComponent` removed. Replaced with `Tag: IgnoreSpiderWeb`.
`SpiderWebVisuals` enumerator removed.

**Changelog**

:cl: aada
- fix: Spiderwebs no longer spawn over other spider webs.
- tweak: Spiderwebs no longer spawn in space.

### DIFF
--- a/Content.Server/WebPlacer/WebPlacerSystem.cs
+++ b/Content.Server/WebPlacer/WebPlacerSystem.cs
@@ -1,4 +1,3 @@
-using System.Linq;
 using Content.Server.Popups;
 using Content.Shared.Maps;
 using Content.Shared.WebPlacer;
--- a/Content.Server/WebPlacer/WebPlacerSystem.cs
+++ b/Content.Server/WebPlacer/WebPlacerSystem.cs
@@ -1,7 +1,7 @@
 using System.Linq;
 using Content.Server.Popups;
 using Content.Shared.Maps;
-using Content.Shared.Spider;
+using Content.Shared.WebPlacer;
 using Content.Shared.Whitelist;
 using Robust.Server.GameObjects;
 using Robust.Shared.Audio;
@@ -9,12 +9,12 @@ using Robust.Shared.Audio.Systems;
 using Robust.Shared.Map;
 using Robust.Shared.Map.Components;
 
-namespace Content.Server.Spider;
+namespace Content.Server.WebPlacer;
 
 /// <summary>
-/// Spawns entities (probably webs) around the component owner. Action handled by <see cref="SharedSpiderSystem"/>.
+/// Spawns entities (probably webs) around the component owner. Action handled by <see cref="SharedWebPlacerSystem"/>.
 /// </summary>
-public sealed class SpiderSystem : SharedSpiderSystem
+public sealed class WebPlacerSystem : SharedWebPlacerSystem
 {
     [Dependency] private readonly EntityLookupSystem _lookup = default!;
     [Dependency] private readonly EntityWhitelistSystem _whitelistSystem = default!;
@@ -26,10 +26,10 @@ public sealed class SpiderSystem : SharedSpiderSystem
     public override void Initialize()
     {
         base.Initialize();
-        SubscribeLocalEvent<SpiderComponent, SpiderWebActionEvent>(OnSpawnNet);
+        SubscribeLocalEvent<WebPlacerComponent, SpiderWebActionEvent>(OnSpawnNet);
     }
 
-    private void OnSpawnNet(EntityUid uid, SpiderComponent component, SpiderWebActionEvent args)
+    private void OnSpawnNet(EntityUid uid, WebPlacerComponent component, SpiderWebActionEvent args)
     {
         if (args.Handled)
             return;
@@ -37,6 +37,7 @@ public sealed class SpiderSystem : SharedSpiderSystem
         var transform = Transform(uid);
         var grid = transform.GridUid;
 
+        // Instantly fail in space
         if (grid == null || !TryComp<MapGridComponent>(grid, out var gridComp))
         {
             _popup.PopupEntity(Loc.GetString(component.MessageOffGrid), args.Performer, args.Performer);
@@ -52,7 +53,7 @@ public sealed class SpiderSystem : SharedSpiderSystem
         foreach (var vect in vects)
             spawnPos.Add(coords.Offset(vect));
 
-        // Spawn webs here
+        // Spawn webs
         bool success = false;
         foreach (var pos in spawnPos)
             if (IsValidTile(pos, component.DestinationWhitelist, component.DestinationBlacklist, grid.Value, gridComp))
@@ -61,6 +62,7 @@ public sealed class SpiderSystem : SharedSpiderSystem
                 success = true;
             }
 
+        // Process success
         if (success)
         {
             _popup.PopupEntity(Loc.GetString(component.MessageSuccess), args.Performer, args.Performer);
--- a/Content.Shared/WebPlacer/SharedWebPlacerSystem.cs
+++ b/Content.Shared/WebPlacer/SharedWebPlacerSystem.cs
@@ -2,12 +2,12 @@ using Content.Shared.Actions;
 using Robust.Shared.Network;
 using Robust.Shared.Random;
 
-namespace Content.Shared.Spider;
+namespace Content.Shared.WebPlacer;
 
 /// <summary>
-/// Gives entities (probably spiders) an action on init.  Spawning handled by <see cref="ServerSpiderSystem"/>..
+/// Gives entities (probably spiders) an action on init.  Spawning handled by <see cref="WebPlacerSystem"/>..
 /// </summary>
-public abstract class SharedSpiderSystem : EntitySystem
+public abstract class SharedWebPlacerSystem : EntitySystem
 {
     [Dependency] private readonly SharedActionsSystem _action = default!;
 
@@ -15,10 +15,10 @@ public abstract class SharedSpiderSystem : EntitySystem
     {
         base.Initialize();
 
-        SubscribeLocalEvent<SpiderComponent, MapInitEvent>(OnInit);
+        SubscribeLocalEvent<WebPlacerComponent, MapInitEvent>(OnInit);
     }
 
-    private void OnInit(EntityUid uid, SpiderComponent component, MapInitEvent args)
+    private void OnInit(EntityUid uid, WebPlacerComponent component, MapInitEvent args)
     {
         _action.AddAction(uid, ref component.ActionEntity, component.SpawnWebAction, uid);
     }
--- a/Content.Shared/WebPlacer/SharedWebPlacerSystem.cs
+++ b/Content.Shared/WebPlacer/SharedWebPlacerSystem.cs
@@ -1,11 +1,9 @@
 using Content.Shared.Actions;
-using Robust.Shared.Network;
-using Robust.Shared.Random;
 
 namespace Content.Shared.WebPlacer;
 
 /// <summary>
-/// Gives entities (probably spiders) an action on init.  Spawning handled by <see cref="WebPlacerSystem"/>..
+/// Gives entities (probably spiders) an action on init.  Spawning handled by <see cref="WebPlacerSystem"/>.
 /// </summary>
 public abstract class SharedWebPlacerSystem : EntitySystem
 {
--- a/Content.Shared/WebPlacer/WebPlacerComponent.cs
+++ b/Content.Shared/WebPlacer/WebPlacerComponent.cs
@@ -5,14 +5,14 @@ using Robust.Shared.GameStates;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype;
 
-namespace Content.Shared.Spider;
+namespace Content.Shared.WebPlacer;
 
 /// <summary>
 /// Gives the entity (probably a spider) an ability to spawn webs around itself.
 /// </summary>
 [RegisterComponent, NetworkedComponent]
-[Access(typeof(SharedSpiderSystem))]
-public sealed partial class SpiderComponent : Component
+[Access(typeof(SharedWebPlacerSystem))]
+public sealed partial class WebPlacerComponent : Component
 {
     /// <summary>
     /// Id of the entity getting spawned.
--- a/Resources/Prototypes/Entities/Objects/Misc/spider_web.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/spider_web.yml
@@ -16,15 +16,14 @@
       sprite: Objects/Misc/spiderweb.rsi
       layers:
         - state: spider_web_1
-          map: ["spiderWebLayer"]
+          map: [ "enum.DamageStateVisualLayers.Base" ]
       drawdepth: WallMountedItems
+    - type: RandomSprite
+      available:
+        - enum.DamageStateVisualLayers.Base:
+            spider_web_1: ""
+            spider_web_2: ""
     - type: Appearance
-    - type: GenericVisualizer
-      visuals:
-        enum.SpiderWebVisuals.Variant:
-          spiderWebLayer:
-            1:  {state: spider_web_1}
-            2:  {state: spider_web_2}
     - type: Clickable
     - type: Transform
       anchored: true
@@ -95,15 +94,14 @@
       sprite: Objects/Misc/spiderweb.rsi
       layers:
         - state: spider_web_clown_1
-          map: ["spiderWebLayer"]
+          map: [ "enum.DamageStateVisualLayers.Base" ]
       drawdepth: WallMountedItems
+    - type: RandomSprite
+      available:
+        - enum.DamageStateVisualLayers.Base:
+            spider_web_clown_1: ""
+            spider_web_clown_2: ""
     - type: Appearance
-    - type: GenericVisualizer
-      visuals:
-        enum.SpiderWebVisuals.Variant:
-          spiderWebLayer:
-            1:  {state: spider_web_clown_1}
-            2:  {state: spider_web_clown_2}
     - type: Clickable
     - type: Transform
       anchored: true
--- a/Resources/Prototypes/Entities/Objects/Misc/spider_web.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/spider_web.yml
@@ -68,13 +68,15 @@
       groups:
         Flammable: [Touch]
         Extinguish: [Touch]
-    - type: SpiderWebObject
     - type: SpeedModifierContacts
       walkSpeedModifier: 0.5
       sprintSpeedModifier: 0.5
       ignoreWhitelist:
-        components:
+        tags:
             - IgnoreSpiderWeb
+    - type: Tag
+      tags: 
+      - SpiderWeb
 
 - type: entity
   id: SpiderWebClown
@@ -150,7 +152,6 @@
       groups:
         Flammable: [Touch]
         Extinguish: [Touch]
-    - type: SpiderWebObject
     - type: FlavorProfile
       flavors:
         - sweet
@@ -162,3 +163,6 @@
           reagents:
           - ReagentId: Sugar
             Quantity: 2
+    - type: Tag
+      tags: 
+      - SpiderWeb
--- a/Content.Server/Spider/SpiderSystem.cs
+++ b/Content.Server/Spider/SpiderSystem.cs
@@ -9,6 +9,9 @@ using Robust.Shared.Map.Components;
 
 namespace Content.Server.Spider;
 
+/// <summary>
+/// Spawns entities (probably webs) around the component owner. Action handled by <see cref="SharedSpiderSystem"/>.
+/// </summary>
 public sealed class SpiderSystem : SharedSpiderSystem
 {
     [Dependency] private readonly EntityLookupSystem _lookup = default!;
@@ -33,7 +36,7 @@ public sealed class SpiderSystem : SharedSpiderSystem
 
         if (grid == null || !TryComp<MapGridComponent>(grid, out var gridComp))
         {
-            _popup.PopupEntity(Loc.GetString("spider-web-action-nogrid"), args.Performer, args.Performer);
+            _popup.PopupEntity(Loc.GetString(component._offGrid), args.Performer, args.Performer);
             return;
         }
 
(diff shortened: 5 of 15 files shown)


witchesfall (approved):

BolloTea:
Could you please leave the web overlay or put restrictions on web overlays if it is a big problem?

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        var coords = transform.Coordinates;
+        List<EntityCoordinates> spawnPos = new();
+
+        var vects = new List<Vector2i>();
+        vects.AddRange(component.OffsetVectors);
+
+        foreach (var vect in vects)
+            spawnPos.Add(coords.Offset(vect));
```
Aeshus, line 53: The local `vects` variable seems to just copy `component.OffsetVectors` and nothing else is added/mutated.

Is there a reason we can't just use `component.OffsetVectors` directly?

```suggestion
        foreach (var vect in component.OffsetVectors)
            spawnPos.Add(coords.Offset(vect));
```

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        // Instantly fail in space
+        if (grid == null || !TryComp<MapGridComponent>(grid, out var gridComp))
+        {
+            _popup.PopupEntity(Loc.GetString(component.MessageOffGrid), args.Performer, args.Performer);
+            return;
+        }
+
+        var coords = transform.Coordinates;
+        List<EntityCoordinates> spawnPos = new();
```
Aeshus, line 47: If we're using linq, this whole thing could be simplified to

```cs
        var spawnPos = component.OffsetVectors.Select(v => coords.Offset(v));
```

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,102 @@
...
+    private void OnSpawnNet(EntityUid uid, WebPlacerComponent component, SpiderWebActionEvent args)
+    {
+        if (args.Handled)
+            return;
+
+        var transform = Transform(uid);
+        var grid = transform.GridUid;
+
+        // Instantly fail in space
+        if (grid == null || !TryComp<MapGridComponent>(grid, out var gridComp))
```
Aeshus, line 40: TryComp should already return false if the grid/uid is null.

```suggestion
        if (!TryComp<MapGridComponent>(grid, out var gridComp))
```

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,102 @@
...
+            spawnPos.Add(coords.Offset(vect));
+
+        // Spawn webs
+        bool success = false;
+        foreach (var pos in spawnPos)
+            if (IsValidTile(pos, component.DestinationWhitelist, component.DestinationBlacklist, grid.Value, gridComp))
+            {
+                Spawn(component.WebPrototype, pos);
+                success = true;
+            }
```
Aeshus, line 62: I would probably invert the if-statement and add braces to the foreach loop. Something something nesting bad

```suggestion
        {
            if (!IsValidTile(pos, component.DestinationWhitelist, component.DestinationBlacklist, grid.Value, gridComp))
                continue;

            Spawn(component.WebPrototype, pos);
            success = true;
        }
```

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,102 @@
...
+        if (success)
+        {
+            _popup.PopupEntity(Loc.GetString(component.MessageSuccess), args.Performer, args.Performer);
+            args.Handled = true;
+
+            if (component.WebSound != null)
+                _audio.PlayPvs(component.WebSound, uid);
+        }
+        else
+            _popup.PopupEntity(Loc.GetString(component.MessageFail), args.Performer, args.Performer);
```
Aeshus, line 74: I would probably invert this one too and make the non-success state an early return.

```suggestion
        if (!success)
        {
            _popup.PopupEntity(Loc.GetString(component.MessageFail), args.Performer, args.Performer);
            return;
        }

        _popup.PopupEntity(Loc.GetString(component.MessageSuccess), args.Performer, args.Performer);
        args.Handled = true;

        if (component.WebSound != null)
            _audio.PlayPvs(component.WebSound, uid);
```

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,102 @@
...
+    [Dependency] private readonly EntityWhitelistSystem _whitelistSystem = default!;
+    [Dependency] private readonly ITileDefinitionManager _tile = default!;
+    [Dependency] private readonly PopupSystem _popup = default!;
+    [Dependency] private readonly SharedAudioSystem _audio = default!;
+    [Dependency] private readonly SharedMapSystem _map = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<WebPlacerComponent, SpiderWebActionEvent>(OnSpawnNet);
```
Aeshus, line 28: nitpick
```suggestion

        SubscribeLocalEvent<WebPlacerComponent, SpiderWebActionEvent>(OnSpawnNet);
```

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,79 @@
...
+    public List<Vector2i> OffsetVectors = new()
+    {
+        Vector2i.Zero,
+        Vector2i.Up,
+        Vector2i.Down,
+        Vector2i.Left,
+        Vector2i.Right,
+    };
+
+    /// Localization files for popup text.
```
Aeshus, line 71: The three slashes makes it a documentation comment for MessageOffGrid, while this is just supposed to be a local region separator thing.

```suggestion
    // Localization files for popup text.
```

Aeshus (changes requested):
Thanks for the PR!

I just have some nitpicks.

0x6273:
> `SpiderWebObjectComponent` and `IgnoreSpiderWebComponent` deleted. Both of these existed to be found by a whitelist in yaml, and have been replaced with tags.

You can use components in whitelists as well as tags. What's the reason for changing them to be tags?

iaada:
> You can use components in whitelists as well as tags. What's the reason for changing them to be tags?

It's because they were empty components. They didn't do anything except exist to be found by a whitelist.

0x6273:
> > You can use components in whitelists as well as tags. What's the reason for changing them to be tags?
> 
> It's because they were empty components. They didn't do anything except exist to be found by a whitelist, which is what tags are for.

But there's no real benefit to using tags over marker comps here. If fact I'm pretty sure tags are slower since it needs to both get the tag component and check if it has the tag, vs just checking for a comp.

Tags can't be used in queries either, so if a downstream relies on querying these comps they would either have to skip or partially merge this PR, or make their code slower and more complicated.

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,79 @@
...
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+[Access(typeof(SharedWebPlacerSystem))]
+public sealed partial class WebPlacerComponent : Component
+{
+    /// <summary>
+    /// Id of the entity getting spawned.
+    /// </summary>
+    [DataField(customTypeSerializer: typeof(PrototypeIdSerializer<EntityPrototype>))]
+    public string WebPrototype = "SpiderWeb";
```
Aeshus, line 21: It's generally best practice to use ProtoIds directly, now.

```suggestion
    [DataField]
    public ProtoId<EntityPrototype> WebPrototype = "SpiderWeb";
```

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    /// <summary>
+    /// Id of the action that will be given.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadOnly)]
+    [DataField(customTypeSerializer: typeof(PrototypeIdSerializer<EntityPrototype>))]
+    public string SpawnWebAction = "ActionSpiderWeb";
```
Aeshus, line 28:
```suggestion
    [DataField]
    public ProtoId<EntityPrototype> SpawnWebAction = "ActionSpiderWeb";
```

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,79 @@
...
+        Vector2i.Up,
+        Vector2i.Down,
+        Vector2i.Left,
+        Vector2i.Right,
+    };
+
+    // Localization files for popup text.
+    [DataField] public LocId MessageOffGrid = "spider-web-action-nogrid";
+    [DataField] public LocId MessageSuccess = "spider-web-action-success";
+    [DataField] public LocId MessageFail = "spider-web-action-fail";
```
Aeshus, line 74: Minor nitpick, I prefer when they're all on their own lines.

```suggestion
    [DataField] 
    public LocId MessageOffGrid = "spider-web-action-nogrid";

    [DataField] 
    public LocId MessageSuccess = "spider-web-action-success";

    [DataField] 
    public LocId MessageFail = "spider-web-action-fail";
```

Aeshus (commented):
Thanks for making the requested changes!

In return, you get some more nitpicks! ;)

Tayrtahn:
> But there's no real benefit to using tags over marker comps here. If fact I'm pretty sure tags are slower since it needs to both get the tag component and check if it has the tag, vs just checking for a comp.
> 
> Tags can't be used in queries either, so if a downstream relies on querying these comps they would either have to skip or partially merge this PR, or make their code slower and more complicated.

Agreed. Marker components are common in ECS and avoid a few issues that come up with the way Tags work (like inheritance). I endorse sticking with using marker components here.

iaada:
I can see your points with `IgnoreSpiderWeb`. Considering how often it gets checked and it being needed by mobs with several tags already, it's probably best to leave it as a marker component. I've re-added it. `SpiderWebObject` however is not inherited anywhere and only queried when spawning the webs every 25 seconds, and has no inherent functionality. It feels like prime use case for a tag.

Content.Shared/WebPlacer/SharedWebPlacerSystem.cs
```diff
@@ -0,0 +1,23 @@
+using Content.Shared.Actions;
+
+namespace Content.Shared.WebPlacer;
+
+/// <summary>
+/// Gives the component owner (probably a spider) an action to spawn entites around itself. Spawning handled by <see cref="WebPlacerSystem"/>.
+/// </summary>
```
Aeshus, line 7: NITPICK: Misspelled entities, and you should probably just use seealso to reference the component rather than the server system that inherits from this.
 
```suggestion
/// <summary>
///     Gives the component owner (probably a spider) an action to spawn entities around itself. Spawning handled by
/// </summary>
/// <seealso cref="WebPlacerComponent" />
```

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,81 @@
...
+        Vector2i.Right,
+    };
+
+    // Localization files for popup text.
+    [DataField]
+    public LocId MessageOffGrid = "spider-web-action-nogrid";
+    [DataField]
+    public LocId MessageSuccess = "spider-web-action-success";
+    [DataField]
+    public LocId MessageFail = "spider-web-action-fail";
```
Aeshus, line 76: You need some spacing and also documentation comments for every field in a component.

Also, I would consider renaming `MessageFail` to be more specific that this is an error message for "no room found" and not some more general failure state.

```suggestion
    /// <summary>
    /// Webs cannot be placed because there is no valid grid.
    /// </summary>
    [DataField]
    public LocId MessageOffGrid = "spider-web-action-nogrid";

    /// <summary>
    /// Webs placed successfully.
    /// </summary>
    [DataField]
    public LocId MessageSuccess = "spider-web-action-success";

    /// <summary>
    /// Webs failed to be placed (e.g. no space).
    /// </summary>
    [DataField]
    public LocId MessageFail = "spider-web-action-fail";
```

Resources/Prototypes/Actions/spider.yml
```diff
@@ -1,3 +1,4 @@
+# Default prototype used by WebPlacerComponent
```
Aeshus, line 1: This comment feels redundant.

```suggestion
```

Resources/Prototypes/Entities/Objects/Misc/spider_web.yml
```diff
@@ -1,3 +1,4 @@
+# Default prototype used by WebPlacerComponent
```
Aeshus, line 1:
```suggestion
```

Content.Shared/IgnoreSpiderWeb/IgnoreSpiderWebComponent.cs
```diff
@@ -0,0 +1,10 @@
+namespace Content.Shared.IgnoreSpiderWeb;
+
+/// <summary>
+/// Marker component for entities that ignore spiderwebs.
+/// </summary>
+[RegisterComponent]
+public sealed partial class IgnoreSpiderWebComponent : Component
+{
+
```
Aeshus, line 9:
```suggestion
```

Aeshus (changes requested):
Thanks for making the requested changes!

I just have some more comments and changes, but the PR is definitely getting to the state of being ready to be merged!

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,81 @@
...
+using Content.Shared.Whitelist;
+using Robust.Shared.Audio;
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.WebPlacer;
+
+/// <summary>
+/// Gives the entity (probably a spider) an ability to spawn entites (probably webs) around itself.
+/// </summary>
```
Aeshus, line 11: Oops, forgot this misspelling too.

```suggestion
/// <summary>
/// Gives the entity (probably a spider) an ability to spawn entities (probably webs) around itself.
/// </summary>
```

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,100 @@
+using System.Linq;
```
TheShuEd, line 1: dont use linq

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Content.Server.WebPlacer;
+
+/// <summary>
+/// Spawns entities (probably webs) around the component owner. Action handled by <see cref="SharedWebPlacerSystem"/>.
+/// </summary>
+public sealed class WebPlacerSystem : SharedWebPlacerSystem
+{
+    [Dependency] private readonly EntityLookupSystem _lookup = default!;
+    [Dependency] private readonly EntityWhitelistSystem _whitelistSystem = default!;
```
TheShuEd, line 20:
```suggestion
    [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
```
codestyle convention

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,100 @@
...
+    [Dependency] private readonly SharedMapSystem _map = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<WebPlacerComponent, SpiderWebActionEvent>(OnSpawnNet);
+    }
+
+    private void OnSpawnNet(EntityUid uid, WebPlacerComponent component, SpiderWebActionEvent args)
```
TheShuEd, line 33:
```suggestion
    private void OnSpawnNet(Entity<WebPlacerComponent> webPlacer, SpiderWebActionEvent args)
```

use modern Entity<>. You can access component by `webPlacer.Comp`

Content.Shared/WebPlacer/SharedWebPlacerSystem.cs
```diff
@@ -0,0 +1,23 @@
...
+    [Dependency] private readonly SharedActionsSystem _action = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<WebPlacerComponent, MapInitEvent>(OnInit);
+    }
+
+    private void OnInit(EntityUid uid, WebPlacerComponent component, MapInitEvent args)
```
TheShuEd, line 19: Entity<>

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// Id of the action that will be given.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadOnly)]
+    [DataField]
+    public EntProtoId SpawnWebAction = "ActionSpiderWeb";
+
+    /// <summary>
+    /// Action given to the player.
+    /// </summary>
+    //[ViewVariables]
```
TheShuEd, line 32: why its commented? Where DataField

TheShuEd (changes requested):
Codestyle review

Content.Shared/WebPlacer/WebPlacerComponent.cs
iaada: I didn't add DataField because this variable is getting set on init, and to my understanding EntityUid can't be set in .yml so it felt unnecessary.
VV being commented was just a mistake on my part.

Content.Server/WebPlacer/WebPlacerSystem.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        // Get the spawn locations.
+        List<EntityCoordinates> spawnPos = new();
+        foreach (var vect in webPlacer.Comp.OffsetVectors)
+            spawnPos.Add(xform.Coordinates.Offset(vect));
+
+        bool success = false;
+        foreach (var pos in spawnPos)
+        {
```
Aeshus, line 55: Ed is a little overzealous with the "no linq" thing (Sloth said it's fine to use select here), but frankly you dont even need a second loop here to map the values.

This runs in fewer cycles and uses less memory. Win-win!

```suggestion
        bool success = false;
        foreach (var vect in webPlacer.Comp.OffsetVectors)
        {
            var pos = xform.Coordinates.Offset(vect);
```

iaada: In the back of my mind I knew I didn't need two loops, but I just didn't think too hard about it.

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// Id of the action that will be given.
+    /// </summary>
+    [DataField]
+    public EntProtoId SpawnWebAction = "ActionSpiderWeb";
+
+    /// <summary>
+    /// Action given to the player.
+    /// </summary>
+    [ViewVariables]
+    public EntityUid? ActionEntity;
```
0x6273, line 34: Shouldn't this be a datafield? It will be null if you save and load, since it's only set on MapInit.

iaada: I guess I'm unclear on what exactly being a datafield does in this case, but you're the second person to say it so it's probably correct.

Tayrtahn: DataField marks a field for serialization. That means it can have a value assigned in YAML and will be written/read when saving/loading.

iaada:
I decided to add back the spiderweb marker component. If spiders ever get more than just webs that slow, it'll be good to have.

I also noticed the fixtures on the clown web were weird, and fixed that as well.

iaada:
`IgnoreSpiderWebComponent` now cancels step triggers with spider webs. If there is a way to cancel collide triggers I couldn't find it, so it's not quite perfect at ignoring webs.

Content.Shared/WebPlacer/WebPlacerComponent.cs
```diff
@@ -0,0 +1,108 @@
...
+    [DataField]
+    public List<Vector2i> OffsetVectors = new()
+    {
+        Vector2i.Zero,
+        Vector2i.Up,
+        Vector2i.Down,
+        Vector2i.Left,
+        Vector2i.Right,
+    };
+
```
metalgearsloth, line 68: You can use directionflags + bitmask to make this smaller.

Content.Shared/WebPlacer/SharedWebPlacerSystem.cs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// <summary>
+    /// Give the entity its spawning action.
+    /// </summary>
+    private void OnInit(Entity<WebPlacerComponent> ent, ref MapInitEvent args)
+    {
+        _action.AddAction(ent.Owner, ref ent.Comp.ActionEntity, ent.Comp.SpawnWebAction, ent.Owner);
+    }
+
```
metalgearsloth, line 44: ActionGrant does the same thing and means you don't need to hardcode the action to the component.

PR #33748, opened 2024-12-06
Engieborg improvements - New lingering plasteel, reinforced glass, black/white tile stacks for construction module + recharging holofan projector for advanced tool module

ArtisticRoomba:
## About the PR
Adds new lingering reinforced glass, plasteel, and white/dark tile stacks. All included in the construction module.

Adds new Holofan Projector borg variant, included in the advanced tool module.

## Why / Balance
Borgs right now don't have the ability to:
- Construct new reinforced walls and reinforced windows, which is suboptimal given the current meteor balance situation.
- Place holofans and assist in atmospheric emergencies, something that an engineering borg should be able to do. Right now, they can't navigate firelocks without causing further gas spread or spacing, which could cause crew harm under the default lawset.

This PR addresses this by adding new lingering variants for reinforced glass, plasteel, dark tiles, and white tiles. This is included in the construction module, which engineering borgs get roundstart.

A self-recharging holofan has also been added to the Advanced Tool Module. I believe it is balanced, but if maintainers disagree, the charge consumed per holofan can be changed to make it so that borgs will have to wait longer in between charges.

I think these changes are good, considering that the engineering borg should be able to specialize and perform station maintenance on reinforced walls and glass.

I *do* understand this is really just increasing borghands spam... but I see this as a necessary compromise before hand whitelisting is coded.

## Technical details
Added new lingering variants and a borg holofan projector, based on previous solutions seen in-game.

## Media
Content Client_TF2YS0p4dm

## Requirements

**Changelog**
:cl:
- add: The construction cyborg module can now handle plasteel, reinforced glass, and white/dark tiles.
- add: The advanced tool cyborg module now includes a self-recharging Holofan Projector.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
@@ -74,9 +74,6 @@
   suffix: borg
   description: Stops suicidal crew from causing more crew harm during atmos emergencies. Powered by a self-recharing power cell.
   components:
-  - type: HolosignProjector
-    signProto: HoloFan
-    chargeUse: 120
   - type: ItemSlots
     slots:
       cell_slot:
--- a/Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
@@ -54,7 +54,7 @@
   parent: Holoprojector
   id: HolofanProjector
   name: holofan projector
-  description: Stop suicidal passengers from killing everyone during atmos emergencies.
+  description: Stops suicidal passengers from killing everyone during atmos emergencies.
   components:
   - type: HolosignProjector
     signProto: HoloFan
@@ -67,6 +67,23 @@
       - HolofanProjector
   - type: StaticPrice
     price: 80
+
+- type: entity
+  parent: HolofanProjector
+  id: HolofanProjectorBorg
+  suffix: borg
+  description: Stops suicidal crew from causing more crew harm during atmos emergencies. Powered by a self-recharing power cell.
+  components:
+  - type: HolosignProjector
+    signProto: HoloFan
+    chargeUse: 120
+  - type: ItemSlots
+    slots:
+      cell_slot:
+        name: power-cell-slot-component-slot-name-default
+        startingItem: PowerCellMicroreactor
+        disableEject: true
+        swap: false
 
 - type: entity
   parent: HolofanProjector
--- a/Resources/Prototypes/Entities/Objects/Materials/Sheets/glass.yml
+++ b/Resources/Prototypes/Entities/Objects/Materials/Sheets/glass.yml
@@ -174,6 +174,15 @@
 
 - type: entity
   parent: SheetRGlass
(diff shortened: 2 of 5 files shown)


SlimmSlamm:
thank god although I think instead of having the specific tiles it should allow you to pick up ANY stack of tiles.

southbridge-fur:
> thank god although I think instead of having the specific tiles it should allow you to pick up ANY stack of tiles.

I also would like this but that seems like s pain to implement. Engi borgs can't toss away unwanted tiles, and if the UI grew over time it could get unruly.

SlimmSlamm:
> > thank god although I think instead of having the specific tiles it should allow you to pick up ANY stack of tiles.
> 
> I also would like this but that seems like a pain to implement. Engi borgs can't toss away unwanted tiles, and if the UI grew over time it could get unruly.
> 
> We would need some method to cycle through all tiles in a single slot like the advanced tool module used to with its tools.

I think being able to drop materials/tiles would be nice. I don't really care about having more than one type all I ask is that I don't need to find/make the one single tile I'm able to hold. I sometimes forget and end up wasting a few materials or at other times I accidentally crowbar the floor and just cannot replace it because it isn't the tile engi Borgs can use.

ArtisticRoomba:
> I think instead of having the specific tiles it should allow you to pick up ANY stack of tiles.

Yeah, it would be nice, but this is out of scope, hand whitelisting needs to be coded. If I were to implement it without hand whitelisting right now, I think someone would shoot me.

Hand whitelisting is also just a pain to develop right now, because:
- Whitelists and blacklists are not intuitive because they aren't communicated to the user at all.
- You would have to manually whitelist/blacklist items for the hands.

Some discussions have taken place, mainly:
- Borgs could get "borg manipulator" hands, and an innate large inventory tied to their frame. Modules add tools to that inventory, and the borg can use their borg hands to fill their toolbar the way they like. The hands won't allow dropping/exchanging.
- For manipulating certain things in the world, borgs could get named "manipulator" hands. For example, a hand that could manipulate machine boards could be called a "board manipulator" and a hand that can manipulate components could be called a "component manipulator". Hell, you could even chalk it under the common term "machine part manipulator". The idea is that this solves the non-communicated whitelist problem, and makes borgs intuitive to play.

There's a lot of stuff to go over. The age old saying applies: it needs a design document.

Resources/Prototypes/Entities/Objects/Devices/holoprojectors.yml
```diff
@@ -68,6 +68,23 @@
...
 
+- type: entity
+  parent: HolofanProjector
+  id: HolofanProjectorBorg
+  suffix: borg
+  description: Stops suicidal crew from causing more crew harm during atmos emergencies. Powered by a self-recharing power cell.
+  components:
+  - type: HolosignProjector
+    signProto: HoloFan
+    chargeUse: 120
```
ArtisticRoomba, line 79: Inherited from `HolofanProjector` parent and is not needed

```suggestion
```

ArtisticRoomba (commented):
Made a mistake, genius me

ArtisticRoomba:
Superseded.

PR #32825, opened 2024-10-15
Fix AME power generation

Golinth:
## About the PR

Adjusted the formula for the AME power generation to not punish larger AMEs. Previously, any AME above 2 core would produce less power, the new formula adjusts it to no longer do that. With the new formula, a 4 core AME (the highest naturally mapped in-game atm) is equivalent to a 2-core AME - 180kW. Additionally, I ensured that having more cores but the same injection amount does lead to decreased power. 

If the AME is returned to cargo for purchase, the realistic cap is around 300kW with ~60 cores, but if you REALLY wanted to and could source 600 cores (and somehow 1200 fuel per injection) you can hit 400kW. 

Check the math [at desmos](https://www.desmos.com/calculator/70hhvmc524) along with some explanatory comments, if you want further proof of this really being an AME nerf in disguise. 

## Why / Balance

Resolves #31758, check it for more details, but at the moment higher core AMEs are less than worthless, as they actively produce less power. This change makes higher core AMEs at least somewhat desireable, as they produce marginally more power, even if they take significantly more fuel.

## Technical details

Adjusted one equation, added some explanatory comments. 
Formula can be messed with and tested [here.](https://www.desmos.com/calculator/70hhvmc524) (leads to desmos). 

## Media

Red line is current formula, Blue line is new one. As shown, at one core the power production is the exact same. 

Y-axis is kW produces, X axis is number of cores, assuming optimal (1:2) core to fuel ratio. 
image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- tweak: AMEs larger than 2 cores now produce more power, instead of less.

PR #29698, opened 2024-07-03
Add basic test of station initial power supply

Tayrtahn:
## About the PR

Adds an integration test that estimates how long each station's initial stored power supply will last with the initial power load. Currently set to a 10 minute target threshold, which none of the maps meet.

## Why / Balance

We should establish a reasonable startup window on all maps so that engineering has a consistent and reasonable amount of time to get things set up.

## Technical details

Pretty simple logic: find the power network with the greatest initial stored charge (this should be the main SMESes and friends, and should exclude stuff like solar SMESes). Then sum up all the desired loads of all APC power consumers (direct users of HV power like the PA are typically not connected at round start), and divide the total charge by that load.

This logic may need some improvement! It's a pretty rough estimate, since it doesn't take into account stuff like the ramp-up delay on batteries, and it could potentially ignore some power storage in cases where the networks are more complicated. Help me improve it!

## Media

Code.
- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

Nah.

PR #33757, opened 2024-12-07
Adds Advanced SMES, an SMES with higher capacity for mapping

ArtisticRoomba:
## About the PR
Adds an Advanced SMES, an SMES with a capacity of 16 MJ (16,000,000 J) intended for mapping. For comparison, the basic SMES has a base capacity of 8 MJ.

This is intended to be used in station battery storage banks and provides more agency for mappers to adjust how their station performs power-wise and for easier future balancing.

If this PR gets merged, I will create a megaissue addressing the current engineering roundflow power issues and how it should be addressed according to the (legacy, but still included, and very much still relevant in my opinion) [engineering design document](https://docs.spacestation14.com/en/space-station-14/departments/engineering/proposals/power-generation.html).

I didn't buff base SMESes as I wanted mappers to have more agency.

~~Requires #33808 or else the sprites will look weird.~~ Merged.

## Why / Balance
Right now, engineering has a horrible power roundflow problem. Engineering power is **intended** to follow this pattern:
- The station is running on battery power at the start of the shift. This will deplete rapidly until engineers setup an AME to help supplement the batteries. Station batteries should reach around 30% capacity at the 5 minute mark, where the AME is started.
- Solars can be setup to further reduce this drain, with them expected to be setup at the 10 minute mark. At this time, batteries should be drained to around 20%. Solars should further offset and supplement the batteries, but there is still a little bit of power drain that warrants a greater power source.
- Now that the batteries are not running out in 2 minutes, engineering can dedicate enough time to setting up a *proper* power source, like the Tesla, Singulo, or TEG. This power source should be setup at the 20 minute mark. Once this is done, engineering will finally be power positive, and can turn off the AME.

Below is a graph from the design document showing the intended flow:
chrome_97lATAtAU9

The RED line is Meta station. It doesn't even last 3 minutes, 30 seconds before the station starts to brown out. Other large stations are worse than this. This is not good at all, and promotes a lot of problems:

- Engineers don't have enough time to teach new TAs engines, they are pushed to rush setup of these engines. I have had to **turn away TAs that want to learn** because taking the moment to explain engines may lead to the station running out of power.
- Engineers get yelled at constantly by people with their departments in brownout because batteries with AME assistance last literally 7 minutes. This often leads to them cryoing and never wanting to play engineering.
- This little buffer time is horrible for lowpop engineering and engineering with only TAs.

This change is intended to give mappers a higher capacity SMES to help bridge that gap to the AME turning on, without having to spam 12 SMESes.

## Technical details
Added and adjusted relevant things in YAML.

## Media
image
image

## Requirements

**Changelog**
:cl: ArtisticRoomba, AugustSun
- add: New Advanced SMES, an SMES with a 16 MJ capacity. It can be researched at Industrial T2, Advanced Powercells.


ArtisticRoomba:
> 1. I would add some red or orange stripes to the top and bottom rims.

I think this would be good to distinguish it, but I am horrible at spriting.
> 2. Does the capacity of the power cell used in substation/SMES construction affect its capacity?

Nope. Component-based upgrades are not desired, explicit machine upgrades should be used instead.
> 3. How is the graph affected when you turn on the roundstart plasma and uranium PACMANs?

I've discussed the usage of generators to help bridge power, and I'll mirror my thoughts about it below:
- The fact that I should have to pry up a floor tile and connect a generator to a grid manually instead of using a station's purpose-built support generator because it sucks is not good game design.
- If mappers want players to use generators to help bridge power and extend the lifetime of the SMES banks, it should be a **clearly, intuitively obvious, mapped feature.** See Fland for an excellent example and see Cog for a horrible example.

> 4. Should we buff substation capacities instead so departments get a longer blackout warning? (Essentially when light flickers and the APCs turn red to show they're losing power.)

I'm partially against giving powernet transformers a larger internal battery for the sake of a band-aid solution to a larger core problem. I think it's good to have this PR's changes reflected in the SMES percentages because it's an easier thing to balance around. However, it could be a good change when things like substation/APC having breakers and APC equipment networks (toggling breakers to equipment, doors, lighting separately) are a thing in the future.

ArtisticRoomba:
> I agree with the point they should already be anchored or positioned on them roundstart like on Cog or Fland. Thus I'm asking what the graph will look like when every map does it.

The underlying issue is that the batteries don't last long enough to support the station's SMES array until the 5 minute mark, where engineers have presumed to have setup the AME and other low-tier sources like PACMANs and the AME.

The solution is mappers taking advantage of dual-core AMEs, something that should be mapped for large stations, as some stations need more power to help assist the batteries than others.

To answer your question, I turned on the AME, PACMAN and SUPERPACMAN at the 5 minute mark (when the SMESes had already ran out for about 2 minutes, but the station was still running off of substation and APC battery power), which brought the load down to 80 kW. Assuming that engineers setup solars by the 10 minute mark, engineering would actually be basically power-neutral, giving them plenty of time to setup an engine (assuming they do it in the timeframe the generators will last).

With my PR, I did testing on Cog, the notorious map where engineers notoriously run into problems. The batteries got to around 38% at the five minute mark, and with a 2-core AME assisting the batteries (compared to the current 1-core), it got to the 10 minute mark at 13%, where solars "kicked in". Once the solars had balanced the load, battery usage was down to around 50-60 kW. The SMESes then fully depleted at the 16 minute mark. A PACMAN generator (which is mapped intuitively) setup alongside the AME would help greatly.

image

In reality, the AME is setup a bit sooner, and solars are set up piecewise, helping to offset the grid much sooner.

thebadman4662:
> * Does the capacity of the power cell used in substation/SMES construction affect its capacity?

Cell substation/SMES upgrades got axed while ago in substation resprite PR. Which was halfish year ago or so.

Golinth:
> > * Does the capacity of the power cell used in substation/SMES construction affect its capacity?
> 
> Cell substation/SMES upgrades got axed while ago in substation resprite PR. Which was halfish year ago or so.

I still have issues with the removal of some part upgrades, like those, to this day. They weren’t all perfect, but especially for substations and SMES it just makes sense

ArtisticRoomba:
> I still have issues with the removal of some part upgrades, like those, to this day. They weren’t all perfect, but especially for substations and SMES it just makes sense

I agree, yeah. For stuff like substations, cell upgrades do indeed make *sense*, however, they aren't explicit.

ArtisticRoomba:
For any maintainer that reviews this, I would also like #32825 looked at as well, as this ties into my larger goals of mappers being able to use multi-core AMEs to help achieve the roundflow SMES power levels as outlined in the current engineering design document.

ArtisticRoomba:
I intend to make some changes soon:
- Add a sprite to the advanced SMES to fully distinguish it between the regular variant.
- Make the power capacity actually double because it uses double the materials. One MJ doesn't reflect to a huge change in time anyways, believe it or not.
- Move the advanced SMES under the advanced cells T2 research because it makes much more sense there. Undesirable to have it available roundstart.

ArtisticRoomba:
I'm finished with my changes. This PR now requires #33808 or else the sprites will look weird. Thanks AugustSun for the sprites.

lexaSvarshik:
You can just overcharge SMES'es, so they will be charged for 400% roundstart, so no power outage for 30 minutes for crew

ArtisticRoomba:
That would require me to code. I could ask a fork if it's okay to implement their solution.

Besides that, I think it's good to give engineering an explicit upgradeable SMES for those that want to bother. Maybe engineering might go around the station and start upgrading them? It's just nice to have something mechanical, "overcharging" feels weird.

Resources/Prototypes/Entities/Structures/Power/smes.yml
```diff
@@ -108,3 +108,48 @@
   components:
   - type: Battery
     startingCharge: 0
+
+- type: entity
+  parent: BaseSMES
+  id: SMESAdvanced
+  suffix: Advanced, 15MJ
```
ScarKy0, line 115:
```suggestion
  suffix: Advanced, 16MJ
```

ScarKy0 (changes requested):
By no means am I a yaml magician but all looks good except that one suffix

ArtisticRoomba:
Lmfao yeah thanks

ArtisticRoomba:
Content Client_HmjLw9pLvK

new sprites once again, thanksies AugustSun
everything seems to work on the sprites side with the lights

I can also confirm that PR #33808 works and they have proper open sprites.

Resources/Prototypes/Entities/Structures/Power/smes.yml
```diff
@@ -111,3 +111,48 @@
   components:
   - type: Battery
     startingCharge: 0
+
+- type: entity
+  parent: BaseSMES
+  id: SMESAdvanced
+  suffix: Advanced, 16MJ
+  name: advanced SMES
+  description: A high-capacity superconducting magnetic energy storage (SMES) unit. This one boasts a larger capacity for powering stations.
```
Aeshus, line 120: Comedy is relative.

```suggestion
  description: An even-higher-capacity superconducting magnetic energy storage (SMES) unit.
```

Resources/Prototypes/Entities/Objects/Devices/Circuitboards/Machine/production.yml
```diff
@@ -640,6 +640,25 @@
...
+    state: power_mod
+  - type: MachineBoard
+    prototype: SMESAdvancedEmpty
+    stackRequirements:
+      Capacitor: 2
+      CableHV: 20
+    componentRequirements:
+      PowerCell:
+        amount: 4
+        defaultPrototype: PowerCellSmall
```
Aeshus, line 660: Thoughts on maybe making it use 4 `PowerCellMedium`s?

Resources/Prototypes/Entities/Structures/Machines/lathe.yml
```diff
@@ -520,6 +520,8 @@
       - MassMediaCircuitboard
       - ReagentGrinderIndustrialMachineCircuitboard
       - JukeboxCircuitBoard
+      - SMESAdvancedMachineCircuitboard
+
```
Aeshus, line 524: None of the other lathes have spacing between these lines, and it can cause confusion if this is left here.

```suggestion
```

Resources/Prototypes/Entities/Objects/Devices/Circuitboards/Machine/production.yml
ArtisticRoomba: I did, but I was suggested to make it slightly cheaper, so I went with smalls. I was thinking of changing it back.

Aeshus: It is double the capacity so I would at least use 6 smalls so that it isn't equivalent.

ArtisticRoomba: Hm, true. I'll use 4 mediums as science can usually print those like candy by the time this is researched.

Resources/Textures/Structures/Power/smes.rsi/meta.json
```diff
@@ -1,16 +1,25 @@
...
+      "name": "advancedsmes-static"
+    },
+    {
+      "name": "advancedsmes-open"
+    },
     {
       "name": "smes"
     },
-		{
+	{
```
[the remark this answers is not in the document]
ArtisticRoomba: rider auto formatting for .json files is actually the 10th circle of hell and you cannot possibly convince me otherwise

SlamBamActionman:
I think the idea is sound (always love a good graph) and  evening out the roundstart power is something we're interested in. I've made a discussion thread to check with maintainers who are better at mapping/Engineering than me, and the sprite will need Art Lead approval.

SlamBamActionman (approved):

ArtisticRoomba:
> I think the idea is sound (always love a good graph) and  evening out the roundstart power is something we're interested in. I've made a discussion thread to check with maintainers who are better at mapping/Engineering than me, and the sprite will need Art Lead approval.

Excellent, let me know how it goes. If it's possible, I'd like PR #32825 included in that discussion as it ties into my larger goal with standardizing station power runtimes.

Partmedia:
Thank you for your work and your detailed analysis. It seems that you concluded that the core issue is mapping -- mappers aren't putting in enough round-start energy storage to make the red curve align with the blue one.

So while the proposed solution (adding a bigger mappable SMES) gives mappers more flexibility, ultimately it doesn't solve the problem of testing power draw on map start. Ultimately, even without this change, mappers can still map a larger round-start SMES bank, no?

ArtisticRoomba:
> Thank you for your work and your detailed analysis. It seems that you concluded that the core issue is mapping -- mappers aren't putting in enough round-start energy storage to make the red curve align with the blue one.
> 
> So while the proposed solution (adding a bigger mappable SMES) gives mappers more flexibility, ultimately it doesn't solve the problem of testing power draw on map start. Ultimately, even without this change, mappers can still map a larger round-start SMES bank, no?

Questioning this PR, I can see your concerns with why it exists in the first place. You are correct that mappers can simply map more roundstart SMES banks in order to compensate for draw, and that this PR isn't really needed. The issue I saw (and what this PR is for) is that using regular 8MJ SMESes leads to a very large battery room for the station. With the Advanced SMESes, the density would be doubled, allowing for a simple drop-in replacement with little requirement for re-mapping SMES rooms on most maps.

For solving the problem of testing power draw on round start, I believe there is a test for station power over time, #29698. However, it needs to be adapted to follow the intended graph. Right now it simply checks for a 10-minute time on batteries alone, which doesn't match with engineering's intended gameplay.

There is a second component to the power issue, which is that the post-AME power graph's slope can't really be modified right now for maps that need it. The AME's current power curve causes core counts over 2 to produce less power, which means that for some large maps, the 20% at 20 minutes target can't be made.

If this PR and the AME PR (#32825) are merged, I'll go ahead and re-tune all maps to conform to the graph.

Partmedia:
[Intuitive simulation](https://docs.spacestation14.com/en/space-station-14/core-design.html) is one of our core design pillars, and I think it makes a lot of intuitive sense that a large station should also be served by a large SMES bank.

That said, I also think that it is useful to give mappers more flexibility, and this advanced SMES is justified as being more "advanced" and also only represents a modest 2× increase in capacity. So I think it's fine.

I do want to emphasize that it is important for mappers to check their round start power draw, because I can see this problem either not getting solved, or small stations getting over-corrected.

Thanks again for your work.

ArtisticRoomba:
To any maintainer who merges this, after the internal discussions with the engineering workgroup about this PR:
This PR's greater goal is to standardize the power roundflow. A huge component that enables this is PR #32825, which rebalances the AME curve to allow core counts higher than 2, with a logarithmic loss. This allows the power curve to be easily fitted for larger stations.

PR #32825 should be merged alongside this PR.

SlamBamActionman:
Emisse has given approval on Discord during the Maintainer Discussion, and seeing as they're the head mapper that means this is Good To Merge+.

PR #33762, opened 2024-12-07
Singularity equipment can now be activated with E interact

PJB3005:
These were coded to only listen to InteractHand, which doesn't follow our interaction model properly.

:cl:
- fix: You can now use Interact (E by default) to toggle radiation collectors, field generators, and emitters. Previously you had to click with an empty hand.


slarticodefast (approved):
Tested and works as intended.

PR #33763, opened 2024-12-07
Give silicons proper lobby/character editor previews

PJB3005:
No more naked dummies, properly show a borg/AI sprite now.

This means taking the JobEntity into account when spawning the dummy. For AIs I had to add a "JobPreviewEntity" field because they'd look like a posibrain otherwise. AI therefore uses a custom dummy entity I defined.

Also I had to add some margins to the UI, because otherwise the 32x32 sprite of the AI would look bad.

image
image
image
image

:cl:
- tweak: Silicons now have a proper preview in the lobby/character editor.

### DIFF
--- a/Content.Client/Lobby/LobbyUIController.cs
+++ b/Content.Client/Lobby/LobbyUIController.cs
@@ -455,7 +455,21 @@ public sealed class LobbyUIController : UIController, IOnStateEntered<LobbyState
     {
         EntityUid dummyEnt;
 
-        if (humanoid is not null)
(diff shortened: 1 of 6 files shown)


slarticodefast (changes requested):
I'm doing some minor cleanup.
Tested and works as intended. No more naked Urists in the lobby.

slarticodefast (approved):

muburu:
now all we need is https://github.com/space-wizards/space-station-14/pull/31303 and silicons will finally be playable :godo:

PR #29574, opened 2024-06-29
Limit air lost by wrenching/breaking pipes

osjarw:
## About the PR
Capped the amount of air lost whenever a pipe is unwrenched/broken. (meteors)

## Why / Balance
Losing 15 minutes of accumulated air because someone unanchored a single pipe is kinda awful. This also applies to explosions. 

## Technical details
The amount of lost air is limited to the volume of the part being removed. 

Tested with Release on Oasis, @ 1500kPa, removing a single pipe used to drop distro by ~1300kPa, now the drop is ~1kPa.

## Media

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

Observe the air before and after the pipe is unanchored

before the change:

https://github.com/space-wizards/space-station-14/assets/62134478/6b23a9a2-f0b4-422d-b165-132c93482f06

after the change:

https://github.com/space-wizards/space-station-14/assets/62134478/b5698c38-bf52-4d1d-b7fb-55a253d157ef

## Breaking changes

**Changelog**

PR #33774, opened 2024-12-08
Fix gas pipe leaking when unanchoring or breaking them

### LINKED ISSUE #32632: Pipe unanchoring is exploitable
## Description
All pipes eternally hold gases even when you unanchor them. The pipe will just eternally be filled with gases until you deconstruct it and build a new one. Pipes will hold the gas indefinitely and can be anchored and unanchored to cause mass destruction

**Reproduction**
Build a set of pipes, fill them with a pump, and unanchor one. There will still be gas in the unanchored pipe, which you can then release by anchoring it elsewhere and unanchoring it again

**Screenshots**
This is after I anchored and unanchored it a few times but the initial pressure was 1400kpa

slarticodefast:
## About the PR
Currently when unanchoring a pipe some of the gas will leak from it, but not all of it. This means you can pick it up as a small 200L tank in your backpack and release the gas into the atmosphere by anchoring and unachoring it a second time, which can be potentially abused.
This PR solves this by making the leak amount equal to $V_{pipe} / V_{total}$ which was suggested by @Partmedia in #29574.
The removed pipe will then always be completely empty.

## Why / Balance
Fixes #32632

## Technical details
I simplified the way the leak amount is calculated, removing the pressure dependency and making it only volume dependend, which makes more sense. The warning popup will now show up when the pipes pressure is above standard atmospheric pressure, below that the gas still leaks, but the amount is insignificant.

When the pipe is unanchored, the node group update will be scheduled for next tick. However, this will mean we cannot yet remove the gas from the pipe element since it is still connected with its neighbors. We cannot do this in the next tick either, because the information about the pipe entity and its node is lost by then. To solve this we force a manual update for the scheduled node groups. This should be fine performance-wise since it only happens once when a player unanchors it and the update won't be repeated in the next tick.

## Media

https://github.com/user-attachments/assets/482efc56-0b86-4e92-b69c-cfa0d15e85e7

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- fix: Fixed pipes still containing gas after unanchoring them.

### DIFF
--- a/Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
+++ b/Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
@@ -1,6 +1,7 @@
 using Content.Server.Atmos.EntitySystems;
 using Content.Server.Atmos.Piping.Components;
 using Content.Server.NodeContainer;
+using Content.Server.NodeContainer.EntitySystems;
 using Content.Server.NodeContainer.Nodes;
 using Content.Server.Popups;
 using Content.Shared.Atmos;
@@ -8,7 +9,6 @@ using Content.Shared.Construction.Components;
 using Content.Shared.Destructible;
 using Content.Shared.Popups;
 using JetBrains.Annotations;
-using Robust.Shared.Player;
 
 namespace Content.Server.Atmos.Piping.EntitySystems
 {
@@ -16,11 +16,12 @@ namespace Content.Server.Atmos.Piping.EntitySystems
     public sealed class AtmosUnsafeUnanchorSystem : EntitySystem
     {
         [Dependency] private readonly AtmosphereSystem _atmosphere = default!;
+        [Dependency] private readonly NodeGroupSystem _group = default!;
         [Dependency] private readonly PopupSystem _popup = default!;
 
         public override void Initialize()
         {
-            SubscribeLocalEvent<AtmosUnsafeUnanchorComponent, BeforeUnanchoredEvent>(OnBeforeUnanchored);
+            SubscribeLocalEvent<AtmosUnsafeUnanchorComponent, UserUnanchoredEvent>(OnUserUnanchored);
             SubscribeLocalEvent<AtmosUnsafeUnanchorComponent, UnanchorAttemptEvent>(OnUnanchorAttempt);
             SubscribeLocalEvent<AtmosUnsafeUnanchorComponent, BreakageEventArgs>(OnBreak);
         }
@@ -30,15 +31,12 @@ namespace Content.Server.Atmos.Piping.EntitySystems
             if (!component.Enabled || !EntityManager.TryGetComponent(uid, out NodeContainerComponent? nodes))
                 return;
 
-            if (_atmosphere.GetContainingMixture(uid, true) is not {} environment)
-                return;
-
             foreach (var node in nodes.Nodes.Values)
             {
                 if (node is not PipeNode pipe)
                     continue;
 
-                if (pipe.Air.Pressure - environment.Pressure > 2 * Atmospherics.OneAtmosphere)
+                if (pipe.Air.Pressure > Atmospherics.OneAtmosphere)
                 {
                     args.Delay += 2f;
                     _popup.PopupEntity(Loc.GetString("comp-atmos-unsafe-unanchor-warning"), pipe.Owner,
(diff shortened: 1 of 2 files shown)


Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
```diff
@@ -30,15 +31,12 @@ private void OnUnanchorAttempt(EntityUid uid, AtmosUnsafeUnanchorComponent compo
...
-            if (_atmosphere.GetContainingMixture(uid, true) is not {} environment)
-                return;
-
             foreach (var node in nodes.Nodes.Values)
             {
                 if (node is not PipeNode pipe)
                     continue;
 
-                if (pipe.Air.Pressure - environment.Pressure > 2 * Atmospherics.OneAtmosphere)
+                if (pipe.Air.Pressure > Atmospherics.OneAtmosphere)
```
Partmedia, line 39: Perhaps this check should stay the same -- because it's the pressure difference between pipe and environment that actually matters, and not the absolute pressures. The proposed change would sometimes warn when there is no pressure difference risk.

Also perhaps good to keep 2 ATM instead of 1.

Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
```diff
@@ -48,56 +46,54 @@ private void OnUnanchorAttempt(EntityUid uid, AtmosUnsafeUnanchorComponent compo
...
+            }
         }
 
         private void OnBreak(EntityUid uid, AtmosUnsafeUnanchorComponent component, BreakageEventArgs args)
         {
-            LeakGas(uid);
+            LeakGas(uid, false);
             // Can't use DoActsBehavior["Destruction"] in the same trigger because that would prevent us
             // from leaking. So we make up for this by queueing deletion here.
             QueueDel(uid);
```
Partmedia, line 67: Here the deletion is queued but not yet done -- so is the copy needed?

Partmedia (commented):
Thanks for the fix. Thanks especially for looking into past related work and giving important technical context. I have a few questions inline.

Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
slarticodefast: It is the absolute pressure that matters now, since that decides how many moles of gas are inside the pipe and will leak out when unanchored. This amount will be the same, no matter the outside pressure.

Changing it to 2 ATM would be reasonable.

Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
slarticodefast: What happens without the copy is this:
- The pipe is broken
- We leak gas from the pipe group, to which it is still connected until next tick
- This will reduce the amount of gas in the pipe group accordingly
- The pipe is deleted
- This code in PipeNet.cs runs, reducing the gas amount a second time
grafik
- this would lower the pressure in the remaining pipes, which should not happen

An alternative would be to do a group update here as well, but that would be really problematic performance-wise. for example when the nuke explodes a lot of pipes will break at once.

Partmedia: Understood, thanks.

Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
Partmedia: I see what you're saying. But I think the old intent was something akin to SS13's warning when unwrenching pipes because the space wind would fling you out if you didn't have magboots on. That is why unwrenching a pipe is "unsafe", not because it would leak gases.

Now, we don't have space wind or the flinging mechanic so that's perhaps that's a case to change the behavior of this warning. But I kind of do like the warning being based on pressure differences -- this is something you could actually theoretically feel when starting to unwrench such a pipe. Otherwise, you're just kind of magically sensing what's inside the pipe.

Up to you what you want to do about this.

Partmedia (approved):
Thanks for the fix. Up to you what you want to do with the pressure warning. Please go ahead and merge when you're finished.

Content.Server/Atmos/Piping/EntitySystems/AtmosUnsafeUnanchorSystem.cs
slarticodefast: Oh, I think I get your point. The warning is not about the amount of gas removed but about its impact on the environment it is realeased into. In vacuum 16 moles is a lot, wheras in a 4500kPA it barely changes the relative gas mix.

I will change it back later today.

slarticodefast:
I reverted the pressure warning popup change as discussed above.

PR #33781, opened 2024-12-08
Signal timer duration limit

Errant-4:
## About the PR
Signal Timers (for example Brig Timers on cells) now have a maximum duration. If any number higher than an hour is input, the countdown will start from 59:59 instead. Max duration is on a component and can be modified by individual prototypes, if necessary.

## Why / Balance
(Technically, there already was a practical maximum, since the largest number that could be typed in was 99m 99s, which becomes 1h 40m. I am merely reducing that limit to an hour instead)

if the current remaining time is over 1h 0m 0s, the display shows hh:mm. This is arguably unintuitive, since people expect it to show mm:ss, and can just look like a short timer got stuck, unless you stare at it long enough to see the minutes change.

There is not really a reasonable usecase for a timer counting down from over an hour. Should that need legitimately arise in the future, or in a fork, MaxDuration can be increased in yaml to the original. But it will still be capped by the limitation of the input field.

This change also makes it possible to set more limitations on specific timers through yaml. For example making it so all brig timers can no longer be set to above a universal maximum sentence time.

## Technical details
Maxduration is on the SignalTimerComponent. When the server receives the value that was input on the client, the lower of the two numbers is used to start the countdown.

## Requirements

**Changelog**

<!--
:cl: Errant
- tweak: Signal timers now have a maximum countdown duration, which can be specified in yaml. The default maximum is one hour, to prevent the displayed time from overflowing into hh:mm .

### DIFF
--- a/Content.Server/DeviceLinking/Components/SignalTimerComponent.cs
+++ b/Content.Server/DeviceLinking/Components/SignalTimerComponent.cs
@@ -1,53 +1,58 @@
 using Content.Shared.DeviceLinking;
 using Robust.Shared.Audio;
 using Robust.Shared.Prototypes;
-using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype;
 
 namespace Content.Server.DeviceLinking.Components;
 
 [RegisterComponent]
 public sealed partial class SignalTimerComponent : Component
 {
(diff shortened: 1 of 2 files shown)


chromiumboy:
Seems reasonable to me

Issue #27650, opened 2024-05-03
Lobby UI isnt as it supposed to be if the server's name is too long


lzk228:
https://github.com/space-wizards/space-station-14/assets/124214523/169e4bab-3508-4ce2-8d9e-b7cbac04c71a

PR #32313, opened 2024-09-19
Disable resizing of lobby character editor

eoineoineoin:
## About the PR

Real trivial bugfix. When attempting to close the character editor, I misclicked and accidentally resized the character editor.

Once the editor has been resized, the container remembers the sizes until application restart, and when the user attempts to manually restore the size, it makes a mess of the main lobby screen.

There's no need for these controls to be user-resizable anyway, so just disable that. 

## Media

Old behaviour; character editor remembers a bad size and is broken whenever you open it:

Lobby after you attempt to resize the character editor back to what it was previously:

Bonus, stops this "resizable" mouse cursor from appearing, even though the panel couldn't be resized:

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

PR #33783, opened 2024-12-08
Lobby chat width and custom lobby titles

Errant-4:
## About the PR
The width of the right side lobby panel is now fixed at 650* or it can be specified via cvar.
(650 was picked somewhat arbitrarily, between Lizard and Salamander. I'm not super attached to it, but it seemed like a reasonable default width.)

Servers can also set a custom title/message to be displayed as the Lobby title via cvar. If it's not specified, then the title will be taken from the server's hub name, similarly* to Live.
(The spacing/layout will theoretically look slightly different for servers with very short names, because the "Lobby" prefix is now part of the same string as the name) 

Having looked at this, ultimately we might want to redesign the layout of lobby chat, so that the vast majority of the screen space does not go wasted while you are actually using it and not basking in the beauty of the lobby art. Or make it draggable so the players can set it.
But quick fix first.

## Why
Currently, the width of the right side panel in the lobby is (in practice) governed by how long a server's name is. This means that basically every server has a slightly different lobby chat width unless their names just happens to be exactly the same lenght.
It turns out some forks already had to modify the layout code because otherwise their exceptionally long names/featurelist would make the chat take up most of the screen. 

The fixed width would now mean that a long server name does not fit onto the panel and runs out of the screen. But why should the lobby title need to be the same as the hub listing, anway? So it can now be specified.

## Live
NOTE! Do not compare the images to anything but each other, as your UI scale and to some degree even Resolution being different from mine will throw off your comparison

Screenshot_2024-12-08_143654
Screenshot_2024-12-08_150254
(Forgive a bit of necromancy, but the difference here is quite huge)
Screenshot_2024-12-08_150041
The logical conclusion of the unmodified code
Screenshot_2024-12-08_143827

## Fixed length at 650 
The same servers with fixed width, no configuration:
Screenshot_2024-12-08_144341
Screenshot_2024-12-08_144644
Screenshot_2024-12-08_181818
Screenshot_2024-12-08_144959

With the lobby_name cvar, servers can fully specify the lobby title as desired.
Screenshot_2024-12-08_144758
Screenshot_2024-12-08_144839
Screenshot_2024-12-08_145104
Screenshot_2024-12-08_145545

## Requirements

## Breaking changes
The chat panel's width in the lobby is now fixed. Server names that are longer than the default width will not fit onto the screen and the end will be cut off. Use your server config file to either increase the width or specify a shorter lobby title to be displayed.
[server] 
lobby_right_panel_width = 675
lobby_name = "Lobby MyServer"

**Changelog**
:cl: Errant
- tweak: Lobby chat width is no longer dependent on the server's name.
- tweak: Lobby title of a server can now be different from the Hub listing's name.

### DIFF
--- a/Content.Client/Lobby/LobbyState.cs
+++ b/Content.Client/Lobby/LobbyState.cs
@@ -60,8 +60,7 @@ namespace Content.Client.Lobby
                 : lobbyNameCvar;
 
             var width = _cfg.GetCVar(CCVars.ServerLobbyRightPanelWidth);
-            Lobby.RightSide.MinWidth = width;
-            Lobby.RightSide.MaxWidth = width;
+            Lobby.RightSide.SetWidth = width;
 
             UpdateLobbyUi();
 
--- a/Content.Client/Lobby/LobbyState.cs
+++ b/Content.Client/Lobby/LobbyState.cs
@@ -54,8 +54,10 @@ namespace Content.Client.Lobby
             LayoutContainer.SetAnchorPreset(Lobby, LayoutContainer.LayoutPreset.Wide);
 
             var lobbyNameCvar = _cfg.GetCVar(CCVars.ServerLobbyName);
+            var serverName = _baseClient.GameInfo?.ServerName ?? "";
+
             Lobby.ServerName.Text = string.IsNullOrEmpty(lobbyNameCvar)
-                ? Loc.GetString("ui-lobby-title") + " "  + _baseClient.GameInfo?.ServerName
+                ? Loc.GetString("ui-lobby-title", ("serverName", serverName))
                 : lobbyNameCvar;
 
             var width = _cfg.GetCVar(CCVars.ServerLobbyRightPanelWidth);
--- a/Content.Client/Lobby/LobbyState.cs
+++ b/Content.Client/Lobby/LobbyState.cs
@@ -5,11 +5,13 @@ using Content.Client.Lobby.UI;
 using Content.Client.Message;
 using Content.Client.UserInterface.Systems.Chat;
 using Content.Client.Voting;
+using Content.Shared.CCVar;
 using Robust.Client;
 using Robust.Client.Console;
 using Robust.Client.ResourceManagement;
 using Robust.Client.UserInterface;
 using Robust.Client.UserInterface.Controls;
+using Robust.Shared.Configuration;
 using Robust.Shared.Timing;
 
 
@@ -18,6 +20,7 @@ namespace Content.Client.Lobby
     public sealed class LobbyState : Robust.Client.State.State
     {
         [Dependency] private readonly IBaseClient _baseClient = default!;
+        [Dependency] private readonly IConfigurationManager _cfg = default!;
         [Dependency] private readonly IClientConsoleHost _consoleHost = default!;
         [Dependency] private readonly IEntityManager _entityManager = default!;
         [Dependency] private readonly IResourceCache _resourceCache = default!;
@@ -49,7 +52,12 @@ namespace Content.Client.Lobby
 
             _voteManager.SetPopupContainer(Lobby.VoteContainer);
             LayoutContainer.SetAnchorPreset(Lobby, LayoutContainer.LayoutPreset.Wide);
-            Lobby.ServerName.Text = _baseClient.GameInfo?.ServerName; //The eye of refactor gazes upon you...
+
+            var serverLobbyName = _cfg.GetCVar(CCVars.ServerLobbyName);
+            Lobby.ServerName.Text = string.IsNullOrEmpty(serverLobbyName)
+                ? Loc.GetString("ui-lobby-title") + " "  + _baseClient.GameInfo?.ServerName
+                : serverLobbyName;
+
             UpdateLobbyUi();
 
             Lobby.CharacterPreview.CharacterSetupButton.OnPressed += OnSetupPressed;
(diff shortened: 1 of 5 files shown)


lzk228:
i think is fixes #27650 
does it?

lzk228:
well basically yeah, it was "fixed" after https://github.com/space-wizards/space-station-14/pull/32313
but the thing that bothers me is you still can't see chat and character editor if chat is too wide

Content.Client/Lobby/LobbyState.cs
```diff
@@ -49,7 +52,16 @@ protected override void Startup()
 
             _voteManager.SetPopupContainer(Lobby.VoteContainer);
             LayoutContainer.SetAnchorPreset(Lobby, LayoutContainer.LayoutPreset.Wide);
-            Lobby.ServerName.Text = _baseClient.GameInfo?.ServerName; //The eye of refactor gazes upon you...
+
+            var lobbyNameCvar = _cfg.GetCVar(CCVars.ServerLobbyName);
+            Lobby.ServerName.Text = string.IsNullOrEmpty(lobbyNameCvar)
+                ? Loc.GetString("ui-lobby-title") + " "  + _baseClient.GameInfo?.ServerName
```
Aeshus, line 58: Shouldn't the concatenation be done in the localization string instead of the order being hard-coded?

Errant-4: I did not think a lot about that part or ever look at it again, because I don't imagine many servers will use the default. But yeah, fair. point

Content.Shared/CCVar/CCVars.Server.cs
```diff
@@ -40,4 +40,17 @@ public sealed partial class CCVars
...
+    /// <summary>
+    ///     This will be the title shown in the lobby
+    ///     If empty, the title will be {ui-lobby-title} + the server's full name from the hub
+    /// </summary>
+    public static readonly CVarDef<string> ServerLobbyName =
+        CVarDef.Create("server.lobby_name", "", CVar.REPLICATED | CVar.SERVER);
+
+    // <summary>
+    //  The width of the right side (chat) panel in the lobby
+    // </summary>
```
Aeshus, line 53: This comment's alignment is a little funky, plus you forgot to make it tripple slashes to make it a doc comment.

```suggestion
    /// <summary>
    ///     The width of the right side (chat) panel in the lobby
    /// </summary>
```

Content.Client/Lobby/LobbyState.cs
```diff
@@ -5,11 +5,13 @@
...
 using Content.Client.UserInterface.Systems.Chat;
 using Content.Client.Voting;
+using Content.Shared.CCVar;
 using Robust.Client;
 using Robust.Client.Console;
 using Robust.Client.ResourceManagement;
 using Robust.Client.UserInterface;
 using Robust.Client.UserInterface.Controls;
+using Robust.Shared.Configuration;
 using Robust.Shared.Timing;
```
Aeshus, line 17: I'm not sure why there's an extra newline here.

```suggestion
```

Content.Client/Lobby/LobbyState.cs
```diff
@@ -49,7 +52,18 @@ protected override void Startup()
 
             _voteManager.SetPopupContainer(Lobby.VoteContainer);
             LayoutContainer.SetAnchorPreset(Lobby, LayoutContainer.LayoutPreset.Wide);
-            Lobby.ServerName.Text = _baseClient.GameInfo?.ServerName; //The eye of refactor gazes upon you...
+
+            var lobbyNameCvar = _cfg.GetCVar(CCVars.ServerLobbyName);
+            var serverName = _baseClient.GameInfo?.ServerName ?? "";
```
Aeshus, line 57: If I recall correctly, both are fine, though I think I see `string.Empty` _slightly_ more.

```suggestion
            var serverName = _baseClient.GameInfo?.ServerName ?? string.Empty;
```

Content.Client/Lobby/LobbyState.cs
```diff
@@ -49,7 +52,18 @@ protected override void Startup()
...
+
+            Lobby.ServerName.Text = string.IsNullOrEmpty(lobbyNameCvar)
+                ? Loc.GetString("ui-lobby-title", ("serverName", serverName))
+                : lobbyNameCvar;
+
+            var width = _cfg.GetCVar(CCVars.ServerLobbyRightPanelWidth);
+            Lobby.RightSide.MinWidth = width;
+            Lobby.RightSide.MaxWidth = width;
```
Aeshus, line 65: This kinda feels like what `SetWidth` is for?

Aeshus (changes requested):
Thanks for addressing the changes I requested!

I just have some nitpicks and a question about if we should be using `SetWidth` instead of doing both `MinWidth` and `MaxWidth`.

Content.Client/Lobby/LobbyState.cs
Errant-4: I suppose it does look more proper

Content.Client/Lobby/LobbyState.cs
Errant-4: Good catch, I only checked if there is a .Width, and then settled for these. Thanks!

Aeshus (approved):
_LGTM_

(I've always wanted to say that)

Resources/Locale/en-US/lobby/lobby-gui.ftl
```diff
@@ -1,4 +1,4 @@
-ui-lobby-title = Lobby
+ui-lobby-title = Lobby {$serverName}
```
slarticodefast, line 1:
```suggestion
ui-lobby-title = Lobby: {$serverName}
```
Since there is no more whitespace between the "Lobby" and the server name as there was before, I'm adding a colon for better readability.

slarticodefast (approved):

slarticodefast:
Looks good to me.
Thanks to Aeshus for the review!

PR #33790, opened 2024-12-09
add "Changes: Audio" label to labeler

slarticodefast:
See title. We got the label already, but it is not applied by the labeler yet.

I have no idea how to best test this before merging, but the code is simple enough.


SlamBamActionman:
You can test labeler PRs by merging them into your own fork and making test PRs to the fork. Might need to enable the workflow first though, but it should even work for branches if you want to avoid messing with your master.

I tested it and it looks like it works! Nice!
image

PR #33792, opened 2024-12-09
Logic Gate Compatibility with Non-Logic Signals bugfix

### LINKED ISSUE #33716: Logic gates don't work correctly with signal buttons
## Description
They work properly with signal switches or door states, but not with pulsed signals.

**Reproduction**
grafik
grafik

- Spawn two signal buttons, an or-gate and a wall light
- link the buttuns to inputs A and B of the or gate
- link the output of the or gate to the toggle of wall light.
- Press button A: The light flashes for one tick.
- Press button B: The light will stay one.

`**Expected behaviour**
The output should be symmetric for both inputs for an or gate. This might affect other gates as well, but I did not test this.

Centronias:
## About the PR
Fixed a bug where B port inputs to logic gates weren't properly reset to `Low` when a non-logic signal is sent to them.
ie. if you link a Signaller to a Logic Gate's B input, it gets stuck as `High` forever.
Fixes #33716

## Why / Balance
I just spent a whole round trying to make a thing and only once I hooked up the debugger did I understand ):

## Technical details
`High` -> `Low`
:)

## Media
Very small and nothing really to show

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- fix: Fixed a bug where the B input on logic gates would get stuck in a high state

### DIFF
--- a/Content.Server/DeviceLinking/Systems/LogicGateSystem.cs
+++ b/Content.Server/DeviceLinking/Systems/LogicGateSystem.cs
@@ -44,7 +44,7 @@ public sealed class LogicGateSystem : EntitySystem
             }
             if (comp.StateB == SignalState.Momentary)
             {
-                comp.StateB = SignalState.High;
+                comp.StateB = SignalState.Low;
             }
 
             // output most likely changed so update it


beck-thompson:
I TOTALLY had this bug happen when trying to make a flip flop. Really good catch I never even realized...

slarticodefast:
Fixes #33716

slarticodefast (approved):
Thanks for the fix!

PR #33794, opened 2024-12-09
Fix 38 non-obsolete warnings

SpaceManiac:
## About the PR
Pursuant to https://github.com/space-wizards/space-station-14/issues/33279.

## Technical details
Fix:
- Fix missing base call in `ClientGameTicker.Initialize` by calling base, and updating base to not apply `roundId: 0` to client-side recordings
- Fix missing base call in `CableVisualizerSystem.Initialize` by detaching it from `VisualizerSystem` entirely
- Remove unused private variables
- Fix private/internal starts-with-underscore naming rule violations
- Remove duplicate dependency on `IAdminLogManager` in `GhostSystem`
- Remove unused redeclaration without use of `new` in `InstrumentBoundUserInterface`
- Remove never-emitted events in `DockObject` UI and their never-called listeners
- Await sidelined `Task`s in a try-catch and log errors
- Remove unused `using`s, but only in files already touched
 
Silence:
- Silence missing base call in `DoAfterSystem.Update` since it seems intentional

This doesn't fix any warning with "obsolete" in the name, RA0026, or these monsters:

> .../space-station-14/BuildChecker/BuildChecker.csproj : warning NU1503: Skipping restore for project '.../space-station-14/BuildChecker/BuildChecker.csproj'. The project file may be invalid or missing targets required for restore. [.../space-station-14/SpaceStation14.sln]
>
> /usr/share/dotnet/sdk/8.0.110/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(284,5): warning NETSDK1206: Found version-specific or distribution-specific runtime identifier(s): debian-x64, fedora-x64, opensuse-x64, rhel-x64, win7-arm, win7-x64, win7-x86. Affected libraries: Libuv. In .NET 8.0 and higher, assets for version-specific and distribution-specific runtime identifiers will not be found by default. See https://aka.ms/dotnet/rid-usage for details. [.../space-station-14/Pow3r/Pow3r.csproj]
>
> CSC : warning AD0001: Analyzer 'NUnit.Analyzers.DisposeFieldsInTearDown.DisposeFieldsAndPropertiesInTearDownAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. [.../space-station-14/Content.IntegrationTests/Content.IntegrationTests.csproj]

## Requirements


Content.Client/Instruments/UI/InstrumentBoundUserInterface.cs
```diff
@@ -13,7 +12,7 @@ public sealed class InstrumentBoundUserInterface : BoundUserInterface
         public IEntityManager Entities => EntMan;
         [Dependency] public readonly IMidiManager MidiManager = default!;
         [Dependency] public readonly IFileDialogManager FileDialogManager = default!;
-        [Dependency] public readonly IPlayerManager PlayerManager = default!;
+        [Dependency] public new readonly IPlayerManager PlayerManager = default!;
```
PJB3005, line 15: Can't you just remove this line entirely?

Content.Client/GameTicking/Managers/ClientGameTicker.cs
```diff
@@ -43,6 +43,8 @@ public sealed class ClientGameTicker : SharedGameTicker
 
         public override void Initialize()
         {
+            base.Initialize();
```
PJB3005, line 46: Looking at the code, this is gonna make client-side replays have invalid 0 round IDs. Should probably guard against that?

Content.Client/Shuttles/UI/DockObject.xaml.cs
```diff
@@ -1,23 +1,22 @@
...
 [GenerateTypedNameReferences]
 public sealed partial class DockObject : PanelContainer
 {
+#pragma warning disable CS0067
     [PublicAPI]
     public event Action? UndockPressed;
 
     [PublicAPI]
     public event Action? ViewPressed;
+#pragma warning restore CS0067
```
PJB3005, line 19: Should these be kept around? There's nothing actually calling these so surely the warning is valid and it shouldn't just be suppressed?

Content.Server/Salvage/SalvageSystem.Magnet.cs
```diff
@@ -45,7 +45,7 @@ private void OnMagnetClaim(EntityUid uid, SalvageMagnetComponent component, ref
             return;
         }
 
-        TakeMagnetOffer((station.Value, dataComp), args.Index, (uid, component));
+        _ = TakeMagnetOffer((station.Value, dataComp), args.Index, (uid, component));
```
PJB3005, line 48: Discarding this task is not the right choice, it'll cause the error result to be poorly handled.

Content.Server/Speech/EntitySystems/ReplacementAccentSystem.cs
```diff
@@ -62,6 +62,7 @@ public string ApplyReplacements(string message, string accent)
                 // this is kind of slow but its not that bad
                 // essentially: go over all matches, try to match capitalization where possible, then replace
                 // rather than using regex.replace
+#pragma warning disable RA0026 // Use of uncached static Regex function
```
PJB3005, line 65: This warning is valid, suppressing it is like this will cause it to go unnoticed rather than getting it fixed properly.

This code is *really bad* and likely causes tons of unnecessary regex parsings, they need to be cached.

Content.Shared/Chat/SharedChatSystem.cs
```diff
@@ -275,7 +275,9 @@ public static string InjectTagInsideTag(ChatMessage message, string outerTag, st
     public static string InjectTagAroundString(ChatMessage message, string targetString, string tag, string? tagParameter)
     {
         var rawmsg = message.WrappedMessage;
+#pragma warning disable RA0026 // Use of uncached static Regex function
         rawmsg = Regex.Replace(rawmsg, "(?i)(" + targetString + ")(?-i)(?![^[]*])", $"[{tag}={tagParameter}]$1[/{tag}]");
+#pragma warning restore RA0026 // Use of uncached static Regex function
```
PJB3005, line 280: Again, this is actually a valid warning that should have its cause fixed, not be suppressed.

Content.Client/Instruments/UI/InstrumentBoundUserInterface.cs
SpaceManiac: On the presumption it was used, `IPlayerManager` is more specific than the parent `ISharedPlayerManager`, but looks like it's actually not used, so sure.

Content.Client/Shuttles/UI/DockObject.xaml.cs
SpaceManiac: I went off the fact they already had `[PublicAPI]` with an apparent meaning of "please do not remove this", but I'm happy to trash them anyway on your say-so.

Content.Client/GameTicking/Managers/ClientGameTicker.cs
SpaceManiac: `replay.yml` has `roundId: 0` instead of not having it at all. Seems harmless to me, but I added a `!= 0` guard before writing it to be on the safe side.

Content.Server/Salvage/SalvageSystem.Magnet.cs
SpaceManiac: Would catch-all-and-log suffice? Seems like it's too late to do anything else of value.

PJB3005: > Would catch-all-and-log suffice? Seems like it's too late to do anything else of value.

Yeah, preferably.

SpaceManiac: Great, done! https://github.com/space-wizards/space-station-14/pull/33794/commits/487edfc6d8cfb58d720f0d49da6adc1deb31cc51. Annotations there are fixed already, remaining red X is a spurious test failure ("system.activatable_u_i: Missing UI Key for entity: alien artifact (64056/n67622, MediumXenoArtifactItem)").

PJB3005 (approved):

PR #33796, opened 2024-12-09
Allow map-editor admin-ghosts to toggle doors

SpaceManiac:
## About the PR
Admin ghosts clicking doors on paused maps (map editor) now actually opens them instead of getting them stuck on the first frame of the opening animation for all time.

## Why / Balance
Admin ghosts have the `InstantDoAfters` tag and so can force open doors instantly even on paused maps. The doors then enter the "Opening" state but never go all the way to "Open", don't animate, can't be closed again, and start open when the map is loaded. It's better if they open properly and can be closed again too.

## Technical details
Skip animation if map is paused and doer has `InstantDoAfters` tag, which only aghosts do.

## Requirements

### DIFF
--- a/Content.Shared/Doors/Systems/SharedDoorSystem.cs
+++ b/Content.Shared/Doors/Systems/SharedDoorSystem.cs
@@ -45,7 +45,6 @@ public abstract partial class SharedDoorSystem : EntitySystem
     [Dependency] protected readonly SharedPopupSystem Popup = default!;
     [Dependency] private readonly SharedMapSystem _mapSystem = default!;
     [Dependency] private readonly SharedPowerReceiverSystem _powerReceiver = default!;
-    [Dependency] private readonly TagSystem _tag = default!;
 
 
     [ValidatePrototypeId<TagPrototype>]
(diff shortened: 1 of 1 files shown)


IProduceWidgets:
I don't know if this is a question for you so much as it is for @metalgearsloth but is there a reason we keep using paused as a proxy for mapping when really (map)init seems much more appropriate? I know this also happens for some other random stuff like device linking.

SpaceManiac:
At least in this case, I think paused is appropriate since the effect is to skip the animation that won't play due to being paused. Admin ghosts toggling doors on a paused post-init map is hypothetically useful.

MilonPL:
Thank you for your PR! After a bit of discussion with the other maintainers, we've decided to close this PR, as it doesn't meet our code standards. 
Checking for both if an entity is paused and if the user has a specific tag is a very "hacky" solution. Doors just shouldn't be allowed to be opened in the paused state, and if anything, this would be better as its own verb allowing you to set the door state.

SpaceManiac:
"Paused" is the natural state that indicates an entity's map is being edited. "InstantDoAfters" is arguably a bad *name*, but it unambiguously identifies aghosts. Would adding a new "MapEditor" tag to aghosts be better? In the meantime, this bug is still trolling mappers.

IProduceWidgets:
> Thank you for your PR! After a bit of discussion with the other maintainers, we've decided to close this PR, as it doesn't meet our code standards. Checking for both if an entity is paused and if the user has a specific tag is a very "hacky" solution. Doors just shouldn't be allowed to be opened in the paused state, and if anything, this would be better as its own verb allowing you to set the door state.

Hmm, gotta disagree here,
We need to be able to map open doors for the non-airlock variants such as wooden doors.

PR #33805, opened 2024-12-09
Add SetPrice to PricingSystem

lzk228:
## About the PR

add a method for setting price to given entity

## Why / Balance

i need it for my next pr and i think it's a pretty useful method 

## Technical details

not a lot, just EnsureComp and change set the price of it to given in method call

## Requirements

## Breaking changes

no

**Changelog**

no cl no fun

### DIFF
--- a/Content.Server/Cargo/Systems/PricingSystem.cs
+++ b/Content.Server/Cargo/Systems/PricingSystem.cs
@@ -218,7 +218,7 @@ public sealed class PricingSystem : EntitySystem
     /// <param name="price">The price that will be given.</param>
     public void SetPrice(EntityUid uid, double price)
     {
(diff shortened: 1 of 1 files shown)


Partmedia:
I'm concerned that this might not always do what you want, for example, when you run into an item which is really in a stack and uses StackPriceComponent. Can you discuss the other cases and why this should ignore that?

slarticodefast:
Yeah, I mentioned this on Dicord already, but if this function is used in unintended ways that might cause more exploits that are not caught from the integration tests because those run only considering the original prototypes.
For example the function in this PR adds the StaticPriceComponent to any entity, which could violate the NoStaticPriceAndStackPrice test.

lzk228:
isn't needed anymore

PR #33808, opened 2024-12-10
Fix: SMES now has wirepanels visuals

ArtisticRoomba:
## About the PR
SMES now has visuals for if the maintenance panel is open.

## Why / Balance
SMES had wirepanel sprites but they were never used because missing code in YAML.
Something I noticed when I was working on #33757 (that PR needs this one to be merged before, or else it'll look weird)

## Technical details
Added like 3 lines of YAML. It was missing the `WiresVisuals` type.
## Media
image

## Requirements

**Changelog**
:cl:
- fix: SMES now gives visual feedback for if the maintenance panel is open.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Power/smes.yml
+++ b/Resources/Prototypes/Entities/Structures/Power/smes.yml
@@ -28,6 +28,8 @@
         - map: ["enum.SmesVisualLayers.Output"]
           state: "smes-op1"
           shader: unshaded
+        - map: [ "enum.WiresVisualLayers.MaintenancePanel" ]
+          state: smes-open
     - type: Smes
     - type: Appearance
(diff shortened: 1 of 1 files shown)


Aeshus (approved):
Works as advertised, thanks!

image

Tangent: I'm not totally sure I like the current sprite as it blocks the most distinctive part of the SMES -- the lights (plus the panel blends in easily with the rest of the SMES)! Maybe in the future we can also give some functionality to the wires so the panel makes sense to open, but both of those are outside the scope of this PR.

Tayrtahn (approved):

slarticodefast (approved):

slarticodefast:
Thanks for the fix!

PR #33810, opened 2024-12-10
Made anchor visuals generic

TGRCdev:
## About the PR

This PR adds the `AnchorVisuals.Anchored` appearance data enum, which stores an entity's anchored status for use with `GenericVisualizer`.

## Why / Balance

There was no way to define anchor-state-dependent sprite states in an entity without writing a bespoke component. I implemented this for downstream content, and I wanted to offer the functionality upstream.

## Technical details

This is intended to allow entities to change their anchored appearance using the `GenericVisualizer` component. When an entity has the `Anchorable` component, their anchor state will be stored in their appearance data under the key `AnchorVisuals.Anchored`. This state can then be retrieved from the `GenericVisualizer` to implement anchor-dependent sprite states without a bespoke component.

I used this functionality to replace the `GasPortableVisuals` appearance enum, which was functionally identical, but intended specifically for anchoring gas canisters to connector ports.

```yml
- type: Anchorable
- type: GenericVisualizer
  visuals:
    enum.AnchorVisuals.Anchored:
      connectedToPort:
        False: { state: can-connector, visible: false }
        True: { state: can-connector, visible: true }
```

## Media

[Screencast_20241210_010947.webm](https://github.com/user-attachments/assets/08886e73-6918-4c03-b07f-de631451ea22)

## Requirements

## Breaking changes

`GasPortableVisuals.ConnectedState` has been removed and replaced with the generic `AnchorVisuals.Anchored` enum.

**Changelog**


metalgearsloth (approved):

Aeshus (approved):
Seems to work as intended, thanks!

I was wondering if this would also be applicable to the nuke and its `NukeVisuals`, which serves a similar purpose (except it does weird stuff outside of the anchoring system).

(Guh, sniped my sloth)

TGRCdev:
I could certainly check! I'm not aware of all the places in the code that have implemented this specific functionality, but it'll be a good opportunity to start looking.

TGRCdev:
From a quick search, looks like the nuke, disposal units, and nav beacons could be simplified to use this. More commits coming.

Aeshus:
I would probably just do those in a second PR as this one is pretty focused and that will just delay this being merged.

TGRCdev:
Good point. In that case, everything is ready to merge.

slarticodefast (approved):
Works as intended. Thank you for your contribution!

PR #33816, opened 2024-12-10
Audio mixers

### LINKED ISSUE #22710: Audio volume sliders should be using auxiliary
This is more or less what it's for, means we don't need to store gain ourselves and just set it once on aux, also makes it easier to update it live.

### LINKED ISSUE #27981: Request: Jukebox volume slider
## Description
The Jukebox is currently tied directly to Master Volume. 
There are multiple better options available:
Ambient music volume, Midi (Instrument) volume or a dedicated volume slider (clientside in jukebox ui)

The jukebox currently overpowers most ANY audio in the game and is to the point of being downright deafening when close enough to it (aka being a barkeep). 
Without an option to change the volume locally it turns into IC arguments that end up in LOOC of 
1: Turn the music off its too loud
2: no I like the song
1: its so loud
2: so leave
1: looc: man its SO loud please
etc, etc.

**Reproduction**
n/a

**Screenshots**
n/a

**Additional context**
n/a

stalengd:
## About the PR
Adds the use of audio mixers for in-game audio. Read the details about mixers in the [PR to Robust Toolbox](https://github.com/space-wizards/RobustToolbox/pull/5554).

There are the proposed mixers (`Resources/Prototypes/audio_mixers.yml`):

- `Music` [No volume slider] (let me know if slider is really needed)
  - `Ambience`
  - `AmbienceMusic`
  - `LobbyMusic`
- `Interface`
- `WorldSounds`
  - `WorldOtherSounds` [Default]
  - `Midi`
  - `Jukebox`

The child mixers volume sliders are visually indented (see media)

## Why / Balance
- fix #22710
- fix #27981
- separates all sounds previously adjustable only via master volume slider to new slider

## Technical details
Requires https://github.com/space-wizards/RobustToolbox/pull/5554

We should probably start assigning more specific mixers to the different sounds (for example separate mixer for weapon sounds) after this PR, this can be done by passing mixer id to AudioParams.

### Important: Entity-based mixers from prototypes problem
*This also is mostly copy-pasted from corresponding part in the PR to Robust Toolbox, so you will not find anything new if you read that one.*

**TL;DR:** Either one integration test should be adjusted/ignored, or I need help to come up with a different way to handle entity-based mixers from prototypes. 

The concept of "prototyped" mixers conflicts with entity-based mixers in terms of lifetimes. 

My first implementation of this was lazily creating mixers for prototype when requested, newly created mixer then stored in dictionary indefinitely. Unfortunately, some SS14 integration tests does not like this idea when any entity creates more other entities than deletes after got deleted.

This implementation creates all mixer entities from prototypes at start of the simulation (check out the funny way how system knows when its time to spawn entities) and when entities flushed. Set of such entities exists separately on client and server though because single-player games may exist and (probably) entities from server may not be sent to the client fast enough. The problem with this implementation is that other SS14 integration test (DeleteAllThenGhost this time) can't handle that deleting all entities on the server leaves mixer entities on client.

So this problem is relevant and PR probably should not be merged like this, either mixers should be implemented differently (but I don't know how at this point) or SS14 integration tests should be adjusted.

## Media
image

## Requirements

## Breaking changes

None (I hope)

**Changelog**

:cl:
- add: Added "world sounds", "other sounds" and "jukebox" volume sliders to the game settings menu.


AJCM-git:
@stalengd  Still interested in getting this in? I can handle this PR if you fix conflicts

PR #33829, opened 2024-12-11
Nerf Nocturine 

lanse12:
## About the PR
Increased the minimum threshold for nocturine to put someone to sleep from 8 to 11
up from 2 hypopen injects to 3 to put someone to sleep

## Why / Balance
Most if not all usages of nocturine are extremely braindead with it being a free kill (minibomb, emagged recycler or just throw into space) so I've decided to make it require more skill to use. With it now requiring 3 injects you will be forced to refill instead of just brainlessly clicking twice on the person you want dead.

## Technical details
yaml change

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Nocturine now needs 11u to put someone to sleep.


vorkathbruh:
this combo is already prohibitively expensive for traitors to get and do much else with. 19u requirement also means you need to inject all 4 times within 2 seconds which with a hypopen may be reasonably difficult because you will need to refill it inbetween uses

anyway, while i do think noct + hypo is bad game design it also pretty much requires an isolated target so there's that

ArtisticRoomba:
This seems like a webedit. Have you done testing in game to assure your changes have the desired effect?

My concern is that 19u to sleep is very, very much pushing it. By the time the second set of injections hit, the person might have already metabolized enough to the point where they're below the threshold and aren't slept.

There's also the fact that buying both a hypopen *and* nocturine is quite expensive, and limits your actions after that. It's a large commitment, and there are plenty of ways for people to mess up, not even considering your changes. Nocturine + hypopen leaves you with 8TC, which is *barely* enough for either an EMAG or a minibomb (points you highlighted in your reasoning).

Personally, I would be very happy if I knew that someone just used all of their TC for the sole purpose of round removing me and me alone.

I would also change your CL to properly state the amount of units required to sleep.

Djungelskog2:
While I agree noct is powerful, I dont think it needs to be nerfed *this* badly.

deltanedas:
it takes 0.5s absolute minimum to inject someone with 10u because of the hypopen use delay
this leaves you with a grand total of 500ms to:
- have good ping
- swap hands to your nocturine soda can
- move your mouse to the hypopen hand
- click it
- swap hands again
- move your mouse to the now-running target
- try to double tap again

you have not even tried this

Stealthbomber16:
This is a prohibitively awful change because it changes the amount of sleeps you can get from one bottle of nocturine from three to one.

SlimmSlamm:
This quite literally sounds like you died to this and made a PR to nerf it. This is already an extremely expensive combination to buy and nerfing it will make it absolutely useless.

Resources/Prototypes/Reagents/narcotics.yml
```diff
@@ -300,7 +300,7 @@
         conditions:
         - !type:ReagentThreshold
           reagent: Nocturine
-          min: 8
+          min: 19
```
Admiral-Obvious-001, line 303: Changing this to 11 will achieve your desired effect of making the user use 3 clicks without totally ruining Noct.

Admiral-Obvious-001 (changes requested):
Mentioned it in the line. Changing noct to 11u forced sleep will achieve the desired effect you want out of it, requiring 3 jabs without completely neutering the item. (I think Noct is bad design as is, but making it 11u will make it work in a syringe too.)

BramvanZijp:
I feel like nocturine itself isnt the issue here. Hypopen+Noct+Minibomb is a VERY expensive syndicate purchase which leaves you with almost nothing to do anything else, access to space is usually in tricky positions so you have to bait them there and make sure you dont get spotted since you only get 8tc to stay alive & uncaptured, and an emagged reycler gibbing somebody is its own balance problem, so I agree that this is most likely a webedit or atleast poorly thought out balance change from somebody who got screwed over by this combo.

If youd be willing to provide a video demonstrating that the combo is still somewhat viable after your nerfs, that may change my opinion, but I believe noct is fine as is.

EthanQix:
This is clearly a mald PR. Can we close this already ?

Resources/Prototypes/Reagents/narcotics.yml
lanse12: > Changing this to 11 will achieve your desired effect of making the user use 3 clicks without totally ruining Noct.

good idea man

BramvanZijp:
Isnt this literally just a webedit?

lanse12:
> Isnt this literally just a webedit?

what do u think it is

deltanedas:
is it a webedit guys i cant tell???

GroggleG:
> This is clearly a mald PR. Can we close this already ?

The balanceguys are what have always killed the creativity of this game. Making chloral hydrate make you 'drowsy' was already a bit drab but this is silly.

BramvanZijp:
> > Isnt this literally just a webedit?
> 
> what do u think it is

You realise those are not allowed right?

superjj18:
> > > Isnt this literally just a webedit?
> > 
> > 
> > what do u think it is
> 
> You realise those are not allowed right?

No opinion on the change but it’s like a 2 character change dude

BramvanZijp:
> > > > Isnt this literally just a webedit?
> > > 
> > > 
> > > what do u think it is
> > 
> > 
> > You realise those are not allowed right?
> 
> No opinion on the change but it’s like a 2 character change dude

Still gotta make & test it ingame

TheShuEd:
While web-edit is considered a bad thing, I believe that much of the negativity towards it is drawn by the discontent of Nocturine's weakening.  :saluting_face: 

Regarding this PR, and in general the balance of nocturine in general. Nocturine is in a very acute position. Yes, it requires high skill on the part of the agent to take advantage of this combo, but the problem is that this combo has no means of defense on the target.

If the agent is robust enough, the victim has no chance or means of escaping the attack. 
This pr weakens this combo even more, raising the bar on skill requirements, which essentially kills this combo, instead of weakening it from the other side - adding some mechanics with which Nocturne could be countered.

TheShuEd:
I don't think this PR should be merged, as it doesn't solve the gameplay design problem

TheShuEd:
Speaking of how this problem could be solved, here's an idea: We can increase the time before falling asleep. After the injection, the target has 10-15 seconds to react, scream, call for help, or run out of maints before the nocturine takes effect. Nocturine can also put you to sleep “gradually”, giving more hints that you are falling asleep. Screen wiggling, “your eyelids are getting tired” popups, slowing down.

deltanedas:
the webedit just means its wasnt tried out at all to see how useless nocturine became

BramvanZijp:
> the webedit just means its wasnt tried out at all to see how useless nocturine became

Nor did they try to apply it on even a static dummy after the changes, at which point they would have realised this is just not a good change.

cohanna:
severely diminishes the use of the hypopen (an already fairly niche item). The game as it stands now has very few (if any) worthwhile lethal reagents that can be utilized in a discrete manner with noct being the prevailing hostile reagent available to traitors (and thieves) due to its synergy with other items. While reagents like heartbreaker exists, it near useless against any experienced player (its damn antidote is given to EVERYONE roundstart) and as a result is hardly ever used. The main usage of chemical warfare currently (outside of noct) is for the cmo to put muta+pax into their hypo and go ham, but that hardly contributes to standard traitor gameplay.

Not every kill method NEEDS the ability for clear counterplay. I really feel the only reason reagents get the short stick is because: 

A: we don't have surgery or a relatively accessible means to revive after 200 poison dmg

B: there isn't a bullet. 

Noct kills are no more difficult than buying a viper and gunning someone down and is just as respondable if you aren't expecting it. 
Both of the "unfair/unfun" kill methods listed in this PR are just as easily achieved with replacing noct with any other firearm. 
But you don't see anyone going "oh the counterplay to getting shot sucks!"

TLDR: chemicals hardly need a nerf. People just don't like losing.

SlamBamActionman:
> Speaking of how this problem could be solved, here's an idea: We can increase the time before falling asleep. After the injection, the target has 10-15 seconds to react, scream, call for help, or run out of maints before the nocturine takes effect. Nocturine can also put you to sleep “gradually”, giving more hints that you are falling asleep. Screen wiggling, “your eyelids are getting tired” popups, slowing down.

I agree with this. 10 seconds should be enough, and we could even add a small speed decrease so that you aren't able to put too much distance away from the injection site by just sprinting.

Anyways, the reason why Noct is at 8 is because the pen is at 10, giving you approximately a 4 second window. Refilling the pen should not need to be your concern when you are using this strategy; it just makes the skill expression into "can you juggle the UI fast enough?" which isn't interesting.

SlamBamActionman:
> Noct kills are no more difficult than buying a viper and gunning someone down and is just as respondable if you aren't expecting it. Both of the "unfair/unfun" kill methods listed in this PR are just as easily achieved with replacing noct with any other firearm. But you don't see anyone going "oh the counterplay to getting shot sucks!"

I think it is worth considering that, by and large, chemicals are silent. Nocturine is almost completely quiet other than the falling over sound (which might as well be a mouse dying) and the occasional snore. Even a cobra is fairly loud with its bullet impact sound, and still leaves blood as evidence.

cohanna:
I agree with this. 10 seconds should be enough, and we could even add a small speed decrease so that you aren't able to put too much distance away from the injection site by just sprinting.

Oo I like this idea! Maybe it could slur your words as well!

> Anyways, the reason why Noct is at 8 is because the pen is at 10, giving you approximately a 4 second window. Refilling the pen should not need to be your concern when you are using this strategy; it just makes the skill expression into "can you juggle the UI fast enough?" which isn't interesting. 

Hard agree on the ui juggling part, noct as it stands now is a pretty good reagent for new players interested in going quiet and forcing the need to refill the hypopen makes it so that new players likely wouldn't inject enough and wonder why it didn't work.

> I think it is worth considering that, by and large, chemicals are silent. Nocturine is almost completely quiet other than the falling over sound (which might as well be a mouse dying) and the occasional snore. Even a cobra is fairly loud with its bullet impact sound, and still leaves blood as evidence.

Sure, but noct itself doesn't kill, if you want to accomplish that task it's likely gonna leave evidence no matter what. I think its role in both stealth kills and not-so-stealthy muggings (both thieves and syndies) should be preserved.

BramvanZijp:
> > Noct kills are no more difficult than buying a viper and gunning someone down and is just as respondable if you aren't expecting it. Both of the "unfair/unfun" kill methods listed in this PR are just as easily achieved with replacing noct with any other firearm. But you don't see anyone going "oh the counterplay to getting shot sucks!"
> 
> I think it is worth considering that, by and large, chemicals are silent. Nocturine is almost completely quiet other than the falling over sound (which might as well be a mouse dying) and the occasional snore. Even a cobra is fairly loud with its bullet impact sound, and still leaves blood as evidence.

Speaking of the cobra, I feel like the bullet impact sound for its caseless rounds should be drastically reduced, but that something for another PR to do.

slarticodefast:
I brought this up for maintainer discussion and while not perfect, I think we can keep the current nocturine balance for now. Like mentioned above, having to refill the pen a second time while in combat to be able to use it is a weird UI clicking skill check and not a good balance mechanic.
A more interesting solution can be implemented once we have a better medical system, for example with a conciousness stat.

PR #33834, opened 2024-12-12
Predict some power PowerReceiver stuff

metalgearsloth:
Need it for some atmos device prediction.

## Breaking changes
Moved some APCPowerReceiver stuff to shared, should use the new methods instead of setting directly (too hard to use access atm).


SlamBamActionman (approved):

SlamBamActionman:
Something something use the PR template

PR #33529, opened 2024-11-24
[BUGFIX] "Ghost" in the lobby lets you see the whole chat

Schrodinger71:
## About the PR
When you type "ghost" while being in the lobby, it lets you see the whole chat, like you're an observer. It doesn't make you a ghost, though. Your chat simply modifies.
What this PR does is checks whether the player is in lobby or not.
If he's in lobby - it will show an error that says that you can't use that in the lobby.

## Why / Balance
To simply fix the bug and prevent players from metainfo, maybe

## Technical details
What this PR does is checks whether the player is in lobby or not.
If he's in lobby - it will show an error that says that you can't use that in the lobby.

## Media
Before the fix:
image
After the fix:
image

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
I don't think the changes might be called "breaking"

**Changelog**
:cl: Schrodinger71
- fix: Fixed a bug letting players type "ghost" in the console and then see the whole chat while being in the lobby.

PR #33838, opened 2024-12-12
fix ghost command error string

lzk228:
really small pr just added missing get loc string
fixes https://github.com/space-wizards/space-station-14/pull/33529#issuecomment-2537571810


slarticodefast (approved):
Thanks for the fix!

Tayrtahn (approved):

PR #33841, opened 2024-12-12
Other colour of the binary channel

Nimfar11:
## About the PR
Changes the colour of the binary channel, making it visually different from the medical channel.

## Why / Balance
The binary and medecine channel have virtually the same colour, which is visually indistinguishable. This presents difficulties for AI, cyborg medics and sometimes syndicates. Now the binary channel will have a different tone of blue that will visually distinguish it from the others.

## Media
Before
изображение
After
изображение

## Requirements

:cl:

- tweak: Changes the colour of the Borg binary channel to a different shade of blue.

### DIFF
--- a/Content.Server/Silicons/Laws/SiliconLawSystem.cs
+++ b/Content.Server/Silicons/Laws/SiliconLawSystem.cs
@@ -67,7 +67,7 @@ public sealed class SiliconLawSystem : SharedSiliconLawSystem
 
         var msg = Loc.GetString("laws-notify");
         var wrappedMessage = Loc.GetString("chat-manager-server-wrap-message", ("message", msg));
-        _chatManager.ChatMessageToOne(ChatChannel.Server, msg, wrappedMessage, default, false, actor.PlayerSession.Channel, colorOverride: Color.FromHex("#2ed2fd"));
+        _chatManager.ChatMessageToOne(ChatChannel.Server, msg, wrappedMessage, default, false, actor.PlayerSession.Channel, colorOverride: Color.FromHex("#5ed7aa"));
 
         if (!TryComp<SiliconLawProviderComponent>(uid, out var lawcomp))
             return;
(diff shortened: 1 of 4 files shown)


SlimmSlamm:
This is extremely unnecessary considering it says the channel at the start of the message and not a ton of people are even going to be seeing Binary and Medical. Not a good PR imo.

slarticodefast:
The darker color seems a little harder to read against the background in my opinion.

ArtisticRoomba:
> This is extremely unnecessary considering it says the channel at the start of the message and not a ton of people are even going to be seeing Binary and Medical. Not a good PR imo.


I disagree, medical borgs interacting with both binary and medical channels is not absolutely unheard of. In fact, I'd say it's a bit common.

Just because it says the name of the channel at the start of the message doesn't mean that we don't need to color the channels. It's quick and intuitive for people to know which channel is which by color alone.

thetolbean:
> The darker color seems a little harder to read against the background in my opinion.

I agree. I think it should be changed for clarity and ease of use, but the showcased tone doesn't stand out clearly on the background

UbaserB:
too dark

slarticodefast:
Ideally we would have icons for the channels anyways in order to help colorblind players.

BigfootBravo:
I do like this new color, actually! And like others have said, would let borgs and AI tell medical and binary apart a bit easier!

NoElkaTheGod:
I would really like the binary channel to look like this, but unfortunately green is already reserved for common channel (and the background isn't quite black).
изображение

Nimfar11:
Made the colour brighter and a little into the green shade.

ScarKy0 (changes requested):
Could you also change the color of the Station AI Announcement to reflect the new binary color?

ScarKy0 (approved):

slarticodefast (changes requested):
The new color looks good to me.
One small thing: The beacon for the AI uses the binary chat color as well, could you update that to the new color?
grafik

slarticodefast (approved):

PR #33842, opened 2024-12-12
Mild annoyance fixes

SpaceLizard24:
## About the PR
Increased the container size of all foods (that I could find, anyways) which had their reagents go over the container size.
Each container is at the nearest multiple of 5, with the exception of:

- Nettle soup, increased to 30.5 as it has .5 Histamine in it, this means that you wont end up with having .5 of the reagent you had spare. It's purely so it looks pretty.
- Super bite burger, it's a big burger, so it can hold a lot of flavouring reagents.
- Tofu, made it consistent with it's slices.
- Laughin' pea pod and Lemoon, made it exactly the amount because that's a thing that produce has. Presumably for mutagen or something, idk.

## Why / Balance
Some foods would have less food then they should have, and things like soups and salads which take bowls would actually  _spill_ when made.

## Technical details
Changed the max container size of a bunch of food items. Not sure if this will cause the universe to explode.

## Media
Do you want a screenshot of bungo curry? get it yourself 

## Requirements

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/burger.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/burger.yml
@@ -583,7 +583,7 @@
   - type: SolutionContainerManager
     solutions:
       food:
-        maxVol: 15
+        maxVol: 20
         reagents:
         - ReagentId: Nutriment
           Quantity: 6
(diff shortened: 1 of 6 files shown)


Everturning:
can you make chicken sandwiches hold more? when I tried putting a certain addictive chemical in it was already full so I just had to serve the sandwiches and glasses of meth as a side

SpaceLizard24:
What I did was only increase the foods that were over capacity, not right on, but yeah, I'll probably do that

SpaceLizard24:
i don't know how git works, but the PR is done. enjoy!

slarticodefast (approved):
Looks good to me!

Issue #33795, opened 2024-12-09
Depowered doors can't be closed.


thebadman4662:
## Description
So you get nearly instant spacing of whole station if engi can't prevent any kind of brownouts from the start of round.

**Reproduction**
Simply crowbar depowered door.

**Screenshots**
![depowered doors can't be closed without power](https://github.com/user-attachments/assets/df9861e2-add0-418e-a369-17854536a487)

Ian321:
Replicated on Leviathan 73152.

https://github.com/user-attachments/assets/8128897c-974b-4d36-85e4-a018412eb466

zHonys:
I found the issue, trying to make a fix

FairlySadPanda:
The bug is basically that prying an unpowered AirlockComponent door open doesn't hit a specific "can change state" function, but when prying it closed it does, and the code in question only works if the door is powered. So specifically an Airlock can't be pried closed.

This smells like cascading rot elsewhere impacting badly-architected code here. I'll see if I can fix this in a way that simplfies the system, but on the face of it the easiest fix is to just get rid of the "is powered" check when attempting to close an airlock.

FairlySadPanda:
Yeah this is basically because there's two reasons a door can close:

1. For reason of power
2. By reason of pry

But there's no handling for recognizing the difference between these two states. Door code is complex and this particular route has been missed.

This needs encapsulating by extending the states a door can be in to prevent this modelling drift. A door can be closing under its own power (Closing) or by an external factor that does not require power (ForceClosing). Encapsulating this will allow for code routes that make sense.

zHonys:
I already fixed the main issue of: the door can't be closed when de powered, some hours ago. But, there already was an bug that went undetected that makes the closing animation not play correctly, because it goes in the Closing -> Open -> Closed, i'll make a PR with the basic fix, but another one still is needed to fix To be clear, this only happens with de powered doors. If someone at least find the issue, i can try making the second fix later

zHonys:
> Yeah this is basically because there's two reasons a door can close:
> 
>     1. Under its own power
> 
>     2. By prying
> 
> 
> But there's no handling for recognizing the difference between these two states. Door code is complex and this particular route has been missed.
> 
> This needs encapsulating by extending the states a door can be in to prevent this modelling drift. A door can be closing under its own power (Closing) or by an external factor that does not require power (ForceClosing). Encapsulating this will allow for code routes that make sense.

[...]

PR #33843, opened 2024-12-12
Fix airlocks can't be closed without power.

c4llv07e:
## About the PR
Right now you can't close unpowered airlocks with crowbar or by hands. This PR fixes it.

## Why / Balance
It's a bug, I guess?

## Technical details
Ignore power state when handling airlock closing event.

I can't say for sure if it's good solution, but it works and I can't find any problematic interactions caused by this fix.

Tested with different airlocks configurations (locked, unpowered, bolted), all of them are working as intended.

## Media

https://github.com/user-attachments/assets/b11fbc54-f954-4dd5-85c3-439cc4db743c

## Requirements

## Breaking changes
None

**Changelog**

:cl: c4llv07e
- fix: Unpowered airlocks can be closed again!

### DIFF
--- a/Content.Shared/Doors/Systems/SharedAirlockSystem.cs
+++ b/Content.Shared/Doors/Systems/SharedAirlockSystem.cs
@@ -43,7 +43,7 @@ public abstract class SharedAirlockSystem : EntitySystem
 
         if (TryComp(uid, out DoorComponent? door)
             && !door.Partial
-            && !CanChangeState(uid, airlock))
+            && DoorSystem.IsBolted(uid))
         {
             args.Cancel();
         }
--- a/Content.Shared/Doors/Systems/SharedDoorSystem.cs
+++ b/Content.Shared/Doors/Systems/SharedDoorSystem.cs
@@ -221,7 +221,7 @@ public abstract partial class SharedDoorSystem : EntitySystem
 
     private void OnBeforePry(EntityUid uid, DoorComponent door, ref BeforePryEvent args)
     {
-        if (door.State == DoorState.Welded || !door.CanPry)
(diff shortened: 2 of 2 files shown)


Content.Shared/Doors/Systems/SharedAirlockSystem.cs
```diff
@@ -43,7 +43,7 @@ private void OnBeforeDoorClosed(EntityUid uid, AirlockComponent airlock, BeforeD
 
         if (TryComp(uid, out DoorComponent? door)
             && !door.Partial
-            && !CanChangeState(uid, airlock))
+            && DoorSystem.IsBolted(uid))
```
slarticodefast, line 46: If I read the code comment above correctly then `!door.Partial` should be there to bypass the power check in `CanChangeState`. So instead of removing this check and potentially breaking something else I would rather find the underlying cause for why that is not working.

c4llv07e:
Okay, it's not fully working, I can close unpowered airlock with E click.
And if I try to close it when there is some entity inside, it won't close, but it will try after the entity is removed.

Errant-4:
Fixes #33843

Errant-4:
Fixes #33795

jamilnielsen:
sorry for popping in, but whats stopping this from being merged?
is it the door.Partial thing? considering the other pr closed after trying to deal with exactly that angle, perhaps this is the best we got for now.

metalgearsloth:
> perhaps this is the best we got for now.

It shouldn't be hard to track down door bugs so better to do that than merge bandaids that may never potentially be unmerged.

c4llv07e:
> sorry for popping in, but whats stopping this from being merged?

Right now I have problem that the door closes itself after an entity leaves it after an unsuccessful pry close.

PR #33844, opened 2024-12-12
Fix the ability to kill clumsy mobs by force climbing

MilenVolf:
## About the PR

One condition was removed due to exploit that can be used to kill clumsy mobs by force climbing them on table non-stop. So now 50% check applies on every climb attempt.
It can also be fixed by reducing stun time, but due to mispredictions it must be reduced to only 1 second (current is 2)

## Why / Balance
Fixes the exploit that can be done by players

## Technical details
Modifies condition
Before:
```C#
if (!_cfg.GetCVar(CCVars.GameTableBonk) && args.PuttingOnTable == ent.Owner && !rand.Prob(ent.Comp.ClumsyDefaultCheck))
   return;
```
After:
```C#
if (!_cfg.GetCVar(CCVars.GameTableBonk) && !rand.Prob(ent.Comp.ClumsyDefaultCheck))
   return;
```
## Requirements

**Changelog**


slarticodefast:
Hmm, I don't like hardcoding the `BiomassReclaimerComponent` into the `ClumsySystem`, that should be an event. We just removed the `ClumsyComponent` from the `GunSystem` for the same reason: It makes systems that have nothing to do with each other depend on each other.

Content.Shared/Clumsy/ClumsySystem.cs
```diff
@@ -97,14 +98,17 @@ private void BeforeGunShotEvent(Entity<ClumsyComponent> ent, ref SelfBeforeGunSh
     private void OnBeforeClimbEvent(Entity<ClumsyComponent> ent, ref SelfBeforeClimbEvent args)
     {
         // checks if ClumsyVaulting is false, if so, skips.
-        if (!ent.Comp.ClumsyVaulting)
+        if (!ent.Comp.ClumsyVaulting
+            || HasComp<BiomassReclaimerComponent>(args.BeingClimbedOn))
```
Tayrtahn, line 102: Yeah, I'm definitely not a fan of hard coding this component check, as slarticodefast said. Is this check even needed with the other check in place?

Tayrtahn (changes requested):

Content.Shared/Clumsy/ClumsySystem.cs
slarticodefast: We probably don't even need to use `SelfBeforeClimbEvent` here and can use normal drag drop without any climbing for inserting into the reclaimer, similar ot the cryopod.

aspiringLich:
I just ran into this bug, and it is, in fact, annoying. Is this still being worked on?

MilenVolf:
> I just ran into this bug, and it is, in fact, annoying. Is this still being worked on?

I was trying to fix it properly by doing what slarti suggested, but have faced with the problem in code, that just don't invoke subscribed method to the dragDrop event. I don't know the reason why it's happening but I think I will just make PR with this then and will ask for help with that

MilenVolf:
I decided to make requested changes in another PR. This PR was converted into small fix for clumsy system

slarticodefast (approved):
Gave it a test and this fixes the stun lock.

Issue #33480, opened 2024-11-23
Moving into a closing airlock mispredicts the animation


metalgearsloth:
It looks pretty bad.

metalgearsloth:
Played around with it and I think the best thing is to still play the audio but don't play any animations by setting the state straight back to open. The closing one is delayed and should never be hit.

PR #33846, opened 2024-12-13
Hotfix: Fixed Closing Unpowered Airlocks

Aeshus:
## About the PR

Did a quick bisect, found that #33481 broke unpowered airlocks from being closed. I just reverted the part that broke it.

## Why / Balance

Not being able to close airlocks is bad.

Alternative to #33843, as this one only tries to revert the change while that one changes the logic to make it work. I don't really know which is better.

## Technical details

The CanClose check fails for unpowered airlocks as it tries to short circuit with `door.Partial`, and as it got moved after, it would be false and cancel the event, and thus fail to close the door.

https://github.com/space-wizards/space-station-14/blob/821ae95dd43ab1e0c18948db36dcce404312be2c/Content.Shared/Doors/Systems/SharedAirlockSystem.cs#L44-L46

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Unpowered airlocks can now be pried closed again.

### DIFF
--- a/Content.Shared/Doors/Systems/SharedDoorSystem.cs
+++ b/Content.Shared/Doors/Systems/SharedDoorSystem.cs
@@ -480,6 +480,8 @@ public abstract partial class SharedDoorSystem : EntitySystem
         if (!Resolve(uid, ref door, ref physics))
             return false;
 
+        door.Partial = true;
+
         // Make sure no entity walked into the airlock when it started closing.
         if (!CanClose(uid, door))
         {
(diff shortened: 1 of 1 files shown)


slarticodefast:
Good investigation! I tested everything in-game and this fixes it without reintroducing #33480.

Errant-4 (approved):
I'm not fully sure this solves everything, but I did not see anything obvious that it broke, and it's probably better than live where in some situations airlocks are unclosable... I say we go with it, as otherwise we would have to either wait for who knows long or revert the prediction fix

Aeshus:
After testing this solution more thoroughly, I believe that the alternative PR is more fruitful (linked in the original PR description).

There's some weird behavior where when crowbaring a door that has an entity inside of it, the door will still close (after the entity is moved) but be no-collide. This was probably because the prediction PR also changes it from the 'opening' state to the 'open' state. I can't just revert this as it reintroduces the graphical glitch that Sloth was trying to fix.

The alternative PR handles this case better as there's no weird graphical no-collide, none of the graphical bugs of the original misprediction, and has a smaller diff. It just seems to work better than this one.

PR #33854, opened 2024-12-14
Show battery level for selected devices in Power Monitoring Console

ArtisticRoomba:
## About the PR
Adds a charge percentage and charge bar to applicable devices in the Power Monitoring Console.

This is a revival of derelict PR #26793, originally coded by Crotalus. I've tried to make it look decent, this is my first PR touching any drop of C#.

## Why / Balance
Extremely needed QOL feature. This tells engineers battery levels across the station and can prove EXTREMELY USEFUL to know. Really gives you a good idea as to where power is moving around in the station, and how the grid behaves as a whole.

## Technical details
* Added battery level indicators to the `PowerMonitoringButton` class, including a progress bar and percentage label for visual representation of battery levels.

* Refactored `GetPrimaryPowerValues` into `PowerStats` class to encapsulate power statistics.

* Updated the `PowerMonitoringConsoleEntry` struct to include a nullable `BatteryLevel` property to store battery level information. (`Content.Shared/Power/SharedPowerMonitoringConsoleComponent.cs`)

* Adds the `BatteryLevel` property to the `PowerMonitoringConsoleEntry` objects that are being created and modified within the `GetSourcesForNode` and `GetLoadsForNode` methods.

## Media
Content Client_weC4i7G7fx

Content Client_ngiUnjU4RL

Content Client_nKd9VEQ4FD

Content Client_2R36TbOB7F

image

## Requirements

## Breaking changes
Shouldn't be any.

**Changelog**
:cl: Crotalus, ArtisticRoomba
- add: Battery levels of applicable devices are now shown in the Power Monitoring Console!


Resources/Locale/en-US/components/power-monitoring-component.ftl
```diff
@@ -13,6 +13,7 @@ power-monitoring-window-unknown-location = Unknown location
 power-monitoring-window-total-sources = Total generator output
 power-monitoring-window-total-battery-usage = Total battery usage
 power-monitoring-window-total-loads = Total network loads
+power-monitoring-window-battery-level = {$percentage}%
```
PJB3005, line 16: This isn't used.

Content.Client/Power/PowerMonitoringWindow.xaml.Widgets.cs
```diff
@@ -104,6 +105,28 @@ public void UpdateWindowEntryButton(NetEntity netEntity, PowerMonitoringButton b
...
+            var batteryPercentage = (int)(entry.BatteryLevel.Value * 100);
+
+            button.BatteryLevel.Value = batteryPercentage;
+            button.BatteryLevel.Visible = true;
+
+            button.BatteryPercentage.Text = $"{batteryPercentage}%";
+            button.BatteryPercentage.Visible = true;
+
+            // Set progress bar color based on percentage
+            var color = Color.FromHsv(new Vector4((batteryPercentage / 100f) * 0.33f, 1, 1, 1));
```
PJB3005, line 121: Why not use `entry.BatteryLevel.Value` directly? You're dividing and then multiplying by 100 here.

Content.Client/Power/PowerMonitoringWindow.xaml.Widgets.cs
```diff
@@ -104,6 +105,28 @@ public void UpdateWindowEntryButton(NetEntity netEntity, PowerMonitoringButton b
...
+
+        // Update battery level if applicable
+        if (entry.BatteryLevel != null)
+        {
+            var batteryPercentage = (int)(entry.BatteryLevel.Value * 100);
+
+            button.BatteryLevel.Value = batteryPercentage;
+            button.BatteryLevel.Visible = true;
+
+            button.BatteryPercentage.Text = $"{batteryPercentage}%";
```
PJB3005, line 117: You should use `.ToString("P0")` instead so this localizes properly.

Content.Client/Power/PowerMonitoringWindow.xaml.Widgets.cs
```diff
@@ -478,6 +505,31 @@ public PowerMonitoringButton()
...
 
+        BatteryLevel = new ProgressBar()
+        {
+            HorizontalAlignment = HAlignment.Right,
+            SetWidth = 60f,
+            Margin = new Thickness(15, 7, 0, 7),
+            MinValue = 0,
+            MaxValue = 100,
+            Page = 0,
+            Value = 0,
```
PJB3005, line 516: You're just setting all of these to the default, there's no point to.

Also set `MaxValue` to 1 so you can avoid multiplying by 100 when setting the value.

Content.Client/Power/PowerMonitoringWindow.xaml.Widgets.cs
```diff
@@ -478,6 +505,31 @@ public PowerMonitoringButton()
...
+            Visible = false,
+        };
+
+        MainContainer.AddChild(BatteryLevel);
+
+        BatteryPercentage = new Label()
+        {
+            HorizontalAlignment = HAlignment.Right,
+            SetWidth = 45f,
```
PJB3005, line 525: You should probably set a MinWidth instead so the layout doesn't break when we get the ability for users to select custom fonts.

Content.Client/Power/PowerMonitoringWindow.xaml.Widgets.cs
```diff
@@ -478,6 +505,31 @@ public PowerMonitoringButton()
 
         MainContainer.AddChild(NameLocalized);
 
+        BatteryLevel = new ProgressBar()
+        {
+            HorizontalAlignment = HAlignment.Right,
```
PJB3005, line 510: This doesn't do anything.

Content.Client/Power/PowerMonitoringWindow.xaml.Widgets.cs
```diff
@@ -478,6 +505,31 @@ public PowerMonitoringButton()
...
+            Page = 0,
+            Value = 0,
+            Visible = false,
+        };
+
+        MainContainer.AddChild(BatteryLevel);
+
+        BatteryPercentage = new Label()
+        {
+            HorizontalAlignment = HAlignment.Right,
```
PJB3005, line 524: This doesn't do anything.

Content.Server/Power/EntitySystems/PowerMonitoringConsoleSystem.cs
```diff
@@ -990,4 +1005,13 @@ private void RefreshPowerMonitoringCableNetworks(EntityUid uid, PowerMonitoringC
 
         Dirty(uid, component);
     }
+
+    private sealed class PowerStats
```
PJB3005, line 1009: This should be a struct.

PJB3005 (changes requested):
> I did some debugging and saw that the PowerMonitoringConsoleEntry[] struct list never has a BatteryLevel float (null), while PowerMonitoringConsoleEntry does when applicable. Because the value is always null, the battery info is never drawn.
> 
> I think this is boiled down to it not being recorded somewhere in PowerMonitoringConsoleSystem, however I don't have experience nor the time right now to go further than what I've already done.

There are other places in the system code where these `PowerMonitoringConsoleEntry` values are being made, you should update them too.

ArtisticRoomba:
> There are other places in the system code where these `PowerMonitoringConsoleEntry` values are being made, you should update them too.

I addressed your review, however I admittedly don't know where exactly these values are being made, after I spent some time searching in the relevant system. I'll look around some more. Sorry.

ArtisticRoomba:
NEVERMIND
image

PJB3005 (approved):

slarticodefast (approved):
Blessed PR.
Tested and works as intended.

superjj18:
I just want to say, I know this change took place on the 12/18/24, but I would still nominate this for best change of 2025 because holy fuck this change revolutionized engineering.  The technical issues you can diagnose just by looking at the console are amazing.  I dont think any other engi PR has changed engineering so drastically with so small a change.  Genuinely bravo

PR #33858, opened 2024-12-14
Chat Refactor [DRAFT]

SlamBamActionman:
## About the PR

This is a refactor of the chat system! It is heavily in development and is *not merge-ready yet*; this PR is currently acting as a way for contributors and maintainers to preview the new system and provide feedback in areas where work has been done. 

The goal of the chat refactor has been to achieve three things:

1. **Modularity**. It should be easy to add new behaviors to chat messages, whether that be modifying text or evaluating conditions.
2. **Linearity**: While chatcode may still be complex, it should be easy to follow it from start to finish, without any offshoots.
3. **Condensed**: Chatcode shouldn't have copy/pasted behavior spread out over multiple systems. Instead it should be concentrated into a single pipeline, and content should be reused where possible.

Note that the PR may preserve some undesirable features that are subject to change by considering them out of scope, e.g. chat being sent as commands. If something can be refactored at a later date without clashing with the core changes, we've likely kept it.

**Due to the sheer scope of this PR and the many files changed just for the code to compile, below I will highlight the files and functions which are of importance in the refactor.**

**Many files have been changed to comply with the refactor, but if you're looking to find the core changes, look at the following files:**
[`ChatManager.MessageSending.cs`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Server/Chat/Managers/ChatManager.MessageSending.cs) - For the main serverside message logic. 
[`ChatCondition.cs`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Shared/Chat/ChatCondition.cs) - Chat conditions; determines if a message may be sent/received by a session/entity
[`ChatModifier.cs`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Shared/Chat/ChatModifier.cs) - Chat modifiers; changes the message in some way, often in the form of adding content markup tags.
[`ContentMarkupTagProcessorBase.cs`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Shared/Chat/ContentMarkupTags/ContentMarkupTagProcessorBase.cs) - Content markup tags; adds to or modifies message nodes inside of the tag.
[`ChatUIController.cs`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs) - For the main clientside message logic.
[`CommunicationChannelPrototype.cs`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs) - Defines chat channels and their behavior. Everything that makes a message not just be regular plaintext is defined in here.
[`communication_channels.yml`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Resources/Prototypes/Chat/communication_channels.yml) - The yaml defining the above communication channels.

---

# Review Guide

To help with reviewing this massive PR, this section should provide a structured description of the important concepts of the PR. I won't cover all the details, but leave some questions as a F.A.Q. to the end.

### Communication Channels

Communication Channels is a concept to condense the many different behaviors of oldchat into a singular pipeline. Channels have three goals:
1. Define who can send (publish) a message.
2. Define who can receive (consume) a message.
3. Modify the message based on the sender & recipient.

[`CommunicationChannelPrototype`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs) defines the prototype, and implementations for channels can be found in [`communication_channels.yml`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Resources/Prototypes/Chat/communication_channels.yml).

The main reason to keep this as yaml is because it allows for a powerful degree of modularity and customization. New channel behavior can be developed on its own and then added to desirable channels, and if a downstream fork wants to modify the default channels they can easily do so without changing core code. 

The main parts to how a communication channel behaves are **chat conditions**, **chat modifiers** and **content markup tags**.

### Chat Conditions

A [`ChatCondition`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Shared/Chat/ChatCondition.cs) is a class that takes an `ICommonSession` *or* `EntityUid` as input and returns a boolean. Chat conditions are used to check if a session is an admin, if an entity is conscious, if a radio server is up etc. This is used to evaluate publishers and consumers, to ensure the right player can send/receive the messages in the various channels. 

For publishing, chat conditions are stored in the `List<IChatCondition>` [`PublishChatConditions`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs#L28), and for consuming they are stored in [`ConsumeChatConditions`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs#L36) in the prototype.

<details>
<summary><b>Chat Condition FAQ</b></summary>
<br> 

- How are the chat condition lists evaluated? 

Chat conditions are evaluated as children of the base conditions [`AllChatCondition`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/ChatCondition.cs#L9) and [`AnyChatCondition`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/ChatCondition.cs#L43), which act as *and*/*or* logical operators. When parsing the channel prototype yaml, all chat conditions defined are included under an `AnyChatCondition`. Chat condition outputs can also be inverted!

- What about conditions that need custom data passed into them?

Chat conditions also take `ChatMessageContext` as input, which is a dictionary of enums and data. This can be the sender entity/session, custom data from the system sending the message (e.g. for announcement colors) or defined in the chat channel prototype (e.g. radio channel id). 

- Isn't this just programming in yaml?

Not any more than construction graphs. Chat conditions are tree structures which are necessary to allow the degree of modularity you'd want from a chat system, and the ability to insert new functionalities or behaviors is much stronger this way. 

</details>

### Chat Modifiers

A `ChatModifier` is a class that takes a `FormattedMessage` message as input and returns a modified version of the message.

Chat modifiers can either be applied serverside or clientside. Serverside modifiers should, as a rule, apply content markup tags or modify the message when the client isn't allowed to know the original contents (e.g. whispering obfuscation).  Clientside modifiers should generally apply formatting (e.g. colors, accents). They are stored in `List<ChatModifier>` `ServerModifiers` and `ClientModifiers` respectively.

Of note is the `List<ConditionalModifiersCollection>` `ConditionalModifiers`, which are serverside modifiers that only apply if they meet certain chat conditions as defined in the `ConditionalModifiersCollection` object.

<details>
<summary><b>Chat Modifiers FAQ</b></summary>
<br> 

- What about modifiers that need custom data passed into them?

Same as conditions, modifiers also take `ChatMessageContext` as input, which is a dictionary of enums and data. 

- What if I need different modifiers depending on the consumer?

For that we have the `List<ConditionalModifiersCollection>` `ConditionalModifiers`. This list contains serverside modifiers that only apply if the consumer meets the chat conditions listed in the `ConditionalModifiersCollection` object. Consumers are split up into groups based on what chat conditions they meet, and modifiers are applied to messages sent to each of those groups.

</details>

### Content Markup Tags

A `ContentMarkupTag` is a fairly simple concept; they are functionally similar to formatting tags like `[bold]hello[/bold]` or `[color=#FF00FF]world[/color]` but are defined in the `Content` namespace and confer some behavior. This can be things like playing audio, replace itself with a string, or apply an accent to its contents. This is the primary way to bake in message-specific metadata.

Chat modifiers are used to insert content markup tags. 

<details>
<summary><b>Content Markup Tags FAQ</b></summary>
<br> 

- What about ChatMessageContexts? Those include metadata too.

The `ChatMessageContext` data for a message is *not* networked between server and client. Some may be accessed on both sides (such as channel-specific data, such as the radio channel) but, in general, including it as a ContentMarkupTag and processing it clientside is safest.

</details>

## The flow of a message

Messages enter the chat pipeline via two different ways:

- Messages are sent from the client to the server using the `chat` command (it supports legacy `say`/`dsay`/`me`/`ooc` commands), and the target channel is given as a command argument. This becomes the input to the `ChatManager.SendChannelMessage` function.
- Entity systems runs the `ChatManager.SendChannelMessage` function, with optional `ChatMessageContext` properties.

`SendChannelMessage` does the following in order:
1. Prevent recursion by checking parent messages.
2. Prepare the chat message context (combine optional properties with default and prototype properties).
3. Evaluate the publishing chat conditions.
4. Evaluate the consumer chat conditions. 
5. Separate the consumers into groups depending on what conditional chat modifiers they should get.
6. Apply server chat modifiers.
8. Apply conditional chat modifiers for each consumer group.
9. Send the message out for each consumer group.
10. Raise a listener event for entities.

On the clientside, the message is received and processed in [`ChatUIController.ProcessChatMessage`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs#L826C17-L826C35):
1. Prepare the chat message context.
2. Apply client chat modifiers.
3. Apply content markup tags.
4. Queue up speech bubbles if necessary.
5. Store the message in history if necessary.

<details>
<summary><b>General FAQ</b></summary>
<br> 

- What is the [`ChatMessageWrapper`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Server/Chat/Managers/ChatMessageWrapper.cs) used as input for `SendChannelMessage`?

This is a record that contains the base data required to properly send a message. It contains the unique ID of a message, any parent message, sender entity/session etc. This is also stored in the chat repository. 

- What's the [`ChatFilter`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs#L68) enum?

This is used to assign what chatbox filter a channel belongs to, allowing users to filter out messages.

- What's the [`ChatMedium`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs#L75) enum?

The chat medium is used for entity listening; it represents the medium a channel message is sent through in-character. For example, a voice trigger should only activate on auditory messages.

- What's the [`AlwaysRelayedToChannels`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs#L83) and [`FallbackChannels`](https://github.com/SlamBamActionman/space-station-14/blob/a3a3c4816d2072be032704de028900db0cee1e0a/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs#L90) lists in [`CommunicationChannelPrototype`](https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Content.Shared/Chat/Prototypes/CommunicationChannelPrototype.cs)?

They represent channels that will also be sent the message for the given channel prototype. AlwaysRelayed are always given the message, while FallbackChannels only send if the message failed. Useful for when you try to do IC speech while dead, falling back to the dead channel.

</details>

## Missing features

- [ ] Sanitizing the messages, making sure they use proper punctuation and capitalization where necessary, as well as emote handling.
- [ ] Accents! Must be moved from Server to Shared, since we want them applied clientside (most of the time).
- [ ] Handheld radios! Should use the listener system.
- [ ] Intercoms! Should also use the listener system.
- [ ] Announcements! Gotta make sure they're properly using custom colors.
- [x] Ensuring admin chat parity! I don't have much user experience of the admin side of SS14, so it's a bit hard to verify everything is working as intended.
- [ ] Radio Jammer! Should be as easy as making a single ChatCondition. 
- [x] Codeword highlighting! Should be as easy as making a single ChatModifier.
- [ ] TVs! Basically same as the intercoms.
- [x] Cleaning up old chatcode, removing everything that is no longer utilized.
- [ ] Cleaning up documentation, adding where it's missed.
- [ ] Fixing inevitable bugs for things I've missed.

## Requirements

## Breaking changes

***Note: this section is outdated and will be updated closer to release.***

`radio_channels.yml` remains untouched, however if you have custom radio channels you will need to create a matching communication channel prototype. Ideally this would be resolved automatically, but the chat refactor must have some scope constraints. To do this, you can look at `communication_channels.yml` in the radio section and copy an existing radio channel; of importance is that the prototype has the id: `Radio[YourRadioPrototypelIdHere]` and the `DefaultChannelParameters.RadioChannel` value be the radio prototype ID.

You will also have to convert accents to Shared, but since I haven't done that part of the refactor yet, I can't explain how to do it. Note to self: remember to edit this part before merging, or it will be embarrassing.

**Changelog**

:cl:
- tweak: There has been a major overhaul to how chat is handled in the game; please report any new chat bugs you may encounter.


FairlySadPanda:
I am here

it is time

[to review](https://www.youtube.com/watch?v=D_2bluVPsb0)

FairlySadPanda:
First things first here's a shorter writeup of Slam's PR summary for those with Subway Surfer brains:

```Hey, here's my notes on the PR above to just condense down the info

Much like "newmed" let's call this something like newchat.

PR information:

1. This isn't done yet
2. This means reviewing for code quality is pointless
3. Slam has prioritized:
-> Modularity, by which he means he want it to be extensible.
-> Linearity, meaning it is generally a consistent point to point process, not a tree.
-> Condensed. No copypasta, one pipeline, KISS.

CommunicationsChannelPrototype
    -> PublishSessionChatConditions
        Lists conditions that a session must fulfil to send on this channel
    -> EntityChatConditions
        A subset of the above that covers entity conditions.
    
The server as auth does not need these auth checks.

ConsumeCollections 
    -> A collection of conditions for receiving on a channel.
        -> Session conditions
        -> Entity conditions
        -> Condition modifiers (i.e. mutators)

ClientChatModifiers
    -> Client-side mods of chat, for accessibility etc.

AlwaysChildCommunicationChannels
BackupChildCommunicationChannels
    -> Subscribers to other channels. e.g. deadchat is a backup to local if you're dead.

SendChannelMessage
    This is the core event that encapsulates a message.
    Contains a message, the channel, and opts of the session and entity sending.
    Some chat conds and mods lean on a context object called ChannelParameters which is a dict of enum to obj.
    Messages can be given specific target sessions.
    
All console chat messages use newchat.

Messages are formatted messages when sent over the network. Formatted messages have a message and markup for that message. Markup uses square brackets - `[]` - for tags. Examples of use are color or entity name tags.

ListenerEntitySystem and ListenerComponent
    -> Non-players get messages via events. These are subbed to by extensions of ListenerComponent. 


TODOs

-> IOC Injection, sim separation
    Some concerns about perf and redundant instantiation.
-> Concerns about weight of chat messages OTA
-> FormattedMessages haven't had a optimization pass
-> Debate over if it's better to have a ProtoID var for senders of SendChatMessage or if strings are ok?

-> Message sanitization
-> Accent handling
-> Handheld radios, intercoms and TVs
-> Announcements
-> Admin chat parity
-> Radio jamming
-> Codeword highlighting
-> Removing chatcode
-> Documentation
-> Bug sweep```

**Comment by Djungelskog2** [comment]
languages, coming to a station near you soon

File: Content.Client/Chat/Managers/ChatManager.cs [hunk 1]
```diff
@@ -12,26 +15,20 @@ internal sealed class ChatManager : IChatManager
...
-    public void SendAdminAlert(EntityUid player, string message)
-    {
-        // See server-side manager. This just exists for shared code.
-    }
-
-    public void SendMessage(string text, ChatSelectChannel channel)
+    public void SendMessage(string text, ChatSelectChannel channel, RadioChannelPrototype? radioChannel)
     {
         var str = text.ToString();
         switch (channel)
```
Fildrance, line 34: this looks like old-old code, is it left there just for backwards compatibility? If so i'd like to propos Proxy or Adapter design pattern here <_< low priority thing ofc...

Content.Client/Chat/Managers/ChatManager.cs
```diff
@@ -81,4 +88,19 @@ public void SendMessage(string text, ChatSelectChannel channel)
...
+        {
+            SendMessage(text, channelPrototype);
+        }
+    }
+
+    public void SendMessage(string text, CommunicationChannelPrototype channel)
+    {
+        var str = text.ToString();
+
+        _consoleHost.ExecuteCommand($"chat {channel.ID} \"{CommandParsing.Escape(str)}\"");
```
Fildrance, line 104: i would like to remove this in favour of direct message/event sending, launching some console commands looks ugly Q_Q
It also can cause unexpected results... but thats of none concern as we were living with this for so long...

Content.Client/Chat/Managers/ContentMarkupTagManager.cs
```diff
@@ -0,0 +1,17 @@
...
+using Content.Shared.Chat;
+using Content.Shared.Chat.ContentMarkupTags;
+
+namespace Content.Client.Chat.Managers;
+
+public sealed class ContentMarkupTagManager : ISharedContentMarkupTagManager
+{
+    public Dictionary<string, IContentMarkupTag> ContentMarkupTagTypes => new IContentMarkupTag[]
+    {
+        new ColorValueContentTag(),
```
Fildrance, line 12: it would be cool if code of DI would just scan assembly for all implementation of IContentMarkupTag and inject them there, and its easy to do! BUT! i am not sure if assembly-scanning is ok for sandboxing, and 2 - i'm not sure if maintainers are okay with that practice.

Content.Client/Chat/Managers/ContentMarkupTagManager.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Linq;
+using Content.Client.Chat.MarkupTags;
+using Content.Shared.Chat;
+using Content.Shared.Chat.ContentMarkupTags;
+
+namespace Content.Client.Chat.Managers;
+
+public sealed class ContentMarkupTagManager : ISharedContentMarkupTagManager
+{
+    public Dictionary<string, IContentMarkupTag> ContentMarkupTagTypes => new IContentMarkupTag[]
```
Fildrance, line 10: u decalare it as Dictionary, so u are OK with changing it in runtime? I'm not 100% sure what about maintainers policy on that but i'd better declare it as IReadOnlyDictionary if u want it initialized and never changed after creation. Maybe some frozen dictionary for bettere perf but same semantics?

Content.Client/Chat/Managers/IChatManager.cs
```diff
@@ -1,9 +1,15 @@
 using Content.Shared.Chat;
+using Content.Shared.Chat.Prototypes;
+using Content.Shared.Radio;
 
 namespace Content.Client.Chat.Managers
 {
     public interface IChatManager : ISharedChatManager
     {
-        public void SendMessage(string text, ChatSelectChannel channel);
+        public void SendMessage(string text, ChatSelectChannel channel, RadioChannelPrototype? radioChannel);
```
Fildrance, line 9: but this ine is cursed :pepeHands: why do we leave it be and not kill it with fre? it is called RadioChannel but is changeling hiveming radio channel? thats all so confusing so i'd be better off without it... but oh well...

Content.Client/Chat/MarkupTags/ColorValueContentTag.cs
```diff
@@ -0,0 +1,62 @@
...
+        ["Radio.Freelance"] = Color.FromHex("#f6ce64"),
+        ["Radio.Service"] = Color.FromHex("#539c00"),
+        ["Radio.Binary"] = Color.FromHex("#2ed2fd"),
+        ["Radio.CentCom"] = Color.FromHex("#2681a5"),
+        ["Radio.Handheld"] = Color.FromHex("#967101"),
+    };
+
+    public List<MarkupNode>? OpenerProcessing(MarkupNode node)
+    {
+        return new List<MarkupNode>() { new MarkupNode("color", new MarkupParameter(ColorValueLookup(node)), null) };
```
Fildrance, line 44: i'm a little bit afraid of perf for heavily loaded server with all thouse lists - i'm not agitating to do some pre-emptive optimizations but it might be interesting to make all chat code allocation-less... maybe later!

Content.Client/Chat/MarkupTags/EntityNameHeaderContentTag.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Content.Client.Chat.MarkupTags;
+
+public sealed class EntityNameHeaderContentTag : IContentMarkupTag
+{
+    public string Name => "EntityNameHeader";
+
+    public List<MarkupNode>? OpenerProcessing(MarkupNode node)
+    {
+
+        var list = new List<MarkupNode>();
```
Fildrance, line 21: dont allocate if u dont use! : D i'm not 100% sure it will be optimized by compiler so better move it down after IF yourself.

Content.Client/Chat/MarkupTags/SessionNameHeaderContentTag.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Content.Client.Chat.MarkupTags;
+
+public sealed class SessionNameHeaderContentTag : IContentMarkupTag
+{
+    public string Name => "SessionNameHeader";
+
+    public List<MarkupNode>? OpenerProcessing(MarkupNode node)
+    {
+
+        var list = new List<MarkupNode>();
```
Fildrance, line 21: same

Content.Client/Chat/UI/SpeechBubble.cs
```diff
@@ -15,6 +15,7 @@ public abstract class SpeechBubble : Control
     {
         [Dependency] private readonly IEyeManager _eyeManager = default!;
         [Dependency] private readonly IEntityManager _entityManager = default!;
+        [Dependency] protected readonly ISharedChatManager _chatManager = default!;
```
Fildrance, line 18: i'd say this one needs updating to adapt usage of our protos >_> so maybe leave todo at least? SpeechType  enum is irritating for example...

Content.Client/Entry/EntryPoint.cs
```diff
@@ -72,6 +73,7 @@ public sealed class EntryPoint : GameClient
         [Dependency] private readonly ILogManager _logManager = default!;
         [Dependency] private readonly DebugMonitorManager _debugMonitorManager = default!;
         [Dependency] private readonly TitleWindowManager _titleWindowManager = default!;
+        [Dependency] private readonly ISharedContentMarkupTagManager _contentMarkupTagManager = default!;
```
Fildrance, line 76: <_< what is that? i mean EntryPoint class... and why is manager needed here...

Content.Client/UserInterface/Systems/Chat/Controls/ChannelFilterPopup.xaml.cs
```diff
@@ -17,6 +17,7 @@ public sealed partial class ChannelFilterPopup : Popup
         ChatChannel.Emotes,
         ChatChannel.Radio,
         ChatChannel.Notifications,
+        ChatChannel.Announcements,
```
Fildrance, line 20: read from protos too? ;w;

Fildrance: maybe leave todo...

Content.Server/Advertise/EntitySystems/AdvertiseSystem.cs
```diff
@@ -62,7 +63,7 @@ public void SayAdvertisement(EntityUid uid, AdvertiseComponent? advert = null)
             return;
 
         if (_prototypeManager.TryIndex(advert.Pack, out var advertisements))
-            _chat.TrySendInGameICMessage(uid, Loc.GetString(_random.Pick(advertisements.Values)), InGameICChatType.Speak, hideChat: true);
+            _chat.SendChannelMessage(Loc.GetString(_random.Pick(advertisements.Values)), "BubbleOnlySpeech", null, uid);
```
Fildrance, line 66: pls no magic strings

Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
```diff
@@ -109,13 +110,7 @@ private void AddAnomalyToBody(Entity<InnerBodyAnomalyComponent> ent)
...
             var message = Loc.GetString(ent.Comp.StartMessage);
             var wrappedMessage = Loc.GetString("chat-manager-server-wrap-message", ("message", message));
-            _chat.ChatMessageToOne(ChatChannel.Server,
-                message,
-                wrappedMessage,
-                default,
-                false,
-                mindComponent.Session.Channel,
-                _messageColor);
+            _chat.SendChannelMessage(wrappedMessage, "GameMessage", null, null, new HashSet<ICommonSession>() { mindComponent.Session });
```
Fildrance, line 113: pls no magic strings and blow too..

Content.Server/Antag/AntagSelectionSystem.API.cs
```diff
@@ -319,8 +319,7 @@ public void SendBriefing(
         if (!string.IsNullOrEmpty(briefing))
         {
             var wrappedMessage = Loc.GetString("chat-manager-server-wrap-message", ("message", briefing));
-            _chat.ChatMessageToOne(ChatChannel.Server, briefing, wrappedMessage, default, false, session.Channel,
-                briefingColor);
+            _chat.SendChannelMessage(wrappedMessage, "GameMessage", null, null, new HashSet<ICommonSession>() { session });
```
Fildrance, line 322: magic string

Content.Server/Bed/Cryostorage/CryostorageSystem.cs
```diff
@@ -290,7 +290,7 @@ protected override void OnInsertedContainer(Entity<CryostorageComponent> ent, re
 
         var msg = Loc.GetString(locKey, ("time", comp.GracePeriod.TotalMinutes));
         if (TryComp<ActorComponent>(args.Entity, out var actor))
-            _chatManager.ChatMessageToOne(ChatChannel.Server, msg, msg, uid, false, actor.PlayerSession.Channel);
+            _chatManager.SendChannelMessage(msg, "GameMessage", null, null, new HashSet<ICommonSession>() { actor.PlayerSession });
```
Fildrance, line 293: magic string

Content.Server/Chat/Commands/LOOCCommand.cs
```diff
@@ -35,7 +35,7 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
             if (string.IsNullOrEmpty(message))
                 return;
 
-            _e.System<ChatSystem>().TrySendInGameOOCMessage(entity, message, InGameOOCChatType.Looc, false, shell, player);
+            _e.System<ChatSystem>().TrySendInGameOOCMessage(entity, message, player);
```
Fildrance, line 38: wierd....  u have communication way (channel) for ooc but there u are using separate method... is that correct?

Content.Client/Chat/Managers/ChatManager.cs
SlamBamActionman: There is a PR by ElectroSr that does this; it would have some minor merge conflicts, but I agree it would be nice to have.

Content.Client/Chat/Managers/ContentMarkupTagManager.cs
SlamBamActionman: There is a similar implementation for RobustToolbox's MarkupTagTypes that does it this way, and cites sandboxing as the reason why. So it might be a limitation we have to live with. I agree it's not optimal.

Content.Server/Content.Server.csproj
```diff
@@ -26,5 +26,8 @@
     <ProjectReference Include="..\RobustToolbox\Robust.Server\Robust.Server.csproj" />
     <ProjectReference Include="..\Content.Shared\Content.Shared.csproj" />
   </ItemGroup>
+  <ItemGroup>
```
Fildrance, line 29: can be safely removed

Content.Client/Entry/EntryPoint.cs
SlamBamActionman: Very good question; I think this is from me trying to get the content markup working, finding another fix and forgetting to remove this line.

Content.Server/GameTicking/GameTicker.cs
```diff
@@ -115,7 +115,7 @@ public override void Shutdown()
         private void SendServerMessage(string message)
         {
             var wrappedMessage = Loc.GetString("chat-manager-server-wrap-message", ("message", message));
-            _chatManager.ChatMessageToAll(ChatChannel.Server, message, wrappedMessage, default, false, true);
+            _chatManager.SendChannelMessage(wrappedMessage, "Server", null, null);
```
Fildrance, line 118: eeeeh, why use overload with exra null parameters? it looks wierd...

Content.Server/Chat/Commands/LOOCCommand.cs
SlamBamActionman: It's just a helper method; could very easily be replaced with SendChannelMessage entirely (as soon as I handle the whole magic strings issue)

Content.Server/Salvage/SalvageSystem.Runner.cs
```diff
@@ -68,16 +68,8 @@ private void Announce(EntityUid mapUid, string text)
...
-        _chat.ChatMessageToManyFiltered(
-            Filter.BroadcastMap(mapId),
-            ChatChannel.Radio,
-            text,
-            text,
-            _mapManager.GetMapEntityId(mapId),
-            false,
-            true,
-            null);
+        // CHAT-TODO: This is wrong - should not take the mapUid as an argument and instead it should be baked into the radio prototype
```
Fildrance, line 71: how is it even used! Q_Q

Content.Client/Chat/Managers/ChatManager.cs
Fildrance: maybe a little of todo with a link to separate issue for cleanup? even if issue will be placeholder - u would be able to fill it up later :godo:

Content.Client/Chat/Managers/ChatManager.cs
Fildrance: sure, valid. i agree.

Content.Client/Chat/Managers/ContentMarkupTagManager.cs
Fildrance: well, i guess we are stuck with this. Well it is readable at least!

Content.Server/Chat/Commands/LOOCCommand.cs
Fildrance: oh sorry then! if its just helper for usual way - its fine! i was just confused coz its kinda looking like old dirty methods.

Content.Server/Administration/Commands/DSay.cs
```diff
@@ -33,8 +35,7 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
             if (string.IsNullOrEmpty(message))
                 return;
 
-            var chat = _e.System<ChatSystem>();
-            chat.TrySendInGameOOCMessage(entity, message, InGameOOCChatType.Dead, false, shell, player);
+            _chat.SendChannelMessage(message, "Dead", player, entity);
```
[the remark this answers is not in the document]
Fildrance: So maybe do some 'public static class ChatChannels' and declare all of channels that are 'usually required' there (stuff that u do not think people gonna easily remove from protos? : D). Declare them with ProtoIdValidate attribute and add xml-doc that this is just helper list for channel names constants (so ppl won't think they NEED to add const there for it to actually work')

FairlySadPanda (commented):
I've left random comments across the files, mostly just flagging magic strings... but the overall review is that this has a few revisions it needs to do:

1. Avoidable API changes should be avoided for external callers. A substantial amount of the file changes are oneliners that change the API call from the old chat management call to the new system. This repetitive work is pointless, it makes the PR harder to review, and can be remedied by keeping the old API around when possible and adapting the old API to the new one via adaption functions.
2. There's a few instances of code that's sitting next to old code when it would be better broken into one or more partial classes. This would allow the PR to be largely append-only, minimizing downstream impact and making what you're trying to do clearer.
3. The "compose channels of specific conditions" design is brittle and prescriptive. Any change to any of the conditions could break the design of any number of channels. You can get the same level of flexibility here by encapsulating the send/receive/mutation rules of a given channel by creating C# classes that control the "rules" that a channel uses, with exposed variables and options to configure that ruleset. This has the advantage of balancing what you want to do - an extensible, modular system - with keeping complexity low.
4. The process of sending a chat message from emission to reception requires domentation and graphing. As part of this you need to demonstrate how other developers would extend or listen in to parts of the system to customize it via raised events.
5. Engineering "server-side" and "game-side" chat using the same systems I think is a dead end. They really should be different systems. It's part of the complexity of the blob-like current system.
6. This really needs an understanding from the admins of what features the admins want. Chat is a key aspect of the game for admins and the current tools for dealing with it are insufficient; sanitization is only one half of the problem. This system needs to support per-message editing and deletion and mass deletion of chat messages. I'd encourage you to sit down with @nikthechampiongr and the rest of the admin team and get the requirements off them.

Overall this is movement in the right direction but it needs reigning in in places. Deffo for the best you didn't make my mistake of finishing an entire end-to-end attempt before asking for feedback.

sleepyyapril:
Does this currently support a BeforeChatSendEvent or similar? If not, it should!

Kadeo64:
holy shit praise slambam

lizelive:
Finally can raise TTS PR after this is merged.

Content.Server/Chat/ChatConditions/ActiveTelecomsEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+[DataDefinition]
+public sealed partial class ActiveTelecomsEntityChatCondition : EntityChatCondition
+{
+
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+    [Dependency] private readonly IEntityManager _entityManager = default!;
+
+    public override HashSet<EntityUid> FilterConsumers(HashSet<EntityUid> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
```
Fildrance, line 28: a thing to consider - maybe its possible to do DI inject on c-tor? Are those re-created often? It feels like it should most likely stay alive whole game session!

Content.Server/Chat/ChatConditions/ActiveTelecomsEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+        var returnConsumers = new HashSet<EntityUid>();
+
+        if (!channelParameters.TryGetValue(DefaultChannelParameters.SenderEntity, out var senderEntity) ||
+            !channelParameters.TryGetValue(DefaultChannelParameters.RadioChannel, out var radioChannel) ||
+            !_prototypeManager.TryIndex((string)radioChannel, out RadioChannelPrototype? radioPrototype) ||
+            !_entityManager.TryGetComponent<TransformComponent>((EntityUid)senderEntity, out var sourceTransform))
+            return new HashSet<EntityUid>();
+
+        var activeServer = radioPrototype.LongRange;
```
Fildrance, line 37: isActiveServcerRequired sounds better

Content.Server/Chat/ChatConditions/ActiveTelecomsEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            var servers = _entityManager.EntityQuery<TelecomServerComponent, EncryptionKeyHolderComponent, ApcPowerReceiverComponent, TransformComponent>();
+            var sourceMapId = sourceTransform.MapID;
+            foreach (var (_, keys, power, transform) in servers)
+            {
+                if (transform.MapID == sourceMapId &&
+                    power.Powered &&
+                    keys.Channels.Contains(radioPrototype.ID))
+                {
+                    activeServer = true;
```
Fildrance, line 49: I'd advice to not reuse variables, this is majorly harder to understand as there is acutally 2 variables - there is marker 'isActiveServcerRequired' which is determined by radioPrototype.LongRange, and can be inlined maybe (into line 39) and there is that makrer, if there is active server. Also if you found that required active server is existing, you might want to stop iterating and return early.

Content.Server/Chat/ChatConditions/ActiveTelecomsEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+public sealed partial class ActiveTelecomsEntityChatCondition : EntityChatCondition
+{
+
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+    [Dependency] private readonly IEntityManager _entityManager = default!;
+
+    public override HashSet<EntityUid> FilterConsumers(HashSet<EntityUid> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+        var returnConsumers = new HashSet<EntityUid>();
```
Fildrance, line 29: don't allocate before you need, move this to where you actually require it

Content.Server/Chat/ChatConditions/AdminOOCEnabledSessionChatCondition.cs
```diff
@@ -0,0 +1,24 @@
...
+[DataDefinition]
+public sealed partial class AdminOOCEnabledSessionChatCondition : SessionChatCondition
+{
+    [Dependency] private readonly IConfigurationManager _configurationManager = default!;
+
+    public override HashSet<ICommonSession> FilterConsumers(HashSet<ICommonSession> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        return _configurationManager.GetCVar<bool>(CCVars.AdminOocEnabled) ? consumers : new HashSet<ICommonSession>();
```
Fildrance, line 22: How does this returning hash set is used? if other code is not adding stuff right into it - it might be better to use some reference to immutable empty entity - then u won't have to re-created hashsets, polluting memory with useless stuff.

Content.Server/Chat/ChatConditions/HasComponentEntityChatCondition.cs
```diff
@@ -0,0 +1,51 @@
...
+    [DataField]
+    public string? Component;
+
+    public override HashSet<EntityUid> FilterConsumers(HashSet<EntityUid> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        var returnConsumers = new HashSet<EntityUid>();
+
+        if (Component != null)
```
Fildrance, line 37: here  i see that early return would benefit readability - its called 'reduce nesting' :3

Content.Server/Chat/ChatConditions/HasComponentEntityChatCondition.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            foreach (var consumer in consumers)
+            {
+                var comp = _componentFactory.GetRegistration(Component, true);
```
Fildrance, line 41: this clearly should be done once. And maybe we can do it not inside FilterConsumers?

Content.Server/Chat/ChatConditions/IsAboveCritEntityChatCondition.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Content.Server.Chat.ChatConditions;
+
+/// <summary>
+/// Checks if the consumer is alive and above crit; does not check for consciousness e.g. sleeping.
+/// </summary>
+[DataDefinition]
+public sealed partial class IsAboveCritEntityChatCondition : EntityChatCondition
+{
+
```
Fildrance, line 23: pls remove empty line

Content.Server/Chat/ChatConditions/IsAboveCritEntityChatCondition.cs
```diff
@@ -0,0 +1,37 @@
...
+    [Dependency] private readonly IEntitySystemManager _entitySystem = default!;
+
+    public override HashSet<EntityUid> FilterConsumers(HashSet<EntityUid> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        if (_entitySystem.TryGetEntitySystem<MobStateSystem>(out var mobStateSystem))
+        {
+            return consumers.Where(x => !mobStateSystem.IsIncapacitated(x)).ToHashSet();
```
Fildrance, line 32: linq! REEE~
pls avoid linq when mentally possible

Content.Server/Chat/ChatConditions/IsGhostSessionChatCondition.cs
```diff
@@ -0,0 +1,22 @@
...
+[DataDefinition]
+public sealed partial class IsGhostSessionChatCondition : SessionChatCondition
+{
+    [Dependency] private readonly EntityManager _entityManager = default!;
+
+    public override HashSet<ICommonSession> FilterConsumers(HashSet<ICommonSession> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        return consumers.Where(x => _entityManager.HasComponent<GhostComponent>(x.AttachedEntity)).ToHashSet();
```
Fildrance, line 20: same here

Content.Server/Chat/ChatConditions/OOCEnabledSessionChatCondition.cs
```diff
@@ -0,0 +1,24 @@
...
+[DataDefinition]
+public sealed partial class OOCEnabledSessionChatCondition : SessionChatCondition
+{
+    [Dependency] private readonly IConfigurationManager _configurationManager = default!;
+
+    public override HashSet<ICommonSession> FilterConsumers(HashSet<ICommonSession> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        return _configurationManager.GetCVar<bool>(CCVars.OocEnabled) ? consumers : new HashSet<ICommonSession>();
```
Fildrance, line 22: Is that also changing when game mode changes? Just not sure : (

Content.Server/Chat/ChatConditions/RadioReceiverEntityChatCondition.cs
```diff
@@ -0,0 +1,53 @@
...
+        IoCManager.InjectDependencies(this);
+        var returnConsumers = new HashSet<EntityUid>();
+
+        if (!channelParameters.TryGetValue(DefaultChannelParameters.SenderEntity, out var senderEntity) ||
+            !channelParameters.TryGetValue(DefaultChannelParameters.RadioChannel, out var radioChannel))
+            return new HashSet<EntityUid>();
+
+        foreach (var consumer in consumers)
+        {
+            if (_entityManager.TryGetComponent<WearingHeadsetComponent>((EntityUid)senderEntity, out var headsetComponent))
```
Fildrance, line 33: extract method so cycle would have just method call

Content.Server/Chat/ChatConditions/RadioReceiverEntityChatCondition.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                if (_entityManager.TryGetComponent(headsetComponent.Headset, out EncryptionKeyHolderComponent? keys) &&
+                    keys.Channels.Contains((string)radioChannel))
```
Fildrance, line 36: hmmm... doesn't this require check for EMP or something?

Content.Server/Chat/ChatConditions/RadioReceiverEntityChatCondition.cs
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    returnConsumers.Add(consumer);
+                }
+            }
+            else if (_entityManager.TryGetComponent<IntrinsicRadioReceiverComponent>((EntityUid)senderEntity, out var intrinsicRadioTransmitterComponent) &&
+                     _entityManager.TryGetComponent<ActiveRadioComponent>((EntityUid)senderEntity, out var activeRadioComponent))
+            {
+                if (activeRadioComponent.Channels.Contains((string)radioChannel))
```
Fildrance, line 44: can be joined with previous statement. Not 100% sure it would be more readable, but it seems like desired thing to do, according overall code base :3

Fildrance, line 44: looks similar to keys check in previous block - maybe extract method like 'AddToHashSetIfChannelKeysContain' (this naming sucks but i hope you get it)

Content.Server/Chat/ChatConditions/RadioTransmissionEntityChatCondition.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Content.Server.Chat.ChatConditions;
+
+/// <summary>
+/// Checks if the consumer has a method to transmit radio messages
+/// </summary>
+[DataDefinition]
+public sealed partial class RadioTransmissionEntityChatCondition : EntityChatCondition
+{
+
```
Fildrance, line 19: empty line not justified

Content.Server/Chat/ChatConditions/RadioTransmissionEntityChatCondition.cs
```diff
@@ -0,0 +1,52 @@
...
+            return new HashSet<EntityUid>();
+
+        foreach (var consumer in consumers)
+        {
+            if (_entityManager.TryGetComponent<WearingHeadsetComponent>((EntityUid)senderEntity, out var headsetComponent))
+            {
+                if (_entityManager.TryGetComponent(headsetComponent.Headset, out EncryptionKeyHolderComponent? keys) &&
+                    keys.Channels.Contains((string)radioChannel))
+                {
+                    returnConsumers.Add(consumer);
```
Fildrance, line 38: emp checks are required? or EMP removes component itself :<

Content.Server/Chat/ChatConditions/RadioTransmissionEntityChatCondition.cs
```diff
@@ -0,0 +1,52 @@
...
+        IoCManager.InjectDependencies(this);
+        var returnConsumers = new HashSet<EntityUid>();
+
+        if (!channelParameters.TryGetValue(DefaultChannelParameters.SenderEntity, out var senderEntity) ||
+            !channelParameters.TryGetValue(DefaultChannelParameters.RadioChannel, out var radioChannel))
+            return new HashSet<EntityUid>();
+
+        foreach (var consumer in consumers)
+        {
+            if (_entityManager.TryGetComponent<WearingHeadsetComponent>((EntityUid)senderEntity, out var headsetComponent))
```
Fildrance, line 33: can be done outside of cycle

Content.Server/Chat/ChatConditions/RangeEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+using Content.Shared.Chat.Prototypes;
+using Robust.Shared.Network;
+using Robust.Shared.Player;
+
+namespace Content.Server.Chat.ChatConditions;
+
+[DataDefinition]
+public sealed partial class RangeEntityChatCondition : EntityChatCondition
+{
+
```
Fildrance, line 14: empty line REEE

Content.Server/Chat/ChatConditions/RangeEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+    [DataField]
+    public float MaximumRange = 5f;
+
+    [Dependency] private readonly IEntityManager _entityManager = default!;
+
+    public override HashSet<EntityUid> FilterConsumers(HashSet<EntityUid> consumers, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        if (channelParameters.TryGetValue(DefaultChannelParameters.SenderEntity, out var senderEntity) &&
```
Fildrance, line 33: maybe add some protected methods to ChatCondition base classes, do this cleaner? just a thing to think about

Content.Server/Chat/ChatConditions/RangeEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+                        distance >= MinimumRange)
+                    {
+                        Logger.Debug(distance.ToString());
+                        returnConsumers.Add(entity);
+                    }
+                }
+            }
+            return returnConsumers;
+        }
+        return new HashSet<EntityUid>();
```
Fildrance, line 57: early return, like in other classes, would seem better. at least be consistent :3

Content.Server/Chat/ChatConditions/RangeEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    if (transform.MapID != sourceTransform.MapID)
+                        continue;
+
+                    // even if they are a ghost hearer, in some situations we still need the range
+                    if (sourceTransform.Coordinates.TryDistance(_entityManager, transform.Coordinates, out var distance) &&
+                        distance < MaximumRange &&
+                        distance >= MinimumRange)
+                    {
+                        Logger.Debug(distance.ToString());
```
Fildrance, line 50: todo to fix that - i don't think u are going to leave that as it is :)

Content.Server/Chat/ChatConditions/RangeEntityChatCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+        IoCManager.InjectDependencies(this);
+
+        if (channelParameters.TryGetValue(DefaultChannelParameters.SenderEntity, out var senderEntity) &&
+            _entityManager.TryGetComponent<TransformComponent>((EntityUid)senderEntity, out var sourceTransform))
+        {
+            var returnConsumers = new HashSet<EntityUid>();
+
+            foreach (var entity in consumers)
+            {
+                if (_entityManager.TryGetComponent<TransformComponent>(entity, out var transform))
```
Fildrance, line 40: reduce nesting by reversing if and continue-ing right away

Content.Server/Chat/Commands/ChatCommand.cs
```diff
@@ -0,0 +1,40 @@
+using Content.Server.Chat.Managers;
+using Content.Server.Chat.Systems;
+using Content.Shared.Administration;
+using Robust.Shared.Console;
+using Robust.Shared.Enums;
+
+namespace Content.Server.Chat.Commands
+{
```
Fildrance, line 8: use file-scoped namespace for new files pls :(

Content.Shared/Chat/ChatChannel.cs
```diff
@@ -19,7 +19,7 @@ public enum ChatChannel : ushort
         Whisper = 1 << 1,
```
Fildrance, line 19: This beauty should be at some point replaced with protos, mostly (and ProtoId i guess). I'm not 100% sure if that should be done on that PR.

Fildrance: flags combinations probably could be replaced by some simple helper methods than

Content.Shared/Chat/ChatConditions/IsAdminSessionChatCondition.cs
```diff
@@ -0,0 +1,23 @@
...
+    [Dependency] private readonly ISharedAdminManager _admin = default!;
+    public override HashSet<ICommonSession> FilterConsumers(HashSet<ICommonSession> consumers, Dictionary<Enum, object>? channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        return consumers.Where(x => _admin.IsAdmin(x, IncludeDeadmin)).ToHashSet();
+    }
+
+    [DataField]
+    public bool IncludeDeadmin;
```
Fildrance, line 22: property before method

Content.Shared/Chat/ChatConditions/IsAdminSessionChatCondition.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Content.Shared.Chat.ChatConditions;
+
+/// <summary>
+/// Checks if the consumers are admins.
+/// </summary>
+[DataDefinition]
+public sealed partial class IsAdminSessionChatCondition : SessionChatCondition
+{
+    [Dependency] private readonly ISharedAdminManager _admin = default!;
+    public override HashSet<ICommonSession> FilterConsumers(HashSet<ICommonSession> consumers, Dictionary<Enum, object>? channelParameters)
```
Fildrance, line 14: empty line to delimit them

Content.Shared/Chat/ChatModifier.cs
```diff
@@ -0,0 +1,309 @@
+﻿using System.Linq;
+using Robust.Shared.Utility;
+
+namespace Content.Shared.Chat;
+
```
Fildrance, line 5: excessive empty line

Content.Shared/Chat/ChatModifier.cs
```diff
@@ -0,0 +1,309 @@
...
+
+/// <summary>
+/// Used to add markup nodes to FormattedMessages, preparing them to later be processed by MarkupTagManager.
+/// </summary>
+[Serializable]
+[DataDefinition]
+[Virtual]
+public partial class ChatModifier
```
Fildrance, line 13: make it abstract maybe if it should not be instatiated by itself?

Content.Shared/Chat/ChatModifier.cs
```diff
@@ -0,0 +1,309 @@
...
+{
+    /// <summary>
+    /// Returns a FormattedMessage after it has been processed by the node supplier.
+    /// </summary>
+    /// <param name="message">The message to be processed.</param>
+    /// <param name="channelParameters">Any parameters that can be handled by the suppliers.</param>
+    /// <returns></returns>
+    public virtual FormattedMessage ProcessChatModifier(FormattedMessage message, Dictionary<Enum, object> channelParameters)
```
Fildrance, line 21: maybe it should be abstract, to mark where in derived types real work code should be placed?

Content.Shared/Chat/ChatModifier.cs
```diff
@@ -0,0 +1,309 @@
...
+[DataDefinition]
+[Virtual]
+public partial class ChatModifier
+{
+    /// <summary>
+    /// Returns a FormattedMessage after it has been processed by the node supplier.
+    /// </summary>
+    /// <param name="message">The message to be processed.</param>
+    /// <param name="channelParameters">Any parameters that can be handled by the suppliers.</param>
+    /// <returns></returns>
```
Fildrance, line 20: Either remove it, or write down specific info about returned object :<

Content.Shared/Chat/ChatModifiers/BoldFulltextChatModifier.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Robust.Shared.Utility;
+
+namespace Content.Shared.Chat.ChatModifiers;
+
+/// <summary>
+/// Wraps the entire message in an [italic] tag.
```
Fildrance, line 6: bold and italic are not same, fix comment)

Content.Shared/Chat/ChatModifiers/BoldTagChatModifier.cs
```diff
@@ -0,0 +1,30 @@
...
+[DataDefinition]
+public sealed partial class BoldTagChatModifier : ChatModifier
+{
+    [DataField]
+    public string? TargetNode = null;
+
+    public override FormattedMessage ProcessChatModifier(FormattedMessage message, Dictionary<Enum, object> channelParameters)
+    {
+
+        if (TargetNode != null && message.TryFirstOrDefault(x => x.Name == TargetNode, out var foundTag))
```
Fildrance, line 23: maybe early return in case field is null will be more readable?

Content.Shared/Chat/ChatModifiers/BubbleProviderChatModifier.cs
```diff
@@ -0,0 +1,30 @@
...
+    [DataField]
+    public SpeechType SpeechType = SpeechType.Say;
+
+    public override FormattedMessage ProcessChatModifier(FormattedMessage message, Dictionary<Enum, object> channelParameters)
+    {
+        return InsertOutsideTag(InsertOutsideTag(message, new MarkupNode("BubbleHeader", new MarkupParameter((int)SpeechType), null), "EntityNameHeader"), new MarkupNode("BubbleMessage", null, null), "MainMessage");
+    }
+}
+
+// This enum needs to be merged with the one in SpeechBubble.cs
```
Fildrance, line 23: should be todo :3

Content.Shared/Chat/ChatModifiers/ColorEntityNameHeaderChatModifier.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Content.Shared.Chat.ChatModifiers;
+
+/// <summary>
+/// Wraps the [EntityNameHeader] tag in a [color] tag, should the player have colored names enabled.
+/// </summary>
+[Serializable]
+[DataDefinition]
+public sealed partial class ColorEntityNameHeaderChatModifier : ChatModifier
+{
+
```
Fildrance, line 17: empty line without reasons

Content.Shared/Chat/ChatModifier.cs
Fildrance, line 21: Having Enum as key is kinda wierd. I honestly am not sure about if thats any better then string or any other type, or is it worse... Maybe make some wrapper around string is better? A thing to think and talk about at least

Content.Shared/Chat/ChatModifiers/ColorEntityNameHeaderChatModifier.cs
```diff
@@ -0,0 +1,59 @@
...
+        IoCManager.InjectDependencies(this);
+
+        var colorName = _config.GetCVar(CCVars.ChatEnableColorName);
+        if (colorName && message.TryFirstOrDefault(x => x.Name == "EntityNameHeader", out var nameHeader))
+        {
+            var name = nameHeader.Value.StringValue;
+            if (name != null)
+            {
+                var color = Color.TryFromHex(GetNameColor(name));
+                if (color != null)
```
Fildrance, line 33: maybe reduce nesting by inverting conditions and early returning? for better readability :<

Content.Shared/Chat/ChatModifiers/EntityNameHeaderChatModifier.cs
```diff
@@ -0,0 +1,44 @@
...
+[DataDefinition]
+public sealed partial class EntityNameHeaderChatModifier : ChatModifier
+{
+    [Dependency] private readonly EntityManager _entityManager = default!;
+
+    public override FormattedMessage ProcessChatModifier(FormattedMessage message, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+
+        if (channelParameters.TryGetValue(DefaultChannelParameters.SenderEntity, out var sender) && sender is EntityUid)
```
Fildrance, line 21: early return would be better for readability :<

Fildrance, line 21: use pattern matching inteaad of casting on line 23

Content.Shared/Chat/ChatModifiers/EntityNameHeaderChatModifier.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            var senderEntity = (EntityUid)sender;
+            MetaDataComponent? metaData = null;
+            string? name;
```
Fildrance, line 25: you are not using it in next block, don't declare it here then :D declare it later

Content.Shared/Chat/ChatModifiers/SanitizeRadioPrefixChatModifier.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Content.Shared.Chat.ChatModifiers;
+
+/// <summary>
+/// Removes any radio chat prefixes from the first text node.
+/// </summary>
+[Serializable]
+[DataDefinition]
+public sealed partial class SanitizeRadioPrefixChatModifier : ChatModifier
+{
+
```
Fildrance, line 19: excessive empty line

Content.Shared/Chat/ChatModifiers/SanitizeRadioPrefixChatModifier.cs
```diff
@@ -0,0 +1,73 @@
...
+    public override FormattedMessage ProcessChatModifier(FormattedMessage message, Dictionary<Enum, object> channelParameters)
+    {
+
+        var returnMessage = new FormattedMessage();
+        if (!message.Nodes.TryFirstOrDefault(x => x.Name == null, out var firstTextNode))
+        {
+            return message;
+        }
```
Fildrance, line 27: new line after if block. maybe make it without braces? :< also move creation of FormattedMessage lower so u create only if u use

Content.Shared/Chat/ChatModifiers/SanitizeRadioPrefixChatModifier.cs
```diff
@@ -0,0 +1,73 @@
...
+        var nodeEnumerator = message.GetEnumerator();
+
+        while (nodeEnumerator.MoveNext())
+        {
+            var node = nodeEnumerator.Current;
+            if (node == firstTextNode)
+            {
+                if (firstTextNode.Value.TryGetString(out var textNodeValue))
```
Fildrance, line 35: join if statements

Content.Shared/Chat/ChatModifiers/InsertLoCChatModifier.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// The node that the string should be inserted next to.
+    /// </summary>
+    [DataField]
+    public string? TargetNode = null;
+
+    /// <summary>
+    /// The string that should be inserted.
+    /// </summary>
+    [DataField]
+    public string LocString = "";
```
Fildrance, line 33: could this be LocId type? then you can name field 'ToInsert' or something

Content.Shared/Chat/ChatModifiers/InsertLoCChatModifier.cs
```diff
@@ -0,0 +1,47 @@
...
+using Content.Shared.CCVar;
+using Content.Shared.Decals;
+using Robust.Shared.Configuration;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
+
+namespace Content.Shared.Chat.ChatModifiers;
+
+/// <summary>
+/// Inserts an LoC string after a specific node. Useful for formatting certain messages, such as whispering.
```
Fildrance, line 11: MAYBE describing what LoC is here would be cool :< at least briefly

Content.Shared/Chat/ChatModifiers/InsertLoCChatModifier.cs
```diff
@@ -0,0 +1,47 @@
...
+/// </summary>
+[Serializable]
+[DataDefinition]
+public sealed partial class InsertLoCChatModifier : ChatModifier
+{
+    /// <summary>
+    /// If false, the string will be inserted before the node.
+    /// </summary>
+    [DataField]
+    public bool AfterNode = true;
```
Fildrance, line 21: a little bit ugly and counterintuitive, maybe base class fro insert and insertAfter/insertBefore derived types would be better? so you would just hide this stuff from ppl who write yaml.

Content.Shared/Chat/ChatModifiers/InsertLoCChatModifier.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// <summary>
+    /// The string that should be inserted.
+    /// </summary>
+    [DataField]
+    public string LocString = "";
+
+    public override FormattedMessage ProcessChatModifier(FormattedMessage message, Dictionary<Enum, object> channelParameters)
+    {
+        if (TargetNode != null)
```
Fildrance, line 37: early return!111

Content.Shared/Chat/ChatModifiers/InsertRadioNameChatModifier.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// If false, the string will be inserted before the node.
+    /// </summary>
+    [DataField]
+    public bool AfterNode = true;
+
+    public override FormattedMessage ProcessChatModifier(FormattedMessage message, Dictionary<Enum, object> channelParameters)
+    {
+        IoCManager.InjectDependencies(this);
+        if (channelParameters.TryGetValue(DefaultChannelParameters.RadioChannel, out var radioChannel) &&
```
Fildrance, line 36: early returns  all the way

PJB3005:
@Fildrance you are spamming this PR. Use GitHub's review feature properly.

Content.Shared/Chat/ChatModifiers/MainMessageChatModifier.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Robust.Shared.Utility;
+
+namespace Content.Shared.Chat.ChatModifiers;
+
+/// <summary>
+/// Wraps a [MainMessage] tag around all text nodes, which can later be utilized by other chat modifiers.
```
Fildrance, line 6: maybe elaborate what actually main message is? it would be a good place for ppl who are trying to find out how all this stuff works :3

Fildrance: other then the actual docs, ofc

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -188,189 +175,333 @@ public void SendHookOOC(string sender, string message)
...
-    {
-        if (!recordReplay && !filter.Recipients.Any())
-            return;
+    #endregion
+
+    #region Private API
 
-        var clients = new List<INetChannel>();
-        foreach (var recipient in filter.Recipients)
+    private List<CommunicationChannelPrototype> HandleChildChannels(CommunicationChannelPrototype communicationChannel, List<ProtoId<CommunicationChannelPrototype>>? childChannels)
```
Fildrance, line 451: its not handle, but filter?

Fildrance (commented):
Added some ideas for minor refactors, waiting for condition refactor (merge session/entity condition) and related changes

Content.Shared/Chat/ChatModifiers/InsertLoCChatModifier.cs
Fildrance: ``` = default! ``` might be more correct to describe that its default value is just to suppres compiler warning

Content.Server/GameTicking/GameTicker.cs
SlamBamActionman: Leaving this in since it indicates it's a server-originating message.

Content.Client/Chat/Managers/ChatManager.cs
```diff
@@ -12,26 +15,20 @@ internal sealed class ChatManager : IChatManager
     [Dependency] private readonly IClientConsoleHost _consoleHost = default!;
     [Dependency] private readonly IClientAdminManager _adminMgr = default!;
     [Dependency] private readonly IEntitySystemManager _systems = default!;
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
```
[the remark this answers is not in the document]
SlamBamActionman: We're ripping out the old system wholesale since it'd be an excessive amount of work to ensure supporting systems allow for both old and new chatcode. I understand this may end up being harder to review, which is why we will be providing assistance for how to do it (especially considering its size). Where possible we're including the old functions in ChatSystem.Obsolete.

Content.Client/Chat/Managers/ChatManager.cs
```diff
@@ -12,26 +15,20 @@ internal sealed class ChatManager : IChatManager
...
-        // See server-side manager. This just exists for shared code.
-    }
-
-    public void SendAdminAlert(EntityUid player, string message)
-    {
-        // See server-side manager. This just exists for shared code.
-    }
-
-    public void SendMessage(string text, ChatSelectChannel channel)
+    public void SendMessage(string text, ChatSelectChannel channel, RadioChannelPrototype? radioChannel)
```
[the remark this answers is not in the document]
SlamBamActionman: I'd rather just rework the entire thing to be cleaner, it's a collision point between newcode and oldcode.

Resources/Prototypes/Chat/communication_channels.yml
```diff
@@ -0,0 +1,456 @@
...
+        - !type:HasComponentEntityChatCondition
+          component: Sleeping
+          inverted: true
+  consumeCollections:
+  - consumeSessionChatConditions:
+    - !type:AllSessionChatCondition
+      useEntityChatConditions: true
+    entityChatConditions:
+    - !type:RangeEntityChatCondition
+      maximumRange: 10
```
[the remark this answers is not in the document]
SlamBamActionman: System should be robust enough at this point to adjust for this at a later stage if needed.

Resources/Prototypes/Chat/communication_channels.yml
```diff
@@ -0,0 +1,456 @@
...
+    locString: ": "
+  - !type:BubbleProviderChatModifier
+    speechType: LOOC
+  backupChildCommunicationChannels:
+  - Dead #If dead & not Moderator, this should attempt the Dead channel
+  chatChannels:
+  - LOOC
+
+- type: communicationChannel
+  id: OOC
```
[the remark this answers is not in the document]
SlamBamActionman: There are different types of OOC communication, and we may want more in the future. Since that means they share chat conditions and modifiers, I think it's fine that we're using the same system for both.

Content.Server/Chat/Managers/ChatManager.cs
```diff
@@ -188,189 +175,333 @@ public void SendHookOOC(string sender, string message)
...
-    public void TrySendOOCMessage(ICommonSession player, string message, OOCChatType type)
+    /// <param name="message">The message that is to be sent.</param>
+    /// <param name="communicationChannel">The communication channel prototype to use as a base for how the message should be treated.</param>
+    /// <param name="senderSession">The session that sends the message. If null, this means the server is sending the message.</param>
+    /// <param name="senderEntity">The entity designated as "sending" the message. If null, the message is coming directly from the session/server.</param>
+    /// <param name="usedCommsChannels">Tracks the communication channels used. Helps prevent infinitely recursive messages.</param>
+    /// <param name="targetSessions">Any sessions that should be specifically targetted (still needs to comply with channel consume conditions). If you are targetting multiple sessions you should likely use a consumeCollection instead of this.</param>
+    /// <param name="supplierParameters">Parameters that may be used by MarkupSuppliers; these are not passed on to the client.</param>
+    public void SendChannelMessage(
```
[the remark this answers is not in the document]
SlamBamActionman: Regions are being used for this at the moment. If enough people yell at me, I'll look into breaking it apart before release.

image

Content.Shared/Chat/ChatChannel.cs
SlamBamActionman: This was moved over to just be for filtering channels in the chatbox UI, just to keep the client functionality similar.

Content.Shared/Chat/ChatModifiers/ObfuscateTextChatModifier.cs
```diff
@@ -0,0 +1,80 @@
...
+        var charArray = message.ToCharArray();
+
+        for (var i = 0; i < charArray.Length; i++)
+        {
+            if (char.IsWhiteSpace((charArray[i])))
+            {
+                continue;
+            }
+
+            if (_random.Prob(chance))
```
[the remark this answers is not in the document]
SlamBamActionman: Apparently! I think it should work the way you describe, but that'll be a future feature.

Content.Shared/Chat/ListenerComponent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Content.Shared.Chat.Prototypes;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Chat;
+
+[RegisterComponent]
+public abstract partial class ListenerComponent : Component
+{
+    [DataField]
+    public ChatChannel? FilteredTypes;
```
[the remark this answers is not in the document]
SlamBamActionman: It's using ChatChannelMedium now instead

Fildrance:
Writing it down just as a reminder, major things left to do:
* recursiveness prevention
* marking every message with unique UUID 
* recording/nuking messages the chat for admin/logscan usage

SlamBamActionman:
Gonna write down a summary on how to review this whole thing (that will eventually replace the PR description). 

And I'll do it in this comment! [WIP]

VerinSenpai:
You're the goat. Think you can have this done in say, 3 hours?

ScarKy0:
> I know I know, I got jokes. Don't everyone laugh at once.

I think we should explode you

VerinSenpai:
*dies with hammers*

Fildrance:
> Is it reasonable to have something like a `ConjuctionChatCondition` so that these other conditions can be provided in a list instead of a deeply nested tree?

was already fixed, look at https://github.com/SlamBamActionman/space-station-14/blob/chat-refactor-tags/Resources/Prototypes/Chat/communication_channels.yml

and better look at last Slam comment as review instruction :>

walksanatora:
so I am confused. `AlwaysRelayedToChannels` is passed around *after* Server modifier right? so making languages could be easy as sending across a channel and if they have the lang you end it directly. if not it falls back. applied the "language" and then passes it back to IC chats right? eitherway amazing PR cant wait for it to be merged.

Resources/Prototypes/Chat/communication_channels.yml
```diff
@@ -0,0 +1,456 @@
...
+  clientModifiers:
+  - !type:InsertLocChatModifier
+    targetNode: SessionNameHeader
+    locString: "OOC: "
+    afterNode: false
+  - !type:BoldTagChatModifier
+    targetNode: SessionNameHeader
+  - !type:InsertLocChatModifier
+    targetNode: SessionNameHeader
+    locString: ": "
```
PJB3005, line 230: The server is responsible for formatting the location of the player name relative to the main message, yet the client is responsible for formatting all the rest. This does not make sense.

PJB3005: Furthermore, this entire approach to formatting chat messages is *completely* impossible to localize.

PJB3005 (changes requested):
This PR overgeneralizes the internals of chat processing to such an extent that it will make it significantly harder to use, understand, develop, and debug.

The external API (for gameplay systems that need to raise chat messages) is also significantly worse: Now instead of using expressive APIs that specify intent (and can easily take parameters sound as sounds), they have to shove everything through a single generic API with 6 parameters and a generic key/value one for everything else.

Furthermore, the simulation layering violation of moving **all** chat processing into `ChatManager` makes the situation *worse* than it already was.

Fildrance:
we'll do it other way

PR #33863, opened 2024-12-14
Remove skeletons

0leshe:
## About the PR
Totaly removes skeletons from the game

## Why / Balance
It was good idea to add skeleton as roleplay character, but its in the most of scenarios just antag who kills everyone who doesn't like him, so whats the point? Are agents not enough? Its also lowpop thing, so he can take control over whole station.

## Technical details
Removes midround event, body parts pictures, accent systems, sound of scream, names dataset, species, and prototype as well.

## Requirements

## Breaking changes

**Changelog**

:cl: Oleshe
- remove: Skeletons


Aeshus:
While I don't personally agree with this change, I would rather suggest that all you remove is the `ClosetSkeleton` gamerule so that they do not spawn normally.

It'll be a smaller diff and still allows for admins to use them if they so wish.

0leshe:
Oh, seems more logic and loyal... Thanks, i'll edit it tomorrow..

Djungelskog2:
dont remove the protos or any of that, just remove the event/ghost role.

ps3moira:
Hopefully these guys can return and be made a lot more easier to kill? Something something limb targeting

SlimmSlamm:
Awful PR. They're a Free Agent they can do whatever they want. There's plenty of times where I see them be friendly and other times at being an antag. They're incredibly fun to do both ways. You shouldn't have to remove an entire ghost role because of the way people play it how it's supposed to be played.

ArtisticRoomba:
Your build failed because you removed skeletons.

You can't just ctrl+f skeletons and delete files and then deal with the mountain of errors after that. If you want to do this right, just delete the event that spawns the ghost role.

Like others have said, I don't think it's really a smart idea to remove an entire ghost role (and the other things that use skeletons!) from existence just because the people that play them play to the rules.

If you want to make them more pivoted towards roleplay, maybe balance them a bit by toning down their innate brute resists? Instead of going for the nuclear option.

Djungelskog2:
The PR being actually broken aside, this is an issue with the free agent roletype and escalation rules in general. (I will continue to glaze beestation escalation rules)

kosticia:
I don't think that we need to remove one of the most free for roleplay role.

> Its also lowpop thing, so he can take control over whole station.

Biggest part of antagonist broken on low-pop, we can got two king-rats without sec or ninja that will summon dragon and break all airlocks

LordEclipse:
Just make the ghost role event happen at a certain pop, like 20 pop or something, many other events work on this logic.

PR #33867, opened 2024-12-15
Zombies keep their anomalies on zombification

pcaessayrs:
## About the PR
Infected people will no longer lose their anomaly when they turn into a zombie. Normally, an entity hosting an anomaly loses that anomaly when dying, which also is when they would turn into a zombie - instead they will turn into a zombie first, and their zombie will still have the anomaly.

## Why / Balance
This allows for anomaly zombies to exist, increasing the variety and interest of zombie types by allowing anomalies and zombies to interact.

## Technical details
InnerBodyAnomalySystem checks for PendingZombieComponent before removing the anomaly in OnMobStateChanged to dead.

## Media
anombies.webm
Shows a person turning into a zombie, still having anomaly powers, and then dying as a zombie and finally losing it.

## Requirements

## Breaking changes

None

**Changelog**

:cl:
- tweak: Anomaly hosts keep their anomaly when turning into a zombie.

### DIFF
--- a/Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
+++ b/Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
@@ -14,7 +14,6 @@ using Content.Shared.Database;
 using Content.Shared.Mobs;
 using Content.Shared.Popups;
 using Content.Shared.Whitelist;
-using Content.Shared.Zombies;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Physics.Events;
 using Robust.Shared.Prototypes;
@@ -187,8 +186,10 @@ public sealed class InnerBodyAnomalySystem : SharedInnerBodyAnomalySystem
         if (args.NewMobState != MobState.Dead)
             return;
 
-        if (EntityManager.HasComponent<PendingZombieComponent>(args.Target))
-            return; // Keep your anomaly when becoming a zombie
+        var ev = new BeforeRemoveAnomalyOnDeathEvent();
+        RaiseLocalEvent(args.Target, ev);
+        if (ev.Cancelled)
+            return;
 
         _anomaly.ChangeAnomalyHealth(ent, -2); //Shutdown it
     }
--- a/Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
+++ b/Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
@@ -187,7 +187,7 @@ public sealed class InnerBodyAnomalySystem : SharedInnerBodyAnomalySystem
             return;
 
         var ev = new BeforeRemoveAnomalyOnDeathEvent();
-        RaiseLocalEvent(args.Target, ev);
+        RaiseLocalEvent(args.Target, ref ev);
         if (ev.Cancelled)
             return;
 
--- a/Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
+++ b/Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
@@ -14,6 +14,7 @@ using Content.Shared.Database;
 using Content.Shared.Mobs;
 using Content.Shared.Popups;
 using Content.Shared.Whitelist;
+using Content.Shared.Zombies;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Physics.Events;
 using Robust.Shared.Prototypes;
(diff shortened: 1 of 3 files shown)


Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
```diff
@@ -186,6 +187,9 @@ private void OnMobStateChanged(Entity<InnerBodyAnomalyComponent> ent, ref MobSta
         if (args.NewMobState != MobState.Dead)
             return;
 
+        if (EntityManager.HasComponent<PendingZombieComponent>(args.Target))
```
beck-thompson, line 190: You should use `HasComp<PendingZombieComponent>(args.Target)` instead (Same thing but neater)! Also, what you should probably also do is instead send an event the target seeing if it wants to cancel, and then then the zombie system can listen for that and then cancel it. If you need some examples of how this is done let me know or ask in the discord!

beck-thompson (commented):
Great first PR!

Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
slarticodefast: Yep, an using an event is the right way to go. The way you are doing it you are hardcoding zombies into the anomaly system. If a fork does not have zombies this will break.

pcaessayrs: Okay! I will make those changes tomorrow.

pcaessayrs: Okay, @beck-thompson, do you have examples of where to look for how this is done?

Content.Shared/Anomaly/Components/InnerBodyAnomalyComponent.cs
```diff
@@ -70,3 +70,12 @@ public sealed partial class InnerBodyAnomalyComponent : Component
...
 }
+
+/// <summary>
+/// Event broadcast when an anomaly is being removed because the host is dying.
+/// Raised directed at the host entity with the anomaly.
+/// Cancel this if you want to prevent the host from losing their anomaly on death.
+/// </summary>
+public sealed class BeforeRemoveAnomalyOnDeathEvent : CancellableEntityEventArgs
+{
+}
```
slarticodefast, line 81:
```suggestion
[ByRefEvent]
public record struct BeforeRemoveAnomalyOnDeathEvent(bool Cancelled = false);
```
Use a record struct instead of a class. A lot of old code is still using `CancellableEntityEventArgs`, but this is more efficient.

Content.Server/Zombies/ZombieSystem.cs
```diff
@@ -64,10 +65,19 @@ public override void Initialize()
...
 
             SubscribeLocalEvent<ZombifyOnDeathComponent, MobStateChangedEvent>(OnDamageChanged);
+
+        }
+
+        private void OnBeforeRemoveAnomalyOnDeath(Entity<PendingZombieComponent> ent, ref BeforeRemoveAnomalyOnDeathEvent args)
+        {
+            // Pending zombies (e.g. infected non-zombies) do not remove their hosted anomaly on death.
+            // Current zombies DO remove the anomaly on death.
+            args.Cancel();
```
slarticodefast, line 80:
```suggestion
            args.Cancelled = true;
```

Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
```diff
@@ -187,8 +186,10 @@ private void OnMobStateChanged(Entity<InnerBodyAnomalyComponent> ent, ref MobSta
         if (args.NewMobState != MobState.Dead)
             return;
 
-        if (EntityManager.HasComponent<PendingZombieComponent>(args.Target))
-            return; // Keep your anomaly when becoming a zombie
+        var ev = new BeforeRemoveAnomalyOnDeathEvent();
+        RaiseLocalEvent(args.Target, ev);
```
slarticodefast, line 190:
```suggestion
        RaiseLocalEvent(args.Target, ref ev);
```

slarticodefast (changes requested):

Content.Shared/Anomaly/Components/InnerBodyAnomalyComponent.cs
beck-thompson: I was spreading misinformation sorry 😢

Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
pcaessayrs: Okay! These changes are done in https://github.com/space-wizards/space-station-14/pull/33867/commits/9c29960f2fcb77b49c112e2b4807ef99103c8149

Content.Server/Anomaly/Effects/InnerBodyAnomalySystem.cs
pcaessayrs: beck pointed at SelfBeforeGunShotEvent and other Before.*Event as examples. thanks!

slarticodefast (approved):
Tested and it works as intended.

Issue #32206, opened 2024-09-16
reinforced glass can't me made on autolathe


lzk228:
it uses ore processor's recipe

![Screenshot_20240916_102706](https://github.com/user-attachments/assets/ac63602c-a77c-4227-be6f-fbca12850e02)

Calecute:
I think this may have been introduced by https://github.com/space-wizards/space-station-14/pull/32069 .It seems the problem might be every machine uses the same recipes in Resources/Prototypes/Recipes/Lathes, if that's the case reinforced glass should only be available either in the ore processor or in the autolathe

VlaDOS1408:
it doesn't work because you can't put coal in the autolight. And the recipe is the same. Write down which solution you would like to see and I will probably do it.

PR #33868, opened 2024-12-15
Add a test to check that lathe recipes are possible

Tayrtahn:
## About the PR

Adds an integration test that makes sure that lathes are able to accept the materials needed by all of their recipes.

## Why / Balance

To avoid issues like #32206 coming up in the future.

## Technical details

Note that this is going to fail on master currently! As reported in #32206, reinforced glass is listed in the autolathe, but the recipe calls for materials that can't be added to an autolathe. Until that is fixed, this test will fail. Edit: see #33876

The test works by checking each lathe and building a hashset of the materials it accepts (found by checking every material-containing entity against the lathe's storage whitelist). The recipes are then checked to make sure that they don't call for any materials that aren't accepted.

There's also a check to make sure that recipes don't call for a material quantity greater than the capacity of the lathe. No lathes currently have a storage limit, but it's a simple enough check that it doesn't hurt to add it.

## Requirements

## Breaking changes

**Changelog**

Not player-facing, so nah.


Content.IntegrationTests/Tests/Lathe/LatheTest.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                        // Mark the lathe as accepting each material in the entity
+                        foreach (var (material, _) in compositionComponent.MaterialComposition)
+                        {
+                            acceptedMaterials.Add(material);
+                        }
+                    }
+
+                    // Check each recipe assigned to this lathe
+                    foreach (var recipeId in latheComp.StaticRecipes)
```
slarticodefast, line 77: The test should also cover the `DynamicRecipes` which are unlocked with research.

slarticodefast (changes requested):
Could you merge master so we can check if the test is successful now that the recipe is fixed?
Also the `AllLatheRecipesValidTest` should probably be moved from `ResearchTest.cs` to the new `Lathetest.cs` for better organisation.

Content.IntegrationTests/Tests/Lathe/LatheTest.cs
slarticodefast: Oh, and the additional recipes in the `EmagLatheRecipesComponent` as well.

southbridge-fur:
Frontier has been looking at limiting lathe storage, so the bit that checks against storage limits could be helpful.

slarticodefast (approved):
Looks good to me.
Very useful PR, thanks!

Issue #33861, opened 2024-12-14
Conveyors cause huge lag if there's a ton of items stuck on them


PJB3005:
@metalgearsloth 

This can be seen easily with corner conveyors in disposals etc. If the lever gets flipped the wrong way the items will never leave the conveyor, which means on large pop servers there will be like a hundred items trying to move every tick.

On a Leviathan round, this caused 15 ms (!!!) of tick time that instantly disappeared when turning off the conveyor.

![image](https://github.com/user-attachments/assets/44ede7d5-caf6-49bb-9671-41ad718dafac)

Kowlin:
<img width="555" height="237" alt="Image" src="https://github.com/user-attachments/assets/63abe1f3-b996-44ee-a8c1-7593c5fdae0b" />

PR #33870, opened 2024-12-15
Conveyor optimisations

metalgearsloth:
- Optimise movement for moving stuff. Much better lookup flags + less resolves + slapped parallelrobustjob on it.
- Avoids tracking contacts on its own.

~~Needs https://github.com/space-wizards/RobustToolbox/pull/5560 + https://github.com/space-wizards/RobustToolbox/pull/5440~~

~~Needs https://github.com/space-wizards/RobustToolbox/pull/5637~~

I want to give it a shot using shapecasts which will also be a prototype for new mob movement similar to how most games handle it (outside of physics).

Current branch still needs subscribing to airtightchanges or whatever to know if we need to start re-conveying entities.

Resolves https://github.com/space-wizards/space-station-14/issues/33861


Content.Shared/Physics/Controllers/SharedConveyorController.cs
```diff
@@ -153,22 +273,72 @@ private bool TryConvey(Entity<ConveyedComponent, PhysicsComponent, TransformComp
...
+
+            var filter = new QueryFilter
+            {
+                LayerBits = fixture.CollisionLayer,
+                MaskBits = fixture.CollisionMask,
+                Flags = QueryFlags.Static,
+            };
+
+            var transform = PhysicsSystem.GetLocalPhysicsTransform(entity.Owner, xform);
+            transform.Position += direction;
```
slarticodefast, line 299: Ideally the actual movement would be done by offsetting the velocity and letting the physics system handle the movement isntead of teleporting it in very small steps. Because currently if an entity is being conveyed it has a velocity of zero even though it is moving, which might cause problems for other systems which need that info (for example my homing projectile PR I'm working on).

Resources/Maps/bagel.yml
```diff
@@ -72,7 +72,7 @@ entities:
     - type: MetaData
```
slarticodefast, line 72: Is this file supposed to be in here?

Content.Shared/Physics/Controllers/SharedConveyorController.cs
metalgearsloth: If a body's InAir ideally it doesn't get conveyed at all (which I imagine a homing projectile would be).

The issue is mob movement entirely dictates the mob's velocity so those systems would need rewriting to handle other input sources.

slarticodefast: The problem is not the projectile (which does not get conveyed), but the velocity of a target object on a conveyor. That relative motion is important for the homing algorithm.

metalgearsloth: I doubt I can cook something up anytime soon as most resources I found on it just adjust position directly because it handles corners better + facilitates not tampering with the CC.

It's just a matter of do we think fixing conveyors tanking the server is more worthwhile than waiting for someone to rewrite conveyor movement as they would have to look at something like velocity sources (which box2d doesn't have innately).

Emisse:
what is the map change for

metalgearsloth:
> what is the map change for

Conveyor comp coz something getting conveyed, I can prolly make it not saveable.

SlamBamActionman:
I am not sure exactly *where* in the code this is caused and I suggest you investigate it further, but I found a bug with the new implementation. 

Walking onto an active conveyor belt from the side and letting go the moment you step on (easiest to do if you stand next to one and lightly tap), there's what appears to be a prediction issue where you jitter between the position you moved to and where the conveyor wants to move you. Is it possible one of the components are not being added when it should clientside, and is instead networked in?

slarticodefast:
>Otherwise it seems to work okay? I know Slart had reservations but at the same time the old conveyor code does the "teleporting" way of moving things, so it's not like we're moving to a worse implementation.

Yeah, that can be fixed in a future PR.
The misprediction should be investigated before merging though.

metalgearsloth:
See https://github.com/space-wizards/space-station-14/pull/35931

PR #33871, opened 2024-12-15
Add a 10u vial of plasma to the chemical locker

Minemoder5000:
## About the PR

Added a 10u plasma vial prototype and put that in the chemical locker.
## Why / Balance

Chemists are impatient and steal a table for the reinforced plasteel. This change makes it so they skip that step and already have the 10u of plasma at round start.
## Technical details

Two changes to two .yml files.
## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl: Minemoder
- add: The chemical locker now has a 10u vial of plasma.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/medical.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/medical.yml
@@ -117,6 +117,7 @@
       - id: BoxPillCanister
       - id: BoxBottle
       - id: BoxVial
+      - id: PlasmaChemistryVial
       - id: ChemBag
       - id: ClothingHandsGlovesLatex
       - id: ClothingHeadsetMedical
--- a/Resources/Prototypes/Entities/Objects/Specific/chemistry-vials.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/chemistry-vials.yml
@@ -149,3 +149,18 @@
         reagents:
           - ReagentId: Chlorine
             Quantity: 5
+            
+- type: entity
+  id: PlasmaChemistryVial
+  parent: BaseChemistryEmptyVial
+  suffix: plasma
+  components:
+  - type: Label
+    currentLabel: reagent-name-plasma
+  - type: SolutionContainerManager
+    solutions:
+      beaker:
+        maxVol: 10
+        reagents:
+          - ReagentId: Plasma
+            Quantity: 10


FATFSAAM2:
The little bit of co operation which chem has with crew (90% of chem players never talk with anyone ) will get taken out by this pr and well i think its a bad idea

ArtisticRoomba (approved):
Tested, works ingame.

I have no idea why the vials for the chemical synthesis kit and the agrichem kit have different max volumes compared to the base vial. This is also going along with that trend, but it's probably fine I guess.

ArtisticRoomba:
> The little bit of co operation which chem has with crew (90% of chem players never talk with anyone ) will get taken out by this pr and well i think its a bad idea

I think you make a good point, but IMO 90% of the cooperation you're referring to is oftentimes just stealing tables from other departments, or chem's own workplace.

It has been highlighted in design-discussions that it's moreso a mapping issue (medical workplaces would realistically have steal and glass tables, rather than reinforced tables) but there is a player component to it.

You're right in that some chemistry players do interact with other departments, but I've heard that the Chemistry Stealing Tables issue has been quite common and it's silly that it has to be this way.

A possible alternative idea: maybe reinforced tables should require welding to deconstruct? It *is* a plasteel table after all.

FATFSAAM2:
> > The little bit of co operation which chem has with crew (90% of chem players never talk with anyone ) will get taken out by this pr and well i think its a bad idea
> 
> I think you make a good point, but IMO 90% of the cooperation you're referring to is oftentimes just stealing tables from other departments, or chem's own workplace.
> 
> It has been highlighted in design-discussions that it's moreso a mapping issue (medical workplaces would realistically have steal and glass tables, rather than reinforced tables) but there is a player component to it.
> 
> You're right in that some chemistry players do interact with other departments, but I've heard that the Chemistry Stealing Tables issue has been quite common and it's silly that it has to be this way.
> 
> A possible alternative idea: maybe reinforced tables should require welding to deconstruct? It _is_ a plasteel table after all.

I think thats a good fix to have them use a welder i feel like artificially removing co operation from the social space game isnt a good idea chem players barely even talk let alone go outside their dep 
Realistically it should require like a 10 second do after for the tables i feel like this change is just a badly thought out and jury rigged solution to a problem which really dosent need a changelog to fix (as in security should be punishing people for stealing tables)

kosticia:
I think the problem in the table, it's too easy to deconstruct to got plasma. Without deconstruction you need to ask for plasma sci or engineers

FATFSAAM2:
I think this pr is just treating the symptom not the problem 
I think the main problem is how easy tables are to take apart because chem will be doing that the second they run out of plasma again

notquitehadouken:
i love completely liquefying sheets of reinforced steel in my Regular Ass blender

everydayeldergod:
Even if tables were easier to disassemble, plasteel would still be significantly easier to extract the plasma from than to use pure plasma for chem. At roundstart, when Chemists are making Dex, the only available plasma is somewhere inside of engi (not sure precisely where, usually near emitter storage). Consequently, all the engineers are busy with the power rush to deliver plasma. 

So, for Dex, asking Engi either a) takes a long time or b) delays power setup. It's much easier to get plasteel from tools, maints, science (I think they usually get a stack?), or a table.

Plus, just 10u of plasma would mean that Chem still needs to ask salv or engi later when they want to make advanced burns or explosives, since the plasma is consumed in those recipes.

TheShuEd:
I agree with the sentiment that this is treating the symptoms, not the problem. 

It needs to be addressed in a different way.

ArtisticRoomba:
> I agree with the sentiment that this is treating the symptoms, not the problem.
> 
> It needs to be addressed in a different way.

I'll see about making the reinforced tables require welding to be deconstructed/constructed. I think this is a good change to prevent this issue, and it makes a bit of sense.

PJB3005 (approved):
In SS13 they start with a plasma sheet they have to grind up, which I kind of prefer. But this is fine too.

beck-thompson (approved):
Tested and it works!

PR #33874, opened 2024-12-15
Block shuttle recall after high zombie percentage

lzk228:
## About the PR

recalling shuttle is unavailable after it's been called automatically cause of high zombie percentage

## Why / Balance

less roundstalling

## Technical details

loc strings moved to the component
added new method just a copy from NukeopsRuleSystem
new bool to check if shuttle has been called

## Media

couldn't reproduce, even when set shuttle call percentage to 0.1

## Requirements

## Breaking changes

no i suppose

**Changelog**

:cl:
- tweak: Automatic shuttle for high zombie percentage now cannot be recalled.

### DIFF
--- a/Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
@@ -25,4 +25,21 @@ public sealed partial class ZombieRuleComponent : Component
     /// </summary>
     [DataField]
     public float ZombieShuttleCallPercentage = 0.7f;
+
+    /// <summary>
+    /// Is the automatical shuttle called.
+    /// </summary>
+    public bool? ZombieShuttleCalled;
+
+    /// <summary>
+    /// The reason when shuttle called.
+    /// </summary>
+    [DataField]
+    public string ShuttleCallReason = "zombie-shuttle-call";
+
+    /// <summary>
+    /// The reason why shuttle can be recalled after certain zombie percentage.
+    /// </summary>
+    [DataField]
+    public string ShuttleCallUnavailableReason = "zombie-shuttle-unavailable";
 }
--- a/Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
@@ -29,7 +29,7 @@ public sealed partial class ZombieRuleComponent : Component
     /// <summary>
     /// Is the automatical shuttle called.
     /// </summary>
-    public bool? ZombieShuttleCalled;
+    public bool ZombieShuttleCalled = false;
 
     /// <summary>
     /// The reason when shuttle called.
--- a/Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
@@ -35,11 +35,11 @@ public sealed partial class ZombieRuleComponent : Component
     /// The reason why the emergency shuttle is called.
     /// </summary>
     [DataField]
-    public string ShuttleCallReason = "zombie-shuttle-call";
+    public LocId ShuttleCallReason = "zombie-shuttle-call";
 
     /// <summary>
     /// The reason why the emergency shuttle cannot be recalled after certain a zombie percentage.
     /// </summary>
     [DataField]
-    public string ShuttleCallUnavailableReason = "zombie-shuttle-unavailable";
+    public LocId ShuttleCallUnavailableReason = "zombie-shuttle-unavailable";
 }
--- a/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
@@ -143,7 +143,7 @@ public sealed class ZombieRuleSystem : GameRuleSystem<ZombieRuleComponent>
         {
             if (zombie is { ZombieShuttleCalled: false })
             {
-                if (zombie.ZombieShuttleCalled.Value)
+                if (zombie.ZombieShuttleCalled)
                 {
                     ev.Cancelled = true;
                     ev.Reason = Loc.GetString(zombie.ShuttleCallUnavailableReason);
--- a/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
@@ -134,7 +134,7 @@ public sealed class ZombieRuleSystem : GameRuleSystem<ZombieRuleComponent>
     }
 
     /// <summary>
-    ///     Doesn't let to recall the shuttle. No roundstalling
+    ///     Doesn't let the shuttle be recalled, stopping round-stalling.
     /// </summary>
     private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEvent ev)
     {
--- a/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
@@ -141,7 +141,7 @@ public sealed class ZombieRuleSystem : GameRuleSystem<ZombieRuleComponent>
         var query = QueryActiveRules();
         while (query.MoveNext(out _, out _, out var zombie, out _))
         {
-            if (zombie is { ZombieShuttleCalled: not null })
+            if (zombie is { ZombieShuttleCalled: false })
             {
                 if (zombie.ZombieShuttleCalled.Value)
                 {
--- a/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
@@ -150,7 +150,6 @@ public sealed class ZombieRuleSystem : GameRuleSystem<ZombieRuleComponent>
         }
     }
 
-
     protected override void Started(EntityUid uid, ZombieRuleComponent component, GameRuleComponent gameRule, GameRuleStartedEvent args)
     {
         base.Started(uid, component, gameRule, args);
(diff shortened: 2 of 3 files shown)


Hanzdegloker:
I kinda wish at a certain point of zombification it would arm the nuke, kinda uh, the station is a very lost cause and we don't want zombies at centcomm. It would also kinda make a reason for NT to have a nuke on station besides nukie gameplay.

lzk228:
we can have the shuttle at lower percentage, like 50
then nuke at 80-90 and immediately round and at 100

Hanzdegloker:
> we can have the shuttle at lower percentage, like 50
> then nuke at 80-90 and immediately round and at 100

Sounds great to me, 80% over 90% though so at least some survivors can enjoy the NT Cover up our mistakes experience. Also the round end at 100% is nice so it ain't just zombies moping around just waiting for the shuttle, and then round end. Makes Initial infected feel like they have an actual proper win condition and zombies feel more pronounced as a gamemode. (Its been too long since I experienced a zombies round, I love me some zeds)

lzk228:
just fyk, we already have immediately round end at 100

Hanzdegloker:
> just fyk, we already have immediately round end at 100

I have literally never seen it happen due to like, CE fartin around in space with a jetpack and or some guy hiding really really well. But hey, glad to see its already a thing lol.

Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
```diff
@@ -25,4 +25,21 @@ public sealed partial class ZombieRuleComponent : Component
     /// </summary>
     [DataField]
     public float ZombieShuttleCallPercentage = 0.7f;
+
+    /// <summary>
+    /// Is the automatic shuttle called.
+    /// </summary>
+    public bool? ZombieShuttleCalled;
```
Aeshus, line 32: It feels weird to have a nullable boolean. 

You only ever set it to true (when the shuttle is called), so there's only really two states you're using: true (called) or null (not called). Instead of using null, you should just use false.

```suggestion
    public bool ZombieShuttleCalled;
```

Resources/Locale/en-US/game-ticking/game-presets/preset-zombies.ftl
```diff
@@ -15,6 +15,7 @@ zombie-infection-underway = Your blood begins to thicken
 zombie-alone = You feel entirely alone.
 
 zombie-shuttle-call = We have detected that the undead have overtaken the station. Dispatching an emergency shuttle to collect remaining personnel.
+zombie-shuttle-unavailable = Shuttle unavailable due to high undead percentage.
```
Aeshus, line 18: Isn't this message for when someone tries to recall?

```suggestion
zombie-shuttle-unavailable = Shuttle recall is unavailable due to high undead percentage.
```

Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
```diff
@@ -25,4 +25,21 @@ public sealed partial class ZombieRuleComponent : Component
...
+
+    /// <summary>
+    /// The reason when shuttle called.
+    /// </summary>
+    [DataField]
+    public string ShuttleCallReason = "zombie-shuttle-call";
+
+    /// <summary>
+    /// The reason why shuttle can be recalled after certain zombie percentage.
```
Aeshus, line 41:
```suggestion
    /// The reason why the emergency shuttle cannot be recalled after certain zombie percentage.
```

Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
```diff
@@ -25,4 +25,21 @@ public sealed partial class ZombieRuleComponent : Component
...
     [DataField]
     public float ZombieShuttleCallPercentage = 0.7f;
+
+    /// <summary>
+    /// Is the automatic shuttle called.
+    /// </summary>
+    public bool? ZombieShuttleCalled;
+
+    /// <summary>
+    /// The reason when shuttle called.
```
Aeshus, line 35:
```suggestion
    /// The reason why the emergency shuttle is called.
```

Aeshus (changes requested):

Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
lzk228: i made it nullable just to use in the next code
image

i suppose with it help, it chooses only components that has the indicated bool
and i'm not sure if it's gonna work if i set it to false and make check for false

Resources/Locale/en-US/game-ticking/game-presets/preset-zombies.ftl
lzk228: i suppose it might be too long but we'll see

Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
Aeshus: Is ZombieShuttleCalled ever false though? 

If it’s not null, it is always true (shuttle is called), otherwise it’s null (shuttle not called)

You’ve basically just turned nullability into a Boolean, when you could’ve just used a regular Boolean.

Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
```diff
@@ -130,6 +133,26 @@ private void CheckRoundEnd(ZombieRuleComponent zombieRuleComponent)
...
+    /// <summary>
+    ///     Doesn't let to recall the shuttle. No roundstalling
+    /// </summary>
+    private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEvent ev)
+    {
+        var query = QueryActiveRules();
+        while (query.MoveNext(out _, out _, out var zombie, out _))
+        {
+            if (zombie is { ZombieShuttleCalled: false })
+            {
```
Aeshus, line 145: Is this part necessary? Currently it's a bug as it's checking if it's false and then true (which is impossible).

lzk228: well it was basically copied from
https://github.com/space-wizards/space-station-14/blob/5958801182b35bb1e78095f18b180d3069ad7394/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs#L305-L323

as i understand i get ZombieGameRule entity only this way

Aeshus: Nukeops does is because it is trying to pattern match if the field is not null, but we don't have to do this because we're using a boolean and not a nullable value.

As we're just working with a boolean, we can access it directly:

```cs
    private void OnShuttleCallAttempt(ref CommunicationConsoleCallShuttleAttemptEvent ev)
    {
        var query = QueryActiveRules();
        while (query.MoveNext(out _, out _, out var zombie, out _))
        {
            if (!zombie.ZombieShuttleCalled)
                continue;

            ev.Cancelled = true;
            ev.Reason = Loc.GetString(zombie.ShuttleCallUnavailableReason);
            return;
        }
    }

```

The current pattern matching is incorrect as it first checks that it is false then it checks that it is true, which is impossible outside of something like a race condition (true != false).

Some examples in other parts of the codebase:
https://github.com/space-wizards/space-station-14/blob/2a6800f35411cf35213f2ef1dfecf2e5c4aa9ed5/Content.Server/Antag/AntagSelectionSystem.cs#L85-L106
https://github.com/space-wizards/space-station-14/blob/2a6800f35411cf35213f2ef1dfecf2e5c4aa9ed5/Content.Server/GameTicking/Rules/RuleGridsSystem.cs#L28-L39

lzk228: will know in the future, thanks!

Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
```diff
@@ -25,4 +25,21 @@ public sealed partial class ZombieRuleComponent : Component
...
+    /// <summary>
+    /// Is the automatic shuttle called.
+    /// </summary>
+    public bool ZombieShuttleCalled = false;
+
+    /// <summary>
+    /// The reason why the emergency shuttle is called.
+    /// </summary>
+    [DataField]
+    public string ShuttleCallReason = "zombie-shuttle-call";
```
Aeshus, line 38: You should be using `LocId` for the type as it'll validate the localization string exists.

```suggestion
    public LocId ShuttleCallReason = "zombie-shuttle-call";
```

Content.Server/GameTicking/Rules/Components/ZombieRuleComponent.cs
```diff
@@ -25,4 +25,21 @@ public sealed partial class ZombieRuleComponent : Component
...
+
+    /// <summary>
+    /// The reason why the emergency shuttle cannot be recalled after certain a zombie percentage.
+    /// </summary>
+    [DataField]
+    public string ShuttleCallUnavailableReason = "zombie-shuttle-unavailable";
```
Aeshus, line 44:
```suggestion
    public LocId ShuttleCallUnavailableReason = "zombie-shuttle-unavailable";
```

Aeshus (changes requested):
Thanks for making the requested changes!

Only thing I noticed is that the localization string fields in the component should be typed as `LocId` rather than `string`.

LocId provides extra validation to ensure that the string exists, so you don't have to worry about weird typos causing the string to not be recognized.

The rest of the code looks perfect, and should be mergable after this last change.

Sorry for not catching this/re-reviewing this earlier!

Content.Server/GameTicking/Rules/ZombieRuleSystem.cs
```diff
@@ -130,6 +133,23 @@ private void CheckRoundEnd(ZombieRuleComponent zombieRuleComponent)
             _roundEnd.EndRound();
     }
 
+    /// <summary>
+    ///     Doesn't let to recall the shuttle. No roundstalling
```
Aeshus, line 137: Just a quick doc comment change

```suggestion
    ///     Doesn't let the shuttle be recalled, stopping round-stalling.
```

Aeshus (approved):

MilonPL:
Thank you for your contribution! I will be closing this PR as stale on account of being a draft and not having any activity for over 6 weeks. If you end up coming back to this, please open a new PR.

Issue #33824, opened 2024-12-11
The birthday cake recipe doesnt require candles


IProduceWidgets:
## Description

Sprite has candles. Recipe no has candles. What gives?
https://github.com/space-wizards/space-station-14/blob/976cae9c785fd8bb4ba1c7476f9a71694250288f/Resources/Prototypes/Recipes/Cooking/meal_recipes.yml#L1468
**Reproduction**

**Screenshots**

![image](https://github.com/user-attachments/assets/fa764063-ebd1-4294-b173-ab55352f8594)

**Additional context**

slarticodefast:
Literally unplayable.
Someone fix this.

TeenSarlacc:
unrelated to this but a crafting recipe for candles would be cool too

PR #33875, opened 2024-12-15
Added small candles to the birthday cake microwave recipe.

DrSpessman:
## About the PR
I changed the microwave recipe for birthday cake so that it now requires three small candles.

## Why / Balance
The birthday cake sprite pictures three lit candles but the recipe did not require candles. The recipe now matches the sprite as three small candles are required to complete the recipe.

## Technical details
I added CandleSmall: 3 to the solids section of the recipeBirthdayCake microwaveMealRecipe in meal_recipes.yml.

## Media
New recipe in action:
MicrowaveRecipe
It even lit the candles!
ItsACake

## Requirements

## Breaking changes
N/A

**Changelog**

:cl:
- tweak: Birthday cake recipe now requires 3 small candles.

### DIFF
--- a/Resources/Prototypes/Recipes/Cooking/meal_recipes.yml
+++ b/Resources/Prototypes/Recipes/Cooking/meal_recipes.yml
@@ -1474,6 +1474,7 @@
     Cream: 5
   solids:
     FoodCakePlain: 1
+    CandleSmall: 3
 
 - type: microwaveMealRecipe
   id: RecipeChocolateCake


DrSpessman:
This PR resolves #33824

slarticodefast (approved):
Looks good to me. Thank you for your first contribution!
Small tip: If you write `resolves #33824` or `fixes #33824` into the PR description itself, then GitHub will automatically link and close that issue when the PR is merged.

SlamBamActionman:
One issue I've found with this PR is that you're required to use the default plain small candles for this; ideally you'd want to be able to use a mix of colored candles to not restrict the recipe. 

That being said, I don't think we have the functionality to support that with the current cooking system. So I'll check with the other maintainers if we can get this through anyways.

IamVelcroboy:
> That being said, I don't think we have the functionality to support that with the current cooking system. So I'll check with the other maintainers if we can get this through anyways.

You should be able to use a "candle" tag, no?

SlamBamActionman:
> > That being said, I don't think we have the functionality to support that with the current cooking system. So I'll check with the other maintainers if we can get this through anyways.
> 
> You should be able to use a "candle" tag, no?

From looking at the prototype we don't seem to be supporting tags for microwave recipes, unfortunately.

IamVelcroboy:
> > You should be able to use a "candle" tag, no?
> 
> From looking at the prototype we don't seem to be supporting tags for microwave recipes, unfortunately.

I see, nm then 🙃

DrSpessman:
> One issue I've found with this PR is that you're required to use the default plain small candles for this; ideally you'd want to be able to use a mix of colored candles to not restrict the recipe.
> 
> That being said, I don't think we have the functionality to support that with the current cooking system. So I'll check with the other maintainers if we can get this through anyways.

Is there no "or" for recipe ingredients?

SlamBamActionman:
Hello! After discussing with the maintainer team we've decided to close this PR for the following reasons:

Conceptually we like needing candles for the cake, but the issue is the implementation. Since recipes specify prototype IDs it means you can't use differently colored candles, which we think could be frustrating and unintuitive for someone who checks the recipe and sees it needing "three small candles", not realizing it's referring specifically to the white ones. 

The way forward for something like this would be to either
a) implement cakes as a FoodSequence recipe. This is the system that tacos and burgers use. This could allow people to make custom cakes and recipes (imagine a cake with 10 candles!). It's the ideal option since cakes can be so multifaceted, but it would also be a fairly extensive PR. 
b) implement tag-based ingredients for recipes. Right now you can only include prototype IDs (the "solids" property) or reagents. Having a new property that take tags as input would be good, where you assign any candle type the Candle tag and then check for that. 

Hopefully one of these options might be worth checking out! Thank you for contribution!

PR #33878, opened 2024-12-15
Fix followers getting sent to nullspace when target is polymorphed

### LINKED ISSUE #28778: Ghost follow is broken by polymorph
## Description
Following Someone as a ghost who polymorphism sends you to nullspace 

**Reproduction**
Follow someone as ghost

they either
- use weh juice to go lizard mode
- Cyro pod
- Cham projector
- Geras as alive

Welcome to nullspace

Tayrtahn:
## About the PR

Followers (ghosts, anomaly cores) orbiting targets will now continue orbiting the polymorph's result rather than getting tucked away in nullspace.

## Why / Balance

Fixes #28778

## Technical details

PolymorphSystem now raises a `PolymorphedEvent` on the targeted entity immediately after making the swap. FollowerSystem subscribes to this event and reacts by setting the new entity as follow target.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Ghosts are no longer sent to the void if the player they are following is polymorphed.


slarticodefast:
I found two bugs while testing:
- The camera rotation is weird when following the polymorphed entity.
grafik

- When unpolymorphing the following ghost gets parented to the target, but does have the rotating animation.
grafik

slarticodefast:
Is it fixed by the engine change now?
https://github.com/space-wizards/RobustToolbox/pull/5595

Tayrtahn:
> Is it fixed by the engine change now? [space-wizards/RobustToolbox#5595](https://github.com/space-wizards/RobustToolbox/pull/5595)

Yep, the first bug you noted is fixed by that engine change. The second is not, but I know what's causing it and will work out a fix.

Tayrtahn:
And there's the fix.

slarticodefast:
Thanks for investingating! The first bug is fixed, but the second one still happens for me:
example

the exact reproduction is:
- Client 1 injects a lot of weh juice using a hypospray until they polymorph
- Client 2 only then starts following them as a ghost
- Client 1 unpolymorphs
- The following animation stops, but player 2 is still bound to the location of player 1

Might be something with the animation. If you prefer we can merge this PR as it is though and open an issue, since this is still a big improvement over the previous behaviour.

Tayrtahn:
Ah, I hadn't tested with multiple clients; I was just following a dummy character. Good catch! I'll try to work out a better fix for that.

slarticodefast (approved):
Works as intended. Thanks for further investigation into the bug!

PR #33879, opened 2024-12-16
Only disable panicbunker for admins with AdminFlags.Admin

### LINKED ISSUE #25260: Panic bunker doesn't check if someone's an admin
Title.

slarticodefast:
## About the PR
Some maintainers have debug permissions, but currently the auto disable function for the panic bunker does not check if someone is a proper admin, so it gets disabled when a maintainer joins.
Fixes #25260

## Technical details
`HasFlag` did not consider the flags if the admin was de-adminned, so I added the `includeDeAdmin` bool similar to the one other functions have.

I was only able to do limited testing since I was not able to figure out how to start a second client without Host permissions as a localhost. But when using breakpoints it seemed to work as intended.

## Requirements

**Changelog**
:cl:
ADMIN:
- tweak: The panic bunker now only automatically disables when an admin with the 'Admin' permission is online instead of any permission.

### DIFF
--- a/Content.Server/Administration/Systems/AdminSystem.cs
+++ b/Content.Server/Administration/Systems/AdminSystem.cs
@@ -339,11 +339,13 @@ public sealed class AdminSystem : EntitySystem
     {
         var hasAdmins = false;
         foreach (var admin in _adminManager.AllAdmins)
+        {
             if (_adminManager.HasAdminFlag(admin, AdminFlags.Admin, includeDeAdmin: PanicBunker.CountDeadminnedAdmins))
             {
                 hasAdmins = true;
                 break;
             }
+        }
 
         // TODO Fix order dependent Cvars
         // Please for the sake of my sanity don't make cvars & order dependent.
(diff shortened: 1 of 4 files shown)


Content.Server/Administration/Systems/AdminSystem.cs
```diff
@@ -337,10 +337,15 @@ private void OnBabyJailMaxOverallMinutesChanged(int minutes)
 
     private void UpdatePanicBunker()
     {
-        var admins = PanicBunker.CountDeadminnedAdmins
-            ? _adminManager.AllAdmins
-            : _adminManager.ActiveAdmins;
-        var hasAdmins = admins.Any();
```
deltanedas, line 343: why not use Any(session => ...)

also this should probably use ban flag instead

deltanedas (changes requested):

Content.Server/Administration/Systems/AdminSystem.cs
slarticodefast: The headmin said to use the Admin flag, not the Ban flag

SlamBamActionman (approved):
yes pls i get jumpscared by the popup every single time i join

PR #33622, opened 2024-11-28
Light verb is now predicted

beck-thompson:
## About the PR

Moved some stuff to shared so the verb is correctly predicted! Same pr as https://github.com/RMC-14/RMC-14/pull/4932

## Why / Balance

This was killing me in game... The lag was making me constantly turn my light on or off when I didn't mean to.

## Technical details

Moved the verb function to shared, made the on and off functions overridable so they can be "predicted" on the client side (Its fake but they need the definitions).

## Media

https://github.com/user-attachments/assets/94fd4900-b73d-484b-86ec-ab23bb4fe44d

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl: Beck Thompson
- fix: Handheld light verbs are now predicted!

PR #32496, opened 2024-09-28
Predict appraisal tool verb!

beck-thompson:
## About the PR

Title! It was annoying to me.

## Technical details

This one is a little weird. Because the systems for determining price and stuff are in server, I had to do some funky override abstract methods! Basically:

Shared is abstract and has all the interaction and verb stuff and also has an abstract definition of `GetPriceOrBounty`.
Client has a simplified version of `GetPriceOrBounty` that only just returns the correct true or false value (But no popups).
Server has the actual implementation of `GetPriceOrBounty`.

This means that the client wont misspredict while also allowing for the verb and interaction events to be predicted, even if the popup isn't.

## Breaking changes

`PriceGunComponent` is now in shared.

**Changelog**

:cl: Beck Thompson
- fix: The appraisal tool verb is now predicted!

PR #32420, opened 2024-09-23
Predict inflatable barriers verb

beck-thompson:
## About the PR

Moved the verb system and stuff into shared so you wouldn't get the annoying lag. Also cleaned up the old version of the do-after system and made it nice.

## Technical details

See above! Nothing too crazy.

## Media

https://github.com/user-attachments/assets/a11079cd-a67f-4a6e-948a-d874b599abcd

## Breaking changes

If you have a `DisassembleOnAltVerb` the fields have been renamed:
`prototype` -> `prototypeToSpawn`
`doAfter` -> `disassembleTime`

If you just change the names everything else works exactly the same!

**Changelog**

:cl: Beck Thompson
- fix: Inflatable barrier disassembly is not not laggy anymore.

PR #33883, opened 2024-12-16
Land mine armament

kaiserbirch:
## About the PR

The landmines are now armable and won't trigger unless armed.

## Why / Balance

By making mines armable they are more safely handled and the amount of accidental explosions will decrease.

## Technical details

Modified the component and system of landmine, added verb to ftl.

## Media

https://github.com/user-attachments/assets/338f5fb8-3836-4d1e-b26d-42f6d567524d

## Requirements

## Breaking changes

**Changelog**

 - tweak: Mines have to be armed before they can be triggered.


lzk228:
i think you should add an already armed version for mapping

Content.Server/LandMines/LandMineSystem.cs
```diff
@@ -14,14 +15,38 @@ public sealed class LandMineSystem : EntitySystem
...
         SubscribeLocalEvent<LandMineComponent, StepTriggeredOnEvent>(HandleStepOnTriggered);
         SubscribeLocalEvent<LandMineComponent, StepTriggeredOffEvent>(HandleStepOffTriggered);
 
         SubscribeLocalEvent<LandMineComponent, StepTriggerAttemptEvent>(HandleStepTriggerAttempt);
     }
 
+    /// <summary>
+    /// Adds a verb to prime the landmine to activate if somebody steps on it.
+    /// </summary>
+    private void OnGetAltVerbs(EntityUid uid, LandMineComponent component, GetVerbsEvent<AlternativeVerb> args)
```
beck-thompson, line 28: If at all possible try to put this in shared so it isn't laggy! You might need to do some funky function overloading but it shouldn't be too hard. If you want help / need more examples let me know or ask in the discord!

Some examples:
#32420
#33622
#32496

Content.Server/LandMines/LandMineComponent.cs
```diff
@@ -10,4 +10,11 @@ public sealed partial class LandMineComponent : Component
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
     public SoundSpecifier? Sound;
+
+    /// <summary>
+    /// Is the land mine armed and dangerous?
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public bool Armed;
```
beck-thompson, line 18: Probably best to set this to false explicitly as the default!

beck-thompson (commented):
Really good first PR! And yeah, it would probably be a good idea to make an already armed version

Content.Server/LandMines/LandMineComponent.cs
Aeshus: Well it already defaults to false, so it would just be an explicit thing so other programmers recognize it.

beck-thompson: Yeah exactly! When I first saw it I was a little confused

Plykiya:
the light shouldn't be blinking while it's not armed

kaiserbirch:
I am not to familiar with mapping but wouldn't it be solved by just giving the entity `armed: true`? Or do you mean a placeable already armed mine?

I could also just make it armed as default so that the behavior is currently unchanged.

Content.Server/LandMines/LandMineSystem.cs
kaiserbirch: I am struggling a bit with understanding exactly the nuances of the implementation of this. I've updated the relevant code with what I *believe* is the right pattern.

Content.Server/LandMines/LandMineComponent.cs
kaiserbirch: Fixed! Added a false.

Content.Shared/LandMines/LandMineComponent.cs
```diff
@@ -10,4 +10,11 @@ public sealed partial class LandMineComponent : Component
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
     public SoundSpecifier? Sound;
+
+    /// <summary>
+    /// Is the land mine armed and dangerous?
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
```
slarticodefast, line 17:
```suggestion
    [DataField]
```
`DataField` includes VV readwrite nowadays, so this is no longer needed.
The datafield and component should be predicted and networked. For that add the `NetworkedComponent`, `AutoGenerateComponentState` and `AutoNetworkedField` attributes.

Content.Server/LandMines/LandMineSystem.cs
```diff
@@ -1,27 +1,42 @@
...
 
+    public override void Arm(EntityUid uid, LandMineComponent component)
+    {
+        component.Armed = true;
+    }
+
+    public override bool IsArmed(LandMineComponent component)
+    {
+        return component.Armed;
+    }
```
slarticodefast, line 32: These should be in the shared system.

kaiserbirch:
> i think you should add an already armed version for mapping

Is this solved by adding an armed version through YAML?

Content.Server/LandMines/LandMineSystem.cs
```diff
@@ -33,11 +48,15 @@ private void HandleStepOnTriggered(EntityUid uid, LandMineComponent component, r
...
+            return;
+
         _trigger.Trigger(uid, args.Tripper);
     }
 
     private static void HandleStepTriggerAttempt(EntityUid uid, LandMineComponent component, ref StepTriggerAttemptEvent args)
     {
         args.Continue = true;
     }
+
```
slarticodefast, line 61:
```suggestion
```

slarticodefast (changes requested):
A few more changes are needed:
- Inherit `SharedLandmineSystem` on the client as well so the verb gets properly predicted. Currently if you right click on a mine you see it pop up after a short moment. Prediction means it will be there immediately.
- The red blinking light on the land mine should only show when armed for visual feedback. Have a look at the `GenericVisualizer` for that.
- Some maps have landmines mapped, for example in front of the bridge. You will have to edit the prototypes to make sure they still start armed.

Content.Server/LandMines/LandMineSystem.cs
kaiserbirch: Fixed

Content.Shared/LandMines/LandMineComponent.cs
kaiserbirch: Fixed

Content.Server/LandMines/LandMineSystem.cs
kaiserbirch: Fixed

lzk228 (changes requested):
leave LandMineExplosive unchanged
and then just add

```yaml
- type: entity
  parent: LandMineExplosive
  id: LandMineExplosiveArmed
  suffix: armed
  components:
  - type: LandMine
    armed: true
```

Resources/Prototypes/Entities/Objects/Misc/land_mine.yml
```diff
@@ -68,9 +82,25 @@
   id: LandMineExplosive
   components:
   - type: ExplodeOnTrigger
+  - type: LandMine
```
lzk228, line 85: you don't need to add it again, because the parent has it

Resources/Prototypes/Entities/Objects/Misc/land_mine.yml
```diff
@@ -68,9 +82,25 @@
...
   - type: Explosive
     explosionType: Default
     maxIntensity: 10
     intensitySlope: 3
     totalIntensity: 120 # about a ~4 tile radius
     canCreateVacuum: false
+
+- type: entity
+  name: explosive mine
```
lzk228, line 94: remove name line, also taken from the parent

kaiserbirch:
I believe the merge conflict is in error due to the error message in the log:
```'Something went wrong while executing your query. This may be the result of a timeout, or it could be a GitHub bug. Please include `8470:1F939B:552F1A0:A877830:676131A5` when reporting this issue.'```

File: Resources/Prototypes/Entities/Objects/Misc/land_mine.yml (hunk 6 above)
Comment (kaiserbirch) [reply Resources/Prototypes/Entities/Objects/Misc/land_mine.yml]: Fixed

File: Resources/Prototypes/Entities/Objects/Misc/land_mine.yml (hunk 7 above)
Comment (kaiserbirch) [reply Resources/Prototypes/Entities/Objects/Misc/land_mine.yml]: Fixed

File: Resources/Prototypes/Entities/Objects/Misc/land_mine.yml [hunk 8]
```diff
@@ -74,3 +88,17 @@
     intensitySlope: 3
     totalIntensity: 120 # about a ~4 tile radius
     canCreateVacuum: false
+
+- type: entity
+  suffix: armed
+  parent: LandMineExplosive
+  id: LandMineExplosiveArmed
```
slarticodefast, line 95: `LandMineExplosive` should be the armed version because it is used in some maps already.
Same with `LandMineKick`, and `LandMineModular`.
Make a new unarmed prototype for each.

slarticodefast (dismissed):
Another thing I found while testing is that the janitor's explosive wet floor sign is now no longer a stealth item. To solve this I would add a bool to the component to disable showing the `Arm` verb and the examination text and make it spawn armed.
grafik

Resources/Prototypes/Entities/Objects/Misc/land_mine.yml
lzk228: yeah it probably would be better than i said

Resources/Prototypes/Entities/Objects/Misc/land_mine.yml
```diff
@@ -22,7 +22,21 @@
...
     sprite: Objects/Misc/landmine.rsi
-    state: landmine
+    state: landmine-inactive
+    layers:
+      - state: landmine-inactive
+        map: [ "disarmed" ]
+        visible: true
+      - state: landmine
+        map: [ "armed" ]
+        visible: false
```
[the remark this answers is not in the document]
kaiserbirch: Uh, sorry. I don't know how to do this. I have tried understanding from the [documentation](https://docs.spacestation14.com/en/ss14-by-example/making-a-sprite-dynamic.html) but failing. The visualizer system is confusing for me. Could you please explain?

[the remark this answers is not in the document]
slarticodefast: Here is an example from the ninja gloves:
grafik
Instead of making the first layer invisible and the second layer visible when armin, just change the state of the layer directly in the generic visualizer. That way you only need one layer instead of two.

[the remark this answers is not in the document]
kaiserbirch: So I kinda understand this YAML but still struggling to make it actually change. The mine shows now with the `landmine-inactive`-sprite but arming it does not change it to `landmine`. And the already armed ones started showing with the `landmine-inactive`-sprite.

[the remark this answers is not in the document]
slarticodefast: The layer settings in the `SpriteComponent` (`- type: Sprite`) determine which sprite is initially spawned and the generic visualizer will update these layers when the Visuals are changed in C#.
For the armed ones you will have to overwrite the SpriteComponent to use the `landmine-inactive` state in the inherited prototype.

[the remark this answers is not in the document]
kaiserbirch: Alright so now I have at least the starting state correct. How do I change the visuals in C#?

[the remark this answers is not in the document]
kaiserbirch: Ah, I found the problem. I was using `ToggleVisuals.Layers` instead of `ToggleVisuals.Toggled`.

Content.Shared/LandMines/SharedLandMineSystem.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    private void OnGetAltVerbs(EntityUid uid, LandMineComponent component, GetVerbsEvent<AlternativeVerb> args)
+    {
+        if (!args.CanInteract || !args.CanAccess || args.Hands == null)
+            return;
+
+        args.Verbs.Add(new AlternativeVerb
+        {
+            Text = Loc.GetString("land-mine-verb-begin"),
+            Disabled = component is {Armed:true},
```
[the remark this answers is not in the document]
kaiserbirch: Fixed

Content.Shared/LandMines/SharedLandMineSystem.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    private void OnExamine(EntityUid uid, LandMineComponent comp, ExaminedEvent args)
+    {
+        if (!args.IsInDetailsRange)
+            return;
+        using (args.PushGroup(nameof(LandMineComponent)))
+        {
+            if(comp.Armed)
+                args.PushMarkup(Loc.GetString("land-mine-armed", ("name", uid)));
+        }
```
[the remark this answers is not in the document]
kaiserbirch: Fixed

Resources/Prototypes/Entities/Objects/Misc/land_mine.yml
kaiserbirch: Fixed

kaiserbirch:
What would the thoughts be on doing armable as a component instead?

slarticodefast:
Good point. It could probably reuse `ToggleVerbComponent` for arming.

kaiserbirch:
Would a disarmament mechanic warrant its own PR or should I expand the scope of this one?

kaiserbirch:
@slarticodefast Are there any more changes that you request?

Content.Shared/Armable/ArmableSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    private void OnExamine(EntityUid uid, ArmableComponent comp, ExaminedEvent args)
+    {
+        if (!args.IsInDetailsRange || !comp.ShowStatusOnExamination || !TryComp<ItemToggleComponent>(uid, out var itemToggle))
+            return;
+
+        if(itemToggle.Activated)
+            args.PushMarkup(Loc.GetString("examine-armed", ("name", uid)));
+        else
+            args.PushMarkup(Loc.GetString("examine-not-armed", ("name", uid)));
```
Aeshus, line 33: Could you move these LocIds into `ArmableComponent`?

Example:

https://github.com/space-wizards/space-station-14/blob/2ff82380301f8ea61f021ecebc8d5dc6daebdeb6/Content.Shared/Wires/WiresPanelComponent.cs#L45-L57

Content.Shared/Armable/ArmableSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<ArmableComponent, ExaminedEvent>(OnExamine);
+        SubscribeLocalEvent<ArmableComponent, ItemToggleActivateAttemptEvent>(TryArming);
+        SubscribeLocalEvent<ArmableComponent, ItemToggledEvent>(ArmingDone);
+    }
+
+    private void TryArming(Entity<ArmableComponent> entity, ref ItemToggleActivateAttemptEvent args)
```
Aeshus, line 19: I was under the impression that `ItemToggleActivateAttemptEvent` is meant for canceling the event if a condition is not made (e.g. the landmine is already armed, so the event should be canceled), not to actually toggle any functionality.

You would put the toggle functionality in the `ItemToggledEvent` instead. 

This would also let you skip the weird `comp.OnActivate = false;` thing that you have in `ArmingDone`, which doesn't have similar functionality anywhere else in the code-base.

Content.Server/LandMines/LandMineSystem.cs
```diff
@@ -14,6 +15,7 @@ public sealed class LandMineSystem : EntitySystem
 
     public override void Initialize()
     {
+        base.Initialize();
```
Aeshus, line 18: We like padding here.

```suggestion
        base.Initialize();
        
```

Content.Shared/LandMines/LandMineComponent.cs
```diff
@@ -1,13 +1,15 @@
 using Robust.Shared.Audio;
+using Robust.Shared.GameStates;
+using Robust.Shared.Serialization;
```
Aeshus, line 3: Are these imports used anywhere?

```suggestion
```

Content.Shared/Armable/ArmableSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+        if (TryComp<ItemToggleComponent>(entity, out var comp))
+            comp.Activated = true;
+    }
+
+    private void OnExamine(EntityUid uid, ArmableComponent comp, ExaminedEvent args)
+    {
+        if (!args.IsInDetailsRange || !comp.ShowStatusOnExamination || !TryComp<ItemToggleComponent>(uid, out var itemToggle))
+            return;
+
+        if(itemToggle.Activated)
```
Aeshus, line 30: You need a space here.

```suggestion
        if (itemToggle.Activated)
```

Content.Shared/Armable/ArmableSystem.cs
```diff
@@ -0,0 +1,50 @@
+﻿using Content.Shared.Examine;
+using Content.Shared.Toggleable;
+using Content.Shared.Item.ItemToggle.Components;
```
Aeshus, line 3: You need to sort your imports

```suggestion
using Content.Shared.Item.ItemToggle.Components;
using Content.Shared.Toggleable;
```

Content.Shared/Armable/ArmableSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Content.Shared.Armable;
+
+public sealed class ArmableSystem : EntitySystem
+{
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
```
Aeshus, line 13:
```suggestion
        base.Initialize();
        
```

Content.Shared/LandMines/LandMineComponent.cs
```diff
@@ -1,13 +1,15 @@
...
 
-namespace Content.Server.LandMines;
+namespace Content.Shared.LandMines;
 
 [RegisterComponent]
 public sealed partial class LandMineComponent : Component
```
Aeshus, line 8: Could we get some documentation here?

Content.Shared/Armable/ArmableSystem.cs
```diff
@@ -0,0 +1,50 @@
+﻿using Content.Shared.Examine;
+using Content.Shared.Toggleable;
+using Content.Shared.Item.ItemToggle.Components;
+
+namespace Content.Shared.Armable;
+
+public sealed class ArmableSystem : EntitySystem
```
Aeshus, line 7: Documentation here too.

Content.Shared/Armable/ArmableComponent.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Robust.Shared.GameStates;
+
+namespace Content.Shared.Armable;
```
Aeshus, line 3: You need a newline here

```suggestion
namespace Content.Shared.Armable;

```

Aeshus (changes requested):
Thanks for the PR!

I have some thoughts about the way you handle the armable activation, as you should move some behavior between your `TryArming` and `ArmingDone` functions to be closer to the intended purpose of those events.

From my reading of the undocumented event code, `ItemToggleActivateAttemptEvent` is for checking/canceling the toggle if some conditions are not met, and `ItemToggledEvent` is for actually activating the functionality given it was not canceled.

The rest are all just minor stylistic nitpicks.

Thanks again.

Content.Shared/Armable/ArmableSystem.cs
kaiserbirch: I have moved the statements to their correct method but the purpose of `comp.OnActivate = false;` is to remove the deactivate verb. I am not sure how to otherwise do that.

Aeshus: Ah, then it should be fine to keep it there.

kaiserbirch: Is it actually possible to try to activate it twice? Should I keep a check if it is already activated in the `TryArming`-method? If not, then the method would be unnecessary and could be cut.

Content.Shared/Armable/ArmableSystem.cs
kaiserbirch: Fixed

Content.Shared/LandMines/LandMineComponent.cs
kaiserbirch: I am admittedly not the best at documentation. Please have a look if the documentation I have added is relevant and helpful.

Content.Shared/Armable/ArmableSystem.cs
kaiserbirch: Documentation feedback requested here too.

Content.Server/LandMines/LandMineSystem.cs
```diff
@@ -14,30 +16,46 @@ public sealed class LandMineSystem : EntitySystem
...
 
-    private static void HandleStepTriggerAttempt(EntityUid uid, LandMineComponent component, ref StepTriggerAttemptEvent args)
+    /// <summary>
+    /// If both ArmableComponent and ItemToggleComponent is present it will not
+    /// continue unless ItemToggleComponent.Activated is true.
+    /// </summary>
+    private void HandleStepTriggerAttempt(EntityUid uid, LandMineComponent component, ref StepTriggerAttemptEvent args)
     {
-        args.Continue = true;
+        if (TryComp<ArmableComponent>(uid, out _) && TryComp<ItemToggleComponent>(uid, out var itemToggle) && !itemToggle.Activated)
```
beck-thompson, line 56: Use `HasComp` here!

beck-thompson (changes requested):
Looks good, just one minor style thing!

beck-thompson (approved):
Looks good! Great first pr :saluting_face:

Issue #30264, opened 2024-07-23
Admins need to be able to interact with health more intuitively


TheDoctor1977:
From discussion on #30114:
> > I kinda think that admins should be able to examine (and edit?) someone's health from the context menu or something. Having to use an item seems clunky.
>
>  And edit the crit / dead / whatever state thresholds as well as max health please.

This could most likely be placed under the admin or debug sections of the context menu, as a panel showing something like:
1. Current health
2. Any chems currently metabolizing (you might actually just able to see this in solution editing? haven't used that much)
3. Buttons to quickly add or remove damage of certain types (rejuvenate could possibly be moved into here)
4. Text boxes or dropdowns to adjust the thresholds for the various health states

bruhmogus:
we can use the damage command as a stopgap for now, it even has autofill

PR #33884, opened 2024-12-16
Fix admins not being able to health scan slimes

### LINKED ISSUE #28300: cant use items on slime people AS AGHOST
## Description
As Aghost you can no longer use the health analyzer in the PDA to check their health or use any item on them as it puts it in their internal storage instead, strangely enough it works as intended as a normal player

**Reproduction**
1 Use any item on hurt slime
2 Be sad as the item in now in them

### LINKED ISSUE #29412: Admin PDA inserts into slimes when health analyzing
## Description
 

when in aghost health analyzing pdas will insert into a slimes internal storage rather than scan them

**Reproduction**
 
this is NOT an issue with pdas while adminned as a urist. This problem occurs exclusively when scanning a slime as a ghost. It's possible that being adminned is irrelevant and the issue is the ghost part, but the only ghosts with hands are admins so *shrugs*

**Additional context**
 
I talked briefly with Shadowcommander about this in the discord and he mentioned it was possibly due to the bypassing of certain "checks" when ghosted or adminned (or both?)

### LINKED ISSUE #30114: Health Analyzers as an Admin Ghost.
## Description

Scanning people with Health Analyzers as an Admin Ghost gives them a notification, when it shouldn't. It also ends up putting our PDA's/Analyzers in slime body storage. This makes it difficult to check people's health quickly during issues without giving things away or being annoying to handle.
Not much else to add here, It's pretty self-explanatory.

slarticodefast:
## About the PR
Fixes a major admin annoyance when trying health scan someone.
Fixes #30114 (the health scanner popup was already made silent for the admin pda)
Fixes #29412
Fixes #28300

## Why / Balance
\>be admin
\>try to use admin pda health analyzer on a slime
\>panic as it inserts into their storage instead

## Technical details
I changed the logic behind the `ClickInsert` storage setting to be no longer affected by the admin's `BypassInteractionChecksComponent` as this should only change *if* you can interact with something, not *how* you interact with it. Additionally I split up the `ClickInsert` datafield into a new `OpenOnActivate` datafield as the former was reused for opening the storage by pressing E for no reason.
Since slimes are the only entity that use `ClickInsert = false` this should have no further gameplay impact.

## Media

https://github.com/user-attachments/assets/8f22f523-a8bb-4b8a-8696-be613028fe4f

## Requirements

## Breaking changes
`ClickInsert = false` no longer disables opening a storage by pressing E.
Use  `OpenOnActivate = false` instead.

**Changelog**
:cl:
ADMIN:
- fix: Health scanning a slime as an aghost will no longer insert the scanner into their inventory.

### DIFF
--- a/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
@@ -360,7 +360,7 @@ public abstract class SharedStorageSystem : EntitySystem
     /// <returns>true if inserted, false otherwise</returns>
     private void OnInteractUsing(EntityUid uid, StorageComponent storageComp, InteractUsingEvent args)
     {
-        if (args.Handled || !CanInteract(args.User, (uid, storageComp), storageComp.ClickInsert, false))
+        if (args.Handled || !storageComp.ClickInsert || !CanInteract(args.User, (uid, storageComp), silent: false))
             return;
 
         var attemptEv = new StorageInteractUsingAttemptEvent();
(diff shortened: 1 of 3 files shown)


slarticodefast:
> Why admins don't get the ability to just examine people and get their entire health is beyond me

Yeah, that would be better and was requested before in #30264
But someone will have to make a new UI for that.

Issue #32642, opened 2024-10-05
Grenades inside item slots cause errors


slarticodefast:
**Reproduction**
1)
- spawn plant pot
- spawn a stinger grenade or incendiary grenade
- activate the grenade and put it inside the plant pot or your pocket slot
- get a ton of these errors:
```
[ERRO] resolve: Can't resolve "Robust.Shared.GameObjects.MetaDataComponent" on entity 2511!
   at System.Environment.get_StackTrace()
   at Robust.Shared.Containers.SharedContainerSystem.IsEntityOrParentInContainer(EntityUid uid, MetaDataComponent meta, TransformComponent xform) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\Containers\SharedContainerSystem.cs:line 333
   at Content.Shared.Damage.Components.RequireProjectileTargetSystem.PreventCollide(Entity`1 ent, PreventCollideEvent& args) in D:\Code\SS14\space-station-14\Content.Shared\Damage\Systems\RequireProjectileTargetSystem.cs:line 37
   at Robust.Shared.GameObjects.EntityEventBus.<>c__DisplayClass57_0`2.<SubscribeLocalEvent>g__EventHandler|0(EntityUid uid, IComponent comp, TEvent& args) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Directed.cs:line 324
   at Robust.Shared.GameObjects.EntityEventBus.<>c__DisplayClass67_0`1.<EntSubscribe>b__0(EntityUid uid, IComponent comp, Unit& ev) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Directed.cs:line 460
   at Robust.Shared.GameObjects.EntityEventBus.EntDispatch(EntityUid euid, Type eventType, Unit& args) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Directed.cs:line 634
   at Robust.Shared.GameObjects.EntityEventBus.RaiseLocalEventCore(EntityUid uid, Unit& unitRef, Type type, Boolean broadcast) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityEventBus.Directed.cs:line 240
[...]

PR #31108, opened 2024-08-16
Cluster grenade refactor and contra markings

Plykiya:
## About the PR

Cluster grenades essentially have two categories.
- Ones that shoot projectiles (projectile grenades)
- Ones that scatter entities that may potentially trigger themselves (scattering grenades)

I split it into two distinct systems to better separate the logic and fixed a few bugs.

## Why / Balance

- Fixes https://github.com/space-wizards/space-station-14/issues/27571
- Fixes cluster grenades and clusterbanana type grenades not triggering when exploded
- Fixes stingers and incendiaries type grenades not triggering when exploded
- Fixes stacking stingers into stingers into stingers into stingers into stingers

## Technical details

Clustergrenades only trigger as a result of their on-use timer. This means that when they were destroyed, nothing happened. I gave both new categories of grenades a damage trigger so that they can be properly chained like regular explosives.

More than half of the variables in the system were unused by projectile-type grenades, so separating the two into two different systems enables easier alteration of logic or behavior.

None of the actual behavior regarding how they function has changed, it's just been separated.

- Scattering grenades and projectile grenades now have an abstract parent to make contra marking sane
- Clusterbang, cluster banana peel, slipocalypse clustersoap, and foam dart grenade migrated to Scattering Grenades and proper contra marking
- Stingers, incendiaries, and shrapnel grenades migrated to Projectile Grenades and proper contra marking

## Media

[...]

PR #33885, opened 2024-12-16
Fix lag spike from stinger grenades

iaada:
## About the PR

`ClusterGrenadeComponent` - specifically the type that spawns bullets - currently causes a lag spike when it goes off. This is because every bullet tries and fails to reference the already deleted grenade. This PR stops the bullets from looking for the grenade.

## Why / Balance

Lag is bad.

## Technical details

Changed `ShootProjectile` in `SharedGunSystem` to make the parameter for gun source nullable.

## Media

Causes ugly admin logging, but previously there was no admin logging at all.
image

## Requirements

## Breaking changes

**Changelog**

:cl: aada
- fix: Stinger grenades no longer cause lag spikes.

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ClusterGrenadeSystem.cs
+++ b/Content.Server/Explosion/EntitySystems/ClusterGrenadeSystem.cs
@@ -97,7 +97,7 @@ public sealed class ClusterGrenadeSystem : EntitySystem
                     switch (clug.GrenadeType)
                     {
                         case GrenadeType.Shoot:
-                            ShootProjectile(grenade, angle, clug, uid);
+                            ShootProjectile(grenade, angle, clug);
                             break;
                         case GrenadeType.Throw:
                             ThrowGrenade(grenade, angle, clug);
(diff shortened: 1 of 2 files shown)


beck-thompson:
fixes: #32642?

Also amazing first contribution

Plykiya:
I took a swing at the problem in my complete cluster grenade refactor and explain it similarly in my latest post, and I mean I guess you can get around the issue by just NEVER setting the ProjectileComponent's gunUid at all, it just feels wrong using gun code to shoot a projectile without providing the gunUid 

https://github.com/space-wizards/space-station-14/pull/31108

iaada:
> I took a swing at the problem in my [complete cluster grenade refactor](https://github.com/space-wizards/space-station-14/pull/31108) and explain it similarly in my latest post, and you can get around the issue by just NEVER setting the ProjectileComponent's gunUid at all which I did at some point, but it also just feels wrong using gun code to shoot a projectile without providing the gunUid because it causes bad admin logs as you've already explained so I didn't go that route

I totally agree that my PR is sloppy. I was just looking at a simple fix, but you're is definitely much better done.

Plykiya:
> > I took a swing at the problem in my [complete cluster grenade refactor](https://github.com/space-wizards/space-station-14/pull/31108) and explain it similarly in my latest post, and you can get around the issue by just NEVER setting the ProjectileComponent's gunUid at all which I did at some point, but it also just feels wrong using gun code to shoot a projectile without providing the gunUid because it causes bad admin logs as you've already explained so I didn't go that route
> 
> I totally agree that my PR is sloppy. I was just looking at a simple fix, but you're is definitely much better done.

I mean I still haven't fixed the problem

Issue #31044, opened 2024-08-15
mob's fixture layer/mask breaks after being in space wind flow


lzk228:
1. be mouse
2. accidentally enter into air pressure flow
3. now you can't go under doors

PR #33888, opened 2024-12-16
Fix space wind layer removal

lzk228:
## About the PR

fixes https://github.com/space-wizards/space-station-14/issues/31044
more in media and tech details

## Why / Balance

bug

## Technical details

now before we remove the table layer removed, we check if it even was in a fixture and write it down
and before adding it back we check if it was there before

## Media

what happens on the video:
- a human gets in a flow
- still can't pass through tables
- a mouse gets in a flow
- still can pass under tables

https://github.com/user-attachments/assets/210b2269-3078-45bf-8341-44a6eab1e8d6

## Requirements

## Breaking changes

no

**Changelog**

:cl:
- fix: Fixed breaking small mobs (e.g. mice) mask after getting into an air flow.

### DIFF
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
@@ -56,12 +56,17 @@ namespace Content.Server.Atmos.EntitySystems
                     _physics.SetBodyStatus(uid, body, BodyStatus.OnGround);
                 }
 
-                if (TryComp<FixturesComponent>(uid, out var fixtures))
+                if (TryComp<FixturesComponent>(uid, out var fixtures)
+                    && TryComp<MovedByPressureComponent>(uid, out var component))
                 {
                     foreach (var (id, fixture) in fixtures.Fixtures)
                     {
-                        _physics.AddCollisionMask(uid, id, fixture, (int) CollisionGroup.TableLayer, manager: fixtures);
+                        if (component.TableLayerRemoved.TryGetValue(id, out var wasRemoved) && wasRemoved)
+                        {
+                            _physics.AddCollisionMask(uid, id, fixture, (int)CollisionGroup.TableLayer, manager: fixtures);
+                        }
                     }
+
                 }
             }
 
@@ -80,7 +85,19 @@ namespace Content.Server.Atmos.EntitySystems
 
             foreach (var (id, fixture) in fixtures.Fixtures)
             {
-                _physics.RemoveCollisionMask(uid, id, fixture, (int) CollisionGroup.TableLayer, manager: fixtures);
+                // Mark fixtures that has TableLayer removed
+                if ((fixture.CollisionMask & (int)CollisionGroup.TableLayer) != 0)
+                    {
+                        component.TableLayerRemoved[id] = true;
+                        _physics.RemoveCollisionMask(uid, id, fixture, (int)CollisionGroup.TableLayer, manager: fixtures);
+                    }
+                else
+                    {
+                        if (!component.TableLayerRemoved.ContainsKey(id))
+                        {
+                            component.TableLayerRemoved[id] = false;
+                        }
+                    }
             }
 
             // TODO: Make them dynamic type? Ehh but they still want movement so uhh make it non-predicted like weightless?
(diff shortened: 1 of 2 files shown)


Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -80,7 +85,19 @@ private void AddMobMovedByPressure(EntityUid uid, MovedByPressureComponent compo
...
+                        component.TableLayerRemoved[id] = true;
+                        _physics.RemoveCollisionMask(uid, id, fixture, (int)CollisionGroup.TableLayer, manager: fixtures);
+                    }
+                else
+                    {
+                        if (!component.TableLayerRemoved.ContainsKey(id))
+                        {
+                            component.TableLayerRemoved[id] = false;
+                        }
+                    }
```
Partmedia, line 100: Fix indentation

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
```diff
@@ -80,7 +85,19 @@ private void AddMobMovedByPressure(EntityUid uid, MovedByPressureComponent compo
...
+                if ((fixture.CollisionMask & (int)CollisionGroup.TableLayer) != 0)
+                    {
+                        component.TableLayerRemoved[id] = true;
+                        _physics.RemoveCollisionMask(uid, id, fixture, (int)CollisionGroup.TableLayer, manager: fixtures);
+                    }
+                else
+                    {
+                        if (!component.TableLayerRemoved.ContainsKey(id))
+                        {
+                            component.TableLayerRemoved[id] = false;
```
Partmedia, line 98: Why store a hash map of layers and removed/not removed, instead of storing a hash set of only layers removed?

Partmedia (commented):
Thanks for the investigation and fix. Feedback inline.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
lzk228: haven't even thought about it lol

Content.Shared/Atmos/Components/MovedByPressureComponent.cs
```diff
@@ -27,5 +27,8 @@ public sealed partial class MovedByPressureComponent : Component
 
     [ViewVariables(VVAccess.ReadWrite)]
     public int LastHighPressureMovementAirCycle { get; set; } = 0;
+
+    [DataField]
+    public HashSet<string> TableLayerRemoved = new();
```
Partmedia, line 32: Let's add a comment to remind us what this is for.

Partmedia (commented):
Thanks for the changes. Do you know what happens if `AddMobMovedByPressure` is called again while the mob is already affected by space wind? Or does `AddMobMovedByPressure` even if a mob gets moved by space wind twice while still moving?

Content.Shared/Atmos/Components/MovedByPressureComponent.cs
Partmedia: Something along the lines of "The table layer gets removed by space wind so that... however we need to remember to put it back..."

lzk228:
> Do you know what happens if `AddMobMovedByPressure` is called again while the mob is already affected by space wind? Or does `AddMobMovedByPressure` even if a mob gets moved by space wind twice while still moving?

i believe the component stays until an entity fully gets out of an airflow

Partmedia:
> > Do you know what happens if `AddMobMovedByPressure` is called again while the mob is already affected by space wind? Or does `AddMobMovedByPressure` even if a mob gets moved by space wind twice while still moving?
> 
> i believe the component stays until an entity fully gets out of an airflow

That's not what I asked. `ExperiencePressureDifference` is called on line 160, and the comp is created by `EnsureComp` which would not prevent it running a second time on an entity that already has the component.

lzk228:
i don't quite understand, does EnsureComp clears all data in the comp to the default? so it will clear to which fixtures we have to add the mask back?

Partmedia:
The bug sequence I'm imagining:

1. Entity (let's call it "mouse") has table layer
2. Enters airflow, gets table layer removed, stashed in the component
3. Before exiting airflow, gets hit by another pressure differential, gets table layer removed, but when stashed in component doesn't have table layer because right now it doesn't have table layer
4. Mouse exits airflow, table layer isn't stashed in component
5. Bug is back

lzk228:
so removing this part should fix it
```cs
else
{
      component.TableLayerRemoved.Remove(id); // Ensure it is no longer marked as removed
}
```

Partmedia:
Maybe, can you do a more throughout analysis of the code and determine if any weird state things don't happen as a result of this?

lzk228:
no, i'm not that good into it to check for stuff like that

Partmedia (approved):
I didn't re-check carefully, but at least we caught and fixed the bug sequence I described earlier. Since this fixes a bug, I'll approve. If something else is broken we'll just squash it again.

MilonPL (approved):
Looks good to me!

lzk228:
yessssssss

PR #33894, opened 2024-12-16
Remove Slim Armor Vests

ps3moira:
## About the PR

Removed slim armor vest.
## Why / Balance

No real reason to have more bloat armor for sec. This should standardize sec's look a bit more.

**Changelog**
:cl:
- remove: Removed slim armor vest


Errant-4:
I don't think it can be objectively said that standardizing their look is an improvement

Everturning:
bartenders

ps3moira:
> bartenders

still in draft, but yeah bartenders should be able to have it but for some reason the contraband on it says otherwise

PR #33899, opened 2024-12-17
Amber station updates

southbridge-fur:
## About the PR

- Ensured all the logic gates are actually wrenched down now
- Removed the footprints in front of the Detective's office
- Added lights into the area outside the vault. (Enough people complained)
- Added some more space and rooms to the north east and north west maints since I felt like they could use them.
- Opened a firelock which was stuck closed. 
- Added more loot
- Gave the Detective and lawyers more room in their rooms. 16% and 33% more room respectively.

## Why / Balance

because I want to

### Render
Amber-0

I felt like this area could use more stuff
image

The maints bar now has a little side room, and is generally a little larger. 
image

## Requirements

## Breaking changes

**Changelog**


southbridge-fur:
I really need to open a new branch at this point. The only commits that actually differ from master at this point are the last three.

PR #33902, opened 2024-12-17
Wood walls from barricades

### LINKED ISSUE #33866: Wooden walls shouldn't require a girder to make.
Wooden walls currently are made using wood on a girder, in 13, they are made by wood on a barricade. 
Also, it just makes no sense to require __metal__ for a __wood__ wall

JustinWinningham:
## About the PR
Resolves #33866

## Why / Balance
Wooden walls should not have metal as part of their construction cost - SS13 creates a wood wall from a barricade, this change reflects that and allows for niche uses for wood when steel is unavailable.

## Technical details
Wood wall construction graph node moved to barricade construction graph and updated woodwall prototype to reflect this.

## Media
woodwallchange

## Requirements
- [ x ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
N/A

**Changelog**
:cl:
- tweak: Wood walls no longer require girders and are built from barricades instead.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Walls/walls.yml
+++ b/Resources/Prototypes/Entities/Structures/Walls/walls.yml
@@ -1106,13 +1106,14 @@
   parent: BaseWall
   id: WallWood
   name: wood wall
+  description: A wooden wall. Built by adding wood to a barricade.
   components:
   - type: Sprite
     sprite: Structures/Walls/wood.rsi
   - type: Icon
     sprite: Structures/Walls/wood.rsi
   - type: Construction
-    graph: Girder
+    graph: Barricade
     node: woodWall
   - type: Destructible
     thresholds:
@@ -1133,7 +1134,7 @@
         sound:
           collection: WoodDestroyHeavy
       - !type:ChangeConstructionNodeBehavior
-        node: girder
+        node: Barricade
       - !type:DoActsBehavior
         acts: ["Destruction"]
   - type: IconSmooth
--- a/Resources/Prototypes/Entities/Structures/barricades.yml
+++ b/Resources/Prototypes/Entities/Structures/barricades.yml
@@ -73,6 +73,9 @@
   - type: Sprite
     sprite: Structures/barricades.rsi
     state: barricade
+  - type: Construction
+    graph: BarricadeCovering
+    node: barricadecover
 
 #Directional Barricade
 - type: entity
--- a/Resources/Prototypes/Entities/Structures/barricades.yml
+++ b/Resources/Prototypes/Entities/Structures/barricades.yml
@@ -38,8 +38,8 @@
       - !type:SpawnEntitiesBehavior
         spawn:
           MaterialWoodPlank1:
-            min: 2 # The construction graphs that use this all override this value on all trees so it's just a fallback
-            max: 2
+            min: 1
+            max: 1
       - !type:DoActsBehavior
         acts: [ "Destruction" ]
   - type: AtmosExposed
--- a/Resources/Prototypes/Entities/Structures/barricades.yml
+++ b/Resources/Prototypes/Entities/Structures/barricades.yml
@@ -38,8 +38,8 @@
       - !type:SpawnEntitiesBehavior
         spawn:
           MaterialWoodPlank1:
-            min: 3
-            max: 3
+            min: 2 # The construction graphs that use this all override this value on all trees so it's just a fallback
+            max: 2
       - !type:DoActsBehavior
         acts: [ "Destruction" ]
   - type: AtmosExposed
--- a/Resources/Prototypes/Recipes/Construction/Graphs/structures/barricades.yml
+++ b/Resources/Prototypes/Recipes/Construction/Graphs/structures/barricades.yml
@@ -72,3 +72,30 @@
           steps:
             - tool: Prying
               doAfter: 5
+
+- type: constructionGraph
+  id: BarricadeCovering
+  start: start
+  graph:
+    - node: start
+      edges:
+        - to: barricadecover
+          steps:
+            - material: WoodPlank
+              amount: 2
+              doAfter: 3
+    - node: barricadecover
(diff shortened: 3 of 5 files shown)


TheShuEd:
Test fails is real

SlamBamActionman:
I conceptually approve of this PR! 👍

Resources/Prototypes/Entities/Structures/Walls/walls.yml
```diff
@@ -1106,13 +1106,14 @@
   parent: BaseWall
   id: WallWood
   name: wood wall
+  description: A wooden wall. Built by adding wood to a barricade.
```
Aeshus, line 1109: There's not much point in putting the name of the item as the description. 

Additionally, we want to avoid putting anything related to how the construction as this will show when examined, where it doesn't need to be. If they need to know how to make it, they can find it from the construction menu.

Here's my suggestion (you don't need to use it), as a play on the whole wall keeping the graytide out.

```suggestion
  description: The traditional graytide defense.
```

Resources/Prototypes/Entities/Structures/Walls/walls.yml
```diff
@@ -1106,13 +1106,14 @@
   parent: BaseWall
   id: WallWood
   name: wood wall
+  description: The traditional graytide defense.
```
slarticodefast, line 1109:
```suggestion
  description: The traditional greytide defense.
```
I did not look at the rest of the code yet.

SlamBamActionman (approved):

slarticodefast:
I gave it a test and it also allows you to build a wall from the small wooden barricade seen on the right. These are used to block doors on some maps and thus allow you to build wooden walls inside an airlock. Please make sure those cannot be upgraded.

JustinWinningham:
> I gave it a test and it also allows you to build a wall from the small wooden barricade seen on the right. These are used to block doors on some maps and thus allow you to build wooden walls inside an airlock. Please make sure those cannot be upgraded. grafik

This is existing functionality. If you open a door with a barricade over it (the built kind, not the premade kind shown in your picture) the exact same behavior occurs. see attached (test done on master, not PR branch)
existingfucnt

slarticodefast:
Ok, did some more testing and we can already build normal walls inside airlocks, so that part is ok.
But I don't think the small barricade (right) should be able to be upgraded into a full wall like the big one does.
grafik

JustinWinningham:
I agree with that. Will update.

JustinWinningham:
will fix tomorrow...

JustinWinningham:
I'm unsure what a fluid unit test failure has to do with my changes. Can somebody more experienced illuminate what might be the issue?

SlamBamActionman:
> I'm unsure what a fluid unit test failure has to do with my changes. Can somebody more experienced illuminate what might be the issue?

The actual failure is a bit further up:

```
A total of 1 test files matched the specified pattern.
  Failed NoMaterialArbitrage [2 s]
  Error Message:
     destroying a BarricadeBlock spawns more Wood than required to construct it.
Assert.That(numSpawned, Is.LessThanOrEqualTo(amount))
  Expected: less than or equal to 200
  But was:  300
```

JustinWinningham:
Ah! thank you

JustinWinningham:
Based off my code changes, I'm not sure why that test is failing.

JustinWinningham:
Based on my testing, I am unable to recreate the scenario where a player can generate infinite wood as the unit test failure indicates.

slarticodefast:
Constructing `BarricadeBlock` costs 2 wood, see the `BarricadeCovering` graph.
But it inherits from `BaseBarricade` which spawns 3 wood when destroyed.
grafik

JustinWinningham:
Ok I understand now. is it even possible for players to create that barricade? It doesn't show up in the build menu so I was under the impression that means its not possible to build.

Resources/Prototypes/Entities/Structures/barricades.yml
```diff
@@ -38,8 +38,8 @@
       - !type:SpawnEntitiesBehavior
         spawn:
           MaterialWoodPlank1:
-            min: 3
-            max: 3
+            min: 2 # The construction graphs that use this all override this value on all trees so it's just a fallback
```
SlamBamActionman, line 41: This isn't fallback; this is what spawns when you _destroy_ the barricade, i.e. hitting it with weapons. 

Using prying to deconstruct the barricade should be considered the more "careful" option compared to just smashing it. Because of that, make smashing the barricade leave 1 plank (because one breaks) while prying via the graph gives back the full 2 planks.

Resources/Prototypes/Recipes/Construction/Graphs/structures/barricades.yml
```diff
@@ -53,3 +72,30 @@
...
+              amount: 2
+              doAfter: 3
+    - node: barricadecover
+      entity: BarricadeBlock
+      edges:
+        - to: start
+          completed:
+            - !type:SpawnPrototype
+              prototype: MaterialWoodPlank1
+              amount: 1 #returns 1 less as one breaks
```
SlamBamActionman, line 94: Change this to 2; deconstructing should be the intended way to get materials back since it requires tools. See the previous comment.

SlamBamActionman (changes requested):

Resources/Prototypes/Entities/Structures/barricades.yml
JustinWinningham: Thanks for the assistance, still learning how this all works :)

JustinWinningham:
@SlamBamActionman Thanks for your patience with my learning on first PR!

SlamBamActionman (approved):

slarticodefast (approved):

PR #33904, opened 2024-12-17
[Hotfix] Add [MRP] tag to the hostname for Salamander.

JIPDawg:
## About the PR

I changed the Hostname to include the [MRP] tag so users can see at a glance it is using the MRP ruleset. 
## Why / Balance

Currently the only way to confirm Salamander is the MRP server is by reading the description. Literally EVERY other MRP server has MRP as a tag. I'm requesting this be a hotfix as we routinely get people that EORG or just forget what server they are on / know which server is the MRP server. Don't see any [MRP] tag and just assume. The only other indication is at the VERY END and it's just "RP".
## Technical details

It's a web edit, sue me.

## Requirements

- [ x ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

no cl

PR #33909, opened 2024-12-17
HOTFIX: Add [MRP] tag to the hostname for Salamander

slarticodefast:
The same as #33904 but properly targeting stable so we don't have to cherry pick.


nikthechampiongr (approved):

SlamBamActionman (approved):
"Mrp" is what my cat says when I come home from work.

PR #33915, opened 2024-12-17
Update PR size labeler line counts

slarticodefast:
grafik

The new line numbers for the labels are:
XS: 0-9
S: 10-99
M: 100-999
L: 1000-4999
XL: >5000

I think these fit nicely to the PR sizes we usually get.


SlamBamActionman:
As the person who set the initial values; agreed. We used the defaults from goonstation, and I think the proposed values are much better.

SlamBamActionman (approved):

PR #33928, opened 2024-12-18
Tweaks for the holopad

chromiumboy:
## About the PR
A collection of minor tweaks and bug fixes for the holopad

## Why / Balance

This PR changes the following:
- The listening range on holopads has been reduced from 4 tiles to 2.5 tiles. This is to reduce the amount of background chatter they might pick up when people make a holo-call in a public area
- Long-range holopads can connect to other long-range holopads that are on different grids, rather than different maps. This will allow ships to communicate with the stations they are docked with 
- AIs can now project themselves from the long-range holopads on their station. It is for parity with shorter-range holopads
- AIs now get an error pop-up if they try to answer a holo-call originating from another map. This is to clear up any confusion AI players might have as to why the call won't connect
- AIs no longer have duplicated holopad UI windows when they answer a summons to another holopad
- The draw depth of the holopad sprite has been corrected to match other floor objects 

## Technical details

N/A

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

N/A

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Machines/holopad.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/holopad.yml
@@ -22,6 +22,7 @@
   - type: StationAiVision
   - type: Sprite
     sprite: Structures/Machines/holopad.rsi
+    drawdepth: FloorObjects
     snapCardinals: true
     layers:
     - state: base
@@ -73,7 +74,7 @@
   - type: Telephone
     transmissionRange: Map
     ringTone: /Audio/Machines/double_ring.ogg
-    listeningRange: 4
+    listeningRange: 2.5
     speakerVolume: Speak
   - type: AccessReader
     access: [[ "Command" ]]
--- a/Content.Server/Holopad/HolopadSystem.cs
+++ b/Content.Server/Holopad/HolopadSystem.cs
@@ -119,7 +119,12 @@ public sealed class HolopadSystem : SharedHolopadSystem
             var source = GetLinkedHolopads(receiver).FirstOrNull();
 
             if (source != null)
+            {
+                if (_stationAiSystem.TryGetStationAiCore(args.Actor, out var stationAiCore) && stationAiCore != null)
+                    _userInterfaceSystem.CloseUi(receiver.Owner, HolopadUiKey.AiRequestWindow, args.Actor);
+
                 ActivateProjector(source.Value, args.Actor);
+            }
 
             return;
         }
--- a/Content.Server/Holopad/HolopadSystem.cs
+++ b/Content.Server/Holopad/HolopadSystem.cs
@@ -1,4 +1,5 @@
 using Content.Server.Chat.Systems;
+using Content.Server.Popups;
 using Content.Server.Power.EntitySystems;
 using Content.Server.Speech.Components;
 using Content.Server.Telephone;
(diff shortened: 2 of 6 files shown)


SlimmSlamm:
I think you should be able to change the listening radius of a holopad in game up to a certain point.

ScarKy0:
1. I feel like holopads shouldn't speak using their names/labels. "Holopad (BotanyStuff) (Kristopher Laborde) said, 'god this is so long...' "
Could probbaly be reduced to saying just holopad, or only the character name
2. Would be nice where youre being requested from(AI) or called from(non-AI) so it would say "[Service-Botany] Kristopher Laborde (Captain) is calling" on the UI or smth along those lines

chromiumboy:
> I think you should be able to change the listening radius of a holopad in game up to a certain point.

Yeah, there could be some additional controls on the 'in-call' screen to toggle a conference mode (which has a wider listening range) as well as a mute toggle.  That would have to wait 'til later though

> 1. I feel like holopads shouldn't speak using their names/labels. "Holopad (BotanyStuff) (Kristopher Laborde) said, 'god this is so long...' "
>    Could probbaly be reduced to saying just the character name
> 2. Would be nice where youre being requested from(AI) or called from(non-AI) so it would say "[Service-Botany] Kristopher Laborde (Captain) is calling" on the UI or smth along those lines

The name + label was to replicate how hand held radios handle relayed message but I see your point, I'll think on it. Having the name of the holopad the caller is coming from is a good idea. Might also have to wait on this one

ScarKy0:
Oh also, its not really needed after all (since the amount of holopads doesnt seem to be an issue) but I still think a search field would be cool

chromiumboy:
> Oh also, its not really needed after all (since the amount of holopads doesnt seem to be an issue) but I still think a search field would be cool

Thats good to know but I'll keep the filter in mind

Resources/Prototypes/Entities/Structures/Machines/holopad.yml
```diff
@@ -104,34 +104,53 @@
...
+    - Grid
+    - Map
+    - Unlimited
 
 # These are spawned by holopads
 - type: entity
   id: HolopadHologram
   categories: [ HideSpawnMenu ]
-  suffix: DO NOT MAP
+  suffix: DO NOT MAP
```
slarticodefast, line 153:
```suggestion
  suffix: DO NOT MAP
```

Resources/Prototypes/Entities/Structures/Machines/holopad.yml
```diff
@@ -90,7 +90,7 @@
   - type: WiresVisuals
   - type: Wires
     boardName: wires-board-name-holopad
-    layoutId: Holopad
+    layoutId: Holopad
```
slarticodefast, line 93:
```suggestion
    layoutId: Holopad
```

slarticodefast (commented):
I noticed don't know enough about this system yet to give it a proper review.
But at least some cleanup :)

chromiumboy:
> I noticed don't know enough about this system yet to give it a proper review. But at least some cleanup :)

Since it's a brand new system, you know as much as everyone else! :D

slarticodefast:
Summoning @SlamBamActionman so we can get it into the next release in time.

PR #33944, opened 2024-12-19
Remove zombie antag role on unzombification

### LINKED ISSUE #20162: Zombies are still considered antags after being cloned and losing all zombie related traits
## Description
- Right click
- Antag ctrl
- Zombie
- Be cloned
- Enable player overlay
- Antag
- No antag related behavior to be found in this new clone

pcaessayrs:
## About the PR
Unzombification (e.g. by cloning) now remove the zombie role from the mind of the zombified entity (which, in the case of cloning, usually won't be attached at the time). 

## Why / Balance
Fixes #20162 
This fixes a problem where unzombified entities will still appear as an antagonist in the Player Overlay and to admins.

## Technical details
CloningEvent now also broadcasts a mind EntityUid.
ZombieSystem::UnZombify now affects the mind.

## Media
za-0
za-1
za-3

The name is wrong due to some pre-existing problem that should only be happening in testing due to cloning myself as an aghost - here's what happens with the current behavior:
za-none

## Requirements

## Breaking changes
CloningEvent now should pass the mind EntityUid. This is generally useful because neither the old body nor the new body had the mind before.

**Changelog**
:cl:
- fix: Fixed cloned bodies keeping the zombie antag status

### DIFF
--- a/Content.Server/Cloning/CloningSystem.cs
+++ b/Content.Server/Cloning/CloningSystem.cs
@@ -213,7 +213,7 @@ namespace Content.Server.Cloning
             var mob = Spawn(speciesPrototype.Prototype, _transformSystem.GetMapCoordinates(uid));
             _humanoidSystem.CloneAppearance(bodyToClone, mob);
 
-            var ev = new CloningEvent(bodyToClone, mob);
+            var ev = new CloningEvent(bodyToClone, mob, mindEnt);
             RaiseLocalEvent(bodyToClone, ref ev);
 
             if (!ev.NameHandled)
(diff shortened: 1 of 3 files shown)


slarticodefast (changes requested):
This approach misses a few cases, for example the new mind swap spell.
Also this removes the role when the clone is made, not when the mind is transferred, as `UnZombify` is used to only revert the visual changes of the zombification for the clone. So if you cancel cloning or the player denies being cloned they will no longer count as a zombie.
The correct approach is to subscribe to `MindAddedMessage` and `MindRemovedMessage` with the `ZombieComponent` and handle the role change there.
I already got that done for my upcoming cloning refactor, but if you want to you can implement it that way here.

pcaessayrs:
That's okay! I will wait for the new cloning refactor and then see if this problem is still around since it's not urgent. Thanks!

PR #33950, opened 2024-12-19
Increase syndicate uplink discount amount to 6.

ScarKy0:
## About the PR

Increased the amount of discounted items syndicate agents have from 3 to 6.

## Why / Balance

The point of the discounts was to encourage trying new items and strategies, however having the amount be only 3 almost never gave anything worth using. I believe this change will help syndie agents actually make cool things with the discounts instead of never looking at the tab.
Chance to roll different discount categories remains unchanged.

## Technical details

Increased amount of uplink discounts to 6 and modified the tests to account for that.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Syndicate uplinks now have 6 discounts instead of 3.

### DIFF
--- a/Content.IntegrationTests/Tests/StoreTests.cs
+++ b/Content.IntegrationTests/Tests/StoreTests.cs
@@ -79,11 +79,11 @@ public sealed class StoreTests
             var discountComponent = entManager.GetComponent<StoreDiscountComponent>(pda);
             Assert.That(
                 discountComponent.Discounts,
-                Has.Exactly(7).Items,
-                $"After applying discount total discounted items count was expected to be '7' "
+                Has.Exactly(6).Items,
+                $"After applying discount total discounted items count was expected to be '6' "
                 + $"but was actually {discountComponent.Discounts.Count}- this can be due to discount "
                 + $"categories settings (maxItems, weight) not being realistically set, or default "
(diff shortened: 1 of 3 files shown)


slarticodefast:
Why the hell is that number hardcoded everywhere in the first place?

beck-thompson:
Love this change! Hopefully this gives some unused items more love. Even with 3 I've chosen items I never would have before, making it 6 will really help

ScarKy0:
> Why the hell is that number hardcoded everywhere in the first place?

Wish I knew. Was hella confusing to find since I thought its in a yml file

ScarKy0:
Next step after this is giving nukies discounts :godo:

ArtisticRoomba:
holy hardcode

ScarKy0:
> Might be pedantic but I think five discounts would be better (especislly if the discounts apply on multiple already-popular items)

Discounts have a decreased chance to happen on the stronger items alrrady (think emags etc), theyre not all the same chance

SlamBamActionman:
I've checked the discount menu every now and then and I have largely been disappointed by the selection. It's either something pointless, something that is still expensive, or something that won't work for my current objective. 

I guess if you end up with 4+ good discounts this might be a bit too powerful, but I am willing to see how this rolls.

SlamBamActionman (approved):

chromiumboy:
Sure, let's give it a go

PR #33956, opened 2024-12-19
Omega Power Rebalance

ArtisticRoomba:
## About the PR
Rebalanced Omega power to be more in-line with the legacy engineering design document.

The commit message says I fixed Packed, but it IS Omega, I promise.

## Why / Balance
Omega power has been in a rough spot for eternity. This doesn't help considering that lowpop maps often face the brunt of consequences when it comes to low engineering numbers. This change should help greatly.

## Media
Omega-0


ArtisticRoomba:
goddammit

PR #33957, opened 2024-12-19
Complex interactions with throwing items into disposals (Maybe overly so...)

beck-thompson:
## About the PR

Overview:
1.) Certain items are now "dunkable" and have an easier time being thrown into item containers (disposals). This includes basketballs and trash!
2.) If you are blind you always miss throwing items into the trash.
3.) If you are partly blind (Welding blindness), you have a lower chance of throwing items in depending on how blind you are.
4.) Clumsy people are very unlikely to throw stuff in the trash.

## Why / Balance

I think small interactions like this are a lot of fun and are really core to what ss14 is. I also was going to make it less likely to throw stuff in depending on how far away you were, but this ended up just not being as fun! I'm also open to upping the default chance (25%) but I'll leave that up to someone else.

## Technical details

Just added some events to `ThrowInsertContainerSystem.cs` and hooked other systems to listen for them!

## Media

https://github.com/user-attachments/assets/5a1b910a-0bf5-4c29-8192-b37c6770ef09

(Excuse the massive lag)

## Requirements

## Breaking changes

**Changelog**

:cl: Beck Thompson
- add: More interactions when throwing items into disposals!

### DIFF
--- a/Content.Shared/Containers/ThrowInsertContainerComponent.cs
+++ b/Content.Shared/Containers/ThrowInsertContainerComponent.cs
@@ -1,22 +1,22 @@
 using Robust.Shared.Audio;
+using Robust.Shared.GameStates;
 
-namespace Content.Server.Containers;
+namespace Content.Shared.Containers;
 
 /// <summary>
 /// Allows objects to fall inside the Container when thrown
 /// </summary>
-[RegisterComponent]
-[Access(typeof(ThrowInsertContainerSystem))]
+[RegisterComponent, NetworkedComponent]
 public sealed partial class ThrowInsertContainerComponent : Component
 {
     [DataField(required: true)]
     public string ContainerId = string.Empty;
 
     /// <summary>
-    /// Throw chance of hitting into the container
+    /// Probability of missing the container
     /// </summary>
     [DataField]
-    public float Probability = 0.75f;
+    public float BaseHitProbability = 0.25f;
 
     /// <summary>
     /// Sound played when an object is throw into the container.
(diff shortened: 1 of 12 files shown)


ArtisticRoomba:
finally, now I can actually drain root beer cans from across the room without it being a Missed! 90% of the time

beck-thompson:
also if a maintainer wants me to un-touch the whitespace issues I'll do it. I'll make another PR to remove them later because its really brutal

Content.Shared/Containers/ThrowInsertContainerComponent.cs
```diff
@@ -1,22 +1,22 @@
...
     [DataField(required: true)]
     public string ContainerId = string.Empty;
 
     /// <summary>
-    /// Throw chance of hitting into the container
+    /// Probability of missing the container
     /// </summary>
     [DataField]
-    public float Probability = 0.75f;
+    public float BaseHitProbability = 0.25f;
```
Coolsurf6, line 19: I would see if it's possible to look at making this more of a 75% chance to hit as the intended code was actually in reverse and should have been a chance of 75% to get it inside the Bin.

I was also seeing if maybe you could just set this to be more like 0.9f so it's a 1/10 chance of getting it in the disposal bin. Make it more of a funny thing when it misses and base all your code of the higher chance to get in, making it severely worse when you have any form of eye damage or condition.

Coolsurf6:
I'd see about changing the value for the hit probability if possible, left a comment about what I think it should be!

Coolsurf6:
Is there a chance you can come back to this? I want to just fix the bug with the code that this patches already if this is stale. I still believe the throw chance should be 75% as that was the original code was _supposed_ to be.

PR #33975, opened 2024-12-21
Feature/contraband fixes

minus1over12:
## About the PR

Numerous items have had contraband status added to them, or modified to reflect who actually uses them.
## Why / Balance

* Clothing: Players expect someone wearing a security officer uniform to be a security officer. That logic should be applicable to other roles. Hence, most department specific clothing has been restricted to their respective department.
  * Brigmedic gear has been marked as restricted to security and medical - brigmedic is currently disabled, so I'm not sure how to restrict this.
* Restricted hardsuits and similar have had the restriction added to their integrated helmets. The helmets sometimes show up by themselves (chameleon), and I think it's reasonable for Nanotrasen to not want someone running around with a blood-red helmet.
* Previously unmarked antagonist items were marked as contraband (holoparasite instructions, nuke codes, syndicate vendors, …).
* A subset of the contraband posters were marked as contraband to give sec something to do while running through maints if no one is doing crime.
* Singularity Generator, Tesla Generator, and Pulse Accelerator boards were marked as engineering contraband.
* Some security gear was changed to allow civilian use, as the bartender gets it roundstart.
* Improvised explosives should have been counted as improvised weapons.
* Weapons ammo is security restricted, except the beanbag shotgun rounds, which are civilian & security restricted.
* Every ID card, except passenger and visitor, have been restricted, as they are what gates access to other department restricted items.
* Mindshields implanters were restricted to security and command, because they are the only ones that get issued them under normal operating conditions.
* Departmental techfabs (such as sec techfab) were restricted to their departments.
* PKA access was widened to include science. This allows science to deal with rock crabs and flesh monsters without needing to call for security or salvage every 5 minutes. Science is already able to make PKAs.
* Centcomm exclusive equipment was restricted to Centcomm (Centcomm stamp, …)
* Head bedsheets are now command restricted. I'm unsure if putting grand theft on them is reasonable or not.
* Headsets were given restrictions that match their encryption keys.
* Surgical tools were restricted to medical and science to discourage unlicensed surgeries (for when that eventually gets finished or upstreamed).

## Technical details

New base contraband entities were added to accommodate restriction patterns that did not yet exist. Contraband statuses are applied by inheriting from the contraband bases, as has been the case.
## Media

Brigmedic Hardsuit on inspection
Nuke Codes on inspection

## Requirements

## Breaking changes

**Changelog**

:cl: WarPigeon
- tweak: Numerous items across 90+ categories have had their contraband status updated.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
+++ b/Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
@@ -113,7 +113,7 @@
     - WhitelistChameleon
 
 - type: entity
-  parent: ClothingEyesBase
+  parent: [ClothingEyesBase, BaseSyndicateContraband]
   id: ClothingEyesGlassesOutlawGlasses
   name: outlaw glasses
   description: A must for every self-respecting undercover agent.
@@ -188,7 +188,7 @@
   parent: [ClothingEyesBase, BaseCommandContraband]
   id: ClothingEyesGlassesCommand
   name: administration glasses
-  description: Upgraded sunglasses that provide flash immunity and show ID card status. 
+  description: Upgraded sunglasses that provide flash immunity and show ID card status.
   components:
   - type: Sprite
     sprite: Clothing/Eyes/Glasses/commandglasses.rsi
--- a/Resources/Prototypes/Entities/Clothing/Head/hats.yml
+++ b/Resources/Prototypes/Entities/Clothing/Head/hats.yml
@@ -171,7 +171,7 @@
     sprite: Clothing/Head/Hats/beret_physician.rsi
 
 - type: entity
-  parent: ClothingHeadBase
+  parent: [ClothingHeadBase, BaseMedicalSecurityContraband]
   id: ClothingHeadHatBeretBrigmedic
   name: brigmedical beret
   description: White beret, looks like a cream pie on the head.
@@ -346,7 +346,7 @@
     - WhitelistChameleon
 
 - type: entity
-  parent: ClothingHeadBase
+  parent: [ClothingHeadBase, BaseSyndicateContraband]
   id: ClothingHeadHatOutlawHat
   name: outlaw's hat
   description: A hat that makes you look like you carry a notched pistol, numbered one and nineteen more.
--- a/Resources/Prototypes/Entities/Clothing/Head/hats.yml
+++ b/Resources/Prototypes/Entities/Clothing/Head/hats.yml
@@ -445,7 +445,7 @@
     accent: SpanishAccent
 
 - type: entity
-  parent: ClothingHeadBase
+  parent: [ClothingHeadBase, BaseMedicalScienceContraband]
   id: ClothingHeadHatSurgcapBlue
   name: surgical cap
   description: A blue cap surgeons wear during operations. Keeps their hair from tickling your internal organs.
@@ -461,7 +461,7 @@
     - WhitelistChameleon
 
 - type: entity
-  parent: ClothingHeadBase
+  parent: [ClothingHeadBase, BaseMedicalScienceContraband]
   id: ClothingHeadHatSurgcapGreen
   name: surgical cap
   description: A green cap surgeons wear during operations. Keeps their hair from tickling your internal organs.
@@ -472,7 +472,7 @@
     sprite: Clothing/Head/Hats/surgcap_green.rsi
 
 - type: entity
-  parent: ClothingHeadBase
+  parent: [ClothingHeadBase, BaseMedicalScienceContraband]
   id: ClothingHeadHatSurgcapPurple
   name: surgical cap
   description: A purple cap surgeons wear during operations. Keeps their hair from tickling your internal organs.
--- a/Resources/Prototypes/Catalog/Fills/Paper/manuals.yml
+++ b/Resources/Prototypes/Catalog/Fills/Paper/manuals.yml
@@ -7,7 +7,7 @@
     content: book-text-ame-scribbles
 
 - type: entity
-  parent: BaseItem
+  parent: [BaseItem, BaseSyndicateContraband]
   id: HoloparasiteInfo
   name: "holoparasite terms and conditions"
   description: A tiny volumetric display for documents, makes one wonder if Cybersun's legal budget is way too high.
@@ -27,7 +27,7 @@
 - type: entity
   id: PaperAgrichemManual
   name: NT "AgriChem Is Fun" manual
-  description: The single sheet of instructions that came in the kit. 
+  description: The single sheet of instructions that came in the kit.
   parent: Paper
   components:
   - type: Paper
--- a/Resources/Prototypes/Entities/Clothing/Back/backpacks.yml
+++ b/Resources/Prototypes/Entities/Clothing/Back/backpacks.yml
@@ -32,7 +32,7 @@
     damageCoefficient: 0.9
 
 - type: entity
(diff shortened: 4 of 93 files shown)


minus1over12:
I guess dealing with the ERT departments is going to be more difficult than I thought. I really didn't want to add CentComm as an allowed department to a bunch of items, but I'm not sure if there is a better way to get around the 1 primary department per role rule.

minus1over12:
Changed it so that CentComm is no longer a primary department, but I am not familiar enough with what @deltanedas wanted to do with those to say that I did not just break something for him.

deltanedas:
theres no centcomm roles roundstart and they probably wouldnt be turf taggers anyway

Hrosts:
Should department-specific survival boxes be marked as contraband? The boxes often get thrown around and picked up by whoever, and unless you want to enforce people disposing of them properly, this will only give some officers stupid reasons to arrest somebody. 

Is nukie survival box even outwardly marked as belonging to the Syndicate? The gas mask inside it is contraband, yes, but otherwise I'm not sure how a person would distinguish those.

Another thing, should you be arrested for carrying a wrong backpack? Should the officer be obligated by Space Law to take away your backpack if it doesn't match? Similarly with boxes, backpacks and duffels are often discarded with little care, or often even shared (e.g. engi bringing a duffel full of materials).

minus1over12:
> Should department-specific survival boxes be marked as contraband? The boxes often get thrown around and picked up by whoever, and unless you want to enforce people disposing of them properly, this will only give some officers stupid reasons to arrest somebody.

You're right, I'll unmark those.

> Is nukie survival box even outwardly marked as belonging to the Syndicate? The gas mask inside it is contraband, yes, but otherwise I'm not sure how a person would distinguish those.

The distinction is small enough that I also in hindsight think you are right. I will unmark their survivial box.

> Another thing, should you be arrested for carrying a wrong backpack? Should the officer be obligated by Space Law to take away your backpack if it doesn't match? Similarly with boxes, backpacks and duffels are often discarded with little care, or often even shared (e.g. engi bringing a duffel full of materials).

I've personally only ever seen the botany plant bags shared, and that is with cargo, who handles large amounts of contraband anyway. My thought when restricting the bags was that a wrong backpack might indicate that it was stolen instead of borrowed with consent; think a Syndie with a steal objective that needed to make a quick getaway, so they just carry the bag with them instead of searching through it. Maybe an in the middle tier of contraband is needed, where having something makes you suspicious but is not criminal? Another possible compromise is restricting the command bags, which can contain high value targets, and leaving the other ones be, which are more likely to get shared around.

Palladinium:
I like the idea that only command bags are contraband (maybe security?), because that reflects how security would generally behave without those markers. Most officers won't shake you down because you've got the wrong bag, but the captain's bag? That super suspicious.

Hrosts:
I agree about command and security bags being restricted. As for the "suspicious" contraband tier I think sec should decide by themselves whether something is only suspicious but not illegal.

Resources/Prototypes/Entities/Clothing/Head/hats.yml
```diff
@@ -85,7 +85,7 @@
       sprite: Clothing/Head/Hats/casa.rsi
 
 - type: entity
-  parent: ClothingHeadBase
+  parent: [ClothingHeadBase, BaseRestrictedContraband]
```
Hrosts, line 88:
```suggestion
  parent: [ClothingHeadBase, BaseCommandContraband]
```
I think you meant to put them into Command contraband.

Resources/Prototypes/Entities/Clothing/Head/hats.yml
```diff
@@ -101,7 +101,7 @@
     - WhitelistChameleon
 
 - type: entity
-  parent: ClothingHeadBase
+  parent: [ClothingHeadBase, BaseRestrictedContraband]
```
Hrosts, line 104:
```suggestion
  parent: [ClothingHeadBase, BaseCommandContraband]
```

Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
```diff
@@ -113,7 +113,7 @@
     - WhitelistChameleon
 
 - type: entity
-  parent: ClothingEyesBase
+  parent: [ClothingEyesBase, BaseSyndicateContraband]
```
Hrosts, line 116: The outlaw glasses are the ones with fake brows and nose. I am not sure if making them contraband is a good idea. They are a joke item, and this only makes them "bad" to ever buy. Technically, stuff like syndie pyjamas is considered contraband, so there's a precedent, but still.

Hrosts:
Another thing I might want to criticize this on is the sheer amount of contraband markers an officer will see on extremely mundane items. Like, the entirety of someone's outfit will be marked as contraband, making it harder to notice when it's actually something you should pay attention to. I fear this will just be a lot of visual noise, but let's see what maintainers will say.

Oh, and they might ask you to split the PR into several parts because of how many files it touches.

Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
minus1over12: My thought with this was to mirror the change to cat ears that was made in #30930 and merged in. Those are also a joke item.

Resources/Prototypes/Entities/Clothing/Head/hats.yml
minus1over12: I meant to put it into science contraband, thanks for the catch.

Resources/Prototypes/Entities/Clothing/Head/hats.yml
minus1over12: I meant to put it into engineering contraband, thanks for the catch.

minus1over12:
I am going to drop the changes to most of the non-mindshield role clothing and come back to those in a different PR.

slarticodefast:
I'm closing this PR since it has been merge conflicted for 6 months. If someone wants to recreate this PR I recommend splitting it up into smaller batches.
Conceptually I had a discussion with the other maintainers and the general opinion was that most of these items don't need to be contraband. The contraband system is there to make it easy for security to decide what is illegal and worth jailtime. Adding it to this many items creates both an information overload, and makes the actual crimes less visible because of the huge amount of contraband crimes security will have to ignore.

Some examples (I did not go through the whole list):
- ClothingBackpackDuffelSurgeryFilled is BaseMedicalScienceContraband: I don't think this is needed. If we start with this then we would have to restrict kitchen knifes and other tools as well, because those are stronger weapons
- BoxMagazine, BoxEncryptionKeyCargo and other boxes: They are only cardbord boxes and the content inside is already restricted, so I don't think it should be contraband
- ClothingBackpackBrigmedic and other brigmedic clothes are sec restricted: It's a normal backpack, no need for it being contraband
- ClothingHandsGlovesHop is command restricted: they are normal gloves, a sec officer can still ask you "why do you have the HOP's gloves?" and jail you if they are stolen, but I see no reason to have them be contraband themselves. Same for other command clothing
- hardsuit helmets: these aren't even their own item, they only spawn in your inventory when you wear them
- ClothingHeadHatHoodBioJanitor: same, it's just drip

I think we could benefit from a design doc outlining what exactly should be contraband and what shouldn't be, before we attempt doing a major cleanup for consistency, in addition to spacelaw changes fixing the problem of
"that one person stole restricted gear, but that other person wore a hat from another department, same punishment according to spacelaw".

PR #33977, opened 2024-12-21
Implement field-deltas for melee

metalgearsloth:
Title.


SlamBamActionman (dismissed):
https://github.com/orgs/community/discussions/4452 Thanks Github.

The following lines change component properties without dirtying them. 
`SharedMeleeWeaponSystem`, line 444, line 838. 
`ZombieSystem.Transform`, lines 128-133, line 178. _(I assume it isn't necessary here since so many fields are being changed it's just easier to dirty the entire comp)_

Can you confirm if these instances are being handled correctly? Otherwise, code looks good.

Also use the PR template wawa

Issue #33954, opened 2024-12-19
Holopad future features


chromiumboy:
Here's a general list of features to consider for holopads in the future:

- [ ] Include a text-based filter at the top of the contacts list so users can quickly narrow down a search for specific holopads
- [ ] Use the view subscription system to preload entities into the AI player's PVS when they get a summons from a holopad
- [ ] Introduce a system to personalize AI holographic avatars
- [ ] Give holopads the ability to change the color of the holograms they project and those to which they are connected too (e.g., syndicate holopads being able to turn the holograms on linked holopads red)
- [ ] Add additional controls to the 'in-call' UI, such as a mute toggle and a conference mode (i.e., change the listening range of the holopad) toggle 
- [ ] Have speech relayed from the holopad holograms rather than the holopad unit itself
- [ ] Include the name of the holopad that is transmitting a call alongside the caller ID on the 'receiving call' UI

spanky-spanky:
Is there any chance we could have emotes transmitted through holopads? Mime communication is very quiet otherwise.

chromiumboy:
> Is there any chance we could have emotes transmitted through holopads? Mime communication is very quiet otherwise.

The current chat system doesn't support it, we'll have to wait until the refactor is done

spanky-spanky:
> > Is there any chance we could have emotes transmitted through holopads? Mime communication is very quiet otherwise.
> 
> The current chat system doesn't support it unfortunately, we'll have to wait until the refactor is done

![image](https://github.com/user-attachments/assets/8069f1a2-903d-4176-a83f-e5d3057c16c5)
Interestingly enough emotes through the emotes menu (Y) work.

ScarKy0:
> Interestingly enough emotes through the emotes menu (Y) work.

Funnily enough, thats the only type of emoting that the chat system supports that works for holopads

chromiumboy:
> > > Is there any chance we could have emotes transmitted through holopads? Mime communication is very quiet otherwise.
> > 
> > 
> > The current chat system doesn't support it unfortunately, we'll have to wait until the refactor is done
> 
[...]

PR #33978, opened 2024-12-21
Speech is relayed by holopad holograms

chromiumboy:
## About the PR

A small change to holopads - speech is now relayed by their hologram rather than the holopad unit itself

## Why / Balance

It seems more natural that the hologram appears to speak, even though the sound is coming from a machine. It also shortens the speaker's name, so it takes up less space in the world space and the chat window

Works towards fulfilling #33954

## Technical details

A new field was added to TelephoneComponent, allowing telephones to nominate another entity with the speech component to act as its speaker. TelephoneSystem checks for these speakers first when attempting to send in character messages, using the telephone as the speaker if it does not find one

## Media

holopad_hologram_speech

## Requirements

## Breaking changes

N/A

**Changelog**

N/A


Content.Shared/Telephone/TelephoneComponent.cs
```diff
@@ -92,6 +93,12 @@ public sealed partial class TelephoneComponent : Component
     [DataField]
     public bool UnlistedNumber = false;
 
+    /// <summary>
+    /// Speech is relayed through this entity instead of the telephone
+    /// </summary>
+    [DataField]
```
chromiumboy, line 99: Needs to be replace with VV read only

PR #33981, opened 2024-12-21
Fix Double Muzzle Flash

sleepyyapril:
## About the PR

Fixes the double muzzle flash from guns shooting by providing the user to all effect functions.

## Why / Balance

The gun is meant to give off one muzzle flash per bullet, not two.

## Technical details

Added a User field to the MuzzleFlashEvent to properly send the user to the CreateEffect method.

## Media

Before
After

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed double muzzle flash for the user.


TheShuEd:
We have guns that can fire without the user (shuttle big guns for example), have you checked how it works with them? If possible can you attach media with them?

sleepyyapril:
> We have guns that can fire without the user (shuttle big guns for example), have you checked how it works with them? If possible can you attach media with them?

While it _seemingly_ worked fine, I'm going to err on the side of safety and make it use the gun uid when user is null.

metalgearsloth:
Do we really need to network the user? It should be getting filtered before getting sent to the client and not after.

sleepyyapril:
> Do we really need to network the user? It should be getting filtered before getting sent to the client and not after.

You don't technically need to, no, considering that filtering does happen server-side, but I wanted to be careful because client-side does contain a "TrackComponent" where, if a user is supplied, it's put in there. I have no knowledge of what that is used for, so if it's nothing I can just remove the networking part entirely.

beck-thompson (changes requested):
This is a nasty one... Good job on finding this! I did a little more digging and I think you could simplify this logic. I'm going to do a quick explanation here:

The reason the event is getting triggered twice is because the client is running the event itself, and then its getting sent an event from the server to run it again.

`CreateEffect` is an overloaded function whose fields are actually different depending if your server or client. On client, the third field tells you the location of the muzzle flash, on server, the third field is the user sending the event (It wont the message to the user as they already ran the command).

This means, whenever the client runs we want the third field to be the gun, and on server we want it to be the user who shot it (Or null if its emplaced).

Your code works here because:

Shooting static emplaced gun has no user, so the location is the gun which works on client. The "user" in server is the gun which is also fine because the gun isn't an actor so it just ignores it.

A user shooting the gun, it also works because on client the location is the user which is is fine (The user and gun are in the same location).
On server, the person not to send a message to is the user which is correct.

It is also probably a good idea to comment the `CreateEffect` functions in client, server, and shared, to explain the differences!

Also, there is another (I'm pretty sure unrelated) bug:
image

To reproduce: drag a cannon in space and then fire it, while moving (If you stand still its fine)

beck-thompson (approved):
Looks good! Tested and everything seems to work. I'll add the comments in another PR (We talked on discord) :saluting_face:

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
```diff
@@ -508,7 +508,8 @@ protected void MuzzleFlash(EntityUid gun, AmmoComponent component, Angle worldAn
             return;
 
         var ev = new MuzzleFlashEvent(GetNetEntity(gun), sprite, worldAngle);
-        CreateEffect(gun, ev, gun);
+        var sending = _netManager.IsServer && user != null ? user : gun;
+        CreateEffect(gun, ev, sending);
```
ElectroJr, line 512: Pretty sure this could and probably should just be
```suggestion
        CreateEffect(gun, ev, user ?? gun);
```

ElectroJr (changes requested):

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
ElectroJr: Apparently they're meant to be different methods on the client & server? In which case that really needs to be fixed, they shouldn't be overriding the same base method if they perform different functions

ElectroJr: Yeah nevermind this isn't right

ElectroJr, line 512: This just be
```suggestion
        CreateEffect(gun, ev, user);
```
And assuming the client-side effect is always meant to track the gun, it should just use the gun argument and rename the `tracked` argument back to `user` so that it's consistent with the shared & server-side methods

SlamBamActionman (approved):
Makes sense to me - server is meant to filter out the third argument from the list of recipients of the muzzle flash event, and this fixes it.

PR #33993, opened 2024-12-21
Action charges refactor

metalgearsloth:
- Fixes the slight godmoding of baseactioncomponent.
- Gets back 1ms of server time.<

## Breaking changes
Charges got moved to LimitedCharges + AutoRechargComps. ChargesComponent also had a bunch of API changes to facilitate combining the 2 systems.


deltanedas:
kid named LimitedChargesComponent

metalgearsloth:
> kid named LimitedChargesComponent

LimitedCharges needs reworking...

keronshb:
DNM until Wizard is released and action plan post-launch is finalized.

iaada (commented):
I hope you don't mind some random contributer nitpicking...

Content.Shared/Charges/Systems/ChargesSystem.cs
```diff
@@ -0,0 +1,212 @@
...
+    {
+        if (!args.IsInDetailsRange)
+            return;
+
+        var rechargeEnt = new Entity<LimitedChargesComponent?, AutoRechargeComponent?>(uid, comp, null);
+        var charges = GetCurrentCharges(rechargeEnt);
+        using var _ = args.PushGroup(nameof(LimitedChargesComponent));
+
+        args.PushMarkup(Loc.GetString("limited-charges-charges-remaining", ("charges", charges)));
+        if (GetCurrentCharges((uid, comp, null)) == comp.MaxCharges)
```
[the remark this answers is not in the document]
metalgearsloth: There was some reason I didn't but it's probably not relevant anymore.

Resources/Prototypes/Magic/projectile_spells.yml
```diff
@@ -29,9 +29,10 @@
   name: Fireball II
   description: Fires a fireball, but faster!
   components:
+  - type: LimitedCharges
+  - type: AutoRecharge
+    rechargeDuration: 10
   - type: WorldTargetAction
-    useDelay: 10
-    renewCharges: true
```
keronshb, line 34: I forgot to remove renew charges from here. The use delay should still be intact.

From discord, @ActiveMammmoth said that "also when used for the first time, the sprite becomes greyed out forever, even after cooldown finishes"

Resources/Prototypes/Magic/projectile_spells.yml
```diff
@@ -51,9 +52,10 @@
   name: Fireball III
   description: The fastest fireball in the west!
   components:
+    - type: LimitedCharges
+    - type: AutoRecharge
+      rechargeDuration: 8
     - type: WorldTargetAction
-      useDelay: 8
-      renewCharges: true
```
keronshb, line 56: I forgot to remove renew charges from here. The use delay should still be intact.

keronshb (changes requested):

keronshb:
@ActiveMammmoth had this feedback from discord:

- ActionOnInteract staves don't work correctly
- Spell upgrades don't work correctly

ActiveMammmoth:
Apologies if I'm reiterating anything that Keron has mentioned.
- ActionOnInteract staves currently don't recognize their charge count and allow for infinite usage with zero delay (Staff of Animation, RGB Wand)
- Action sprites appear to be greyed out despite being available for use
- Action cooldown and action charge cooldown seem to be two separate things which causes issues. Example being Fireball lvl 3 which has a charge cooldown of 8 seconds as intended but a regular cooldown of 15 seconds which prevents it from being usable despite being recharged.

Content.Client/Actions/ActionsSystem.cs
```diff
@@ -152,7 +128,7 @@ public override void UpdateAction(EntityUid? actionId, BaseActionComponent? acti
             if (!ResolveActionData(actionId, ref action))
                 return;
 
-            action.IconColor = action.Charges < 1 ? action.DisabledIconColor : action.OriginalIconColor;
+            action.IconColor = _charges.GetCurrentCharges(actionId.Value) < 1 ? action.DisabledIconColor : action.OriginalIconColor;
```
keronshb, line 131: If something doesn't have a limited charges component, like Scream or Combat Mode, this makes those actions start greyed out.

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -416,21 +316,17 @@ private void OnActionRequest(RequestPerformActionEvent ev, EntitySessionEventArg
...
             return;
 
-        var curTime = GameTiming.CurTime;
-        if (IsCooldownActive(action, curTime))
-            return;
-
-        // TODO: Replace with individual charge recovery when we have the visuals to aid it
-        if (action is { Charges: < 1, RenewCharges: true })
-            ResetCharges(actionEnt, true, true);
-
```
keronshb, line 433: This makes an unintended huge nerf.
If you have something with 2 charges and a use delay, the action fires off, so does the use delay, and then you're unable to use an action _until_ the use delay is up. 

The behavior should be:
If an action still has charges remaining, use delay should not trigger until the charges are used up.

metalgearsloth: Charges didn't work before and I'm not super interested in design changes to it in the same PR, I'm mainly worried about the big performance issues charges are causing atm. This is effectively a selective revert without having to completely remove everything.

keronshb: We need some sort of information to the player that a charge is recharging. Even if it's just a timer on the examine text.

And logic should be:
Use delay is never longer/equal to the recharge timer.

metalgearsloth:
> Apologies if I'm reiterating anything that Keron has mentioned.
> 
>     * Action cooldown and action charge cooldown seem to be two separate things which causes issues. Example being Fireball lvl 3 which has a charge cooldown of 8 seconds as intended but a regular cooldown of 15 seconds which prevents it from being usable despite being recharged.

This was already existing with the charges PR.

I think we need to prioritise fixing the performance problems and should either just revert charges entirely if this pr is insufficient or leave it for now. Content shouldn't take priority over 1ms of server time being wasted. I pointed out this could happen in the original PR but let it slide and then it did anyway so there's precedent that it was known in advance it would need reverting.

keronshb:
> > Apologies if I'm reiterating anything that Keron has mentioned.
> > ```
> > * Action cooldown and action charge cooldown seem to be two separate things which causes issues. Example being Fireball lvl 3 which has a charge cooldown of 8 seconds as intended but a regular cooldown of 15 seconds which prevents it from being usable despite being recharged.
> > ```
> 
> This was already existing with the charges PR.
> 
> I think we need to prioritise fixing the performance problems and should either just revert charges entirely if this pr is insufficient or leave it for now. Content shouldn't take priority over 1ms of server time being wasted. I pointed out this could happen in the original PR but let it slide and then it did anyway so there's precedent that it was known in advance it would need reverting.

Previously the UseDelay also conveyed Charges. We didn't have support (at the time) to do individual charges recharging, so instead I intentionally had all charges recharge at once with the usedelay.

Now that the Use Delay and Recharge Timer are independent of each other, we now have a scenerio where Use Delay times could be larger than or equal to the Recharge Timer.

So we 100% need to convey this to the player somehow or this will look unfinished.

If doing an animation is out of scope, I'm asking for you to just convey the time on the examine screen (like how remaining / charges are). And I'm asking to do another pass to see which Use Delays are longer or equal to the Recharge Times and to reduce the Use Delays significantly so the Charges recharge properly.

Content.Shared/Charges/Systems/SharedChargesSystem.cs
```diff
@@ -1,103 +1,216 @@
...
+            return;
+        }
+
+        action.Comp.LastCharges = adjusted;
+        action.Comp.LastUpdate = _timing.CurTime;
+        Dirty(action);
+    }
+
+    [Pure]
+    private TimeSpan GetNextRechargeTime(Entity<LimitedChargesComponent?, AutoRechargeComponent?> entity)
```
keronshb, line 172:
```suggestion
    public TimeSpan GetNextRechargeTime(Entity<LimitedChargesComponent?, AutoRechargeComponent?> entity)
```

public for where this will be used.

Content.Client/UserInterface/Systems/Actions/Controls/ActionButton.cs
```diff
@@ -195,10 +201,12 @@ private void OnUnpressed(GUIBoundKeyEventArgs args)
...
 
-        if (_action is { Charges: not null })
+        if (_entities.TryGetComponent(ActionId, out Shared.Charges.Components.LimitedChargesComponent? actionCharges))
         {
-            var charges = FormattedMessage.FromMarkupPermissive(Loc.GetString($"Charges: {_action.Charges.Value.ToString()}/{_action.MaxCharges.ToString()}"));
-            return new ActionAlertTooltip(name, decr, charges: charges);
+            var charges = _sharedChargesSys.GetCurrentCharges((ActionId.Value, actionCharges, null));
+
+            var chargesText = FormattedMessage.FromMarkupPermissive(Loc.GetString($"Charges: {charges.ToString()}/{actionCharges.MaxCharges}"));
+            return new ActionAlertTooltip(name, decr, charges: chargesText);
```
keronshb, line 209:
```suggestion
            var charges = _sharedChargesSys.GetCurrentCharges((ActionId.Value, actionCharges, null));
            var chargesText = FormattedMessage.FromMarkupPermissive(Loc.GetString($"Charges: {charges.ToString()}/{actionCharges.MaxCharges}"));

            if (_entities.TryGetComponent(ActionId, out AutoRechargeComponent? autoRecharge))
            {
                var chargeTimeRemaining = _sharedChargesSys.GetNextRechargeTime((ActionId.Value, actionCharges, autoRecharge));
                chargesText.AddText(Loc.GetString($"{Environment.NewLine}Time Til Recharge: {chargeTimeRemaining}"));
            }

            return new ActionAlertTooltip(name, decr, charges: chargesText);
```

This way we relay the info back to the user how long the next charge will be.

image

image

Content.Shared/Charges/Systems/SharedChargesSystem.cs
```diff
@@ -1,103 +1,216 @@
...
+        var timeRemaining = fullTime - _timing.CurTime;
+
+        if (timeRemaining < TimeSpan.Zero)
+        {
+            return TimeSpan.Zero;
+        }
+
+        var nextChargeTime = timeRemaining.TotalSeconds % entity.Comp2.RechargeDuration.TotalSeconds;
+        return TimeSpan.FromSeconds(nextChargeTime);
     }
```
keronshb, line 190: What happens right now is if this is called, it gets the charge time right?
But if another charge is used, this is reset.

So if you had limited charge to 3/3 and auto charge to 20 seconds and you used an action, the action goes to 2/3 and the timer starts. Works as expected.
If you fire off another charge while the timer is active, the timer will reset. So the action goes to 1/3 and the timer goes back to 20 seconds. The first charge still doesn't continue counting down and instead you need to wait another full 20 for the charge to recharge.

I don't think this is breaking enough for me to want to hold up the PR, but please put a TODO: with the bug information in this method so someone can look at it in the future.

keronshb (dismissed):
New changes requested.
Staff of Animation still has an issue and me and Mammmoth are looking into it.

Content.Client/UserInterface/Systems/Actions/Controls/ActionButton.cs
metalgearsloth: This shouldn't even be hardcoded in here to begin with?

keronshb: Can you explain more why you think this is hardcoded here?
This wouldn't be changed in the action refactor I have planned unless I had someone to work on an actual animation for charges & recharge time on the refactor.

What really is the issue of showing people the information they need instead of keeping it hidden? I do not want recharge time to _not_ be conveyed to the user.

metalgearsloth: For future archaeologists tl;dr is that any number of systems may want to modify tooltips like itemstatus, alerts, or actual tooltips, but for now it's just being slapped in.

PR #34016, opened 2024-12-22
Fix weird camera rotation/movement in mechs

### LINKED ISSUE #30528: Odd camera rotation with RelayInputMover
## Description

When a player is controlling an item with relayed input movement (e.g. a mech), rotating the camera results in odd angles and movement.

**Reproduction**

1. Spawn a Ripley.
2. Climb into the Ripley.
3. Rotate the camera, it will not resolve to a cardinal direction of the grid you're on.
4. Rotate the camera in the same direction three more times.  The first two will have the same issue at different angles, and the third goes back to a cardinal direction.

**Screenshots**

In the video below, a pattern of (left, right, left, right, up, down, up, down) is input, and the screen is rotated clockwise.  This is repeated four times.  This was recorded in a build generated on `5bc9c04a`.

https://github.com/user-attachments/assets/0a927d43-9c15-42f8-9fb0-a58327c627b3

**Additional context**

Looked into this a little - when in the odd rotations, the target rotation for both the relayer and the relayed seem to be set well, but the rotation may be reset each update and the lerp never settles.  Unsure why.

### LINKED ISSUE #34022: Changing viewport rotation while not attached to a grid causes movement to behave in unexpected ways
## Description
The title. Also renders you unable to correct the viewport rotation. This can make getting back aboard the station difficult if you're in space as it can cause movement controls to become inverted of sorts.

**Reproduction**
Head to space. While adrift, use numpad 7 or 9 to rotate the camera. Notice that the view port is tilted at an odd angle, cannot be reverted, and movement direction is not as expected.

**Screenshots**
I can't provide a video just yet. Ill do so tomorrow.

Tayrtahn:
## About the PR

Rotating the camera while piloting a mech or other entity using a `RelayInputMoverComponent` will no longer cause the camera to wind up at weird angles or alter movement directions.

## Why / Balance

Fixes #30528.
Might also fix #34022?

## Technical details

The Server version of MoverController has logic in UpdateBeforeSolve that makes it not update entities that have a RelayInputMoverComponent, so their RelativeRotation field is essentially frozen while they are relaying input. This frozen values is included in RelayInputMoverComponent's state data and synced to the client. Additionally, the frozen RelativeRotation value is passed from the relayer to the relayed entity, and that state is also networked synced. This also interacts with EyeLerpingSystem and its networking to make for a confusing mess of interactions.

This PR simply removes the guard clause that prevents the server version of MoverController from updating entities with a RelayInputMoverComponent. This enables the serverside RelativeRotation field to be lerped just like the clientside version.

**NOTE**: I'm a bit uncomfortable deleting what seems like a purposefully-written guard clause, but I can't work out why it needs to exist. Any explanations are appreciated.

**Edit**: after some discussion, it seems that the original purpose of the guard clause was to avoid issues with ordering, since the relay logic runs while iterating through all the movers and affects other movers. This PR now removes the relay logic from the movers loop into a separate loop which runs after all the movers have been updated, ensuring consistent order.

## Media

https://github.com/user-attachments/assets/4a6f90bb-22ca-436b-8b53-337e57ab6e54

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Client/Physics/Controllers/MoverController.cs
+++ b/Content.Client/Physics/Controllers/MoverController.cs
@@ -94,7 +94,10 @@ public sealed class MoverController : SharedMoverController
             return;
 
         if (RelayQuery.TryGetComponent(player, out var relayMover))
+        {
             HandleClientsideMovement(relayMover.RelayEntity, frameTime);
+            HandleRelayMovement(relayMover.RelayEntity);
+        }
 
         HandleClientsideMovement(player, frameTime);
     }
--- a/Content.Server/Physics/Controllers/MoverController.cs
+++ b/Content.Server/Physics/Controllers/MoverController.cs
@@ -67,9 +67,6 @@ public sealed class MoverController : SharedMoverController
         {
             var physicsUid = uid;
 
-            if (RelayQuery.HasComponent(uid))
-                continue;
-
             if (!XformQuery.TryGetComponent(uid, out var xform))
             {
                 continue;
--- a/Content.Server/Physics/Controllers/MoverController.cs
+++ b/Content.Server/Physics/Controllers/MoverController.cs
@@ -98,6 +98,12 @@ public sealed class MoverController : SharedMoverController
                 frameTime);
         }
 
+        var movementRelayTargetEnumerator = AllEntityQuery<MovementRelayTargetComponent, InputMoverComponent>();
+        while (movementRelayTargetEnumerator.MoveNext(out var uid, out var relay, out var mover))
+        {
+            HandleRelayMovement((uid, relay, mover));
+        }
+
         HandleShuttleMovement(frameTime);
     }
 
--- a/Content.Shared/Movement/Systems/SharedMoverController.cs
+++ b/Content.Shared/Movement/Systems/SharedMoverController.cs
@@ -108,6 +108,32 @@ public abstract partial class SharedMoverController : VirtualController
         UsedMobMovement.Clear();
     }
 
+    protected void HandleRelayMovement(Entity<MovementRelayTargetComponent?, InputMoverComponent?> entity)
+    {
+        if (!Resolve(entity, ref entity.Comp1, ref entity.Comp2))
+            return;
+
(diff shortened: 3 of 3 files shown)


Kadeo64:
rejoice

TheShuEd (approved):
checked in game, looks as intented

code lgtm

VerinSenpai:
definitely fixes #34022 appreciate that a bunch its been giving me a headache for a while now.

metalgearsloth:
Note that the relayquery check was specifically to avoid order-of-operations problems so not sure if it can be removed safely.

Tayrtahn:
> Note that the relayquery check was specifically to avoid order-of-operations problems so not sure if it can be removed safely.

Yep! That's why I moved all the relay-applying code into a separate loop.

If I understand the issue correctly, the problem was that relayed inputs were applied to MoverControllers while iterating all the MoverControllers, which meant that the results would be different depending on whether or not the target of a relay had already been processed by the loop. This change makes it so that ALL of the relay updates wait to be applied until after all of the MoverControllers have been processed. So I'm pretty sure the check isn't needed any more?

VerinSenpai:
Praying everyday for this ;(

VerinSenpai:
@metalgearsloth ^ worth far more than any personal testing I can provide

metalgearsloth:
> @metalgearsloth ^ worth far more than any personal testing I can provide

See my comment on why it being fine is at the whims of entity ordering and relays spawning first, which is liable to change. There's another pr that makes it more deterministic.

VerinSenpai:
You're breakin my heart ;(

I understand

VerinSenpai:
don't suppose this is superseded by the movement refactor?

Tayrtahn:
Superceded by https://github.com/space-wizards/space-station-14/pull/36847

whatston3:
> Superceded by #36847

@Tayrtahn, even as a temporary fix, it was still greatly appreciated.  Thanks again.

PR #34028, opened 2024-12-23
Fix battery charging stopping just short of being full

### LINKED ISSUE #18090: cap's laser gun wont fully charge
## Description

unless you have it in your active hand the gun wont recharge to full and will just stay at 9 until you fire it again and hold it in your active hand until it charges back to 10
**Reproduction**

simply fire the gun to empty and wait with it on your belt or in your inactive hand
**Screenshots**

N/A
**Additional context**

I don't think this needs much more explaining

Errant-4:
## About the PR
In some rare arithmetic alignments, batteries can currently be charged close-to-but-not-exactly 100%, and refuse to "finish". This is usually irrelevant and unnoticeable, with the exception of battery guns, which will report a missing shot in these cases. The gun will simply not fully charge, until it is fired again at least once.

## Why / Balance
Bug
Fixes #18090

## Technical details
Batteries are currently considered full at various* values close to 100% (Different codepaths use a few different calculations, between 99 and 100). With this change, they must be exactly at the MaxCharge value to count as full. Thus, they will never stop charging before 100%.

To make the calculations consistent, the BatteryComponent.IsFullyCharged (which calculated from the batterycomponent's values) is removed, and every use of it (as well as a few places that calculated it for themselves directly) now relies on BatterySystem.IsFull()

## Requirements

## Breaking changes
BatteryComponent.IsFullyCharged has been removed, use BatterySystem.IsFull instead.
BatterySystem.IsFull no longer reports values between 99 and 100% as "full". Only the CurrentCharge = MaxCharge counts as full.

**Changelog**
:cl: Errant
- fix: Battery-powered weapons will no longer occasionally refuse to recharge the last missing shot.


Errant-4:
@deltanedas do you remember why we needed the fuzzy comparison to MaxCharge in the first place why not just strictly compare like this?

deltanedas:
the 99% thing was for ninja charging since its always drawing power it would try to charge forever i think??

Errant-4:
> the 99% thing was for ninja charging since its always drawing power it would try to charge forever i think??

Is this about draining power from an APC with the ninja gloves? It stopped fine

deltanedas:
ok good shitcode begone

Centronias (approved):

PR #34032, opened 2024-12-23
Remove deprecated sensor monitoring console

slarticodefast:
## About the PR
It was added in #18840 as a way to debug the TEG, but never fully implemented and at the moment it just crashes the game when spawned. It can spawn randomly from chrsitmas presents, so we should remove it.
The functionality (monitoring SMES battery level and atmos devices) has been replaced by the power monitoring and atmos monitoring consoles.

## Why / Balance
Remove broken and unused code.

## Technical details
Remove all the relevant files and components.
Migrate the entities to null.

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
Removed all systems, components and UIs relevant for the sensor monitoring console console.
Removed `BatterySensorComponent`, which was part of the SMES, but unused.

**Changelog**
not player facing

PR #34035, opened 2024-12-23
Fix the sensor monitoring console

PJB3005:
Still isn't really suitable to just map but at least it doesn't outright NRE anymore.

Alternative to #34032


Tayrtahn (approved):

slarticodefast (approved):
Tested and works as intended.

PR #28902, opened 2024-06-12
Make SSD character fall asleep

lzk228:
## About the PR

If entity is SSD, it got asleep with force (cannot be awaked)
When entity no longer SSD, it can be awaked
Resolves #21906 

## Why / Balance

Cool

## Technical details

no

## Media

https://github.com/space-wizards/space-station-14/assets/124214523/2568d015-849c-456e-858a-547c0bb6ef3a

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- tweak: SSD character now will fall asleep until they are no longer SSD.

Issue #25623, opened 2024-02-26
Character has SSD indicator on aghost


lzk228:
1. aghost
2. everyone can see youre aghosted

deltanedas:
they can still examine and see catatonic

lzk228:
> they can still examine and see catatonic

![image](https://github.com/space-wizards/space-station-14/assets/124214523/d6e2e7db-5e27-40cf-90b3-6e182159d186)

VasilisThePikachu:
This sounds... Intended? If you ghost our of your body normally this would happen still. I don't even think as an admin you should be aghosting and returning to your body in the first place

PR #34039, opened 2024-12-23
SSD sleep take 2

### LINKED ISSUE #21906: Feature request: SSD characters should really fall asleep

lzk228:
## About the PR

If an entity is SSD, it gets a "timer" from cvar (default 10 mins) and after it, the entity gets forced to sleep
When entity is no longer SSD, it can be awakened

Reopened https://github.com/space-wizards/space-station-14/pull/28902 (was unable to reopen, because force-pushed)
Resolves #21906 

## Why / Balance

Nice feature

## Technical details

Not really much, two new cvars, bool for ssd sleep and float for the timer
then some easy checks with time and giving or removing of the ForcedSleepComponent

## Media

https://github.com/user-attachments/assets/f5471cc0-7287-4514-aefc-60f16d04aaa0

## Requirements

## Breaking changes

no

**Changelog**

:cl:
- tweak: When an entity becomes SSD, it will be forced to sleep after 10 minutes. It can be awakened if no more SSD.

### DIFF
--- a/Content.Shared/CCVar/CCVars.Ic.cs
+++ b/Content.Shared/CCVar/CCVars.Ic.cs
@@ -45,4 +45,17 @@ public sealed partial class CCVars
     /// </summary>
     public static readonly CVarDef<bool> ICShowSSDIndicator =
         CVarDef.Create("ic.show_ssd_indicator", true, CVar.CLIENTONLY);
+
+    /// <summary>
+    ///     Forces SSD characters to sleep after
+    /// </summary>
+    public static readonly CVarDef<bool> ICSSDSleep =
+        CVarDef.Create("ic.ssd_sleep", true, CVar.SERVER);
+
+    /// <summary>
+    ///     Time between character getting SSD status and falling asleep
+    ///     Won't work without ICSSDSleep
+    /// </summary>
+    public static readonly CVarDef<float> ICSSDSleepTime =
+        CVarDef.Create("ic.ssd_sleep_time", 600f, CVar.SERVER);
 }
--- a/Content.Shared/CCVar/CCVars.Ic.cs
+++ b/Content.Shared/CCVar/CCVars.Ic.cs
@@ -47,7 +47,7 @@ public sealed partial class CCVars
         CVarDef.Create("ic.show_ssd_indicator", true, CVar.CLIENTONLY);
 
     /// <summary>
-    ///     Forces SSD characters to sleep after
+    ///     Forces SSD characters to sleep after ICSSDSleepTime seconds
     /// </summary>
     public static readonly CVarDef<bool> ICSSDSleep =
         CVarDef.Create("ic.ssd_sleep", true, CVar.SERVER);
--- a/Content.Shared/SSDIndicator/SSDIndicatorComponent.cs
+++ b/Content.Shared/SSDIndicator/SSDIndicatorComponent.cs
@@ -9,7 +9,7 @@ namespace Content.Shared.SSDIndicator;
 ///     Shows status icon when player in SSD
 /// </summary>
 [RegisterComponent, NetworkedComponent]
-[AutoGenerateComponentState]
+[AutoGenerateComponentState, AutoGenerateComponentPause]
 public sealed partial class SSDIndicatorComponent : Component
 {
     [ViewVariables(VVAccess.ReadWrite)]
@@ -23,6 +23,13 @@ public sealed partial class SSDIndicatorComponent : Component
     /// <summary>
     ///     When the entity should fall asleep
     /// </summary>
-    [DataField, Access(typeof(SSDIndicatorSystem))]
+    [DataField, AutoPausedField, Access(typeof(SSDIndicatorSystem))]
     public TimeSpan FallAsleepTime = TimeSpan.Zero;
+
+    /// <summary>
+    ///     Required to don't remove forced sleep from other sources
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite)]
+    [AutoNetworkedField]
+    public bool ForcedSleepAdded = false;
 }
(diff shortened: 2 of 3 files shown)


Content.Shared/SSDIndicator/SSDIndicatorComponent.cs
```diff
@@ -19,4 +19,10 @@ public sealed partial class SSDIndicatorComponent : Component
     [ViewVariables(VVAccess.ReadWrite)]
     [DataField]
     public ProtoId<SsdIconPrototype> Icon = "SSDIcon";
+
+    /// <summary>
+    ///     When the entity should fall asleep
+    /// </summary>
+    [DataField, Access(typeof(SSDIndicatorSystem))]
+    public TimeSpan FallAsleepTime = TimeSpan.Zero;
```
Tayrtahn, line 27: Autopause

Hrosts:
The Shift+Click at 0:14 in the video doesn't show the yellow text on the character, which tells that the person has disconnected. Will it be seen in a normal game?

Content.Shared/SSDIndicator/SSDIndicatorSystem.cs
```diff
@@ -16,12 +23,46 @@ public override void Initialize()
     private void OnPlayerAttached(EntityUid uid, SSDIndicatorComponent component, PlayerAttachedEvent args)
     {
         component.IsSSD = false;
+
+        // Removes force sleep and resets the time to zero
+        if (_cfg.GetCVar(CCVars.ICSSDSleep))
+        {
+            component.FallAsleepTime = TimeSpan.Zero;
+            EntityManager.RemoveComponent<ForcedSleepingComponent>(uid);
```
Tayrtahn, line 31: Won't this remove ForcedSleepingComponents from other sources too?

lzk228:
> The Shift+Click at 0:14 in the video doesn't show the yellow text on the character, which tells that the person has disconnected. Will it be visible in a normal game?

yes, in video it is so, because i aghosted, tho the character should not became ssd, as it was before this issue https://github.com/space-wizards/space-station-14/issues/25623

SlamBamActionman:
This PR has conceptual approval as it is based on a maintainer decision from the previous PR linked in the description. As such it does not need to undergo a maintainer discussion; code review and 2 approvals are still required, however.

Content.Shared/SSDIndicator/SSDIndicatorSystem.cs
lzk228: well damn, firstly i wanted to do that but thought there will be a problem that cannot even happen
yeah, gonna fix that

NoElkaTheGod:
Free revenant kill

ScarKy0:
> Free revenant kill

For the better tbh...

NoElkaTheGod:
> For the better tbh... Reventant already is hella unfun since all bodies are always guarded

True, but i don't think that this will make revenant any more fun. People will just be encouraged to bring everyone to cryosleep even more.

ScarKy0:
> True, but i don't think that this will make revenant any more fun. People will just be encouraged to bring everyone to cryosleep even more.

Rev needs to be able to eat animal souls like in ss13 but thats a discussion for another place :godo:

Content.Shared/SSDIndicator/SSDIndicatorSystem.cs
```diff
@@ -16,12 +23,52 @@ public override void Initialize()
...
+
+        var query = EntityQueryEnumerator<SSDIndicatorComponent>();
+
+        while (query.MoveNext(out var uid, out var ssd))
+        {
+            // Forces the entity to sleep when the time has come
+            if(ssd.IsSSD &&
+                ssd.FallAsleepTime <= _timing.CurTime &&
+                !TerminatingOrDeleted(uid) &&
+                !TryComp<ForcedSleepingComponent>(uid, out _)) // Don't add the component if the entity has it from another sources
```
Tayrtahn, line 67: Use HasComp instead of TryComp with a discard.

Content.Shared/CCVar/CCVars.Ic.cs
```diff
@@ -45,4 +45,17 @@ public sealed partial class CCVars
     /// </summary>
     public static readonly CVarDef<bool> ICShowSSDIndicator =
         CVarDef.Create("ic.show_ssd_indicator", true, CVar.CLIENTONLY);
+
+    /// <summary>
+    ///     Forces SSD characters to sleep after
```
Tayrtahn, line 50: After...?

Tayrtahn (commented):
A couple minor things. 

Also, I'm going to make some changes to the test that's failing to try to get some more useful info out of it.

Tayrtahn:
So, with some improvements to the test (#34069), I can now see that MobHumanSyndicateAgentMedic fails the test because the loadout makes it spawn with the combat medical kit in-hand. When the ForcedSleepingComponent is added, the medic falls over and drops any in-hand items. The medic mob is then deleted, but it leaves behind the medical kit and its contents, which counts as leaked entities and fails the test.

Kinda funny, but a bit inconvenient. Gonna have to think about the best way to resolve this.

lzk228:
maybe i can fix it here? because as it  maybe can be seen on the video, the urists are sleeping as i join the server, and i see why
they don't go through the PlayerDetached method, so their sleeptime is always zero and on update it makes them sleep
i just need to add some kind of OnMapInit or something so they would go through sleeptime setting

lzk228:
added the method i described above
updated the media part in pr desc

Tayrtahn (approved):
Excellent solution! This was a fun one to work out together.

lzk228:
it happened

LaCumbiaDelCoronavirus:
Is this intended to affect AI NPCs too?

lzk228:
yes

PR #29320, opened 2024-06-21
refactor event schedulers to use explicit game rules

IProduceWidgets:
## About the PR

Refactors basic & ramping event scheduler to take explicit event definitions in yml instead of just doing whatever the hell they wanted to. Also ported meteor scheduler over to the other scheduler types with more variety.

Added two (really just tweaked) gamemodes as rare chances in secret to flex the scheduler utility a bit.

## Why / Balance

Shuttle events, gift events, the desire to add things that shouldn't be in event schedulers.
I cannot overstate how many levers this adds to event balancing by using combinations of multiple schedulers, table selectors & station event weights.

## Technical details

Now uses EntityTableSelectors. Also, since it allows for multiple uses of the different event schedulers I removed the ccvars for meteors and ramping schedulers and placed them on the components.

## Media

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

The cvars for ramping event scheduler severity and for meteor timing have been removed, and moved to fields on their respective components.
If you have events you want run via the basic/ramping event scheduler, they need to be defined explicitly now.

**Changelog**

:cl:
- tweak: shuttle events have been separated from other midround events and can now happen concurrently.
- tweak: Cargo Gift events should be less likely.
- tweak: meteors have been changed to be more responsive to player count.
- add: Game mode: Kessler syndrome - What if survival but also the apocalypse?
- add: Game mode: Zombeteors - Zombies and a meteor shower all at once!

PR #34051, opened 2024-12-24
Remove kessler and zombeteors gamemodes from the secret pool

ArtisticRoomba:
## About the PR
This PR removes Kesser and Zombeteors from the secret pool, tuning the Survival chance to match the void.

Zombeteors, for those who don't know, is just the Kessler gamemode with an added Zombies gamerule.

## Why / Balance
For context, Kessler and Zombeteors were implemented along with the refactoring of event schedulers to use explicit game rules (PR #29320). The reasoning seen in that PR for adding them was moreso a "showing off" of what the refactor is capable of.

Kessler is unfortunately just an annoying meteor repair simulator. Basically:
- Round progression is very bland. Engineering will just keep repairing ever-increasing waves of meteors until the situation reaches critical mass and the amount of incoming damage outweighs the amount of damage Engineering can repair, and then evac gets called.
- Meteors have this amazing problem where they target structures that have a large surface area (stick out from the station) which usually means Evac or Arrivals. Arrivals shouldn't be targeted (this is a problem for another PR but alas) and usually both structures are made entirely out of glass, which is annoying to repair and takes a while, which is plenty of time for another round to come in before you're done.
- Engineering can do nothing to prevent meteors from coming in, they can only repair the damage once it's done. This leads to a very annoying feeling of fighting against entropy which isn't fun. This is also annoying considering in most other gamemodes there's a real, physical antagonist that the crew can fight against. Engineering can't really fight against meteors other than repairing the constant stream of damage that comes in.
- This affects lowpop engineering to an annoying degree.

I'm more inclined to believe that Kessler and Zombeteors was snuck in as part of the rework, with the benefits not properly evaluated:

- What roleplay does this add to the game?
- What can the crew do to combat this?
- Is this fun for other departments, or just one?

In terms of core design principles,
- Meteors aren't really unique chaos. The same thing happens over and over again. Evac gets hit -> repair, arrivals gets hit -> repair, repeat with other locations over and over again until you can't keep up anymore and call evac. _"Mechanics that [guarantee] a chaotic situation will always occur should be avoid[ed] since this negatively impacts player agency."_
- _"Part of the unique experience of Spacestation 14 is interacting with its indepth simulation gameplay while dealing with the chaos that comes from antagonists and random events."_ While I do agree that meteor damage leads to interaction with the many systems this game has to offer (atmos simulation, construction), the negatives outweigh the benefits.

If I want to cryo if I knew it's Kessler or Zombeteors, then the gamemode shouldn't be included in the pool.

In my opinion, the idea of Kessler would be much more suited if Engineering had a real, potentially station-wide system they could deploy over the course of a round that could intercept incoming meteors, as a form of reactionary measure.

Lacking this, it would be better to not have Kessler (and Zombeteors, as a component of that).

## Technical details
Removed the Zombeteors and Kessler roll chances from the secret gamemode table, increased Survival to compensate, cause that's what should be rolled instead. The gamemodes were kept in the files to enable admin torture.

## Requirements

**Changelog**
:cl:
- remove: Kessler Syndrome and Zombeteors are removed from the possible gamemodes in the Secret pool.

### DIFF
--- a/Resources/Prototypes/secret_weights.yml
+++ b/Resources/Prototypes/secret_weights.yml
@@ -3,6 +3,6 @@
   weights:
     Nukeops: 0.20
     Traitor: 0.60
-    Zombie: 0.04
-    Survival: 0.11
+    Zombie: 0.05
+    Survival: 0.10
     Revolutionary: 0.05
--- a/Resources/Prototypes/secret_weights.yml
+++ b/Resources/Prototypes/secret_weights.yml
@@ -4,7 +4,5 @@
     Nukeops: 0.20
     Traitor: 0.60
     Zombie: 0.04
-    Zombieteors: 0.01
-    Survival: 0.09
-    KesslerSyndrome: 0.01
+    Survival: 0.11
     Revolutionary: 0.05
--- a/Resources/Prototypes/game_presets.yml
+++ b/Resources/Prototypes/game_presets.yml
@@ -13,6 +13,21 @@
     - BasicRoundstartVariation
 
 - type: gamePreset
+  id: KesslerSyndrome
+  alias:
+    - kessler
+    - junk
+    - meteorhell
+  name: kessler-syndrome-title
+  showInVote: false # secret
+  description: kessler-syndrome-description
+  rules:
+    - KesslerSyndromeScheduler
+    - RampingStationEventScheduler
+    - SpaceTrafficControlEventScheduler
+    - BasicRoundstartVariation
+
+- type: gamePreset
   id: AllAtOnce
   alias:
   - all
@@ -201,5 +216,21 @@
   - Zombie
   - BasicStationEventScheduler
   - MeteorSwarmScheduler
+  - SpaceTrafficControlEventScheduler
+  - BasicRoundstartVariation
+
+- type: gamePreset
+  id: Zombieteors
+  alias:
+  - zombieteors
+  - zombombies
+  - meteombies
+  name: zombieteors-title
+  description: zombieteors-description
+  showInVote: false
+  rules:
+  - Zombie
+  - BasicStationEventScheduler
+  - KesslerSyndromeScheduler
   - SpaceTrafficControlEventScheduler
   - BasicRoundstartVariation
--- a/Resources/Prototypes/game_presets.yml
+++ b/Resources/Prototypes/game_presets.yml
@@ -13,21 +13,6 @@
     - BasicRoundstartVariation
 
 - type: gamePreset
-  id: KesslerSyndrome
-  alias:
-    - kessler
-    - junk
-    - meteorhell
-  name: kessler-syndrome-title
-  showInVote: false # secret
-  description: kessler-syndrome-description
-  rules:
-    - KesslerSyndromeScheduler
-    - RampingStationEventScheduler
-    - SpaceTrafficControlEventScheduler
-    - BasicRoundstartVariation
-
-- type: gamePreset
   id: AllAtOnce
   alias:
   - all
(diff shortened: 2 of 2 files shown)


august-sun:
Mentioned in the Discord, still would like to ddoc this because I think it could be cool if done right, but yeah, it disrupts roundflow significantly in a bad way at the moment.

august-sun:
> > Mentioned in the Discord, still would like to ddoc this because I think it could be cool if done right, but yeah, it disrupts roundflow significantly in a bad way at the moment.
> 
> I disagree because there's not much mechanics for the rest of the crew to help out or survive

The crew is able to construct and repair damage the station takes. It won't be as efficient as full engineers, but you can, as a crew member, absolutely fix your own workspaces when supplied with the materials to do so.

cohanna:
would not miss Kessler (least fun rounds by far IMO), but Zombmeteors is hilarious. Its like a mini admeme, and is rare enough to not be disruptive.

IProduceWidgets:
As the OP, I agree the kessler gamerule is a little lackluster alone. It needs more meteor related mechanics we unfortunately missed out on due to some staff departures.

I don't think the zombie one is bad though. The round sort of needs to end after zombies regardless. I could see adding the basic station events into zombieteros to add more stuff for the first half of shift however.

Resources/Prototypes/secret_weights.yml
```diff
@@ -4,7 +4,5 @@
     Nukeops: 0.20
     Traitor: 0.60
     Zombie: 0.04
-    Zombieteors: 0.01
-    Survival: 0.09
-    KesslerSyndrome: 0.01
+    Survival: 0.11
```
slarticodefast, line 7:
```suggestion
    Survival: 0.10
```

Resources/Prototypes/secret_weights.yml
```diff
@@ -4,7 +4,5 @@
     Nukeops: 0.20
     Traitor: 0.60
     Zombie: 0.04
```
slarticodefast, line 6:
```suggestion
    Zombie: 0.05
```
At least divide up the chance nicely. A 4 and 11% chance is annoying me way too much 😄

slarticodefast (changes requested):

Resources/Prototypes/secret_weights.yml
ArtisticRoomba: Increasing zombie chance??? Evil man, you are...
Pushed requested changes.

slarticodefast: The 1 additional percent of zombies was already in zombieteors, so this keeps is the same actually.

slarticodefast (approved):
Approval from my side. I brought it up for maintainer discussion and maybe we want to add a very small chance for combinations of other gamemodes (or even AllAtOnce if we feel particularly evil) to keep some variety. I'll get back to you on that.

ArtisticRoomba:
Thanks. Extremely rare gamemodes like AllAtOnce or perhaps a new one called Thievery where it's just all thieves could be something funny for security to deal with.

IProduceWidgets:
> Thanks. Extremely rare gamemodes like AllAtOnce or perhaps a new one called Thievery where it's just all thieves could be something funny for security to deal with.

I've run some oops all thieves rounds before, where everyone but command and sec are thieves.
They're fun, but they need to be really uncommon. 
It takes about 15 minutes for everyone to figure out everyone's a thief and then the game becomes competitive scavenger hunting. 
Kinda sucks as command though because every department just stops functioning.

ArtisticRoomba:
Yeah, although keeping this restricted to a 1% chance is generally how this should be. Would get annoying if it felt like it was thieves/allatonce back to back.

Having it as a low chance makes for a nice twist on standard roundflow and gameplay. It can always be tuned later if it shows itself to be too annoying.

cohanna:
> Kinda sucks as command though because every department just stops functioning.

That's when you awaken command as sleepers with only kill targets :godo:

PR #34056, opened 2024-12-24
Refactor light mask prototypes for compatibility with illumination measuring system of RobustToolbox

Chubbygummibear:
## About the PR
light masks for PointLightComponents are expanded to be their own prototype so that in addition to the texture which will be applied to the light, the specifications of those cones, such as angle of light being cast, and the number of cones, is stored in that prototype. This is so that while the client overlays the texture for rendering, the server has usable data for calculating exposure to those lights which would otherwise be treated as a full circle, rather than a cone.

## Why / Balance
The LightSensitiveSystem I created for RobustToolbox needs usable numbers in the form of light cone specifics to accompany textures, because the server can't utilize a texture for calculating what entities are and aren't in a light cone being pointed around the game world.

Important to note that the data which accompanies each texture is still eyeballed by me. So if it feels like maybe you're in or out of the light and not getting the right exposure level, it's because there's no way to actually read in the texture to the server to get an exact shape for it 😔 

## Technical details
Requires https://github.com/space-wizards/RobustToolbox/pull/5574

## Media

https://github.com/user-attachments/assets/82d3049a-50ed-4ef4-9f24-6bb890c956c1

https://github.com/user-attachments/assets/062c71d0-1b07-4475-b4e0-2e4b13f9fefe

## Requirements


Chubbygummibear:
i should probably close this and make a new pr later since the sibling pr for this on the engine has changed substantially, but i'm still using this branch/pr to verify it works

PR #34062, opened 2024-12-25
bagel update

Emisse:
changelog
fix a lot of button sillies/add borders to buttons so theyre clear what they do
add 2 baby maints to arrivals so if main door is blocked for whatever reason theres other path
move chaplain office right door down 1 so its in line with the other doors and stops making me mad
rewire the new secondary smes input so it doesnt go through a wall but goes under a window instead
skub


ArtisticRoomba:
i would never commit mapping sins while complaining about mappers committing mapping sins, all while leaving a maptainer to fix my sloppy work

PR #34072, opened 2024-12-26
Arrivals blacklist for bluespace lockers and QSIs

crazybrain23:
## About the PR

Added the arrivals blacklist to bluespace lockers and quantum spin inverters.

## Why / Balance

Both bluespace lockers and quantum spin inverters could and have been used to get back to the arrivals terminal. This is a problem because players have recently been building "colonies" out on the arrivals planet map.
This change prevents the method commonly used to get to the terminal (bluespace lockers) and another method I knew of (quantum spin inverters).

## Technical details

In the bluespace locker event, just after the lockers are configured, I made it ensure the locker has the arrivals blacklist component. As for the quantum spin inverter, I just added the component in the yaml.

## Media

image
image
(QSI worked in exactly the same way so screenshots are pointless)

## Requirements

~~## Breaking changes~~

**Changelog**

:cl: crazybrain
- tweak: Bluespace lockers and quantum spin inverters can no longer go on the arrivals shuttle.

### DIFF
--- a/Content.Server/StationEvents/Events/BluespaceLockerRule.cs
+++ b/Content.Server/StationEvents/Events/BluespaceLockerRule.cs
@@ -47,6 +47,8 @@ public sealed class BluespaceLockerRule : StationEventSystem<BluespaceLockerRule
 
             Sawmill.Info($"Converted {ToPrettyString(potentialLink)} to bluespace locker");
 
+            EnsureComp<ArrivalsBlacklistComponent>(potentialLink); // To stop people getting to arrivals terminal
+
             return;
         }
     }
--- a/Content.Server/StationEvents/Events/BluespaceLockerRule.cs
+++ b/Content.Server/StationEvents/Events/BluespaceLockerRule.cs
@@ -6,7 +6,6 @@ using Content.Shared.Access.Components;
 using Content.Shared.Station.Components;
 ﻿using Content.Shared.GameTicking.Components;
 using Content.Shared.Coordinates;
-using Content.Server.Shuttles.Components;
 
 namespace Content.Server.StationEvents.Events;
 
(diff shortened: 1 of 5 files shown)


Content.Server/StationEvents/Events/BluespaceLockerRule.cs
```diff
@@ -45,6 +46,8 @@ protected override void Started(EntityUid uid, BluespaceLockerRuleComponent comp
             _bluespaceLocker.GetTarget(potentialLink, comp, true);
             _bluespaceLocker.BluespaceEffect(potentialLink, comp, comp, true);
 
+            EnsureComp<ArrivalsBlacklistComponent>(potentialLink); // To stop people getting to arrivals terminal
```
slarticodefast, line 49: Wouldn't this only blacklist one of the two lockers since the other one is assigned in `GetTarget`?
This should probably be moved to the `OnStartup` function for the `BluespaceLockerComponent` so that all bluespace lockers are arrival shuttle immune.

slarticodefast:
Also storing the QSI inside a storage container will make it teleport to the center of the shuttle instead of staying at the position of the container. But that can be fixed in a separate PR.

Content.Server/StationEvents/Events/BluespaceLockerRule.cs
crazybrain23: You were right there, good catch. One though I had for not putting it in `OnStartup` was that to allow a bluespace locker on arrivals, you'd have to manually remove the component, but I now reaslise that's not really an issue.
image
image

crazybrain23: > This should probably be moved to the OnStartup function for the BluespaceLockerComponent so that all bluespace lockers are arrival shuttle immune.

Yeah so now tests fail because it gains a component on spawn.

slarticodefast (approved):
I didn't even know the integration tests don't like prototypes gaining components on spawn. But adding it to the prototype beforehand seems to work and keeping it in the startup subscription ensures that all bluespace lockers are arrivals immune, so it looks good to me.

TheShuEd (approved):

crazybrain23:
> I didn't even know the integration tests don't like prototypes gaining components on spawn. 

My guess would be that emaintainers want the components explicitly listed, as opposed to implicitly gained from another component.

PR #34077, opened 2024-12-26
Add PreventFTLComponent and changed FTL check

Betterwithketchup:
## About the PR

Added PreventFTLComponent  and changed the way grids check if they can FTL.

## Why / Balance

There wasn't a way before to have a shuttle that wasn't able to FTL, but still pilotable. With this change, now there is.

## Technical details

Added PreventFTLCompnent.cs to shuttle components and changed the check in ShuttleSystem.FasterThanLight.cs to include it.

## Requirements

## Breaking changes

**Changelog**


IProduceWidgets:
While this does work, I cant help but feel its a bandaid for not having an FTL drive of some sort and all grids just magically being able to FTL because they have a console.

MilonPL (approved):
Looks good, the lack of feedback to the player is my only nitpick, but that's out of scope for this probably, since we're not using the result of CanFTL apparently. Could also change the UI to just grey out the button. But I assume this isn't going to be used that often outside of maybe admin events, so I'd say that's fine, and should be done as a part of a larger rework.

PR #34078, opened 2024-12-26
Removed Power Monitoring Computer boards from research and lathe recipes. Added default engineering sprites for atmos computer boards.

Alpaccalypse:
## About the PR
Set the default sprite for Atmospheric Alerts Computer boards and Atmospheric Network Monitoring boards to `cpu_engineering`. The 'power monitoring computer board' has been removed from industrial research and lathe recipes.

## Why / Balance
The atmos boards defaulted to the base green sprites. The power monitoring computer can be resource intensive when used, and was not intended to be able to be researched and printed, due to this concern.

## Technical details
A few lines added to and removed from the relevant industrial/lathe/computer/electronics .yml files.

## Media
The boards are now yellow, screenshot for example (they are not actually printable):
{CA3B86A2-B175-40D8-97D9-0415489C8A00}

## Requirements

**Changelog**
:cl:
- remove: Power monitoring computer boards can no longer be researched or printed, as originally intended.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/Circuitboards/computer.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/Circuitboards/computer.yml
@@ -24,15 +24,19 @@
   name: atmospheric alerts computer board
   description: A computer printed circuit board for an atmospheric alerts computer.
   components:
+    - type: Sprite
+      state: cpu_engineering
     - type: ComputerBoard
       prototype: ComputerAlert
-      
+
 - type: entity
   parent: BaseComputerCircuitboard
   id: AtmosMonitoringComputerCircuitboard
   name: atmospheric network monitor board
   description: A computer printed circuit board for an atmospheric network monitor.
   components:
+    - type: Sprite
+      state: cpu_engineering
     - type: ComputerBoard
       prototype: ComputerAtmosMonitoring
 
--- a/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
@@ -474,7 +474,6 @@
       - SolarControlComputerCircuitboard
       - SolarTrackerElectronics
       - TurboItemRechargerCircuitboard
-      - PowerComputerCircuitboard
       - AutolatheHyperConvectionMachineCircuitboard
       - ProtolatheHyperConvectionMachineCircuitboard
       - CircuitImprinterHyperConvectionMachineCircuitboard
--- a/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
@@ -521,6 +521,8 @@
       - ReagentGrinderIndustrialMachineCircuitboard
       - JukeboxCircuitBoard
       - SMESAdvancedMachineCircuitboard
+      - AlertsComputerCircuitboard
+      - AtmosMonitoringComputerCircuitboard
   - type: EmagLatheRecipes
     emagDynamicRecipes:
       - ShuttleGunDusterCircuitboard
--- a/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
@@ -520,8 +520,6 @@
       - ReagentGrinderIndustrialMachineCircuitboard
       - JukeboxCircuitBoard
       - SMESAdvancedMachineCircuitboard
-      - AlertsComputerCircuitboard
-      - AtmosMonitoringComputerCircuitboard
   - type: EmagLatheRecipes
     emagDynamicRecipes:
       - ShuttleGunDusterCircuitboard
--- a/Resources/Prototypes/Recipes/Lathes/electronics.yml
+++ b/Resources/Prototypes/Recipes/Lathes/electronics.yml
@@ -431,11 +431,6 @@
 
 - type: latheRecipe
   parent: BaseCircuitboardRecipe
-  id: PowerComputerCircuitboard
-  result: PowerComputerCircuitboard
-
-- type: latheRecipe
-  parent: BaseCircuitboardRecipe
   id: CloningConsoleComputerCircuitboard
   result: CloningConsoleComputerCircuitboard
 
--- a/Resources/Prototypes/Recipes/Lathes/electronics.yml
+++ b/Resources/Prototypes/Recipes/Lathes/electronics.yml
@@ -609,3 +609,13 @@
   parent: BaseCircuitboardRecipe
   id: CutterMachineCircuitboard
   result: CutterMachineCircuitboard
+
+- type: latheRecipe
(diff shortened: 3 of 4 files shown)


slarticodefast:
I'm pretty sure these are not printable on purpose because they are very networking expensive (same as the power monitoring console).

Alpaccalypse:
> I'm pretty sure these are not printable on purpose because they are very networking expensive (same as the power monitoring console).

Ahh, I thought there might have been a reason, though the power monitoring can be printed, which is why I went ahead and added these too. I acutally looked into this due to a bug I haven't been able to replicate, where I dismantled the alerts console to find the name of the board forsci to print; it didn't drop a board, so couldn't be rebuilt.

The reason I've persoanlly wanted these, was to build an all-in-one monitoring station in atmospherics, on maps where there was a lot of walking between the different alerts/monitoring stations in engineering. Hopefully there would rarely be a call to have more than one of these built as extras, though that wouldn't stop someone spamming them just for server grief I suppose, if that is an issue.

It did occur to me that they could be given a slightly higher material cost to print, than the default, as a "specialist" board.

slarticodefast:
Let me ping @metalgearsloth to get some info about the performance aspect. Not sure how problematic someone spamming them would be. The consoles can only be used by one player at a time for the same reason by the way.

chromiumboy:
> > I'm pretty sure these are not printable on purpose because they are very networking expensive (same as the power monitoring console).
> 
> Ahh, I thought there might have been a reason, though the power monitoring can be printed, which is why I went ahead and added these too. I acutally looked into this due to a bug I haven't been able to replicate, where I dismantled the alerts console to find the name of the board forsci to print; it didn't drop a board, so couldn't be rebuilt.
> 
> The reason I've persoanlly wanted these, was to build an all-in-one monitoring station in atmospherics, on maps where there was a lot of walking between the different alerts/monitoring stations in engineering. Hopefully there would rarely be a call to have more than one of these built as extras, though that wouldn't stop someone spamming them just for server grief I suppose, if that is an issue.
> 
> It did occur to me that they could be given a slightly higher material cost to print, than the default, as a "specialist" board.

If power monitoring circuit boards can be printed, that's a issue, and that recipe will need to be removed from the circuitboard printer. The atmos computers have lower network and server computational costs than power monitoring but they probably shouldn't be printable either, TBH. Destroyed consoles should always drop their respective circuitboards though, so if thats not happening it's a bug that needs to be fixed

slarticodefast:
Alright, thanks for the info, chrome. @Alpaccalypse We can still merge the sprite change, but could you remove the board lathe recipes and their research, including the power monitoring board?

Alpaccalypse:
Sure. It also turns out, I can replicate that "not dropping boards from deconstructed computers" bug on live servers, just doesn't seem to hapen with latest build I was testing on, so maybe it has been fixed? I will also remove the power monitoring boards from research and lathes, if that is desired by consensus?

Alpaccalypse:
Latest version should now only add sprites for the atmos boards, and also removes Power Monitoring Computer board from research and the lathe menu. WAI when tested.

Note: Removing the line for researching Power board also included the comment `#the actual solar panel itself should be in here` and I am not sure what that was referring to, is it somehow important for researching and building solar panels?

slarticodefast (approved):
Looks good to me, thank you for your first contribution!

TheShuEd (approved):

SpaceRox1244:
Wait a minute, isnt there already a precaution to reduce the load on the server caused by the power monitoring console, since only one person at a time is allowed to use one? Last time I was playing engineering, if someone was using the console, everybody else was locked out of it.

Alpaccalypse:
> Wait a minute, isnt there already a precaution to reduce the load on the server caused by the power monitoring console, since only one person at a time is allowed to use one? Last time I was playing engineering, if someone was using the console, everybody else was locked out of it.

I think the concern was, if the boards were printable, multiple consoles could be built to get around this. Probably not a major concern for one or two, but if with no limits on number of board printed, it could become a problem if enough people worked together to grief.

chromiumboy:
> > Wait a minute, isnt there already a precaution to reduce the load on the server caused by the power monitoring console, since only one person at a time is allowed to use one? Last time I was playing engineering, if someone was using the console, everybody else was locked out of it.
> 
> I think the concern was, if the boards were printable, multiple consoles could be built to get around this. Probably not a major concern for one or two, but if with no limits on number of board printed, it could become a problem if enough people worked together to grief.

Yup pretty much this. I need to revisit the code someday and see if I can improve its performance further

SpaceRox1244:
> > Wait a minute, isnt there already a precaution to reduce the load on the server caused by the power monitoring console, since only one person at a time is allowed to use one? Last time I was playing engineering, if someone was using the console, everybody else was locked out of it.
> 
> I think the concern was, if the boards were printable, multiple consoles could be built to get around this. Probably not a major concern for one or two, but if with no limits on number of board printed, it could become a problem if enough people worked together to grief.
> 
> I've got a half-baked idea on a way to allow boards to be printed but also address this potential problem, I imagine it might take some careful consideration before passing review though.

You mean if a significant portion of active players decided to all do that at the same time and nobody did anything about them, players or admins? Are these edge cases really that worth it? :?

SpaceRox1244:
Don't people get really frustrated when design decisions are made to mitigate edge case griefing but cause the majority to have less mechanics to play with? Don't motivations like that, which could be difficult on a technical level to understand as a player, create more misunderstanding and distrust between dev and community, thus pushing some to hostile action, especially considering community spaces made to stew hostility about this precautionary philosophy? I kinda think the better road here is just unclenching about it. I know it sounds like conceding, but you dont know the people youre anticipating in your head and they dont know you. Either only knows what theyre told the other thinks of them. We cant feasibly police everything.

Aisu9:
> Don't people get really frustrated when design decisions are made to mitigate edge case griefing but cause the majority to have less mechanics to play with? Don't motivations like that, which could be difficult on a technical level to understand as a player, create more misunderstanding and distrust between dev and community, thus pushing some to hostile action, especially considering community spaces made to stew hostility about this precautionary philosophy? I kinda think the better road here is just unclenching about it. I know it sounds like conceding, but you dont know the people youre anticipating in your head and they dont know you. Either only knows what theyre told the other thinks of them. We cant feasibly police everything.

get use to it, the end goal of wizden is to be griefing-proof no matter the cost

Alpaccalypse:
I'd like to see an in-game way to solve the problem, in future, though let's keep things in perspective; this only came up because I wanted to color the boards in question, and I am literally the only person I've heard of even trying to legitimately build a new atmos monitor. I was actually hoping to add atmos boards to printing, not remove power ones.

If I'm honest, I wouldn't have mentioned it if I'd known the boards would get removed, but I wouldn't categorise it as greifing-proof no matter the cost, considering how few people are even going to know this change even happened without reading the changelog.

PJB3005:
> get use to it, the end goal of wizden is to be griefing-proof no matter the cost

@Aisu9 This is a **formal warning** that this behavior will not be tolerated on our repo. You will be banned from the repo if you post disrespectful comments like this again.

Issue #33965, opened 2024-12-19
Clothing HideLayerComponent breaks on clothing with multiple equipment slots


paige404:
## Description

Clothing with a HideLayerComponent will fail to properly hide the designated layers if the clothing being equipped has multiple potential equip slots.

I initially reported this [here](https://github.com/RMC-14/RMC-14/issues/5173) on RMC as their gas mask has both Mask and SuitStorage equip slots, so that they can be attached to a hazard vest. We tracked the issue down to [Content.Shared/Clothing/EntitySystems/ClothingSystem.cs line 104](https://github.com/space-wizards/space-station-14/blob/0462993153200e654c5da47ecaacf18835b295b8/Content.Shared/Clothing/EntitySystems/ClothingSystem.cs#L104). I don't know if this currently effects any items on WizDen, but it definitely impacts items on other forks and it would be helpful to sort it out here. 

**Reproduction**

1. Add an extra equipment slot to any piece of clothing with a HideLayerComponent
2. Build, run, and put on the piece of clothing in either of the allowed spots
3. The layer will not be hidden.

**Screenshots**

**Additional context**

It may be tempting to simply change this line to see if the clothing item's slots list contains the relevant slot for that layer. However, this would result in a snout layer being hidden when a mask is equipped on the SuitStorage slot, for instance.

The solution here will need to check that specifically the slot the item is equipped in applies to that layer.

paige404:
> Is this fixed now that the PR is merged

Yeah, this looks good.

PR #34080, opened 2024-12-26
Fix broken layer hiding on clothes with multiple equipment slots

paige404:
## About the PR

Adjusted ClothingSystem logic to confirm that a slot to be hidden by a HideLayerComponent matches the slot that an item is equipped in.

Before PR: ClothingSystem compared the sum combination of slots an item can be equipped in to the slot the HideLayerComponent's slot. This would always be false for clothing with more than one equip slot.

After PR: ClothingSystem specifically checks for a match between the item's current equipped slot and the HideLayerComponent's slot.

## Why / Balance

Addresses issue #33965

## Technical details

I haven't noticed any problems on local testing, but this logic _is_ somewhat less performant than the old logic. I had initially hoped to use `HasFlag` to keep it simple, but this resulted in goofy things like hiding the snout layer when equipping a mask to suit storage.

I don't think it's going to cause issues at scale, but if anyone has some clever bitshift magic with the slotflags that could do this more performantly I think that would be ideal solution.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
+++ b/Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
@@ -101,7 +101,8 @@ public abstract class ClothingSystem : EntitySystem
                 {
                     if (comp.Slots.Contains(layer))
                     {
-                        if (TryComp(item, out ClothingComponent? clothing) && clothing.Slots == slot.SlotFlags)
+                        if (TryComp(item, out ClothingComponent? clothing) && clothing.InSlot != null
+                                                                           && clothing.InSlot.Equals(slot.Name))
                         {
                             //Checks for mask toggling. TODO: Make a generic system for this
                             if (comp.HideOnToggle && TryComp(item, out MaskComponent? mask))


Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
```diff
@@ -101,7 +101,8 @@ private void ToggleVisualLayers(EntityUid equipee, HashSet<HumanoidVisualLayers>
                 {
                     if (comp.Slots.Contains(layer))
                     {
-                        if (TryComp(item, out ClothingComponent? clothing) && clothing.Slots == slot.SlotFlags)
+                        if (TryComp(item, out ClothingComponent? clothing) && clothing.InSlot != null
+                                                                           && clothing.InSlot.Equals(slot.Name))
                         {
```
ElectroJr, line 106: I don't think this is the right way to fix this, and AFAIK `clothing.InSlot.Equals(slot.Name)` should always be true here? I think this likely should just be a has-flag check?

Though clothing & visuals also seem to generally have shitty documentation, but IIRC `ClothingComponent.Slots` is meant to define the slots in which a piece of clothing is considered "worn" i.e., magboots in pockets are clothing in a slot, but not the right kind of slot.

But if you want even more control about what layers clothing will hide when in certain slots, then `HideLayerClothingComponent` probably either needs to have its own set of slot-flags to decide whether it should be enabled or not, or for even more control the hidden layers should be specified per-slot by slot(`i.e., replace `HashSet<HumanoidVisualLayers>` with a `Dictionary<SlotFlags, HashSet<HumanoidVisualLayers>>`).

In general ToggleVisualLayers() seems pretty cursed / convoluted and hard to parse and should probably be re-written.

ElectroJr (changes requested):

Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
paige404: Nope. I tried HasFlag, but you can also try it if you like. Use `clothing.Slots.HasFlag(slot.SlotFlags)` in that section I replaced instead. Add suitstorage to a mask's slots, play a lizard, equip the mask in your suit storage, and watch your nose disappear despite the mask not being on your face. It's because HasFlag is checking whether the piece _can_ cover that layer, not whether it _is_ covering that layer.

`ClothingComponent.Slots` is the sum of the SlotFlags of all its equippable slots. `ClothingComponent.InSlot` is the name of the slot it is currently equipped in.

So with the example of RMC ([where I discovered this issue](https://github.com/RMC-14/RMC-14/issues/5173)), the gas mask can go onto the suit storage slot (for... CM13 parity reasons I guess). `ClothingComponent.Slots` will always contain Mask and Suit Storage SlotFlags. But `ClothingComponent.InSlot` will only be Mask if it's currently in the Mask slot, and only be SuitStorage when it's currently in the SuitStorage slot.

A cleaner way might be to add a SlotFlag equivalent to `ClothingComponent.InSlot`--like `ClothingComponent.InSlotFlag` or something, so we could just do `clothing.InSlotFlag == slot.SlotFlags`, But that would mean adding the code to update that variable everywhere it needs to be updated.

paige404: Anyway, `HideLayerClothingComponent` _does_ have a slot-flags in a sense. It has "slots", though that is a misnomer--what it calls slots are actually layers. But those layers have associated slots.

So gas mask's HideLayerClothingComponent has several "slots" (layers) including "snout". Each of those layers has corresponding actual slots (snout has Mask and Head IIRC) and then in this method we're checking the equipped clothing against each of these slots. For clothes that can only be worn in one slot, this is fine. And I can't think of any other examples of clothes that can be worn in multiple slots--it's only the RMC gas mask as far as I know.

paige404: Okay I take it back. I thought I understood the code, I did not not understand the code. Also somehow I had foolishly tested the change with a default snout rodentia model on RMC (i.e. flat-faced) and so hadn't noticed that my proposed fix didn't even actually fix the issue. And thinking that it had fixed the issue, I _really_ misunderstood the code.

I went back over it with a debugger and you're absolutely right, this is an even bigger mess than I'd realized. HideLayerClothingComponent would need a refactor to fix this issue. I'm not 100% sure how to manage that refactor without breaking compatibility with every yaml definition that uses it though. I'll think on it and see what I can come up with.

ElectroJr: Yeah like I said:

> In general ToggleVisualLayers() seems pretty cursed / convoluted and hard to parse and should probably be re-written.

If you do just change it to use its own slot flags, you could make them default to using the clothing flags which might help with compatibility? But either way the code probably needs rewriting to actually make it intelligible.

paige404: Things got delayed for the holidays, but I've made some progress.

As you'll see, I've begun the refactor you recommended and then some. Item prototypes that use HideLayerClothingComponent.Slots still work the same (and will continue to exhibit the bug this is fixing--no way around that). Item prototypes using the new HideLayerClothingComponent.Layers will not exhibit the bug. Item prototypes using both will default to Layers and ignore Slots.

I've also made ToggleVisualLayers much more CPU efficient by avoiding the nested loops. I've discovered why that loop was nested though, as this does introduce a _new_ bug: wearing multiple clothes that hide the same layer (e.g. gas mask + welding mask) can mess with the layer visibility, especially if you then begin toggling the mask on and off.

I have an idea for how to address it--by similarly dressing up HumanoidAppearanceComponent.HiddenLayers into a Dictionary, so that we can track all slots currently hiding a given layer. I didn't want to commit to this before first getting feedback on the approach I've already taken though.

paige404:
It required some bigger refactors than I'd intended, but this solves the problems and removes some excessive looping over the full inventory in `ToggleVisibleLayers` as a bonus.

I have noticed one strange behavior, though, which is somewhat inconsistent: occasionally, when equipping and then removing multiple layer-hiding pieces, the layer will stay hidden for an extra moment. Just long enough to be noticeable, but not long enough to be a serious issue.

For instance, equipping a bandana or a gas mask, then equipping a welding mask on top, then removing both. Sometimes, there will be a brief delay--less than a second--before the snout appears again. I _think_ it has something to do with client and server mismatch, but I'm not sure where that's occurring.

Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
```diff
@@ -85,49 +85,77 @@ private void QuickEquip(
...
+                {
+                    if (!appearanceLayers.Contains(layer))
+                        continue;
 
-            bool shouldLayerShow = true;
-            while (enumerator.NextItem(out EntityUid item, out SlotDefinition? slot))
+                    bool shouldLayerShow = true;
+                    SlotFlags hideSlot = hideLayer.Layers[layer];
+                    if (clothing.EquippedInSlot == hideSlot)
+                    {
```
ElectroJr, line 109: IMO, this should support more than one slot. I.e., it should optionally allow you to hide the same layer even when equipped to a number of different slots.

```suggestion
                    var validSlots = hideLayer.Layers[layer];
                    if (validSlots.HasFlag(clothing.EquippedInSlot))
                    {
```

Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
```diff
@@ -85,49 +85,77 @@ private void QuickEquip(
...
+                {
+                    return false;
+                }
+            }
+            else
+            {
+                return false;
             }
-            _humanoidSystem.SetLayerVisibility(equipee, layer, shouldLayerShow);
+            return true;
```
ElectroJr, line 151: IMO its nicer to read with less if/else nesting. I.e.,
```cs
bool MaskToggleCheck()
{
    if (!hideLayer.HideOnToggle || !TryComp(equipment, out MaskComponent? mask))
        return false;

    return clothing.EquippedPrefix == mask.EquippedPrefix;
}
```

Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
ElectroJr: Actually there's another issue here. `ClothingComponent.EquippedInSlot` gets cleared by `OnGotUnequipped` before `CheckEquipmentForLayerHide` is raised.

The overall logic would also just be cleaner if layer hiding were handled by a separate system that listens for `ClothingGotUnequippedEvent`, `ItemMaskToggledEvent` and `ClothingGotUnequippedEvent` events directed at entities that have a HideLayerClothingComponent.

ElectroJr (changes requested):
Other than `ClothingComponent.EquippedInSlot` gteting cleared too early, it mostly seems fine.  Though there are a few other changes I want to make to just clean up the system a bit, do you mind if I push some changes to your branch? Otherwise I can also just open a separate PR later

paige404:
> Other than `ClothingComponent.EquippedInSlot` gteting cleared too early, it mostly seems fine. Though there are a few other changes I want to make to just clean up the system a bit, do you mind if I push some changes to your branch? Otherwise I can also just open a separate PR later

I don't mind at all! Better to have it clean the first time and fewer PRs IMO

Content.Shared/Clothing/EntitySystems/ClothingSystem.cs
paige404: I can see what you mean. I think the double negative OR is a little harder to read, but I'm not particular about it.

ElectroJr:
I've pushed some changes, including:
- Splitting the layer hiding logic out of ClothingSystem and into a separate `HideLayerClothingSystem`.
- Allowing a single layer to be hidden from more than one slot.
- Reworking parts of `MaskSystem`, mainly to fix some networking & prediction bugs
  - It wasn't actually marking the component as dirty when changing `IsToggled`, leading to prediction issues.
  - There's parts here that I still don't really understand, and there weren't any comments. I've just tried to keep the functionality mostly unchanged, though I have renamed some fields like `MaskComponent.IsEnabled` to `IsToggleable`, because I think that's what it was meant to do??

Content.Shared/Clothing/Components/ClothingComponent.cs
```diff
@@ -60,9 +67,15 @@ public sealed partial class ClothingComponent : Component
...
 
     /// <summary>
-    /// Name of the inventory slot the clothing is in.
+    /// Name and flags of the inventory slot the clothing is currently in.
+    /// Note that this being non-null does not mean the clothing is considered "worn" or "equipped" unless the slot
+    /// satisfies the <see cref="Slots"/> flags.
     /// </summary>
-    public string? InSlot;
+    [DataField]
+    public (string Name, SlotFlags Flag)? InSlot;
```
paige404, line 75: This is a cleaner solution, I agree. But I'd wanted to avoid changing `InSlot` itself purely because I wanted to ensure that if it's being used somewhere downstream this won't break anything.

Content.Shared/Humanoid/SharedHumanoidAppearanceSystem.cs
```diff
@@ -114,70 +116,95 @@ private void OnExamined(EntityUid uid, HumanoidAppearanceComponent component, Ex
...
-                dirty |= humanoid.PermanentlyHidden.Add(layer);
+            if (source is not { } slot)
+            {
+                dirty |= ent.Comp.PermanentlyHidden.Add(layer);
+            }
+            else
+            {
+                var oldSlots = ent.Comp.HiddenLayers.GetValueOrDefault(layer);
+                ent.Comp.HiddenLayers[layer] = slot | oldSlots;
+                dirty |= oldSlots == SlotFlags.NONE;
```
paige404, line 205: I think we need dirty to be true regardless here, as otherwise the client and server get slightly out of sync. I had this problem in my 3rd commit on this PR as well, which I fixed in the 4th by ensuring that dirty was true every time a change was made to `HiddenLayers`.

You can see the bug in effect on the current state of the branch if you put on a gas mask and welding mask at the same time, then remove the gas mask. The snout will stick out for a moment before the client and server sync up again and it gets hidden.

paige404 (commented):
Forgive me the liberty of reviewing your code in your review of my code, but there is one bug to fix here. I don't mind fixing it though, if you want.

ElectroJr (approved):

Content.Shared/Humanoid/SharedHumanoidAppearanceSystem.cs
paige404: I tested the latest one, no bug anymore. I then tested your suggested logic for dirtying, and that also worked. So I've changed it to your suggested logic. I think we're good to ship?

paige404:
... I found another bug, but I have no idea what's causing it. Folding a bandana to wear on the head, the bandana doesn't render when worn.

If you spawn a folded bandana and put it on your head, it renders appropriately. If you unfold that bandana and wear it on your face, that renders too. If you then refold the bandana and put it on your head, the bandana no longer renders on the sprite.

I'll dig into this some more, but this is not ready to ship.

ElectroJr:
> If you spawn a folded bandana and put it on your head, it renders appropriately. If you unfold that bandana and wear it on your face, that renders too. If you then refold the bandana and put it on your head, the bandana no longer renders on the sprite.

The cause was that toggling a mask sets a prefix for the clothing RSI state. Previously foldable clothing was being given special treatment in the mask system without any comments or explanation, and was seemingly not setting its rsi state specifically if the mask was being toggled due to folding, which I accidentally removed in an earlier commit.

I've fixed it now by making bandanas use the existing `FoldableClothingComponent.FoldedEquippedPrefix` instead of having some unintuitive hidden interaction with mask system. I've also fixed another bug in foldable clothing where it could just modify hidden layers while equipped. Every time I look at a new clothing system I find more spaghetti & weird interactions.

paige404:
I can find no bugs. LGTM.

paige404:
Oh hell. Merge conflicts. @ElectroJr it's in the code you changed. Could I ask you to fix it?

paige404:
Alright, thanks for that. What's next? When do we send this?

paige404:
For real though, how long is this PR gonna hang in limbo?

paige404:
For real, though. What's holding this up?

paige404:
This PR says it's still awaiting review from @Jezithyr and @ElectroJr despite that @ElectroJr already approved it. Is this why this process has been halted?

ElectroJr:
Sorry about the delay, that's largely my fault, I've been pretty inactive for the last while. PRs are meant to have [two maintainer approvals](https://docs.spacestation14.com/en/wizden-staff/maintainer/review-procedure.html). I did put a message in the maintainer reviews channel in discord over a month ago asking for a second review, but never followed up on it and should've been more proactive. I'll send another message there now.

The 2-review requirement has since been relaxed for some PRs like bugfixes, but still disallows self merges and I've contributed enough to the PR where it probably blurs the line, but if no one else gets around to reviewing it I can probably justify just merging it without another review?

Regarding GH's reviewers list, it doesn't necessarily mean anything. Unless people assign themselves it's probably just based on the CODEOWNERS file, but that might be out of date (e.g., Jez probably shouldn't be there anymore?) and doesn't mean that those people are required to give approval, though you could probably use it to figure out who might be knowledgeable about a part of the codebase.

metalgearsloth:
> This PR says it's still awaiting review from @Jezithyr and @ElectroJr despite that @ElectroJr already approved it. Is this why this process has been halted?
> 
> The bot has labeled it as Needs Review, but I don't know that it still does?

Needs 2nd review, also needs breaking changes section filled out.

PR #34089, opened 2024-12-27
Fix wagging action name and desc

lzk228:
like there is no need for it to be with locale strings
downstreams can translate it via `ent-ActionToggleWagging` key


slarticodefast (approved):
Makes sense.
Thanks for the cleanup!

PR #34092, opened 2024-12-28
Remove discard functionality from the ChemMaster

sowelipililimute:
## About the PR
The ChemMaster no longer has the ability to discard chemicals.

## Why / Balance
This duplicates other methods of discarding chemicals, including the drain conventionally given to chemistry labs.

Any problems it solves are miniscule compared to the severe UI error that is having:
- a destructive action
- with no confirmation of the destructive action
- attached to a mode
- that is capable of changing without the user looking at it
- with no explicit indication that a completed action was destructive

In plainer words, the way discard in the UI works encourages irreversible mistakes via a bad UI design, and in enough ways that it would require a restructuring of how the ChemMaster UI works to eliminate these mistakes.

Instead of adding a band-aid solution such as a "Are you Sure?" dialogue or attempting to improve the UI, I took a step back and realised that this functionality doesn't solve many problems for chemists - unlike other features of the ChemMaster that are mostly unique to it at round-start, the discard functionality has a litany of alternatives, including but not limited to:
- dump it in a drain
- dump it on the floor and have it evaporate
- dump it on the floor with space cleaner
- put the chemical in a container and send it to disposals

## Recommended Reading On UI Design

As always, the Nielsen-Norman group is the gold standard for introductory articles to these concepts:

- https://www.nngroup.com/articles/slips/
- https://www.nngroup.com/articles/user-mistakes/

Bruce Tognazzini also has some good elaboration on guidelines and examinations of specific examples:
- https://www.asktog.com/columns/069ScottAdamsMeltdown.html
- https://asktog.com/atc/principles-of-interaction-design/#protectUsersWork

If you're more into published books, Don Norman has an entire chapter on design and user errors in his seminal work *The Design Of Everyday Things* titled *Human Error? No, Bad Design*, which opens with:
> Most industrial accidents are caused by human error: estimates range between 75 and 95 percent. How is it that so many people are so incompetent? Answer: They aren’t. It’s a design problem. If the number of accidents blamed upon human error were 1 to 5 percent, I might believe that people were at fault. But when the percentage is so high, then clearly other factors must be involved. When something happens this frequently, there must be another underlying factor.
> 
> When a bridge collapses, we analyze the incident to find the causes of the collapse and reformulate the design rules to ensure that form of accident will never happen again. When we discover that electronic equipment is malfunctioning because it is responding to unavoidable electrical noise, we redesign the circuits to be more tolerant of the noise. But when an accident is thought to be caused by people, we blame them and then continue to do things just as we have always done.
> 
> Physical limitations are well understood by designers; mental limitations are greatly misunderstood. We should treat all failures in the same way: find the fundamental causes and redesign the system so that these can no longer lead to problems. We design equipment that requires people to be fully alert and attentive for hours, or to remember archaic, confusing procedures even if they are only used infrequently, sometimes only once in a lifetime. We put people in boring environments with nothing to do for hours on end, until suddenly they must respond quickly and accurately. Or we subject them to complex, high- workload environments, where they are continually interrupted while having to do multiple tasks simultaneously. Then we wonder why there is failure.
> 
> Even worse is that when I talk to the designers and administrators of these systems, they admit that they too have nodded off while supposedly working. Some even admit to falling asleep for an instant while driving. They admit to turning the wrong stove burners on or off in their homes, and to other small but significant errors. Yet when their workers do this, they blame them for “human error.” And when employees or customers have similar issues, they are blamed for not following the directions properly, or for not being fully alert and attentive.

## Technical details
Removed all of the UI, functionality, and translation strings associated with the mode functionality in the ChemMaster.

## Media

grafik

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Removed Discard mode from ChemMaster


Princess-Cheeseballs:
If merge conflicts are fixed I can approve and merge.

Pgriha:
> * dump it in a drain
> * dump it on the floor and have it evaporate
> * dump it on the floor with space cleaner
> * put the chemical in a container and send it to disposals

From all this variants really workable only first:
* Dump on floor? Only water can evaporate quickly (but it will slippery).
* Dump on flor with space cleaner?. . More water, more slips - that what realy need every chemist.
* Disposals. . . All what sended to them not just disappear. What janitors need to do with huge puddle of chems (and if you add potassium... Boom)?

And still with first you can have issues, when need to utilize really big amount of chems (1000+ units, cause 1000u limit of drain)

TheRealMasterChief117:
> If merge conflicts are fixed I can approve and merge.

got it

PR #34096, opened 2024-12-28
Biome refactor

metalgearsloth:
Few goals with this:
- CPUJobs so people walking into the void doesn't lag the server.
- Cleanup marker layers so it's not its own thing.
- Combine it with dungeons so we can focus on making dungeons good instead of 2 discrete systems. This also makes doing stuff like infinite dungeon spawns easy.

## Breaking changes
Biomes got entirely re-done but it shouldn't be that hard to update.


metalgearsloth:
I really need to finish this but I ADHD'd entity spawn prediction (which is arguably more important) and will come back to it.

PR #34101, opened 2024-12-28
Add pricegun sound

### LINKED ISSUE #33473: Appraisal tool needs sound
## Description

**Reproduction**

**Screenshots**

**Additional context**

DylanWhittingham:
## About the PR
Fixes #33473
Adds a sound that plays when the Appraisal Tool successfully appraises an entity.

## Why / Balance
Previously, the Appraisal Tool had no sound, this improves feedback of the tool with audio. 

## Technical details
added SoundPathSpecifier to "Machines/chime.ogg" to PriceGunComponent class 
added PlayPredicted to SharedPriceGunSystems class in the OnAfterIntereact method.

## Media

https://github.com/user-attachments/assets/c847db24-9942-4407-99d4-a021c1539469

## Requirements
- [X] I have read and am following the Pull Request and Changelog Guidelines

## Breaking changes
None

**Changelog**
:cl:
add: Added appraisal tool sound.


Content.Shared/Cargo/SharedPriceGunSystem.cs
```diff
@@ -10,6 +11,7 @@ namespace Content.Shared.Cargo.Systems;
 /// </summary>
 public abstract class SharedPriceGunSystem : EntitySystem
 {
+    [Dependency] private readonly SharedAudioSystem _audio = null!;
     public override void Initialize()
```
slarticodefast, line 15:
```suggestion
    [Dependency] private readonly SharedAudioSystem _audio = null!;

    public override void Initialize()
```
formatting

Content.Shared/Cargo/SharedPriceGunSystem.cs
```diff
@@ -40,6 +42,7 @@ private void OnAfterInteract(Entity<PriceGunComponent> entity, ref AfterInteract
         if (!args.CanReach || args.Target == null || args.Handled)
             return;
 
+        _audio.PlayPredicted(entity.Comp.AppraisalSound, entity.Owner, args.User);
```
slarticodefast, line 45: This bypasses the use delay, meaning you can spam the sound by clicking repeatedly.
Also the sound does not play if you use the verb in the right click menu.
I would move playing the sound to the server into the `GetPriceOrBounty` function to fix this, which will also better sync up the sound with the popup since it is not predicted. You may have to use `_audio.PlayPvs` in that case.

slarticodefast (changes requested):
Is there maybe a better sound than the chime? It feels a little too loud and more like a warning sound to me.

slarticodefast:
Also your changelog is formatted wrong. Please have a look at the PR guidelines to see how to do it correctly, so that the bot can add it to the in-game changelog.

beck-thompson:
Fun idea! I love little things like this 👍

slarticodefast (changes requested):
Small suggestion that saves us the `TryComp`.
You will have to adjust the `GetPriceOrBounty` function in `SharedPriceGunSystem.cs` accordingly.
Also the new sound file will need to be added to the `attributions.yml` file in the same folder.

Content.Shared/Cargo/SharedPriceGunSystem.cs
```diff
@@ -18,6 +18,7 @@ public override void Initialize()
     }
 
     private void OnUtilityVerb(EntityUid uid, PriceGunComponent component, GetVerbsEvent<UtilityVerb> args)
+
```
slarticodefast, line 21:
```suggestion
```

Content.Server/Cargo/Systems/PriceGunSystem.cs
```diff
@@ -25,10 +27,15 @@ protected override bool GetPriceOrBounty(EntityUid priceGunUid, EntityUid target
...
-            _popupSystem.PopupEntity(Loc.GetString("price-gun-pricing-result", ("object", Identity.Entity(target, EntityManager)), ("price", $"{price:F2}")), user, user);
+            _popupSystem.PopupEntity(Loc.GetString("price-gun-pricing-result",
+                    ("object", Identity.Entity(target, EntityManager)),
+                    ("price", $"{price:F2}")),
+                user,
+                user);
         }
 
-        _useDelay.TryResetDelay((priceGunUid, useDelay));
+        _audio.PlayPvs(entity.Comp.AppraisalSound, target);
```
slarticodefast, line 37:
```suggestion
        _audio.PlayPvs(entity.Comp.AppraisalSound, entity.Owner);
```
Did some testing, and found a minor nitpick. The sound should originate from the apraisal tool's location not from the target, otherwise the loudness of the sound will vary depending on the distance.
I'll make a quick commit for you.

slarticodefast (approved):
The code looks good and the in-game test went well.
Should be good to merge.
Thank you for your for first contribution!

PR #34110, opened 2024-12-29
Adds additional holopads

IamVelcroboy:
## About the PR
Adds additional holopads for
- Command - Evac
- Engineering - PA Control
- General - Park
- Medical - Outpost
- Medical - Psychologist
- Security - Checkpoint
- Service - Breakroom
- Service - Toolroom

## Why / Balance
Mapping/localisation

## Technical details
n/a

## Media
n/a

## Requirements

## Breaking changes
n/a

**Changelog**
n/a

### DIFF
--- a/Resources/Locale/en-US/holopad/holopad.ftl
+++ b/Resources/Locale/en-US/holopad/holopad.ftl
@@ -52,6 +52,7 @@ holopad-general-lounge = General - Lounge
 holopad-general-arcade = General - Arcade
 holopad-general-evac = General - Evac
 holopad-general-arrivals = General - Arrivals
+holopad-general-park = General - Park
 
 # Command
 holopad-command-bridge = Command - Bridge
@@ -66,6 +67,7 @@ holopad-command-qm = Command - QM
 holopad-command-ce = Command - CE
 holopad-command-rd = Command - RD
 holopad-command-hos = Command - HoS
+holopad-command-evac = Command - Evac
 
 # Science
 holopad-science-anomaly = Science - Anomaly
@@ -87,6 +89,8 @@ holopad-medical-paramedic = Medical - Paramedic
 holopad-medical-virology = Medical - Virology
 holopad-medical-front = Medical - Front
 holopad-medical-breakroom = Medical - Breakroom
+holopad-medical-psychologist = Medical - Psychologist
+holopad-medical-outpost = Medical - Outpost
 
 # Cargo
 holopad-cargo-front = Cargo - Front
@@ -109,6 +113,7 @@ holopad-engineering-tech-vault = Engineering - Tech Vault
 holopad-engineering-ame = Engineering - AME
 holopad-engineering-power = Engineering - Power
 holopad-engineering-main = Engineering - Main
+holopad-engineering-pa-control = Engineering - PA Control
 
 # Security
 holopad-security-front = Security - Front
@@ -123,6 +128,7 @@ holopad-security-lawyer = Security - Lawyer
 holopad-security-armory = Security - Armory
 holopad-security-locker-room = Security - Locker Room
 holopad-security-brig-med = Security - Brig Med
+holopad-security-checkpoint = Security - Checkpoint
 holopad-security-evac-checkpoint = Security - Evac Checkpoint
 holopad-security-arrivals-checkpoint = Security - Arrivals Checkpoint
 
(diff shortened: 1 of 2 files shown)


Resources/Locale/en-US/holopad/holopad.ftl
```diff
@@ -66,6 +67,7 @@ holopad-command-qm = Command - QM
 holopad-command-ce = Command - CE
 holopad-command-rd = Command - RD
 holopad-command-hos = Command - HoS
+holopad-command-evac = Command - Evac
```
ScarKy0, line 70: What station has Evac as a command room?

Resources/Locale/en-US/holopad/holopad.ftl
```diff
@@ -123,6 +128,7 @@ holopad-security-lawyer = Security - Lawyer
 holopad-security-armory = Security - Armory
 holopad-security-locker-room = Security - Locker Room
 holopad-security-brig-med = Security - Brig Med
+holopad-security-checkpoint = Security - Checkpoint
```
ScarKy0, line 131: Some stations have so many security checkpoints a single "Security - Checkpoint" might give the wrong idea of adding it everywhere, thus having like 5 "Security - Checkpoint" per map with no means to differanciate them.
The better idea would be to just make a holopad specific to every checkpoint on every map but thatd be huge entity bloat

Resources/Locale/en-US/holopad/holopad.ftl
```diff
@@ -141,6 +147,8 @@ holopad-service-musician = Service - Musician
 holopad-service-mime = Service - Mime
 holopad-service-clown-mime = Service - Clown/Mime
 holopad-service-gameroom = Service - Game Room
+holopad-service-breakroom = Service - Breakroom
```
ScarKy0, line 150: isnt this just General - Lounge?

Resources/Locale/en-US/holopad/holopad.ftl
```diff
@@ -141,6 +147,8 @@ holopad-service-musician = Service - Musician
...
+holopad-service-toolroom = Service - Toolroom
```
ScarKy0, line 151: Isnt this just General - Tools?

ScarKy0 (commented):
I'm kinda against adding more holopads unless accompanied by station changes because their purpose isn't immediately apparent and just leads to entity bloat.
If you plan to add those to a station you should add them in the same PR that makes the change.

Resources/Locale/en-US/holopad/holopad.ftl
IamVelcroboy: Downstream. We have several of them.

Resources/Locale/en-US/holopad/holopad.ftl
IamVelcroboy: Again downstream. It's standard on our maps to have service locked tool rooms.

Resources/Locale/en-US/holopad/holopad.ftl
IamVelcroboy: Perhaps. It's possible I've not really heard them referred to as lounges before. If it's redundant, I can remove it.

IamVelcroboy:
> EDIT: Ah I missed its some DV ones getting upstreamed. If so I think its fine but im still a little confused about Command - Evac, Security - Checkpoint and Service Breakroom and Tools

Command evac rooms are fairly common on our maps. 
Sec checkpoint I can remove. I see your point there.
I'll remove service breakroom in favor of lounge.
We have service locked tool rooms on our maps that are separate from public tool rooms.

If you don't want any of these here. That's ok. I assumed they would be common enough. I can move the ones you don't want back to DV if that's the case.

Resources/Locale/en-US/holopad/holopad.ftl
ScarKy0: What I mean is those rooms with a table and some chairs that are usually around dorms. Or at least thats how ive been mapping it.

ScarKy0:
>Command evac rooms are fairly common on our maps.
Sec checkpoint I can remove. I see your point there.
I'll remove service breakroom in favor of lounge.
We have service locked tool rooms on our maps that are separate from public tool rooms.
If you don't want any of these here. That's ok. I assumed they would be common enough. I can move the ones you don't want back to DV if that's the case.

Its not that I "dont want them" but seeing as some have no use on the upstream I think they shouldnt be moved here. I don't really play downstream or see how things are mapped elsewhere so a maintainer should probably speak up about this, this is mostly just what I think.

IamVelcroboy:
> Its not that I "dont want them" but seeing as some have no use on the upstream I think they shouldnt be moved here. I don't really play downstream or see how things are mapped elsewhere so a maintainer should probably speak up about this, this is mostly just what I think.

Oh, ha, 😄I thought you were reviewing as a maintainer. I guess I'll wait for an official maintainer review before making changes so I don't have to make multiple ones. I see your point though.

IamVelcroboy:
Moving this downstream

PR #34129, opened 2024-12-30
Anchored pipe sprite layers.

### LINKED ISSUE #33495: Make anchored pipes visibly different
## Description
Building pipes is basically the primary thing atmospheric technicians do, and yet, pipes are very annoying to deal with. If you're modifying something (I.E. the weird teg setups that always come with almost every station) you're gonna have some pipes unanchored, and some anchored. There's no visual difference between them.

**potential solution**
brackets. Two little grey rectangles with dots at the ends perpendicular to the pipe that only appear on anchored pipes. Very simple.

Alpaccalypse:
## About the PR
Resolves #33495. Displays bolted brackets when pipes are anchored.

## Why / Balance
Visual feedback for anchored vs. unanchored sprites. Unanchored pipes are almost identical to anchored; when in place, it can be very hard to tell if individual pipe sections are unanchored, in a pipe network. This should make it much more obvious to the player, while maintaining a minimalist aesthetic to the pipe sprites.

## Technical details
Allows a single brace and rivet sprite layer to be overlayed onto each anchored pipe type.

## Media
Anchored pipe appearance:
pipeStraightAnchoredDemo
pipeBendAnchoredDemo
pipeTJunctionAnchoredDemo
pipeFourwayAnchoredDemo

In-game example:

With painted and unanchored pipes:

## Requirements

## Bugs
* (Minor/Subjective): Two straight pipes anchored crossways cause brackets look odd, as they appear bolted onto the opposing pipe.

**Changelog**
Cosmetic, though relatively significant for atmos gameplay. Presumably no changelog needed.


Alpaccalypse:
It is annoying that a different sprite/alignment is needed for some pipes, or this could just be added to the pipeConnector.png and I assume would just work, without needing any extra code.

Right now I'm tacking down how that actually works, as I assume using a similar method would be the best way to implement the anchoring functionality for these layers, into the game. I assume a Visualizer component for related entities in pipes.yml would also work, though so far, I've struggled to get this working using anchoring as a state.

Alpaccalypse:
> You can simply use a GenericVisualizer to get the sprite change working, no C# needed. See #33810

Thanks so much, that was exactly what I was looking and couldn't work out for myself. You saved me a lot of head scratching.

Alpaccalypse:
Well I uploaded a working version of the pipe.yml, but not sure it is quite right.

I was hoping I could get `visible: true` working, as with the gas cannisters, but it wouldn't work, just displayed the anchored version for all states. It seems to me like the uploaded version is putting down a second layer of the default pipe sprite when unanchored, but I may be misinterpreting how Visualizer layers work. It certainly seems to WAI in the game at least.

Alpaccalypse:
It appears the zfighting bug I identified is unique to the test environment. Opon map loading, or anchoring/constructing a new pipe, the game correctly picks one pipe and bolt as the top layer.

UbaserB (changes requested):
I like the concept a lot, but there's a couple small issues. 
Firstly, the 'screws' (as in, individual bordered pixels on the sides) should be less separated from the sheet of metal wrapping over the pipe *in terms of contrast*, not position. Also make sure it's not to the point of fully looking like a part of the sheet. 

Second issue is during 3-4 way pipe intersections. The contrast is way too intense as compared to the 1-2 way ones, and it's simply jarring to see. Takes up too much attention on the screen and it should be similar to the others.

Alpaccalypse:
Thank for the feedback, both seem like fair points to me. I'll do a revision to address this and post an update.

Emisse:
feels visually messy to me

Emisse:
brackets should probably stick out from the sides of the pipe less

Alpaccalypse:
I was on the fence about how much they should stick out. I think if I follow Ubasers advice about contrast, it will allow them to be pulled back in without merging too much with the pipe outline.

Alpaccalypse:
Updated pipes to pull focus less, and blend in better. Have also shortened brackets slightly for straight and corner pipes, without losing too much fidelity I think.

image
image

Alpaccalypse:
Before

Alpaccalypse:
After (too subtle?)

Alpaccalypse:
Had a go at trying to make the Fourway/TJunction brackets less dominating. Problem is, reducing the width leaves very little room for any filler, it's all outline needed for definition. Stil, might be of interest. Am basically pre-empting queries. 
pipeFourwayAnchoredv2

Emisse:
sticks out too far on the sides and interrupting the color every tile makes it feel very visually messy. Not sure how I would go about this.

PR #34130, opened 2024-12-30
Stop greytide virus spacing

### LINKED ISSUE #33798: Greytide virus can space the station due to improper checks.
## Description

The new Greytide virus is made to not target airlocks with external access, which isn't good enough on some airlocks. Salvage airlocks for example may be salvage access and not external, meaning the check allows all of cargo to be bolted open to space.
**Reproduction**

1. Have greytide target Engineering or Cargo
2. Engineering or Cargo airlocks such as solars or salvage are now bolted open
3. The entire station is spaced.
**Screenshots**

image

**Additional context**

Maybe make the event blacklist AirlockExternal prototypes?

Booblesnoot42:
## About the PR
Fixed the Greytide virus event spacing the station by opening airlocks with access to space.

## Why / Balance
Resolves #33798

## Technical details
-Created a new ExternalAccessMarker component for airlocks with access to space and added it to the AirlockExternal prototype.
-Added a check in GreytideVirusRule.cs to skip over airlocks with the ExternalAccessMarker component.

## Media
Uneeded(?), but I did test it on a local server and it did fix the problem.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: The Greytide virus event is now unable to open external airlocks.

### DIFF
--- a/Content.Server/StationEvents/Events/GreytideVirusRule.cs
+++ b/Content.Server/StationEvents/Events/GreytideVirusRule.cs
@@ -60,6 +60,7 @@ public sealed class GreytideVirusRule : StationEventSystem<GreytideVirusRuleComp
 
         var firelockQuery = GetEntityQuery<FirelockComponent>();
         var accessQuery = GetEntityQuery<AccessReaderComponent>();
+        var externalAccessQuery = GetEntityQuery<ExternalAccessMarkerComponent>();
 
         var lockQuery = AllEntityQuery<LockComponent, TransformComponent>();
         while (lockQuery.MoveNext(out var lockUid, out var lockComp, out var xform))
@@ -86,10 +87,7 @@ public sealed class GreytideVirusRule : StationEventSystem<GreytideVirusRuleComp
         while (airlockQuery.MoveNext(out var airlockUid, out var airlockComp, out var doorComp, out var xform))
         {
             // don't space everything
-            if (firelockQuery.HasComp(airlockUid))
-                continue;
-
-            if (HasComp<ExternalAccessMarkerComponent>(airlockUid))
+            if (firelockQuery.HasComp(airlockUid) || externalAccessQuery.HasComp(airlockUid))
                 continue;
 
             // make sure not to hit CentCom or other maps
--- a/Content.Server/StationEvents/Events/GreytideVirusRule.cs
+++ b/Content.Server/StationEvents/Events/GreytideVirusRule.cs
@@ -89,6 +89,9 @@ public sealed class GreytideVirusRule : StationEventSystem<GreytideVirusRuleComp
             if (firelockQuery.HasComp(airlockUid))
                 continue;
 
+            if (HasComp<ExternalAccessMarkerComponent>(airlockUid))
+                continue;
+
             // make sure not to hit CentCom or other maps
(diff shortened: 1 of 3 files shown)


Content.Server/StationEvents/Events/GreytideVirusRule.cs
```diff
@@ -89,6 +89,9 @@ protected override void Started(EntityUid uid, GreytideVirusRuleComponent virusC
             if (firelockQuery.HasComp(airlockUid))
                 continue;
 
+            if (HasComp<ExternalAccessMarkerComponent>(airlockUid))
```
slarticodefast, line 92: This should use a query, same as the firelockQuery above

lzk228 (commented):
i think it should be completely different
a list in GreytideVirusComponent with accesses, and if a door has the access, we skip it

slarticodefast:
>i think it should be completely different
a list in GreytideVirusComponent with accesses, and if a door has the access, we skip it

It already works that way and EVA access is blacklisted to prevent most spacing. Also everything with the FirelockComponent is skipped. However, due to inconsistent mapping and some doors facing space not having EVA access this is not enough.

lzk228:
well yeah, makes sense then

EthanQix:
At this point we have what, 3 different blacklists for the greytide virus ? It might be best to just make one unified greytide virus blacklist for more granular control over which doors cannot get affected, in case we want to add more special cases later.

Content.Server/StationEvents/Events/GreytideVirusRule.cs
Booblesnoot42: Changed it, does it look better now?

slarticodefast:
Might also make sense to make a single GreyTideVirusTargetComponent instead as a whitelist. But keeping the access blacklist is still useful. I'll take a look at the yaml files and figure out which approach is better once I'm back home on the 2nd of January.

Booblesnoot42:
> At this point we have what, 3 different blacklists for the greytide virus ? It might be best to just make one unified greytide virus blacklist for more granular control over which doors cannot get affected, in case we want to add more special cases later.

Something like a boolean function that takes in the airlock uid and returns whether it's blacklisted?

SlimmSlamm:
Greytide virus spacing rooms is what makes it funny though. Why do we feel the need to remove the best part of random events?

slarticodefast:
> Greytide virus spacing rooms is what makes it funny though. Why do we feel the need to remove the best part of random events?

The problem is that our atmos simulation is currently broken and airlocks don't work as intended. They only trigger *after* both sides have already been spaced. So excluding the space facing airlocks is a bandaid fix for now.

ArtisticRoomba:
> Greytide virus spacing rooms is what makes it funny though. Why do we feel the need to remove the best part of random events?

I disagree. Unfortunately firelocks don't work correctly right now, and as a result spacing can easily spread between rooms and adjacent hallways. I've heard of this reportedly ending rounds, and while I take all engineering-related complaints with a grain of salt (as it's usually a player issue) this event does introduce problems and I have seen the described issues happening.

ArtisticRoomba:
This PR might not be needed if #38918 gets merged.

slarticodefast:
As of #38918 firelocks actually work again, so this isn't really needed anymore.

PR #34055, opened 2024-12-24
UI improvements for holopads

chromiumboy:
## About the PR

This PR makes some improvements to the holopad UI (see #33954)

## Why / Balance

Changes made:
- A search filter has been added to the holopad contacts list
- Incoming calls now also specify the name of the device that the call is coming from

## Technical details

N/A

## Media

Screen Shot 2024-12-24 at 2 09 27 PM
Screen Shot 2024-12-24 at 2 10 16 PM

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

TelephoneComponent.LastCallerId is now a tuple with three strings; the third string is the name of the device the last called the telephone

**Changelog**

:cl:
- tweak: The holoapd UI has been updated to include a text-based filter for the contacts list, and also to specify the name of the holopad that incoming calls originate from

PR #34137, opened 2024-12-30
Holopad UI tweak for incoming calls

chromiumboy:
## About the PR

This PR tweaks one of the UI changes brought about in #34055, specifically removing the base name from the call source and displaying just its label. If the device is unlabelled, the call will be listed as having an 'unknown source'

## Why / Balance

Including the base name of the call source makes it difficult to disguise. This would be problematic, for example, for a syndicate agent making holocalls from a listening output

## Technical details

N/A

## Media

holopad_incoming_call_v1
holopad_incoming_call_v4

## Requirements

## Breaking changes

N/A

**Changelog**

N/A


Cojoke-dot:
It should display as unknown source when emaged too, but that it's not important to this pr specifically I guess.

PJB3005 (approved):

PR #34142, opened 2024-12-31
Id console fix

### LINKED ISSUE #34100: Id card console cannot change access if the privileged ID is missing any access tag the other ID has
## Description
This is problematic, because if you microwave your ID and gain any access tag the HOP does not have in his own ID (wizard, syndicate etc.), then the microwaved card can no longer be modified at all.
Expected behaviour:
You should be able to add or remove any access tag in the priviliged ID. All tags that are unavailable should keep the buttons disabled and greyed out as they are right now.

**Reproduction**
- Spawn an ID card console, a HOP ID card and a captain ID card.
- Use the HOP ID card as the priviliged ID, then insert the captain ID card after it
- Try to remove some of the access the HOP card has available.
- The button will mispredict, then reset to active.

Peque-e:
## About the PR

Fixed a bug where if the priviliged id card had missing accesses compared to the id card to be modified it would not change the resulting accesses.

Fixes #34100

## Technical details

Fixed the way the difference is calculated, wich caused the issue.

## Requirements

## Breaking changes

**Changelog**


PJB3005 (changes requested):
This still doesn't seem to work properly. From testing, trying to modify an ID card with centcomm access causes that access level to be removed silently.

Pequemax:
Is there a place where all the "protected" access are saved?(Like syndicate, centcomm, etc) My main issue are with these ones

PJB3005:
> Is there a place where all the "hidden" access are saved?(Like syndicate, centcomm, etc) My main issue are with these ones

I feel like any solution that requires knowing all hidden accesses (instead of simply solely working off the ones available to the ID console) is fundamentally flawed?

Pequemax:
I was planning to do a datafield with all the access that are unknown to the user of the computer, its less than an ideal solution since they would need to be updated everytime a new traitor role is added

Pequemax:
I got 2 ideas to fix the issue about the missing hidden access, the first one is add them to the id console as normal access, but they will stick out like a sore thumb(though the computer user would not be able to see if they have it or not), and the other one is to add them to the console component as hidden access, but that would have the problem outlined in the previous comment, plus the fact that i am not too proficient with the codebase

Pequemax:
All right, I think I managed to get it working, I tested HoP removing Cap access, Centcom and Syndies keeping their hidden access, and HoP trying to remove access they dont have, but i think I overcorrected something since the UI/UX label popped. I had to add a list of all hidden access to make it work tho, so it might not be good to pull

PJB3005 (changes requested):
Adding a list of unknown accesses is not a proper way to do this. Just make the console not remove access levels on the ID card that the console can't give out.

Pequemax:
Apologies for the back and forth, fixed without the use of a list for the hidden access, checked with:

- HoP removing Cap access, both permited and prohibited
- Hop giving syndies access, and keeping Nuke Ops access
- Hop giving centcomm access, and keeping Centcomm access

Content.Server/Access/Systems/IdCardConsoleSystem.cs
```diff
@@ -151,25 +151,32 @@ private void TryWriteToTargetId(EntityUid uid,
...
-        var difference = newAccessList.Union(oldTags).Except(newAccessList.Intersect(oldTags)).ToHashSet();
+        var ableToModify = component.AccessLevels.ToHashSet();
+
+        var addedTags = newAccessList.Except(oldTags).Select(tag => tag).ToHashSet();
+        var removedTags = oldTags.Except(newAccessList).Select(tag => tag).ToHashSet();
+
+        //Probably there is a cleaner way to do this, visible difference is the difference between the new access and the old access the user of the computer could see
+        var difference = addedTags.Union(removedTags).ToHashSet();
+        var visibleDifference = difference.Intersect(ableToModify).ToHashSet();
+        var hiddenAccess = difference.Except(visibleDifference).ToList();
```
PJB3005, line 162: It makes more sense to calculate this as `difference.Except(ableToModify)` instead.

Content.Server/Access/Systems/IdCardConsoleSystem.cs
```diff
@@ -151,25 +151,32 @@ private void TryWriteToTargetId(EntityUid uid,
...
-        // I hate that C# doesn't have an option for this and don't desire to write this out the hard way.
-        // var difference = newAccessList.Difference(oldTags);
-        var difference = newAccessList.Union(oldTags).Except(newAccessList.Intersect(oldTags)).ToHashSet();
+        var ableToModify = component.AccessLevels.ToHashSet();
+
+        var addedTags = newAccessList.Except(oldTags).Select(tag => tag).ToHashSet();
+        var removedTags = oldTags.Except(newAccessList).Select(tag => tag).ToHashSet();
+
+        //Probably there is a cleaner way to do this, visible difference is the difference between the new access and the old access the user of the computer could see
+        var difference = addedTags.Union(removedTags).ToHashSet();
```
PJB3005, line 160: `.ToHashSet()` is needlessly expensive.

Content.Server/Access/Systems/IdCardConsoleSystem.cs
```diff
@@ -151,25 +151,32 @@ private void TryWriteToTargetId(EntityUid uid,
...
+        if (!visibleDifference.IsSubsetOf(privilegedPerms))
         {
             _sawmill.Warning($"User {ToPrettyString(uid)} tried to modify permissions they could not give/take!");
             return;
         }
 
-        var addedTags = newAccessList.Except(oldTags).Select(tag => "+" + tag).ToList();
-        var removedTags = oldTags.Except(newAccessList).Select(tag => "-" + tag).ToList();
-        _access.TrySetTags(targetId, newAccessList);
+        _access.TrySetTags(targetId, newAccessList.Union(hiddenAccess));
```
PJB3005, line 172: The way this code works is extremely non-obvious, and I had to check your code out in a debugger to even understand it. You are implicitly relying on `hiddenAccess` only ever containing negative differences due to removal of these tags by the UI, which works, but is extremely unclear.

Please either restructure this code to make the behavior more obvious, or make it heavily commented.

Content.Server/Access/Systems/IdCardConsoleSystem.cs
```diff
@@ -151,25 +151,32 @@ private void TryWriteToTargetId(EntityUid uid,
...
+
+        var changes = addedTags.Select(tag => "+" + tag).Union(removedTags.Except(hiddenAccess).Select(tag => "-" + tag));
 
         /*TODO: ECS SharedIdCardConsoleComponent and then log on card ejection, together with the save.
         This current implementation is pretty shit as it logs 27 entries (27 lines) if someone decides to give themselves AA*/
         _adminLogger.Add(LogType.Action, LogImpact.Medium,
-            $"{ToPrettyString(player):player} has modified {ToPrettyString(targetId):entity} with the following accesses: [{string.Join(", ", addedTags.Union(removedTags))}] [{string.Join(", ", newAccessList)}]");
+            $"{ToPrettyString(player):player} has modified {ToPrettyString(targetId):entity} with the following accesses: [{string.Join(", ", changes)}] [{string.Join(", ", newAccessList)}]");
```
PJB3005, line 179: This doesn't correctly log the resulting access list, as `newAccessList` won't show hidden accesses.

PJB3005 (changes requested):

whatston3:
Should be closed, superceded [by] #44016.  Cheers for the bugfix!

PR #34148, opened 2024-12-31
Adds purchase delay to nukie medical bundle, max purchase limt of 1

K-Dynamic:
## About the PR

- Adds a 15-minute delay and purchase limit of 1 to nukie medical bundle

### Blitzops
Reduces the effectiveness of roundstart blitzops, which are already difficult to counter by nature and cause short rounds.

This doesn't stop nukies from buying medical kits, raiding medical or simply getting the drop on crew before they attack back. However, it does mean blitzops become riskier with less available healing chemicals, thus discouraging them.

### Warops
Doubling the TC amount distorts balance, making expensive purchases more readily available (keeping in mind expenses are generally the only balancing factor). This is especially apparent with medical bundles as you stock up on multiple autoinjectors, reducing the need for a nukie medic's hypospray.

A single purchase limit should prevent nukies from stocking up on several autoinjectors at once, making healing more tense and impactful.

## Technical details

yml changes only

- `Resources/Prototypes/Catalog/uplink_catalog.yml` (adds restockTime to uplink catalog for medical bundle, adds stock condition)

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Nukie medical bundles now have a 15 minute delay before purchase, as well as a stock limit of one.

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -714,6 +714,8 @@
       blacklist:
         components:
           - SurplusBundle
+    - !type:ListingLimitedStockCondition
+      stock: 1
 
 # Deception
 


beck-thompson:
I always knew this thing would be extremely powerful, especially for round start attacks. I think this is a great skill check for those kind of attacks (The medic actually has to make chems quickly)

Cojoke-dot:
> Added a one purchase limit so nukies don't stock up on injectors during warops so easily.

That's not going to stop someone from just using another nukies uplink...

Stewie523:
Terrible change.

Cojoke-dot:
> Added a one purchase limit so nukies don't stock up on injectors during warops so easily.

I thought about this again, isn't that the whole point of the item? Stock up on injectors so everyone can take a couple or the medic just uses them instead of jugs. Limiting it just seems like it's against the intent of the item being used by less experienced medics. The timer seems fine to me just the limit does not make much sense.

Cojoke-dot:
> Basically imagine trying to shoot down a jug suit with 3 or 4 punct pens, while a nukie medic and their team rotates around the station.

A punc pen has 10u of punc and 5u of Tranacid. That's 2 uses of a hypo for the punc which is the more important part. I've been and seen jug suits just buy a hypo pen to do this. From what I've seen if a nukie is skilled enough to use combat chems and fight they can generally make a jug of punc within 15 minutes, so they would just take the cheaper option if the hypo pen.

I think it's just going to affect players not as skilled in chem from getting meds they need if topped with the timer change.

BramvanZijp:
Another downside with pens is that if unskilled nukies use them and an agent doesnt stress the importance of calling out their usage, ODs and/or Razorium can happen easily if the agent is trying to make chems to do their actual job.

Keep in mind, 24 TC is a lot, and if nukies want to spend 48TC on 2 med bundles, let them, since that is a massive blow to the rest of their gear, and even then probably will not sustain them for an entire operation.

Alongside the 15 minute delay, which is actually reasonable IMO, the only change I could see being good here without nerfing the bundle to hell is maybe a slight cost increase to 25, but limiting it to just 1 seems like an arbitrary restriction.

SlimmSlamm:
> Could elaborate further so I can make better changes? I am not a mind reader.

I dont think there should be a change at all?

cohanna:
This seems like a good way to kill an item. I frankly see no reason anyone would bother waiting for the timer to finish, bar incompetent/absent agents. At best this will be used as a convenient way to see if you are moving fast enough for war ops (if you can buy this while still on nukie planet you aren't moving fast enough). 

Blitzops are already a very rare occurrence, and while it is a very effective strat it's hardly frequent enough to deserve a "nerf". 
In my experience I see blitzops somehow less often than zombie ops despite it being many times more effective. It's a neat gimmick and is hard to coordinate. 

This just kinda seems like a PR made in response to losing to a blitzops...

cohanna:
You could make it so that the shuttle can't leave within the first 10-15 mins unless war is declared. 

This would make it so that the crew is at least expecting nukies.

PR #27640, opened 2024-05-03
Containment Field Alarms

Vermidia:
## About the PR

Adds an item that when used on containment field generators makes them send messages over the engineering radio (or common when low) when a containment field is losing power or loses a connection because of power. They can be ordered from cargo.

## Why / Balance

Gives engineering a heads up to either fix the problem or get everyone else to start running for their lives for one of the simpler ways to sabotage the tesla/singulo (cutting a few wires and waiting).

Implements part of [this design doc](https://docs.spacestation14.com/en/space-station-14/departments/engineering/proposals/engine-containment.html).

This PR does not replace the current containment fields engineering starts with, but variants and crates with alarms are available for mapping in.

## Technical details

Uses a new component that the field generator checks for and sends power data to to keep it updated. The component uses the radio and navmap systems to send messages about it's location.

## Media

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

https://github.com/space-wizards/space-station-14/assets/32827189/82faa3f6-4b1e-4120-85f2-9a9baa512ab7

## Breaking changes

Hope not

**Changelog**

:cl:
- add: Containment field generators can now be upgraded with power alarms that alert over radio when losing power or breaking connections.
- add: You can buy these power alarms from cargo.

PR #34171, opened 2025-01-02
Add alarms to containment field generators

minus1over12:
## About the PR

* Containment field generators play an alarm sound when their buffer is draining.
* Containment field generators play a different alarm and broadcast a warning on the engineering radio on failure.
## Why / Balance

This makes an imminent containment failure harder to miss, and allows faster reaction in the event that a containment failure does happen. Additionally, since blind characters may have trouble checking the field generator power indicator, this gives an audible warning for them that something is incredibly wrong.
## Technical details

The Entity System for the Field Generators was changed to play sounds on failure and buffer power loss. Additionally, a radio message was added to the Entity System for when failure happens.

3 engine alert sounds from /tg/ were added to use (both for this and the other station wide destruction engine events).
## Media

https://www.youtube.com/watch?v=37U9KWsVCRE

## Requirements

## Breaking changes

**Changelog**

:cl: WarPigeon
- add: Containment Field Generators now have alarms for containment failure.


SlimmSlamm:
this will make it essentially impossible to loose from incompetence which i think should be fine. looses are fun and we rarely ever see DAGDs loose because its considered lame and boring. i dont wanna make looses impossible to see or rare when it already is really rare right now. we also just recently had a change to where you cannot loose to incompetency setup. saboing containment fields will also be so much harder for no reason either.

Vermidia:
Just noting  #27640 is still up as well

minus1over12:
> Just noting #27640 is still up as well

Didn't notice that, I'll shut this in favor of yours, but I do think you should add in the engine alarm sounds on your PR.

PR #34173, opened 2025-01-02
Fix sink draining sound

themias:
## About the PR

This fixes the sink draining sound continuing to play forever.

I also saw that floor drains did not play the draining sound when manually emptying into them, so I fixed that too.

Reported on discord here:
https://discord.com/channels/310555209753690112/1323872026288590889/

## Technical details

The check was checking for available volume instead of volume, so it would only enter if completely full.

We don't want to exit early if we're a floor drain with puddle logic, and I added a volume check down below so that they can play the sound when emptying manually.

I've tested manually emptying into a sink, manually emptying into a floor drain, and dumping liquid near a floor drain, and it all seems to work as expected.

## Media

Dumping a water bottle into a sink:
<video src='https://github.com/user-attachments/assets/79f70c5d-a6eb-49d3-8719-27a60227f4e6'/>

Dumping a water bottle into a drain:
<video src='https://github.com/user-attachments/assets/f77bcbf1-fd54-4226-9a4d-e41e8e226143'/>

Spilling whiskey near a drain: 
<video src='https://github.com/user-attachments/assets/9a233303-cda8-4f2e-a130-8784bf4a695f'/>

## Requirements

**Changelog**
:cl:
- fix: Fixed sink draining noise continuous playback

### DIFF
--- a/Content.Server/Fluids/EntitySystems/DrainSystem.cs
+++ b/Content.Server/Fluids/EntitySystems/DrainSystem.cs
@@ -141,7 +141,7 @@ public sealed class DrainSystem : SharedDrainSystem
             if (!_solutionContainerSystem.ResolveSolution((uid, manager), DrainComponent.SolutionName, ref drain.Solution, out var drainSolution))
                 continue;
 
-            if (drainSolution.AvailableVolume <= 0)
+            if (drainSolution.Volume <= 0 && !drain.AutoDrain)
             {
                 _ambientSoundSystem.SetAmbience(uid, false);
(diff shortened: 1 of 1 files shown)


themias:
> Looks reasonable. Where media?

I've updated with videos showing the scenarios, manual sink emptying, manual drain emptying, and auto drain.
(Mop bucket works as expected but takes longer due to larger volume.)

whatston3 (approved):
Small set of changes, works well in testing.  Good stuff.

PR #34179, opened 2025-01-02
Fixing bowl foods

### LINKED ISSUE #34172: Salads can't be eaten with a fork
## Description ##
Salads cannot be eaten with a fork. Attempts to do so lead to a "... cannot be eaten with a fork" message.

**Reproduction**
Attempt to eat any salad with a fork.

**Screenshots**
Salad Problem

**Additional context**
The issue is caused by salad prototypes (located at Resources/Prototypes/Entities/Objects/Consumable/Food/soup.yml) inheriting from FoodBowlBase (in the same file), which requires a spoon to eat because it seems to be based on the assumption of being soup. I believe the best solution is to create a new abstract prototype FoodSaladBase that inherits from FoodBowlBase and overrides its Food component to require a fork, then change all the salads to inherit from that instead of FoodBowlBase. 

I'll probably fix it tomorrow if someone doesn't beat me to it.

Booblesnoot42:
## About the PR
Cleaned up the prototypes referring to foods that are eaten out of bowls. Previously, everything that inherited from FoodBowlBase, including salads and rice dishes, was considered a soup (had a soup tag and could only be eaten with a spoon). This changes that so the different dishes have their own tags and can be eaten with proper utensils. The file containing these prototypes, soup.yml, was also renamed to bowl_foods.yml to properly describe what it contains.

### Testing
- Spawn any salad (for example, Caesar salad)
- Spawn a fork
- Eat the salad with the fork.
- Repeat with any rice dish (for example, boiled rice)

## Why / Balance
Resolves #34172 and makes the tags on bowl foods better represent them.

## Technical details
-Removed the "soup" tag from FoodBowlBase,
-Created "Salad" and "Rice" tags in tags.yml.
-Created three new abstract prototypes: FoodSaladBase, FoodRiceBase, and FoodSoupBase. All of these inherit from FoodBowlBase and have the appropriate tags and utensils.
-Changed the parents of all relevant foods to inherit from the correct base class.
-Renamed soup.yml to bowl_foods.yml.

## Media
No real in game changes aside from being able to eat salad and rice with a fork.

## Requirements

## Breaking changes
Breaks forks which modified soup.yml.

**Changelog**
:cl:
- fix: Salads and rice dishes can now be eaten with forks.


Resources/Prototypes/Entities/Objects/Consumable/Food/bowl_foods.yml
```diff
@@ -372,10 +380,22 @@
           Quantity: 5
 
 # Rice
```
Alpaccalypse, line 382: As you are renaming the file and adding new parent entities anyway, you may as well move `FoodRiceBase` and the entities that inherit it to be above `FoodSaladBase`, so it's all nicely alphabetical. Total nitpick I know, but I find it satisfying to see if there's no reason not to.

Alpaccalypse (changes requested):
I don't see any mistakes here, and other than my one nitpick this all looks like it should work as intended and would be a greatly satisying update, for something that might seem insignificant.

Any other changes I could think of are really just nitpicking to the point of pedantry, as this does what it sets out to do just fine.

Alpaccalypse:
I've a couple of questions, that are unrelated to this PR actually working:

- Is there a reason not to add `- salad` as a tag to anything that is conceivably a salad, e.g. coleslaw, kimchi salad?
- Is there a problem with adding more than one utensil? I would have thought most rice dishes could be eaten with a spoon or fork. In a lot of places it is not uncommon to eat rice with chopsticks even (which seem to be a utensil type, but in their own file for some reason).

Booblesnoot42:
> I've a couple of questions, that are unrelated to this PR actually working:
> 
>* Is there a reason not to add `- salad` as a tag to anything that is conceivably a salad, e.g. coleslaw, kimchi salad?

Just to clarify, anything that inherits from FoodSaladBase has the "Salad" tag because it inherits the tag component, including both coleslaw and kimchi salad. The foods that explicitly have the "Salad" tag do so because they need to add another tag (e.g. fruit salad needs the "Fruit" tag, but if it had a tag component with just "Fruit" then it would lose the "Salad" tag when the new tag component overwrote the inherited one, so it needs to have them both on the new tag component).
>* Is there a problem with adding more than one utensil? I would have thought most rice dishes could be eaten with a spoon or fork. In a lot of places it is not uncommon to eat rice with chopsticks even (which seem to be a utensil type, but in their own file for some reason).

The "utensil" field on the "Food" component (at Content.Server/Nutrition/Components/FoodComponent.cs) only allows one value at the moment. I agree that it would make more sense for there to be multiple possible utensils, though. I'll make a new issue for it.

Alpaccalypse:
Thanks for salad explanation, that all makes perfect sense. I forgot about salad inheritance.

Resources/Prototypes/Entities/Objects/Consumable/Food/bowl_foods.yml
Booblesnoot42: I tried to do this, but for some reason switching the order around causes everything in the file to not be loaded into the game, i.e. I can't find them in the entity spawner. I'll give it another try some other time, but it might be best to do it in a cleanup pr.

Alpaccalypse: Yeah, if it works one way and not the other entity ordering is not really a critical issue. Also, as you say, might be better for a cleanup PR as I'm certain there are plenty of files just in this subfolder that could do with a tweak on that front.

Alpaccalypse (approved):
Given that my requested change was just a style suggestion, with reasoning why it wasn't done, and I don't see any errors with the .yml, this gets my non-maintainer approval. I think it's a good update to foods, even if you can just scoff anything with your hands instead if you want to.

slarticodefast:
Looks, like this was fixed in the food refactor, but it seems there is now a similar bug the other way around and you can eat soup with a fork. I'll open a new issue for that.

PR #34184, opened 2025-01-02
Give glasses a small chance to break when slipping

thetolbean:
## About the PR
This PR gives vision correcting glasses a very small chance, 1%, to break when slipping.
This also adds a new pair of glasses: Broken Glasses.
Broken glasses still provide a small amount of vision correction, but only a fraction compared to standard glasses.
This also adds a crafting recipe to replace the glass in a pair of broken glasses.

In general, the `breakonslip` component can be used in general for any piece of clothing and can be customized with a different end equipment, message, or chance. 

## Why / Balance
image

Adds RP potential around glasses as they are a notoriously fragile piece of clothing.
Makes glasses and their care more vital.
Gives the ability to repair broken glasses with a very simple recipe in order to prevent _some_ frustration with them breaking.

## Media

https://github.com/user-attachments/assets/92d6a362-27d9-418f-8e59-5ac96c0875d1

https://github.com/user-attachments/assets/6cb17857-a7f4-445d-b4ac-846a3e0ae8c9

_These videos do not show the broken glasses giving vision correction._

## Requirements

**Changelog**

:cl:
- add: Vision correcting glasses now have a chance to break when slipping.
- add: Broken glasses can now be repaired via the crafting menu.


ps3moira:
Would be cool if there was a visual change to the glasses when they break

thetolbean:
> Would be cool if there was a visual change to the glasses when they break

There is a very subtle change to the glasses which isn't shown in the videos.
My goal was to just have a crack in each lens. but each lens is only a 2x2 sprite when worn

Content.Shared/Clothing/EntitySystems/BreakOnSlipSystem.cs
```diff
@@ -0,0 +1,31 @@
...
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<BreakOnSlipComponent, InventoryRelayedEvent<SlippedEvent>>(OnSlip);
+    }
+
+    private void OnSlip(EntityUid uid, BreakOnSlipComponent component, InventoryRelayedEvent<SlippedEvent> args)
+    {
+        if (!_random.Prob(component.BreakChance) || _net.IsClient)
+            return;
+        EntityManager.DeleteEntity(uid);
```
TheShuEd, line 26: we already have the Breakable system, which is used for vending machines, for example. It makes sense to rework the system so that it deals damage to objects when player slips.

in this case, the replacement of the sprite and behavior will depend on another system.

this is also useful because hitting the glasses with a club will also break them according to the same system

TheShuEd (changes requested):

Hrosts:
> The sprites are so small differences are hard to notice

You can make it as if they got somewhat bent around the bridge, shift one of the eyes a pixel lower.

beck-thompson:
I can finally make velma :laughing:
super fun pr

Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
```diff
@@ -77,10 +77,43 @@
...
+  - type: Destructible
+    thresholds:
+    - trigger:
+        !type:DamageTypeTrigger
+        damageType: Blunt
+        damage: 10
+      behaviors:
+      - !type:SpawnEntitiesBehavior
+        spawn:
+          ClothingEyesGlassesBroken:
```
TheShuEd, line 101: image
you can try did it like this with `Breakage`, and also add Repairable component, io be able to fix the glasses back up. But that's up to you, I don't require it.

thetolbean: The glasses are currently "repairable" via a crafting recipe using the broken glasses and a piece of glass. I'll experiment with using the breakage and repairable

Content.Shared/Clothing/Components/DamageOnSlipComponent.cs
```diff
@@ -0,0 +1,37 @@
...
+/// Pairs best with a `Destructible` component
+/// </remarks>
+[RegisterComponent]
+public sealed partial class DamageOnSlipComponent : Component
+{
+    /// <summary>
+    /// Chance the clothing will be damaged when slipped
+    /// </summary>
+    [DataField]
+    [ViewVariables(VVAccess.ReadWrite)]
```
slarticodefast, line 18:
```suggestion
```
DataField includes VV redwrite access nowadays. Old files still do this manually though.

Content.Shared/Clothing/Components/DamageOnSlipComponent.cs
```diff
@@ -0,0 +1,37 @@
...
+    public float DamageChance = 0.01f;
+
+    /// <summary>
+    /// Damage per instance of unlucky slip
+    /// </summary>
+    [DataField(required: true)]
+    [ViewVariables(VVAccess.ReadWrite)]
```
slarticodefast, line 25:
```suggestion
```

Content.Shared/Clothing/Components/DamageOnSlipComponent.cs
```diff
@@ -0,0 +1,37 @@
...
+    public DamageSpecifier Damage;
+
+    /// <summary>
+    /// Damage multiplier maximum
+    /// </summary>
+    /// <remarks>
+    /// Will multiply the damage specifier by a random integer from 1 to maximum (non inclusive)
+    /// </remarks>
+    [DataField]
+    [ViewVariables(VVAccess.ReadWrite)]
```
slarticodefast, line 35:
```suggestion
```

Content.Shared/Clothing/Components/DamageOnSlipComponent.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    [DataField(required: true)]
+    [ViewVariables(VVAccess.ReadWrite)]
+    public DamageSpecifier Damage;
+
+    /// <summary>
+    /// Damage multiplier maximum
+    /// </summary>
+    /// <remarks>
+    /// Will multiply the damage specifier by a random integer from 1 to maximum (non inclusive)
```
slarticodefast, line 32: why a random integer and not a float?

Content.Shared/Clothing/EntitySystems/DamageOnSlipSystem.cs
```diff
@@ -0,0 +1,36 @@
...
+public sealed class BreakOnSlipSystem : EntitySystem
+{
+    [Dependency] private readonly IRobustRandom _random = default!;
+    [Dependency] private readonly INetManager _net = default!;
+    [Dependency] private readonly DamageableSystem _damageableSystem = default!;
+
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<DamageOnSlipComponent, InventoryRelayedEvent<SlippedEvent>>(OnSlip);
+    }
```
slarticodefast, line 19:
```suggestion
    {
        base.Initialize();

        SubscribeLocalEvent<DamageOnSlipComponent, InventoryRelayedEvent<SlippedEvent>>(OnSlip);
    }
```

Content.Shared/Inventory/InventorySystem.Relay.cs
```diff
@@ -40,6 +40,7 @@ public void InitializeRelay()
         SubscribeLocalEvent<InventoryComponent, TargetBeforeHyposprayInjectsEvent>(RelayInventoryEvent);
         SubscribeLocalEvent<InventoryComponent, SelfBeforeGunShotEvent>(RelayInventoryEvent);
         SubscribeLocalEvent<InventoryComponent, SelfBeforeClimbEvent>(RelayInventoryEvent);
+        SubscribeLocalEvent<InventoryComponent, SlippedEvent>(RefRelayInventoryEvent);
```
slarticodefast, line 43: best move this directly below `SlipAttemptEvent`

Resources/Locale/en-US/clothing/glasses.ftl
```diff
@@ -0,0 +1 @@
+﻿clothing-glasses-shatter = The glasses shatter
```
slarticodefast, line 1:
```suggestion
﻿clothing-glasses-shatter = Your glasses shatter.
```

Resources/Prototypes/Entities/Clothing/Eyes/glasses.yml
slarticodefast: I think making them repairable would be the nicer way to do this rather than adding a crafting recipe specifically for glasses.

slarticodefast (changes requested):

Resources/Locale/en-US/clothing/glasses.ftl
thetolbean: I used `The` instead of `Your` because this popup is shown to everyone, and this popup will also show if the glasses are broken even when not equipped, including if they are thrown or are caught in an explosion. It saying `"Your glasses shatter."` when a random pair of glasses break would be very confusing

slarticodefast: Ah, makes sense.
Add the dot at the end though.

Content.Shared/Clothing/Components/DamageOnSlipComponent.cs
thetolbean: My intention was to give the possibility of slipping once to shatter your glasses despite not damaging them previously by doubling the damage. This only required a multiplier of 1 or 2. Otherwise, it very much could be a float.

ArtisticRoomba:
Let us know when you come back to it.

PR #34205, opened 2025-01-03
Make reagent dispensers gridinv-based instead of pseudo-listinv

sowelipililimute:
This simplifies the code and makes the experience of examining contents easier without the reagent dispenser UI, as well as adding the possibility for dispensers to have items of heterogeneous sizes in them, which would allow configuring reagent dispensers to accept smaller containers such as beakers or vials in order to allow for more types of smaller quantities of reagents, or other flexibilities brought by using a standard storage component.

## About the PR

- reagent dispensers now use a standard StorageComponent to have a gridinv-based system for inserting containers instead of the custom quasi-listinv thing accomplished by generating a bunch of item slots at runtime
- the reagent dispenser components no longer care about duties handled by other components, such as whitelisting the slots or prefilling the reagent dispenser, so this is trimmed out

## Why / Balance
Codewise, having a custom approximation of a normal storage thing via a bunch of runtime-generated item slots identified by strings is somewhat smelly, and using the standard storage component allows code to be reduced.

UIwise, this allows examining a reagent dispenser's contents to look like this:
grafik
instead of this:
grafik

Balancewise, not much has changed, except for a slightly larger capacity brought on by the grid and smaller bottles in the booze and soda dispensers take less space than they did before.

## Technical details
- remove whitelist and storage slots handling from ReagentDispenser system/component
- offload that to the StorageComponent
- remove the now-unused reagent dispenser inventory prototype

## Requirements

## Breaking changes
- `ReagentDispenser` no longer has `NumSlots` or `StorageWhitelist`, configure a `Storage` on your entity prototype instead
- `ReagentDispenser` no longer has `PackPrototypeId` and `ReagentDispenserInventory` no longer exists, use a `StorageFill` component to preload a reagent dispenser

**Changelog**

:cl:
- tweak: Reagent dispensers (Chem Dispenser, Booze Dispenser, Soda Dispenser) now use a grid inventory for their internal storage


Alpaccalypse:
Having been looking into this on the .yml side and in the game, recently, I have to say, I can't see a downside to this. Looks like a desirable update. Unfortunately, I'm not really experienced enough to give a credible review on the code, or I would.

Content.Server/Chemistry/EntitySystems/ReagentDispenserSystem.cs
```diff
@@ -82,32 +86,31 @@ private void UpdateUiState(Entity<ReagentDispenserComponent> reagentDispenser)
 
         private List<ReagentInventoryItem> GetInventory(Entity<ReagentDispenserComponent> reagentDispenser)
         {
+            if (!TryComp<StorageComponent>(reagentDispenser.Owner, out var storage))
+            {
+                return [];
```
ciaran, line 91: Should there be some logging here?

Content.Server/Chemistry/EntitySystems/ReagentDispenserSystem.cs
```diff
@@ -122,22 +125,28 @@ private void OnSetDispenseAmountMessage(Entity<ReagentDispenserComponent> reagen
 
         private void OnDispenseReagentMessage(Entity<ReagentDispenserComponent> reagentDispenser, ref ReagentDispenserDispenseReagentMessage message)
         {
+            if (!TryComp<StorageComponent>(reagentDispenser.Owner, out var storage))
+            {
+                return;
```
ciaran, line 130: And here

ciaran (approved):
LGTM but I wonder if the precedent is to log if a component is missing

Content.Server/Chemistry/EntitySystems/ReagentDispenserSystem.cs
sowelipililimute: this is not usually done

Content.Shared/Chemistry/SharedReagentDispenser.cs
```diff
@@ -66,11 +67,22 @@ public ReagentDispenserSetDispenseAmountMessage(String s)
...
-        public ReagentDispenserDispenseReagentMessage(string slotId)
+        public ReagentDispenserDispenseReagentMessage(ItemStorageLocation storageLocation)
         {
-            SlotId = slotId;
+            StorageLocation = storageLocation;
+        }
+    }
+
+    [Serializable, NetSerializable]
+    public sealed class ReagentDispenserEjectContainerMessage : BoundUserInterfaceMessage
```
beck-thompson, line 79: add a comment!

Content.Server/Chemistry/EntitySystems/ReagentDispenserSystem.cs
beck-thompson: I think its fine either way!

beck-thompson (changes requested):
OK looks good! I don't have time right now to track down this specific issue but there seems to be some weird prediction stuff between syncing up the UI and the storage. It seems to be like always one behind of what's actually in the storage (but only sometimes?) I can't quite figure out exactly whats wrong!

beck-thompson (approved):
Looks good! That was an odd bug

PR #34221, opened 2025-01-05
Theater Sign Generalization

ps3moira:
## About the PR

Removed the three theater signs and replaced them with one single theater sign.
## Why / Balance

I discussed on the discord asking why there was three different theater signs for one of each job and thought it would be better if it was more general and not job specific for a department sign.
## Media

image
image
image

**Changelog**
:cl:
- add: Added new theater sign
- remove: Removed mime, clown, and musician theater signs


ArtisticRoomba:
I kinda liked the previous doodled/drawn on sign. Gave it real ss14 vibes as compared to a regular sign, and it fit with the OOC description as well. But this sprite is also good/

IamVelcroboy:
Why?

ps3moira:
> Why? Edit: I like it. But why remove the random others ones? is what I meant to say.

To more clearly show where the theater room is. The multiple ones make it confusing for players if it's looking like one of the individual jobs in the sub-department.

Alpaccalypse:
> > Why? Edit: I like it. But why remove the random others ones? is what I meant to say.
> 
> To more clearly show where the theater room is. The multiple ones make it confusing for players if it's looking like one of the individual jobs in the sub-department.

I'm not sure what policy is, so this might be completely wrong, but might it be better to leave the other sign .png in the folder and just not reference them, rather than remove them. I'd agree that you don't really want 100s of .pngs that are unused lying around, but these seem like they could likely still have some use in a future PR.

Other than that, I think your changes make sense, and don't really take anything away from the game at this point.

Boaz1111:
Changelog should be clear on multiple signs being removed

ps3moira:
> Changelog should be clear on multiple signs being removed

Is this better?

Boaz1111:
Yes

EmoGarbage404 (changes requested):
I would be fine merging this is an alternative theater sign but i like the trio too. Could you just add this as another sign and keep the old ones? Maybe have it say "stage" or something slightly different?

TheShuEd:

PR #34224, opened 2025-01-05
Add ItemSlot Inserted and Ejected events

lzk228:
## About the PR

add two new events that are called after an item successfully inserted/ejected in/from an entity with ItemSlotsComponent

## Why / Balance

need it in downstream
good feature

## Technical details

just two new events

## Media

no

## Requirements

## Breaking changes

no 

**Changelog**

no

### DIFF
--- a/Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
+++ b/Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
@@ -269,13 +269,7 @@ namespace Content.Shared.Containers.ItemSlots
 
                 if (slot.InsertSuccessPopup.HasValue)
                     _popupSystem.PopupClient(Loc.GetString(slot.InsertSuccessPopup), uid, args.User);
-
                 args.Handled = true;
-
-                var ev = new ItemSlotInsertedEvent(uid, args.Used, args.User, slot);
-                RaiseLocalEvent(uid, ref ev);
-                RaiseLocalEvent(args.Used, ref ev);
-
                 return;
             }
         }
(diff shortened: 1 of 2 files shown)


deltanedas (changes requested):
add to doc that its raised on both the slot entity and item

TheShuEd (approved):

Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
```diff
@@ -271,6 +271,11 @@ private void OnInteractUsing(EntityUid uid, ItemSlotsComponent itemSlots, Intera
                     _popupSystem.PopupClient(Loc.GetString(slot.InsertSuccessPopup), uid, args.User);
 
                 args.Handled = true;
+
+                var ev = new ItemSlotInsertedEvent(uid, args.Used, args.User, slot);
+                RaiseLocalEvent(uid, ref ev);
```
slarticodefast, line 276: This should be moved to the `Insert` function. Currently it would not be raised when inserted via the alt verbs or other methods calling the insert function.

slarticodefast (changes requested):

0x6273:
> need it in downstream
> good feature

Can you explain what this is needed for? Why can't you use the existing storage events?

lzk228:
> > need it in downstream
> > good feature
> 
> Can you explain what this is needed for? Why can't you use the existing storage events?

i need to change variables of the SlotEntity, but ItemSlotInsertAttemptEvent runs event for every slot, so if i have 4 slots, it runs 4 times when i inserted only one item, the same with Eject i suppose, so i need events that are called after successfully inserting an item

0x6273:
> > > need it in downstream
> > > good feature
> > 
> > 
> > Can you explain what this is needed for? Why can't you use the existing storage events?
> 
> i need to change variables of the SlotEntity, but ItemSlotInsertAttemptEvent runs event for every slot, so if i have 4 slots, it runs 4 times when i inserted only one item, the same with Eject i suppose, so i need events that are called after successfully inserting an item

Why can't you use the `EntInsertedIntoContainerMessage` event for this?

lzk228:
damn, i didn't know about this one

Issue #34031, opened 2024-12-23
mini jetpacks are the same size as normal jetpacks 


Carolyn3114:
put jetpack in backpack, 4x4 size, put mini jetpack in backback, 4x4 size
it is not miniature

Errant-4:
This sounds more like a name issue than an inventory issue. Do mini jetpacks actually NEED a buff? If not, I think it  would be more reasonable to just rename them to something else to resolve the discrepancy between the name and the item size

Errant-4:
Maintainer consensus is that mini jetpacks are powerful enough to warrant their large inventory size, they already have advantages over the "normal jetpack", they do not need to also be easier to carry in a backpack. 

If someone still feels this is an issue that must be fixed, then propose a better name for the mini jetpack

PR #34228, opened 2025-01-05
Make mini jetpack actually mini

NoElkaTheGod:
## About the PR
Mini jetpack is now 2x4 (regular jetpacks are 4x4).
Void jetpack is also 2x4, because it goes into the same slots as a mini jetpack, and is exclusive to CE, so i think it makes sense.

## Why / Balance
Logic. Also resolves https://github.com/space-wizards/space-station-14/issues/34031.

## Technical details
YAML changes.

## Media
Not really.

## Requirements

**Changelog**
:cl:
- tweak: Mini jetpacks and void jetpacks are now actually smaller than a normal jetpack

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Tools/jetpacks.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/jetpacks.yml
@@ -187,6 +187,8 @@
   components:
     - type: Item
       sprite: Objects/Tanks/Jetpacks/mini.rsi
+      shape:
+        - 0,0,3,1
     - type: Sprite
       sprite: Objects/Tanks/Jetpacks/mini.rsi
     - type: Clothing
@@ -259,6 +261,8 @@
   components:
   - type: Item
     sprite: Objects/Tanks/Jetpacks/void.rsi
+    shape:
+      - 0,0,3,1
   - type: Sprite
     sprite: Objects/Tanks/Jetpacks/void.rsi
   - type: Clothing


pheenty:
If anything, balance-wise I would make regular jetpacks smaller, mini-jetpacks are fine as they are

NoElkaTheGod:
> If anything, balance-wise I would make regular jetpacks smaller, mini-jetpacks are fine as they are

So you're saying that mni jetpacks should be larger than normal jetpacks? That doesn't make any sense at all. They're called "mini" for a reason.

pheenty:
> > If anything, balance-wise I would make regular jetpacks smaller, mini-jetpacks are fine as they are
> 
> So you're saying that mni jetpacks should be larger than normal jetpacks? That doesn't make any sense at all. They're called "mini" for a reason.

Well yes it doesn't, but again, balance-wise mini-jetpacks are much better than regular ones already. Also you still basically never (at least you shouldn't) wear them outside of belt or suit storage anyway

NoElkaTheGod:
> > So you're saying that mni jetpacks should be larger than normal jetpacks? That doesn't make any sense at all. They're called "mini" for a reason.
> 
> Well yes it doesn't, but again, balance-wise mini-jetpacks are much better than regular ones already. Also you still basically never (at least you shouldn't) wear them outside of belt or suit storage anyway

They aren't "better", both have their uses. If you're juat a salvager who needs to go to and from wrecks, mini-jet is perfectly fine, but you really don't want to run out of fuel while chasing a dragon in space, and mini-jets have a rather low volume. Mini-jets are considered "better", because:
1. Having to fight someone in space is rather rare, and there aren't many other situations where a full jetpack is needed.
2. To properly use a normal jetpack you need a lot of preparations (put your gun into suit storage, all the ammo into your belt, and dispose of everything you won't need in a fight), whereas a mini-jet you can simply put on a belt/suit slot and call it a day.
3. They're just cheaper.

As for "you never store the mini-jet in your bag", well, that's what this PR is aiming to solve. Now it xan actually be viable to put your jetpack in your bag when you're not using it, and use belt/suit slots for something else (like a gun, most of which are 4x4)

ruddygreat:
fwiw void jetpacks aren't exclusive to the CE, 2 of the roundstart savlage wrecks have them.
though also it does still make sense for them to be 2 * 4, they still go in the belt slot and all that.

Errant-4:
Thank you for taking the time to contribute to the codebase, but I am closing this PR (along with the issue it was adressing.)
Maintainer consensus is that mini jetpacks are powerful enough to warrant their large inventory size, they already have advantages over the "normal jetpack", they do not need to also be easier to carry in a backpack.

PR #32423, opened 2024-09-23
Change the syndicate charge to start a timer on signal

SaphireLattice:
## About the PR
Made the charge start the timer countdown, rather than getting detonated instantly

Mutually exclusive with #32429 that instead changes minibomb to start the timer, but keeps C4 instant. For maximum fun, consider them non-exclusive.

## Why / Balance
Minibomb + C4 can not be dodged, predicted or counterplayed by any way other than extreme paranoia towards anyone being nearby. If the instant detonation succeeds, which depends on item dropping skills of the player and the target being in the right spot (e.g. distracted by roleplaying and typing), the target will be not just killed but completely gibbed. This shouldn't be something that any traitor can do with a minibomb after spending just extra 2 TC. At this point just give them a one-time-use throwable instagib Throngler for 8 TC. Same deal.

It affects the instant C4 strategy of nukies, but pre-cooking the C4 sounds pretty doable. Does make it less of an instant grenade though.

Alternative to this PR would be to give the minibomb way more explosive resistance, make it start a timer when damaged by an explosion, and just completely break without explosion if it gets damaged to destruction. Rather than the current behavior of deleting itself explosively.

## Technical details
Feels like shitcode, basically copy-paste of the existing `TriggerOnSignalComponent` code, but with `StartTimer` instead of `Trigger`. Unsure if I got the `EntityUid user` right for that method though. Also added new localisation strings, woo!

## Media
It's visually minimal but can attach a recording later.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
:cl:
- tweak: Syndicate C4 now starts a countdown on signal, rather than exploding instantly.

PR #32429, opened 2024-09-24
Change minibomb to be explosion resistant and start timer on damage

SaphireLattice:
## About the PR
Prevents C4+Minibomb combo from being instant and impossible to do anything about without excessive paranoia. Same idea as #32423, but done from the other end. Mutually exclusive with it, unless maintainers will want to go for maximum fun.

For even more fun, see #32428 that makes the container fly into a random direction before the second explosion.

## Why / Balance
See #32423, but without the note about nukies. Instagib throwable shouldn't be near impossible to counterplay *in an RP game full of distractions lasting 5+ seconds*. This isn't a twitch shooter.

## Technical details
Creates a new damage threshold behavior that works on explosives and causes them to start counting down. In combination with extreme explosive resistance, this allows for a bomb that's not deleted and is counting down, after being triggered by another explosive. Current version of this PR only does this to minibomb, though. So no funny business with C4, unless someone requests that. Also it will just disappear without exploding if damaged too much. Gotta be careful!

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- tweak: Minibomb is now explosion resistant and will start counting down if damaged by a C4

PR #34230, opened 2025-01-05
Remove Signal Trigger, adds signal to Timer Trigger 

K-Dynamic:
## About the PR

- Removes Signal Trigger for modular grenades
- Adds signal to Timer Trigger, renames Timer Trigger to Timer-Signal Trigger

## Why / Balance

Directly nerfs remote toolbox bombs by giving crew/antags time to escape the toolbox or throw it back before it detonates. Remote bombs are still possible as the Timer Trigger can accept signals, which activates a countdown to trigger payloads.

Nerf applies similar principle to C-4 being given a delay rather than instant detonation (#32423).

Does not affect voice triggers, which can still instantly detonate upon stating a designated codeword. Unfortunately I am unable to add a timer delay on voice triggers with my current skillset.

## Technical details

yml changes only
- `Resources/Prototypes/Entities/Objects/Devices/Electronics/triggers.yml` (removes Signal Trigger and adds timer device link to Timer Trigger)

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Removed Signal Triggers for modular grenades.
- tweak: Timer Triggers are renamed to Timer-Signal Triggers as they can now accept signals, which activates a countdown.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/Electronics/triggers.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/Electronics/triggers.yml
@@ -20,6 +20,9 @@
     size: Small
   - type: StaticPrice
     price: 40
+  - type: Tag
+    tags:
+    - TimerSignalTrigger
   - type: PayloadTrigger
     components:
     - type: OnUseTimerTrigger
@@ -39,30 +42,6 @@
     - type: DeviceLinkSink
       ports:
       - Timer
-
-#- type: entity
-#  parent: TimerTrigger
-#  id: SignalTrigger
-#  name: signal trigger
-#  description: Adds a machine link that is triggered by signals.
-#  components:
-#  - type: Sprite
-#    sprite: Objects/Devices/signaltrigger.rsi
-#    state: signaltrigger
-#  - type: StaticPrice
-#    price: 40
-#  - type: Tag
-#    tags:
-#    - SignalTrigger
-#  - type: PayloadTrigger
-#    components:
-#    - type: TriggerOnSignal
-#    - type: DeviceNetwork
-#      deviceNetId: Wireless
-#      receiveFrequencyId: BasicDevice
-#    - type: WirelessNetworkConnection
-#      range: 200
-#    - type: DeviceLinkSink
 
 - type: entity
   parent: BaseItem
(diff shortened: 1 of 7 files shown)


TytosB:
bad change. instant bombs are the great equalizer, and one such items thats not antag-restricted needs to exist.

TytosB:
perhaps, but warops is already busted, the nukies do not need more help in that department. if they want to do warops then let them suffer the consequences of doing so.

Hrosts:
> This already exists (https://github.com/space-wizards/space-station-14/pull/32429), I just felt this approach is 'better' since the other would introduce too long of a delay (3 seconds + another 3 seconds). I also don't like the idea of making bombs unpredictable.

That PR doesn't solve the chain reaction toolbox issue, as you know, and it's your approach which introduces a longer delay - in case of "unstable chain reaction", you would get some damage in immediately, as the first grenade explodes on a signal, then the rest will follow, while in your case all blows up later, making it _less_ reliable, but also less chaotic. Both less reliable and less chaotic are bad for PVP combat, IMO.

Your change will also nerf or mess with other uses for modular grenades - putting in a chem charge with any dangerous foam/smoke mix would be pointless, as all potential victims will run away before it detonates. Making plushie-on-signal or signal-on-signal grenades will be impossible.

Also, in what cases is it even valuable to activate a timer remotely for a modular grenade? Most of the time it would be easier to just throw a timer trigger grenade, removing one step. Is there even a point in such trigger?

It will also fail to completely address the chain reaction issue - you will still be able to record a phrase on a voice trigger, chug a toolbox, then Ctrl+V the trigger into the chat for the instant explosion. That, in turn, will require the voice trigger to also be reworked into a voice-timer trigger, which is ridiculous, removes the oomph and the coolness and much of the point of a voice trigger and also kills many non-explosive-charge uses in the process.

Instead of giving players less control over when and where the explosion happens, make them have less control over the maximum amount of damage it can deal when not reacted to on time. Using bags/toolboxes full of explosive charges isn't technically an intended way to use those charges, so make them not behave their best when such things are done. The "by the book" uses remain predictable and good, the messy duct taped together solutions - less so.

Errant-4:
I don't think the signal trigger should be changed like this, because it can be used for things other than triggering explosives - why change the functionality there?

PR #32711, opened 2024-10-09
Holopads

chromiumboy:
## About the PR

This PR introduces the holopad. The holopad is a semi-private communication device that can be used by both the station AI and the station's crew members. As interactions with this device are somewhat complex, I've listed them all out here

_**AI usage of holopads**_
- Holopads provide vision to AIs, just like surveillance cameras do
- The AI can activate an idle holopad, projecting themselves as a hologram
- Upon activating a holopad, the AI's vision is shifted from its free-roaming eye to that of the holopad, seeing things from its perspective
- The AI is immobile in this state but can still interact at a distance with white-listed devices within its new line of sight
- The AI's local chat will be relayed to the holopad, letting them speak with those nearby. They will also be able to hear crew members talking around the holopad while it is active
- The AI, or another crew member, can use the holopad UI to end the call and return the AI to its free roam mode
- AIs can also answer a holopad that another crew member is ringing; the AI will be transferred to the calling holopad and start projecting from it
- The AI can also initiate an emergency broadcast from a holopad (see below)
- The AI can only project itself from a single holopad at a time, unless it has triggered an emergency broadcast

_**Crew usage of holopads**_
- Crew can use the holopad in three ways: to request the presence of the station's AI (see below), to call another holopad, or to initiate an emergency broadcast (see below)
- To call another holopad, the user selects the desired receiver from a list that is auto-populated with holopads that are in its communication range (see below)
- When a call is placed to a holopad, if it is already in use or unpowered, the caller will get a busy tone. Otherwise, the receiving holopad will start to ring
- If the station AI or crew member interacts with a ringing holopad, they will see the ID of the person making the call and can elect to answer or hang up
- If no response is received within 30 seconds of a call being placed, the call will end automatically
- If a call is accepted within the time limit, holograms of the caller and the recipient will appear on their counterpart's holopad and the two crew members can converse with each other
- Speech from other people nearby (within 4 tiles) will also be transmitted; the appropriate hologram will be automatically updated to match the appearance of the last speaker
- If a user moves too far away from their holopad, their hologram will be replaced with a generic 'in-call' hologram to signal to the person on the other end that they may be speaking with empty air

_**Requesting the AI**_
- If the AI is already using a holopad when a request for the AI is submitted the caller will get a busy signal
- Otherwise, the AI player will get a small pop-up in the bottom right-hand corner of the screen
- The pop-up lists the ID of the caller and an answer and end call button
- If the AI answers the call, it will be transferred to the calling holopad and active its projector
[...]

PR #34248, opened 2025-01-06
fix: Allow Station AI to hear speech during regular calls

DisposableCrewmember42:
## About the PR
The AI can now hear speech during regular holopad calls, but not broadcasts.

## Why / Balance
It seemed unintentional for the AI not to be able to hear people who call it, or when using the regular projection feature.

**Edit**: It works here, for some reason. It shouldn't. See comments.

## Technical details
I removed the attribute deafening the AI. Copy-paste mistake?

## Media
None

## Requirements

## Breaking changes

**Changelog**

No CL. Internal change. It shouldn't work here, but it does. Downstream, the setting correctly mutes the receiving end when built in Release configuration, and only in Release configuration. Here, it doesn't mute the receiver irrespective of Configuration.

### DIFF
--- a/Content.Server/Holopad/HolopadSystem.cs
+++ b/Content.Server/Holopad/HolopadSystem.cs
@@ -705,8 +705,7 @@ public sealed class HolopadSystem : SharedHolopadSystem
 
         var callOptions = new TelephoneCallOptions()
         {
-            ForceConnect = true,
-            MuteReceiver = true
+            ForceConnect = true
         };
 
         _telephoneSystem.CallTelephone(source, receiver, user, callOptions);


DisposableCrewmember42:
Very much unintentional. According to #32711:
> The AI's local chat will be relayed to the holopad, letting them speak with those nearby. They will also be able to hear crew members talking around the holopad while it is active

ScarKy0:
I just tested on master and it doesn't seem to be an issue?

DisposableCrewmember42:
Odd. You're right. It isn't an issue here, which I don't think makes sense, since the receiver is supposed to be muted with the current settings...

It does certainly cause issues downstream, which are fixed by setting MuteReceiver to false, so I assumed it would be the same here.

DisposableCrewmember42:
Sure enough, it also seems to be ignored during Emergency Broadcasts.

ScarKy0:
Confusing stuff...

DisposableCrewmember42:
So, downstream it seems like the settings are ignored in Debug configuration but work properly in Release.

Here, the settings are ignored in both Debug and Release.

chromiumboy:
This is the first time I'm hearing of this issue from downstream. Who's being affected?

I don't think this is the correct fix as it may lead to duplicated messages upstream. I'm out of town atm so I can't test to confirm

When the AI is being projected from the holopad, it hears local speech by subscribing to ExpandICChatRecipientsEvent. The relevant code is in Server.StationAISystem

DisposableCrewmember42:
@chromiumboy 
There have been issues with AI hearing speech on Delta-V.
Here's the [relevant issue](https://github.com/DeltaV-Station/Delta-v/issues/2602). When projecting to a holopad, the AI can hear neither it's own voice nor anyone else's when out of the hearing range of the AI core.

DisposableCrewmember42:
Oh, and, yeah, it causes duplicated messages downstream too, I just somehow didn't notice because I was paying attention to the speech bubbles, so I'll be closing this.

chromiumboy:
> Oh, and, yeah, it causes duplicated messages downstream too, I just didn't notice because I was paying attention to the speech bubbles, so I'll be closing this.

OK, thanks for checking. I can have a look at this bug in a about a week if you need assistance in getting a fix worked out. But I'd look at the station AI code I mentioned above and see if the issue is originating there

deltanedas:
its most likely just a missing bugfix commit yeah

chromiumboy:
> its most likely just a missing bugfix commit yeah

Do you mean delta V is missing a bug fix commit? This was never an issue upstream, so I never made one for this issue

deltanedas:
yes

chromiumboy:
> yes

@deltanedas 

Think I might have an answer. Do the affected Delta V maps have more than one AI core on them?

In OnExpandICChatRecipients, in StationAISystem, the query loop there has two conditions with return statements that should actually be continue statements; I reckon that the empty core is being check first and terminating the loop early

PR #34262, opened 2025-01-07
Fix ChemMaster's weird interactions with heat

sowelipililimute:
## About the PR

This changes the ChemMaster to use the same codepaths as the ChemDispenser and pouring together beakers manually, which makes the behaviour less glitchy-seeming as it follows the same logic that everything else does: adding something cold to something hot makes it colder, and adding something hot to something cold makes it hotter.

## Why / Balance

This makes the behaviour of the ChemMaster more predictable, because it doesn't arbitrarily work differently from everything else to the game that relates to mixing chemicals, especially when there are reactions that may cause the temperature to change in an otherwise unchanging thing. This also means that the energy from hot beakers doesn't magically vanish if they're emptied by the ChemMaster, as that heat energy is conserved instead of ignored in favour of whatever was currently existing.

## Technical details
Instead of specifically adjusting reagent amounts, this makes the ChemMasterSystem use split-solution-with-only and then adds two solutions together.

## Media

https://github.com/user-attachments/assets/ae480bee-e6b4-4872-961e-569c29fcba17

## Requirements

**Changelog**
:cl: 
- tweak: ChemMasters now have proper heat conservation, instead of keeping buffer and inserted container at fixed temperatures

### DIFF
--- a/Content.Server/Chemistry/EntitySystems/ChemMasterSystem.cs
+++ b/Content.Server/Chemistry/EntitySystems/ChemMasterSystem.cs
@@ -38,6 +38,7 @@ namespace Content.Server.Chemistry.EntitySystems
         [Dependency] private readonly StorageSystem _storageSystem = default!;
         [Dependency] private readonly LabelSystem _labelSystem = default!;
         [Dependency] private readonly ISharedAdminLogManager _adminLogger = default!;
+        [Dependency] protected readonly IPrototypeManager _prototypeManager = default!;
 
         [ValidatePrototypeId<EntityPrototype>]
         private const string PillPrototypeId = "Pill";
@@ -139,14 +140,14 @@ namespace Content.Server.Chemistry.EntitySystems
             if (fromBuffer) // Buffer to container
             {
                 amount = FixedPoint2.Min(amount, containerSolution.AvailableVolume);
-                amount = bufferSolution.RemoveReagent(id, amount, preserveOrder: true);
-                _solutionContainerSystem.TryAddReagent(containerSoln.Value, id, amount, out var _);
+                var split = bufferSolution.SplitSolutionWithOnly(amount, id.Prototype);
+                _solutionContainerSystem.TryAddSolution(containerSoln.Value, split);
             }
             else // Container to buffer
             {
                 amount = FixedPoint2.Min(amount, containerSolution.GetReagentQuantity(id));
-                _solutionContainerSystem.RemoveReagent(containerSoln.Value, id, amount);
-                bufferSolution.AddReagent(id, amount);
+                var split = _solutionContainerSystem.SplitSolutionWithOnly(containerSoln.Value, amount, id.Prototype);
+                bufferSolution.AddSolution(split, _prototypeManager);
             }
 
             UpdateUiState(chemMaster, updateLabel: true);
--- a/Content.Shared/Chemistry/EntitySystems/SharedSolutionContainerSystem.cs
+++ b/Content.Shared/Chemistry/EntitySystems/SharedSolutionContainerSystem.cs
@@ -379,6 +379,19 @@ public abstract partial class SharedSolutionContainerSystem : EntitySystem
         return splitSol;
     }
 
+    /// <summary>
+    /// Splits a solution with only the specified reagent(s).
+    /// </summary>
+    public Solution SplitSolutionWithOnly(Entity<SolutionComponent> soln, FixedPoint2 quantity, params string[] reagents)
+    {
+        var (uid, comp) = soln;
+        var solution = comp.Solution;
+
+        var splitSol = solution.SplitSolutionWithOnly(quantity, reagents);
+        UpdateChemicals(soln);
+        return splitSol;
+    }
+
     public void RemoveAllSolution(Entity<SolutionComponent> soln)
     {
         var (uid, comp) = soln;


DaeconWawa:
If I'm reading this right this will hugely alter chemist routines (slow them down a lot) due to this fixing lava beakers which is well frustrating as a chem player, and having three chemists sitting around waiting on the hotplate is gonna be miserable so in that regard more hotplates should be mapped if this goes through.

Qwazori:
> If I'm reading this right this will hugely alter chemist routines (slow them down a lot) due to this fixing lava beakers which is well frustrating as a chem player, and having three chemists sitting around waiting on the hotplate is gonna be miserable so in that regard more hotplates should be mapped if this goes through.

This is not at all frustrating because while heating one chemical, the chemist can parallel produce another that does not require heating, and you can also bring a microwave, for chemicals that do not require strong heating, turning the gameplay of the chemist in a kind of management simulator, which in contrast to what is happening now at least interesting. I'm personally sick of chemists robusters who pour all the chemicals in the chemmaster and then without taking out beakers do just click click click even without using the hotplate, which in my opinion kills the whole point of chemist's gameplay, because you can produce all the drugs in the first 10-15 minutes of game. After all, chemistry is supposed to be fun and with different reactions and ways of producing substances. But I agree that it is desirable to make one hotplate for each one chemist, alternatively you can reduce the boiling point of some drugs

mqole:
> Really excited for whenever this fix goes through, heads up that it breaks the sort function from #34763 ;;

found that after some digging this is because `SplitSolutionWithOnly()` removes reagents from the CM solution, performs the function, then re-adds the reagents in reverse order, which is... not ideal lol. might need to approach this differently.

Qwazori:
@sowelipililimute Hopefully this PR won't be abandoned.

southbridge-fur (approved):
oh hey I can officially approve this now.

2DSiggy:
As a chemist main, I really, really, really don't want this merged. Removing mechanical depth for sake of slowing down a job is not exactly great, especially when said job is depended upon for other jobs to be able to work. People Lavabeaker because the hot plate sucks and it's more fun to. It's unintuitive but I'd rather it be kept instead of removing it and not introducing a replacement for the hot plate's being sucky when trying to heat large contents. Also this would kill bluespace beakers outside of storage.

2DSiggy:
> > If I'm reading this right this will hugely alter chemist routines (slow them down a lot) due to this fixing lava beakers which is well frustrating as a chem player, and having three chemists sitting around waiting on the hotplate is gonna be miserable so in that regard more hotplates should be mapped if this goes through.
> 
> This is not at all frustrating because while heating one chemical, the chemist can parallel produce another that does not require heating, and you can also bring a microwave, for chemicals that do not require strong heating, turning the gameplay of the chemist in a kind of management simulator, which in contrast to what is happening now at least interesting. I'm personally sick of chemists robusters who pour all the chemicals in the chemmaster and then without taking out beakers do just click click click even without using the hotplate, which in my opinion kills the whole point of chemist's gameplay, because you can produce all the drugs in the first 10-15 minutes of game. After all, chemistry is supposed to be fun and with different reactions and ways of producing substances. But I agree that it is desirable to make one hotplate for each one chemist, alternatively you can reduce the boiling point of some drugs

very few meds require not being heated. There's dylo, dex+, bic, arith, and derm. Every other chem that's worth using needs heating. And dylovene is a joke. Microwave heating is a joke too.

ArtisticRoomba:
> People Lavabeaker because the hot plate sucks and it's more fun to.

if the hotplate sucks so much, buff it or make it more interesting (milon is working on turning it into a thermobath but he's on break right now). leaving in this weird buggy behavior because its supposedly more interesting and fun isn't really that good of an idea when the underlying problem should be fixed instead (and made more interesting and fun).

2DSiggy:
> > People Lavabeaker because the hot plate sucks and it's more fun to.
> 
> if the hotplate sucks so much, buff it or make it more interesting (milon is working on turning it into a thermobath but he's on break right now). leaving in this weird buggy behavior because its supposedly more interesting and fun isn't really that good of an idea when the underlying problem should be fixed instead (and made more interesting and fun).

or just wait for that PR to be done to merge this one with.

southbridge-fur:
that PR by the way is https://github.com/space-wizards/space-station-14/pull/37070

OxidisedGearz:
image

Preservation of beaker temperature can be used to produce Infinity Kelvin and NaN Kelvin solutions. Anything that's been heated that far is stuck there and you can't cool it back down even by washing it with sink water, the container is going to be NaN forever. Thankfully this doesnt impact gameplay since NaN permits reactions to occur and there's no adverse health effects to infinitely hot blood, but it generally just feels silly...


Going to concur with a lot of what's already been said by others. I've been playing without the bug (washing beakers with cold water between uses to reset temp to 293) and chemistry still feels fine. Certain recipes (looking at you, pyra) suddenly are much more difficult, but that's not necessarily a bad thing and makes non-advanced medicines more useful. Also you can recycle the leftover heat from the advanced medicines to warm up other reactants. To that end, it raises the skill expression of the role, not lower it. Making temperature requirements matter again is not removing depth.


It does make it obvious that chemlabs often have fewer hotplates than chemists, and requesting extras from science every round may get tedious. It would be nice to have more premade, but similar to how deltaP was rolled out, I don't see a reason why we'd have to update maps before fixing the chemmaster.


I don't think thermobaths should be a prerequisite. Realistically, thermobaths are hotplates that can cool and automatically turn off. For a lot of the concerns people seem to have with the fix, that functionality isn't going to be beneficial with current medicine recipes.


Also worth noting that the list of chemicals not requiring heating neglects a fair few like most botanical supplies, trico, inap, tranex, and cryox. Plus a number of recipes that require heat don't require much, and other recipes are exothermic.

PR #34263, opened 2025-01-07
Update vessel_warning.ogg

Pangogie:
Remove DC offset and apply short fade out.

I hate this beep. Make the beep slightly less bad by having it not cause DC offset clicking.


slarticodefast (changes requested):
Yep, sounds much better to me. Please mention that you modified the file in the corresponding attributions.yml file.

PR #33160, opened 2024-11-04
Half Revert #31978

cohanna:
## About the PR
~~The ability to upgrade reinforced windows to shuttle windows has been removed.~~

~~**You can still construct shuttle windows from scratch**~~
Partially reverts #31978 

## Why / Balance
~~This specific upgrade is largely used to upgrade the windows to brig and command offices which is very powergamey and
not very cool. Itd be one thing if these were meant to be destroyed but its 500 hp is comparable to a reinforced wall which makes it unreasonable for a 2 plasteel upgrade with no tools required.~~
Revert was requested by @keronshb in comments below

## Technical details
~~removed the upgrade edge on the reinforced window graph~~
Partially reverts #31978 

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: clinux
- remove: Normal and reinforced windows can no longer be directly upgraded using rods, plasma, uranium, or plasteel.

PR #31978, opened 2024-09-08
Add easy window upgrading, fix some other window issues.

Blackern5000:
## About the PR
**First off, the main thing:**
Windows can now have rods/plasma/uranium/plastic added to them to quickly and easily upgrade them to reinforced/plasma/uranium/tinted versions. Plasteel can also turn reinforced windows into shuttle windows.
**_ENGINEERS REJOICE!!!_**

**Secondly:**
Window HP is now a bit more standardized and not completely all over the place.
- window: 50 (Glass)
- tinted window: 50 (Glass) (these use plastic now to allow them to fit in the upgrade tree alongside reinforced windows)
- plasma window: 75 (Glass)
- uranium window: 75 (Glass)
- clockwork window: 75 (RGlass)
- reinforced window: 75 (RGlass)
- reinforced plasma window: 150 (RGlass)
- reinforced uranium window: 150 (RGlass)
- shuttle window: 500 (RGlass)

Thindows have half the HP and the same damage mods as their full version. (600 HP reinforced plasma glass thindow is dead)

**Thirdly:**
Window damage visuals are fixed, for awhile now they've displayed very strangely due to not having their values updated when windows got their HP nerfed. Visuals display at 20% damage, 40% damage, and 80% damage now.

## Why / Balance
So basically, trying to upgrade ANYTHING as an engineer has just been absolutely horrible. You'd first need to get the materials, craft them together, remove the old window, possibly mess with some grilles and float around in space, and chances are you won't even get everything done in time. Now you can get the materials, bring them to the place you want to upgrade, and now it's just a few doafters and boom, done! Engineers can pre-emptively upgrade windows before meteors hit evac now without going through an hour and a half of pain!

## Media

https://github.com/user-attachments/assets/b22d8715-5f00-4bfb-8a93-59929b013add

## Requirements
- [X] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes
**Changelog**
:cl:
- add: Normal and reinforced windows have been made directly upgradable using rods, plasma, uranium, or plasteel.
- tweak: Reinforced plasma and uranium windows have been made tougher.
- fix: Windows have been made to correctly display their damage visuals.
[...]

PR #34268, opened 2025-01-07
Allow upgrading windows by adding materials halfway through deconstruction

Blackern5000:
## About the PR
You can now upgrade windows similarly to #31978, which was removed by #33160, however now there are constructions steps involved so it's not just magically applying materials for stronger window. These steps are the same for all windows:
Start the regular deconstructions steps, but halfway through add the material for upgrading. Upon completing the deconstruction steps the window will now upgrade instead of deconstructing.
The time needed to complete the latter half of the steps after adding the material is decreased (except for the shuttle window upgrade) only if the material was added, so it is significantly more convenient than fully deconstructing and building the new window in place of the old one.

I was thinking of adding a tip for this since it's a pretty hard feature to come across on your own, but I'm not sure how to word it well. I'm open to suggestions.

(Also adds directly building directional uranium windows since uhh that wasn't a thing for some reason. Also directly building frosted windows but they don't function yet.)

## Why / Balance
Quote from my original PR for window upgrading:
> So basically, trying to upgrade ANYTHING as an engineer has just been absolutely horrible. You'd first need to get the materials, craft them together, remove the old window, possibly mess with some grilles and float around in space, and chances are you won't even get everything done in time.

With the original pr, upgrading windows wasn't just convenient, it was pathetically easy. You could just walk right up to a window, apply rods/plasma/plasteel, and suddenly in just 5 seconds or less you have a SIGNIFICANTLY stronger window for a fraction of the time investment.
With this, it's still much faster and much easier (the entire point of having this, so engineers don't have to struggle as much trying to pre-emptively guard against station damage such as meteors.), but you still need to use welding tools for upgrading reinforced windows and can't just slap some plasteel onto the permabrig, cell windows, or sec front door as hos or warden and make it effectively immune to escape attempts, riots, or break-ins.

## Media

https://github.com/user-attachments/assets/7eeaaa3e-b8f1-45ad-9fb1-19ad8fbc78e0

## Requirements

**Changelog**
:cl:
- add: Normal and reinforced windows have once again been made directly upgradable by adding rods, plasma, uranium, or plasteel exactly half-way through the deconstruction process of the window being upgraded.
- add: You can now build directional uranium windows and directional reinforced uranium windows from the construction menu.
- add: You can now build directional frosted windows from the construction menu, despite them being non-functional and only acting as regular directional windows.


dffdff2423 (approved):
Did not see anything wrong with the code. However, I have not really done much with construction graphs. I think this is a great QOL feature.

Blackern5000:
FINALLY got around to fixing the conflicts in this

AJCM-git:
I think this is fine conceptually, could argue about at which point in the graph the upgrade should be possible. Though this is very intuitive behavior so at minimum there should be some visual feedback when the upgrade is possible. Adding that would be outside the scope of this though, so closing. Feel free to reopen if thats figured out

PR #34270, opened 2025-01-07
Darkened Service job interface icons for better contrast

frobnic8:
## About the PR
The service job interface icons had low contrast, so this PR changes the ID cards and interface job icons to use the next darker shades from the prototypes advanced palette.

## Why / Balance
This should have no impact on game balance.

## Technical details
The colors were taken from `Resources/Prototype/Palettes/advanced.yml`. The current colors were replaced with the next two darker shades from the service palette.

## Media
DarkerServiceIcons

## Requirements

## Breaking changes

This should have no breaking changes.

**Changelog**

This is not worth a changelog entry.


frobnic8:
The outstanding question for me is if all the map changes are correct (do some of those greens just HAPPEN to match service green without being service related?) and if there are any other areas where Service green is used outside the resources.

lzk228:
> The outstanding question for me is if all the map changes are correct (do some of those greens just HAPPEN to match service green without being service related?) and if there are any other areas where Service green is used outside the resources.

i think it's better to change just the id cards and job icons

slarticodefast:
I don't think manually editing all the maps is a good idea, it will cause way too many merge conflicts.

Alpaccalypse:
The change was good though, just a few shades darker makes all the difference.

Alpaccalypse:
> The outstanding question for me is if all the map changes are correct (do some of those greens just HAPPEN to match service green without being service related?) and if there are any other areas where Service green is used outside the resources.

Not sure if it even uses the same shade anyway, but if you think the difference is enough to be worth bothering with, service signs use service green too.

janitor
chapel

frobnic8:
I'll work on reverting the map changes and hitting just the ID cards. I'll also try to double check the signs. I think the change in the guidebook for the terms like "Service" and "Head of Personnell" are keepers, though.

frobnic8:
After checking `Resources/Prototypes/Palattes/advanced.yml` I've adjusted the images to use it's "service black" color for the shade and made it's "service dark" color match what I was calling the "light" color here. While it might be "more correct" to adjust the whole palette for service in there, it feels like too much of a stretch for the others. 

Maps reverted, along with a couple other small file changes that were not service related. Left the Guidebook and the ones that were clearly labelled as service in the entities, palettes, and roles under prototypes.

frobnic8:
Okay, after force-pushing to undo a small squash since that's not cool here (yay git!), I realized I should just use the existing darker shades in the `Prototypes/Palettes/advanced.yml` instead of the colors I picked, since they are close enough, and it means changing less, and it means I can leave the question about the other files and their service colors for another PR.

Alpaccalypse (approved):
Being eminently more qualified to review this than the rest of you uncultured peasants, I, Alpaccalapse, approve of this tasteful and discenring update to the game. (I see no errors and this sets out what it aimed to do).

Alpaccalypse:
> It's so much darker. Compare it with all the other departments and it creates a jarring difference. It's dull.

As am accessbility issue, I think it is fair to make some concessions. If you look at other departments the shade was actually more mismatched before than it is now. I actually think it was more jarring before.

If others think this is still an issue though, I think that either upping the tone, or changing the hue is the way to go, otherwise we end up back with the contrast being an issue again. Only other way around the contrast issue is to change the white to a grey, which would make service icons stand out from the others even more.

onesch:
maybe a little lighter?

frobnic8:
Using colors other than those already defined in `Prototypes/Palettes/advanced.yml` would make this a much bigger change across a lot more areas. (Which I learned the hard way with my first few commits.)

Yes, the palettes have a strong drop off in value past the half-way point, but that's the current design and I wouldn't feel comfortable changing that without input from the larger group that manages the visual design in general.

These match the value levels seen in wall signs already in use and there isn't a real way around the contrast issue other than making either the lime green or the white darker, and making the white darker would make them REALLY not match the other icons.

That said, Command and Admin icons are already darker than these in value, so they don't come across as out of place to me.

Emisse (approved):

Emisse:
art approval

slarticodefast (approved):

PR #34271, opened 2025-01-07
fix holopads with multiple ai cores dying

deltanedas:
## About the PR
chromiumboy suggested this

reason why it worked sometimes was arbitrary entity order, it might pick the station's ai core vs a different one (in this case on centcomm) so it might never have to return

imagine if return inside a loop had a warning the amount of bugs its created is insane...


chromiumboy:
Can you move this PR to the stable branch so we can hot fix it?

deltanedas:
done

deltanedas:
killed by phoneposting

Errant-4:
So to clarify, no one actually tested this, then?

chromiumboy:
> So to clarify, no one actually tested this, then?

It's already been merged on Delta V, not sure how extensively it was tested beforehand

deltanedas:
it does work just have to remake it on stable

PR #34278, opened 2025-01-07
Sec can open firelocks

Errant-4:
## About the PR
Make triggered firelocks openable with Security access

## Why / Balance
Currently, only Engineers can instantly open a triggered firelock without any tools. This is because they are meant to work in depressurised areas. However, Sec is meant to keep crew safe. Thus, they should be able to open firelocks to get people out of depressurised areas. Additionally, not being able to open firelocks is apparently specifically a problem for them when facing China Lake users.

Sec access is regulated enough that they can probably be trusted with this ability.

## Requirements

**Changelog**
:cl: Errant
- tweak: Firelocks can now be opened with Security access.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Doors/Firelocks/firelock.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/Firelocks/firelock.yml
@@ -106,7 +106,7 @@
     - type: StaticPrice
       price: 150
     - type: AccessReader
-      access: [ [ "Engineering" ] ]
+      access: [ ["Engineering"], ["Security"] ]
     - type: PryUnpowered
       pryModifier: 0.5
     - type: PointLight


Errant-4:
> I'd rather make firelocks be openable for all crew, especially since it's confusing when a firelock rescinds its accesses

Frankly, after having played atmostech and seeing the breached areas, I just don't trust most of the crew with firelocks. I am not against discussing a revamp of how they work in general, but this is not that PR

ArtisticRoomba:
> I'd rather make firelocks be openable for all crew, especially since it's confusing when a firelock rescinds its accesses

No. Nonononononono, god please no. 

Right now, spacing across a station is comprised of this:
- 25% of the spacing is the actual spacing.
- 75% of the spacing is caused by tiders, people who powergame crowbars, etc, who just open all of the damm firelocks and spread the spacing to other rooms.

I don't even trust regular engineers to handle spacing sometimes.

Right now, no space wind has enforced this feeling that opening firelocks between spaced rooms is fine and doesn't cause more damage. There's no imminent threat to the player, especially if that player has space protection.

Alternative phrasing of this question: should regular visitors, passengers, janitors of the station get access to override a vital safety feature of an entire station?

JustinWinningham:
I don't see a reason why command would not also have this ability.

Errant-4:
Paramedic was also requested. I'm considering them

deltanedas:
since everyone powergames crowbars anyway you might as well just get rid of the access requirement, all it does it punish people for not powergaming via subjecting them to slowly dying from space unless they find a maints door and just go into maints

Emisse:
agree

Nox38:
> * 25% of the spacing is the actual spacing.
> * 75% of the spacing is caused by tiders, people who powergame crowbars, etc, who just open all of the damm firelocks and spread the spacing to other rooms.

Id argue that at least 80% of the crew will have crowbars anyway and are just as inclined to open firelocks to get where they are going as if they had access. To stop the inevitable chain spacing of the station with every hull breach or greytide event, firelocks would need to be reworked to be more dangerous to open.

Nox38:
As for adding multiple jobs accesses to firelocks, why not just make the ‘external’ ID access open them?

poklj:
This was a rather large discussion in developer chat, and having played a fair bit of security, the most common cause of death is someone not holding the door after a quick push out into a spaced area, or a firelock closing at the worst possible moment (and the people who just fled through it continuing to flee leaving those who are stuck, behind).

from personal anecdote A Nukie or other space fighter has less of an issue pushing a firelock and keeping the position held than a secoff has fleeing through one, and fights are far more interesting and last longer when a HOP provides security with the ability to open firelocks, Secoffs are also far more likely to be willing to hold the door to allow for retreating fighters to make it through.

Having played a fair bit of atmos tech, people are rather prone to opening firelocks as though they don't exist, consequences be damned and then will quickly complain about the lack of oxygen on the station (that they directly made worse), so full access to firelocks to the general station populous is probably a poor idea, however command's need to flee is probably a good enough reason for them to have it as well.

Errant-4:
> Secoffs are also far more likely to be willing to hold the door to allow for retreating fighters to make it through.
> 
> Having played a fair bit of atmos tech, people are rather prone to opening firelocks as though they don't exist, consequences be damned 

Holding firelocks open for a length of time IS treating them as if they don't exist, and greatly contributes to LARGE areas losing pressure! or do you mean they are holding them when they are not given access by the HOP and so could not reopen it quickly?

Either way, I am starting to suspect that this PR would just make things worse, and the answer is indeed some sort of redesign of how firelocks work, instead

deltanedas:
firelocks serve 2 purposes right now:
- being ridiculously tanky, stronger than RWall
- killing people who dont powergame crowbars

because of a bug nobody wants to fix meaning they only close *after* the area gets spaced unlike SS13, it means they are as good as doors for stopping atmos emergencies.

Errant-4:
I have decided that this needs a proper fix so there is no point to this change, as it would either make things worse, or if it makes things slightly better, it might postpone an actually good solution

Errant-4:
@deltanedas can you elaborate on what that bug you mentioned is, do we have an issue for it? It's not obvious to me how anything about firelock activation could be different

deltanedas:
if you have rooms A and B with a firelock between them, then space A, B will get spaced too before the firelock closes

Errant-4:
When I went to look at this, it was difficult to find any areas where this bug would even matter. It seems rare for two sections to be separated by ONLY a firelock, and not airlock+firelock

poklj:
> > Secoffs are also far more likely to be willing to hold the door to allow for retreating fighters to make it through.
> > Having played a fair bit of atmos tech, people are rather prone to opening firelocks as though they don't exist, consequences be damned
> 
> Holding firelocks open for a length of time IS treating them as if they don't exist, and greatly contributes to LARGE areas losing pressure! or do you mean they are holding them when they are not given access by the HOP and so could not reopen it quickly?
> 
> Either way, I am starting to suspect that this PR would just make things worse, and the answer is indeed some sort of redesign of how firelocks work, instead

I was saying holding the lock open during a fight, Usually air pressure is less of a concern when you have 5 highly dangerous shooters bearing down on you, A secoff would be holding it open for the other Non-Access granted fighters (see The passengers who may take up arms). they would do this for both when granted and not granted.

Outside of fighting, a Secoff probably isn't holding that door open (it's two very different situations)

PR #34283, opened 2025-01-08
Adds DumpableSolution to more solution containers

SpaceRox1244:
## About the PR

Adds the DumpableSolution component to chemistry beakers, jugs, buckets, mop carts, and the welding fuel generator. This means that the player can drag any other solution container onto the container with DumpableSolution to pour it all in.

## Why / Balance

I enjoy the drag-and-drop function of solution transferring as a shorthand for "move as much as this stuff as possible into here". I think its more comfortable in some cases than either spam clicking with a beaker in hand or manually adjusting transfer amount. This PR makes it a more common option.

## Media

Probably doesnt need a video of me doing it all since it's just giving an existing functionality to a handful of items.

## Requirements

**Changelog**

:cl: SpaceRox
- add: You can now drag-and-drop solution containers onto more things to dump in the solution, like chemistry beakers, mop buckets, and generators.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/chemical-containers.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/chemical-containers.yml
@@ -30,6 +30,8 @@
       solution: beaker
     - type: InjectableSolution
       solution: beaker
+    - type: DumpableSolution
+      solution: beaker
     - type: SolutionTransfer
       canChangeTransferAmount: true
     - type: SolutionItemStatus
--- a/Resources/Prototypes/Entities/Objects/Specific/chemistry.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/chemistry.yml
@@ -137,6 +137,8 @@
     solution: beaker
   - type: InjectableSolution
     solution: beaker
+  - type: DumpableSolution
+    solution: beaker
   - type: SolutionTransfer
     canChangeTransferAmount: true
   - type: SolutionItemStatus
--- a/Resources/Prototypes/Entities/Objects/Specific/chemistry.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/chemistry.yml
@@ -39,6 +39,8 @@
     solution: beaker
   - type: InjectableSolution
     solution: beaker
+  - type: DumpableSolution
+    solution: beaker
   - type: SolutionTransfer
     canChangeTransferAmount: true
   - type: SolutionItemStatus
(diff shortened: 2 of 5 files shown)


slarticodefast:
Does this bypass the transfer amount limit?

SpaceRox1244:
> Does this bypass the transfer amount limit?

Yeah it seems to. Is that a problem? :?

slarticodefast:
I assume it is there for a reason, probably to limit the size of chemical reactions you can create at once. But in my opinion this is an intuitive way to interact with containers, so it would be very useful for things like jug mixing meds.
Also I did not do any in-game testing yet, but irrc there were bugs with drag and dropping, for example you can drag food into a chemmaster to transfer all the reagents inside. That would have to get fixed first.

NoElkaTheGod:
> I assume it is there for a reason, probably to limit the size of chemical reactions you can create at once.

That doesn't really work, because you can just instantly spill any container. Back in the day it was a very popular way of creating pretty damn big ClF3 explosions.

slarticodefast:
Gave it a test and yes, you can just drag drop food into a beaker to get the reagents out.
grafik

>That doesn't really work, because you can just instantly spill any container. Back in the day it was a very popular way of creating pretty damn big ClF3 explosions.

But wouldn't that still be limited by the puddle size? I have seen it cause a bunch of small ClF3 explosions when you throw beakers for example.

SpaceRox1244:
> Gave it a test and yes, you can just drag drop food into a beaker to get the reagents out.

Oh yeah I remember this now. That's also already possible with the ChemMaster.

slarticodefast:
Yeah, but allowing everyone to do it with any container would kinda make it much worse. So I would prefer if that was fixed first. Not this PR's fault of course.

NoElkaTheGod:
> > That doesn't really work, because you can just instantly spill any container. Back in the day it was a very popular way of creating pretty damn big ClF3 explosions.
> 
> But wouldn't that still be limited by the puddle size? I have seen it cause a bunch of small ClF3 explosions when you throw beakers for example.

Pretty sure puddle size is a soft cap, so if you manage to spill all the containers fast enough (or just explode them), all the reagents will react at once.

insoPL:
> Gave it a test and yes, you can just drag drop food into a beaker to get the reagents out.
> grafik
> 
> >That doesn't really work, because you can just instantly spill any container. Back in the day it was a very popular way of creating pretty damn big ClF3 explosions.
> 
> But wouldn't that still be limited by the puddle size? I have seen it cause a bunch of small ClF3 explosions when you throw beakers for example.

For anyone that comes back to this PR. Food drag and drop was fixed.

AJCM-git:
stale

PR #34303, opened 2025-01-08
Fix Mixed puddles not updating slips when evap

poklj:
## About the PR

When a puddle evaporates it does so in steps.
These steps will never be displayed to players cause the solution container is never told to update.
Further, If a solution is a "mixed solution" (i.e has 2 reagents in it) it will not update it's state when an evaporatable reagent evaporates out of the solution, causing it to be displayed in the larger puddle state, With an incorrect description and will act as though it is still slippery for at least 1 slip/external forces firing an SolutionContainerUpdateEvent
## Why / Balance

Mixed solution puddles are and have always been broken

## Technical details

Update the solution every evaporation tick

## Media

https://github.com/user-attachments/assets/f3404643-bed7-4c8a-8f02-b29a2ff9c5bf

https://github.com/user-attachments/assets/0ac15342-4493-4939-b1c2-38d709cd28f8

## Requirements

## Breaking changes

**Changelog**

:cl: Kickguy223
- fix: Puddles will now correctly evaporate

### DIFF
--- a/Content.Server/Fluids/EntitySystems/PuddleSystem.Evaporation.cs
+++ b/Content.Server/Fluids/EntitySystems/PuddleSystem.Evaporation.cs
@@ -1,6 +1,4 @@
-using Content.Server.Chemistry.Containers.EntitySystems;
 using Content.Shared.Chemistry.Components;
-using Content.Shared.Chemistry.EntitySystems;
 using Content.Shared.FixedPoint;
 using Content.Shared.Fluids.Components;
 
--- a/Content.Server/Fluids/EntitySystems/PuddleSystem.Evaporation.cs
+++ b/Content.Server/Fluids/EntitySystems/PuddleSystem.Evaporation.cs
@@ -1,3 +1,4 @@
+using Content.Server.Chemistry.Containers.EntitySystems;
 using Content.Shared.Chemistry.Components;
 using Content.Shared.Chemistry.EntitySystems;
 using Content.Shared.FixedPoint;
@@ -57,9 +58,7 @@ public sealed partial class PuddleSystem
                 QueueDel(uid);
             }
 
-            var ev = new SolutionContainerChangedEvent(puddleSolution, puddle.SolutionName);
-            RaiseLocalEvent(uid, ref ev);
-            Dirty(puddle.Solution.Value.Owner, puddle.Solution.Value.Comp);
+            _solutionContainerSystem.UpdateChemicals(puddle.Solution.Value);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Content.Server/Fluids/EntitySystems/PuddleSystem.Evaporation.cs
```diff
@@ -55,6 +56,10 @@ private void TickEvaporation()
                 Spawn("PuddleSparkle", xformQuery.GetComponent(uid).Coordinates);
                 QueueDel(uid);
             }
+
+            var ev = new SolutionContainerChangedEvent(puddleSolution, puddle.SolutionName);
+            RaiseLocalEvent(uid, ref ev);
+            Dirty(puddle.Solution.Value.Owner, puddle.Solution.Value.Comp); // Fix the misprediction
```
Centronias, line 62:
```suggestion
            Dirty(puddle.Solution.Value.Owner, puddle.Solution.Value.Comp);
```

The comment seems like an explanation for why this change is in this PR, and it won't make sense once this is committed.

Content.Server/Fluids/EntitySystems/PuddleSystem.Evaporation.cs
```diff
@@ -55,6 +56,10 @@ private void TickEvaporation()
                 Spawn("PuddleSparkle", xformQuery.GetComponent(uid).Coordinates);
                 QueueDel(uid);
             }
+
+            var ev = new SolutionContainerChangedEvent(puddleSolution, puddle.SolutionName);
+            RaiseLocalEvent(uid, ref ev);
+            Dirty(puddle.Solution.Value.Owner, puddle.Solution.Value.Comp);
```
metalgearsloth, line 62: This should just call solutioncontainersystem instead which handles the events rather than manually doing solution handling.

Emisse (approved):

slarticodefast (commented):
I gave it a test and I'm still slipping when walking of a puddle that was space lube, but has been turned into water by space cleaner.
I think you need to do the same for `CleanTileReaction`.

poklj:
> I gave it a test and I'm still slipping when walking of a puddle that was space lube, but has been turned into water by space cleaner spray bottle. I think you need to do the same for `CleanTileReaction`.

Discussed this in discord: essentially the original solution was assumed to solve the linked issue, however `PuddleSystem::UpdateSlip` _never updates the Super Slippery component_ based on the Puddles solution contents.

The original problem this PR was meant to solve was: 
When a puddle evaporates it does so in steps.
These steps will never be displayed to players cause the solution container is never told to update.
Further, If a solution is a "mixed solution" (i.e has 2 reagents in it) it will not update it's state when an evaporatable reagent evaporates out of the solution, causing it to be displayed in the larger puddle state, With an incorrect description and will act as though it is still slippery for at least 1 slip;

I will remove the issue note and do another PR to solve the super slippery problem

slarticodefast:
Thanks for investigating!
This should be good to merge as its own bugfix.

PR #34304, opened 2025-01-08
Decouple water from evaporation code 

poklj:
## About the PR

Fixed the hardcoding limiting evaporation to only Water. Left out applying this to anything (like space lube) to allow other PR's to handle adding it to reagents.
## Why / Balance

Evaporation is hardcoded to the water Prototype ID via a magic variable. By adding helper functions and a Prototype field, evaporation is able to be applied to any reagent.

## Technical details

added 2 helper functions to cover all cases where Evaporation either utilizes The magic variable or otherwise would need to check if a reagent is evaporatable

## Media

https://github.com/user-attachments/assets/565ef564-06fa-470f-a194-0057d22d20e4

## Requirements

## Breaking changes

**Changelog**

No CL No Fun

### DIFF
--- a/Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
+++ b/Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
@@ -106,7 +106,7 @@ namespace Content.Shared.Chemistry.Reagent
         public bool Slippery;
 
         /// <summary>
-        /// If this reagent is part fo a puddle does it evaporate
+        /// If this reagent is part of a puddle does it evaporate
         /// </summary>
         [DataField]
         public bool Evaporates;
--- a/Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
+++ b/Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
@@ -106,6 +106,12 @@ namespace Content.Shared.Chemistry.Reagent
         public bool Slippery;
 
         /// <summary>
+        /// If this reagent is part fo a puddle does it evaporate
+        /// </summary>
+        [DataField]
+        public bool Evaporates;
+
+        /// <summary>
         /// How easily this reagent becomes fizzy when aggitated.
         /// 0 - completely flat, 1 - fizzes up when nudged.
         /// </summary>
--- a/Content.Shared/Fluids/SharedPuddleSystem.Evaporation.cs
+++ b/Content.Shared/Fluids/SharedPuddleSystem.Evaporation.cs
@@ -1,5 +1,6 @@
 using Content.Shared.Chemistry.Components;
 using Content.Shared.Chemistry.Reagent;
+using Content.Shared.FixedPoint;
 
 namespace Content.Shared.Fluids;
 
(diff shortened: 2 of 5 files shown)


Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
```diff
@@ -105,6 +105,12 @@ public sealed partial class ReagentPrototype : IPrototype, IInheritingPrototype
         [DataField]
         public bool Slippery;
 
+        /// <summary>
+        /// If this reagent is part fo a puddle does it evaporate
```
0x6273, line 109:
```suggestion
        /// If this reagent is part of a puddle does it evaporate
```

Content.Shared/Fluids/SharedPuddleSystem.Evaporation.cs
```diff
@@ -10,8 +11,31 @@ public abstract partial class SharedPuddleSystem
...
+
+        foreach (var (reagent, _) in solution.Contents)
+        {
+            var reagentProto = _prototypeManager.Index<ReagentPrototype>(reagent.Prototype);
+            if (reagentProto.Evaporates)
+                evaporationReagents.Add(reagentProto.ID);
+        }
+        return evaporationReagents.ToArray();
+    }
+    public bool SolutionHasEvaporation(Solution solution)
```
0x6273, line 26:
```suggestion
    }
    
    public bool SolutionHasEvaporation(Solution solution)
```

Content.Shared/Fluids/SharedPuddleSystem.Evaporation.cs
```diff
@@ -10,8 +11,31 @@ public abstract partial class SharedPuddleSystem
...
+    {
+        foreach (var (reagent, quantity)  in solution.Contents)
+        {
+            var reagentProto = _prototypeManager.Index<ReagentPrototype>(reagent.Prototype);
+            if(reagentProto.Evaporates && quantity > FixedPoint2.Zero)
```
0x6273, line 31:
```suggestion
            if (reagentProto.Evaporates && quantity > FixedPoint2.Zero)
```

Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
```diff
@@ -105,6 +105,12 @@ public sealed partial class ReagentPrototype : IPrototype, IInheritingPrototype
...
+        /// </summary>
+        [DataField]
+        public bool Evaporates;
```
0x6273, line 112: I think it would be better to make this a number for how fast it evaporates, instead of a bool. So different reagents can evaporate at different rates. 0 would be the same as false.

lzk228:
at this point, can we add evaporate speed?

Content.Shared/Chemistry/Reagent/ReagentPrototype.cs
poklj: It would simply be making it a 0,1 where 1 is true and 0 is false as adding that functionality would creep the scope beyond the magic variable decoupling. 

But yes this can be done

slarticodefast:
Similar to https://github.com/space-wizards/space-station-14/pull/33399 I think?
We would habe to see which implementation is better.

dvir001:
> Similar to #33399 I think? We would habe to see which implementation is better.

If you can mop blood with puke / space lube / anything.
This PR then missing the other fix to the issue related to mop code that https://github.com/space-wizards/space-station-14/pull/33399 fixed.

EmoGarbage404:
Superseded by #33399

PR #34330, opened 2025-01-09
Fix access configurator debug assert

ScarKy0:
## About the PR

Fixes a debug assert raised by the access configurator on modifying door access.

## Technical details

Made GetMainAccessReader return an entity instead of just the component, then fixing the dirty to raise on that entity.

## Requirements

## Breaking changes

AccessReaderSystem.GetMainAccessReader now gives an entity instead of just the access reader comp.

**Changelog**

changelogn't

### DIFF
--- a/Content.Server/Access/Systems/AccessOverriderSystem.cs
+++ b/Content.Server/Access/Systems/AccessOverriderSystem.cs
@@ -245,7 +245,7 @@ public sealed class AccessOverriderSystem : SharedAccessOverriderSystem
             $"{ToPrettyString(player):player} has modified {ToPrettyString(accessReaderEnt.Value):entity} with the following allowed access level holders: [{string.Join(", ", addedTags.Union(removedTags))}] [{string.Join(", ", newAccessList)}]");
 
         accessReaderEnt.Value.Comp.AccessLists = ConvertAccessListToHashSet(newAccessList);
-        Dirty(accessReaderEnt.Value, accessReaderEnt.Value.Comp);
+        Dirty(accessReaderEnt.Value);
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


slarticodefast (approved):
Code looks good and works as intended. Thank you for fixing it!

Content.Server/Access/Systems/AccessOverriderSystem.cs
```diff
@@ -242,10 +242,10 @@ private void TryWriteToTargetAccessReaderId(EntityUid uid,
         var removedTags = oldTags.Except(newAccessList).Select(tag => "-" + tag).ToList();
 
         _adminLogger.Add(LogType.Action, LogImpact.Medium,
-            $"{ToPrettyString(player):player} has modified {ToPrettyString(component.TargetAccessReaderId):entity} with the following allowed access level holders: [{string.Join(", ", addedTags.Union(removedTags))}] [{string.Join(", ", newAccessList)}]");
+            $"{ToPrettyString(player):player} has modified {ToPrettyString(accessReaderEnt.Value):entity} with the following allowed access level holders: [{string.Join(", ", addedTags.Union(removedTags))}] [{string.Join(", ", newAccessList)}]");
 
-        accessReader.AccessLists = ConvertAccessListToHashSet(newAccessList);
-        Dirty(component.TargetAccessReaderId, accessReader);
+        accessReaderEnt.Value.Comp.AccessLists = ConvertAccessListToHashSet(newAccessList);
+        Dirty(accessReaderEnt.Value, accessReaderEnt.Value.Comp);
```
Tayrtahn, line 248:
```suggestion
        Dirty(accessReaderEnt.Value);
```
Nitpick. I'm pretty sure you can just pass the Entity<T> to Dirty instead of breaking it up.

slarticodefast: I think it needs a `Dirty<AccessReaderComponent>(accessReaderEnt.Value)` in that case.

ScarKy0: Rider says its redundant, is this really a good way to do it? I don't think ive seen anything dirtied this way before.

slarticodefast: `accessReaderEnt` is an `Entity<AccessReaderComponent>`
so only that component gets dirtied
it's just a more compact way of writing both the uid and component separately

Tayrtahn (approved):

PR #34334, opened 2025-01-09
Fix multiple research consoles exploit

### LINKED ISSUE #34160: Opening multiple R&D computers lets you unlock every T3 at once
## Description

Having multiple R&D consoles open at once makes it so that doing research on any console does not update the available unlocks on any other console UNLESS the console is updated via a player opening it again. Additionally, selecting a new T3 changes the current Main Discipline, allowing for the full tech tree of a Discipline to be unlocked before moving onto the next tech tree.

**Reproduction**

1. Have enough research points to unlock wanted T3s
2. Open at least 2 R&D computers and have no one else update their displays by interacting with them
3. Unlock the wanted upgrades from one Discipline using one console
4. Unlock the wanted upgrades from a different Discipline using another console
(Repeat for as many branches of T3s as you want)
5. Science Major Victory!

**Screenshots**

image

**Additional context**

I have been able to perform this multiple times across Lizard and Leviathan, and others have witnessed me doing so. Potential mark as the first Bug of the Year 2025, in more than one way.
image

themias:
## About the PR

This fixes an exploit where available technologies are not updated on one open console if unlocked on another. This could allow unlocking tier 3 technologies from different disciplines.

Fixes #34160 

## Technical details

Research consoles have a local copy of a research server's database. This gets updated when unlocking a technology. 
The server sends an update event to all of its clients (research consoles, lathes, etc.) However, upon receiving this event, other research consoles would only update their available points, only syncing their DBs when reopening the UI.

This change makes sure that all the open research consoles for a given research server will update their available technologies/disciplines if another console unlocks them.

## Media

https://github.com/user-attachments/assets/60abf7dd-5785-4337-a40d-3d89fa64db45

## Breaking Changes
- Research console syncing now raises TechnologyDatabaseSynchronizedEvent instead of TechnologyDatabaseModifiedEvent

## Requirements

**Changelog**
:cl:
- fix: Fixed an exploit using multiple R&D computers to research multiple discipline tier 3 technologies


Content.Server/Research/Systems/ResearchSystem.Console.cs
```diff
@@ -97,6 +97,7 @@ private void OnConsoleRegistrationChanged(EntityUid uid, ResearchConsoleComponen
 
     private void OnConsoleDatabaseModified(EntityUid uid, ResearchConsoleComponent component, ref TechnologyDatabaseModifiedEvent args)
     {
+        SyncClientWithServer(uid, raiseEvents: false);
```
Centronias, line 100: This kind of boolean is the start of the path to unmaintainable spaghetti.
Why not extract the portions of `SyncClientWithServer` that you do want to call into a new function and call that from here and reuse it in `SyncClientWithServer`?
Basically, one function would be "just synchronize the DB components" and the other would be "synchronize the DB components and raise `TechnologyDatabaseModifiedEvent`"?

Centronias: Maybe the real issue here is that `TechnologyDatabaseModifiedEvent` is raised on the console _even though the local DB copy hasn't actually changed_, so a more correct solution would be to add a new event to discriminate between "The server I sync with changed" and "my internal DB copy changed"?

Centronias (changes requested):

Centronias (approved):
Nice and clean 👍

Content.Server/Research/Systems/ResearchSystem.Technology.cs
```diff
@@ -23,7 +23,7 @@ public void Sync(EntityUid primaryUid, EntityUid otherUid, TechnologyDatabaseCom
 
         Dirty(primaryUid, primaryDb);
 
-        var ev = new TechnologyDatabaseModifiedEvent();
+        var ev = new TechnologyDatabaseSynchronizedEvent();
         RaiseLocalEvent(primaryUid, ref ev);
```
metalgearsloth, line 27: This is a breaking change and should be listed on the PR please.

themias: Derp, I originally wasn't solving it this way and forgot to go back and add one. Thanks!

Emisse (approved):

Terraspark4941:
can't t3maxx anymore, its so joever (thanks for the fix)

PR #34338, opened 2025-01-09
Bottled drinks and dispensers standardised

Alpaccalypse:
## About the PR
Corrects a variety of inconsistencies with bottled drinks and the soda dispeser.

## Why / Balance
There was a wild array of different containers and volumes used on for the soda dispenser, and the list of drinks they came from.

Drinks like Bloody Mary and Cosmopolitan can now be mixed at round start.

## Technical details
Changes the values for most drinks in drinks_bottles.yml
Adds some missing drinks to beverage.yml for the soda dispenser
Makes soda and booze dispensers inherit from ReagendDispenserBase

## Requirements

**Changelog**
Probably. TBA


Alpaccalypse:
I think, if nothing else, this could benefit from some feedback and discussion.

For instance, just with the soda dispenser: 

- If I just add every missing drink used in recipes to the soda dispenser, it overflows the bottom of the default window height in the UI.
- I'm also not sure about adding milk and soy milk to the dispenser... but cream is there and they are both used in common cocktail recipes.
- If I remove unused sodas from the dispenser, it will fit, but then again, it is a "soda" dispenser, so why aren't they there?

A normal small water bottle is still the odd one out, at only 30u volume, but as everyone starts with this I wanted to check before increasing it to 50u, which is what the rest of the soda bottles have. Soda and tonic water both had 100u bottles, so a I made a new 100u water bottle, that they inherit from.

Alpaccalypse:
It would kind of make sense to have a juice dispenser and a hot drinks dispenser. It all coming out of one soda dispenser seems a bit strange. It would also help solve UI bloat, and remove a lot of stuff that isn't used very often with bartending from the single dispenser.

On the other hand, even 1 more dispenser to add to all maps is a bit of a hassle. Not sure how bartenders would feel either. Less clutter, but more dispensers to deal with.

Alpaccalypse:
Sorry to be just venting my thoughts here, but it ocurred to me, that the neatest solution might be to simplify some of the drink and reactions involved in bartending, in it's current state. One shining example is that we have two branded lemon sodas used in cocktails but lemonade is only used in one cocktail, requires its own reaction to mix, and is not included in the soda dispenser.

Prole0:
Hope you see this soon but I'm using your code as a base to fix https://github.com/space-wizards/space-station-14/issues/36024. 

Gonna use this to separate water bottle from being used as a parent to soda/tonic bottles. Credit will be given on the PR, thank you for your work. 
image

EmoGarbage404:
Pretty sure this is no longer needed

Issue #29306, opened 2024-06-21
Rimworld floor exploded


HolyEclipsee:
Something exploded in the rimworld (i guess its name i cant remember it correctly) it was wood floor (changed) before explosion and if you look at picture its black like a void no dirt or grass or stone and you can walk on it with no problem

![20240621151131_1](https://github.com/space-wizards/space-station-14/assets/173493171/39ffc57d-352f-4a93-8aee-4296e6347fe0)

MACMAN2003:
this is because placing tiles on a planetmap requires replacing the grass/dirt/sand with the endless void of space

PR #34339, opened 2025-01-09
Make indestructible tiles not breakable by explosions

poklj:
## About the PR

Make explosions no longer break indestructible tiles, fixing AI spacing and allowing shuttles to no longer get spaced by Shuttlebombing (Tiles not included)
## Why / Balance

They're indestructible.
## Technical details

add guard statement to DamageFloorTile to stop damage when encountering a indestructible floor tile.
## Media

https://github.com/user-attachments/assets/6eb3f51c-01ff-41ab-b3bc-46d56e952f9c

## Requirements

## Breaking changes

**Changelog**

:cl: Kickguy223
- fix: Indestructible tiles no longer break due to explosions

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
+++ b/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
@@ -512,9 +512,8 @@ public sealed partial class ExplosionSystem
         List<(Vector2i GridIndices, Tile Tile)> damagedTiles,
         ExplosionPrototype type)
     {
-        if (_tileDefinitionManager[tileRef.Tile.TypeId] is not ContentTileDefinition tileDef)
-            return;
-        if(tileDef.Indestructible)
+        if (_tileDefinitionManager[tileRef.Tile.TypeId] is not ContentTileDefinition tileDef
+            || tileDef.Indestructible)
             return;
 
         if (!CanCreateVacuum)
--- a/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
+++ b/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
@@ -514,6 +514,8 @@ public sealed partial class ExplosionSystem
     {
         if (_tileDefinitionManager[tileRef.Tile.TypeId] is not ContentTileDefinition tileDef)
             return;
+        if(tileDef.Indestructible)
+            return;
 
         if (!CanCreateVacuum)
             canCreateVacuum = false;


SlimmSlamm:
I dont think we should take away shuttle bombing because this will highly affect nukies/DAGD

minus1over12:
Does this also fix #29306?

NoElkaTheGod:
> I dont think we should take away shuttle bombing because this will highly affect nukies/DAGD

Um, no? DADG is not "kill yourself while causing as much harm as possible", it's "die with glory". You don't have to blow yourself up on evac to complete it. In fact, blowing up evac is the most boring way of completing DADG, for the love of god, be more creative.

Errant-4:
Possible future uses of indestructible tiles, and its ramifications, are beyond the scope of this PR. 
We WANT tiles like this at the very least for Arrivals terminal/shuttle, where this is mechanically enforcing a rule.

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
```diff
@@ -514,6 +514,8 @@ public void DamageFloorTile(TileRef tileRef,
     {
         if (_tileDefinitionManager[tileRef.Tile.TypeId] is not ContentTileDefinition tileDef)
             return;
+        if(tileDef.Indestructible)
```
Errant-4, line 517: Please combine the new check with the existing one, since the outcome is the same

        if (_tileDefinitionManager[tileRef.Tile.TypeId] is not ContentTileDefinition tileDef
            || tileDef.Indestructible)
            return;

Errant-4 (changes requested):
Tested and it works, code just needs to be streamlined a bit

slarticodefast:
Woudn't this block all tiles from being damaged if the explosion originates on an indestructible tile?
For example a nukie being anchored on one.

Errant-4:
> Woudn't this block all tiles from being damaged if the explosion originates on an indestructible tile? For example a nuke being anchored on one exploding.

I just tried it, and it did not stop a nuke, merely the indestructible tiles directly under it have survived as expected. The explosion crashed my client but that nearly always happens when I detonate a test nuke so it's probably normal

poklj:
> > Woudn't this block all tiles from being damaged if the explosion originates on an indestructible tile? For example a nuke being anchored on one exploding.
> 
> I just tried it, and it did not stop a nuke, merely the indestructible tiles directly under it have survived as expected. The explosion crashed my client but that nearly always happens when I detonate a test nuke so it's probably normal

I also tested it with smaller explosives and it works the same.
Content Client_2025-01-10_12-12-10

I've also ensured that Singularity still eats them 
Content Client_2025-01-09_17-53-20

Errant-4 (approved):

slarticodefast (approved):
Works as intended.
The nukie planet is still destructible, but that is probably a mapping PR.

PR #34340, opened 2025-01-09
Mapping rotation fix

honeyed-lemons:
## About the PR
Fixes a blatant issue where ties between maps with the same priority are broken with alphabetical order, this is really egregious with maps like Amber, which can roll constantly.

## Why / Balance
It is a horrible bug, really bad especially for downstream servers with more maps

## Technical details
Previously, the GetFirstInRotationQueue function in GameMapManager.cs sorted alphabetically last, this has been changed to no longer sort alphabetically, and instead choose a map randomly (with a fail safe if maps have the same weird)

## Media
No media required 
## Requirements

- [x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
No known breaking changes 

**Changelog**
:cl:
- fix: Fixed map rotations not being truely random

### DIFF
--- a/Content.Server/Maps/GameMapManager.cs
+++ b/Content.Server/Maps/GameMapManager.cs
@@ -217,18 +217,22 @@ public sealed class GameMapManager : IGameMapManager
     {
         _log.Info($"map queue: {string.Join(", ", _previousMaps)}");
 
-        var eligible = CurrentlyEligibleMaps()
+        var eligibleList = CurrentlyEligibleMaps()
             .Select(x => (proto: x, weight: GetMapRotationQueuePriority(x.ID)))
+            .ToArray();
+
+        var thresholdList = eligibleList
             .Where(x => x.weight > 0)
             .ToArray();
 
-        _log.Info($"eligible queue: {string.Join(", ", eligible.Select(x => (x.proto.ID, x.weight)))}");
+        _log.Info($"eligible queue: {string.Join(", ", eligibleList.Select(x => (x.proto.ID, x.weight)))}");
 
         // YML "should" be configured with at least one fallback map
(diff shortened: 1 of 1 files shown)


Content.Server/Maps/GameMapManager.cs
```diff
@@ -217,20 +217,22 @@ private GameMapPrototype GetFirstInRotationQueue()
...
-        Debug.Assert(eligible.Length != 0, $"couldn't select a map with {nameof(GetFirstInRotationQueue)}()! No eligible maps and no fallback maps!");
+        Debug.Assert(eligibleList.Length != 0, $"couldn't select a map with {nameof(GetFirstInRotationQueue)}()! No eligible maps and no fallback maps!");
 
-        var weight = eligible[0].weight;
-        return eligible.Where(x => x.Item2 == weight)
-            .MinBy(x => x.proto.ID)
-            .proto;
+        // Choose a random map based on the threshold condition
+        var selectedList = thresholdList.Any() ? thresholdList : eligibleList;
+        return _random.Pick(selectedList).proto;
```
metalgearsloth, line 235: Code doesn't really make sense to me:
1. Why are you getting a list with weights then immediately discarding it if the weight is <= 0
2. Why does it randomly pick from a *weighted* list?
3. Why use LINQ for Any instead of count.

honeyed-lemons: let me fix this rq :)

honeyed-lemons: Actually just noticed, while fixing this i accidentally forgot what the function was originally for HAHA, I'll close this and revisit later to completely recode, but i just want to let y'all know that function doesnt work properly

PR #34358, opened 2025-01-10
Remove the ability to print the station anchor circuit board

ArtisticRoomba:
## About the PR
Removes the ability to print the station anchor machine board.

Note that this board can still be obtained by admemery.

## Why / Balance
Right now, the Station Anchor can be printed and flatpacked by Science.

Unfortunately, this can be abused. People can easily multitool a station anchor in the halls, a tight space, etc. You're not supposed to be able to make and easily deploy machines of this size. This is exactly why you can't make the station Gravity Generator for example.

Recently I've seen people making a station anchor flatpack and then unpacking it on Evac. Evac will still launch, but it doesn't move while in FTL space.

While I do agree that this violates some core design principles (removing player agency, and working against the dynamic environment SS14 should be), the station anchor isn't a necessary part of the station, and if destroyed, isn't really an issue. Except for maybe salvage. But not really.

The station gravity generator acts similarly in this regard, however it is much more important and usually leads to a shuttle-call if destroyed. Science can research shuttlecraft and make a mini gravity generator, which works, however I haven't actually seen this occur in a round before evac arrives.

image

## Technical details
Removes the recipe from the lathe.

## Media
image

## Requirements

**Changelog**
:cl:
- remove: The station anchor machine board can no longer be printed at the circuit imprinter.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/lathe.yml
@@ -451,7 +451,6 @@
     - SodaDispenserMachineCircuitboard
     - SpaceHeaterMachineCircuitBoard
     - CutterMachineCircuitboard
-    - StationAnchorCircuitboard
     - SalvageMagnetMachineCircuitboard
     dynamicRecipes:
       - ThermomachineFreezerMachineCircuitBoard
--- a/Resources/Prototypes/Recipes/Lathes/electronics.yml
+++ b/Resources/Prototypes/Recipes/Lathes/electronics.yml
@@ -586,12 +586,6 @@
 
 - type: latheRecipe
   parent: BaseGoldCircuitboardRecipe
-  id: StationAnchorCircuitboard
-  result: StationAnchorCircuitboard
-  completetime: 8
-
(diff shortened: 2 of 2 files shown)


everydayeldergod:
I've never actually seen the station anchor used as a pop-up barrier before, but I don't see why it's necessarily a problem. Like you said, it's extremely rare to sabotage the station anchor, since it's pretty much inconsequential from a gameplay perspective.

I think that removing the station anchor flatpack has no real roleplay or gameplay advantages, while keeping pop-up station anchors provides a niche disruption option for antagonists.

HyperB1:
Maybe someone should make a compact version of a station anchor for smaller grids in analogy to the gravity generator? I see this being very useful downstream.

slarticodefast (approved):
Yeah, it's basically an instand barricade when used with a flatpack and clips through walls and other objects when constructed.
grafik
grafik
If someone wants to add it back we would need some way to construct larger machine frames or make it similar to the AME or PA.

ScarKy0:
Tested in game and works as intended. Thank you for your contribution!

PR #34368, opened 2025-01-10
Let ghosts sometimes make certain devices say creepy things

Tayrtahn:
## About the PR

The "Boo!" ability for ghosts can now sometimes make vending and arcade machines randomly speak.

## Why / Balance

Making a couple of lights flicker isn't very exciting these days. The Boo action seems to have been designed to be expanded to have some more effects, so I'm adding one.

Obviously ghosts shouldn't be able to affect the round too dramatically. I think that occasionally making a vending machine say something creepy is a pretty cosmetic thing, and I can see some fun RP coming out of it.

The 2-minute cooldown on the Boo action combined with the configurable probabilities and cooldowns between responses should keep these event fairly uncommon and avoid spam. That said, it's a bit hard to estimate how frequent these will wind up being without playtesting, so it may be necessary to adjust the values (with simple YAML tweaks) in the future.

## Technical details

Adds `SpookySpeaker` component and system. The system subscribes to `GhostBooEvent` and (after a cooldown and probability check) picks a random entry from a `LocalizedDataset` of messages.

Additionally, a small tweak was made to GhostSystem, where the HashSet of nearby entities is converted to a List and shuffled, ensuring that the spawn order of entities no longer affects the priority of `GhostBooEvent` targets.

## Media

Screenshot 2025-01-08 at 3 45 18 PM
Screenshot 2025-01-08 at 3 46 44 PM
Screenshot 2025-01-10 at 2 21 32 PM
Screenshot 2025-01-10 at 2 21 47 PM
Screenshot 2025-01-10 at 2 21 54 PM

The recycler can (rarely) react too:
Screenshot 2025-01-08 at 3 48 09 PM

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Arcades, vending machines and the recycler now sometimes say creepy things when a ghost uses the "Boo!" action near them.


SoulFN:
Can revenant do the same if he clicks on vendor? Like he does with the lights.

Tayrtahn:
> Can revenant do the same if he clicks on vendor? Like he does with the lights.

It looks like the revenant code specifically checks for PoweredLightComponent, so no. I don't think that would be hard to change though.

slarticodefast (approved):
Fun idea!
Code looks good and works as intended. Maybe the exact numbers will need some tweaking, but we can just try them out.

PR #34369, opened 2025-01-10
Add a popup message when ghost Boo action does nothing

Tayrtahn:
## About the PR

Adds a simple notification to the player when they use the "Boo!" action as a ghost but it doesn't affect anything.

## Why / Balance

Activating the Boo ability and seeing no result feels bad and has led people to think that the ability is broken. Giving the player feedback when the ability fails makes it clear that at least the game is working correctly.

## Technical details

The code for the Boo action already tracks the number of entities that reacted. So this just checks if the number is zero and displays a popup on the specific client if it is.

## Media

Screenshot 2025-01-10 at 3 03 38 PM

## Requirements

## Breaking changes

**Changelog**

Nah.

### DIFF
--- a/Content.Server/Ghost/GhostSystem.cs
+++ b/Content.Server/Ghost/GhostSystem.cs
@@ -24,6 +24,7 @@ using Content.Shared.Mobs.Components;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.Movement.Events;
 using Content.Shared.Movement.Systems;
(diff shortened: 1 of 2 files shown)


beck-thompson:
love this!!

slarticodefast (approved):
Yeah, that was confusing sometimes. Thanks for the PR!

lzk228:
maybe we could cooldown time if it does nothing?

PR #34375, opened 2025-01-11
Fix nonsensical RegEx for name restriction

HyperB1:
## About the PR

Basically I cleaned up the RestrictedNameRegex, mainly by removing unnecessary delimiters (commas) and a backslash, then moved hyphen to the end of the expression for it to not be interpreted as a range of characters.

## Why / Balance

Correct regular expression is better than an incorrect one.

## Technical details

Unfortunately, for C# and .NET I cannot shorten [A-Za-z0-9] to just \w, as it will include characters from most alphabets and writing systems in unicode.

**Clarification for the edited characters:**
- `-` is a range character, has to be taken literally or be at the end, otherwise causes an erroneous expression interpretation.
- No need for `,` to repeat so much, it's not a separator and names should not have commas in them.
- `\\` is just why? It was added alongside the apostrophe symbol in #28652 for no particular reason. Backslash repeated twice would escape itself and does not seem to affect anything. Replaced by a single backslash `\`.
- Added a `@` delimiter for proper interpretation of meta escape characters and for consistency to avoid further problems.

Tested and works as intended (no warnings as well).

## Media

You can see how the new RegEx filters improper characters in this example.
unknown_2025 01 13-06 28

## Requirements

## Breaking changes

Uh, IDK, your characters with a name containing a backslash or a comma will no longer contain them.

**Changelog**
No CL, no fun.

### DIFF
--- a/Content.Shared/Preferences/HumanoidCharacterProfile.cs
+++ b/Content.Shared/Preferences/HumanoidCharacterProfile.cs
@@ -25,7 +25,7 @@ namespace Content.Shared.Preferences
     [Serializable, NetSerializable]
     public sealed partial class HumanoidCharacterProfile : ICharacterProfile
     {
-        private static readonly Regex RestrictedNameRegex = new("[^A-Za-z0-9 '-]");
+        private static readonly Regex RestrictedNameRegex = new(@"[^A-Za-z0-9 '\-]");
         private static readonly Regex ICNameCaseRegex = new(@"^(?<word>\w)|\b(?<word>\w)(?=\w*$)");
 
         public const int MaxNameLength = 32;
--- a/Content.Shared/Preferences/HumanoidCharacterProfile.cs
+++ b/Content.Shared/Preferences/HumanoidCharacterProfile.cs
@@ -25,7 +25,7 @@ namespace Content.Shared.Preferences
     [Serializable, NetSerializable]
     public sealed partial class HumanoidCharacterProfile : ICharacterProfile
     {
-        private static readonly Regex RestrictedNameRegex = new("[^A-Z,a-z,0-9, ,\\-,']");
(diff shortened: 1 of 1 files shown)


0x6273:
> `\\` is just why?

I think it's to escape the `-`. Putting it at the end also works but it's common practice to always escape to prevent mistakes if more characters are added at the end.

HyperB1:
> I think it's to escape the `-`.
Unfortunately it just... Escapes itself by repeating twice. I should've mentioned it, my bad.

0x6273:
> > I think it's to escape the `-`.
> 
> Unfortunately it just... Escapes itself by repeating twice. I should've mentioned it, my bad.

Really? I'm pretty sure the double backslash is needed in C# string literals, and the actual string that regex sees will only contain one backslash before the `-`.

HyperB1:
> I'm pretty sure the double backslash is needed in C# string literals, and the actual string that regex sees will only contain one backslash before the `-`.
> 
> Edit: It seems like the best practice for regex strings is to use the `@" "` string literals, which don't need escaping so it avoids having to double escape. The regex on the next line does that.

Good to know. I'll test it. How would you recommend me to edit the expression to properly escape `-`?

0x6273:
> > I'm pretty sure the double backslash is needed in C# string literals, and the actual string that regex sees will only contain one backslash before the `-`.
> > Edit: It seems like the best practice for regex strings is to use the `@" "` string literals, which don't need escaping so it avoids having to double escape. The regex on the next line does that.
> 
> Good to know. I'll test it. How would you recommend me to edit the expression to properly escape `-`?

Change it to a `@" "` string literal (like the regex on the next line) and escape it with a single backslash.

Emisse (approved):

0x6273 (approved):

PR #18177, opened 2023-07-20
Arachnids: New markings, new organs sprites and chelicerae

PixelTheKermit:
## About the PR

Adds a few organ sprites, some new markings and gives arachnids a pair of chelicerae.
Also removed the webs that used to be on appendages. I didn't really consider them to look that good.
The arachnid sprite also had very small changes but they're really not noteworthy.

**Media**

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

https://github.com/space-wizards/space-station-14/assets/85175107/73983c3c-3865-47cf-a2bd-c5fa357a18d8

image
image

**Changelog**

:cl: PixelTK
- add: Arachnids have a few more appendages, chest markings, and now show off their chelicerae.
- add: Arachnids now have a few different organ sprites
- remove: Arachnids no longer have webs strapped to their appendages

PR #34381, opened 2025-01-11
fix arachnid lungs

### LINKED ISSUE #34364: Arachnid Lungs problem
## Description
I noticed a problem when I was studying the respiratory organs of the races. Namely, that arachnids have AnimalLungs instead of ArachnidLungs. What does not give one of the features of the race, namely, more frequent breathing compared to other races.
**Screenshot**
Снимок экрана 2025-01-10 204700

lzk228:
## About the PR

change arachnid lungs from animal to arachnid

## Why / Balance

fixes #34364

## Technical details

no

## Media

no

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Arachnids now have actual arachnid lungs, instead of animal.

### DIFF
--- a/Resources/Prototypes/Body/Prototypes/arachnid.yml
+++ b/Resources/Prototypes/Body/Prototypes/arachnid.yml
@@ -14,7 +14,7 @@
       part: TorsoArachnid
       organs:
         heart: OrganArachnidHeart
-        lungs: OrganAnimalLungs
+        lungs: OrganArachnidLungs
         stomach: OrganArachnidStomach
         liver: OrganArachnidLiver
         kidneys: OrganArachnidKidneys


themias:
I see in the original PR, the lungs are meant to use more oxygen. #18177
However, the lungs use an update interval, similar to what was removed recently from arachnid stomachs. #34298 

Do we need to update how the lungs work to achieve the desired effect?

frobnic8 (approved):
This is clearly the intended state.

ArtisticRoomba (dismissed):
`OrganArachnidLungs` needs to be adjusted to actually behave like how we want them to. The currently implementation of arachnid lungs doesn't do anything. Unsure if this is easy to do or not.

lzk228:
> `OrganArachnidLungs` needs to be adjusted to actually behave like how we want them to

how do you want them to behave?

TheShuEd:
@ArtisticRoomba You've been expected to respond for months

ArtisticRoomba:
Sorry for not responding. In testing arachnids seem to consume gas at the same rate as moths or humans, they run out of air from an emergency oxygen tank at the same time. Ideally we want arachnids to consume a bit more oxygen.

lzk228:
> Sorry for not responding. In testing arachnids seem to consume gas at the same rate as moths or humans, they run out of air from an emergency oxygen tank at the same time. Ideally we want arachnids to consume a bit more oxygen.

ok, got it, but i noticed that arachnid lungs actually have human metabolizer and after changing it to arachnid i don't really know how to make it consume more oxy

ArtisticRoomba:
I would assume that the `rateModifier` would change the rate at which gas is consumed, and its set way beyond the default, but nothing's changed, so I'm not sure where to go from here.

Princess-Cheeseballs:
Lungs currently consume 100% of the Oxygen which is inhaled so there's not any way to make Arachnids consume more Oxygen through the lungs with only yaml. 

Breathing rate and Breath volume are both stored in the respirator component and have nothing to do with the lungs.

For the desired result you have three options:
- 1: Add a way for lungs to modify either breathing rate or breath size with events (Would be appreciated API).
- 2: Decrease the breathing interval on the respirator component for Arachnids 
- 3: Increase the breath volume on the respirator component for Arachnids.

Errant-4:
Seeing as this does make a required change, I am going to merge it rather than keep it in limbo forever. 

It does not implement the difference in breathing, though, so I will keep the issue open and remove the CL. Nothing changed from a player perspective. 
Thank you for your contribution. If you or anyone wants to also try fixing the lung, just open a new PR and link the same issue

Errant-4 (approved):

PR #34387, opened 2025-01-11
Weapon resizing

Nox38:
## About the PR
Added custom dimensions to every firearm, for better realism. Slightly nerfs the satchel too, by making it less able to carry a full-length assault rifle.

## Why / Balance
This started as simply "why isn't the Enforcer 1x5?" and was quickly followed by other strange questions like "why is the assault rifle a square?". This has led me to change every single gun to have a somewhat more realistic and balanced inventory size.

## Technical details
Custom dimensions were added to the item component of every firearm.

## Media
image
image

## Requirements

**Changelog**
:cl: Nox38
- Add: All firearms now have custom inventory sizes.


Resources/Prototypes/Entities/Objects/Weapons/Guns/Snipers/snipers.yml
```diff
@@ -53,6 +57,10 @@
   id: WeaponSniperHristov
   description: A portable anti-materiel rifle. Fires armor piercing 14.5mm shells. Uses .60 anti-materiel ammo.
   components:
+  - type: Item
+    size: Huge
```
kosticia, line 61: Hristov bundle didn't have enough space to contain rifle and lawyer costume at one time

kosticia:
Great PR! But in my opinion satchels will be so much useless for antagonists and security, because biggest part of guns will be so much big for them.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Snipers/snipers.yml
Nox38: oopsies, ill fix that!

Nox38:
> Great PR! But in my opinion satchels will be so much useless for antagonists and security, because biggest part of guns will be so much big for them.

I'm not sure what guns would still work as 4 long besides the SMGs and the bulldog. The goal isn't to make the satchel useless for those roles, but make satchel vs backpack more of a serious question than satchel being better always. If you have any change suggestions let me know!

Emisse (approved):

IProduceWidgets:
You'll have to double check it works because I remember there being issues with some visualizers & storage sprite changes in the past, but the pistols would look a lot better offset. https://github.com/space-wizards/space-station-14/blob/5fff6bb1d2ae770e395c0d21ae3c1de409388826/Content.Shared/Item/ItemComponent.cs#L62

Nox38:
> You'll have to double check it works because I remember there being issues with some visualizers & storage sprite changes in the past, but the pistols would look a lot better offset.

I wasn't sure how to offset them, I will get on that now.

> Similarly, the mosin and hristov probably need non-diagonal sprites instead of just making the inv size bigger

unfortunately I do not know how to make/edit sprites, but I like the idea. 
The Hristov isn't a huge deal in my opinion because you'd need to disassemble it to fit it in a bag in the first place, explaining the square-er shape, but the Mosin could definitely use a horizontal sprite.

LaCumbiaDelCoronavirus:
I feel like the enforcer, bulldog and Drozd should keep their original size, as they took large amounts of space in bags to account for how good they are as weapons by themselves. The SMGs (other than drozd, but especially C20s) should really also be longer so that you can't fit, what, 6 of them in one satchel, let alone a bag.
Other than that it looks like a nice PR!

ArtisticRoomba:
> I feel like the point of gridinv was to standardise sizes, this is going in the opposite direction

The reason why GridInv has the same sizes for a lot of items is actually because they're inherited from the default sizes of small, medium, large, etc. Expanding the sizes into more unique shapes encourages creativity and funny tetris inventory management which is more fun IMO.

image
image

pheenty:
Syndie shoulder holster should definitely be changed somehow if this gets merged, something like a literal slot for a gun would work (like the jackboots have a slot for a pistol)

pheenty:
I think that captain's laser should stay L-shaped, fitting into the pocket

Nox38:
> I think that captain's laser should stay L-shaped, fitting into the pocket

It wouldnt fit the sprite that way, what if it was 2x2 but still fit in pockets? (I think thats how it is right now, but id have to check)

7-62x54r:
Satchel nerf real
> The Hristov folds up to just 3x5 to fit into a briefcase. The 28-inch barrel probably splits in half or something for this, idk.

Then let me fit it into the satchel??? It's literally taking less space in this PR (4 by 4 origianlly makes it 16 tiles big, now it's 15), but that space is _different_ so it won't fit.
SMGs being 2x3 is kinda strong btw, that 1x2 fits either another mag or ammo box or meds.
Sawed off being 1x3... makes sense. Don't like it but it's logical i guess (DB being 1x5 and sawed off looks like 2/3 of it without the barrel)
You know, if we are arguing about realism here, you wouldn't just put a rifle in the bag VERTICALLY so it would stick out, you put it diagonally, that's the feel i got from bigger guns being 4x4... tbh i would add a toggle for that in inspect menu (like "place in the bag diagonally/vetrically") so you could pick the ones you need. Satchels still get 4x4 and duffels get that one extra inventory tile from 3x5 and more fitting shape.
_please don't make hristov even worse, that gun sees no use as is..._

Nox38:
> _please don't make hristov even worse, that gun sees no use as is..._

Your right I suppose. I’ll make the Hrisrov 4x4 until one of the buffs for it gets finished. As for the fitting guns in vertically, sounds interesting, but also sounds exactly like what your back slot is for! Perhaps I could make guns fit in the belt slot instead?

Pumkin69:
i like the PR a LOT! However the advanced laser pistol should probably stay normal shaped as that fits the sprite better.

Nox38:
> i like the PR a LOT! However the advanced laser pistol should probably stay normal shaped as that fits the sprite better.

It uh... doesn't actually fit correctly with the pistol shape. It's sprite clips into other items.

pheenty:
I feel like Hristov should stay 4x4, you probably disassemble it to fit it into a bag anyway, also would make it actually unique in this as the other guns are no longer like that. Or at least 2x6, it doesn't look like 3x5 at all. Also use ItemComponent: StoredRotation for Mosin (and Hristov if you don't make it 4x4 again).

Nox38:
> I feel like Hristov should stay 4x4, you probably disassemble it to fit it into a bag anyway, also would make it actually unique in this as the other guns are no longer like that. Or at least 2x6, it doesn't look like 3x5 at all. Also use ItemComponent: StoredRotation for Mosin (and Hristov if you don't make it 4x4 again).

Yeah, I changed it back to 4x4. I just never updated the picture, ill fix that.

Nox38:
Updated the images and PR!

Nox38:
Code could be improved, and I can't find a better way to reset everything then closing and making a new PR.

PR #34404, opened 2025-01-12
Dynamic localization change.

c4llv07e:
## About the PR
Thanks to bhijn for the idea.

Introduces a cvar to change localization at runtime without requiring a client or server to reload.
Additionally, this allows to set localization in the configuration file without needing to rebuild it.

## Why / Balance
Downstreams would be very happy! No, really, hardcoded language isn't good, especially for multi-language builds.

## Media

https://github.com/user-attachments/assets/ef907cf4-ec06-41b3-8343-ebd37981e941

(Took an example localization from [corvax](https://github.com/space-syndicate/))

## Requirements

## Breaking changes
Downstream implementations must make minor updates to their localization code.
All locale specific code (like adding new functions) should be placed in the `OnCultureUpdate` function now.

Корвакс, прости

**Changelog**
No CL, no fun.


TheShuEd (approved):
Looks very good

Content.Shared/Localizations/ContentLocalizationManager.cs
```diff
@@ -46,8 +46,22 @@ public void Initialize()
              */
             var cultureEn = new CultureInfo("en-US");
 
+            if (!_loc.HasCulture(cultureEn))
+                _loc.LoadCulture(cultureEn);
+            _loc.SetFallbackCluture(cultureEn); // I don't think there's any reason to change the fallback culture.
             _loc.AddFunction(cultureEn, "MAKEPLURAL", FormatMakePlural);
             _loc.AddFunction(cultureEn, "MANY", FormatMany);
+
+            _cfg.OnValueChanged(CCVars.Language, OnCultureUpdate, true);
```
slarticodefast, line 55: Shouldn't this use `Subs.CVar` or is that not available in a manager?

TheShuEd:
uhh, how it works with https://github.com/space-wizards/RobustToolbox/pull/5227?

c4llv07e:
> uhh, how it works with [space-wizards/RobustToolbox#5227](https://github.com/space-wizards/RobustToolbox/pull/5227)?

For my surprise, it works really good. That PR allows to change client locale, and this change allows to change server locale. If both client and server switch to one locale, it predicts correctly. If client has different locale, it will first predict client's locale and then server will update it by its own locale.

I think it works really good, but I need to change the cvar name because of this.

c4llv07e:
@TheShuEd I think it will work. Functions now should be defined in `OnCultureUpdate` function, so it will be repopulated on each change. Don't see any problems with this, because locale changes not so often and function addition is not a resource intensive operation.

c4llv07e:
Daaamn... That's not good. I still think tt is an RT bug, because `EntryPoint.Init` uses uninitialized engine managers (`LocalizationManager`) and it tries to use nulled sawmill.

Arctic-Husky:
Would it be a bad idea to just do a `_loc.Initialize()` in ContentLocalizationManager Initialize method?

c4llv07e:
> Would it be a bad idea to just do a `_loc.Initialize()` in `ContentLocalizationManager` `Initialize` method?

It would be a bad idea, because loc is an engine manager and content loc is a content manager. It will either require to make it default behaviour to init engine managers inside content (which is really bad). Or initing it twice (first in content loc, second in RT itself). But we can't guarantee that twice initializing loc won't have any side effects in the future.

> It would be better if there was a way to check if the LocalizationManager is initialized before running `_loc.ReloadLocalizations();`

There's some ways to check it, but only with a lot of duct tape.

IMHO, the best way to fix it is to initialize RT managers before content init, so when execution goes into the game code, it can't break anything uninitialized.

c4llv07e:
@metalgearsloth sorry for ping, but I don't know who to ask at this point. Is it normal that RT managers are uninitialized during content init? If so, how can it be checked not to trigger an execution of uninitialized parts?

PR #29545, opened 2024-06-28
Remove rads from bananium

DogZeroX:
webedit ops

## About the PR

This allows people to power the station infinitely with rad absortion with absolutely no drawbacks.
Also it's annoying to deal with. (I'm surprised bananium has it but not literal uranium, wouldn't want that anyways, not yet at least)
## Media

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- remove: Bananium no longer emit rads

PR #34436, opened 2025-01-14
Make radioactive material radioactive

minus1over12:
## About the PR

Uranium is now weakly radioactive.
## Why / Balance

People expect uranium to be radioactive.
## Technical details

Yaml was changed to make uranium and related entities radioactive. ~~The ore rocks look like they have high intensity in code (2.3), but are actually very low because the rock's radiation shielding weakens it immensely.~~
## Media

Uranium in hand producing radiation

## Requirements

## Breaking changes

**Changelog**

:cl: WarPigeon
- add: Uranium is now slightly radioactive.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/pizza.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/pizza.yml
@@ -581,6 +581,9 @@
           Quantity: 4
         - ReagentId: Uranium
           Quantity: 16
+  - type: RadiationSource
+    intensity: 0.1
+    slope: 3
 
 - type: entity
   name: slice of spicy rock pizza
@@ -616,6 +619,9 @@
           Quantity: 0.5
         - ReagentId: Uranium
           Quantity: 2
+  - type: RadiationSource
+    intensity: 0.1
+    slope: 3
 
 # Tastes like crust, tomato, cheese, radiation.
 
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/pizza.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/pizza.yml
@@ -583,7 +583,7 @@
           Quantity: 16
   - type: RadiationSource
     intensity: 0.1
-    slope: 3
+    slope: 300
 
 - type: entity
   name: slice of spicy rock pizza
@@ -621,7 +621,7 @@
           Quantity: 2
   - type: RadiationSource
     intensity: 0.1
-    slope: 3
+    slope: 300
 
 # Tastes like crust, tomato, cheese, radiation.
 
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/pizza.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/pizza.yml
@@ -583,7 +583,7 @@
           Quantity: 16
   - type: RadiationSource
     intensity: 0.1
-    slope: 300
+    slope: 3
 
 - type: entity
   name: slice of spicy rock pizza
@@ -621,7 +621,7 @@
           Quantity: 2
   - type: RadiationSource
     intensity: 0.1
-    slope: 300
+    slope: 3
 
 # Tastes like crust, tomato, cheese, radiation.
 
--- a/Resources/Prototypes/Entities/Structures/Walls/asteroid.yml
+++ b/Resources/Prototypes/Entities/Structures/Walls/asteroid.yml
@@ -275,6 +275,9 @@
           state: rock_asteroid_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
 
 - type: entity
   id: AsteroidRockUraniumCrab
@@ -284,6 +287,9 @@
     - type: OreVein
       oreChance: 1.0
       currentOre: OreUraniumCrab
+    - type: RadiationSource
+      intensity: 2.4
+      slope: 0.1
 
 - type: entity
   id: AsteroidRockBananium
@@ -564,6 +570,9 @@
           state: ironrock_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
 
 - type: entity
   id: IronRockBananium
@@ -937,6 +946,9 @@
         state: rock_west
       - state: rock_uranium
         map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+  - type: RadiationSource
+    intensity: 2.3
+    slope: 0.1
 
 
 - type: entity
@@ -1234,6 +1246,9 @@
           state: rock_wall_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
 
 
 - type: entity
@@ -1510,6 +1525,9 @@
           state: rock_snow_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
 
 
 - type: entity
@@ -1786,6 +1804,9 @@
           state: rock_sand_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
 
 - type: entity
   id: WallRockSandBananium
@@ -2061,6 +2082,9 @@
           state: rock_chromite_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
 
 
 - type: entity
@@ -2337,6 +2361,9 @@
           state: rock_andesite_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
 
 
 - type: entity
--- a/Resources/Prototypes/Entities/Structures/Walls/asteroid.yml
+++ b/Resources/Prototypes/Entities/Structures/Walls/asteroid.yml
@@ -275,9 +275,6 @@
           state: rock_asteroid_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
-    - type: RadiationSource
-      intensity: 2.3
-      slope: 0.1
 
 - type: entity
   id: AsteroidRockUraniumCrab
@@ -287,9 +284,6 @@
     - type: OreVein
       oreChance: 1.0
       currentOre: OreUraniumCrab
-    - type: RadiationSource
-      intensity: 2.4
-      slope: 0.1
 
 - type: entity
   id: AsteroidRockBananium
@@ -570,9 +564,6 @@
           state: ironrock_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
-    - type: RadiationSource
-      intensity: 2.3
-      slope: 0.1
 
 - type: entity
   id: IronRockBananium
@@ -946,9 +937,6 @@
         state: rock_west
       - state: rock_uranium
         map: [ "enum.MiningScannerVisualLayers.Overlay" ]
-  - type: RadiationSource
-    intensity: 2.3
-    slope: 0.1
 
 
 - type: entity
@@ -1246,9 +1234,6 @@
           state: rock_wall_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
-    - type: RadiationSource
-      intensity: 2.3
-      slope: 0.1
 
 
 - type: entity
@@ -1525,9 +1510,6 @@
           state: rock_snow_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
-    - type: RadiationSource
-      intensity: 2.3
-      slope: 0.1
 
 
 - type: entity
@@ -1804,9 +1786,6 @@
           state: rock_sand_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
-    - type: RadiationSource
-      intensity: 2.3
-      slope: 0.1
 
 - type: entity
   id: WallRockSandBananium
@@ -2082,9 +2061,6 @@
           state: rock_chromite_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
-    - type: RadiationSource
-      intensity: 2.3
-      slope: 0.1
 
 
 - type: entity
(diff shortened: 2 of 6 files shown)


Aeshus (commented):
Thank for the contribution!

**praise:** Your PR looks good, and I don't see any issues that would block this code-wise.

**issue (blocking):** I am wondering the problems that were brought up in https://github.com/space-wizards/space-station-14/pull/29545 have been addressed.

Bananium used to be radioactive, but it was removed as you could create a pile of them next to a radiation collector for essentially infinite power.

That PR even mentions the lack of uranium's radioactivity, and states that's a good thing, as it would have the same infinite power problem.

**issue (non-blocking):** Also, is there any thing that may require mappers may need to think about when placing the uranium on a map? 

For maps that place it in the engineering vault, they made need to have extra considerations like moving it away from the plasma to minimize radiation damage.  

**suggestion (non-blocking):** It also seems that lighting is inconsistent between the uranium sources. I believe that if it is radioactive, it should glow green as an indicator. Some of the uranium sources already do this, but not all (e.g. the sheets).

**question:** Also, is there any formula behind the slope values, or was that simply something you tested and worked well?

minus1over12:
> **praise:** Your PR looks good, and I don't see any issues that would block this code-wise.

Thanks!

> **issue (blocking):** I am wondering the problems that were brought up in #29545 have been addressed.
> 
> Bananium used to be radioactive, but it was removed as you could create a pile of them next to a radiation collector for essentially infinite power.
> 
> That PR even mentions the lack of uranium's radioactivity, and states that's a good thing, as it would have the same infinite power problem.

I will test with this a bit and get back to you. I can scale down the stackable item's radiation values if it turns out to be useful for power.

> **thought:** Also, is there any thing that may require mappers may need to think about when placing the uranium on a map?
> 
> For maps that place it in the engineering vault, they made need to have extra considerations like moving it away from the plasma to minimize radiation damage.

Engineers have radiation resistant hard suits to slow down the damage, and I also tried to address this using the slope values. Mappers may want to consider moving engineering's uranium into a lead box to neutralize the radiation threat.

> **suggestion (non-blocking):** It also seems that lighting is inconsistent between the uranium sources. I believe that if it is radioactive, it should glow green as an indicator. Some of the uranium sources already do this, but not all (e.g. the sheets).

I can add this.

> **question:** Also, is there any formula behind the slope values, or was that simply something you tested and worked well?

I set $\text{intensity}$ to $0.1$ since that is the smallest number in game Geiger counters show.

For $\text{slope}$, the goal was to meet a player expectation (uranium is radioactive) without making it too useful as a [demon core](https://www.youtube.com/watch?v=BdeG2iDIICA). I went with an assumption on how the game is doing the calculation ($\text{rads} = \text{intensity} - \text{slope} * \text{distance}$) and the thought that a full stack of uranium should not have an effect more than $1$ tile away; hence the slope of $3$ since I was using an intensity of $0.1$. From testing, this seemed to be correct. I am highly open to nerfing this further if 1 tile is too much.

minus1over12:
Changes the slope from 3 to 300 on the items, which:
* Makes mapping no longer a concern, as uranium is now only dangerous when held.
* Prevents the demon core scenario.
* Makes it not useful for power.

minus1over12:
Added the glow effect from the ore item to the sheets. I am leaving the rocks dark since I don't want to detract from the value of the mineral scanner.

Uranium glowing

image

HyperB1:
And radium (which is a million times more radioactive than uranium) still emits no rads...

MozarteanChaos:
> And radium (which is a million times more radioactive than uranium) still emits no rads...

As far as I'm aware, radium currently only exists as a liquid reagent, rather than a solid material. I don't think any liquids passively emit radiation at present, and it doesn't look like this PR changes that.

slarticodefast:
Personally I like there being some more dangers to watch out for when mining. And at the moment we don't have a lot of radiation sources.
The problem when bananium was radioactive was that carrying a bag full of it around could be used to quickly and stealthily perma kill someone without them noticing. Ideally there would be some sort of warning sign of radiation poisoning, like vomiting at a certain damage threshold so you don't have to keep an eye on your health bar at all times.
Making uranium weakly radioactive should be fine though in my opinion.

ilovehans10:
> Ideally there would be some sort of warning sign of radiation poisoning, like vomiting at a certain damage threshold so you don't have to keep an eye on your health bar at all times.

**you taste metal**
**you suddenly feel hot**
**your muscles spasm uncontrollably**
**you feel lightheaded**
**you feel nauseous**
**a clump of your hair falls out**

I tried to base all of these off of symptoms shown by victims that were directly exposed to high dosages of radiation like [Cecil Kelley](https://en.wikipedia.org/wiki/Cecil_Kelley_criticality_accident).

Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/pizza.yml
```diff
@@ -581,6 +581,9 @@
           Quantity: 4
         - ReagentId: Uranium
           Quantity: 16
+  - type: RadiationSource
+    intensity: 0.1
+    slope: 300
```
Aeshus, line 586: **nitpick (non-blocking):** After testing in-game, it seemed that a slope of 50 seems to be the minimum required to stop it from generating power. If you feel like changing, make sure to test it first. ;-)

Aeshus:
> > Ideally there would be some sort of warning sign of radiation poisoning, like vomiting at a certain damage threshold so you don't have to keep an eye on your health bar at all times.
> 
> **you taste metal** **you suddenly feel hot** **your muscles spasm uncontrollably** **you feel lightheaded** **you feel nauseous** **a clump of your hair falls out**
> 
> I tried to base all of these off of symptoms shown by victims that were directly exposed to high dosages of radiation like [Cecil Kelley](https://en.wikipedia.org/wiki/Cecil_Kelley_criticality_accident).

I think that would be better suited for another PR.

Resources/Prototypes/Entities/Structures/Walls/asteroid.yml
```diff
@@ -275,6 +275,9 @@
           state: rock_asteroid_west
         - state: rock_uranium
           map: [ "enum.MiningScannerVisualLayers.Overlay" ]
+    - type: RadiationSource
+      intensity: 2.3
+      slope: 0.1
```
Aeshus, line 280: **thought:** Theoretically this can be used to generate infinite power, but it is significantly easier to get a damaged RTG, so I think this is acceptable. The amount of setup needed isn't worth the amount of power.

Plus, it's generally in inconvenient places, so it's a balance of crew risk verses infinite power.

Aeshus (approved):
Thanks for making my requested changes!

**praise:** I think the changes all look good, and I couldn't create infinite power with the sheets when testing in-game.

**question (non-blocking):** Should we make uranium glass and its descendants radioactive as well (e.g. uranium shiv)? This should probably be done in a separate PR, as that may have more balance ramifications. 

**note:** And as slarticodefast and ilovehans10 brought up, the lack of warning signs of radiation is quite a problem.

I think that should be done in a different PR as this has been a problem for quite a while and isn't directly related to this change. Things like the Singularity and Cancer Mouse are highly deadly in their radiation damage, yet don't have any direct symptoms.

Just like the other radiation sources, I believe having our current visual indicator (Cancer Mouse glows green, Singularity is the singularity) should be sufficient for now.

slarticodefast:
Yeah, I guess the warning signs from radiation damage should be done in a separate PR.
I did not test it in-game yet, but a slope of 300 sounds a little too extreme if the falloff is linear.
- In case you do want to use radiation for something, for example triggering artifacts or giving someone radiation damage you would need perfect accuraccy to the point where it may feel bugged.
- This might actually make it more dangerous because the geiger counter will no longer detect it
- I think an effective radius of roughly one tile is reasonable

Aeshus:
> Yeah, I guess the warning signs from radiation damage should be done in a separate PR. I did not test it in-game yet, but a slope of 300 sounds a little too extreme if the falloff is linear.
> 
> * In case you do want to use radiation for something, for example triggering artifacts or giving someone radiation damage you would need perfect accuracy to the point where it may feel bugged.

Yeah, that’s true. I think this would need to be solved as it seems pretty important to not cause unneeded confusion.

Currently it’s kept so low as to stop it from generating power using the radiation collectors. The underlying problem wasn’t solved since Bananium’s radioactivity removal. A singular stack, when split into a pile, can basically power a whole station.

> * This might actually make it more dangerous because the geiger counter will no longer detect it

Geiger counters that are on your person will still detect it just fine, and you are in no danger if you are not directly holding them.

> * I think an effective radius of roughly one tile is reasonable

I think if we could stop the power generation problem, this would be the appropriate distance.

slarticodefast:
> Currently it’s kept so low as to stop it from generating power using the radiation collectors. The underlying problem wasn’t solved since Bananium’s radioactivity removal. A singular stack, when split into a pile, can basically power a whole station.

I don't think it's that much of a problem in this case since the intensity is relatively low compared to bananium and with all the uranium you could jsut run a bunch of super pacmans instead, which would also power the whole station.

Aeshus:
> > Currently it’s kept so low as to stop it from generating power using the radiation collectors. The underlying problem wasn’t solved since Bananium’s radioactivity removal. A singular stack, when split into a pile, can basically power a whole station.
> 
> I don't think it's that much of a problem in this case since the intensity is relatively low compared to bananium and with all the uranium you could just run a bunch of super pacmans instead, which would also power the whole station.

True. I think Bananium was intensity 1, while this is only intensity 0.1.

A Slope 3 should be fine then, as that seemed to have the desired effect earlier. 

I'll go look into maybe seeing how we can maybe whitelist radiation sources for powergen as there's also the damaged RTG problem for infinite power.

minus1over12:
Brought the slope back to 3, if it becomes an issue we can always make a hotfix to bring it up to 50.

Optimally we would have 3 different types of radiation (alpha, beta, gamma) and change the effects based on that. However, that is out of scope for this PR.

Aeshus (changes requested):
**issue (blocking):** After discussing with some other maintainers, there may be an issue with the fact that the uranium rocks may harm the mobs on the vgroid if they happen to spawn in proximity to it.

This is probably for the best as the current radiation system tries to send rays from the radiation sources to all mobs within 50 tiles, so this could potentially add unnecessary strain to the server.

Slarticodefast suggested that it can just be explained by the rocks shielding the radiation.

**suggestion:** The solution is just (unfortunately) deleting the rocks from the patch.

Alpaccalypse:
FYI, this inspired me to try some things myself and I can across what seems to be a pretty consistent bug, that if you have intensity set to 0.1, no radiation seems to be generated at all, or at least you take 0 toxin damage and the geiger counter doesn't register anything. #34456

Aeshus:
> FYI, this inspired me to try some things myself and I can across what seems to be a pretty consistent bug, that if you have intensity set to 0.1, no radiation seems to be generated at all, or at least you take 0 toxin damage and the geiger counter doesn't register anything. #34456

I think we've tracked the reason down, though I can confirm that the `0.1` is the cutoff. Any stacks larger than 1 will work fine, but if it is exactly 1, it will not register any radiation. 

The solutions would either be to change the `<=` in the link to just `<` (to be more in line with the describing CVAR comment, see issue for info), or just set all of the intensities to any value greater than `1.0f`.

https://github.com/space-wizards/space-station-14/blob/e094b79ee4636cbd91e760091420017e81222ebe/Content.Server/Radiation/Systems/RadiationSystem.GridCast.cs#L136-L137

It may be a little out of scope for this no C# PR, but the change seems small enough and is more of a bug-fix that I think it could be fine (though I can't really make any definitive statement).

ElectroJr:
While it'd be nice feature, and more realistic, for uranium to be radioactive, I don't think it should be made radioactive, mainly because it'd probably be pretty bad for performance if there's a decent number of sheets and ores around the station (unless its much rarer than I assume it is).

It'd probably require a radiation rewrite. Either to make it better handle many sources by using some kind of radiation tilemap instead of iterating over sources & receivers, or at the very least to rewrite so that it no longer assumes that there's more receivers than sources. E.g., at the moment it needs to call `Irradiate()` 6400 times per radiation tick if there's 80 players & 80 sheets.

A compromise might be to make it radioactive only if it is also unstable and rapidly decays away to effectively limit the total number that can exist at any one time, but that obviously also has significant balance consequences and removes some of the realism.

minus1over12:
> there may be an issue with the fact that the uranium rocks may harm the mobs on the vgroid if they happen to spawn in proximity to it.

Valid consideration. While I think it would make sense for space creatures to have evolved resistance to solar radiation (and by extension what Uranium can put out), I'll leave that for a different place and remove the radiation from the rocks.

> the current radiation system tries to send rays from the radiation sources to all mobs within 50 tiles

While not something I am going to touch in this PR, I don't see why it is hardcoded for 50 instead of calculating the maximum effective distance and using that as a limit.

> mainly because it'd probably be pretty bad for performance if there's a decent number of sheets and ores around the station (unless its much rarer than I assume it is).

I don't see sheets and ores around that often. Engineering always gets a stack at round start, but otherwise it's limited to what salvage can get from mining. And this if anything will encourage people to put uranium into lathes instead of leaving it out to irradiate people.

> It'd probably require a radiation rewrite.

Based on the hardcoded 50 tile range issue, the code probably needs an optimization pass over it anyway.

> It may be a little out of scope for this no C# PR

I vote that C# is out of scope for no C# PRs.

minus1over12:
Update: the code does check for if the ray will have any effect right after it checks the 50 tile distance limit

```csharp
        // get direction from rad source to destination and its distance
        var dir = destWorld - sourceWorld;
        var dist = dir.Length();

        // check if receiver is too far away
        if (dist > GridcastMaxDistance) // ← This checks if the entities are more than 50 tiles apart.
            return null;

        // will it even reach destination considering distance penalty
        var rads = incomingRads - slope * dist;

        // Apply rad modifier if the source is enclosed within a radiation blocking container
        rads = GetAdjustedRadiationIntensity(sourceUid, rads);

        if (rads <= MinIntensity) // ← Distance effectiveness check is here. Also slightly incorrect as @Aeshus pointed out.
            return null;
```

slarticodefast:
> I don't see sheets and ores around that often. Engineering always gets a stack at round start, but otherwise it's limited to what salvage can get from mining. And this if anything will encourage people to put uranium into lathes instead of leaving it out to irradiate people.

Might be fine since we did not have those problems with bananium either, when it was radioactive

Aeshus:
Electro has made a PR for some radiation system changes that should hopefully alleviate some of the performance concerns.

#34459

ElectroJr:
> I don't see sheets and ores around that often. Engineering always gets a stack at round start, but otherwise it's limited to what salvage can get from mining. And this if anything will encourage people to put uranium into lathes instead of leaving it out to irradiate people.

I didn't take into account that this would be pretty effective at incentivising people to get rid of sheets as quickly as possible, so that would definitely help. I don't really have a feel for how common uranium sheets/ore are, but if they're really not all that commonplace it's probably fine? I just checked midround on lizard, and it seems like there were only 10 so it seems ok

SS14 Loader_BjAzXMU92u


> Update: the code does check for if the ray will have any effect right after it checks the 50 tile distance limit, so it will stop itself from calculating rays needlessly.

Currently that code is actually abysmally slow for what it should be doing, but that's a separate issue. Though even if if a check is cheap, it can still hurt performance just because it gets run a large number of times, especially if it scales badly like radiation does.

> I vote that C# is out of scope for no C# PRs.

Sure, but, that could still result in a PR being blocked until C# issues get fixed. Though like I said probably fine here.

Djungelskog2:
is it still dangerous if dragged or shoved in a locker or crate near you? I feel that should probably be a thing

Aeshus (approved):
Thanks for working on this!

The maintainer discussion was unanimous in approving this conceptually, so this is ready to be merged once a second maintainer checks this over and grants approval!

slarticodefast (approved):

githubuser4141:
Is it actually realistic most of google tends to agree it's alpha radiation which is blocked by skin and it's really only dangerous if you happen to breathe it in or eat the dust

deltanedas:
no

minus1over12:
> Is it actually realistic most of google tends to agree it's alpha radiation which is blocked by skin and it's really only dangerous if you happen to breathe it in or eat the dust

You're right. However, we also don't have any way to differentiate between radiation types currently.

PR #34441, opened 2025-01-14
Fix glass duplication with beakers

Alice4267:
Fixing glass dupe.

## About the PR
I have changed the amount of glass sheets required to create beakers to the 2 glass sheets by default instead of one. Large beakers still require twice as much as small beakers.

## Why / Balance
This minor change is made with the single purpose of preventing so-called glass dupe. Since hyperconvectional autolathes need only half of the resources of the original receipt, it made glassware cost only 0.5 glass sheets. At the same time, glass shards that could be obtained through breaking glassware still can be turned into glass sheets by using a welding tool. It made it so that you spent 1 sheet of glass and got back 2. This abuse can basically create an infinite amount of building resources and money completely for free. I do believe that it is a glitch and wasn't originally meant, so I decided to fix it.

## Technical details
Simple change of numbers.

## Media
No needed.

## Requirements

## Breaking changes
Only minor changes.

**Changelog**
:cl:
- tweak: Changed the amount of glass sheets needed to create beakers to 2 on normal autolathe and 1 on hyperconvectional autolathe.
- tweak: Changed the amount of glass sheets needed to create large beakers to 4 on normal autolathe and 2 on hyperconvectional autolathe.

### DIFF
--- a/Resources/Prototypes/Recipes/Lathes/chemistry.yml
+++ b/Resources/Prototypes/Recipes/Lathes/chemistry.yml
@@ -3,14 +3,14 @@
   result: Beaker
   completetime: 2
   materials:
-    Glass: 100
+    Glass: 200
 
 - type: latheRecipe
   id: LargeBeaker
   result: LargeBeaker
   completetime: 2
   materials:
-    Glass: 200
+    Glass: 400
 
 - type: latheRecipe
   id: CryostasisBeaker


pheenty:
Just make a beaker spawn 0 to 1 glass shards on destruction instead

SpaceRox1244:
> Since hyperconvectional autolathes need only half of the resources of the original receipt, it made glassware cost only 0.5 glass sheets.

Maybe the real problem was progression systems all along.....

ElectroJr:
> Maybe the real problem was progression systems all along.....

The real real problem is people changing how lathe progression works without updating the tests that specifically exist to prevent this sort of thing, and then giving lathes a 50% discount on materials. Though the test seemingly never accounted for welder refinable resources either

ElectroJr:
There are a bunch more similar issues (see #34449). I don't know what the best fix is, maybe the 50% discount should just be nerfed. But if people want to keep it, there's a lot more recipes or destruction spawns that need fixing.

PR #34444, opened 2025-01-15
Muzzles reduce emote sound

themias:
## About the PR

This PR extends the new separated muzzle accent by reducing the volume and distance of vocal emote sounds, (screaming, crying, laughing, etc.)

## Why / Balance

Someone wearing a muzzle has their speech muffled, but they are able to scream their lungs out. This change is intuitive and useful for keeping kidnapped crew and unruly prisoners unheard.

## Technical details

We listen to the EmoteEvent and simply pass audio params when playing vocal emotes.
Ideally it would sound 'muffled' rather than just quieter, but that would require many new sound variations.

## Media

https://github.com/user-attachments/assets/6a96e3ec-51a2-405a-a0bc-196b410a644c

## Requirements

**Changelog**
:cl:
- tweak: Wearing a muzzle now reduces the sound of vocal emotes


SlamBamActionman (approved):
Tested ingame and works as advertised. Non-spoken emotes (tail thumps, clapping etc) aren't muted, which is nice.
I think that there is an argument to be made that we might wanna fully disable sound emotes (like laughing, crying) when wearing a muffle, but this works too.

It touches chatcode, but specifically emote code, so I can approve it despite the restriction.

Content.Server/Speech/Components/MumbleAccentComponent.cs
```diff
@@ -1,7 +1,14 @@
...
 
 [RegisterComponent]
 public sealed partial class MumbleAccentComponent : Component
 {
-
+    /// <summary>
+    /// This modifies the audio parameters of emote sounds, screaming, laughing, etc.
+    /// By default, it reduces the volume and distance of emote sounds.
+    /// </summary>
+    [DataField, ViewVariables]
```
MilonPL, line 12: The ViewVariables attribute is redundant, you can just use DataField

Content.Server/Speech/EntitySystems/MumbleAccentSystem.cs
```diff
@@ -1,16 +1,33 @@
...
 using Content.Server.Speech.Components;
+using Content.Shared.Chat.Prototypes;
+using Content.Shared.Speech.Components;
 
 namespace Content.Server.Speech.EntitySystems;
 
 public sealed class MumbleAccentSystem : EntitySystem
 {
     [Dependency] private readonly ReplacementAccentSystem _replacement = default!;
+    [Dependency] private readonly ChatSystem _chat = default!;
```
MilonPL, line 11: Dependencies should be sorted alphabetically.

Content.Server/Speech/EntitySystems/MumbleAccentSystem.cs
```diff
@@ -1,16 +1,33 @@
...
 
     public override void Initialize()
     {
         base.Initialize();
 
         SubscribeLocalEvent<MumbleAccentComponent, AccentGetEvent>(OnAccentGet);
+        SubscribeLocalEvent<MumbleAccentComponent, EmoteEvent>(OnEmote, before: new[] { typeof(VocalSystem) });
```
MilonPL, line 18: I think a collection expression would be cleaner here, `[typeof(VocalSystem)]`

Content.Server/Speech/EntitySystems/MumbleAccentSystem.cs
```diff
@@ -1,16 +1,33 @@
...
+    }
+
+    private void OnEmote(EntityUid uid, MumbleAccentComponent component, ref EmoteEvent args)
```
MilonPL, line 21: use the `Entity<T> ent, ref event args` pattern for event handlers. Here it would be `Entity<MumbleAccentComponent>`. You can then get the EntityUid with `ent` and the component with `ent.Comp`.

slarticodefast: not resolved yet

Content.Server/Speech/EntitySystems/MumbleAccentSystem.cs
```diff
@@ -1,25 +1,42 @@
...
 {
+    [Dependency] private readonly ChatSystem _chat = default!;
     [Dependency] private readonly ReplacementAccentSystem _replacement = default!;
 
     public override void Initialize()
     {
         base.Initialize();
 
         SubscribeLocalEvent<MumbleAccentComponent, AccentGetEvent>(OnAccentGet);
+        SubscribeLocalEvent<MumbleAccentComponent, EmoteEvent>(OnEmote, before: [typeof(VocalSystem)]);
```
slarticodefast, line 18: Wouldn't this subscription conflict with the one in MutingSystem when handled in the wrong order? MutingSystem might want to cancel it, but the mumbe accent would already play the sound before that.

slarticodefast (dismissed):

Content.Server/Speech/EntitySystems/MumbleAccentSystem.cs
themias: oops

Content.Server/Speech/EntitySystems/MumbleAccentSystem.cs
themias: I wasn't able to get it to make sound while muted, although I've now updated the MutingSystem subscription to hopefully prevent that scenario from ever happening.

PR #34446, opened 2025-01-15
Mjollnir and Singularity Hammer for Wizard

keronshb:
## About the PR

This is going to require a few different PRs. I'm atomizing it, but leaving this as the master PR so I'll be adding the requires as I go through.

Major Changes -
Mjollnir and Singularity Hammers were added! Both wizard items,

Mjollnir - A hammer that stuns and knocks back the targets it hit until they hit a wall. Also works when thrown!
Singularity Hammer - Attracts everyone and every item nearby even on a miss!

WieldComponent - noUseDelayOnWield - conditional to not activate use delay on wield if configured to do so. Useful where you have a use delay tied to something melee wise and you don't want wield to activate it.

MeleeThrowOnHit  - Added stuns and the ability to activate on throw.

UseDelayOnMeleeHit - Activates use delay on melee hit. Useful where you might want an ability to trigger on melee hit but don't want it spammed. Also works on throw hit.

## Why / Balance

Two wizard items ported from SS13. Needed to flesh out the wizard more before release.
Both hammers require wielding to melee. Mjollnir gets away with this a bit by being able to activate the effect on throw.

The other components are nice to have, especially UseDelayOnMeleeHit and NoUseDelayOnWield.

Why Repulse Attract instead of using Gravity Well? I needed something to work with the _throw system and not activate on every tick. It's way more instant. Also this will be useful for the Repulse spell I plan on adding to Wizards.

## Technical details

The RepulseAttractSystem works by getting all entities in range and either pulling them towards the user or pushing them away, depending how it's configured.

UseDelayOnMeleeHit activates when a melee hit occurs or when a throw hit occurs and it triggers UseDelay.

MeleeThrowOnHit now has stun support and now subscribes to the on throw hit event to work on hit if configured.

WieldComp now supports toggling usedelay off.

## Media

https://github.com/user-attachments/assets/fd449ff1-7d6f-4d84-812d-e85e2d24a3d2

## Requirements

## Breaking changes

- UseDelayOnWield - conditional to not activate use delay on wield if configured to do so. Useful where you have a use delay tied to something melee wise and you don't want wield to activate it.
- MeleeThrowOnHit - Added stun conditional. Added throw conditional and support to activate on throw. Enabled field was removed entirely and AttemptMeleeThrowOnHitEvent handles the check.  Lifetimes & Endtimes removed. Distance added instead. Uses throw system internally now.
- Wield- Added support to disable use delay on wield
- Sledgehammers.yml was renamed to hammers.yml
- removed MeleeThrownComponent
- removed MeleeThrowOnHitEndEvent, use LandEvent instead
- IsDelayed takes in a nullable comp and resolves it.

**Changelog**


Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            if (_whitelist.IsWhitelistFail(comp.Whitelist, target) || _whitelist.IsBlacklistPass(comp.Blacklist, target))
+                continue;
+
+            var targetXForm = Transform(target);
+
+            if (targetXForm.Anchored || HasComp<GhostComponent>(target))
+                continue;
+
+            var userWorldPos = _xForm.GetWorldPosition(xForm);
```
slarticodefast, line 81: can be done outside the loop

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+            var targetWorldPos = _xForm.GetWorldPosition(target);
+
+            var direction = targetWorldPos - userWorldPos;
+
+            if (comp.Attract)
+                direction = userWorldPos - targetWorldPos;
+
+            _throw.TryThrow(target, direction, comp.Strength, doSpin: true);
```
slarticodefast, line 89: You are using the parameters for the function wrong.
The length of the `direction` vector matters, it decides how far an item is thrown until it lands.
What you call strength is the velocity of the throw.
And with how you calculate the direction vector this means items get thrown further if they are further away from the epicenter (in case you copied this from explosions, they do this wrong as well).

slarticodefast (changes requested):
I did not look at the rest of the code yet.

slarticodefast (approved):
Will need some more testing with my changes and art approval from emisse.
The breaking changes section needs to be updated.

keronshb:
Breaking Changes changed no response from Emisse.

Emisse (approved):

Emisse:
art approval

Content.Shared/RepulseAttract/RepulseAttractComponent.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityWhitelist? Whitelist;
+
+    /// <summary>
+    ///     What kind of entities should be excluded from the effect?
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityWhitelist? Blacklist;
+}
```
metalgearsloth, line 41: Isn't this just a duplicate of gravitywell?

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+            // attract: throw all items directly to to the epicenter
+            // repulse: throw them up to the maximum range
+            var throwDirection = attract ? -direction : direction.Normalized() * (range - direction.Length());
+
+            _throw.TryThrow(target, throwDirection, speed);
+        }
+
+        return true;
+    }
+}
```
metalgearsloth, line 80: See above on gravitywell, at the very least gravitywell just needs extension for what's needed and optimising.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
```diff
@@ -15,113 +16,72 @@ namespace Content.Shared.Weapons.Melee;
...
-
-            if (!CanThrowOnHit(ent, hit))
-                continue;
-
-            if (comp.UnanchorOnHit && HasComp<AnchorableComponent>(hit))
-            {
-                _transform.Unanchor(hit, Transform(hit));
-            }
+        if (_delay.IsDelayed(weapon.Owner))
+            return;
```
metalgearsloth, line 38: This should be cancelling the entire melee hit I would think.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
```diff
@@ -15,113 +16,72 @@ namespace Content.Shared.Weapons.Melee;
...
-        var ev = new AttemptMeleeThrowOnHitEvent(target);
-        RaiseLocalEvent(uid, ref ev);
-
-        if (ev.Handled)
-            return !ev.Cancelled;
-
-        return comp.Enabled;
-    }
+        if (direction == Vector2.Zero)
+            return;
```
metalgearsloth, line 77: This should really be an extension of throwing given how generic is rather than having each system having to handle it.

Content.Shared/Wieldable/SharedWieldableSystem.cs
```diff
@@ -166,6 +166,9 @@ private void OnUseInHand(EntityUid uid, WieldableComponent component, UseInHandE
             args.Handled = TryWield(uid, component, args.User);
         else if (component.UnwieldOnUse)
             args.Handled = TryUnwield(uid, component, args.User);
+
+        if (HasComp<UseDelayComponent>(uid) && !component.UseDelayOnWield)
+            args.ApplyDelay = false;
```
metalgearsloth, line 171: Bit suss on why we're even mixing usedelay into wieldable and not using a 3rd system? Now forks are reliant upon keeping the 2 as they're coupled.

Content.Shared/RepulseAttract/RepulseAttractComponent.cs
keronshb: It's not. Gravity well is consistent, this is using throw as a backend.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
keronshb: Same as above.

Content.Shared/Wieldable/SharedWieldableSystem.cs
keronshb: idk they've always been coupled im just adding a toggle to bypass the use delay.
There's different uses where you want use delay to either activate on wield or activate on something else.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
keronshb: Wouldn't that be handled in the melee system? This isn't really like "if use delay is active don't melee" it's "if use delay is active don't throw this person"

Content.Shared/RepulseAttract/RepulseAttractComponent.cs
```diff
@@ -0,0 +1,41 @@
...
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(RepulseAttractSystem))]
+public sealed partial class RepulseAttractComponent : Component
+{
+    /// <summary>
+    ///     Attracts if true, Repulse if false.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Attract;
+
```
metalgearsloth, line 17: This can just be derived from whether speed is positive or negative.

Content.Shared/RepulseAttract/RepulseAttractComponent.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float Speed = 1.0f;
+
+    /// <summary>
+    ///     How close do the entities need to be?
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float Range = 1.0f;
+
```
metalgearsloth, line 29: Placeholder values should instead be marking the datafield as required instead; neither of these defaults are being used.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+    private void OnMeleeAttempt(Entity<RepulseAttractComponent> ent, ref AttemptMeleeEvent args)
+    {
+        if (args.Cancelled)
+            return;
+
+        if (_delay.IsDelayed(ent.Owner))
+            return;
+
+        TryRepulseAttract(ent, args.User);
+    }
```
metalgearsloth, line 36: Attempt events should never be making state changes, they should just be returning if something is possible. Just calling "CanAttack" one would not expect to suddenly trigger this.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+        return TryRepulseAttract(ent.Comp.Attract, position, ent.Comp.Speed, ent.Comp.Range, ent.Comp.Whitelist, ent.Comp.Blacklist);
+    }
+
+    public bool TryRepulseAttract(bool attract, MapCoordinates position, float speed, float range, EntityWhitelist? whitelist = null, EntityWhitelist? blacklist = null)
+    {
+        var entsInRange = _lookup.GetEntitiesInRange(position, range, flags: LookupFlags.Dynamic | LookupFlags.Sundries);
+        var epicenter = position.Position;
+        var bodyQuery = GetEntityQuery<PhysicsComponent>();
+        var xformQuery = GetEntityQuery<TransformComponent>();
+
```
metalgearsloth, line 50: Obsolete to use these as local variables please remove.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+    public bool TryRepulseAttract(Entity<RepulseAttractComponent> ent, EntityUid user)
+    {
+        var position = _xForm.GetMapCoordinates(ent.Owner);
+        return TryRepulseAttract(ent.Comp.Attract, position, ent.Comp.Speed, ent.Comp.Range, ent.Comp.Whitelist, ent.Comp.Blacklist);
+    }
+
+    public bool TryRepulseAttract(bool attract, MapCoordinates position, float speed, float range, EntityWhitelist? whitelist = null, EntityWhitelist? blacklist = null)
+    {
+        var entsInRange = _lookup.GetEntitiesInRange(position, range, flags: LookupFlags.Dynamic | LookupFlags.Sundries);
```
metalgearsloth, line 46: Same here, you should be re-using a set instead of getting one every time. Also adjust the flags to make the static check not required.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+        var bodyQuery = GetEntityQuery<PhysicsComponent>();
+        var xformQuery = GetEntityQuery<TransformComponent>();
+
+        foreach (var target in entsInRange)
+        {
+            if (!bodyQuery.TryGetComponent(target, out var physics)
+                || physics.BodyType == BodyType.Static
+                || physics.CollisionLayer == (int)CollisionGroup.GhostImpassable) // don't affect ghosts
+                continue;
+
```
metalgearsloth, line 57: Remove the static check it is unnecessary + the collisionlayer should be on the component and not hardcoded to ghosts.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+            if (_whitelist.IsWhitelistFail(whitelist, target) || _whitelist.IsBlacklistPass(blacklist, target))
+                continue;
+
```
metalgearsloth, line 60: Blacklists in general are a code smell, especially including both, and with the above changes you can easily remove the blacklist and avoid this massive performance cost.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+            var targetXForm = xformQuery.GetComponent(target);
+
+            var targetPos = _xForm.GetMapCoordinates(target, targetXForm).Position;
+
+            // vector from epicenter to target entity
+            var direction = targetPos - epicenter;
+
+            if (direction == Vector2.Zero)
+                continue;
+
```
metalgearsloth, line 70: EntityQuery is not even being used so you can just remove the transform one.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+            // attract: throw all items directly to to the epicenter
+            // repulse: throw them up to the maximum range
+            var throwDirection = attract ? -direction : direction.Normalized() * (range - direction.Length());
+
+            _throw.TryThrow(target, throwDirection, speed);
+        }
```
metalgearsloth, line 76: As above you can easily just derive the property from existing fields.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
metalgearsloth: Having the melee swing decoupled from the melee effect is bizarre I would think? Especially when it's using generic usedelays (which are standardised to around 1 - 1.5s) and not its own usedelay.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
metalgearsloth: Yeah, the stun part is prolly fine.

metalgearsloth (changes requested):

Content.Shared/RepulseAttract/RepulseAttractComponent.cs
slarticodefast: The vector math is different in both cases, so it is not simply a sign that is carried through. But I guess it taking the absolute value of the velocity for the calculations and using its sign instead of the bool would work. I preferred the other way because it is closer to the parameters TryThrow uses.

Content.Shared/RepulseAttract/RepulseAttractComponent.cs
slarticodefast: Personally I always preferred to do it this way because it makes it easier to add the component using VV for debugging purposes or admin shenanigans. But if you prefer making it required I can change it.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
keronshb: I still think it's different, and we don't really have a good way to display effect use delays (like this) vs what's on an item already.

For example if UseDelay is on a bat, what is it used for? Wielding uses the Bat's UseDelay to show that there's a use delay in effect. And in that effect use delay still doesn't prevent you from attacking, unless the "requires wield" was on the bat as well.

And if it wasn't on the bat and instead was on the comp, how do you effectively:
1) Show that this is the usedelay coming from a different comp
2) Tell when a usedelay should be applying towards something.

That being said I think that's completely out of scope for this.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
keronshb: Can you point me to a melee event that works on attempt? 
Meaning, you swing and miss.

This is supposed to work if you miss or not. I could just add a conditional to change the behavior.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
keronshb: Can you explain why it's a code smell?
What if you wanted to have a repulse/attract where you included everything except a few comps?

Content.Shared/RepulseAttract/RepulseAttractComponent.cs
keronshb: I preferred leaving it as placeholder. The defaults might not be used now but may be used later.
Could always just raise the defaults for these if there's an issue.

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
keronshb: To get this to work on melee attempt then we'd have to call getting ents in range on the melee check instead. 
We could make it a parameter instead but then we'd need to think of where we'd move the ents in range call.

Content.Shared/Throwing/ThrowingSystem.cs
```diff
@@ -131,12 +136,13 @@ public void TryThrow(EntityUid uid,
...
         bool playSound = true,
-        bool doSpin = true)
+        bool doSpin = true,
+        bool unanchor = false)
     {
         if (baseThrowSpeed <= 0 || direction == Vector2Helpers.Infinity || direction == Vector2Helpers.NaN || direction == Vector2.Zero || friction < 0)
             return;
 
-        if ((physics.BodyType & (BodyType.Dynamic | BodyType.KinematicController)) == 0x0)
+        if ((physics.BodyType & (BodyType.Dynamic | BodyType.KinematicController)) == 0x0 && !unanchor)
```
slarticodefast, line 145: This will still cause a warning everytime mjollnir is used on an anchored object, because it is static.

slarticodefast (changes requested):
The sinularity hammer could use slightly lower speed due to the physics speed changes. Currently it overshoots when attracting items, causing them to fly though the wizard.

Content.Shared/Throwing/ThrowingSystem.cs
keronshb: Removed the warning since we have unanchor support in throws.

Content.Shared/Throwing/ThrowingSystem.cs
```diff
@@ -131,16 +136,14 @@ public void TryThrow(EntityUid uid,
...
+        bool doSpin = true,
+        bool unanchor = false)
     {
         if (baseThrowSpeed <= 0 || direction == Vector2Helpers.Infinity || direction == Vector2Helpers.NaN || direction == Vector2.Zero || friction < 0)
             return;
 
-        if ((physics.BodyType & (BodyType.Dynamic | BodyType.KinematicController)) == 0x0)
-        {
-            Log.Warning($"Tried to throw entity {ToPrettyString(uid)} but can't throw {physics.BodyType} bodies!");
+        if ((physics.BodyType & (BodyType.Dynamic | BodyType.KinematicController)) == 0x0 && !unanchor)
```
slarticodefast, line 145:
```suggestion
        if (unanchor && HasComp<AnchorableComponent>(uid))
            _transform.Unanchor(uid);

        if ((physics.BodyType & (BodyType.Dynamic | BodyType.KinematicController)) == 0x0)
```

Now that I think about it again, I would move the unanchoring before this body type check. This way it also accounts for cases where unanchoring fails for some reason. And it will be more future proof in case body types are changed in the future.

Remove these two lines below.

Content.Shared/Throwing/ThrowingSystem.cs
```diff
@@ -152,6 +155,9 @@ public void TryThrow(EntityUid uid,
             Animate = animated,
         };
 
+        if (unanchor && HasComp<AnchorableComponent>(uid))
+            _transform.Unanchor(uid);
+
```
slarticodefast, line 160:
```suggestion
```

slarticodefast (changes requested):

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
```diff
@@ -15,113 +16,64 @@ namespace Content.Shared.Weapons.Melee;
...
-    private void OnThrownShutdown(Entity<MeleeThrownComponent> ent, ref ComponentShutdown args)
-    {
-        if (TryComp<PhysicsComponent>(ent, out var body))
-            _physics.SetBodyStatus(ent, body, ent.Comp.PreviousStatus);
-        var ev = new MeleeThrowOnHitEndEvent();
-        RaiseLocalEvent(ent, ref ev);
+        var userPos = _transform.GetMapCoordinates(args.User).Position;
+        foreach (var target in args.HitEntities)
+        {
+            var targetPos = _transform.GetMapCoordinates(target).Position;
```
metalgearsloth, line 43: As above using the wrong method, should be using getworldposition. This is also expensive af so you should really be checking if there are any hitentities first before even calling it.

Content.Shared/Weapons/Melee/UseDelayOnMeleeHitSystem.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Content.Shared.Throwing;
+using Content.Shared.Timing;
+using Content.Shared.Weapons.Melee.Components;
+using Content.Shared.Weapons.Melee.Events;
+
+namespace Content.Shared.Weapons.Melee;
+
+public sealed class UseDelayOnMeleeHitSystem : EntitySystem
+{
+    [Dependency] private readonly UseDelaySystem _delay = default!;
```
metalgearsloth, line 10: New code needs documentation please.

metalgearsloth (dismissed):

Content.Shared/Weapons/Melee/UseDelayOnMeleeHitSystem.cs
keronshb: Does something this obvious need documentation?

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (direction == Vector2.Zero)
+                continue;
+
+            // attract: throw all items directly to to the epicenter
+            // repulse: throw them up to the maximum range
+            var throwDirection = speed < 0 ? -direction : direction.Normalized() * (range - direction.Length());
+
+            _throw.TryThrow(target, throwDirection, Math.Abs(speed));
+        }
```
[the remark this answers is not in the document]
keronshb: 1 & 2 Why. Please give a detailed explanation.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
keronshb: Why should it be using WorldPos instead of map?

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        TryRepulseAttract(ent, args.User);
+    }
+
+    public bool TryRepulseAttract(Entity<RepulseAttractComponent> ent, EntityUid user)
+    {
+        var position = _xForm.GetMapCoordinates(ent.Owner);
+        return TryRepulseAttract(position, ent.Comp.Speed, ent.Comp.Range, ent.Comp.Whitelist, ent.Comp.CollisionLayer);
+    }
+
```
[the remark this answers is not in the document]
keronshb: User's going to be used in a future PR (Repulse Spell)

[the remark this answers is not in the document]
slarticodefast: fixed, it was intended to be passed to TryThrow for the admin log

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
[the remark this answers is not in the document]
slarticodefast: done

Content.Shared/RepulseAttract/RepulseAttractSystem.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        _entSet.Clear();
+        var epicenter = position.Position;
+        _lookup.GetEntitiesInRange(position.MapId, epicenter, range, _entSet, flags: LookupFlags.Dynamic | LookupFlags.Sundries);
+
+        foreach (var target in _entSet)
+        {
+            if (!_physicsQuery.TryGetComponent(target, out var physics)
+                || (physics.CollisionMask & (int)layer) == 0x0) // don't affect ghosts
+                continue;
```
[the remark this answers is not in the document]
slarticodefast: Like this? The documentation for this is awful and every system does it differently to exclude ghosts.

Content.Shared/Weapons/Melee/MeleeThrowOnHitSystem.cs
slarticodefast: done

Content.Shared/Weapons/Melee/UseDelayOnMeleeHitSystem.cs
slarticodefast: done

PR #33318, opened 2024-11-14
Adds paper label visuals to closets and lockers

SpaceRox1244:
## About the PR

Adds visuals for attaching a paper, like a cargo bounty or invoice, to a closet.

## Why / Balance

Closets can accept paper labels the same way that crates can, but they dont currently have a visual for it. I want to be able to label my department locker so nobody touches it!!

## Media

labels

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: SpaceRox1244
- add: Closets and lockers now have visuals for being labeled with papers.

PR #34450, opened 2025-01-15
allow paper labels on seeds

Ian321:
## About the PR

This PR add the ability to add a paper label to seeds.

## Why / Balance

- Seeds can have a wide variety of traits a botanist might want to document.
- Preparation for future PR ("print" button on the plant analyzer)

## Technical details

- Added the PaperLabelComponent to SeedBase
- Gave the base examine text for seeds the priority 1 so that it appears above the paper label.

## Media

image

## Requirements

## Breaking changes

None

**Changelog**

:cl:
- add: Seeds can now have paper labels attached to them.


beck-thompson:
Great idea! Test fail is real though, I think you need to add an ItemSlots component on the seeds themselves so it doesn't automatically add it when its spawned.

Resources/Prototypes/Entities/Objects/Specific/Hydroponics/seeds.yml
```diff
@@ -11,6 +11,20 @@
       size: Tiny
     - type: StaticPrice
       price: 20
+    - type: ItemSlots
+    - type: ContainerContainer
+      containers:
+        paper_label: !type:ContainerSlot
+    - type: PaperLabel
+      labelSlot:
+        insertVerbText: Attach Label
```
beck-thompson, line 20: Localize these in an FTL file! The class isn't clear (It should be a locid) but they are localized at the end.

Boaz1111 (approved):
Works fine in game, love the idea!

ScarKy0 (approved):
Looks all good!
But could you change the changelog to something along the lines of "Seeds can now have paper labels attached to them"? I feel like the current one might be confused with the labeller

beck-thompson (changes requested):
Yeah tested works great! The only actual change I think is important is to make the label appear on the item itself (Like lockers see this pr #33318). It shouldn't be too hard but if you need help ask here or the discord 🫡

Resources/Prototypes/Entities/Objects/Specific/Hydroponics/seeds.yml
Ian321: Should I reuse the
```
# Verb text
hand-labeler-remove-label-text = Remove label
hand-labeler-add-label-text = Apply label
```
from `hand-labeler.ftl` or create new ones?

beck-thompson: Probably make new ones? I guess you could reuse it but it seems a little weird because they are like different things ya know?

Ian321:
image

Boaz1111:
You'll need a new sprite for the labels

Ian321:
image

beck-thompson (changes requested):
OK tested and everything looks good! The only issue is that when you plant seeds with a tag, the tag is just deleted. Is that intended (I think its OK if it is, just wanted to make sure)?

Ian321:
It should probably drop. (Here is a little feature I imagine it being useful for: "Alien" seed, when you find em they have a label attached with growing conditions (high/low pressure/temperate, specific gasses it needs) so it does not instantly die).

beck-thompson (approved):
Not using an event is a little sussy but I think its fine :thinking:
I don't think there would be many other checks for a system like this

(Just waiting on art approval now)

beck-thompson:
> I did not find a way to use TryEjectToHands via a event. Pray tell if you know how.
>All the ways I found required the seeds to be destructible or breakable (which they are not).

Yeah you'd probably have to make a new event and stuff which seems like a lot of work for something that wont be used anywhere else!

beck-thompson:
image

Hey! I was told by an art lead to try to make it a little more like the image above. I think its probably fine to make it a little bigger than "realistic" to make it have better visibility. Let me know if you have any questions or need help!

Ian321:
image

Ian321:
image

PR #34452, opened 2025-01-15
added chemMaster search bar

PROG-MohamedDwidar:
makes it easy for chemists to do their work by having a search bar for the buffer

why:
a job is dependent on it

what i did:
edited the chemMasterWindow.xaml and chemMasterWindow.xaml.cs for a simple search bar


PJB3005:
This falls inline with my other thoughts about the chemmaster: it would be better if the chemmaster wasn't used for this purpose. Having a search bar is not a good solution IMO.

Princess-Cheeseballs:
Closing as derelict, merge conflicted for nearly a year now. Feel free to reopen if you come back to this.

Issue #34456, opened 2025-01-15
RadiationSource behaves stangely at low intensity


Alpaccalypse:
## Description
If you have the `RadiationSource` component on an entity set to `intensity: 0.1` they will never register on the geiger counter or cause any radiation poisoning. Changing `slope` to any value still results in no radiation. If you set intensity at 0.105, 0.11 etc, then it will behave as expected.

If you make a stack of 2 of this 0.1 intensity entity and place it in the environment or carry it, it will emit radiation, at approximately the correct values. Changing the slope value seems to cause minor variation from expected values when comparing the output of different stacks.

I have exhausted myself testing for now, but I think there is also a problem with something in the calculation mutliplying when it should be adding, or maybe some kind of compound error with the sum of the base values.

**Reproduction**

You can start testing it with something like this:
```
- type: entity
  parent: OreBase
  id: RadBase1
  suffix: rad ore test 11
  components:
  - type: Sprite
    state: uranium
  - type: Material
  - type: RadiationSource
    intensity: 0.11
    slope: 0.1
```

**Screenshots**
Intensity 0.11, slope 09, stack of two.
![image](https://github.com/user-attachments/assets/8b5793b0-105d-495e-a951-60f26958bdf2)
Intensity 0.1, slope 09, stack of two.
![image](https://github.com/user-attachments/assets/66ce7c64-dd58-42b3-beb9-9bea78bd370e)

This might be expected, but it always seems to be off, even accounting for .01 and multiplication of values.

**Additional context**
I am so tired of testing this, and I might have either missed something obvious or made an obvious mistake somewhere, but I am certain that at least something is wrong here with it not registering 0.1 intensity correctly.

Aeshus:
I suspect that this could be caused by `RadiationMinIntensity` (which defaults to 0.1).

https://github.com/space-wizards/space-station-14/blob/e094b79ee4636cbd91e760091420017e81222ebe/Content.Shared/CCVar/CCVars.Radiation.cs#L7-L12

Aeshus:
I think the CCVAR's comments are misleading at that's the not the minimum that can be received, but rather the maximum that will not be accepted.

https://github.com/space-wizards/space-station-14/blob/e094b79ee4636cbd91e760091420017e81222ebe/Content.Server/Radiation/Systems/RadiationSystem.GridCast.cs#L136-L137

Winkarst-cpu:
@Aeshus, was the issue fixed in the #34459?

PR #34459, opened 2025-01-16
Optimize & clean up RadiationSystem

ElectroJr:
## About the PR
Over time some PRs have carelessly added some expensive checks to code in radiation system that could be a hot path. This PR removes that, while also just cleaning up parts of radiation system.

## Why / Balance
Performance is good

## Technical details
- `GetAdjustedRadiationIntensity` now avoids allocating a list and returns early. Also no more linq.
- `GetAdjustedRadiationIntensity` is now only called once per source & receiver, instead of once per `(source+1)*receiver`
- Component fetching and Intensity scaling by stack count is now computed once per source instead of per `source*receiver`
- `OnIrradiatedEvent` is now a struct event
- `RadiationRay` is now a struct, and various debug related collections are no longer allocated unless debugging is enabled.
- `RadiationRay` now uses EntityUid instead of NetEntity. Debugging uses a separate `DebugRadiationRay` struct that uses NetEntity.

## Media

AFAICT It still seems to work, through I didn't do any thorough testing.

https://github.com/user-attachments/assets/fe6b3bc0-91fa-4c10-9947-e7881d63fafc

## Requirements

## Breaking changes
I guess the radiation ray changes are technically breaking, but Its pretty small and I doubt downstreams use the class directly, so probably fine W/O #codebase-changes.

### DIFF
--- a/Content.Server/Radiation/Systems/RadiationSystem.GridCast.cs
+++ b/Content.Server/Radiation/Systems/RadiationSystem.GridCast.cs
@@ -145,7 +145,7 @@ public partial class RadiationSystem
 
         // will it even reach destination considering distance penalty
         var rads = source.Intensity - source.Slope * dist;
-        if (rads <= MinIntensity)
+        if (rads < MinIntensity)
             return null;
 
         // create a new radiation ray from source to destination
--- a/Content.Server/Radiation/Systems/RadiationSystem.GridCast.cs
+++ b/Content.Server/Radiation/Systems/RadiationSystem.GridCast.cs
@@ -1,12 +1,9 @@
-using System.Linq;
 using System.Numerics;
 using Content.Server.Radiation.Components;
 using Content.Server.Radiation.Events;
 using Content.Shared.Radiation.Components;
 using Content.Shared.Radiation.Systems;
-using Content.Shared.Stacks;
 using Robust.Shared.Collections;
-using Robust.Shared.Containers;
 using Robust.Shared.Map.Components;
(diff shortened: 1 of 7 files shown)


Content.Server/Radiation/Systems/RadiationSystem.GridCast.cs
```diff
@@ -108,82 +123,87 @@ private void UpdateGridcast()
...
             return null;
 
         // will it even reach destination considering distance penalty
-        var rads = incomingRads - slope * dist;
-
-        // Apply rad modifier if the source is enclosed within a radiation blocking container
-        rads = GetAdjustedRadiationIntensity(sourceUid, rads);
-
+        var rads = source.Intensity - source.Slope * dist;
         if (rads <= MinIntensity)
```
Aeshus, line 148: **thought (non-blocking):** Could we get this changed to being `<` instead of `<=`? The CCVAR comment kinda implies that `0.1f` is the minimum accepts, while it is actually the maximum that it will reject.

#34456 is relevant.

https://github.com/space-wizards/space-station-14/blob/e094b79ee4636cbd91e760091420017e81222ebe/Content.Shared/CCVar/CCVars.Radiation.cs#L7-L12 
 
```suggestion
        if (rads < MinIntensity)
```

ElectroJr: Sure, though it won't make much of a difference unless the distance is 0

juliangiebel (approved):

PR #34471, opened 2025-01-16
Adds Parcel Wrap

### LINKED ISSUE #4833: Need package wrap
## Description

- Should be able to wrap an entity below a given max size, change package sprite based on size, and then return the entity when the package is unwrapped (visualizer)

**Screenshots**
image

this stuff

Centronias:
## About the PR
Adds Parcel Wrap, which can wrap items into nice little parcel boxes :)
(Many thanks to @lvvova1 , who put together a previous PR which I referenced quite a lot: https://github.com/space-wizards/space-station-14/pull/7974 )

- Parcel wrap can be used directly on other items (`AfterInteractEvent`) or used via a `UtilityVerb`, which enables wrapping things which have storage, which'd eat the wrap rather than use it.
- Wrapped parcels can be unwrapped in hand (`UseInHandEvent`) or via an `InteractionVerb`, which enables unwrapping things which are too big to pick up (currently not strictly necessary).
- Wrapped parcels drop their contents on being destroyed
- Wrapped parcels are recyclable, so that disposals doesn't fill up with them
- Wrap and parcels are relatively extensible via yaml only changes, enabling other visuals.
- ~~I did pull over all of TG's parcel sprites, but only use about a third of them since there's a lot of handling for giftwrap visuals that I didn't deal with.~~

## Why / Balance
More ways to hide bombs is more fun. Also I want mail to be more of a thing, and this moves in that direction.
Resolves #4833 .

## Technical details
- `ParcelWrapComponent`
  - Makes an entity capable of wrapping things in parcel wrap
  - Contains information about what kind of parcels are created
    - size
    - shape
   - Specifies the sound made when wrapping things
   - Has a blacklist so that you can't, eg., wrap the nuke disk
- `WrappedParcelComponent`
  - Marks an entity as being a wrapped parcel
  - Has a container for the wrapped item
  - Specifies the entity to create when the item's unwrapped, eg. some trash
  - Specifies a sound made when unwrapping it
- `ParcelWrappingSystem`
  - Handles the logic for all of the above
  - Makes `WrappedParcel`s in such a way that they copy the size of the wrapped item, which guards against size-arbitrage.
- `WrappedParcel` entity prototype
  - Is a wrapped parcel
  - Has appearance stuff to change the sprite used based on the size of the parcel

## Media

https://github.com/user-attachments/assets/35df1d43-580e-4fe3-b7e7-dcfcc84d9ce8
(Shows off the `doafter`s and examine text)

https://github.com/user-attachments/assets/f8dc8d13-f295-468c-9c17-7973a1f37e26
(Known bugs in this media: popup refers to the parcel rather than the contents; wrapping in a container can be jank when things have unusual shapes)

https://github.com/user-attachments/assets/fef7de45-ab09-47f6-a571-a8af8d3cd782
(Demos fixes to the bugs in the previous)

image
(Cargo order + crate)

image
(@ps3moira 's new cabinet perspective sprites)

## Requirements

## Breaking changes
All new stuff

**Changelog**

:cl:
- add: Parcel wrap, purchasable from cargo, which allows you to wrap items in paper so you can give people surprise gifts... like bombs.


Centronias:
The RSI test failure is complaining about the licenses I've used.

TG station's repo's license is AGPL-3.0 . Does that not cover the sprites as well? What should I use instead?

Content.Server/ParcelWrap/EntitySystems/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,236 @@
...
+    /// <summary>
+    /// Spawns a WrappedParcel containing <paramref name="target"/>.
+    /// </summary>
+    /// <param name="user">The entity using <paramref name="wrapper"/> to wrap <paramref name="target"/>.</param>
+    /// <param name="wrapper">The wrapping being used. Determines appearance of the spawned parcel.</param>
+    /// <param name="target">The entity being wrapped.</param>
+    /// <returns>The newly created parcel. Returns null only in exceptional failure cases.</returns>
+    private Entity<WrappedParcelComponent>? WrapInternal(EntityUid user, ParcelWrapComponent wrapper, EntityUid target)
+    {
+        var spawned = Spawn("WrappedParcel", Transform(target).Coordinates);
```
beck-thompson, line 149: Don't hard code this! If you add it to a field in the `ParcelWrapComponent` then you could have different parcels wrap with different sprites.

Resources/Prototypes/Entities/Objects/Misc/ParcelWrap.yml
```diff
@@ -0,0 +1,69 @@
+- type: entity
```
beck-thompson, line 1: Rename this file to match how the rest are done (e.g like `parcel_wrap.yml`)

Resources/Prototypes/Entities/Objects/Misc/ParcelWrap.yml
```diff
@@ -0,0 +1,69 @@
...
+            "Large": { state: "parcel-medium" }
+            "Huge": { state: "parcel-large" }
+            "Ginormous": { state: "parcel-large" }
+    - type: Sprite
+      sprite: Objects/Misc/ParcelWrap/wrapped_parcel.rsi
+      layers:
+        - state: parcel-medium
+          map: [ "enum.WrappedParcelVisuals.Layer" ]
+    - type: WrappedParcel
+      unwrapSound: /Audio/Effects/poster_broken.ogg
```
beck-thompson, line 47: This seems odd. The noise sounds great but I'm not sure that its a good idea to reuse like this (Same with the handcuffs)! I'll ask to get confirmation.

Content.Server/ParcelWrap/EntitySystems/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,236 @@
...
+    /// Despawns <paramref name="parcel"/>, leaving the contained entity where the parcel was.
+    /// </summary>
+    /// <param name="parcel">The entity being unwrapped.</param>
+    /// <returns>
+    /// The newly unwrapped, contained entity. Returns null only in the exceptional case that the parcel contained
+    /// nothing, which should be prevented by not creating such parcels.
+    /// </returns>
+    private EntityUid? UnwrapInternal(Entity<WrappedParcelComponent> parcel)
+    {
+        var parcelCoords = Comp<TransformComponent>(parcel).Coordinates;
```
beck-thompson, line 205: Use `Transform` here like you did below!

Content.Server/ParcelWrap/EntitySystems/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,236 @@
...
+            Act = () => UnwrapInternal(entity),
+        });
+    }
+
+    private void OnDestroyed<T>(Entity<WrappedParcelComponent> parcel, ref T args)
+    {
+        // Unwrap the package and if something was in it, show a popup describing "wow something came out!"
+        if (UnwrapInternal(parcel) is { } contents)
+        {
+            var parcelId = Identity.Name(contents, EntityManager);
```
beck-thompson, line 116: You never used `parcelId` - I think its OK to just pass in the contents like you do here and you could just remove this line.

Content.Server/ParcelWrap/EntitySystems/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,236 @@
...
+        args.Verbs.Add(new UtilityVerb
+        {
+            Text = Loc.GetString("parcel-wrap-verb-wrap"),
+            Act = () => WrapInternal(user, entity, target),
+        });
+    }
+
+    private void OnComponentInit(Entity<WrappedParcelComponent> entity, ref ComponentInit args)
+    {
+        entity.Comp.Contents = _container.EnsureContainer<ContainerSlot>(entity, WrappedParcelContainerId);
```
beck-thompson, line 84: Store the id in the component! I think that's the right way to do it now :thinking:

Content.Server/ParcelWrap/EntitySystems/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,236 @@
...
+                PopupType.MediumCaution);
+        }
+    }
+
+    #region accessors
+
+    /// <summary>
+    /// Returns whether or not <paramref name="wrapper"/> can be used to wrap <paramref name="target"/>.
+    /// </summary>
+    public bool IsWrappable(Entity<ParcelWrapComponent> wrapper, EntityUid target) =>
```
beck-thompson, line 129: Could you just make this a normal function? It makes sense what your doing but I think for consistency it would be better not to be the arrow function!

beck-thompson (commented):
Overall, code looks great! Lot of minor nitpick stuff. The only actual issues:
1.) Move as much of this as you can into shared (Especially the verbs)
2.) I think a doafter would make a lot of sense. They are annoying to make but it would feel a lot better!
3.) It feels weird that one wrap has an infinite amount of uses. I think making it like an item stack or something would be nice (Or at least some kind of maximum usages).

If you need help with any of this ask on discord!

Also, it would also be fun if PAIs had their name changed if they were in a parcel! Just add the component to the list in pai.yml:77 - Fully up to you though (I'm also biased :laughing:)

Content.Server/ParcelWrap/EntitySystems/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,236 @@
...
+    /// <returns>The newly created parcel. Returns null only in exceptional failure cases.</returns>
+    private Entity<WrappedParcelComponent>? WrapInternal(EntityUid user, ParcelWrapComponent wrapper, EntityUid target)
+    {
+        var spawned = Spawn("WrappedParcel", Transform(target).Coordinates);
+
+        // If this wrap maintains the size when wrapping, set the parcel's size to the target's size. Otherwise use the
+        // wrap's fallback size.
+        ItemComponent? targetItemComp = null;
+        var size = wrapper.FallbackItemSize;
+        if (wrapper.WrappedItemsMaintainSize && Resolve(target, ref targetItemComp, logMissing: false))
```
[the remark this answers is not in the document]
Centronias: Huh, this is confusing -- both `TryComp` and `Resolve` ultimately delegate to `EntityManager.TryGetComponent(uid, out comp)`, so it's not clear to me when to use which...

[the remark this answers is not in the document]
Centronias: Though actually, there is a difference in that `Resolve` will short-circuit if the `out` arg is not null, which is specifically what I wanted in this case (so that we don't waste time looking up a component that's already been looked up.)

Centronias:
@beck-thompson , I kinda sidestepped a few of your organizational comments by shattering the system into three files.
I wanted to keep the `ParcelWrap` versus `WrappedParcel` organizational distinction (that's why `ComponentInit` and the `GetVerb` handlers were organized kinda weirdly).
Lemme know if this is undesireable.

Content.Server/ParcelWrap/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,62 @@
...
+        base.Initialize();
+
+        SubscribeLocalEvent<ParcelWrapComponent, ParcelWrapItemDoAfterEvent>(OnWrapItemDoAfter);
+
+        SubscribeLocalEvent<WrappedParcelComponent, UnwrapWrappedParcelDoAfterEvent>(OnUnwrapParcelDoAfter);
+        SubscribeLocalEvent<WrappedParcelComponent, DestructionEventArgs>(OnDestroyed);
+        SubscribeLocalEvent<WrappedParcelComponent, GotReclaimedEvent>(OnDestroyed);
+    }
+
+    private void OnWrapItemDoAfter(Entity<ParcelWrapComponent> _, ref ParcelWrapItemDoAfterEvent args)
```
beck-thompson, line 26: I think you can move all of these functions to shared! Just make sure `WrapInternal` and `UnwrapInternal` are properly predicted on client and your golden :ok_hand:

Content.Shared/ParcelWrap/Components/ParcelWrapComponent.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// The <see cref="EntityPrototype"/> of the parcel created by using this component.
+    /// </summary>
+    [DataField(required: true)]
+    public EntProtoId ParcelPrototype = default!;
+
+    /// <summary>
+    /// How many times this wrap can be used to create parcels.
+    /// </summary>
+    [DataField(required: true), AutoNetworkedField, Access(typeof(SharedParcelWrappingSystem))]
```
beck-thompson, line 27: You might not need the access after you move stuff around! Also if this has a default don't bother making it required

Content.Shared/ParcelWrap/Components/WrappedParcelComponent.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <summary>
+    /// Sound played when unwrapping this parcel.
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? UnwrapSound;
+
+    /// <summary>
+    /// The ID of <see cref="Contents"/>.
+    /// </summary>
+    public const string ContainerId = "contents";
```
beck-thompson, line 43: Just make this a datafield not any const stuff! I'm a little more shaky if this is the correct approach but I'm pretty sure it is :thinking:

I'll ask to double check I'm right on this one

Content.Shared/ParcelWrap/EntitySystems/ParcelWrappingSystem.ParcelWrap.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    private void OnAfterInteract(Entity<ParcelWrapComponent> entity, ref AfterInteractEvent args)
+    {
+        if (args.Handled ||
+            args.Target is not { } target ||
+            !args.CanReach ||
+            !IsWrappable(entity, target))
+            return;
+
+        TryStartWrapDoAfter(args.User, entity, target);
```
beck-thompson, line 34: Maybe try setting handled to `TryStartWrapDoAfter`? I'm pretty sure that'll do what you want :thinking:

If it works, do it for all of em!

Content.Shared/ParcelWrap/EntitySystems/SharedParcelWrappingSystem.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// <returns>The newly created parcel. Returns null only in exceptional failure cases.</returns>
+    protected Entity<WrappedParcelComponent>? WrapInternal(EntityUid user,
+        Entity<ParcelWrapComponent> wrapper,
+        EntityUid target)
+    {
+        var spawned = Spawn(wrapper.Comp.ParcelPrototype, Transform(target).Coordinates);
+
+        // If this wrap maintains the size when wrapping, set the parcel's size to the target's size. Otherwise use the
+        // wrap's fallback size.
+        ItemComponent? targetItemComp = null;
```
beck-thompson, line 73: OK this is also responding to your old comment on the difference between resolve and trycomp!

1.) You are correct, both basically do the same thing so using it here or not is just a style difference!
2.) You want to use Resolve when you have a function that has optional inputs that could be filled. E.g (You can find many other examples as well):
image
So if whoever is calling this function by chance already has the component, they can just pass it in so the system doesn't have to recompute it! But again, its an optional so they don't need to pass it. If they don't it works like a trycomp.

For your case here this is what I would do:

`TryComp<ItemComponent>(target, out var targetItemComp);` on line 73.

Then on line 75 and 86 just do `targetItemComp != null`.

You'll only need to compute it once this way as well! Let me know if that makes sense :saluting_face:

Content.Shared/ParcelWrap/EntitySystems/SharedParcelWrappingSystem.cs
```diff
@@ -0,0 +1,165 @@
...
+        var size = wrapper.Comp.FallbackItemSize;
+        if (wrapper.Comp.WrappedItemsMaintainSize && TryComp(target, out targetItemComp))
+        {
+            size = targetItemComp.Size;
+        }
+
+        var item = Comp<ItemComponent>(spawned);
```
beck-thompson, line 80: Good use of comp! Maybe add a simple comment explaining why its safe to use it here for new contribs?

beck-thompson (changes requested):
Splitting the system up into mutliple files like that is much cleaner good job :+1:

Some more prediction issues but they shouldn't be too difficult to fix! Again, if you have any questions let me know here or discord :saluting_face:

Content.Shared/ParcelWrap/EntitySystems/ParcelWrappingSystem.WrappedParcel.cs
```diff
@@ -0,0 +1,56 @@
+using Content.Shared.DoAfter;
+using Content.Shared.Interaction.Events;
+using Content.Shared.ParcelWrap.Components;
+using Content.Shared.Verbs;
+using Robust.Shared.Containers;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.ParcelWrap.EntitySystems;
```
MilonPL, line 8: Should be `.Systems` not `.EntitySystems`

Content.Server/ParcelWrap/ParcelWrappingSystem.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        if (args is { Target: { } target, Used: { } used } &&
+            TryComp<ParcelWrapComponent>(used, out var wrapper))
+        {
+            WrapInternal(args.User, (used, wrapper), target);
+            args.Handled = true;
+        }
+    }
+
+    private void OnUnwrapParcelDoAfter(Entity<WrappedParcelComponent> _, ref UnwrapWrappedParcelDoAfterEvent args)
```
MilonPL, line 39: You can't use discard in methods, it doesn't work like this, you're creating a new variable called `_` which might cause other attempts to use a discard fail.

Content.Shared/ParcelWrap/Components/ParcelWrapComponent.cs
```diff
@@ -0,0 +1,74 @@
...
+/// <seealso cref="Components.WrappedParcelComponent"/>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access] // Default readonly, except for VV editing
+public sealed partial class ParcelWrapComponent : Component
+{
+    /// <summary>
+    /// The <see cref="EntityPrototype"/> of the parcel created by using this component.
+    /// </summary>
+    [DataField(required: true)]
+    public EntProtoId ParcelPrototype = default!;
```
MilonPL, line 22: You don't need to initialize EntProtoId.

Content.Shared/ParcelWrap/Components/ParcelWrapComponent.cs
MilonPL: All components should have restricted access by default, but you can move `typeof(SharedParcelWrappingSystem)` to the access attribute on the component, not just this field.

Content.Shared/ParcelWrap/Components/ParcelWrapComponent.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// </summary>
+    [DataField]
+    public ProtoId<ItemSizePrototype> FallbackItemSize = "Ginormous";
+
+    /// <summary>
+    /// If true, parcels created by this will have the same shape as the item they contain. If false, parcels created by
+    /// this will have the default shape for their size.
+    /// </summary>
+    [DataField]
+    public bool WrappedItemsMaintainShape = false;
```
MilonPL, line 49: Bools are false by default, you don't need to set this.

Content.Shared/ParcelWrap/EntitySystems/ParcelWrappingSystem.ParcelWrap.cs
```diff
@@ -0,0 +1,73 @@
...
+            wrapper)
+        {
+            NeedHand = true,
+            BreakOnMove = true,
+            BreakOnDamage = true,
+        });
+}
+
+[Serializable, NetSerializable]
+public sealed partial class ParcelWrapItemDoAfterEvent : SimpleDoAfterEvent;
```
MilonPL, line 73: I think all of these events would be better off in their own file.

Content.Shared/ParcelWrap/EntitySystems/SharedParcelWrappingSystem.cs
```diff
@@ -0,0 +1,165 @@
...
+    [Dependency] private readonly SharedItemSystem _item = default!;
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
+
+    /// <inheritdoc/>
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<ParcelWrapComponent, ExaminedEvent>(OnExamined);
```
MilonPL, line 36: For partial systems, I think it'd be better to only subscribe to events that have handlers in the specific file. So for this you'd want to create a new initialization method in the `ParcelWrap` part that would contain this handler, then call `InitializeParcelWrap()` from here, and so on.

MilonPL:
One more thing I forgot to mention (because I hate looking over yaml), for entity prototypes you should stick to the `type > abstract > parent > id > name > description > components` order.

Content.Shared/ParcelWrap/EntitySystems/ParcelWrappingSystem.ParcelWrap.cs
Centronias: I followed [the `DrinkSystem`'s doafter stuff](https://github.com/space-wizards/space-station-14/blob/b82605b18530c0d4398b9ad41a108868613cf81e/Content.Server/Nutrition/EntitySystems/DrinkSystem.cs#L226), so I am unsure.
I don't think the doafter should ever fail to start, though, so maybe it doesn't make a difference?

Content.Shared/ParcelWrap/EntitySystems/ParcelWrappingSystem.ParcelWrap.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        // "Wrap" verb for when just left-clicking doesn't work.
+        args.Verbs.Add(new UtilityVerb
+        {
+            Text = Loc.GetString("parcel-wrap-verb-wrap"),
+            Act = () => TryStartWrapDoAfter(user, entity, target),
+        });
+    }
+
+    private bool TryStartWrapDoAfter(EntityUid user, Entity<ParcelWrapComponent> wrapper, EntityUid target) =>
```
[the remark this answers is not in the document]
Centronias: Definitely not necessary as it's just syntax. I'll relent as SS14 attracts a lot of new C# coders.
I do just wanna say that it feels meh to have a function that's a single return statement -- but it doesn't change the meaning at all, so eh.

Content.Shared/ParcelWrap/EntitySystems/SharedParcelWrappingSystem.cs
```diff
@@ -0,0 +1,165 @@
...
+            if (_container.TryGetContainingContainer((parcel, null, null), out var outerContainer))
+            {
+                // Make space in the container for the parcel contents.
+                _container.Remove((parcel, null, null), outerContainer, force: true);
+                _container.InsertOrDrop((parcelContents, null, null), outerContainer);
+            }
+        }
+
+        // Make some trash and play an unwrapping sound.
+        var trash = Spawn(parcel.Comp.UnwrapTrash, parcelCoords);
```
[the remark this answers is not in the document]
Centronias: Wow, I looked at the signature of `Spawn`, saw the protoID is nullable, and assumed that meant if `null`'s passed in that it'd just do nothing (like a lot of other functions do).
Buuut apparently not. It just spawns some... empty entity? Huh.
Imagine the doc comment explaining that 😭

Content.Shared/ParcelWrap/EntitySystems/SharedParcelWrappingSystem.cs
Centronias: Oh oh oh, I made the change you suggested, then remembered why I did it the way I did:

```csharp
ItemComponent? targetItemComp = null;
if (wrapper.Comp.WrappedItemsMaintainSize && TryComp(target, out targetItemComp))
{
    // ...
}
if (wrapper.Comp.WrappedItemsMaintainShape && targetItemComp is { Shape: { } shape })
{
    // ...
}
```
If `WrappedItemsMaintainSize` is false, the `if`'ll short circuit the condition and never run `TryComp`, meaning `targetItemComp` will always be null, even if the target does have an `ItemComponent`.

So, either we can do the "double look up"
```csharp
ItemComponent? targetItemComp = null;
if (wrapper.Comp.WrappedItemsMaintainSize &&
    TryComp(target, out targetItemComp)) // Or Resolve
{
    // ...
}

if (wrapper.Comp.WrappedItemsMaintainShape &&
    TryComp(target, out targetItemComp) && // Or `Resolve`
    targetItemComp.Shape is { } shape)
{
    // ...
}
```

Or we can look up once and null check it in each `if`
```csharp
TryComp<ItemComponent>(target, out var targetItemComp);
if (wrapper.Comp.WrappedItemsMaintainSize && targetItemComp is { } targetItem)
{
    // ...
}
if (wrapper.Comp.WrappedItemsMaintainShape && targetItemComp is { Shape: { } shape })
{
    // ...
}
```

Gonna go with the latter, I think, since it's not as bad as I remembered...

Content.Shared/ParcelWrap/Components/ParcelWrapComponent.cs
Centronias: It seems like if the `[Access]` on the class is omitted, the default is unrestricted. (eg. `DrinkSystem` was able to modify this field directly with no RA complaints).

And, following the principle of least privilege, since only this `Uses` field needs to be changed by systems, the `friend` should be set on just this field.

Centronias:
@slarticodefast indicated this implementation has some issues: https://discord.com/channels/310555209753690112/770682801607278632/1331719351283351563

Follow the replies on https://discord.com/channels/310555209753690112/770682801607278632/1333271598676971570 to see the back and forth, but I'm of the opinion that there's not really any components that can just solve this, so check out my proposed changes to generalize the behavior here away from being tied to parcels:
https://github.com/Centronias/space-station-14/pull/1

beck-thompson (approved):
Good job on the PR! The prediction stuff is nasty but everything seems to work now :saluting_face: 

(Someone else with more experiencing might need to weigh in on if there is a way to make it more generalizable)

Resources/Textures/Objects/Misc/ParcelWrap/wrapped_parcel.rsi/meta.json
```diff
@@ -0,0 +1,27 @@
+{
+  "version": 1,
+  "license": " CC BY-SA 4.0",
+  "copyright": "wrapped_parcel © 2025 by Centronias (https://github.com/centronias) is licensed under CC BY 4.0",
```
Centronias, line 4: I made the sprites myself, so attribution's to just me. If somebody can come up with a way to adds some attribution to the `box.rsi` `box` sprite's creator, I'd be happy to do that, but I'm not sure who did that and the [attribution in that file is, uh, unintelligible](https://github.com/space-wizards/space-station-14/blob/master/Resources/Textures/Objects/Storage/boxes.rsi/meta.json#L4) to me.

Winkarst-cpu:
Will close #4833. Add "Closes #4833" into your description.

EmoGarbage404:
Stupendous

PR #34472, opened 2025-01-16
Fix two zombie bugs

IProduceWidgets:
## About the PR

I identified two bugs with zombies and fixed them.
- Vox health regen from #33722 was the same strength as the poisoning effect from the PendingZombieComponent meaning Vox would never naturally succumb to the virus. This is bad because it prevents zombies from snowballing, but also because it means Vox could be II forever with no issues which was not intended. Solution: make the poison effect from zombie virus slightly stronger to compensate.
- All zombies were erroneously initialized with the grace period of Initial Infected. This made the zombie virus really really slow by letting anyone bit by a zombie wait 15 minutes before they start taking damage. Solution: Non-Initial Infected zombies no longer get a grace period.

## Why / Balance

I squash bug.
This will make zombies virus scarier and faster.
It only matters if you survive a fight with a zombie.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Vox are no longer immune to the zombie virus under certain circumstances
- fix: The zombie virus no longer provides non-Initial Infected an undue grace period before poisoning them.

### DIFF
--- a/Content.Server/Zombies/ZombieSystem.cs
+++ b/Content.Server/Zombies/ZombieSystem.cs
@@ -83,6 +83,13 @@ namespace Content.Server.Zombies
         private void OnPendingMapInit(EntityUid uid, IncurableZombieComponent component, MapInitEvent args)
         {
             _actions.AddAction(uid, ref component.Action, component.ZombifySelfActionPrototype);
+
+            if (HasComp<ZombieComponent>(uid) || HasComp<ZombieImmuneComponent>(uid))
+                return;
+
+            EnsureComp<PendingZombieComponent>(uid, out PendingZombieComponent pendingComp);
+
+            pendingComp.GracePeriod = _random.Next(pendingComp.MinInitialInfectedGrace, pendingComp.MaxInitialInfectedGrace);
         }
 
         private void OnPendingMapInit(EntityUid uid, PendingZombieComponent component, MapInitEvent args)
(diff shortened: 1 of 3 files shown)


pheenty:
Buffing it two times is definitely not "make it slightly stronger", I don't even think Vox being kinda immune to it is an issue, yet the second fix is nice

Content.Server/Zombies/ZombieSystem.cs
```diff
@@ -83,6 +83,13 @@ private void OnBeforeRemoveAnomalyOnDeath(Entity<PendingZombieComponent> ent, re
         private void OnPendingMapInit(EntityUid uid, IncurableZombieComponent component, MapInitEvent args)
         {
             _actions.AddAction(uid, ref component.Action, component.ZombifySelfActionPrototype);
+
+            if (HasComp<ZombieComponent>(uid) || HasComp<ZombieImmuneComponent>(uid))
+                return;
+
+            EnsureComp<PendingZombieComponent>(uid, out PendingZombieComponent pendingComp);
```
slarticodefast, line 90: Hmm, adding a component in the initialization of another component is a little messy and afaik there are some integration tests that can get fail due to this for certain prototypes.
And as far as I can see it should already be added where ever the IncurableZombieComponent is added.

IProduceWidgets: Yes, but that causes a race condition between the two. Its what caused the bug.

IProduceWidgets: I can admit that they should probably just be the same component, but tbh all of zombie system and its components and zombiesystem.transform needs a rewrite. Its a mess in there.

PR #34474, opened 2025-01-17
Fix ghost visibility

ElectroJr:
## About the PR
Updates ghost system so that ghosts are invisible by default.

## Technical details
Revenants made ghosts default to being visible instead of invisible, while relying on PVS visibility masks to prevent clients from seeing them, but that obviously doesn't work when PVS is disabled and can introduce bugs (e.g., see space-wizards/RobustToolbox/pull/5598).

Ghost sprite visibility and PVS visibility masks are now configurable via `GhostSystem.SetVisible`, and revenants now use that.

## Requirements

:cl:
- fix: Fixed ghosts sometimes being visible when they shouldn't be.

### DIFF
--- a/Content.Server/Administration/Commands/ShowGhostsCommand.cs
+++ b/Content.Server/Administration/Commands/ShowGhostsCommand.cs
@@ -28,11 +28,7 @@ namespace Content.Server.Administration.Commands
                 return;
             }
 
-            var ghostSys = _entities.EntitySysManager.GetEntitySystem<GhostSystem>();
-            var revSys = _entities.EntitySysManager.GetEntitySystem<RevenantSystem>();
-
-            ghostSys.MakeVisible(visible);
-            revSys.MakeVisible(visible);
+            _entities.System<GhostSystem>().MakeVisible(visible);
         }
     }
 }
--- a/Content.Client/Ghost/GhostSystem.cs
+++ b/Content.Client/Ghost/GhostSystem.cs
@@ -17,7 +17,7 @@ namespace Content.Client.Ghost
 
         public int AvailableGhostRoleCount { get; private set; }
 
-        private bool _ghostVisibility = true;
+        private bool _ghostVisibility;
 
         private bool GhostVisibility
         {
@@ -32,9 +32,9 @@ namespace Content.Client.Ghost
                 _ghostVisibility = value;
 
                 var query = AllEntityQuery<GhostComponent, SpriteComponent>();
-                while (query.MoveNext(out var uid, out _, out var sprite))
+                while (query.MoveNext(out var uid, out var ghost, out var sprite))
                 {
-                    sprite.Visible = value || uid == _playerManager.LocalEntity;
+                    UpdateVisibility((uid, ghost, sprite));
                 }
             }
         }
@@ -70,8 +70,15 @@ namespace Content.Client.Ghost
 
         private void OnStartup(EntityUid uid, GhostComponent component, ComponentStartup args)
         {
-            if (TryComp(uid, out SpriteComponent? sprite))
-                sprite.Visible = GhostVisibility || uid == _playerManager.LocalEntity;
+            UpdateVisibility((uid, component));
+        }
+
+        private void UpdateVisibility(Entity<GhostComponent?, SpriteComponent?> ghost)
+        {
+            if (!Resolve(ghost.Owner, ref ghost.Comp1, ref ghost.Comp2))
+                return;
+
+            ghost.Comp2.Visible = GhostVisibility || ghost.Comp1.Visible || ghost.Owner == _playerManager.LocalEntity;
         }
 
         private void OnToggleLighting(EntityUid uid, EyeComponent component, ToggleLightingActionEvent args)
@@ -130,7 +137,9 @@ namespace Content.Client.Ghost
         private void OnGhostState(EntityUid uid, GhostComponent component, ref AfterAutoHandleStateEvent args)
         {
             if (TryComp<SpriteComponent>(uid, out var sprite))
-                sprite.LayerSetColor(0, component.color);
+                sprite.LayerSetColor(0, component.Color);
+
+            UpdateVisibility((uid, component, null));
 
             if (uid != _playerManager.LocalEntity)
                 return;
--- a/Content.Server/Ghost/GhostSystem.cs
+++ b/Content.Server/Ghost/GhostSystem.cs
@@ -124,6 +124,33 @@ namespace Content.Server.Ghost
             Dirty(uid, component);
         }
 
+        public override void SetVisible(Entity<GhostComponent?> ghost, bool visible)
+        {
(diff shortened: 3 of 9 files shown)


keronshb (changes requested):
Sanity check me on this but I think the Wizard Summon Ghost event needs to be updated to include the new method as well.

keronshb:
> > Sanity check me on this but I think the Wizard Summon Ghost event needs to be updated to include the new method as well.
> 
> The action / `ToggleGhostVisibilityToAllEvent` seems to just use `MakeVisible(true)` (same as the end-of-round stuff), so it seems like its behaviour should be unchanged. While I didn't test wizard specifically, the end of round visibility still worked.
> 
> Though there's a separate issue that also exists on master, where `MakeVisible` only affects currently existing ghosts, and won't affect any newly created ones. This probably doesn't matter much for the end-of-round ghost visibility thing, but probably matters somewhat more for wizard. Also the action description is wrong ("invisible" instead of "visible").

The eventual goal is to get it to work like that, but I'm honestly not sure how or where to start for that.
The text is fixed in #34994 (which is in limbo for some reason?)

slarticodefast:
@metalgearsloth Is this even needed anymore after your vismask changes?

ElectroJr:
> @ metalgearsloth Is this even needed anymore after your vismask changes?

If you mean #35347, that only touched the ghost's eye masks, not their visibility layers.

I've updated the PR to fix conflicts, and also fixed how the show-ghosts wizard action works so that it also applies to newly created ghosts. In the process of fixing conflicts, I've introduced some more breaking changes:
- Renamed `MakeVisible()` to `SetAllObserversVisible()` for clarity.
- Replaced the `AllowGhostShownByEvent` tag introduced in #34994 with a bool datafield on ghost component.
  - Its a tag that exists purely to toggle ghost comp related behaviour, it should've just been a field instead of tag.

ElectroJr:
I've re-written most of the PR to remove the revenant code duplication and its more of a refactor than a bugfix now.

Content.Client/Ghost/Commands/ToggleGhostVisibilityCommand.cs
```diff
@@ -12,15 +12,15 @@ public sealed class ToggleGhostVisibilityCommand : IConsoleCommand
 
     public void Execute(IConsoleShell shell, string argStr, string[] args)
     {
-        var ghostSystem = _entSysMan.GetEntitySystem<GhostSystem>();
+        var ghostSystem = _entSysMan.GetEntitySystem<GhostVisibilitySystem>();
```
MilonPL, line 15: Turn it into a LocalizedEntityCommand and use dependencies to get the system?

Content.Client/Ghost/GhostVisibilitySystem.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private void OnDetached(Entity<GhostVisibilityComponent> ent, ref PlayerDetachedEvent args)
+    {
+        UpdateVisibility(ent!);
+    }
+
+    private void OnAttached(Entity<GhostVisibilityComponent> ent, ref PlayerAttachedEvent args)
+    {
+        UpdateVisibility(ent!);
```
MilonPL, line 54: This is converting from `Entity<T>` into `Entity<T1?, T2?>` so we can use the `AsNullable` method. There's already implicit conversion happening from `T` to `T1, T2`.

Content.Client/Ghost/GhostVisibilitySystem.cs
```diff
@@ -0,0 +1,81 @@
+using Content.Shared.Ghost;
+using Robust.Client.GameObjects;
+using Robust.Shared.Player;
+
+namespace Content.Client.Ghost;
+
+public sealed class GhostVisibilitySystem: SharedGhostVisibilitySystem
```
MilonPL, line 7: `GhostVisibilitySystem : SharedGhostVisibilitySystem`

Content.Server/Administration/Commands/ShowGhostsCommand.cs
```diff
@@ -28,11 +27,7 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
                 return;
             }
 
-            var ghostSys = _entities.EntitySysManager.GetEntitySystem<GhostSystem>();
-            var revSys = _entities.EntitySysManager.GetEntitySystem<RevenantSystem>();
-
-            ghostSys.MakeVisible(visible);
-            revSys.MakeVisible(visible);
+            _entities.System<SharedGhostVisibilitySystem>().SetAllVisible(visible);
```
MilonPL, line 30: Same here I guess, could use dependency injection instead.

Content.Shared/Ghost/SharedGhostVisibilitySystem.cs
```diff
@@ -0,0 +1,137 @@
...
+    protected virtual void SetVisible(Entity<GhostVisibilityComponent?, VisibilityComponent?> ghost, bool visible)
+    {
+        if (!Resolve(ghost.Owner, ref ghost.Comp1, ref ghost.Comp2))
+            return;
+
+        if (ghost.Comp1.Visible == visible && ghost.Comp1.LifeStage >= ComponentLifeStage.Running)
+            return;
+
+        // VisibilityComponent might not exist yet, and will not get added on client
+        Resolve(ghost.Owner, ref ghost.Comp2, false);
```
MilonPL, line 65: What's the point of this? The Resolve above will return if Comp2 is null or can't be found on the entity.

Content.Client/Ghost/GhostVisibilitySystem.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        base.Initialize();
+        SubscribeLocalEvent<GhostVisibilityComponent, AfterAutoHandleStateEvent>(OnGhostVisState);
+        SubscribeLocalEvent<GhostVisibilityComponent, PlayerAttachedEvent>(OnAttached);
+        SubscribeLocalEvent<GhostVisibilityComponent, PlayerDetachedEvent>(OnDetached);
+    }
+
+    private void OnDetached(Entity<GhostVisibilityComponent> ent, ref PlayerDetachedEvent args)
+    {
+        UpdateVisibility(ent!);
```
MilonPL, line 49: Same here.

MilonPL (changes requested):

Content.Client/Ghost/Commands/ToggleGhostVisibilityCommand.cs
ElectroJr: Done in #38338

Content.Shared/Ghost/SharedGhostVisibilitySystem.cs
ElectroJr: I just forgot to remove the second component from the earlier Resolve in the if-return check, which I've fixed now, thanks for spotting that bug.

Issue #25038, opened 2024-02-08
You cannot rotate selected drawings when using crayons


deepdarkdepths:
Mainly a QOL fix. I suggest the rotation functionality of the construction menu be integrated into drawings. 
This is an issue when you're trying to for example draw an arrow but it can only point north instead of where you want it to point due to lack of ability to rotate..

PR #34492, opened 2025-01-17
Crayon rotation and preview mode overlay

themias:
## About the PR

This PR allows crayon drawings to be rotated and adds a Preview Mode overlay that allows you to see the rotation and exactly where it will be drawn.

I'm opening this as a draft as I haven't worked on overlays before and this will likely require plenty of revisions. I'm also likely to modify it based on design feedback.

## Why / Balance

I was looking to add rotation for crayon decals (#25038), but it seemed difficult to use without being able to clearly see the orientation before placing it. I figured this would also be useful in general for crayon artists to more precisely draw, if they wish.

## Technical details

The rotation is pretty simple and is just a float in degrees. It is passed/stored similar to color/selected decal.
The UI element can be incremented 90 degrees with +/-, or exact degrees can be entered (requires hitting Enter key to update)

The preview overlay is largely based on the existing decal placement overlay.
Design decisions:
- The preview mode is tied to a BUI button, as I figured it would be annoying to be on by default when holding a crayon.
- The mode can only be turned on from the active-hand crayon's window
- The mode is disabled when closing the BUI, switching hands, or unequipping the crayon.
- These events were placed on the server. Can/should they be in shared instead?
- BeforeDraw() in the overlay checks if the location is within range (to avoid confusion as to why the click didn't draw anything). Is this too expensive, and is there a better way to do it?

## Media

https://github.com/user-attachments/assets/43cfa61f-4ea2-469b-b078-d271424ff695

## Requirements

**Changelog**
:cl:
- tweak: Crayon drawings can be rotated
- tweak: Crayons have a Preview Mode to see precise location and orientation


themias:
Response seems positive to the general concept, so I'll mark this as ready for review.

I've done some testing with grids, and the preview mode shows the correct (new) orientation if standing in one grid and drawing in another (e.g. standing in airlock and drawing in shuttle.)

EmoGarbage404 (approved):

Content.Client/Crayon/CrayonComponent.cs
```diff
@@ -10,5 +10,8 @@ public sealed partial class CrayonComponent : SharedCrayonComponent
         [ViewVariables(VVAccess.ReadWrite)] public bool UIUpdateNeeded;
         [ViewVariables] public int Charges { get; set; }
         [ViewVariables] public int Capacity { get; set; }
+        [ViewVariables] public string State { get; set; }
+        [ViewVariables] public float Rotation { get; set; }
+        [ViewVariables] public bool PreviewMode { get; set; }
```
sowelipililimute, line 15:
```suggestion
        [DataField] public string State;
        [DataField] public float Rotation;
        [DataField] public bool PreviewMode;
```
no need to write out get/set, also we use DataFields these days

Content.Client/Crayon/CrayonComponent.cs
```diff
@@ -10,5 +10,8 @@ public sealed partial class CrayonComponent : SharedCrayonComponent
         [ViewVariables(VVAccess.ReadWrite)] public bool UIUpdateNeeded;
         [ViewVariables] public int Charges { get; set; }
         [ViewVariables] public int Capacity { get; set; }
+        [ViewVariables] public string State { get; set; }
```
sowelipililimute, line 13: this property appears to be a duplicate of `SharedCrayonComponent.SelectedState`

Content.Client/Crayon/CrayonSystem.cs
```diff
@@ -64,4 +93,55 @@ protected override void FrameUpdate(FrameEventArgs args)
                 ("capacity",_parent.Capacity)));
         }
     }
+
+    private DecalPrototype? GetDecal(string decalId)
+    {
+        return decalId != null ?
+            _protoMan.Index<DecalPrototype>(decalId) : null;
+    }
```
sowelipililimute, line 101:
```suggestion
    private DecalPrototype? GetDecal(ProtoId<DecalPrototype>? decalId)
    {
        return decalId is { } id ? _protoMan.Index(id) : null;
    }
```
type safety is good to have here

Content.Client/Crayon/CrayonSystem.cs
```diff
@@ -64,4 +93,55 @@ protected override void FrameUpdate(FrameEventArgs args)
...
+
+    private void UpdateOverlayInternal(string state, float rotation, Color color, bool previewMode)
```
sowelipililimute, line 103:
```suggestion
    private void UpdateOverlayInternal(ProtoId<DecalPrototype>? state, float rotation, Color color, bool previewMode)
```

Content.Server/Crayon/CrayonComponent.cs
```diff
@@ -24,5 +24,11 @@ public sealed partial class CrayonComponent : SharedCrayonComponent
         [ViewVariables(VVAccess.ReadWrite)]
         [DataField("deleteEmpty")]
         public bool DeleteEmpty = true;
+
+        [ViewVariables(VVAccess.ReadWrite)]
+        public float Rotation { get; set; }
+
+        [ViewVariables(VVAccess.ReadWrite)]
+        public bool PreviewMode { get; set; }
```
sowelipililimute, line 32:
```suggestion
        [DataField]
        public float Rotation;

        [DataField]
        public bool PreviewMode;
```
also given that these are also seen on the client component, i would put these in shared + network them

Content.Server/Crayon/CrayonSystem.cs
```diff
@@ -125,6 +132,55 @@ private void OnCrayonBoundUIColor(EntityUid uid, CrayonComponent component, Cray
...
+            _uiSystem.SetUiState(uid, SharedCrayonComponent.CrayonUiKey.Key, new CrayonBoundUserInterfaceState(component.SelectedState, component.SelectableColor, component.Color, component.Rotation, component.PreviewMode));
+        }
+    }
+
+    private void OnBuiClosed(EntityUid uid, CrayonComponent component, BoundUIClosedEvent args)
+    {
+        component.PreviewMode = false;
+        Dirty(uid, component);
+        _uiSystem.SetUiState(uid, SharedCrayonComponent.CrayonUiKey.Key, new CrayonBoundUserInterfaceState(component.SelectedState, component.SelectableColor, component.Color, component.Rotation, component.PreviewMode));
+        RaiseNetworkEvent(new CrayonOverlayUpdateEvent(component.SelectedState, component.Rotation, component.Color, component.PreviewMode));
```
sowelipililimute, line 165: refactor this out into a helper method & deduplicate the calls

sowelipililimute (changes requested):

metalgearsloth (changes requested):

metalgearsloth:
Will finish review later but noticed more stuff.

slarticodefast:
`CrayonDecalPlacementOverlay` looks like it is just copy pasting code from `DecalPlacementOverlay`, can't you reuse that one?
Also if I remember correctly sloth said `DecalPlacementOverlay` already was a bandaid for `PlacementManager` not working well, but that one needs a complete rewrite.

Content.Server/Crayon/CrayonComponent.cs
```diff
@@ -24,5 +24,11 @@ public sealed partial class CrayonComponent : SharedCrayonComponent
         [ViewVariables(VVAccess.ReadWrite)]
         [DataField("deleteEmpty")]
         public bool DeleteEmpty = true;
+
+        [DataField]
+        public float Rotation;
+
+        [DataField]
+        public bool PreviewMode;
```
slarticodefast, line 32: add documentation

Content.Server/Crayon/CrayonSystem.cs
```diff
@@ -125,6 +132,54 @@ private void OnCrayonBoundUIColor(EntityUid uid, CrayonComponent component, Cray
...
+    {
+        DisablePreviewMode(uid, component);
+    }
+
+    private void DisablePreviewMode(EntityUid uid, CrayonComponent component)
+    {
+        component.PreviewMode = false;
+        Dirty(uid, component);
+        _uiSystem.SetUiState(uid, SharedCrayonComponent.CrayonUiKey.Key, new CrayonBoundUserInterfaceState(component.SelectedState, component.SelectableColor, component.Color, component.Rotation, component.PreviewMode));
+        RaiseNetworkEvent(new CrayonOverlayUpdateEvent(component.SelectedState, component.Rotation, component.Color, component.PreviewMode));
```
slarticodefast, line 180: this should use component states instead of manually networking it
Use `AfterAutoHandleState` if you switch to autonetworking or `ComponentHandleState` otherwise for that

Content.Shared/Crayon/SharedCrayonComponent.cs
```diff
@@ -70,6 +70,33 @@ public CrayonUsedMessage(string drawn)
...
     }
 
+    [Serializable, NetSerializable]
+    public sealed class CrayonRotationMessage : BoundUserInterfaceMessage
+    {
+        public readonly float Rotation;
+        public CrayonRotationMessage(float rotation)
+        {
+            Rotation = rotation;
+        }
```
slarticodefast, line 80: just use the primary constructor
also add documentation
same below

Content.Shared/Crayon/SharedCrayonComponent.cs
```diff
@@ -80,13 +107,17 @@ public sealed class CrayonComponentState : ComponentState
         public readonly string State;
         public readonly int Charges;
         public readonly int Capacity;
+        public readonly float Rotation;
+        public readonly bool PreviewMode;
 
-        public CrayonComponentState(Color color, string state, int charges, int capacity)
+        public CrayonComponentState(Color color, string state, int charges, int capacity, float rotation, bool previewMode)
```
slarticodefast, line 113: I would just use AutoNetworking and combine the server, client and shared components into a single shared one.
No clue why it is done like that, but I assume it's just really old code.

Content.Server/Crayon/CrayonSystem.cs
```diff
@@ -125,6 +132,54 @@ private void OnCrayonBoundUIColor(EntityUid uid, CrayonComponent component, Cray
...
+            _uiSystem.SetUiState(uid, SharedCrayonComponent.CrayonUiKey.Key, new CrayonBoundUserInterfaceState(component.SelectedState, component.SelectableColor, component.Color, component.Rotation, component.PreviewMode));
+        }
+    }
+
+    private void OnBuiClosed(EntityUid uid, CrayonComponent component, BoundUIClosedEvent args)
+    {
+        DisablePreviewMode(uid, component);
+    }
+
+    private void OnHandDeselected(EntityUid uid, CrayonComponent component, ref HandDeselectedEvent args)
```
slarticodefast, line 165:
```suggestion
    private void OnHandDeselected(Entity<CrayonComponent> ent, ref HandDeselectedEvent args)
```
use the `Entity<Component>` pattern
same for all the other methods here
In case you don't know about that, I got a writeup here https://hackmd.io/@Slart/S1hsoGFm1l

slarticodefast (changes requested):

slarticodefast:
Not a full review from me either, just some stuff I noticed.

Content.Client/Crayon/CrayonSystem.cs
```diff
@@ -1,24 +1,50 @@
...
+
+        SubscribeLocalEvent<LocalPlayerDetachedEvent>(OnPlayerDetached);
+        SubscribeLocalEvent<RoundRestartCleanupEvent>(OnRoundRestart);
+        SubscribeLocalEvent<CrayonComponent, ComponentShutdown>(OnShutdown);
+    }
+
+    public override void Shutdown()
+    {
+        base.Shutdown();
     }
```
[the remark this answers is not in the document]
themias: I think I originally had something in there and then it got removed. Thanks for the catch.

themias:
> `CrayonDecalPlacementOverlay` looks like it is just copy pasting code from `DecalPlacementOverlay`, can't you reuse that one? Also if I remember correctly sloth said `DecalPlacementOverlay` already was a bandaid for `PlacementManager` not working well, but that one needs a complete rewrite.

IIRC the issue is that DecalPlacementOverlay has its state set from the decal system, so I couldn't reuse it. I agree that it isn't a great solution, so I can wait until PlacementManager is rewritten if we want. (Maybe just separate out the Rotation part and leave the Preview for later?)

sowelipililimute (approved):
this looks fine by me, do address sloth's components regarding code duplication of the decal preview system though

themias:
Closing in favor of #39448

PR #34493, opened 2025-01-17
HOTFIX: patch a certain mech related exploit

Ian321:
## About the PR

YouTube recently recommended a video to me that detailed the execution of a exploit.

I am keeping certain information hidden as to not spread how it can be executed. (If you are a maintainer, feel free to DM me.)

## Why / Balance

Fixes a exploit.

## Technical details

Kinda dirty fix. There is something broken deep down in the physics system...

## Requirements

## Breaking changes

None

**Changelog**

No *honest* player facing changes.

### DIFF
--- a/Content.Shared/Mech/EntitySystems/SharedMechSystem.cs
+++ b/Content.Shared/Mech/EntitySystems/SharedMechSystem.cs
@@ -2,6 +2,7 @@ using System.Linq;
 using Content.Shared.Access.Components;
 using Content.Shared.ActionBlocker;
 using Content.Shared.Actions;
+using Content.Shared.Climbing.Components;
 using Content.Shared.Destructible;
 using Content.Shared.DoAfter;
 using Content.Shared.DragDrop;
@@ -18,6 +19,7 @@ using Content.Shared.Weapons.Melee;
 using Content.Shared.Whitelist;
 using Robust.Shared.Containers;
 using Robust.Shared.Network;
+using Robust.Shared.Physics.Components;
 using Robust.Shared.Serialization;
 using Robust.Shared.Timing;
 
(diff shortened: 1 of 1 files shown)


ScarKy0:
Maintainer here, are you on the discord server? Give me a ping if so

Content.Shared/Mech/EntitySystems/SharedMechSystem.cs
```diff
@@ -365,6 +368,12 @@ public bool TryInsert(EntityUid uid, EntityUid? toInsert, MechComponent? compone
         if (!CanInsert(uid, toInsert.Value, component))
             return false;
 
+        if (!_entity.TryGetComponent<ClimbingComponent>(toInsert, out var climbing))
+            return false;
+
+        if (climbing.IsClimbing)
+            return false;
+
```
ScarKy0, line 376: Could you combine these two IF statements into one? Also please add a TODO comment explaining what they are for, as this is more of a bandaid fix

ScarKy0 (changes requested):

slarticodefast:
This also will have to be merged into staging instead of stable since we already branched off the new release.
Or we wait for that and hotfix it then.

ScarKy0:
I think we can wait until after release to merge it

ScarKy0 (approved):

metalgearsloth:
I would much rather have a bug report and a fix than a vague hint in code comments for this please.

slarticodefast:
Looks like the same bug happens with disposal units and cryo pods by the way. So only doing this in the mech system would not achieve much.

Ian321:
Here is what I have found so far: Somehow some of the `Contact` in the players body get nulled; and it crashes/bugs because `DestroyContact` assumes that all values inside a contact are valid. One fix would be to handle null `Contact` values inside the destroy function. A better fix would involve finding where they get nulled.

The only places where the nullable values get set are:

- `SharedPhysicsSystem.SetContact` > `SharedPhysicsSystem.Return` (no idea why that method exists)
- `SharedPhysicsSystem.SetContact` > `SharedPhysicsSystem.CreateContact` > `SharedPhysicsSystem.AddPair` (which does not accept null)

Ian321:
Could this be the same issue that's described here? https://github.com/space-wizards/RobustToolbox/issues/3524

slarticodefast:
Probably fixed by https://github.com/space-wizards/RobustToolbox/pull/5619
We will have to wait for the next engine release and can do some testing then.

slarticodefast:
I can confirm that the new engine release fixed the underlying issue, so this PR is no longer needed.
Anyways, thank you for your investigation and for informing us of the bug!

PR #34496, opened 2025-01-18
Fix `emergency_elkridge` being saved as a map

ElectroJr:
## About the PR
Its a shuttle, it should be saved as a grid not a map.

## Requirements

### DIFF
--- a/Resources/Maps/Shuttles/emergency_elkridge.yml
+++ b/Resources/Maps/Shuttles/emergency_elkridge.yml
@@ -18,7 +18,7 @@ entities:
       name: NT Evac Log
     - type: Transform
       pos: -0.42093527,-0.86894274
-      parent: 637
+      parent: invalid
     - type: MapGrid
       chunks:
         0,0:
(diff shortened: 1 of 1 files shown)


slarticodefast:
Does this have to be hotfixed?

IProduceWidgets:
> Does this have to be hotfixed?

Probably a good idea since the shuttle will never arrive. I don't think the emergencyshuttlesystem has any safeguards for ending the round in such a case. Admins sometimes mess up the shuttle and have to end it manually.

ElectroJr:
> Does this have to be hotfixed?

I'm not sure how critical it is because I didn't test it. I **think** the changes to shuttle loading made in #22767 should make it still work even if its saved incorrectly, but I didn't actually test it. But probably better to be save. Like, I think the `loadgrid` command might still wipe the map?

Emisse:
target stable with this one

Emisse (approved):

ArtisticRoomba:
> Probably a good idea since the shuttle will never arrive.

should probably cherry pick to stable

PR #32586, opened 2024-10-01
Borg type switching.

PJB3005:
## About the PR

This allows borgs (new spawn or constructed) to select their chassis type on creation, like in SS13. This removes the need for the many different chassis types, and means round-start borgs can actually play the game immediately instead of waiting for science to unlock everything.

New borgs have an additional action that allows them to select their type. This opens a nice window with basic information about the borgs and a select button. Once a type has been selected it is permanent for that borg chassis.

These borg types also immediately start the borg with specific modules, so they do not need to be printed. Additional modules can still be inserted for upgrades, though this is now less critical. The built-in modules cannot be removed, but are shown in the UI.

The modules that each borg type starts with:

* Generic: tools
* Engineering: tools, construction, RCD, cable
* Salvage: Grappling gun, appraisal, mining
* Janitor: cleaning, light replacer
* Medical: treatment
* Service: music, service, clowning

Specialized borgs have 3 additional module slots available on top of the ones listed above, generic borgs have 5.

The ability to "downgrade" a borg so it can select a new chassis, like in SS13, is something that would be nice, but was not high enough priority for me to block the feature on. I did keep it in mind with some of the code, so it may be possible in the future.

There is no fancy animation when selecting borg types like in SS13, because I didn't think it was high priority, and it would add a lot of complex code.

## Why / Balance
Playing borg round-start is just awful. The only module you have access to is the basic tools, and you basically have to pray science cares enough to upgrade you. Not to mention that having to *always* be disassembled to become a specialized borg is just insane. In SS13 there is none of this friction, you just select what borg type you want to play this round and hit go.

## Technical details
Borg types are specified in a new BorgTypePrototype. These prototypes specify all information about the borg type. It is assigned to the borg entity through a mix of client side, server, and shared code. Some of the involved components were made networked, others are just ensured they're set on both sides of the wire.

The most gnarly change is the inventory template prototype, which needs to change purely to modify the borg hat offset. I managed to bodge this in with an API that *probably* won't explode for specifically for this use case, but it's still not the most clean of API designs.

Parts for specific borg chassis have been removed (so much deleted YAML) and specialized borg modules that are in the base set of a type have been removed from the exosuit fab as there's no point to printing those.

Supersedes #30406

## Media
https://github.com/user-attachments/assets/59776058-a4a2-4aa7-b537-1220bf318dc6

## Requirements
[...]

PR #34501, opened 2025-01-18
Station AI customizations

chromiumboy:
## About the PR

The station AI can now customize the core and holographic projections from a list of predefined options. These customizations can be accessed when the AI is inserted in an AI core, right-clicks on it, and chooses the appropriate verb from the menu

## Why / Balance

This grants AI players a way of personalizing their appearance 

This PR works towards fulfilling #33954

## Technical details

The system works similarly to loadouts. The customization options are prototypes, which are defined in yaml for ease of adding new / modifying old options in the future. The majority of the logic that handles the players changing their appearance and applying to their avatar is contained in SharedStationAiSystem.Customization

## Media

[Demonstration of the system](https://youtu.be/xY-XdVI3pz8)

Screenshots
Capture

Capture2

ai-cutomization ai-cutomization2

## Requirements

- [X I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).

## Breaking changes

N/A

**Changelog**

:cl:
- add: Station AIs can personalize the appearance of their cores and holograms by right-clicking the AI core in which they are inserted and selecting 'AI customization' from the list


TheShuEd (commented):
I don't think this approach is quite right, and it will be difficult to expand. I would suggest implementation through some Generic !type: loadouts entries that are applied to the player entity as EntityEffects or something similar when connecting to the game. In this case, extending the new selection functionality via loadouts will be much easier and more convenient

chromiumboy:
> I don't think this approach is quite right, and it will be difficult to expand. I would suggest implementation through some Generic !type: loadouts entries that are applied to the player entity as EntityEffects or something similar when connecting to the game. In this case, extending the new selection functionality via loadouts will be much easier and more convenient

Station AI behave very differently to crew in that they don't carry items and their entity is hidden away in their core, so their customization data needs to be handled differently. Also loadouts are intended for storing data between rounds, while AI customizations are done in round and do not persist (PJB was insistent that the AI not have loadouts for these customizations and they should be dynamically changeable)

Resources/Prototypes/AppearanceCustomization/station_ai.yml
```diff
@@ -0,0 +1,145 @@
...
+  previewKey: Hologram
+  protoIds:
+  - StationAiHologramFemale
+  - StationAiHologramMale
+  - StationAiHologramFace
+  - StationAiHologramCat
+  - StationAiHologramDog
+
+# Iconography
+- type: stationAiCustomization
```
slarticodefast, line 29: I didn't look at all of the code, but couldn't these simply be a normal entity prototype instead of a custom one?
With the hologram shader and a generic visualizer for the dead state.

chromiumboy: Oh, that's a good idea. I can spawn a dummy entity and use sprite view to give a preview with the hologram shader too. I'll have a look. Can the client see entities in null space?

chromiumboy: Not sure if the iconography would work well as an entity proto though

slarticodefast: >Can the client see entities in null space

Oh, that part might be a little more tricky. You may be able to simply spawn them clientside?

IProduceWidgets:
Cat and corgi I'm not sure are a good idea because they're not easily distinguishable from real cats or real corgis. The humans at least have a unique "outfit" on.

IProduceWidgets:
> > Cat and corgi I'm not sure are a good idea because they're not easily distinguishable from real cats or real corgis.
> 
> they're blue

always are
image

Resources/Locale/en-US/silicons/station-ai.ftl
```diff
@@ -22,3 +22,25 @@ toggle-light = Toggle light
...
+station-ai-icon-angel = Guardian angel
+station-ai-icon-bliss = Simpler times
+station-ai-icon-clown = Clownin' around
+station-ai-icon-dorf = Adventure awaits
+station-ai-icon-heartline = Lifeline
+station-ai-icon-smiley = All smiles
+
+station-ai-hologram-female = Female appearance
+station-ai-hologram-male = Male appearance
+station-ai-hologram-face = Disembodied head
```
Tayrtahn, line 44: I'm wondering if these should specify female *human* appearance, etc. Maybe have options for the AI to appear as other playable races too?

chromiumboy:
> Cat and corgi I'm not sure are a good idea because they're not easily distinguishable from real cats or real corgis. The humans at least have a unique "outfit" on.

Fair point, but they're cute!

Resources/Locale/en-US/silicons/station-ai.ftl
chromiumboy: I can update the description but I don't have plans to make options for other races, I'm using already available sprites for the holograms

Tayrtahn: Sounds good. If they're wanted, someone can add them later.

Emisse:
do borgs have the tool/chassis selection on spawn yet? if yes, this should be wrapped into that system. otherwise this should be expanded into a generic system that fits that. i havent kept up on the silicons sphere of things in a while though due to busy but thats my input.

vladkaron14R:
Hello, can I find out if I can upload my sprite to the kernel so that only I can see it?

SlimmSlamm:
> do borgs have the tool/chassis selection on spawn yet? if yes, this should be wrapped into that system. otherwise this should be expanded into a generic system that fits that. i havent kept up on the silicons sphere of things in a while though due to busy but thats my input.

they got chassis selection on spawn a month ago i believe.

chromiumboy:
> do borgs have the tool/chassis selection on spawn yet? if yes, this should be wrapped into that system. otherwise this should be expanded into a generic system that fits that. i havent kept up on the silicons sphere of things in a while though due to busy but thats my input.

Yes, borg type selection is in, see #32586

I'm not sure about merging these two systems together as they function somewhat differently / have different uses. Borg switching is a one off event that changes the entire entity, while station AI customization is repeatable and ony affects appearance. The customizations also have to follow the AI so if they are inserted into a new core, the customizations get reapplied, while borgs don't need to worry about that

But if we really want a single generic system to handle all this, I'm gunna leave it to someone else. I'm not really interested in rewriting this code again, I've done that once already to meet PJB's requirements

ruddygreat:
Could you add AI name customisation to this as well, or would that be too out of scope?
Being addled with a random name at the start of a round is kinda just annoying.

he1acdvv:
> Cat and corgi I'm not sure are a good idea because they're not easily distinguishable from real cats or real corgis. The humans at least have a unique "outfit" on.

You might be colorblind.

aeoli-fly:
maybe adding something like a positronic brain to the hologram options? I'd reckon some people would prefer to go completely inorganic/inhuman

Mixelz:
> Could you add AI name customisation to this as well, or would that be too out of scope? Being saddled with a random name at the start of a round is kinda just annoying.

Choosing a name is already something you can do in the load out for station AI.

EmoGarbage404 (approved):
code looks good to me. just needs to compile

ArtisticRoomba (approved):
Looks all good to me.

M1tht1c:
> > is this one-time or is it something the AI can do as long as they can click on themselves?
> 
> 
> 
> You can do it as often as you like :)

I think you might have misunderstood their question? What they're asking is whether you can change it while in the round - rather than just in the lobby. (If that is what you're saying, sorry, but your answer seemed to be answering a different question to me)

chromiumboy:
To clarrify, all settings are changed in round, not in the lobby - these customizations are not loadouts. See the video linked in the media section of the PR description for a demonstration

Issue #34482, opened 2025-01-17
Entities are losing their fixtures in space.


kanopus952:
## Description

When you put an entity in space, it disappears shape, aabbs and becomes invulnerable to RMB and hitscan attacks. But projectiles can still hit it. This problem is critical enough to the servers that they decided to convert all the projectiles to hitscan. 
**Reproduction**

1. Put any entity into space
2. Send command physics shapes or physics aabbs into console
3. Try to shoot with any hitscan rifle
4. See that entity lost it shape
**Screenshots**

![Image](https://github.com/user-attachments/assets/e7821a9a-24ca-4a3c-986f-74b210232772)

kanopus952:
I tried tracking down any changes to the FixturesComponent, but didn't find anything super unusual. This bug appeared about 3-4 weeks ago, and may be related to code changes in FixturesComponent

PJB3005:
@metalgearsloth @ElectroJr

PR #28797, opened 2024-06-09
Fix laser beam visuals in space

n00kii:
## About the PR

fixed the visual effects for hitscan weaponry when fired between grids or in space (resolves #23699)
## Why / Balance

bugfix

## Technical details

add an extra field to `HitscanEvent` to keep track of the grid that the shot was fired from, attach all effect segments to that grid. because post-spawn grid traversal, requires an engine change to allow traversal to be bypassed and keep the effects attached to the source grid.

Requires https://github.com/space-wizards/RobustToolbox/pull/5365

## Media

previous behavior
<video src="https://github.com/space-wizards/space-station-14/assets/57325298/bced9e39-b962-4f75-b160-6745dc10bce4"></video>

new behavior
<video src="https://github.com/space-wizards/space-station-14/assets/57325298/0e64ea19-afe0-4f56-80e6-51c76c7e54d6"></video>

[X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- fix: Fixed misalignment of laser effects when shot across grids.

Issue #5365, opened 2021-11-16
Trying to sync a protolathe with an R&D server past a certain distance throws an exception.


TimrodDX:
## Description

How to replicate:

- Be on Packedstation, as I don't think this is possible to trigger on Saltern due to how small it is.
- Build (or spawn) a Protolathe in Cargo or Engineering. I've had it happen in both places.
- Use the Protolathe and try to choose a server to sync from.
- VS crashes with the following exception:

![image](https://user-images.githubusercontent.com/7907891/141876804-bf96f315-f90b-48cd-a19c-9f4d1b92c082.png)

PR #34515, opened 2025-01-19
Fix hitscan visuals

### LINKED ISSUE #23699: Strange laser visual behaviour in open space
## Description

When I was playtesting shuttle lasser cannons, I noticed that lasers behave very strangely in space.

Apparently a shot from a moving shuttle causes errors in ray mapping, because of which visually the shots are displayed at the wrong angles. The raycast and damage dealing works correctly.

**Reproduction**

fire hitscan weapons while on a moving and rotating grid

**Screenshots**

image
 
https://youtu.be/0ZcT5bHQkt8

ElectroJr:
## About the PR
Fixes hitscan visuals when shooting across grids.

## Technical details
This PR just makes sure that the angle passed in the event is always relative to the entity associated with the `fromCoordinate`, and makes sure that all the parts of the beam have the same world rotation.

I finished this PR after fixing #34482, but before realising that there is already an open PR that also tries to fix this (#28797), though it's been stuck waiting on an engine PR that was waiting for reviews/responses. I'm not sure about the required engine changes and think a better fix would be to just combine the laser effect into a singular entity (which this neither this PR or the other PR does).

In the meantime, this PR still fixes the main issue w/o requiring any engine changes, though it doesn't fix the problem that the beam can split when grids move or rotate.

fixes #23699

## Media

Before: see video in #5365

After:

https://github.com/user-attachments/assets/00128ef1-997f-4aa5-9651-3ec5042fd8f3

## Requirements

**Changelog**

IMO @n00kii still deserves the credit for initially working on a fix for this bug, but if they don't want their name on the CL I can remove it

:cl:  n00kii
- fix: Fixed laser weapon visuals being broken when shooting across grids or into space.


Content.Server/Weapons/Ranged/Systems/GunSystem.cs
```diff
@@ -396,23 +396,24 @@ private void FireEffects(EntityCoordinates fromCoordinates, float distance, Angl
...
         var angle = mapDirection;
 
         // We'll get the effects relative to the grid / map of the firer
         // Look you could probably optimise this a bit with redundant transforms at this point.
-        var xformQuery = GetEntityQuery<TransformComponent>();
 
-        if (xformQuery.TryGetComponent(gridUid, out var gridXform))
+        var gridUid = fromXform.GridUid;
+        if (gridUid != fromCoordinates.EntityId && TryComp(gridUid, out TransformComponent? gridXform))
         {
```
slarticodefast, line 407: Didn't all entities have a TransformComponent? So no need to do a TryComp
```suggestion
        if (gridUid != fromCoordinates.EntityId)
        {
            var gridXform = Transform(gridUid);
```

Content.Server/Weapons/Ranged/Systems/GunSystem.cs
```diff
@@ -396,23 +396,24 @@ private void FireEffects(EntityCoordinates fromCoordinates, float distance, Angl
         // Forgive me for the shitcode I am about to do
         // Effects tempt me not
         var sprites = new List<(NetCoordinates coordinates, Angle angle, SpriteSpecifier sprite, float scale)>();
-        var gridUid = fromCoordinates.GetGridUid(EntityManager);
+        var fromXform = Transform(fromCoordinates.EntityId);
         var angle = mapDirection;
```
slarticodefast, line 400:
```suggestion
```
This is a new variable and not referencing the list above, right? In that case we can just delete this line and rename the variable.

slarticodefast (changes requested):

Content.Server/Weapons/Ranged/Systems/GunSystem.cs
ElectroJr: Yeah, I was just trying to keep changes to a minimum, but might as well clean it up a bit.

Content.Server/Weapons/Ranged/Systems/GunSystem.cs
ElectroJr: gridUid might be null here, so the `TryComp` also acts as a ` != null` check while simultaneously getting the component and I don't think it makes it any harder to read

slarticodefast (approved):
Like you metioned in the code comment, this is still very jank and needs some rewriting in the future, but at least it works as intended for now, so this is an improvement.
Thank you for fixing this ancient bug!

PR #34521, opened 2025-01-19
Metabolizing bloodstream and hypervolemia

nkokic:
## About the PR

Bloodstream now acts as metabolizing solution.
Changed blood loss damage scaling to be linear. Having blood loss now does double asphyxiation instead.
Added hypervolemia damage which causes minor slash with bleeding effect and asphyxiation.
Hypervolemia damage scales similarly with blood loss after HyperVolemiaThreshold in opposite direction.

## Why / Balance

To increase medical immersion, punish reckless chemical dumping and make gluttony an actual concern.
Having disrupted blood level is now a big enough concern that it is not something to be shruged off.

## Technical details

Mainly edited BloodstreamSystem and BloodstreamComponent to implement hypervolemia similarly to how blood loss is implemented.
Changed wherever DefaultChemicalsSolutionName was used in context of transferring chemicals or metabolizing, DefaultBloodSolutionName was put instead.
Due to how BloodstreamSystem treats BloodMaxVolume differently, I have changed its name to BloodReferenceVolume because it does not mean maximum volume anymore. (in C# and YAML files)
Maximum bloodstream solution is double of BloodReferenceVolume.
Blood level is calculated by bloodsolution volume / bloodReferenceVolume.
Made sure that charcoal's cleansing of chemicals does not flush blood reagents.
Made sure rejuvenating removes everything from bloodstream and then refills it with right amount of blood.
Saline also has been also turned down to be equal with blood.

## Media

How hypervolemia looks on scanner and what's under the car hood:
image

Visual representation of damage scaling:
image

## Requirements

## Breaking changes

For forked repositories:
May cause issues if forks have custom mobs.
Change bloodMaxVolume into bloodReferenceVolume.

**Changelog**

:cl:
- add: Added guide entry about blood pressure and it's management
- add: Added hypervolemia damage. Consists of asphyxiation and minor slash.
- tweak: Bloodloss damage is applied linearly. Potentially more dangerous than usually.
- tweak: Changed bloodloss damage to stronger asphyxiation damage only
- tweak: Maximum blood level is now double of healthy blood level
- tweak: Rerouted all chemicals that mob can take into bloodstream instead of chemical solution


Hrosts:
Time to bring back the true and tried practice of bloodletting! 🔪

Resources/Prototypes/Entities/Mobs/base.yml
```diff
@@ -229,11 +229,15 @@
...
+    solution: bloodstream
   - type: Bloodstream
     bloodlossDamage:
       types:
-        Bloodloss: 0.5
+        Asphyxiation: 2
+    hypervolemiaDamage:
+      types:
+        Slash: 0.1
+        Asphyxiation: 1.90
```
0x6273, line 240:
```suggestion
        Asphyxiation: 1.9
```

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -189,7 +208,7 @@ private void OnComponentInit(Entity<BloodstreamComponent> entity, ref ComponentI
             return;
 
         chemicalSolution.MaxVolume = entity.Comp.ChemicalMaxVolume;
-        bloodSolution.MaxVolume = entity.Comp.BloodMaxVolume;
+        bloodSolution.MaxVolume = entity.Comp.BloodReferenceVolume * 2; // enables high fluid volume
```
0x6273, line 211: This value probably shouldn't be hardcoded.

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -356,6 +386,14 @@ public void SetBloodLossThreshold(EntityUid uid, float threshold, BloodstreamCom
         comp.BloodlossThreshold = threshold;
     }
 
+    public void SetHypervolemiaThreshold(EntityUid uid, float threshold, BloodstreamComponent? comp = null)
```
0x6273, line 389: Make it take `Entity<BloodstreamComponent?> ent` instead of separate args for uid and comp.

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -120,8 +120,9 @@ public override void Update(float frameTime)
...
                 continue;
 
-            // Adds blood to their blood level if it is below the maximum; Blood regeneration. Must be alive.
-            if (bloodSolution.Volume < bloodSolution.MaxVolume && !_mobStateSystem.IsDead(uid))
+            var bloodPercentage = GetBloodLevelPercentage(uid, bloodstream);
+            // Adds blood to their blood level if it is below the reference; Blood regeneration. Must be alive.
+            if (bloodPercentage < 1 && !_mobStateSystem.IsDead(uid))
             {
                 TryModifyBloodLevel(uid, bloodstream.BloodRefreshAmount, bloodstream);
             }
```
0x6273, line 128: This could fill a bit above `BloodReferenceVolume` if you're missing less blood than `BloodRefreshAmount`. It should probably be capped.

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -302,11 +321,14 @@ private void OnRejuvenate(Entity<BloodstreamComponent> entity, ref RejuvenateEve
...
-            TryModifyBloodLevel(entity.Owner, bloodSolution.AvailableVolume, entity.Comp);
-
         if (_solutionContainerSystem.ResolveSolution(entity.Owner, entity.Comp.ChemicalSolutionName, ref entity.Comp.ChemicalSolution))
             _solutionContainerSystem.RemoveAllSolution(entity.Comp.ChemicalSolution.Value);
+
+        if (_solutionContainerSystem.ResolveSolution(entity.Owner, entity.Comp.BloodSolutionName, ref entity.Comp.BloodSolution))
+            _solutionContainerSystem.RemoveAllSolution(entity.Comp.BloodSolution.Value);
+
+        if (_solutionContainerSystem.ResolveSolution(entity.Owner, entity.Comp.BloodSolutionName, ref entity.Comp.BloodSolution, out var bloodSolution))
+            TryModifyBloodLevel(entity.Owner, entity.Comp.BloodReferenceVolume, entity.Comp);
```
0x6273, line 331: Don't resolve the same solution twice.

0x6273:
Do things like iron pills cause hypervolemia? I haven't tested it, but from reading the code it seems like that would be the case. Not sure if that's intended.

nkokic:
Everything has potential to cause hypervolemia, which is most of the time temporal. However, all chemicals that raise actual blood reagent volume (iron, copper, saline and even protein) causes a more permanent hypervolemia.
It is partially intentional so medical players have to draw blood to cure it.
Currently if reagent is the same as blood reagent, it will be skipped for metabolizing.
I was thinking of diuretic feature of bloodstream. To make kidneys (or metabolism) flush all excess chemicals IF the chemicals are hydrophilic, while in state of hypervolemia. For example: blood level is 150%, saline is hydrophilic (literally salt water) and could be flushed, skipping metabolizing part or still wait as if metabolized, just not have any function. While, let's say bicardine is not hydrophilic and sticks around until metabolized fully.
Could do same with blood reagent.
Another solution, would be to check for blood reagent amount and let blood to be metabolized anyway if blood reagent amount (not blood level) is above reference.

Solutions that could be implemented:
- specific behavior/condition for chemicals when blood level is under/over X (I highly recommend)
- let blood be metabolized if blood reagent volume is above reference (quick fix)

What are your opinions.

crazybrain23:
Does this mean you can drink blood to cure blood loss?

nkokic:
Yes. It is possible, but in dire situations, drinking
blood from floor is impossible to cure blood loss in time.

It is also possible to turn stomach into centrifuge which could solve this problem. And have already managed to get this part working on my fork. However, chemical ANY mixes in stomach do not get passed onto metabolizing solution. There is an open issue about it... 

My another idea was adding very tiny amount of "sepsis" reagent to chemicals on floor, or everytime they get spilled. Making drinking blood not sanitary, but also everything else too.

Content.Server/Body/Systems/BloodstreamSystem.cs
nkokic: Should I do it to `SetBloodLossThreshold` as well?

nkokic:
NOTES:
1 saline = 12 blood
1 protein = 2 blood
1 iron/copper = 3.2 blood (looking at code it should be 4, but something suspicious is going on while counting updates)

CHANGES:
1 saline = 1 blood
1 protein = 0 blood

Realistically, saline should be less than equivalent than blood, but we do not have IV stands to make huge injections of saline possible.
Increase of protein should not directly increase blood. While blood cells do require proteins to be made, it is mostly the proteins that we already have stored and not sudden increase of protein in bloodstream.

ScarKy0:
Heya! We brought this up for maintainer discussion and decided to close this PR.
While we really love this idea, as it stands our systems aren't really ready for this sort of thing and they would need updating, which would increase the scope of this PR too much.
Ideally we would need to have proper support and implementation for different methods of taking medicine first. (injection could put things in the bloodstream and have its own effect, but digestion to a seperate container where it has different effects)

PR #34530, opened 2025-01-20
Componentize Rat King "Rummaging."

widgetbeck:
## About the PR

Separated out the "rummage" verb and its associated systems from the Rat King component. Also added several options to make it more flexible for developer use. 

## Why / Balance

Many older components (e.g. Rat King, Dragon, Drone) combine several unrelated systems into one monolithic component with very little YML customizability, which I feel goes against one of the core aims of an ECS - modularity. 
My goal in my next few upstream PRs is to separate out these (niche, but useful) mechanics so that they can be more easily used by developers without having to develop their own systems that do the same thing, and in so doing provide more variables which can be manipulated without delving into CS.

## Technical details

In this PR, I:
- Create two new components, CanRummage and Rummageable, which replace the same parts of Rat King component.
- Create a new system, RummagingSystem, which performs the functions previously confined to Rat King component.
- Add several options to CanRummage, including:
  - The ability to define a default rummage loot table per-Rummager, as opposed to per-Rummageable object -  this can be overridden on the Rummageable object.
  - The ability to customize the LocId of the verb.
  - A rummage speed modifier.
- Add several options to Rummageable, including:
  - The ability to mark a Rummageable entity as "re-lootable", and a definable cooldown for when its loot will be reset.  
  - The ability to override the default loot table defined in Rummaging.

## Requirements

## Breaking changes

**Changelog**

I have not included a changelog because these changes do not immediately affect gameplay.


widgetbeck:
will fix the linter errors shortly.

Content.Shared/Rummaging/RummagingComponent.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// A weighted random entity prototype containing the different loot that rummaging can provide. 
+    /// Defining this on the rummager so different things can get different stuff out of the same container type.
+    /// Can be overridden on the entity with Rummageable. 
+    /// </summary>
+    [DataField("rummageLoot", customTypeSerializer: typeof(PrototypeIdSerializer<WeightedRandomEntityPrototype>)), ViewVariables(VVAccess.ReadWrite)]
+    [AutoNetworkedField]
+    public string RummageLoot = "RatKingLoot";
+
+    [DataField("rummageVerb"), ViewVariables(VVAccess.ReadWrite), AutoNetworkedField]
+    public LocId RummageVerb = "verb-rummage-text";
```
iaada, line 26: This probably doesn't need to be networked.

iaada (commented):
I like this PR! Here's a few things I noticed that I've gotten feedback on before.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <summary>
+    /// Runs on getting the verbs of a rummageable entity, raised on that entity.
+    /// </summary>
+    /// <param name="uid"></param>
+    /// <param name="rummageable"></param>
+    /// <param name="args"></param>
+    private void OnGetVerb(Entity<RummageableComponent> entity, ref GetVerbsEvent<AlternativeVerb> args)
+    {
+        // if the entity is relootable and the cooldown has passed, reset looted status.
+        if (entity.Comp.Relootable && entity.Comp.NextRelootable < _gameTiming.CurTime)
```
beck-thompson, line 36: Just make a function called like `IsLootable` that does this check and use that here and on the doafter! The setting the looted field to false is suspect when its actually the time that effect it.

Content.Shared/Rummaging/RummageableComponent.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    [DataField, AutoNetworkedField]
+    public TimeSpan RelootableCooldown = TimeSpan.FromSeconds(60);
+    public TimeSpan NextRelootable;
+
+    /// <summary>
+    /// A weighted loot table. 
+    /// Overrides the same setting on RummagingComponent.
+    /// </summary>
+    [DataField, AutoNetworkedField]
```
beck-thompson, line 35: Doesn't need to be auto networked!

This is what is causing your test fails specifically. When you `Dirty` a component what that means is it'll send all the auto networked fields from the server to the client. However, to do this, it has to "serialize" the data (E.g pack it up and put it in a nice form so it can be sent over the network. Think of it kind of like pass by value vs pass by reference but even more intensive). Apparently `EntityTableSelector`s aren't able to be networked so when you made this `AutoNetworked` it failed!

Content.Shared/Rummaging/RummageableComponent.cs
```diff
@@ -0,0 +1,49 @@
...
+/// </summary>
+[RegisterComponent, NetworkedComponent, Access(typeof(RummagingSystem))]
+[AutoGenerateComponentState]
+public sealed partial class RummageableComponent : Component
+{
+    /// <summary>
+    /// Whether or not this entity has been rummaged through already.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Looted;
```
beck-thompson, line 19: Should be removed! (See below comment about it)

Content.Shared/Rummaging/RummageableComponent.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    /// Whether or not this entity can be rummaged through multiple times.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Relootable = false;
+
+    [DataField, AutoNetworkedField]
+    public TimeSpan RelootableCooldown = TimeSpan.FromSeconds(60);
+    public TimeSpan NextRelootable;
```
beck-thompson, line 29: Add `AutoGenerateComponentPause` to the top of the component and `[AutoPausedField]` here. That'll make it so this properly gets stops when the game is paused!

Content.Shared/Rummaging/RummageableComponent.cs
```diff
@@ -0,0 +1,49 @@
...
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Rummaging;
+
+/// <summary>
+/// This is used for entities that can be
+/// rummaged through to get loot.
+/// </summary>
+[RegisterComponent, NetworkedComponent, Access(typeof(RummagingSystem))]
+[AutoGenerateComponentState]
```
beck-thompson, line 12: I don't any of these fields need to be auto networked! Probably remove it and all the corresponding tags.

beck-thompson (changes requested):
Only other thing is, could you rename `Rummagning`component/system to something like "CanRummage" or something? It was kind of confusing at first!

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,112 @@
...
+                    DistanceThreshold = 2f
+                });
+            }
+        });
+    }
+
+    /// <summary>
+    /// Runs after the do-after. Handles spawning items from YML-definable loot tables.
+    /// </summary>
+    /// <param name="uid"></param>
```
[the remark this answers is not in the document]
widgetbeck: The comment? I'd rather have more comments than necessary than fail to document stuff, to be honest.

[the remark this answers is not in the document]
beck-thompson: Yeah I agree but this is a generic event listener and normally those don't get comments! Every event listener is basically the same so unless there is something weird going they don't really have them.

If you do want to keep it, just remove the param tags because they are empty and not necessary

Content.Shared/Rummaging/RummageableComponent.cs
widgetbeck: I actually still need it to track the lootable status of non-Relootable entities. I did remove its datafield and autonetworkedfield tags because they were unnecessary.

Content.Shared/Rummaging/RummagingSystem.cs
widgetbeck: good idea. sometimes i forget about return types other than void lmao

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,112 @@
...
+        if (entity.Comp.Relootable)
+            entity.Comp.NextRelootable = _gameTiming.CurTime + entity.Comp.RelootableCooldown;
+    }
+
+    // checks Relootable status
+    public bool IsCurrentlyLootable(Entity<RummageableComponent> entity)
+    {
+        if (entity.Comp.Relootable)
+        {
+            if (entity.Comp.NextRelootable < _gameTiming.CurTime)
```
beck-thompson, line 106: Simplify all this to like `return !entity.CompRelootable || entity.Comp.NextRelootable < _gameTiming.CurTime`. If you want just add a comment above explaining

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,112 @@
...
+        if (!entity.Comp.Looted)
+            entity.Comp.Looted = true;
+
+        // and set the next refresh if the entity is relootable.
+        if (entity.Comp.Relootable)
+            entity.Comp.NextRelootable = _gameTiming.CurTime + entity.Comp.RelootableCooldown;
+    }
+
+    // checks Relootable status
+    public bool IsCurrentlyLootable(Entity<RummageableComponent> entity)
```
beck-thompson, line 102: If this is public, add full comments for all inputs and outputs!

beck-thompson (changes requested):
Just really minor stuff! Doing some final tests now

Content.Shared/Rummaging/RummagingSystem.cs
widgetbeck: fair enough. i just find the way i've written it here easier to follow the logic flow of

beck-thompson: The only one that I found confusing is the second if statement. If you want to keep the first one that would make sense!

Content.Shared/Rummaging/RummagingSystem.cs
widgetbeck: No real reason for it to be public so I just made it private lol

widgetbeck:
Your way is cleaner I reckon.

beck-thompson (approved):
Tested and it works great! Good pr :+1:

Content.Shared/Rummaging/CanRummageComponent.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    [DataField]
+    public float RummageModifier = 1f;
+}
+
+[Serializable, NetSerializable]
+public sealed partial class RummageDoAfterEvent : SimpleDoAfterEvent
+{
+
+}
```
MilonPL, line 39: This doesn't need a body.

Content.Shared/Rummaging/RummageableComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+/// This is used for entities that can be
+/// rummaged through to get loot.
+/// </summary>
+[RegisterComponent, NetworkedComponent, Access(typeof(RummagingSystem)), AutoGenerateComponentPause]
+public sealed partial class RummageableComponent : Component
+{
+    /// <summary>
+    /// Tracks whether or not this particular object has been looted.
+    /// </summary>
+    public bool Looted = false;
```
MilonPL, line 18: Don't initialize bools like this, they're false by default.

Content.Shared/Rummaging/RummageableComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Whether or not this entity can be rummaged through multiple times.
+    /// </summary>
+    [DataField]
+    public bool Relootable = false;
```
MilonPL, line 24: Above.

Content.Shared/Rummaging/RummageableComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    [DataField, AutoPausedField]
+    public TimeSpan RelootableCooldown = TimeSpan.FromSeconds(60);
+    public TimeSpan NextRelootable;
```
MilonPL, line 28: Why do you only have the attribute for the top field?
Static TimeSpans that aren't runtime-modified can be just marked as datafields, while `NextRelootable` should have *both* the `TimeOffsetSerializer` and `AutoPausedField`.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Network;
+using Robust.Shared.Timing;
+
+
+namespace Content.Shared.Rummaging;
+public sealed class RummagingSystem : EntitySystem
+{
+    [Dependency] private readonly INetManager _net = default!;
+    [Dependency] private readonly IGameTiming _gameTiming = default!;
```
MilonPL, line 14: Just `_timing` here.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+    [Dependency] private readonly SharedAudioSystem _audio = default!;
+    [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
+    [Dependency] private readonly EntityTableSystem _entityTable = default!;
```
MilonPL, line 17: Sort these alphabetically.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+        // otherwise, add the verb.
+        args.Verbs.Add(new AlternativeVerb
+        {
+            Text = Loc.GetString(rummaging.RummageVerb),
+            Priority = 100, // needs to be the highest-prio alt verb, otherwise it just doesn't show up.
+            Act = () =>
+            {
+                var rummageDuration = entity.Comp.RummageDuration * rummaging.RummageModifier;
+                _doAfter.TryStartDoAfter(new DoAfterArgs(EntityManager, user, rummageDuration,
+                    new RummageDoAfterEvent(), entity.Owner, entity.Owner)
```
MilonPL, line 49: You don't need `.Owner` here because `Entity<T>` casts to uid.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        });
+    }
+
+    /// <summary>
+    /// Runs after the do-after. Handles spawning items from YML-definable loot tables.
+    /// </summary>
+    /// <param name="uid"></param>
+    /// <param name="rummageable"></param>
+    /// <param name="args"></param>
```
MilonPL, line 65: Docs don't match the method signature, either remove all this (which is fine for private methods) or update it.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        // this is mostly here to grab the rummaging component.
+        if (!TryComp<CanRummageComponent>(args.User, out var rummaging))
+            return;
+
+        if (args.Cancelled || !IsCurrentlyLootable(entity))
+            return;
+
+        Dirty(entity.Owner, entity.Comp);
+        _audio.PlayPredicted(entity.Comp.Sound, entity.Owner, args.User);
```
MilonPL, line 76: Remove `.Owner`.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+            var spawn = _entityTable.GetSpawns(entity.Comp.RummageLoot).First();
+            if (_net.IsServer)
+                Spawn(spawn, Transform(entity.Owner).Coordinates);
+        }
+        // otherwise, uses the user's settings.
+        else if (rummaging.RummageLoot != null)
+        {
+            var spawn = _entityTable.GetSpawns(rummaging.RummageLoot).First();
+            if (_net.IsServer)
+                Spawn(spawn, Transform(entity.Owner).Coordinates);
```
MilonPL, line 90: Remove `.Owner`.

Content.Shared/Rummaging/RummagingSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+    private void OnDoAfterComplete(Entity<RummageableComponent> entity, ref RummageDoAfterEvent args)
+    {
+        // this is mostly here to grab the rummaging component.
+        if (!TryComp<CanRummageComponent>(args.User, out var rummaging))
+            return;
+
+        if (args.Cancelled || !IsCurrentlyLootable(entity))
+            return;
+
+        Dirty(entity.Owner, entity.Comp);
```
MilonPL, line 75: When calling Dirty on `Entity<T>`, you can just do it as `Dirty(entity)`, as we get both the component and the EntityUid from it. This signature should be used either when you want to dirty a component that is different from the Type in Entity, or when using EntityUid.

MilonPL (changes requested):

MilonPL:
Please address the review.

MilonPL:
I am going to be closing this PR as Stale. If you ever end up coming back to this, please open a new PR.

Issue #29859, opened 2024-07-09
Slipping prediction issues


slarticodefast:
## Description
If you throw slippery items at a moving mob or player, they sometimes get slipped even while the item is in the air. This is not supposed to happen since #24494.
There are some prediction problems happening with `SlipCausingAttemptEvent` in `ThrownItemSystem.cs`.

**Reproduction**
Case 1:
- Run into one direction and throw soap in the direction you are running to. This will cause you to slip when it flies above you.

Case 2: 
- Build a monkey corridor like this:
![Screenshot (189)](https://github.com/space-wizards/space-station-14/assets/161409025/335fb648-8b37-416d-9127-58eed35a24db)

- Spawn an AI monkey and immediately throw soap at him
- It will run towards you while the soap flies above him
- Sometimes he slips on the client but not on the server, causing him to fall down but keep running towards the gas analyzer. Afterwards his sprite is bugged and will stay horizontal.

**Additional context**
If a slippery item like soap is thrown above a moving mob or player, the `SlipCausingAttemptEvent` is supposed to cancel the slip, but this sometimes only happens on the server, but not on the client. This prediction problem was already mentioned in #24494, but disregarded as uncommon. My recent change to throwing mechanics (#29726) made thrown items fly less far and therefore a little slower, causing more slip events that have to get cancelled and more of these mispredictions to happen. You can compare the current upstream to c8a87ceaab which is the last commit before my PR. The bug happens in both, but is more rare before.

luizwritescode:
Recently I've seen an artifact spawn slippery items (banana peels iirc) as multiple players slipped and got "stuck" with a horizontal sprite at the same time. It was pretty funny, but I guess this should be fixed since visual bugs will invite players to break RP and chat in LOOC.
![image](https://github.com/user-attachments/assets/a2a5a06b-3599-4176-8780-2ca97a7d8bb9)
While Case 2 is pretty easy to reproduce, Case 1 is not. After testing many soap throws while running, depending on the distance:
- far: normal behaviour
- midrange: will wrongfully predict slipping on the client but you will almost instantly get up
- close: will slip you but not play the slipping sound

PR #34531, opened 2025-01-20
Wizard Touch Spells (Smite, Cluwne's Curse, Slippery Slope)

keronshb:
## About the PR

Smite - Gibs target

Cluwne's Curse - Turns the target into a Cluwne.

Slippery Slope - Makes the target slippery, and they slip other people by simply walking into them!

All of these spells will require Wizard Hat and Robes, so you can't just polymorph into something or mind swap and use it.

## Why / Balance

Smite is a classic spell. It round removes, but that's the point.

Cluwne has become a staple spell as well, arguably you're worse off effected by this than smite.

Slippery is something I came up with. It can definitely backfire on the Wizard and that exact chaotic nature is a perfect fit for the Wizard.

## Media

https://github.com/user-attachments/assets/2773716b-6828-43ea-bca9-b44c470a938e

There's a slipping misprediction also in the video. Been reported for awhile - https://github.com/space-wizards/space-station-14/issues/29859

Can't quite figure it out.

## Requirements

## Breaking changes

**Changelog**


IProduceWidgets:
I'm a little worried the slippery one is going to be prediction hell

IProduceWidgets:
Also, would be funny to make the judge robe count as a robe for wizard purposes. 
*You're a legal wizard!*

Resources/Prototypes/Magic/touch_spells.yml
```diff
@@ -0,0 +1,74 @@
...
+    interactOnMiss: false
+    sound: !type:SoundPathSpecifier
+      path: /Audio/Magic/disintegrate.ogg
+    icon:
+      sprite: Objects/Magic/magicactions.rsi
+      state: gib
+    event: !type:SmiteSpellEvent
+      speech: action-speech-spell-smite
+  - type: Magic
+    requiresClothes: true
```
IProduceWidgets, line 22: probably breaks the smite snail, which is whatever. But we should either have a non-clothes smite or remove the snail.

keronshb:
I've free'd the snail

Golubgik:
Maybe this will be merged in the near future? Within a week, for example. @keronshb

ScarKy0 (changes requested):
Other than this small thing it looks all fine to me
Also, Slippery Slope seems to play the slip sound effect whenever an affected mob walks, should probably be fixed.

Content.Shared/Slippery/SlipperySystem.cs
```diff
@@ -79,11 +80,21 @@ private void OnGetSlowedOverSlipperyModifier(EntityUid uid, SlowedOverSlipperyCo
         args.Args.SlowdownModifier *= comp.SlowdownModifier;
     }
 
+    private void OnCollide(Entity<SlipOnCollideComponent> ent, ref StartCollideEvent args)
+    {
+        if (args.OtherEntity == args.OurEntity)
+            return;
+
+        var slipComp = EnsureComp<SlipperyComponent>(ent.Owner);
+
```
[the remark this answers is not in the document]
slarticodefast: first one is fine

Content.Shared/Slippery/SlipOnCollideComponent.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Robust.Shared.GameStates;
+
+namespace Content.Shared.Slippery;
+
+/// <summary>
+///     Put on entities to make them slip others when they walk into them (Separate from the step trigger). Also works on throwing ents like Hamlet, where regular <see cref="SlipperyComponent"/> won't
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class SlipOnCollideComponent : Component
```
slarticodefast, line 9: I'm not sure I like this one, why not just add `SlipperyComponent` directly?

ScarKy0: I believe its cause when you walk into mobs they dont slip unless theyre moving, this spell is meant to make you slip mobs on walking into them

slarticodefast: steptrigger uses `StartCollideEvent` as well, so I don't see why the subscription needs to be duplicated. And I'm worried that we are bypassing some important events getting raised this way.

keronshb: This is needed because StepTrigger is handled differently.
We're not skipping any extra checks because we run TrySlip, which has all the checks we need.
Which I should probably add the CanSlip check as well.

ScarKy0 (approved):
Code looks okay, works in game. Only issue is that slipping with Slippery Slope happens twice, but that is an issue with how soap works in general so not really a blocker or anything like that.
Wizard Soon(tm)

ScarKy0:
Oh also, additionally approved by Slarti on the wizard thread on discord.

PR #33267, opened 2024-11-12
Fixed issue where entities in containers would be deleted when the construction process advances to another node

kbarkevich:
## About the PR
The `Content.Server/Construction/ConstructionSystem.Graph.cs` file was changed to allow entities inside containers to persist during construction as the entity is changed from node to node.

## Why / Balance
The development docs (specifically https://docs.spacestation14.com/en/space-station-14/core-tech/construction.html#containers) state that "When the entity changes due to reaching a node with a different entity prototype, all those containers *will be transferred to the new entity*. " I discovered this bug when creating a new entity that involved a multi-node/multi-entity construction process, and also necessitated that the specific entities stored within the construct are available to be edited and retrieved in order to function. I found that when the construction involves more than one step, the entities stored within the containers are deleted/not carried over to the next entity.

This also fixes issue #32590

Before: 
image

After:
image

## Technical details
Within the previously mentioned file, the `ChangeEntity` method was altered. This alteration is within the "Container transferring" section of the method. The alteration iterates over all containers in the previous construction entity and ensures that they are transferred along to the next entity.

## Media
N/A, unless requested

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**
N/A

PR #34542, opened 2025-01-20
Fixed Intercom Encryption Key Deletion

### LINKED ISSUE #32590: Secure intercoms (command and sec) deleting comms keys
## Description

When removing the security pannel from the intercom via welding then prying, the keys inside the intercom get deleted. This issue is not replicated at master intercoms which do not have security paneling. 

**Reproduction**

1. Go to a secured command intercom
2. open the pannel
3. de-weld the security plate
4. Pry the plate off
5. POOF no more keys
NOTE: the pry action to remove the keys was never taken (because the keys were deleted before it was possible).

**Screenshots**

**Additional context**

This was noticed on Core station on Lizard in the bridge (both command intercoms, left and right side). I have not checked to see if the security intercom has the same issue.

### LINKED ISSUE #34054: Opening secure intercoms maintenance panel (such as an intercom with command comms), and then unwelding the security plate, removes all encryption keys inside (disappear)
## Description

About the title. Any secure intercoms upon unwelding plate will make their keys dissapear

**Reproduction**

Screwdriver to intercom, notice the secure panel, attempt to unweld the secure panel as the examination prompt indicates,, the secure plate is unwelded successfully but the encryption keys disappear and you can no longer get them.

unsure about normal intercoms

**Screenshots**

**Additional context**

Booblesnoot42:
## About the PR
Fixed an issue wherein using a welding tool on any intercom which inherited from BaseIntercomSecure, which should remove the security plate and allow access to the encryption keys, would delete all the encryption keys within it.

## Why / Balance
Fixes #32590
Fixes #34054
(Duplicate Issues)

## Technical details
Change: 
Changes the construction system to check abstract parents when checking if an entity inherits from a construction node's desired prototype. 

Explanation:
The problem is in how the "doNotReplaceInheritingEntities" code works. When an entity changes nodes on a construction graph, it is supposed to change to match the new node's prototype unless it is already that type. If "doNotReplaceInheritingEntities" is set to true, however, than it only needs to change if it doesn't inherit from the node's prototype rather than needing to match the type exactly. The function currently used to find the entity's parents doesn't check abstract parents or their grandparents, however, resulting in the error. 

For example, the CommandIntercom inherits from BaseIntercomSecure (an abstract prototype) which itself inherits from IntercomConstructed. The BaseIntercomSecure prototype sets the construction node to "intercomReinforced." When a player uses the welding tool on the command intercom, it changes from the "intercomReinforced" node to the "intercom" node. The "intercom" node has its prototype set as "IntercomConstructed", and the doNotReplaceInheritingEntities field set to true; as a result, the intercom shouldn't change prototypes because the construction system should recognize that it inherits from IntercomConstructed through BaseIntercomSecure. Because BaseIntercomSecure is abstract, however, the system won't realize that the intercom inherits from IntercomConstructed and will change the CommandIntercom into an IntercomConstructed, deleting the keys in the process.

## Media

https://github.com/user-attachments/assets/4b883a67-a8a6-4feb-bb2a-9ff0f4b69ab1

## Requirements

## Breaking changes

None that I'm aware of.
**Changelog**

:cl:
- fix: Removing the security plate from an intercom will no longer delete its encryption keys.

### DIFF
--- a/Content.Server/Construction/ConstructionSystem.Graph.cs
+++ b/Content.Server/Construction/ConstructionSystem.Graph.cs
@@ -310,9 +310,9 @@ namespace Content.Server.Construction
             if (GetCurrentNode(uid, construction)?.DoNotReplaceInheritingEntities == true &&
                 metaData.EntityPrototype?.ID != null)
             {
-                var parents = PrototypeManager.EnumerateParents<EntityPrototype>(metaData.EntityPrototype.ID)?.ToList();
+                var parents = PrototypeManager.EnumerateAllParents<EntityPrototype>(metaData.EntityPrototype.ID)?.ToList();
 
-                if (parents != null && parents.Any(x => x.ID == newEntity))
+                if (parents != null && parents.Any(x => x.id == newEntity))
                     return null;
             }
 


cohanna:
#33267 seeks to achieve a similar thing (linked for visibility)

Booblesnoot42:
Closed: found a better solution -> #34911

PR #34557, opened 2025-01-21
Fixed: Ore now correctly drops the right amount of ore

beck-thompson:
## About the PR

Ore was broken! The new drop rates are:
```
Steel, Quartz, Coal:                 1-4 -> 1-5  (+20% increase)
Gold, Silver, Salt:                  1-2 -> 1-3  (+33% increase)
Plasma, Uranium, Bananium, Diamond:  1-1 -> 1-2  (+66% increase)
```
## Technical details

`random.Next` has an exclusive max so max was never chosen! Added 1 so the max actually happens. 

**Changelog**

:cl:
- fix: Ore rocks now drop the correct amount of ore when mined.

### DIFF
--- a/Content.Server/Mining/MiningSystem.cs
+++ b/Content.Server/Mining/MiningSystem.cs
@@ -35,7 +35,7 @@ public sealed class MiningSystem : EntitySystem
             return;
 
         var coords = Transform(uid).Coordinates;
-        var toSpawn = _random.Next(proto.MinOreYield, proto.MaxOreYield);
+        var toSpawn = _random.Next(proto.MinOreYield, proto.MaxOreYield+1);
         for (var i = 0; i < toSpawn; i++)
         {
             Spawn(proto.OreEntity, coords.Offset(_random.NextVector2(0.2f)));


ArtisticRoomba:
It was broken????

ArtisticRoomba:
It was broken????

beck-thompson:
> It was broken????

Yeah somehow :laughing:

AgentSmithRadio:
I'm not sure if this brings mining to the power level it needs to be, debris is still the best pick 99% of the time for raw resource generation, but catching this bug is certainly going to help and will be appreciated by Salvage players. Good catch!

beck-thompson:
> I'm not sure if this brings mining to the power level it needs to be, debris is still the best pick 99% of the time for raw resource generation, but catching this bug is certainly going to help and will be appreciated by Salvage players. Good catch!

The only reason I caught this bug was because I was going to buff silver and gold drops! If mining is still to weak for salvage let me know (Obviously let this settle in for a week or two first :laughing:).

AgentSmithRadio:
> The only reason I caught this bug was because I was going to buff silver and gold drops! If mining is still to weak for salvage let me know (Obviously let this settle in for a week or two first 😆).

I'm at work and distracted, but what I want to do is simulate some "average" debris runs for resource output, contrasted against mining output, and get some averages. Mining _should_ be worse than debris because the loot rewards are targeted (you literally select your asteroid's makeup from a list) and it's zero risk (no carp or sharkminnows), but if I'm going to get the same average uranium from a single uranium generator wreck that I am getting from a poor uranium asteroid, something feels wrong about that exchange. 

This is great for now, windmill slam that merge button, but hopefully I can find some time to examine this with some detail and report back.

beck-thompson:
> This is great for now, windmill slam that merge button, but hopefully I can find some time to examine this with some detail and report back.

This would be amazing! Post your results in the development channel and ping me

PR #33385, opened 2024-11-18
Job contraband rework

sporkyz:
## About the PR

Added functionality to the contraband system to allow for restriction based on job, not just department. Added a majority of the unfinished changes in https://github.com/space-wizards/space-station-14/issues/31047

Fixes https://github.com/space-wizards/space-station-14/issues/32432
Fixes https://github.com/space-wizards/space-station-14/issues/33227
## Why / Balance

There were a multitude of places where items were seen by players as being restricted by job, but the in-game systems didn't allow for this (bartender shotguns/bandoliers/armor, lawyers having access to security PDAs, janitor galoshes). I've added the necessary systems and made the contraband changes listed above. Requested in: https://github.com/space-wizards/space-station-14/issues/31047

## Technical details

Added the allowedJobs field to ContrabandComponent, and added the necessary logic for processing it in ContrabandSystem.cs. Modified the yml files for the corresponding items to give them the correct contraband classification.

## Media

image
(Example of double barrel for a non bartender or security member)

image
(Example of a double barrel for a bartender)

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
The `AllowedDepartments` datafield in `ContrabandComponent` is no longer nullable. Use an empty list instead.

**Changelog**

:cl:
- tweak: Galoshes are now restricted to janitor.
- tweak: Double barrel shotguns, basic armor vests, beanbag/flare shotgun shells, and bandoliers are now allowed for Bartenders, not just Security.
- fix: Lawyers now have access to the Security headset they spawn with, as well as the encryption key inside.
- fix: Sawn-off shotguns are now properly considered contraband instead of being unrestricted.

PR #34558, opened 2025-01-21
added missing allowed department to the restricted severity

### LINKED ISSUE #34581: This item is restricted to ,
## Description
Some items (i think guns) count as department-restricted items with no eligible departments for them
#33385 is the culprit
also this is on master, not live yet

**Reproduction**
1. pick gun
2. illegal

**Screenshots**
Image

**Additional context**

Ian321:
## About the PR

The #33385 PR removed the default allowedDepartments member from the ContrabandComponent class causing all entities with it to have a broken description and status.

## Why / Balance

Bugfix; Reverting it to how it was before it PR. Fixes #34581

## Technical details

- Added `allowedDepartments: [ Security ]` to `BaseRestrictedContraband`
- Added a test to check if any entity is affected by this.

## Media
code-change

before

after

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Restricted contraband now gets displayed correctly.

### DIFF
--- a/Content.IntegrationTests/Tests/ContrabandTest.cs
+++ b/Content.IntegrationTests/Tests/ContrabandTest.cs
@@ -1,5 +1,6 @@
 using System.Linq;
 using Content.Shared.Contraband;
+using Content.Shared.Prototypes;
 using Robust.Shared.GameObjects;
 using Robust.Shared.Prototypes;
 
@@ -21,9 +22,7 @@ public sealed class ContrabandTest
             var protos = protoMan.EnumeratePrototypes<EntityPrototype>()
                 .Where(p => !p.Abstract)
                 .Where(p => !pair.IsTestPrototype(p))
-                .Where(p => p.TryGetComponent<ContrabandComponent>(out _, componentFactory))
-                .OrderBy(p => p.ID)
-                .ToList();
+                .Where(p => p.HasComponent<ContrabandComponent>(componentFactory));
 
             foreach (var proto in protos)
             {
--- a/Content.IntegrationTests/Tests/ContrabandTest.cs
+++ b/Content.IntegrationTests/Tests/ContrabandTest.cs
@@ -1,6 +1,4 @@
-using System.Linq;
 using Content.Shared.Contraband;
-using Content.Shared.Prototypes;
 using Robust.Shared.GameObjects;
 using Robust.Shared.Prototypes;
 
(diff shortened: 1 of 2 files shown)


Ian321:
You sure? As far as I can see #33385 has only been merged into master.

Content.IntegrationTests/Tests/ContrabandTest.cs
```diff
@@ -0,0 +1,41 @@
...
+        var client = pair.Client;
+        var protoMan = client.ResolveDependency<IPrototypeManager>();
+        var componentFactory = client.ResolveDependency<IComponentFactory>();
+
+        await client.WaitAssertion(() =>
+        {
+            var protos = protoMan.EnumeratePrototypes<EntityPrototype>()
+                .Where(p => !p.Abstract)
+                .Where(p => !pair.IsTestPrototype(p))
+                .Where(p => p.TryGetComponent<ContrabandComponent>(out _, componentFactory))
```
slarticodefast, line 24:
```suggestion
                .Where(p => p.HasComponent<ContrabandComponent>())
```
No need to get the component since we don't use it.

Content.IntegrationTests/Tests/ContrabandTest.cs
```diff
@@ -0,0 +1,41 @@
...
+                .OrderBy(p => p.ID)
```
slarticodefast, line 25:
```suggestion
```
no need to order them

Content.IntegrationTests/Tests/ContrabandTest.cs
```diff
@@ -0,0 +1,41 @@
...
+                .ToList();
```
slarticodefast, line 26: Instead of creating a giant list of all prototypes using linq just do these checks inside the foreach loop with a `if (...) continue`

slarticodefast (changes requested):

Content.IntegrationTests/Tests/ContrabandTest.cs
slarticodefast: This one is still unresolved.

Content.IntegrationTests/Tests/ContrabandTest.cs
```diff
@@ -0,0 +1,43 @@
...
+            foreach (var proto in protoMan.EnumeratePrototypes<EntityPrototype>())
+            {
+                if (proto.Abstract || pair.IsTestPrototype(proto))
+                    continue;
+
+                if (!proto.TryGetComponent<ContrabandComponent>(out var contraband, componentFactory))
+                    continue;
+
+                if (!protoMan.TryIndex(contraband.Severity, out var severity))
+                    Assert.Fail(@$"{proto.ID} has a ContrabandComponent with a unknown severity.");
```
Tayrtahn, line 29:
```suggestion
                Assert.That(protoMan.TryIndex(contraband.Severity, out var severity), Is.False, @$"{proto.ID} has a ContrabandComponent with a unknown severity.");
```

Content.IntegrationTests/Tests/ContrabandTest.cs
```diff
@@ -0,0 +1,43 @@
...
+
+                if (severity.ShowDepartmentsAndJobs)
+                {
```
Tayrtahn, line 32: Invert this and make it a guard clause (`if (!foo) continue`)

Content.IntegrationTests/Tests/ContrabandTest.cs
```diff
@@ -0,0 +1,43 @@
...
+                    if (contraband.AllowedDepartments.Count + contraband.AllowedJobs.Count == 0)
+                    {
+                        Assert.Fail(@$"{proto.ID} has a ContrabandComponent with ShowDepartmentsAndJobs but no allowed departments or jobs.");
+                    }
```
Tayrtahn, line 36:
```suggestion
                    Assert.That(contraband.AllowedDepartments.Count + contraband.AllowedJobs.Count, Is.Not.EqualTo(0), @$"{proto.ID} has a ContrabandComponent with ShowDepartmentsAndJobs but no allowed departments or jobs.");
```

Tayrtahn (changes requested):
Some minor tweaks.

Content.IntegrationTests/Tests/ContrabandTest.cs
Ian321: You flipped that, if its false its a error.

Tayrtahn: Yep, you're right. Use `Is.True` instead.

slarticodefast (approved):
Looks good to me!

Tayrtahn (approved):

PR #34559, opened 2025-01-21
pluralize the job name in the contra description

Ian321:
## About the PR

Change the jobname in the contraband description to its plural form.

## Why / Balance

Correct grammar.

## Technical details

Please tell me if you know a way to use the `MAKEPLURAL` ftl function directly.

- Added a `generic-plural` localization string and used it on every job in the list.

## Media

|before|after|
|-|-|
|387075525-e9f89b29-d873-41e9-af9d-e755fec08301|Screenshot From 2025-01-21 19-34-13|

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Job specific contraband descriptions are now grammatically sound.

### DIFF
--- a/Content.Shared/Contraband/ContrabandSystem.cs
+++ b/Content.Shared/Contraband/ContrabandSystem.cs
@@ -56,8 +56,8 @@ public sealed class ContrabandSystem : EntitySystem
         using (args.PushGroup(nameof(ContrabandComponent)))
         {
             // TODO shouldn't department prototypes have a localized name instead of just using the ID for this?
-            var localizedDepartments = ent.Comp.AllowedDepartments.Select(p => Loc.GetString($"department-{p.Id}"));
-            var localizedJobs = ent.Comp.AllowedJobs.Select(p => Loc.GetString("generic-plural", ("thing", _proto.Index(p).LocalizedName)));
+            var localizedDepartments = ent.Comp.AllowedDepartments.Select(p => Loc.GetString("contraband-department-plural", ("department", Loc.GetString($"department-{p.Id}"))));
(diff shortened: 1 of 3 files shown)


slarticodefast:
Hmm, I wonder if this may cause problems for some languages that do not pluralize or list the names like this (if there are any), since it is still a little hardcoded in C#. Ideally we would pass two lists of strings into the loc string, one for the departments and one for the jobs, and then let fluent do the rest.
Otherwise the code looks good.

Ian321:
I could maybe make the translation key less generic, so that langues that do not want it pluralized can change it in their ftl.

slarticodefast:
> I could maybe make the translation key less generic, so that languages that do not want it pluralized can change it in their ftl.

Yeah, I think that would be good. Put it into the same ftl file, make one for the departments and one for the jobs.
In english the department ftl string would do nothing, and the job string would pluralize.

slarticodefast (approved):
Yep, that looks more flexible. Thank you for your contribution!

beck-thompson (changes requested):
Looks like there are some glitches
Before:
image

After:
image

I couldn't quite figure out why this was happening specifically (curse you linq!!)

beck-thompson:
> @beck-thompson that's unrelated to this PR, check my other PR https://github.com/space-wizards/space-station-14/pull/34558

OK so that other PR needs to get merged first before this one works?

Ian321:
> > @beck-thompson that's unrelated to this PR, check my other PR #34558
> 
> OK so that other PR needs to get merged first before this one works?

Order does not matter. This PR fixes the grammar, the other the missing department.
They don't even touch the same files, so there won't be a conflict.

slarticodefast:
Yep, tested it again and can confirm that only the security restricted contra is affected. So the other PR will fix that part.

slarticodefast:

PR #34562, opened 2025-01-21
Removes fire rate buffer from equipping the grapple gun

SpaceRox1244:
## About the PR

Removes the 1-2 seconds wait before being able to fire from the grappling gun when you equip it (pick it up, swap to active hand, etc).

## Why / Balance

The grapple gun is not a lethal weapon in need of fire rate control, and it feels better to be able to more quickly launch the hook. Currently, I have to hold the grappling gun out and keep my active hand on it if I even think I might have to use it. This tool has saved me from being flung, but only when I was already holding it.

## Media

None

## Requirements

## Breaking changes

**Changelog**

:cl: SpaceRox
- tweak: The grapple gun can now fire immediately when swapped to.

Feel free to omit the changelog if it is too small of a change, I don't really mind. It's just here so that merging isnt delayed by asking for it.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Launchers/launchers.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Launchers/launchers.yml
@@ -254,6 +254,7 @@
       soundGunshot: /Audio/Weapons/Guns/Gunshots/harpoon.ogg
       fireRate: 0.5
       clumsyProof: true
+      resetOnHandSelected: false
     - type: BasicEntityAmmoProvider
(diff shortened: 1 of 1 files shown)


slarticodefast:
This does make it more responsive, which is good, but I found a bug while testing.
If you pick it up while in combat mode it gets fired immediately now. This probably needs a fix for how the user input is handled.

TheShuEd:
How's the progress? Do you plan to continue working on PR?

SpaceRox1244:
Unfortunately I dont plan on doing anything with this right now. Ill close the PR.

Issue #26071, opened 2024-03-13
You can play two (or more?) wizden servers once


UbaserB:
## Description

Joining a second doesn't close the first and its not kicking you out from joining a second. This seems... abusable. Wouldn't you be able to afk grind with several characters over all 4 servers and get 4x the speed in getting jobs?

**Screenshots**

![image](https://github.com/space-wizards/space-station-14/assets/134914314/f2cb36de-264f-4d35-a665-16fe289eef5c)

metalgearsloth:
> ## Description
> 
> Joining a second doesn't close the first and its not kicking you out from joining a second. This seems... abusable. Wouldn't you be able to afk grind with several characters over all 4 servers and get 4x the speed in getting jobs?

PlayTimeTrackingManager does set operations, not add, so you're just overwriting the same data.

UbaserB:
> PlayTimeTrackingManager does set operations, not add, so you're just overwriting the same data.

well even if it's not mechanically an issue you still shouldn't be able to join several servers at once. Raiders could use this as well.

deltanedas:
i recall this kicking you from old server after joining on second one a while ago

Titian3:
We are currently discussing this in the admin team also, People are using it to antag-roll mid round ghost antags but having multiple open and shopping for a ghost role taking it from people who were playing the round and may have perished.
So a mechanical prevention or time out between changing servers could be really good.

deepdarkdepths:
I only use this to port over characters to a new server with ease. Not having this would make it a very big pain, but its understandable why it would need to be fixed

LordEclipse:
This would be almost a non-issue with ghost lottery

metalgearsloth:
I should also point out it's useful for adminning multiple servers so you don't have to rely on reading the relay.

PR #34563, opened 2025-01-21
Add system to kick people if they connect to multiple servers at once.

PJB3005:
## About the PR
This uses the database notification system to, effectively, send a kick message to every other server whenever player connects.

The system is disabled by default, and admins are exempt.

Fixes https://github.com/space-wizards/space-station-14/issues/26071

## Why / Balance
Being connected to multiple servers at once is used a lot by people antagrolling. Bad. Admins decided they wanted to disallow this.

## Technical details
The logic is implemented in a new manager. There is also a new method for C# code to send database notifications to Postgres. I also made DB notifications easier to deal with thanks to a new helper method, and the ban manager also uses it now.

## Breaking changes
None.

**Changelog**

:cl:
ADMIN:
- add: Introduced an option to automatically disconnect players from their current server if they join another one within the same server community. Admins are always exempt.

### DIFF
--- a/Content.Server/Administration/Managers/BanManager.Notification.cs
+++ b/Content.Server/Administration/Managers/BanManager.Notification.cs
@@ -1,6 +1,4 @@
-using System.Text.Json;
 using System.Text.Json.Serialization;
-using Content.Server.Database;
 
 namespace Content.Server.Administration.Managers;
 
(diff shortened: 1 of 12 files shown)


juliangiebel:
sloggers

VasilisThePikachu (approved):

juliangiebel (approved):
Lgtm 👍 
Though when we eventually get more notification channels we might want to subscribe the notification channel based on whether the feature is enabled or not but that's also out of scope so that should be done in a different PR

Simyon264:
sloggers

PR #34566, opened 2025-01-22
Add PKA and PTK-800 shuttle gun recipes to sec techfab

K-Dynamic:
## About the PR

- Adds sec techfab recipes for the PKA and PTK-800 shuttle gun when unlocked

## Why / Balance

PKA and PTK-800 are arsenal weapons and shouldn't be missing from the sec techfab, even if there's better weapons to print.

## Technical details

yml changes only
- `Resources/Prototypes/Entities/Structures/Machines/lathe.yml` (adds magboots, PKA, PTK-800 to sec lathe dynamic recipes)

## Media
image

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Proto-Kinetic Accelerators and PTK-800 Shuttle Gun recipes are now available to the security techfab when researched by science.


IProduceWidgets:
Probably fine?
My hesitancy is mostly around the PKA, and its terrible on-station balance.

Aisu9:
> > i am not sure security absolutely need noslips outside of spacewalks, you should consider that they are engineer contraband. Granting the security hardsuit itself the ability to magnetize just like magboots does would be smarter and not encourage security to break space laws every days because magboots are better than their standard issued shoes/boots
> 
> Magboots have no contraband markers. They're also used by science and salvage so no one breaks space law with them to begin with.
> 
> I don't know of any hardsuits that have in-built magboots currently, so I can't just add that action. I'm not even sure if you need them for spacewalks unless gravity goes out, and they stack horribly with current hardsuit speed debuffs.
> 
> > (and again you do now want security threatening science for silly research)
> 
> Is this an actual thing or just an anecdote everybody brings up as though it happens every shift or at all?

it was happening often on salamander when cloning was a T3, used to hear on radio "if science don't research cloning were going in, taking science hostage and execute anyone that doesn't follow our orders"

ScarKy0 (changes requested):
I understand PKA and PTK as they are weapons, but Magboots shouldn't be in the fab

FairlySadPanda:
I would have to check again @K-Dynamic but the last time I played Sci actively, magboots were a tech unlock that could be made inside science fabs? So it's just sec coming over and asking for magboots nicely; in general science should actually need to talk to people to provide services, same as cargo.

slarticodefast (approved):
Without the magboots this should be fine.

ScarKy0 (approved):
Shitcode. Never merge. :godo:

PR #34572, opened 2025-01-22
bagel update

Emisse:
completely redid bagel security
rejigged a lot of areas fire alarms/air alarms to be more sane
audited firelocked areas, most should contain an escape route, 1 fire extinguisher, appropriate fire alarms, etc
converted many buttons to signal switches where appropriate/labelled all the unlabelled ones i could find
skub

image
image
image

some candid photos


Emisse:
forgot an mv wire but its throwin some wack errors at me

minthydra:
I like this a lot though will probs take some getting used to lol

PR #34575, opened 2025-01-22
CondiMaster Neo

SG6732:
## About the PR
Re-introducing the Condiment Master Neo, or CondiMaster Neo for short. Effectively a ChemMaster but for exclusively civilian service use, it offers filtration and mixing capabilities as well as the ability to bottle up condiments and other cooking reagents. 

However, due to both the expense of actually giving a useful tool to Service and the potential security risk of having a publicly-available chemical mixer, Nanotrasen designed the CondiMaster Neo to come with some notable differences: It requires 50 watts less power but has a maximum buffer capacity of 300 u, as well as no ability to make pills. Nevertheless, it should prove to be a worthwhile machine to have around the kitchen!

## Why / Balance
- Various forks of SS13 have similar machines, if not outright just reskinned ChemMasters. I figured having some notable differences like the bottle dispensing and limited buffer would ensure it isn't abused game-wise.
- Given the rather hectic nature of cooking, there are dozens of recipes that may require mixing various reagents together, and with limited beakers it can be challenging to keep track of everything. The CondiMaster Neo allows for organized and pure mixing to ensure that people don't accidentally eat food harboring toxins...
- The CondiMaster Neo can also reduce the need for constant trips to Chemistry asking for chemists to help out with their mixes. Simpler requests can be taken up by the civilian machine while Chemistry can focus more on medicines and other pet projects.
- This PR also integrates new C# code w/ the ChemMaster, enabling the ability to turn both the pill and bottle UIs in the output screen off. 

## Media
image
image

## Requirements

**Changelog**

:cl:
- add: Added the CondiMaster Neo, a civilian-grade ChemMaster with reduced power consumption in exchange for limited buffer capacity and the ability to only fill up bottles.


Content.Client/Chemistry/UI/ChemMasterWindow.xaml.cs
```diff
@@ -138,19 +138,30 @@ public void UpdateState(BoundUserInterfaceState state)
...
+            {
+                PillTypeButtonGroup.Visible = false;
+                PillLabelGroup.Visible = false;
+            }
+
+            if (!castState.CanMakeBottles)
+            {
+                BottleLabelGroup.Visible = false;
+            }
+
```
Tayrtahn, line 151: Just disabling the buttons clientside is insufficient, as an altered client could spoof the messages to the server. You'll need to add checks on the server before running the pill/bottle making code.

Tayrtahn (changes requested):

NoElkaTheGod:
Friendly reminder, that vhemmaster doesn't require research, and is, in fact, dirt cheap. You can just ask sci to make you one and they'll probably cooperate.

Either way though, good PR. Going to sci every single round is mildly annoying.

insoPL:
So this is just reskin of the regular ChemMaster, without bottles and pills (that is rarely used eitherway) and storage limited to 300? There is hardly any difference between them.

EmoGarbage404:
Ping to re-open if you ever come back to this

PR #34577, opened 2025-01-22
Fixed Attachbodypart Console Command

Coolsurf6:
## About the PR
Fixed a console command that would state it was functioning as intended but never actually attached a selected limb to an entity.

_Massive thanks for Milon for helping out with this code!_

## Why / Balance
Want to mess with attaching body parts and would be good to have a command working.

## Technical details
First portion of the implementation to connect a limb checked if the root container wasn't null, which in the right circumstance should be the case. This meant that the code to simply check was always running, and never the code to **actually** connect the limp.

## Media
N/A

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- fix: Attachbodypart command now functions as intended.


Coolsurf6:
I may need some help with a better output for the WriteError command, but that's about it.

cfif126:
You sure this is not admin cl?

PR #34580, opened 2025-01-22
Mob collisions

metalgearsloth:
Title.

Currently POC stage but it should work with prediction.

https://github.com/user-attachments/assets/844d04f0-6156-46fa-b393-35984c8c22bc

A later video though I polished it a bit from here and am happy.

https://github.com/user-attachments/assets/e79e7461-b4c5-46d2-9af0-2c5545a252f8

Requires https://github.com/space-wizards/RobustToolbox/pull/5560

**Changelog**
- add: Added mob collisions so you can no longer walk inside people.


AgentePanela:
can you make a cvar?

Rinary1:
What about the backlog, tikreit? Won't that cause problems?

Content.Shared/Movement/Systems/SharedMobCollisionSystem.cs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private void OnCollisionStartup(Entity<MobCollisionComponent> ent, ref ComponentStartup args)
+    {
+        _fixtures.TryCreateFixture(ent.Owner,
+            ent.Comp.Shape,
+            "mob_collision",
+            hard: false,
+            collisionLayer: (int) CollisionGroup.MidImpassable,
+            collisionMask: (int) CollisionGroup.MidImpassable);
```
Tayrtahn, line 29: Should layer/mask be configurable on the component, so flying mobs could ignore ground mobs but still collide with other flying mobs?

keronshb:
Finally we can change riot shields after this is implemented.

ImRenQ:
maints in 2025
maints

Content.Shared/Movement/Systems/SharedMobCollisionSystem.cs
metalgearsloth: I ended up using flammable as it's supposed to be a generic non-hard event subscriber though still not entirely happy with it.

If I ever get around to my planned collision layer refactor I might revisit it or if we have mobs that don't have flammable that also need pushing.

metalgearsloth:
Ready for review

lzk228:
your cl is broken, :cl: symbol is missing

Carolyn3114:
ideally when moving around in help mode, you could have collision that works like how it does in tf2 or something, where standing in your teammates will push you both out of eachother, but you can still walk through eachother

when in harm mode, you should have full collision like what you've shown

Errant-4:
It feels like nothing is happening when I run through a mob moving the opposite  way in a 1-wide corridor, we seemingly just pass through each other with no resistance. Tested by spawning kobolds in a corridor

Content.Shared/Movement/Systems/SharedMoverController.cs
```diff
@@ -105,6 +104,14 @@ public override void Shutdown()
     public override void UpdateAfterSolve(bool prediction, float frameTime)
     {
         base.UpdateAfterSolve(prediction, frameTime);
+
+        var query = AllEntityQuery<InputMoverComponent, PhysicsComponent>();
+
+        while (query.MoveNext(out var uid, out var _, out var physics))
+        {
+            //PhysicsSystem.SetLinearVelocity(uid, Vector2.Zero, body: physics);
```
slarticodefast, line 112: is this supposed to be commented out?

Content.Shared/Movement/Systems/SharedMobCollisionSystem.cs
```diff
@@ -0,0 +1,226 @@
...
+
+    protected void MoveMob(Entity<MobCollisionComponent, TransformComponent> entity, Vector2 direction)
+    {
+        // Wake it so we don't clip into a wall.
+        Physics.WakeBody(entity.Owner);
+        var xform = entity.Comp2;
+
+        // Alternative though needs tweaks to mob movement code.
+        // Physics.ApplyLinearImpulse(entity.Owner, direction);
+        _xformSystem.SetLocalPosition(entity.Owner, xform.LocalPosition + direction);
```
slarticodefast, line 136: Continous movement should be handled using velocities, not by teleporting in small steps.
What tweaks would be needed for mob movement?

metalgearsloth:
> In general, doesn't Box2d support repulsive forces following the inverse square law? This is a very common physics interacation (gravity, coulomb force) and could be used instead of a custom implementation.

The issue is doing it across separate ticks between client and server due to prediction and box2d is not built for networking.

metalgearsloth:
Got some ideas on how to rework mob movement a bit to use velocity better but need to test them still.

KaiserMaus:
Do I understand correctly that the heavy player will now push the lighter player?

SlamBamActionman (approved):
The system is probably not ready to be merged as-is (e.g. ratking feels pretty bad to play with all the minions pushing), but it's definitely cool and would be a nice addition to the game. Left some thoughts in the code.

metalgearsloth:
I have implemented some basic mass diffs but we will need to play around with it.

ksen0morph:
image

RedBookcase:
Not sure if feedback is best left here or not but current implementation feels very clunky, especially in places that are often busy or crammed like the evac shuttle. Being stuck in an area and being physically unable to move from the collision of other players feels really bad. Personally I think the collision radius should either be reduced drastically or it should be kept on hold until something like crawling gets added so players have some sort of way to avoid collisions in those scenarios.

lzk228:
oh yea, are you guys also crawling in crowded mall to avoid collisions and not bump into strangers

deltanedas:
presumably combat mode will be taken into account in the future making that better

PR #34582, opened 2025-01-22
Fixing intercom encryption key dropping

### LINKED ISSUE #34571: Intercom encryption key removal spawns keys in the wall
## Description
When removing encryption keys from an intercom with a crowbar, any keys which don't fit in your hand will be dropped on top of the intercom. Because the intercom is inside a wall, this causes them to act unpredictably. 

**Reproduction**
1. Find an intercom (except ones with a built-in metal plate, see issue #32590).
2. Use a screwdriver on it to open the maintenance panel.
3. Use a crowbar to remove the encryption keys (make sure you won't be able to pick up all the keys).

**Screenshots**

https://github.com/user-attachments/assets/3a127990-914b-4f25-a04f-0bad236daf4d

**Additional context**
The [encryption key system](https://github.com/space-wizards/space-station-14/blob/20acef16ee9f6d0ef247253b65760e772a5f70df/Content.Shared/Radio/EntitySystems/EncryptionKeySystem.cs#L49) uses PickupOrDrop to remove the keys from the intercom. When the user doesn't have space to pickup the keys, this causes them to be dropped on top of the intercom. Having them drop on top of the user instead would solve the problem, but I don't know if there's an easy way to do that.

Booblesnoot42:
## About the PR
When removing encryption keys from an intercom, any that don't fit in the user's hands will be dropped at their feet instead of dropping inside the wall.

## Why / Balance
Resolves #34571

## Technical details
Adds an optional parameter to PickupOrDrop called dropNear which causes the item to move next to the target entity when they are unable to pick up the item.

## Media

https://github.com/user-attachments/assets/1412cca7-5378-427b-8cd6-24daac1ffa8e

## Requirements

## Breaking changes

Adds an optional parameter to PickupOrDrop. 

(Not sure if this breaks stuff?)

**Changelog**

:cl:
- fix: When removing encryption keys from an intercom, excess keys will now be dropped at the remover's feet instead of dropping inside the wall.

### DIFF
--- a/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Pickup.cs
+++ b/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Pickup.cs
@@ -197,12 +197,14 @@ public abstract partial class SharedHandsSystem : EntitySystem
     /// <summary>
     ///     Puts an item into any hand, preferring the active hand, or puts it on the floor.
     /// </summary>
+    /// <param name="dropNear">If true, the item will be dropped near the owner of the hand if possible.</param>
     public void PickupOrDrop(
         EntityUid? uid,
         EntityUid entity,
         bool checkActionBlocker = true,
         bool animateUser = false,
         bool animate = true,
+        bool dropNear = false,
         HandsComponent? handsComp = null,
         ItemComponent? item = null)
     {
@@ -214,6 +216,11 @@ public abstract partial class SharedHandsSystem : EntitySystem
             // TODO make this check upwards for any container, and parent to that.
             // Currently this just checks the direct parent, so items can still teleport through containers.
             ContainerSystem.AttachParentToContainerOrGrid((entity, Transform(entity)));
+
+            if (dropNear && TryComp<TransformComponent>(uid, out var transformComponent))
+            {
+                TransformSystem.PlaceNextTo((entity, Transform(entity)), ((EntityUid)uid, transformComponent));
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


Content.Shared/Hands/EntitySystems/SharedHandsSystem.Pickup.cs
```diff
@@ -214,6 +216,11 @@ public void PickupOrDrop(
             // TODO make this check upwards for any container, and parent to that.
             // Currently this just checks the direct parent, so items can still teleport through containers.
             ContainerSystem.AttachParentToContainerOrGrid((entity, Transform(entity)));
+
+            if (dropNear && TryComp<TransformComponent>(uid, out var transformComponent))
+            {
+                TransformSystem.PlaceNextTo((entity, Transform(entity)), ((EntityUid)uid, transformComponent));
+            }
```
slarticodefast, line 223:
```suggestion
            if (dropNear)
                TransformSystem.PlaceNextTo(entity, uid.Value);
```
No need to do a `TryComp` for the `TransformComponent`, since every entity has it.
Also the `Resolve` inside the `PlaceNextTo` function gets it for you if you don't pass it.
Also wouldn't this fail in the case that `uid` is `null`? You are just casting it from a nullable to an `EntityUid`, but that will only remove the IDE warning which is there for a purpose.

Booblesnoot42: Didn't know they all had transform, neat. 

As for the null issue, TryComp accepts a nullable uid and will return false if it doesn't have a value (if I'm reading it correctly), so it shouldn't be possible to reach the PlaceNextTo with uid being null.

I'll replace the TryComp with a normal null check; I had it there because I thought it would be a good way to a) verify the uid wasn't null, b) check the entity had a transform component, and c) get the transform component if it existed so it didn't need to be found again later.

Also, is using uid.Value generally better than casting when switching to a non-nullable type? A google search gives me some results about easier maintenance if the type gets changed later; are there other benefits I should know?

slarticodefast: `uid.Value` is more readable and clearly states the intent in my opinion. Also the IDE gives you a warning if you are using it where it might be null.
Not sure how the compiler handles these in detail, but it's probably smart enough that there is no difference in performance.

slarticodefast (approved):
Looks good to me!

PR #34584, opened 2025-01-22
Update wizden config to disallow multiple connections to multiple wizden servers

VasilisThePikachu:
## Why / Balance

Admins wanted and had voted this a while ago, the feature is disabled by default. We are enabling it on our servers.

## Technical details

#34563


slarticodefast:
Nik approved
grafik

PR #34587, opened 2025-01-22
Pulled item interaction fix

### LINKED ISSUE #30157: Prying component entities works while pulling
1. Pull entity with PryingComponent
2. use virtual item in your hand like an actual tool
3. (optional) pry captain room with dead tarantula :trollface:

https://github.com/user-attachments/assets/b316d69f-7154-4bc8-ae3c-6336ff83d96a

### LINKED ISSUE #30161: pulling cleanbot can be used as a mop
1. pull cleanbot
2. mop

https://github.com/user-attachments/assets/9364bd31-45fe-43cf-bc64-bf5d5bd63caf

### LINKED ISSUE #30166: A pulled entity can be used as a tool
## Description

When you are pulling a tool/mob, you can use it as if it was a tool in your hands.
For example: 
- if you pull a syringe, you can inject people while holding it's virtualitem in your hands. 
- you can use a pulled defib with one hand
- dragged mobs which have required components can be used as tools, such as tarantula's as crowbars and cleanbots as mops(cleanbots are essentially mops on wheels from the yml perspective)
- spray chems with spray bottles or fire extinguishers by using their virtualitem

**Reproduction**

pull something
click on something in the world

**Additional context**
related issues
#30157
#30161

themias:
## About the PR

This PR fixes an issue where pulled items, represented by a virtual item in hand, could be used in interactions as if they were being held. This PR still allows a virtual item to be used in hand interactions if its blocking (real) item is held in another hand.

Fixes #30157 
Fixes #30166 
Fixes #30161 

## Why / Balance

We want to allow virtual items to be used if the real item they point to is also in your hands. For example, if we are holding a multi-handed item like a defibrillator, or an item wielded in both hands (currently switching hands causes it to un-wield, but maybe there is some future case), we want to be able to use it even if our active hand is the off-hand (holding the virtual item.)

However, if we are merely pulling the item, we don't want to be able to use it in hand interactions, as this allows us to use a dead zombie's strength to pry doors, or use a pulled defib with only one hand, etc.

## Technical details

We raise the GetUsedEntityEvent on the active-hand item. The SharedVirtualItemSystem subscribes and checks if the user is holding the blocking (real) item in some other hand. If so, we allow the blocking item to be used in the interaction. Otherwise, we default to using the active-hand item itself.

## Media

I'm pretty sure right-click wielding from the off-hand isn't intentional, but I just wanted to show it here as POC.

https://github.com/user-attachments/assets/48181ad3-5254-4f21-9013-f0943c468496

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: You can no longer use pulled objects as tools.

### DIFF
--- a/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
+++ b/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
@@ -188,11 +188,13 @@ public abstract partial class SharedHandsSystem : EntitySystem
         if (args.Handled)
             return;
 
-        // TODO: this pattern is super uncommon, but it might be worth changing GetUsedEntityEvent to be recursive.
-        if (TryComp<VirtualItemComponent>(component.ActiveHandEntity, out var virtualItem))
-            args.Used = virtualItem.BlockingEntity;
-        else
-            args.Used = component.ActiveHandEntity;
+        if (component.ActiveHandEntity.HasValue)
+        {
+            // allow for the item to return a different entity, e.g. virtual items
+            RaiseLocalEvent(component.ActiveHandEntity.Value, ref args);
+        }
+
+        args.Used ??= component.ActiveHandEntity;
     }
 
     //TODO: Actually shows all items/clothing/etc.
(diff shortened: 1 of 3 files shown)


Booblesnoot42:
Changelog should end with a period.

metalgearsloth:
Bit weird but looks good.

PR #34594, opened 2025-01-23
Fix: Scrap generators can now be fultoned

beck-thompson:
## About the PR

It makes sense! They are structures and fulton can teleport those kinds of things.

## Technical details

Just added a parent component. There were also a few random components that I forgot to remove that don't do anything so I removed those as well.

## Media

[Screencast from 01-22-2025 03:48:27 PM.webm](https://github.com/user-attachments/assets/8b381371-bff8-47f6-8372-8778d969f569)

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Scrap generators can now be fultoned!

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Materials/scrap.yml
+++ b/Resources/Prototypes/Entities/Objects/Materials/scrap.yml
@@ -446,13 +446,20 @@
 
 - type: entity
   parent: BaseScrapLarge
+  id: BaseScrapGenerator
+  abstract: true
+  components:
+  - type: Anchorable
+  - type: Sprite
+    sprite: Objects/Materials/Scrap/generator.rsi
+
+- type: entity
+  parent: BaseScrapGenerator
   id: ScrapGeneratorPlasmaLeaking
   name: leaking P.A.C.M.A.N. generator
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
Struck down by thy heisentest

slarticodefast:
I can't seem to push them into the recycler, except when pulling from a weird angle to make them glitch in. Was that an issue before or added by making them anchorable?

slarticodefast:
Are there other entities with a collision box that collides with rhe recycler that can be recycled?
Maybe we should have recycling pnly for items and not for structures. The generators could be dismantled instead.

beck-thompson:
> Are there other entities with a collision box that collides with rhe recycler that can be recycled?
Maybe we should have recycling only for items and not for structures. The generators could be dismantled instead.

I like this idea a lot! I will change them to work like this

YoungThugSS14:
I get the vision behind needing the generators to be dismantled before Fultoning them, but I really wish it was done in a separate PR, salvagers need this so bad!

MilonPL:
Is this being worked on?

EmoGarbage404:
reopen if you come back to this

PR #34604, opened 2025-01-23
Return Drozd full-auto and semi-auto firing modes

K-Dynamic:
## About the PR

- Returned Drozd full-auto and single-auto with previous firerate while retaining burst mode
- Rearranged available modes in alphabetical order

## Why / Balance

Drozds became underused as the SMGs are locked to burst mode only, making the SMG less versatile. Re-adding full-auto and semi-auto (in addition to burst) should make the gun more interesting, as players will be able to choose between spraying at lower fire rate or burst-firing with higher fire-rate.

Made the firing modes alphabetical because it annoyed me to see them out of order (not that rearranging them has an effect in-game).

## Technical details

yml changes only
- `Resources/Prototypes/Entities/Objects/Weapons/Guns/SMGs/smgs.yml` (Add semi-auto and full-auto to Drozd)

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: The Drozd SMG can now fire in full-auto, burst or semi-auto

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/SMGs/smgs.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/SMGs/smgs.yml
@@ -106,9 +106,9 @@
     maxAngle: 32
     shotsPerBurst: 5
     availableModes:
-    - SemiAuto
     - Burst
     - FullAuto
+    - SemiAuto
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/c-20r.ogg
   - type: ChamberMagazineAmmoProvider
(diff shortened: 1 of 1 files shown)


Emisse (approved):

KaiserMaus:
We need semi-auto on SMG ?

SlamBamActionman:
As the burst fire PR author:

I think that reverting it on the Drozd is good; the initial choice to make the Drozd burst-only was due to concerns that burst would be too oppressive for long-range peek combat, and therefore putting it on a gun with a big spread would balance things.

Turns out, that just made the gun suboptimal in either scenario! It's too inaccurate for long-range, and you don't get any benefit from the burst behavior in short-range. 

Personally I think that having a gun restricted to burst-only is still desirable since that pushes players to engage with the gun's unique attributes, just that the Drozd was the wrong choice for it. Personally I'd like to see the behavior moved over to the Lecter, which is a proper long-range assault rifle that should be able to actually benefit from the burst, but I recognize letting things lie for a bit might be best.

minthydra:
Personally I play from Australia so the ping is so bad in most cases I straight up cant use burst fire guns, so I think its a good change.

PR #34608, opened 2025-01-23
Remove traitor as fallback

sleepyyapril:
## About the PR

Removed Traitor as a fallback preset.

## Why / Balance

Alright, so, lets say Secret gets selected. It picks Traitor. It fails due to low player count, going for Fallback. Oh no! It selects traitor again!

Repeat. You see the issue. This is a major issue for lower-pop servers and we should not have gamemodes with a minimum player requirement as a fallback regardless.

## Technical details

Changes a CCVar default.

## Requirements

**Changelog**

### DIFF
--- a/Content.Shared/CCVar/CCVars.Game.cs
+++ b/Content.Shared/CCVar/CCVars.Game.cs
@@ -45,7 +45,7 @@ public sealed partial class CCVars
         ///     The preset for the game to fall back to if the selected preset could not be used, and fallback is enabled.
         /// </summary>
         public static readonly CVarDef<string>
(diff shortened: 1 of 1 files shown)


sleepyyapril:
> Reach already has systems in place to prevent traitors from spawning while it is the current map, and the other stations typically only occur when there is enough people to warrant an actual antag.

yet, this issue still occurs!

sleepyyapril:
genuinely the whole antag system needs a massive refactor, but that's outside the scope of this PR

Princess-Cheeseballs:
Maintainers had a discussion about this PR. The issue this is fixing isn't very common at the moment and the fix just masks the issue rather than properly solving it. We'd rather a more permanent solution be found as part of the gamerule selection system so that this CVAR can be depreciated. 

PR #34620, opened 2025-01-24
revenant refactor

MilonPL:
## About the PR

refactors the revenant system
made documentation more robust
made examines predicted, entity<t>, etc etc

added two new revenant abilities:
- Cold Snap (removes energy from the nearby atmos, lowering the temp)
- Energy Drain (drains nearby SMESes, APCs, batteries, etc etc)

ideally merged after https://github.com/space-wizards/space-station-14/pull/34474 so i dont have to deal with the corporeal system bullshit that much

## Why / Balance

modern code good 👍 
more abilities also good, but I have no idea where they land in terms of cost/balance

## Technical details

since abilities were entities anyways they are now all split into separate components for each one, instead of being stored on the revenantcomponent

Revenant abilities (and hopefully soon other abilities) are now handled on two layers:
- RevenantSystem processes the ability events before the AbilitySystem, checking the requirements if an entity with the RevenantComponent performs the action AND the action has the RevenantActionComponent
   - If we don't have enough essence, we display the popup and mark it as handled, so that the AbilitySystem doesn't perform it. 
- AbilitySystem handles the core logic, making these reusable for any other entity

## Requirements

## Breaking changes

A lot of the revenant namespace changes, and forks with custom revenant abilities will need to modify their code and prototypes with the new component references. Actions like Defile, Overload Lights, Malfunction, Cold Snap, and Energy Drain are now generic abilities that can be added to other entities. 

**Changelog**

:cl:
- add: Added two new abilities for revenants!


MilonPL:
i will come back to this eventually trust

PR #34625, opened 2025-01-24
safe embeddable checks

TheShuEd:
## About the PR

Just making code more crash-safe

## Why / Balance

image
i catch random integration tests thats fails in this point, and i see unsafe .! calls. I dont like unsafe !!!


deltanedas:
might as well use is {} shooter to get rid of the .Value funny, same for projectile

TheShuEd:
Merged with #34736

Issue #13475, opened 2023-01-15
Unbuckle via verb does contact interaction but via alert does not


metalgearsloth:
title.

PR #34634, opened 2025-01-25
small buckle cleanup

lzk228:
i noticed that OnUnbuckleAlert sends the entity to TryUnbuckle as a component, so i fixed it so send the component
also cleaned up another call of TryUnbuckle
i suppose it fixes https://github.com/space-wizards/space-station-14/issues/13475

### DIFF
--- a/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
+++ b/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
@@ -92,7 +92,7 @@ public abstract partial class SharedBuckleSystem
     {
         if (args.Handled)
             return;
-        args.Handled = TryUnbuckle(ent, ent, ent);
+        args.Handled = TryUnbuckle(ent, ent, ent.Comp);
     }
 
     #endregion
--- a/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
+++ b/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
@@ -92,7 +92,7 @@ public abstract partial class SharedBuckleSystem
     {
         if (args.Handled)
             return;
-        args.Handled = TryUnbuckle(args.User, args.User, ent.Comp);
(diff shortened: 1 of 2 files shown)


Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -92,7 +92,7 @@ private void OnUnbuckleAlert(Entity<BuckleComponent> ent, ref UnbuckleAlertEvent
     {
         if (args.Handled)
             return;
-        args.Handled = TryUnbuckle(ent, ent, ent);
+        args.Handled = TryUnbuckle(ent, ent, ent.Comp);
```
metalgearsloth, line 95: Pre sure this should be args.user instead.

metalgearsloth: Actually no that won't change anything but uhh this code is suss?

lzk228: it shouldn't

Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -92,7 +92,7 @@ private void OnUnbuckleAlert(Entity<BuckleComponent> ent, ref UnbuckleAlertEvent
     {
         if (args.Handled)
             return;
-        args.Handled = TryUnbuckle(ent, ent, ent);
+        args.Handled = TryUnbuckle(ent, args.User, ent.Comp);
```
slarticodefast, line 95: I did some breakpointing and this seems to be identical to me.
The issue is caused by verbs always doing contact interaction, see SharedVerbSystem:
grafik

You will have to do this manually for the alert.

slarticodefast (changes requested):

slarticodefast:
Are you coming back to this?

lzk228:
well i don't understand what contact interaction is, so i'll leave for smart fellas

Issue #22141, opened 2023-12-05
crit people shouldn't be gasping if wearing a gas mask connected to gas tank


lzk228:

Brandon-Huu:
How do you turn on the gas tank someone else is wearing?

Anzarot121:
> How do you turn on the gas tank someone else is wearing?

context menu

slarticodefast:
See linked PR

PR #34637, opened 2025-01-25
make crit entities breath to connected gas tanks

lzk228:
## About the PR

title

## Why / Balance

dunno, more ways to survive?
to fix https://github.com/space-wizards/space-station-14/issues/22141

## Technical details

just add a check for connected gas tank

## Media

https://github.com/user-attachments/assets/c9da29cf-38c2-4989-8adf-0a0156157b8b

## Requirements

## Breaking changes

no

**Changelog**

:cl:
- tweak: People in crit will not be gasping if they have a connected gas tank to their internals

### DIFF
--- a/Content.Server/Body/Systems/RespiratorSystem.cs
+++ b/Content.Server/Body/Systems/RespiratorSystem.cs
@@ -76,7 +76,7 @@ public sealed class RespiratorSystem : EntitySystem
 
             UpdateSaturation(uid, -(float) respirator.UpdateInterval.TotalSeconds, respirator);
 
-            if (!_mobState.IsIncapacitated(uid)) // cannot breathe in crit.
+            if (!_mobState.IsIncapacitated(uid) || (TryComp<InternalsComponent>(uid, out var internals) && internals.GasTankEntity != null))
             {
                 switch (respirator.Status)
                 {


slarticodefast:
I always wondered why there was no artificial respiration.
Since this has some balance implications I'll open a maintainer discussion thread, but I like it.

SpaceRox1244:
> But I wonder how much people will start powergaming with internals to increase their longevity if they know they're about to go crit, so they have more time for allies to find them and save them (and if that's even really a problem that will turn the tables in a round). Although in the case of Nukie shootouts, civilians already typically start putting on eva suits or breath masks anyway for the inevitable spacings

It's not powergaming to make smart informed decisions based on your knowledge of mechanics. We should discard the notion that relying on game knowledge is a failure of roleplay and that roleplay exists despite mechanics. We will never be able to and should never separate the mechanics from the roleplay.

Djungelskog2:
> It's not powergaming to make smart informed decisions based on your knowledge of mechanics. We should discard the notion that relying on game knowledge is a failure of roleplay and that roleplay exists despite mechanics. We will never be able to and should never separate the mechanics from the roleplay.

I mean it kinda is

SpaceRox1244:
> I mean it kinda is

I'd like to hear more about that so I can discuss it.

Moryachok136:
Wouldn't it regenerate damage from asphyxiation toxins like lexorin? It think shouldn't.

NoElkaTheGod:
> Wouldn't it regenerate damage from asphyxiation toxins like lexorin? I don't think it should.

Don't think this is an issue. If you use lexorin offensively (i.e. lexorin hypodarts), you can just finish the target off while they're in crit. If you want to use it stealthily, just use 11 units or more, it will be more than enough to kill someone, even if they're breathing while in crit.

lzk228:
have you discussed about how it can be added in another way? e.g. as a medical equipment

slarticodefast:
> have you discussed about how it can be added in another way? e.g. as a medical equipment

No, but it could be part of a larger medical rework.

PR #34639, opened 2025-01-25
refactor: simple radial menu for easier creation

Fildrance:
## About the PR

So i cooked a bit on radial menu, to make it a bit easier to create for people who are not really close friends with UI code.
To fix that i tried to create unified SimpleRadialMenu with xaml/xaml.cs and separate models for buttons. This means:
on the positive side:
* no more xaml+xaml.cs if u just want usual radial menu with buttons and layers, its already done once - why duplicate it. You only need BUI to do server interactions.
* layers are described by how buttons are nested inside each other
* button and layer declaration and button interaction is done in c#, and passed as c-tor parameters to SimpleRadialMenu, so its not part of UI code anymore
* all existing code can be easily converted
on the negative side:
* you cannot make your own UI, derived from RadialMenu and declare layers with xaml (only c#) due to layers now using references to each other instead of magic string
* new way is a bit suboptimal memory-wise
* all key properties of radial menu buttons should be duplicated on model, in case people would like to change them... and description of 'key' is vague here
* choosing TYPE of radial menu button (with segment, with only icon, with only borders, etc) would be... kind of a pain?

Also entity tracking (when menu center is following some entity on screen) is implemented inside SimpleRadialMenu.

## Why / Balance
 I detected several pain sources on RadialMenu (mostly for newcomers):
* any amount of radial menu requires BUI (for server interaction) xaml + xaml.cs, which is a lot of files when u want to have 3 options and interactions on click.
* xaml layout is unintuitive when it comes to layers - different 'layers' cannot nest into each other logically, they should stay directly as child of RadialMenu
* binding between RadialMenuContainer and button that redirects to it is done using magic string, which is error-prone, but is KINDA intuitive when you really get into it
* each radial menu HAVE TO do the icon related code and choose radial menu button type by itself, which is not something that contributors who are oblivious to UI code would do successfully. More then that, existing radial UI menus have all different code for that, zero reusability, and it really makes it even harder to 'just copy paste the code u need'
* most of existing uis have button-related code (fill buttons and button click reactions) mixed up with UI realted code a little bit, in a bad way.

I belive my solution can address those problems.

## Technical details

SimpleRadialMenu is a UI element added, models for filling it created:
* RadialMenuOption - base class for menu button **model**
* RadialMenuActionOption - class for button that is doing something on even of clicking and closes menu.
* RadialMenuNestedLayerOption - container for cases where button should lead to other layer, can contain 1 or more buttons (RadialMenuActionOption) or other layer redirections (RadialMenuNestedLayerOption)

Old existing radial menu uis migrated to SimpleRadialMenu.

## Media

well... it works the same way it was, it is refactor after all 

## Requirements

## Breaking changes

**Changelog**


Fildrance:
Tested that old way of declaring radial menu still works too! Even with control names used in TargetLayer property (buttons declared in xaml).

Fildrance:
I think i missed ACTUALLY putting OpenCenteredAt into SimpleRadialMenu, but won't there will always an option to just call Open or other stuff?.. I wouldn't like the idea to override all respective methods (and i'm not sure they are virtual) so i'm not 100% sure what's the best way to handle such situation...

Fildrance:
got some comments from **[here](https://github.com/space-wizards/space-station-14/pull/35152)** - resolved all of them (fixed AI stuff, removed all LINQ from models generation)

metalgearsloth (approved):

PR #34650, opened 2025-01-26
Reduced speed penalty for engineering/atmos hardsuit

Compilatron144:
## About the PR
This PR reduces the speed penalty of the atmospherics and engi hardsuits, and reduces the speed penalty of the firesuit while removing it's barotrauma resistance.

## Why / Balance
Engineers are the most common users of hardsuits (aside salvage) and are likely to be wearing them nearly constantly, particularly during eventful rounds where spacing is constant. Although both hardsuits provide relevant protection against specific hazards, they will only be used for such specific protections in specific scenarios. Given the hardsuits are used primarily for their space protection, the 30% speed penalty is harsh. 

This penalty would make sense if the hardsuit was used explicitly for engine maintenance, and another alternative was available for station repairs, but this alternative does not exist - meaning that going to work sites requires engineers endure the 30% speed penalty always.

This PR reduces the speed penalty to 20%, which is still very noticable, but makes slogging between engineering and breach sites more bearable. This brings it in line with other non-combat hardsuits.

For atmos techs specifically, they had access to the firesuit, which was better than the hardsuit for almost all situations because it had less speed penalty. In line with the speed penalty reduction for the hardsuits, the atmos tech's firesuit has had it's speed penalty reduced from 20% to 10%, but removes it's spacing resistance. It will still protect the wearer from high pressures, which makes it perfect for quickly responding to station fires - but not outright better than the hardsuit.

## Technical details
Changed engineering hardsuit speed modifier from 0.7 to 0.8
Changed atmospheric hardsuit speed modifier from 0.7 to 0.8
Changed atmospheric firesuit speed modifier from 0.8 to 0.9
Changed atmospheric firesuit low pressure multiplier from 1000 to 1

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Reduced the engineering and atmospheric hardsuit speed penalty to 20%
- tweak: Reduced the atmospheric firesuit speed penalty to 10%
- tweak: The atmospheric firesuit no longer provides space protection

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/hardsuits.yml
@@ -58,8 +58,8 @@
         Heat: 0.8
         Radiation: 0.5
   - type: ClothingSpeedModifier
-    walkModifier: 0.7
-    sprintModifier: 0.7
+    walkModifier: 0.8
+    sprintModifier: 0.8
   - type: HeldSpeedModifier
   - type: ToggleableClothing
     clothingPrototype: ClothingHeadHelmetHardsuitAtmos
@@ -90,8 +90,8 @@
         Caustic: 0.5
         Radiation: 0.2
   - type: ClothingSpeedModifier
-    walkModifier: 0.7
-    sprintModifier: 0.7
+    walkModifier: 0.8
+    sprintModifier: 0.8
   - type: HeldSpeedModifier
   - type: ToggleableClothing
     clothingPrototype: ClothingHeadHelmetHardsuitEngineering
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/suits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/suits.yml
@@ -83,7 +83,7 @@
     sprite: Clothing/OuterClothing/Suits/atmos_firesuit.rsi
   - type: PressureProtection
     highPressureMultiplier: 0.02
-    lowPressureMultiplier: 1000
+    lowPressureMultiplier: 1 # Protects from fires, not space
   - type: TemperatureProtection
     heatingCoefficient: 0.001
     coolingCoefficient: 0.05
@@ -96,8 +96,8 @@
         Heat: 0.8
         Cold: 0.8
   - type: ClothingSpeedModifier
-    walkModifier: 0.8
-    sprintModifier: 0.8
+    walkModifier: 0.9
+    sprintModifier: 0.9
   - type: HeldSpeedModifier
   - type: GroupExamine
   - type: ProtectedFromStepTriggers


metalgearsloth:
> are likely to be wearing them nearly constantly

Isn't the whole point to encourage them to not wear it 24/7.

ArtisticRoomba:
I'm a little indifferent on this change. One of the complaints about engineering is that it's a walking simulator, however this is usually from people that are always wearing their suit (which is discouraged).

Personally I can't draw a line between a suit that is somehow fireproof but not pressure-proof. It can somehow resist insane pressures (5000 kPa+!!!) but suddenly it can't hold air to save its life.

SpaceRox1244:
> Isn't the whole point to encourage them to not wear it 24/7.
> 
> Firesuit also has no pressure protection in 13 so not really comparable.

They still wear it 24/7, whether there is a speed penalty or not, because it enables them to respond to spacing issues immediately. Not only do engineers feel expected to remain space-faring to be efficient at their job, they are racing each other to claim repairing gameplay because of how little they otherwise have to do.

metalgearsloth:
> they are racing each other to claim repairing gameplay because of how little they otherwise have to do.

Isn't that more of an indictment that we should finish porting 13 gameplay first before balancing?

SpaceRox1244:
> Isn't that more of an indictment that we should finish porting 13 gameplay first before balancing?

I mean, we've made larger temporary changes for the sake of alleviation until further development, like how we changed spacing's lethality.

Compilatron144:
If you want to stop people wearing a hardsuit 24/7 then make the hardsuit more portable while not worn. Even if hardsuits had a 50% slowdown, you'd find people wearing them anyway, as otherwise you'll need to go back to engi, grab a suit, walk out of engi, go to a breach site - or just be wearing it already and go to the breach site. Might as well make it less of a slog to get there.

ilovehans10:
I swear I had seen a PR or issue suggesting this, but I can't seem to find it now. Someone else had proposed adding an option to fold engineering hardsuits on a doafter to make it so they can fit in bags more easily, and unfold on another doafter to be wearable. Once folded the suit would be 3x3 or 4x4 so they take up a reasonable amount of inventory space, but still provide enough room for other materials.

Another option would be to have an item that when turned on provides a temporary pressure shield to the player, but no other protection. This would mean that the engineer would still have to wear a mask for air, and probably wouldn't want to use this anywhere with active danger, but could use it to replace broken windows. In my mind this item would be built on the same technology as the holofan, and as such would run on a battery, but be bigger than it's holofan counterpart in order to "intelligently move the holofan with its user". In terms of size, I think 2x3 or even 2x2 with an extra block stuck on it to make inventory Tetris more prevalent would be fair. I imagine this looking something like the energy spacesuits shown in the Guardians of the Galaxy movie, but yellow instead of blue to match the coloring of the holofan.

End of the day, I fully support this change, and just wanted to give my two cents on things that could be implemented in the future to allow this change to be reverted.

semensponge:
Personally I disagree with removing Firesuit space protection. Some time as Atmos you need to mess with your space-facing burn chambers and this would prevent them from doing that safely. What I would prefer, if a nerf to the Firesuit is absolutely necessary, would be to remove it's slash damage resistance entirely. Make it only useful for fires.

Alternatively, make the Firesuit speed penalty match the Hardsuit one, and give them more meaningful differences. As you said, a lot of Atmos players (myself included) just default to Firesuit because you move faster. Having them all share the same speed penalty should reinforce their unique strengths. Not sure what buffs it would necesitate but something to make it a meaningful choice.

Regardless, I agree with lowering the speed reduction as a whole. Engineering is a slog to play, and since getting more comfortable with Atmos, I default to that because when the station gets hit with meteors for the 20th time, I don't have to drag myself through the halls like a slug as often. As @SpaceRox1244 said, the Engineers are fighting for something to do and I let them fight amongst themselves while I make funny gas.

In regards to Engineers wearing Hardsuits constantly, how often are you called to fix something that isn't spaced? 9/10 times if a bomb has blown up or meteors hit the station, there's spacing. 24/7 Hardsuits are a necessary evil due to how the gameplay loop progresses.

Nox38:
Barotrauma is extreme pressure, not lack of pressure.

Nox38:
> Probably better off mapping engi hardsuits and engi rooms around the map, such as on Amber

Engi checkpoints??

slarticodefast:
Hey, I brough this up for maintainer discussion and it was decided not to merge this. Hardsuits are already quite powerful and should come with a serious downside, or players would be even more encouraged to wear them at all times. Speed is a stat that is relevant for station traversal and combat. While there is an argument for the former, Engineers tend to be in a position where the emergency they respond to isn't immediate (if a hallway is spaced, it's not gonna matter if they get there 5 seconds earlier), and when they move to areas they tend to spend relatively long time there stationary.
As such I will be closing this PR, thank you for your contribution.

deltanedas:
wizden doesnt have space damage so you can just use a winter coat and spam rods/tiles to fill a hole

Mixelz:
> wizden doesnt have space damage so you can just use a winter coat and spam rods/tiles to fill a hole

Dunno if this is something we should want to have happen since it seems like a pretty reductive way of playing the game compared to an "intended" form of play with regards to doing repairs.

ArtisticRoomba:
Yes, it's not intended. Spacing damage is nerfed because firelocks keep getting trolled by monstermos, among other better reasons.

PR #34661, opened 2025-01-26
Adding the ability and implementing using job pools for trainee and regular jobs

KickTheWhale:
## About the PR
Non destructively changed the station YAMLs and added shared job pools logic, so the trainee and regular jobs share the same job pool
Backwards compatible with the old styling.

## Why / Balance
There were a bit of a balancing issue between starter and advanced servers in terms of map design. Example: Grasshopper could enjoy massive security crews, while Salamander almost never had any. 
This was due us having a starter jobs separate from regular (they are exclusive based on the time played), so when server population has some time played, almost never there are any starter job slots taken.
This addresses the issue and uses a common job pool for all starter/regular jobs that are exclusive
Now mappers can just figure out amount of regular workers for a map and forget about it. 

## Technical details

Added a new field in Station Jobs named jobGroups:
YAML example
```
- type: StationJobs
   availableJobs:
...
      #engineering
      ChiefEngineer: [ 1, 1 ]
      AtmosphericTechnician: [ 3, 3 ]
      StationEngineer: [ 0, 0 ]
      TechnicalAssistant: [ 0, 0 ]
    jobGroups:
      EngineeringTeam:
        roundStart: 7
        midRound: 7
        members:
          - StationEngineer
          - TechnicalAssistant
```
Made new passes in job system to adjust for jobgroups them being
In OnStationInitialized to handle new datafield and add job slots in jobgroups
In TryAdjustJobSlot to handle jobs being taken from a group correctly
In TryGetJobSlot to handle jobs being displayed properly

Tested by loading new map, checking that job slots are correctly showing start/midround values
Respawning a bunch of times until the slots are exhausted
Observed 0 slots

Also tested loading the old-style prototype, all been handled correctly (like it did before)

## Media
image
image

## Requirements

## Breaking changes
Should not break anything
One edge case would be using the new YAML JobGroups implementation and the old one with values different from [0, 0] for the grouped jobs.
One should check if its present before adding new jobslots
I have added comments in YAMLs to minimize the chance of this happening

**Changelog**

### DIFF
--- a/Content.Server/Station/Components/StationJobsComponent.cs
+++ b/Content.Server/Station/Components/StationJobsComponent.cs
@@ -79,4 +79,29 @@ public sealed partial class StationJobsComponent : Component
     /// </summary>
     [DataField("availableJobs", required: true)]
     public Dictionary<ProtoId<JobPrototype>, int[]> SetupAvailableJobs = default!;
+
+    /// <summary>
+    /// Dictionary contatining jobs that are in a shared job pools, containing the usual roundstart, midround or negative value.
+    /// JobGroup class described separately
+    /// </summary>
+    [DataField("jobGroups")]
+    public Dictionary<string, JobGroup> JobGroups = new();
+
+}
+
+/// <summary>
+/// Containing list of jobs that are part of shared job pool
+/// Also containing numbers for the usual roundstart/midround amount logic
+/// </summary>
+[DataDefinition]
+public sealed partial class JobGroup
+{
+    [DataField("roundStart")]
+    public int RoundStart;
+
+    [DataField("midRound")]
+    public int MidRound;
+
+    [DataField("members")]
+    public List<ProtoId<JobPrototype>> Members = new();
 }
--- a/Content.Server/Station/Systems/StationJobsSystem.cs
+++ b/Content.Server/Station/Systems/StationJobsSystem.cs
@@ -75,6 +75,18 @@ public sealed partial class StationJobsSystem : EntitySystem
             x => x.Key,
             x=> (int?)(x.Value[1] < 0 ? null : x.Value[1]));
 
+        /// JobGroup logic
+        foreach (var (_, group) in stationJobs.JobGroups)
+        {
+            foreach (var member in group.Members)
+            {
+                if (stationJobs.JobList.TryGetValue(member, out _))
+                {
+                    stationJobs.JobList[member] = group.MidRound;
+                }
+            }
+        }
+
         stationJobs.TotalJobs = stationJobs.JobList.Values.Select(x => x ?? 0).Sum();
 
         UpdateJobsAvailable();
@@ -160,6 +172,29 @@ public sealed partial class StationJobsSystem : EntitySystem
 
         if (amount == 0)
             return true;
+
+        //Adjustment logic for jobs that are in JobGroups
+        foreach (var (_, group) in stationJobs.JobGroups)
+        {
+            if (group.Members.Contains(jobPrototypeId))
+                {
+                    var newMidRound = group.MidRound + amount;
+                
+                    if (newMidRound < 0 && !clamp)
+                        return false;
+
+                    group.MidRound = Math.Max(newMidRound, 0);
+                
+                    // Update all group members
+                    foreach (var member in group.Members)
+                    {
+                        stationJobs.JobList[member] = group.MidRound;
+                    }
+                
+                    UpdateJobsAvailable();
+                    return true;
+                }
+        }
 
         switch (jobList.TryGetValue(jobPrototypeId, out var available))
         {
@@ -343,6 +378,16 @@ public sealed partial class StationJobsSystem : EntitySystem
     {
         if (!Resolve(station, ref stationJobs))
             throw new ArgumentException("Tried to use a non-station entity as a station!", nameof(station));
+
+        /// Logic to update slots for jobs in JobGroups
+        foreach (var (_, group) in stationJobs.JobGroups)
+        {
+            if (group.Members.Contains(jobPrototypeId))
+            {
+                slots = group.MidRound;
+                return true;
+            }
+        }
 
         return stationJobs.JobList.TryGetValue(jobPrototypeId, out slots);
     }
--- a/Content.Server/Station/Systems/StationJobsSystem.cs
+++ b/Content.Server/Station/Systems/StationJobsSystem.cs
@@ -82,7 +82,7 @@ public sealed partial class StationJobsSystem : EntitySystem
             {
                 if (stationJobs.JobList.TryGetValue(member, out _))
                 {
-                    stationJobs.JobList[member] = group.MidRound;
+                    stationJobs.JobList[member] = group.RoundStart;
                 }
             }
         }
--- a/Resources/Prototypes/Maps/amber.yml
+++ b/Resources/Prototypes/Maps/amber.yml
@@ -31,25 +31,25 @@
             #engineering
             ChiefEngineer: [ 1, 1 ]
             AtmosphericTechnician: [ 2, 2 ]
-            StationEngineer: [ 2, 4 ]
-            TechnicalAssistant: [ 1, 1 ]
+            StationEngineer: [ 0, 0 ] # Handled in its own jobgroup
+            TechnicalAssistant: [ 0, 0 ] # Handled in its own jobgroup
             #medical
             ChiefMedicalOfficer: [ 1, 1 ]
             Chemist: [ 2, 2 ]
-            MedicalDoctor: [ 2, 4 ]
+            MedicalDoctor: [ 0, 0 ] # Handled in its own jobgroup
             Paramedic: [ 2, 2 ]
-            MedicalIntern: [ 1, 2 ]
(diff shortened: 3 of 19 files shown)


KickTheWhale:
If its too much adjusting all current map prototypes to use the system, I can dissect just the code changes, as its backwards compatible and old style jobs will be handled like they were before

BramvanZijp:
What if alternatively, they remained seperate, but if the total amount of job slots for the regular job (IE Security Officer) runs out, you can still join as SO and it will subtract 1 open Cadet slot.

The main issue with this system otherwise would be that you could end up with 5 people with 0 clue what they are doing, and only 1 (plus the dpt head) with prior experience.

KickTheWhale:
> What if alternatively, they remained seperate, but if the total amount of job slots for the regular job (IE Security Officer) runs out, you can still join as SO and it will subtract 1 open Cadet slot.
> 
> The main issue with this system otherwise would be that you could end up with 5 people with 0 clue what they are doing, and only 1 (plus the dpt head) with prior experience.

I mean fair, so just we limit bottom line trough the cadets?

KickTheWhale:
On some maps there can be like 6 cadets and 0 officers, or 12 cadets yes. Weird to me aswell
But its a step towards global limits as in, as in some YAMLs have "I want 9-12 security"
I can add some sort of a barrier to jobpool so it can get a certain job as a limit above overall amount, but it feels weird, if say, cadets will be occupied instantly, so you have 4, then 1 officer joins and its 5, then one more and its 6

KickTheWhale:
It will still need some sort of relationship of proper/trainee described and I def not touching job prototypes
Maybe as in job gradation in yaml? :sob:

KickTheWhale:
Just tried to do that, its gonna mean hiding the real values behind some structure so roundstart works uhuhh
Will think of a better implementation if this one is not good

Centronias:
If all maps use the same groups, then it's probably more correct to define the groups as prototypes and then reference the prototypes in the map defs.

Content.Server/Station/Systems/StationJobsSystem.cs
```diff
@@ -75,6 +75,18 @@ private void OnStationInitialized(StationInitializedEvent msg)
             x => x.Key,
             x=> (int?)(x.Value[1] < 0 ? null : x.Value[1]));
 
+        /// JobGroup logic
+        foreach (var (_, group) in stationJobs.JobGroups)
```
Centronias, line 79:
```suggestion
        foreach (var group in stationJobs.JobGroups.Values)
```

Content.Server/Station/Systems/StationJobsSystem.cs
```diff
@@ -75,6 +75,18 @@ private void OnStationInitialized(StationInitializedEvent msg)
...
+        {
+            foreach (var member in group.Members)
+            {
+                if (stationJobs.JobList.TryGetValue(member, out _))
+                {
+                    stationJobs.JobList[member] = group.RoundStart;
+                }
+            }
```
Centronias, line 87:
```suggestion
            foreach (var member in group.Members.Where(member => stationJobs.JobList.ContainsKey(member)))
            {
                stationJobs.JobList[member] = group.RoundStart;
            }
```
Definitely use `Contains` instead of a `TryGetValue` which throws away the `out` result.

Consider using a LINQ expression instead of a nested `if` and `foreach`.

Content.Server/Station/Systems/StationJobsSystem.cs
```diff
@@ -161,6 +173,29 @@ public bool TryAdjustJobSlot(EntityUid station,
         if (amount == 0)
             return true;
 
+        //Adjustment logic for jobs that are in JobGroups
+        foreach (var (_, group) in stationJobs.JobGroups)
```
Centronias, line 177:
```suggestion
        foreach (var group in stationJobs.JobGroups.Values)
```

KickTheWhale:
Oh lots of feedback thanks! Will go trough it tomorrow!

Emisse:
awesome work

Centronias:
I thought about it some more, and while I think it's kinda bleh that it's gonna be the same pools for _every_ station, I think it's better to have the option for flexibility in the pool defs per map rather than being forced to use the same pool prototypes for everything.
If this proves to be the wrong decision, it'll be easy to swap to pool proto definitions anyway.

Content.Server/Station/Systems/StationJobsSystem.cs
KickTheWhale: Does the same thing, less readable
After checking the other job code, thats like super hard to contribute, wanna make it look as simple as possible

Content.Server/Station/Systems/StationJobsSystem.cs
```diff
@@ -75,6 +75,18 @@ private void OnStationInitialized(StationInitializedEvent msg)
...
 
+        /// JobGroup logic
+        foreach (var (_, group) in stationJobs.JobGroups)
+        {
+            foreach (var member in group.Members)
+            {
+                if (stationJobs.JobList.TryGetValue(member, out _))
+                {
+                    stationJobs.JobList[member] = group.RoundStart;
+                }
```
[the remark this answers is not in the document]
KickTheWhale: Yeah, if its not declared in previous jobs, will be ignored. Good side is it makes it so you have to declare it (in case we make a massive group prototype, and I have placed a comment everywhere so it should be ok

KickTheWhale:
> I thought about it some more, and while I think it's kinda bleh that it's gonna be the same pools for every station, I think it's better to have the option for flexibility in the pool defs per map rather than being forced to use the same pool prototypes for everything.
If this proves to be the wrong decision, it'll be easy to swap to pool proto definitions anyway.

Yeah for now it can be easy to adjust, if we see noone ever uses it in any other way, we can shrink the yaml somehow, but I dont think we added much bloat

KickTheWhale (commented):
So far looking like ready to get merged, not sure if I want to adjust anything, updated to have conflicts with upstream

Nox38:
I, Joe Meta, have fixed the job pools thingy for meta

Content.Server/Station/Components/StationJobsComponent.cs
```diff
@@ -79,4 +79,29 @@ public sealed partial class StationJobsComponent : Component
...
+{
+    [DataField("roundStart")]
+    public int RoundStart;
+
+    [DataField("midRound")]
+    public int MidRound;
+
+    [DataField("members")]
+    public List<ProtoId<JobPrototype>> Members = new();
+}
```
Centronias, line 107:
```suggestion
}

```
My inner pedant wants a trailing newline like the rest of the files.

Centronias (approved):
Code looks fine.

IMO it's confusing to need the empty `StationEngineer: [ 0, 0 ]` entries for jobs to exist at all, even if they're defined in a group, but I'll leave that for somebody who actually handles map defs to decide 🤷

KickTheWhale:
@Emisse summoning you to review new code, basically no changes

crazybrain23:
> Jesus fucking christ I fucked up a few merges and summoned everyone it seems :D

You have alerted the codeowners

KickTheWhale:
Ok holy f---. I think this is correct now. Not sure what issue was and why it summoned basically all code owners.
Forcepushed last "normal" version and merged upstream in

Do we have a tag for "cursed" PRs? :sob:

KickTheWhale:
Also can someone fix labels to not show the unrelevant changes? I think my first commit somehow put all 138 commits that I were behind as MY CHANGES so I summoned everyone

KickTheWhale:
@Emisse hi! Summoning you for vibe check and to make sure we have enough signs off happening for this change to ever be looked upon.

BramvanZijp:
@KickTheWhale Please try to resolve that merge conflict, also, this is a substantial enough change to where you will be expected to provide a changelog entry, which I assume you forgot to un-comment out, as in the standard PR format it starts that way.

That being said, while I personally think this inheritance should be one-way (meaning if all regular slots are full, you can join as regular but it subtracts 1 trainee slot, since that way you cant have 10 cadets and 1 officer), I do think this is a change that warrants raised priority, especially given people are locked out of trainee roles if they play the dpt enough.

DISCLAIMER: I am not a maintainer, I am a triager, I cannot merge/approve/officially review this PR, or take any official actions beyond adding/changing triage labels, though I can chip in with my feedback and suggestions just like anyone else can.

KickTheWhale:
> That being said, while I personally think this inheritance should be one-way (meaning if all regular slots are full, you can join as regular but it subtracts 1 trainee slot

Oh yea its a good idea actually. Can change the yaml to just have inheritance order and pooled jobs, instead of describing mixed jobs maximum amount and having same amount for all of them

> which I assume you forgot to un-comment out
Yes D:

Have not paid attention for a while and accumulated conflicts, will fix those too, probably after doing the inheritance thing, should be easier for map maintainers to understand and will save me work keeping it conflict-free

KickTheWhale:
Going to draft it for now, til I implement the ordering of same pools

slarticodefast:
Is this still being worked on?

slarticodefast:
Closing as derelict. Leave a comment when you come back to this and I can reopen the PR.

Issue #34659, opened 2025-01-26
APC light switches are broken


Booblesnoot42:
## Description
APC net switches are supposed to control all the light switches that are connected to the same APC as they are. They don't.

**Reproduction**
1. Make an APC net switch near an APC that powers lights.
2. Use the switch.
3. Observe the lack of darkness.

**Additional context**
As part of pr #13645, [light's network ids](https://github.com/space-wizards/space-station-14/blob/2214857809c7eaae5ea0d703875efd6330696a41/Content.Server/Light/EntitySystems/PoweredLightSystem.cs#L375) were changed from "Apc" to "Wireless," while APC net switches stayed on "Apc". Because they are on different networks, they can't communicate. There are two quick solutions and one harder one that I can think of; unfortunately, the easy ones are either Incredibly Bad™ or will probably break something.

Incredibly Bad™: Just change the APC switch network id to "Wireless" 😃 
Why bad: This will cause the switch to affect, as far as I can tell, _almost every light on the station._ There might be a way to set a max range with a wireless network component, though.

Probably break something: Change lights back to "Apc" and remove their wireless network component.
Why bad: This prevents lights from receiving wireless signals. I don't know if anything currently does that, but it would mean players couldn't control them using signal switches or remote signalers.

Hard, I think: Allow devices to receive from multiple network id's.

In any case, I think the APC net switch should be removed from the build menu until it gets fixed since it currently does nothing at all.

PR #34664, opened 2025-01-26
hide light-switches from build menu

Booblesnoot42:
## About the PR
Hides apc net switches (lightswitches) from the build menu.

## Why / Balance
Until issue #34659 is resolved, they should be hidden so people don't build them when they don't do anything.

## Technical details

Uses the "Hide" field to hide them from the construction menu.

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Due to a lack of functionality, light-switches have temporarily been removed from the construction menu.

### DIFF
--- a/Resources/Prototypes/Recipes/Construction/machines.yml
+++ b/Resources/Prototypes/Recipes/Construction/machines.yml
@@ -60,6 +60,7 @@
   canBuildInImpassable: true
   conditions:
     - !type:WallmountCondition
+  hide: true #TODO: Fix the lightswitch, issue #34659. Until then, keep hidden so people don't build it and get confused.
 
 - type: construction
   name: signal switch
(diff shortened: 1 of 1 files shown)


lzk228:
instead of removing something, fix it

Booblesnoot42:
> instead of removing something, fix it

1. Busy.
2. If it were something where part of it's functionality was broken but it was still useful, like how you can't get encryption keys out of secured intercoms but the intercom itself still works, then I'd agree it's better to leave it until the fix is made. In this case, though, the button literally does nothing. Temporarily hiding it until it's fixed so it doesn't confuse people who want to make a light switch ~so their satanic chapel can have proper mood lighting~ seems like the better option to me.

juliangiebel (approved):
Apc based networking will probably never work so you might as well remove the switches entirely but maybe someone will change the to use something else.
You could probably make them use a device list or smth, like air alarms.
For now they shouldn't be constructible though.

MilonPL (approved):

PR #34671, opened 2025-01-27
Store Refund - Add more disable scenerios & time to disable refund.

keronshb:
## About the PR

Added more routes for us to disable refunds after some entity events and so a time can be provided to disable events instead of just relying on the grid.

## Why / Balance

Just relying on grid changes is not great and is really mostly for the wizard.
As an extra security measure we also now check if the allotted time has passed before also disabling refunds & checking grid.

## Technical details

Method was made to check if we should disable refunds.
Subscriptions to events for using an item, using an action, and shooting a gun were added so we can disable refunds.

## Requirements

## Breaking changes

Removed action check from refund disable method, so we can disable refund on action use.
Refund disable method now checks grid & if the time is eligible.

**Changelog**


Content.Server/Store/StoreRefundComponent.cs
```diff
@@ -1,13 +1,33 @@
 ﻿using Content.Server.Store.Systems;
+using Robust.Shared.GameStates;
```
ScarKy0, line 3: GameStates is not needed.

Content.Server/Store/StoreRefundComponent.cs
```diff
@@ -1,13 +1,33 @@
...
 
 namespace Content.Server.Store.Components;
```
ScarKy0, line 4: Would there be any issue with just putting this component in shared?

ScarKy0 (changes requested):
Tested and works, just minor stuff.

ScarKy0 (approved):
Looks good.

Content.Server/Store/StoreRefundComponent.cs
keronshb: Accesstype moment

keronshb:
1 maint approval + 1 PM making PR = merg :trollface:

PR #34674, opened 2025-01-27
add a chem dispenser to the nukie planet

ArtisticRoomba:
## About the PR
Adds a filled chem dispenser to the nukie planet.

## Why / Balance
As of right now, it's a universal meme that the nukie agent usually isn't up to the task and doesn't know how to make chems.

This usually isn't due to the nukie agent not *knowing* how to make chems, this is due to the player being forced into an unfamiliar environment. Players who learned chemistry through heavy interaction with the chem dispenser would suddenly throw their hands up in the air and effectively not know what to do, as these players have likely never jugmixed before, or used the chemmaster solely as a mixer.

This creates a skill gap:
- Newkie agents who don't know how to jugmix or aren't familiar with mixing chems using the chemmaster get their entire team killed because they don't know how to be an effective chem medic.
- Veteran nukie players can just jugmix their hyperoptimized agent chems and have enough chems for a lovely party.

By having a chem dispenser on the planet, agents have a sense of familiarity and can use the knowledge they're expected to have, instead of being thrown into an unfamiliar environment and being forced to use the chemmaster.

It's important to note that this does affect balance: nukies technically will get one more jug of each base reagent. However I don't think it's a super big deal as nukies usually face a carrying capacity problem, not a lack of reagents problem.

## Media
Content Client_oJd2qnJmUw

## Requirements

**Changelog**
No CL not really any fun

### DIFF
--- a/Resources/Maps/Nonstations/nukieplanet.yml
+++ b/Resources/Maps/Nonstations/nukieplanet.yml
@@ -1630,6 +1630,7 @@ entities:
     - type: MetaData
     - type: Transform
     - type: Map
+      mapPaused: True
     - type: PhysicsMap
     - type: GridTree
     - type: MovedGrids
@@ -1661,6 +1662,15 @@ entities:
         - 0
         - 0
     - type: LoadedMap
+- proto: ActionStethoscope
+  entities:
+  - uid: 1739
+    components:
+    - type: Transform
+      parent: 1336
+    - type: EntityTargetAction
+      originalIconColor: '#FFFFFFFF'
+      container: 1336
 - proto: AirlockExternalGlassNukeopLocked
   entities:
   - uid: 49
@@ -6576,12 +6586,12 @@ entities:
   - uid: 2031
     components:
     - type: Transform
-      pos: 4.0109396,-12.223828
+      pos: 3.3130646,-11.479682
       parent: 104
   - uid: 2095
     components:
     - type: Transform
-      pos: 4.017678,-12.520861
+      pos: 3.6099396,-11.479682
       parent: 104
 - proto: Bed
   entities:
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
> Do they still need a chemvend if they start with chemdispenser?

No, they still do. Nukies would then only be stuck with one carbon jug which isn't a lot of adv. brute meds and would affect balance significantly. That chemvend also has welding fuel in it.

Aisu9:
i'm pretty sure they don't need ALL THESE extra jugs, they have an uplink for bonus medicine. what is the next step giving them a free double e-sword on spawn?

southbridge-fur:
> i'm pretty sure they don't need ALL THESE extra jugs, they have an uplink for bonus medicine. what is the next step giving them a free double e-sword on spawn?

As was mentioned in the description, the issue most Agents have is more of a carrying capacity one rather than a supply one. I've played Agent a fair amount and have never run out of any reagents I needed. I usually leave the planet with four or five entire jugs stuffed into my bag, all filled with medicine.

ArtisticRoomba:
> i'm pretty sure they don't need ALL THESE extra jugs, they have an uplink for bonus medicine. what is the next step giving them a free double e-sword on spawn?

Missing the point of the PR.

BramvanZijp:
> i'm pretty sure they don't need ALL THESE extra jugs, they have an uplink for bonus medicine. what is the next step giving them a free double e-sword on spawn?

If you as an agent are relying on uplink meds outside of blitzops, you are a bad agent. As mentioned, the limiting factor for agents most is bag space, not lack of reagents. And thats a good thing, we shouldnt arbitrarily be restricting what meds agents can make by for example only giving them 1-2 jugs of carbon

ArtisticRoomba:
We can still balance it in order to force nukies to weigh options (chems or lube) however this really isn't necessary as of right now.

Issue #29377, opened 2024-06-23
Cryogenic Gasses Aren't Harmful To Brrathe


Admiral-Obvious-001:
## Description

Simply, breathing a gas that's been frozen to extremely low temperatures don't do any harm to anyone. This is somewhat being used in atmos to cheese the longevity of air tanks and canisters to the point that they can run a single air tank for hours on end. Far more than any reasonable shift will be run for.

**Reproduction**

Setup a freezing loop for your breathable gas of choice, optionally also including spaced radiators to get the gas temperature as close to 0k as possible. Add a gas connector port to the loop. Apply storage canister. Add your air tank to the canister. Fill the tank, and have enough mols to last literal days with no consequence for breathing liquid gas.

**Additional context**

Wasn't sure if this should be made as a feature request but it definitely feels like a gameplay exploit/issue.

WarMechanic:
thisll probably see some change either when newmed hits or sometime after because newmed changes how breathing is simulated

deltanedas:
when breathing cold gases is changed to kill you there should be a rare salv item and/or a tech to decompress liquid oxygen for breathing so you cant just make it in 2 minutes from roundstart

thebadman4662:
Cryo pods use same gas to let patient breathe and freeze them, so it might be neccesary evil till they are changed.

slarticodefast:
![Screenshot (177)](https://github.com/space-wizards/space-station-14/assets/161409025/d71a10a2-9298-4e54-a31b-5a99db91833a)

Similar thing happening with extremely hot gases, no damage taken from breathing them. In this case the player cannot breath at all because there are not enough moles of gas in the tank (even though is has full pressure).

Errant-4:
> Similar thing happening with extremely hot gases, no damage taken from breathing them. In this case the player cannot breath at all because there are not enough moles of gas in the tank (even though is has full pressure).

I think it's not the tank, it's the lungs. Lungs fill with the same volume of hot gas, which will contain less moles than the game system needs since gas expands when hot. Thus, airloss will increase despite the lung being full of air

thebadman4662:
> > Similar thing happening with extremely hot gases, no damage taken from breathing them. In this case the player cannot breath at all because there are not enough moles of gas in the tank (even though is has full pressure).
> 
[...]

PR #34682, opened 2025-01-27
make mobs take damage from breathing in hot or cold gases

MilonPL:
## About the PR

This PR implements temperature damage when breathing extremely hot or cold gases. It calculates the damage based on the energy transfer between inhaled air and body temperature. Applies to gases below -20C or above 60C, with damage scaling up to 5 times based on the thermal energy difference.

## Why / Balance

closes https://github.com/space-wizards/space-station-14/issues/29377

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Supercooled or superheated gases are now harmful to breathe in.

### DIFF
--- a/Content.Server/Body/Systems/RespiratorSystem.cs
+++ b/Content.Server/Body/Systems/RespiratorSystem.cs
@@ -114,7 +114,7 @@ public sealed class RespiratorSystem : EntitySystem
         }
     }
 
-    public void Inhale(Entity<RespiratorComponent?, BodyComponent?> ent)
+    public void Inhale(Entity<RespiratorComponent?, BodyComponent?, ThermalRegulatorComponent?> ent)
     {
         if (!Resolve(ent, ref ent.Comp1, ref ent.Comp2, logMissing: false))
             return;
@@ -140,8 +140,10 @@ public sealed class RespiratorSystem : EntitySystem
         if (actualGas.TotalMoles > Atmospherics.GasMinMoles)
         {
             var gasTemp = actualGas.Temperature;
-            const float normalBodyTempK = Atmospherics.T0C + Atmospherics.NormalBodyTemperature;
 
+            // Take the normal body temperature from the mob's ThermalRegulatorComponent
+            // We don't resolve this component because it shouldn't be required, so we fall back to the atmos consts
+            var normalBodyTempK = ent.Comp3?.NormalBodyTemperature ?? Atmospherics.T0C + Atmospherics.NormalBodyTemperature;
             var heatCapacity = _atmosSys.GetHeatCapacity(actualGas, false);
 
             // Calculate thermal energy difference between inhaled gas and normal body temperature


lzk228:
can you make not a solid line where mobs will take damage but a gradient
like the colder a gas the more damage a mob gets

MilonPL:
I'll wait for some maintainer input first, but making it scale shouldn't be an issue

ArtisticRoomba:
Just to preemptively answer some comments:
- Yes, supercooled oxygen and nitrogen is broken. I shouldn't be able to get 40 minutes on an emergency tank.
- No, you don't need supercooled oxygen/nitrogen for patrolling outside with two hands free.
- No, this won't kill Atmos content. Atmos has more content than you think.
- No, we don't have to wait until newmed to implement this. The previous reason why we nerfed tanks was because antags could sit outside forever, supercooled oxygen and nitrogen enables this.

If you want to give supercooled nitrogen/oxygen a buff or promote its usage, you can:
- Add the ability to adjust the mass flow rate (L/s) on jetpacks. Right now they indiscriminately consume propellant no matter the density. Realistically, your reaction mass is much more dense so you can get a lot more speed out of it or you can go for longer.
- Encourage atmospherics to purchase supercooled oxygen and nitrogen tanks from cargo a heat them to refill the stations oxygen and nitrogen stores. However without Flowmos and with gas miners still implemented this is far, far in the future.

TheRaidGo (approved):

kosticia:
How this works with cryopods?

Moryachok136:
I think you need at least -30 or -40 for damage. Lol -20 isn't even too cold.

ArtisticRoomba:
In real life, breathing -20C gas will result in lung damage.

HerCoyote23:
-20C is -4 F.
A few days ago, it was around -10F (-23 C) where I live. I went outside and it sure was cold, but my lungs were fine.

Errant-4:
thb I'm not super interested in bikeshedding the exact values to be used, especially based on arguments from realism, before we have even tested the feature. Crewmembers frequently experience body temperatures of negative celsius degrees without dying. Who is to say we can't err in the opposite way here, if it turns out better for gameplay

deltanedas:
> Atmos has more content than you think.

you really overestimate how much content there is for atmos (zero)
instead of set up distro and make liquid oxy for engi then go cryo, now its set up distro and go cryo

> antags going into space forever
they can still do this by dragging an oxygen canister, which if they can get liquid gases, they can get that too obviously. an oxygen canister will last you *forever*

> proposed jetpack buff

already exists, cold gases make jetpacks last longer

this is just another "hidden" nerf to salv and giving atmos even less gameplay than there already is

NoElkaTheGod:
> How this works with cryopods?

I assume that you will have to use an oxygen/nitrogen tank now. Shouldn't be a big deal, and even kinda makes sense.

metalgearsloth:
> you really overestimate how much content there is for atmos (zero)

Is it really possible to comment on atmos content when half the department still isn't predicted in 2025.

deltanedas:
no content and what is only 5% is predicted so you go cryo ss13 style

pcaessayrs:
Does this include space at absolute 0, and how does that affect the time to death for most species?

Errant-4:
> Does this include space at absolute 0, and how does that affect the time to death for most species?

Hasn't been tested yet but theoretically it wouldn't make much of an impact because in space there is nothing to breathe. So the Inhale code will exit early, and no damage is applied from breathing

Content.Server/Body/Components/RespiratorComponent.cs
```diff
@@ -29,6 +29,30 @@ public sealed partial class RespiratorComponent : Component
...
+        /// <see cref="DamageSpecifier"/> of damage taken when breathing in gas at a low temperature.
+        /// </summary>
+        [DataField]
+        public DamageSpecifier LowTemperatureDamage = default!;
+
+        /// <summary>
+        ///     The maximum temperature of gas the entity can breathe before it starts taking damage.
+        /// </summary>
+        [DataField]
+        public float HighTemperatureDamageThreshold = 253.15f; // -20C
```
pcaessayrs, line 48: This should the LowTemperatureDamageThreshold, right? It's lower than the other value.

Content.Server/Body/Components/RespiratorComponent.cs
```diff
@@ -29,6 +29,30 @@ public sealed partial class RespiratorComponent : Component
...
+
+        /// <summary>
+        ///     The minimum temperature of gas the entity can breathe before it starts taking damage.
+        /// </summary>
+        [DataField]
+        public float LowTemperatureDamageThreshold = 333.15f; // 60C
```
pcaessayrs, line 54: Likewise, this is the HighTemperatureDamageThreshold?

pcaessayrs (changes requested):
High and Low temp thresholds are reversed.

pcaessayrs:
Testing with Debug Counters, a naked person in space will go crit in 35 seconds now (down from 65 seconds before). A person on a spaced hull tile will go crit in 43 seconds (down from 105 seconds before).

This is also affected by the thresholds being wrong (see the above review), so - a person just standing in a regular room will go crit in about 5 minutes either way...


before34682_65_105
after34682_36_43

Pronana:
> they can still do this by dragging an oxygen canister, which if they can get liquid gases, they can get that too obviously. an oxygen canister will last you _forever_

You only need three full sized oxygen tanks to last you the entire shift outside, and having externals access means you can just pop in every 30 or so minutes to refill at a can for free, and security definitely isn't going to have the time, manpower, or patience to react to you doing this. 

This doesn't actually prevent antags from camping outside in space, you can do it really easily without liquid air, I know I've done it before when someone needed the AI disabled and I didn't want to round remove them so I just hung out with them outside until the round ended and chatted with them. 

A better solution would be to just make space more dangerous for antags. Either by giving security more tools to deal with an antag who is in space, or adding random events that make space more dangerous even if you have a hardsuit. Personally I think the former is better. 

Alternatively, you could make it so, as suggested in the original issue, _some_ oxygen tanks "decompress" the liquid o2 making it safe to breathe, that way people can't use liquid air in emergency tanks, only specialized tanks or jetpacks.

And I know we shouldn't use realism as a suggestion for balance changes, but liquid oxygen tanks for breathing exist in the real world and have existed for a long time this is definitely technology that should exist, and as stated by others you can breathe very cold air without it really being harmful because it just doesn't have a very good heat transfer coefficient.

Lastly this removes atmos content which is incredibly lacking at the moment, and harms salvage and Voxes pretty directly who are very big users of liquid air. I don't think this is going to do anything to stop antags from being able to camp out in space, and is instead going to be a change that players will malign.

pcaessayrs:
> And I know we shouldn't use realism as a suggestion for balance changes, but liquid oxygen tanks for breathing exist in the real world and have existed for a long time this is definitely technology that should exist, and as stated by others you can breathe very cold air without it really being harmful because it just doesn't have a very good heat transfer coefficient.

Speaking of, it could be very fun if heat / cold damage was indeed based off the heat exchange such that breathing liquid oxygen is fine, but breathing supercooled water vapor (which has twice the heat capacity) would be dangerous. This would also work to fix the above problem with the current implementation where you take extreme cold damage from breathing space even though it's a vacuum.

Errant-4:
Yeah I was mistaken about vacuum cancelling respiration, because it's not actually a null gas, just gas with 0 moles but still, notably, with a temperature. Perhaps damage should be scaled not (just) by temperature, but also by density. That way 0 density would translate to 0 damage. Unsure about the upper cap of this scaling at this moment, because it could lead to very cold and dense gases being instantly lethal

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -109,24 +109,61 @@ public override void Update(float frameTime)
...
+
+            var scaledDamage = ent.Comp1.LowTemperatureDamage * scaleFactor;
+            _damageableSys.TryChangeDamage(ent, scaledDamage, true);
+
+            // Log damage for admins
+            _adminLogger.Add(LogType.Temperature,
+                $"{ToPrettyString(ent):entity} took {scaledDamage} breathing damage from {gasTemp:F1}K gas");
+        }
+
         var actualGas = ev.Gas.RemoveVolume(Atmospherics.BreathVolume);
```
slarticodefast, line 167: I would do the damage checks below this line and use actualGas for the temperature (if its volume isn't zero).
That way you prevent the player from taking additional cold damage in space as if you are holding your breath.
This is important as vacuum is currently coded to be extremely cold gas (at TCMB) in order to make radiators work in space.

slarticodefast (commented):
You got way too many magic numbers hardcoded in there.

MilonPL:
oops forgot to fill in all the components so tests will fail 
still good to review i will fix later

Resources/Locale/en-US/alerts/alerts.ftl
```diff
@@ -22,6 +22,12 @@ alerts-too-cold-desc = You're [color=cyan]freezing cold![/color] Get somewhere w
 alerts-too-hot-name = [color=red]Too Hot[/color]
 alerts-too-hot-desc = It's [color=red]too hot![/color] Get somewhere colder, take off any insulating clothing like a space suit, or at least get away from the flames.
 
+alerts-air-too-cold-name = [color=cyan]Too Cold Air[/color]
+alerts-air-too-cold-desc = The air you're inhaling is [color=cyan]too cold cold![/color]
```
Simyon264, line 26: Too cold cold.

Resources/Prototypes/Alerts/alerts.yml
```diff
@@ -121,6 +121,24 @@
...
+  - sprite: /Textures/Interface/Alerts/temperature.rsi # TODO custom sprites
+    state: hot2
+  name: alerts-air-too-hot-name
+  description: alerts-air-too-hot-desc
+
+- type: alert
+  id: AirCold
+  category: AirTemperature
+  icons:
+  - sprite: /Textures/Interface/Alerts/temperature.rsi # TODO custom sprites
```
Simyon264, line 137: For accessibility and new player friendliness these should have a different indicator.

Simyon264 (changes requested):
Works ingame, have not looked at the code.

Resources/Prototypes/Alerts/alerts.yml
MilonPL: todo take it from https://github.com/lzk228/space-station-15/tree/0d199073dc03147d24179cc3912672f8cec63fcf/Resources/Textures/Interface/Alerts/breathing.rsi

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -109,26 +114,92 @@ public override void Update(float frameTime)
...
+                var scaledDamage = ent.Comp1.LowTemperatureDamage * energyScaleFactor;
+                _damageableSys.TryChangeDamage(ent, scaledDamage, true);
+
+                // Show the alert for the player
+                _alertsSystem.ShowAlert(ent, LowTempAlertId);
+
+                // Log damage for admins with energy information
+                _adminLogger.Add(LogType.Temperature,
+                    $"{ToPrettyString(ent):entity} took {scaledDamage.GetTotal():damage} breathing damage ");
+            }
```
ljm862, line 190: As this is essentially the same, why not extract it into a separate function and pass in the alertId and temperature damage? Would be easier to maintain in the future!

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -109,26 +114,92 @@ public override void Update(float frameTime)
...
         {
             return;
         }
 
+        // Remove a volume of gas from the environment
         var actualGas = ev.Gas.RemoveVolume(Atmospherics.BreathVolume);
 
+        // Only apply temperature damage if there are actually moles to breathe
+        // This prevents damage from vacuum which has temperature but no actual gas
+        if (actualGas.TotalMoles > Atmospherics.GasMinMoles)
```
ljm862, line 140: Speaking of extracting, you could probably extract everything this `if` pertains to into a `ApplyGasTemperatureDamage` or something similar. There may be other uses for it in the future and would save extracting it, or potential duplication

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -109,26 +114,92 @@ public override void Update(float frameTime)
...
+        {
+            var gasTemp = actualGas.Temperature;
+            const float normalBodyTempK = Atmospherics.T0C + Atmospherics.NormalBodyTemperature;
+
+            var heatCapacity = _atmosSys.GetHeatCapacity(actualGas, false);
+
+            // Calculate thermal energy difference between inhaled gas and normal body temperature
+            // This represents how much thermal energy would be transferred to/from the body
+            var normalThermalEnergy = heatCapacity * normalBodyTempK;
```
Partmedia, line 149: Shouldn't this be the heat capacity of the body?

Content.Server/Body/Components/RespiratorComponent.cs
```diff
@@ -29,6 +29,30 @@ public sealed partial class RespiratorComponent : Component
...
+        ///     The maximum temperature of gas the entity can breathe before it starts taking damage.
+        /// </summary>
+        [DataField]
+        public float HighTemperatureDamageThreshold = 333.15f; // 60C
+
+        /// <summary>
+        ///     The minimum temperature of gas the entity can breathe before it starts taking damage.
+        /// </summary>
+        [DataField]
+        public float LowTemperatureDamageThreshold = 253.15f; // -20C
```
Partmedia, line 54: This duplicates a lot of fields from `TemperatureComponent`. Consider basing it off the values there, instead of reinventing `TemperatureComponent`.

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -109,26 +114,92 @@ public override void Update(float frameTime)
...
+            var gasThermalEnergy = _atmosSys.GetThermalEnergy(actualGas);
+            var energyDifference = Math.Abs(gasThermalEnergy - normalThermalEnergy);
```
Partmedia, line 151: I see. This a roundabout way of saying $\delta E = c_{body} (T_{gas} - T_{body})$.

Except this is not how thermal transfer works. Enough energy is transferred to equalize the temperatures. What you breath out isn't at absolute zero. Take a look how atmos code calculates this.

Content.Server/Body/Components/RespiratorComponent.cs
MilonPL: This is intentional, currently not used by anything, but exist in case of mobs that have insulation but could be immune to gas temperature damage. Same for mobs that can take temperature damage but do not have a RespiratorComponent. In general I think having it split is reasonable, they aren't that similar or related, other than both storing temperature thresholds.

Content.Server/Body/Systems/RespiratorSystem.cs
MilonPL: We aren't actually modelling the accurate thermal transfer between the gas and the body for gameplay reasons, because the heat capacity of the body is significantly larger than the heat capacity of the gas you breathe in. 

In order to realistically model it using the equilibrium temperature formula, we would have to cheat with the numbers and scale thermal transfer significantly in order to have it do anything.

Content.Server/Body/Components/RespiratorComponent.cs
Partmedia: I think I see what you're saying, but I still don't think this split makes sense. Long term, I'm concerned about the maintainability of so many different temperature thresholds for a single entity.

Besides, the whole point of ECS is to compose behaviors. `TemperatureComponent` is designed to store the temperature limits of an entity.

You give two examples:
- in case of mobs that have insulation but could be immune to gas temperature damage: You could add a bool field `IgnoresGasTemperature`, or, extend `RespiratorComponent` e.g. with `RepiratorWithTempLimitComponent`
- mobs that can take temperature damage but do not have a RespiratorComponent: why is this relevant here?

Content.Server/Body/Systems/RespiratorSystem.cs
Partmedia: Then why go through the motions of bothering with a specific heat capacity at all? Just make it a `temperatureScaleFactor`.

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -109,26 +114,94 @@ public override void Update(float frameTime)
...
+            // Oxygen at -200°C (73.15K) with its heat capacity will cause maximum damage (5x multiplier)
+            const float baseDamageEnergyThreshold = 25f;
+
+            // Cap damage multiplier at 5x to prevent extremely hot/cold small amounts of gas from dealing excessive damage
+            const float maxDamageMultiplier = 5.0f;
+
+            // Calculate a damage scale factor based on thermal energy difference
+            // The factor starts at 1.0 at the base threshold and increases linearly with energy
+            var energyScaleFactor = 1.0f + (energyDifference / baseDamageEnergyThreshold);
+            energyScaleFactor = Math.Min(energyScaleFactor, maxDamageMultiplier);
```
Partmedia, line 165: Consider using some combination of `Clamp` and `Lerp` to make this more readable. Or, put the whole equation on one line so someone can easily understand what the formula is and plot it.

Content.Server/Body/Systems/RespiratorSystem.cs
Partmedia: Yes, I agree. Besides the `LowTempAlertId` vs `HighTempAlertId`, everything else just looks like copy/paste. I think it would be better to merge this into one code path.

Content.Server/Body/Components/RespiratorComponent.cs
MilonPL: I could *move* it to the `TemperatureComponent`, however I think it should still remain as separate fields, especially considering that:
- The thresholds at which a mob starts taking cold/hot damage from gases are different enough from those on the TemperatureComponent to warrant it being a different customizable field.
- Similarly, the damage taken from gases is significantly higher than damage from fires or hot atmosphere itself, and is tied to respiration cycles, which in turn is tied to the RespiratorComponent. 

Given how closely related the damage and thresholds are to the RespiratorComponent - and the fact that they are also handled by RespiratorSystem - does it make sense to have it on TemperatureComponent instead?

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -109,26 +114,94 @@ public override void Update(float frameTime)
...
+
+            // Apply high temperature damage
+            if (gasTemp > ent.Comp1.HighTemperatureDamageThreshold)
```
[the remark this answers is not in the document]
MilonPL: I agree that scaling it linearly with temperature rather than energy difference itself is better for this use case, considering that we would have to cheat with the numbers anyways for it to work the way we want to.
To be clear, that was the original implementation. Using heat transfer seems massively over-engineered here, but it was requested by @slarticodefast.

Content.Server/Body/Systems/RespiratorSystem.cs
MilonPL: Could you please look at this version of the code? This was the previous implementation before Slarti requested it to use heat capacity rather than linear scaling. We still wouldn't want to deal damage if there's not enough moles for a full breath (like in vacuum, the check is missing in this version), but I feel like this approach was better here.
https://github.com/space-wizards/space-station-14/blob/26adaeb3bb4e2e7b222cd7df5c62db7f1b623464/Content.Server/Body/Systems/RespiratorSystem.cs

Partmedia: I think simple temperature scaling gives us the behavior that we want to model without doing too many mental gymnastics. Why bother dealing with heat capacity if you're just going to ignore the heat capacity-equilibrium equation?

If you want specific feedback on the old version:
- The same comments about the if statement applies
- The same readability/one-line still applies
- We may want to more closely look at the equation

PJB3005: "I am taking burn damage in an 0.1 kPa environment" has historically been a bug in SS13. I'm not following this convo entirely, so drive-by-commenting: whatever you do, make sure that doesn't happen.

MissKay1994:
Something to note I don't know if it was considered, is it's not unusual at times for the hallway to be -30c on the station. Which will mean in turn the thresholds you want would force internals crates to be purchased round start every round and handed out in mass the moment any spacings or gas leaks happen so people aren't dropping dead from the cold.

NoElkaTheGod:
> Something to note I don't know if it was considered, is it's not unusual at times for the hallway to be -30c on the station. Which will mean in turn the thresholds you want would force internals crates to be purchased round start every round and handed out in mass the moment any spacings or gas leaks happen so people aren't dropping dead from the cold.

I mean, everyone has a survival box. If you get rid of it roundstart, and then die because you don't have an oxygen tank, that's entirely your problem tbh.

Content.Server/Body/Systems/RespiratorSystem.cs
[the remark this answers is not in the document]
slarticodefast: Using temperature rather than heat capacity will mean you will be able to kill someone with 0.00001 moles of supercooled/superheated gas. This will be highly problematic in cases where larger parts of the station are spaced and space can quickly cool down the remaining, low pressure air (because that does respect heat capacity). I think this would be a bad idea since airlocks are still broken at the moment.
Also it could be abused to quickly kill a whole room by first spacing it and then opening a gas tank of supercooled gas. Even if it only contains a few moles of gas, its temperature would then deal full damage.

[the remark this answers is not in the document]
Partmedia: Given all this, I think the current implementation is going to needlessly duplicate a lot of code that already exists in `TemperatureComponent` and `TemperatureSystem`. I could suggest investigating splitting out the code from the existing `OnAtmosExposedUpdate` in `TemperatureSystem` and adding a public method that this component can then inject into the same air-handling code, while skipping any resistances if applicable.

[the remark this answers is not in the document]
Partmedia: Let's discuss on Discord if it's not clear what I mean by this.

MissKay1994:
> I mean, everyone has a survival box. If you get rid of it roundstart, and then die because you don't have an oxygen tank, that's entirely your problem tbh.

The survival box is not particularly useful for the problem of whole station is cold when that only holds 4 minutes of air

crazybrain23:
Instead of damage, why not just bring the body temperature towards that of the gas? 
If you know the temperature and moles of gas breathed, and the mass and temperature of the mob, can't you calculate the change?

retequizzle:
me when i set the station's AC to cold and accidentally (or not accidentally) kill the entire station

i am aware it would take a significant amount of time to take effect because of how wizden atmos currently works (air cycling by default is minimal so the volume replaced is equally minimal, without depressurization or tinkering with air alarms to panic siphon areas or other forms of user input) but it would take an even longer time to fix considering temperature equalization would require you to either pump a ton of ambient temp air into distro, or hotter air into distro and then run the risk of hot spots/cold spots until temps eventually stabilize. and by then the round's been over for an hour.

"oops you breathed spicy/minty air so now you're dying" also doesn't feel great for random station inhabitants who aren't doing the thing brought up in the issue this PR is meant to close

[ideaguys]

would it not make more sense to allow for personal tanks to warm to ambient temperatures (or cool to, in event of spacewalking) and then adjust the volume in personal tanks to compensate for the variable pressure value at that point? it'd even allow for us to introduce overpressurization of personal tanks. that'd _also_ allow for us to make more differentiation between emergency personal tanks and large capacity personal tanks (aka da big bois) by making the latter pressure resistant

[/ideaguys]

it's entirely possible i don't see the vision behind this, and i'll admit i've not read through the comments to see how much of this was addressed since the PR's been open since january, but i gotta admit i don't feel quite right about this much of an overarching changing just to target players who are cooling their air tanks before going out on salvages or traitors hiding out in space for extended periods

PR #34711, opened 2025-01-28
"DO NOT MAP" items integration test

Killerqu00:
## About the PR
If an item with "DO NOT MAP" gets mapped on a map, it will get seen via test fail

## Why / Balance
g

## Technical details
the test is currently inside the NoSavedPostMapInitTest, because the code is practically the same. I can separate those if requested

## Requirements

## Breaking changes
I have to mention that Meta currently has a warden stamp explicitly mapped in warden's locker for some reason, which will result in failure of this test upon merging

**Changelog**

### DIFF
--- a/Content.IntegrationTests/Tests/PostMapInitTest.cs
+++ b/Content.IntegrationTests/Tests/PostMapInitTest.cs
@@ -42,6 +42,11 @@ namespace Content.IntegrationTests.Tests
             "/Maps/Shuttles/infiltrator.yml",
         };
 
+        private static readonly string[] DoNotMapWhitelist =
+        {
+            "/Maps/centcomm.yml",
+        };
+
         private static readonly string[] GameMaps =
         {
             "Dev",
@@ -120,6 +125,7 @@ namespace Content.IntegrationTests.Tests
             var server = pair.Server;
 
             var resourceManager = server.ResolveDependency<IResourceManager>();
+            var protoManager = server.ResolveDependency<IPrototypeManager>();
             var mapFolder = new ResPath("/Maps");
             var maps = resourceManager
                 .ContentFindFiles(mapFolder)
@@ -151,6 +157,21 @@ namespace Content.IntegrationTests.Tests
                 var postMapInit = meta["postmapinit"].AsBool();
 
                 Assert.That(postMapInit, Is.False, $"Map {map.Filename} was saved postmapinit");
+
+                // testing that maps have nothing with the "DO NOT MAP" suffix
+                // I do it here because it's basically copy-paste code for the most part
+                var yamlEntities = root["entities"];
+                foreach (var yamlEntity in (YamlSequenceNode)yamlEntities)
+                {
+                    var protoId = yamlEntity["proto"].AsString();
+                    protoManager.TryIndex(protoId, out var proto, false);
+                    if (proto is null || proto.EditorSuffix is null)
+                        continue;
+                    if (proto.EditorSuffix.ToUpper().Contains("DO NOT MAP") && !DoNotMapWhitelist.Contains(map.ToString()))
+                    {
+                        Assert.Fail($"\nMap {map} has the DO NOT MAP prototype {proto}");
+                    }
+                }
             }
             await pair.CleanReturnAsync();
         }


Pixeltheaertist:
Does this stop maps from initting or anything? Or just a log

Killerqu00:
> Does this stop maps from initting or anything? Or just a log

It will fail the integration test and tell which prototype on what map did that

Killerqu00:
failed test runs fine locally :fire:

IamVelcroboy:
Failing tests might be a frustrating way to approach this. As downstreams don't all follow the same guidelines or sometimes we make exceptions for certain situations. 

While this would overall be helpful to flag these items. I wonder if there's another way of logging this w/out making tests fail?

Killerqu00:
> Failing tests might be a frustrating way to approach this. As downstreams don't all follow the same guidelines or sometimes we make exceptions for certain situations. 
> 
> While this would overall be helpful to flag these items. I wonder if there's another way of logging this w/out making tests fail? 

remove suffix

Pixeltheaertist:
I agree with the downstream thing imo, I’ll probably just remove this PR next time we merge but it’s a bit tedious for everyone to do. Just my opinion however

Killerqu00:
> I agree with the downstream thing imo, I’ll probably just remove this PR next time we merge but it’s a bit tedious for everyone to do. Just my opinion however

I mean, you can remove the suffixes. if you have an alternative suggestion, I can try doing that instead

Pixeltheaertist:
> > I agree with the downstream thing imo, I’ll probably just remove this PR next time we merge but it’s a bit tedious for everyone to do. Just my opinion however
> 
> I mean, you can remove the suffixes. if you have an alternative suggestion, I can try doing that instead

I have specific objects labeled do not map because there's meant to be one in a very specific spot. But you're right, i don't have a better suggestion, so it's not really the end of the world to delete the system on my side.

ArtisticRoomba:
IMO [DO NOT MAP] items are very seldom mapped. Usually stuff like this is caught during mapping review and most mappers are smart enough to listen to the suffix.

While forks can choose to not merge it, I don't really see the benefit of saving some time versus causing some annoying issues for downstreams. Like others have said, this test really only tests *our* policy and annoys other servers with changed policies.

Killerqu00:
Also, there is a map whitelist built into the test - you can just add the path to your map to make the test ignore it. This should solve the issue for downstreams, since (I suppose) most usecases are related to custom maps.

NotSoDana:
Very bruh moment. [DO NOT MAP] can be ignored in some situations and this fall of tests will be more annoying than useful. Like said above, stuff like this usually getting caught on mapping review and don't really need to ruin tests if adding was approved.

Killerqu00:
> [DO NOT MAP] can be ignored in some situations

list at least one situation like this, please

> Like said above, stuff like this usually getting caught on mapping review

Meta has a warden stamp mapped right now, meaning it wasn't caught

Ko4ergaPunk:
> > [DO NOT MAP] can be ignored in some situations
> 
> list at least one situation like this, please

I know it might not be yours problem, but a lot of servers upstream from Wizards. On Corvax, for example, you can mapping shuttle guns

poeMota:
I think you should also add testing for debug, admin, admem, etc items. It is also worth adding dev.yml to the whitelist

poeMota:
And by the way, does this test work correctly with localization? After all, on forks DO NOT MAP suffix can be translated

Killerqu00:
> And by the way, does this test work correctly with localization? After all, on forks DO NOT MAP suffix can be translated

sadly not, because the suffix is directly written in the prototypes

TheShuEd:
I also think it shouldn't be based on a suffix, as it's an unsafe line. An entity could have the suffix [DO NOT MAP!] and it won't be handled by this integration test. I think it makes sense to make an entity category like with HideSpawnMenu or ForkFiltered.

Killerqu00:
> I also think it shouldn't be based on a suffix, as it's an unsafe line. An entity could have the suffix [DO NOT MAP!] and it won't be handled by this integration test. I think it makes sense to make an entity category like with HideSpawnMenu or ForkFiltered.

it will actually be handled, because suffix only has to contain "DO NOT MAP" in any case. I'll look into making it an entity category, though

TheShuEd:
Yes, that would be a much better solution.

TheShuEd:
> > I also think it shouldn't be based on a suffix, as it's an unsafe line. An entity could have the suffix [DO NOT MAP!] and it won't be handled by this integration test. I think it makes sense to make an entity category like with HideSpawnMenu or ForkFiltered.
> 
> it will actually be handled, because suffix only has to contain "DO NOT MAP" in any case. I'll look into making it an entity category, though

There's still no typo check or a hundred other reasons that the suffix is misspelled.

Killerqu00:
> > > I also think it shouldn't be based on a suffix, as it's an unsafe line. An entity could have the suffix [DO NOT MAP!] and it won't be handled by this integration test. I think it makes sense to make an entity category like with HideSpawnMenu or ForkFiltered.
> > 
> > 
> > it will actually be handled, because suffix only has to contain "DO NOT MAP" in any case. I'll look into making it an entity category, though
> 
> There's still no typo check or a hundred other reasons that the suffix is misspelled.

that's true, golden honk has "no mapping" suffix

Content.IntegrationTests/Tests/PostMapInitTest.cs
```diff
@@ -151,6 +157,23 @@ public async Task NoSavedPostMapInitTest()
                 var postMapInit = meta["postmapinit"].AsBool();
 
                 Assert.That(postMapInit, Is.False, $"Map {map.Filename} was saved postmapinit");
+
+                // testing that maps have nothing with the "DO NOT MAP" suffix
```
TheShuEd, line 161: suffix :trollface:

Resources/Locale/en-US/entity-categories.ftl
```diff
@@ -3,3 +3,6 @@ entity-category-name-game-rules = Game Rules
 entity-category-name-objectives = Objectives
 entity-category-name-roles = Mind Roles
 entity-category-name-mapping = Mapping
+entity-category-name-donotmap = Do not map
+
+entity-category-suffix-donotmap = DO NOT MAP
```
TheShuEd, line 8: Is this used?

TheShuEd (approved):
Looks good to me

Resources/Locale/en-US/entity-categories.ftl
Killerqu00: yes, it's the default suffix that is applied to all entities in the DoNotMap group

slarticodefast:
With the suffix removed, do mappers have a way to see which entities should not be mapped from the spawn menu?

Killerqu00:
> With the suffix removed, do mappers have a way to see which entities should not be mapped from the spawn menu? 

the suffix is not removed - it is applied as a default suffix. All items with additional suffixes have their suffixes set manually (this is shitYML, but I can't fix this in the same PR)
basically that means that all suffixes are the same in editor as before the PR

image

> Also, can you merge master to fix the tests?

k

Killerqu00:
the test fail is intended and needs a mapping fix for Meta station

ElectroJr (approved):
Regarding some of the comments/concerns people have raised:

- Any entity in the "DoNotMap" category automatically get given a "DO NOT MAP" suffix (unless that entity already defines a custom suffix).
- Nothing prevents someone from giving an entity the suffix without adding it to the category if they want to discourage mappers from mapping it without making that entity fail the test. I.e., the test/category isn't meant for entities that are actually allowed to be mapped, but only in a specific spot or for a specific map. 
- If downstreams want to allow these entities to be mapped, they can just disable or modify this part of the test. The tests don't get updated that frequently, so it shouldn't be a significant maintenance burden. They can also modify the relevant entities to remove them from the category, but I guess that might result in merge conflicts in the future.
- It's possible the test might be updated in future to automatically exclude shuttles or other non-standard maps. Half of the current failures are due to misc evac shuttles containing disallowed entities.

> the test fail is intended and needs a mapping fix for Meta station

For now I've just expanded the whitelist. I'll open a separate issue with a list of maps that need fixing.

PR #34713, opened 2025-01-28
Fix thrust animation rotation

### LINKED ISSUE #34692: Animation WeaponArcThrust applies weapon sprites backwards
## Description

WeaponArcThrust applies sprites backwards when used on melee weapons. This makes stun prods or spears appear they are hitting people backwards with the butt of the items rather than the lethal end.

**Reproduction**

1. Use a melee weapon with WeaponArcThrust (e.g. stun prod, spear)
2. Apply precise attack (LMB)
3. Sprite appears 'backwards'

**Screenshots**

Image

themias:
## About the PR

This fixes the rotation for the thrust animation that certain weapons use upon a precision attack (LMB).

Fixes #34692 

## Technical details

The WideAnimationRotation (passed in through spriteRotation) was not being applied to the animation for thrusts.

(It's  called 'WideAnimationRotation', but its purpose is to simply point the weapon out from the player. We also want that same behavior when thrusting. This param is already defined for thrusting weapons for their wide attacks.)

## Media

https://github.com/user-attachments/assets/4bc8fed4-cd28-47ac-9888-da6960fcee74

## Requirements

**Changelog**
:cl:
- fix: Fixed precision attack animation for thrust weapons

### DIFF
--- a/Content.Client/Weapons/Melee/MeleeWeaponSystem.Effects.cs
+++ b/Content.Client/Weapons/Melee/MeleeWeaponSystem.Effects.cs
@@ -138,22 +138,13 @@ public sealed partial class MeleeWeaponSystem
         const float length = 0.15f;
         var startOffset = sprite.Rotation.RotateVec(new Vector2(0f, -distance / 5f));
         var endOffset = sprite.Rotation.RotateVec(new Vector2(0f, -distance));
-        var rotation = sprite.Rotation + spriteRotation;
+        sprite.Rotation += spriteRotation;
 
         return new Animation()
         {
             Length = TimeSpan.FromSeconds(length),
             AnimationTracks =
             {
-                new AnimationTrackComponentProperty()
-                {
-                    ComponentType = typeof(SpriteComponent),
-                    Property = nameof(SpriteComponent.Rotation),
-                    KeyFrames =
-                    {
-                        new AnimationTrackProperty.KeyFrame(rotation, 0f)
-                    }
-                },
                 new AnimationTrackComponentProperty()
                 {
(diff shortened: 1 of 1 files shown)


Content.Client/Weapons/Melee/MeleeWeaponSystem.Effects.cs
```diff
@@ -138,12 +138,22 @@ private Animation GetThrustAnimation(SpriteComponent sprite, float distance, Ang
...
             Length = TimeSpan.FromSeconds(length),
             AnimationTracks =
             {
+                new AnimationTrackComponentProperty()
+                {
+                    ComponentType = typeof(SpriteComponent),
+                    Property = nameof(SpriteComponent.Rotation),
+                    KeyFrames =
+                    {
+                        new AnimationTrackProperty.KeyFrame(rotation, 0f)
```
slarticodefast, line 154: I think there is no need to put this inside an animation, since this value is constant throughout.
Can't you directly set the sprite rotation?

themias: Ah, I was worried doing so might cause the rotation to persist/accumulate, but it works fine! Done.

slarticodefast (approved):
Works as intended, thanks for the bugfix!

Content.Client/Weapons/Melee/MeleeWeaponSystem.Effects.cs
slarticodefast: Yeah, it spawns a temporary entity that copies the sprite and only that copy is being changed here, so it's safe.

PR #34079, opened 2024-12-26
Fake mindshield componentry and Implanter

poklj:
## About the PR

Added a a Component and Implanter that injects a Togglable fake Mindshield into someone, allowing them to deceive anyone with the ability to see mindshielded status.

This functionality is also later required for Changelings ["Mindshield"](https://docs.spacestation14.com/en/space-station-14/round-flow/proposals/changeling.html#stealth) ability at a later date.

As well due to how Implanting works, the Check for if a implant is already implanted has been moved to AFTER the DoAfter to prevent sufficiently Metagaming Secoffs from instantly identifying if a particular individual is faking their Implant.

## Why / Balance

At the moment any attempt to Disguise or impersonate a Command or Security staff is foiled by a lack of Mindshield, and while an antag CAN acquire them via duplicitious means via the ATS, that is a higher bar to cross and further means they might otherwise be unable to shed their disguise to blend back into the crowd.

A Togglable fake is therefor able to Allow an antag with sufficient speech skills to impersonate Command or Security staff and further blend back into the crowd, however such a method is still not foolproof and would require the individual to pick their moment.
 
## Technical details

Moved the SameImplant check into the shared system and into the `Implant` method, as before it would check at the action of trying to implant. it won't use up the implant but gives some breathing room to the implantee to act (and further make it obvious if someone is attempting to metagame), this can probably also be made to Harm the Implanter a-la #32136 to further punish attempted metagaming.

Added a check inside of ShowMindShieldIconsSystem to check for enabled Fake mindshields.

## Media

https://github.com/user-attachments/assets/cb45c49f-17fd-45f9-a335-b206c5803522

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

Moved `CheckSameImplant` from `Content.Server.Implants.ImplanterSystem` to `Content.Shared.Implants.SharedImplanterSystem`
**Changelog**

:cl: Kickguy223 
- add: A new Syndicate and Thief togglable implant is available capable of Impersonating a Mindshield
- tweak: All implants will now inform you if an implant is already present after the attempt instead of before

PR #34718, opened 2025-01-28
Fake Mindshield Implant cleans itself up on draw

poklj:
## About the PR

Fake mindshield (#34079) has a minor bug on draw where the implant drawing doesn't clean up the components added to the implanted entity.

Adding an event that's relayed to the implant when it's drawn and having the implant clean itself up solves this problem (and further allows other Implants to have unique draw effects
## Why / Balance

Fixes a bug introduced with (#34079) where if you draw a Fake mindshield while it is toggled on, the Mindshield info icon will stay on.
## Technical details

- Give Fake mindshield a cleanup method, to remove the component added fixing a problem where the icon persists after implant draw
## Requirements

## Breaking changes

**Changelog**

No CL No Fun


beck-thompson (approved):
Looks good :saluting_face:

PR #34723, opened 2025-01-28
Fixed bola effect stacking

### LINKED ISSUE #34694: Bolas can apply multiple times on the same mob
## Description

Bolas can stack and collide with the same target. This is a problem as there might be multiple targets you want to slow down.

This can also be abused as you can only remove one bola at a time, so you may be perpetually movelocked if multiple people throw bolas. At about 5 or 6 bolas you can barely move, and good luck taking the time to remove all of them.

Ideally bolas should only apply once and ignore players that are already hit or restrained.

**Reproduction**

1. Throw multiple bolas at the same target
2. Target has some sort of additive or multiplicative speed reduction going on

**Screenshots**

Image

impubbi:
## About the PR
Bolas had a bug that allowed them to stack and apply to the same target ([#34694](https://github.com/space-wizards/space-station-14/issues/34694))
I changed the shared ensnareable system to only allow one bola to be applied as mentioned in ([#34694](https://github.com/space-wizards/space-station-14/issues/34694))

Fixes #34694

## Why / Balance
Removes the ability to movelock and makes it less annoying to remove many bolas.

## Technical details
Changed the checks in the SharedEnsnareableSystem's TryEnsnare method to only ensnare if they aren't already ensnared. (Try saying that 3 times fast lol)

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Bolas will now only be applied once.

### DIFF
--- a/Content.Shared/Ensnaring/SharedEnsnareableSystem.cs
+++ b/Content.Shared/Ensnaring/SharedEnsnareableSystem.cs
@@ -256,15 +256,15 @@ public abstract class SharedEnsnareableSystem : EntitySystem
         if (!TryComp<EnsnareableComponent>(target, out var ensnareable))
             return false;
 
-        // Need to insert before free legs check.
-        Container.Insert(ensnare, ensnareable.Container);
 
         var legs = _body.GetBodyChildrenOfType(target, BodyPartType.Leg).Count();
-        var ensnaredLegs = (2 * ensnareable.Container.ContainedEntities.Count);
-        var freeLegs = legs - ensnaredLegs;
+        var ensnaredLegs = (ensnareable.Container.ContainedEntities.Count);
 
-        if (freeLegs > 0)
+        //If both legs are ensnared, don't let it ensnare more
+        if (legs == ensnaredLegs)
             return false;
+
+        Container.Insert(ensnare, ensnareable.Container);
 
         // Apply stamina damage to target if they weren't ensnared before.
         if (ensnareable.IsEnsnared != true)
--- a/Content.Shared/Ensnaring/SharedEnsnareableSystem.cs
+++ b/Content.Shared/Ensnaring/SharedEnsnareableSystem.cs
@@ -262,13 +262,10 @@ public abstract class SharedEnsnareableSystem : EntitySystem
 
         Container.Insert(ensnare, ensnareable.Container);
 
-        // Apply stamina damage to target if they weren't ensnared before.
-        if (ensnareable.IsEnsnared != true)
+        // Apply stamina damage to target
+        if (TryComp<StaminaComponent>(target, out var stamina))
         {
-            if (TryComp<StaminaComponent>(target, out var stamina))
-            {
-                _stamina.TakeStaminaDamage(target, component.StaminaDamage, with: ensnare, component: stamina);
-            }
+            _stamina.TakeStaminaDamage(target, component.StaminaDamage, with: ensnare, component: stamina);
         }
 
         component.Ensnared = target;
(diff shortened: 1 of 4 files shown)


Content.Shared/Ensnaring/SharedEnsnareableSystem.cs
```diff
@@ -256,16 +256,12 @@ public bool TryEnsnare(EntityUid target, EntityUid ensnare, EnsnaringComponent c
...
-
-        if (freeLegs > 0)
+        //Don't do anything if they are already ensnared
+        if (ensnareable.IsEnsnared)
             return false;
 
+        Container.Insert(ensnare, ensnareable.Container);
+
         // Apply stamina damage to target if they weren't ensnared before.
         if (ensnareable.IsEnsnared != true)
```
slarticodefast, line 266: This if condition isn't needed anymore.

slarticodefast (dismissed):

Content.Shared/Ensnaring/SharedEnsnareableSystem.cs
impubbi: Thank you, fixed

lzk228:
you can link your pr to the issue so it could be autoclosed along with pr merging 
https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

impubbi:
Thanks for the info!

slarticodefast:
Now that I look a the code in more detail it does not seem to be a bug to be able to get ensnared more than once, and more like a design choice. Before it was limited to one bola for each leg. According to the linked issue it was possible to get ensnared 5 or 6 times, so this check is failing somewhere
grafik

impubbi:
Thanks for the input!
It's now limited to 2 legs how it was designed
This implementation had some weird network problems:
Content Client_ZaVXOJl4S6

So I also made the sprint and walk speeds AutoNetworkedFields in the ensnareable component:
Content Client_lprFT3dMbm


Results:
Content Client_yA6whuenEu
Content Client_rAWNlOlVbk
Content Client_uf3aY5v1d6

keronshb:
> Now that I look a the code in more detail it does not seem to be a bug to be able to get ensnared more than once, and more like a design choice. Before it was limited to one bola for each leg (which will become more relevant once we have surgery). According to the linked issue it was possible to get ensnared 5 or 6 times, so this check is failing somewhere grafik

No, it should be a singular bola.

It doesnt matter if they have 2 legs or 1, the singular bola counts.

keronshb (changes requested):
It was always intended to be 1 bola per person. It's never meant to be 1 bola per leg, that's too overpowered.

impubbi:
If it's still in scope for this PR, I can make the maximum number of ensnares a property in the ensnare component and let another PR handle the balance side of things. I'm not sure if the ensnare system will be used for something else.

impubbi:
Thinking on it now, if another system wanted to use the ensnare component, it might get messy if the number of times an ensnare can be applied is hard-coded

keronshb:
> If it's still in scope for this PR, I can make the maximum number of ensnares a property in the ensnare component and let another PR handle the balance side of things. I'm not sure if the ensnare system will be used for something else.

It's not exclusive to bolas but I had designed it to be 1 ensnare at a time.

If you want to make a config in the ECS to allow multiple I think that's fine, but by default it should just be 1 per person.

slarticodefast:
It's fine to change it to 1 bola per person, but this should be labeled as a balance change and not a bugfix then.

keronshb:
> It's fine to change it to 1 bola per person, but this should be labeled as a balance change and not a bugfix then.

Arguably it's a bugfix because it was never intended to work like that in the first place.
I feel it's a little minutia to discuss it any further and we can leave it as a balance change, but as the person who ported this from 13 it was never ever intended to have more than 1 at once.

impubbi:
Content Client_aTHIZhS8ze
Should be working now

keronshb (approved):

PR #34734, opened 2025-01-29
Nuke Timer MinimumTime

southbridge-fur:
## About the PR

Nuke timer now resets to a minimum value if it is disabled with a time remaining lower than that minimum value.

## Why / Balance

Disincentivize people "cooking" the timer on the nuke.

## Technical details

Added a new attribute to the NukeComponent named `MinimumTime` which is utilized in `NukeSystem.cs::DisarmBomb`.
In `DisarmBomb` it checks if the RemainingTime is less than the `MinimumTime` and if so, sets the `RemainingTime` to the `MinimumTime`.

The current `MinimumTime` value is 180 seconds (three minutes) which I set since I felt like that gives crew plenty of time to respond, assuming they are already aware of the threat and have already geared up.

## Media

[A video showcasing the nuke behaving as described](https://youtu.be/rQ36Alp8Scg)

## Requirements

## Breaking changes

This could interfere with anything that utilizes or relies on a linear change in the RemainingTime value since the RemainingTime could jump up in value.

**Changelog**

:cl: Southbridge
- tweak: The Nuke will now reset to a minimum countdown value when disarmed if the countdown is below the minimum value.

### DIFF
--- a/Content.Server/Nuke/NukeComponent.cs
+++ b/Content.Server/Nuke/NukeComponent.cs
@@ -26,6 +26,14 @@ namespace Content.Server.Nuke
         public int Timer = 300;
 
         /// <summary>
+        ///     If the nuke is disarmed, this sets the minimum amount of time the timer can have.
+        ///     The remaining time will reset to this value if it is below it.
+        /// </summary>
+        [DataField("minimumTime")]
+        [ViewVariables(VVAccess.ReadWrite)]
+        public int MinimumTime = 180;
+
+        /// <summary>
         ///     How long until the bomb can arm again after deactivation.
         ///     Used to prevent announcements spam.
         /// </summary>
--- a/Content.Server/Nuke/NukeComponent.cs
+++ b/Content.Server/Nuke/NukeComponent.cs
@@ -29,8 +29,7 @@ namespace Content.Server.Nuke
         ///     If the nuke is disarmed, this sets the minimum amount of time the timer can have.
         ///     The remaining time will reset to this value if it is below it.
         /// </summary>
-        [DataField("minimumTime")]
-        [ViewVariables(VVAccess.ReadWrite)]
+        [DataField]
         public int MinimumTime = 180;
 
         /// <summary>
--- a/Content.Server/Nuke/NukeSystem.cs
+++ b/Content.Server/Nuke/NukeSystem.cs
@@ -522,6 +522,12 @@ public sealed class NukeSystem : EntitySystem
         _sound.PlayGlobalOnStation(uid, _audio.GetSound(component.DisarmSound));
         _sound.StopStationEventMusic(uid, StationEventMusicType.Nuke);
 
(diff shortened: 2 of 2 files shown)


Content.Server/Nuke/NukeComponent.cs
```diff
@@ -25,6 +25,14 @@ public sealed partial class NukeComponent : SharedNukeComponent
         [ViewVariables(VVAccess.ReadWrite)]
         public int Timer = 300;
 
+        /// <summary>
+        ///     If the nuke is disarmed, this sets the minimum amount of time the timer can have.
+        ///     The remaining time will reset to this value if it is below it.
+        /// </summary>
+        [DataField("minimumTime")]
+        [ViewVariables(VVAccess.ReadWrite)]
+        public int MinimumTime = 180;
```
metalgearsloth, line 34: 1. datafield strings obsolete and not needed
2. VVRW same if it's a datafield.

Content.Server/Nuke/NukeSystem.cs
```diff
@@ -522,6 +522,12 @@ public void DisarmBomb(EntityUid uid, NukeComponent? component = null)
         _sound.PlayGlobalOnStation(uid, _audio.GetSound(component.DisarmSound));
         _sound.StopStationEventMusic(uid, StationEventMusicType.Nuke);
 
+        // reset nuke remaining time to the minimum time if it is shorter than the minimum time.
+        if (component.RemainingTime < component.MinimumTime)
+        {
+            component.RemainingTime = component.MinimumTime;
+        }
```
metalgearsloth, line 529: Use Math.Max(remaining time, min time)

metalgearsloth (changes requested):

Content.Server/Nuke/NukeComponent.cs
southbridge-fur: for 2. do you mean VVRW is also obsolete?

Djungelskog2:
pretty sure nukies can still win by just letting it detonate elsewhere, even off-station

metalgearsloth (approved):

southbridge-fur:
> pretty sure nukies can still win by just letting it detonate elsewhere, even off-station

This isn't about that. This is to stop the very specific action of cooking the nuke off station then bringing it back on station with seconds remaining so it goes off without any chance for the crew to respond

Errant-4 (approved):
Sounds reasonable, and worked as intended

Djungelskog2:
> > pretty sure nukies can still win by just letting it detonate elsewhere, even off-station
> 
> This isn't about that. This is to stop the very specific action of cooking the nuke off station then bringing it back on station with seconds remaining so it goes off without any chance for the crew to respond
> 
> Edit: Plus like some other edge cases I could think of like when the nuke gets disarmed in the last couple seconds and a troll comes up later to arm it again.

It does literally the exact same thing, just without all the fireworks

PR #34763, opened 2025-01-30
Adding sorting to chem master

catdotjs:
## About the PR
I've added the ability to sort chemicals in Chem Master, due to me and several chemist players requesting this feature.

With this addition chem master will be able to sort by
- Alphabetically
- Quantity
- Time Added/Lastest Added (Basically when the chemical was first added to the machine)

## Why / Balance
Chemistry requires you to know where chemicals are. Without a sorting mechanism it is pretty easy to lose where you left what.

This should not really affect balance and make playing chemist just a bit easier.

## Technical details
It is a pretty straightforward addition. However I had to change how chemicals are displayed in the buffer to not repeat the same code twice. Before my PR, the element was built on a foreach loop and then immidetly added to the buffer UI.

Now there is a step to extract only needed data, then sort, then add to the UI. The new first step is necessary to make sure that Alphabetic sorting works across languages. It also allows the code to stay DRY.

## Media
https://github.com/user-attachments/assets/663e550b-e448-4944-9250-68bc9c6b3fa0

## Requirements
- [X] I have added media to this PR ~~or it does not require an ingame showcase~~.

## Breaking changes
None I hope! From my testing it works pretty smooth 

**Changelog**

* Chem Master can now sort based on following categories
 - Alphabetical
 - Quantity
 - Time Added to Machine

* Sorting is disabled by default on new machines

* Sorting settings persist in the machine for everyone


ArtisticRoomba:
Looks good. I dunno C# though, so you'll have to wait for a proper review (godo).

SlamBamActionman:
Overall I think the PR is good conceptually, but I am not too sure I understand the usecase for sorting by "None"? It feels like that should be a fallback option in code only, and instead the ingame sorting defaults to latest added option.

SlamBamActionman:
> I think none is misleading name. None basically is First in stays at top and last in goes down. It is how chem master does it currently. I could rename it to "Unsorted" to make it more clear. I assumed there will be some grumpy people who will want it "How it was".

How is this different to `Time Added/Latest Added (Basically when the chemical was first added to the machine)`? Or is it the same list but in reverse?

If that's the case, I would go with `Newest First` & `Oldest First` for the sake of consistency, and have `Oldest First` as default.

catdotjs:
> Or is it the same list but in reverse?
Yep exactly that

> If that's the case, I would go with Newest First & Oldest First as names for the sake of consistency, and have Oldest First as default.
I didn't do it that way due to fearing that the text might be too long. However now that I think about it. It makes more sense.

I'll push a commit with the change

catdotjs:
image
image

I think the reason I called it "Time added" was due to it not being about most recent used chemical but most recently added chemical. EG if you have 9 carbon and add 100u more it won't move from where it is. However if have 0u of carbon then put 100u then it will be on the top

Content.Client/Chemistry/UI/ChemMasterWindow.xaml.cs
```diff
@@ -213,6 +214,16 @@ private void UpdatePanelInfo(ChemMasterBoundUserInterfaceState state)
 
             BufferInfo.Children.Clear();
 
+            // This has to happen here due to people possibly
+            // setting sorting before putting any chemicals
+            BufferSortButton.Text = state.SortingType switch {
```
MilonPL, line 219: Line breaks around expression braces, this should be moved into a newline.

Content.Server/Chemistry/EntitySystems/ChemMasterSystem.cs
```diff
@@ -93,6 +94,16 @@ private void OnSetModeMessage(Entity<ChemMasterComponent> chemMaster, ref ChemMa
             ClickSound(chemMaster);
         }
 
+        // The message is currently empty. However It is planned to later on be filled for future additions
+        private void OnCycleSortingTypeMessage(Entity<ChemMasterComponent> chemMaster, ref ChemMasterSortingTypeCycleMessage message){
```
MilonPL, line 98: Move the brace to a new line.

Content.Server/Chemistry/EntitySystems/ChemMasterSystem.cs
```diff
@@ -93,6 +94,16 @@ private void OnSetModeMessage(Entity<ChemMasterComponent> chemMaster, ref ChemMa
...
+            chemMaster.Comp.SortingType++;
+            if (chemMaster.Comp.SortingType > ChemMasterSortingType.Latest){
```
MilonPL, line 100: above

MilonPL (commented):
Looks fine otherwise.

catdotjs:
@MilonPL I've done the changes as requested. Let me know if there is anything more needed ^^

catdotjs:
The fail on the latest ubuntu test is not related to my commits.

MilonPL (approved):
Looks good to me.

gusxyz (approved):
All the code looks clean to me.

SlamBamActionman:
I should've caught this when I merged so this is on me, but for the future make sure that your changelog is formatted properly as per Pull Request guidelines.

```
:cl:
- add: The Chem Master now has sorting categories.
```

PR #34765, opened 2025-01-30
Randomized Markings

widgetbeck:
## About the PR

Added fully randomized markings and color palettes to HumanoidCharacterAppearance.Random.

## Why / Balance

Randomized characters have always felt very lacking, because they were - All species without hair would generate looking identical and flat. This fixes that.

## Technical details

My comments on the code itself are very thorough, but here's the gist:
HumanoidCharacterAppearance.Random now creates a palette of three complementary colors, and assigns them to skin and markings differently depending on the species' SkinColor type.

It also now loops through every marking category and assigns markings from it at random. Hair and facial hair work mostly the same as they used to, but every other category has a random chance to pick a marking, assign it a color complementary to the skin color, and add it to the output HumanoidCharacterAppearance's list of markings.

Only real quirks are:
- I have to manually check to see if the marking ID says "gauze" in it to make sure gauze markings don't look bad. This could be fixed if gauze and other non-bodypart markings had their own category, but I determined that was out of scope for this PR.
- Tails have slightly different probability and color rules. You can see why in my comments.
- Slime People have a tendency to look bad because their arm and leg gradients can roll without the corresponding hand/foot markings. I looked into fixing that, but determined it was out of scope for this PR.

## Media

https://github.com/user-attachments/assets/4d9c60b7-270a-4540-a5bd-ead3d654fdb5

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Randomized characters will now roll markings.

### DIFF
--- a/Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
+++ b/Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
@@ -126,51 +126,219 @@ public sealed partial class HumanoidCharacterAppearance : ICharacterAppearance, 
     {
         var random = IoCManager.Resolve<IRobustRandom>();
         var markingManager = IoCManager.Resolve<MarkingManager>();
-        var hairStyles = markingManager.MarkingsByCategoryAndSpecies(MarkingCategories.Hair, species).Keys.ToList();
-        var facialHairStyles = markingManager.MarkingsByCategoryAndSpecies(MarkingCategories.FacialHair, species).Keys.ToList();
 
-        var newHairStyle = hairStyles.Count > 0
-            ? random.Pick(hairStyles)
-            : HairStyles.DefaultHairStyle;
+        var newFacialHairStyle = HairStyles.DefaultFacialHairStyle;
+        var newHairStyle = HairStyles.DefaultHairStyle;
+        List<Marking> newMarkings = [];
 
-        var newFacialHairStyle = facialHairStyles.Count == 0 || sex == Sex.Female
-            ? HairStyles.DefaultFacialHairStyle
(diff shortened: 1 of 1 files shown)


TheShuEd (commented):
I have only 2 complaints about this PR, and both are unnecessary. The first is that I don't like another switch case regarding the type of layer coloring. This shitty code is only getting bigger, and adding new types of color restrictions for races is becoming more and more difficult. The second complaint is the helper functions that work with color... I'm not sure that they should be here. These functions would be useful not only for this system, and they should be moved somewhere

widgetbeck:
The switch case does kind of suck, but I couldn't figure out another way to do it without going in and refactoring skin color and stuff, which is beyond the scope of what I'm trying to do here.
Those helper functions are in here because the only other place I could think to put them (and I agree, they would be useful there) is in the Color type, but that would be an engine change, which is both beyond the scope of this PR and beyond my personal area of confidence as a programmer. 

I will also point out a complaint I personally have with this PR, which is the fact that I'm checking the name of the marking prototype to ensure gauze doesn't generate in weird bright colors. 

My logic in being kind of bodge with this system is that it doesn't actually *need* to be super modular - It only runs when you click the Randomize button, so optimization isn't a big concern, and any issues with the markings and colors it generates can be adjusted by the player in the customization menu after the fact.

widgetbeck:
For the record - we've had this in the game for a couple months on Impstation, where we have significantly more markings. The players love it and we haven't run into any issues with it.

TheShuEd (approved):

beck-thompson (approved):
code is sus but I think its fine for something specific like this. Not really worse than before and it looks a lot better :shrug:

PR #34773, opened 2025-01-30
Radioxadone

kosticia:
## About the PR
So, I have added new reagent - radioxadone. That reagent can be used in cryogenics in a pretty low temperatures, and works even on dead.

## Why / Balance
Right now gas condenser actually useless. Frezon can be used just for one useless recipe. Tritium can be used for creating flamethrower, but nothing else. This is sad, because potential of interaction for atmos and chem is a pretty big. We also have radiation damage that can actually gibs players so much easy, because deals damage even on dead one. With that pull request I want add a new reagent that can be created used tritium and treats radiational damage. This will slightly increase the amount of interaction between med and engineering.

## Requirements

**Changelog**

:cl:
- add: Added radioxadone, reagent that can be used for treating the radiation damage but should be created from tritium.

### DIFF
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -1169,7 +1169,7 @@
         conditions:
         - !type:ReagentThreshold
           min: 12
-          
+
 - type: reagent
   id: Opporozidone #Name based of an altered version of the startreck chem "Opporozine"
   name: reagent-name-opporozidone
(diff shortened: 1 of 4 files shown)


Resources/Prototypes/Reagents/medicine.yml
```diff
@@ -1384,3 +1384,25 @@
...
+  metabolisms:
+    Medicine:
+      effects:
+        - !type:HealthChange
+          conditions:
+          - !type:Temperature
+            max: 180.0
+          damage:
+            groups:
+              Brute: 10
```
ArtisticRoomba, line 1406: Very high brute damage for forcing Atmos to give you Tritium as well as consuming plasma in the reaction. Personally I would remove the brute damage, you're going through a lot of trouble already and right now the only benefit is it cures radiation slightly faster with the downside of dealing with cryogenics.

Moryachok136:
I think it should heal corpses in cryo

kosticia:
Bloaty

superjj18:
Yeah I agree this is nice.  Medbays would actually use this.  It makes atmos work actually help the station aside from only money and terrorism.

kosticia:
I changed my mind.

ArcaneOcto:
> I changed my mind.

Welcome back

slarticodefast:
Thank you for your contribution. I talked to roomba and we decided to close this PR for now, since we really need better medical mechanics before adding more medical reagents. Currently they mostly have the same effects, but with slightly different stats and will need to be completly reworked in the future, so I would rather avoid adding more work to that list.

PR #34774, opened 2025-01-30
Update shutter sounds

themias:
## About the PR

This PR updates the sound for knocking on and opening/closing shutters. It also replaces the '\*knock knock\*' text with '\*rattle rattle\*'.

## Why / Balance

It sounded odd for 'knocking' on shutters to sound like glass instead of rattling shutters. The dramatic blast door opening noise also always sounded out of place to me. I feel like the new sounds are more natural.

## Media

https://github.com/user-attachments/assets/40850672-7d1b-4deb-9cfb-59e104215ae0

## Requirements

**Changelog**
:cl:
- tweak: Updated sound effects for shutters


Resources/Audio/Machines/attributions.yml
```diff
@@ -191,3 +191,8 @@
   license: "CC0-1.0"
   copyright: "by AftrLite (Github). Uses audio from hypospray.ogg and hiss.ogg (Found in Resources/Audio/Items)"
   source: "https://github.com/space-wizards/space-station-14/pull/33097"
+
+- files: ["shutter"]
+  license: "CC-BY-4.0"
```
slarticodefast, line 196: I'll need someone with licencing knowledge here.
One of the used files is licenced as CC-BY-3.0, the other one as CC-BY-4.0
Can the combine file be licenced under CC-BY-4.0 in that case?

themias: I went ahead and removed one of the files, so there's no more license conflict.

ps3moira:
bit too loud

slarticodefast (changes requested):
The blast door should keep the old sound effect and interaction popup.
It also inherits from BaseShutter at the moment, so it got changed, too.

themias:
> The blast door should keep the old sound effect and interaction popup. It also inherits from BaseShutter at the moment, so it got changed, too.

Good catch, thanks!

slarticodefast (approved):
Looks good to me!

>bit too loud

I tried it out in-game and the volume is fine in my opinion. It is actually a little less loud than the old blast door sound.

Simyon264 (approved):

PR #34793, opened 2025-01-31
Fix hitting through directional windows (and more!)

SlamBamActionman:
## About the PR

This PR fixes the ability to hit certain objects (like grilles) through thin windows - also solves some edgecases where wideswings hit through normal windows.

## Why / Balance

bugs!

## Technical details

To start with, directional windows have a larger hitbox that matches the sprite; previously objects underneath could poke out on the other side when doing physics raycasts. 

This solved it somewhat, but not completely. 
1) When doing a lightswing, the physics check would automatically accept the result if you were pressed up right against the window, since that means the distance to the grille is also 0 which caused the function to return early without checking for obstructions. This distance check is now not performed when doing weapon attacks. 

2) When doing wideswings, the raycast for what object you hit only considers the first item. This is fine most of the time, except for when there's exact overlap. The check now also looks to see if there are any other objects with the same distance as the first item so they get added to the wideswing.

3) In addition, wideswing objects now clientside get checked for obstructions. This fixes the bug where objects would be erroneously highlighted as being hit. 

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Hitting objects through directional windows should not longer be possible.
- fix: Wideswing attacks should no longer make objects flash red that weren't actually hit.

### DIFF
--- a/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
+++ b/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
@@ -170,7 +170,7 @@ public sealed partial class MeleeWeaponSystem : SharedMeleeWeaponSystem
         var targetCoordinates = xform.Coordinates;
         var targetLocalAngle = xform.LocalRotation;
 
-        return Interaction.InRangeUnobstructed(user, target, targetCoordinates, targetLocalAngle, range);
+        return Interaction.InRangeUnobstructed(user, target, targetCoordinates, targetLocalAngle, range, overlapCheck: false);
     }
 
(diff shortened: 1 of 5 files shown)


FairlySadPanda:
Yes, you did watch Liltenhead's video :godo:

VerinSenpai:
Don't suppose this fixes the issue where you could burst into flames if you bumped a directional that was covered by flames (ie a wall for a burn chamber)

SlamBamActionman:
> Yes, you did watch Liltenhead's video :godo:

Can you imagine my groan when I booted up the IDE to fix this after putting it off for weeks, saw a Liltenhead notification on my phone, and then the video included this exact specific bug?

PR #34797, opened 2025-01-31
Ghost orbit jitter fix

Tayrtahn:
## About the PR

Removed the "jittering" visual glitch seen when starting to follow an entity as a ghost. Also made the orbit begin at a "random" point in the circle instead of always starting from the right.

These changes also apply to the orbiting effect of anomaly cores (since they use the same code).

## Why / Balance

The jitter was ugly.

## Technical details

The random number generator is now seeded with the current time before generating the orbit speed and distance values, which means it will produce consistent values throughout prediction (previously it would generate completely new values each predicted frame, which caused the visual jittering).

After fixing this, another issue became much more apparent: there is a significant pause at the start of the orbit before the ghost begins circling. This is due to prediction again: the orbit animation is stopping and restarting on each predicted frame, causing the ghost to repeatedly get sent back to the start of the animation. The orbit "animation" just linearly interpolates a single float value (`OrbitVisualsComponent.Orbit`) from 0 to 1 over `OrbitVisualsComponent.OrbitLength` seconds. To fix this issue, I removed the orbit animation, and instead the system now derives the ghost's position in the orbit based on the current time. (`(float)(_timing.CurTime.TotalSeconds / orbit.OrbitLength) % 1`, which still produces a linear interpolation between 0 and 1, lasting `OrbitVisualsComponent.OrbitLength` seconds). This has the side effect of causing the orbit to start at a (seemingly) random point in the circle (actually based on the current time). This can be returned to the old behavior of always starting at the right (0 degrees) if desired, but I think it's kind of nice.

## Media

https://github.com/user-attachments/assets/077653a0-7cac-4292-bf7e-8bcbad2b6c0a

https://github.com/user-attachments/assets/46397ea7-ab9d-44de-aaa2-f8549a798191

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Ghosts and anomaly cores no longer jitter when starting to follow a target.

### DIFF
--- a/Content.Client/Orbit/OrbitVisualsSystem.cs
+++ b/Content.Client/Orbit/OrbitVisualsSystem.cs
@@ -77,7 +77,8 @@ public sealed class OrbitVisualsSystem : EntitySystem
 
         foreach (var (orbit, sprite) in EntityManager.EntityQuery<OrbitVisualsComponent, SpriteComponent>())
         {
-            var angle = new Angle(Math.PI * 2 * orbit.Orbit);
+            var progress = (float)(_timing.CurTime.TotalSeconds / orbit.OrbitLength) % 1;
+            var angle = new Angle(Math.PI * 2 * progress);
             var vec = angle.RotateVec(new Vector2(orbit.OrbitDistance, 0));
 
             sprite.Rotation = angle;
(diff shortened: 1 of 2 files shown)


slarticodefast:
If we use the CurTime as a basis for the phase of the orbit, wouldn't that sync the position of multiple ghosts instead of giving the current ghost tornado effect?

Tayrtahn:
> If we use the CurTime as a basis for the phase of the orbit, wouldn't that sync the position of multiple ghosts instead of giving the current ghost tornado effect?

Dividing by the randomized orbit length seems to give it enough variety. When I tested with anomaly cores, it looked good, but I haven't tested with multiple ghosts yet.

Tayrtahn:
Yeah, they don't really synchronize like you'd expect:

https://github.com/user-attachments/assets/b6f5b935-8484-4740-932f-2c7299d6506f

slarticodefast (approved):
Blessed PR. It feels so much smoother.

PR #34800, opened 2025-02-01
The "Obvious Fake Disk" mapping item

Compilatron144:
## About the PR
This PR adds an "obvious fake disk", which is a variant of the fake disk which is given a different description by default and lacks the contraband component.

## Why / Balance
When Amber was first released, it included a fake disk which did not have a custom description. This made it indistinguishable to players (and even some admins), which was problematic. Although this has been fixed, there should be some measure to ensure that a mistake like that doesn't happen again. The intention behind the "obvious fake disk" is to provide a fake nuclear authentication disk which is already obviously not a real disk upon inspection - ideal for mappers to include as an easter egg. This also means that fake disks can be given other interactions in the future without such interactions being transferred into a round start item.

## Technical details
Added an "obvious fake disk" which is a regular item that looks like the disk, but has a description mocking the reader and lacks a contraband component.
Added "DO NOT MAP" to the real nuke disk (even if this is included as part of the PR guidelines, best to be certain)
Added "DO NOT MAP" to the fake nuke disk to encourage use of the "obvious fake disk"

## Media

image

## Requirements

## Breaking changes

**Changelog**

Since this change is purely for mappers and won't show up in game until mapped in, I assume this doesn't need a PR.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Misc/dat_fukken_disk.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/dat_fukken_disk.yml
@@ -2,6 +2,7 @@
   name: nuclear authentication disk
   parent: [BaseItem, BaseGrandTheftContraband]
   id: NukeDisk
+  suffix: DO NOT MAP
   description: A nuclear auth disk, capable of arming a nuke if used along with a code. Note from nanotrasen reads "THIS IS YOUR MOST IMPORTANT POSESSION, SECURE DAT FUKKEN DISK!"
   components:
   - type: NukeDisk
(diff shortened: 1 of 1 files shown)


SlamBamActionman:
I can agree that there are a lot more items that should have a "DO NOT MAP" (see: Loop having a CC pen that can edit stamped papers, and like you say Amber with the Uplink fake disk), but I don't think it's necessary to make prototypes for map-specific joke items. Making one is as simple as Rightclick -> Edit Name & Description for mappers. I'd rather not nudge mappers to include fake disks and instead encourage them to come up with their own gimmick items (see: Marathon's galoshesn't)

Compilatron144:
> I can agree that there are a lot more items that should have a "DO NOT MAP" (see: Loop having a CC pen that can edit stamped papers, and like you say Amber with the Uplink fake disk), but I don't think it's necessary to make prototypes for map-specific joke items. Making one is as simple as Rightclick -> Edit Name & Description for mappers. I'd rather not nudge mappers to include fake disks and instead encourage them to come up with their own gimmick items (see: Marathon's galoshesn't)

Fair valid.... yeah I agree with this actually.

Issue #34796, opened 2025-01-31
AI can pet itself


NoElkaTheGod:
## Description
Title. Station AI can pet itself.

**Reproduction**
1. Be AI.
2. Click on your core.
3. You somehow pet yourself, even though you don't have arms.

**Screenshots**
Sorry, but nah.

ArtisticRoomba:
Extremely pressing issue, quick someone call scarky

aox2:
**THIS IS HORRIBLE PLEASE FIX NOW IT RUINS MY IMMERSION!!!**

impubbi:
I have no hands and I must pet

TheGreyDiamond:
So the soultion would be to add arms to the AI core only for petting, right?

PR #34802, opened 2025-02-01
Fix: AI can pet itself (#34796)

olliebdev:
## About the PR

If the player is station AI don't let them pet themselves

## Why / Balance

Issue(#34796)

## Technical details

check if user has component StationAICoreComponent if so don't let them pet themselves.

## Requirements

**Changelog**

:cl:
- fix: AI can pet itself

### DIFF
--- a/Content.Shared/Interaction/InteractionPopupSystem.cs
+++ b/Content.Shared/Interaction/InteractionPopupSystem.cs
@@ -5,6 +5,7 @@ using Content.Shared.Interaction.Events;
 using Content.Shared.Mobs.Components;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.Popups;
+using Content.Shared.Silicons.StationAi;
 using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Network;
@@ -68,6 +69,8 @@ public sealed class InteractionPopupSystem : EntitySystem
         {
             return;
         }
+        if (HasComp<StationAiCoreComponent>(Transform(user).ParentUid))
+            return;
 
         args.Handled = true;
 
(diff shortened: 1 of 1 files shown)


Content.Shared/Interaction/InteractionPopupSystem.cs
```diff
@@ -68,6 +69,8 @@ private void SharedInteract(
         {
             return;
         }
+        if (HasComp<StationAiCoreComponent>(Transform(user).ParentUid))
```
slarticodefast, line 72: I don't like hardcoding this component into this system here. If a fork does not have the AI, then this would break.

slarticodefast (changes requested):

NoElkaTheGod:
The most important bugfix

Content.Shared/Interaction/InteractionPopupSystem.cs
```diff
@@ -69,7 +68,7 @@ private void SharedInteract(
         {
             return;
         }
-        if (HasComp<StationAiCoreComponent>(Transform(user).ParentUid))
+        if (Prototype(Transform(user).ParentUid)?.ID == "PlayerStationAi")
```
slarticodefast, line 71: Sorry, maybe I was unclear. Checking for the prototype is even worse, as prototypes should only be used for spawning entities.
This bug will need a more complex solution. We got different systems handling player interaction, and these raise a number of events which are checking for things like having free hands, being able to do complex interactions, being in range and unobstructed. One of these should prohibit the AI interacting this way somewhere along the event chain from the user's mouse click to the popup appearing. Adding a check for having hands, which should be an argument parameter in `ActivateInWorldEvent` and `InteractHandEvent` might work in this case, but it should be tested if it breaks other interactions of this type.

slarticodefast (changes requested):

slarticodefast:
Closing as derelict. Leave a comment if you come back to this and I will re-open.

Issue #34508, opened 2025-01-19
Guns (ballistics and energy) need a pop-up when empty


K-Dynamic:
## Description

See title. Louder dry-fire sounds were added by #34447, but an empty gun pop-up is also needed for accessibility.

ArtisticRoomba:
Empty gun how so? Like the gun is empty when you pick it up, or when the last round is expended and the gun bolts open?

- We have distinct sounds and a visual queue when a gun runs out of ammo and bolts open, making it unable to be fired.
- Players should just be forced to check if a gun is empty/full when they pick it up by looking at the bullet view or inspecting the gun.

PR #34447, opened 2025-01-15
New dry fire sound

K-Dynamic:
## About the PR

- New dry fire sound

## Why / Balance

Partially fixes #33189.

Uses the much louder dry fire sound from /tg/ station 13 and Delta-V, which audibly alerts both the shooter and nearby players. This reduces reliance on 'counting' bullets or looking at the UI to tell when the gun is empty, which should improve player responsiveness in combat.

## Technical details

ogg changes only

## Media

https://github.com/user-attachments/assets/1c70645c-3f3f-4158-8469-dd8da6f24a95

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Empty guns have a louder dry fire sound (i.e. when the chamber or power cell is empty).

PR #34816, opened 2025-02-02
Adds popup when firing gun while gun has no ammo

Plykiya:
## About the PR

Adds a popup under the cursor when firing a gun that has no ammo remaining

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/34508

## Technical details

Adds an additional localization string for the no ammo left message, shows the no ammo message if there is no other reasoning available

Test video uses the same two guns in the same setting as PR https://github.com/space-wizards/space-station-14/pull/34447

## Media

https://github.com/user-attachments/assets/0c744a6f-fd9b-414c-8f26-420b9be9d285

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: A message now pops up when attempting to fire a gun with no ammo.


slarticodefast (approved):
Looks good to me!

slarticodefast (approved):
I pushed a small improvement.
`IsFirstTimePredicted` is already checked in PopupCursor so it is not needed here.

PR #34824, opened 2025-02-02
[Admin] Omni Accent Smite

ScarKy0:
## About the PR

Please make it stop.
Adds a smite for omni accent, which applies most accents in the game to a player. Also 1/8th chance to add the backwards accent.

## Why / Balance

I got bribed into it :>

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- add: Added the omni-accent smite. It adds most accents onto the target.

### DIFF
--- a/Content.Server/Administration/Systems/AdminVerbSystem.Smites.cs
+++ b/Content.Server/Administration/Systems/AdminVerbSystem.Smites.cs
@@ -892,5 +892,36 @@ public sealed partial class AdminVerbSystem
             Message = string.Join(": ", superslipName, Loc.GetString("admin-smite-super-slip-description"))
         };
         args.Verbs.Add(superslip);
+
+        var omniaccentName = Loc.GetString("admin-smite-omni-accent-name").ToLowerInvariant();
+        Verb omniaccent = new()
+        {
+            Text = omniaccentName,
+            Category = VerbCategory.Smite,
+            Icon = new SpriteSpecifier.Rsi(new("Interface/Actions/voice-mask.rsi"), "icon"),
+            Act = () =>
+            {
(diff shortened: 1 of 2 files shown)


ArtisticRoomba:
probably gould get an `ADMIN:` changelog so that this change can face the players indirectly

ScarKy0:
> probably gould get an ADMIN: changelog so that this change can face the players indirectly

If anyone ever witnesses this in-game I feel sorry for them

Sparlight:
I think you're missing the RussianAccentComponent.

ArtisticRoomba:
i'm not even sure if the target can become even more unintelligible

ScarKy0:
> I think you're missing the RussianAccentComponent.

It's supposed to be bad but not THIS bad :(

Sparlight:
My two cents without being the kind of person that is allowed to go around writing reviews on code like this:  it might be better (though not terribly cost-efficient) to first make an abstract 'AccentComponent' class that all accents inherit from, then use reflection to target all classes that inherit from AccentComponent and create instances of those.  That way, this code doesn't go stale the moment someone adds another accent.

slarticodefast:
Oh god, it's terrible.
Screenshot (361)
Approved.

> My two cents without being the kind of person that is allowed to go around writing reviews on code like this: it might be better (though not terribly cost-efficient) to first make an abstract 'AccentComponent' class that all accents inherit from, then use reflection to target all classes that inherit from AccentComponent and create instances of those. That way, this code doesn't go stale the moment someone adds another accent.

I think it's fine in this case since some accents are incompatible with each other. For example the russian accent would make some of the other ones not apply anymore due to the cyrillic letters.

slarticodefast (approved):

PR #34831, opened 2025-02-02
Revert "Add taped logo back for 10th anniversary"

slarticodefast:
Reverts space-wizards/space-station-14#34486

Anniversary is over. Also we missed doing it in time for the current release, so you all can look at it for two more weeks :)


Iamgoofball:
I kinda like this logo better. It should be a 5% easter egg or something to render if the logo is shown.

VasilisThePikachu:
> I kinda like this logo better, the tape is super funny. It should be a 5% easter egg or something to render if the logo is shown.

The Easter egg idea is welcome, but it's out of scope of this pr.

PR #34838, opened 2025-02-02
Fix solutions flickering when transferring contents

### LINKED ISSUE #32367: Solution transfers mispredict a lot
Get like a bottle and try to fill it up and watch it mispredict hella bad.

Tayrtahn:
## About the PR

Solution containers (beakers, glasses, etc) no longer rapidly flicker through different fill level visuals when transferring their contents.

## Why / Balance

It's a really ugly glitch and it's been around for quite a while now.
Fixes #32367
Fixes https://github.com/space-wizards/space-station-14/issues/38512

## Technical details

`SolutionComponent`s hold a `Solution` class reference in an entity. For networking and clientside prediction, component states are copied and later reapplied. The problem is that the copied data is only a reference to the solution, and not the solution's current state. This means that when solutions are modified during prediction, they are not properly rewound for the next predicted frame.

~To fix this, `SolutionComponent` now no longer uses AutoGenerateComponentState and does state handling the oldschool way. Most importantly, when applying component state, a clone is now made of the `Solution` data, so changes to it do not persist.~
`Solution` now implements ~`ICloneable`~ `IRobustCloneable<Solution>`, which the engine change will detect and properly copy the data. ~This required the existing `Solution.Clone` method to be renamed, since `ICloneable.Clone` must return an `object`. The old method is now `Solution.CloneSolution`.~
It was also necessary to make the `Clone` method copy the `CanReact` field, otherwise the client would mispredict solution reactions in nonreactive containers.

This does produce a bit more junk for the garbage collector to clean up, but `Solution`s aren't huge and it's only clientside on solutions that change within PVS.

## Media

Before:

https://github.com/user-attachments/assets/c903066c-d149-42c7-bc95-4346bf71c40d

After:

https://github.com/user-attachments/assets/0c6d2c64-8fdf-4963-9d3b-c3413909deb0

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Solution containers (beakers, glasses, etc) no longer rapidly flicker through different fill level visuals when transferring their contents.


Booblesnoot42 (approved):
It fixed the issue when I ran it on my computer.

Content.Shared/Chemistry/EntitySystems/SharedSolutionContainerSystem.cs
```diff
@@ -757,6 +760,19 @@ private void OnSolutionShutdown(Entity<SolutionComponent> entity, ref ComponentS
...
+    }
+
+    private void OnSolutionHandleState(Entity<SolutionComponent> entity, ref ComponentHandleState args)
+    {
+        if (args.Current is not SolutionComponentState cast)
+            return;
+
+        entity.Comp.Solution = cast.Solution.Clone();
+    }
+
```
metalgearsloth, line 775: Should the componentnetworkgenerator be updated instead?
 E.g. implement ICloneable here and make it handle it?

metalgearsloth (changes requested):

Content.Shared/Chemistry/EntitySystems/SharedSolutionContainerSystem.cs
Tayrtahn: Oo, fancy! I dunno how many other cases it would be useful for, but that sounds like a pretty nice idea.

slarticodefast:
https://github.com/user-attachments/assets/0009431b-9bde-44fe-ab0a-38505d0c78ee

I'm raising this to critical because this is pretty damn bad and impacting basic gameplay mechanics in addition to the screen flashing and UI glitches.

Tayrtahn:
For the record, I have verified that this change fixes the above issue.

PJB3005 (approved):

PR #34846, opened 2025-02-03
Forced sleep threshold for Chloral Hydrate

DuckManZach:
## About the PR

Chloral Hydrate now has a forced sleep threshold. 

## Why / Balance

Chloral Hydrate has a reputation of being a knockout drug, meant to forcefully put people to sleep (as mentioned in its description). This purpose was effectively replaced with Nocturine, a drug which is effectively antag-only. This PR brings back the knockout effect at a much higher threshold, making it useful again while limiting its effectiveness in combat. Additionally it no longer gives poison damage.

- Forced sleep threshold is at 25u

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Chloral Hydrate now has a forced sleep threshold

### DIFF
--- a/Resources/Prototypes/Reagents/toxins.yml
+++ b/Resources/Prototypes/Reagents/toxins.yml
@@ -69,19 +69,11 @@
         time: 4
         type: Add
         refresh: false
-      - !type:HealthChange
-        conditions:
-        - !type:ReagentThreshold
-          reagent: ChloralHydrate
-          min: 30
-        damage:
-          types:
-            Poison: 1.5
       - !type:GenericStatusEffect
         conditions:
         - !type:ReagentThreshold
           reagent: ChloralHydrate
-          min: 30
+          min: 25
         key: ForcedSleep
         component: ForcedSleeping
         refresh: false
--- a/Resources/Prototypes/Reagents/toxins.yml
+++ b/Resources/Prototypes/Reagents/toxins.yml
@@ -73,10 +73,19 @@
         conditions:
         - !type:ReagentThreshold
           reagent: ChloralHydrate
-          min: 20
+          min: 30
         damage:
           types:
             Poison: 1.5
+      - !type:GenericStatusEffect
+        conditions:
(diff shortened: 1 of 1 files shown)


NoElkaTheGod:
Nice. It was really useless without the forced sleep.

slarticodefast:
Seems reasonable to me at a high threshold. But I would set refresh to true, so that the status effect does not stack to high numbers if you inject a large amount.

NoElkaTheGod:
> Seems reasonable to me at a high threshold. But I would set refresh to true, so that the status effect does not stack to high numbers if you inject a large amount.

Why though? It's hard to use it offensively (hypospray is pretty much the only reliable option, as foam/smoke is easy to avoid, and evertything else isn't enough), abd if you're drinking a full bottle of it... well, you know what you're doing, don't you?

Pumkin69:
25u seems like a great amount imo.

mutanteggs:
Keep in mind that the tranq rounds that are a research have Chloral Hydrate in them too. Not that they’re particularly common, just a consideration to keep in mind as “combat potential”. Could be strong loaded into an Enforcer.

DuckManZach:
> Keep in mind that the tranq rounds that are a research have Chloral Hydrate in them too. Not that they’re particularly common, just a consideration to keep in mind as “combat potential”. Could be strong loaded into an Enforcer.

Thats a good point, with this PR you need to be hit 4 times with tranqs to get knocked out, which each shot giving 7 chloral hydrate. Personally I think this is still pretty balanced, but I can tweak the threshold if people think it's necessary.

jamilnielsen:
checked out chloral hydrate today, yea entirely useless (i was able to knock someone out for 0,2 second)
removing the poison i like.
personally i'd love to see the recipe buffed a bit so its not 4u to 1u but perhaps 1:1 but im not a combat person, i just figured it would be a neat way for a medical borg to stop fights, currently its a tat expensive for that though.

PR #34853, opened 2025-02-03
Add new complex armor component, tweaks for CE hardsuit

kosticia:
## About the PR
Added new component that provides resist to the damage when clothing with this component being worn with specific item

## Media

https://github.com/user-attachments/assets/4291e5f8-b2c8-4eca-b477-fe514df38c13

## Why / Balance
More ways to manage damage resist. Thats cool.

## Requirements


KaiserMaus:
100% a feature of the CE hardsuit, since Chief Engi can work with the singularity without fear, this is also the unofficial goal of the agent when stealing the Plutonium Core, of course you can take other spacesuits or medicine from radiation, but this will not save for long

MozarteanChaos:
actually, wait, this PR just adds resistances to the CE helmet... the helmet didn't have radiation resistance in the first place. are you sure you're PRing this to the right place?

MozarteanChaos:
> actually, wait, this PR just adds resistances to the CE helmet... the helmet didn't have any resistances in the first place, including radiation resistance. are you sure you're PRing this to the right place?

i have been informed that they inherit resistances from ClothingHeadHardsuitBase, which for some reason has 75% radiation resistance by default.
this baffles and vexes me. however, it does also remove the problem i had with this PR.

slarticodefast:
Since it does not change the stats at all I don't really see the need for this. It only adds extra yaml code.

VerinSenpai:
Honestly I feel like the suit should provide like 75 and 100 with the helmet

ArtisticRoomba:
The hardsuit should provide the bulk of the protection and the helmet bring it up to 100- we have flavor text for telling you to equip your helmet before doing stuff like this, however in practice it's actually unnecessary. This change should be made so it lines up with expected behavior.

slarticodefast:
> Honestly I feel like the suit should provide like 75 and 100 with the helmet

Yeah, that makes sense. It should not give you full protection without the helmet.

kosticia:
I have better idea...

HelpTheSteak:
> > Honestly I feel like the suit should provide like 75 and 100 with the helmet
> 
> Yeah, that makes sense. It should not give you full protection without the helmet.


It would not be possible to have fully rad proof armor if you split them up. The resistances don't add up with eachother they take away damage separately so you would still take a very little amount of rad each second

slarticodefast:
> It would not be possible to have fully rad proof armor if you split them up. The resistances don't add up with eachother they take away damage separately so you would still take a very little amount of rad each second

Pretty sure there was a bool changing it to be additive.

deltanedas:
helmet with 0 radiation coefficient makes it radiation proof

kosticia:
Draft because it doesn't work for some reason

kosticia:
Work now. Billions CE will die.

PR #34854, opened 2025-02-03
Update submodule to 243.0.0

metalgearsloth:
:cl:
- tweak: Interfaces will now save their position when closed.


ArtisticRoomba:
Is stable also going to see the RT update to fix the BUI position not being saved, or will have to wait until next release?

PR #34860, opened 2025-02-03
Add sandbox support for rotatable/mirrored tiles

SlamBamActionman:
## About the PR

This PR gives some nice clientchange QoL with the new rotatable/mirrored tiles:
1) When using the sandbox copy option, this ensures it copies rotation & mirroring as well.
2) Variantize command properly keeps mirroring/rotation.
3) Rotated entities are set to the closest cardinal direction when using sandbox copy. 

https://github.com/space-wizards/RobustToolbox/pull/5652

## Why / Balance

Nice QoL for mapping.

## Technical details

Just ensuring Direction & Mirror properties are carried over in the copy/variantize functionality

## Requirements

## Breaking changes

**Changelog**

Nope


Content.Client/Sandbox/SandboxSystem.cs
```diff
@@ -103,6 +103,8 @@ public bool Copy(ICommonSession? session, EntityCoordinates coords, EntityUid ui
                 if (_placement.Eraser)
                     _placement.ToggleEraser();
 
+                _placement.Direction = _transform.GetWorldRotation(uid).GetCardinalDir();
```
whatston3, line 106: This should be relative to the grid if possible, ~~not the map~~.

PR #34863, opened 2025-02-03
Predicted dice rolls

Plykiya:
## About the PR

Updates dice system rolls to be predicted

## Why / Balance

I'm just browsing for things to make predicted at random and noticed that dice rolls were done in Content.System.

## Technical details

The previous implementation said that the client couldn't predict rolling, but it technically can if you pass it a seed. Currently the SharedMaskSystem and the SharedClumsySystem make use of `_timing.CurTick.Value` as their seed, so that's what I did as well.

Updated all functions in SharedDiceSystem to use the `Entity<T>` function format.

## Media

https://github.com/user-attachments/assets/9300930e-2f08-418f-8dfc-e0a64010ad9d

image

## Requirements

## Breaking changes

SharedDiceSystem function parameters changed from `EntityUid uid, DiceComponent comp` format to `Entity<DiceComponent> entity`, use entity for the EntityUid and entity.Comp for the DiceComponent.

SharedDiceSystem's SetCurrentSide() and Roll() visibility changed from public to private.

### DIFF
--- a/Content.Server/Dice/DiceSystem.cs
+++ b/Content.Server/Dice/DiceSystem.cs
@@ -1,28 +1,9 @@
 using Content.Shared.Dice;
-using Content.Shared.Popups;
 using JetBrains.Annotations;
-using Robust.Shared.Audio;
-using Robust.Shared.Audio.Systems;
-using Robust.Shared.Random;
 
 namespace Content.Server.Dice;
 
 [UsedImplicitly]
 public sealed class DiceSystem : SharedDiceSystem
 {
-    [Dependency] private readonly IRobustRandom _random = default!;
-    [Dependency] private readonly SharedAudioSystem _audio = default!;
-    [Dependency] private readonly SharedPopupSystem _popup = default!;
-
-    public override void Roll(EntityUid uid, DiceComponent? die = null)
-    {
-        if (!Resolve(uid, ref die))
-            return;
-
-        var roll = _random.Next(1, die.Sides + 1);
-        SetCurrentSide(uid, roll, die);
-
-        _popup.PopupEntity(Loc.GetString("dice-component-on-roll-land", ("die", uid), ("currentSide", die.CurrentValue)), uid);
-        _audio.PlayPvs(die.Sound, uid);
-    }
 }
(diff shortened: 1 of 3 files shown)


Content.Server/Dice/DiceSystem.cs
```diff
@@ -1,28 +1,9 @@
...
-using Content.Shared.Popups;
 using JetBrains.Annotations;
-using Robust.Shared.Audio;
-using Robust.Shared.Audio.Systems;
-using Robust.Shared.Random;
 
 namespace Content.Server.Dice;
 
 [UsedImplicitly]
 public sealed class DiceSystem : SharedDiceSystem
```
TheShuEd, line 7: you can just delete server side system, and make shared not abstract

Content.Shared/Dice/SharedDiceSystem.cs
```diff
@@ -1,12 +1,18 @@
...
 using Content.Shared.Interaction.Events;
+using Content.Shared.Popups;
 using Content.Shared.Throwing;
-using Robust.Shared.GameStates;
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Timing;
 
 namespace Content.Shared.Dice;
 
 public abstract class SharedDiceSystem : EntitySystem
```
TheShuEd, line 10:
```suggestion
public sealed class SharedDiceSystem : EntitySystem
```

TheShuEd (approved):

Content.Client/Dice/ClientDiceSystem.cs
```diff
@@ -0,0 +1,28 @@
+using Content.Shared.Dice;
```
slarticodefast, line 1: @TheShuEd Shouldn't this file be called DiceSystem.cs and the client system override the shared one?
I'm pretty sure having two systems like this is against the file convention.

slarticodefast (changes requested):

slarticodefast (approved):

slarticodefast:
I changed the file name back to fit our naming convention. Thank you for your contribution.

PR #34867, opened 2025-02-04
Update staging submodule to 243.0.0

metalgearsloth:


Errant-4:
Don't update the submodule on staging unless we have a version that fixes the borg law issue, we are considering reverting instead to version 240.1.2

Errant-4:
We are doing the release, so this is no longer necessary

PR #34868, opened 2025-02-04
Reveal settings for light blur and shadow blur

notquitehadouken:
## About the PR

Added extra options for light blur and soft shadows to the options menu

## Why / Balance

Because all 3 settings were previously all in one setting (lighting quality) which limited what effects you could get because every quality setting other than high lowered the light resolution scale

Also I want more people to discover the magic of turning off light blur

## Media

image

image

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: There are now extra settings for lighting in the graphics tab of options

### DIFF
--- a/Content.Client/Options/UI/Tabs/GraphicsTab.xaml
+++ b/Content.Client/Options/UI/Tabs/GraphicsTab.xaml
@@ -9,9 +9,15 @@
                 <CheckBox Name="VSyncCheckBox" Text="{Loc 'ui-options-vsync'}" />
                 <CheckBox Name="FullscreenCheckBox" Text="{Loc 'ui-options-fullscreen'}" />
 
+                <!-- Lighting -->
+                <Label Text="{Loc 'ui-options-lighting-label'}" StyleClasses="LabelKeyText"/>
+                <ui:OptionDropDown Name="DropDownLightingQuality" Title="{Loc 'ui-options-lighting-quality-label'}" />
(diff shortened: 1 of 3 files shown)


metalgearsloth:
Is blurring the lights even expensive?

notquitehadouken:
> Is blurring the lights even expensive?

lmao i have no idea but i think it looks ugly

metalgearsloth:
That should be done at the discretion of the server and not a client cvar if it has no material performance benefit.

notquitehadouken:
> That should be done at the discretion of the server and not a client cvar if it has no material performance benefit.

i put it in there (and soft shadows) because i think the game looks better with it off

notquitehadouken:
and want to give others the option to see

notquitehadouken:
> That should be done at the discretion of the server and not a client cvar if it has no material performance benefit.

why should a lighting effect with no effect on gamplay be controlled by the server

that's like if you joined a Minecraft server and whether or not you had view bobbing on was up to the server instead of you

notquitehadouken:
ykw i have a better idea

notquitehadouken:
better idea turned out to be worse

notquitehadouken:
By the Way once this is merged (or not) I'm going to add an option for normals in a different PR 

they are separate prs because this isn't related to normals so i don't want it to be put in hold or closed entirely solely on the basis that lighting normals need more review

notquitehadouken:
betterer idea

PR #34878, opened 2025-02-04
Reduces handheld security radios range for picking people's messages up.

Boaz1111:
## About the PR

Handheld security radios are now only able to pick up noise from one tile away, instead of four. Change not applied to salvage radios.
## Why / Balance

This is actually a buff, since using handheld security radios was previously really annoying, since it picks up messages from like.. four tiles away, and that just makes it spam security radio a lot.
## Media

https://github.com/user-attachments/assets/aa3c7765-a1b9-4c8e-bcaf-86b0ff551e65

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Reduced the range at which security radios pick up noise from 4 tiles to 1 tile.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/radio.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/radio.yml
@@ -33,6 +33,7 @@
   components:
   - type: RadioMicrophone
     broadcastChannel: Security
+    listenRange: 1
   - type: RadioSpeaker
     channels:
     - Security


slarticodefast:
Makes it a little inconsistent between radios, but I guess they have different use cases that justify different stats.
I'll leave the PR open a little isntead of directly merging in case anyone else got feedback.

slarticodefast (approved):

Boaz1111:
> Makes it a little inconsistent between radios, but I guess they have different use cases that justify different stats. I'll leave the PR open a little instead of directly merging in case anyone else got feedback.

I can also reduce it on the salv radio, if you'd like?

slarticodefast:
> I can also reduce it on the salv radio, if you'd like?

I kinda like leaving that one as it is, as it allows you to hide it somewhere to listen in on conversations. Since this is also useful for detectives it might be more interesting to add a verb that allows you to set the range of the microphone.

Pumkin69:
thank you less sec radio spam is always appreciated.

PR #33367, opened 2024-11-17
Rename Space Heater to Space HVAC Unit.

ScarKy0:
## About the PR

Changed the name of the Space Heater.

## Why / Balance

The Space Heater apparently also has a cooling option. I only learned of it today. I think the name should portray the fact it can go both ways a bit more.

## Technical details

Just all name changes. No ID touched.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Space Heater renamed to Space HVAC Unit to show it can also cool rooms.

PR #34879, opened 2025-02-04
Changed immovable rod announcement to reflect immobility

Booblesnoot42:
## About the PR
Changes the station announcement for the immovable rod event to better reflect the situation.

## Why / Balance
There was some discussion on Discord that it was confusing for the "immovable rod" to move with "high velocity." The usual explanation is that the rod isn't moving towards the station, but instead the station is moving towards the rod. Due to the existence of station anchors, however, this doesn't make as much sense as it did in ss13. The idea this PR supposes is that both the station and the rod are "anchored" to specific reference frames, but to different ones, leading to a relative velocity between them.

## Technical details
What is C#? :clueless:

## Media
image

## Requirements

## Breaking changes

**Changelog**
Minor change, unnecessary.

### DIFF
--- a/Resources/Locale/en-US/station-events/events/immovable-rod.ftl
+++ b/Resources/Locale/en-US/station-events/events/immovable-rod.ftl
@@ -1,1 +1,1 @@
-station-event-immovable-rod-start-announcement = High velocity unidentified object is on a collision course with the station. Impact imminent.
+station-event-immovable-rod-start-announcement = Scanners have detected an unidentified object anchored to a spatial reference frame with non-zero velocity relative to the station. Impact imminent.


Resources/Locale/en-US/station-events/events/immovable-rod.ftl
```diff
@@ -1 +1 @@
-station-event-immovable-rod-start-announcement = High velocity unidentified object is on a collision course with the station. Impact imminent.
+station-event-immovable-rod-start-announcement = Scanners have detected an unidentified object anchored to a spatial reference frame with non-zero velocity relative to the station. Impact imminent.
```
FairlySadPanda, line 1: "Scanners have detected an unidentified object anchored to a spatial reference frame with a high relative velocity to the station. Impact imminent."

Errant-4:
The idea that the station is moving towards the rod actually makes perfect sense if you consider orbital mechanics. That said, this still falls apart the moment an immovable rod launcher is spawned. A much more logical solution would be if we renamed it to unstoppable rod

slarticodefast:
Due to relativity you can't tell if the station is moving and the rod is standing still or the other way around. That exactly is the joke.
"What happens when an unstoppable force meets an immovable object"
Changing this just makes it boring.

ArtisticRoomba:
Personally I viewed the usage of "immovable" in this context as "unable to be moved beyond its present course".

Unstoppable rod is probably more fitting and intuitive, but to be honest, I think the name should stay the same... it's been like this for a while and we don't really have to rename everything to be "realistic". The game is far from that. Think of it as a classic.

See #33367 for a similar case.

Errant-4:
> Personally I viewed the usage of "immovable" in this context as "unable to be moved beyond its present course".

That's... literally the definition of "unstoppable". Although admittedly, with different reference frames the two ideas are essentially interchangeable

ArtisticRoomba:
> > Personally I viewed the usage of "immovable" in this context as "unable to be moved beyond its present course".
> 
> 
> 
> That's... literally the definition of "unstoppable". Although admittedly, with different reference frames the two ideas are essentially interchangeable
> 
> 

Let me use my madman logic okay (godo)

FairlySadPanda:
Immovable Rod is a DnD magic item and one of the more commonly memed items due to it being a "wait, hang on, what" overpowered nonsense item when thought about scientifically. The Rod is immovable _on some frame of reference_. This is not a problem when the Rod is being used in some DnD game with geocentricity, but the joke is that, if the Rod is set up in space and anchored to some far-distant planet, it essentially functions as a potentially superluminal missile to anything that is not also anchored to that frame of reference.

Booblesnoot42:
> The idea that the station is moving towards the rod actually makes perfect sense if you consider orbital mechanics. That said, this still falls apart the moment an immovable rod launcher is spawned. A much more logical solution would be if we renamed it to unstoppable rod

Considering it's a reference to an item from D&D which has the same name, I think renaming it would weaken the reference to the point it would be better to make it a different item entirely.

Resources/Locale/en-US/station-events/events/immovable-rod.ftl
Booblesnoot42: Changed.

VerinSenpai:
I'm confused. The only thing I see that identifies the rod as immovable is the loc string >.>

Booblesnoot42:
> Due to relativity you can't tell if the station is moving and the rod is standing still or the other way around. That exactly is the joke. "What happens when an unstoppable force meets an immovable object?" Changing this just makes it boring.

Before station anchors were a thing, I'd agree. With the anchors, though, it's weird to have two objects claiming to be immobile that are moving relative to one another, and without an explanation it just seems like a mistake.

Booblesnoot42:
> I'm confused. The only thing I see that identifies the rod as immovable is the loc string >.>

It's an item in-game.
https://github.com/space-wizards/space-station-14/blob/8482767109f6a24f0922b3187c5d2b92478b7fb7/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml#L3

slarticodefast:
> it's weird to have two objects claiming to be immobile that are moving relative to one another

That's exactly what relativity means in physics.
The station anchor prevents acceleration, not movement.

Booblesnoot42:
> > With the anchors, though, it's weird to have two objects claiming to be immobile that are moving relative to one another
> 
> That's exactly what relativity means in physics. You cannot say if object A or object B is the one standing still. The station anchor prevents acceleration, not movement.

I'm still not entirely sure why you are against this change. It does nothing but explicitly state the situation that you are describing.

Errant-4:
It feels way overcomplicated to me, to the point of potentially being confusing. 
Why not just something like "unidentified object with high relative velocity"

slarticodefast:
I asked around and the feedback was that the old message is funnier, so we will keep that one.

PR #34880, opened 2025-02-04
MindRemoveRole refactor

Errant-4:
## About the PR
Reorganised MindRemoveRole and it's overloads. Mind roles can now be removed based on prototypeId

## Why / Balance
Need this for an upcoming PR.
Until now, it was only possible to remove them by specifying a component. Now the same data that can create them can also remove them.

## Technical details
MindRemoveRole can now be called without <T>, instead specifying a protoId as the second parameter.
Both the old and the new version runs their own foreach on the mind's roles, and creates a deletion list. They then call the new MindRemoveRoleDo(), which contains parts of the code that could be shared between the different versions.
The `Found` variable which was used to mark that deletable entities were found has been removed, in favor of simply checking if the Delete list contains any elements.

## Requirements

## Breaking changes
MindTryRemoveRole() was replaced with an overload of MindRemoveRole(). Simply remove "Try" from every existing use, and it should work right away.

**Changelog**

No player or admin facing changes.


slarticodefast:
Who dares summon me?

Content.Shared/Roles/SharedRoleSystem.cs
```diff
@@ -284,25 +283,84 @@ public bool MindRemoveRole<T>(Entity<MindComponent?> mind) where T : IComponent
...
+        return MindRemoveRole<T>((mindId, mind));
+    }
+
+        /// <summary>
+    /// Finds and removes all mind roles of a specific type
+    /// </summary>
+    /// <param name="mind">The mind entity and component</param>
+    /// /// <param name="protoId">The prototype ID of the mind role to be removed</param>
+    /// <returns>True if the role existed and was removed</returns>
+    public bool MindRemoveRole(Entity<MindComponent?> mind, EntProtoId<MindRoleComponent> protoId)
```
slarticodefast, line 327: I'm not sure if removing components depending on a prototype is the right approach. Usually we don't want prototypes to be used for anything other than spawning entities.
What is the usecase you will need this for?

Errant-4: Specifying a temporary mindrole by datafield. 
Specific usecase is making a brain/pAI insertable into a  borg ((whether this is a desired feature is not relevant for the purposes of functionality)). It must get the Silicon mindrole when put into a borg, and lose it when removed. The most straightforward way for this is to specify the prototypeID in yaml, and use it both for adding and removal.
The alternative would be to store some type of "termination identifier" on each MindRoleComponent, then you would have to Index the prototypeID supplied for creation, read it's MindRoleComponent, and search the target's mind role entities for that, and remove. That was how I was first going to do it, but it seems needlessly complicated

slarticodefast: Using prototypes for this looks still a little sus, but the idea makes sense.
We best ask sloth if he can think of a different approach.

Errant-4: My backup idea would be to use Tags, but then I would have to make a bunch of tags, seems like a waste of time and space

Content.Shared/Roles/SharedRoleSystem.cs
```diff
@@ -284,25 +283,84 @@ public bool MindRemoveRole<T>(Entity<MindComponent?> mind) where T : IComponent
...
+                continue;
+            }
+
             if (!HasComp<T>(role))
                 continue;
 
+            delete.Add(role);
+        }
+
+        return MindRemoveRoleDo(mind, delete, "type " + typeof(T).Name);
```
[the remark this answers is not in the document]
Errant-4: I guess that's a fair point, though to my knowledge nothing ever changes or has any reason to change the name of the mind roles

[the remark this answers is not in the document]
Errant-4: (continued: https://discord.com/channels/310555209753690112/900426319433728030/1367146509417578556 )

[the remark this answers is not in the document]
Errant-4: I did the thing, now the component's name is only used if there were no mind role matches so there are no entity names to use (but the log still needs to say what _failed_ to be deleted)

PR #34335, opened 2025-01-09
Slight instrument tweaks

SpaceLizard24:
## About the PR
Put two saxophones in the woodwind crate, removed two saxophones from the brass crate, made the kalimba percussion, changed the price of woodwind instrument crate from 2500 --> 3000, and changed the price of the brass instrument crate from 2500 --> 2000 

Also increased the percussion instrument crate from needing 4 instruments to 7.

## Why / Balance
DESPITE BEING MADE OF BRASS, the saxophone is a woodwind instrument. This is because it uses a reed, and has a system like a flute. The kalimba is a percussion instrument, despite having keys. If the kalimba is a keyed instrument, by that logic so should a glockenspiel and xylophone. 
Because the woodwind instrument crate now has three more instruments than the brass instrument crate, and because the brass instrument crate lost it's two saxophones, the prices changed.

The increased percussion instrument bounty is so that you can't just buy one crate and immediately resell it for a profit. Now you have to buy two crates, like how it was before.e

## Technical details
Changed the number "2000" to "3000" and "2500." May cause the world to explode.

## Media
No media for you. Use your imagination, or listen to saxophone solos with rain in the background.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

🆑 fix: A Nanotrasen jazz enthusiast found out the ATS was packing saxophones into Brass Ensemble Crates. After an angry call, the saxophones are now put into the Woodwind Ensemble crates instead.
🆑 fix: The Kalimba is now a percussion instrument.
🆑 tweak: Increased the cost of the Woodwind Ensemble Crate from 2500 to 3000
🆑 tweak: Decreased the cost of the Brass Ensemble Crate from 2500 to 2000
🆑 tweak: Increased the amount of instruments needed for the "Percussion instrument bounty" from 4 to 7

PR #34851, opened 2025-02-03
LogSystem/SharpSystem replaced by SliceableSystem/ButcherableSystem, etc.

Plykiya:
## About the PR

I wanted to work towards this issue https://github.com/space-wizards/space-station-14/issues/28141 when I realized how deep I had to go to fix this, so I haven't actually touched SliceableFoods yet.

- `LogComponent` and `LogSystem` in Content.Server replaced by more generic `SliceableComponent` and `SliceableSystem` in Content.Shared
- `SharpComponent` replaced by `ToolComponent` and `ToolQualityProto` "Slicing" and all instances of it replaced as such
- `SharpSystem` in Content.Server replaced by `ButcherableSystem` in Content.Shared

LogComponent/System entire purpose was "if sharp object used, spawn these items". Sliceable does the same but in shared. I wanted to merge Sliceablefoods into it, but considering the current amount of changed files, I chose not to do that yet.

SharpComponent itself had comments saying to just make it a tool type. I deleted the SharpComponent and replaced every instance of it by adding the ToolComponent with the "Slicing" quality. When code checked for the presence of the SharpComponent, I checked for the ToolComp and the Slicing quality instead.

SharpSystem, despite the name being quite generic, was used solely for the do-after and verbs related to butchering. I moved this all to the ButcherableSystem in Content.Shared instead, along with any changes required to make it not use the SharpComponent.

## Why / Balance

LogComponent isn't generic enough, and SliceableFood should be merged into the new Sliceable stuff after this PR to fix https://github.com/space-wizards/space-station-14/issues/28141
SharpComponent had a TODO comment saying it didn't need to be a component, could just be a tool quality. So I did that.

## Technical details

- `SharpComponent` deleted, replaced by `ToolComponent` with the "Slicing" `ToolQualityPrototype`. This is most of the yml changes.
- Any checks for `SharpComponent` replaced by a check for `ToolComponent` and using the `SharedToolSystem` to check if the `ToolComponent` has the "Slicing" `ToolQualityPrototype`.
- `SharpSystem` in Content.Server replaced by `ButcherableSystem` in Content.Shared. This makes the butcher verb predicted. Spawning and deletion of entities are only run on the server.
- `SharpSystem` AfterInteractEvent subscription changed to `ButcherableSystem` InteractUsingEvent subscription. Previously, the knife checked if it interacted on a butcherable target. Now, the butcherable target checks if it was interacted with by a slicing tool.
[...]

PR #34883, opened 2025-02-04
Fix/ Improvement for NoCargoBountyArbitrageTest

GansuLalan:
## About the PR

Fix for the cargo arbitrage test regarding bounties as it was not covering many cases
## Why / Balance

Was reading other PRs and saw in #34335 mention of a bounty arbitrage integration test that, considering the prevalence of flipping order for bounties right now, wasn't working. This is an attempt to rectify this though there are concerns regarding runtime as the number of permutations is quite a bit. I don't think it currently covers restock breaking that you can do for the donut bounty and some other methods involving multiple orders.

Now also tweaks the cargo bounties in order to pass this new test
## Technical details

The current code works by spawning in a single box of each product and checking if it fulfils any bounties while costing less, while this catches some cases it fails to account for the idea that you can simply order 2 of a product and flip it for the bounty. This code works in a similar way except it will check with multiple boxes of a product and will skip any irrelevant bounties as a time saving measure. This has increased the run time for the test substantially but not overwhelmingly and helps to spot more cases, there are a lot of cases...

There is probably further room for optimisation if needed and I can continue work if this is desired but the current speed should be sufficient for a PR I hope...

My favourite discovery is that you can fulfil the laser bounty worth 28500 by ordering two firefighter boxes for 3000 as batteries count for this bounty as it lacks a blacklist...
## Requirements

## Breaking changes

None that I am aware of, only potential thing is making a method public since I didn't want to do a copy paste but I fail to see how this would ever cause a problem

**Changelog**

:cl: Gansu
- tweak: Changed certain bounty rewards to prevent the ability to flip products on the ATS for a profit


Plykiya:
yeah I ran into that same kind of silly test fail in https://github.com/space-wizards/space-station-14/pull/34851 regarding plastic knives

ArtisticRoomba:
> My favourite discovery is that you can fulfil the laser bounty worth 28500 by ordering two firefighter boxes for 3000 as batteries count for this bounty as it lacks a blacklist...

you can do what???

deltanedas:
batteries are battery ammo providers so you can spend 6 steel and plastic for it

ArtisticRoomba:
incredible

PR #29544, opened 2024-06-28
Fix air vent and scrubber not requiring power

slarticodefast:
## About the PR
Found this bug while working on #29527

If you build or spawn a gas vent or scrubber outside the range of a LV power network they work without power. This happens because Enabled is true by default and only updated via a PowerChangedEvent, which does not happen without a cable connected.

As far as I am aware other atmos devices do not have this bug.

## Why / Balance
bugfix

## Technical details
true -> false

## Media

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes
I hope I'm not breaking any maps with this.

**Changelog**

- fix: Fix some gas vents and scrubbers working without power.

PR #34887, opened 2025-02-05
Fix air devices ignoring settings after power cycle

### LINKED ISSUE #34852: Air alarm settings cease to control connected devices after a power cycle.
## Description
As the title says. Air alarm settings need to be reselected upon an air alarm losing power. For example, if an air alarm is set to fill when power is lost, when power resumes none of the connected devices will behave as such, even though the alarm is still set to fill.

**Reproduction**
Set an air alarm to fill and observe that its connected vents and scrubbers are respecting the current selection.
Turn off the connected apc so the air alarm shuts down and reenable the apc.
Observe that the connected vents and scrubbers are no longer respected the fill selection.

**Screenshots**
https://github.com/user-attachments/assets/4db885ce-8df0-4ea4-98d3-9870d44197f8

**Additional context**
~Don't mind the room pressure in that video. I wasn't conducing any nefarious experiments. Not at all.~

Checked for an existing issue for this but didn't find one so heres hoping its not a dupe.

VerinSenpai:
## About the PR
Fixes #34852

## Why / Balance
Vents and scrubbers currently revert to their default settings when the power cycles. Alongside being annoying, this means they're no longer in sync with their connected air alarm. For example, if an alarm is set to fill, the vent should be enabled and the scrubber disabled. However after a power cycle, the scrubber will be back online, which at the very least gives the impression the alarm is no longer on fill.

## Technical details
Reworks GasVentPumpSystem and GasVentScrubberSystem to use ApcPowerReceiverComponent to determine power state.

## Media
I don't think this is needed? I can add a video if need be.

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- fix: Air alarm devices now return to their previous settings after a power cycle.


Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
```diff
@@ -11,8 +11,11 @@ namespace Content.Server.Atmos.Piping.Unary.Components
     [RegisterComponent]
     public sealed partial class GasVentPumpComponent : Component
     {
+        [DataField]
+        public bool Powered { get; set; } = false;
+
```
metalgearsloth, line 16: Use the power components please.

Content.Server/Atmos/Piping/Unary/Components/GasVentScrubberComponent.cs
```diff
@@ -10,7 +10,10 @@ namespace Content.Server.Atmos.Piping.Unary.Components
     public sealed partial class GasVentScrubberComponent : Component
     {
         [DataField]
-        public bool Enabled { get; set; } = false;
+        public bool Powered { get; set; } = false;
+
```
metalgearsloth, line 14: As above.

metalgearsloth (dismissed):

Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
VerinSenpai: pretend I'm stupid. don't suppose theres an example of this >.>

VerinSenpai: I'm running myself ragged trying to find an existing powered component. The only thing I'm seeing that has a powered variable is ApcPowerReceiverComponent. Is that what you're referring to? I also noticed that the particle accelerator is enabling/disabling power in a similar manner.

ArtisticRoomba:
crazy first PR verin

VerinSenpai:
> crazy first PR verin

Roomba!! please!! I'm gonna cry.... Metal wants me to use a powered component but I can't find one ;(

Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
ArtisticRoomba: I believe for the most part you can do a trycomp to check power like so.
```C#
TryComp<ApcPowerReceiverComponent>(uid, out var power) && power.Powered
```

You can see an example in GasPressurePumpSystem, where we check if the device is not powered, if so we stop making sound and return early.
```C#
private void OnPumpUpdated(EntityUid uid, GasPressurePumpComponent pump, ref AtmosDeviceUpdateEvent args)
    {
        if (!pump.Enabled
            || (TryComp<ApcPowerReceiverComponent>(uid, out var power) && !power.Powered)
            || !_nodeContainer.TryGetNodes(uid, pump.InletName, pump.OutletName, out PipeNode? inlet, out PipeNode? outlet))
        {
            _ambientSoundSystem.SetAmbience(uid, false);
            return;
        }
```

Here's an example for checking if a scrubber is powered:

```C#
            else if (!scrubber.Enabled || TryComp<ApcPowerReceiverComponent>(uid, out var power) && !power.Powered)
            {
                _ambientSoundSystem.SetAmbience(uid, false);
                _appearance.SetData(uid, ScrubberVisuals.State, ScrubberState.Off, appearance);
            }
```

That might work. Feel free to shoot me if it doesn't sloth (godo)

VerinSenpai: I thought it may be something along those lines but thought that may be a bit bloated. I'll write it up and give it a shot. Thanks Roomba.

VerinSenpai: I think I've got it right? Do I have to use an if statement for the trycomp in OnPowerChanged? That bugs me a tad.

VerinSenpai:
Appears to function as desired. I did spot another issue though while writing this. If a powered air alarm is changed to a different setting while a vent/scrubber is not powered, the state of the vent/scrubber doesn't update.

VerinSenpai:
Gonna correct myself on that. Seems the setting changes, but the appearance doesn't.

ArtisticRoomba:
I think I have a reason as to why that's happening

Content.Server/Atmos/Piping/Unary/EntitySystems/GasVentPumpSystem.cs
```diff
@@ -207,7 +208,9 @@ private void OnAtmosAlarm(EntityUid uid, GasVentPumpComponent component, AtmosAl
 
         private void OnPowerChanged(EntityUid uid, GasVentPumpComponent component, ref PowerChangedEvent args)
         {
-            component.Enabled = args.Powered;
+            if (TryComp<ApcPowerReceiverComponent>(uid, out var power))
+                power.Powered = args.Powered;
```
ArtisticRoomba, line 212: In this diff you nuked component.Enabled, which is used a lot around the system. I also have absolutely no clue why you're setting power.Powered here as well... you're setting power.Powered to the value you get from looking at the PowerChangedEvent.

You're probably best just leaving this as-is. But that's my review. Sloth will probably correct me.

ArtisticRoomba (changes requested):

Content.Server/Atmos/Piping/Unary/EntitySystems/GasVentPumpSystem.cs
VerinSenpai: Right so I'm running through and testing this again. That was a dunder on my part setting power.Powered to what was basically itself. As for enabled however, the issue is that it was reenabling scrubbers and vents that were disabled by air alarm settings during a power cycle (IE powercheckevent) So far vents and scrubbers still appear to function with enabled removed from the OnPowerChange event.

VerinSenpai: Righty, changing to having OnPowerChange only update the state has had no visible impact. Everything still appears to function. I do still however have the issue where the visual state can be incorrect when settings are changed on an air alarm while a device is depowered. Did you have an idea for what was causing that? I was gonna ask if it was because the packet wasn't being sent for a depowered device but given that the settings still seem to take affect, its just visually out of sync.

ArtisticRoomba: Okay, what are you trying to do? You're confusing me.

Right now the first issue is resolved. Vents no longer return to default settings if power is cycled. They'll keep their settings through power cyclings.

Are you trying to make it so that the settings on vents and scrubbers can be changed when the power is off? Because when the power for that device is off, it's not able to be sent data, it disappears off of the device list. Refresh your air alarm when the devices are depowered to see this in action.

VerinSenpai: So I was testing on Reach. On reach, an APC that covers the two scrubbers in the hall running from the north most hall to atmos covers both scrubbers in this hall but does not cover the air alarm they're linked to. This allows the settings on the alarm to be changed while the scrubbers themselves are depowered. It appears that when you change settings on the air alarm and reenable power, the actual function of the devices has changed to the new settings, but the sprite is out of sync. So if you set it to fill while it was off but it was on panic, the sprite will continue to be on panic. Does that make sense?

ArtisticRoomba: Are you sure?

I've been reproducing this on Dev atmos. What I do is set up an air alarm looking at a scrubber and an air vent. I depower the scrubber/vent, change settings, and then re-power. No changes take effect, because they can't possibly receive updates, because they're depowered. I don't know how you're seeing the behavior in your testing.

> So if you set it to fill while it was off but it was on panic, the sprite will continue to be on panic.

The scrubber should still be siphoning. I just tested it and this is what happens.

VerinSenpai: forgive me roomba sensei I know not what I speak. behavior is as expected.

VerinSenpai: question for you. out of scope of this PR probably, but should the resync button on air alarms update the settings on the devices as well? I think that'd be a nice to have, instead of having to "change the setting to something else and then right back" which feels jank.

ArtisticRoomba: It should, yeah. Have fun looking at air alarm code though.

VerinSenpai: Good to mark this as resolved?

ArtisticRoomba: Yeah, you're free to mark this and everything else as resolved if you feel so, and I think you're good to do so IMO.

You'll obviously have to wait for the real review from Sloth to come in, I don't know C# :godo:.

ArtisticRoomba (approved):
LGTM and works as intended.

Content.Server/Atmos/Piping/Unary/EntitySystems/GasVentPumpSystem.cs
VerinSenpai: > It should, yeah. Have fun looking at air alarm code though.

I have infected the air alarm code with my presence. Every day I grow within it. Soon, there we be only Verin.

ArtisticRoomba: Keep in mind that the Resync button is commonly used by elder atmosians to get more data per second on the sensors that it governs, so making resync push a packet to the network overwriting everything may not be optimal at all.

VerinSenpai: I realized this. namely because I do that myself. I was thinking maybe updating the device setting when it connects to the alarm. This would also cause newly built devices to automatically adjust as well. I'm not sure yet though, I'm just toying with ideas.

ArtisticRoomba: This will cause problems during a power cycle. You'll need to make it so that this functionality is disabled when Auto mode is disabled, otherwise you'll overwrite custom tuned sensors when they reconnect after an outage.

VerinSenpai:
drats, I've encountered another issue. The number of devices connected to an air alarm doesn't change when one device gets depowered.

ArtisticRoomba:
Have you pressed the Resync button and observed this behavior? The device list only updates very, very infrequently.

VerinSenpai:
I must be tripping. Hang tight, double checking now.

VerinSenpai:
ignore me, I'm just paranoid

ArtisticRoomba:
I get that you're been working on your frezon setups for the (just merged!) frezon buff, but man, don't get high on your own supply (godo).

Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
```diff
@@ -12,7 +12,7 @@ namespace Content.Server.Atmos.Piping.Unary.Components
     public sealed partial class GasVentPumpComponent : Component
     {
         [ViewVariables(VVAccess.ReadWrite)]
-        public bool Enabled { get; set; } = false;
+        public bool Enabled { get; set; } = true;
```
Partmedia, line 15: Could you explain why the default state was changed?

It seems to me that the meaning of this variable has changed -- which has prompted this flip. Could you please document what the new `Enabled` field means?

Content.Server/Atmos/Piping/Unary/Components/GasVentScrubberComponent.cs
```diff
@@ -10,7 +10,7 @@ namespace Content.Server.Atmos.Piping.Unary.Components
     public sealed partial class GasVentScrubberComponent : Component
     {
         [DataField]
-        public bool Enabled { get; set; } = false;
+        public bool Enabled { get; set; } = true;
```
Partmedia, line 13: Same here.

Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
VerinSenpai: if I recall correctly, the state was being changed by OnPowerChange to true round start. Since we're instead using ApcPowerReceiver to handle the power state, this is no longer the case. Essentially vents and scrubbers will start the shift disabled with this set to false. Given that vents and scrubbers are no longer using enabled to determine if they're powered or not, setting this to true allows the devices to maintain being enabled on round start and when placed but leaves powered to apcpowerreceiever. Hopefully I explained that well enough. I'll test it without this change real quick but I'm pretty sure the result was that vents and scrubbers started offline.

VerinSenpai: Aye, vents and scrubbers start the shift disabled if this is set to false. Enabled is still used by alarm modes to control whether the devices are... enabled, so from what I can tell behavior hasn't changed.
{1713AFC9-E3EA-415B-AE5C-7EA74D9D209E}

VerinSenpai: OH
I should've read the comment fully, my bad

VerinSenpai:
My first ever merge conflict on a project that isn't mine ;(

I'm weeping tears of joy.

Partmedia (approved):

Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
VerinSenpai: ignore that last message, we're no longer setting anything but the visual state in OnPowerChange.

Partmedia (changes requested):
When you fixed conflicts, it looks like you reverted the change to `Content.Server/Atmos/Piping/Unary/EntitySystems/GasVentPumpSystem.cs`. Remember to review the diff text.

VerinSenpai:
a bit on the tired side. heres hoping I didn't screw that up too.

Partmedia (approved):
Looks good, thanks!

Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
```diff
@@ -11,8 +11,12 @@ namespace Content.Server.Atmos.Piping.Unary.Components
     [RegisterComponent]
     public sealed partial class GasVentPumpComponent : Component
     {
+        /// <summary>
+        /// Identifies if the device is enabled by an air alarm. Does not indicate if the device is powered.
+        /// By default, all air vents start enabled, whether linked to an alarm or not.
+        /// </summary>
         [ViewVariables(VVAccess.ReadWrite)]
-        public bool Enabled { get; set; } = false;
+        public bool Enabled { get; set; } = true;
```
slarticodefast, line 19: This will cause them to not need any power when spawned outside the range of an LV cable since the status is never updated.
https://github.com/space-wizards/space-station-14/pull/29544

Partmedia: Per the comment, `Enable` is now a user power switch instead of conveying actually being enabled or not. The code that determines if the powered functionality is run is `if (TryComp<ApcPowerReceiverComponent>(uid, out var power) && !power.Powered || !vent.Enabled)`.

Partmedia: (But a quick in-game test will verify if this is right or not... if someone gets around to doing that, please post your result!)

VerinSenpai: Already on it.

VerinSenpai: (just as soon as this hunk of a game finishes cloning)

Partmedia:
@VerinSenpai I'm forwarding review comments from @metalgearsloth (via maintainer review channel on Discord), and the comment was to consider using `IsPowered()` from `PowerReceiverSystem`.

If you're able to make that revision, great. Otherwise when I get back from travels I'll update and merge. Thanks for your help.

VerinSenpai:
@Partmedia I believe I have successfully not shitcoded the requested changes. Also @slarticodefast I have tested in game (post changes) and the vents and scrubbers power and dont as expected. I can post a video of this if so desired.

VerinSenpai:
https://github.com/user-attachments/assets/8feb40b2-c80a-4aa1-8876-a571dbf60ee0

uploading a video anyway because it makes me feel good.

Partmedia (approved):

VerinSenpai (commented):
This just gave me the scare of a life time.

slarticodefast (approved):
Works as intended. Thank you for your first contribution!

PR #34888, opened 2025-02-05
give security tasers back

Emisse:
As it is now, antags are too easy to murderbone on. Admins ban murderboning antags. Security gets wiped routinely by traitors. This solves all of the above issues by letting security be powerful and forcing antags to have to be more careful/play better. 
I expect this pr to be a shitstorm so please be civil.

🆑 
- tweak: Security starts with tasers again.

PR #34890, opened 2025-02-05
Sec Balance Part 1: Buffs Disabler and Disabler SMG

keronshb:
## About the PR

Disablers:
- Fire Cost reduced to 50
- Projectile speed increased to 35

Disabler SMGs:
- Fire Cost reduced to 25
- Fire Rate increased to 3.5

## Why / Balance

Alternative to https://github.com/space-wizards/space-station-14/pull/34888

Disablers are actually awful. Probably the worst performing I've seen them across Space Station history.

Increasing the projectile speed any higher is an engine limitation. Needs prediction and we might be able to go to 40 projectile speed.

Disabler SMGs still aren't great but are a little better than they were before. They're delicate to balance so they don't become too good.

## Media

https://github.com/user-attachments/assets/08a98aa3-c88b-4925-b325-fba78d0d9b88

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Disabler - decreased fire cost, increased projectile speed.
- tweak: Disabler SMG - decreased fire cost, increased fire rate.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -229,7 +229,7 @@
       fly-by: *flybyfixture
   - type: Ammo
   - type: StaminaDamageOnCollide
-    damage: 30
+    damage: 47
   - type: Projectile
     impactEffect: BulletImpactEffectDisabler
     damage:
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -229,7 +229,7 @@
       fly-by: *flybyfixture
   - type: Ammo
   - type: StaminaDamageOnCollide
-    damage: 47
+    damage: 40
   - type: Projectile
     impactEffect: BulletImpactEffectDisabler
     damage:
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -229,7 +229,7 @@
       fly-by: *flybyfixture
   - type: Ammo
   - type: StaminaDamageOnCollide
-    damage: 35
+    damage: 33
   - type: Projectile
     impactEffect: BulletImpactEffectDisabler
     damage:
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -229,7 +229,7 @@
       fly-by: *flybyfixture
   - type: Ammo
   - type: StaminaDamageOnCollide
-    damage: 40
+    damage: 35
   - type: Projectile
     impactEffect: BulletImpactEffectDisabler
     damage:
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -434,6 +434,7 @@
     price: 100
   - type: Gun
     fireRate: 3
+    projectileSpeed: 35
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/taser2.ogg
   - type: ProjectileBatteryAmmoProvider
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -433,7 +433,7 @@
   - type: StaticPrice
     price: 100
   - type: Gun
-    fireRate: 2.5
+    fireRate: 2
     projectileSpeed: 35 # any higher and this causes issues in lag
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/taser2.ogg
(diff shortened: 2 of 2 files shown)


Pronana:
> I'd put the firing cost to 62.5 for 16 shots. You would almost never have to recharge with 20 shots unless you got EMP'd.

I think giving them more ammo doesn't really matter, the baton already has 20 hits and it puts it in parity with the baton. This also means new security officers are punishes less for being bad at aiming which is good, firerate increase helps with that a lot too.

> I'm also not sure at having them stun people in 3 hits. Two hits deals 94 stam damage, which already slows them down heavily. This also makes them superior to stun batons in most situations.

The damage is a bit intense yeah, I like the disabler downing in four shots. If it's still not good enough I'd rather see the firerate increased to 4.0 or make the disabler hitscan. At most I think it should do equal damage to the baton (35). The baton shouldn't become worse than the disabler for stunning considering you need to be in melee range. 

> This would also incentivise antags to use EMPs, which currently have no counter.

EMP implants have horrible range, and do nothing to protect you from bullets. Plus you only get three uses of them and can't implant yourself twice which does matter. EMPs grenades also don't get bought often because they require a hand to use and take 5 seconds to activate which is more than enough time for a good secoff to stun and cuff you.

ImRenQ:
```
Disabler Bolts now deal 47 stam damage per hit, roughly putting someone down in 3-4 shots
Fire rate increased to 3
```
stunning in a second?
Great!
But when will I finally have a sniper rifle in my pocket instead of a pistol?

Pronana:
> EMPs themselves aren't too bad against unarmoured opponents, but become extremely dangerous against nukies or web vest users.

To me that's fine because Nukies need some way to defend themselves from stuns, and webbed vest gets SHREDDED by heat and has no spacing resistance.

Anyways did some testing on a devmap where I spawned a salvager with the disabler as programmed (except I gave them infinite ammo to compensate for the fact they have no trigger disipline) and the changes are a huge improvment especially the firerate. 

I do think the damage buff is a bit too strong though, it's basically a two shot stun since the slowdown guarantees the third shot will hit you and with the high firerate, it'll hit you quick.

keronshb:
Based on feedback:
Lowered stam damage to 40 per, which requires 4 hits.
Increased projectile speed to 35

https://github.com/user-attachments/assets/90251da1-d7c0-4719-9d11-aed240a6e862

Pronana:
> Based on feedback: Lowered stam damage to 40 per, which requires 4 hits. Increased projectile speed to 35
> Content.Client_jqJn74zAgu.mp4

Good changes, feels good from playing. I'd probably drop the stam damage to 35 though so it's identical to the stun baton. Doesn't change the time it takes to stun at all so really just a consistency change. 

Also, this does minorly affect the Energy Shotgun which now has a worse time to stun by comparison. Should I open a PR to give it some minor buffs in the form of:

Reduced energy shot cost(?)
Narrower spread for disabling mode (?)
Faster projectile speed(?)

Because it shoots several bolts and disables in three shots (1.5 seconds) if you hit all your shots makes it hard to balance effectively.

ImRenQ:
It's still 3 hits...
I don't know what's wrong with the charge status, but even the video shows 3 hits

keronshb:
> > Based on feedback: Lowered stam damage to 40 per, which requires 4 hits. Increased projectile speed to 35
> > Content.Client_jqJn74zAgu.mp4
> 
> Good changes, feels good from playing. I'd probably drop the stam damage to 35 though so it's identical to the stun baton. Doesn't change the time it takes to stun at all so really just a consistency change.
> 
> Also, this does minorly affect the Energy Shotgun which now has a worse time to stun by comparison. Should I open a PR to give it some minor buffs in the form of:
> 
> Reduced energy shot cost(?) Narrower spread for disabling mode (?) Faster projectile speed(?)
> 
> Because it shoots several bolts and disables in three shots (1.5 seconds) if you hit all your shots makes it hard to balance effectively.

Go for it, Spreads and SMGs are notoriously harder to balance.

keronshb:
Oops my math bad.

Pronana:
> I'm really sorry, but... It's still 3 hits

35x3 = 105 
Damn they're right, I actually had the damage set to 30 on my branch and didn't realize. Oops.

keronshb:
Ok so this discovered a bug for the live version.

This is back down to 33 (current default)

2 hits = slowdown as expected
3rd hit = normal speed
4th = stun

Tested on my master branch and replicated there as well.

https://github.com/user-attachments/assets/65294eb3-ae0e-4cdf-b825-b92a0df3708c

DuckManZach:
I don't agree with the 4 hit down, once you are stunned you are completely at the mercy of sec (or whoever is using the disabler) and I think there should be a lot of leeway given when interacting with a ranged stun weapon. 
If you are set on 4 shots, I would at least decrease the ammo to 10. This would make officers more reluctant to start firing off shots carelessly, and stress the importance of aiming.

pheenty:
It IS the four-hit stun currently.
And it's absolutely fine, leave these 30 damage as they are.
Firerate, fire cost and the projectile speed buffs are fine though.
Also the disabler SMG should have like 50 shots, 20 damage, 6 firerate I think

Boaz1111:
CL has a minor issue
image

keronshb:
PR & Video updated to match new stats.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
```diff
@@ -229,7 +229,7 @@
       fly-by: *flybyfixture
   - type: Ammo
   - type: StaminaDamageOnCollide
-    damage: 30
+    damage: 33
```
SlamBamActionman, line 232: Not sure what changing from 30 to 33 actually does, but I guess it doesn't hurt.

SlamBamActionman (approved):
I think these values are reasonable.

gusxyz:
This > tazers

SlamBamActionman:
Merging this now;
Maintainer Emisse has also expressed approval in an internal discussion thread on Discord.

Pumkin69:
good change

Zalycon:
neutral change

ImRenQ:
evil change

PR #34892, opened 2025-02-05
Adds bleed reduction to bruise packs, lowers gauze bleed reduction

### LINKED ISSUE #28831: Excessive brute damage causes bleeding, yet there is no way to stop bleeding caused by brute
## Description
When a biological entity sustains a great amount of brute damage, they will start bleeding. However, the bleeding appears to be impossible to stop without cauterization. (impossible on corpses and risky on crit)

**Reproduction**
1. Hit someone with a blunt object multiple times OR use the 200 dmg stick
2. Try to use gauze to stop bleeding
3. Fail

**Additional context**
None

### LINKED ISSUE #33020: Gauze should not stop all bleeding at once
## Description

A single unit of gauze will stop ALL bleeding no matter how much you actually have. If anything, it should be similar to cauterisation in which it requires multiple uses.

Tourniquets are basically overshadowed by gauze (it doesn't help you can't stack them or obtain them outside of security but that's another issue).

Not sure if the same logic should be applied to sutures as they're much rarer.

**Reproduction**

1. Obtain 1 unit of gauze.
2. Stab yourself multiple times until near crit.
3. Apply gauze.
4. Magically stop all bleeding (you will still die of bloodloss unless you have saline syringes).

**Screenshots**

image

K-Dynamic:
## About the PR

- Bruise pack now reduces 1 bleed* (formerly none)
- Gauze now reduces 3 bleed (formerly 10)

*bleed referring to bleeding after injury, not bloodloss damage

## Why / Balance

Resolves #33020, would have resolved #28831

Bruise packs should reduce bleeding as it heals all three brute damage groups. Bleeding from blunt was unhealable prior to #28831, and is still an issue without access to gauze or sutures.

One gauze is effective at stopping ALL bleeding (bleeding maxes out at 10). Bleed reduction has been lowered to 3.

## Technical details

yml changes only

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: Bruise packs now reduce bleeding by 1 (formerly none)
- tweak: Gauze now reduce bleeding by 3 (formerly 10)

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Medical/healing.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Medical/healing.yml
@@ -279,7 +279,7 @@
       types:
         Slash: -5
         Piercing: -10
-    bloodlossModifier: -5
+    bloodlossModifier: -3
     healingBeginSound:
       path: "/Audio/Items/Medical/brutepack_begin.ogg"
     healingEndSound:
--- a/Resources/Prototypes/Entities/Objects/Specific/Medical/healing.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Medical/healing.yml
@@ -124,6 +124,7 @@
     damage:
       groups:
         Brute: -15 # 5 for each type in the group
+    bloodlossModifier: -1
     healingBeginSound:
       path: "/Audio/Items/Medical/brutepack_begin.ogg"
     healingEndSound:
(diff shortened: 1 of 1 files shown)


pheenty:
1 and 3 is too little, I'd make it 2.5 and 5 at the very least.
Also that means that skeletons wouldn't be able to use bruise packs anymore, which is bad

Pumkin69:
i would make gauze 5 so it takes 2 uses of gauze to stop a max bleed which makes logical sense to me.

PR #34893, opened 2025-02-05
Reduces Viper and echis fire rate down to 4

keronshb:
## About the PR

Viper rate 6 > 4
echis rate 5 > 4

## Why / Balance

The current full auto fire rate is ridiculous for a pistol and can down a full armor wearing sec relatively fast. This gives sec a little bit more of a chance.

## Media

https://github.com/user-attachments/assets/04a9e872-6b2f-47fd-a812-00704ee44b15

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Vesper fire rate lowered
- tweak: echis fire rate lowered

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Pistols/pistols.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Pistols/pistols.yml
@@ -74,6 +74,17 @@
   id: WeaponPistolViper
   description: A small, easily concealable, but somewhat underpowered gun. Retrofitted with a fully automatic receiver. Uses .35 auto ammo.
   components:
+  - type: Gun
+    fireRate: 4
+    selectedMode: SemiAuto
+    availableModes:
+    - SemiAuto
+    - FullAuto
+    soundGunshot:
+      path: /Audio/Weapons/Guns/Gunshots/pistol.ogg
+  - type: ChamberMagazineAmmoProvider
+    soundRack:
+      path: /Audio/Weapons/Guns/Cock/pistol_cock.ogg
   - type: Sprite
     sprite: Objects/Weapons/Guns/Pistols/viper.rsi
   - type: ItemSlots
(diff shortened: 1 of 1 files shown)


1TALERD:
Why not set the rate of fire to 5, like the MK58? Why make it worse than a standard pistol

keronshb:
> Why not set the rate of fire to 5, like the MK58? Why make it worse than a standard pistol

Reason being it and the Echis are the only two pistols that have an auto fire mode and we don't have a way (right now) to separate Semi Auto and Auto Fire firing speeds.

FairlySadPanda:
1. This risks being a disordered set of simultaneous PRs and it feels like an issue - or better yet, a doc hub that doesn't require PRs to edit - would be the best way to coordinate.
2. This is yet more stuff that I think needs test server balance passes rather than allowing untested balance modifications ship to stable.

keronshb:
> 1. This risks being a disordered set of simultaneous PRs and it feels like an issue - or better yet, a doc hub that doesn't require PRs to edit - would be the best way to coordinate on sec balance.
> 
>     2. This is yet more stuff that I think needs test server balance passes rather than allowing untested balance modifications ship to stable.

Honestly the real fix to this is to adjust fire rate independent of mode. Then this can stay at 6 but auto fire goes down to 4.

Everturning:
why is there all of a sudden a bunch of PRs specifically for the benefit of security and the detriment of everyone else how fucking stupid does the average secoff player appear

TurboTrackerss14:
>     2. This is yet more stuff that I think needs test server balance passes rather than allowing untested balance modifications ship to stable.

We test on prod here OORAH

keronshb:
> Why the syndicate pistol should be worse than the most basic pistol in the game? There's nothing wrong with 6,0 RoF, it should be better, as intended

Remove auto fire and the fire rate can stay the same.
Keep auto fire and it gets removed.
Introduce a separate system where you can adjust fire rate for each fire rate mode.

Pick one and PR it please.

PR #34898, opened 2025-02-05
Energy Shotgun buffs with #34890 in mind

Pronana:
## About the PR
#34890 buffed disablers to have a faster firerate, faster projectiles and more ammo capacity, which means the energy shotgun (which inherits its disabler projectile from the DisablerSMG) needs to be tweaked by comparison. These changes go a bit beyond the scope that to make the energy shotgun a bit more consistent in combat. No damage values have been tweaked, purely energy consumption and spread.

## Why / Balance
To bring parity with disabler buffs and to also make the HoS shotgun perform a bit more like other shotguns.

The disabler now (if the PR gets merged) has 20 shots meaning if you hit all four shots required to stun, you have enough bolts to stun about 5 people. This makes the energy shotgun worse by comparison as it has less ammo capacity for its disabler shots and takes longer to disable a target. 

To compensate, I've decreased ammo consumption so now it holds 20 disabler shots just like the new disabler. I've also tightened the spread a bit on the shotgun overall.

In terms of ranges I opted to try and make it two tiles worse than a shotgun alternative because heat is a really good damage type so:

2 tiles optimal range for wide (Just barely out of melee reach) and 1 tile worse than an enforcer with regular shells

6 tiles optimal range for narrow (The same as disabling was before) which is 2 tiles worse than the enforcer

7 tiles optimal range for disabling

At standard shell range for the enforcer, the narrow barrage is slightly better than wide with wide barely missing four projectiles hitting.

## Technical details
- energy consumption of energy shotgun disabler shots reduced from 120 -> 80
- spread of wide barrage reduced from 30 -> 25 making it usable just barely outside of melee range and better than narrow up to enforcer range
- spread of narrow barrage has been reduced from 10 -> 9 the previous range of the disabling barrage
- spread of disabling barrage has been reduced from 9 -> 8 
- updated disabler, disabler SMG, and disabler projectile to match that of #34890

## Requirements

## Breaking changes
None I am aware of

**Changelog**

:cl:
- tweak: energy shotgun made more accurate, and halved energy consumption of disabler bolts
-->


keronshb:
This will be reviewed after the initial disabler PR is added and merged, So we don't have too many changes at once and make sure that balance is fine with the initial round.

Pronana:
> This will be reviewed after the initial disabler PR is added and merged, So we don't have too many changes at once and make sure that balance is fine with the initial round.

Works for me also makes it easier to avoid merge conflicts

PR #34903, opened 2025-02-05
Add attack delay to weapon throwing

SlamBamActionman:
## About the PR

This PR restricts melee throwing weapons after using them to melee attack, requiring the user to wait out the melee cooldown before being able to throw. 

Additionally, if a stamina damage melee weapon is thrown, the throw cooldown is equivalent to the melee swing delay (no 1 second stuns using throw keybinds).

Unarmed attacks are exempted from this.

## Why / Balance

Doing a melee swing and then throwing is a technique most commonly seen used in conjunction with stunbatons, as it reduces the normal 2 second stamcrit time to just 1 second (hit > wait > hit > throw). A 1 second stamcrit is of course too short, so this PR fixes that.

It was kinda annoying to not be able to throw immediately upon missing an unarmed attack, since you might've been trying to pick up the item and misclicked, so it makes an exception for that. 

## Technical details

At first I just checked the melee swing timer but that ends up being set when you pick up an item, so it added an annoying delay to picking things up from the ground and immediately throwing, which can be important for transporting stuff. 

So instead we're subscribing to the `MeleeAttackEvent` and that sets the throw timer instead. 

Additionally for `StaminaDamageOnCollideComponent` we check `ThrownEvent` as well.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Melee weapons with throwing attributes can no longer be thrown immediately after swinging.


Content.Server/Hands/Systems/HandsSystem.cs
```diff
@@ -228,6 +232,22 @@ hands.ActiveHandEntity is not { } throwEnt ||
             return true;
         }
 
+        /// <summary>
+        /// Fires when a melee weapon is swung and delays the throw timer.
+        /// Makes an exception for when the entity is the weapon itself, i.e. unarmed attacks.
+        /// </summary>
+        private void OnMeleeAttack(Entity<HandsComponent> ent, ref MeleeAttackEvent args)
```
slarticodefast, line 239: this should be in shared, even if throwing is not predicted yet

SlamBamActionman: Sharing is caring

slarticodefast:
A problem with this I can see is the lack of communication to the player, which makes it feel like an unintended bug. You press the button and sometimes nothing happens, depending on your exact timing for the throw (and I imagine on a live server on your ping as well, due to throwing not being predicted).

ArtisticRoomba:
> A problem with this I can see is the lack of communication to the player, which makes it feel like an unintended bug. You press the button and sometimes nothing happens, depending on your exact timing for the throw (and I imagine on a live server on your ping as well, due to throwing not being predicted), especially since this new cooldown is very short.

I'm sure a simple "You can't throw this item yet!" or something along those lines would tell the player that the block is intended.

slarticodefast:
Having the melee weapon cooldown show up as a doafter circle in the hand slot would work as well, but I think the cooldown is currently using its own implementation for that.

SlamBamActionman:
It's worth noting there are fairly few weapons where you'd consider throwing after hitting, so any overarching attack cooldown changes are probably not necessary.

I will look into improving the UX though!

FairlySadPanda:
The do-after is representing the amount of time it really takes to complete an attack. So in-universe the reason you can't throw the weapon should be because you're still mid-swing.

Realistically a throw is also an attack, so it makes sense without additional UX, IMO.

Content.Shared/Throwing/ThrowingSystem.cs
```diff
@@ -170,7 +170,7 @@ public void TryThrow(EntityUid uid,
 
         ThrowingAngleComponent? throwingAngle = null;
 
-        // Give it a l'il spin.
+        // Give it a li'l spin.
```
Tayrtahn, line 173: Absolutely vital change! 😄

Tayrtahn (approved):
Code looks good; change makes sense.

UX improvement would be nice. Could be added here or separately.

SlamBamActionman:
I have looked into this further and there are some minor not-so-good feeling stuff with it. E.g. if you swing a crowbar there is a 1 second window before you can throw it, which just doesn't *feel* good. I tried a pop-up but that honestly didn't seem to solve anything. It's a crowbar, why shouldn't I be able to just throw it?

So I am pivoting with the code. Instead of having a general `OnMeleeAttack` event in the handcode, I am moving it to any component that should block throwing. This seems to be `DamageOtherOnHit` and `StaminaDamageOnCollide`. Still testing it but theoretically it would mean only objects where throwing after hitting matters have it blocked.

Then I'll add a general `You are not ready to throw!` pop-up to a failed throw caused by the throw cooldown, which will pop up either after a relevant melee swing, or if you try to spam the throw button (i.e. the normal throw cooldown).

SlamBamActionman:
It was a wee bit trickier than I would've liked (why is `DamageOnCollide` a server-only component?) but I settled on a solution that should be future-proof; `ResetThrowCooldown` can be used by other systems to reset/set a new throw cooldown, and the relevant systems do that now. Also, pop-up!

Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
```diff
@@ -213,4 +216,23 @@ private void HandleExamined(EntityUid examinedUid, HandsComponent handsComp, Exa
             args.PushMarkup(Loc.GetString(locKey, locUser, locItems));
         }
     }
+
+    /// <summary>
+    /// Resets the throw cooldown to allow for throwing, and optionally sets a new throw tim if given.
```
Tayrtahn, line 221:
```suggestion
    /// Resets the throw cooldown to allow for throwing, and optionally sets a new throw time if given.
```

Tayrtahn (approved):
lgtm

SlamBamActionman:
Hello! Please update this PR to resolve merge conflicts!

SlamBamActionman:
Okay, I'll try to get around to it sometime this week hopefully!

Errant-4:
But nobody came

lexaSvarshik:
bruh

Resources/Locale/en-US/hands/hands-system.ftl
```diff
@@ -4,3 +4,5 @@ comp-hands-examine-empty = { CAPITALIZE(SUBJECT($user)) } { CONJUGATE-BE($user)
 comp-hands-examine-wrapper = { INDEFINITE($item) } [color=paleturquoise]{$item}[/color]
 
 hands-system-blocked-by = Blocked by
+
+hands-system-throwing-cooldown = You are not ready to throw yet!
```
ScarKy0, line 8: My teammates at the start of the match in Deadlock.

Content.Shared/Damage/Systems/SharedDamageOtherOnHitSystem.cs
```diff
@@ -29,4 +34,17 @@ private void OnAttemptPacifiedThrow(Entity<DamageOtherOnHitComponent> ent, ref A
...
+    /// </summary>
+    [SubscribeLocalEvent]
+    private void OnMeleeCollideHit(Entity<DamageOtherOnHitComponent> ent, ref MeleeHitEvent args)
+    {
+        if (!TryComp<HandsComponent>(args.User, out var hands) || !TryComp<MeleeWeaponComponent>(ent.Owner, out var melee))
+            return;
+
+        if (hands.NextThrowTime < melee.NextAttack)
+            _hands.ResetThrowCooldown(args.User, hands, melee.NextAttack);
+    }
```
ScarKy0: Wouldn't it be better as a general thing where you cannot throw after doing a melee attack instead of being part of DamageOtherOnHit?

Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
```diff
@@ -209,4 +212,22 @@ private void HandleExamined(EntityUid examinedUid, HandsComponent handsComp, Exa
...
     }
+
+    /// <summary>
+    /// Resets the throw cooldown to allow for throwing, and optionally sets a new throw time if given.
+    /// </summary>
+    /// <param name="customTime">If set, this will be the new NextThrowTime.</param>
+    public void ResetThrowCooldown(EntityUid uid, HandsComponent? handsComp, TimeSpan? customTime = null)
+    {
+        if (!Resolve(uid, ref handsComp))
+            return;
```
ScarKy0, line 223:
```suggestion
    public void ResetThrowCooldown(Entity<HandsComponent?> ent, TimeSpan? customTime = null)
    {
        if (!Resolve(ent.Owner, ref ent.Comp, false))
            return;
```

will need the rest of the function updated for `ent.Comp` too

Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
```diff
@@ -1,11 +1,14 @@
...
+using Content.Shared.Damage.Components;
 using Content.Shared.Examine;
 using Content.Shared.Hands.Components;
 using Content.Shared.IdentityManagement;
 using Content.Shared.Input;
 using Content.Shared.Interaction;
 using Content.Shared.Inventory.VirtualItem;
 using Content.Shared.Localizations;
+using Content.Shared.Weapons.Melee;
+using Content.Shared.Weapons.Melee.Events;
```
ScarKy0, line 11: unused

Content.Shared/Hands/EntitySystems/SharedHandsSystem.Interactions.cs
```diff
@@ -1,11 +1,14 @@
 using System.Linq;
+using Content.Shared.Damage.Components;
```
ScarKy0, line 2: unused

Content.Shared/Damage/Systems/SharedStaminaSystem.cs
```diff
@@ -233,6 +236,27 @@ private void OnThrowHit(EntityUid uid, StaminaDamageOnCollideComponent component
...
+    [SubscribeLocalEvent]
+    private void OnThrown(Entity<StaminaDamageOnCollideComponent> ent, ref ThrownEvent args)
+    {
+        if (!TryComp<HandsComponent>(args.User, out var hands) || !TryComp<MeleeWeaponComponent>(ent.Owner, out var melee))
+            return;
+
+        var throwCooldown = Timing.CurTime + TimeSpan.FromSeconds(1 / melee.AttackRate);
+        if (hands.NextThrowTime < throwCooldown)
+            _hands.ResetThrowCooldown(args.User.Value, hands, throwCooldown);
+    }
```
ScarKy0, line 258: Wouldn't it be better to be a general MeleeWeapon thing instead of only applying to specific weapons?

ScarKy0 (changes requested):

Resources/Locale/en-US/hands/hands-system.ftl
sleepyyapril: true

PR #34912, opened 2025-02-06
Fix references to obsolete TryGetContainingContainer override

TemporalOroboros:
## About the PR

Changed two calls to `TryGetContainingContainer` to be compatible with the `Entity<T>` override.

## Why / Balance

It decreases the number of compile warnings that we get spammed with when we compile.

## Technical details

Changed one call to explicitly use the `.Owner` for the previously used entity as it can be implicitly converted to the required `Entity<T>` and the entity it was already passing could not.
Changed the other call to pass the correct `Entity<T>` instead of passing the uid and components separately.

## Requirements


slarticodefast (approved):
Thanks for the cleanup!

PR #34921, opened 2025-02-06
Fix IDs without station records not being able to have nonstandard contents

### LINKED ISSUE #33738: IDs without Station Record Keys cannot have nonstandard contents
## Description
IDs without Station Record Keys cannot have nonstandard contents, eg. a `CMO` prototype, added `theater` access, and a title like `CMO (Visiting)`.
The modified title and modified access work fine, but the job prototype is reverted to `Passenger`, meaning the icon will always be the `P`.
Functionally, this isn't very disruptive, but the number of times I've given out IDs as HoP and people get harassed by sec for having a `P` icon is at least two.

**Reproduction**
* wear something like a sechud which shows ID icons
* have ID computer
* have replacement passenger ID (eg. vended from HoP ID stuff vendor)
* insert captain ID in id computer, then replacement passenger ID
* select new job for the passenger ID, eg. `Botanist`
  * *BUG:* Note that the UI bugs and immediately reverts the job to `Passenger`
* eject the passenger ID
  * Note that the icon matches what was selected, eg. `Botanist`
* insert the passenger ID again
* modify title, name, or access
* eject the passenger ID
  * *BUG:* Note that the icon has reverted to `P` for passenger, the job proto the UI reverted to

At this point, if access or a title that's different from what's associated with the desired job proto is configured on the card, attempting to fix the job icon by selecting a job proto will overwrite the desired title/access, leaving us in the situation of "IDs without Station Record Keys cannot have nonstandard contents".

**Screenshots**

https://github.com/user-attachments/assets/fb1a78d0-cf25-4cb8-8bc8-f802c445e405

* Jobspawn / Roundstart ID is modifiable just fine
* Blank passenger ID has modification issues:
  * Change just name, title access -- works fine
  * Changing the job prototype, the UI immediately reverts the selection to `Passenger`, but if you eject the card, the icon matches what was selected (Botanist in this case)
  * Change anything else on the card after changing the prototype, and the `Passenger` job prototype is applied for real (the icon reverts to `P`)

**Additional context**
I looked into the implementation of the ID console stuff, and it seems like the discrepancy between jobspawn/roundstart IDs working fine and replacement IDs not working comes down to the former having a Station Record Key and associated station record. When the job on an ID is modified, the console tries to update the associated station record, and it sort of "back-enforces" the record's content onto the console's UI. In the case that there's no associated station record, however (like when you're working with a freshly vended blank ID), there's no station record to look up, so it ends up "back-enforcing" to a fallback job, `"passenger"`, no matter what is selected. This one thing completely explains all of the behavior described.

As for fixing, well, it doesn't seem easy. It's not as trivial as "just put a station record key onto blank cards" because that'd have a lot of additional ramifications. It might be as simple as doing something different in the case that the ID card doesn't have a station record key on it besides "back-enforcing" the UI to use `Passenger`, but I am not sure what kind of effects that may have on other things.

CheesePlated:
## About the PR

The ID card computer will now not reset to passenger constantly when editing an ID without an associated station record

## Why / Balance

fixes #33738

## Technical details

added a JobPrototype field to the IdCardComponent which is null when there's a valid station record for that card

## Media

https://github.com/user-attachments/assets/1a594ab8-891d-411e-bb85-f9756336b0d5

## Requirements

## Breaking changes

**Changelog**

:cl: CheesePlated
- fix: The ID card console now works properly when setting job presets for replacement IDs

### DIFF
--- a/Content.Client/Access/UI/IdCardConsoleWindow.xaml.cs
+++ b/Content.Client/Access/UI/IdCardConsoleWindow.xaml.cs
@@ -174,7 +174,8 @@ namespace Content.Client.Access.UI
                                        new List<ProtoId<AccessLevelPrototype>>());
 
             var jobIndex = _jobPrototypeIds.IndexOf(state.TargetIdJobPrototype);
-            // If the job index is < 0 that means they don't have a job registered in the station records.
+            // If the job index is < 0 that means they don't have a job registered in the station records
+            // or the IdCardComponent's JobPrototype field.
             // For example, a new ID from a box would have no job index.
             if (jobIndex < 0)
             {
--- a/Content.Server/Access/Systems/IdCardConsoleSystem.cs
+++ b/Content.Server/Access/Systems/IdCardConsoleSystem.cs
@@ -83,9 +83,9 @@ public sealed class IdCardConsoleSystem : SharedIdCardConsoleSystem
             var targetIdComponent = EntityManager.GetComponent<IdCardComponent>(targetId);
             var targetAccessComponent = EntityManager.GetComponent<AccessComponent>(targetId);
 
-            var jobProto = new ProtoId<AccessLevelPrototype>(string.Empty);
+            var jobProto = targetIdComponent.JobPrototype ?? new ProtoId<AccessLevelPrototype>(string.Empty);
             if (TryComp<StationRecordKeyStorageComponent>(targetId, out var keyStorage)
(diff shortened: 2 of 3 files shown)


southbridge-fur:
instamerge for all the HoP players out there

BackeTako:
As a avid HoP main I need this, this bug is just the most annoying thing ever

MilonPL (approved):

PR #34922, opened 2025-02-06
cleanup not needed locale string

lzk228:
just removed unneeded locale
color still working
image

### DIFF
--- a/Resources/Prototypes/Actions/types.yml
+++ b/Resources/Prototypes/Actions/types.yml
@@ -331,8 +331,8 @@
 
 - type: entity
   id: FakeMindShieldToggleAction
-  name: action-name-toggle-fake-mindshield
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/Actions/types.yml
```diff
@@ -331,8 +331,8 @@
 
 - type: entity
   id: FakeMindShieldToggleAction
-  name: action-name-toggle-fake-mindshield
-  description: action-description-toggle-fake-mindshield
+  name: '[color=green]Toggle Fake Mindshield[/color]'
+  description: Turn the Fake Mindshield implants transmission on/off
```
slarticodefast, line 335:
```suggestion
  description: Turn the Fake Mindshield implant's transmission on/off
```

beck-thompson (approved):
:+1: Tested and it works!

PR #33493, opened 2024-11-23
Make storage implant drop items on gibbing

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
No more items round removing using storage implant. Resolves #25524.

## Media

https://github.com/user-attachments/assets/02375f14-9027-488f-b3e2-1ec0b0a4dbab

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Now items from the storage implant drop on the floor upon gibbing of the owner.

PR #34923, opened 2025-02-06
HOTFIX: fix minibomb implant and syndicats not exploding

slarticodefast:
## About the PR
See title.

## Technical details
https://github.com/space-wizards/space-station-14/pull/33493 was a bugfix that intended to drop storage implant contents when being gibbed instead of deleting them. It did so by deleting all implants, which caused the implanter to drop its storage. However, this also caused the microbomb implant to be deleted before exploding because it first gibs the user.
We solve this by only emptying the storage implant on gibbing.
It is still deleted when the user's body is deleted.

Also added readonly VV access to the ImplantedComponent because that annoyed me every time I had to debug implants.
(probably should be a DataField, but that is out of scope)

## Media

https://github.com/user-attachments/assets/d16b0ed7-58e1-4345-80d1-ccaf8307fec7

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- fix: Fixed microbomb implants and syndiecats not exploding.

### DIFF
--- a/Content.Server/Implants/ImplantedSystem.cs
+++ b/Content.Server/Implants/ImplantedSystem.cs
@@ -1,5 +1,6 @@
 using Content.Server.Body.Components;
 using Content.Shared.Implants.Components;
+using Content.Shared.Storage;
 using Robust.Shared.Containers;
 
 namespace Content.Server.Implants;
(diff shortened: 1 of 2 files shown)


slarticodefast:
We should fix the template to conform with the guidelines :)

deltanedas:
shouldnt be datafield since its a container which gets serialized in container container

MilonPL (approved):
Simple enough, looks good.

Issue #33596, opened 2024-11-27
Winter Coats continue to kill you with heat damage long after a fire is extinguished


superjj18:
## Description

If you ignite while wearing a winter coat, even if you extinguish the flame you will still continue to take large amounts of heat damage due to the coat preventing you from cooling down.

I would think clothing would protect you from burning rather than being a lethal combination.  It's unintuitive and its heat retention should only turn on when you're below a temperature that hurts the player.

**Reproduction**

>wear winter coat

>burn(in my example I touched the corner of an atmos burn chamber)

>immediately take a few seconds to extinguish

>continue to cook due to coat heat retention

>crit 

>continue to take heat damage 

>reach medbay with 800 burn damage for a fire that was extinguished minutes ago

Errant-4:
The same thing happens if you get into a spaced corridor, start freezing to death, and then put on a winter coat in an attempt to stave off further cold damage

Aisu9:
no issues. winter coat now make it easier to gain temperature and harder to lose temperature, they no longer modifier temperature thresholds

MozarteanChaos:
This isn't super intuitive, yeah. I've seen a lot of people ash - or get very close to ashing - because nobody around them connected the dots that their winter coat might be preventing them from cooling down.
Maybe fully removing this functionality would be a bit extreme - in part because it'd just make them a safe way to avoid getting cold, instead of a potential risk in event of a gas fire - but it could benefit from being toned down a little. It feels odd that a single, perfectly normal coat can make the difference between being readily revivable, and fully ashing.
I mean, grain of salt since I'm not a contrib, but that's my opinion on the matter.

superjj18:
> no issues. winter coat now make it easier to gain temperature and harder to lose temperature, they no longer modifier temperature thresholds

that’s not how physics work

Typically things stop burning when the fire gets put out and typically fires burn you from the outside not the inside so insulation should reduce burn damage not increase and preserve it

Aisu9:
> > no issues. winter coat now make it easier to gain temperature and harder to lose temperature, they no longer modifier temperature thresholds
> 
> that’s not how physics work
> 
> Typically things stop burning when the fire gets put out and typically fires burn you from the outside not the inside so insulation should reduce burn damage not increase and preserve it

[...]

PR #34929, opened 2025-02-06
Chameleon vests now act like winter coats

beck-thompson:
## About the PR

Title! I gave chameleon vests the same stats as winter coats. They now protect you from the cold, and you get hot slightly faster. I also gave them the minor protection (10% heat, 5% slash) protection.

## Why / Balance

Chameleon clothing isn't used much in general, and on certain maps (Glacier for example but that's Delta), not having the cold protection is actually really annoying.

## Technical details

I mostly just added a new field to the armor component. I didn't use events because of how the armor component works (I'd have to change the event and also think how I'd want it to interact with explosion resistance). I think this is probably the cleanest solution :thinking:

I'm also not 100% sure I should do it this way or parent from the original winter coat. I'd need to rearrange a few things but maybe its worth it?

## Media

https://github.com/user-attachments/assets/2843ce59-e0e0-474c-86e1-212965aa4625

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Chameleon vests now have the same stats as winter coats (Cold protection, minor heat increase, 5% slash resistance, 10% heat resistance)

### DIFF
--- a/Content.Shared/Armor/ArmorComponent.cs
+++ b/Content.Shared/Armor/ArmorComponent.cs
@@ -22,6 +22,12 @@ public sealed partial class ArmorComponent : Component
     /// </summary>
     [DataField]
     public float PriceMultiplier = 1;
+
+    /// <summary>
+    /// If true, you can examine the armor to see the protection. If false, the verb won't appear.
+    /// </summary>
+    [DataField]
+    public bool ShowArmorOnExamine = true;
 }
 
 /// <summary>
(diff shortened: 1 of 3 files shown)


superjj18:
Fire is now the anti-chameleon tool since winter coats round remove people who catch fire and don't take off their coat.

If a person with a Cham vest catches fire they either need to break their disguise before they go crit or get ashed by heat.

#33596

beck-thompson:
> Fire is now the anti-chameleon tool since winter coats round remove people who catch fire and don't take off their coat.
If a person with a Cham vest catches fire they either need to break their disguise before they go crit or get ashed by heat.
https://github.com/space-wizards/space-station-14/issues/33596

That is an issue for sure! I don't think it'll come up very much though. Lighting people on fire already isn't common and if someone takes off their vest I don't think many people would think much of it.

ScarKy0 (approved):
I'm not good with combat balance changes or armor tweaks like this but this is okay

Issue #14747, opened 2023-03-20
penguins can be used to tatorcheck


deltanedas:
## Description
since they attack from lockers and bags you can stuff them in and tatorcheck by putting next to them with next to no risk

**Reproduction**
1. buy pingu
2. pingu stolen
3. pingu put in locker
4. pingu will only attack crewmates so set him up in a hallway or something to tatorcheck

**Screenshots**

**Additional context**
the cor issue cant be fixed without making pingu attack everyone (maybe except who bought it since its usually obvious, see FactionException in fish pr)
but safe tatorchecking can be stopped by not letting it attack outside containers, so you have to put him on the ground which is more dangerous

SlamBamActionman:
The circumstances for this to take place are so incredibly niche with the current behavior that I can't see how it would be utilized to any degree. Penguins keep chomping at their container long after a valid target is gone, so the only way you could achieve this is if the person putting the penguin in the locker is also a Traitor, and the first person who walks by that you "check" is also a Traitor (or else it gets stuck in the chomping state). 

If someone can find a setup where this is actually viable, this issue can be reopened.

PR #34935, opened 2025-02-07
Grenade penguin htn

iaada:
## About the PR

This PR creates a new HTN compound for grenade penguins. They no longer try to escape when in a bag held by a syndicate, don't juke, and don't change their melee target when released. It now acts something like a homing missile.

Also reduced the intensity of the explosion, since the penguin is going to be closer to its target on average.

## Why / Balance

Grenade penguin is fun, but it's not good for much more than chaos. This PR gives it a bit more of a niche, and diversifies NPC behavior. Penguins attacking their buyer is also super annoying.

Normally without having the juke NPCs are difficult to run away from, but since grenade penguins are items they can be picked up and thrown to escape. 

## Technical details

Creates a new precondition `InFriendlyContainerPrecondition` that checks the faction of all parent containers.

## Media

https://github.com/user-attachments/assets/b53becc0-fcb5-436e-b8a8-ba6eec2b1eb6

https://github.com/user-attachments/assets/64b30962-553f-4303-a60f-db9d16cd5b9a

## Requirements

## Breaking changes

**Changelog**

:cl: aada
- tweak: Grenade penguins have new AI. They won't bite your hand while holding them, and hone in on a single target when released.


lzk228:
do you think it fixes https://github.com/space-wizards/space-station-14/issues/14747 ?

iaada:
> do you think it fixes #14747 ?

I would say it's better, but not fixed. Firstly, the repo described in #14747 doesn't work. With how EscapeCompound works mobs will attack the locker the first time they see an enemy, and won't ever stop attacking that locker. You can't really traitor check like that.

As for actually being a traitor checker, it's use as a checker is heavily constrained by only traitors being able to acquire it, and it exploding. Improvement wise, the current penguin clears everyone it attacks of suspicion, and this version will only clear the one person it zeroes in on. It can be used to confirm a traitor if that traitor is holding the penguin, but again it's a bomb. Handing a traitor a bomb to see if they're a traitor is a pretty comical idea, and the current penguin can still confirm a traitor if you put them together alone in a room.

Content.Server/NPC/HTN/Preconditions/InFriendlyContainerPrecondition.cs
```diff
@@ -0,0 +1,51 @@
...
+    private NpcFactionSystem _npcFaction = default!;
+
+    [DataField] public bool IsInFriendlyContainer = true;
+
+    public override void Initialize(IEntitySystemManager sysManager)
+    {
+        base.Initialize(sysManager);
+        _container = sysManager.GetEntitySystem<ContainerSystem>();
+        _npcFaction = sysManager.GetEntitySystem<NpcFactionSystem>();
+    }
```
metalgearsloth, line 25: Just use the entitymanager passed in.

metalgearsloth (changes requested):

iaada:
I haven't had `RestartTest` fail on me before but it seems unrelated.

SlamBamActionman:
Conceptual approval from a Combat workgroup member (me). I'm not familiar with HTN code at all so I can't speak on that front sadly.

Grenade penguins are rarely used and a bit of a noob trap because they're so inconsistent. This PR would make the more consistent, which would make them more good, and there is some counterplay in that you can pick them up (albeit that fact is kind of hidden).

Some notes:
1) When does the grenade penguin lock onto the target? Is it when you put it down? If so, what happens if the traitor (or anyone really) picks the penguin up and puts them down again? 
2) The penguin's uplink store description should probably be updated to align with this behavior, i.e. something along the lines of "Relentlessly attacks the target it is closest to when placed on the floor."? Just to clearly convey that this is how you best use it.
3) Similarly, this now ignores Syndicates. Which might be a bit of a problem if your target... is a Syndicate. The description should include this, similar to how the automated turret does.

iaada:
1. They choose their target when they're placed down. This also means that they change target if they're picked up and put down again. With my understanding of HTN I can't change that behavior, but it might be possible by doing something with `operator: !type:UtilityOperator proto: NearbyMeleeTargets` to have it check if a previously chosen target is still nearby.
2. The description was already changed for both of these, but I'll update it to be a bit more descriptive of how they work.

iaada:
OP updated to show the explosion changes better.

slarticodefast:
> When does the grenade penguin lock onto the target? Is it when you put it down? If so, what happens if the traitor (or anyone really) picks the penguin up and puts them down again?

When it leaves a friendly container, in this case your hand. In case it's in your backpack it won't attack either. If you pick it up again it will become idle again.

> Similarly, this now ignores Syndicates

Already did the same before before, it's how factions work, The only difference is that it will do nothing when a friendly mob holds it or has it in the backpack.

slarticodefast (approved):

PR #34938, opened 2025-02-07
Purge uses of TransformComponent.Anchored.set. Also adds parentheses.

TemporalOroboros:
Switches over all existing content uses of TransformComponent.Anchored.set to use either SharedTransformSystem.AnchorEntity or SharedTransformSystem.Unanchor
Also removes all references to obsolete SharedTransformSystem anchoring API methods.

## About the PR

Removes all existing uses of `TransformComponent.Anchored.set` in the content modules.
Removes all existing uses of `SharedTransformSystem.AnchorEntity` or `SharedTransformSystem.Unanchor` which take separate `EntityUid` and components.
Requires https://github.com/space-wizards/RobustToolbox/pull/5681

## Why / Balance

Part of the slow ECSing of TransformComponent

## Technical details

All uses of `TransformComponent.Anchored.set` were switched over to use `SharedTransformSystem.AnchorEntity` , `SharedTransformSystem.Unanchor`, or both as needed.
All changes assume that the component has been initialized as of the call site.

## Requirements


Content.Server/Construction/Completions/SetAnchor.cs
```diff
@@ -12,7 +13,17 @@ public sealed partial class SetAnchor : IGraphAction
...
+                return;
+
+            var mapManager = IoCManager.Resolve<IMapManager>();
+            var transformSystem = entityManager.System<SharedTransformSystem>();
+
+            if (Value && mapManager.TryFindGridAt(transformSystem.GetMapCoordinates(transform), out var gridUid, out var grid))
+                transformSystem.AnchorEntity((uid, transform), (gridUid, grid));
+            else
+                transformSystem.Unanchor(uid, transform);
         }
```
metalgearsloth, line 27: It would be better for engine to do this instead of content having to manually do it every time, whether via TryAnchor or the likes.

TemporalOroboros: Moved to the engine.

PR #34948, opened 2025-02-07
Hotfix: Hristov wielding delay

SlamBamActionman:
## About the PR

This PR fixes a missing mechanic with the Hristov; upon wielding the gun, there should be a delay before being able to shoot. It was assumed this was already the case during testing, but as it turns out the delay that was observed was caused by making the gun hand the active hand, which triggered a firing delay. By holding the gun in your active hand but not wielding you could bypass the delay.

Also provides this delay in the examine text.

## Why / Balance

Hristov is intended to require setup and punishing if you miss. Being able to instantly wield and fire makes the weapon busted.

This delay is optional and as such, no other gun is affected.

## Technical details

GunRequiresWieldComponent now has a delay property, which when set causes that delay upon wield (unless a longer cooldown is active).

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed bug with Hristov being able to fire instantly upon wielding.

### DIFF
--- a/Content.Shared/Weapons/Ranged/Components/GunRequiresWieldComponent.cs
+++ b/Content.Shared/Weapons/Ranged/Components/GunRequiresWieldComponent.cs
@@ -18,4 +18,10 @@ public sealed partial class GunRequiresWieldComponent : Component
 
     [DataField]
     public LocId? WieldRequiresExamineMessage  = "gunrequireswield-component-examine";
+
+    [DataField, AutoNetworkedField]
+    public float WieldDelay;
+
+    [DataField]
+    public LocId? WieldDelayExamineMessage  = "gunrequireswield-component-delay-examine";
 }
--- a/Content.Shared/Weapons/Ranged/Components/GunRequiresWieldComponent.cs
+++ b/Content.Shared/Weapons/Ranged/Components/GunRequiresWieldComponent.cs
@@ -20,7 +20,7 @@ public sealed partial class GunRequiresWieldComponent : Component
     public LocId? WieldRequiresExamineMessage  = "gunrequireswield-component-examine";
 
     [DataField, AutoNetworkedField]
-    public float WieldDelay;
+    public TimeSpan WieldDelay;
 
     [DataField]
     public LocId? WieldDelayExamineMessage  = "gunrequireswield-component-delay-examine";
--- a/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
+++ b/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
@@ -22,6 +22,7 @@ using Content.Shared.Weapons.Melee.Events;
 using Content.Shared.Weapons.Ranged.Components;
 using Content.Shared.Weapons.Ranged.Events;
 using Content.Shared.Whitelist;
+using Content.Shared.Wieldable;
 using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Containers;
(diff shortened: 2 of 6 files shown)


slarticodefast:
Didn't wielding already add a use delay?
Also this should be maybe unified for all wieldables and not gun specific.

MozarteanChaos:
> Didn't wielding already add a use delay? Also this should be maybe unified for all wieldables and not gun specific.

liltenhead recently demonstrated that the delay was functioning very weirdly, and could be readily bypassed in the way described in this PR

slarticodefast:
Yeah, I have seen that video, but I was wondering why it was working inconsistently. So it would make more sense to find the underlying problem and fix that instead of adding a second delay on top to hide it.

SlamBamActionman:
> Yeah, I have seen that video, but I was wondering why it was working inconsistently. So it would make more sense to find the underlying problem and fix that instead of adding a second delay on top to hide it.

I have already described in the PR description why any perceived "wielding delay" is incorrect; you are not seeing a delay caused by wielding, you are seeing the delay caused by switching your active hand to your gun hand.

It is only that Hristov is the one gun that has a long enough firing delay for it to seem like there's a wielding delay.

- Swap hand + immediately wield = "wield delay" (actually hand swap delay)
- Swap hand + wait a bit + wield = no delay

keronshb:
Im making somewhat similar changes in the hammer pr: https://github.com/space-wizards/space-station-14/pull/34446

Content.Shared/Weapons/Ranged/Components/GunRequiresWieldComponent.cs
```diff
@@ -18,4 +18,10 @@ public sealed partial class GunRequiresWieldComponent : Component
 
     [DataField]
     public LocId? WieldRequiresExamineMessage  = "gunrequireswield-component-examine";
+
+    [DataField, AutoNetworkedField]
+    public float WieldDelay;
```
PJB3005, line 23: Make this a `TimeSpan` instead.

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
```diff
@@ -96,6 +97,7 @@ public override void Initialize()
         SubscribeLocalEvent<GunComponent, ExaminedEvent>(OnExamine);
         SubscribeLocalEvent<GunComponent, CycleModeEvent>(OnCycleMode);
         SubscribeLocalEvent<GunComponent, HandSelectedEvent>(OnGunSelected);
+        SubscribeLocalEvent<GunComponent, ItemWieldedEvent>(OnGunWielded);
```
PJB3005, line 100: Why isn't this event subscribed on `GunRequiresWieldComponent`? That's the actual component involved.

PJB3005 (changes requested):

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
SlamBamActionman: Because `GunComponent.NextFire` is access-restricted to `SharedGunSystem`.

I can make a publicly available function to modify `NextFire` if that's the better way to go about it, though it would probably also involve refactoring `OnGunMelee` which does the same thing (but with `MeleeWeaponComponent`) for the same reason.

PJB3005: Yes, do that.

SlamBamActionman: Turns out the thing `OnGunMelee` was doing was a bit different. Anyhow, changed it to its own public function now.

Errant-4 (commented):
I have a UX concern about this, the zero feedback as to why the gun is not firing (or indication for when it will be able to fire) makes it feel like a bug, the first time you encounter it. If you know, or once you figured it out, there will still be some friction from not knowing WHEN you will be able to fire, if you have to rapid-wield it in a hectic situation (even if that is not the intended use of the gun)

Errant-4:
I guess this was not that useful of an exploit if it never became an issue, then? Should we just leave it in

SlamBamActionman:
I'm closing this in favor of a new solution using UseDelays. #42256 is being set as a pre-requisite, due to how UseDelaySystem uses default IDs, so I'll wait for that first.

PR #34959, opened 2025-02-07
Add chatty lathes

sowelipililimute:
## About the PR
Lathes now have a `LatheChattyComponent` that can be added to cause them to communicate when they gain new recipes *from a research server*, i.e. emagged recipes aren't announced.

## Why / Balance
This is mostly a PR that I intend to facilitate departamental lathes in downstream servers to avoid needing to patch existing code in the research system that would be prone to merge conflicts if something was changed upstream later on.
However, I think this is a nice-to-have for the two departamental lathes that exist upstream, so I am submitting it with that content change.

## Technical details
- LatheChattyComponent exists
- LatheSystem now listens to (LatheChattyComponent, TechnologyDatabaseModifiedEvent) and broadcasts on the provided radio channels
- TechnologyDatabaseModifiedEvent now has a field for carrying which recipes were unlocked by the research

## Media
grafik

## Requirements

:cl:
- add: The medical and security techfabs now announce new recipes that they can fabricate as they are researched

**Changelog**

### DIFF
--- a/Content.Server/Lathe/LatheSystem.cs
+++ b/Content.Server/Lathe/LatheSystem.cs
@@ -8,6 +8,7 @@ using Content.Server.Materials;
 using Content.Server.Popups;
 using Content.Server.Power.Components;
 using Content.Server.Power.EntitySystems;
+using Content.Server.Radio.EntitySystems;
 using Content.Server.Stack;
 using Content.Shared.Atmos;
 using Content.Shared.Chemistry.Components;
@@ -20,6 +21,7 @@ using Content.Shared.Emag.Systems;
 using Content.Shared.Examine;
 using Content.Shared.Lathe;
 using Content.Shared.Lathe.Prototypes;
+using Content.Shared.Localizations;
 using Content.Shared.Materials;
 using Content.Shared.Power;
 using Content.Shared.ReagentSpeed;
@@ -53,6 +55,7 @@ namespace Content.Server.Lathe
         [Dependency] private readonly SharedSolutionContainerSystem _solution = default!;
         [Dependency] private readonly StackSystem _stack = default!;
         [Dependency] private readonly TransformSystem _transform = default!;
+        [Dependency] private readonly RadioSystem _radio = default!;
 
         /// <summary>
         /// Per-tick cache
@@ -67,6 +70,7 @@ namespace Content.Server.Lathe
             SubscribeLocalEvent<LatheComponent, MapInitEvent>(OnMapInit);
             SubscribeLocalEvent<LatheComponent, PowerChangedEvent>(OnPowerChanged);
             SubscribeLocalEvent<LatheComponent, TechnologyDatabaseModifiedEvent>(OnDatabaseModified);
+            SubscribeLocalEvent<LatheAnnouncingComponent, TechnologyDatabaseModifiedEvent>(OnTechnologyDatabaseModified);
             SubscribeLocalEvent<LatheComponent, ResearchRegistrationChangedEvent>(OnResearchRegistrationChanged);
 
             SubscribeLocalEvent<LatheComponent, LatheQueueRecipeMessage>(OnLatheQueueRecipeMessage);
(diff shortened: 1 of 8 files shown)


walksanatora:
so this is similar to goobstation where when a lathe gets a new recipe it will say above it "new recipies recieved" or something.

Content.Server/Lathe/LatheSystem.cs
```diff
@@ -364,6 +368,40 @@ private void OnDatabaseModified(EntityUid uid, LatheComponent component, ref Tec
...
+            var recipeNames = new List<string>();
+            foreach (var recipeId in args.NewlyUnlockedRecipes)
+            {
+                if (!potentialRecipes.Contains(new(recipeId)))
+                    continue;
+
+                if (!_proto.TryIndex(recipeId, out LatheRecipePrototype? recipe))
+                    continue;
+
+                recipeNames.Add(Loc.GetString("lathe-unlock-recipe-radio-broadcast-item", ("item", GetRecipeName(recipe!))));
```
lzk228, line 388: try to use here thing like this
`FormattedMessage.FromMarkupPermissive($"[bold]{itemName}[/bold]")`

lzk228 (commented):
only one little thing, and i don't know about the name "LatheChatty"
otherwise lgtm

Content.Server/Lathe/LatheSystem.cs
sowelipililimute: i don't like to hardcode string formatting in code, even for "simple" cases such as bolding text

lzk228: i took it from 
https://github.com/space-wizards/space-station-14/blob/f6c0e6d3d2b197c9d18508a00730bcb2ae22a7b0/Content.Client/Examine/ExamineSystem.cs#L240
https://github.com/space-wizards/space-station-14/blob/f6c0e6d3d2b197c9d18508a00730bcb2ae22a7b0/Content.Client/Examine/ExamineSystem.cs#L248

this string just for formatting, like in my examples, so it won't just bloat locale strings

sowelipililimute: if you insist, sure

notquitehadouken:
what if there's more than one lathe

Content.Server/Lathe/LatheSystem.cs
```diff
@@ -67,6 +70,7 @@ public override void Initialize()
             SubscribeLocalEvent<LatheComponent, MapInitEvent>(OnMapInit);
             SubscribeLocalEvent<LatheComponent, PowerChangedEvent>(OnPowerChanged);
             SubscribeLocalEvent<LatheComponent, TechnologyDatabaseModifiedEvent>(OnDatabaseModified);
+            SubscribeLocalEvent<LatheAnnouncingComponent, TechnologyDatabaseModifiedEvent>(OnDatabaseNotification);
```
slarticodefast, line 73:
```suggestion
            SubscribeLocalEvent<LatheAnnouncingComponent, TechnologyDatabaseModifiedEvent>(OnTechnologyDatabaseModified);
```
naming convention

Content.Server/Lathe/LatheSystem.cs
```diff
@@ -364,6 +368,41 @@ private void OnDatabaseModified(EntityUid uid, LatheComponent component, ref Tec
...
+            foreach (var recipeId in args.NewlyUnlockedRecipes)
+            {
+                if (!potentialRecipes.Contains(new(recipeId)))
+                    continue;
+
+                if (!_proto.TryIndex(recipeId, out LatheRecipePrototype? recipe))
+                    continue;
+
+                var itemName = GetRecipeName(recipe!);
+                recipeNames.Add($"[bold]{itemName}[/bold]");
```
slarticodefast, line 389: this should be a loc string

Content.Server/Lathe/LatheSystem.cs
```diff
@@ -364,6 +368,41 @@ private void OnDatabaseModified(EntityUid uid, LatheComponent component, ref Tec
...
+            var recipeNames = new List<string>();
+            foreach (var recipeId in args.NewlyUnlockedRecipes)
+            {
+                if (!potentialRecipes.Contains(new(recipeId)))
+                    continue;
+
+                if (!_proto.TryIndex(recipeId, out LatheRecipePrototype? recipe))
+                    continue;
+
+                var itemName = GetRecipeName(recipe!);
```
slarticodefast, line 388:
```suggestion
                var itemName = GetRecipeName(recipe.Value);
```

slarticodefast (changes requested):

Content.Server/Lathe/LatheSystem.cs
sowelipililimute: this isn't a value object so it's just T not Nullable<T>

walksanatora:
I think starlight does this kinda better. (what is done is each tech has a field that says what extra radio channels to broadcast to. so it broadcast military researches to sec, meat manipulation to service etc. which is much better then just "new recipes in X" as you know what recipes have been added)

PR #34965, opened 2025-02-08
Docking allocs reduc

metalgearsloth:
Stuff I had unstaged and saw on dotmemory.

### DIFF
--- a/Content.Server/Shuttles/Systems/DockingSystem.Shuttle.cs
+++ b/Content.Server/Shuttles/Systems/DockingSystem.Shuttle.cs
@@ -323,7 +323,9 @@ public sealed partial class DockingSystem
             // If it's a map check no hard collidable anchored entities overlap
             if (isMap)
             {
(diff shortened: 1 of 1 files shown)


slarticodefast:
Had a short discussion about this one with sloth on discord. To me it looks identical and should not be fully allocated since it uses yield, but maybe someone with more C# knowledge can chime in.

Further reading:
these say it should use lazy evaluation
https://learn.microsoft.com/en-us/dotnet/standard/linq/deferred-execution-lazy-evaluation
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/yield
this article says there might be some weird things going on with foreach and interfaces:
https://pikhota.com/posts/unity-foreach/

PJB3005:
> To me it looks identical and should not be fully allocated since it uses yield, but maybe someone with more C# knowledge can chime in.

The allocation is in the `IEnumerable<T>` instance itself, not allocating an array or something to hold the enumeration's contents.

PJB3005 (approved):

PR #34970, opened 2025-02-08
Make Ignite tryget flammablecomp

metalgearsloth:
The flammable effect does it but we don't do it here for reasons. Sometimes this heisentests.


slarticodefast (approved):
It is a little sus to me to not fix the situation which causes this to be called instead (rock wall being generated above lava in dungeon generation), but at least we get rid of the annoying heisentest.
Since you said it's ok here, I'll merge.
grafik

PR #34971, opened 2025-02-08
Double bullet speeds

metalgearsloth:
The only reason we had it at 20m/s was tunneling reasons.

The quick maths is (BodyA size + BodyB size) / frametime. Currently the game is doing substepping AFAIK so we can use 0.016 as the frametime therefore with a 0.40 width body we can get up to 40.625 (assuming the target is no narrower than 0.25 which given my thindows change last year should be a decently safe assumption).

Draft because I need to adjust the other values and want to put them all in the one spot.

:cl:
- tweak: Doubled bullet speeds.

### DIFF
--- a/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
+++ b/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
@@ -417,7 +417,7 @@ public abstract partial class SharedGunSystem : EntitySystem
         EntityUid? user = null,
         bool throwItems = false);
 
-    public void ShootProjectile(EntityUid uid, Vector2 direction, Vector2 gunVelocity, EntityUid gunUid, EntityUid? user = null, float speed = 20f)
+    public void ShootProjectile(EntityUid uid, Vector2 direction, Vector2 gunVelocity, EntityUid gunUid, EntityUid? user = null, float speed = 40f)
     {
         var physics = EnsureComp<PhysicsComponent>(uid);
         Physics.SetBodyStatus(uid, physics, BodyStatus.InAir);
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -42,7 +42,7 @@
       projectile:
         shape:
           !type:PhysShapeAabb
-          bounds: "-0.1,-0.1,0.1,0.1"
+          bounds: "-0.1,-0.10,0.1,0.30"
         hard: false
         mask:
         - Impassable


Nox38:
Gawd we need this

Errant-4:
This doesn't seem to do anything. The default value is probably getting overwritten somewhere

metalgearsloth:
> This doesn't seem to do anything. The default value is probably getting overwritten somewhere

Some bullets use their own values which is why it's draft coz it's a bit of the wild west in there.

SlamBamActionman:
Conceptually I am willing to playtest this on Vulture, but I'm gonna want to see some media comparison in the PR description since it's a major change. A pass to make sure relevant bullets overriding the values have appropriate values is probably warranted too (e.g. disablers probably don't need a projectile speed boost). 

This will make precision guns like the Lecter, Hristov and Estoc more lethal at longer range since it'll be easier to lead your shots, and honestly considering how bad it can feel to use those in that usecase (the thing they're meant for!!!) I am all for that kind of buff. Should have little to no impact on how effective close-range guns are, other than perhaps trying to chip damage someone with a longrange shotgun pellet.

Kittygyat:
Does this PR affect non-hitscan laser projectiles, (such as the wards laser shotgun, or HoS laser revolver, or the AI turret projectile etc) or only ballistic projectiles?

SlamBamActionman:
> Does this PR affect non-hitscan laser projectiles, (such as the wards laser shotgun, or HoS laser revolver, or the AI turret projectile etc) or only ballistic projectiles?

Any gun that does not have `ProjectileSpeed` set to something different.

Errant-4:
The disabler and practice disabler currently have a projectile speed of 35, while the disabler SMG and energy shotgun have the default. This seems a bit inconsistent, and means that apparently disablers had a bit of an edge in projectile specs compared to all other guns before, but will (after this PR) be near-identical-actually-very-slightly-slower

We might want to just set the disabler blobs to the default velocity at this point

Errant-4:
Here is a comparison. Left is 40, right is 25
I don't think it's really that informative from video

https://github.com/user-attachments/assets/71e73c86-1fbc-4fb4-8625-fba8cfb08d10

Content.Server/Projectiles/ProjectileSystem.cs
```diff
@@ -122,7 +127,13 @@ private void OnStartCollide(EntityUid uid, ProjectileComponent component, ref St
...
         if (component.ImpactEffect != null && TryComp(uid, out TransformComponent? xform))
         {
-            RaiseNetworkEvent(new ImpactEffectEvent(component.ImpactEffect, GetNetCoordinates(xform.Coordinates)), Filter.Pvs(xform.Coordinates, entityMan: EntityManager));
+            var coords = xform.Coordinates;
+            // This moves the impact coordinates back along the bullet's flight path to prevent the impact from visually overshooting too deep/behind the object.
+            // Maybe not the best solution but unless we want to implement raycasts to find the exact point of impact, it works.
+            if (TryComp(uid, out PhysicsComponent? physComp))
+                coords = new EntityCoordinates(xform.MapUid!.Value, _transform.GetWorldPosition(xform) - physComp.LinearVelocity * (float)_gameTiming.FrameTime.TotalSeconds * 0.5f);
+
+            RaiseNetworkEvent(new ImpactEffectEvent(component.ImpactEffect, GetNetCoordinates(coords)), Filter.Pvs(xform.Coordinates, entityMan: EntityManager));
```
Princess-Cheeseballs: This is caused by an engine bug as a result of buffering contacts. On the newest engine version this is fixed so rather than using this workaround we should just accept the visuals will be broken till the next engine version.

ThatGuyUSA:
Thanks Gun God!

PR #34972, opened 2025-02-08
Fix winter coats turning you to ash

### LINKED ISSUE #33596: Winter Coats continue to kill you with heat damage long after a fire is extinguished

beck-thompson:
## About the PR

Fixes #33596! There are a few ways I could have approached this, but I think this is the simplest solution. I added a maximum and minimum values for the heating coefficients, this means if you are super heated (E.g 600K) you will cool down like normal instead of 10x slower until you hit the upper limit. There are probably a few other clothing items that need this but I'm not sure what they are and what their values should be.

## Media

**Changelog**

:cl:
- fix: Winter coats will now not turn you to ash at extremely high temperatures.

### DIFF
--- a/Content.Server/Temperature/Components/TemperatureProtectionComponent.cs
+++ b/Content.Server/Temperature/Components/TemperatureProtectionComponent.cs
@@ -13,10 +13,34 @@ public sealed partial class TemperatureProtectionComponent : Component
     public float HeatingCoefficient = 1.0f;
 
     /// <summary>
+    ///     If the temperature is above this value, the heating coefficient will not be applied.
+    /// </summary>
+    [DataField]
+    public float? HeatingCoefficientMaxTemp;
+
+    /// <summary>
+    ///     If the temperature is below this value, the heating coefficient will not be applied.
+    /// </summary>
+    [DataField]
+    public float? HeatingCoefficientMinTemp;
+
+    /// <summary>
     ///     Multiplier for the transferred heat when cooling down
     /// </summary>
     [DataField]
     public float CoolingCoefficient = 1.0f;
+
+    /// <summary>
+    ///     If the temperature is above this value, the cooling coefficient will not be applied.
+    /// </summary>
+    [DataField]
+    public float? CoolingCoefficientMaxTemp;
+
+    /// <summary>
+    ///     If the temperature is below this value, the cooling coefficient will not be applied.
+    /// </summary>
+    [DataField]
+    public float? CoolingCoefficientMinTemp;
 }
 
 /// <summary>
--- a/Content.Server/Temperature/Systems/TemperatureSystem.cs
+++ b/Content.Server/Temperature/Systems/TemperatureSystem.cs
@@ -135,7 +135,7 @@ public sealed class TemperatureSystem : EntitySystem
 
         if (!ignoreHeatResistance)
         {
-            var ev = new ModifyChangedTemperatureEvent(heatAmount);
+            var ev = new ModifyChangedTemperatureEvent(heatAmount, temperature.CurrentTemperature);
             RaiseLocalEvent(uid, ev);
             heatAmount = ev.TemperatureDelta;
         }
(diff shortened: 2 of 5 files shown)


ArtisticRoomba:
Beautiful.

> There are probably a few other clothing items that need this but I'm not sure what they are and what their values should be.

Funny thing about hardsuits/softsuits? They don't change your body temp if you're on fire. But you still take heat damage from being on fire. It's weird.

LaCumbiaDelCoronavirus:
Winterboots too

beck-thompson:
> Winterboots too

Thank you! I forgot about those

slarticodefast:
I'm not sure I like this approach. This adds way too many parameters that need to be tuned, and breaks energy conservation even more than it already is. The non-symmetrical coefficient for heating and cooling is already wrong from a physics perspective, I don't want to make it worse.

beck-thompson:
> I'm not sure I like this approach. This adds way too many parameters that need to be tuned, and breaks energy conservation even more than it already is. The non-symmetrical coefficient for heating and cooling is already wrong from a physics perspective, I don't want to make it worse.

1.) Yeah I agree there are a lot of parameters and its not good. The issue is there isn't really a way to reduce them. I really tried thinking of a way to limit it to only two but I don't think there is.
2.) Physics wise I agree it doesn't make much sense but it does make sense "realism" wise. The only other way I could think of fixing this is to just change the minimum and maximum comfortable temperatures of a specifies that is wearing a coat. That *also* doesn't make sense from a physics perspective but might be the lesser of two evils? I personally don't like it as much because then everyone, coat or not, will be the same temperature which feels weird.

If you have some good way of resolving this please let me know its been really bothering me :laughing:

beck-thompson:
Screenshot from 2025-02-12 14-21-57
Screenshot from 2025-02-12 14-26-17

Unfortunately it doesn't look like that fixes the issue. The problem is that 10x reduction in heat change is just so much it takes just too long too cool back down to normal temperatures. The other issue is that sweating and shivering only work when you are alive and but the coat works even after you died.

ArtisticRoomba:
Is heat transfer between bodies and the atmosphere modeled correctly? I don't have access to my dev environment right now so I can't check. But realistically, the hotter something is relative to its environment, the quicker it will cool down.

beck-thompson:
> Is heat transfer between bodies and the atmosphere modeled correctly? I don't have access to my dev environment right now so I can't check. But realistically, the hotter something is relative to its environment, the quicker it will cool down.

I'm pretty sure it does but the 10x slowdown is just too much for it to effect the result much

slarticodefast:
>Is heat transfer between bodies and the atmosphere modeled correctly? I don't have access to my dev environment right now so I can't check. But realistically, the hotter something is relative to its environment, the quicker it will cool down.

The code for that is a crime against physics. Here is a quote from me on an older PR modifying it:
grafik

We should do the opposite of this PR and remove a bunch of magic number used for balancing and just enforce proper energy conservation everywhere. That should result in things becoming more balanced on their own.

ArtisticRoomba:
HAHAHAHAHA

beck-thompson:
> We should do the opposite of this PR and remove a bunch of magic numbers used for balancing and just enforce proper energy conservation everywhere. That should result in things becoming more balanced on their own.

> In the meantime, can we just adjust the existing parameters a little?

Probably makes sense :laughing:
The issue is that the only way to tweak the parameters is to just make the coat worse and from what I see in game its pretty balanced right now as is. The only idea I have would be to just not use this component at all and change the comfortable range of temperatures someone has when they are wearing the coat. Again though, the issue with that is then even though your wearing a coat you'll be the same temperature as someone without one you just wont get damaged.

slarticodefast:
grafik
grafik
The underlying issue with thermal regulation was fixed on master, so you can now properly cool down even when wearing a winter coat and you will no longer turn into ash.
So I will be closing this PR.

PR #34973, opened 2025-02-08
Add proper weather to expeditions

LaCumbiaDelCoronavirus:
## About the PR
Completes adding weather (exact same as the weather in the 'weather' console command) to salvage expeditions

available weather:
snow biome:
-heavy, medium and light snowfall
-rain (not storm)
grasslands biome:
-rain
-storm
lava biome:
-heavy, medium and light ashfall

weather can also just be clear

## Why / Balance
It looks nice + it's cool

## Technical details
added nullable ProtoId 'weather' to SalvageMission

## Media
https://www.youtube.com/watch?v=XG-ONWKSjQI

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added weather to salvage expeditions.


ArtisticRoomba:
extremely cool, good stuff

FairlySadPanda (approved):
LGTM, think this is a trivial merge for a maint.

FairlySadPanda:
Testing expeditions:

1. Change your `development.toml` file to load Reach by default, because Dev is a hacked station that doesn't behave properly.
2. Load into Reach.
3. Open the console and `loadgrid` a shuttle. `loadgrid 1 /Maps/Shuttles/infiltrator.yml` or whichever shuttle you like the most
4. Aghost and place down an expedition console on the station, open the console, find a world - for this PR presumably one with a weather system - and print out the disc
5. `tp 0 0 1` to go to the shuttle which spawned at 0,0
6. Load the disc into the shuttle console
7. Open the shuttle console, go to Maps, scan, click the bottom dropdown, click the planet, go to the map, hover over the weird The Sims diamond logo thing, it should be green, click
8. The shuttle FTLs to the expedition world
9. Arrive
10. Behold space weather

For @beck-thompson

loksill:
The wizards killed the expeditions, and you're trying to revive them. It would have been better to make Lavaland, because no one cares about expeditions anymore.

metalgearsloth:
> The wizards killed the expeditions, and you're trying to revive them. It would have been better to make Lavaland, because no one cares about expeditions anymore.

Brother I got 10,000 projects to handle we'll make em better trust.

Content.Shared/Salvage/Expeditions/SalvageExpeditions.cs
```diff
@@ -133,6 +135,11 @@ public sealed record SalvageMission(
     /// </summary>
     public readonly float Temperature = Temperature;
 
+    /// <summary>
+    /// Weather to be displayed on the planet.
+    /// </summary>
+    public readonly string? Weather = Weather;
```
beck-thompson, line 141: Use a proto id here instead of string! I know the rest of them don't do it, but its better to have one done correctly than none :)

beck-thompson (changes requested):
OK so I have tested and it does appear to work but I'm confused: Is the weather supposed to stop after 30 seconds? From my testing and from the function you seem to have hard coded a 30 second maximum.
image
This comment seems to imply that isn't what you wanted? Let me know!

Content.Shared/Salvage/Expeditions/SalvageExpeditions.cs
LaCumbiaDelCoronavirus: I have no clue how but I will try 🙏

LaCumbiaDelCoronavirus:
> OK so I have tested and it does appear to work but I'm confused: Is the weather supposed to stop after 30 seconds? From my testing and from the function you seem to have hard coded a 30 second maximum.

SetWeather has a parameter called EndTime (or something like that, i cant check right now). When you initially call it, it will take EndTime amount of time to change into the new weather specified; its basically how long it takes to transition between the old/no weather into different weather. Yes, it's very poorly named. And yes, this effect is intended.
If you're wondering why I didn't just make this effect instant, this parameter is required and I couldn't get it to be instant without it looking weird, so I settled on 30 seconds.

TL;DR: It takes 30 seconds (hardcoded) from the start of the expedition to transition from clear weather to whichever weather the expedition is supposed to be on.

LaCumbiaDelCoronavirus:
> I'm not so sure this is true 🤔 I've tested this ~5 times at this point at at the 30 second mark the weather always goes from whatever it is supposed to be (Snow, rain, etc...) to nothing. I put breakpoints and it does appear to end after 30 seconds code wise as well. I know nothing about expeditions so maybe this is expected? Let me know 🫡

This is not intended at all. I always got the intended effect when testing, though

beck-thompson (approved):
I'm going to assume it works now, I don't have time to fully test it again but from everything before it should work. :+1:

LaCumbiaDelCoronavirus:
**_This might not be the best idea to merge right now_** considering there could probably be a bug with the weather audio occasionally (very randomly) persisting after the expedition ends, but i have no idea if that's actually the case considering I can't test it (or code a fix if it is actually bugged) right now.. oops :trollface:

beck-thompson:
> This might not be the best idea to merge right now considering there could probably be a bug with the weather audio occasionally (very randomly) persisting after the expedition ends, but i have no idea if that's actually the case considering I can't test it (or code a fix if it is actually bugged) right now.. oops :trollface:

How common is this? If its rare then I would actually think that this PR would be BETTER to merge then we can actually have people replicate the bug

LaCumbiaDelCoronavirus:
> How common is this? If its rare then I would actually think that this PR would be BETTER to merge then we can actually have people replicate the bug

On the debug & tools builds both, it occurs basically every time. However on the release build, this has a chance (twice out of 4 or 5 times when i was testing) to happen once and never again for that round. I _can_ make a very slightly janky fix by adding a delay inbetween the last shuttle on the expedition FTLing away and the expedition planet being deleted, which i am pretty confident will work. I'll do that and try to solve the actual underlying problem in another PR.

LaCumbiaDelCoronavirus:
Can't merge this until I (or someone else who's kind enough :D ) get around to #35081 . Hearing snow ambience forever after finishing an exped isn't something I would say is good, now after testing with more realistic conditions than before I realise that it happens basically 100% of the time.

If anyone likes to know the direct issue: Shutting down a WeatherComponent from a map entity while there's ongoing weather will cause everyone who was on the map to continue hearing it forever no matter what; so by extension this happens when expeds shut down too.

Pixeltheaertist:
> > This might not be the best idea to merge right now considering there could probably be a bug with the weather audio occasionally (very randomly) persisting after the expedition ends, but i have no idea if that's actually the case considering I can't test it (or code a fix if it is actually bugged) right now.. oops :trollface:
> 
> How common is this? If its rare then I would actually think that this PR would be BETTER to merge then we can actually have people replicate the bug

It may also persist in lobby, potentially. That's what i've seen happen anyways. It happens on Trieste, you can hear the planet's rainstorm even when you haven't joined yet... real weird

LaCumbiaDelCoronavirus:
> It may also persist in lobby, potentially. That's what i've seen happen anyways. It happens on Trieste, you can hear the planet's rainstorm even when you haven't joined yet... real weird

It probably won't (im 90% sure) persist in lobby _unless_ the round ends while you are experiencing weather (aka being on an exped planet with active weather).

This is certainly an odd bug though - testing it leads me to believe it's an engine related bug... but I am probably mistaken.

LaCumbiaDelCoronavirus:
This pr is cooked until The Curse Of #35081 is lifted

PR #34987, opened 2025-02-09
Make pressure damage respect species damage modifiers

LaCumbiaDelCoronavirus:
## About the PR
Make species damage resistances (innate damage resistances for species, e.g. blunt resistance for dionas/slimes) apply to pressure damage, without applying armor/etc. damage resistances

## Why / Balance
This is a bug.

## Technical details
Adds new bool, alwaysApplyBaseDamageModifier to TryChangeDamage() of DamageableSystem. if true, base damage modifiers (the ones applied by ApplyModifierSet, like those of species, not the ones applied by armor etc.) will ignore the ignoreResistances bool; TLDR lets you apply damage onto something with it's inherent damage modifiers while still ignoring other damage modifiers such as armor (see video if you didnt understand). Yes, this can be used for more than just pressure damage

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Fix barotrauma damage not respecting

### DIFF
--- a/Content.Server/Atmos/EntitySystems/BarotraumaSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/BarotraumaSystem.cs
@@ -237,7 +237,7 @@ namespace Content.Server.Atmos.EntitySystems
                 if (pressure <= Atmospherics.HazardLowPressure)
                 {
                     // Deal damage and ignore resistances. Resistance to pressure damage should be done via pressure protection gear.
-                    _damageableSystem.TryChangeDamage(uid, barotrauma.Damage * Atmospherics.LowPressureDamage, true, false);
+                    _damageableSystem.TryChangeDamage(uid, barotrauma.Damage * Atmospherics.LowPressureDamage, true, false, damageable, alwaysApplyBaseDamageModifier: true);
 
                     if (!barotrauma.TakingDamage)
                     {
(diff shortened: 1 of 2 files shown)


LaCumbiaDelCoronavirus:
I accidentally entered the PR early while typing smh. one moment

LaCumbiaDelCoronavirus:
should be fixed I think

Partmedia:
Thank you for your investigation and proposed fix. I'm pretty unfamiliar with damage code and I'm going to admit that I read and re-read your comment, and the old code, and still don't understand what's going on and how this is supposed to work. I can imagine that folks in the future are going to have a lot of trouble figuring out when to call it with `ignoreResistances` and `alwaysApplyBaseDamageModifier`.

Shouldn't the nested `if` just be lifted out of the `(!ignoreReistances)` to make it more clear?

Partmedia:
I pushed to the PR branch with an example of what I had in mind. Would you mind double-checking it and testing it to make sure that this works correctly?

LaCumbiaDelCoronavirus:
> I pushed to the PR branch with an example of what I had in mind. Would you mind double-checking it and testing it to make sure that this works correctly?

Thanks, code _looks_ like it will work as intended, haven't tested yet though.

LaCumbiaDelCoronavirus:
Tested, works as intended!

PR #34991, opened 2025-02-09
Add new freezer atmos devices and fix freezer fixgridatmos marker

ArtisticRoomba:
## About the PR
Adds new Freezer atmos devices for mapping. These devices have their sensors tuned to the roundstart environment of a freezer, with auto mode off. The solution is similar to that employed by custom Vox alarms that are no longer used.

Fixes the Freezer FixGridAtmos marker to properly pressurize freezer rooms. Previously they were only at 80 kPa due to their molar amounts not being adjusted to a colder environment. Freezers now start at 101.325 kPa at their desired temp.

Maps will need to have FixGridAtmos reran in order for the changes to take effect when this PR is merged.

## Why / Balance
Commonly mappers map atmospheric devices into freezers with a connected air alarm. This air alarm will commonly scream at you over the atmospheric alerts computer, because the freezer is not within a safe temperature.

## Technical details
Added new threshold, freezerTemperature.
Added new freezer.yml file for holding all of our custom atmos devices with their changed sensor values.

Added new float, FreezerTemp.
Added new float, MolesCellFreezer for determining how many moles we need for a colder room (we'll need more than standard because the room is colder).
Added new floats, OxygenMolesFreezer and NitrogenMolsFreezer, for usage in FixGridAtmos.
Adjusted mixtures[6].AdjustMoles to include new OxygenMolsFreezer and NitrogenMolsFreezer.

## Media
image

## Requirements

**Changelog**
This unfortunately can't get a CL right now because the player-facing changes haven't been made yet.

### DIFF
--- a/Resources/Prototypes/Atmospherics/thresholds.yml
+++ b/Resources/Prototypes/Atmospherics/thresholds.yml
@@ -114,3 +114,14 @@
     threshold: 0.8 # danger below 80% nitrogen
   lowerWarnAround: !type:AlarmThresholdSetting
     threshold: 1.125 # warning below 90%
+
+- type: alarmThreshold
+  id: freezerTemperature
+  upperBound: !type:AlarmThresholdSetting
+    threshold: 335.15 # T-38.15C (235) + 100
+  lowerBound: !type:AlarmThresholdSetting
+    threshold: 135.15 # T-38.15C (235) - 100
+  upperWarnAround: !type:AlarmThresholdSetting
+    threshold: 0.8
+  lowerWarnAround: !type:AlarmThresholdSetting
+    threshold: 1.1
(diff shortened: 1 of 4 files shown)


Partmedia (approved):
Looks good to me, thanks!

Resources/Prototypes/Atmospherics/thresholds.yml
```diff
@@ -114,3 +114,14 @@
     threshold: 0.8 # danger below 80% nitrogen
   lowerWarnAround: !type:AlarmThresholdSetting
     threshold: 1.125 # warning below 90%
+
+- type: alarmThreshold
+  id: freezerTemperature
+  upperBound: !type:AlarmThresholdSetting
+    threshold: 335.15 # T-38.15C (235) + 100
+  lowerBound: !type:AlarmThresholdSetting
+    threshold: 135.15 # T-38.15C (235) - 100
```
Partmedia, line 123: These bounds are quite wide, but in line with the current bounds further up in this file. Someone else can eventually adjust this later.

ArtisticRoomba: They are, and they'll have to be manually adjusted if the freezer temp is changed.

My plan is to remove the directional fans in freezers when I give every map's freezer a proper atmos setup. Might be nice.

PR #34992, opened 2025-02-09
Wizard Fix - Rod Polymorph - Prevent gibbing & slows speed

keronshb:
## About the PR

Reduced the damage Wizard Rod does from 200 to 190 so it doesn't auto gib players, but will still kill them.
Max Speed capped at 25 so they don't go TOO far.

## Why / Balance

Mass dying is fine, gibbing is not. Use Smite if you want to gib.
Slight speed adjustment so you don't go extremely far.

## Media

https://github.com/user-attachments/assets/35a5c278-311e-4c39-8a42-5a98b5059eb3

## Requirements

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
+++ b/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
@@ -66,7 +66,6 @@
   suffix: Wizard
   components:
   - type: ImmovableRod
-    minSpeed: 25
     maxSpeed: 25
     destroyTiles: false
     randomizeVelocity: false
--- a/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
+++ b/Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
@@ -66,13 +66,14 @@
   suffix: Wizard
   components:
   - type: ImmovableRod
-    minSpeed: 35
+    minSpeed: 25
+    maxSpeed: 25
     destroyTiles: false
     randomizeVelocity: false
     shouldGib: false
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
```diff
@@ -66,13 +66,14 @@
   suffix: Wizard
   components:
   - type: ImmovableRod
-    minSpeed: 35
+    minSpeed: 25
+    maxSpeed: 25
     destroyTiles: false
     randomizeVelocity: false
```
rokudara-sen, line 72:
```suggestion
    maxSpeed: 25
    destroyTiles: false
    randomizeVelocity: true
```
I’d recommend setting `maxSpeed` to 25 and removing `minSpeed`, so it falls back on the default 10f. That way, speed isn’t fixed. Also, consider setting `randomizeVelocity` to true so speed can vary instead of staying at a constant value.

rokudara-sen:
```suggestion
    maxSpeed: 25
    destroyTiles: false
    randomizeVelocity: true
```
Another option is to set `randomizeVelocity` to true and remove `minSpeed` (or tweak it if you don’t want a 10f–25f range), so the rod’s speed varies instead of staying fixed.

rokudara-sen (approved):
LGTM

Emisse (approved):

keronshb:
Really small & shameless self merge because of 1 maint requirement.

PR #34994, opened 2025-02-09
Wizard Fix - Summon Ghost Event no longer shows Admin Ghosts or Revenants

keronshb:
## About the PR

Summon Ghosts no longer shows admin ghosts or Revenants.

## Why / Balance

AGhosts shouldn't be shown and Revenants would be too easy to target if they're always visible.

## Technical details

We continue in the MakeVisible query if the ghost can interact or if it has a Revenant component.

## Media

https://github.com/user-attachments/assets/6076baba-5124-4396-a3f8-d5954e00d063

## Requirements

**Changelog**

### DIFF
--- a/Content.Server/Ghost/GhostSystem.cs
+++ b/Content.Server/Ghost/GhostSystem.cs
@@ -25,6 +25,7 @@ using Content.Shared.Mobs.Systems;
 using Content.Shared.Movement.Events;
 using Content.Shared.Movement.Systems;
 using Content.Shared.Popups;
+using Content.Shared.Revenant.Components;
 using Content.Shared.Storage.Components;
 using Robust.Server.GameObjects;
 using Robust.Server.Player;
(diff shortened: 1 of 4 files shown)


nighoggDatatype:
On a related note, I do believe that the tool tip to summon ghost should say 'permanently visible', not 'permanently invisible'

keronshb:
> On a related note, I do believe that the tool tip to summon ghost should say 'permanently visible', not 'permanently invisible'

Good catch

Content.Server/Ghost/GhostSystem.cs
```diff
@@ -399,8 +400,12 @@ private void OnToggleGhostVisibilityToAll(ToggleGhostVisibilityToAllEvent ev)
         public void MakeVisible(bool visible)
         {
             var entityQuery = EntityQueryEnumerator<GhostComponent, VisibilityComponent>();
-            while (entityQuery.MoveNext(out var uid, out _, out var vis))
+            while (entityQuery.MoveNext(out var uid, out var ghostComp, out var vis))
             {
+                // Don't show Revenants, best way to check for admin ghost is if they can interact
+                if (ghostComp.CanGhostInteract || HasComp<RevenantComponent>(uid))
+                    continue;
```
metalgearsloth, line 407: Don't put specific component onto generic system, bad.

Almost every fork is going to have ghostsystem but almost none of them going to have revenantsystem.

metalgearsloth: Eventbus is good, bool property, anything else that doesn't tie an antag to ghosts. Ghosts just handle state changes to / from another entity when a mob dies.

keronshb: So don't check for ghostcomp or is it specifically rev? IIRC there was some discussion about using a tag in discord last night. Think that might be better instead.

ScarKy0 (approved):
Looks good and works. Go my squash and merge button

PR #34999, opened 2025-02-09
Some changes to a few of the visitor ships

TytosB:
## About the PR

i increased the spawn weights for two visitor ships, the cossack man o' war and the syndie evac pod, to 7 [default for all shuttles is 10, so its still lower than average] and moved the man o' war from the antag ship table to the freelance ship table

## Why / Balance

the man o' war move i think explains itself, the cossack ghosts roles it has are free agents, not antags. and they arent particularly strong either, their ship literally just has bow and arrows and other medieval weaponry. really not sure why it was under antag shuttles to begin with.

as for increasing the spawn weights, while i understand that there is always some anxiety for making free agents more common, and i get that the main reason for the visitor ships is to let dead or latejoin observers get into the round, i think there's a bit of opportunity lost there. most of the time the visitor will just dock to the station [almost always blocking evac] beg hop for access, and join the crew quietly. while this DOES achieve the goal of getting them back into the round, i think a big opportunity is lost here. the visitor ghost roles, especially free agents, can have a much more varied and unpredictable series of interactions with the crew and station. the idea of syndie civilians being forced to seek refuge on an NT station is a killer idea and some of my most memorable moments were from this role, even when it was nonantag. its sad to see it so insanely rare now. as for the cossacks, i've literally never even seen them spawn, but the idea of nomadic travelling groups of space feudalists stopping by the station sounds too good for them to be the next to impossible spawn they currently are. at a weight of 7 both are still rarer than the nonantag ships while being common enough that you can actually see them appear in rounds more than once per millennia. and if they prove to just keep being discount antags that murderbone with no nuance we can always turn the weight back down.

## Requirements

## Breaking changes

**Changelog**


Emisse:
id prefer them to be kept very rare

TytosB:
> id prefer them to be kept very rare

whys that?

slarticodefast:
I brought it up for maintainer discussion and we decided this shuttle should be kept rare.
The original author of it mentioned that it is only a 10 minute combat gag and is quite shallow design-wise, but that is fine as long as it's not too common. The admins mentioned that 4 free agent ghost roles with a bunch of weapons can cause a lot of havoc.

PR #35030, opened 2025-02-10
use manual component state for BaseEmitSoundComponent

MilonPL:
## About the PR

:death:
abstract components cannot be networked, attributes arent inherited, you cant put the AutoGenerateComponentState attribute on a component that has no AutoNetworkedFields because those exist in the other class
this isnt how ECS was supposed to be played
but it is what it is and thats better than no networking at all, especially since sounds are predicted

## Technical details

minor cleanup, added manual handling for most components that inherit off of BaseEmitSoundComponent
we cannot do it for `SpamEmitSound` because it uses autonetworking for something else, and `EmitSoundOnTrigger` is in Server so not predicted 
i swear there was a better way of subscribing to multiple events but maybe im tripping?? the Initialize method is a mess either way

## Requirements

## Breaking changes

**Changelog**

no fun

### DIFF
--- a/Content.Shared/Sound/SharedEmitSoundSystem.cs
+++ b/Content.Shared/Sound/SharedEmitSoundSystem.cs
@@ -12,6 +12,7 @@ using Content.Shared.Whitelist;
 using JetBrains.Annotations;
 using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
+using Robust.Shared.GameStates;
 using Robust.Shared.Map;
 using Robust.Shared.Map.Components;
 using Robust.Shared.Network;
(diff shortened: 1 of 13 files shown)


Content.Shared/Sound/SharedEmitSoundSystem.cs
```diff
@@ -54,6 +55,38 @@ public override void Initialize()
...
+    private static void GetBaseEmitState<T>(EntityUid uid, T component, ref ComponentGetState args) where T : BaseEmitSoundComponent
+    {
+        args.State = new EmitSoundComponentState(component.Sound);
+    }
+
+    private static void HandleBaseEmitState<T>(EntityUid uid, T component, ref ComponentHandleState args) where T : BaseEmitSoundComponent
+    {
+        if (args.Current is EmitSoundComponentState state)
+            component.Sound = state.Sound;
     }
```
metalgearsloth, line 90: As these are classes need to copy it (new or whatever else) or else running prediction will mutate the copied state and mispredict.

PR #35025, opened 2025-02-10
Sentry turrets - Part 1: Art assets

chromiumboy:
## About the PR

This PR adds the art assets required for sentry turrets, their projectiles/effects, and their control panels

## Why / Balance

This PR will be required for later PRs in this series

## Technical details

N/A

## Media

Turret sprites in-game (inactive, stun, and lethal settings)
turrets

Video clip of turret animations

https://github.com/user-attachments/assets/f31911b8-bab0-488b-b47f-db7bafe9ede3

Turret control panel sprites in-game (inactive [green], stun [yellow], and lethal [red] settings)
turret_controls

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

N/A

**Changelog**

N/A

PR #35031, opened 2025-02-10
Sentry turrets - Part 2: Basic prototype

chromiumboy:
## About the PR

This PR reorganizes the turret prototypes and creates parent prototypes from which both ballistic and energy weapon turrets can inherit. It also adds prototypes for a pair of new energy-based projectiles and their accompanying muzzle flashes.

A new field has also been added to BatterySelfRechargerComponent, allowing the specification of whether the attached entity needs to be connected to an APC power net to self-recharge (used by energy turrets).

This PR requires #35025

## Why / Balance

This PR will be required for later PRs in this series

## Technical details

N/A

## Media

N/A

## Requirements

## Breaking changes

Ballistic turrets must now inherit from BaseWeaponBallisticTurret instead of BaseWeaponTurret

**Changelog**

N/A


chromiumboy:
Oops, I tried to rename the branch and it closed it :P

ArtisticRoomba (changes requested):
image

I don't know why Scar thought that I would be a good reviewer for this, but nevertheless here is your review.

Mostly stylistic and petty spacing stuff.

This is a "glossing this over" review because 3AM so obviously an actual maintainer needs to give this a proper do-over. Ha.

ArtisticRoomba:
Bless you for atomizing the turrets implementation btw

ScarKy0:
> Bless you for atomizing the turrets implementation btw

It would never have been merged otherwise

ArtisticRoomba (approved):
The YAML is perfect.

:clueless:

Content.Server/Power/Components/BatterySelfRechargerComponent.cs
```diff
@@ -32,5 +32,10 @@ public sealed partial class BatterySelfRechargerComponent : Component
         /// Do not auto recharge if this timestamp has yet to happen, set for the auto recharge pause system.
         /// </summary>
         [DataField] public TimeSpan NextAutoRecharge = TimeSpan.FromSeconds(0f);
+
+        /// <summary>
+        /// Does this device require an external APC network connection to charge?
+        /// </summary>
+        [DataField] public bool RequiresApcNetwork = false;
     }
```
metalgearsloth, line 40: This seems like a bit of a code smell? A battery entity that doesn't actually charge from the network but needs the network to self-recharge?

At the very least it should be its own component as most batteries don't interact with LV power at all.

chromiumboy: It is a bit smelly, but there doesn't appear to be a battery system that supports devices on the APC network. The closest is the PowerChargeSystem that the gravity generator and station anchor use. I can look at implementing something if required

PR #35037, opened 2025-02-10
Tweak viewport nullspace check

metalgearsloth:
WorldPos being 0,0 is sussy-a. Code itself is sussy anyway.


slarticodefast (approved):
Was it really just checking if you are at coordinates 0,0 ?
:csgrad: moment

PR #35043, opened 2025-02-10
Mime can no longer write on paper without breaking their vow

SpeltIncorrectyl:
## About the PR
Mime is unable to write on a piece of paper with a pen/crayon while their vow is active.
They can break their vow to regain the ability to write, and lose it upon remaking their vow.
They can still read pieces of paper regardless of vow.
They can still stamp pieces of paper regardless of vow.
They can still draw on the floor with crayons regardless of vow.

Mime no longer starts with paper.

## Why / Balance

> job is called mime
> looks inside
> no miming
> mfw

I've noticed a habit of mime players just writing everything they want to say on a piece of paper and showing it to people. This sort of defeats the point doesn't it? The point of mime is to play with this interesting restriction of not being able to speak. Restricting yourself to communicating by just writing technically fulfils this restriction, but it's mechanically identical to speaking since you are just typing a literal message out and then other players are reading it. This way of playing is far less interesting then actually miming whatever you want to say and playing charades.

Too often I have run into mimes who only only ever communicate by writing something on a piece of paper, and only ever emote to frown - which they do when you don't read their piece of paper.

## Technical details
* Added IlliterateComponent.
* PaperSystem stops the paper-writing UI from opening if the writer has the IlliterateComponent, and ignores messages from the UI telling it to add content to the paper when the writer has the IlliterateComponent.
* MimePowersComponent has field to decide if the mime is forbidden from writing, called PreventsWriting.
* This field defaults to false but is set to true in the mime job prototype. Downstreams can simply remove this line of yml if they don't want this feature.
* MimePowersSystem adds the IlliterateComponent alongside MutedComponent if PreventsWriting is true, and removes it when the vow is broken and readds it when the vow is remade.

## Media

Cannot write because of vow.
Screenshot from 2025-02-10 18-50-54

Mime no longer starts with paper.
Screenshot from 2025-02-10 18-51-28

Vow text changed.
Screenshot from 2025-02-10 18-49-46
Screenshot from 2025-02-10 18-51-40

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Mimes can no longer write on a piece of paper without breaking their vow first.

### DIFF
--- a/Content.Shared/Paper/BlockWritingSystem.cs
+++ b/Content.Shared/Paper/BlockWritingSystem.cs
@@ -4,7 +4,7 @@ namespace Content.Shared.Paper;
 /// A system that prevents those with the IlliterateComponent from writing on paper.
 /// Has no effect on reading ability.
 /// </summary>
-public sealed class BlockWritingSystem: EntitySystem
+public sealed class BlockWritingSystem : EntitySystem
 {
     public override void Initialize()
     {


Emisse:
hero

Moryachok136:
How about disabling the ability to write with crayons?)

ArtisticRoomba:
Thank god, now we'll get to experience the bread and butter "_Urist McMime can i have externals access_"

VerinSenpai:
> Thank god, now we'll get to experience the bread and butter "_Urist McMime can i have externals access_"

Lmfao

SpeltIncorrectyl:
> How about disabling the ability to write with crayons?)

Mimes can't write on paper with crayons but can write on the floor with them.
I decided to keep it in because drawing stuff on the floor as mime is quite a big part of the job.
Besides no one is going to draw a mural outside the hop to ask for externals access, and even if they did it would be silly because crayons have only 15 charges.

B1essedknight:
How does it affect the PDA note keeper? 
I have had mute individuals use that as their main way of communicating.

Content.Server/Abilities/Mime/MimePowersSystem.cs
```diff
@@ -146,6 +156,8 @@ public void RetakeVow(EntityUid uid, MimePowersComponent? mimePowers = null)
             mimePowers.ReadyToRepent = false;
             mimePowers.VowBroken = false;
             AddComp<MutedComponent>(uid);
+            if (mimePowers.PreventWriting)
+                AddComp(uid, new IlliterateComponent { FailWriteMessage = mimePowers.FailWriteMessage });
```
Errant-4, line 160: This should be EnsureComp to make sure it doesn't try to add a new comp to a mob that somehow alredy has it, through dev shenanigans or something

ScarKy0:
Based PR

Errant-4 (changes requested):
Looks good, just a few minor things

Errant-4:
> How does it affect the PDA note keeper? I have had mute individuals use that as their main way of communicating.

It does not, it still works. And you know what, I am okay with this for two reasons
- From an in-universe perspective, it's hilarious that the mime vow is so old that it did not/could not factor in typing on devices
- Game design-wise, having to _hand you PDA_ to someone is risky enough that I think it should be allowed, just so people can punish you for it. In before all mimes powergame by getting a second PDA, but if that becomes an issue I guess we can just do something about it then :shrug:

Errant-4 (approved):
I like this, I would like to see mimes actually mime. If a server doesn't want it they can just change the mime yaml

Content.Shared/Paper/PaperSystem.cs
```diff
@@ -113,6 +114,14 @@ private void OnInteractUsing(Entity<PaperComponent> entity, ref InteractUsingEve
                     args.Handled = true;
                     return;
                 }
+
+                if (TryComp<IlliterateComponent>(args.User, out var illiterateComp))
```
beck-thompson, line 118: It would be a lot neater if you used events! If you do it that way you could also get rid of the `IlliterateComponent` and just have the mime power system listen for the event and cancel it if the oath isn't broken. If you need examples let me know but it should probably be like `TryWriteOnPaperEvent` and it could be canceled and also would have a field for the popup FTL string.

Content.Shared/Paper/PaperSystem.cs
```diff
@@ -156,6 +165,9 @@ private static StampDisplayInfo GetStampInfo(StampComponent stamp)
 
     private void OnInputTextMessage(Entity<PaperComponent> entity, ref PaperInputTextMessage args)
     {
+        if (HasComp<IlliterateComponent>(args.Actor))
```
beck-thompson, line 168: Probably the same event here? I haven't tested it so I'm not quite sure if you'd need a second one for this or if its like the same thing.

beck-thompson (changes requested):
Looks pretty good! Only thing is I think you should try using events and not have the component. If you need any help or think that isn't the correct approach let me know 🫡

Skarletto:
I always liked mimes being able to write "..." on papers, a shame
but I agree most mimes do not mime at all

Minemoder5000:
This will block those rare few mimes that use paper to draw, like those trading cards, but I agree this is a good change.

ArtisticRoomba:
It will be a shame leaving out those Mimes that mimic drawing and write "[extremely unfathomably detailed description]" on papers, was very creative. But I do agree that most Mimes don't mime.

Content.Shared/Paper/PaperSystem.cs
SpeltIncorrectyl: I am just extending what the MimePowersSystem already does.
It adds and removes a MutedComponent from the mime.
So I made my own IlliterateComponent that is also added and removed by the MimePowersSystem.

SpeltIncorrectyl: I think what you want is for me to make a new IlliteracySystem and have a AttemptWriteEvent and refactor the cancelling out of the PaperSystem and into this new IlliteracySystem where I subscribe to and cancel the AttemptWriteEvent if the mime has the IlliterateComponent.
Is this correct?

Content.Shared/Paper/PaperSystem.cs
SpeltIncorrectyl: If you want me to move this to its own system and use event handling I can do so.
But it is necessary to cancel the writing here because it's a nice server-side check that the writer isn't illiterate.

Content.Shared/Paper/PaperSystem.cs
beck-thompson: > I think what you want is for me to make a new IlliteracySystem and have a AttemptWriteEvent and refactor the cancelling out of the PaperSystem into this new IlliteracySystem where I subscribe to and cancel the AttemptWriteEvent if the mime has the IlliterateComponent.
Is this correct?

Yes you could do that! However, It'll be just simpler if you just listen with the `MimePowerSystem` instead.

SpeltIncorrectyl: Ok so remove illiterate component.
I'm going to have to make cancellable events for the Paper system though, is that ok?
The only event the PaperSystem has is one after a pen has been used to write on something.

beck-thompson: Yes that's totally fine and it makes sense! Yeah if there are already events, put the new one(s) next to the old ones

Content.Shared/Paper/IlliterateComponent.cs
```diff
@@ -0,0 +1,18 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Paper
```
Aeshus, line 3: **nitpick:** We use file-scoped namespaces. 

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces

**suggestion:** Make this file/class file-scoped.

Content.Shared/Paper/PaperSystem.cs
slarticodefast: > Ok so remove illiterate component

I would keep the component.
That way it is a separate thing from mime powers (ECS, yay!)
and it will allow it to be added as a selectable character trait later
or to be added to animals

slarticodefast: But using an event instead of a HasComp is good!

beck-thompson: Yes I agree with slart :)

Sorry for steering you the wrong way like that, your original commen is correct: 
>I think what you want is for me to make a new IlliteracySystem and have a AttemptWriteEvent and refactor the cancelling out of the PaperSystem into this new IlliteracySystem where I subscribe to and cancel the AttemptWriteEvent if the mime has the IlliterateComponent.
Is this correct?

That is what you should do!

pheenty:
Add this to monkeys!!

Killerqu00:
This is a bandaid change that not only doesn't solve the problem, but also removes methods of expression (such as mentioned before - "drawings" on paper, book writing and other stuff).
Those abusing the paper writing will just use notewriter in PDA by placing the ID directly in the ID slot and using the IDless PDA as paper (PDA is 2x1 in storage, paper and pen are both 1x1, so it's basically the same).

Errant-4:
> This is a bandaid change that not only doesn't solve the problem, but also removes methods of expression (such as mentioned before - "drawings" on paper, book writing and other stuff). Those abusing the paper writing will just use notewriter in PDA by placing the ID directly in the ID slot and using the IDless PDA as paper (PDA is 2x1 in storage, paper and pen are both 1x1, so it's basically the same).

We'll see

Content.Shared/Paper/IlliterateComponent.cs
```diff
@@ -0,0 +1,17 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Paper;
+
+/// <summary>
+/// An entity with this component cannot write on paper.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class IlliterateComponent : Component
```
beck-thompson, line 9: Illiterate means you can't read or write not just write! Because it isn't really part of this PR I'm OK with leaving it like this especially because your comment makes it clear that its only for blocking writing.

Content.Shared/Paper/PaperSystem.cs
```diff
@@ -229,3 +250,9 @@ private void UpdateUserInterface(Entity<PaperComponent> entity)
 /// </summary>
 [ByRefEvent]
 public record struct PaperWriteEvent(EntityUid User, EntityUid Paper);
+
+public class PaperWriteAttemptEvent(EntityUid paper): CancellableEntityEventArgs
```
beck-thompson, line 254: Write a comment explaining the event! Also, you should use structs for events not classes (Basically copy the PaperWriteEvent). Yes, its weird that you have to manually add the cancelled field and not extend the `CancellableEntityEventArgs` but trust me on this one its slightly better performance :)

beck-thompson (changes requested):
Only a few minor things. Looks great! I think if people end up using the PDA too much we can always just block it. I think this is fine for now!

Content.Shared/Paper/IlliteracySystem.cs
```diff
@@ -0,0 +1,16 @@
+namespace Content.Shared.Paper;
+
+public sealed class IlliteracySystem: EntitySystem
```
Aeshus, line 3: **issue:** You'll need a doc comment here (doesn't need to be anything fancy, just describe what it does).

Content.Shared/Paper/PaperSystem.cs
```diff
@@ -229,3 +250,21 @@ private void UpdateUserInterface(Entity<PaperComponent> entity)
 /// </summary>
 [ByRefEvent]
 public record struct PaperWriteEvent(EntityUid User, EntityUid Paper);
+
+/// <summary>
+/// Cancellable event for attempting to write on a piece of paper.
+/// </summary>
+/// <param name="paper"></param>
```
beck-thompson, line 257: Fill this out!

beck-thompson (approved):
Besides this and Aeshus review it looks good! :+1:

slarticodefast (dismissed):

Content.Shared/Paper/IlliterateComponent.cs
Centronias, line 1: I'm here to be a pedant:
"Illiterate" means "[Unable to read AND write](https://en.wiktionary.org/wiki/illiterate#English)".
Perhaps a better term would be "dysgraphic", though I accept that this is not a word that'd be recognized as illiterate would, and perhaps there would be legitimate concerns with "dysgraphia" describing a learning disability.

SpeltIncorrectyl: I understand concerns about IlliterateComponent being named as it is.
However it's not player facing (player never even seen the word illiterate).
If someone gives me a good name to change it to I will.
DysgraphiaComponent is a bad name I think because it's something you will have to look up to understand, which defeats the purpose of a meaningful name.

ArtisticRoomba: Just name it something like BlockWriting component, I agree with the name change because there might be a future case for an IlliterateComponent. Although you could probably just add a BlockReading component instead of wrapping it under a single component.

SpeltIncorrectyl: That is good name.
I have changed it.

beck-thompson (approved):
test fail fake still looks good to me!

SpeltIncorrectyl:
Main branch fails tests again...
🚬👌😔

Simyon264 (changes requested):

SpeltIncorrectyl:
If this commit makes the build fail I'm gonna crash out

walksanatora:
also if this is added as a trait (I didn't check). I feel like it should be dysgraphia (learning disability that makes it hard to write) instead of illiterate (as you are still able to read)

Errant-4:
> also if this is added as a trait (I didn't check). I feel like it should be dysgraphia (learning disability that makes it hard to write) instead of illiterate (as you are still able to read)

The mimes are too illiterate to know what illiterate means, and so they don't know they are not supposed to not be able to read

Simyon264 (approved):

Simyon264:
There are 26 Hours left on the vote for approval. Code wise, it all looks good to me.

Emisse (approved):

Content.Shared/Paper/PaperSystem.cs
```diff
@@ -3,6 +3,7 @@
 using Content.Shared.UserInterface;
 using Content.Shared.Database;
 using Content.Shared.Examine;
+using Content.Shared.Ghost;
```
slarticodefast, line 6: what is this import for?

SpeltIncorrectyl: no reason, I assume I autocompleted the wrong thing and Rider put it in
i've removed it

PR #35044, opened 2025-02-10
Fland Armory Restock

Nox38:
## About the PR
Restocked armory to meet the new standard (added a few more lasers, added magazines for the Drozds), did some QOL button changes, added more armor sets, and gave warden a Enforcer. Oh also there were only two cyborg spawns for a maxpop station, so I increased it to 5, and added radiation suits to cargo so they don't die of radiation poisoning while handing out uranium lol

## Why / Balance
Brings Fland up to date in terms of armory equipment, and fixes a few minor issues.

## Media
image

## Requirements

### DIFF
--- a/Resources/Prototypes/Maps/fland.yml
+++ b/Resources/Prototypes/Maps/fland.yml
@@ -24,15 +24,15 @@
             Bartender: [ 2, 2 ]
             Botanist: [ 3, 3 ]
             Chef: [ 2, 2 ]
-            Janitor: [ 3, 3 ]
+            Janitor: [ 4, 4 ]
             Chaplain: [ 1, 1 ]
             Librarian: [ 1, 1 ]
             ServiceWorker: [ 2, 2 ]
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
Has the Warden getting an enforcer thing actually been approved? Also it would kinda make much more sense to put it in Warden's locker.

If we map it in, we'll have to unmap it on every map when Warden gets the Energy Shotgun. Rather than just changing one YAML line.

Nox38:
> Has the Warden getting an enforcer thing actually been approved? Also it would kinda make much more sense to put it in Warden's locker.
> 
> If we map it in, we'll have to unmap it on every map when Warden gets the Energy Shotgun. Rather than just changing one YAML line.

I'm willing to manually unmap it, and I figured mapping it in directly makes more sense as that's how the HOS's WT550 is added

Emisse:
i think its probably fine

PR #35046, opened 2025-02-10
Display radio frequencies with a decimal place

Tayrtahn:
## About the PR

Radio frequencies are now displayed with a decimal point before the last digit (1459 -> 145.9).

## Why / Balance

SS13 parity. Looks better too, in my opinion.

## Technical details

Divide the value by 10 on the C# side, then format it to display one decimal place in Fluent.

I haven't tested, but by using the `NATURALFIXED` Fluent function, cultures that use a different decimal separator should be supported automatically.

## Media

Command Headset

Cargo Encryption Key

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Radio frequencies are now displayed with a decimal point.


metalgearsloth:
Radios are weird because we use an int but treat it like a FixedPoint1 so idk. Maybe if we get more radio parity with 13 it might be changed but fine for now and mildly annoys me.

PR #35048, opened 2025-02-10
Don't force-show storage windows when their contents change

sowelipililimute:
## About the PR
This PR makes it so that the client StorageSystem doesn't force a storage BUI to show when its state changes.

## Why / Balance
The only time a storage BUI is instantiated but hidden is in the case of nested storage, and code wants parent storages to be hidden whilst a child storage is displayed. The force-showing here causes glitchy behaviour as other code attempts to enforce that invariant.

## Technical details
- comment out a storageBui shown

## Requirements

## How to reproduce the bug
- have 2 storages able to opened
- open your bag storage with something nested in it (e.g. roundstart backpack with survival box)
- open nested storage
- take an item out
- shift-b to smart equip it into the bag containing the nested container that's currently open
- observed w/o PR: parent window is re-shown, and glitches out due to code fighting over whether it should be hidden or shown
- observed w/ PR: parent window is not re-shown

**Changelog**

:cl:
- fix: Smart equipping into your backpack w/ a nested storage window open no longer bugs out

### DIFF
--- a/Content.Client/Storage/Systems/StorageSystem.cs
+++ b/Content.Client/Storage/Systems/StorageSystem.cs
@@ -74,10 +74,6 @@ public sealed class StorageSystem : SharedStorageSystem
             {
                 containerBui.Hide();
             }
-            else
-            {
-                storageBui.Show();
-            }
         }
     }
 


metalgearsloth:
Thanks for the fixes! I put the other ones in the review channel as well.

PR #35054, opened 2025-02-11
Remove storage hotbar GUI bandaid

metalgearsloth:
With BUI deferred opens it also happened to fix this issue which is nice.

:cl:
- fix: Fix storage windows being duplicated when reconnecting.

### DIFF
--- a/Content.Client/Storage/StorageBoundUserInterface.cs
+++ b/Content.Client/Storage/StorageBoundUserInterface.cs
@@ -69,12 +69,5 @@ public sealed class StorageBoundUserInterface : BoundUserInterface
 
         _window.Visible = true;
     }
-
-    public void ReOpen()
-    {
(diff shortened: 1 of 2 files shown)


Content.Client/Storage/StorageBoundUserInterface.cs
```diff
@@ -92,12 +92,6 @@ public void Show(Vector2 position)
...
-
-    public void ReOpen()
-    {
-        _window?.Orphan();
-        _window = null;
-        Open();
-    }
 }
 
+
```
metalgearsloth, line 97:
```suggestion
```

SlamBamActionman (approved):
Don't reopen window when screen changes because it adds another one. Screen changes when reconnecting.

Makes sense to me.

PR #35066, opened 2025-02-11
Exclude "Die a Glorious Death" traitors from "Keep alive" target pool

psykana:
## About the PR
"Keep alive" objective now excludes traitors with "Die a Glorious Death" objective.

## Why / Balance
On paper, this combination has comedic value, but in practice such players often ghost after a fight/bombing to greentext and it ends up being frustrating for the helping traitor.

## Technical details
`KeepAliveConditionSystem` now excludes traitors with `DieConditionComponent` objective.

## Requirements

## Breaking changes
Not that I know of

**Changelog**
:cl:
- tweak: "Keep alive" objective now excludes traitors with "Die a Glorious Death" objective

### DIFF
--- a/Content.Server/Objectives/Systems/KeepAliveCondition.cs
+++ b/Content.Server/Objectives/Systems/KeepAliveCondition.cs
@@ -58,6 +58,18 @@ public sealed class KeepAliveConditionSystem : EntitySystem
             }
         }
 
+        // Not gonna save someone who's aimed to die a glorious death
+        foreach (var traitor in traitors)
+        {
+            foreach (var objective in traitor.Mind.Objectives)
+            {
+                if (TryComp<DieConditionComponent>(objective, out var _))
+                {
+                    traitors.RemoveWhere(x => x.Id == traitor.Id);
+                }
+            }
+        }
+
         // You are the first/only traitor.
         if (traitors.Count == 0)
         {


beck-thompson (changes requested):
Pretty brutal hard coding of the component :pensive:. This will fix this specific issue but there are a lot of other cases where certain objectives just don't mix well. It might be ok to hard code this for now and just wait for a full refactor of how objectives are assigned (There is a lot to be desired with the current system.)

At the minimum, add a TODO comment explaining that this needs to be fixed!

If another maintainer says its OK then I'll approve

Pronana:
I actually like this interaction though, it's very funny and chaotic to have to try and keep a DAGD target alive. Traitors are solo antagonists and them getting conflicting objectives is fine in my opinion. I had a funny game where a person and I were looking for the same target only for us to realize at the last moment one of us needed them alive and the other needed them dead. I've also had the same happen where I needed to die or someone I needed alive needed to die and the conflict usually lead to an interesting round.

psykana:
@Pronana, I agree, this is up for debate

beck-thompson:
> @Pronana, I agree, this is up for debate

I'll up it up for maintainer discussion for you :saluting_face:

DieselMohawk:
I think removing this from the game removes friction between traitors from the game, it's genuinely a fun interaction

beck-thompson:
Hello! The maintainer vote concluded that even though it might seem weird, it brings up some potentially fun RP opportunities and its worth the downsides of keeping it as is.

Thanks for the PR!

Coolsurf6:
I wanted to poke a bit of a comment here. 

I personally haven't had a DAGD objective yet and RP for dealing with a traitor with this is actually quite nice. However, those traitors can just /suicide to prevent anything the other player can do to stop. And there is also no way to know if your objective has it to prepare something like nocturine, and a long enough supply for nitrous oxide to force them asleep. That is to say, if you can't suicide while asleep. 

I'd honestly feel this is warranted because those objectives become nearly impossible if someone ends up just taking the literal 7 or 8 buttons to finish their objective to type out a command that the other player can't feasibly stop in time. Its an instant death, no counterplay. 

I feel like this may be warranted cause even if we add some delay, they can just hide in maint and die easily out of sight for even a second. Making the suicide command not give the DAGD win might be a good option, but even then ita very easy to throw yourself out an airlock or beat yourself into crit and succumb (yes you could also make succumbing stop it, but then there's easy counterplay to DAGD like fire or ODing on epi bottles and hiding in a locker)

PR #33800, opened 2024-12-09
Predicted internals

metalgearsloth:
Predicts:
- Canisters
- gas tanks
- internals toggling

Requires https://github.com/space-wizards/RobustToolbox/pull/5558

## Breaking changes
Going to pull this PR out but pretty much everything in internals + gastanksystem + the limited sharedatmossystem it touches.

**Changelog**
:cl:
- tweak: Predict gas canisters, gas tanks, and internals.

PR #35068, opened 2025-02-11
Internals: prioritize gas tanks over jetpacks

psykana:
## About the PR
Internals now use jetpacks only if there's no other gas tanks.

## Why / Balance
Jetpack internals can be rather confusing and frustrating/deadly if not noticed in time.

## Technical details
`FindBestGasTank` now keeps searching if it finds a jetpack, and uses it only if there's no proper gas tank equipped.

## Requirements

## Breaking changes
Not that I know of

**Changelog**
:cl:
- tweak: Toggle internals now uses jetpack only if there's no proper gas tank equipped

### DIFF
--- a/Content.Shared/Body/Systems/SharedInternalsSystem.cs
+++ b/Content.Shared/Body/Systems/SharedInternalsSystem.cs
@@ -277,7 +277,7 @@ public abstract class SharedInternalsSystem : EntitySystem
             _gasTank.CanConnectToInternals((backEntity.Value, backGasTank)))
         {
             found = (backEntity.Value, backGasTank);
-            if (!TryComp<JetpackComponent>(backEntity.Value, out var _))
+            if (!HasComp<JetpackComponent>(backEntity.Value))
             {
                 return found;
             }
@@ -288,7 +288,7 @@ public abstract class SharedInternalsSystem : EntitySystem
             _gasTank.CanConnectToInternals((entity.Value, gasTank)))
         {
             found ??= (entity.Value, gasTank);
-            if (!TryComp<JetpackComponent>(entity.Value, out var _))
+            if (!HasComp<JetpackComponent>(entity.Value))
             {
                 return (entity.Value, gasTank);
             }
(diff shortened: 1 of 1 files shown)


metalgearsloth:
Merge after https://github.com/space-wizards/space-station-14/pull/33800

psykana:
hmmm

psykana:
Someone please cleanup reviewers list and tags since I somehow merged half the repo into my branch

Errant-4 (approved):

RedBookcase:
Honestly if you feel up for it in the future, maybe consider just removing Jetpacks as an option for internals altogether? I legitimately believe there would be nothing lost from that change and it would go a long way to removing some action button clutter.

Errant-4:
> Honestly if you feel up for it in the future, maybe consider just removing Jetpacks as an option for internals altogether? I legitimately believe there would be nothing lost from that change and it would go a long way to removing some action button clutter.

The option of using a jetpack to breathe is currently not considered a viable option by anyone, because of other implementation issues with our jetpacks. But this is kinda offtopic

Content.Shared/Body/Systems/SharedInternalsSystem.cs
```diff
@@ -252,22 +257,36 @@ protected short GetSeverity(InternalsComponent component)
             TryComp<GasTankComponent>(backEntity, out var backGasTank) &&
             _gasTank.CanConnectToInternals((backEntity.Value, backGasTank)))
         {
-            return (backEntity.Value, backGasTank);
+            found = (backEntity.Value, backGasTank);
+            if (!TryComp<JetpackComponent>(backEntity.Value, out var _))
```
sowelipililimute, line 261: you can use HasComp here

sowelipililimute (changes requested):
this looks mostly fine, just use `HasComp` instead of `TryComp out var _` in those places

VerinSenpai:
Praise be 🙏

Princess-Cheeseballs (approved):
Looks fine. Doesn't run enough to warrant using entityqueries I think, but if it does become an issue that's not a hard fix to make.

PR #35071, opened 2025-02-11
Metabolizing bloodstream

nkokic:
## About the PR

Simplified version of https://github.com/space-wizards/space-station-14/pull/34521, without causing or letting issues be caused by other systems.
Blood transfusion is now possible.
Chemicals can now react with blood.
Full saline shot is now good as 2 blood packs.
Proteins don't increase blood.
Vomiting does not take anything from bloodstream.

## Why / Balance

Increasing medical/biological accuracy.
Simplifying the code's idea of metabolism.
Making a path towards better features regarding blood and bloodstream.
Opens a way for chemicals to react with blood as it should.
(Instead of applying blood loss damage, you could now actually remove blood)

## Technical details

C#:
Chemical solution is no more.
Metabolisms refrains from eating its own blood.
Charcoal refrains from flushing blood reagents.
Rerouted from chemical solution towards bloodstream.
Separated bleeding and direct blood subtraction behavior.
Bloodstream has bigger capacity to take chemicals.
Blood is regulated both ways to usual numbers.
Renamed BloodMaxVolume to BloodReferenceVolume.
Added BloodMaxFactor default.

YAML:
Renamed bloodMaxVolume to bloodReferenceVolume.
Removed references to chemical solution.
Adjusted blood gain from saline and protein.

## Requirements

## Breaking changes

For forked repositories:
May cause issues if forks have custom mobs.
Change bloodMaxVolume into bloodReferenceVolume.

**Changelog**

:cl:
- remove: Removed blood generation from protein.
- tweak: Saline (15u) is now equivalent to 2 blood packs.
- tweak: All chemicals are lead to bloodstream.


nkokic:
Regarding the opening path towards better features.
Toxins that apply blood loss as an actual damage, could apply blood level modification/subtraction instead. 
It is counter intuitive to have blood level and then blood loss damage which do not necessarily lead to each other while it should.
This PR also opens space for new toxins to be created, that are specifically effective against certain blood types with their reactions.
However, this PR also puts a concern on chemicals that are (easily) reactive with blood. Such chemicals would be labeled as toxins or medication that requires more attention when handling.

Quantum-cross:
This could also maybe make drunkenness more severe when you have less blood?

nkokic:
@Quantum-cross 
> This could also maybe make drunkenness more severe when you have less blood?

Drunkenness is not affected by blood level. However, if you wish, you can go to Drunk.cs C# file for drunk effect and insert dependency on bloodstream and do math. It would go something like this:

`var boozePower = BoozePower * (1 - _bloodstream.GetBloodLevel(args.TargetEntity));`

Do mind you would need to polish around for null checks and replace mentioned 1 with BloodstreamComponent.BloodMaxFactor which is 2 by default, for proper linear inversion.

ThanosDeGraf:
@ArtisticRoomba

ArtisticRoomba:
Sorry, I am not experienced with the metabolization system enough to offer a meaningful review. Maybe @Princess-Cheeseballs can offer a good overview?

Princess-Cheeseballs:
File is pretty merge conflicted currently @nkokic are you still working on this? If so please resolve the merge conflicts and I'd split off the saline balance changes to their own PR.

nkokic:
> File is pretty merge conflicted currently @nkokic are you still working on this? If so please resolve the merge conflicts and I'd split off the saline balance changes to their own PR.

Oh yes definitely. I wanted to wait if my PR would be closed idea wise. I will solve merge conflicts soon.

nkokic:
> I'd split off the saline balance changes to their own PR.
 
 @Princess-Cheeseballs
 Just separate the saline balance or should I also separate protein balance into new PR?

Princess-Cheeseballs:
> > I'd split off the saline balance changes to their own PR.
> 
> @Princess-Cheeseballs Just separate the saline balance or should I also separate protein balance into new PR?

Both please. Would rather keep this purely mechanical changes with minimal balance.

Content.Server/Body/Systems/MetabolizerSystem.cs
```diff
@@ -159,6 +160,11 @@ private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, Solutio
                 if (reagents >= ent.Comp1.MaxReagentsProcessable)
                     return;
 
+                if (TryComp<BloodstreamComponent>(solutionEntityUid.Value, out var bloodstream))
+                {
+                    if (reagent.Prototype == bloodstream.BloodReagent.Id)
+                        continue;
+                }
```
Princess-Cheeseballs: I know that below this there' a TryComp for MobState in a nested for each, but we really shouldn't have multiple TryComps running in a for loop especially when the entity doesn't change and we can just cache the component elsewhere.

Secondly, I'm not sure how I feel about hardcoding a TryComp for bloodstream into this system. Ideally the list should be edited before coming into the loop since it's different from the solution, and that could probably be event based.

Princess-Cheeseballs (changes requested):
Not a full review but this stood out to me.

Content.Server/Body/Systems/MetabolizerSystem.cs
nkokic: Oh I did think this was something similar to Unity's GetComponent<T> function. Insanely bad for performance if put in loops. Unfortunately, I didn't know how to let Metabolizer reach out to Bloodstream.

I assume we could just cache the reference on initialisation. 
Events are my weakest point here so I would like guidance on that if we are possible to solve it with that.

Princess-Cheeseballs: Sure, events are pretty simple. My first idea would be to have it so when the list of reagents to react is created, before shuffling, raise that list as an event to the solution entity and the containing entity itself. 

Events are just structs that contain data so you'd define and raise a new one by making a 
```
var ev = new SomeKindOfEvent(list);
RaiseLocalEvent(entity, ref ev);
```

And as a struct it would look like
```
[ByRefEvent]
public record struct SomeKindOfEvent(ReagentQuantity[] List);
```

Optionally you could also put a helper method inside the struct like "Remove reagent from list" that way you could subscribe to the event with BloodstreamComponent and within the struct call the "RemoveReagentFromList(reagent);" method so you don't have to copypaste code. 

Lastly, you need to raise the event to the entity with the component, so you'll be raising it to solutionEntityUid.Value in your RaiseLocalEvent.

nkokic: I think it would be nicer if we only fetched singular reagentID from a component for exclusion, than a list of reagents for inclusion in metabolizing.

Unless there are other ways, I will be caching the reference to the bloodsteam since metabolizer and bloodstream are often together.

Princess-Cheeseballs: With events you can have different components exclude reagentIDs from the list per component, which would be better in the long run.

nkokic: I would honestly need an example with that. I cannot see who is invoking the event, who is processing and such.

However, I thought of an workaround: make TryMetabolize function use BloodstreamComponent instead of SolutionManagerComponent.

Content.Server/Body/Systems/MetabolizerSystem.cs
```diff
@@ -95,7 +96,7 @@ public override void Update(float frameTime)
             }
         }
 
-        private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, SolutionContainerManagerComponent?> ent)
+        private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, SolutionContainerManagerComponent?, BloodstreamComponent?> ent)
```
Princess-Cheeseballs:
```suggestion
        private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, SolutionContainerManagerComponent?> ent)
```

unused.

Content.Server/Body/Systems/MetabolizerSystem.cs
```diff
@@ -138,6 +139,9 @@ private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, Solutio
             var list = solution.Contents.ToArray();
             _random.Shuffle(list);
 
+            TryComp<BloodstreamComponent>(solutionEntityUid.Value, out var bloodstream);
+            TryComp<MobStateComponent>(solutionEntityUid.Value, out var state);
+
```
Princess-Cheeseballs: Can just cache mobstate by doing: 
```suggestion
            var dead = _mobStateSystem.IsDead(solutionEntityUid.Value);
```

Content.Server/Body/Systems/MetabolizerSystem.cs
```diff
@@ -138,6 +139,9 @@ private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, Solutio
             var list = solution.Contents.ToArray();
             _random.Shuffle(list);
 
+            TryComp<BloodstreamComponent>(solutionEntityUid.Value, out var bloodstream);
```
Princess-Cheeseballs: Event should be raised here. Organs and such use Metabolizer and don't have bloodstreams. For example: Lungs, which metabolize as lungs regardless of bloodstream. 

```suggestion
            var ev = new MetabolismExclusionEvent(list);
            RaiseLocalEvent(solutionEntityUid.Value, ref ev);
            
            _random.Shuffle(ev.list);
```

If you need me to I can push changes to get it working.

Princess-Cheeseballs (changes requested):
Events please, if you need help I can explain or I can push changes for you. Feel free to ask.

Content.Server/Body/Systems/MetabolizerSystem.cs
nkokic: What is a good file example to look up about defining events in SS14?
I see I would need to make a new file/class.

PS: Even if we create new event, how would event know what is the component? This sounds like component fetch problem again.

Princess-Cheeseballs: EntitySystems do a lot of it automatically for you.

Take a look at IngestionSystem.Blockers for how subscribing to and handling events works and IngestionSystem.API for how events are raised and used to modify data.

nkokic:
@Princess-Cheeseballs
I have done the events. It seems to "work".
And by work I mean that metabolising foreach loop is throwing an exception for modifying the reagent list during the process.
Turned the list into array and it just does't pick up the exclusion fast enough.

My conclusion is that the event is too slow to modify the reagent list before the metabolism starts processing it.

It's not working and I need more help with events.

PS: quick fix didn't do much

Princess-Cheeseballs:
> @Princess-Cheeseballs I have done the events. It seems to "work". And by work I mean that metabolising foreach loop is throwing an exception for modifying the reagent list during the process. Turned the list into array and it just does't pick up the exclusion fast enough.
> 
> My conclusion is that the event is too slow to modify the reagent list before the metabolism starts processing it.
> 
> It's not working and I need more help with events.
> 
> PS: quick fix didn't do much

I can take a look at it and push some changes later today.

Princess-Cheeseballs:
> @Princess-Cheeseballs I have done the events. It seems to "work". And by work I mean that metabolising foreach loop is throwing an exception for modifying the reagent list during the process. Turned the list into array and it just does't pick up the exclusion fast enough.
> 
> My conclusion is that the event is too slow to modify the reagent list before the metabolism starts processing it.
> 
> It's not working and I need more help with events.
> 
> PS: quick fix didn't do much

Congrats you found a really bad bug with reagent code that I'm gonna fix real quick in a different PR. I patched it on your branch and made appropriate changes. I saw NukeOps failed which is unrelated to the enumeration and probably something to do with metabolism changing so I'll leave that for you to debug if it fails again.

arenaconspiracy:
This could revolutionize medical gameplay with only a few meager followups. I can't wait to see this. I have been thinking about the possibilities for a few days now.

nkokic:
> Not a full review (...)

I completely forgot only one file was reviewed. Hahah.

@Princess-Cheeseballs 
When could I expect a full review?

Content.Server/Medical/VomitSystem.cs
```diff
@@ -74,22 +74,7 @@ public void Vomit(EntityUid uid, float thirstAdded = -40f, float hungerAdded = -
...
-                    var vomitChemstreamAmount = _solutionContainer.SplitSolution(bloodStream.ChemicalSolution.Value, vomitAmount);
-                    vomitChemstreamAmount.ScaleSolution(chemMultiplier);
-                    solution.AddSolution(vomitChemstreamAmount, _proto);
-
-                    vomitAmount -= (float)vomitChemstreamAmount.Volume;
-                }
-
-                // Makes a vomit solution the size of 90% of the chemicals removed from the chemstream
-                solution.AddReagent(new ReagentId("Vomit", _bloodstream.GetEntityBloodData(uid)), vomitAmount); // TODO: Dehardcode vomit prototype
+                solution.AddReagent(new ReagentId("Vomit", _bloodstream.GetEntityBloodData(uid)), solutionSize); // TODO: Dehardcode vomit prototype
```
Princess-Cheeseballs: This should still take chems from the bloodstream so ipecac isn't useless. It doesn't make sense but it's what we live with until we get a bigger medical refactor.

Content.Shared/Body/Components/BloodstreamComponent.cs
```diff
@@ -139,17 +138,18 @@ public sealed partial class BloodstreamComponent : Component
...
 
     /// <summary>
-    /// Max volume of internal blood storage,
-    /// and starting level of blood.
+    /// Factor which multiplied with <see cref="BloodReferenceVolume"/>
+    /// gives maximum volume for bloodstream.
     /// </summary>
     [DataField]
-    public FixedPoint2 BloodMaxVolume = FixedPoint2.New(300);
+    public float BloodMaxFactor = 2f;
```
Princess-Cheeseballs, line 144: this name feels confusing to me as it implies it's a Blood max volume, but it's the bloodstream's total max volume. It should be changed to reflect that.

Content.Shared/Body/Systems/SharedBloodstreamSystem.cs
```diff
@@ -71,24 +73,31 @@ public override void Update(float frameTime)
             if (!SolutionContainer.ResolveSolution(uid, bloodstream.BloodSolutionName, ref bloodstream.BloodSolution, out var bloodSolution))
                 continue;
 
-            // Adds blood to their blood level if it is below the maximum; Blood regeneration. Must be alive.
-            if (bloodSolution.Volume < bloodSolution.MaxVolume && !_mobStateSystem.IsDead(uid))
+            var bloodLevel = GetBloodLevel(uid);
+
+            // Blood level regulation. Must be alive.
+            if (bloodLevel != 1f && !_mobStateSystem.IsDead(uid))
```
Princess-Cheeseballs: Use MathHelpers.CloseTo to avoid floating point errors.

Princess-Cheeseballs (changes requested):

Princess-Cheeseballs:
Pushed some changes and left a review. Should be good once changes are addressed unless I missed something

ThanosDeGraf:
I love how this could also pave the way for a dialysis machine, letting doctors continuously monitor and filter out contaminates from a patient's bloodstream.

nkokic:
> I love how this could also pave the way for a dialysis machine, letting doctors continuously monitor and filter out contaminates from a patient's bloodstream.

@ThanosDeGraf 
About dialysis machine.
You can check blood content even with current system. Syringes pull from bloodstream and chemstream at the same time in respective ratios. And so you could in every blood pull iteration purge the impurities from blood and return the good blood.
However, that progress is inverse exponential (very very slow and never ending).
So best option for dialysis is a mechanic that calls for flushing foreign chemicals like what charcoal does, which to work is independent of this PR.

arenaconspiracy:
I have been stalking this PR ever since I found it. Words cannot describe how much I need this.

Content.Server/Medical/VomitSystem.cs
Princess-Cheeseballs: Still waiting on this to be addressed before I review again.

arenaconspiracy:
The changelog needs to be updated to remove the reagent changes which have been factored out of this PR, and the solution line should be reworded, since it's technically not even player facing.

nkokic:
Updated changelog to reflect the current state of the PR.

Content.Server/Body/Systems/MetabolizerSystem.cs
```diff
@@ -133,13 +133,24 @@ private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, Solutio
                 return;
             }
 
+            // Copy the solution do not edit the original solution list
+            var list = solution.Contents.ToList();
+
+            // excluding chemicals which bloodstream declares as blood
+            var ev = new MetabolismExclusionEvent(list);
+            RaiseLocalEvent(solutionEntityUid.Value, ref ev);
```
sowelipililimute: i'm not sure how I feel about this event given that it just outright lets any subscriber modify the entire list of what the metabolizer thinks it has instead of gathering a set to ignore

Content.Shared/Body/Systems/SharedBloodstreamSystem.cs
```diff
@@ -289,25 +294,53 @@ private void OnRejuvenate(Entity<BloodstreamComponent> ent, ref RejuvenateEvent
...
+        var totalBloodLevel = 0.0f;
+
+        for (var i = bloodSolution.Contents.Count - 1; i >= 0; i--)
+        {
+            // Unlike with the metabolism exclusion, we treat all blood as blood, so long as it's "close enough"
+            var (reagentId, quantity) = bloodSolution.Contents[i];
+            if (reagentId.Prototype == entity.Comp.BloodReagent)
+            {
+                totalBloodLevel += quantity.Float();
+            }
```
sowelipililimute: so foreign blood in your bloodstream counts as blood for the purposes of blood level but will be metabolized and won't be affected by methods that modify the blood level explicitly...?

Content.Shared/Body/Systems/SharedBloodstreamSystem.cs
```diff
@@ -289,25 +294,53 @@ private void OnRejuvenate(Entity<BloodstreamComponent> ent, ref RejuvenateEvent
...
-        if (SolutionContainer.ResolveSolution(ent.Owner, ent.Comp.ChemicalSolutionName, ref ent.Comp.ChemicalSolution))
-            SolutionContainer.RemoveAllSolution(ent.Comp.ChemicalSolution.Value);
+        // Copy to list again to avoid modifying during enumeration!
+        foreach (var reagentQuantity in args.ReagentList.ToList())
+        {
+            if (reagentQuantity.Reagent != referenceBlood)
+                continue;
+
+            args.ReagentList.Remove(reagentQuantity);
+        }
```
sowelipililimute: use [`List<T>.RemoveAll`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.removeall?view=net-9.0) instead of duplicating the list

Content.Shared/Body/Systems/SharedBloodstreamSystem.cs
```diff
@@ -364,31 +402,51 @@ public bool FlushChemicals(Entity<BloodstreamComponent?> ent, ProtoId<ReagentPro
...
-        var newSol = SolutionContainer.SplitSolution(ent.Comp.BloodSolution.Value, -amount);
+        foreach (var (reagentId, _) in bloodSolution.Contents)
+        {
+            // Only remove our blood specifically.
+            if (reagentId != bloodRef)
+                continue;
+
+            SolutionContainer.RemoveReagent(ent.Comp.BloodSolution.Value, reagentId, -amount);
+            return true;
+        }
```
sowelipililimute: this loop is extraneous and can be simplified to
```suggestion
    return SolutionContainer.RemoveReagent(ent.Comp.BloodSolution.Value, bloodRef, -amount) > FixedPoint2.Zero;
```

Content.Shared/Body/Events/MetabolismExclusionEvent.cs
```diff
@@ -0,0 +1,6 @@
+using Content.Shared.Chemistry.Reagent;
+
+namespace Content.Shared.Body.Events;
+
+[ByRefEvent]
+public readonly record struct MetabolismExclusionEvent(List<ReagentQuantity> ReagentList);
```
sowelipililimute: needs a doc comment explaining what this does, what it's raised on by what & for what purposes, etc.

Content.Shared/Body/Systems/SharedBloodstreamSystem.cs
```diff
@@ -328,34 +361,39 @@ public void SetBloodLossThreshold(Entity<BloodstreamComponent?> ent, float thres
...
-            if (reagentId.Prototype != excludedReagentID)
-            {
-                SolutionContainer.RemoveReagent(ent.Comp.ChemicalSolution.Value, reagentId, quantity);
-            }
+            if (reagentId.Prototype == ent.Comp.BloodReagent || reagentId.Prototype == excludedReagentId)
+                continue;
+
+            var reagentFlushAmount = SolutionContainer.RemoveReagent(ent.Comp.BloodSolution.Value, reagentId, quantity);
+            flushedSolution.AddReagent(reagentId, reagentFlushAmount);
         }
```
sowelipililimute: the backwards iteration was to allow flushing chemicals w/o allocating an extra list, please keep it

sowelipililimute (dismissed):
not going to block on the event in the metabolizer system, but the way overfilling works is kinda suboptimal and allows a bottle of saline to completely fill someone's metabolism

i would advise adding a default-on-but-able-to-be-disabled parameter for ModifyBloodlevel that prevents it from going over the reference volume unless the user explicitly specifies so, as atm there's not really any interaction/self-regulation to hypervolemia besides clogging metabolism and requiring bloodletting to metabolize medicines

a pass is also needed to ensure that `GetBloodLevel` is used by everything that it should be, that being said, the only thing i see scanning through the code is `HealingSystem` not stopping healing at 100% blood level like it used to

nkokic:
> as atm there's not really any interaction/self-regulation to hypervolemia besides clogging metabolism and requiring bloodletting to metabolize medicines

No, there is a way. I have rewritten blood generation so it works both ways. Blood level will drop the same way it will rise when not on reference value. I used proportional regulator with a clamp.

Saline is valid point. I could instead write a condition for effects which checks blood level. I think such effect condition for reagents would be very important for the future.

ThanosDeGraf:
> > not going to block on the event in the metabolizer system, but the way overfilling works is kinda suboptimal and allows a bottle of saline to completely fill someone's metabolism

> Saline is valid point. I could instead write a condition for effects which checks blood level. I think such effect condition for reagents would be very important for the future.

You can inplement the check, but should let hypervolemia (taking brute damage when going above some percentage above 100%, because cells and bloodvessels are overstretched and popping) be an effect of heavy saline overdose.

Princess-Cheeseballs:
I'd rather just prevent saline and blood packs from letting you go over 100% blood level until we have polycythaemia. 

Polycythaemia is out of scope for this PR.

ThanosDeGraf:
For a mostly refactor, yea I guess so. Would you kindly add a TODO where relevant to expand on the *above 100% blood level* later?

nkokic:
>i would advise adding a default-on-but-able-to-be-disabled parameter for ModifyBloodlevel

Doesn't that break YAGNI principle? We are currently using only one way of filling the blood.

nkokic:
@sowelipililimute 
What's the status on this PR?
How should I proceed?

ThanosDeGraf:
@nkokic With so many PRs open rn, try re-requesting reviews?

nkokic:
> @nkokic With so many PRs open rn, try re-requesting reviews?

I had contacted one of the maintainers about it for some quick responses and I was told this in the DMs:
> I think PR is probably fine as is? If you can change the event to only allow removal of reagents that would also be good. 
Looking at the PR I'm probably gonna need to place a freeze on it cause it would have big conflicts with the current offmed test and also the entityeffects refactor so it'll have to wait till after those before it can be merged. Which will add more merge conflicts.

So @ThanosDeGraf, I am just waiting for maintainers to do their refactors and the huge offmed update.

ThanosDeGraf:
> > @nkokic With so many PRs open rn, try re-requesting reviews?
> 
> I had contacted one of the maintainers about it for some quick responses and I was told this in the DMs:
> 
> > I think PR is probably fine as is? If you can change the event to only allow removal of reagents that would also be good.
> > Looking at the PR I'm probably gonna need to place a freeze on it cause it would have big conflicts with the current offmed test and also the entityeffects refactor so it'll have to wait till after those before it can be merged. Which will add more merge conflicts.
> 
> So @ThanosDeGraf, I am just waiting for maintainers to do their refactors and the huge offmed update.

Shoot yea completely forgot that a month a go offmed wasn't even a thing, let alone an entityeffects refactor x.x. But it's no big deal, it just means there's even more things to look forward to!

Princess-Cheeseballs:
Alright so here's where we're at with this PR. 
Hot liquids melting Diona's bloodstream into syrup is an issue that needs resolving in either one of two ways.
1. Remove the interaction temporarily for balance reasons
2. Merge a bunch of shit to punish hot liquids elsewhere (Thermobaths, Lavabeaker removal)

Thermobaths are needed and the PR is derelict so if you wanted to pick it up that'd be appreciated, otherwise it might be best to just comment out the sap -> syrup reaction. 

Hot liquids in bloodstream should be bad but it shouldn't be easy to do accidentally.

nkokic:
> Hot liquids melting Diona's bloodstream into syrup is an issue that needs resolving in either one of two ways.
> 
> 1. Remove the interaction temporarily for balance reasons
> 2. Merge a bunch of shit to punish hot liquids elsewhere (Thermobaths, Lavabeaker removal)

There's another way:
image
I hope this appeals to diona AND the chefs as they gotta heat and stir like the real ones!

While testing I have found regarding issues:
* Blood/chemstream does not exchange heat with the environment, making it an almost fully closed thermal system.
* All reactions are instant, total conversion happens the moment conditions are met. There should be a max cap how fast a reaction can be done per second.
With those done we could add back the feature of cooking diona from inside out.

Princess-Cheeseballs:
> > Hot liquids melting Diona's bloodstream into syrup is an issue that needs resolving in either one of two ways.
> 
> > 
> 
> > 1. Remove the interaction temporarily for balance reasons
> 
> > 2. Merge a bunch of shit to punish hot liquids elsewhere (Thermobaths, Lavabeaker removal)
> 
> 
> 
> There's another way:
> 
> image
> 
> I hope this appeals to diona AND the chefs as they gotta heat and stir like the real ones!
> 
> 
> 
> While testing I have found regarding issues:
> 
> * Blood/chemstream does not exchange heat with the environment, making it an almost fully closed thermal system.
> 
> * All reactions are instant, total conversion happens the moment conditions are met. There should be a max cap how fast a reaction can be done per second.
> 
> With those done we could add back the feature of cooking diona from inside out.

Guh how did I miss this comment? Yeah this is a good solution, it also needs to be applied to all other blood reactions which currently exist. 

If you're still working on this please update the PR and I can try to get it pushed through. Otherwise I can just fix it up myself.

nkokic:
Seeing multiblood reagents in action and feedback I have the following in plan:
- [x] make blood regulate per blood reagent.
- [X] make events fetch the list of blood reagents and present them to metabolizer system to filter itself instead
- [x] make ~EVIL~ FUN blood level calculation `BloodLevel = (VolA / RefVolA) * (VolB / RefVolB) * ...`

nkokic:
Made blood level calculation work in a multiplicative way, for example:
If your body is supposed to have 100 water and 50 blood (cells), but it only has 50 water and 50 blood cells, it would count as 50% blood level. If there's 300 water and 0 blood cells, 0% blood level.
This only works for cookie people for now, as from what I saw, they are the only mobs who got multi reagent blood.

Tell me if it should be just removed or be a separate change.

nkokic:
I have noticed that there's some stuff that got overlooked in https://github.com/space-wizards/space-station-14/pull/41489
There's currently redundant information now:
- `maxBloodVolume`
- `bloodReagents`

`maxBloodVolume` data can be literaly stored into `bloodReagents` while still giving information about proper ratios.
It would also help me a lot in calculating the regulating blood part so I am gonna have to do that too.

Princess-Cheeseballs:
> Tell me if it should be just removed or be a separate change.

Makes sense. Previously this scenario wasn't possible due to how blood volume worked but it's important now.

nkokic:
It should be done now.

ThanosDeGraf:
This is like watching a rocket launch, a year in the making 🔥

deltanedas:
blessed pr

PR #35087, opened 2025-02-11
Hotfix: Camera offset for "Separated Chat" HUD fix & targetting fix

SlamBamActionman:
## About the PR

This PR fixes some major bugs with the camera offset introduced in `EyeCursorOffsetSystem`:

- Previously the offset was calculated using the full application window size, which worked fine for the standard HUD view but broke the direction/offset for the separated chat HUD, where the game viewport is offset from the middle of the application window. 
- Previously trying to aim at specific targets (e.g. prone players) would fail. This was more noticeable when entering placement mode while offset, as the placement ghost would flip back and forth between the non-offset and offset positions your cursor hovered over. 

## Why / Balance

bugs bad

## Technical details

For the window size, it now queries the UserInterfaceManager's ActiveScreen and gets the correct viewport. There are many different ways to get window sizes and viewports; this was the only way I found that corresponded with the actual game view (other methods I tried either gave the full application window size, or some unsized windows).

For the aiming issue, it turns out that the rendering is ran after every FrameUpdate, however some selection stuff ends up running after Update. This caused a hidden issue where the offset kept being reset to `0,0`, selection code ran, and then the FrameUpdate caught up and adjusted the rendering, making everything appear visually legit. So now we just run the offset query on both the FrameUpdate and Update. It felt a bit iffy just copying code, so if there's a better solution, let me know!

## Steps to reproduce bugs

Note: Since this is a *fix*, these are only noticeable in the current master. 

1. Set your HUD to Separated in the General options menu.
2. Wield the Hristov
3. Try to shoot diagonally

Expected behavior: The bullet flies towards the direction you're aiming.
Result: The bullet will be slightly offset from the direction you're aiming.

1. Wield the Hristov
2. Press P on an entity to enter sandbox Copy placement mode

Expected behavior: The item placement ghost is located underneath the cursor
Result: The item placement ghost flickers between underneath the cursor and whatever location the cursor would hover if the Hristov wasn't wielded.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Aiming with the Hristov is now working correctly when using the Separated Chat HUD mode.
- fix: Cursor targetting with the Hristov now correctly selects the target underneath the cursor.

### DIFF
--- a/Content.Client/Movement/Systems/ContentEyeSystem.cs
+++ b/Content.Client/Movement/Systems/ContentEyeSystem.cs
@@ -63,4 +63,14 @@ public sealed class ContentEyeSystem : SharedContentEyeSystem
             UpdateEyeOffset((entity, eyeComponent));
         }
     }
+
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+        var eyeEntities = AllEntityQuery<ContentEyeComponent, EyeComponent>();
+        while (eyeEntities.MoveNext(out var entity, out ContentEyeComponent? contentComponent, out EyeComponent? eyeComponent))
+        {
+            UpdateEyeOffset((entity, eyeComponent));
+        }
+    }
 }
--- a/Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
+++ b/Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
@@ -1,12 +1,12 @@
 using System.Numerics;
 using Content.Client.Movement.Components;
+using Content.Client.UserInterface.Controls;
 using Content.Shared.Camera;
-using Content.Shared.Inventory;
-using Content.Shared.Movement.Systems;
 using Robust.Client.Graphics;
 using Robust.Client.Input;
 using Robust.Shared.Map;
 using Robust.Client.Player;
+using Robust.Client.UserInterface;
 
 namespace Content.Client.Movement.Systems;
 
@@ -14,11 +14,9 @@ public partial class EyeCursorOffsetSystem : EntitySystem
 {
     [Dependency] private readonly IEyeManager _eyeManager = default!;
     [Dependency] private readonly IInputManager _inputManager = default!;
+    [Dependency] private readonly IUserInterfaceManager _uiManager = default!;
     [Dependency] private readonly IPlayerManager _player = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
-    [Dependency] private readonly SharedContentEyeSystem _contentEye = default!;
-    [Dependency] private readonly IMapManager _mapManager = default!;
-    [Dependency] private readonly IClyde _clyde = default!;
 
     // This value is here to make sure the user doesn't have to move their mouse
     // all the way out to the edge of the screen to get the full offset.
@@ -42,9 +40,16 @@ public partial class EyeCursorOffsetSystem : EntitySystem
 
     public Vector2? OffsetAfterMouse(EntityUid uid, EyeCursorOffsetComponent? component)
     {
+        // We need the main viewport where the game content is displayed, as certain UI layouts (e.g. Separated Chat)
+        // can make it a different size to the game window.
+        if (_uiManager.ActiveScreen == null ||!_uiManager.ActiveScreen!.TryGetWidget<MainViewport>(out var mainViewport))
+            return null;
+
         var localPlayer = _player.LocalPlayer?.ControlledEntity;
         var mousePos = _inputManager.MouseScreenPosition;
-        var screenSize = _clyde.MainWindow.Size;
+        var screenSize = mainViewport.Size;
+
+        // Defines the circle at which the offset is capped.
         var minValue = MathF.Min(screenSize.X / 2, screenSize.Y / 2) * _edgeOffset;
 
         var mouseNormalizedPos = new Vector2(-(mousePos.X - screenSize.X / 2) / minValue, (mousePos.Y - screenSize.Y / 2) / minValue); // X needs to be inverted here for some reason, otherwise it ends up flipped.


Content.Client/Movement/Systems/ContentEyeSystem.cs
```diff
@@ -63,4 +63,14 @@ public override void FrameUpdate(float frameTime)
             UpdateEyeOffset((entity, eyeComponent));
         }
     }
+
+    public override void Update(float frameTime)
```
slarticodefast, line 67: I still think doing exactly the same thing in both the Update and FrameUpdate is extremely sus.

Errant-4 (approved):
Fix the merge conflict

PJB3005 (changes requested):
This code is broken with UI scaling.

Furthermore, the fix shouldn't even be able to fix the diagonal shooting. This requires root causing properly.

Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
```diff
@@ -38,9 +40,16 @@ private void OnGetEyeOffsetEvent(EntityUid uid, EyeCursorOffsetComponent compone
 
     public Vector2? OffsetAfterMouse(EntityUid uid, EyeCursorOffsetComponent? component)
     {
+        // We need the main viewport where the game content is displayed, as certain UI layouts (e.g. Separated Chat)
+        // can make it a different size to the game window.
+        if (_uiManager.ActiveScreen == null ||!_uiManager.ActiveScreen!.TryGetWidget<MainViewport>(out var mainViewport))
```
PJB3005: Use `IEyeManager.MainViewport`.

Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
```diff
@@ -38,9 +40,16 @@ private void OnGetEyeOffsetEvent(EntityUid uid, EyeCursorOffsetComponent compone
...
+            return null;
+
         var localPlayer = _player.LocalEntity;
         var mousePos = _inputManager.MouseScreenPosition;
-        var screenSize = _clyde.MainWindow.Size;
+        var screenSize = mainViewport.Size;
```
PJB3005: This doesn't account for UI scaling properly.

PJB3005:
> The `Update()` loop hack is not a proper fix, it'll cause various unacceptable mispredicts in the future.

Gonna walk this one back. The update loop is like... acceptable I guess. But put a big-ass comment saying "this is to work around gun code being broken."

Content.Client/Movement/Systems/ContentEyeSystem.cs
Princess-Cheeseballs: I looked into this a bit myself, Slam is correct in having the EyeOffset update with Update() but I'm mostly wondering what the purpose of EyeOffset updating during FrameUpdate is? 

Switching FrameUpdate to Update fixes the Offset targeting issues and didn't break the singularity's offset compensation.

SlamBamActionman: Basically my understanding is this:
1. The server sends an update of the base camera position. Most of the time (not always) this will be 0,0. The client should then apply its own clientside offset (i.e. mouse offset) on top of that. So you need to apply this effect every time the position variable gets reset in Update.

2. If you only run it during Update, you will get the correct behavior *but* it will not look good visually. This is because Update runs on every tick, as opposed to FrameUpdate which runs every frame. If Update is slower than FrameUpdate, you'll get a noticeable lower "animation" of the camera moving by only running Update.

3. Simialrly, if you only do it in FrameUpdate, there will be a short tiny frame gap where the position is still set to 0,0 via Update before FrameUpdate fixes it. The camera updates every FrameUpdate, so you won't have any visual inconsistencies, but other systems like interaction clicks update in the Update loop and therefore can trigger when the position is still at 0,0.

SlamBamActionman:
The fix has now been updated to (hopefully) account for all UI scaling and sizes. If I've missed something let me know, I think I tested all cases but I am not fully familiar to the extent of which things can very. It also fixes some aiming offsets that were incorrect in the process. 

I also changed the offset to be 0.8 of the viewport because 0.9 seems to bump against UI elements in separated view. 

---

This offset feature demands something that I don't think any other feature does, which is the mouse position bounded by the *visible* game content's viewport. Some things get close (like interactions) but they aren't impacted if the game content is scaled to be outside the game's window and therefore ignore it. 

This can't, because the circle that determines the offset needs to the maximum interactable area the player can reach. If the green rectangle is the game viewport, you can see that the circle must always be bounded by the viewport or the window<sup>1</sup> itself; otherwise you either get a situation where you have to go into the black padding to reach max offset, or you won't reach max offset because the circle extends outside the game window.

So the changed code basically ensures the offset circle behaves like below:
image

This means that regardless of how your game content is being displayed in terms of positioning and zoom, you will always be able to achieve max offset by moving your mouse to the edge of what's interactable for you.<sup>2</sup>

<sup>1</sup>I use "window" for simplicity here but of course this applies to the separated view where the chatbox is part of the game's *window* but is considered *outside* the game's content viewport.

<sup>2</sup>If these bounds already exists somewhere I am throwing my computer off a bridge.

Princess-Cheeseballs (approved):
I looked into fixing this myself and from what I found I think that this actually just is the proper fix. 

PJB said it's from gun code being broken but it's not just gun code, for example the velocity overlay on players from showplayervelocity is also affected and to me demonstrates why we need two update loops (since it's also broken for more than just camera offset).

Having a loop for frameUpdate ensures that we are getting an accurate position for entities on screen which is good since we shouldn't tie the game's framerate to its update rate. 

On the other end, we don't want to purely use frameUpdate since the game might update in between frames and we don't want the client sending data that's incorrect because a game state change causes the camera offset to decrease. 

A better example of why we should have two loops is because of the other thing this is broken for: Movement. 

If you walk while showplayervelocity is up, you see it flickering because it's not accounting for LERPing properly. In this scenario we would want two different methods for each update loop since the frame update is less predictable than the game update. 

If no one else objects I think this PR should just be merged as is. Right now relying on frameUpdate purely doesn't work since the Update loop resets offset and runs its own calculations and there are good reasons for why it should do that.

southbridge-fur (approved):
*Nothing more permanent than a temporary solution*

Princess-Cheeseballs:
Should be good to go.

PR #35093, opened 2025-02-12
mega datafield names cleanup

lzk228:
cleanup datafield names as they can be taken from the variable name

may be some breaking changes


lzk228:
well damn

Tayrtahn:
Was this done with the code fixer or some find and replace magic?

lzk228:
magic of course

Tayrtahn:
Ah, okay, good. Wanted to make sure the code fixer isn't messed up. 

For the record, I dunno if a bulk change of these is a great idea, since it's merge conflict bait. If the count ever gets low enough though, I'd love to promote it to an actual warning.

PR #35097, opened 2025-02-12
Fix build ghosts not being placeable on a new round.

### LINKED ISSUE #33623: Can't Place Construction Ghost
## Description
When I open the construction menu and left click 'Place construction ghost' under any construction, then click a tile, I cannot place the ghost. I have tried this with many constructables. I also cannot right click to exit construction mode at this point, I have to open the crafting menu by clicking it's icon and left click 'Place construction ghost' to exit the mode. Please help.

VerinSenpai:
## About the PR
Fixes an issue where left over ghosts from the previous shift cause you to be inable to place new ones until you either reconn or clear ghosts.

Fixes #33623

## Why / Balance
It's a bug.

## Technical details
Adds a handler for ConstructionGhostComponent ComponentShutdown to clear the ghost attached to the component.
Attaches the ghost id to the component so this is possible.

## Media
Before:

https://github.com/user-attachments/assets/7d6f453a-bc08-4e22-bebb-6e24bb1861f1

After:

https://github.com/user-attachments/assets/2274ddf9-3418-4084-afcc-614812efd607

## Requirements

## Breaking changes
None.

**Changelog**
I don't believe this needs a cl? Smite me if I'm wrong.

### DIFF
--- a/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
+++ b/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
@@ -91,7 +91,10 @@ namespace Content.Client.Construction.UI
 
             // This is required so that if we load after the system is initialized, we can bind to it immediately
             if (_systemManager.TryGetEntitySystem<ConstructionSystem>(out var constructionSystem))
+            {
+                constructionSystem.ClearAllGhosts();
                 SystemBindingChanged(constructionSystem);
+            }
 
             _systemManager.SystemLoaded += OnSystemLoaded;
             _systemManager.SystemUnloaded += OnSystemUnloaded;
--- a/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
+++ b/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
@@ -91,10 +91,7 @@ namespace Content.Client.Construction.UI
 
             // This is required so that if we load after the system is initialized, we can bind to it immediately
             if (_systemManager.TryGetEntitySystem<ConstructionSystem>(out var constructionSystem))
-            {
-                constructionSystem.ClearAllGhosts();
                 SystemBindingChanged(constructionSystem);
-            }
 
             _systemManager.SystemLoaded += OnSystemLoaded;
             _systemManager.SystemUnloaded += OnSystemUnloaded;
--- a/Content.Client/Construction/ConstructionGhostComponent.cs
+++ b/Content.Client/Construction/ConstructionGhostComponent.cs
@@ -7,6 +7,7 @@ namespace Content.Client.Construction
     [RegisterComponent]
     public sealed partial class ConstructionGhostComponent : Component
     {
+        public int GhostId { get; set; }
         [ViewVariables] public ConstructionPrototype? Prototype { get; set; }
     }
 }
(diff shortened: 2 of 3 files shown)


Content.Client/Construction/UI/ConstructionMenuPresenter.cs
```diff
@@ -91,7 +91,10 @@ public ConstructionMenuPresenter()
 
             // This is required so that if we load after the system is initialized, we can bind to it immediately
             if (_systemManager.TryGetEntitySystem<ConstructionSystem>(out var constructionSystem))
+            {
+                constructionSystem.ClearAllGhosts();
                 SystemBindingChanged(constructionSystem);
+            }
```
metalgearsloth, line 98: 1. What happens exactly with the bug?
2. Can't you just remove the ghost from the dictionary on constructionghostcomponent shutdown?

VerinSenpai: So the bug is that _ghost is persistent between rounds currently. This causes GhostPresent to error out because its checking an entity that doesn't exist anymore, preventing the placement of ghosts until the player either reconnects or clears the ghosts themselves.
As for your suggestion, I'll give it a shot.

VerinSenpai: changed it to use ConstructionGhostComponent ComponentShutdown in constructionsystem.cs to call ClearAllGhosts. This appears to function as desired and doesn't raise an error for predictions in the new round.

metalgearsloth: You need to clear just the relevant ghost not all of them. Clearing all on 1 shutdown will nuke references for every ghost.

VerinSenpai: How's this look? I searched but couldn't find away to get the ghostid from the component without adding a variable for it myself.

VerinSenpai: Upon further review in my more awake state, I believe neither the ghost component nor the construction prototype have access to the ghost id which is needed to remove the ghost from _ghosts. So I'd wager adding ghost id to the component is probably an okay way to go about it.

VerinSenpai: Alternatively, I could use the hash of the component as the key in _ghosts. Though, if gethashcode changes its response based on the contents of the component, this could cause the hash to change and thus break. Assuming I'm reading the doc for gethash properly of course.

VerinSenpai: Ignore that last message, that probably doesn't work how I was thinking. Also, is it cool if I mark this as fixing #33623? I noticed you opened an issue on RT for the key exception but I wasn't sure if it was directly related or just another issue spotted.

VerinSenpai:
Looking further, it seems like its separated so the client and server can have different ghost ids for the same item.

VerinSenpai:
Yeah I'm gonna back off that idea. I've been thoroughly confuddled.

VerinSenpai:
oh shit. nice.

VerinSenpai:
Its been bugging me *wink* for months. Just doin my part.

PR #35098, opened 2025-02-12
Fixing ContrabandSystem to check for Syndicate and Centcomm explicitly

Coolsurf6:
## About the PR
Syndicate items that inherit from other items that have contraband components seem to allow departments from that previously inherited item to "be able to carry the item". This now blanket bans all syndicate items to never be allowed to be on the station.

This also changes it so anyone who is centcomm is fully allowed to use all items, being that centcomm is a rare thing and they _make all da rules_

## Why / Balance
Only a few items had this bug, this just fixes it and gives centcomm more power for those who may use them down stream.

## Technical details
Edited ContrabandSystem to check if the severity is syndicate, and if so would always make the check fail to show the dis-allow string. 

Added a seperate check that skips all others that if you are CentralCommand, always allow.

## Media
Screenshot_20250212_181829

## Requirements

## Breaking changes
Those downstream who may rename CentralCommand may have this break.

**Changelog**
:cl:
- tweak: Central Command are now exempt from all Contraband.
- fix: Syndicate Items that are based on other contraband items now display properly

### DIFF
--- a/Content.Shared/Contraband/ContrabandSystem.cs
+++ b/Content.Shared/Contraband/ContrabandSystem.cs
@@ -86,8 +86,10 @@ public sealed class ContrabandSystem : EntitySystem
 
         String carryingMessage;
         // either its fully restricted, you have no departments, or your departments dont intersect with the restricted departments
-        if (departments.Intersect(component.AllowedDepartments).Any()
-            || jobs.Contains(jobId))
+        if (departments.Contains("CentralCommand") ||
+            severity.ID != "Syndicate" &&
+            departments.Intersect(component.AllowedDepartments).Any() ||
+            jobs.Contains(jobId))
         {
             carryingMessage = Loc.GetString("contraband-examine-text-in-the-clear");
         }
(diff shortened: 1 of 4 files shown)


Coolsurf6:
Should be fixed and cleaned up. I don't know if the hardcoded "CentralCommand" is the best way for this, or maybe making a group of high command that if you have that ID, you overwrite any contraband.

For Wizden, I don't see this being necessarily the case, but may be an option in the future for a downstream if they have that sort of thing.

ArtisticRoomba:
Yeah, even though it works I'm not sure if this should be the solution here, as this is hardcoded. Iffy for forks and if we ever want to change this ourselves down the line.

Coolsurf6:
I think a good way to do this is.

Severity: Minor, Moderate, High; seperate syndicate severity off this and make it its own job. 

Make allowed departments have Syndicate which by default is not allowed, and any items that inherits that depsrtment comes up as the syndicate text.

This would make items like the Ninja gear not major, but high (maybe even extreme). I am tempted to classify all ninja equipment as syndicate as the syndicate as a term in my opinion is a label that NT places on opposing companies or people. Basically a label they have ingrained into their workers to tell them who is bad.

Forks can then make seperate listings for departments. We can make a positive override (if you are that job, you are allowed everything) and a negative one (if your item is allowed in that job, but is against NT/station, it is always disallowed).

slarticodefast:
this should not be hardcoded like this

Coolsurf6:
Going to close this and come back to it later when I have an idea on what to do.

PR #35101, opened 2025-02-12
adds remote signaller error message when too far away

### LINKED ISSUE #17389: Remote signaller needs a distinctive clicking sound when you use it
## Description

Feels like it lacks feedback
**Reproduction**

**Screenshots**

**Additional context**

### LINKED ISSUE #34821: Remote signallers don't warn users if devices are out of range
## Description

Remote signaller range is a bit arbitrary and requires estimating range to work properly. (15-tile range doesn't translate well to player viewport and requires manual counting.)

**Reproduction**
1. Link remote signaller to a device (e.g. C4).
2. Move beyond 15-tile range.
3. Activate remote signaller.
4. No pop-up or warning stating at least one device is out of range.

robinthedragon:
## About the PR
remote signallers make a small popup (*Device too far away!*) when they fail to send a signal due to range
remote signallers make a different small popup (*Device activated.*) when they succeed to send a signal
if you have multiple devices, it'll give multiple popups, stacking as popups usually too, eg *Device too far away! x3*

i'd have liked to add sound to it too (then i could resolve #17389 as well), but when i added `EmitSoundOnUse` to the signaller in yaml, it stopped working as a signaller. so getting sound to work is non-trivial

## Why / Balance
it's nice QoL. remote signallers give very little feedback to the user; this helps alleviate that problem

fixes #34821

## Technical details
new event, `SignalFailedEvent`
`WirelessNetworkSystem` now raises this event when it rejects a packet due to the range being too high
`SignallerSystem` subscribes to this event & makes a popup upon hearing it

update: `WirelessNetworkSystem` *always* raises the event, and attaches a boolean to it
`SignallerSystem` uses this boolean to decide which of the two popups to make

## Media

(video outdated, but i have tested all the new features)
https://github.com/user-attachments/assets/a894cde2-fbf2-4bc2-87d4-dd0a5b2a061e

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Remote signallers now provide a popup upon activation, telling you if you're out of range or not.

### DIFF
--- a/Content.Server/DeviceLinking/Events/SignalFailedEvent.cs
+++ b/Content.Server/DeviceLinking/Events/SignalFailedEvent.cs
@@ -4,4 +4,4 @@ using Content.Shared.DeviceNetwork;
 namespace Content.Server.DeviceLinking.Events;
 
 [ByRefEvent]
-public readonly record struct SignalFailedEvent(EntityUid? uid = null);
+public readonly record struct SignalFailedEvent(EntityUid? uid, bool failed);
--- a/Content.Server/DeviceLinking/Events/SignalFailedEvent.cs
+++ b/Content.Server/DeviceLinking/Events/SignalFailedEvent.cs
@@ -1,7 +1,4 @@
-using Content.Server.DeviceNetwork;
-using Content.Shared.DeviceNetwork;
-
 namespace Content.Server.DeviceLinking.Events;
 
 [ByRefEvent]
-public readonly record struct SignalFailedEvent(EntityUid? uid, bool failed);
+public readonly record struct SignalFailedEvent(EntityUid? uid);
--- a/Content.Server/DeviceLinking/Events/SignalFailedEvent.cs
+++ b/Content.Server/DeviceLinking/Events/SignalFailedEvent.cs
@@ -1,0 +1,7 @@
+using Content.Server.DeviceNetwork;
+using Content.Shared.DeviceNetwork;
+
+namespace Content.Server.DeviceLinking.Events;
+
+[ByRefEvent]
+public readonly record struct SignalFailedEvent(EntityUid? uid = null);
--- a/Content.Server/DeviceLinking/Systems/SignallerSystem.cs
+++ b/Content.Server/DeviceLinking/Systems/SignallerSystem.cs
@@ -7,8 +7,6 @@ using Content.Shared.Database;
 using Content.Shared.Interaction.Events;
 using Content.Shared.Popups;
 using Content.Shared.Timing;
-using Robust.Shared.Map;
-using Microsoft.CodeAnalysis.Scripting.Hosting;
 using Robust.Shared.Toolshed.Commands.Math;
 
 namespace Content.Server.DeviceLinking.Systems;
@@ -19,7 +17,6 @@ public sealed class SignallerSystem : EntitySystem
     [Dependency] private readonly UseDelaySystem _useDelay = default!;
     [Dependency] private readonly IAdminLogManager _adminLogger = default!;
     [Dependency] private readonly PopupSystem _popup = default!;
-    [Dependency] private readonly SharedTransformSystem _transform = default!;
 
     public override void Initialize()
     {
@@ -42,6 +39,7 @@ public sealed class SignallerSystem : EntitySystem
             return;
 
         _adminLogger.Add(LogType.Action, LogImpact.Low, $"{ToPrettyString(args.User):actor} triggered signaler {ToPrettyString(uid):tool}");
+        component.LastUser = args.User;
         _link.InvokePort(uid, component.Port);
         args.Handled = true;
     }
@@ -58,15 +56,9 @@ public sealed class SignallerSystem : EntitySystem
         args.Handled = true;
     }
 
-    private void OnSignalFailed(Entity<SignallerComponent> ent, ref SignalFailedEvent args) // called by WirelessNetworkSystem
+    private void OnSignalFailed(EntityUid uid, SignallerComponent component, ref SignalFailedEvent args) // called by WirelessNetworkSystem
     {
-        if (args.failed) // makes a popup if the signaller is out of range
-        {
-            _popup.PopupEntity(Loc.GetString("signaller-interact-out-of-range-text"), ent.Owner, _transform.GetParentUid(ent.Owner), PopupType.SmallCaution);
-        }
-        else // diff popup if it's in-range
-        {
-            _popup.PopupEntity(Loc.GetString("signaller-interact-in-range-text"), ent.Owner, _transform.GetParentUid(ent.Owner));
-        }
+        // makes a popup if the signaller fails to trigger
+        _popup.PopupEntity(Loc.GetString("signaller-interact-failed-to-activate-text"), uid, component.LastUser, PopupType.SmallCaution);
     }
 }
--- a/Content.Server/DeviceLinking/Systems/SignallerSystem.cs
+++ b/Content.Server/DeviceLinking/Systems/SignallerSystem.cs
@@ -5,7 +5,9 @@ using Content.Server.Explosion.EntitySystems;
 using Content.Server.Popups;
 using Content.Shared.Database;
 using Content.Shared.Interaction.Events;
+using Content.Shared.Popups;
 using Content.Shared.Timing;
+using Robust.Shared.Map;
 using Microsoft.CodeAnalysis.Scripting.Hosting;
 using Robust.Shared.Toolshed.Commands.Math;
 
@@ -17,6 +19,7 @@ public sealed class SignallerSystem : EntitySystem
     [Dependency] private readonly UseDelaySystem _useDelay = default!;
     [Dependency] private readonly IAdminLogManager _adminLogger = default!;
     [Dependency] private readonly PopupSystem _popup = default!;
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
 
     public override void Initialize()
     {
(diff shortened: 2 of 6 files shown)


robinthedragon:
> make the code messier

nvm, just had an idea for a more elegant solution (send event always, attach a true/false to it, base message off of that), will implement tomorrow

Booblesnoot42 (approved):
Does what it says it does when ran on my machine. Something that is shown in the video but not stated in the pr description: if the signaller is connected to multiple devices, then the message will be displayed with however many devices are out of range. For example, "Too far away! x3" if three devices are out of range.

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
```diff
@@ -48,4 +54,10 @@ private void OnTrigger(EntityUid uid, SignallerComponent component, TriggerEvent
         _link.InvokePort(uid, component.Port);
         args.Handled = true;
     }
+
+    private void OnSignalFailed(Entity<SignallerComponent> ent, ref SignalFailedEvent args)
+    {
+        // makes a popup if the signaller is out of range
+        _popup.PopupEntity(Loc.GetString("signaller-interact-out-of-range-text"), ent.Owner);
```
Booblesnoot42, line 61: As is, this causes the popup to appear for all players, not just the user of the signaller.

Booblesnoot42 (changes requested):

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
SpeltIncorrectyl: Yeah you should use PopupClient

Booblesnoot42: > Yeah you should use PopupClient

PopupEntity also has the ability to only show for specific clients. 
`/// <summary>`
`        ///     Variant of <see cref="PopupEntity(string, EntityUid, PopupType)"/> that shows the popup only to some specific client.`
`/// </summary>`
`public abstract void PopupEntity(string? message, EntityUid uid, EntityUid recipient, PopupType type = PopupType.Small);`

PopupClient states:
`/// Variant of <see cref="PopupEntity(string, EntityUid, EntityUid, PopupType)"/> that only runs on the client, outside of prediction.`
`/// Useful for shared code that is always ran by both sides to avoid duplicate popups.`

I don't know if it's necessary in this situation.

robinthedragon:
> Does what it says it does when ran on my machine. Something that is shown in the video but not stated in the pr description: if the signaller is connected to multiple devices, then the message will be displayed with however many devices are out of range. For example, "Too far away! x3" if three devices are out of range.

I didn't even realize this was behaviour until I was recording the video and thought "wait, fuck, this is a possible failure state", lol. thankfully it just happened to resolve itself in a neat way

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
robinthedragon: my main roadblock in making it a client-only popup; i have no idea how to get the UID of the character currently holding an object

robinthedragon:
image

image

all requested changes made

robinthedragon:
image

oh. that's bad. need to fix that

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
```diff
@@ -1,9 +1,15 @@
...
 using Content.Server.DeviceLinking.Components;
+using Content.Server.DeviceLinking.Events;
 using Content.Server.Explosion.EntitySystems;
+using Content.Server.Popups;
 using Content.Shared.Database;
 using Content.Shared.Interaction.Events;
+using Content.Shared.Popups;
 using Content.Shared.Timing;
+using Robust.Shared.Map;
+using Microsoft.CodeAnalysis.Scripting.Hosting;
```
slarticodefast, line 11: remove this

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
```diff
@@ -48,4 +57,16 @@ private void OnTrigger(EntityUid uid, SignallerComponent component, TriggerEvent
         _link.InvokePort(uid, component.Port);
         args.Handled = true;
     }
+
+    private void OnSignalFailed(Entity<SignallerComponent> ent, ref SignalFailedEvent args) // called by WirelessNetworkSystem
+    {
+        if (args.failed) // makes a popup if the signaller is out of range
+        {
+            _popup.PopupEntity(Loc.GetString("signaller-interact-out-of-range-text"), ent.Owner, _transform.GetParentUid(ent.Owner), PopupType.SmallCaution);
```
slarticodefast, line 65: Doing this via the parent uid feels very hacky.

Content.Server/DeviceNetwork/Systems/WirelessNetworkSystem.cs
```diff
@@ -29,8 +30,15 @@ private void OnBeforePacketSent(EntityUid uid, WirelessNetworkComponent componen
             if (xform.MapID != args.SenderTransform.MapID
                 || (ownPosition - _transformSystem.GetWorldPosition(xform)).Length() > sendingComponent.Range)
             {
+                var eventArgs = new SignalFailedEvent(args.Sender, true); // event so signallers throw an error if they're out of range
+                RaiseLocalEvent(args.Sender, ref eventArgs, false);
```
slarticodefast, line 34: Not sure if I like raising this event here rather than where `BeforePacketSentEvent` has been raised. The signal could fail for other reasons as well, not only due to range.
Best let @juliangiebel have a look at this.

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
robinthedragon: the alternative solution i have was to save the uid of the person to trigger the remote, and then send the popups to them

i suppose that would be more robust (it'd work if you activated the remote while it was sitting on the floor, for instance), but it feels more expensive and complicated

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
robinthedragon: done on local, will push in next commit

Content.Server/DeviceNetwork/Systems/WirelessNetworkSystem.cs
robinthedragon: full disclosure, it's implemented this way purely because, when i was first looking at the files, `WirelessNetworkSystem` was the easiest one to wrap my head around

knowing what i do now, it would not be hard to implement a more generalized `SignalFailedEvent`, but it would make the overlapping popups problem even worse

juliangiebel: I agree with slarticodefast. Please move raising the event to where the `BeforePacketSentEvent` gets raised

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
robinthedragon: alternative solution adopted

robinthedragon:
> image
> 
> oh. that's bad. need to fix that

i just removed the popup on successful activation; this behaviour seems annoying to fix, and it's much more important to have feedback on failure, imo. since success already has the feedback of... succeeding.

SpeltIncorrectyl:
Is this stuck waiting for review?

Content.Server/DeviceLinking/Systems/SignallerSystem.cs
```diff
@@ -48,4 +55,10 @@ private void OnTrigger(EntityUid uid, SignallerComponent component, TriggerEvent
         _link.InvokePort(uid, component.Port);
         args.Handled = true;
     }
+
+    private void OnSignalFailed(EntityUid uid, SignallerComponent component, ref SignalFailedEvent args) // called by WirelessNetworkSystem
+    {
+        // makes a popup if the signaller fails to trigger
+        _popup.PopupEntity(Loc.GetString("signaller-interact-failed-to-activate-text"), uid, component.LastUser, PopupType.SmallCaution);
+    }
```
EmoGarbage404, line 63: instead of storing the last person who used it, just get the person holding the item through ContainerSystem

Content.Server/DeviceLinking/Events/SignalFailedEvent.cs
```diff
@@ -0,0 +1,4 @@
+namespace Content.Server.DeviceLinking.Events;
+
+[ByRefEvent]
+public readonly record struct SignalFailedEvent(EntityUid? uid);
```
EmoGarbage404, line 4: should have a doc comment for the event + parameters

Content.Server/DeviceNetwork/Systems/DeviceNetworkSystem.cs
```diff
@@ -364,7 +365,11 @@ private void SendToConnections(ReadOnlySpan<DeviceNetworkComponent> connections,
                 if (!beforeEv.Cancelled)
                     RaiseLocalEvent(connection.Owner, packet, false);
                 else
+                {
                     beforeEv.Uncancel();
+                    var eventArgs = new SignalFailedEvent(packet.Sender); // event so signallers throw an error if they're out of range
+                    RaiseLocalEvent(packet.Sender, ref eventArgs, false);
```
EmoGarbage404, line 371:
```suggestion
                    RaiseLocalEvent(packet.Sender, ref eventArgs);
```
This defaults to false

EmoGarbage404 (changes requested):

TheShuEd:

PR #35102, opened 2025-02-12
Updated Syndicate and Ammo Contraband

Coolsurf6:
## About the PR
Updates some Syndicate Related Items

## Why / Balance
Most likely most of these items were missed originally, and can cause some confusions with players expecially with Syndicate IDs and PDAs. 

## Technical details
Added BaseSyndicateContraband to some items.
Added to BaseRestrictedContraband to BaseMagazineBoxAntiMateriel (.60 Antimaterial Rifle Ammo). Other syndicate related ammo items are only security contraband.

## Media
N/A

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- tweak: Updated more contraband items for Syndicate and Security.


Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
```diff
@@ -650,7 +650,7 @@
     - SyndicateAgent
 
 - type: entity
-  parent: IDCardStandard
+  parent: [ IDCardStandard, BaseSyndicateContraband ]
```
RedBookcase, line 653:
```suggestion
  parent: [ IDCardStandard, BaseMajorContraband ]
```

Pirate gear isn't Syndicate contra, and shouldn't be marked as such. If you want it to be contra make it either minor/major.

RedBookcase (changes requested):

Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
Coolsurf6: I believe Major here works. I see Syndicate as any **corporation/entity that goes against NanoTrasen**. It's weird that something maybe like a gun or even the ninja sword is only considered Major when those are directly against NT the same way a syndicate soap is. 

Even with that in mind, Major is a good suggestion and I'll change it to that now.

Coolsurf6:
Web Edit my beloved --> I checked this in game and it did change correctly. I can see that the Pirate PDA is still considered Syndicate contraband so I will quickly change that as well.

image

The Pirate PDA was actually parented off the SyndiPDA which is why it was syndicate contra, which I believe is why I made it syndicate contraband for the ID itself.

ArtisticRoomba (approved):
Looks good to me, thanks for the fix!

PR #35104, opened 2025-02-12
Cleanup: Superbonk cleanup

### LINKED ISSUE #32179: SuperBonk needs cleanup
- Removing comp on its own eventbus subs.
- `if (!(comp.TimeRemaining <= 0))`
- delay isn't used.
- Should use EntityT

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Resolves #32179.

## Requirements

## Breaking changes

**Changelog**


Content.Server/Administration/Systems/SuperBonkSystem.cs
```diff
@@ -59,49 +56,69 @@ public override void Update(float frameTime)
...
+            RemComp<SuperBonkComponent>(uid);
         }
+
+        _toRemove.Clear();
     }
 
-    private void Bonk(SuperBonkComponent comp)
+    public void StartSuperBonk(EntityUid target, bool stopWhenDead = false)
     {
-        var uid = comp.Tables.Current.Key;
```
beck-thompson, line 80: Remove the extra white space

Content.Server/Administration/Systems/SuperBonkSystem.cs
```diff
@@ -12,44 +13,40 @@ public sealed class SuperBonkSystem : EntitySystem
     [Dependency] private readonly SharedTransformSystem _transformSystem = default!;
     [Dependency] private readonly ClumsySystem _clumsySystem = default!;
     [Dependency] private readonly SharedAudioSystem _audioSystem = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+
+    private readonly HashSet<EntityUid> _toRemove = [];
```
beck-thompson, line 18: Could you remove this? I think you could just remove the super bonk component immediately right?

Content.Server/Administration/Systems/SuperBonkSystem.cs
```diff
@@ -59,49 +56,69 @@ public override void Update(float frameTime)
...
+        // This is done so we don't crash if something like a new table is spawned.
+        while (tables.MoveNext(out var uid, out var comp))
+        {
+            bonks.Add(uid);
+        }
+
+        var hadClumsy = EnsureComp<ClumsyComponent>(target, out _);
 
-        _clumsySystem.HitHeadClumsy((comp.Target, clumsyComp), uid);
+        var component = new SuperBonkComponent
```
beck-thompson, line 96: Use `EnsureComp`! If someone already being bonked it breaks. Also probably don't allow two to get started at once in the first place :thinking:. I feel like it'll lead to more issues :laughing:

beck-thompson (changes requested):
I haven't done a full review but just a few things I noticed!

Content.Server/Administration/Systems/SuperBonkSystem.cs
Winkarst-cpu: You pointed at the remove code.

Content.Server/Administration/Systems/SuperBonkSystem.cs
Winkarst-cpu: See linked issue #32179.

beck-thompson: I don't think its necessary! Why can't you just use RemCompDeferred?

Content.Server/Administration/Systems/SuperBonkSystem.cs
```diff
@@ -59,49 +54,64 @@ public override void Update(float frameTime)
...
+        while (tables.MoveNext(out var uid, out var comp))
+        {
+            bonks.Add(uid);
+        }
 
-        _clumsySystem.HitHeadClumsy((comp.Target, clumsyComp), uid);
+        var hadClumsy = EnsureComp<ClumsyComponent>(target, out _);
 
-        _audioSystem.PlayPvs(clumsyComp.TableBonkSound, comp.Target);
+        var component = new SuperBonkComponent
```
beck-thompson, line 89: I still think just using ensurecomp would be better. Doing it this way doesn't save any lines or anything and feels weirder because addcomp is basically never used. Up to you or the next reviewer though it doesn't matter because of the first check

beck-thompson (approved):
Looks good! The only other thing you could do is just remove the clumsy component from this thing entirely. You'd have to update the public function the clumsy system but it would be easy. Just make it only give a default stun time if the entity has the clumsy component already - because your iterating over all boinkable entities it literally will never get used anyway)

I don't think its ever used and keeping clusmy after the smite feels a little weird in the first place

Winkarst-cpu:
@beck-thompson, if Mob had no clumsy before then it's remove once SuperBonk is https://github.com/space-wizards/space-station-14/pull/35104/files#diff-05f68187ef55aafd77064ed6871b761fa26c2401232158f8b53c528b2cf4af89R47.

beck-thompson:
> @beck-thompson, if Mob had no clumsy before then it's remove once SuperBonk is https://github.com/space-wizards/space-station-14/pull/35104/files#diff-05f68187ef55aafd77064ed6871b761fa26c2401232158f8b53c528b2cf4af89R47.

Yeah but there isn't any reason for them to have clumsy while they are being boinked! The public method in the clumsy system doesn't actually need it because your only bonking against tables which have the bonkable component. The admin smites that call this also never keep clumsy after the smite is over so its not really necessary.

It is a fun thing that could be turned on in the future though so I think its OK to leave it if you want.

beck-thompson (approved):

PR #35105, opened 2025-02-12
Add ability for admins to change certain cvars via command.

Simyon264:
Requires https://github.com/space-wizards/RobustToolbox/pull/5674

## About the PR

Adds a command that downstreams can use to allow more fine grained control about cvar changes.
## Why / Balance

Currently, the only way to change CVars is via the config files or the `cvar` command which is +HOST only. +HOST is a dangerous permission to give. This PR adds a framework to allow certain CVars to be changed simply via the `changecvar` content command. The CVars define their own min and max as well as the AdminFlag required to execute the command.

This is to allow config changes (such as playercount, for example) from admins that do not have +HOST. It will also reduce boilerplate code (see https://github.com/space-wizards/space-station-14/blob/master/Content.Server/Administration/Commands/PanicBunkerCommand.cs as an example). This PR does not actually add any CVars to the list as this is only the framework to do it.

Possible CVars to allow changing in the future:
- The CVars that change when running the `clientsidemappingsetup` command.
- Panic punker settings
- MOTD
- Other CVars that have simple command switches.

## Technical details

This PR adds a new command. The command fetches all CVars that can be changed via reflection once autocomplete is attempted. Why not on command creation? `IPostInjectInit` does not trigger on commands and dependencies are not available on construction.

## Media

Screenshots are slightly outdated.
image
image
image
image

## Requirements

## Breaking changes

(This is not breaking changes, just an FYI for forks!)
A new command `changecvar` has been added to allow non +HOST admins to change CVar values. To achieve this add the `[CVarControl]` attribute. Example:
```cs
namespace Content.Shared.MyAwesomeFork.CCVar;

[CVarDefs]
public sealed partial class MyAwesomeForkCCVars : CVars
{
    /// <summary>
    /// If true, makes people sell their souls on connection.
    /// </summary>
	[CVarControl(AdminFlags.Admin)]
    public static readonly CVarDef<bool> GameAwesomeForkSellSoulOnJoin =
        CVarDef.Create("game.awesome_fork.sell_soul_on_join", true, CVar.SERVERONLY);

    /// <summary>
    /// Whats the chance that the soul gets consumed?
    /// </summary>
	[CVarControl(AdminFlags.Server, min: 0, max: 100)]
    public static readonly CVarDef<int> GameAwesomeForkSellSoulOnJoin =
        CVarDef.Create("game.awesome_fork.chance_to_consume_soul", 20, CVar.SERVERONLY);
}
```

`min` and `max` currently support the following values: `int`,  `float`, `long` and `ushort`.

Optionally you can define a help text for the command to show. Like so:

```ftl
changecvar-simple-debug_test_cvar = Does nothing.
changecvar-full-debug_test_cvar = Just a simple testing cvar. Does nothing.
```

Both the simple and the full text must be present or both null. Replace any `.` with `_`.


FairlySadPanda:
You've listed the reason as:

"This is useful if you do not want to give everyone +HOST for simple playercount changes."

Could you explain why this is tied to admin privilages and not its own privilage? Do all admins need the ability to do this?

Simyon264:
This PR was made specifically after Slam wanted a way to playtest damage values via CVars. He then brought up making a set of commands (Much boilerplate) to change the damage values. I suggested making something like this instead, to which he requested me to do as he was going on vacation. And now here we are.

Resources/Locale/en-US/administration/commands/change-cvar-command.ftl
```diff
@@ -0,0 +1,9 @@
+﻿cmd-changecvar-no-arguments = You must specify a cvar.
```
SlamBamActionman, line 1: Some minor spelling inconsistentcies, is all.

Simyon264: I'll quickly fix them.

SlamBamActionman (approved):
Search tested nicely! Good QoL!

0x6273:
Why is a new command needed for this? Can't you just make the existing command not require HOST ?

Simyon264:
The existing command is in engine and admin flags are in content.

Simyon264:
To further explain:
The base CVar command contains no checks for admin flags. The only way it's gated is behind https://github.com/space-wizards/space-station-14/blob/master/Resources/engineCommandPerms.yml#L130. You can either give a person the permissions to change all CVars or none. 
Botching more access control into the engine for this one specific command is unnecessary work, a new command made specifically in content is much more pratical.

0x6273:
Having two commands that have the exact same function but with different permissions seems a bit hacky. Wouldn't it be better to have a permissions system in the engine instead?

Simyon264:
Yes, I agree that the commands share functionality, but implemting a permission system in RT is a much higher workload and might not even be watned. Also RT supports singleplayer. How would it work with that?

PJB3005:
Yeah this isn't a feature that should be in RT.

PR #35112, opened 2025-02-12
Render suit storage under back layer

### LINKED ISSUE #30816: Gas canister equipped sprites need to be a layer below backpacks
## Description

I wanted to do so in #30809 but I couldn't figure out how. Ideally, it should be above eva/outer clothing/belt but _below_ back slot.

**Screenshots**

**This is what it should look like:**

image

themias:
## About the PR

This PR makes the suit storage layer render under the back layer.
This was desired largely to have gas tanks render under backpacks and hair.

Fixes #30816 

## Why / Balance

IMO, it looks better. Having the layer appear above hair was also a bit weird.

Note: this does mean that other items worn in the suit storage slot would be hidden by big hair, e.g. weapons. However, these are already hidden under big hair if you wear them on your back slot.

It looks like there are a few jetpacks that would be rendered under backpacks, like the mini-jetpack, so that's also a consideration.

## Media

Before: 
before-back-tank
hair-before

After: 
backpack-tank
hair-after

## Requirements

## Breaking changes

I'm not sure if downstreams would have changed the render order here, but might be worth an announcement?

**Changelog**

Probably no CL needed


slarticodefast:
Hmm, from the one screenshot it looks like the sprite is now below the lizard tail, not sure if that is ideal. Or does it have any other conflicts when it's above?

themias:
> Hmm, from the one screenshot it looks like the sprite is now below the lizard tail, not sure if that is ideal. Or does it have any other conflicts when it's above?

I've now moved the tail layer below the back and suit-storage layer. The only conflict I can see is that it is now below the 'neck' layer, so things like capes will now cover tails. This seems okay/logical to me, though.

image
image

I don't think any species that uses the 'tail' layer uses the base layering map besides Reptilians, but it might be worth moving their map out of base.yml at some point if we start adding species again.

themias:
merge conflicted. I might fix and reopen later

Issue #34982, opened 2025-02-08
Map renderer reports inaccurate/misleading UID for objects it can't render


Vortebo:
## Description
When an EffectHearts object (and presumably other particle effect entities) gets saved on the map, the map renderer doesn't seem to know how to deal with it. It will report
`System.Collections.Generic.KeyNotFoundException: Entity XXXX does not have a component of type Robust.Shared.GameObjects.TransformComponent`,
but the ID will not be for the correct object. The object it _does_ point to will most likely have the Transform component, leading to much confusion on the part of the mapper.

As far as I can tell, what's happening is that between the `Painting map [MAP]` and `Loaded client and server in XXXXX ms` messages, every object in the map gets loaded into memory and assigned an ID. For some reason, the EffectHearts object doesn't get the ID it has in the YML. Couldn't tell you why it gets the one that it does, but it seems to be consistent as long as you don't change anything in the map file. Once the renderer makes it to the EntityPainter phase, it will hit this error and fail.

I can think of a few possible end goals here. Whatever is causing the unrenderable object to get a different UID could respect the UID it has in the file, the part of code that throws this exception ( located in RobustToolbox->Robust.Shared->GameObjects->EntityManager.Components.cs->EntityQuery->GetComponent() ) could also report the object type, or a new error could be added which gets thrown when objects like this are present on the map.

**Reproduction**
Open a map using `mapping`. Enter MappingState. Add an `Ian [AI]` object. Click on them. They should emit hearts like you pet them, but the hearts should just sit there rather than animating and going away. Save the map. Attempt to render it.

**Additional context**
This error is mentioned several times in the cartography Discord channel, and it doesn't look like folks have identified a fix for it to this point. I feel it is fairly unlikely other folks made the exact same mistake I made, so I don't know what objects might have been causing the error for them or how they managed to spawn them.

PR #35113, opened 2025-02-12
Fixing (an instance of) MapRenderer "Entity Does Not Have a Component" Error Having Misleading UID

Vortebo:
## About the PR
MapRenderer now reports the name of the problem prototype when it can't be rendered rather than (what I assume is) a client UID which does not correspond to the map file UID. It also skips the object in question and renders the map.

## Why / Balance
There have been a handful of instances where mappers were unable to render their maps due to an entity allegedly not having the transform component. When they checked the given UID in the map file, it would be an object that did indeed have that component, which wasn't particularly helpful for solving the issue.

Addresses https://github.com/space-wizards/space-station-14/issues/34982

## Technical details
I added a check in GridPainter's GetEntities() function after it gets an entity prototype. If that prototype's ID starts with "Effect," we print a message saying to get rid of the object and then ignore it. I checked in the map editor for anything with a name that starts with "Effect" that someone might actually want to place, and I didn't see anything, so I'm fairly confident this shouldn't cause any issues.

I should note that the fix is entirely based on my own experience; I don't know what was causing the issue for other mappers. I expect if there are any other broad groups of objects that generate the error, they can be fairly easily handled in the same if statement.

## Media
N/A

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: MapRenderer should now point the mapper to the correct problem prototype.

### DIFF
--- a/Content.MapRenderer/Painters/GridPainter.cs
+++ b/Content.MapRenderer/Painters/GridPainter.cs
@@ -83,6 +83,11 @@ namespace Content.MapRenderer.Painters
                 {
                     continue;
                 }
+                if (prototype.ID.StartsWith("Effect"))
+                {
+                    Console.WriteLine($"Skipping unrenderable entity {prototype.ID}; please remove it from the map.");
+                    continue;
+                }
 
                 var transform = _sEntityManager.GetComponent<TransformComponent>(serverEntity);
                 if (_sEntityManager.TryGetComponent(transform.GridUid, out MapGridComponent? grid))
--- a/Content.MapRenderer/Painters/GridPainter.cs
+++ b/Content.MapRenderer/Painters/GridPainter.cs
@@ -85,7 +85,7 @@ namespace Content.MapRenderer.Painters
                 }
                 if (prototype.ID.StartsWith("Effect"))
                 {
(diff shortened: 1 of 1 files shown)


Content.MapRenderer/Painters/GridPainter.cs
```diff
@@ -83,6 +83,11 @@ private ConcurrentDictionary<EntityUid, List<EntityData>> GetEntities()
                 {
                     continue;
                 }
+                if (prototype.ID.StartsWith("Effect"))
+                {
+                    Console.WriteLine($"Skipping unrenderable entity {prototype.ID}; please remove it from the map");
+                    continue;
+                }
```
metalgearsloth, line 90: ...Effect?

Vortebo: For me the issue was caused by EffectHearts, so I sort of ran on the assumption that there is a class of entities spawned for quick animations, and their names all start with "Effect"

metalgearsloth: I don't think we should be doing magic string checks here, if anything these should be flagged as not map saveable or the likes.

Vortebo: Oh, if that's an option then yeah definitely

PR #35117, opened 2025-02-12
Makes the Long-Range Holopad Constructible

Centronias:
## About the PR

- Added Long-Range holopad circuit board
- Long range holopads can be constructed with their board
- Normal holopad boards and long-range boards can swap into each other, like freezer/heater boards

## Why / Balance

Made a shuttle last night, was sad when I couldn't put a long range holopad in it

"Balance": LR pads take 8 capacitors while normal ones take 4 because, uh, I guess the capacitance determines the range or something.

## Technical details

- Added the LR board
- Made it so the LR pad uses the LR board
- Made a construction graph so that the boards can swap between each other, and applied the graph to the boards

## Media

## Breaking changes

No C# changes.

**Changelog**

:cl:
- add: Long-Range Holopad circuitboards, which can be obtained by using a screwdriver on a holopad board (similar to thermomachine boards), and which can be used to construct Long-Range Holopads.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/Circuitboards/Machine/holopad.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/Circuitboards/Machine/holopad.yml
@@ -2,7 +2,7 @@
   id: HolopadMachineCircuitboard
   parent: BaseMachineCircuitboard
   name: holopad machine board
-  description: A machine printed circuit board for a holopad.
+  description: A machine printed circuit board for a holopad. Looks like you could use a screwdriver to change the board type.
   components:
     - type: MachineBoard
       prototype: Holopad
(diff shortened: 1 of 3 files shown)


ArtisticRoomba (approved):
LGTM

chromiumboy:
I'd say 8 capacitors is a bit steep. Maybe reduce the standard holopad to 2 capacitors and the long range to 4 instead.

Not sure if long range and standard holopad curcuitboards should be interchangeable. We might want some rarer materials required to print the long range version

Centronias:
I will say without reservations that I don't play enough science to have any feel for balance when it comes to material costs. My thinking with making them interchangeable was that it means there's not yet another thing showing up in the R&D interface, and not yet another thing extending the board printer's list.

Issue #8524, opened 2022-05-29
Current Freezes


wrexbe:
This is a list of things that will probably not get merged right now, while you can still make a PR, you may have to resolve conflicts after the PR that is causing the freeze is merged.

Freezes / PRs that are in progress
* Maps - https://github.com/space-wizards/space-station-14/pull/8301
* UI - https://github.com/space-wizards/space-station-14/pull/7202

Zumorica:
Instruments - Currently refactoring this entirely and moving all of the networking logic to engine. I'm also refactoring most of the engine MIDI code.

EmoGarbage404:
EntityStorage - Undergoing a refactor at #9291

ZeroDayDaemon:
Bartending drinks / reagents / sprites - Undergoing full conversion to /tg/ with the new /tg/ sprites

EmoGarbage404:
Uplink as of #10201

metalgearsloth:
Antags - We haven't actually had a meeting since I proposed it so it's tentatively here.
Reason is people work on a new antag (gameplay for one person occasionally) instead of job content (gameplay actively in every round for several players) so we continue our spiral of having no job content for departments like science.

Fixing an existing antag is okay.

Injazz:
xenosci was merged so consider lifting freeze from research

metalgearsloth:
do_afters as keron is currently refactoring them to be less poopy.

juliangiebel:
Anything related to Machile linking. So using or changing calls to methods of [SignalLinkerSystem](https://github.com/space-wizards/space-station-14/blob/master/Content.Server/MachineLinking/System/SignalLinkerSystem.cs).
I have a refactor PR up that moves linking on top of device networking. #13645

EmoGarbage404:
Gonna ask that people don't touch gamemodes/station events while i rework on gamerule refactor.

keronshb:
Restoring the antag freeze.

deltanedas:
does borging count as species i wanna try it

metalgearsloth:
> does borging count as species i wanna try it

We're discussing freezes more in discord but no. Borgs need to use body system though.

deltanedas:
yeah ill be basing it off current body rework code, seems like brain will be fun fun fun

EmoGarbage404:
[...]

PR #35119, opened 2025-02-12
Make it impossible to talk while suffocating

### LINKED ISSUE #33485: You shouldn't be able to speak, scream, etc. when there is nothing in your lungs
The voice and other sounds are the result of air going through the vocal cords, so when your lungs are empty, you are muted.

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Resolves #33485. Reasons:

1. Realism - voice is the result of air going through the vocal cords, so when you are suffocating you can't talk.

2. Gameplay - you can use it in combat to prevent your enemies from communicating or prepare a vacuum trap for somebody.

3. Roleplay - if you work in space with somebody and your oxygen tank runs out, then you will have to communicate non-verbally explain your problem to the other person.

## Media

https://github.com/user-attachments/assets/2e62f43c-5d77-4757-b6f7-a5be250e5fe6

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Now you can't talk while you are suffocating.


Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -59,6 +69,49 @@ private void OnUnpaused(Entity<RespiratorComponent> ent, ref EntityUnpausedEvent
...
+    private void OnEmote(Entity<RespiratorComponent> ent, ref EmoteEvent args)
+    {
+        if (args.Handled)
+            return;
+
+        var (_, component) = ent;
+
+        if (component.Saturation < component.SuffocationThreshold)
+            args.Handled = true;
+    }
```
lzk228, line 95: does it work on body emotes?

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -18,6 +18,12 @@
 using JetBrains.Annotations;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Timing;
+using Content.Shared.Speech;
+using Content.Server.Popups;
+using Robust.Shared.Player;
+using Content.Shared.IdentityManagement;
+using Content.Shared.Chat.Prototypes;
+using Content.Server.Speech.EntitySystems;
```
Aeshus, line 26: **nitpick:** Please make sure that the imports are ordered!

UbaserB:
i like this a lot

ArtisticRoomba:
extremely based PR, and so the roleplay door opens....

LaCumbiaDelCoronavirus:
I like the concept, but being completely unable to talk, and also not making it clear to others that you are even attempting to talk doesn't seem very interesting; how about something like just for forcing you to whisper?
And, does this effect also apply when gasping for air after you were suffocating? (like how you will gasp if you still have asphyxiation damage)

ArtisticRoomba:
You're gasping when you're doing this so it should be visible to players that you don't have any air, also you can emote.
*Roomba Angle points vigorously at their oxygen tank!*

Hrosts:
Having an emote like "tries to speak, but makes no sound" play for people around would be better than having people see you typing only to receive nothing afterwards and think you changed your mind.

Coolsurf6:
> Having an emote like "tries to speak/shout, but makes no sound" play for people around would be better than having people see you typing only to receive nothing afterwards and think you changed your mind.

Maybe instead, just have them gasp if they try and type, giving the critical message to the entity of them having no air to warn people, but also just means when people type, they gasp.

I think there should be a range though, those who are gasping from 0 air should not be able to speak, those gasping when there is still air around should be able to speak, but still maybe gasping, or even have their text slightly muffled or censored not being able to string words correctly.

This makes fully sense for spaced environments, but if the atmos is just under what you can breath, I feeeeel like you should still be able to cry out to atmos for help.

Winkarst-cpu:
> Having an emote like "tries to speak/shout, but makes no sound" play for people around would be better than having people see you typing only to receive nothing afterwards and think you changed your mind.

If you try to talk, then others will see a popup with text "{THE($target)} tries to speak, but nothing comes out of {POSS-ADJ($target)} mouth!".

FairlySadPanda:
As a warning: https://github.com/space-wizards/space-station-14/issues/8524

Specifically the final freeze about chat code:

`Chatcode - PRs involving chatcode or adding/changing formatting functionality are at risk of being closed due to impacting refactoring chatcode.`

Whilst this is tied into RespiratorSystem this is adding a new dependency to `Content.Shared.Chat.Prototypes` (and some Speech code).

slarticodefast:
I would like to see some reasoning in the PR description on why this is good for the game other than realism.
It does add a few interesting scenarios for roleplay and emergent gameplay in my opinion.
For example two salvagers out in space, one runs out of oxygen and now has to emote his problem to the other one, which is a fun situation. Being forced to use imotes instead of speaking forces you to be creative, which is why playing mime is fun (although that might not something everyone likes).
In combat it can be creatively used by spacing areas to prevent your opponent from communicating, which means this is a huge buff for nukies.
It also allows you could prepare a vacuum trap for someone, which prevents the victim from calling for help over radio, similar to the radio jammer.
On the other hand this can be very frustrating if the whole station is spaced.

Simyon264:
Hello,

this PR was brought up to a maintainer discussion. 
We wish that you provide a clear reason why you would propose this change.

Please edit your PR description to include a reason on why this is good for the game.

Thank you for your contribtuon.

Kadeo64:
>Realism - voice is the result of air going through the vocal cords, so when you are suffocating you can't talk.

when has ss13/14 ever put realism at the forefront of balance

>Gameplay - you can use it in combat to prevent your enemies from communicating or prepare a vacuum trap for somebody.

nukie spacing is already cancerous this will make it 300x worse i agree with slarti's '_`On the other hand this can be very frustrating if the whole station is spaced. Being unable to talk in a multiplayer roleplay game might make it less fun, so this could be a case where gameplay is more important than realism.`_' point

>Roleplay - if you work in space with somebody and your oxygen tank runs out, then you will have to communicate non-verbally explain your problem to the other person.

i do agree that this will be great for roleplay.. maybe on downstream forks that have a focus on roleplay (or salamander). I don't think it's needed on LRP as i'm gonna be honest this will increase people abusing LOOC/Emotes to speak more than they already do 

overall good PR but maybe not the best server to PR it to? idk turning this into a Cvar would be cool. also chatcode refactor being on the horizon means that this might break when that gets merged.

SlamBamActionman:
Re: chat refactor, it would break this. But should hopefully not be difficult to transfer over either. If this PR gets approved I would suggest a Conceptually Approved tag until the refactor is done.

FairlySadPanda:
To provide an upside: this PR proposes making it impossible to speak _when you are suffocating_, not when there's no atmos to convey the speech. I.e. if the nukies have busted through the walls like kool-aid men, you just need to be on internals to speak, and only once your lungs have dumped their atmos. You should be on internals to _live_, so in the heirarchy of spess needs someone who is screaming NUKIES CARGO should be getting their internals on anyway.

Winkarst-cpu:
Will prevent you from talking while breathing for example CO2, bad implementation over all.

Simyon264:
This PR has been conceptually approved by the maintaners.

This means that if anyone PRs this concept in the future, it'll have to fulfill the following conditions:
- CVar toggle for easy testing, defaults to off

PR #35121, opened 2025-02-13
Added Coin Flipping And More

GrownSamoyedDog:
## About the PR
- Added in Coin Flipping as a Mechanic
- Remade the existing Coin Sprites (Sprites were from VGstation)
- Added in New Coins! One can be found in Good Clean Fun vendor next to dice.

## Why / Balance
Fun. Lots of RP potential. Is a feature in SS13.

We already have Dice, but not coin flipping. Can't say Dice are complete if you don't have a D2.

Coins already exist as treasure/loot, but is largely useless except for as a small benefit to Cargo.

## Technical details
- I moved coin stuff from /treasure/ to /fun/ to stay consistent with dice.
- A lot of the coin flipping code is similar to dice rolling code, but different enough to warrant new files.

Coins have a 0.1% chance (1 in 1000 odds) to land on their side instead of Heads/Tails.

I added coins to migration.yml which mean the few that started as salvager loot treasure should still appear as such. Updated, but similar appearances.

## Media
diagram

https://github.com/user-attachments/assets/a03ab196-e6af-4e1b-a6ec-378b599f1693

## Requirements

## Breaking changes
- Changed many coin IDs (but added to migration.yml)
- Changed coin folder from /treasure/ to /fun/

**Changelog**

:cl:
- add: coin flipping and new coins
- tweak: moved coin folders from /treasure/ to /fun/
- tweak: changed coin IDs and added to migration.yml


lzk228:
keep only player facing changes in the changelog
e.g.
- add: Added ability to flip coins

slarticodefast:
I only looked briefly at the code, but this will need some changes:
- Move it to shared and predict it. Look at the open PR that makes dice predicted.
- You can use a generic visualizer for this instead of directly setting layers.
- The three sides and their probabilites are hardcoded. I would generalize it to any set of weighted states, which could be defined in a dictionary. This would allow for example a weighted coin that lands more often on one side. This can maybe even be unified with the DiceSystem, which is very similar, to avoid code duplication.
- Same for the strings, those should all be localized and not hardcoded.

GrownSamoyedDog:
> I only looked briefly at the code, but this will need some changes:
> 
> * Move it to shared and predict it. Look at the open PR that makes dice predicted.
> * You can use a generic visualizer for this instead of directly setting layers.
> * The three sides and their probabilites are hardcoded. I would generalize it to any set of weighted states, which could be defined in a dictionary. This would allow for example a weighted coin that lands more often on one side. This can maybe even be unified with the DiceSystem, which is very similar, to avoid code duplication.
> * Same for the strings, those should all be localized and not hardcoded.

I will try to add all those though I am slow

Mixelz:
> * The three sides and their probabilites are hardcoded. I would generalize it to any set of weighted states, which could be defined in a dictionary. This would allow for example a weighted coin that lands more often on one side. This can maybe even be unified with the DiceSystem, which is very similar, to avoid code duplication.

With this implementation, I would love to see a 1 TC Syndicate Coin that can be forced to always land either heads or tails.

Just-a-Unity-Dev:
Changelog should be like

- add: Added new coin flipping/etc etc...

slarticodefast:
Hello, and thank you for your contribution!
At the moment maintainers are facing a high backlog and unfortunately can’t review every single pull request. As this PR pertains to the addition of new content and/or balance changes deemed low-impact, we have opted to close it instead of leaving it open indefinitely.
PRs may be reopened in the future once the backlog has been reduced, and we encourage you to keep contributing in other areas of the codebase.

PR #35058, opened 2025-02-11
Sentry turrets - Part 3: Turret AI

chromiumboy:
## About the PR

This PR adds the code needed for the sentry turrets' AI. Namely, it adds the following:
- The ability for turrets to attack targets that are standing behind windows and other non-opaque entities (so they can fire their lasers through windows)
- The ability for turrets to ignore a target if they have specific access levels on their ID card or body
- The ability to switch the turret's AI on/off (so they don't attack when inactive)

This PR also makes the following changes / additions:
- NPCs will no longer make ranged attacks against targets in a critical health state (I can remove this, or make a separate check specific to turrets rather than ranged NPCs in general)
- Three new access levels have been added: Station AI, Cyborg and Robot (these are needed for turrets to determine if borgs and bots are valid targets; the station AI access level will be used in a later PR to control AI-specific turrets). These new access levels fall under the new Silicon access group. The station AI and admin ghosts have had the Silicon access group added to their access lists, as well as the universal ID card and universal access configurator
- The access levels under the Silicon access group, as well as EmergencyShuttleRepealAll, have been marked as  'canAddToIdCard: false'. In practice, it means that these access levels cannot be gained when microwaving an ID card.
- The Command access level group has been expanded to include all the heads of the departments. This will be used in a later PR for populating the turret control panel UI

## Why / Balance

This PR will be required for later PRs in this series

## Technical details

N/A

## Media

N/A

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

N/A

**Changelog**

N/A

PR #35123, opened 2025-02-13
Sentry turrets - Part 4: The sentry turret and its primary systems

chromiumboy:
## About the PR

This PR adds the sentry turret, an autonomous weapon system designed to defend critical areas of the station. Sentry turrets are stowed within the floor while inactive for safekeeping. However, when interacted with using the 'E' key or right-click menu they will emerge and enter a combat-ready state

Sentry turrets will fire upon any nearby valid biological or mechanical targets until deactivated. Only those who have the proper access credentials are exempt from being fired upon (any default settings are determined in the entity prototype). The sentry turret's weapon systems typically draw their power from an APC power net, however, they do possess an internal battery to maintain weapons fire for a time if their external power supply is interrupted

There are two sentry turret prototypes included in this PR. The first type is the AI sentry turret. Only the station AI can interact with these devices - it's intended to defend the AI core from any intruders, whether they are crew or not. Only cyborgs or robots can pass by these turrets safely. The second type is intended for a more general defense of the station. Any individual with security access can safely interact with them

A subsequent PR will introduce wall-mounted sentry turret control panels. These will be used to synchronize the weapon settings of any turrets linked to it, as well as which personnel are permitted to pass by unscathed

This PR requires PRs #35025, #35031 and #35058

Requires https://github.com/space-wizards/RobustToolbox/pull/5755

## Why / Balance

Sentry turrets have the following stats/features
- Two weapon settings: disabler and laser bolts (toggleable using the right-click menu)
- Disabler bolts reduce stamina by 33 stamina (the same as a disabler) but deal no physical damage
- Laser bolts deal 28 heat damage (the same as a laser cannon)
- These projectiles will pass through glass (sentry turrets will also engage targets that are behind glass, see PR #35058)
- Projectile speed: 25 units per second (shared between disabler and laser bolts. Note that disabler shots were recently increased to 35 units per second)
- Fire rate: 1.5 shots per second (shared between disabler and laser bolts)
- Battery capacity: 20 shots (the battery will recharge while connected to APC power, providing effectively infinite shots)
- Health
	- Broken threshold: 300 units (becomes inoperable)
	- Destroyed threshold: 600 units (returns to its machine frame state)
- Damage modifiers
	- Blunt: 70% (plus a flat 5 blunt damage reduction while retracted)
	- Slash: 50%
	- Piercing: 70%
	- Shock: 120%
	- Heat: 100% (plus a flat 5 heat damage reduction while retracted)
	- Structural 50% (while retracted)
- Can be repaired with a welding tool (10s do after)
- Deployment/retraction time: 1.2 seconds
- APC load: 5 (idle) / 250 (recharging battery)
- Access-locked maintenance panel (only accessible while the turret is inactive)
- Wires panel: 4 dummy wires, 2 power wires, AI interaction wire, access wire
- Construction: requires a machine frame, a sentry turret machine board, a laser cannon, a proximity sensor, and a battery (battery quality has no effect)

## Technical details

The majority of the sentry turrets' behavior is contained in the SharedDeployableTurretSystem. Its server-side system handles tracking of the turrets' current state, when they are damaged/repaired, and HTNs while the client-side system plays the turrets' animations and applies visual state changes

Changes have been made to BatteryWeaponFireModesSystem to accommodate the turrets. These include the ability to prevent  individuals from changing a weapon's fire mode unless they have the appropriate access credentials and applying visual state changes when the fire mode is changed

## Mapping
AI sentry turrets should only be mapped to the station AI core and/or the AI satellite. Regular sentry turrets can be mapped to any other locations on the station (e.g., the station vault, the armory) at the maptainers discretion (if any are mapped at all). Do not map sentry turrets of different types to the same room

Sentry turret prototypes only differ by who can access them (set under the Access component) and who they will not target (set under the TurretTargetSettings component). Additional prototypes with different default settings can be easily created if certain use cases crop up

Note that in the future there will be a follow-up PR which will introduce wall-mounted sentry turret control panels. These devices will be able to configure the settings of multiple turrets simultaneously, including what personnel are authorized to pass the turrets while they are active

## Media

[Old demonstration video of the sentry turret in action](https://youtu.be/amWWmgHgaKU)

See PR #35025 for additional media.

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Sentry turrets can potentially be found guarding sensitive areas of the station. When deployed, they will shoot unauthorized personnel on sight with either stunning or lethal laser bolts. Note that only cyborgs and robots can safely pass sentry turrets that protect the station AI core


ScarKy0:
Im not entirely sure if we need a changelog here
if we merge this and not get everything else done and mapped by next release people will be confused

chromiumboy:
> Im not entirely sure if we need a changelog here if we merge this and not get everything else done and mapped by next release people will be confused

I don't think this will be a issue, as players know by now that maps can take a while to update with new features. I'll tweaked the changelog though

Vinkoln:
we are waiting

Content.Shared/Weapons/Ranged/Systems/BatteryWeaponFireModesSystem.cs
```diff
@@ -74,12 +80,12 @@ private void OnGetVerb(EntityUid uid, BatteryWeaponFireModesComponent component,
...
-    private void OnInteractHandEvent(EntityUid uid, BatteryWeaponFireModesComponent component, ActivateInWorldEvent args)
+    private void OnUseInHandEvent(EntityUid uid, BatteryWeaponFireModesComponent component, UseInHandEvent args)
     {
-        if (!args.Complex)
+        if (component.FireModes.Count < 2)
             return;
 
-        if (component.FireModes.Count < 2)
+        if (TryComp<AccessReaderComponent>(uid, out var accessReader) && !_accessReaderSystem.IsAllowed(args.User, uid, accessReader))
             return;
```
metalgearsloth, line 89: This should be in cycle or a trycycle method rather than every caller having to handle it.

Content.Shared/Weapons/Ranged/Systems/BatteryWeaponFireModesSystem.cs
```diff
@@ -44,12 +47,15 @@ private BatteryWeaponFireMode GetMode(BatteryWeaponFireModesComponent component)
...
-        if (!args.CanAccess || !args.CanInteract || args.Hands == null)
+        if (!args.CanAccess || !args.CanInteract || !args.CanComplexInteract)
             return;
 
         if (component.FireModes.Count < 2)
             return;
 
+        if (TryComp<AccessReaderComponent>(uid, out var accessReader) && !_accessReaderSystem.IsAllowed(args.User, uid, accessReader))
+            return;
+
```
metalgearsloth, line 58: See below.

Content.Shared/Turrets/SharedDeployableTurretSystem.cs
```diff
@@ -0,0 +1,170 @@
...
+        if (TryComp<HitscanBatteryAmmoProviderComponent>(ent, out var hitscanBatteryAmmo) &&
+            (hitscanBatteryAmmo.Shots > 0 || ent.Comp.Powered))
+            return true;
+
+        if (TryComp<BallisticAmmoProviderComponent>(ent, out var ballisticAmmo) &&
+            ballisticAmmo.Count > 0)
+            return true;
+
+        return false;
+    }
```
metalgearsloth, line 169: This is really hacky and not ECS as it's just going to break if any new components get used, it should be using events to check.

Content.Shared/Turrets/DeployableTurretComponent.cs
```diff
@@ -0,0 +1,172 @@
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Broken = false;
+
+    /// <summary>
+    /// Whether the turret is currently powered (required until power is predicted)
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Powered = false;
+
```
metalgearsloth, line 33: We have components for this, components shouldn't be storing it.

Content.Server/Turrets/DeployableTurretSystem.cs
```diff
@@ -0,0 +1,174 @@
...
+
+    private void OnRepaired(Entity<DeployableTurretComponent> ent, ref RepairedEvent args)
+    {
+        ent.Comp.Broken = false;
+        Dirty(ent);
+
+        if (TryComp<AppearanceComponent>(ent, out var appearance))
+            _appearance.SetData(ent, DeployableTurretVisuals.Broken, false, appearance);
+    }
+
```
metalgearsloth, line 79: Wouldn't it be better to just combine this with other broken components rather than duplicating it?

metalgearsloth (changes requested):
As is the deployableturret stuff is too singleton and hardcoded specifically for AI turrets so nothing else can really use the code.

Content.Server/Turrets/DeployableTurretSystem.cs
chromiumboy: Do we have a 'broken' component? The destructible component doesn't store the state of the entity, it just sends an event when it passes a threshold

Content.Shared/Turrets/DeployableTurretComponent.cs
metalgearsloth: The compstate is networked. Even if the power change itself isn't predicted the client will still know if it's powered or not 99% of the time.

Content.Shared/Weapons/Ranged/Systems/BatteryWeaponFireModesSystem.cs
metalgearsloth: Thinking about it I would just split this out to a new component, make an event like "CycleAttemptEvent", and then also make the other cycle code use it.

Content.Client/Turrets/DeployableTurretSystem.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        if (args.Key != DeployableTurretComponent.AnimationKey)
+            return;
+
+        if (!TryComp<SpriteComponent>(ent, out var sprite))
+            return;
+
+        if (!TryComp<AnimationPlayerComponent>(ent, out var animPlayer))
+            return;
+
```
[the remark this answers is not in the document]
chromiumboy: The animation player system doesn't require the sprite comp, so I'll just add the animation player comp to AnimationCompletedEvent

Content.Server/Turrets/DeployableTurretSystem.cs
metalgearsloth: I asked in maints chat but atm we have 4 different components storing it manually.

chromiumboy: I addressed this by adding the bool to the destructible component. It gets  set true when a breakage or destroyed behavior triggers, so its should be more widely useable

chromiumboy:
Build errors are because the engine PR needs to be merged first

ArtisticRoomba:
Put "Requires (PR)" in your PR comment in a newline so the bot picks it up and reruns tests with the engine PR.

Content.Shared/Turrets/SharedDeployableTurretSystem.cs
```diff
@@ -0,0 +1,163 @@
...
+
+    private void OnGetVerb(Entity<DeployableTurretComponent> ent, ref GetVerbsEvent<Verb> args)
+    {
+        if (!args.CanAccess || !args.CanInteract || !args.CanComplexInteract)
+            return;
+
+        if (TryComp<AccessReaderComponent>(ent, out var accessReader) && !_accessReader.IsAllowed(args.User, ent, accessReader))
+            return;
+
```
metalgearsloth, line 46: Don't think the trycomp on accessreader required as IsALlowed should be handling the resolve.

Content.Shared/Access/Components/AccessComponent.cs
```diff
@@ -57,3 +57,17 @@ public void AddGroup(ProtoId<AccessGroupPrototype> group)
...
+[ByRefEvent]
+public sealed class AccessAttemptedEvent : CancellableEntityEventArgs
+{
+    public EntityUid? User;
+
+    public AccessAttemptedEvent(EntityUid? user)
+    {
+        User = user;
+    }
+}
```
metalgearsloth, line 73: GunSystem should just be calling IsAllowed but inside the TryX methods, this is just an event wrapping a method call atm.

Content.Shared/Turrets/DeployableTurretComponent.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// </summary>
+    [DataField]
+    public float RetractionLength = 1.19f;
+
+    /// <summary>
+    /// The time that the current animation should complete (in seconds)
+    /// </summary>
+    [DataField]
+    public TimeSpan AnimationCompletionTime = TimeSpan.Zero;
+
```
metalgearsloth, line 93: Needs AutoPausedField so it handles persistence properly.

Content.Shared/Turrets/DeployableTurretComponent.cs
```diff
@@ -0,0 +1,161 @@
...
+    public ProtoId<DamageModifierSetPrototype>? DeployedDamageModifierSetId;
+
+    #region: Sound data
+
+    /// <summary>
+    /// Sound to play when denied access to the turret.
+    /// </summary>
+    [DataField]
+    public SoundSpecifier AccessDeniedSound = new SoundPathSpecifier("/Audio/Machines/custom_deny.ogg");
+
```
metalgearsloth, line 59: Can this definitely not go on access?

Content.Shared/Turrets/SharedDeployableTurretSystem.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            _physics.SetHard(ent, fixture, enabled);
+        }
+
+        var msg = enabled ? "deployable-turret-component-activating" : "deployable-turret-component-deactivating";
+        _popup.PopupClient(Loc.GetString(msg), ent, user);
+
+        ent.Comp.Enabled = enabled;
+        Dirty(ent);
+    }
```
metalgearsloth, line 154: If using delta fields need to call dirtyfield instead for all calls. ATM the state is so small it's not worth it but if we want it to be editable by admins then more of the fields would need autonetworkedfield for it.

metalgearsloth (changes requested):

Content.Shared/Turrets/DeployableTurretComponent.cs
chromiumboy: I mean, it can, but this is how other access denied sounds are currently being handled - it'll need to be cleaned up, but I think that's outside the scope of this PR

chromiumboy:
[The PR has been approved by Sloth on Discord](https://discord.com/channels/310555209753690112/900426319433728030/1354658451686555783)

Waiting for tests to finish running

ScarKy0:
> Was the cool menu where you could control the turrets cut out?

Its part of another PR. They will be out soon tm

chromiumboy:
> Was the cool menu where you could control the turrets cut out?

Nope, it's right here - #35235

PR #35136, opened 2025-02-13
Filter prototypes checked by NoStaticPriceAndStackPrice test and skip spawning

Tayrtahn:
## About the PR

NoStaticPriceAndStackPrice now only checks entity prototypes that have a `StaticPriceComponent`. It also no longer bothers to spawn and delete each entity; instead it just examines the prototype data.

## Technical details

Currently the test spawns and checks every `EntityPrototype` that does not have a `MapGridComponent` (and isn't abstract or a test prototype), but the logic of the test only actually cares about entities that have a `StaticPriceComponent`.

Now the prototypes are filtered before spawning, which massively cuts down on the number of entities that need to be spawned (from ~9000 to ~4000).

I suspect that this was also the reason why this test sometimes randomly failed, since it wasn't filtering out some prototypes that could do weird things (possibly with `RoomFillComponent`?).

There's also no reason to actually spawn in the entities - the test only examines a couple of component fields that are set in YAML. So the test no longer spawns a test map or any of the entities; it just iterates through the prototypes and checks the values there. I don't think this was a particularly slow test before, but it should be faster now without having to do that work.

## Requirements

## Breaking changes

**Changelog**

Nah.

### DIFF
--- a/Content.IntegrationTests/Tests/CargoTest.cs
+++ b/Content.IntegrationTests/Tests/CargoTest.cs
@@ -128,23 +128,20 @@ public sealed class CargoTest
             {
                 var ent = entManager.SpawnEntity(proto, coord);
 
+                Assert.That(entManager.TryGetComponent<StaticPriceComponent>(ent, out var staticPriceComp), Is.True,
+                    $"The prototype {proto} has a StaticPriceComponent but somehow does not when spawned?"); // Sanity check
+
                 if (entManager.TryGetComponent<StackPriceComponent>(ent, out var stackpricecomp)
                     && stackpricecomp.Price > 0)
                 {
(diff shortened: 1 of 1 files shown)


slarticodefast:
> I suspect that this was also the reason why this test sometimes randomly failed, since it wasn't filtering out some prototypes that could do weird things

I added some logging to the test before and let it run over night to catch that heisentest. For some cursed reason it only happens if a random crystal spawner spawns a crystal over a space shroom that is spawned by another random spawner, which is why it was so rare. I wasn't able to reproduce it manually though.

However, while this does not solve the underlying problem for the heisentest, that is not what the test was supposed to test for and this PR is a good optimization.

slarticodefast (approved):

PR #35088, opened 2025-02-11
Fix thrusters not dealing damage

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Fixes #34685.

## Media

https://github.com/user-attachments/assets/8d0ac9b6-de66-40b8-b681-b7e0021746a6

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- fix: Thrusters now deal heat damage.

PR #35143, opened 2025-02-13
minor thruster system cleanup

### LINKED ISSUE #35141: ThrusterSystem NextFire needs fixing
One-time values should be done in MapInit never ComponentInit or else persistence breaks.

See https://github.com/space-wizards/space-station-14/pull/35088

slarticodefast:
Fixes a minor thing I missed in the review of https://github.com/space-wizards/space-station-14/pull/35088.
Fixes #35141

The intial time should be set on map init for persistence reasons.


metalgearsloth (approved):

Winkarst-cpu:
#35142

slarticodefast:
Oh, did not see that https://github.com/space-wizards/space-station-14/pull/35142 was faster
I like this version more though since it does not do the `var (_, component) = ent;`

PR #35159, opened 2025-02-14
Fix: AI can pet itself

### LINKED ISSUE #34796: AI can pet itself

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Fixes #34796.

## Media

https://github.com/user-attachments/assets/7f4bdaef-4d82-4327-88fd-2b8dd21e4bbf

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/Interaction/InteractionPopupSystem.cs
+++ b/Content.Shared/Interaction/InteractionPopupSystem.cs
@@ -1,4 +1,5 @@
 using Content.Shared.Bed.Sleep;
+using Content.Shared.Hands.Components;
 using Content.Shared.IdentityManagement;
 using Content.Shared.Interaction.Components;
 using Content.Shared.Interaction.Events;
@@ -59,6 +60,9 @@ public sealed class InteractionPopupSystem : EntitySystem
 
         //Handling does nothing and this thing annoyingly plays way too often.
         // HUH? What does this comment even mean?
+
+        if (!HasComp<HandsComponent>(user))
+            return;
 
         if (HasComp<SleepingComponent>(uid))
             return;
--- a/Content.Shared/Interaction/InteractionPopupSystem.cs
+++ b/Content.Shared/Interaction/InteractionPopupSystem.cs
@@ -45,6 +45,9 @@ public sealed class InteractionPopupSystem : EntitySystem
 
     private void OnInteractHand(EntityUid uid, InteractionPopupComponent component, InteractHandEvent args)
     {
+        if (!HasComp<HandsComponent>(args.User))
+            return;
+
         SharedInteract(uid, component, args, args.Target, args.User);
     }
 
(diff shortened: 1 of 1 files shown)


Content.Shared/Interaction/InteractionPopupSystem.cs
```diff
@@ -60,6 +61,9 @@ private void SharedInteract(
         //Handling does nothing and this thing annoyingly plays way too often.
         // HUH? What does this comment even mean?
 
+        if (!HasComp<HandsComponent>(user))
+            return;
+
```
slarticodefast, line 66: Hmm, this should probably be handled by `args.Complex` above, which seems to exist exactly for that. Or does the AI need that for other interactions?
If that does not work these two lines should be moved into `OnActivateInWorld` since `InteractHandEvent` already ensures that the user has hands.

Winkarst-cpu: > Hmm, this should probably be handled by `args.Complex` above, which seems to exist exactly for that. Or does the AI need that for other interactions? If that does not work these two lines should be moved into `OnActivateInWorld` since `InteractHandEvent` already ensures that the user has hands.

AI can do complex interactions since it has ``ComplexInteractionComponent`` (https://github.com/space-wizards/space-station-14/blob/0227afec4cd9061c17336db37c858d2cb09bc35b/Resources/Prototypes/Entities/Mobs/Player/silicon.yml#L428), also it raises ``InteractHandEvent`` instead of ``OnActivateInWorld``.

slarticodefast: > Also it raises InteractHandEvent instead of OnActivateInWorld.

Wait, InteractHandEvent can be raised without having any hands? That sound wrong.

Winkarst-cpu: https://github.com/space-wizards/space-station-14/blob/0227afec4cd9061c17336db37c858d2cb09bc35b/Content.Shared/Interaction/SharedInteractionSystem.cs#L362

SlamBamActionman (approved):
Works as advertised

no petta da doggo...

slarticodefast:
@SlamBamActionman I'm not sure this implementation is a good approach.
We used to have a check for `HandsComponent` everywhere and we removed them on purpose in favor of complex interaction to have things more consistent. This would just hardcode hands back in.

SlamBamActionman:
> @SlamBamActionman I'm not sure this implementation is a good approach. We used to have a check for `HandsComponent` everywhere and we removed them on purpose in favor of complex interaction to have things more consistent. This would just hardcode hands back in.

Alright, looking deeper into the issue I see I misunderstood the purpose of having it be a complex interaction. Makes sense to avoid looking for hands specifically 👍

Winkarst-cpu:
``InteractHandEvent`` should be reworked to accommodate it's name since right now it only checks for complex interactions. I'm closing this PR since we need a better solution.

PR #35162, opened 2025-02-14
change radiation stacking logic

Killerqu00:
## About the PR
This implements a system that allows stacks of items to apply rads only on each Nth item in the stack.

## Why / Balance
Before it was reverted, large amounts of uranium on a person with high/full rad resist could kill people in a minute or so without almost any symptoms. Even if you get symptoms, time to kill is still too high.

## Technical details
Radiation from stackable items now has an optional field RadiationScaleMultiplier - this increases radiation intensity on every Nth item in the stack. For example, if set to 5, stack with <5 items are not radioactive, stacks with 5-9 items have 1x intensity... and stack with 30 items (full) applies 6x intensity.

## Media
image
image

## Requirements

## Breaking changes
nuh uh

**Changelog**
non-player facing unless something implements this


VerinSenpai:
How am I only just learning you get rads from uranium

ArtisticRoomba:
I do still believe that you should require a radiation suit to handle large amounts of uranium, it seems quite silly that you can just carry the refined stuff around and it gives a use for radiation suits. But yeah, there is a problem with it being too spread out apparently.

deltanedas:
believe it or not unenriched uranium isnt shooting out 5gw of xrays or whatever, its mostly alpha radiation which is blocked by paper

Resources/Prototypes/Entities/Objects/Materials/Sheets/other.yml
```diff
@@ -234,14 +234,6 @@
         - ReagentId: Radium
           Quantity: 2
         canReact: false
-  - type: RadiationSource
```
ciaran, line 237: Isn't this a player-facing change?

Killerqu00: no, #35330 already did that, it's just solving the merge conflict

PR #35165, opened 2025-02-14
Flippo Engraved Lighter is now in detective locker

### LINKED ISSUE #35163: Detective's lighter difficult to find for thieves
## Description
Apparently, the lighter spawns in the detective's spare coat, in their closet.
Many thieves do not know this, and I'd bet most detectives don't know this. It's an obscure item, in an obscure location, which makes it very difficult for thieves to even figure out where it is on their own.
An item being hard to steal is fine. An item being so hard to locate that even the role that's meant to have it might not know it exists, though... that sounds like it might be a problem.

**Reproduction**
Look for a flippo engraved lighter.
Realize it's buried in the pockets of an unused coat.
Contemplate why the world is like this.

**Screenshots**
N/A

**Additional context**
As a disclaimer - this isn't a mald issue, I don't play thief. I've just seen a lot of thieves baffled by how hard it is to find the lighter.
I think a viable solution might be to move the lighter to spawning loose in the closet, or just lying in the detective's office. This also means detectives might actually notice and use it, instead of a regular lighter.

Zokkie:
## About the PR
This pull request was made to resolve #35163 

Summary of the issue is that the detective coat in the locker is actually a unique coat that contains the flippo engraved lighter (steal objective). The coat is like a regular detective coat with nothing special about it so people never look inside of it.

This PR will make it so that the detective's flippo engraved lighter is in the locker right away instead of a coat. I also made sure the coat that is in the locker does not contain the special lighter anymore.

## Why / Balance
I saw the issue and kind of agreed with the original poster. As a detective player myself I also did not know this thing even existed! This PR will help make the item visible instead of being in some obscure coat pocket.

## Technical details
The item with id FlippoEngravedLighter is now added to the LockerDetectiveFilled.
The item with id ClothingOuterCoatDetective is now replaced with ClothingOuterCoatDetectiveLoadout in the LockerDetectiveFilled. Which is basically the same coat but without the flippo engraved lighter.

## Media
https://github.com/user-attachments/assets/806de646-a3cc-49fc-8e7a-64714f2179cf

## Requirements

## Breaking changes
Tested it. No breaking changes.

**Changelog**
This is such a small change. I don't think it requires a changelog addition.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/security.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/security.yml
@@ -125,7 +125,8 @@
       - id: ClothingHeadHatFedoraBrown
       - id: ClothingNeckTieDet
       - id: ClothingOuterVestDetective
-      - id: ClothingOuterCoatDetective
+      - id: ClothingOuterCoatDetectiveLoadout
+      - id: FlippoEngravedLighter
       - id: FlashlightSeclite
       - id: ForensicScanner
       - id: LogProbeCartridge


ArtisticRoomba (approved):
Looks good to me.

PR #35169, opened 2025-02-14
Removed Raffle from space carp.

ToastEnjoyer:
## About the PR
Removed the 10 second raffle timer from space carp.

## Why / Balance
It is a team antagonist with so many other carps with them, that this change will let dead players play as a carp much faster.

## Technical details
Removed the raffle: component

## Media

## Breaking changes
If anything breaks, add the raffle component and timer: short

**Changelog**


beck-thompson (changes requested):
This doesn't work. Unsure why but I'm pretty sure the default is short or it inherits from something that is short!

image

PR #35172, opened 2025-02-14
New Feature: Slot blockers

### LINKED ISSUE #12098: Can equip masks even with hardsuit helmets on
I could've sworn in ss13 you had to remove your hardsuit helmet to put a mask on / off.

Winkarst-cpu:
## About the PR
Added feature that prohibits from unequipping and equipping items in certain slots, because of the other equipped items.

## Why / Balance
Resolves (partially) #12098. Gameplay-wise will force people to find a place to safely change their glasses, jumpsuit, etc and could potentially make stealing harder.

## Media

https://github.com/user-attachments/assets/6add5f86-7f81-4c62-bd25-53c2febc954b

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Now certain clothes (like hardsuits and their helmets) can prevent from changing clothes, goggles, etc.


qrwas:
maybe it would be cool to hide closed clothing?

lzk228:
i agree with clothing hiding, but only in strip ui, not for yourself

Winkarst-cpu:
> maybe it would be cool to hide closed clothing?

Should be done in a separate PR since this would require changing ``SharedStrippableSystem.cs``.

Errant-4:
> maybe it would be cool to hide closed clothing?

Shouldn't do so _universally_. Consider for example the tank harness, which is literally just some straps, or even the high-vis vest. Those should definitely not block seeing the jumpsuit underneath

ArtisticRoomba:
Yeah, but in the future we should definitely make a PR to hide technically non visible clothing, like eyes and face behind a hardsuit helmet. Got plenty of secoffs that open your strip menu to see that syndie mask you're trying to hide from them.

Content.Shared/Inventory/SlotBlockComponent.cs
```diff
@@ -0,0 +1,16 @@
...
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(SlotBlockSystem))]
+public sealed partial class SlotBlockComponent : Component
+{
+    /// <summary>
+    /// Slots that this entity should block.
+    /// </summary>
+    [DataField(required: true), AutoNetworkedField]
+    public string[] Slots = [];
+}
```
metalgearsloth, line 16: ... Slotflag? Much smaller.

EmoGarbage404 (approved):

beck-thompson (changes requested):
Looks pretty good but two things:
1.) I think it would be better to make `IsEquippingAttemptEvent` and `IsUnequippingAttemptEvent` inventory relay events instead! Then you can just listen for the event and cancel it. You wouldn't need the `GetBlocker` function and stuff it would be a lot cleaner.
2.) I think it would also make sense to block headsets when you have like an EVA suit equipped - did you exclude it for balance reasons? If so, that's fair and feel free to leave it how it is, but if you just forgot I think it makes sense to block!

Let me know if you need any examples or help!

Winkarst-cpu:
PR is cooked 💀

beck-thompson (approved):
A little bit of web eddit ops but looks good! I couldn't push to your repo for some reason

EmoGarbage404 (approved):

PR #35175, opened 2025-02-15
Automatically split stacks when inserting into a storage that wouldn't otherwise fit the whole stack

Alkheemist:
## About the PR

Upstreamed code from this PR https://github.com/new-frontiers-14/frontier-station-14/pull/2891
Modifies _TryInsertMaterialEntity_ to calculate the maximum amount that will fit in a materialStorage, and then Use that many sheets from the stack in hand.
 
## Why / Balance

So that attempting to put more materials that can fit inside a storage doesn't fail silently and so the player doesn't have to manually split the stacks to make them fit.

## Technical details

To SharedMaterialStorageSystem.cs
- added sharedStackSystem as a dependency
- modified _TryInsertMaterialEntity_ with stack splitting functionality
- refactored _TryInsertMaterialEntity_ in general.

To Client MaterialStorageSystem
- removed override, predicted deletion is now handled by `_sharedStackSystem.TryUse`

To Server MaterialStorageSystem
- added functionality to report amount inserted in the popup

## Media

https://github.com/user-attachments/assets/12f5ce0f-39d9-4999-8c5e-0842dafecb96

## Requirements

## Breaking changes

- _TryInsertMaterialEntity_ args are now
```
TryInsertMaterialEntity(EntityUid user,
        Entity<MaterialComponent?, PhysicalCompositionComponent?> toInsert,
        Entity<MaterialStorageComponent?> receiver)
```

All optional args (`storage`, `material` and `component`) have been removed.

**Changelog**

:cl:
- tweak: Inserting fuel into a generator by hand will automatically insert the maximum amount that will fit.


ArtisticRoomba:
Really good stuff, another checked box in the "non-communicated blocked player interactions" issues.

Nitpick, could you raise text that says how much you put into the machine? For example, stuff like:
- You insert (amount) (item) into the (target generator).
- (Target generator) is full!

Alkheemist:
https://github.com/user-attachments/assets/125f08b0-f504-4037-88fa-a52857c61847

Good call. Not sure if I should modify the base TryInsertMaterialEntity to support reporting the amount inserted as well.

Content.Shared/Materials/SharedMaterialStorageSystem.cs
```diff
@@ -291,6 +296,83 @@ public virtual bool TryInsertMaterialEntity(EntityUid user,
...
+        if (storage.StorageLimit is null || !HasComp<StackComponent>(toInsert))
+            return false;
+
+        int multiplier;
+        var availableVolume = (int)storage.StorageLimit - GetTotalMaterialAmount(receiver, storage);
+        var volumePerSheet = 0;
+        foreach (var (_, vol) in composition.MaterialComposition)
+        {
+            volumePerSheet += vol;
+        }
```
metalgearsloth, line 331: .Values is better than iterating both.

Content.Server/Materials/MaterialStorageSystem.cs
```diff
@@ -114,6 +114,35 @@ public override bool TryInsertMaterialEntity(EntityUid user,
...
+        _audio.PlayPvs(storage.InsertingSound, receiver);
+        _popup.PopupEntity(Loc.GetString("machine-insert-item-amount", ("user", user), ("machine", receiver),
+            ("item", toInsert), ("amount", count)), receiver);
+
+        // Logging
+        _adminLogger.Add(LogType.Action, LogImpact.Low,
+            $"{ToPrettyString(user):player} inserted {count} {ToPrettyString(toInsert):inserted} into {ToPrettyString(receiver):receiver}");
+        return true;
+    }
+
```
metalgearsloth, line 145: 1. Is this even predicted properly? All of this is in server and using non-predicted code.
2. Why is power even coupled into this?

Content.Shared/Materials/SharedMaterialStorageSystem.cs
```diff
@@ -291,6 +296,83 @@ public virtual bool TryInsertMaterialEntity(EntityUid user,
...
+
+        if (_whitelistSystem.IsWhitelistFail(storage.Whitelist, toInsert))
+            return false;
+
+        if (HasComp<UnremoveableComponent>(toInsert))
+            return false;
+
+
+        if (storage.StorageLimit is null || !HasComp<StackComponent>(toInsert))
+            return false;
```
metalgearsloth, line 323: Why is stackcomponent even being checked if the component isn't being used?

metalgearsloth (changes requested):
I realised you just copy-pasted everything from lines 241 to 292. Please do not do this as it makes the codebase unmaintainable. You must make your code re-useable.

Alkheemist:
> I realised you just copy-pasted everything from lines 241 to 292. Please do not do this as it makes the codebase unmaintainable. You must make your code re-useable.

Apologies for that, I've been trying to minimally touch upstream functions in my code but I realise that I don't need to do that when I'm on the upstream codebase.

I've refactored the functionality into an optional arg to _TryInsertMaterialEntity_, setting trySplitStacks to true will make it attempt to split the stack if the whole stack doesn't otherwise fit. Currently only set to true in _OnInteractUsing_

slarticodefast:
Are you coming back to this?

Content.Shared/Materials/SharedMaterialStorageSystem.cs
```diff
@@ -350,23 +354,59 @@ public virtual bool TryInsertMaterialEntity(EntityUid user,
 
         // Material Whitelist checked implicitly by CanChangeMaterialAmount();
 
+        // test if the whole stack fits
         var multiplier = TryComp<StackComponent>(toInsert, out var stackComponent) ? stackComponent.Count : 1;
```
iaada:
```suggestion
        var multiplier = _sharedStackSystem.GetCount(toInsert);
```

Content.Server/Materials/MaterialStorageSystem.cs
```diff
@@ -93,28 +93,41 @@ public override bool TryInsertMaterialEntity(EntityUid user,
         EntityUid receiver,
         MaterialStorageComponent? storage = null,
         MaterialComponent? material = null,
-        PhysicalCompositionComponent? composition = null)
+        PhysicalCompositionComponent? composition = null,
+        bool trySplitStacks = false)
     {
         if (!Resolve(receiver, ref storage) || !Resolve(toInsert, ref material, ref composition, false))
             return false;
         if (TryComp<ApcPowerReceiverComponent>(receiver, out var power) && !power.Powered)
```
iaada: This is the only thing keeping this method from being moved to shared, which can be solved with an API call.
```suggestion
        if (!SharedPowerReceiverSystem.IsPowered(receiver))
```

Content.Shared/Materials/SharedMaterialStorageSystem.cs
```diff
@@ -334,7 +337,8 @@ public virtual bool TryInsertMaterialEntity(EntityUid user,
         EntityUid receiver,
         MaterialStorageComponent? storage = null,
         MaterialComponent? material = null,
-        PhysicalCompositionComponent? composition = null)
+        PhysicalCompositionComponent? composition = null,
+        bool trySplitStacks = false)
```
iaada: This really spooks me with optional components for different entities. This should be included in breaking changes.
```
    public bool TryInsertMaterialEntity(EntityUid user,
        EntityUid<MaterialStorageComponent?> toInsert,
        EntityUid<MaterialComponent?, PhysicalCompositionComponent?> receiver)
```

iaada (changes requested):
This can be predicted which removes the uncomfortable code of trying to extract information from the base method.

As for the shared method I feel it needs a different strategy. There's a lot of redundant checks being done from overused API functions in the original. The nested ifelse inside a foreach with 3 variables at the head also feels very wrong.

First we should be getting the volume of `toInsert` using `IsMaterialWhitelisted`. Then get the available space and calculate how many sheets to use. Then call `TryChangeMaterialAmount`.

Alkheemist:
Well, the requested changes have been made to move most of the functionality to shared, but for some reason it will flash the UI briefly before the insertion happens. Doesn't occur on master so definitely it's something about my changes causing it but I have no idea what.

PR #35180, opened 2025-02-15
Fix colornetwork command not checking for correct permissions

### LINKED ISSUE #32466: colornetwork is available always and for everyone
the code says that this shouldn't be possible, but somehow it is

https://github.com/space-wizards/space-station-14/blob/bf80f3abb8a2ba329636f9e288f70e3bc501b8b0/Content.Server/Sandbox/Commands/ColorNetworkCommand.cs#L22-L25

also players can get all entities list with dumpentities command which also is available for everyone (i'll create issue for this)
and there probably can be pipes in the list

https://github.com/user-attachments/assets/a6a99c65-7c6c-4a49-a0ea-ae81efd0e9e4

Simyon264:
What is shell.IsClient even?? And why is it checked on a server side command?

???

Fixes #32466


SlamBamActionman (approved):
A bit sus on IsClient even being there since the Command seems to only be possible to run on server... But otherwise the change makes sense.

PR #35182, opened 2025-02-15
New Feature: Information about fields copying in PDA

### LINKED ISSUE #28588: PDA menu should has some info about copying on clicking info buttons
Copying data from pda was introduced in #23533 but there is no way to find out that all info lines actually are buttons that can copy text in your clipboard, so there should be some feedback after pressing or info
what i suggest:
- Popup "Copied" after pressing any button
- Add somewhere "?" button with text on hover or just text with "Click to copy" (this one would be better because why someone would press on info to get clue about copying)

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Resolves #28588.

## Media

https://github.com/user-attachments/assets/9bad99b8-0b07-464c-8121-8c8023567c1c

## Requirements

## Breaking changes

**Changelog**


Content.Client/PDA/PdaMenu.xaml.cs
```diff
@@ -97,35 +101,41 @@ public PdaMenu()
             PdaOwnerButton.OnPressed += _ =>
             {
                 _clipboard.SetText(_pdaOwner);
+                _popupSystem.PopupCursor(Loc.GetString("comp-pda-ui-copied"));
```
beck-thompson, line 104: Maybe just make a helper function so if someone wants to change the popup or add something else they won't have to change it in so many places? Besides that I don't see how to make it better :laughing:

beck-thompson (changes requested):

beck-thompson (approved):
wow that's better than I was thinking good job :+1:

Centronias:
IMO "field" is a term the average SS14 player wouldn't understand (similar rationale to `Avoid Technical Jargon` [here](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html#changelog)).
Rather than adding more text to the PDA UI, would something like making the copiable fields underlined (so as to look like clickable hyperlinks, which everyone is familiar with) make the feature more discoverable and therefore not need an explanation?

PR #35185, opened 2025-02-15
Refactor: Arcades

### LINKED ISSUE #2532: Block Game™️ code improvements
As discussed with @PaulRitter 

- Messages like `BlockGameVisualUpdateMessage` that have an internal enum to determine the purpose of their data, done to avoid having 1 C# type per variant
    - This is a rather baseless fear, there's nothing bad about having lots of classes for messages (just stick them in a `.messages` sub namespace if you're worried about pollution)
    - The message size increases to be able to hold the enum, thankfully with byte enums this shouldn't be too bad (The enums in Block Game™️ aren't byte though!), but it's still unnecessary size
    - The "variant"/"purpose" of the data can be determined through the type of the message if each message were its own concrete C# type -- this is effectively free as the type of the message is already sent over the network

- State changes everywhere
    - The current piece can be moved in multiple "ticking" methods such as ProcessInput 
    - Makes it hard to determine what the state of the game is
    - Bad for prediction

- There's no 2d grid
    - Block Game™️ is a 2d rectangle grid game, the simplest storage form for that would be a 2d array
        - the 2d array can be easily stored as a 1d array
             - Positions of blocks no longer need to be stored, if a block's x and y need to be looked up from an index, it's as simple as:
                ```cs
                    var x = i % widthOfGrid;
                    var y = i / widthOfGrid;
                ```
           - Blocks can be looked up via their x/y coords still
                ```cs
                    var i = widthOfGrid * y + x;
                ```
    - 2d grid would allow for easy removal of the `BlockGameBlock` struct, simply have the grid be an array of `BlockGameBlock` enum, eg:
    ```cs
    public enum BlockGameBlock : Byte
    {
        Red,
        Blue,
        ...,
        Empty
    }
    ```
    - Collisions against the grid can simply be done by looking up if an x y index isn't `BlockGameBlock.Empty`

### LINKED ISSUE #24811: Arcade code sucks
1. Not predicted.
2. It's a mess.
3. It's just increasing tech debt if people add features to it.

At least nothing depends on it but it barely works.

Winkarst-cpu:
## Why / Balance
Resolves #24811, Resolves #2532.

### New UI for Space Villian Arcade and WireActions fix
https://github.com/user-attachments/assets/2a2012ce-2751-4b8e-bedd-0c2881a8a709

## Requirements

## Breaking changes

**Changelog**


Winkarst-cpu:
Nuh, I better make a new PR. This is in conflict hell by now.

PR #35193, opened 2025-02-15
Shove down a person on uncuff if harm mode is on

### LINKED ISSUE #29007: uncuffing someone should keep them in place for a few seconds
## Description
to make the lives of sec easier when dealing with shitters holding down movement keys so as soon as you uncuff theyve escaped

from @SlamBamActionman 
> Want to improve Security's handling of prisoners? Make people cuffed to chairs/beds remain stuck for an additional 2 seconds after being uncuffed so that we can exit the damn brig cell without the detained immediately rushing out through us. Now that is annoying.

Killerqu00:
## About the PR
Uncuffing someone with combat mode turned on will shove them down. Ingame tip included.
Note: popups being fucked in media is not due to this PR, see #35247 

## Why / Balance
Resolves #29007 

## Media
https://github.com/user-attachments/assets/d338c180-0f5e-463c-8a5a-cba1a984b82a

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Uncuffing someone with combat mode on will shove them down.


ArtisticRoomba:
What are the balance implications for people stealthily uncuffing themselves in a cell on the station or evac? Seems like they could simply be put down instantly and re-cuffed, even though security was supposed to be punished for letting a cuffed person wiggle away.

Note that security can use flashes to run out of a cell, and use zipties to restrain prisoners for long enough to exit perma before they escape without giving the prisoner cuffs.

A wall flasher might also be an alternative solution.

keronshb:
This is highly unbalanced.

If prisoners getting up is an issue, introduce wall flashers inside the cell. Or increase the unbuckle > uncuff time.
Or they can just baton the prisoner until they're stunned, take off the cuffs, and get out.

SlimmSlamm:
It's not that hard to flash someone right after you uncuff them.

SlamBamActionman:
I think this PR's heart is in the right place; shitters love to fuck with you the moment you uncuff them, and while there are strategies against it they aren't reliable, unintuitive for new players and most of the time you end back up in the status quo of the prisoner being back in handcuffs so you have to do the whole song and dance again.

For those of you who don't play Sec regularly: *uncuffing prisoners in a cell is, without hyperbole, the worst-feeling part of the role and feels like it's full of jank*.

We used to have people do uncuffing through windoors to be safe which was deemed an acceptable exploit by admins until it got patched. The current safest method I know of is to place down a barrier in front of the door, jump onto it and uncuff from there. But shitters will try to disarm you the moment they get free, likely making you drop the cuffs. Pull out a baton? They get 3 chances to disarm you. Flashes are good until the prisoner wedges themselves into the windoor and prevents it from closing, and then back to cuffs they go. Don't even think about talking inside of a cell to tell them to stop, or you will be dragged to the door using a keycard.

My changed would be the following:
1) Make it only apply on *being* uncuffed from a buckled state, not self-uncuffs.
2) Instead of 99% stamcrit, just stun them fully. That leaves a short window the SecOff can leave the cell and would be close to the same as just stun batoning them beforehand, just with a more forgiving window.

This would still allow stealthy escapes, and give a simple timer visual for when the prisoner can get back up from being buckled.

SlamBamActionman:
And to pre-empt the argument that fighting back against Security is part of the game:

Agreed! But causing trouble *the moment you're uncuffed in a cell* is almost exclusively a shitter/self-antag move. You're in Sec after all, even if you get your hands on a stun baton the odds of escaping are fairly low. The most realistic result is getting back into cuffs, meaning you have lost nothing but Sec has had to waste time wrangling you - and now you get to do it again. And again. And again.

Chases, pulling out a hidden weapon and perma-breakouts are fun for both sides. This isn't, it's just a timewaster.

Errant-4:
Why not just put an interaction cooldown/doafter on interactions for 1-2 seconds after someone was uncuffed? Feels much less arbitrary. Stamina damage here feels rather unintuitive

Killerqu00:
> My changes would be the following:
> 
>     1. Make it only apply on _being_ uncuffed from a buckled state, not self-uncuffs.
> 
>     2. Instead of 99% stamcrit, just stun them fully. That leaves a short window the SecOff can leave the cell and would be close to the same as just stun batoning them beforehand, just with a more forgiving window.

done

LaCumbiaDelCoronavirus:
you can hold a flash in the same hand that youre uncuffing someone with which gives you ample time to flash the prisoner and get out of the cell without them being able to do anything, regardless of your reaction time

not a fan of this for same reasons as above

SlamBamActionman:
Hello! We've had a fairly extensive discussion in the maintainer team, and these are the results:

Although the concept of the idea had support, just stunning someone was raised as feeling like an odd result of the uncuff interaction. We discussed alternatives but what ended up with major approval was something that should increase affordance and fit within existing gameplay verbs. 

> Uncuffing someone else while in Harm Mode shoves them down on the ground upon completion.

This solution would be in line with shoves being part of the Harm Mode, allow for it to be easily toggled with minimal impact on non-problem prisoners, and works regardless of whether the cell has a buckleable object. It ends up being very similar to your solution in practice (in effect you can still stun someone who you uncuff who is buckled, you just gotta be in harm mode). The "main" thing needed would be a different pop-up from the regular uncuff pop-up. `You start unrestraining Urist McFluff, readying to shove [them].` or something.

Killerqu00:
> Hello! We've had a fairly extensive discussion in the maintainer team, and these are the results:
> 
> Although the concept of the idea had support, just stunning someone was raised as feeling like an odd result of the uncuff interaction. We discussed alternatives but what ended up with major approval was something that should increase affordance and fit within existing gameplay verbs.
> 
> > Uncuffing someone else while in Harm Mode shoves them down on the ground upon completion.
> 
> This solution would be in line with shoves being part of the Harm Mode, allow for it to be easily toggled with minimal impact on non-problem prisoners, and works regardless of whether the cell has a buckleable object. It ends up being very similar to your solution in practice (in effect you can still stun someone who you uncuff who is buckled, you just gotta be in harm mode). The "main" thing needed would be a different pop-up from the regular uncuff pop-up. `You start unrestraining Urist McFluff, readying to shove [them].` or something.

sounds feasible

Errant-4:
We probably only want to customize the popup message at the end that is generated after we already know if the uncuff was executed with harm mode on or off, to avoid confusing people who toggle it during the uncuff and get a different result than what was advertised. The popup message then just has to achieve the goal of letting a player who is not familiar with how this works know why their criminal/hostage just fell to the ground

Killerqu00:
> We probably only want to customize the popup message that happens at the end that is generated after we already know if the uncuff was executed with harm mode on or off, to avoid confusing people who toggle it during the uncuff and get a different result than what was advertised in the earlier popup message. The popup message then just has to achieve the goal of letting a player who is not familiar with how this works know why their criminal/hostage just fell to the ground

ideally there should be a way to find that out other than a popup _on actually doing the thing_ + ingame tip. im not sure what, though; guidebook, maybe (but last time I checked, due to the structure of guidebook, random facts wouldn't really fit there)

Errant-4:
> ideally there should be a way to find that out other than a popup _on actually doing the thing_ + ingame tip. im not sure what, though; guidebook, maybe (but last time I checked, due to the structure of guidebook, random facts wouldn't really fit there)

I think the the sec guide would be the logical place for this. Either put a short mention at the end of the Gear section (similar to how the Flash already mentions its multiple modes of operation) or separate nonlethals and the cuffs+bolas into their own sections

Screenshot_2025-03-09_100900

Errant-4 (changes requested):
Put it in the guidebook and we are good to go

Resources/ServerInfo/Guidebook/Security/Security.xml
```diff
@@ -15,7 +15,7 @@ They face [textlink="Syndicate Agents" link="Traitors"], [textlink="Nuclear Oper
   </Box>
 
 ## Gear
-First we have non-lethals a step above simply telling someone to cooperate with instructions. Both the stunbaton and disabler are capable of limiting the movement of an assailant, whereas handcuffs can be applied to deny a criminal free movement and access to their hands.
+First we have non-lethals a step above simply telling someone to cooperate with instructions. Both the stunbaton and disabler are capable of limiting the movement of an assailant, whereas handcuffs can be applied to deny a criminal free movement and access to their hands. Cuffs will also shove the person down if your [color=red]harm mode[/color] is active during uncuff.
```
ArtisticRoomba, line 18: I would maybe bold it just so it gets noticed

Killerqu00: it has red text already in it + there is an ingame tip + changelog

Killerqu00:
i made sure #35743 is still working btw

SlamBamActionman (approved):

Errant-4 (approved):
Thank you for your contribution

PR #35228, opened 2025-02-16
[Bugfix] ContrabandSystem checks jobs correctly

Killerqu00:
## About the PR
Right now ContrabandSystem makes job checks incorrectly: for example, janitors are told that they're not allowed to have galoshes by the examine text. This PR fixes that.

## Why / Balance
bug bad, me fix bug

## Technical details
Right now, it checks if player's job is in the list of allowed jobs *in plural*, which is most likely not true. Instead, I rewrited it to check if player's list is in actual list of allowed jobs (by names). This is still not ideal, but it works and further improvements need refactors in IdCardComponent, so not today.

## Media
image
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Contraband examines now correctly tell you if you can possess the object.

### DIFF
--- a/Content.Shared/Contraband/ContrabandSystem.cs
+++ b/Content.Shared/Contraband/ContrabandSystem.cs
@@ -56,7 +56,8 @@ public sealed class ContrabandSystem : EntitySystem
         // one, the actual informative 'this is restricted'
         // then, the 'you can/shouldn't carry this around' based on the ID the user is wearing
         var localizedDepartments = component.AllowedDepartments.Select(p => Loc.GetString("contraband-department-plural", ("department", Loc.GetString(_proto.Index(p).Name))));
-        var localizedJobs = component.AllowedJobs.Select(p => Loc.GetString("contraband-job-plural", ("job", _proto.Index(p).LocalizedName)));
+        var jobs = component.AllowedJobs.Select(p => _proto.Index(p).LocalizedName).ToArray();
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
@Coolsurf6 might be interested

Content.Shared/Contraband/ContrabandSystem.cs
```diff
@@ -86,7 +86,7 @@ private void OnDetailedExamine(EntityUid ent,ContrabandComponent component, ref
         String carryingMessage;
         // either its fully restricted, you have no departments, or your departments dont intersect with the restricted departments
         if (departments.Intersect(component.AllowedDepartments).Any()
-            || localizedJobs.Contains(jobId))
+            || component.AllowedJobs.Select(p => _proto.Index(p).LocalizedName).Contains(jobId))
```
slarticodefast, line 89: Looks like this broke when https://github.com/space-wizards/space-station-14/pull/34559/files pluralized the `localizedJobs` variable.
Could you adjust this so we don't have to index all the prototypes twice (see line 59)? Just save them in another variable.

slarticodefast (changes requested):

Content.Shared/Contraband/ContrabandSystem.cs
Killerqu00: true

slarticodefast (approved):
Works as intended, thanks for the bugfix!

PR #35233, opened 2025-02-17
Fix MappingCommand grid loading

ElectroJr:
## About the PR
Since engine v247.0.0, the `mapping` command fails to load grid files with an error explaining that it was expecting a map file. This PR fixes that, though it will still generally result in server-side errors being logged unless a new optional argument is used to explicitly specify that the given file is a grid instead of a map.

This PR also changes the mapping auto-save logic so it auto-saves grids as grids, instead of saving them as maps.


IProduceWidgets:
I don't have any complaints, but I do wonder if it makes sense to have this *and* loadgrid?

Issue #35243, opened 2025-02-17
need centcomm/bluespace announcement console


Emisse:
kinda like syndie/wizard comms console where it can broadcast everywhere

VerinSenpai:
Content PR *does some weird shit with his hands* MERGED!!

PR #35248, opened 2025-02-17
Add CentComm Comms Console + Change admin announcement color

ScarKy0:
## About the PR

Added a centcomm comms console for #35243

## Why / Balance

Requested for new centcomm.

## Technical details

Just a new comm console with updated sprites and colors. Regular events etc still use the old yellow.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- tweak: Aghost's intrinsic comms console now uses a new color, announces globally and has significantly lowered delays. Announce command and admin menu option remain unchanged.


lzk228:
feels like should be a little brighter

ScarKy0:
> feels like should be a little brighter

This is the color that centcomm clothes use, but im all down if someone proposes a better one because i honestly have no idea what to use

YoungThugSS14:
> > feels like should be a little brighter
> 
> This is the color that centcomm clothes use, but im all down if someone proposes a better one because i honestly have no idea what to use

Maybe Centcomm radio blue?

ScarKy0:
> Maybe Centcomm radio blue?

The main color of centcomm is green so id wonna stick to that, i think the color i managed to get right now is okay

YoungThugSS14:
> > Maybe Centcomm radio blue?
> 
> The main color of centcomm is green so id wonna stick to that, i think the color i managed to get right now is okay

Yes, I understand that, but I feel like it’s a little similar to the normal radio color so people won’t pay attention to it.

ScarKy0:
Here it is when blue, is it better?

image

YoungThugSS14:
> Here it is when blue, is it better?
> 
> image

Yeah In my opinion that looks much more recognisable.

walksanatora:
it would also be nice if it had the ability to call gamma/delta alerts. just so a random player at central could send announcements and call those alerts. (even if the lack the ability to spawn the deathsquad)

ScarKy0:
> it would also be nice if it had the ability to call gamma/delta alerts. just so a random player at central could send announcements and call those alerts. (even if the lack the ability to spawn the deathsquad)

No chance, comms consoles cannot set alert levels of other stations, so it would only change the alert of centcomm
also, it feels like a bit much

walksanatora:
> > it would also be nice if it had the ability to call gamma/delta alerts. just so a random player at central could send announcements and call those alerts. (even if the lack the ability to spawn the deathsquad)
> 
> No chance, comms consoles cannot set alert levels of other stations, so it would only change the alert of centcomm also, it feels like a bit much

i mean starlight has this as the "sectoral console"
and it is able to call gamma and delta alerts. this is because they are working to make central playable as non-admins (it will require more playtime then captain. since it would be presumed that if you are in this role you know spacelaw and their SoPs)

VasilisThePikachu (approved):

PR #35255, opened 2025-02-17
Add ability for +VVEDIT users to scale damage/healing in the game

SlamBamActionman:
## About the PR

This PR adds a whole bunch of CVars that allow admins (or rather, those with VVedit perms) to scale the damage dealt/healing done in the game. 

Changing these is done via the recently introduced `changecvar` command and takes a float as input. The following categories can be scaled:
- *All* damage dealt
- *All* healing done
- Melee damage
- Projectile damage
- Hitscan damage
- Reagent damage
- Reagent healing
- Explosion damage
- Thrown damage
- Topicals healing

## Why / Balance

This would allow testing for different kill speeds, how the game feels when healing is more/less effective and explosion scaling, to name a few. 

There's been a recent concern that it can be far too easy for a single player to dominate a round, and that the time to kill unexpecting players is too fast. Playtesting a higher Time To kill would be interesting to see how it plays out.

## Technical details

`DamageableSystem` is subbed to if any of the cvars change, and other systems when dealing damage can multiply using those public variables. 

To make sure guidebooks/examines are consistent with these modifiers, they also grab these values. However due to how guidebook chemistry entries don't update dynamically they have to be manually updated - so we add a new function that reloads them all. 

## Media

https://github.com/user-attachments/assets/3b8ad0e0-b0be-484f-8683-977a4a2cbe74

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- add: Users with +VVEDIT perms can now use `changecvar playtest.[modifier] [float]` to modify damage/healing scaling live.


SlamBamActionman:
> What are some interesting cvar numbers to try?
> 
> Also at some point we should look at the various health systems in ss13, including implementing soft and hard crit states

I, for one, am looking forward to seeing how 0.5x multiplier to all damage plays out.

Simyon264 (approved):

IProduceWidgets:
Very cool, been wanting to mess with projectile damage for a while. This is so much easier than uploading all new projectile prototypes.

IProduceWidgets:
Bad news, unless Im just skill issuing real hard I dont think admins can set these, because even Game Masters don't have perms to set server cvars.

PR #35262, opened 2025-02-17
allow to publish news without ui

lzk228:
## About the PR

split OnWriteUiDeleteMessage in two, allowing to use a new AddNews method for publishing news from other systems, without need to go through client

## Why / Balance

do it for a fork, but i suppose it could have some usage in upstream in the future or may be helpful to other forks, as Slam said

image

## Technical details

described in about pr

## Media

n/a

## Requirements

## Breaking changes

OnWriteUiDeleteMessage method has been split in two.
Now it's OnWriteUiDeleteMessage, which raised on client, and AddNews, which can be accessed from other systems.

**Changelog**

no playerfacing cl no fun

### DIFF
--- a/Content.Server/MassMedia/Systems/NewsSystem.cs
+++ b/Content.Server/MassMedia/Systems/NewsSystem.cs
@@ -99,7 +99,7 @@ public sealed class NewsSystem : SharedNewsSystem
             _adminLogger.Add(
                 LogType.Chat, LogImpact.Medium,
                 $"{ToPrettyString(msg.Actor):actor} deleted news article {article.Title} by {article.Author}: {article.Content}"
-                );
+            );
 
             articles.RemoveAt(msg.ArticleNum);
             _audio.PlayPvs(ent.Comp.ConfirmSound, ent);
@@ -130,9 +130,6 @@ public sealed class NewsSystem : SharedNewsSystem
         if (!ent.Comp.PublishEnabled)
             return;
 
-        if (!TryGetArticles(ent, out var articles))
-            return;
-
         if (!CanUse(msg.Actor, ent.Owner))
             return;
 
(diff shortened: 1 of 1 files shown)


Content.Server/MassMedia/Systems/NewsSystem.cs
```diff
@@ -146,32 +143,46 @@ private void OnWriteUiPublishMessage(Entity<NewsWriterComponent> ent, ref NewsWr
...
-            ("author", article.Author ?? Loc.GetString("news-read-ui-no-author"))
+            _chatManager.SendAdminAnnouncement(Loc.GetString("news-publish-admin-announcement",
+                                                             ("actor", msg.Actor),
+                                                             ("title", article.Value.Title),
+                                                             ("author", article.Value.Author ?? Loc.GetString("news-read-ui-no-author"))
             ));
+        }
+    }
+
+    public void AddNews(EntityUid uid, string title, string content, string? author, out NewsArticle? article)
```
slarticodefast, line 166: Why use `out` and not just make it return value?
Unless you want to turn it into a `public bool TryAddNews` function that returns if it was successful

slarticodefast (changes requested):

Content.Server/MassMedia/Systems/NewsSystem.cs
```diff
@@ -146,38 +143,53 @@ private void OnWriteUiPublishMessage(Entity<NewsWriterComponent> ent, ref NewsWr
...
+            _chatManager.SendAdminAnnouncement(Loc.GetString("news-publish-admin-announcement",
+                                                             ("actor", msg.Actor),
+                                                             ("title", article.Value.Title),
+                                                             ("author", article.Value.Author ?? Loc.GetString("news-read-ui-no-author"))
             ));
+        }
+    }
 
-        articles.Add(article);
+    public bool TryAddNews(EntityUid uid, string title, string content, string? author, [NotNullWhen(true)] out NewsArticle? article)
```
slarticodefast, line 164: add a documentation summary and a `<param name="uid"> description </param>` line for each of the parameters since they can be a little confusing

Content.Server/MassMedia/Systems/NewsSystem.cs
```diff
@@ -146,38 +143,53 @@ private void OnWriteUiPublishMessage(Entity<NewsWriterComponent> ent, ref NewsWr
...
+    {
+        if (!TryGetArticles(uid, out var articles))
+        {
+            article = default;
```
slarticodefast, line 168: should return `null` if it fails

slarticodefast (changes requested):

PJB3005 (approved):

Issue #34980, opened 2025-02-08
China lake is difficult to counter as crew vs loneop.


Pumkin69:
The china lake in its current form does not feel like it follows the same standard of gameplay as most if not all other syndicate weaponry. Its combat flow and counterplay is not intuitive and i will give my reasons for that now.

Currently the china lake bundle is a 25TC purchase that comes with a china lake, 4 blast grenades and 7 frag grenades so 11 total grenades blasts being a 6 tile wide area with incredibly high damage and frags being an 11 tile wide area with high damage. 

The current counter play to the china lake boils down to hoping the operative makes a mistake instead of being able to take the initiative to gain an advantage, this simply does not feel very good at all gameplay wise.

If a loneop decides they want to use the china lake and a jugsuit they have multiple ways of completely removing security from trying to stop them. A loneop uplink has 60TC in it and loneops start with a viper (important). Currently as a loneop you could purchase 
1 energy sword
3 magazines of .35 auto for your viper
1 access breaker
1 agent id card
1 pair of no-slip shoes
1 china lake bundle 
1 EMP implant (only 2TC purchase which removes a lot of options for security)
1 cybersun juggernaut suit (can tank 3 frag rounds from the China lake literally point blank without going crit)

This loadout currently solves every possible problem the loneop will face if they even slightly use their china lake ammo sparingly.
The china lake will straight up instantly crit any security officer daring enough to attempt to get into the operatives line of sight.
The emp implant completely removes any chance of stun batons or stun prods 3 TIMES.
The juggernaut suit makes most damage types do nothing more than tickle you.
The access breaker lets you walk through doors
The energy sword lets you break open windows or other objects in tight spots as well as killing anybody in melee.
The viper with 4 total magazines is easily enough to take down most other problems such as stray crew members with firearms and whatnot.

There is VERY LITTLE to no counterplay using this loadout. The major problems in this loadout that feel disproportionately strong are the china lake and the EMP implant. 

The China lake has 11 rounds of grenades that take 1 shot to crit MOST people on the station and holds 3 rounds in it with NO slowdown and a very good rate of fire, considering the fact that its a grenade launcher.

The EMP implant is a 3 use 2TC instant EMP around you with a range of 2.5 tiles. The power that this 2TC purchase holds is genuinely incredible and is basically a must buy for all operatives no matter what situation you are in.
It completely removes stun batons as an option for security and completely drains batteries of everything else they have.
[...]

PR #35264, opened 2025-02-17
China Lark Nerf Issue: #34980

YoungThugSS14:
## About the PR

Fixes Issue #34980 I made the Chinalake only have the ability to contain 1 round, along with making it a requirement to weld the weapon.
## Why / Balance

Alot of people in the comments of the original issue, thought it would be a good idea to reduce it to 1 round, I also decided to make the wield change because I feel like somebody shouldn't be able to use a grenade launcher with one hand.
## Technical details

Capacity of Chinalake reduced to 1, along with the gun having the GunRequiresWield component.
## Media

N/A
## Requirements

## Breaking changes

N/A

**Changelog**

🆑 
- tweak: The Chinalake can only hold 1 explosive at a time, along with it needing to be wielded to use it.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Launchers/launchers.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Launchers/launchers.yml
@@ -37,6 +37,7 @@
     - Back
     - suitStorage
   - type: AmmoCounter
+  - type: GunRequiresWield
   - type: Gun
     fireRate: 1
     selectedMode: SemiAuto
@@ -48,7 +49,7 @@
     whitelist:
       tags:
         - Grenade
-    capacity: 3
+    capacity: 1
     proto: GrenadeFrag
     soundInsert:
       path: /Audio/Weapons/Guns/MagIn/batrifle_magin.ogg


keronshb:
I don't think this is going to be the solution for this.
There's a bigger balance pass that needs to be made and just changing this value isn't going to really fix anything other than be an omega nerf.

Item Balance isn't frozen but it is restricted - https://github.com/space-wizards/space-station-14/issues/8524

YoungThugSS14:
> I don't think this is going to be the solution for this. There's a bigger balance pass that needs to be made and just changing this value isn't going to really fix anything other than be an omega nerf.
> 
> Item Balance isn't frozen but it is restricted - #8524

This likely won’t be the full solution, and obviously people have better solutions than me, but I think the chinalake needs a temporary change to give more options for lone ops and nukies, to encourage creativity.

keronshb:
> > I don't think this is going to be the solution for this. There's a bigger balance pass that needs to be made and just changing this value isn't going to really fix anything other than be an omega nerf.
> > Item Balance isn't frozen but it is restricted - #8524
> 
> This likely won’t be the full solution, and obviously people have better solutions than me, but I think the chinalake needs a temporary change to give more options for lone ops and nukies, to encourage creativity.

We're currently not interested in making temporary changes and smaller balance passes like this without looking at the bigger picture.

foboscheshir:
just delete it?

LaCumbiaDelCoronavirus:
suggestion:
let chinalake stay as being fireable even when unwielded 
and keep the original ammo capacity
but just increase the delay between each shot

Pumkin69:
the actual balance change would likely need to touch a lot of the loneops kit.

keronshb:
> > > I don't think this is going to be the solution for this. There's a bigger balance pass that needs to be made and just changing this value isn't going to really fix anything other than be an omega nerf.
> > > Item Balance isn't frozen but it is restricted - #8524
> > 
> > 
> > This likely won’t be the full solution, and obviously people have better solutions than me, but I think the chinalake needs a temporary change to give more options for lone ops and nukies, to encourage creativity.
> 
> We're currently not interested in making temporary changes and smaller balance passes like this without looking at the bigger picture.

Based on this and the restriction, we're going to close this PR. 
We're currently actively looking at Traitor/Operative balance.

Issue #35210, opened 2025-02-16
The wizard recall spell should show a popup when teleporting an item.


slarticodefast:
Something like
"The {$item} vanishes into thin air!"

Without this it might look like a bug that randomly deleted the item.

lzk228:
related https://github.com/space-wizards/space-station-14/pull/34983

keronshb:
as of right now it cant

slarticodefast:
> as of right now it cant

Any reason why it fails?
I assume it has something to do with PVS?

keronshb:
> > as of right now it cant
> 
> Any reason why it fails? I assume it has something to do with PVS?

youd need to ask scar.

ScarKy0:
The popup appears at the destination of the teleport instead of at the origin, so it ends up just showing above your head. This could probably be fixed with PopupCoordinates but I havent tried that one yet.

PR #35272, opened 2025-02-18
Wizard Recall shows Disappearing text at Coordinate

CooperWallace:
## About the PR

Added a Popup prediction based upon a certain coordinate
Display disappearing text at the last location of the item being recalled
Addresses https://github.com/space-wizards/space-station-14/issues/35210

## Why / Balance

Text can now be produced at a certain location with an event triggered by an action. 
In this specific PR, the recall of an item has a message displayed where it was last located.

## Technical details

PopupPredictedCoordinates merges PopupPredicted and PopupCoordinates

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: A popup is now displayed at the item's previous location when it is recalled by the recall wizard spell.

### DIFF
--- a/Content.Client/Popups/PopupSystem.cs
+++ b/Content.Client/Popups/PopupSystem.cs
@@ -125,7 +125,7 @@ namespace Content.Client.Popups
 
         public override void PopupPredictedCoordinates(string? message, EntityCoordinates coordinates, EntityUid? recipient, PopupType type = PopupType.Small)
         {
-            if (recipient != null && _timing.IsFirstTimePredicted)
+            if (_playerManager.LocalSession == recipient & _timing.IsFirstTimePredicted)
                 PopupCoordinates(message, coordinates, type);
         }
 
--- a/Content.Client/Popups/PopupSystem.cs
+++ b/Content.Client/Popups/PopupSystem.cs
@@ -123,6 +123,12 @@ namespace Content.Client.Popups
                 PopupMessage(message, type, coordinates, null, true);
         }
 
+        public override void PopupPredictedCoordinates(string? message, EntityCoordinates coordinates, EntityUid? recipient, PopupType type = PopupType.Small)
+        {
+            if (recipient != null && _timing.IsFirstTimePredicted)
+                PopupCoordinates(message, coordinates, type);
+        }
+
         private void PopupCursorInternal(string? message, PopupType type, bool recordReplay)
         {
             if (message == null)
--- a/Content.Client/Popups/PopupSystem.cs
+++ b/Content.Client/Popups/PopupSystem.cs
@@ -125,8 +125,8 @@ namespace Content.Client.Popups
 
         public override void PopupPredictedCoordinates(string? message, EntityCoordinates coordinates, EntityUid? recipient, PopupType type = PopupType.Small)
         {
-            if (_playerManager.LocalEntity == recipient && _timing.IsFirstTimePredicted)
-                PopupCoordinates(message, coordinates, type);
+            if (recipient != null && _timing.IsFirstTimePredicted)
+                PopupCoordinates(message, coordinates, recipient.Value, type);
         }
 
         private void PopupCursorInternal(string? message, PopupType type, bool recordReplay)
--- a/Content.Client/Popups/PopupSystem.cs
+++ b/Content.Client/Popups/PopupSystem.cs
@@ -125,7 +125,7 @@ namespace Content.Client.Popups
 
         public override void PopupPredictedCoordinates(string? message, EntityCoordinates coordinates, EntityUid? recipient, PopupType type = PopupType.Small)
         {
-            if (_playerManager.LocalSession == recipient & _timing.IsFirstTimePredicted)
+            if (_playerManager.LocalEntity == recipient && _timing.IsFirstTimePredicted)
                 PopupCoordinates(message, coordinates, type);
         }
 
--- a/Content.Server/Popups/PopupSystem.cs
+++ b/Content.Server/Popups/PopupSystem.cs
@@ -75,18 +75,13 @@ namespace Content.Server.Popups
                 return;
 
             var mapPos = _transform.ToMapCoordinates(coordinates);
+            var filter = Filter.Empty().AddPlayersByPvs(mapPos, entManager: EntityManager, playerMan: _player, cfgMan: _cfg);
             if (recipient != null)
             {
                 // Don't send to recipient, since they predicted it locally
-                var filter = Filter.PvsExcept(recipient.Value, entityManager: EntityManager);
-                RaiseNetworkEvent(new PopupCoordinatesEvent(message, type, GetNetCoordinates(coordinates)), filter);
+                filter = filter.RemovePlayerByAttachedEntity(recipient.Value);
             }
-            else
-            {
-                // No prediction, send to everyone including recipient.
-                var filter = Filter.Empty().AddPlayersByPvs(mapPos, entManager: EntityManager, playerMan: _player, cfgMan: _cfg);
-                RaiseNetworkEvent(new PopupCoordinatesEvent(message, type, GetNetCoordinates(coordinates)), filter);
-            }
+            RaiseNetworkEvent(new PopupCoordinatesEvent(message, type, GetNetCoordinates(coordinates)), filter);
         }
 
         public override void PopupEntity(string? message, EntityUid uid, PopupType type = PopupType.Small)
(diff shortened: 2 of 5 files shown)


CooperWallace:
There should be text from onlookers to describe whose hand the book showed up in.

I can see this being super handy in a variety of situations. Showing when an artifact disappears would be really useful (That used to confuse me). Just wasn't too sure how to test

Content.Shared/ItemRecall/SharedItemRecallSystem.cs
```diff
@@ -85,6 +85,8 @@ private void RecallItem(Entity<RecallMarkerComponent?> ent)
 
         _popups.PopupPredicted(Loc.GetString("item-recall-item-summon", ("item", ent)), actionOwner.Value, actionOwner.Value);
```
slarticodefast, line 86:
```suggestion
        _popups.PopupClient(Loc.GetString("item-recall-item-summon", ("item", ent)), actionOwner.Value, actionOwner.Value);
```
While we are already at it we can also fix this one. Currently everyone sees the "the item appears in your hand" popup. With `popupClient` only the action user does.

slarticodefast (changes requested):

Content.Shared/ItemRecall/SharedItemRecallSystem.cs
CooperWallace: Agreed, I think there should be a third popup to others with "the item appears in X's hand". Lmk

slarticodefast: Good idea!
In that case you can switch to the other overload of `PopupPredicted` that takes two string arguments.

CooperWallace:
@slarticodefast Think that should do it!
image

Resources/Locale/en-US/item-recall/item-recall.ftl
```diff
@@ -5,5 +5,7 @@ item-recall-item-marked = You draw a magical sigil on {THE($item)}.
 item-recall-item-already-marked = {CAPITALIZE(THE($item))} is already marked!
 item-recall-item-mark-empty = You must be holding an item!
 item-recall-item-summon = {CAPITALIZE(THE($item))} appears in your hand!
+item-recall-item-summon-witness = {CAPITALIZE(THE($item))} appears in {$name}'s hand!
```
slarticodefast, line 8:
```suggestion
item-recall-item-summon-self = {CAPITALIZE(THE($item))} appears in your hand!
item-recall-item-summon-others = {CAPITALIZE(THE($item))} appears in {$name}'s hand!
```
the usual naming convention

Content.Shared/ItemRecall/SharedItemRecallSystem.cs
```diff
@@ -83,7 +83,8 @@ private void RecallItem(Entity<RecallMarkerComponent?> ent)
         if (TryComp<EmbeddableProjectileComponent>(ent, out var projectile))
             _proj.EmbedDetach(ent, projectile, actionOwner.Value);
 
-        _popups.PopupPredicted(Loc.GetString("item-recall-item-summon", ("item", ent)), actionOwner.Value, actionOwner.Value);
+        _popups.PopupPredicted(Loc.GetString("item-recall-item-summon", ("item", ent)), Loc.GetString("item-recall-item-summon-witness", ("item", ent), ("name", actionOwner)), actionOwner.Value, actionOwner.Value);
```
slarticodefast, line 86:
```suggestion
        _popups.PopupPredicted(Loc.GetString("item-recall-item-summon", ("item", ent)), Loc.GetString("item-recall-item-summon-witness", ("item", ent), ("name", Identity.Entity(actionOwner, EntityManager))), actionOwner.Value, actionOwner.Value);
```
You will have to pass the user's identity instead or the popup will reveal their identity even if it is hidden by a mask.

slarticodefast (changes requested):

lzk228:
your cl is broken, but you don't need it for this pr anyway, as those are not playerfacing changes

slarticodefast:
You forgot to import the namespace and to adjust the loc string where it is used.

CooperWallace:
couldnt sleep, thought I'd fix this morning.
image

Content.Shared/ItemRecall/SharedItemRecallSystem.cs
```diff
@@ -5,6 +5,7 @@
 using Content.Shared.Projectiles;
 using Robust.Shared.GameStates;
 using Robust.Shared.Player;
+using Content.Shared.IdentityManagement;
```
slarticodefast, line 8: sort alphabetically

slarticodefast (commented):
I gave it another test and the vanishing popup does not appear on a second client when the recaller is outside of PVS range of the first client.
To reproduce:
- Client 1 marks an item
- Put the item on the floor next to the player of Client 2
- Client 1 moves far away, on the dev map best drift off into space
- Client 1 recalls the item
- Client 2 sees it disappear without the popup

Not sure why this is happening as the PVS filter for PopupPredictedCoordinates is using the coordinates where the item was located.

CooperWallace:
@slarticodefast Interesting I couldn't actually reproduce this issue. Although the receipient in this case would never be null, so we wouldn't ever call the else case. 

I think the better solution would be to use a dual filter. 
```
var filter = Filter.PvsExcept(recipient.Value, entityManager: EntityManager).AddPlayersByPvs(mapPos, entManager: EntityManager, playerMan: _player, cfgMan: _cfg);
```
I'll look later today

Content.Server/Popups/PopupSystem.cs
```diff
@@ -70,6 +69,26 @@ public override void PopupCoordinates(string? message, EntityCoordinates coordin
...
+        public override void PopupPredictedCoordinates(string? message, EntityCoordinates coordinates, EntityUid? recipient, PopupType type = PopupType.Small)
+        {
+            if (message == null)
+                return;
+
+            var mapPos = _transform.ToMapCoordinates(coordinates);
+            if (recipient != null)
+            {
+                // Don't send to recipient, since they predicted it locally
+                var filter = Filter.PvsExcept(recipient.Value, entityManager: EntityManager);
```
Tayrtahn, line 81: I'm pretty sure the issue is here, since PvsExcept is getting players in PVS range around *recipient*, rather than around the coordinates. I think you can just use the logic from the else case and remove recipient from it instead.

CooperWallace: I looked at it briefly over lunch, thanks for getting back to me! I'll have that up shortly

CooperWallace: Should be good!

slarticodefast (approved):

Content.Client/Popups/PopupSystem.cs
```diff
@@ -123,6 +123,12 @@ public override void PopupCoordinates(string? message, EntityCoordinates coordin
                 PopupMessage(message, type, coordinates, null, true);
         }
 
+        public override void PopupPredictedCoordinates(string? message, EntityCoordinates coordinates, EntityUid? recipient, PopupType type = PopupType.Small)
+        {
+            if (recipient != null && _timing.IsFirstTimePredicted)
```
slarticodefast, line 128: Oh wait, I missed something.
Shouldn't this check for the correct recipient instead of any recipient?
```suggestion
            if (_playerManager.LocalSession == recipient & _timing.IsFirstTimePredicted)
```

slarticodefast (changes requested):

Content.Client/Popups/PopupSystem.cs
CooperWallace: Thanks!

Content.Client/Popups/PopupSystem.cs
```diff
@@ -123,6 +123,12 @@ public override void PopupCoordinates(string? message, EntityCoordinates coordin
                 PopupMessage(message, type, coordinates, null, true);
         }
 
+        public override void PopupPredictedCoordinates(string? message, EntityCoordinates coordinates, EntityUid? recipient, PopupType type = PopupType.Small)
+        {
+            if (_playerManager.LocalSession == recipient & _timing.IsFirstTimePredicted)
```
slarticodefast, line 128:
```suggestion
            if (_playerManager.LocalSession == recipient && _timing.IsFirstTimePredicted)
```
typo

slarticodefast (changes requested):

Content.Client/Popups/PopupSystem.cs
CooperWallace: This doesn't resolve, so I'm using LocalEntity. I belive that's what PopupPredicted uses.

CooperWallace: Oh wait. This isn't needed. The original check was okay, the entity was being compared in PopupCoordinate.

CooperWallace: Okay so that's more of a sane check.

slarticodefast (approved):
Gave it another test and this time it looks good for sure!

Tayrtahn (approved):

PR #35277, opened 2025-02-18
Sentry turrets - Part 8: AI notifications

chromiumboy:
## About the PR

This PR adds a generic system for sending predefined chat notifications to players. At present, it can notify the station AI when 1) someone attempts to download its consciousness into an intellicard and 2) one of its sentry turrets opens fire on a target.

Support has been added to notify the AI when and where its vision and interaction wires have been snipped (there was legacy code for this functionality, but it doesn't appear to work), but this has been left disabled because this should be a MALF ability and/or a purchased upgrade for station AIs (the appropriate checks just need to be added when these features are added)

## Why / Balance

Rather than having chat notifications for the AI being handled in a bespoke manner, they should be handled under a generic notification system so the code can be reused 

## Technical details

Notifications are defined using a ChatNotificationPrototype. It sets what message is sent, the text coloration, and what delays should be used between notifications. ChatNotificationSystem handles the messaging logic

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: The station AI will receive chat notifications when its sentry turrets start firing upon hostile targets.

### DIFF
--- a/Content.Server/Chat/Systems/ChatNotificationSystem.cs
+++ b/Content.Server/Chat/Systems/ChatNotificationSystem.cs
@@ -1,0 +1,82 @@
+using Content.Server.Chat.Managers;
+using Content.Shared.Chat;
+using Content.Shared.Chat.Prototypes;
+using Content.Shared.Mind;
+using Content.Shared.Roles;
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Timing;
+
+namespace Content.Server.Chat.Systems;
+
+public sealed partial class ChatNotificationSystem : EntitySystem
+{
+    [Dependency] private readonly IPrototypeManager _proto = default!;
+    [Dependency] private readonly IChatManager _chats = default!;
+    [Dependency] private readonly SharedMindSystem _mind = default!;
+    [Dependency] private readonly SharedRoleSystem _roles = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+
+    // The following data does not need to be saved
+
+    // For storing chat notifications by source
+    // (Recipient, ChatNotification) -> Dictionary<Source, last TOA>
+    Dictionary<(EntityUid, ProtoId<ChatNotificationPrototype>), Dictionary<EntityUid, TimeSpan>> _chatNotificationsBySource = new();
+
+    // For storing chat notifications by type
+    // (Recipient, ChatNotification) -> last TOA
+    Dictionary<(EntityUid, ProtoId<ChatNotificationPrototype>), TimeSpan> _chatNotificationsByType = new();
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<ActorComponent, ChatNotificationEvent>(OnChatNotification);
+    }
+
+    public void OnChatNotification(Entity<ActorComponent> ent, ref ChatNotificationEvent args)
+    {
+        if (!_proto.TryIndex(args.ChatNotification, out var chatNotification))
+            return;
+
+        var source = args.Source;
+
+        // If we receive a notification before the delay time has elasped, update the timer and exit
+
+        if (chatNotification.NotifyBySource)
+        {
+            if (!_chatNotificationsBySource.TryGetValue((ent, chatNotification), out var trackedSources))
+                trackedSources = new();
+
+            trackedSources.TryGetValue(source, out var timeSpan);
+            trackedSources[source] = _timing.CurTime;
+
+            _chatNotificationsBySource[(ent, chatNotification)] = trackedSources;
+
+            if (_timing.CurTime < (timeSpan + TimeSpan.FromSeconds(chatNotification.NextDelay)))
+                return;
+        }
+
+        else
+        {
+            _chatNotificationsByType.TryGetValue((ent, chatNotification), out var timeSpan);
+            _chatNotificationsByType[(ent, chatNotification)] = _timing.CurTime;
+
+            if (_timing.CurTime < (timeSpan + TimeSpan.FromSeconds(chatNotification.NextDelay)))
+                return;
+        }
+
+        var sourceName = args.SourceNameOverride ?? Name(source);
+        var userName = args.UserNameOverride ?? (args.User.HasValue ? Name(args.User.Value) : string.Empty);
+        var targetName = Name(ent);
+
+        var message = Loc.GetString(chatNotification.Message, ("source", sourceName), ("user", userName), ("target", targetName));
+        var wrappedMessage = Loc.GetString("chat-manager-server-wrap-message", ("message", message));
+
+        _chats.ChatMessageToOne
+            (ChatChannel.Notifications, message, wrappedMessage, default, false, ent.Comp.PlayerSession.Channel, colorOverride: chatNotification.Color);
+
+        if (chatNotification.Sound != null && _mind.TryGetMind(ent, out var mindId, out _))
+            _roles.MindPlaySound(mindId, chatNotification.Sound);
+    }
+}
--- a/Content.Server/Silicons/StationAi/StationAiSystem.cs
+++ b/Content.Server/Silicons/StationAi/StationAiSystem.cs
@@ -1,12 +1,10 @@
-using System.Linq;
-using Content.Server.Chat.Managers;
 using Content.Server.Chat.Systems;
-using Content.Shared.Chat;
-using Content.Shared.Mind;
-using Content.Shared.Roles;
+using Content.Server.DeviceNetwork.Components;
+using Content.Shared.Chat.Prototypes;
 using Content.Shared.Silicons.StationAi;
 using Content.Shared.StationAi;
-using Robust.Shared.Audio;
+using Content.Shared.Turrets;
+using Content.Shared.Weapons.Ranged.Events;
 using Robust.Shared.Map.Components;
 using Robust.Shared.Player;
 using static Content.Server.Chat.Systems.ChatSystem;
@@ -15,19 +13,17 @@ namespace Content.Server.Silicons.StationAi;
 
 public sealed class StationAiSystem : SharedStationAiSystem
 {
-    [Dependency] private readonly IChatManager _chats = default!;
     [Dependency] private readonly EntityLookupSystem _lookup = default!;
     [Dependency] private readonly SharedTransformSystem _xforms = default!;
-    [Dependency] private readonly SharedMindSystem _mind = default!;
-    [Dependency] private readonly SharedRoleSystem _roles = default!;
 
-    private readonly HashSet<Entity<StationAiCoreComponent>> _ais = new();
+    private readonly HashSet<Entity<StationAiCoreComponent>> _stationAiCores = new();
 
     public override void Initialize()
     {
         base.Initialize();
 
         SubscribeLocalEvent<ExpandICChatRecipientsEvent>(OnExpandICChatRecipients);
+        SubscribeLocalEvent<StationAiTurretComponent, AmmoShotEvent>(OnAmmoShot);
     }
 
     private void OnExpandICChatRecipients(ExpandICChatRecipientsEvent ev)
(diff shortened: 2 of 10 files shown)


RangerXVII:
What if the captain wants to change the ai's laws? Can they just not do that anymore?

chromiumboy:
> What if the captain wants to change the ai's laws? Can they just not do that anymore?

Sure they can, it's just advisable that they do so while the turrets are inactive ;) If the turrets _are_ active, well, things get a little more complicated. Either convince the AI somehow to turn the turrets off or find a way to brute-force your way in

RangerXVII:
> > What if the captain wants to change the ai's laws? Can they just not do that anymore?
> 
> Sure they can, it's just advisable that they do so while the turrets are inactive ;) If the turrets _are_ active, well, things get a little more complicated. Either convince the AI somehow to turn the turrets off or find a way to brute-force your way in

That will make it more fun. The AI is always powerless to stop me when I'm captain because of the door remote.

Content.Shared/Silicons/StationAi/SharedStationAiSystem.cs
```diff
@@ -263,7 +263,7 @@ private void OnHolderInteract(Entity<StationAiHolderComponent> ent, ref AfterInt
         if (!TryComp(args.Target, out StationAiHolderComponent? targetHolder))
             return;
 
-        //Don't want to download/upload between several intellicards. You can just pick it up at that point.
+        //Don't want to download/upload between several s. You can just pick it up at that point.
```
LevitatingTree, line 266: Accidentally deleted part of the comment here?

Content.Shared/Turrets/StationAiTurretComponent.cs
```diff
@@ -0,0 +1,12 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Turrets;
+
+/// <summary>
+/// This component designates a turret that is under the direct control of the station AI
```
Fildrance: dots at the end of comment are good practice :3

Content.Shared/Silicons/StationAi/SharedStationAiSystem.cs
```diff
@@ -568,7 +566,7 @@ public sealed partial class JumpToCoreEvent : InstantActionEvent
 }
 
 [Serializable, NetSerializable]
-public sealed partial class IntellicardDoAfterEvent : SimpleDoAfterEvent;
+public sealed partial class DoAfterEvent : SimpleDoAfterEvent;
```
Fildrance: this class name is already in use - i am not sure we want them to be reused. other then that - all of them have unique names. If that would be nested class - well that would be other story..

Content.Shared/Silicons/StationAi/SharedStationAiSystem.cs
```diff
@@ -275,24 +275,24 @@ private void OnHolderInteract(Entity<StationAiHolderComponent> ent, ref AfterInt
 
         if (cardHasAi && coreHasAi)
         {
-            _popup.PopupClient(Loc.GetString("intellicard-core-occupied"), args.User, args.User, PopupType.Medium);
+            _popup.PopupClient(Loc.GetString("-core-occupied"), args.User, args.User, PopupType.Medium);
```
Fildrance: how will this work? there is not loc string such as that  and you are not concat-ing it, or so it seems

Content.Shared/Silicons/StationAi/SharedStationAiSystem.cs
```diff
@@ -275,24 +275,24 @@ private void OnHolderInteract(Entity<StationAiHolderComponent> ent, ref AfterInt
...
             args.Handled = true;
             return;
         }
         if (!cardHasAi && !coreHasAi)
         {
-            _popup.PopupClient(Loc.GetString("intellicard-core-empty"), args.User, args.User, PopupType.Medium);
+            _popup.PopupClient(Loc.GetString("-core-empty"), args.User, args.User, PopupType.Medium);
```
Fildrance: same

Content.Shared/Silicons/StationAi/SharedStationAiSystem.cs
```diff
@@ -275,24 +275,24 @@ private void OnHolderInteract(Entity<StationAiHolderComponent> ent, ref AfterInt
...
             args.Handled = true;
             return;
         }
 
-        if (TryGetHeld((args.Target.Value, targetHolder), out var held) && _timing.CurTime > intelliComp.NextWarningAllowed)
+        if (TryGetHeld((args.Target.Value, targetHolder), out var held))
```
Fildrance, line 291: Now if spam limit is reached (previously intention was that it was limited by NextWarningAllowed) - it will just silently fail, i assume? I have no idea how to do it 'properly' with the best outcome, to be honest, but at least thats a thing we should probably talk about...

Content.Shared/Chat/Prototypes/ChatNotificationPrototype.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Determines whether notification delays should be determined by the source
+    /// entity or by the notification prototype (i.e., individual notifications
+    /// vs grouping the notifications together)
+    /// </summary>
+    [DataField]
+    public bool NotifyBySource = false;
+}
+
+[ByRefEvent]
+public record ChatNotificationEvent(ProtoId<ChatNotificationPrototype> ChatNotification, EntityUid Source, EntityUid? User = null)
```
Fildrance, line 61: xml-doc would not hurt (for fields more then for other stuff)

Content.Shared/Chat/Prototypes/ChatNotificationPrototype.cs
```diff
@@ -0,0 +1,58 @@
...
+    public Color Color = Color.White;
+
+    /// <summary>
+    /// Sound played upon receiving the notification
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? Sound;
+
+    /// <summary>
+    /// Amount of time that must elapse before the next notification
```
Fildrance: Maybe better have it as TimeSpan? ;w; or we dont have deserialization for such stuff from proto? Then at least have 'Seconds' in field name. Also xml-doc should say more clearly what is it
'Time span in seconds for which notifications of same prototype will be blocked for same player after one was already delivered.' or something like that. Maybe adding info on what will happen to notificaitons that were passed in this blocked period - will they be delivered after delay? will they stack? will they be ignored with no clear indiction of that?

Content.Server/Silicons/StationAi/StationAiSystem.cs
```diff
@@ -61,15 +57,33 @@ private void OnExpandICChatRecipients(ExpandICChatRecipientsEvent ev)
...
 
+    private void OnAmmoShot(Entity<StationAiTurretComponent> ent, ref AmmoShotEvent args)
+    {
+        var xform = Transform(ent);
+
+        if (!TryComp(xform.GridUid, out MapGridComponent? grid))
+            return;
+
+        var ais = GetStationAIs(xform.GridUid.Value);
+
```
Fildrance, line 71: this empty line does not seem to fit well :3

Content.Server/Silicons/StationAi/StationAiSystem.cs
```diff
@@ -80,54 +94,59 @@ public override bool SetWhitelistEnabled(Entity<StationAiWhitelistComponent> ent
...
+        var xform = Transform(uid);
+
+        if (!TryComp(xform.GridUid, out MapGridComponent? grid))
             return;
 
-        var msg = Loc.GetString("ai-consciousness-download-warning");
-        var wrappedMessage = Loc.GetString("chat-manager-server-wrap-message", ("message", msg));
-        _chats.ChatMessageToOne(ChatChannel.Server, msg, wrappedMessage, default, false, actor.PlayerSession.Channel, colorOverride: Color.Red);
+        var ais = GetStationAIs(xform.GridUid.Value);
+
```
Fildrance, line 113: same, remove empty line here

Content.Server/Silicons/StationAi/StationAiSystem.cs
```diff
@@ -80,54 +94,59 @@ public override bool SetWhitelistEnabled(Entity<StationAiWhitelistComponent> ent
...
+        foreach (var ai in ais)
+        {
+            if (!StationAiCanDetectWireSnipping(ai))
+                continue;
+
+            var ev = new ChatNotificationEvent("AiWireSnipped", uid);
```
Fildrance: extract as static readonly field with ProtoId?

Content.Server/Silicons/StationAi/StationAiSystem.cs
```diff
@@ -61,15 +57,33 @@ private void OnExpandICChatRecipients(ExpandICChatRecipientsEvent ev)
...
+        var xform = Transform(ent);
+
+        if (!TryComp(xform.GridUid, out MapGridComponent? grid))
+            return;
+
+        var ais = GetStationAIs(xform.GridUid.Value);
+
+        foreach (var ai in ais)
+        {
+            var ev = new ChatNotificationEvent("TurretIsAttacking", ent);
```
Fildrance: extract as static readonly ProtoId

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+    [Dependency] private readonly IChatManager _chats = default!;
+    [Dependency] private readonly SharedMindSystem _mind = default!;
+    [Dependency] private readonly SharedRoleSystem _roles = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+
+    // The following data does not need to be saved
+
+    // For storing chat notifications by source
+    // (Recipient, ChatNotification) -> Dictionary<Source, last TOA>
+    Dictionary<(EntityUid, ProtoId<ChatNotificationPrototype>), Dictionary<EntityUid, TimeSpan>> _chatNotificationsBySource = new();
```
Fildrance: private readonly

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    [Dependency] private readonly IPrototypeManager _proto = default!;
+    [Dependency] private readonly IChatManager _chats = default!;
+    [Dependency] private readonly SharedMindSystem _mind = default!;
+    [Dependency] private readonly SharedRoleSystem _roles = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+
+    // The following data does not need to be saved
+
+    // For storing chat notifications by source
```
Fildrance: i'd add 'Local cache for rate limiting'

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<ActorComponent, ChatNotificationEvent>(OnChatNotification);
+    }
+
+    public void OnChatNotification(Entity<ActorComponent> ent, ref ChatNotificationEvent args)
+    {
+        if (!_proto.TryIndex(args.ChatNotification, out var chatNotification))
+            return;
```
Fildrance, line 55: shouldnt we at least log some problem? thats clearly not 'intended use' of system :<

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        var sourceName = args.SourceNameOverride ?? Name(source);
+        var userName = args.UserNameOverride ?? (args.User.HasValue ? Name(args.User.Value) : string.Empty);
+        var targetName = Name(ent);
+
+        var message = Loc.GetString(chatNotification.Message, ("source", sourceName), ("user", userName), ("target", targetName));
+        var wrappedMessage = Loc.GetString("chat-manager-server-wrap-message", ("message", message));
+
+        _chats.ChatMessageToOne
```
Fildrance: erm, wierd line break, prolly better with something like
```
_chats.ChatMessageToOne(
    ChatChannel.Notifications, 
    message, 
    wrappedMessage, 
    default, 
    false, 
    ent.Comp.PlayerSession.Channel, 
    colorOverride: chatNotification.Color
);
```

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            trackedSources.TryGetValue(source, out var timeSpan);
+            trackedSources[source] = _timing.CurTime;
+
+            _chatNotificationsBySource[(ent, chatNotification)] = trackedSources;
+
+            if (_timing.CurTime < (timeSpan + TimeSpan.FromSeconds(chatNotification.NextDelay)))
+                return;
+        }
+
```
Fildrance: invalid empty line

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+        if (chatNotification.NotifyBySource)
+        {
+            if (!_chatNotificationsBySource.TryGetValue((ent, chatNotification), out var trackedSources))
+                trackedSources = new();
+
+            trackedSources.TryGetValue(source, out var timeSpan);
+            trackedSources[source] = _timing.CurTime;
+
+            _chatNotificationsBySource[(ent, chatNotification)] = trackedSources;
```
Fildrance: you re-create (ent, chatNotification) here. And it'd be better for readability to add colliection in if block that does not find it. As you dont need to assign it again if you found it.
Also probably would be better to use ProtoId that already was created (args.ChatNotification) and not cast to it from prototype

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        // If we receive a notification before the delay time has elasped, update the timer and exit
+
+        if (chatNotification.NotifyBySource)
+        {
+            if (!_chatNotificationsBySource.TryGetValue((ent, chatNotification), out var trackedSources))
+                trackedSources = new();
+
+            trackedSources.TryGetValue(source, out var timeSpan);
+            trackedSources[source] = _timing.CurTime;
```
Fildrance: we are saving current time even if we reached limit - thats a bit off. probably we should set time only if we are really sending data. Also maybe its better to save 'next time when we will be okay with another notification' so in case of spam - we will only compare cur time with saved time and no calculation will be required unless delivery is okay.

Content.Server/Chat/Systems/ChatNotificationSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+using Content.Shared.Chat.Prototypes;
+using Content.Shared.Mind;
+using Content.Shared.Roles;
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Timing;
+
+namespace Content.Server.Chat.Systems;
+
+public sealed partial class ChatNotificationSystem : EntitySystem
```
Fildrance, line 15: xml-doc for system would be nice

Fildrance (changes requested):
Limiter for messages and how it handles limits looks a bit off - might be worth a little bit of a disucssion. I agree that it is required in some way, but there are options on how we can do it.

Fildrance:
We can do rate limiting by buffering messages for N seconds and then publishing them as string like ```{message} xN times``` but i guess that delay could feel off for some scenarios. Dunno.

Content.Shared/Silicons/StationAi/SharedStationAiSystem.cs
chromiumboy: It seems that I have accidentally removed all mention of 'intellicard' from this file. I'm not sure how I managed to mess that up. This is the cause for a lot of the other weirdness in the file. I've fixed the issue.

Content.Shared/Silicons/StationAi/SharedStationAiSystem.cs
chromiumboy: The system effectively works as it did for previous intellicard announcements, it's just handled by the ChatNotificationSystem now. If an identical message was received by the player in the past X seconds, do not send it again. I'm not sure what else you'd want here. The chat refactor is still on hold AFAIK, so there has to be a spam limiter

Content.Server/Silicons/StationAi/StationAiSystem.cs
chromiumboy: I like my white space :P

Content.Server/Chat/Systems/ChatNotificationSystem.cs
chromiumboy: The idea behind resetting the timer when a notification is received, even if the message isn't displayed to the player, is so that spammed events (such as turrets attacking someone - it could potentially take a long time for a turret to stop firing depending on circumstances) don't result in new notifications popping up every X seconds. Basically, the event needs to be resolved, or at least settled into a lull, before new notifications about that event are received

PR #33190, opened 2024-11-06
Adds space ticks to vent spawn event

K-Dynamic:
## About the PR
Adds space ticks to vent spawn event.

## Why / Balance

Not enough variety in the vent spawn event, and ticks seemed to make sense as infestations on space stations.

They are fast, bunch up and inject a lot of poison, but are easy to take down with two or so crowbar swings.

## Technical details

yml changes only

- `Resources/Prototypes/GameRules/events.yml` (adds space ticks game rule and event itself)

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

N/A

**Changelog**

:cl:
- add: Vent spawns may now include space ticks.

PR #32742, opened 2024-10-11
Reduce chance of random angry animals event spawning

Stealthbomber16:
## About the PR
Massively reduce the weight of the random angry animals spawn events. 

## Why / Balance
There are currently four random spawn events in the game with a combined weight of 16.5. These random enemies are frustrating for players, do not lead to any roleplay, prey disproportionately on new players, encourage powergaming, and cause rule issues when players take the open ghost roles and destroy station infrastructure or try to become station pets. I do not understand why these events should be so frequent.

Worth noting that I did not increase any of the other event weights in comparison. Let me know if I need to or if that breaks things.

## Technical details
I wrote this in webedit in like five minutes and haven't tested it. I'm welcome to making other changes to stop things from breaking.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- tweak: Significantly reduced the frequency of the random killer animal spawn events.

PR #35278, opened 2025-02-18
A new event, Holopad Malfunctions!

YoungThugSS14:
## About the PR

A new station event was added, known as Holopad Malfunctions, its similar to the vent critter events, however it has a different announcement and it spawns holocarps.
Future plans would be to make it so they only spawn out of holopads
## Why / Balance

More variety in station events, along with them not being an issue with balancing because they are pretty weak.
## Technical details

Made a new station announcement and added a new station event, using the vent critter code right now but that shouldn't be too much of a problem.
## Media

N/A (cant record media due to my system not being able to handle it)
## Requirements

## Breaking changes

N/A

**Changelog**

🆑 
- add: A new station event known as the Holopad malfunction, spawning Holocarps.

### DIFF
--- a/Content.Server/StationEvents/Components/VentCrittersRuleComponent.cs
+++ b/Content.Server/StationEvents/Components/VentCrittersRuleComponent.cs
@@ -1,17 +1,89 @@
 ﻿using Content.Server.StationEvents.Events;
 using Content.Shared.Storage;
+using Robust.Shared.GameObjects;
+using Content.Server.GameObjects.Components;
+using Content.Server.GameObjects.EntitySystems;
+using Robust.Shared.IoC;
 
-namespace Content.Server.StationEvents.Components;
-
-[RegisterComponent, Access(typeof(VentCrittersRule))]
-public sealed partial class VentCrittersRuleComponent : Component
+namespace Content.Server.StationEvents.Components
 {
-    [DataField("entries")]
-    public List<EntitySpawnEntry> Entries = new();
+    [RegisterComponent, Access(typeof(VentCrittersRule))]
+    public sealed partial class VentCrittersRuleComponent : Component
+    {
+        public List<EntitySpawnEntry> Entries = new();
 
-    /// <summary>
-    /// At least one special entry is guaranteed to spawn
-    /// </summary>
-    [DataField("specialEntries")]
-    public List<EntitySpawnEntry> SpecialEntries = new();
+        /// <summary>
+        /// At least one special entry is guaranteed to spawn
+        /// </summary>
+        public List<EntitySpawnEntry> SpecialEntries = new();
+
+        /// <summary>
+        /// Defines the maximum number of critters that can be spawned
+        /// </summary>
+        public int CritterCapacity { get; set; } = 10; // Default capacity is 10 critters
+
+        // List of holopad spawn points
+        public List<EntityUid> HolopadSpawnPoints { get; set; } = new();
+
+        // Method to check if the capacity has been reached
+        public bool CanSpawnMoreCritters(int currentCritterCount)
+        {
+            return currentCritterCount < CritterCapacity;
+        }
+
+        // Method to spawn a specific special entity on holopads
+        public void SpawnSpecialEntityOnHolopads(EntitySpawnEntry specialEntity, int numberToSpawn)
+        {
+            var entitySystemManager = IoCManager.Resolve<IEntitySystemManager>();
+            var entitySystem = entitySystemManager.GetEntitySystem<EntitySpawnSystem>();
+
+            int spawnedCount = 0;
+            foreach (var holopad in HolopadSpawnPoints)
+            {
+                for (int i = 0; i < numberToSpawn; i++)
+                {
+                    // Actual spawn logic
+                    entitySystem.SpawnEntity(specialEntity.EntityId, holopad.Transform.Coordinates);
+
+                    spawnedCount++;
+                    if (spawnedCount >= numberToSpawn)
+                    {
+                        break;
+                    }
+                }
+
+                if (spawnedCount >= numberToSpawn)
+                {
+                    break;
+                }
+            }
+        }
+
+        // Method to spawn multiple instances of a specific entity on holopads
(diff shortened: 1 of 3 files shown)


ArtisticRoomba:
@ScarKy0 you wanted this, you know what you must do.

Resources/Prototypes/GameRules/events.yml
```diff
@@ -584,3 +585,22 @@
...
+      path: /Audio/Announcements/attention.ogg
+    earliestStart: 20
+    minimumPlayers: 20
+    weight: 4
+    duration: 60
+  - type: VentCrittersRule
+    entries:
+    - id: MobCarpHolo
+      prob: 0.05
+
```
metalgearsloth, line 605: Isn't it better to spawn these on holopads given it's a holopad event? Just need to add component support to the event.

YoungThugSS14: > Isn't it better to spawn these on holopads given it's a holopad event? Just need to add component support to the event.

I'll be honest I have no idea how to do that.

metalgearsloth: Can try finding a coder on discord willing to do it or try to tackle it yourself.

YoungThugSS14: > Can try finding a coder on discord willing to do it or try to tackle it yourself.

I'll try tackling it myself, if I dont figure it out i'll ask on discord.

1TALERD:
Given that holocarps are not taken by regular bullets, this will bring a little more variety and chaos to the game. 
I really want to see it. 

In addition, how rare is this event?

Djungelskog2:
might i suggest you just take the %chance for this to happen out of the vent creature event? its a slightly rarer/more unique version of that event from what im looking at it functions basically the same gameplay wise

Djungelskog2:
> might i suggest you just take the %chance for this to happen out of the vent creature event? its a slightly rarer/more unique version of that event from what im looking at it functions basically the same gameplay wise

Minor thing but id also make it spawn less in general because carps (especially holocarp) are some of the more strong mobs (in my opinion)

YoungThugSS14:
> might i suggest you just take the %chance for this to happen out of the vent creature event? its a slightly rarer/more unique version of that event from what i'm looking at it functions basically the same gameplay wise

Yeah I am trying to make this its own event rule, but I know little to nothing about C# and components, so it'll probably take me a good while.

Mixelz:
> Sorry to pile on, but I had an idea that holopads that are about to spawn a holocarp would begin ringing like it's receiving a call: if the call is answered or after enough time the carp will be summoned, but if the call is declined it will prevent that carp from spawning
> 
> This will give some more agency to crew to be able to deal with the threat, and would also add some fun RP moments

Where would the calls be sourced from though? A random other department?

slarticodefast:
> might i suggest you just take the %chance for this to happen out of the vent creature event? its a slightly rarer/more unique version of that event from what im looking at it functions basically the same gameplay wise

Agreed. A current problem is that each vent critter event we add increases the total chance of them being picked out of the event pool. https://github.com/space-wizards/space-station-14/pull/33190 adds yet another type and https://github.com/space-wizards/space-station-14/pull/32742 reduces the chance to compensate. But to completely solve this and make it future proof it should be changed first so that there is only one event that picks the type from an entity table selector.

TiniestShark:
Holodeck malfunction episodes real...

Content.Server/StationEvents/Components/VentCrittersRuleComponent.cs
```diff
@@ -1,17 +1,89 @@
 ﻿using Content.Server.StationEvents.Events;
 using Content.Shared.Storage;
+using Robust.Shared.GameObjects;
+using Content.Server.GameObjects.Components;
+using Content.Server.GameObjects.EntitySystems;
+using Robust.Shared.IoC;
 
-namespace Content.Server.StationEvents.Components;
+namespace Content.Server.StationEvents.Components
+{
```
Tayrtahn, line 9: Use file scoped namespaces.

Content.Server/StationEvents/Components/VentCrittersRuleComponent.cs
```diff
@@ -1,17 +1,89 @@
...
+    [RegisterComponent, Access(typeof(VentCrittersRule))]
+    public sealed partial class VentCrittersRuleComponent : Component
+    {
+        public List<EntitySpawnEntry> Entries = new();
```
Tayrtahn, line 13: These all need to be DataFields

Content.Server/StationEvents/Components/VentCrittersRuleComponent.cs
```diff
@@ -1,17 +1,89 @@
...
+        {
+            foreach (var specialEntry in SpecialEntries)
+            {
+                if (specialEntry.EntityId == entityId)
+                {
+                    SpawnSpecialEntityOnHolopads(specialEntry, numberToSpawn);
+                    break; // Assuming you only need to spawn a specific number of this entity
+                }
+            }
+        }
```
Tayrtahn, line 73: Components should only be used to hold data, never to perform logic. Game logic should be handled in EntitySystems.

Please read up on how our ECS system works: https://docs.spacestation14.com/en/robust-toolbox/ecs.html

Tayrtahn (changes requested):

SlamBamActionman:
As slart has said, this encroaches on the vent spawn event area, and as such please refer to #35554.

YoungThugSS14:
> As slart has said, this encroaches on the vent spawn event area, and as such please refer to #35554.

Alright, I’m guessing I should close the PR till this refractor occurs?

slarticodefast:
No need. It's called a refactor, but switching the events to use entity tables should be a simple change. Someone just has to code it.

YoungThugSS14:
Im going to be later remaking this PR when the vent critter refractor is implemented, and when I am more experienced with C#, or somebody else can revive it... I dunno.

PR #35281, opened 2025-02-18
Fix contraband parenting defaulting to security.

BramvanZijp:
## About the PR
This PR makes the following changes:
1. BaseRestrictedContraband now has empty lists for both allowedDepartments and allowedJobs, as opposed to having Security in allowedDepartments.
2. BaseRestrictedContraband now clarifies it should only be used to parent other contraband prototypes, not actual items.
3. All items that previously parented off of BaseRestrictedContraband have been updated to now use BaseSecurityContraband instead.

## Why / Balance
BaseRestrictedContraband seems to have been intended to be used as a baseline for the other department restricted contrabands, yet for some reason defaulted to security. This not only resulted in a lot of security gear using BaseRestrictedContraband, despite the fact that BaseSecurityContraband exists, it also meant that BaseJanitorContraband, which parents off of BaseRestrictedContraband but did not override the allowedDepartments field, was restricted to Janitors and Security, effectively condoning or even encouraging security powergaming by using galoshes.

I was actually tempted to outright delete BaseRestrictedContraband due to how little the other contraband prototypes inherit from it, but that could cause more severe downstream issues, which this change already does somewhat.

## Media
image

## Requirements

## Breaking changes
BaseRestrictedContraband no longer restricts an item to security, and as such should NOT be used for security restricted gear, use BaseSecurityContraband instead.

**Changelog**
:cl: BramvanZijp
- fix: Fixed Janitor-Restricted contraband unintentionally stating Security may use it too.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Crates/armory.yml
+++ b/Resources/Prototypes/Catalog/Fills/Crates/armory.yml
@@ -1,6 +1,6 @@
 - type: entity
   id: CrateArmorySMG
-  parent: [ CrateWeaponSecure, BaseRestrictedContraband ]
+  parent: [ CrateWeaponSecure, BaseSecurityContraband ]
   name: SMG crate
   description: Contains two high-powered, semiautomatic rifles with four mags. Requires Armory access to open.
   components:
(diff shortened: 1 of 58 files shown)


slarticodefast (approved):
Looks good to me and works as intended.
While this is a breaking change, this fits the existing naming convention for the contraband parents much better. Also the existing integration test will find any cases of this being used in the wrong way.
Thank you for the bugfix!

ScarKy0 (approved):
I see no issues here

PR #35284, opened 2025-02-18
version watermark

Errant-4:
## About the PR
Makes it possible via cvar to display the current version on top of the screen. Additionally, servers can force all clients to show the watermark, ignoring the client setting.   Both cvars default to off, but Vulture has it enabled.

Ideally, this should also be a toggle in Options, but I didn't do that yet since it did not seem like a very high-demand feature

## Why / Balance
This would be useful to automatically mark screenshots/videos from test servers so we can skip the whole "was this on stable or master" minigame

## Technical details
Moved the existing code to generate the version string from the changelogwindow (which already used it) to ChangelogManager and made both systems refer to this new helper function

## Media

version number faked in for the shot (dev build doesn't have any)
Screenshot_2025-02-18_152532

wiiide chat
Screenshot_2025-02-18_190350

newchat
Screenshot_2025-02-18_190413

high res with no scaling makes my lazy placement code apparent, but who would choose to live like that
Screenshot_2025-02-18_190445
Screenshot_2025-02-18_190507

## Requirements

**Changelog**

:cl: Errant
- add: The game version can now be displayed as an optional watermark on the hud. Servers can force this watermark to be displayed.

### DIFF
--- a/Content.Client/Changelog/ChangelogManager.cs
+++ b/Content.Client/Changelog/ChangelogManager.cs
@@ -1,6 +1,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using Content.Shared.CCVar;
+using Robust.Shared;
 using Robust.Shared.Configuration;
 using Robust.Shared.ContentPack;
 using Robust.Shared.Serialization.Manager;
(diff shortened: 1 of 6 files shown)


lzk228:
how does it look in the "Standard" ui mode? and how does it look with wide chat?

Errant-4:
None of that really matters, since it's not pushed around, it's on a "layer" apart from everything else. But I have added more images

Errant-4:
Might as well also switch it on on Vulture to begin with, that is the whole idea after all

PR #35287, opened 2025-02-18
fix: grappling gun is ready right after putting it in hand

Fildrance:
## About the PR

Fixed grappling gun getting delay after picking it up.

## Why / Balance

Sad salv cannot do sick panic shot right after they get pushed into space! No reason really visible for delay, at least on grappling gun (which is not really a gun). 1 ammo capacity also restricts any kind of spam.

## Technical details

yaml change!

## Media

before change

[grappling_before.webm](https://github.com/user-attachments/assets/8b5b4efb-bddd-400a-8b7a-384ed166e2f8)

after change 

[grappling_after.webm](https://github.com/user-attachments/assets/33725a32-2da3-45fc-a0fb-2b10e6659976)

## Requirements

## Breaking changes

**Changelog**

:cl: Fildrance
- tweak: grappling gun shot delay on equip was removed

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Launchers/launchers.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Launchers/launchers.yml
@@ -254,6 +254,7 @@
       soundGunshot: /Audio/Weapons/Guns/Gunshots/harpoon.ogg
       fireRate: 0.5
       clumsyProof: true
+      resetOnHandSelected: false
     - type: BasicEntityAmmoProvider
       proto: GrapplingHook
       capacity: 1


slarticodefast:
Identical to https://github.com/space-wizards/space-station-14/pull/34562/files

This causes a bug when picking up picking it up while in combat mode, which makes it fire directly. This should be fixed as well.

Fildrance:
> Identical to #34562
> 
> This causes a bug when picking up picking it up while in combat mode, which makes it fire directly. This should be fixed as well.

Could add specific field for GunComponent, delay for picking up  - Slam said he added some specific stuff for Hristov to not be able to shoot right away... may be the reverse thing? eh...

slarticodefast:
It's probably a missing `Handled = true` somewhere. The mouse click triggers both picking up the gun and shooting it, but the second one gets only cancelled because of the delay.

Fildrance:
> It's probably a missing `Handled = true` somewhere. The mouse click triggers both picking up the gun and shooting it, but the second one gets only cancelled because of the delay.

ah, that sounds like nice idea. Yeah, gonna look at that..

Fildrance:
Well thats kinda sad. The way shooting is done - is by checking that thing inside GunSystem.Update loop

image

and if its false - then it creates new fire shoot event (well, RequestShootEvent but meh).
I guess its because button could be held for some time and it would mean that you don't want to stop shooting (for automatic weapons that is). There is no direct relationship between this check and picking item up, sadly... or so it looks at least.

Fildrance:
Looks like proper fix would be to use separate component, AutoShootGunComponent, for checking auto/burst shooting, and enable/disable it using common UseInteraction event, that can be marked as 'Handled', like RMC14 is doing it.

Fildrance:
This is not coming any time soon :sadge:
Will revive it when i'll have time to look at auto-shooting and gun-related events that could prevent this auto-fire on pickup behaviour

PR #35288, opened 2025-02-18
Don't open the Radial Emote Wheel if there's nothing in it

Centronias:
## About the PR
Prevents the emote wheel from displaying if there're no emotes to press in it.

## Why / Balance
Empty menu make me go 😕

## Technical details
It's just a quick and dirty implementation which:
- When constructing the emote wheel UI tracks if any emotes have been put into it
- and if none have been, it aborts the initialization of the radial menu and cleans up what it's done so far.

I am certain there's a better way to deal with this, but I haven't looked at it too hard.

## Media

https://github.com/user-attachments/assets/211edc60-a719-4ec3-abe9-0db8413add49

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: It is no longer possible to open the emote wheel in the case that it'd contain no emotes (e.g. as a ghost)

### DIFF
--- a/Content.Client/Chat/UI/EmotesMenu.xaml.cs
+++ b/Content.Client/Chat/UI/EmotesMenu.xaml.cs
@@ -20,6 +20,7 @@ public sealed partial class EmotesMenu : RadialMenu
     [Dependency] private readonly ISharedPlayerManager _playerManager = default!;
 
     public event Action<ProtoId<EmotePrototype>>? OnPlayEmote;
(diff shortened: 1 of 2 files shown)


Centronias:
Not like it was about to be insta-merged, but hang on because I, unsurprisingly, broke toggling.

Centronias:
After looking at how this interacts with the top menu button, I think it'd actually be worse to not show the empty menu, for reasons of consistency.
So bleh.

PR #35293, opened 2025-02-19
centcomm update

Emisse:
added navmap beacons/warps for: Centcomm, Afterhours, ERT, and Thunderdome
skub


ArtisticRoomba:
looks like linter is crying about the new metadata that was added to the new map saving feature

PR #35297, opened 2025-02-19
Make CharacterPickerButton the same height for dwarves as for non-dwarves

Absotively:
## About the PR
CharacterPickerButton was shorter for dwarves than for other species. This change makes it the same height for all characters.

## Why / Balance
I just thought it looked odd.

## Technical details
CharacterPickerButton's height is determined by the height of its contents, the tallest of which is usually the SpriteView showing the character sprite at double size. Dwarf sprites are shorter than other character sprites. I gave the SpriteView a SetSize so that it is 64 pixels tall even if the sprite it is displaying is shorter than 32 pixels.

## Media
Before:
image
After:
image

## Requirements

### DIFF
--- a/Content.Client/Lobby/UI/CharacterPickerButton.xaml
+++ b/Content.Client/Lobby/UI/CharacterPickerButton.xaml
@@ -8,7 +8,8 @@
         <SpriteView Scale="2 2"
                     Margin="0 4 4 4"
                     OverrideDirection="South"
-                    Name="View"/>
+                    Name="View"
+                    SetSize="64 64"/>
         <Label Name="DescriptionLabel"
                ClipText="True"
                HorizontalExpand="True"/>


Absotively:
I copied the SetSize attribute from the SpriteView in HealthAnalyzerWindow, so it's not introducing a new assumption about maximum sprite height, though it is certainly spreading the existing one.

Aeshus:
> I copied the SetSize attribute from the SpriteView in HealthAnalyzerWindow, so it's not introducing a new assumption about maximum sprite height, though it is certainly spreading the existing one.

Me when my old code comes back to haunt me...

P.S. That assumption there is kinda a mistake as being in the Cryopod can glitch out the picture, but it's honestly better than having the whole UI break to try to fit everything in...

PR #35302, opened 2025-02-19
Update RSI JSON schema to have copyright info on every line

ArtisticRoomba:
## About the PR
Updates the RSI JSON schema to allow appending copyright information to every sprite state, individually.

Let me know if you want the "license" line extended as well however the license is usually applied to all files and never individual ones.

Also changed the whitespace in the file from 3 spaced to 4 spaced. You can choose to hide whitespace changes in github when reviewing. Let me know if you want it reverted, but I thought that our schema should at least match our defined conventions.

## Why / Balance
https://github.com/space-wizards/space-station-14/blob/c62ed868546a3251d7f9867bac016534080a1a7d/Resources/Textures/Objects/Fun/toys.rsi/meta.json#L4

image

Meme explanation aside, having a one-line copyright attribution for a collection of files is truly painful. It's horrible for VCS, and it causes a million bajillion merge conflicts when multiple backlogged PRs are targeting the file (see `toys.rsi`).

This change will lead to less painful merge conflicts and VCS history.

I've been wanting to clean up JSON files and do some YAML atomization on high traffic files. I need this because when I go to do my atomizations this allows me to knock out two birds with one stone.

## Technical details
Added new copyright fields in `rsi-schema.json`.

## Requirements

**Changelog**
no cl no fun

### DIFF
--- a/.github/rsi-schema.json
+++ b/.github/rsi-schema.json
@@ -1,185 +1,190 @@
 {
-   "$schema":"http://json-schema.org/draft-07/schema",
-   "default":{
-      
-   },
-   "description":"JSON Schema for SS14 RSI validation.",
-   "examples":[
-      {
-         "version":1,
-         "license":"CC-BY-SA-3.0",
-         "copyright":"Taken from CODEBASE at COMMIT LINK",
-         "size":{
-            "x":32,
-            "y":32
-         },
-         "states":[
-            {
-               "name":"basic"
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "default": {
+
+    },
+    "description": "JSON Schema for SS14 RSI validation.",
+    "examples": [
+        {
+            "version": 1,
+            "license": "CC-BY-SA-3.0",
+            "copyright": "Taken from CODEBASE at COMMIT LINK",
+            "size": {
+                "x": 32,
+                "y": 32
             },
-            {
-               "name":"basic-directions",
-               "directions":4
-            },
-            {
-               "name":"basic-delays",
-               "delays":[
-                  [
-                     0.1,
-                     0.1
-                  ]
-               ]
-            },
-            {
-               "name":"basic-delays-directions",
-               "directions":4,
-               "delays":[
-                  [
-                     0.1,
-                     0.1
-                  ],
-                  [
-                     0.1,
-                     0.1
-                  ],
-                  [
-                     0.1,
-                     0.1
-                  ],
-                  [
-                     0.1,
-                     0.1
-                  ]
-               ]
-            }
-         ]
-      }
-   ],
-   "required":[
-      "version",
-      "license",
-      "copyright",
-      "size",
-      "states"
-   ],
-   "title":"RSI Schema",
-   "type":"object",
-   "properties":{
-      "version":{
-         "$id":"#/properties/version",
-         "default":"",
-         "description":"RSI version integer.",
-         "title":"The version schema",
-         "type":"integer"
-      },
-      "license":{
-         "$id":"#/properties/license",
-         "default":"",
-         "description":"The license for the associated icon states. Restricted to SS14-compatible asset licenses.",
-         "enum":[
-            "CC-BY-SA-3.0",
-            "CC-BY-SA-4.0",
-            "CC-BY-NC-3.0",
-            "CC-BY-NC-4.0",
-            "CC-BY-NC-SA-3.0",
-            "CC-BY-NC-SA-4.0",
-            "CC0-1.0"
-         ],
-         "examples":[
-            "CC-BY-SA-3.0"
-         ],
-         "title":"License",
-         "type":"string"
-      },
-      "copyright":{
-         "$id":"#/properties/copyright",
-         "type":"string",
-         "title":"Copyright Info",
-         "description":"The copyright holder. This is typically a link to the commit of the codebase that the icon is pulled from.",
-         "default":"",
-         "examples":[
-            "Taken from CODEBASE at COMMIT LINK"
-         ]
-      },
-      "size":{
-         "$id":"#/properties/size",
-         "default":{
-            
-         },
-         "description":"The dimensions of the sprites inside the RSI.  This is not the size of the PNG files that store the sprite sheet.",
-         "examples":[
-            {
-               "x":32,
-               "y":32
-            }
-         ],
-         "title":"Sprite Dimensions",
-         "required":[
-            "x",
-            "y"
-         ],
-         "type":"object",
-         "properties":{
-            "x":{
-               "$id":"#/properties/size/properties/x",
-               "type":"integer",
-               "default":32,
-               "examples":[
-                  32
-               ]
-            },
-            "y":{
-               "$id":"#/properties/size/properties/y",
-               "type":"integer",
-               "default":32,
-               "examples":[
-                  32
-               ]
-            }
-         },
-         "additionalProperties":true
-      },
-      "states":{
-         "$id":"#/properties/states",
-         "type":"array",
-         "title":"Icon States",
-         "description":"Metadata for icon states. Includes name, directions, delays, etc.",
-         "default":[
-            
-         ],
-         "examples":[
-            [
-               {
-                  "name":"basic"
-               },
-               {
-                  "name":"basic-directions",
-                  "directions":4
-               }
+            "states": [
+                {
+                    "name": "basic",
+                    "copyright": "Taken from CODEBASE at COMMIT LINK"
+                },
+                {
+                    "name": "basic-directions",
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
@SaphireLattice at request

ArtisticRoomba:
I should also mention that tests pass when actually using this (appending copyright stuff under sprite states) as well.

slarticodefast:
Should we maybe allow multiple licences per .rsi as well?
Because we had weird cases where a single sprite state had to be put into a separate .rsi because of that.

ArtisticRoomba:
Sure, I'll do that later today.

.github/rsi-schema.json
```diff
@@ -1,185 +1,190 @@
...
+            "title": "Copyright Info",
+            "description": "The copyright holder. This is typically a link to the commit of the codebase that the icon is pulled from.",
+            "default": "",
+            "examples": [
+                "Taken from CODEBASE at COMMIT LINK"
+            ]
+        },
+        "size": {
+            "$id": "#/properties/size",
+            "default": {},
```
ArtisticRoomba, line 110: Reminder to revert autoformatting

.github/rsi-schema.json
```diff
@@ -1,185 +1,190 @@
...
+                }
+            },
+            "additionalProperties": true
+        },
+        "states": {
+            "$id": "#/properties/states",
+            "type": "array",
+            "title": "Icon States",
+            "description": "Metadata for icon states. Includes name, directions, delays, etc.",
+            "default": [],
```
ArtisticRoomba, line 149: Reminder to revert autoformatting

ArtisticRoomba:
Made changes:
- Added License info before Copyright for all applicable properties
- Added a missing `"delays"` property to icon states as it wasn't there even though we use it. This change was committed separately and can be reverted if need be.

.github/rsi-schema.json
```diff
@@ -1,185 +1,209 @@
...
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "default": {
+
+    },
+    "description": "JSON Schema for SS14 RSI validation.",
+    "examples": [
+        {
+            "version": 1,
+            "license": "CC-BY-SA-3.0",
+            "copyright": "Taken from CODEBASE at COMMIT LINK",
```
SaphireLattice, line 11: would be better to note that those fields are not to be used together with the per-file option. I think it would be better to require one or the other, and NOT allow mixing this. "Default for the file" will get messy easily, and will lead to people forgetting to change things. Though yeah it adds extra work if someone wants to add a new file into existing RSI.

It can too easily lead to confusion in terms of "which applies?". Best to deprecate this.

In general, I am unsure if it's better to go for having a separate array for licensing/copyright that follows RGA format, or if it should be part of the image states array.

Also missing RGA's `links` property. Please use that instead of "taken from COMMIT LINK".

.github/rsi-schema.json
```diff
@@ -1,185 +1,235 @@
...
+            "title": "License",
+            "type": "string"
+        },
+        "copyright": {
+            "$id": "#/properties/copyright",
+            "type": "string",
+            "title": "Copyright Info",
+            "description": "The copyright holder. This is typically a link to the commit of the codebase that the icon is pulled from.",
+            "deprecated": true,
+            "deprecationMessage": "Whole-file attribtuion is not recommended. You should be doing individual attribution on each state!",
```
ArtisticRoomba, line 107:
```suggestion
            "deprecationMessage": "Whole-file attribution is not recommended. You should be doing individual attribution on each state!",
```

.github/rsi-schema.json
SaphireLattice: Still need to move this over to use `links` :)

.github/rsi-schema.json
```diff
@@ -1,185 +1,244 @@
...
-                  [
-                     0.1,
-                     0.1
-                  ]
-               ]
+            "states": [
+                {
+                    "name": "basic",
+                    "license": "CC-BY-SA-3.0",
+                    "copyright": "Taken from CODEBASE at COMMIT LINK"
```
SaphireLattice, line 19: Would be nice to also remove "put links into copyright field" from here given we do have that now

.github/rsi-schema.json
```diff
@@ -1,185 +1,244 @@
...
+            "deprecated": true,
+            "deprecationMessage": "Whole-file licencing is not recommended. You should be doing individual licencing on each state!",
+            "enum": [
+                "CC-BY-SA-3.0",
+                "CC-BY-SA-4.0",
+                "CC-BY-NC-3.0",
+                "CC-BY-NC-4.0",
+                "CC-BY-NC-SA-3.0",
+                "CC-BY-NC-SA-4.0",
+                "CC0-1.0"
```
SaphireLattice, line 93: Missing `CC-BY-4.0` and `CC-BY-3.0` from

https://github.com/space-wizards/RobustToolbox/blob/76b46479b6a1ad1d960fba7c8df878469590d1ea/Schemas/rga_validators.py#L6-L18

Also `MIT` and `Custom`. However applicable that is to *image files*. But never hurts to have around.

SaphireLattice (changes requested):
Honestly, just nitpicking over the `links` field not being used in examples, and "COMMIT LINK" in `copyright` fields.

As for actual "mechanical" issues: missing a few licenses that RGA can take

SaphireLattice (approved):
Looks good now, though might want to check that the verifiers work okay with all this? There's some in the RobustToolbox project IIRC.

Actually I wonder if this necessitates a version bump? Might want to ask @PJB3005 about this. (Also I wonder if you can require a static value in JSON schema?)

PJB3005 (changes requested):
This needs RSIEdit support and a sane way to handle licenses in the file. Right now you have it so *every* state needs license/copyright info. That is not practical for authoring purposes so you are gonna need to figure out how to make that work.

SaphireLattice:
The second option I had was to have something like RGA's "entry with an array of file names". Could go with that?

PJB3005:
> The second option I had was to have something like RGA's "entry with an array of file names". Could go with that?

Yes, this is already preferable.

PJB3005:
> So we're all good doing an RGA attribution.yml in each RSI, and move away from attribution in the meta.json?
> 
> If it's as simple as that, I'll close this PR and start atomizing horror files when I have the chance.

No? What Saphire said is "something **like** RGA", not *literally* RGA.

ArtisticRoomba:
I'm assuming your goal with having it contained in the meta.json is so that it can be edited really easily via RSIEdit. After talking with Saphire the direction I'm going to be going is doing attribution on the root in the form of something like:
```JSON
{
  "version": 2,
  "license": "N/A",
  "copyright": "N/A",
  "size": {
    "x": 32,
    "y": 32
  },
  "states": [
    {
      "name": "smokeaction"
    }
  ],
  "attributions": [
    {
      "files": ["smokeaction"],
      "license": "CC-BY-SA-3.0",
      "copyright": "Taken from https://github.com/discordia-space/CEV-Eris/blob/81b3a082ccdfb425f36bbed6e5bc1f0faed346ec/icons/effects/chemsmoke.dmi modified by keronshb"
    },
    {
      "files": ["secondset"],
      "license": "CC-BY-SA-3.0",
      "copyright": "second copyright..."
    }
  ]
}
```

Looks good?

Do note that my C# experience is nonexistent so adding a section in RSIEdit to modify this stuff would be a project for me. Code would be a certified :csgrad:.

ArtisticRoomba:
Something that Aeshus raised on the discord: Do we have to incorporate an RGA-like attribution system in the metadata, or can we just use RGA? It seems silly to copy an already fairly robust and standardized system for attribution into the metadata as a second format, when we could just use the one we already made for the codebase as standard.

Couldn't we just have a seperate RGA file and have RSIEdit pick up on that instead?

StormyDragon:
I admit to have tackled attribution requests without first looking at all the issues surrounding it. I also came to the conclusion that sticking everything in the `meta.json` was a horrible idea. See #39462 . Especially with the many disparate icons and overlapping changes that happened.

I feel that the least painful way to fix it was to use the RGA file since it has no conflicts with the RSI format and it seems trivial even to extend the editor to process and allow individual attributions on media in the set. Then adding the `attributions.yml` will be consistent with all other attributions also.

Sticking all the attributions in the `meta.json` feels forced and with "attributions but in json" I cringe a little also. 
I propose that the schema is changed to make `license` and `copyright` in `meta.json` optional, but the checker must assert that the fields exist if no attributions file exists. It would also be nice for a checker to make sure that if a media file is added or modified that a corresponding new attribution line is also present within `attributions.yml` this could apply to all added or modified media.

PR #29730, opened 2024-07-05
Fix Antag Bans

Titian3:
## About the PR
Fixes the currently half built Antag role banning system to actually work.
Currently working:
- Nukeops(Commander and Nukeops medic)
- Thief
- Traitor (Including SleeperAgents mid-round)
- Head Rev
- Initial infected
- Zombie (banned players will just ghost)
- Converted Rev (banned players will not convert and randomly die a random time after.)

This is a starting point for me to complete all antags and ghost roles but just this set will help admin load and issues a great deal.

Added a Random Death Smite.

Fixes #24781 #16032 #14356 #10784 #20391

## Why / Balance
Several commonly occurring admin issues relating to abuse of antag roles or ignoring game flow to just chase down antag roles can be effectively deterred with antag bans.

## Technical details
List AntagPrototype's in roleban menu and allow it to be saved as a roleban. Check to remove players from the potential selection pool on AntagSelectionSystem.

## Media
image

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes
none

**Changelog**
Make sure to take this Changelog template out of the comment block in order for it to show up.
:cl: Repo
ADMIN:
- fix: Fixed antag bans.
- add: Added random death smite.

PR #35308, opened 2025-02-19
[HOTFIX] Antag ban fix

### LINKED ISSUE #35266: Placing role bans should sanitize that jobs are valid
see ss below

Image

Killerqu00:
## About the PR
fixes #35266 
hopefully I am doing a hotfix properly this time

## Technical details
Nukeops isn't a job

## Media
server in dev doesn't crash anymore so that's a win, I guess

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Administration/Managers/BanManager.cs
+++ b/Content.Server/Administration/Managers/BanManager.cs
@@ -239,7 +239,7 @@ public sealed partial class BanManager : IBanManager, IPostInjectInit
     {
         if (!_prototypeManager.HasIndex<JobPrototype>(role))
         {
-            _chat.SendAdminAlert(Loc.GetString("cmd-roleban-unknown-role", ("role", role)));
+            _chat.SendAdminAlert(Loc.GetString("cmd-roleban-job-parse", ("job", role)));
             return;
         }
 
--- a/Content.Server/Administration/Managers/BanManager.cs
+++ b/Content.Server/Administration/Managers/BanManager.cs
@@ -237,7 +237,7 @@ public sealed partial class BanManager : IBanManager, IPostInjectInit
     // Removing it will clutter the note list. Please also make sure that department bans are applied to roles with the same DateTimeOffset.
     public async void CreateRoleBan(NetUserId? target, string? targetUsername, NetUserId? banningAdmin, (IPAddress, int)? addressRange, ImmutableTypedHwid? hwid, string role, uint? minutes, NoteSeverity severity, string reason, DateTimeOffset timeOfBan)
     {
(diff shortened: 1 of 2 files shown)


Content.Server/Administration/Managers/BanManager.cs
```diff
@@ -237,7 +237,7 @@ private void KickForBanDef(ICommonSession player, ServerBanDef def)
     // Removing it will clutter the note list. Please also make sure that department bans are applied to roles with the same DateTimeOffset.
     public async void CreateRoleBan(NetUserId? target, string? targetUsername, NetUserId? banningAdmin, (IPAddress, int)? addressRange, ImmutableTypedHwid? hwid, string role, uint? minutes, NoteSeverity severity, string reason, DateTimeOffset timeOfBan)
     {
-        if (!_prototypeManager.TryIndex(role, out JobPrototype? _))
+        if (!_prototypeManager.TryIndex(role, out JobPrototype? _) && !_prototypeManager.TryIndex(role, out AntagPrototype? _))
```
slarticodefast, line 240: These should be using `HasIndex` since the `out` variable is discarded anyways.

Errant-4 (approved):

slarticodefast:
Why are both of these prototypes handled by a single id string by the way? That looks a little sus.

Killerqu00:
> Why are both of these prototypes handled by a single id string by the way? That looks a little sus.

wish i knew

SlamBamActionman (approved):

Errant-4:
Updating severity as per issue

PJB3005 (changes requested):
This doesn't make any sense. The function is still adding the `Job:` prefix to the role name, so this would just create invalid database data.

Killerqu00:
> This doesn't make any sense. The function is still adding the `Job:` prefix to the role name, so this would just create invalid database data.

the antag bans aren't even real from the looks of it (#29730 is closed). i can just make it so it ignores the antag bans instead of throwing

PJB3005 (changes requested):
This should be getting validated (and logged) by the UI that allows admins to place bans instead.

Killerqu00:
> This should be getting validated (and logged) by the UI that allows admins to place bans instead.

fair

PJB3005:
Given that antag bans clearly aren't even a thing at all I'm gonna change the original issue description and lower the priority. I suspect the "server crashes" is just a thing caught by exception tolerance so it's not too important.

Probably best to make this not a hotfix anymore but you can do whateevr.

Killerqu00:
sure, I'll make a separate PR for this then

PR #35310, opened 2025-02-19
fix liar accent adding quotes to message

qwerltaz:
## About the PR

fix pathological liar accent changing words into double quotes `""` instead of changing into nothing. 

## Media

before
image

after
image

## Technical details
fluent empty string needs to be encased in curly brackets, otherwise it's treated as actual quote characters.

If a message like "dont" is changed into empty message nothing is sent, as expected.


lzk228 (approved):

Simyon264:
Superseded by #39370

Issue #13270, opened 2023-01-01
Admin mass action wishlist


Chief-Engineer:
Even barebones this has some use cases, many of the things listed are a wishlist within the wishlist. Primary use case for non-admemes is marking people involved in revolutions.

- [ ] A window "mass actions" that allows admins to add players to a list displayed in the window
- [ ] The list is server-side but not persistent across server restarts. A client relog by the admin does not clear the list but a server restart does clear the list
- [ ] Admins can create and name lists and select between the lists
- [ ] Lists can be created by duplicating any other list
- [ ] Default lists exist for
  - [ ] Each antag type
  - [ ] All antags combined
- [ ] Lists can be set to either be for entities or for players. Player lists are useful for admin tasks while entity lists would be useful for admemes
- [ ] Admins can see other admin's lists
- [ ] Admins can make their lists editable by other admins
- [ ] When viewing a list, an admin can set that list to mark its members in the overlay. This is clientside with no persistence beyond the session
- [ ] Admins can perform mass actions for player lists:
  - [ ] Kick
  - [ ] Roleban
  - [ ] Ban
  - [ ] Note
  - [ ] Open notes
  - [ ] Ahelp message
- [ ] Admins can perform mass actions for entity lists, or for player lists using their occupied entity:
  - [ ] Smite
  - [ ] Revive
  - [ ] Antag ctrl
  - [ ] Freeze
  - [ ] Subtle message
  - [ ] TP here
  - [ ] BQL
- [ ] Clicking on an item in the list TPs to them
- [ ] Right clicking opens the right click context menu you'd get by right clicking their entity
- [ ] A button next to each list item can be used to remove it from the list, with confirmation
- [ ] Players/entities can be added to the list with
  - [ ] The right click context menu
  - [ ] A mode similar to the spawn menu's erase that allow clicking either specific entities or selecting all in an area
- [ ] List items can have a note added under them
- [ ] Autogenerated lists
  - [ ] Connected new accounts
  - [ ] Connected accounts with active watchlist notes

ZoldorfTheWizard:
- [ ] pls add chekbox to remove disconnected players from the players tab.  should be enabled by default

kin98:
>Clicking on an item in the list TPs to them

i suggest changing it to double click instead and make click a selction like in a file explorer so you could do actions to some poeple in a list or make a new list out of the selceted people in that list and so on

PR #35311, opened 2025-02-19
New Feature: ``masskick`` command

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
It will be useful for admins to kick several players with just one command (can also help with #13270 in the future).

## Media

https://github.com/user-attachments/assets/34019e99-4171-4a2d-aca4-df70530379e1

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- add: Added ``masskick`` command.


slarticodefast (changes requested):
I don't think this implementation is a good approach.
#13270 mentions it should be possible to add players into a list using an UI and then perform various mass actions on them. This implementation is just exactly the same as the existing kick command, which can simply be run multiple times for the same effect.
What you could do instead is instead modify the existing `kick` command to be able to take multiple arguments. But that would have to be done in the RT repository.

Winkarst-cpu:
> I don't think this implementation is a good approach. #13270 mentions it should be possible to add players into a list using an UI and then perform various mass actions on them. This implementation is just exactly the same as the existing kick command, which can simply be run multiple times for the same effect. What you could do instead is instead modify the existing `kick` command to be able to take multiple arguments. But that would have to be done in the RT repository.

Reasonable.

PR #35314, opened 2025-02-19
Fix debug asserts in WoolySystem and UdderSystem

### LINKED ISSUE #34282: Debug Assert in  UdderSystem
## Description
It looks like it fails to resolve the solution container when an entity with UdderComponent is outside PVS range.
Likely introduced in https://github.com/space-wizards/space-station-14/pull/32905 which moved UdderSystem to shared.
```
Exception has occurred: CLR/Robust.Shared.Utility.DebugAssertException
An unhandled exception of type 'Robust.Shared.Utility.DebugAssertException' occurred in Robust.Shared.dll
   at Robust.Shared.Utility.DebugTools.Assert(Boolean condition) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\Utility\DebugTools.cs:line 37
   at Content.Shared.Animals.UdderSystem.Update(Single frameTime) in D:\Code\SS14\space-station-14\Content.Shared\Animals\UdderSystem.cs:line 57
   at Robust.Shared.GameObjects.EntitySystemManager.TickUpdate(Single frameTime, Boolean noPredictions) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntitySystemManager.cs:line 306
   at Robust.Shared.GameObjects.EntityManager.TickUpdate(Single frameTime, Boolean noPredictions, Histogram histogram) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntityManager.cs:line 273
   at Robust.Client.GameObjects.ClientEntityManager.TickUpdate(Single frameTime, Boolean noPredictions, Histogram histogram) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Client\GameObjects\ClientEntityManager.cs:line 194
   at Robust.Client.GameStates.ClientGameStateManager.ApplyGameState() in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Client\GameStates\ClientGameStateManager.cs:line 459
   at Robust.Client.GameController.Tick(FrameEventArgs frameEventArgs) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.cs:line 552
   at Robust.Client.GameController.<StartupContinue>b__63_0(Object sender, FrameEventArgs args) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.cs:line 243
   at Robust.Shared.Timing.GameLoop.Run() in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Shared\Timing\GameLoop.cs:line 240
   at Robust.Client.GameController.ContinueStartupAndLoop(DisplayMode mode) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.Standalone.cs:line 162
   at Robust.Client.GameController.GameThreadMain(DisplayMode mode) in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.Standalone.cs:line 147
   at Robust.Client.GameController.<>c__DisplayClass105_0.<Run>b__0() in D:\Code\SS14\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.Standalone.cs:line 106
```

**Reproduction**
- spawn a cow
- walk away until it leaves PVS range
- get a debug assert

Tayrtahn:
## About the PR

`WoolySystem` and `UdderSystem` no longer trip debug asserts when entities move out of PVS range.

## Why / Balance

Fixes #34282

## Technical details

This one's a bit of a doozy. Here's my best explanation, though there are a few parts that still don't fully make sense to me:

`UdderComponent` keeps a cached reference to the solution entity. When the cow leaves PVS, the cow is moved to nullspace and the solution is removed from the cow's solution container, but the reference in `UdderComponent` isn't cleared - it's still a valid (non-null) `Entity<SolutionComponent>` object even though the entity it points to isn't valid. When `UdderSystem` goes through the next update, it calls `ResolveSolution` with the invalid reference. Since the reference is not null, `ResolveSolution` follows the code path that checks for the debug assert, and `TryGetSolution` tries to get the container, but it fails, returning false and tripping the assert.

The fix here is to have `UdderSystem` notice (by event subscription) when the contained solution entity is removed from the container and set its cached reference to null. When `ResolveSolution` is called after that, it will properly return false without tripping the assert. When the cow comes back into PVS range, `ResolveSolution` will find the solution and recache it in `UdderComponent`.

All the above applied to `WoolySystem` too.

I'm really not a fan of copying the exact same bug fix into two classes, but it's temporary - `WoolySystem` and `UdderSystem` will be merged soon. The debug assert is really annoying for development though, so it's worth a little temporary repetition to get rid of it now.

## Media

Cows and goats, but no crashing.

## Requirements

## Breaking changes

**Changelog**

Nah.


Tayrtahn:
Actually, I might have a better fix... gimme a bit to try it out.

Content.Shared/Animals/WoolySystem.cs
```diff
@@ -24,13 +25,24 @@ public override void Initialize()
...
+    private void OnEntRemoved(Entity<WoolyComponent> entity, ref EntRemovedFromContainerMessage args)
+    {
+        // Make sure the removed entity was our contained solution
+        if (entity.Comp.Solution == null || args.Entity != entity.Comp.Solution.Value.Owner)
+            return;
+
+        // Clear our cached reference to the solution entity
+        entity.Comp.Solution = null;
+    }
+
```
metalgearsloth, line 45: I would've thought you'd just use trygetsolution and not resolvesolution for the below.

Tayrtahn: That was my initial plan, but the comments in the Solution classes recommend caching a reference to the solution rather than searching for it every time its needed. Using `ResolveSolution` lets this keep using the cached reference, for what that's worth. When the cached solution is null, the `ResolveSolution` call will cache the new reference when it gets it.

Tayrtahn: Yeah, I hear ya.

One thing that would work would be finding a way to detect that the solution entity passed into `ResolveSolution` is valid. I'm pretty sure this bug could be fixed more generally by adding that condition to the if statement on line 119:
https://github.com/space-wizards/space-station-14/blob/112d7c320f105cca1f863c55f45a00ab5d580dc3/Content.Shared/Chemistry/EntitySystems/SharedSolutionContainerSystem.cs#L119-L124

slarticodefast: >Might need a second look as I imagine this is already a widespread problem sort of getting swept under the rug.

Not really that widespread yet since the Uddersystem is one of the few predicted systems that use SharedSolutionContainer in an update loop, but it will become more common when we move more stuff to shared.
I think we should investigate why the cow is moved to nullspace, but the solution entity is deleted when moved out of PVS range.

juliangiebel: > One thing that would work would be finding a way to detect that the solution entity passed into ResolveSolution is valid. I'm pretty sure this bug could be fixed more generally by adding that condition to the if statement on line 119:

@Tayrtahn Are you going to implement this more general fix? Otherwise this comment should really be resolved and the PR can probably get merged

Tayrtahn: Sure, I'll put together another PR with the alternative fix and we can figure out which one we like better.

Tayrtahn: @juliangiebel #35872

Errant-4: Since the other solution had performance concerns, are we good to mark this resolved and proceed?

slarticodefast: I'm still not really happy with this solution because it adds boilerplate to every system using this, and it is hard to understand for contributors why this is needed.
What I would like to know is why the cow is moved into whitespace whereas the solution entity is deleted.
And if pausing entities that are out of PVS range could be a better alternative. I don't see a reason why they still need to be updated in the update loop on the client.

ElectroJr (approved):
Its fine as a fix for now, though I think we should eventually fix it some other way as well.

I.e., probably by pausing pvs-detached entities & maybe just changing the debug assert in ResolveSolution so it ignores  detached ents?


EDIT:
To clarify, just pausing entities outside of PVS range doesn't fully fix the issue, because the solution-holder might re-enter pvs range a few ticks before the solution entity. So it might still get un-paused while the solution ent is sitll missing. But it would still prevent the issue in most situations.

Issue #29821, opened 2024-07-08
TryStopNukeOpsFromConstantlyFailing test is flaky


ShadowCommander:
## Description

https://github.com/space-wizards/space-station-14/actions/runs/9837766958/job/27156273176?pr=29819
```
  Failed TryStopNukeOpsFromConstantlyFailing [12 s]
  Error Message:
   Robust.Shared.Utility.DebugAssertException : Exception of type 'Robust.Shared.Utility.DebugAssertException' was thrown.

  1) Test was dirty-disposed.

  Stack Trace:
     at Content.Server.Atmos.Piping.EntitySystems.AtmosDeviceSystem.Update(Single frameTime) in /home/runner/work/space-station-14/space-station-14/Content.Server/Atmos/Piping/EntitySystems/AtmosDeviceSystem.cs:line 124
   at Robust.Shared.GameObjects.EntityManager.TickUpdate(Single frameTime, Boolean noPredictions, Histogram histogram) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.cs:line 256
   at Robust.Server.GameObjects.ServerEntityManager.TickUpdate(Single frameTime, Boolean noPredictions, Histogram histogram) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameObjects/ServerEntityManager.cs:line 195
   at Robust.Server.BaseServer.Update(FrameEventArgs frameEventArgs) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/BaseServer.cs:line 702
   at Robust.UnitTesting.RobustIntegrationTest.IntegrationGameLoop.SingleThreadRunUntilEmpty() in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.UnitTesting/RobustIntegrationTest.cs:line 1097
   at Robust.UnitTesting.RobustIntegrationTest.IntegrationGameLoop.Run() in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.UnitTesting/RobustIntegrationTest.cs:line 1084
   at Robust.UnitTesting.RobustIntegrationTest.ServerIntegrationInstance._serverMain() in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.UnitTesting/RobustIntegrationTest.cs:line 655
```

```
[...]

PR #35318, opened 2025-02-20
Corrupt borg speech if they are damaged or low power

Quantum-cross:
## About the PR

Adds an accent component to borgs that randomly corrupts their speech according to how damaged they are or how low their power cell is.

## Why / Balance

I think it's really cool, and interesting from an RP perspective.

From a gameplay perspective, all organics are unable to speak after entering critical state at 100 damage. Borgs have a unique ability to be 100% coherent in speech and radio until they gib at 300 damage.

This change makes it difficult to communicate exact facts when they are heavily damaged, but it still allows borgs to clearly alert crew and listeners that "something" is wrong.

When the borg is close to 300 damage, the messages can be quite unintelligible and desperate sounding.

As for the charge based corruption, I'll often find myself standing still having a deep conversation with someone while my battery completely drains without me paying attention. This way there is some more interesting effects as the power cell drains.

Additionally it makes a borg in a 0% power state feel more like it's in an "emergency" or "power save" mode as it can only say extremely short messages without any error. Right now the probabilities for low charge corruption are quite conservative.

Example:
"Help" -- "Open" -- "Bring to" -- "Robotics" -- "Charge"

## Technical details

Add `DamagedSiliconAccentSystem` and `DamagedSiliconAccentComponent`, with the latter attached to prototype `BaseBorgChassis`

`OnAccent` will perform two separate corruption routines.

If the power cell is less than 15%, then any messages longer than 8 characters will start randomly start dropping characters on longer messages with increasing probability which scales with power cell charge and the position of the character (maximum probability at 40 characters).

The second routine will iterate on the message characters and roll to "corrupt" each character with a probability scaled to total damage. The probability of corruption ranges from 0% at 0 damage to 45% at 300 damage.

If a character is "damaged" there is a 50% chance to turn it into a random punctuation mark, a 25% to delete the character, or a 25% chance to repeat the character (with a continuing 25% to repeat further, "exploding dice" style).

## Media

Screenshot 2025-02-19 190050
Screenshot 2025-02-19 190134
Screenshot 2025-02-19 190235
Screenshot 2025-02-19 190352

Example output for borg with full charge and damage increasing from 0 - 300:
(Do remember that a loud alarm sounds at 75, and it is downed at 100!)
```
D: 0	'The quick brown fox jumps over the lazy dog'
D: 25	'The quuick brown f%x jumps over .he lazy dog'
D: 50	'The quick brown fox jumps over the lazy dog'
D: 75	'The qu$ck brwn f!x jumps over the lazy@dog'
D: 100	'The[qick  b\own  fox#jumps ve~ the lz` \og'
D: 125	'The  quiick b>own fox jjumps o.err   the lazy doog'
D: 150	'T( ui-k#brown fx jummps over the l.z  dog'
D: 175	'Theee qu\k brown fo%jumpss over theee lazy do<'
D: 200	'\hee quu" ?`o@n  fox jumps}"e\}t&e lazy dog'
D: 225	'The quu:c"}&ow) fox ju<ps      o?r +h laz     doggg'
D: 250	'e qu(c% brrro|n ffox }umps  over he+laa{y dog'
D: 275	'The qquic@ b"o)oox -umppps oove -: lazy d`'
D: 300	'/h quicckbr^wn  =ox  jj[mps ooveerrr the lllazyy  dg'
```
Screenshot 2025-02-19 193154

Example output for borg no damage and charge decreasing from 15% - 0%
```
C: 0.15	'The quick brown fox jumps over the lazy dog'
C: 0.12	'The quick brown fox .umps over the lz. d.g'
C: 0.09	'The quick brown fox jumps ver the laz..dog'
C: 0.06	'The quick brown fox jum.s over.th la.y dog'
C: 0.03	'The quick brown fx jumps .ver tela.y d.g'
C: 0.00	'The quick brown fox jumps ove. t. ay.d.'
```

## Requirements

## Breaking changes

None anticipated

**Changelog**

:cl:
- add: Corrupt borg speech if they are damaged or low power


Content.Server/Speech/EntitySystems/DamagedSiliconAccentSystem.cs
```diff
@@ -0,0 +1,151 @@
...
+    private string CorruptRepeat(char letter)
+    {
+        // If we got here that means we're at least repeating it once
+        var res = new StringBuilder($"{letter}{letter}");
+        // 25% chance to add another character in the streak
+        // (kind of like "exploding dice")
+        while (_random.Prob(0.25f))
+        {
+            res.Append(letter);
+        }
```
metalgearsloth, line 148: Doing randoms in a while loop is a bad idea due to RNG.

Ratonero500:
Love this.

Content.Server/Speech/EntitySystems/DamagedSiliconAccentSystem.cs
Quantum-cross: I was thinking about that, in case there is an implementation of Robust.Random that might not actually put out random numbers....

I'll just numerically calculate up to 5 repeats and hard code those in...

Vinkoln:
looks cool, hope to see it in the game

Content.Server/Speech/EntitySystems/DamagedSiliconAccentSystem.cs
Quantum-cross: removed the RNG based loop

BigfootBravo:
This is very cool!

Quantum-cross:
One thing I want to look over when I get home from work is how accents are stacked.

If a borg is wearing a cowboy hat, will that accent completely override this one or will they both be applied? 

If they are both applied, in what order? Will the corruption happen first and break word replacements? Or will we get a "H%wdddy paaa+$er"?

Quantum-cross:
easy, fixed that

image

SlamBamActionman (approved):
I think it's a fun idea, and testing it made it very fun. The fact you can't do longer messages when low-bat is a creative and good way to do it.

We have a restriction on chat content, however most of this code should be good to reuse in the chat refactor version. Nice stuff!

Quantum-cross:
> I think it's a fun idea, and testing it made it very fun. The fact you can't do longer messages when low-bat is a creative and good way to do it.
> 
> We have a restriction on chat content, however most of this code should be good to reuse in the chat refactor version. Nice stuff!

So will this have to wait until the chat refactor? Is there any discussion or tracking or that?

superjj18:
Does this get turned off with accentless?  Wouldn't mind the option

Quantum-cross:
> Does this get turned off with accentless? Wouldn't mind the option

I briefly tested it and I don't think it gets removed, which personally I think is appropriate if we consider the gameplay considerations important.

Quantum-cross:
rebased off master

Quantum-cross:
Fun fact, I have seen a few Liltenhead nukie replays in which this would have drastically altered the ability for a borg to communicate a surprise/stealth nukie threat to the crew. I really want to see this in game.

Quantum-cross:
Feedback from @MilonPL in discord:

- I feel like these values would be better as datafields on the component rather than being defined in the system itself? 
- Event listeners should use the Entity<T> pattern.
- There's some extra newlines above dependencies.
- You're missing the base call in the Initialize method.
- In the OnAccent method it might be better to define the vars first then use if statements to change them only if we get the component, rather than using ?? operators. I think it would be more readable.

Quantum-cross:
All of @MilonPL's feedback items should be resolved.

Content.Server/Speech/Components/DamagedSiliconAccentComponent.cs
```diff
@@ -0,0 +1,90 @@
...
+    [DataField]
+    public bool OverrideChargeLevel = false;
+
+    /// <summary>
+    ///     Value of charge level to use for charge level corruption effects if <see cref="OverrideChargeLevel" /> is true.
+    /// </summary>
+    [DataField]
+    public float OverriddenChargeLevel = 0.0f;
+
+
```
metalgearsloth, line 56: Nullable does the same thing.

Content.Server/Speech/Components/DamagedSiliconAccentComponent.cs
```diff
@@ -0,0 +1,90 @@
+﻿using Content.Shared.FixedPoint;
+namespace Content.Server.Speech.Components;
+
+[RegisterComponent]
+public sealed partial class DamagedSiliconAccentComponent : Component
+{
```
metalgearsloth, line 6: Shared + predicted, just because chat isn't doesn't mean we can't do it here in anticipation.

Content.Server/Speech/Components/DamagedSiliconAccentComponent.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// </summary>
+    [DataField]
+    public bool OverrideTotalDamage = false;
+
+    /// <summary>
+    ///     Value of total damage to use for damage corruption effects if <see cref="OverrideTotalDamage" /> is true.
+    /// </summary>
+    [DataField]
+    public FixedPoint2 OverriddenTotalDamageValue = 0;
+
```
metalgearsloth, line 24: As below nullable.

metalgearsloth (changes requested):

Content.Server/Speech/Components/DamagedSiliconAccentComponent.cs
Quantum-cross: Wouldn't this require synchronizing the state of the random generator? If so, I'm guessing I'll need to seed the random generator on the tick value or something like in #34863.

Going to work on this, let me know if I'm barking up the wrong tree...

Quantum-cross: Yeah I'm kind of unsure how to approach this, because battery charge levels and total damage need to be used to calculate the corruption on the client side, and I don't think those values are accessible client side, right?

Content.Server/Speech/Components/DamagedSiliconAccentComponent.cs
Quantum-cross: not sure why I didn't think to use nullables as a datafield...

Content.Server/Speech/Components/DamagedSiliconAccentComponent.cs
Quantum-cross: Resolved per discord conversation hopefully

slarticodefast (commented):
Please add `DamagedSiliconAccentComponent` to the cloning whitelist of accents to be copied on cloning in `clone.yml`.
Just in case it will ever be added to a humanoid (for example IPCs).

Quantum-cross:
> Please add `DamagedSiliconAccentComponent` to the cloning whitelist of accents to be copied on cloning in `clone.yml`. Just in case it will ever be added to a humanoid (for example IPCs).

done

Quantum-cross:
sorry I did a dumb

Quantum-cross:
I have no idea why adding DamagedSiliconAccent to the whitelist is breaking a test on the nukeops gamerule test.

See results:

- https://github.com/space-wizards/space-station-14/actions/runs/13813986826
- https://github.com/space-wizards/space-station-14/actions/runs/13815443008



> Please add `DamagedSiliconAccentComponent` to the cloning whitelist of accents to be copied on cloning in `clone.yml`. Just in case it will ever be added to a humanoid (for example IPCs).

Can this be added at a later date when such a species is implemented?

Quantum-cross:
ugh it's an intermittent test failure, I had been repeating the test for about an hour and it JUST failed locally after I posted that.

I will add it in again and hopefully the test will pass.

Quantum-cross:
see #29821

Quantum-cross:
there it goes...

slarticodefast:
Yeah, it was just a random, unrelated test fail

Quantum-cross:
Hooray! Thanks for the merge 🎊 


> I love this but one weird thing comes to mind is: Borg is not the really one talking. it's the positronic brain / man-machine interface. Even dead borg because the brain can talk can talk.

I did think of this as well. My lore explanation is that when inserted into a borg, the speech is processed through the borg's audio systems.

Gameplay wise, bare positronic brains and MMIs don't have access to radio, so it's probably fine that they can chat without corruption.

Do they have access to binary chat when outside of a borg? I don't think so but I can't remember...

Quantum-cross:
Ok, I forgot that PBs and MMIs could access binary.

Here are my thoughts on applying this effect to PBs and MMIs, it should be easy to implement it and I will probably work on it soon.

When I refer to PB I also mean to include MMIs...

- PBs currently are not damageable, this is fine, I don't think we want PBs to be round removable via damage.
- Add the damageable component to PBs, with no destruction trigger
- Add the DamagedSiliconAccent component to PBs (possibly with different corruption thresholds)
- Damage value will be automatically used by the DamagedSiliconAccent component
- Heal damage over time

This makes it so that damage on an PB will cause an acute period of time where it is unable to speak. The ability to speak clearly will return over the course of say, 10 seconds or some other time period.

I will open an issue for this.

Thanks again for the merge!

PR #31312, opened 2024-08-22
ItemToggle + slots stuff

metalgearsloth:
- Add component for itemslot locks to match LockComponent (surprised this didn't exist).
- Add thing for pointlight to match itemtoggle. In future should be used for PDAs and stuff but need to fix some other stuff first.

PR #35319, opened 2025-02-20
Bugfix for headphones item so it can play sprites properly again

### LINKED ISSUE #31755: Headphones will no longer give the alternative activated sprite
## Description
Regardless of activated or deactivated, headphones no longer have the little notes.

**Reproduction**
Turn on headphones
Wear them
Cry

**Screenshots**

**Additional context**
This physically hurts me.

Zokkie:
## About the PR
The headphones are broken. It used to work in the past in a more "hacky" way I heard but new codebase changes now prevent it from working in this way. That is why I made headphones into a proper item (just like the magboots for example) which makes them working again. 

This is my first attempt at writing C# code so I hope I did everything properly and that this is an acceptable solution. 

## Why / Balance
The headphones are broken. If you activated it the sprite would not play properly while wearing it. This fixes #31755 

## Technical details
I checked how the magboots item worked and basically tried to implement it the same way for headphones. Headphones are a super simple item so the HeadphonesSystem.cs and HeadphonesComponent.cs have a lot less code compared to the magboots .cs files for example. 

In the misc.yml file changes have been made to the headphones entity to work the same way items of this type are suppose to work (like magboots for example).

One of the headphones.rsi images has been renamed ("off-equipped-NECK" is now "equipped-NECK") 
Also two of the headphones.rsi images have been duplicated which are "inhand-left" and "inhand-right" and named "on-inhand-left" and "on-inhand-right". 
These file changes in the headphones.rsi have been made to ensure that headphones sprites work in the item system it now part of. The artwork of the images have not been changed in any way.

## Media
https://github.com/user-attachments/assets/d32696d9-e3dd-454b-a4af-cff5ee1c17c1

## Requirements

## Breaking changes
I tested it and I did not see any breaking changes.  The changed and added files also should not have any impact on other things. So all in all it should be fine.

**Changelog**
:cl:
- fix: Activated headphones now show little notes playing again when worn.


slarticodefast (changes requested):
Having a single component and system for headphones, that is only there to fix the sprite toggle not working and has nothing to do with headphones at all goes against the idea of having code be reusable, for example the telescopic shield has the exact same bug affecting it.
The sprite toggle is already part of `ItemToggleSystem`, but something broke in this PR https://github.com/space-wizards/space-station-14/pull/31312, which needs to be fixed instead. So I'm afraid this is completely the wrong approach.

Zokkie:
@slarticodefast Thanks for the review and describing where to look to fix it the proper way. I appreciate it! This will probably be beyond my current skill level but I can always give it a try.

PR #33142, opened 2024-11-03
invert the shuttle law

deltanedas:
## About the PR
changed it to THE SHUTTLE MUST BE CALLED BECAUSE OF WHATEVER ON THE STATION

## Why / Balance
while round stalling may not be against the rules for borgs with that law to do, its still just as annoying if a borg makes a hidey hole in some impossible to find area and constantly recalls it

it still can lead to the same borg hunt as before, but also if its like because of the 10 gazillion passengers on the station, the borg might be incentivised to kill all passengers :trollface:

it also means the borg doesnt have to wait for someone to call evac to start the shenanigans, as soon as you get the law you can do it

**Changelog**
:cl:
- tweak: Changed the shuttle-related ion storm law so it forces you to call the shuttle, rather than recall it.

PR #35321, opened 2025-02-20
Replace Ion law shuttle cannot be called with must be called

Minemoder5000:
## About the PR

Ion storms now give laws stating the shuttle must be called, instead of must not be called.
## Why / Balance

The law falls in a weird rule gray area where silicons must follow their laws which is the first roleplay rule, and conflicting with roleplay rule 7 stating you cant stall the round. Instead of removing the law outright which I tried before, forcing cyborgs to call the law means they will be revealed instantly, will still cause chaos regarding the emergency shuttle, and will do the exact opposite of round stalling.
## Technical details

replaced ion-storm-law-no-shuttle with ion-storm-law-call-shuttle
## Media

maZFtGR

## Requirements

## Breaking changes

The localization for ion-storm-law-no-shuttle has been replaced with ion-storm-law-call-shuttle, anything referencing ion-storm-law-no-shuttle should be replaced.

**Changelog**

:cl: Minemoder
- tweak: Ion storms will now require silicons to call the shuttle, instead of preventing it from being called.


walksanatora:
I am stealing the comms console and hiding. the shuttle WILL be called.

ArtisticRoomba (approved):
woe. approval be upon ye.

Aeshus (dismissed):
Thanks for the PR!

I wonder though if it would be appropriate to say that a localization key changing would be a breaking change. I'm leaning more on that being yes, but I don't think we have a consensus on that yet.

Aeshus:
> it would be more breaking to not change the loc id since english and translated locales would differ (english says always call russian says never call)

Yeah of course, though I’m more so wondering if we should notify them that they’ll need to localize a new key and now their old key/translation is unused.

Minemoder5000:
I added the breaking change, I figure if its not important enough then it wont be posted.

Aeshus:
Could you merge master into your branch?

I'm assuming that is what is causing the miscellaneous test failures.

slarticodefast:
Duplicate of https://github.com/space-wizards/space-station-14/pull/33142 which was denied before
I'll ping everyone in the existing discussion thread

Minemoder5000:
Closing in favor of #35751

PR #35322, opened 2025-02-20
Put Neckwear above Backpacks

TiniestShark:
## About the PR
Changes sprite layering so the Neck layer is above the Backpack layer.

## Why / Balance
A good chunk of the clothing items on the Neck Layer tend to get squashed or buried by Backpacks, Satchels, and Duffel bags. Backpacks are easily the worst of the bunch, but there's plenty of items the others also affect. This changes it so the neck layer is always sitting above backpacks so they're much easier to see.

This admittedly does create a situation where backpacks can be hidden underneath cloaks, but cloaks are fairly uncommon to the point where I feel it's a minor issue (They can also still be seen when a person is facing to the side anyway).

## Technical details
Just swaps the Neck and Back layers.

## Media
Screenshot 2025-02-19 202339
Screenshot 2025-02-19 201221

Screenshot 2025-02-19 195508
Screenshot 2025-02-19 195530

Screenshot 2025-02-19 195351
Screenshot 2025-02-19 203247

Screenshot 2025-02-19 203019
Screenshot 2025-02-19 203413

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Scarves, pins and other neckwear now render above the backpack.


Nairodian:
Side tangent: God we really need a system where we can define layering to change based on the direction.

In this case if we had a system like that, we could have backpacks layer over neck items when facing north, but when south facing they layer under neck items. Best of both worlds.

SlamBamActionman:
also why tf is the yaml map schema validator failing? >:(

TiniestShark:
> Not a fan of backpacks being hidden from behind but this is one of those things where we have to pick which looks best. Imo this is the better option.

Yeah that's my line of thinking too. It's way more common for people to have neckwear getting obscured than neckwear that'd hide the backpack. It just looks awkward when people are putting backpacks over their scarves and such.

Emisse:
i think this is fine

PR #28747, opened 2024-06-08
Buffs the particle decelerator

Flareguy:
## About the PR
Buffs the particle decelerator to be 6.5x stronger (-10 energy to -65 energy.)
In theory, this is still not good enough to do much with 1 player firing, but makes teams of players actually useful.

Also adds debug prototypes for buffed particle decelerators, to simulate multiple players.

## Why / Balance
-10 singularity energy is ridiculously terrible - you'd need a team of 10+ people to even put a dent in a low level singularity. This should make particle decelerators actually usable for situations outside of admin-spawned ERTs.

## Media
- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**
- tweak: Particle decelerators are now significantly stronger.

PR #35325, opened 2025-02-20
Portable decelerator buff

K-Dynamic:
## About the PR

- Portable decelerator now fires a spread of three anti particle projectiles
- Slightly reduce individual anti particle projectile stats (-10 to -4, 10 radiation damge to 4)
- Reduce particle decelerator crate cost from 15000 to 10000, reduce decelerators from 3 to 2

## Why / Balance

Decelerators were interesting when they got added but ultimately are never used. This is because they work too slowly in killing a singularity and are impossible to use against the smaller and faster-moving teslas.

Making them fire three particles at a time in a spread allows players to actually use decelerators to kill a singuloloose (more energy overall) or teslaloose (spread allows for margin of error when missing).

Reducing the upfront cost of the crate means cargo is able to purchase them in emergencies.

## Technical details

yml changes only

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: particle decelerators now fire three particles at a time
- tweak: particle decelerator crate cost reduced from 15000 to 10000 spesos, contents reduced from three decelerators to two

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_engineering.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_engineering.yml
@@ -134,7 +134,7 @@
     sprite: Objects/Weapons/Guns/Battery/particle_decelerator.rsi
     state: base
   product: CrateParticleDecelerators
-  cost: 15000
+  cost: 10000
   category: cargoproduct-category-name-engineering
   group: market
 
--- a/Resources/Prototypes/Catalog/Fills/Crates/engineering.yml
+++ b/Resources/Prototypes/Catalog/Fills/Crates/engineering.yml
@@ -199,12 +199,12 @@
   id: CrateParticleDecelerators
   parent: CrateEngineeringSecure
   name: particle decelerators crate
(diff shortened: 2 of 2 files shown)


ArtisticRoomba:
doesn't make sense for them to fire 3 at a time when there's only one accelerator (decelerator) port on the sprite, I would just keep it 1 and buff the singufood removal

good change otherwise, usually killing a singularity requires science to have researched the technology and prepared in advance (the thing is expensive) or cargo needs to have spare cash on hand (they spent it all on pizzas for their stay on the ATS after they heard the call)

ArtisticRoomba:
I've never had a problem with aiming a single particle, I don't think it's hard given the speed.

Tesloose is plenty possible with 3 people (although that's extremely rare). Maybe if you buff it enough you can get down to 2.

slarticodefast:
Previous discussion here: https://github.com/space-wizards/space-station-14/pull/28747
TLDR: They should deal damage proportional to the singulo's health, in addition to a flat damage value.

PR #35330, opened 2025-02-20
Revert "Make radioactive material radioactive"

metalgearsloth:
Reverts space-wizards/space-station-14#34436

This has had a cascading effect and is causing a bit too many issues:
- Doomstacks of uranium.
- Rad collectors.

It needs more finesse if it were to be brought back in the future. https://github.com/space-wizards/space-station-14/pull/35162 is just a partial revert of the change but it would be better to just have a more comprehensive design for a minor feature to stop it getting abused.

:cl:
- tweak: Removed radioactivity from uranium.


kosticia:
Looks like 🆑 is broken

slarticodefast:
Yeah, in general I like the concept of it being radioactive, but it need proper balancing to be reintroduced.
What I would like to see is an early warning sign of radiation poisoning, for example vomiting at 50 rad damage, which means you will no longer have to actively observe the health alert at all times.

Rad collectors are not a problem for uranium, because you can simply put the uranium into a superpacman, which is much more effective at power generation.

slarticodefast (approved):
Code for the revert looks good, but for some reason master is broken and all tests are failing.

slarticodefast:
@metalgearsloth Can you merge master to fix the tests? I tried to push it myself but you did not put the checkmark for that.

superjj18:
Yeah the one time I choose to not look at the patch notes, my ass gets fried grabbing engi's generator uranium

Rad damage is a lil problematic because it's one of the few damage types that can kill you with zero feedback as to what's actually harming you.

PunishedJoe:
> Yeah, in general I like the concept of it being radioactive, but it needs proper balancing with stacks to be reintroduced. What I would like to see is an early warning sign of radiation poisoning, for example vomiting at 50 rad damage, which means you will no longer have to actively observe the health alert at all times.
> 
> Rad collectors are not a problem for uranium, because you can simply put the uranium into a superpacman, which is much more effective at power generation.

feds glow
do this and its balanced

PR #35331, opened 2025-02-20
Require hwid

nikthechampiongr:
## About the PR

Stops clients who don't provide a modern hwid from connecting.

## Why / Balance

Allows for easier investigations of accounts since they will always have a hwid. Also stops someone from evading by just not providing a hwid.

## Technical details

If no modern hwids are provided we deny the connection. 

Guest connections are exempted from this.

## Requirements

## Breaking changes

If a server still relies solely on legacy hwid this cvar should not be enabled.
**Changelog**

no cl no fun

### DIFF
--- a/Content.Server.Database/Model.cs
+++ b/Content.Server.Database/Model.cs
@@ -987,6 +987,7 @@ namespace Content.Server.Database
         BabyJail = 4,
         /// Results from rejected connections with external API checking tools
         IPChecks = 5,
+        NoHwId = 6
     }
 
     public class ServerBanHit
--- a/Content.Server/Connection/ConnectionManager.cs
+++ b/Content.Server/Connection/ConnectionManager.cs
@@ -220,6 +220,11 @@ namespace Content.Server.Connection
 
             var modernHwid = e.UserData.ModernHWIds;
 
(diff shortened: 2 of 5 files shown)


metalgearsloth (approved):

Geekyhobo:
Doesn't this bar linux machines from connecting? Or did we fix that?

nikthechampiongr:
> Doesn't this bar linux machines from connecting? Or did we fix that?

linux machines have hwid

Content.Server.Database/Model.cs
```diff
@@ -987,6 +987,7 @@ public enum ConnectionDenyReason : byte
         BabyJail = 4,
         /// Results from rejected connections with external API checking tools
         IPChecks = 5,
+        NoHwId = 6
```
PJB3005, line 990: This should be called `NoHwId`.

Also, doc comment.

PJB3005 (changes requested):

Content.Server.Database/Model.cs
nikthechampiongr: It's already called NoHwId. Do you mean something else?

PJB3005: Oops, meant `NoHwid` (notice capitalization).

PJB3005 (approved):

PR #35334, opened 2025-02-20
Hotfix: Fix uplinks allowing buying conditionally restricted items

### LINKED ISSUE #35274: nukie uplink blacklist doesnt work
## Description
buyer getting changed to mind broke it

Image

SlamBamActionman:
## About the PR

Fixes #35274.
The bug where Nukies can buy job-restricted items.

## Why / Balance

Bugfix 

## Technical details

#33553 changed so that conditions check for the `MindComponent` when evaluating whether something should show in the uplink. This works for PDA uplinks, since they register the owner's mind as the owner, but Nukie uplinks have no owner and thus use the buyer entity for evaluations. Problem is, the buyer entity is the mind entity container, and therefore doesn't have a `MindComponent` to evaluate against. 

You can't guarantee that the buyer has a mind however, since surplus crates "buy" items.

So now we check if the buyer has a `MindComponent` (PDA uplinks), and if they don't we check to see if they contain a mind entity. If they do, we make that the buyer instead (Nukie uplinks). If they don't, they're a non-mind entity buying (Surplus crate). 

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Nukies no longer have access to job-restricted items.

### DIFF
--- a/Content.Server/Store/Systems/StoreSystem.Listings.cs
+++ b/Content.Server/Store/Systems/StoreSystem.Listings.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics.CodeAnalysis;
+using Content.Shared.Mind;
 using Content.Shared.Store;
 using Content.Shared.Store.Components;
 using Robust.Shared.Prototypes;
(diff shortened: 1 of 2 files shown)


Errant-4 (approved):
Works as advertised

PJB3005 (changes requested):
Do not copy paste this logic.

slarticodefast (approved):
Let's hope no one will ever make surplus crates sentient.

PR #35337, opened 2025-02-20
fix ion storm code readability

lzk228:
it was inconvenient to read, because if `plural = true` you actually get singular version, tweaked it
also a small fix in the dataset `HAS -> HAVE` for plural dataset


slarticodefast (approved):
Looks good to me!

PR #35342, opened 2025-02-20
Fix missing locales for localized forks

MilenVolf:
## About the PR

Swapped the order of parents of the ClothingOuterArmorBasic prototype

## Why / Balance

For some forks that have localized prototypes, this caused the name to be missing due to inheriting the name of the first parent that didn't have one.

Besides, I think that this order is more correct (see other prototypes in armor.yml file below as an example)

## Requirements

## Breaking changes

**Changelog**


MilenVolf:
Nvm, found better solution

PR #35354, opened 2025-02-21
Swap price of EMAG and AD.

ScarKy0:
## About the PR

Simply swaps the price and discount chance of the AD and EMAG.

## Why / Balance

The point of the original EMAG split was to lower the popularity of the EMAG and try to keep it on par with the popularity of the AD. While this was successful, the EMAG is still the 4th most bought item on the wizden servers, just behind C4 (earlier it was 3rd and C4 was 4th). I was also initially unaware people were using the emag for the interactions so much and assumed the more used effect was door and locker opening.
This is meant to be a small nudge to keep the AD and EMAG slightly closer in popularity while not keeping the EMAG such an easy option.

## Technical details

Just swapped numbers around.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Increased the price of the EMAG to 5TC. It is also now part of the "very rare discounts" instead of "rare discounts".
- tweak: Decreased the price of the Access Breaker to 4TC. It is also now part of the "rare discounts" instead of "very rare discounts".

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -870,11 +870,11 @@
   name: uplink-access-breaker-name
   description: uplink-access-breaker-desc
   productEntity: AccessBreaker
-  discountCategory: veryRareDiscounts
+  discountCategory: rareDiscounts
   discountDownTo:
-    Telecrystal: 3
+    Telecrystal: 2
   cost:
-    Telecrystal: 5
+    Telecrystal: 4
   categories:
   - UplinkDisruption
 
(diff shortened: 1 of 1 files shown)


ScarKy0:
> Makes sense to me, the emag is the more powerful one considering it gives you borg allies, free guns, ammo, and more.
Splitting the emag and access breaker made it even more powerful in my opinion since you got twice the charges for only one extra TC, and you only need to buy the one you need, making it much cheaper. We can try out this balance for now and maybe increase the price further in the future.

I want to only start increasing the price once we start doing some cleanup for the interactions. Ideally the EMAG will not recharge but the interactions will be way stronger.

ScarKy0:
> When you say most bought, does this include thieves who purchase the syndie kit bundle
I'm also not sure you'd be able to decrease popularity with one extra tc as the functions it has are quite important. It'll probably also only get more popular once we add even more emag functions for devices.

This does not include thief kits.
It isn't meant to be a full on popularity decrease, just a small nudge. And emag interactions are frozen so there will not be any new emag interactions till the current ones are cleaned up.

slarticodefast:
No one else commenting, so I'm merging.

PR #35356, opened 2025-02-21
Improve TryAllReactionsTest

TheShuEd:
## About the PR

The integration test, which checks that all reactions work properly, now checks that all reactions can be achieved by players by gradually heating or cooling. This prevents accidental situations where 2 recipes with the same reactants but different temperature range requirements can interfere with each other. 

No media.
No CL.
No fun.
No.


TheShuEd:
image
Сathced!

PR #35359, opened 2025-02-21
Role subtypes

Errant-4:
## About the PR
Role types can now have a subtype that will be visible to admins on the antag overlay and playerlist.

They are not shown to players in order to not complicate the concept/presentation of role types. It should already be obvious to them what specific antag they are anyway, from the ghost role info, chat briefing and the character window.

The antag overlay and Admin Window Playerlist both have various configuration options can both be configured independently to show either the role type, or the subtype. The Playerlist can also show both, in either order. (This option was not provided for the overlay due to space considerations, I doubt anyone would want to see both type and subtype there anyway). If subtype is not available on a mind when the display is set to show that, it will display the role type instead as a fallback. (It will not show the role type twice, if the playerlist is set to show both)

Also included is a rework of the Admin Options UI because the checkboxes were getting out of hand. The new features (and related old controls) are now controlled by dropdown lists

This PR only adds subtypes to existing mindroles. Creating a large number of new mind roles for all other misc ghostroles should be a separate PR to avoid bloat

## Why / Balance
It is currently tedious for admins to determine exactly what kind of antagonist someone is.

## Technical details
Subtypes are locid-s on the MindRoleComponents. They can be configured as part of the mind role prototype in yaml.
The currently active subtype is also kept on MindComponent and in PlayerInfo just like how RoleType already is.

## Media

Subtypes included at this time:
image

https://github.com/user-attachments/assets/bf2d6e0a-f75f-46f6-8a14-c854178ef853

## Requirements

## Breaking changes
Mind Roles can now be configured with a Subtype that provides more information about what they are, specifically. This information is currently only visible to admins, for easier quick identification of what kind of antagonist a particular player is. If a mind role does not have their subtype specified, they will only be able to show their role type (as before the change). Downstream antagonists (or any other mindroles that is to take advantage of this feature) must have their subtypes specified, see `mind_roles.yml `

Some CVars and locale identifiers related to admin playertab and admin overlays have changed, see `CCVars.Interface.cs` and `options-menu.ftl` (the ## Admin menu part)

The PlayerInfo, AdminNameOverlay and PlayerTabEntry constructors need some new input types.

**Changelog**

:cl: Errant
ADMIN:
- tweak: Mind Roles now have subtypes, which can more specifically indicate what role they are. For example, depending on clientside settings a Traitor can either show as "SOLO ANTAGONIST" or "TRAITOR" on both the admin overlay and playerlist. See Admin Options for the available settings. (Note, many minor roles don't ((currently)) have a subtype configured, and will default to showing their role type instead)


lzk228:
image

i am looking at this and thinking...
could you maybe add colors for yes and no?

slarticodefast:
Wasn't the whole point of role types to categorize them instead of having one type for every single traitor? For example the `TraitorRoleComponent` is already handling this for the `Solo Antagonist Traitor`. This feels like duplication and unneccessary overhead for adding new antags.

Errant-4:
> image
> 
> i am looking at this and thinking... could you maybe add colors for yes and no?

If I ever further touch the "Antagoinist" column, it will be to remove it

Errant-4:
> Wasn't the whole point of role types to categorize them instead of having one type for every single traitor? For example the `TraitorRoleComponent` is already handling this for the `Solo Antagonist Traitor`. This feels like duplication and unneccessary overhead for adding new antags.

This is not exactly an overhead, because you can still add new antags without giving them a subtype, same as currently. 

The alternative to subtypes being parented role types would be to define subtypes on the MindRoleComponent. It would be easier to configure in yaml when making a new antag, but it just felt kinda clumsy as a solution?

slarticodefast:
What I don't like about this is how we add yet another layer onto the already complicated mind ECS.

Mob (entity)
-> MindContainerComponent (Component on the mob)
-> Mind (entity stored in nullspace, but linked from the container)
-> MindComponent (Component on the mind entity)
-> Mind role entities (referenced in a list in the mind component)
-> traitor specific mind role components, for example TraitorRoleComponent (Component on the mind role entity)
-> MindRoleComponent (Component on the mind role entity)
-> Role types (prototype stored in the MindRoleComponent)
-> and now another layer with the sub role types, which is basically doing the same as the traitor specific mind role components above

This is getting way too convoluted and confusing for anyone who did not code this themselves.

Errant-4:
Ok moved the stored data to MindRoleComponent and got rid of the subtype role entities, it's much simpler to configure them now. Back to cleaning up the code

Content.Client/Administration/Systems/AdminSystem.Overlay.cs
```diff
@@ -22,7 +25,7 @@ public sealed partial class AdminSystem
 
         private void InitializeOverlay()
         {
-            _adminNameOverlay = new AdminNameOverlay(this, EntityManager, _eyeManager, _resourceCache, _entityLookup, _userInterfaceManager);
+            _adminNameOverlay = new AdminNameOverlay(this, EntityManager, _eyeManager, _resourceCache, _entityLookup, _userInterfaceManager, _config, _roles);
```
Simyon264, line 28: This should be split up into multiple lines for code cleanliness.

Content.Client/UserInterface/Systems/Character/CharacterUIController.cs
```diff
@@ -221,18 +221,11 @@ private void UpdateRoleType()
...
-        var color = Color.White;
-        if (_prototypeManager.TryIndex(mind.RoleType, out var proto))
-        {
-            roleText = Loc.GetString(proto.Name);
-            color = proto.Color;
-        }
-        else
-            _sawmill.Error($"{_player.LocalEntity} has invalid Role Type '{mind.RoleType}'. Displaying '{roleText}' instead");
+        if (!_prototypeManager.TryIndex(mind.RoleType, out var proto))
+            _sawmill.Error($"{_player.LocalEntity} has invalid Role Type '{mind.RoleType}'. Displaying default instead");
```
Simyon264, line 225: This error will just print the client entity ID to the console. This isn't really that bad, as the RoleType will still be what matters, but we can't really do anything with the client entity ID.

Content.Server/Administration/Systems/AdminSystem.cs
```diff
@@ -252,8 +256,18 @@ private PlayerInfo GetPlayerInfo(SessionData data, ICommonSession? session)
...
 
-        return new PlayerInfo(name, entityName, identityName, startingRole, antag, roleType, GetNetEntity(session?.AttachedEntity), data.UserId,
-            connected, _roundActivePlayers.Contains(data.UserId), overallPlaytime);
+        return new PlayerInfo(name,
+            entityName,
+            identityName,
+            startingRole,
+            antag,
+            roleType,
+            subtype,
```
Simyon264, line 265: This does an implicit conversion from `String` to `LocId`, should probably just initalize the subtype as a nullable LocId
`LocId? subtype = null;`

Content.Shared/Roles/SharedRoleSystem.cs
```diff
@@ -202,47 +201,53 @@ private bool MindRolesUpdate(Entity<MindComponent?> ent)
             return false;
 
         //get the most important/latest mind role
-        var roleType = GetRoleTypeByTime(ent.Comp);
+        var role = GetCurrentRole(ent.Comp);
```
Simyon264, line 204: For code readability you should deconstruct the returned tuple like so:
`var (protoId, locId) = GetCurrentRole(ent.Comp);`.
You don't need to name the variables like I did.

Content.Shared/Roles/SharedRoleSystem.cs
```diff
@@ -202,47 +201,53 @@ private bool MindRolesUpdate(Entity<MindComponent?> ent)
...
         {
             var comp = Comp<MindRoleComponent>(role);
             if (comp.RoleType is not null)
-                roles.Add(comp.RoleType.Value);
+                roles.Add((comp.RoleType.Value, comp.Subtype));
         }
 
-        ProtoId<RoleTypePrototype> result = (roles.Count > 0) ? roles.LastOrDefault() : "Neutral";
+        var result = (roles.Count > 0) ? roles.LastOrDefault() : ("Neutral", null);
         return (result);
```
Simyon264, line 232: You can just directly return the result. No need to put it in a variable which is only assigned once and then never changed.

Content.Shared/Roles/SharedRoleSystem.cs
```diff
@@ -620,6 +625,30 @@ public void MindPlaySound(EntityUid mindId, SoundSpecifier? sound, MindComponent
 
         return antag.Requirements;
     }
+
+    /// <summary>
+    /// Returns the localized name of a role type's subtype and/or role type.
+    /// </summary>
+    /// <param name="roleType">The locale string of the role type</param>
+    /// <param name="subtype">The locale string of the subtype</param>
```
Simyon264, line 633: These two can be made to require a `LocId` as input instead of a string to avoid implicit conversions. All current usages of this method already pass in a LocId.

Simyon264 (changes requested):

Content.Client/Administration/AdminNameOverlay.cs
```diff
@@ -14,22 +15,29 @@ namespace Content.Client.Administration;
 
 internal sealed class AdminNameOverlay : Overlay
 {
-    [Dependency] private readonly IConfigurationManager _config = default!;
```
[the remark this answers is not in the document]
Errant-4: This was collateral damage for SharedRoleSystem

Content.Shared/Roles/SharedRoleSystem.cs
Errant-4: At one point it was all changed to string, and I forgot to change these back

Content.Client/Administration/AdminNameOverlay.cs
```diff
@@ -14,22 +15,29 @@ namespace Content.Client.Administration;
...
     private readonly IUserInterfaceManager _userInterfaceManager;
+    private readonly IConfigurationManager _config;
+    private readonly SharedRoleSystem _roles;
     private readonly Font _font;
+    private readonly Font _fontBold;
+    private bool _subtypesOnly;
 
     //TODO make this adjustable via GUI
     private readonly ProtoId<RoleTypePrototype>[] _filter =
         ["SoloAntagonist", "TeamAntagonist", "SiliconAntagonist", "FreeAgent"];
```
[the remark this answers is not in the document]
Errant-4: It was imagined as a QoL feature that neutrals and familiars don't get their role types displayed, keeps the antag overlay slightly less visually busy. Additionally, it was the intent that at some point the filter will be customizable by individual admins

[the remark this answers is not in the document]
Simyon264: Yeah, seems fine then.

Content.Client/UserInterface/Systems/Character/CharacterUIController.cs
Errant-4: Replaced it with player local session string

Errant-4:
I made the new cvar more consistent with how the last cvar for the overlay was named/placed

Errant-4:
I put back the legacy overlay

Resources/Prototypes/Roles/MindRoles/mind_roles.yml
```diff
@@ -113,18 +113,19 @@
   - type: MindRole
     antagPrototype: SubvertedSilicon
     roleType: SiliconAntagonist
+    subtype: "rule-subtype-subverted"
```
slarticodefast, line 116:
```suggestion
    subtype: rule-subtype-subverted
```
these don't need to be in speech marks I think
same for the others below

slarticodefast:
Wouldn't it be more readable if we split up this column into role types and the role itself instead of using subrole types?
grafik

So instead of 
| Role type |
|--|
| Team Antagonist - Head Rev|
| Crew Aligned |

We have
| Role type | Role |
|--|--|
| Team Antagonist | Head Rev |
| Crew Aligned | |

That would be much more readable in my opinion than subtypes.
I did not follow the admin discussion for that though.

Errant-4:
I'll get some unrelated overlay and playerlist changes and improvements out of the way first, then get back to this

Content.Client/Administration/AdminNameOverlay.cs
```diff
@@ -21,15 +22,26 @@ internal sealed class AdminNameOverlay : Overlay
...
-    public AdminNameOverlay(AdminSystem system, IEntityManager entityManager, IEyeManager eyeManager, IResourceCache resourceCache, EntityLookupSystem entityLookup, IUserInterfaceManager userInterfaceManager)
+    public AdminNameOverlay(AdminSystem system,
+        IEntityManager entityManager,
+        IEyeManager eyeManager,
+        IResourceCache resourceCache,
+        EntityLookupSystem entityLookup,
+        IUserInterfaceManager userInterfaceManager,
+        SharedRoleSystem roles)
     {
         IoCManager.InjectDependencies(this);
```
Simyon264, line 46: After asking PJB:
Overlays should use the constructor to pass in dependencies. Remove this call to `IoCManager`.

Content.Shared/CCVar/CCVars.Interface.cs
```diff
@@ -37,6 +37,14 @@ public sealed partial class CCVars
     public static readonly CVarDef<bool> OutlineEnabled =
         CVarDef.Create("outline.enabled", true, CVar.CLIENTONLY);
 
+    /// <summary>
+    /// If true, the antag overlay will show role subtype.
+    /// If false, it will show both role type and subtype.
+    /// If subtype is not set for a mob's mind then role type will be shown regardless of setting.
```
Simyon264, line 43: Intend with one tab.

Simyon264 (changes requested):
Looks mostly good now! Really like this.

Errant-4:
Some usability concerns were raised about this PR in it's current form, which I try to address in changes included in/made possibe by #35538 . As such, that PR should be merged before this one

Content.Shared/CCVar/CCVars.Interface.cs
Errant-4: Not sure what you mean here?

Errant-4:
I don't want to alarm anyone but this is finally happening

Resources/Locale/en-US/mind/role-types.ftl
```diff
@@ -7,3 +7,17 @@ role-type-silicon-name = Silicon
...
+role-subtype-ninja = Ninja
+role-subtype-nukie = Nukie
+role-subtype-traitor-reinforcement = Reinforcement
+role-subtype-revolutionary = Rev
+role-subtype-head-revolutionary = Head Rev
+role-subtype-initial-infected = Infected
+role-subtype-zombie = Zombie
+role-subtype-dragon = Dragon
+role-subtype-survivor = Survivor
+rule-subtype-subverted = Subverted
```
slarticodefast, line 23: where paradox clone 😞

Errant-4: Fair, just added them

Simyon264 (approved):

PR #35368, opened 2025-02-21
Fix virtual items appearing in cryo storage.

### LINKED ISSUE #35317: Putting people into cryo with virtual items (hand blockers or otherwise) shows them in cryo (and might obliterate servers)
## Description
Putting an entity in cryo with virtual items in their hands (like handcuffs) allows you to see the virtual items in cryo.

**Reproduction**
Handcuff a person.
Toss them into cryo and have them fully cryo.
Access their inventory and see virtual items.

**Screenshots**
image

**Additional context**
Reportedly the Lizard server got obliterated ~5 minutes after this was done so it may be related? But still a bug nonetheless. I haven't reproduced it I'm just relaying it.

VerinSenpai:
## About the PR
Fixes #35317

Removes virtual items from persons in cryostorage.

## Why / Balance
Current understanding of the bug is as such, cuffs and wielded items show virtual items which shouldn't be visible at all in the gui for cryo storage. If you attempt to remove the cuffs, an error is raised on the server because the entity that is cuffed no longer exists. A wielded item can still be removed if the non-virtual side is removed. Adding on to this, otherwise non-removable hardsuit helmets can be removed from a person in cryo.
Currently theres a bug where virtual items will appear and be removable from persons in cryo storage. This typically raises errors.

## Technical details
~~Implements an event EnterCryostorageEvent and adds handlers for it to shared wieldablesystem and cuffablesystem. In order to make this work, we first uncuff or unwield, and then delete the virtual items.~~
Moved the code that handles transferring items to InventorySystem and HandSystem respectively. Replaced code in CryostorageSystem and SharedStrippableSystem with calls to those functions. Also needed to add a check for virtual items that get added to the UI to ensure virtual items aren't shown unless they're handcuffs.

## Media

https://github.com/user-attachments/assets/56d2c136-6ea6-4db9-812e-c71a9ddae0b2

## Requirements


VerinSenpai:
Beck is suggesting using an InventoryRelayedEvent but I'm unsure how to even begin implementing that.

VerinSenpai:
Update. Succeeded in getting HeldRelayedEvent to work for deleting the virtual items. Issue now is that an error raises on the client. I *believe* the error is being raised because its trying to delete an item that used to exist but no longer does.

metalgearsloth:
Having to hardcode this for cryo is the wrong approach as any other similar systems now suffer the same bug.

I'm more interested in why this even happens and why the virtual entities are even being deparented to begin with when nothing should be happening to them besides a map change (which we handle in every other instance just fine).

VerinSenpai:
> Having to hardcode this for cryo is the wrong approach as any other similar systems now suffer the same bug.
> 
> I'm more interested in why this even happens and why the virtual entities are even being deparented to begin with when nothing should be happening to them besides a map change (which we handle in every other instance just fine).

any chance I could get elaboration on that deparenting part? the entities are still virtual items when the player character is moved to cryostorage. I've also been looking around and I noticed that in the strippablesystem a check is present for virtual items and within that check is another for cuffs. I know you said you didn't want it hardcoded for cryostorage but I figured I'd bring it up just to see what hooked.

VerinSenpai:
actually, I think I understand. The handcuffs for instance aren't removable because they are no longer linked to an entity that exists, or something like that.

VerinSenpai:
@metalgearsloth when you're able to, can I get your input on this again? The approach I took with it currently was moving the code that handles transferring items between people to their respective systems and calling that code within strippable and cryostorage.

VerinSenpai:
somEWHEEREREREER oVER THE PRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR somEWHEREEEER OveR THE SKYYYYYY

PR #33968, opened 2024-12-20
Fixed de powered airlocks/doors not being able to be closed

zHonys:
## About the PR

Fixed https://github.com/space-wizards/space-station-14/issues/33795

## Why / Balance

Fixes a Bug

## Technical details

Added new property "BeingPried" in Content.Shared.Doors.DoorComponent

Changed Content.Shared.Doors.Systems.SharedAirlockSystem OnBeforeDoorClosed() and CanChangeState() to check for BeingPried door flag

Changed Content.Shared.Doors.Systems.SharedDoorSystem OnAfterPry to set BeingPried to true when closing

Changed Content.Shared.Doors.Systems.SharedDoorSystem set BeingPried to false after CanClose is called

## Media

https://github.com/user-attachments/assets/f2714a4b-e590-4a35-a650-d4756a3b4879

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

No breaking change is expected

**Changelog**

:cl:
- fix: Fixed Airlocks and FireLocks not being able to be closed when de powered

PR #35380, opened 2025-02-22
Fix prying shut unpowered doors

### LINKED ISSUE #33795: Depowered doors can't be closed.

PJB3005:
Anyways, yes point and laugh at the dysfunctional Wizden maintainer team because nobody stepped up to fix a critical game-breaking bug like this. Jesus christ.

The core of the issue is that door prying code is complete spaghetti, and prying code bypasses all normal door closing checks, instead substituting it with its own. Except door closing checks are done *twice*, once when the door starts closing and once halfway through. Prying only skips the first.

Airlocks had special code to skip the second by checking whether the door was already in "Partial" state. This is of course a complete nightmare, and unsurprisingly it broke when Sloth rearranged the state logic in #33481. At least it says in a common clearly what the intent of that logic was.

Because I don't want to sit here and rewrite all this prying code spaghetti for what should be a CRITICAL HOTFIX TO A GAME BREAKING BUG WE'VE BEEN SLACKING ON FOR THE PAST **2 FUCKING MONTHS**, I'm just gonna fix the original hack by passing the "second check" explicitly as a boolean in the event instead of storing it as state in the component data.

WHY THIS IS BETTER THAN THE OTHER PRS:

#33968 adds even more state to the component to add another layer of cruft, and did not root cause the issue to remove the original (now dysfunctional) hack.

#33843 similarly does not root cause the issue and also does not remove the original dysfunctional hack.

Christ man. We're maintainers, this wasn't even that hard.

Fixes #33795

:cl:
- fix: Fixed prying shut unpowered doors with a crowbar.

### DIFF
--- a/Content.Shared/Doors/DoorEvents.cs
+++ b/Content.Shared/Doors/DoorEvents.cs
@@ -44,15 +44,20 @@ namespace Content.Shared.Doors
     /// </summary>
     /// <remarks>
     /// This event is raised both when the door is initially closed, and when it is just about to become "partially"
-    /// closed (opaque & collidable). If canceled while partially closing, it will start opening again. Useful in case
+    /// closed (opaque &amp; collidable). If canceled while partially closing, it will start opening again. Useful in case
     /// an entity entered the door just as it was about to become "solid".
     /// </remarks>
     public sealed class BeforeDoorClosedEvent : CancellableEntityEventArgs
     {
+        /// <summary>
+        /// If true, this check is being performed when the door is partially closing.
+        /// </summary>
+        public bool Partial;
         public bool PerformCollisionCheck;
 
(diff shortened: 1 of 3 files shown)


VasilisThePikachu (approved):

VerinSenpai:
by the gods

lzk228:
let me guess
pjb decided to play the game again :godo:

deltanedas:
(big big mistake)

zHonys:
Indeed a better fix, good work! Finally something was merged, i was almost just waiting for the door System rework to fix it

PR #35382, opened 2025-02-22
Clean up the implementation of the Plasma / Oxygen Burn Reaction

Centronias:
## About the PR

Made the implementation of the plasma / oxygen burn reaction more readable and imperceptibly more performant.

I suggest reviewers review the two commits separately because the `namespace { ... }` to `namespace ...;` change is mostly just whitespace changes.

## Why / Balance

No balance changes the actual reaction should work exactly the same.
Rewritten for easier maintenance and reverse engineering of game mechanics.

## Technical details

Just look at it. But the big things are:
- simplify conditionals because there were some codepaths that were implicitly impossible (eg. energy released when no burn happened)
- move vars closer to where they're actually needed (except of course when grabbing values prior to changing the mix)
- add a helper for the common `Clamp((value - min) / (max - min), 0, 1)` ratio calculation

## Media

N/A
Unless you want me to demo fires still working 🤷 

## Requirements

## Breaking changes

None

**Changelog**

N/A

### DIFF
--- a/Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
+++ b/Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
@@ -1,3 +1,4 @@
+using System.Runtime.CompilerServices;
 using Content.Server.Atmos.EntitySystems;
 using Content.Shared.Atmos;
 using Content.Shared.Atmos.Reactions;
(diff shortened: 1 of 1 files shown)


Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
```diff
@@ -9,77 +10,79 @@ namespace Content.Server.Atmos.Reactions;
...
-                plasmaBurnRate = temperatureScale * (initialOxygenMoles / Atmospherics.PlasmaOxygenFullburn) / Atmospherics.PlasmaBurnRateDelta;
+            // The maximum rate based on the relative proportions of reactants, expressed as moles of plasma burned.
+            var maximumReactionRate =
+                Math.Min(initialPlasmaMoles, initialOxygenMoles / Atmospherics.PlasmaOxygenFullburn);
+            var reactionRate = maximumReactionRate * temperatureReactionScale / Atmospherics.PlasmaBurnRateDelta;
 
-            if (plasmaBurnRate > Atmospherics.MinimumHeatCapacity)
+            // TODO I have no fucking idea why the rate is compared against a heat capacity here.
+            // ... but it has the effect of not doing little finicky adjustments, I guess.
+            if (reactionRate > Atmospherics.MinimumHeatCapacity)
```
Centronias, line 37: Should `Atmospherics.MinimumHeatCapacity` be replaced by something else here?
Not to say the value should change, but it really doesn't make sense that this is comparing a number of moles to a specific heat.

Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
```diff
@@ -1,86 +1,88 @@
...
-        }
+    /// <summary>
+    /// Normalizes <paramref name="value"/> as a number in the range of <paramref name="min"/> to <paramref name="max"/>,
+    /// returning a value from 0.0, when <paramref name="value"/> is <paramref name="min"/>, to 1.0, when
+    /// <paramref name="value"/> is <paramref name="max"/>.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static float ProportionOfRange(float value, float min, float max)
+    {
+        return Math.Clamp((value - min) / (max - min), 0.0f, 0.1f);
```
whatston3, line 86: 1.0f and not 0.1f?

Centronias: Oh yeah, that's a big dumb on my part.

Centronias:
```suggestion
        return Math.Clamp((value - min) / (max - min), 0.0f, 1.0f);
```

Partmedia:
Per our [pull request guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html), formatting changes must be in a separate PR. The reason is that when this PR gets merged by squashing, it will be very difficult to track what changed.

Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
```diff
@@ -9,78 +10,81 @@ namespace Content.Server.Atmos.Reactions
...
+            float heatScale)
         {
-            var energyReleased = 0f;
-            var oldHeatCapacity = atmosphereSystem.GetHeatCapacity(mixture, true);
             var temperature = mixture.Temperature;
-            var location = holder as TileAtmosphere;
             mixture.ReactionResults[(byte)GasReaction.Fire] = 0;
 
-            // More plasma released at higher temperatures.
-            var temperatureScale = 0f;
```
Partmedia, line 21: It would be more clear if you didn't change the variable names for a variable that shouldn't functionally change.

Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
```diff
@@ -9,78 +10,81 @@ namespace Content.Server.Atmos.Reactions
...
-                    plasmaBurnRate = initialPlasmaMoles * temperatureScale / Atmospherics.PlasmaBurnRateDelta;
-                else
-                    plasmaBurnRate = temperatureScale * (initialOxygenMoles / Atmospherics.PlasmaOxygenFullburn) / Atmospherics.PlasmaBurnRateDelta;
+                // The maximum rate based on the relative proportions of reactants, expressed as moles of plasma burned.
+                var maximumReactionRate =
+                    Math.Min(initialPlasmaMoles, initialOxygenMoles / Atmospherics.PlasmaOxygenFullburn);
+                var reactionRate = maximumReactionRate * temperatureReactionScale / Atmospherics.PlasmaBurnRateDelta;
 
-                if (plasmaBurnRate > Atmospherics.MinimumHeatCapacity)
+                // TODO I have no fucking idea why the rate is compared against a heat capacity here.
```
Partmedia, line 35: Maybe because rate times an implicit tickRate is a change in heat capacity?

Partmedia: Can you find out if the rate is in unit/tick or unit/sec?

Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
Centronias: To make sure I understand, "it would be more clear" refers to the change set rather than the actual implementation?

"Rate" doesn't seem like the right word to me since it's affecting both the amount of reactants and the relative proportion. So neither name seems best to me.
Would `temperatureFactor` or something similar be alright, or is leaving it as it was your preference?

Partmedia: Try to put yourself in the shoes of a reviewer. I see a pretty dense and substantial diff and I need to prove to myself that there are no functional changes. Keeping variable names that haven't changed functionally would help with this.

Centronias: Believe me, I can empathize with a reviewer faced with dense reviews. My intention was to, admittedly, place a bit of one-time strain on reviewers of this PR to make the strain on maintainers of the code lighter. It's very much a subjective thing, so I've reverted the name of this local.

Content.Server/Atmos/Reactions/PlasmaFireReaction.cs
Centronias: As far as I can tell, it's units of gas per [AtmosTick](https://github.com/space-wizards/space-station-14/blob/68ce53ae17985876d6d112b764b2144964a9f42e/Content.Server/Atmos/EntitySystems/AtmosphereSystem.CVars.cs#L26).

`Atmospherics.MinimumHeatCapacity` is only compared to other heat capacities or used as a default heat capacity except in this case, and in the case of the Freezon + N2 reaction, which seems to be very similar to the plasmox reaction, so it may be a copy/paste.

So if `reactionRate`'s dimension `mass / time` and `Atmospherics.MinimumHeatCapacity`'s is (real world) `energy / (mass * temperature)`, it reaaaaally seems to me like comparing them does not describe anything physically meaningful.

PR #35388, opened 2025-02-22
Roof data rework

metalgearsloth:
Requested by Ed, separate data structure using bitmasks and not stored directly on tile.

breaking changes:
- The TileFlag for rooves got wrapped into RoofComponent. RoofSystem has its own method for "IsRooved" to determine if a tile is rooved, SetRoof API is identical.


TheShuEd:
Ok, code looks good for me.
Conceptual approve from me too

TheShuEd (approved):

NoElkaTheGod:
So you can actually look at the ceiling now? :godo:

PR #35394, opened 2025-02-22
[HOTFIX] Borg emag notification and sound fix

ScarKy0:
## About the PR

It turns out I did a silly and forgot to mark methods as override in my EMAG split PR.
This fixes that so borgs get notified when they get emagged again.

## Why / Balance

Bugfix.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Cyborgs now correctly get notified when they get emagged again.

### DIFF
--- a/Content.Server/Silicons/Laws/SiliconLawSystem.cs
+++ b/Content.Server/Silicons/Laws/SiliconLawSystem.cs
@@ -172,14 +172,18 @@ public sealed class SiliconLawSystem : SharedSiliconLawSystem
         });
     }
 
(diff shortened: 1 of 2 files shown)


Errant-4 (approved):
Works as advertised, with a caveat. Unfortunately, there is a bug where the Altered Silicon role will not _go away_ when the brain is taken out of the borg body. This is caused by an unrelated bug in SharedRoleSystem that was simply not visible due to this bug (and might have been around previously, just be too obscure to notice). It will get its own hotfix

PR #35399, opened 2025-02-22
[HOTFIX] Altered Silicon removal fails to trigger role update

Errant-4:
## About the PR
Removing an emagged borg's brain now correctly reverts it to Silicon on the Character window and admin UI.

(Note, currently the Altered Silicon role isn't **added** in the first place for emagged borgs, that will be fixed by #35394)

## Why / Balance
bug

## Media

35394 + 35399

https://github.com/user-attachments/assets/5e852476-5a8e-47b0-84af-55591049d5dc

## Technical details
A check that was originally for safeguarding against the rare/unexpected situation where Mind.OwnedEntity is null, is being triggered when called as the borg's brain is removed, as apparently at that point it's not owning either the old or the new mob entity.  This check was there to prevent the update event from being raised on a null entity, but the event was at some point changed to be broadcast, so it's irrelevant _what_ entity it gets raised on - all the important details are in the event args and no component is used to subscribe to them. Thus, I have simply changed the target entity to the mind entity itself and removed the check for OwnedEntity

## Requirements

## Breaking changes
RoleAddedEvent and RoleRemovedEvent messages are now raised on the mind entity, rather than the mind's owned entity. If any codebase is sending these events as non-broadcast, the events will now be sent to a different entity and probably no longer do what they were changed to do.

**Changelog**
:cl: Errant
- fix: Removing an emagged borg's brain now correctly reverts it to Silicon mind role.

### DIFF
--- a/Content.Shared/Roles/SharedRoleSystem.cs
+++ b/Content.Shared/Roles/SharedRoleSystem.cs
@@ -168,11 +168,9 @@ public abstract class SharedRoleSystem : EntitySystem
         var update = MindRolesUpdate((mindId, mind));
 
         // RoleType refresh, Role time tracking, Update Admin playerlist
-        if (mind.OwnedEntity != null)
-        {
-            var message = new RoleAddedEvent(mindId, mind, update, silent);
-            RaiseLocalEvent(mind.OwnedEntity.Value, message, true);
-        }
+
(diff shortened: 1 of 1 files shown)


ScarKy0 (approved):
o7

PJB3005 (approved):

Issue #28353, opened 2024-05-28
Some stuff in HumanoidCharacterProfile.cs should be CVARs


lzk228:
all, but `ICNameCaseRegex` (i suppose) should be defined in cvars for easy changing on other servers
https://github.com/space-wizards/space-station-14/blob/1777eea9a42a8bdbc4a865327bb9ef3e10fe57de/Content.Shared/Preferences/HumanoidCharacterProfile.cs#L28-L32

lzk228:
partially fixed by https://github.com/space-wizards/space-station-14/pull/35407

PR #35407, opened 2025-02-22
HumanoidCharacterProfile and IdCardConsoleComponent constants moved to cvar. Sync id card length with character name length

iaada:
## About the PR

PR moves constants in `HumanoidCharacterProfile` and `IdCardConsoleComponent` to cvar. Specifically character name lengths, id card name/job lengths, and flavor text (character description) length.

Additionally makes ID name length use the same value as character name length.

Advances #28353. I attempted moving the regex to cvar, but cvars do not support regex as a type.

## Why / Balance

Bugfix and more cvar. Id card was lengthened instead of loadout being shortened because there might be existing loadouts with 32 character names.

## Technical details

Calls to ccvar in several files.
LineEdit controls in `IdCardConsoleWindow.xaml.cs`, `HumanoidProfileEditor.xaml.cs`, and `LoadoutWindow.xaml.cs` responsible for entering appropriate values have had a line added to prevent entering more characters than can be accepted.

## Media

Before:
image

After:
image

## Requirements

## Breaking changes

Some hard coded constants for characters and IDs have been moved to CCVar. Specifically name length, loadout name length, description (flavor text) length, ID name length, and ID job title length. ID name length has been rolled into the same ccvar as name length.

`HumanoidCharacterProfile.MaxNameLength`, `.MaxLoadoutNameLength`, and `.MaxDescLength` replaced with CCvars `MaxNameLength`, `MaxLoadoutNameLength`, and `MaxFlavorTextLength`.
`IdCardConsoleComponent.MaxFullNameLength` and `.MaxJobTitleLength` replaced with CCvars `MaxNameLength` and `MaxIdJobLength`.

**Changelog**

:cl: aada
- fix: Id cards now have the same max length as character names.


beck-thompson:
Maybe this should be a ccvar? It feels weird that you have to just manually set them both to the same value :thinking:

lzk228:
> Maybe this should be a ccvar? It feels weird that you have to just manually set them both to the same value 🤔

https://github.com/space-wizards/space-station-14/issues/28353

PJB3005 (changes requested):
Do not copy paste this magic number, make it use a shared constant.

Content.Shared/Preferences/Loadouts/RoleLoadout.cs
```diff
@@ -79,9 +82,9 @@ public void EnsureValid(HumanoidCharacterProfile profile, ICommonSession session
         {
             var name = EntityName.Trim();
 
-            if (name.Length > HumanoidCharacterProfile.MaxNameLength)
+            if (name.Length > configManager.GetCVar(CCVars.MaxNameLength))
```
beck-thompson, line 85: Just save the variable so you don't call it twice! Do it like this pr does it: #35651

I only commented here but just do it on all the files

beck-thompson (changes requested):

beck-thompson (changes requested):
Looks good but add a breaking changes section to your PR explaining the new ccvars and stuff and that you removed the old fields.

beck-thompson (approved):
looks good to me!

iaada:
Genpop lockers broke it, I'll fix this weekend.

Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
```diff
@@ -131,6 +135,14 @@ public HumanoidProfileEditor(
             _requirements = requirements;
             _controller = UserInterfaceManager.GetUIController<LobbyUIController>();
             _sprite = _entManager.System<SpriteSystem>();
+
+            #region CCVar
```
PJB3005, line 139: This really does not need a `#region` block.

PJB3005 (approved):

Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
iaada: I added it because this file is dense and chock-full of regions already, but I can remove it.

PR #35408, opened 2025-02-22
Emp spell

beck-thompson:
## About the PR

Wizard needs more cool spells! Added a simple emp spell.

## Technical details

Code is a little weird but because of the upcoming refactor I think its fine!

## Media

https://github.com/user-attachments/assets/9c464f78-6bbd-4b9c-b96d-444f79c40b7f

## Requirements


Warpzoned:
I like it! But it doesn't feel as "grandiose" as some other spells I think? Maybe you could make it overcharge nearby lights/defibrillators to have them break/explode (Sorta like plasmafying defibs)? Usually for Wizards you'd expect their spells to do something more uniquely rather than just doing what everyday traitor items do (Not a diss btw! It's just my thoughts regarding what should be made of the most whimsical antagonist's abilities)

beck-thompson:
> I like it! But it doesn't feel as "grandiose" as some other spells I think? Maybe you could make it overcharge nearby lights/defibrillators to have them break/explode (Sorta like plasmafying defibs)? Usually for Wizards you'd expect their spells to do something more uniquely rather than just doing what everyday traitor items do (Not a diss btw! It's just my thoughts regarding what should be made of the most whimsical antagonist's abilities)

Yeah fully agree! The issue is I currently just do not have the time to make a full unique system in a nice way :(
Keron wanted this spell before release so I think its OK if its a little underbaked! Apparently the spell system is going to get a refactor soon TM. Maybe then I can dedicate a little more time to make it feel better!

Resources/Prototypes/Magic/emp_spell.yml
```diff
@@ -0,0 +1,16 @@
+- type: entity
+  id: ActionEmp
+  name: Magnetic blast
+  description: Disable nearby electronics.
+  components:
+  - type: EmpOnTrigger
```
slarticodefast, line 6: I'm not sure if I like reusing the trigger component to store the EMP settings here.
On the one hand it avoids code duplication nicely.
On the other hand it has nothing to do with triggers.

beck-thompson: > I'm not sure if I like reusing the trigger component to store the EMP settings here.
> On the one hand it avoids code duplication nicely.
> On the other hand it has nothing to do with triggers.

I tried really hard to do it the "right way" with triggers but its not possible sadly. Its either this or I make a new component :(

Keron said there will be a refactor that'll make it nicer so I didn't want to make a custom system if it will be replaced in the future

slarticodefast: Yeah, the trigger will cause the emp to happen at the location of the entity with the `EmpOnTrigger` component.
And I think this is in nullspace in this case since the action is in a container on the mind entity.

slarticodefast: At least leave a `TODO` comment

beck-thompson: > Yeah, the trigger will cause the emp to happen at the location of the entity with the EmpOnTrigger component.
And I think this is in nullspace in this case since the action is in a container on the mind entity.

Yep exactly! I tried to be tricky with it as well but it doesn't quite work out unfortunately :pensive: 

> At least leave a TODO comment

:saluting_face:

Resources/Textures/Objects/Magic/magicactions.rsi/meta.json
```diff
@@ -36,6 +36,9 @@
     },
     {
       "name": "repulse"
+    },
+    {
+        "name": "spell_emp"
```
ArtisticRoomba, line 41: Fix indentation

Resources/Prototypes/Magic/emp_spell.yml
```diff
@@ -0,0 +1,18 @@
+- type: entity
+  id: ActionEmp
+  name: Magnetic blast
+  description: Disable nearby electronics.
```
ArtisticRoomba, line 4:
```suggestion
  description: Disables nearby electronics.
```

Resources/Locale/en-US/store/spellbook-catalog.ftl
```diff
@@ -47,6 +47,9 @@ spellbook-slip-desc = Learn the ancient ways of the Janitor and curse your targe
 spellbook-item-recall-name = Item Recall
 spellbook-item-recall-description = Mark a held item and summon it back at any time with just a snap of your fingers!
 
+spellbook-item-emp-name = Magnetic blast
+spellbook-item-emp-description = Disables nearby electronic equipment!
```
ArtisticRoomba, line 51: I guess if you're localizing the grimore stuff you might as well localize the YAML/actions stuff.

slarticodefast: actions are entities and those get an automatically generated loc string forks can overwrite

Content.Server/Emp/EmpSystem.cs
```diff
@@ -108,6 +110,15 @@ private void HandleEmpTrigger(EntityUid uid, EmpOnTriggerComponent comp, Trigger
         args.Handled = true;
     }
 
+    private void HandleEmpSpellAction(Entity<EmpOnTriggerComponent> ent, ref EMPSpellActionEvent args)
```
metalgearsloth, line 113: Can we not just make a trigger event instead? Having to manually do 100 events for these is not good

beck-thompson: I tried that but I could not get it to work sadly :(

The issue is that there is technically nothing to trigger it on. For stuff like implants it triggers on the implant itself, but for spells the *only* thing it could trigger on is like the action which doesn't work because its in nullspace.

I could be just thinking about it wrong though!

slarticodefast: The problem is that the EmpComponent is on the mind entity, not on the wizard's body (without that it would not get transferred with mind swap). So if you trigger it would activate in nullspace.

TokenStyle: Yep, it is. Mind's don't touch body's triggers and vice versa [https://github.com/space-wizards/space-station-14/blob/master/Content.Server/Explosion/EntitySystems/TriggerSystem.cs](https://github.com/space-wizards/space-station-14/blob/master/Content.Server/Explosion/EntitySystems/TriggerSystem.cs)

Issue #22669, opened 2023-12-17
Hardcoded locale "and" when player holding two or more items


lzk228:
tried to find out where its written, but cant

![image](https://cdn.discordapp.com/attachments/921760547207266314/1175197045972283432/image.png?ex=656a5a89&is=6557e589&hm=92ab8281dc4dc62a2bed58aefd4f0b6303290a20a7373420f8dfc3125d57e060&)

EmoGarbage404:
Its using ContentLocalization.FormatList iirc

lzk228:
> ContentLocalization.FormatList

cant find it both in ss14 and engine

EmoGarbage404:
https://github.com/space-wizards/space-station-14/blob/master/Content.Shared%2FHands%2FEntitySystems%2FSharedHandsSystem.Interactions.cs#L195

ContentLocalizationManager.FormatList

lzk228:
so i can just change "and" to locale string?

lzk228:
https://github.com/space-wizards/space-station-14/blob/118cb5dc7a4c1b6559de59470b7ce15376e8ec52/Content.Shared/Localizations/ContentLocalizationManager.cs#L111-L120
here

mirrorcult:
the entire thing should just be a loc string, but if you cant get that to work then changing `and` is probably good enough

PR #35415, opened 2025-02-23
dehardcode FormatList

lzk228:
fixes https://github.com/space-wizards/space-station-14/issues/22669

Screenshot_20250223_030704
Screenshot_20250223_030711
Screenshot_20250223_030720
Screenshot_20250223_030725

### DIFF
--- a/Content.Shared/Localizations/ContentLocalizationManager.cs
+++ b/Content.Shared/Localizations/ContentLocalizationManager.cs
@@ -105,33 +105,31 @@ namespace Content.Shared.Localizations
             }
         }
 
-        // TODO: allow fluent to take in lists of strings so this can be a format function like it should be.
         /// <summary>
         /// Formats a list as per english grammar rules.
         /// </summary>
         public static string FormatList(List<string> list)
         {
-            return list.Count switch
-            {
-                <= 0 => string.Empty,
-                1 => list[0],
-                2 => $"{list[0]} and {list[1]}",
-                _ => $"{string.Join(", ", list.GetRange(0, list.Count - 1))}, and {list[^1]}"
-            };
+            if (list.Count == 0) return string.Empty;
+
+            return Loc.GetString("zzz-list-format",
+                                 ("count", list.Count),
+                                 ("item1", list.ElementAtOrDefault(0) ?? ""),
+                                 ("item2", list.ElementAtOrDefault(1) ?? ""),
+                                 ("items", string.Join(", ", list.Take(list.Count - 1))),
+                                 ("last", list.LastOrDefault() ?? ""));
         }
 
-        /// <summary>
-        /// Formats a list as per english grammar rules, but uses or instead of and.
-        /// </summary>
         public static string FormatListToOr(List<string> list)
         {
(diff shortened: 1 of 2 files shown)


Resources/Locale/en-US/_lib.ftl
```diff
@@ -33,4 +33,18 @@ zzzz-fmt-power-joules = { TOSTRING($divided, "F1") } { $places ->
 }
 
 # Used internally by the PLAYTIME() function.
-zzzz-fmt-playtime = {$hours}H {$minutes}M
\ No newline at end of file
+zzzz-fmt-playtime = {$hours}H {$minutes}M
+
```
slarticodefast, line 37:
```suggestion

# Used internally by the FormatList function.
```

Resources/Locale/en-US/_lib.ftl
```diff
@@ -33,4 +33,18 @@ zzzz-fmt-power-joules = { TOSTRING($divided, "F1") } { $places ->
...
+zzz-list-format =
```
slarticodefast, line 38:
```suggestion
zzzz-fmt-list=
```
keep the naming scheme consistent
no idea what the zzzz is for though
same for the other one below

slarticodefast (commented):
The delimiter should probably be localized as well.

Content.Shared/Localizations/ContentLocalizationManager.cs
```diff
@@ -107,33 +107,31 @@ private ILocValue FormatMakePlural(LocArgs args)
             }
         }
 
-        // TODO: allow fluent to take in lists of strings so this can be a format function like it should be.
```
Tayrtahn: Keep this todo comment. This refers to modifying the Fluent implementation so you can directly pass in a list as a variable. Like:
```cs
List<EntityUid> myList = [];
var message = Loc.GetString(locId, ("items", myList));
```
And then handle the formatting entirely in the FTL files.

Fildrance:
Usage of LINQ here is highly questionable.. its not problematic perf-wise in this case - as there are List-related optimizations, but that creates opportunity for copy-paste lovers to use same methods in cases where it would actively hurt performance.
Changes are band-aid-looking - .ftl part looks very wordy and quite hard to read too, easy to mess up. Attempting to hide this behind custom localization function is much-much more preferrable - maybe you could look at it? Fluent itself does not work with collection-typed arguments but possibly custom function injected into linguini would? It would be very nice to have something like https://cldr.unicode.org/development/development-process/design-proposals/list-formatting but thats an ideal scenario.

I am not sure it is good to merge unless we tried all other solutions.

lzk228:
@Fildrance well i know there is a similar thing
https://github.com/space-wizards/space-station-14/blob/e2d96f1f4921d229ef92a3b6dc7850ce720b9748/Content.Shared/Nutrition/EntitySystems/FlavorProfileSystem.cs#L70-L75

and i don't know any other solutions

Fildrance:
> @Fildrance well i know there is a similar thing

Use index of element and Index-from-end Operator

PR #35418, opened 2025-02-23
Make holoparasite's damage transfer ignore the host's armor

NoElkaTheGod:
## About the PR
Hitting a holoparasite will now apply the full* damage to the host, ignoring whatever armor host may be wearing.
*Holoparasites still have an innate 35% resist to all damage.

## Why / Balance
- It logicaly makes sense (holoparasite directly transfers any damage to the host's body, why would it care about armor?)
- Combining a holoparasite with a good armor (like a jugsuit) makes the host way too strong.
- It already has 35% resistance, very good melee damage (36 blunt and 36 structural per second), ignores bullets and can act semi-independently from it's host, which is pretty damn good for 14 TC.

## Technical details
изображение

## Media
изображение
A person in bloodred has taken 65 damage when their holoparasite was hit with a 100 damage stick due to the holoparasite's 35% resist.

## Requirements

**Changelog**
:cl:
- tweak: Hitting a holoparasite will now ignore the host's armor.

### DIFF
--- a/Content.Server/Guardian/GuardianSystem.cs
+++ b/Content.Server/Guardian/GuardianSystem.cs
@@ -269,6 +269,7 @@ namespace Content.Server.Guardian
                 component.Host,
                 args.DamageDelta * component.DamageShare,
                 origin: args.Origin,
+                ignoreResistances: true,
                 interruptsDoAfters: false);
             _popupSystem.PopupEntity(Loc.GetString("guardian-entity-taking-damage"), component.Host.Value, component.Host.Value);
 


PJB3005 (approved):
Seems reasonable to me but I don't know game balance so.

Pronana:
I don't really like this change.

This is a significant nerf to holoparasites for traitors to try and make them potentially balanced on nukeops. 

It doesn't feel right, and even with the change holos would still be really good on nukeops, as they could break walls, peek corners, and protect your crit body from being looted and possibly revenge kill really easily. 

A genuine downside to Holos as well is that they die when you die and can't be revived, but that's not a downside on Nukeops since you won't be getting revived either.

NoElkaTheGod:
> I don't really like this change.
> 
> This is a significant nerf to holoparasites for traitors to try and make them potentially balanced on nukeops.

It's really hard for a traitor to get good armor **and** a holoparasite, as the latter is fairly expensive considering their budget, and even if they do somehow get both, that's not really useful for them, as traitors only have 1 or 2 kill objectives.

Also, right now a holoparasite with good armor is so immensely powerful, that it hoesntly needs a nerf even if the other PR gets rejected. A holoparasite with a jugsuit literaly makes you an (almost) unstoppable walking gib machine.

> It doesn't feel right, and even with the change holos would still be really good on nukeops, as they could break walls, peek corners, and protect your crit body from being looted and possibly revenge kill really easily.

While they do provide good utility, they can easily become a nukie's weak spot, because holoparasite only has 35% resist, while nukie has at least 50% from a blood-red hardsuit. Not to mention that there is a bajillion ways of breaking walls, and you have other nukies who can protect your fallen body even better than a holoparasite.

> A genuine downside to Holos as well is that they die when you die and can't be revived, but that's not a downside on Nukeops since you won't be getting revived either.

This is true, but is out of scope of this PR.

slarticodefast:
> It's really hard for a traitor to get good armor and a holoparasite, as the latter is fairly expensive considering their budget, and even if they do somehow get both, that's not really useful for them, as traitors only have 1 or 2 kill objectives.

>Also, right now a holoparasite with good armor is so immensely powerful, that it hoesntly needs a nerf even if the other PR gets rejected. A holoparasite with a jugsuit literaly makes you an (almost) unstoppable walking gib machine.

Yeah, I agree with this, usually you don't have both, at least when I saw it used by traitors. It also does not make any sense from a logical standpoint that armor should protect you from the damage transfer.

slarticodefast (approved):

pheenty:
> PR is already merged but I'll mention traitors can purchase the web vest for 3 tc, which provides 70% pierce resist and 40% blunt/slash resist with no speed penalty. This would've stacked with the 35% damage reduction from holoparasites.

Holoparasites do not take pierce damage anyway (bullets pass through them), unless you hit them with a spear of course, but how often do you do that

pheenty:
And while this change makes sense, holoparasites were already laughably weak. I'd suggest to buff their damage resistance up to 50% and to make them immune to explosions. And maybe buff their attack speed a bit too.

NoElkaTheGod:
> And while this change makes sense, holoparasites were already laughably weak. I'd suggest to buff their damage resistance up to 50% and to make them immune to explosions. And maybe buff their attack speed a bit too.

While i do agree, this is a balance change, and thus would require 4-8 years of discussion which i do not want to participate in.

pheenty:
> > And while this change makes sense, holoparasites were already laughably weak. I'd suggest to buff their damage resistance up to 50% and to make them immune to explosions. And maybe buff their attack speed a bit too.
> 
> While i do agree, this is a balance change, and thus would require 4-8 years of discussion which i do not want to participate in.

fair enough

PR #35439, opened 2025-02-24
Fix arrivals

metalgearsloth:
Fixes arirals.

:cl:
- fix: Fix arrivals.


lzk228:
what exactly does it fix

ScarKy0:
> what exactly does it fix

Arrivals Terminal does not spawn at all on master

Issue #35270, opened 2025-02-18
Strange plant draining bug


AjexRose:
## Description

You can drag-click a plant into the chem-master to instantly drain reagents from the plant without having to use the reagent grinder
**Reproduction**

Get poppies
Get a chem master
click and drag poppies onto the chem master
poppy reagent appears in the chemmaster
Grind poppy
Get nothing.
**Screenshots**

**Additional context**

Quantum-cross:
Works with drains too, they kind of act like buckets...

![Image](https://github.com/user-attachments/assets/8bad49fd-3079-4383-8c07-e1aab8141ec6)

![Image](https://github.com/user-attachments/assets/bff6497c-cb23-43f8-8b8a-439fc5f1d4cd)

Eating the result:

![Image](https://github.com/user-attachments/assets/452735f2-8add-48fe-b77f-b31918b980a4)

Quantum-cross:
Tracked it down to this:

https://github.com/space-wizards/space-station-14/blob/f3c681153ac0d4c351dccd125846bb512d79f074/Resources/Prototypes/Entities/Objects/Consumable/Food/food_base.yml#L39-L40

RefillableSolutionComponent --

```Reagents that can be added easily. For example like
pouring something into another beaker, glass, or into the gas
tank of a car.```

NAAAAAAH

I'd remove this myself but I don't know the ramifications of it, perhaps there should be a second solution in addition to "food" for "injected_reagents".

CooperWallace:
@Quantum-cross Yeah it seems to be on a lot of food bases. I was experimenting with it yesterday, thanks for pointing in the right direction.

Quantum-cross:
### Proposal

- Add a `condiments` solution to `FoodBase` with volume of 10
- Have RefillableSolution target `condiments` instead of `food`

The condiments can still be emptied which is a little silly but it will work. This still also allows for fun antag stuff because you can totally put 10 units of amatoxin on food...

### Bonus points

Add `FillableSolutionComponent` that doesn't allow draining

if you accidentally dump a whole bunch of salt in your soup, it ain't coming out...

CooperWallace:
@Quantum-cross Most of the meals have extra room for solutions in them already. Interestingly enough, Condiments aren't allowed to be dragged. So I feel like removing the RefillableSolution component from produce would work, (and probably others too) and only add it when specifically needed.

Quantum-cross:
[...]

PR #34683, opened 2025-01-27
Small OnRefillableDragged bugfix

insoPL:
## About the PR
OnRefillableDragged - drag and drop system for solution containers
now checks if container being droped is open and Drainable
Drainable component is defined as:
"Denotes the solution that can be easily removed through any reagent container.
Think pouring this or draining from a water tank."
It makes check for that when drag and dropping container.

You could make closed but empty soda cans using drains which is also goofy.

## Why / Balance
- small bugfix of #33115

## Technical details
OnRefillableDragged checks for DrainableComponent and checks if it is open if openable

## Media

https://studio.youtube.com/video/waQ_NSvvgt4

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

PR #35445, opened 2025-02-24
Fix Refillable solutions being drained via drag and drop

CooperWallace:
## About the PR

Fixes https://github.com/space-wizards/space-station-14/issues/35270

## Why / Balance

The extraction of the solutions within produce without grinding. Makes the use of Industrial Grinders unneeded.

## Technical details

Changed the Drag-and-drop event for Fluid Tranfers to be initialized on `DrainableSolution` instead of `RefillableSolution`.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Produce and others can no longer have their solutions extracted by drag-and-drop.


RedBookcase:
Said this on the discord as well, but I think this bug should remain given the relative lack of easy alternatives for mass plant processing. 

The regular grinder is very slow to use of course, which makes sense given it's pretty basic.

The industrial reagent grinder requires a bunch of extra setup. It only has 400u of internal storage. It's not possible to see it's current capacity without chem goggles. If you put more than 400u into the grinder, it starts dumping liquids out on the ground, which means you can't really dump out a plant bag for processing without worrying about spillage. There's no way to separate liquids from the grinder. It's an improvement over the basic blender but it's still very tedious to use.

The drag and drop bug is a bug, but it's also the least tedious way of processing plants. Reagents are immediately dropped in the chemmaster without any risk of unintended mixtures. You don't have to worry about overfilling the container and causing a spill. There's no need to constantly pause between filling and emptying containers and putting the plants into the machine. Botanists can either drag their produce into the Chemist's chemmaster, or get Science to build one for them and bring pure reagents to chem by the bucket easily. It reduces so much hassle.

I would really prefer to see the Industrial Reagent Grinder reworked to have a larger capacity and the ability to separate chems - or even just allowing us to dump plants directly into the Chemmaster - instead of having this loophole fixed with no other changes. In my eyes this change would only add more tedium and repetitive clicking instead of alleviating it. 

(Also - would this remove the ability to drag and drop regular fluid containers into the chemmaster? If so - that would only make the tedium problem much worse.)

NoElkaTheGod:
I do think it would be really nice if there was a device that automaticaly transfers any reagents from an industrial grinder into a chemmaster, but not fixing a bug because of this is dumb.

Quantum-cross:
Does this affect dragging mop buckets and trolleys into drains? Probably not because they have the Drainable component.

ArtisticRoomba:
Me when I hold my plants up the the chem master screen and magically transfer the plants contents, and then I juice them for the same contents and put it in the chemmaster

If the issue is with the industrial reagent grinder not having enough capacity, then increase the capacity of it to handle more juices. That is a very easy PR. But I don't think we should justify this bug (that has been in the game for a long time!) existing because "doing things the inefficient regular way is painful". When it usually isn't if you think smart.

insoPL:
this is duplicate of my PR #34683

MilonPL:
Closing as duplicate.

PR #35429, opened 2025-02-23
Cargo Mail System

ScarKy0:
## About the PR

Added a delivery system for cargo.
Cargo now has access to a mail teleporter which will periodically spawn mail addressed to a random crew member(based on crew manifest).
Delivering said mail and having it be opened grants cargo some bonus money and gives the recipient a little reward (or spam)

## Why / Balance

Currently cargo's main way of earning money is the bounty system and selling random objects around the station. This can lead to situations such as other departments not cooperating or some bounties simply being not possible to make within a reasonable timeframe, as well as getting repetitive after a while.
While those issues can be fixed by making changes to the bounty system (like lowering the bounty skip cooldown) that feels like a bandaid fix to the fact cargo has basically only one way to earn money.

Delivering mail is meant to serve as a secondary option of income cargo can take care of during their shift.

## Technical details

Added a new system for Deliveries. It handles being assigned to a random crewmember and allowing them to open it by fingerprint.
It also handles spawning deliveries on DeliverySpawners. By default the amount of mail spawned is 1 per 7 members of crew, which then gets distributed randomly across all spawners on station.

## Media

image

https://github.com/user-attachments/assets/52d6ab27-d845-4096-b652-2751c640c802

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

SpaceGarbageComponent has been moved from Content.Server.Shuttles.Components to Content.Shared.Shuttles.Components.

**Changelog**

:cl: ScarKy0, MilonPL
- add: Added deliveries! Cargo now has access to mail teleporters, which periodically spawn mail. Deliver it to the correct recipients for bonus spesos!

PR #35450, opened 2025-02-24
GettingUsedAttemptEvent

ScarKy0:
## About the PR

Added a new event raised on the item when it is being used.
Taken from #35429 for atomization purposes.

## Technical details

Added a new event.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/ActionBlocker/ActionBlockerSystem.cs
+++ b/Content.Shared/ActionBlocker/ActionBlockerSystem.cs
@@ -109,10 +109,16 @@ namespace Content.Shared.ActionBlocker
         /// </remarks>
         public bool CanUseHeldEntity(EntityUid user, EntityUid used)
         {
-            var ev = new UseAttemptEvent(user, used);
-            RaiseLocalEvent(user, ev);
+            var useEv = new UseAttemptEvent(user, used);
+            RaiseLocalEvent(user, useEv);
 
-            return !ev.Cancelled;
+            if (useEv.Cancelled)
+                return false;
+
+            var usedEv = new GettingUsedAttemptEvent(user);
+            RaiseLocalEvent(used, usedEv);
+
+            return !usedEv.Cancelled;
         }
 
 
--- a/Content.Shared/Interaction/Events/GettingUsedAttemptEvent.cs
+++ b/Content.Shared/Interaction/Events/GettingUsedAttemptEvent.cs
@@ -1,0 +1,10 @@
+﻿namespace Content.Shared.Interaction.Events
+{
+    /// <summary>
+    /// Version of UseAttemptEvent raised on the used item.
+    /// </summary>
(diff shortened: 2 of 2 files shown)


Content.Shared/Interaction/Events/GettingUsedAttemptEvent.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Content.Shared.Interaction.Events
```
slarticodefast, line 1: use file scoped namespace

slarticodefast (changes requested):

Content.Shared/Interaction/Events/GettingUsedAttemptEvent.cs
```diff
@@ -0,0 +1,10 @@
...
+{
+    /// <summary>
+    /// Version of UseAttemptEvent raised on the used item.
+    /// </summary>
```
FairlySadPanda, line 5: This should state what this event does, not how it relates to some other event.

If you need to reference another class, struct, etc inside a summary block, you should make sure it's linked via a tag (which I am not fresh on for syntax but you should be able to find in the project in many places)

Content.Shared/ActionBlocker/ActionBlockerSystem.cs
```diff
@@ -109,10 +109,16 @@ public bool CanInteract(EntityUid user, EntityUid? target)
...
-            RaiseLocalEvent(user, ev);
+            var useEv = new UseAttemptEvent(user, used);
+            RaiseLocalEvent(user, useEv);
 
-            return !ev.Cancelled;
+            if (useEv.Cancelled)
+                return false;
+
+            var usedEv = new GettingUsedAttemptEvent(user);
+            RaiseLocalEvent(used, usedEv);
```
FairlySadPanda, line 119: I do think you may need two events for this, but this is confusing.

You have two options, I think?
1. Drop the targeted raise from UseAttemptEvent. We already know who the User is in UseAttemptEvent anyway.
2. Pass both user and used into GettingUsedAttemptEvent, rename it to be understandable, and fix the docstring on both the new event and this sending function.

Content.Shared/Interaction/Events/GettingUsedAttemptEvent.cs
Tayrtahn: `<see cref="My Class" />`

Content.Shared/Interaction/Events/GettingUsedAttemptEvent.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Content.Shared.Interaction.Events;
+
+/// <summary>
+/// Cancellable event raised on used item. Cancelling it cancels the interaction.
```
slarticodefast, line 4:
```suggestion
/// Event raised on an item when attempting to use it in your hands. Cancelling it stops the interaction.
```

PR #35454, opened 2025-02-24
Cleanup: Fix field naming rule violation in ``GhostComponent``

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Naming rule violation. #33279.

## Requirements

## Breaking changes
Field ``color`` of the ``GhostComponent`` class was renamed to ``Color``.

**Changelog**
Nope.

### DIFF
--- a/Content.Shared/Ghost/GhostComponent.cs
+++ b/Content.Shared/Ghost/GhostComponent.cs
@@ -87,7 +87,7 @@ public sealed partial class GhostComponent : Component
     /// </summary>
     /// <remarks>Used to allow admins to change ghost colors. Should be removed if the capability to edit existing sprite colors is ever added back.</remarks>
     [DataField("color"), ViewVariables(VVAccess.ReadWrite), AutoNetworkedField]
-    public Color color = Color.White;
+    public Color Color = Color.White;
 
     [DataField("canReturnToBody"), AutoNetworkedField]
     private bool _canReturnToBody;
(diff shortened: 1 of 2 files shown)


Content.Shared/Ghost/GhostComponent.cs
```diff
@@ -87,7 +87,7 @@ public bool CanReturnToBody
     /// </summary>
     /// <remarks>Used to allow admins to change ghost colors. Should be removed if the capability to edit existing sprite colors is ever added back.</remarks>
     [DataField("color"), ViewVariables(VVAccess.ReadWrite), AutoNetworkedField]
```
slarticodefast, line 89:
```suggestion
    [DataField, AutoNetworkedField]
```

slarticodefast (approved):
Code looks good, but not sure if this will break some admin scripts.

Winkarst-cpu:
> Code looks good, but not sure if this will break some admin scripts.

This field is only used once in upstream.

slarticodefast:
> This field is only used once in upstream.

Afaik a lot of admins have custom scripts they run when going online, changing color, speed etc.
Those might break, so we should at least warn them.

Winkarst-cpu:
> Afaik a lot of admins have custom scripts they run when going online, changing color, speed etc. Those might break, so we should at least warn them.

Just ping them on Discord ig. Also there is a breaking change.

Simyon264 (approved):

Issue #30259, opened 2024-07-22
Some cooking meal recipes containing RawEgg reagent results both the cooked dish and CookedEgg reagent in beaker


SamGithubAccount:
## Description

When cooking recipes in the microwave that use egg liquid, `Egg`, 
**Reproduction**

Cook any recipe requiring raw egg, and check the beaker after it is done cooking
**Screenshots**

**Additional context**

I was testing this on debug client for a PR, but did not make any changes that would affect this. I might make a PR to remove the cookedEgg recipe as it has no use currently, that could be re-added if recipe prioritization or something similar gets added as part of a cooking refactor/rework.

Katzenminer:
i Have fixed  the Egg Issue on my branch im just Busy Testing every recepie for the pr so i will mention this issue in my PR wehn i make it.

PR #30294, opened 2024-07-23
Microvawe  Recepies  and Egg interaction´s Fix

Katzenminer:
## About the PR
I  have  changed almost all Recepies That used only  Solid Eggs as in `FoodEgg` to now accept the Reagent `Egg`, making it possible to use either whole eggs,  or raw egg  in a  container. I have also the addresed the issue that Came with it that Whole Eggs in the brownie Recepie always  survive the cooking  effectivly cloning them.
i Have now also added  that Should a  Food Type container(Container  with the FoodComponent) be used up they will spawn  there junk version. 

## Why / Balance
It is very inconvinent for chef´s wehn cooking to be forced to use a  whole  a egg,  and also it makes no sense to  force  them to in many recepies. 
fixes https://github.com/space-wizards/space-station-14/issues/30259

## Technical details
I Have Added Two  Foreach loops to the  `SubtractContents `Methode inside MircowaveSystems.cs. The first  foreach loop check for Containers  that have the type Food and if they are empty, If that is the case they get added to a Stack Called `entsToRemove `. 
The  second foreach loop removes  each item in the Stack.
I have also removed the construction graph and node aswell as the reaction wich created boiled egg from raw egg. This was causing the most issues in terms of causing duplication  as 2 reactions happend at once

## Media
This media  below  is me  going through every  recepie making sure it doesnt duplicate the eggs.

- [X ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

https://github.com/user-attachments/assets/7b9fec00-047e-4945-b678-5a0a3f177159

**Changelog**
:cl:
- add: Added a Recepie for Boiled Egg.
- remove: Removed Eggs being Cooked as A Chemical Reaction by Heat.
- fix: Fixed Eggs Not  disapearing after Microwaving with them sometimes.

PR #30262, opened 2024-07-22
Four cooking recipes changed to use liquid egg instead of whole egg, and 5u of milk added to cake batter

SamGithubAccount:
## About the PR

Changed the recipe for brownies, cannabis brownies/special brownies, chow mein, and egg-fried rice to to require liquid egg instead of a whole egg. 5u of milk were added to the CreateCakeBatter reaction.
## Why / Balance

To better fit with other recipes, such as pancakes, nettle soup, and noodles. Solves issue #30123, as the other three recipes mentioned seem to make sense with a whole egg, either fried or boiled.
Cake batter was created from the required reagents in beaker; probably why the recipe was what it was previously.  To fix this, 5u of milk were added to its recipe, which as far as I know shouldn't affect chefs too much.
## Technical details

Added 6 (raw) `Egg` reagent per previously used whole egg and removed previous FoodEgg requirements in the four previously mention recipes.
Added 5u of milk to cake batter recipe to prevent it occurring in beaker before brownies can be made
## Media

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl: TheKittehJesus
- tweak: The recipe for chow mein, egg-fried rice, and both brownies now use liquid egg instead of a whole egg. 
- tweak: Cake batter now also requires 5u of milk

PR #35459, opened 2025-02-24
Fix egg cooking and make microwave code a little less bad

slarticodefast:
## About the PR
You no longer get both cooked egg and the recipe result when cooking any recipe that uses the egg reagent.
Fixes https://github.com/space-wizards/space-station-14/issues/30259
The microwave can no longer use reagents while they are inside a food item, you need to crack eggs into a beaker first.

Replaces https://github.com/space-wizards/space-station-14/pull/30294
Thanks to @Katzenminer for the initial investigation into the bug.

## Why / Balance
bugfix

## Technical details
The microwave decides which recipe it will spawn when it is activated, however the reagents used for this are only removed at the end of the cooking time. This results in any reagents that are transformed during cooking to be duplicated. For example when cooking a pancake (5 milk, 5 flour, 5 raw egg) the egg turns into cooked egg from being heated up.
We solve this by stopping the reaction when the reagent has been reserved for a recipe. Other reactions still happen normally.

Some recipes have been changed to use liquid eggs instead of whole eggs. These were either overlooked in https://github.com/space-wizards/space-station-14/pull/30262 or added later.

Microwave code has been changed to no longer use *any* solution for the recipes, but only drainable (i.e. accessible) ones. Before you could put whole eggs inside and the reagents would magically be used without damaging the shell. Now you need to crack the eggs first if you want to use them in a recipe.
Similarly other food items could be drained of their reagents before.

Previously eggs had an construction graph replacing them with a boiled egg at a certain temperature. However, this deleted any other reagents inside. We simply removed this contruction graph step, since the egg boiling is already handled by the reagent inside reacting from the heat.

Eggs have another construction graph step, which is supposed to cause them to explode when heated up too much, but it did not trigger correctly because the values in the `InternalTemperatureComponent` were off, causing the internal temperature to only raise by a few degrees when microwaved.
Essentially eggs had 3 different temperatures before: Internal, external, and the reagent temperature inside.
Since the reagent temperature already handles the cooking correctly now, we removed the redundant `InternalTemperatureComponent`. This causes the construction graph to use the external temperature instead and we adjusted the threshold accordingly.
Eggs will now cook when microwaved for 5 seconds and explode when microwaved longer (or they get heated up by atmos).

## Media

https://github.com/user-attachments/assets/5a278ff6-3c71-4dd4-b0e9-41e4b269b395

## Requirements

## Breaking changes

**Changelog**
:cl: Katzenminer, slarticodefast
- tweak: You now need to break eggs when using them for microwave recipes. Some recipes have been adjusted accordingly.
- fix: Fixed a reagent duplication bug when microwaving eggs.

### DIFF
--- a/Content.Server/Kitchen/Components/ActivelyMicrowavedComponent.cs
+++ b/Content.Server/Kitchen/Components/ActivelyMicrowavedComponent.cs
@@ -1,5 +1,3 @@
-using Content.Shared.Kitchen;
-
 namespace Content.Server.Kitchen.Components;
 
 /// <summary>
@@ -8,4 +6,9 @@ namespace Content.Server.Kitchen.Components;
 [RegisterComponent]
 public sealed partial class ActivelyMicrowavedComponent : Component
 {
+    /// <summary>
+    /// The microwave this entity is actively being microwaved by.
+    /// </summary>
+    [DataField]
+    public EntityUid? Microwave;
 }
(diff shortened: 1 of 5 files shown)


metalgearsloth:
this code scares me

PR #35461, opened 2025-02-24
Make the C-20R no longer wieldable, and have full accuracy one handed.

BramvanZijp:
## About the PR
The C-20R can no longer be wielded, and its accuracy when firing one handed is now the same as what it used to be when wielded.

## Why / Balance
The only reason to use the C-20r as a nukie (not loneop) right now is because you want to budget with the one thats mapped on the table outside for free and combine that with an ammo bundle or to save inventory space, which while the first scenario is fairly common, let me show you why I think this weapon needs some love, and also explain why the above is even true at all, because if as a nukie (excluding loneop) you are BUYING a C-20r bundle, you are truly making a mistake. 

You spend 17 TC on the C-20r.
You get a weapon that has mediocre accuracy when wielded, and horrible accuracy when not wielded, with a firerate of 8 per second, 16 pierce a hit, resulting in 128 DPS, 30 rounds per mag, and a total of 91 rounds in the bundle. You can also buy additional 30 round mags for 2 TC a mag (1tc = 15 rounds), and an ammo bundle comes with another 120 rounds.

However, when I say the above out loud, another weapon comes to mind with mediocre accuracy when wielded, horribly accuracy when not wielded, and a firerate of 8 per second, and while it does cost 30 TC (13 more expensive), heres some improvements that the L6 SAW offers over the C-20r that more then makes up for the steeper price point:
You get 19 pierce a hit (+3), and as you maintain 8 rounds per second, the DPS climbs to 152 (+24), you also get 100 rounds per mag (+70) and a total of 201 rounds in the bundle (+110), and it gets better, as you can buy an ammo box with 60 rounds for 1 TC per box (1 TC = 60 rounds, thats +45 or 4x.) with the only downside being it takes a bit to load em into an empty mag. You also get 200 rounds in the ammo bundle (+80), already in mags.

Now seeing all of those upgrades, the only reason to use the C-20 as a nukie is either as a loneop who would prefer not to blow half of their budget on just the primary weapon, or to grab the one on the table of the nukie planet and work with an ammo bundle, but other then that there is not really any excuse for this beyond horrible inventory management and an occupied back slot, which is just a skill issue atp.

I feel like making the weapon work just fine with one hand opens up a LOT of possibilities for it to actually give it a lot of unique use cases, such as with an energy shield, energy sword, easier reloads, etc. While also giving it its unique position as a one handed weapon back, since the L6 was not changed with the wield update despite pretty much every long-arm except for said L6 getting nerfed with that update. I also don't like how its current status is just "A worse version of the l6 that costs less".

Also, another prime reason is to match the WT550, which I have always seen as the crew variant of the C-20r.
While the C20 has a higher firerate and the 550 is more accurate, there are also plenty of simularities. They each have 30 round mags, they each fire .35 caliber rounds, both nukies AND security can get one for free (HoS Room & Nukie planet gun range) with the ability to buy more if needed (Cargo & Uplink), they are both a 2x4 in your inventory, etc. 
Yet when the change was made that resulted in most weapons having to be wielded in order to be fired accurately, the C-20r got this change while the WT550 did not, despite the fact that IMO the C20 looks like it'd be somewhat easier to use one handed compared to the 550, and since the one handed ability is now a large part of what makes the WT550 so unique, meaning removing that would just be lame, I am buffing the C-20r to work one handed instead.

TLDR: This makes it match the WT550 and gives it an upside over the L6 saw beyond simply being cheaper.

## Technical details
Rest in peace 7 lines of code, you may or may not be missed.

## Media
Not really needed, just shoot the C-20 while its wielded, and now imagine it doing that without being wielded, yippee.

## Requirements

**Changelog**
:cl: BramvanZijp
- tweak: The syndicate C-20r submachine gun has been updated to have its standard accuracy while firing it with only one hand.
- remove: The syndicate C-20r submachine gun can no longer be wielded with 2 hands.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/SMGs/smgs.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/SMGs/smgs.yml
@@ -96,14 +96,7 @@
       map: ["enum.GunVisualLayers.Mag"]
   - type: Clothing
     sprite: Objects/Weapons/Guns/SMGs/c20r.rsi
-  - type: Wieldable
-    unwieldOnUse: false
-  - type: GunWieldBonus
-    minAngle: -19
-    maxAngle: -16
   - type: Gun
-    minAngle: 21
-    maxAngle: 32
     shotsPerBurst: 5
     availableModes:
     - Burst


BramvanZijp:
> I think the bigger concern for me is traitor balance and balance around the bulldog. The bulldog is very similar to the C-20 in terms of weapon DPS, damage output, and price.
> 
> C-20: .35 Auto (16 damage) 8 firerate = 128 DPS. 480 damage per mag. 1440 damage per bundle.
> 
> Bulldog: .50 shell (6 pellets at 10 damage each 60 damage) 2 firerate = 120 DPS. 480 damage per mag. 1440 damage per bundle.
> 
> They’re basically the same gun just dealer’s choice as to if you prefer a smg or shotgun if you exclude the fact that the bulldog costs more TC and both traitors and nukies buy the bulldog less than the c-20.
> 
> The C-20 also has the benefit of being able to take magazines from the Drozd which is a decently common crew weapon which does matter and makes its ammo economy a bit better sometimes.
> 
> That being said, it being 1 handed would be interesting, it makes it more comparable to the WT-550 rather than the Drozd in terms of traitor weapon choice and I don't think that's necessarily a bad thing. It just makes the Bulldog's price even less justified.
> 
> Lastly in terms of balance compared to L6.
> 
> L6: .30 rifle (19 damage) 8 firerate = 152 DPS. 1900 Damage per mag. 3800 Damage per bundle.
> 
> Considering its cost the L6 should be better, but ammo economy matters a lot of Nukies and at the moment the best ammo economy weapon is the L6. Making the C-20 a one handed weapon means you could pair it with an e-shield and also gives agent a one handed gun they could use to defend themselves with.
> 
> I think it's definitely an interesting idea and I don't think it would have that bad of an impact on traitors since they can't buy e-shields. But something would also need to happen to the bulldog. I'm not sure what, because the bulldog sits in a nice place in terms of balance compared to the Enforcer, both have the same DPS but the bulldog had better uptime. Maybe the bulldog needs to be cheaper, maybe it needs more mags per bundle.

I feel like giving the bulldog an extra 0.5-1 firerate would be justified given the increased price and the fact that it has to be wielded because nobody has bothered to code inaccuracy for spread projectiles (Unless we want to drop the firerate on the c-20 and l6 if powercreep is the concern.), but thats a seperate PR

BramvanZijp:
But yeah I agree, especially on the agent defense weapon. I still believe from an optimal strategy standpoint the c-20 probably should not be bought outside of warops, and the free one should be given to the agent for the ammo management, probably in exchange for them buying an ammo bundle and emag/ab or smth. (and if somebody does buy another one, they should buy their own mags seperate from the ammo bundle)

But for warops, or maybe some 5 man nukie teams idk, this would actually make the c-20 a potentially worth it option. Also, I am not sure if the monkey/kobold reinforcements are able to fire guns, but if so, this would actually be an interesting option for them too.

NoElkaTheGod:
Finaly i can use grenades with a C20 without palying RUSH E on my keyboard.

Also, yes, monke reinforcements can use guns, so giving them a C20 is an option... good luck trying to reload it with one hand though.

KaiserMaus:
Why not make one-handed shooting from C20, as it is now in two hands, and two-handed accuracy will have the accuracy of a lecter?

Pumkin69:
i think the C20-R is a really really good weapon because the ammo is so common and prints so fast you can emag a lathe or get the secfab you can just print smg mags and have nearly infinite ammo. Compared to the L6 which you need to take boxes of ammo and pack the magazines cause you cant print the mags. Being fast as nukies is really important so this is a major major downside to the L6. The bulldog is really great at close quarters burst damage and thats its use case and its good at that, alongside mag printing with the bulldog. If anything is to change i would see if maintainers would let you add a syndicate assault rifle as a middle ground between the L6 and the C20-R. All of that aside i think the change is fine.

superjj18:
Wt550 feels closer to the hos's version of cap's gun(At least before the energy shotgun was added) Than crew's version of the c-20.

I have no issue either way, it's really mostly balancing concerns, but it is good to point out that one handed firing is the one trait that makes hos' unique SMG, unique.

NoElkaTheGod:
> I have no issue either way, it's really mostly balancing concerns, but it is good to point out that one handed firing is the one trait that makes hos' unique SMG, unique.

It isn't unique to HOS though, cargo can buy them. I don't really know why does HOS still have it, since energy shotgun is a thing now.

Pumkin69:
> I don't really know why does HOS still have it, since energy shotgun is a thing now.

HoS would be utterly useless without the WT550 imo

shamblestf:
I think its good that the WT550 is unique with this, and making c20 one handed would be a bit too powerful considering nukies have access to eswords and eshields, i would instead make the c20r super accurate while wielded and it will really shine when bullet speeds are doubled (soon surely) as the nukie single target laser beam weapon

BramvanZijp:
> I think its good that the WT550 is unique with this, and making c20 one handed would be a bit too powerful considering nukies have access to eswords and eshields, i would instead make the c20r super accurate while wielded and it will really shine when bullet speeds are doubled (soon surely) as the nukie single target laser beam weapon

Reloading will be a massive pain if you are using a sword or shield alongside it, and there are plenty of other high power weapons that can do this, think of a traitor hacking a cargo console and buying an smg crate.

Making the c-20r super accurate not only makes zero logical sense for the weapon, as it is an SMG, not a fully automatic sniper rifle, or any rifle at all, but it would make the weapon FAR too overpowered without nerfing the firerate, which puts the weapon in a completely different state/use-case, which also competes with the newly buffed hristov. 

And before you bring up the lecter, that only has a firerate of 5 shots per second, and with its .20 rifle bullets, thats 95 DPS, yet it is still a very popular weapon for security, giving that level of accuracy to a 128 DPS SMG does not seem like a good idea.

There is a reason high firerate and accuracy arent found on the same weapon except for debug items or the pulse carbine (admin event exclusive item, and even then the firerate is pretty low until you factor in the damage per shot), because its be horribly unbalanced. Imagine if the C-20, which can already deal over 100 damage in a single second, could also do this at the edge of your screen. It just doesn't make sense for balance OR logic/game design.


As somebody who has played on frontier where the NFSD (security) can get a pulse carbine mid-late game, it is extremely strong due to its DPS and accuracy, primarily kept in check by the massive downtime which requires a recharger. Doing this with any ballistic weapon seems like a bad idea in general, but ESPECIALLY on the c-20 it just does not make sense in my opinion.

The048:
This seems like a really bad idea. For one, it's an SMG clearly modeled after the old Drozd, it would make no sense to not be able to one-hand. For two, the C-20R is designed as an all-rounder weapon, oriented towards being able to do nearly any task reasonably well, while not specializing in anything too hard, and making it one-handed would make it too flexible. For three, this would create a very toxic potential meta for nukies, since the very likely combo of Eshield+C-20R would be absolutely awful. And lastly, the C-20R is already a very viable weapon, if not one of the best syndicate weapons in the game. I don't see any reason to buff it even more.

BramvanZijp:
> This seems like a really bad idea. For one, it's an SMG clearly modeled after the old Drozd, it would make no sense to not be able to one-hand. For two, the C-20R is designed as an all-rounder weapon, oriented towards being able to do nearly any task reasonably well, while not specializing in anything too hard, and making it one-handed would make it too flexible. For three, this would create a very toxic potential meta for nukies, since the very likely combo of Eshield+C-20R would be absolutely awful. And lastly, the C-20R is already a very viable weapon, if not one of the best syndicate weapons in the game. I don't see any reason to buff it even more.

Firstly, eshield+C20, while powerfull, runs out of ammo very quick, and having to juggle the eshield in order to reload it would get messy. Secondly, the model actually looks like itd be pretty easy to use one handed, ESPECIALLY when compared to the WT550, which looks a lot harder to use one handed, yet that one also has this. I dont see why this weapon cannot be used properly with one hand while the 550 can, and it feels odd if you look at them side by side. 

If we have to end up slightly reducing the firerate, then thats one thing, but I still think this is fair, the crew get the WT550 which has solid firerate but amazing accuracy, while the syndies&nukies get the c-20, which has amazing firerate but mediocre accuracy, I think it also needing to be wielded makes it far too simular to the L6 saw, as mentioned in the PR description.

SlamBamActionman:
Hello! As a Combat workgroup member I will be closing this PR.

The C-20r is not underperforming and is strong as a "default" Nukie ambush weapons. Yes, you are more exposed without something like a shield but the strength of the shield is balanced into other one-handed weapons such as the e-sword. While the gun is a bit of an ammo hog and this becomes a lot worse if you one-hand it, that's something you can end up negating with controlled bursts and adds to the skill expression of the weapon.

If the issue is that the WT550 is overperforming, I'd rather tweak that one down than make the C-20r stronger.

PR #35480, opened 2025-02-24
Cleanup: Use ``Volatile.Read`` instead of ``Thread.VolatileRead`` in ``RoundEndTest``

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Obsolete. #33279.

## Requirements

## Breaking changes

**Changelog**
Nope.


PJB3005 (changes requested):
All this manual thread volatile code should be removed instead. It's completely nonsensical that high-level code like this is using these low-level memory primitives.

MilonPL:
Is this being worked on? If so, please address the review.

PR #35484, opened 2025-02-24
Make the version watermark less annoying

Errant-4:
## About the PR
Watermark is now in the top left corner, after the FPS counter (it's position is fixed whether the FPS counter is active or not). It's also highly transparent now. The screen space it takes up is "not useful" to begin with, because it's covered by the hotbar or some UI buttons, depending on Hud mode. It's also transparent enough to not meaningfully obstruct these UI elements.

## Why / Balance
People complained

## Media
Before:
Screenshot_2025-02-24_234912

After:
Screenshot_2025-02-24_234511
Screenshot_2025-02-24_234529

## Requirements

## Breaking changes

**Changelog**
:cl: Errant
- tweak: The version watermark is now less visually annoying.

### DIFF
--- a/Content.Client/Gameplay/GameplayState.cs
+++ b/Content.Client/Gameplay/GameplayState.cs
@@ -61,10 +61,9 @@ namespace Content.Client.Gameplay
             _version = new Label();
             _version.FontColorOverride = Color.FromHex("#FFFFFF20");
             _version.Text = _changelog.GetClientVersion();
-            _version.Visible = VersionVisible();
             UserInterfaceManager.PopupRoot.AddChild(_version);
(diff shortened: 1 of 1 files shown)


slarticodefast:
Much better for now, but maybe we can move it in the bottom right corner or somewhere else in the future.

slarticodefast (approved):

Errant-4:
I genuinely think the top left corner is as unintrusive as it gets. Bottom right for example is the chat input field on separated, so the user will be actively looking for text there

slarticodefast:
I mean here
Screenshot (365)
But having that adaptive to the chat window size is probably complicated.

PR #29493, opened 2024-06-26
fixed vents pumping air into space

osjarw:
## About the PR
Fixed vents pumping air into rooms which are actively spacing. 
This is accomplished by keeping track of the changes in pressure, and causing a lockout if pressure is dropping.

## Why / Balance
As far as I'm aware, vents should go into a lockout if the room gets spaced. 
Because of this bug, a single hull breach may temporarily empty distro, or cause a constant strain on distro pressure until distro runs out, or the vent somehow gets fixed.
Because spacing speed was slowed down, vents pump out air fast enough to keep their immediate surroundings at high pressures, which doesn't allow the vents to go into a underpressureLockout.

## Technical details
Added a new method in GasVentPumpSystem that keeps track of the air pressure, and skips pumping air for a tick if the pressure is dropping. 
With the current values it seems to be working, though a underpressurelockout failure can still happen if you put 20 vents in the corner of some spacing room, and decrease underpressurelockouttreshold to under 10. 

Also large rooms may repressurize slightly slower, since airflows can trigger the fix I added. The variable `pressurizationLockout` can be changed to somewhat mitigate this. 

Assuming I did my profiling correctly, this fix should have negligible effect on atmospheresystem's usage.

There's probably other ways to fix this bug, like just saving the pressure from last tick and calculating a change in pressure over one tick, and using that instead of doing this averaging stuff I wrote.

## Media

Some footage of the bug in question:

https://github.com/space-wizards/space-station-14/assets/62134478/d032088c-7514-43e2-830c-c10415155793

And here's my fix: 

https://github.com/space-wizards/space-station-14/assets/62134478/893f2f97-b8e6-4507-a52e-b06cdeb3528c

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

PR #35489, opened 2025-02-25
Smart gas vents, open/close with approximate derivative of external pressure

Quantum-cross:
Add a virtual "internal valve" to the gas vents that _open and close_ depending on the _change_ in external pressure over time.

## About the PR

Right now this is a super draft, just want to get some feedback.

Gas vents in this mode have a variable from 0.0 to 1.0 that represents how "open" the vent is from 0% to 100%, this can restrict the gas release flow. This value will get updated depending on the change of external pressure.

Additionally, the old lockout behavior would probably still be useful on shuttles and such where air is a super valuable commodity, as the strict lockout system is much more conservative.

Todo:
- [ ] Clean up variable names, open/close distictions
- [ ] Add air alarm UI to toggle "velocity based lockout" vs "static/strict lockout"
- [ ] Ensure old mapped air alarms use the strict lockout behavior
- [ ] Test minimum "open" value and measure how many mols get leaked compared to old lockout
- [ ] Graph pressure values and "open" value over time for sanity checks
- [ ] Test with larger room sizes (oasis junction)
- [ ] add more todo items when I think of them

## Why / Balance

I hate fixing depressurization when I'm atmos.

With power, when you fix it, the power comes back nearly immediately everywhere and everyone claps.

With atmos, when there is an issue, it takes so much time to fix it, and pressure needs to slowly equalize, and you have to literally click every single air alarm on the station until there's enough pressure to handle the drops via equalization when people open doors.

I feel like any time anyone gasps people beg to call for evac because atmos just cannot get pressure back fast enough.

Vents should be smart enough to detect if they are in vacuum or not...

## Technical details

On every update, the gas vent will compare the external pressure with the last update, then the vent will "open" or "close" by an amount scaled to this difference (open if the pressure is increasing, close if the pressure is decreasing) and also scaled to an "acceleration" parameter. This acceleration parameter is currently multiplied by 10 for positive (opening) increases in order to minimize the effects of room pressure equalization ticks by the Monstermos system.

## Media

https://github.com/user-attachments/assets/e192fdf0-3ece-4fc5-858e-75fa8d5e806c

## Requirements

## Breaking changes
When it's finished I hope there won't be any breaking changes, I'd like to make it so that mapped air alarms default to the old behavior.

**Changelog**

TBD

### DIFF
--- a/Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
+++ b/Content.Server/Atmos/Piping/Unary/Components/GasVentPumpComponent.cs
@@ -36,6 +36,23 @@ namespace Content.Server.Atmos.Piping.Unary.Components
         [DataField]
         public bool UnderPressureLockout { get; set; } = false;
 
+        [DataField]
+        public float VelocityPressureLockoutValue { get; set; } = 1.0f;
+
+        [DataField]
+        public bool VelocityPressureLockoutEnabled { get; set; } = true;
+
+        public float VelocityPressureModifier => VelocityPressureLockoutEnabled && !PressureLockoutOverride ? VelocityPressureLockoutValue : 1.0f;
+
+        [DataField]
+        public float VelocityPressureValueMin { get; set; } = 0.0001f;
+
+        [DataField]
+        public float PreviousPressure { get; set; } = 0.0f;
+
+        [DataField]
+        public float VelocityPressureAcceleration { get; set; } = 2.0f;
+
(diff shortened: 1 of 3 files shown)


deltanedas:
theres another pr that does something similar #29493

Quantum-cross:
going to examine the other pr more and think about this more while working on other things. this barely works in larger environments and if I revisit it, I will come to a different solution.

PR #35491, opened 2025-02-25
Staff of Animation Fixes

ActiveMammmoth:
## About the PR

staf of anime
## Why / Balance

less buggy
## Technical details

- Object now has hard collision set to true, prevents certain objects from moving underneath walls
- Object now has BlockMovement component removed, allows for objects with that component to now move as intended since they're animate
- Object now has Item component removed, prevents object from being picked up / placed into an inventory again
- Moved the bulk of the Animate Spell code to a new AnimateSpellSystem that does all of the physics shenanigans on MapInit
- Got rid of the old AnimateSpellEvent and use ChangeComponentsSpellEvent instead
- Added TEG components to blacklist
- Added Artifact component to blacklist
- Added Animate component which is added to an object when the Animation spell is cast on them
- New AnimateSpellEvent that's just a readonly record struct raised when the spell is cast for other systems to listen to, plan to use in near future PR
- Object is forced out of any container it's inside of when it's initially animated by using AttachParentToContainerOrGrid, prevents animate object from being stuck in a bag
- Wizard faction is now hostile to everything that isn't Wizard, prevents entities like Pun Pun from being ignored by animate objects
- Added DoSpeech variable to ChangeComponentsSpellEvent so the staff can cast the spell without any magic words being spoken
- Added Staff of Animation and RGB Staff to the Summon Magic items list, removed Smite spellbook from the list because crew members don't normally have access to Wizard clothing and Smite requires Wizard clothing
## Requirements


Content.Shared/Magic/Systems/AnimateSpellSystem.cs
```diff
@@ -0,0 +1,55 @@
...
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<AnimateComponent, ComponentStartup>(OnAnimate);
+        SubscribeLocalEvent<AnimateComponent, GettingPickedUpAttemptEvent>(OnPickupAttempt);
+    }
+
+    private void OnAnimate(EntityUid uid, AnimateComponent comp, ComponentStartup args)
+    {
+        // Physics bullshittery necessary for object to behave properly
+
```
ScarKy0, line 27:
```suggestion

    private void OnAnimate(Entity<AnimateComponent> ent, ref OnMapInit args)
    {
        // Physics bullshittery necessary for object to behave properly

```

We use Entity<T> now instead. Code below should be updated to use ent.Owner instead of uid and ent.Comp instead of comp

Content.Shared/Magic/Systems/AnimateSpellSystem.cs
```diff
@@ -0,0 +1,55 @@
...
+        var ev = new AnimateSpellEvent();
+        RaiseLocalEvent(ref ev);
+    }
+
+    private void OnPickupAttempt(EntityUid uid, AnimateComponent comp, GettingPickedUpAttemptEvent args)
+    {
+        args.Cancel();
+    }
+}
```
ScarKy0, line 52:
```suggestion
    private void OnPickupAttempt(Entity<AnimateComponent> ent, GettingPickedUpAttemptEvent args)
    {
        args.Cancel();
    }
}
```

Same deal, also wouldn't removing ItemComponent be better instead of just cancelling GettingPickedUpAttemptEvent? Honestly no clue, could you test that?

Resources/Prototypes/Magic/event_spells.yml
```diff
@@ -181,8 +181,6 @@
         orGroup: Magics
       - id: BlinkBook
         orGroup: Magics
-      - id: SmiteBook
-        orGroup: Magics
```
ScarKy0, line 185: Instead the smite book should be updated to use the SmiteNoReq version

Content.Shared/Magic/Components/AnimateComponent.cs
```diff
@@ -0,0 +1,11 @@
...
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Magic.Components;
+
+// Added to objects when they are made animate
+[RegisterComponent, NetworkedComponent]
+public sealed partial class AnimateComponent : Component
+{
+
+}
```
ScarKy0, line 11:
```suggestion
public sealed partial class AnimateComponent : Component;
```

Content.Shared/Magic/Systems/AnimateSpellSystem.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    [Dependency] private readonly SharedPhysicsSystem _physics = default!;
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly SharedContainerSystem _container = default!;
+
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<AnimateComponent, ComponentStartup>(OnAnimate);
+        SubscribeLocalEvent<AnimateComponent, GettingPickedUpAttemptEvent>(OnPickupAttempt);
+    }
```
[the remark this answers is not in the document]
Tayrtahn: Scar is right.

Just to add some info, the reason is that ComponentInit is raised when an entity is restored from a save, and (for client and shared code) when an entity re-enters a client's PVS. MapInit, on the other hand, is only ever raised once, when the component is first added to the entity.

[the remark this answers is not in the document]
ActiveMammmoth: I see, thank you for the info, made that change

Content.Shared/Magic/Systems/AnimateSpellSystem.cs
ActiveMammmoth: I removed ItemComponent instead, seems to work fine, so got rid of listening to onpickup attempt.

Resources/Prototypes/Magic/staves.yml
```diff
@@ -46,7 +46,7 @@
     actions:
     - ActionAnimateSpell
   - type: Item
-    size: Normal
+    size: Ginormous
```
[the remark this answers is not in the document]
ActiveMammmoth: uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

ScarKy0 (changes requested):

keronshb (approved):
ECS Good.

PR #35496, opened 2025-02-25
Cleanup: Replace uses of the ``AudioHelpers`` with ``AudioParams``

Winkarst-cpu:
## About the PR
Title.

## Why / Balance
Obsolete. #33279.

## Requirements

## Breaking changes

**Changelog**
Nope.


slarticodefast (commented):
The obsoletion comment says "Use AudioParams.Variation data-field".
So I think these should be moved into the sound specifier datafields.

Winkarst-cpu:
> The obsoletion comment says "Use AudioParams.Variation data-field". So I think these should be moved into the sound specifier datafields.

Not possible with every instance, since some are just ``null`` or  ``default!``.

Should I do ``new SoundPathSpecifier()`` to every like that?

PR #35500, opened 2025-02-25
New component for lathes, make processor require nitrogen

kosticia:
## About the PR
Added ore processor prototype that will be require nitrogen, requires mapping.

## Why / Balance
All according to the salvage doc

## Technical details
Added component

## Media
Maybe later

## Requirements

**Changelog**
Actually need mapping so no :cl:


jimmy12or:
I think i'm missing something. What's the point of this one?

kosticia:
> I think i'm missing something. What's the point of this one?

This prevent salvage from using processor as infinite ore storage.

kosticia:
> I still don't get how nitrogen makes salvage not use it as infinite storage, they just need to drag a nitrogen tank from maints or a wreck.
> 
> Don't get me wrong, I like the idea of ore processors needing atmospheric gases to smelt and refine ores, I just don't think it'll solve your issue.

Hmmmmmm. I think you right. I will increase amount of required gas, so it will need atmos system instead of two pipes and nitrogen tank.

VerinSenpai:
My worry with that is, what is happening to the nitrogen its using? if you're taking from atmos, you're gonna drain nitrogen from the distronet.

ArtisticRoomba:
you really need to explain the PR waaaaaaay more just to help with review alone, just having it say "it uses nitrogen" isn't enough. how much nitrogen does it use? does it act like a filter or is it just a static input? etc.

Pronana:
This PR feels weird. I think the component could be useful and opens up the opportunity for more unique machines but the salvage design doc this was made for was made for a salvage that is well in the past now and I dare say outdated. 

With the way salvage currently works, this is just adding extra complexity to a machine that doesn't need it to solve a problem that doesn't exist anymore.

RedBookcase:
> This PR feels weird. I think the component could be useful and opens up the opportunity for more unique machines but the salvage design doc this was made for was made for a salvage that is well in the past now and I dare say outdated.
> 
> With the way salvage currently works, this is just adding extra complexity to a machine that doesn't need it to solve a problem that doesn't exist anymore.

Generally in agreement with this. I don't see much reason to add this to the basic Ore Processor given other basic machines don't require this sort of additional resource to function.

I could maybe see it as an additional component on the Industrial Ore Processor? Like, the lathe/circuit printer upgrades require heat sinks in order to work at more efficient rates, maybe the Industrial Ore Processor requires certain gases to make ore at a more efficient rate. But it may require some additional buff to offset the added complexity.

slarticodefast:
Having such a component for lathes is a great feature that gives us a lot of new options for advanced lathe recipes.
For that reason I think the gas type and amount consumed should be part of the lathe recipe itself instead of having it a constant value in the component.
If we want the to use it for the ore processor can be brought up for discussion, but is not required.

kosticia:
> Having such a component for lathes is a great feature that gives us a lot of new options for advanced lathe recipes. For that reason I think the gas type and amount consumed should be part of the lathe recipe itself instead of having it a constant value in the component. If we want the to use it for the ore processor can be brought up for discussion, but is not required.

💀

PR #35501, opened 2025-02-25
kronkus + other new chemicals + FUN!

FlapstheMoth:
## About the PR
Adds kronkus, kronkus extract, kronkaine.
Adds morphine.
Nerfs and buffs.
Fun added.

## Why / Balance
Wanted by the community.


lzk228:
please write everything that was in the template for your pr

MozarteanChaos:
seems like most of these are under the ronstation namespace. you sure you're PRing this to the right place?

PR #35506, opened 2025-02-25
DocumentParsingManager: Ignore XML comments in guidebook pages

whatston3:
## About the PR

Adds a parser to ignore XML comments in guidebook pages.  Ports https://github.com/new-frontiers-14/frontier-station-14/pull/2979 with a minor fix for end tag parsing.

## Why / Balance

Currently, adding a comment will be treated as a "!--" tag, which will fail to parse and cause a large error window in the guidebook.  It's useful to be able to document changes within a document itself, especially for forks, without causing errors in the guidebook.

## Technical details

A parser was added that consumes both whitespace and comments if they exist.  Any combination of whitespace and complete XML comments, even split between lines, should be accepted.

The test bench used to check this feature is as follows:
1. Edit APE.xml to the following text.
```xml
<Document>
# A.P.E.
<!-- - - > --><!-- A comment?  In my APE guidebook entry?  At this time of year?

 Does it really work?
 are you sure? --><!--really?
-->The Anomalous Particle Emitter is a machine used to interface with anomalies. It is the primary way of manipulating the stats of an anomaly.

<Box>
<GuideEntityEmbed Entity="MachineAPE"/>
</Box>

<!-- Jank Station: Added Psi particles, Collapse effect -->
The A.P.E. can shoot four different kinds of anomalous particles: Delta, Epsilon, Sigma, Psi and Zeta. Each particle type has certain effects:

- [color=crimson]Danger:[/color] Increases the severity of an anomaly.
- [color=plum]Unstable:[/color] Increases the instability of an anomaly.
- [color=goldenrod]Containment:[/color] Increases anomaly stability at the cost of its health - if an anomaly is struck by too many containment particles, it will collapse, leaving behind an anomaly core.
- [color=#6b75fa]Transformation:[/color] Increases anomaly severity with a chance of changing its behavior.
- [color=#33ee33]Collapse:[/color] Greatly reduces the instability of an anomaly, but may delete any accumulated research and the anomaly core with it.
<!-- End Jank Station -->

Which particle type corresponds to each effect is unique and can be learned by scanning the anomaly.

The A.P.E. can also be locked by anyone with science access, which prevents it from being deconstructed, turned on or off, or having the particle type switched.

</Document>
```
2. Run the game, view the page.  There should be no errors, and the page should look like the Media result below.  Note that the comments are not visible.

## Media

The result of step 2 of the test bench above.
image

## Requirements

## Breaking changes

**Changelog**

No in-game effect, no changelog.

### DIFF
--- a/Content.Client/Guidebook/DocumentParsingManager.cs
+++ b/Content.Client/Guidebook/DocumentParsingManager.cs
@@ -36,15 +36,17 @@ public sealed partial class DocumentParsingManager
             .Assert(_tagControlParsers.ContainsKey, tag => $"unknown tag: {tag}")
             .Bind(tag => _tagControlParsers[tag]);
 
+        Parser<char, Unit> whitespaceAndCommentParser = SkipWhitespaces.Then(Try(String("<!--").Then(Parser<char>.Any.SkipUntil(Try(String("-->"))))).SkipMany());
+
         _controlParser = OneOf(_tagParser, TryHeaderControl, ListControlParser, TextControlParser)
-            .Before(SkipWhitespaces);
+            .Before(whitespaceAndCommentParser);
 
         foreach (var typ in _reflectionManager.GetAllChildren<IDocumentTag>())
         {
             _tagControlParsers.Add(typ.Name, CreateTagControlParser(typ.Name, typ, _sandboxHelper));
         }
 
-        ControlParser = SkipWhitespaces.Then(_controlParser.Many());
+        ControlParser = whitespaceAndCommentParser.Then(_controlParser.Many());
 
         _sawmill = Logger.GetSawmill("Guidebook");
     }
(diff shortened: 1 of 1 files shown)


Aidenkrz (approved):
LGTM 👍 👍 👍 👍 👍 👍 👍 
👍 
👍 👍 👍 
pls merg mnatiner i ned drownstream

Content.Client/Guidebook/DocumentParsingManager.cs
```diff
@@ -36,15 +36,17 @@ public void Initialize()
             .Assert(_tagControlParsers.ContainsKey, tag => $"unknown tag: {tag}")
             .Bind(tag => _tagControlParsers[tag]);
 
+        Parser<char, Unit> whitespaceAndCommentParser = SkipWhitespaces.Then(Try(String("<!--").Then(Parser<char>.Any.SkipUntil(Try(String("-->"))))).SkipMany());
```
Simyon264, line 39: For the future, use `var` instead of typed for a variable declaration.

Simyon264 (approved):

whatston3:
Just for the sake of documenting bugs (frankly, forgot I still had this open), there's an issue with whitespace not being properly consumed.  Even in the picture above, note that the "The A.P.E can shoot four different..." is indented by a single space.

Try making an XML comment and adding a list directly underneath, the first item in the list should break the list item parser and appear with a hyphen rather than the angular bracket.

PR #35531, opened 2025-02-26
Starting glasses for Captain and HoP

K-Dynamic:
## About the PR

- Adds starting Administration Glasses to Captain and Centcomm Official, Administration HUD to Head of Personnel

## Why / Balance

Captains still wore sunglasses despite administration glasses being available in their locker. I also gave Centcomm Officials administration glasses as opposed to secglasses make them more reliant on escorts and security.

HoP could've gotten administration glasses too but I felt that would introduce powercreep, especially since their locker has HUD rather than glasses anyway.

This also makes Captains and HoPs less likely to get secglasses from security because their glasses go missing or there's an organic hiring error.

## Technical details

yml changes only

## Media

image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Captain and Head of Personnel starts with administration glasses and hud respectively.
ADMIN:
- tweak: Centcomm Official now starts with administration glasses.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/CentComm/official.yml
+++ b/Resources/Prototypes/Roles/Jobs/CentComm/official.yml
@@ -19,7 +19,7 @@
     jumpsuit: ClothingUniformJumpsuitCentcomOfficial
     shoes: ClothingShoesBootsCombatFilled
     head: ClothingHeadHatCentcom
-    eyes: ClothingEyesGlassesSunglasses
+    eyes: ClothingEyesGlassesCommand
     gloves: ClothingHandsGlovesColorBlack
     outerClothing: ClothingOuterArmorBasic
     id: CentcomPDA
--- a/Resources/Prototypes/Roles/Jobs/Command/captain.yml
+++ b/Resources/Prototypes/Roles/Jobs/Command/captain.yml
@@ -35,7 +35,7 @@
   id: CaptainGear
   equipment:
     shoes: ClothingShoesBootsLaceup
-    eyes: ClothingEyesGlassesSunglasses
+    eyes: ClothingEyesGlassesCommand
     gloves: ClothingHandsGlovesCaptain
     id: CaptainPDA
     ears: ClothingHeadsetAltCommand
--- a/Resources/Prototypes/Roles/Jobs/Command/head_of_personnel.yml
+++ b/Resources/Prototypes/Roles/Jobs/Command/head_of_personnel.yml
@@ -63,6 +63,7 @@
     gloves: ClothingHandsGlovesHop
     ears: ClothingHeadsetAltCommand
     belt: BoxFolderClipboard
+    eyes: ClothingEyesHudCommand
   storage:
     back:
     - Flash
--- a/Resources/Prototypes/Roles/Jobs/Fun/visitors_startinggear.yml
+++ b/Resources/Prototypes/Roles/Jobs/Fun/visitors_startinggear.yml
@@ -7,7 +7,7 @@
   equipment:
     jumpsuit: ClothingUniformJumpsuitCaptain
     shoes: ClothingShoesBootsLaceup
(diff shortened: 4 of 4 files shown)


slarticodefast:
We really need to come up with consistent rules on what equipment should be in lockers and what should be in loadouts.

slarticodefast:
Personally I think keeping them locker only is better.
That way someone can be promoted to command staff.
And having them both a locker and loadout item is not optimal, since it will lead to having a spare one that someone else might grab for powergaming.

slarticodefast:
If they loose their equipment that's a skill issue :godo:

K-Dynamic:
I should stress a lot of jobs have glasses roundstart, there's not much reason why command doesn't.

ScarKy0:
Does the admin hud/glasses have flash immunity?

K-Dynamic:
> Does the admin hud/glasses have flash immunity?

Admin sunglasses has flash protection, the admin hud does not

ScarKy0 (approved):
yeah

ScarKy0:
It doesn't let me re-run checks, youll have to do an empty commit

FairlySadPanda:
Missing an ADMIN CL as this touches on a common admeme role.

ScarKy0 (changes requested):
ah shit, can you also make the changes for the visitor Captain since they get sunglasses?
Afaik visitor HoP doesn't get glasses or a hud so no need to add them there.

K-Dynamic:
> ah shit, can you also make the changes for the visitor Captain since they get sunglasses? Afaik visitor HoP doesn't get glasses or a hud so no need to add them there.
> 
> And as slarti said, for the regular cap/HoP just put them in their lockers instead.

Added them to visitor cap

Admin glasses and hud are already in the captain and hop lockers respectively

ScarKy0:
> Admin glasses and hud are already in the captain and hop lockers respectively

If they're already in the lockers then no need to add them to the loadouts by default then

K-Dynamic:
> If they're already in the lockers then no need to add them to the loadouts by default then

Yeah but the problem is that:
1) Every other job starts with their job-equivalent glasses
2) If they don't start with them then they'll go to security - the whole point of this PR is to let them start with their admin glasses so they're not tempted to go get secglasses instead

Fildrance:
> > If they're already in the lockers then no need to add them to the loadouts by default then
> 
> Yeah but the problem is that:
> 
> 1. Every other job starts with their job-equivalent glasses
> 2. If they don't start with them then they'll go to security - the whole point of this PR is to let them start with their admin glasses so they're not tempted to go get secglasses instead

Other jobs are using non-uniq glasses and they are jobs with more then 1 position usually, no?

ScarKy0 (approved):
I mean I don't think it matters a lot. It's literally just +1 pair of glasses...
Either on person or in locker or both seem fine to me.

pheenty:
> > > If they're already in the lockers then no need to add them to the loadouts by default then
> > 
> > Yeah but the problem is that:
> > 
> > 1. Every other job starts with their job-equivalent glasses
> > 2. If they don't start with them then they'll go to security - the whole point of this PR is to let them start with their admin glasses so they're not tempted to go get secglasses instead
> 
> Other jobs are using non-uniq glasses and they are jobs with more then 1 position usually, no?

 Well yes but these (administration) glasses are unique only because they're only needed by certain jobs, not because they're powerful. They're just a worse sec glasses if anything

K-Dynamic:
> Well yes but these (administration) glasses are unique only because they're only needed by certain jobs, not because they're powerful. They're just a worse sec glasses if anything

I would normally be against extra unique items except secglasses are better and far more ubiquitous.

Fildrance (approved):
Not happy about bloat of uniq items, and feeling pretty pessimistic about this changing command peeps rushing for sec glasses... but okay.
We probably need to just nerf sec glasses at some point but thats harder :godo:

PR #35532, opened 2025-02-26
Allow dump out bags into lathes

kosticia:
## About the PR
Now dumpable items can be dumped out into structures with material storage component.

## Why / Balance
Right now inserting ore into processor is really feels bad, you need to dump it on the floor, them take it in hands, them insert into processor. This pull request allows dump it directly into lathe.

## Technical details
Added some checks into DumpableSystem.cs

## Media
https://github.com/user-attachments/assets/786d8d05-0150-458b-948a-761cc7530302

## Requirements

**Changelog**

<!--
:cl:
- add: Ore bags now can be dumped out directly into ore processor.

### DIFF
--- a/Content.Shared/Storage/EntitySystems/DumpableSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/DumpableSystem.cs
@@ -3,11 +3,11 @@ using Content.Shared.Disposal;
 using Content.Shared.DoAfter;
 using Content.Shared.Interaction;
 using Content.Shared.Item;
+using Content.Shared.Materials;
 using Content.Shared.Placeable;
 using Content.Shared.Storage.Components;
 using Content.Shared.Verbs;
 using Robust.Shared.Audio.Systems;
-using Robust.Shared.Containers;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
 using Robust.Shared.Utility;
@@ -22,6 +22,7 @@ public sealed class DumpableSystem : EntitySystem
     [Dependency] private readonly SharedDisposalUnitSystem _disposalUnitSystem = default!;
     [Dependency] private readonly SharedDoAfterSystem _doAfterSystem = default!;
     [Dependency] private readonly SharedTransformSystem _transformSystem = default!;
+    [Dependency] private readonly SharedMaterialStorageSystem _materialStorage = default!;
 
     private EntityQuery<ItemComponent> _itemQuery;
 
@@ -108,6 +109,20 @@ public sealed class DumpableSystem : EntitySystem
             };
             args.Verbs.Add(verb);
         }
+
+        if (HasComp<MaterialStorageComponent>(args.Target))
+        {
+            UtilityVerb verb = new()
+            {
+                Act = () =>
+                {
+                    StartDoAfter(uid, args.Target, args.User, dumpable);
+                },
+                Text = Loc.GetString("dump-lathe-verb-name", ("lathe", args.Target)),
+                IconEntity = GetNetEntity(uid)
+            };
+            args.Verbs.Add(verb);
+        }
     }
 
     private void StartDoAfter(EntityUid storageUid, EntityUid targetUid, EntityUid userUid, DumpableComponent dumpable)
@@ -164,6 +179,15 @@ public sealed class DumpableSystem : EntitySystem
             foreach (var entity in dumpQueue)
             {
                 _transformSystem.SetWorldPositionRotation(entity, targetPos + _random.NextVector2Box() / 4, targetRot);
+            }
+        }
+        else if (HasComp<MaterialStorageComponent>(args.Args.Target))
+        {
+            dumped = true;
+
+            foreach (var entity in dumpQueue)
+            {
+                _materialStorage.TryInsertMaterialEntity(args.Args.User, entity, args.Args.Target.Value);
             }
         }
         else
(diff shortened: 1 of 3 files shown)


Content.Shared/Storage/EntitySystems/DumpableSystem.cs
```diff
@@ -108,6 +111,29 @@ private void AddUtilityVerbs(EntityUid uid, DumpableComponent dumpable, GetVerbs
             };
             args.Verbs.Add(verb);
         }
+
+        if (TryComp<MaterialStorageComponent>(args.Target, out var comp))
```
beck-thompson, line 115: Use an event for this instead! Create something like `GetDumpableVerbEvent` and have it contain the default looc string and uid of the dumpable item. Then systems can override that with their own custom values (Or nothing if the verb shouldn't appear!)

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
```diff
@@ -166,6 +192,15 @@ private void OnDoAfter(EntityUid uid, DumpableComponent component, DumpableDoAft
                 _transformSystem.SetWorldPositionRotation(entity, targetPos + _random.NextVector2Box() / 4, targetRot);
             }
         }
+        else if (HasComp<MaterialStorageComponent>(args.Args.Target))
```
beck-thompson, line 195: This cannot go on any longer... I'm conscripting you to be the hero that will refactor this!

Instead of doing a massively changed if else statement like this, use events!
1.) Create a `OnDumpedEvent` or something similar. Pass in the uid of the thing dumping!
2.) In each of these systems (E.g `PlaceableSurfaceComponent`, `MaterialStorageComponent` etc...) listen for that event and handle it (Remove the items and stuff)

This will be a LOT nicer and allow each system to do its own thing instead of being a big if statement in the middle of this system.

Resources/Locale/en-US/storage/components/dumpable-component.ftl
```diff
@@ -1,3 +1,4 @@
 dump-verb-name = Dump out on ground
 dump-disposal-verb-name = Dump out into {$unit}
+dump-lathe-verb-name = Dump out into {$lathe}
```
beck-thompson, line 3: you can just combine all three of these into one!

beck-thompson (changes requested):
This also doesn't work if you just click on the lathe itself. 

image

Your also going to need to refactor that to use events or something as well. You might need two or three total!

If you need any help or have questions ask here or the discord :saluting_face:

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
```diff
@@ -128,58 +129,69 @@ private void StartDoAfter(EntityUid storageUid, EntityUid targetUid, EntityUid u
...
 
-            foreach (var entity in dumpQueue)
-            {
-                _transformSystem.SetWorldPositionRotation(entity, targetPos + _random.NextVector2Box() / 4, targetRot);
-            }
-        }
-        else
-        {
-            var targetPos = _transformSystem.GetWorldPosition(uid);
+    public byte DumpTypeCheck(EntityUid uid)
```
Centronias, line 184: Make an enum for this, magic integers are not maintainable.

Resources/Locale/en-US/storage/components/dumpable-component.ftl
```diff
@@ -1,3 +1,2 @@
 dump-verb-name = Dump out on ground
-dump-disposal-verb-name = Dump out into {$unit}
-dump-placeable-verb-name = Dump out onto {$surface}
```
Centronias, line 3: Removing this'll make "Dump out onto steel table" into "Dump out into steel table", which is incorrect in US English.

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
```diff
@@ -128,58 +129,69 @@ private void StartDoAfter(EntityUid storageUid, EntityUid targetUid, EntityUid u
...
-
-    private void OnDoAfter(EntityUid uid, DumpableComponent component, DumpableDoAfterEvent args)
+    private void OnDoAfter(EntityUid uid, DumpableComponent comp, DumpableDoAfterEvent args)
     {
-        if (args.Handled || args.Cancelled || !TryComp<StorageComponent>(uid, out var storage) || storage.Container.ContainedEntities.Count == 0)
+        if (args.Handled
+        || args.Cancelled
+        || args.Args.Target == null
+        || !TryComp<StorageComponent>(uid, out var storage)
+        || storage.Container.ContainedEntities.Count == 0)
```
Centronias, line 146: Usually multi line expressions in the project are broken up with the operator at the end of the line, not the start.
```suggestion
        if (args.Handled ||
            args.Cancelled ||
            args.Args.Target == null ||
            !TryComp<StorageComponent>(uid, out var storage) ||
            storage.Container.ContainedEntities.Count == 0)
```

Resources/Locale/en-US/storage/components/dumpable-component.ftl
beck-thompson: bruh I keep reading them as the same my goober brain cannot handle this

kosticia: > bruh I keep reading them as the same my goober brain cannot handle this

Same💀

PR #35535, opened 2025-02-26
Rifle crate order

K-Dynamic:
## About the PR

- Adds rifle crate (two Lecters and four magazines) for both cargo catalogue and as part of Lethal Weapons gift
- Changes SMG crate description to be more accurate

## Why / Balance

Firearm was missing from cargo inventory when considering pistols, shotguns and SMGs. 

May have implications for nukies and antags as Lecters have high accuracy, high fire rate and high stopping power.

SMG crate description was changed to avoid confusion with the Rifle crate and to remove reference to 'semiautomatic rifles.'

## Technical details

yml changes only

## Media

dotnet_c0WuBfQXmj
image
Outdated: rifle crate now costs now 8000

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Rifle crates with two Lecters and four magazines can be ordered from Cargo for 8000 spesos.
- tweak: Lethal Weapons gift event now includes a Rifle crate.
- tweak: SMG crates are properly described and no longer reference 'semiautomatic rifles.'

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_armory.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_armory.yml
@@ -67,3 +67,13 @@
   cost: 5200
   category: cargoproduct-category-name-armory
   group: market
+
+- type: cargoProduct
+  id: ArmoryRifle
+  icon:
+    sprite: Objects/Weapons/Guns/Rifles/lecter.rsi
+    state: icon
+  product: CrateArmoryRifle
+  cost: 11000
+  category: cargoproduct-category-name-armory
+  group: market
--- a/Resources/Prototypes/Catalog/Cargo/cargo_armory.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_armory.yml
@@ -74,6 +74,6 @@
     sprite: Objects/Weapons/Guns/Rifles/lecter.rsi
     state: icon
   product: CrateArmoryRifle
-  cost: 11000
+  cost: 8000
   category: cargoproduct-category-name-armory
   group: market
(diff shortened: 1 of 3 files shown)


Pumkin69:
its good but i would reduce the price of the crate down to 8000 cause IMO the wt550 is the strongest weapon in the game. It is a one handed, high accuracy, good damage, fully automatic smg. Its easy to reload and you can use a shield with it. Doesn't make much sense that the worse gun would be more expensive.

ps3moira:
Wasn't WT550 removed from cargo for a reason?

Resources/Prototypes/Catalog/Cargo/cargo_armory.yml
```diff
@@ -67,3 +67,13 @@
...
   category: cargoproduct-category-name-armory
   group: market
+
+- type: cargoProduct
+  id: ArmoryRifle
+  icon:
+    sprite: Objects/Weapons/Guns/Rifles/lecter.rsi
+    state: icon
+  product: CrateArmoryRifle
+  cost: 11000
```
ArtisticRoomba, line 77: This costs a lot, which makes this crate unattractive compared to other options (right now the WT550 crate is very popular due to it being a one handed-gun, so this won't be bought as often). It's still a good option for buffing the lethal weapons gift though. I would reduce the price to make it very similar.

ArtisticRoomba (changes requested):
Everything else looks fine

BramvanZijp:
@ArtisticRoomba You forgor

K-Dynamic:
I still find it funny a SMG with a pistol caliber is considered *better* than a battle rifle

PR #35539, opened 2025-02-26
Add AttachmentSystem

notquitehadouken:
## About the PR

Adds a generic attachment system

## Why / Balance

Because I'm going to use it to add gun attachments

## Media

Not player facing

## Requirements

## Breaking changes

**Changelog**

no cl no fun


slarticodefast:
This should probably be unified with `ContainerCompComponent`

notquitehadouken:
> This should probably be unified with `ContainerCompComponent`. Can't you already do all of that by the way? You only need to make an entity prototype and edit its components instead of having the components directly in the AttachmentComponent.

`ContainerCompComponent` modifies entities that go into containers, not entities that have something put into them. AttachmentSystem accomplishes the latter.

notquitehadouken:
also yes i will be using prototypes because this componentregistry idea has not gone even slightly well

Content.Server/Attachments/AttachmentSystem.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            return fieldInfoExisting;
+        }
+        else
+        {
+            throw new ArgumentException(
+                $"Field '{field}' does not exist publicly in component type '{type}'");
+        }
+    }
+
```
metalgearsloth, line 30: 1. We have copycomp now, and component copying should only ever extend ISerializationManager.CopyTo, this is omegacursed.
2. Attachments should either use entities added to the target to store the data OR copying components and storing the original somewhere.

Content.Shared/Attachments/Components/AttachmentHolderComponent.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    ///     If a component name exists in this list, every field will be copied from the reference.
+    /// </summary>
+    [ViewVariables]
+    [DataField]
+    public List<string>? CopyAllFields;
+
+    [ViewVariables(VVAccess.ReadOnly)]
+    public List<(EntityUid, Type)> AddedComps = new ();
+}
```
metalgearsloth, line 38: This is trying to be a godcomp way, way too much, at best it should be a componentregistry and leveraging the engine copycomps / having a method to restore them if desired.

metalgearsloth (changes requested):

TheShuEd:

PR #35543, opened 2025-02-26
Admin Options tab

Errant-4:
## About the PR
Adds an Admin tab on the options menu for selecting/configuring admin stuff. It's only visible to admins, since regular players have no need for it.

It's, uuuh. A little bare right now, I only had the antag overlay classic mode cvar off the top of my head to put on it. Though I have multiple PRs open that add new admin cvars, so I will be able to put those here as well. Do we currently have anyhting else to go here?

## Why / Balance
So admins actually see what settings they can configure and don't have to do it via console

## Media
image

## Requirements

**Changelog**
:cl: Errant
ADMIN:
- add: Some admin settings are now configurable from the Options menu's new Admin tab.


SlamBamActionman (approved):
Go forth, young padawan, and add options.

PR #35557, opened 2025-02-27
Prevent crates, pet carriers and other things from going into disposals

SlamBamActionman:
## About the PR

People found that you could put someone inside of a pet carrier and then put that pet carrier inside of a disposals bin, and then they can't escape. Extra points if you drop a minibomb and get a free gib.

Also the throwing system bypassed any insertion checks, which means that any object that could be considered thrown (which can be done by being launched at high velocity) could end up inside of a disposals bin. 

## Why / Balance

Stuff was inside of disposals bin that shouldn't be in disposals bin. And also toilets.

## Technical details

Blacklisted EntityStorage component from disposals/toilets; as far as I could tell that was only really seen on boxes, crates, lockers and pet carriers.

Also added a cancellable event to the ThrowInsertComponent that other systems can subscribe to for cancelling a throw insert.

## Media

https://github.com/user-attachments/assets/ea58ca3e-bf15-4f1a-90a6-eaa852b78454

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Pet carriers can no longer be put into toilets.
- fix: You can now escape containers (such as pet carriers) inserted into other containers.
- fix: Thrown items no longer bypasses insertion checks for toilets and disposal units.

### DIFF
--- a/Content.Server/Containers/ThrowInsertContainerSystem.cs
+++ b/Content.Server/Containers/ThrowInsertContainerSystem.cs
@@ -31,7 +31,7 @@ public sealed class ThrowInsertContainerSystem : EntitySystem
             return;
 
         var beforeThrowArgs = new BeforeThrowInsertEvent(args.Thrown);
-        RaiseLocalEvent(ent, beforeThrowArgs);
+        RaiseLocalEvent(ent, ref beforeThrowArgs);
 
         if (beforeThrowArgs.Cancelled)
         {
@@ -59,12 +59,5 @@ public sealed class ThrowInsertContainerSystem : EntitySystem
 /// Sent before the insertion is made.
 /// Allows preventing the insertion if any system on the entity should need to.
 /// </summary>
-public sealed class BeforeThrowInsertEvent : CancellableEntityEventArgs
-{
-    public readonly EntityUid ThrownEntity;
-
-    public BeforeThrowInsertEvent(EntityUid thrownEntity)
-    {
-        ThrownEntity = thrownEntity;
-    }
-}
+[ByRefEvent]
+public record struct BeforeThrowInsertEvent(EntityUid ThrownEntity, bool Cancelled = false);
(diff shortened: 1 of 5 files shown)


TheShuEd (approved):
Code looks good for me. I love beforechecks event

Content.Server/Containers/ThrowInsertContainerSystem.cs
```diff
@@ -46,3 +54,17 @@ private void OnThrowCollide(Entity<ThrowInsertContainerComponent> ent, ref Throw
             _adminLogger.Add(LogType.Landed, LogImpact.Low, $"{ToPrettyString(args.Thrown)} thrown by {ToPrettyString(args.Component.Thrower.Value):player} landed in {ToPrettyString(ent)}");
     }
 }
+
+/// <summary>
+/// Sent before the insertion is made.
+/// Allows preventing the insertion if any system on the entity should need to.
+/// </summary>
+public sealed class BeforeThrowInsertEvent : CancellableEntityEventArgs
```
slarticodefast, line 62: should be a record struct

SlimmSlamm:
Why is this a problem though? It's very creative and it's not that easy to do considering just moving even once gets you out of the pet carrier. Not to mention you have to Fortnite build in front of the person just to do it.

slarticodefast:
You cannot escape from the carrier when inside disposals.

SlamBamActionman:
https://github.com/user-attachments/assets/9cd3ac15-9d53-4fd9-95b9-4fb42e7828b0

I changed the underlying behavior so now you can leave containers that are inside of other containers. So pet carriers can still be put inside of disposal units (not toilets, that's stupid), you just can escape them immediately like normal. 

If you somehow get put inside of a locked locker, you'll exit the carrier and immediately begin the doafter to break down the locker door.

Killerqu00:
https://github.com/user-attachments/assets/d2eb5126-cc05-4b04-a7af-1c51b9f6f874

this is the demonstration of the bug in question - it is impossible to move out of the disposals bin once you're in

slarticodefast (approved):

Issue #35493, opened 2025-02-25
Admin Command Shopping List


ArtisticRoomba:
## Description
A list of admin commands desired by the admin team. Obviously #3684 exists, however... yea!

- [ ] A better ghost role UI
    - With fields to select role types, add subtypes, select standard rules and time requirements

- [ ] Command to put an entity into an entity storage
- [ ] Spawn as job verb and command from RMC14
- [ ] A way to ban from taking ghost roles
- [ ] The ability to manually adjust someone's job
- [ ] A way to subscribe to a players local chat listening range
- [ ] Commands that let you adjust, add to, remove body parts
- [ ] `Setmind` but instead of yelling at me for sentience it just does it
- [ ] `Listantags` / `lsantags` (list the antags)
- [ ] A command akin to joingame but that can target another player
    - Example a command like `sendtogame <player>` with optional ID args to set their role and/or station ID
    - Allows admins to bypass UI bugs and expand admemery potential

- [ ] A UI for `adminwho` that refreshes on join/leave
- [ ] "Give me my `togglehealthoverlay` back"
- [ ] Command to spawn ERT/Deathsquad (Something that spawns a prefab shuttle with a variation of the ERT roles ready to go)
- [ ] If a TP is done and you're on a grid, it just uses the current grid as reference instead of requiring a grid input
- [ ] MARK thing could mark more than one entity
- [ ] A command to set outfit
- [ ] `Blcrate [ent] [ent]` to set up a bluespace crate between crate/locker
- [ ] `Bloneway [ent] [ent]` to set up a bluespace crate between crate/locker, but makes it one-way
- [ ] `customghostvote [message] [option 1] ... [option n]`
- [ ] More comprehensive objectives menu, preferably UI
- [ ] A `rerolltraitorobjectives [user]`
- [ ] `forceghost [user]` to force ghost someone
- [ ] `swapmind [user] [user]`
- [ ] Command to add or remove enumerables
- [ ] `clearuploadedprototypes`
- [ ] A way to add a new solution to something's `SolutionContainerManagerComponent` or add a regenerating solution to something, for example `makesolutionregenerate [entid]/solutioncomponent/solution/reagentid amount`
- [ ] `addaccess [entuid] [accesstag]` and `rmaccess [entuid] [accesstag]` for adding/removing access tags from an ID present in someone
- [ ] `checkaccess [entuid]` for returning the list of access tags of the ID someone is holding
[...]

Issue #13099, opened 2022-12-20
Commands should be using dependency fields


metalgearsloth:
No more resolves.

VerinSenpai:
This is the holy book from which all my future PRs shall derive. Til the day it glows a purple hue, may its praises be sang and its verses recited.

SlamBamActionman:
Quick check shows that this is still present in some commands, issue is still relevant.

VerinSenpai:
Work? Never done. Timbers, shivered.

PR #35558, opened 2025-02-27
[ADMIN] Automatically Add Minds.(setmind command)

Schrodinger71:
## About the PR
This PR modifies the `set-mind` command to automatically add a mind to an entity that doesn't have one. If the target entity lacks a `MindContainerComponent`, it now provides feedback to the user indicating that a mind is being added, and calls `MakeSentientCommand.MakeSentient` to make the entity sentient.

## Why / Balance
This change makes the `set-mind` command more flexible by allowing administrations to set minds in entities that aren't initially sentient. This enhances the functionality of the command, enabling it to be used on a broader range of entities, including non-sentient ones. It also improves user feedback by letting them know when the system is automatically making an entity sentient.
- This also solves the problem mentioned here (https://github.com/space-wizards/space-station-14/issues/35493)

## Technical details
- The check for the `MindContainerComponent` was modified so that if the target entity does not have one, it will add a mind to the entity.
- The message shown to the user was updated to indicate that the mind is being added, and it includes the entity's UID.
- The `MakeSentientCommand.MakeSentient` function is called to ensure the entity becomes sentient if it is not already.

## Media
image

## Requirements

## Breaking changes
None.

**Changelog**
:cl: Schrodinger71
ADMIN:
- tweak: `setmind` command now automatically adds a mind to entities that are not sentient and provides feedback when doing so.

### DIFF
--- a/Content.Server/Administration/Commands/SetMindCommand.cs
+++ b/Content.Server/Administration/Commands/SetMindCommand.cs
@@ -1,4 +1,4 @@
-using Content.Server.Players;
+using Content.Server.Mind.Commands;
 using Content.Shared.Administration;
 using Content.Shared.Mind;
 using Content.Shared.Mind.Components;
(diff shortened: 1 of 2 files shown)


Content.Server/Administration/Commands/SetMindCommand.cs
```diff
@@ -1,88 +1,87 @@
...
+[AdminCommand(AdminFlags.Admin)]
+sealed class SetMindCommand : IConsoleCommand
 {
-    [AdminCommand(AdminFlags.Admin)]
-    sealed class SetMindCommand : IConsoleCommand
+    [Dependency] private readonly IEntityManager _entManager = default!;
+
+    public string Command => "setmind";
+
+    public string Description => Loc.GetString("set-mind-command-description", ("requiredComponent", nameof(MindContainerComponent)));
```
Centronias, line 18: This description needs to be updated to indicate that the `MindContainerComponent` isn't required, but is instead added if it's absent.

Resources/Locale/en-US/administration/commands/set-mind-command.ftl
```diff
@@ -1,5 +1,5 @@
 set-mind-command-description = Transfers a mind to the specified entity. The entity must have a {$requiredComponent}. By default this will force minds that are currently visiting other entities to return (i.e., return a ghost to their main body).
 set-mind-command-help-text = Usage: {$command} <entityUid> <username> [unvisit]
 set-mind-command-target-has-no-content-data-message = Target player does not have content data (wtf?)
-set-mind-command-target-has-no-mind-message = Target entity does not have a mind (did you forget to make sentient?)
+set-mind-command-target-add-mind-message = Target entity does not have a mind. For {$uid} adding mind.
```
Centronias, line 4: "For someEntity adding mind" is ungrammatical.
```suggestion
set-mind-command-target-add-mind-message = Target entity {$uid} did not have a mind previously, so one has been added.
```

Content.Server/Administration/Commands/SetMindCommand.cs
```diff
@@ -1,88 +1,87 @@
...
 
-        public CompletionResult GetCompletion(IConsoleShell shell, string[] args)
+        if (!_entManager.HasComponent<MindContainerComponent>(eUid))
         {
-            if (args.Length == 2)
-            {
-                return CompletionResult.FromHintOptions(CompletionHelper.SessionNames(), Loc.GetString("cmd-mind-command-hint"));
-            }
+            shell.WriteLine(Loc.GetString("set-mind-command-target-add-mind-message", ("uid", _entManager.ToPrettyString(eUid))));
+            MakeSentientCommand.MakeSentient(eUid.Value, _entManager);
```
Centronias, line 53: Is it typical for commands to delegate some implementation to another command?

Content.Server/Administration/Commands/SetMindCommand.cs
Schrodinger71: Yeees, thanks!!

Content.Server/Administration/Commands/SetMindCommand.cs
Schrodinger71: I did it according to the examples, I think it is not critical, and is quite justified
image

Content.Server/Administration/Commands/SetMindCommand.cs
```diff
@@ -1,88 +1,87 @@
...
+        }
 
-            return CompletionResult.Empty;
+        if (!IoCManager.Resolve<IPlayerManager>().TryGetSessionByUsername(args[1], out var session))
```
VerinSenpai, line 56: You should convert this to a dependency. See #13099

I don't think this needs to be a localized entity command for that to be depend'd here but I may be wrong.

VerinSenpai (changes requested):

Content.Server/Administration/Commands/SetMindCommand.cs
VerinSenpai: Ah, you know what. The diff got me messed up. Ignore me, this can be handled separately.

Pok27 (changes requested):
It's the same as in #35649 (I'm tired D:)

Schrodinger71:
A?

Pok27:
> A?

I'm talking about the review, this PR has the same problems as that one.

AJCM-git:
@Schrodinger71 coming back to this?

PR #35562, opened 2025-02-27
fix subwizard gamerule

slarticodefast:
Caused an Heisentest at 5% chance.
https://github.com/space-wizards/space-station-14/actions/runs/13572577451/job/37941034977#step:10:407
Should not be abstract so that it can actually be spawned.
No clue why the linter did not catch that.

### DIFF
--- a/Resources/Prototypes/GameRules/subgamemodes.yml
+++ b/Resources/Prototypes/GameRules/subgamemodes.yml
@@ -33,7 +33,6 @@
 
 # Needs testing
 - type: entity
-  abstract: true
   parent: BaseWizardRule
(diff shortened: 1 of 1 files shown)


Tayrtahn:
Good catch, interesting bug.

The linter doesn't catch it because there's nothing *wrong* with an abstract gamerule entity prototype - the problem is trying to spawn it. We have a test that spawns all the (non abstract) game rules, but the problem here was that an abstract rule was added to a subgamemodes rule. It would be pretty easy to add a quick test that looks at any prototypes with the subgamemodesrule component and makes sure that they're valid.

PR #35564, opened 2025-02-27
[HOTFIX] - Players with unknown playtimes now are tagged as new players

beck-thompson:
## About the PR

Title! The admins wanted it.

## Technical details

The playtime of players is sometimes null (This sometimes can occur when the players are first joining the server but can also randomly happen to older players as well). This doesn't fix the root issue, but when it does occur it will tag these players as having unknown playtime and the new player icon in the ahelp menu

## Media

Screenshot from 2025-02-27 11-23-58

## Requirements

### DIFF
--- a/Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
+++ b/Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
@@ -74,7 +74,7 @@ namespace Content.Client.Administration.UI.Bwoink
                 if (info.Antag && info.ActiveThisRound)
                     sb.Append(new Rune(0x1F5E1)); // 🗡
 
-                if (info.OverallPlaytime <= TimeSpan.FromMinutes(_cfg.GetCVar(CCVars.NewPlayerThreshold)))
+                if (info.OverallPlaytime == null || info.OverallPlaytime <= TimeSpan.FromMinutes(_cfg.GetCVar(CCVars.NewPlayerThreshold)))
                     sb.Append(new Rune(0x23F2)); // ⏲
 
                 sb.AppendFormat("\"{0}\"", text);
--- a/Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
+++ b/Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
@@ -36,6 +36,9 @@ namespace Content.Client.Administration.UI.Bwoink
             RobustXamlLoader.Load(this);
             IoCManager.InjectDependencies(this);
 
+            var newPlayerThreshold = 0;
+            _cfg.OnValueChanged(CCVars.NewPlayerThreshold, (val) => { newPlayerThreshold = val; }, true);
+
             var uiController = _ui.GetUIController<AHelpUIController>();
             if (uiController.UIHelper is not AdminAHelpUIHandler helper)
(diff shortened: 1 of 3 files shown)


Resources/Locale/en-US/generic.ftl
```diff
@@ -13,5 +13,6 @@ generic-hours = hours
 generic-minutes = minutes
 
 generic-playtime-title = Playtime
+unknown-playtime = UNKNOWN
```
Errant-4, line 16: There is already a generic-unknown, should we make a new one?

Errant-4:
Although there is precedence for doing so, lol
image

Resources/Locale/en-US/generic.ftl
beck-thompson: I didn't know that cool!

slarticodefast:
Staging is already branched off, so you should target that

Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
```diff
@@ -74,7 +74,8 @@ public BwoinkControl()
                 if (info.Antag && info.ActiveThisRound)
                     sb.Append(new Rune(0x1F5E1)); // 🗡
 
-                if (info.OverallPlaytime <= TimeSpan.FromMinutes(_cfg.GetCVar(CCVars.NewPlayerThreshold)))
+                var newPlayerThreshold = _cfg.GetCVar(CCVars.NewPlayerThreshold);
```
Errant-4, line 77: I think this declaration should be moved "higher" in the code branch, because here, it's going to be called a _lot_ and while the actual cost of this is probably not significant, there is just no reason to be constantly reloading the value of this cvar, it's not going to change

Errant-4:
> Staging is already branched off, so you should target that

I don't really see a reason to wait for the release for a hotfix, I think stable is fine

slarticodefast:
Won't this possibly cause very weird merge conflicts? Because we have to merge stable into master after this PR, but staging will be missing it. And the new staging we will merge into stable won't have the changes master has.

Errant-4:
> Won't this possibly cause very weird merge conflicts? Because we have to merge stable into master after this PR, but staging will be missing it. And the new staging we will merge into stable won't have the changes master has. Pretty sure we said to avoid stuff like this when possible.

100% of the recent Releases caused weird merge conflicts anyway, so I'm not sure that should be a great consideration. But then, if we are willing to wait until release, I propose that it's also easier to just target stable and not merge until after release, and before stable mergeback. Even less chance for any conflict that way

Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
Errant-4:
```cs
var newPlayerThreshold = 0;
_cfg.OnValueChanged(CCVars.NewPlayerThreshold, (show) => { newPlayerThreshold = show; }, true);
```

slarticodefast:
If we wait for release targeting stable should be perfectly fine.

beck-thompson:
I trust whatever you guys think :laughing:

Errant-4 (approved):
Okay, I say let's merge this after release

slarticodefast (approved):

PR #35575, opened 2025-02-28
Greatly reduced how much the stealth shimmer changes over time for ninja and stealth boxes

MidZik:
## About the PR

The wavy shimmer effect that player-used stealth entities have no longer changes much over time and it is now a more static distortion.

## Why / Balance

I've always found it too easy to spot stealth entities, even when they're not moving. This change makes it more difficult, but not impossible, to notice a stealth entity that isn't moving. The distortion amount is the same, but it no longer changes over time significantly. Does not affect how easy it is to spot a moving stealth entity much, if at all, since the movement of the entity dominated the shimmering effect.

I'm focusing on player-used stealth entities to make them a bit stronger (stealth boxes and the ninja). Non player entities (space cobras, anomalies) are unchanged.

## Technical details

Added a new parameter to the stealth shader that determines shimmer frequency. 
Added shimmer frequency field to the Stealth component.
Ninja, Stealth Cardboard Box, and Invisible Crate had their shimmer frequency reduced by a factor of 20.

## Media

Comparison:

[shimmercomparison2.webm](https://github.com/user-attachments/assets/ec553d9b-be15-4b0a-a297-658507e0e9c7)

## Requirements

**Changelog**

:cl:
- tweak: Shimmer effect on an invisible ninja, stealth cardboard box, and invisible crate is now much subtler.


millionse:
It's already disgusting to play against any invisible entity/item/player, please no

Pumkin69:
i like this a lot.

Everturning:
> It's already disgusting to play against any invisible entity/item/player, please no

that sounds like a skill issue, there is a very noticeable distortion in the floor tiles with all invisible things and theres still the sound of movement (if its an entity/player)

slarticodefast:
This should not be a hardcoded change in the shader, but just adjust its parameters instead for the relevant entities you want buffed. Otherwise this will break stuff in other places where it is used, especially for forks.
Conceptually I think stealth should not be too overpowered. It is not supposed to be undetectable.

MidZik:
> This should not be a hardcoded change in the shader, but just adjust its parameters instead for the relevant entities you want buffed. Otherwise this will break stuff in other places where it is used, especially for forks.

If you mean to not change the shader at all and instead modify the `visibility` parameter, then no, I can't do that for this PR. The point is to dramatically reduce the frequency of the shimmering, and changing `visibility` will not accomplish that. The values in the shader were arbitrarily hardcoded in the first place, so it should be fine to adjust them and keep all the stealth entities consistent as they are now.

But if you want, I can add another parameter to the shader that adjusts the shimmer frequency.

> Conceptually I think stealth should not be too overpowered. It is not supposed to be undetectable.

They still aren't undetectable. The distortion remains as before. The problem is that I felt it was _too_ detectable, and every time I spot a stealth object I feel like I'm supposed to pretend I don't see it because it's just so obvious. And keep in mind that this change is very limited in scope. It only really affects things that aren't moving. Stealthed things that are actively trying to kill you will not be standing still.

My goal for this change is that, if there is a stealth box in the room I'm in which I'm not aware of, I shouldn't notice it the majority of the time. It's supposed to be stealthy after all, and if I see it most of the time, I think it has patently failed at being so. I think it is worth giving this change a shot and seeing how it plays out.

slarticodefast:
Yeah, then make them parameters for the shader.

Mixelz:
This wouldn't just affect the player based invisible effects, but also the anomalies that roll the stealth property as well. Something to keep in mind with this change.

> My goal for this change is that, if there is a stealth box in the room I'm in which I'm not aware of, I shouldn't notice it the majority of the time.

At that point why not remove the shimmer and make it truly invisible? Obviously, because that then makes it extremely unfun to play against as it removes the ability to know it's there.

But the problem with stealth obfuscation like that isn't that we see it in the first place, it's that once we learn the pattern to see it becomes so much more obvious to us. There isn't really a fix for that beyond changing the pattern (which doesnt make much sense anyways for this scenario), and I don't believe making the eyesight check even more difficult is the way to go in either case as it may greatly impede on other players who already have an impossible time seeing them.

MidZik:
I added the shader parameter and made a separate commit that changes only the ninja and invisible box and crate.

Not sure why the integration test failed.

MidZik:
Added a better comparison video that shows the old and new side by side.

Xeri7 (approved):
Ran locally and didnt see any issues, might give the invisible people a bit better chance of hiding in a room haha

SlamBamActionman:
Giving this a conceptual approval. The point of the stealth shader is that you're never completely invisible but the binary of being found or not right now heavily leans in favor of experienced players who can recognize the wave-y effect even when not actively looking for it. Movement is a lot easier to detect than just a warp of the sprite and since SS14 is a game that doesn't feature a lot of moving animations the stealth shimmer can be kind of eye-catching.

I've tested this ingame and movement still wrecks the stealth aspect, as it should, and sometimes the shader still distorts things under it to make the invisibility more obvious. 

I have not reviewed the code side with this comment, but conceptually, this is a fine change.

Pok27 (approved):
I cleaned up the code a bit, but everything is fine on your end

PR #35583, opened 2025-02-28
Remove cellular resistance for slimes

SlamBamActionman:
## About the PR

Removes the 80% cellular resistance for slimes.

## Why / Balance

Cellular damage has been moving towards being an choice-balancing damage type, where it's being used as a way to negatively impact certain actions the player chooses to do (e.g. eating brains, incorrectly drawing an implant). If this impact is reduced to irrelevance it tilts the scales too much in the action's favor. Simply put, slimes shouldn't be able to mass-consume brains or implant-check freely.

Sadly this means slimes can no longer chug benzene. I hope the playerbase can survive that.

## Technical details

Just removed the cellular damage resistance in the modifierset. 

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Slimes no longer have cellular damage resistance.

### DIFF
--- a/Resources/Prototypes/Damage/modifier_sets.yml
+++ b/Resources/Prototypes/Damage/modifier_sets.yml
@@ -161,7 +161,6 @@
     Piercing: 1.2
     Cold: 1.5
     Poison: 0.8
-    Cellular: 0.2
 
 - type: damageModifierSet
   id: Infernal
--- a/Resources/ServerInfo/Guidebook/Mobs/SlimePerson.xml
+++ b/Resources/ServerInfo/Guidebook/Mobs/SlimePerson.xml
@@ -20,6 +20,6 @@
   moderately filling food for other species, although drinking the blood of your coworkers is usually frowned upon.
   They suffocate 80% slower, but take pressure damage 9% faster. This makes them by far the species most capable to survive in hard vacuum. For a while.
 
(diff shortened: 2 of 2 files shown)


MozarteanChaos:
AFAIK, slimes were unable to chug benzene or get away with brain-eating in the first place, since consumed reagents don't care about damage resistances.
If the future implant-checking system cares about damage resistances, though, then this is probably a good change.

SlamBamActionman:
TIL! Thanks for checking! ❤️

slarticodefast:
Shouldn't the implanter draw damage already ignore the resistances?
grafik
Or does that only affect clothing and not inherent resistences?

slarticodefast:
I just checked and slimes take the same damage when drawing the wrong implant as other species.
grafik
So this PR can be closed.

SlamBamActionman:
That's on me for not doublechecking my sources.

Regardless, with the intended purpose of cellular as a damage type meant to offset player-taken actions, removing cellular from slimes should still be desirable.

slarticodefast:
> Regardless, with the intended purpose of cellular as a damage type meant to offset player-taken actions, removing cellular from slimes should still be desirable.

That makes sense, especially since there is currently no method of taking damage where this resistance is even applied.

slarticodefast (approved):

slarticodefast:
Ok since we merge this into master it only needs one approval now.
And before anyone complains about a nerf: This resistance is currently not used anywhere and removing it has no gameplay impact.

PR #35587, opened 2025-02-28
Save Space Station 14 from the Toilet Gibber Forever

FairlySadPanda:
## About the PR
Pet carriers are an obvious vector for abuse that requires reworking how EntityStorage actually works to fix, a wonderful change that has a high risk of unintended consequences in other code! The description of the pet carrier has been updated to reference the infamous Toilet Gibber of 2545.

## Why / Balance
Currently an obvious vector for kidnapping and insta-gibbing people thanks to a wonderful combination of toilets and minibombs. Only rarely used for their intended purpose, and without crew objectives to protect pets their "intended unintended" purpose to facilitate dognapping Ian ends up feeling a bit of a weak reason to cope with their abuse.

## Technical details
EntityStorage now actually uses its whitelist properly.
A redundant event has been removed.

## Media

https://github.com/user-attachments/assets/b8890468-7178-41c6-b086-68af2b3fc95d

## Requirements

## Breaking changes
The event ```StoreMobInItemContainerAttemptEvent``` has been removed. Use ```InsertIntoEntityStorageAttemptEvent``` instead, checking if the item being inserted is a mob via ```HasComp<BodyComponent>``` or another solution of your preference.

**Changelog**
:cl:
- fix: Pet carriers now only carry pets. preventing various unintended, albeit hilarious, activities.

### DIFF
--- a/Content.Shared/Foldable/FoldableSystem.cs
+++ b/Content.Shared/Foldable/FoldableSystem.cs
@@ -1,3 +1,4 @@
+using Content.Shared.Body.Components;
 using Content.Shared.Buckle;
 using Content.Shared.Buckle.Components;
 using Content.Shared.Storage.Components;
@@ -23,7 +24,7 @@ public sealed class FoldableSystem : EntitySystem
 
         SubscribeLocalEvent<FoldableComponent, ComponentInit>(OnFoldableInit);
         SubscribeLocalEvent<FoldableComponent, ContainerGettingInsertedAttemptEvent>(OnInsertEvent);
-        SubscribeLocalEvent<FoldableComponent, StoreMobInItemContainerAttemptEvent>(OnStoreThisAttempt);
+        SubscribeLocalEvent<FoldableComponent, InsertIntoEntityStorageAttemptEvent>(OnStoreThisAttempt);
         SubscribeLocalEvent<FoldableComponent, StorageOpenAttemptEvent>(OnFoldableOpenAttempt);
 
         SubscribeLocalEvent<FoldableComponent, StrapAttemptEvent>(OnStrapAttempt);
@@ -45,10 +46,8 @@ public sealed class FoldableSystem : EntitySystem
             args.Cancelled = true;
     }
 
-    public void OnStoreThisAttempt(EntityUid uid, FoldableComponent comp, ref StoreMobInItemContainerAttemptEvent args)
+    public void OnStoreThisAttempt(EntityUid uid, FoldableComponent comp, ref InsertIntoEntityStorageAttemptEvent args)
     {
-        args.Handled = true;
-
         if (comp.IsFolded)
             args.Cancelled = true;
     }
--- a/Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
@@ -331,7 +331,20 @@ public abstract class SharedEntityStorageSystem : EntitySystem
         if (component.Contents.ContainedEntities.Count >= component.Capacity)
             return false;
 
-        return CanFit(toInsert, container, component);
+        // Allow other systems to prevent inserting the item: e.g. the item is actually a ghost.
+        var attemptEvent = new InsertIntoEntityStorageAttemptEvent(toInsert);
+        RaiseLocalEvent(toInsert, ref attemptEvent);
+
+        if (attemptEvent.Cancelled)
+            return false;
+
+        // Consult the whitelist. The whitelist ignores the default assumption about how entity storage works.
+        if (component.Whitelist != null)
+            return _whitelistSystem.IsValid(component.Whitelist, toInsert);
+
+        // Mobs can be inserted into entity storage if the container isn't an item. All other entities must be an item.
+        // Being a mob has a higher priority than being an tem.
+        return HasComp<BodyComponent>(toInsert) ? !HasComp<ItemComponent>(container) : HasComp<ItemComponent>(toInsert);
     }
 
     public bool TryOpenStorage(EntityUid user, EntityUid target, bool silent = false)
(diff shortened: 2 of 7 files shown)


Killerqu00 (commented):
changelog is half in-character, should probably be changed

suggestion

- remove: Pet carriers are now replaced with broken pet carriers until they get a better implementation.
- remove: Thieves no longer get pet steal objectives.

also, the recipe for pet carriers exists, should probably be removed as well

dvir001:
Here is the real fix:
https://github.com/new-frontiers-14/frontier-station-14/commit/fe65c353bc37e9f3b25e47072c57b909393d2a01

Please just use this.

FairlySadPanda:
No, that is shitcode. You don't fix _fundamental design problems with a system_ by just going "hmm yes let's just add a whitelist". That's hacking forward. There is _no good reason_ to use EntityStorage to store pets when we already have the sufficient code to support this in gridinv, a system that does not require constant future maintenance and awareness of this bizarre edge-case.

I have been directly told and instructed, at length, by the project leadership here about how they want this stuff fixed, and you do not fix it via additional complexity to patch a design problem. That's how we get door code.

FairlySadPanda:
OK, went off and then talked more about this:

There is a "Whitelist" available for use on EntityStorage. It's used currently in two situations:
1. Artifact Crushers (only artifacts can be crushed)
2. Borg Rechargers (only borgs can be recharged)

Given there are not any reports of Urist McInnocent being crushed to death inside a Borg Recharger, I'm assuming the whitelist code works.

However, there is a bool called "canStoreMobs" which was added to get pet carriers working. It is only used in _one_ place: the pet carrier (with a joke reference on the Arabian Lamp).

Frontier currently adds a _second_ bool to cope with this - the forceWhitelist bool linked in that commit. I don't think going from "one bool to set up this pre-existing component to two" is a good idea.

I do think that the decision to use entitystorage for pet carriers is inherently flawed. Picking pets up and carrying them around is a thing on SS13 servers, and we already have an item-pet in Hamlet, a well-proven playable character. Gridinv gives us the needed tools.

I'll see what I can do, though, to avoid actually bricking this item.

FairlySadPanda:
image

Hmm yes I love this codebase

FairlySadPanda:
```cs
    public bool CanInsert(EntityUid toInsert, EntityUid container, SharedEntityStorageComponent? component = null)
    {
        if (!ResolveStorage(container, ref component))
            return false;

        if (component.Open)
            return true;

        if (component.Contents.ContainedEntities.Count >= component.Capacity)
            return false;
        
        // Allow other systems to prevent inserting the item: e.g. the item is actually a ghost.
        var attemptEvent = new InsertIntoEntityStorageAttemptEvent(toInsert);
        RaiseLocalEvent(toInsert, ref attemptEvent);

        if (attemptEvent.Cancelled)
            return false;

        // Consult the whitelist. The whitelist ignores the default assumption about how entity storage works.
        if (component.Whitelist != null)
            return _whitelistSystem.IsValid(component.Whitelist, toInsert);

        // Mobs can be inserted into entity storage if the container isn't an item. All other entities must be an item.
        // Being a mob has a higher priority than being an tem.
        return HasComp<BodyComponent>(toInsert) ? !HasComp<ItemComponent>(container) : HasComp<ItemComponent>(toInsert);
    }
    ```
 Lots of single-use hackery surrounding this that can be dumped.

File: Content.Shared/Foldable/FoldableSystem.cs [hunk 1]
```diff
@@ -45,10 +46,8 @@ private void OnFoldableOpenAttempt(EntityUid uid, FoldableComponent component, r
             args.Cancelled = true;
     }
 
-    public void OnStoreThisAttempt(EntityUid uid, FoldableComponent comp, ref StoreMobInItemContainerAttemptEvent args)
+    public void OnStoreThisAttempt(EntityUid uid, FoldableComponent comp, ref InsertIntoEntityStorageAttemptEvent args)
```
FairlySadPanda, line 49: This is the sole use of "StoreMobInItemContainerAttemptEvent". Given we already _always_ raise an event when we attempt to insert something into a container, a listening system can just determine if an entity it cares about is a mob. In this case, an assumption is made that you should not be able to insert _any_ item into a folded EntityStorage.

FairlySadPanda:
Yes, I'm aware force-pushing is a no-no here but considering it's had no real reviews I risked it.

I've rebuilt this PR based on advice. Do I know every way EntityStorage is used? No. Have I made an effort to find where the whitelist and can-affect-mobs bool were used and account for them? Yes. Am I making a dangerous assumption about whitelists? Yes.

SlimmSlamm:
I cant believe I caused two PRs over my toilet gibber strat... anyway this is probably the solution to anything pet carrier related as sad as it is to see it leave.

Kadeo64:
the demon is vanquished and shitcode has been vanquished.. temporarily. (unfortunately mechcode still exists)

FairlySadPanda:
> I cant believe I caused two PRs over my toilet gibber strat... anyway this is probably the solution to anything pet carrier related as sad as it is to see it leave.

I've given you a tribute in the pet carrier item description for your service.

Pumkin69:
damn no more permanently trapping people in disposal chutes with pet carriers.

dvir001:
> Frontier currently adds a _second_ bool to cope with this - the forceWhitelist bool linked in that commit. I don't think going from "one bool to set up this pre-existing component to two" is a good idea.

No need to rewrite this code on Frontier when we can one liner fix this mess and avoid code changes, but do actually fix it properly here 👍

FairlySadPanda:
Yeah, I wrote my original comment about shitcode in the heat of the moment (it's been a day). For downstreams like Frontier coping with an upstream bug is fine, but it's always nice when a fix is done properly on a downstream and it makes its way upstream.

Resources/Prototypes/Entities/Objects/Misc/pet_carrier.yml
```diff
@@ -36,7 +36,9 @@
   - type: EntityStorage
     capacity: 1
     airtight: false
-    itemCanStoreMobs: true
+    whitelist:
+      tags:
+        - VimPilot # If it can fit in a Vim it can fit in a pet carrier.
```
Centronias, line 41: Worth it to rename the `VimPilot` tag since it's used more as a "mob can be put into things" tag now?

Centronias (approved):
"Forever" is a strong word.

Code look nice, me like.

Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
```diff
@@ -331,7 +331,20 @@ public bool CanInsert(EntityUid toInsert, EntityUid container, SharedEntityStora
...
+
+        if (attemptEvent.Cancelled)
+            return false;
+
+        // Consult the whitelist. The whitelist ignores the default assumption about how entity storage works.
+        if (component.Whitelist != null)
+            return _whitelistSystem.IsValid(component.Whitelist, toInsert);
+
+        // Mobs can be inserted into entity storage if the container isn't an item. All other entities must be an item.
+        // Being a mob has a higher priority than being an tem.
```
Coolsurf6, line 346: Just quickly saw this.
"being an tem" > "being an item"

Please no hurt.

deltanedas:
if the issue is putting pet carriers etc in toilets then surely the toilet pr is enough

FairlySadPanda:
Thanks Simyon, will take a look.

@deltanedas - although I've referenced the toilet exploit before, there are at least four disparate issues on upstream content related to this:

1. Use a toilet, a carrier and a minibomb to round-remove people
2. Use an unpowered disposals bin and a carrier to softlock people
3. Use a powered mail delivery system and a carrier to send a friend to any part of the station 
4. Completely hide anybody you like inside a carrier to evade Sec, easily sneak off to have Tot conversations, etc.

We can just try and patch issues as they come up, or we can make the EntityStorage whitelist system work cleanly which offloads the work to the YAML static data, which is way easier to maintain.

deltanedas:
number 2 isnt specific to pet carriers, as seen with "throwing" lockers in
number 3 is a non issue, you can make a tagger and go anywhere solo
number 4 is a non issue, should lockers have nametags above them too so you cant hide? (obviously not the same but still)

Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
FairlySadPanda: that's it

i'm off to become a tibetan monk

tell my family "weh" and then refuse to elaborate

Resources/Prototypes/Entities/Objects/Misc/pet_carrier.yml
FairlySadPanda: That would be a polluting impact of this PR, making this harder to revert in the future if needed.

Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
```diff
@@ -331,7 +332,23 @@ public bool CanInsert(EntityUid toInsert, EntityUid container, SharedEntityStora
         if (component.Contents.ContainedEntities.Count >= component.Capacity)
             return false;
 
-        return CanFit(toInsert, container, component);
+        var aabb = _lookup.GetAABBNoContainer(toInsert, Vector2.Zero, 0);
```
FairlySadPanda, line 335: Collecting all the actual "can insert" logic into "CanInsert".

Simyon264 (approved):

Simyon264 (changes requested):
This PR has breaking changes. Please include them in the PR description. Otherwise it's merge ready.

FairlySadPanda:
Added a note about the removed, very useless, event.

Simyon264 (approved):

Resources/Prototypes/Entities/Objects/Specific/Medical/morgue.yml
```diff
@@ -44,6 +44,10 @@
   - type: EntityStorage
     capacity: 1
     isCollidableWhenOpen: true
+    whitelist: # Use a tag whitelist rather than filtering for PerishableComponent to avoid chefs using body bags for food preservation
```
deltanedas, line 47: freezers hold 30 items and can easily be mass produced so this has no meaning balance wise. it just goes against the immersive sim aspect

FairlySadPanda: That's true. I think I'm just kinda a hater about body bags. If downstreams don't want this lmk, I'll put in a PR to remove this row

IProduceWidgets:
Seems like this actually just broke pet carriers. They can no longer move mobs, which was their intended purpose. *unless* those mobs are pets, but most mobs are not pets, very few are.

PR #35592, opened 2025-02-28
[Draft] Replacing mole to unit conversions with mass to unit conversions.

Alpha-Two:
This is my first C# PR, but I may have bitten off a bit more than I can chew (feel free to close if you don't like it)

## About the PR
In four places, direct conversions between moles and units have been replaced with conversions between mass and units (mass being equal to moles * relative molecular mass). The current method is tuned such that one mole of carbon-12 would be equal to 1u (because chemistry).

Tritium and Plasma have also been given slight buffs and nerfs to their damage, as they have a very low and high molecular mass respectively. These aren't final, however (see Merge Requirements)

## Why / Balance
Why? Because mass -> units is more ✨ realistic ✨.
Balance? Nope, see Merge Requirements.

## Technical details
- Adds a constant (Atmospherics.MolarMassToReagentMultiplier) that is just ~1/12
- Slightly reworks various areas of the code to include molar mass and the constant to the calculations
- Yeets the CreateGas Multiplier constant and the GasCondenser MolesToReagentMultiplier constant, in favour of the above method.
- Plasma + Trit YAML balance change

## Merge Requirements
The main issue with this PR, and the reason it is still a draft, is that the breathing system is fucked and I can't understand it. With the modifications made to the breathing system, saturation decreases at ~18.25kPa (as opposed to ~18.0kPa). Additionally, the "Factor" constant required raising to ~500 for _some_ reason to prevent everyone from suffocation, but doesn't affect the survivable pressure??

Atmos code is cursed. The issue isn't that major, but if the maintainers want it reverted, I'm not going to be the one figuring out how the breathing system works.

Also, the tritium and plasma buffs/nerfs should be playtested, I'll do that myself, but only if the above issue is fixed/accepted as a reasonable change.

## Requirements

## Breaking changes
Since this is still a draft, the changes aren't final but here's the current breaking changes:
Atmospherics.cs: Atmospherics.BreathMolesToReagentMultiplier has been renamed to Atmospherics.MolarMassToReagentMultiplier, and has been expanded in scope.
GasCondenserComponent.cs: The MolesToReagentMultiplier field (previously VV ReadWrite) has been removed

**Changelog**
IDK if it needs a changelog, but I'll amend it with one if the PR gets approval/conceptual approval and it is needed.

### DIFF
--- a/Content.Server/Atmos/Piping/Unary/Components/GasCondenserComponent.cs
+++ b/Content.Server/Atmos/Piping/Unary/Components/GasCondenserComponent.cs
@@ -27,13 +27,4 @@ public sealed partial class GasCondenserComponent : Component
     /// </summary>
     [ViewVariables]
     public Entity<SolutionComponent>? Solution = null;
-
-    /// <summary>
-    /// For a condenser, how many U of reagents are given for a molar mass of 1
-    /// </summary>
-    /// <remarks>
-    /// Taken based on the idea that one mole of Carbon-12 should equal 1u, because chemistry
-    /// </remarks>
-    [DataField, ViewVariables(VVAccess.ReadWrite)]
-    public float MolarMassToReagentMultiplier = 0.0833f;
 }
--- a/Content.Server/Atmos/Piping/Unary/Components/GasCondenserComponent.cs
+++ b/Content.Server/Atmos/Piping/Unary/Components/GasCondenserComponent.cs
@@ -29,12 +29,11 @@ public sealed partial class GasCondenserComponent : Component
     public Entity<SolutionComponent>? Solution = null;
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


0x6273:
> Why? Because mass -> units is more ✨ realistic ✨.

Don't you need the density also to calculate volume from mass?

Alpha-Two:
> > Why? Because mass -> units is more ✨ realistic ✨.
> 
> Don't you need the density also to calculate volume from mass?

Hey I said it was _more_ realistic, never said it _was_ realistic (also it doesn't even really matter as units are so fucked up already as it is)

slarticodefast:
Leave a comment when you come back to this and I'll reopen.

PR #35597, opened 2025-03-01
Reorganize speech traits into accents and traits

Minemoder5000:
## About the PR

Accents are now split into Speech Accents and Speech Traits
## Why / Balance

Allows for better point distribution balancing between language accents and allows for more flexibility with future accents or traits
## Technical details

Added a new trait category
Reorganized speech.yml
Added and modified localization
## Media

image

## Requirements

## Breaking changes

SpeechTraits was split into SpeechTraits and SpeechAccents, including localization changes.
**Changelog**

:cl: Minemoder
- tweak: Accents have been split into Accents and Traits.

### DIFF
--- a/Resources/Locale/en-US/preferences/ui/humanoid-profile-editor.ftl
+++ b/Resources/Locale/en-US/preferences/ui/humanoid-profile-editor.ftl
@@ -61,5 +61,5 @@ humanoid-profile-editor-trait-count-hint = Points available: [{$current}/{$max}]
 
 trait-category-disabilities = Disabilities
 trait-category-speech-accents = Speech Accents
-trait-category-speech-traits = Speech traits
+trait-category-speech-traits = Speech Traits
 trait-category-quirks = Quirks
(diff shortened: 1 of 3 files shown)


UbaserB:
speech traits should go right under speech accents, not disabilities

Minemoder5000:
I agree. However, the game seems to organize it based on the name of the category ID rather than how they're ordered in the yml.

slarticodefast:
Speech Accents should be reduced to 1 available trait point in that case. Having multiple ones and two speech traits on top of that makes your speech pretty much unreadable.

Minemoder5000:
> Speech Accents should be reduced to 1 available trait point in that case.


Would you accept me making the category be 3 points and increasing the cost of high impact accents to 2 points?

slarticodefast:
> Would you accept me making the category be 3 points and increasing the cost of high impact accents to 2 points?

That would still have the same result. Before you were able to select 2 speech modifying traits. Now with splitting the categories you get up to 4. If the speech accents category get 3 points it might even be 5 different ones.

HyperB1:
"Accentless" trait still should be two points, because, well, you probably shouldn't select more accents if you are accentless.
But in case this trait does not block any other accent but species-specific, I take back what I just said.

MozarteanChaos:
> "Accentless" trait still should be two points, because, well, you probably shouldn't select more accents if you are accentless. But in case this trait does not block any other accent but species-specific, I take back what I just said.

Accentless doesn't affect accents applied via traits, only the species' accent.

PR #35599, opened 2025-03-01
Add stamina and mob damage playtest modifiers

SlamBamActionman:
## About the PR

This PR adds two playtest cvars for mob damage and stamina damage.

Mob damage is for all damage dealt to mobs (specifically those with the MobStateComponent).
Stamina damage is all stamina damage dealt.

## Why / Balance

Changing the modifier for damage dealt via damage source means it also impacts damaging structures and objects. This ended up being fairly impactful, so hopefully only affecting mobs will make for more interesting balance tests.

Also stamina damage needed a modifier.

## Technical details

Cvar subscription => update a system value => modify the correct damage.

## Requirements

## Breaking changes

**Changelog**

ADMIN: 
:cl:
- add: Added playtest cvars for mob and stamina damage. Can be changed via changecvar with VVEDIT perms.

### DIFF
--- a/Content.Shared/Damage/Systems/DamageableSystem.cs
+++ b/Content.Shared/Damage/Systems/DamageableSystem.cs
@@ -214,9 +214,6 @@ namespace Content.Shared.Damage
 
             damage = ApplyUniversalAllModifiers(damage);
 
-            if (UniversalMobDamageModifier != 1f && HasComp<MobStateComponent>(uid))
-                damage *= UniversalMobDamageModifier;
-
             // TODO DAMAGE PERFORMANCE
             // Consider using a local private field instead of creating a new dictionary here.
             // Would need to check that nothing ever tries to cache the delta.
--- a/Content.Shared/Damage/Systems/DamageableSystem.cs
+++ b/Content.Shared/Damage/Systems/DamageableSystem.cs
@@ -40,6 +40,7 @@ namespace Content.Shared.Damage
         public float UniversalExplosionDamageModifier { get; private set; } = 1f;
         public float UniversalThrownDamageModifier { get; private set; } = 1f;
         public float UniversalTopicalsHealModifier { get; private set; } = 1f;
+        public float UniversalMobDamageModifier { get; private set; } = 1f;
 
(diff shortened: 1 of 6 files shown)


Content.Shared/Damage/Systems/DamageableSystem.cs
```diff
@@ -212,6 +214,9 @@ public void DamageChanged(EntityUid uid, DamageableComponent component, DamageSp
 
             damage = ApplyUniversalAllModifiers(damage);
 
+            if (UniversalMobDamageModifier != 1f && HasComp<MobStateComponent>(uid))
```
slarticodefast, line 217: Not sure if I like hardcosing this into this function, especially if we might remove it again at some point.
You can just do a subscription to `DamageModifyEvent` with `MobStateComponent` instead.

slarticodefast (commented):
Are all the multiplied damage values (including the existing cvars) adjusted in the guidebook?

Content.Shared/Damage/Systems/DamageableSystem.cs
SlamBamActionman: Yeah that's probably better in this instance, yes

Content.Shared/Mobs/Systems/MobStateSystem.Subscribers.cs
```diff
@@ -186,5 +188,13 @@ private void OnAttemptPacifiedAttack(Entity<MobStateComponent> ent, ref AttemptP
         args.Cancelled = true;
     }
 
+    private void OnDamageModify(Entity<MobStateComponent> ent, ref DamageModifyEvent args)
+    {
+        if (_damageable.UniversalMobDamageModifier != 1)
+        {
+            args.Damage *= _damageable.UniversalMobDamageModifier;
+        }
```
slarticodefast, line 196:
```suggestion
        args.Damage *= _damageable.UniversalMobDamageModifier;
```
I doubt there is any noticeable performance difference from this, so let's do it the simple way

slarticodefast (approved):

PR #35600, opened 2025-03-01
Fingerprint Reader System

ScarKy0:
## About the PR

Added a fingerprint reader system.
Atomized out of #35429

## Why / Balance

Needed for cargo mail. Just atomized this out.

## Technical details

Added FingerprintReaderSystem and FingerprintReaderComponent.
The component stores allowed fingerprints and the system has a public API for checking if they match.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
+++ b/Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
@@ -1,0 +1,46 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.FingerprintReader;
+
+/// <summary>
+/// Component for checking if a user's fingerprint matches allowed fingerprints
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class FingerprintReaderComponent : Component
+{
+    /// <summary>
+    /// Whether the fingerprint reader is enabled
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Enabled = true;
+
+    /// <summary>
+    /// The fingerprints that are allowed to access this entity
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public HashSet<string> AllowedFingerprints = new();
+
+    /// <summary>
+    /// Whether to ignore gloves when checking fingerprints
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool IgnoreGloves;
+
+    /// <summary>
+    /// The popup to show when access is denied due to fingerprint mismatch
+    /// </summary>
+    [DataField]
+    public LocId? FailPopup;
+
+    /// <summary>
+    /// The popup to show when access is denied due to wearing gloves
+    /// </summary>
+    [DataField]
+    public LocId? FailGlovesPopup;
+}
+
+/// <summary>
+/// Event raised when a fingerprint reader's configuration changes
+/// </summary>
+[ByRefEvent]
+public readonly record struct FingerprintReaderConfigurationChangedEvent;
--- a/Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
+++ b/Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
@@ -9,38 +9,33 @@ namespace Content.Shared.FingerprintReader;
 public sealed partial class FingerprintReaderComponent : Component
 {
     /// <summary>
-    /// Whether the fingerprint reader is enabled
+    /// Whether the fingerprint reader is enabled.
+    /// If false, all access attempts will always be allowed.
     /// </summary>
     [DataField, AutoNetworkedField]
     public bool Enabled = true;
 
     /// <summary>
-    /// The fingerprints that are allowed to access this entity
+    /// The fingerprints that are allowed to access this entity.
     /// </summary>
     [DataField, AutoNetworkedField]
     public HashSet<string> AllowedFingerprints = new();
 
     /// <summary>
-    /// Whether to ignore gloves when checking fingerprints
+    /// Whether to ignore gloves when checking fingerprints.
     /// </summary>
     [DataField, AutoNetworkedField]
     public bool IgnoreGloves;
 
     /// <summary>
-    /// The popup to show when access is denied due to fingerprint mismatch
+    /// The popup to show when access is denied due to fingerprint mismatch.
     /// </summary>
     [DataField]
     public LocId? FailPopup;
 
     /// <summary>
-    /// The popup to show when access is denied due to wearing gloves
+    /// The popup to show when access is denied due to wearing gloves.
     /// </summary>
     [DataField]
     public LocId? FailGlovesPopup;
 }
-
-/// <summary>
-/// Event raised when a fingerprint reader's configuration changes
-/// </summary>
-[ByRefEvent]
-public readonly record struct FingerprintReaderConfigurationChangedEvent;
--- a/Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
+++ b/Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
@@ -6,15 +6,9 @@ namespace Content.Shared.FingerprintReader;
 /// Component for checking if a user's fingerprint matches allowed fingerprints
 /// </summary>
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(FingerprintReaderSystem))]
 public sealed partial class FingerprintReaderComponent : Component
 {
-    /// <summary>
-    /// Whether the fingerprint reader is enabled.
-    /// If false, all access attempts will always be allowed.
-    /// </summary>
-    [DataField, AutoNetworkedField]
-    public bool Enabled = true;
-
     /// <summary>
     /// The fingerprints that are allowed to access this entity.
     /// </summary>
--- a/Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
+++ b/Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using Content.Shared.Forensics.Components;
 using Content.Shared.Inventory;
 using Content.Shared.Popups;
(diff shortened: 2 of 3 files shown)


Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
Tayrtahn, line 1: Nitpick: add a period to the end of all the xmldoc descriptions.

Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
```diff
@@ -0,0 +1,46 @@
...
+
+/// <summary>
+/// Component for checking if a user's fingerprint matches allowed fingerprints
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class FingerprintReaderComponent : Component
+{
+    /// <summary>
+    /// Whether the fingerprint reader is enabled
+    /// </summary>
```
Tayrtahn, line 13: Explain how this affects behavior, like:
```suggestion
    /// <summary>
    /// Whether the fingerprint reader is enabled.
    /// If false, access attempts will always be allowed.
    /// </summary>
```

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
...
+            return true;
+
+        if (target.Comp.AllowedFingerprints.Count == 0)
+            return true;
+
+        // Check for gloves first
+        if (HasBlockingGloves(user) && !target.Comp.IgnoreGloves)
+        {
+            if (target.Comp.FailGlovesPopup != null)
+                _popup.PopupEntity(Loc.GetString(target.Comp.FailGlovesPopup), target, user);
```
Tayrtahn, line 33: This should probably use PopupPredicted.

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
...
+            return false;
+        }
+
+
+        // Check fingerprint match
+        if (!TryComp<FingerprintComponent>(user, out var fingerprint) || fingerprint.Fingerprint == null ||
+            !target.Comp.AllowedFingerprints.Contains(fingerprint.Fingerprint))
+        {
+            if (target.Comp.FailPopup != null)
+                _popup.PopupEntity(Loc.GetString(target.Comp.FailPopup), target, user);
```
Tayrtahn, line 43: PopupPredicted here too

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// <summary>
+    /// Adds an allowed fingerprint to a fingerprint reader
+    /// </summary>
+    [PublicAPI]
+    public void AddAllowedFingerprint(Entity<FingerprintReaderComponent> target, string fingerprint)
+    {
+        target.Comp.AllowedFingerprints.Add(fingerprint);
+        RaiseFingerprintReaderChangedEvent(target);
+    }
```
Tayrtahn, line 81: It would also be nice to have an `AddAllowedFingerprints` method for adding multiple fingerprints without raising the event each time.

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
...
+            return true;
+
+        if (!target.Comp.Enabled)
+            return true;
+
+        if (target.Comp.AllowedFingerprints.Count == 0)
+            return true;
+
+        // Check for gloves first
+        if (HasBlockingGloves(user) && !target.Comp.IgnoreGloves)
```
Tayrtahn, line 30: Micro-optimization: reverse the order of these conditions, since checking a bool value is cheaper.

Resources/Locale/en-US/fingerprint-reader/fingerprint-reader.ftl
```diff
@@ -0,0 +1,2 @@
+fingerprint-reader-fail = Your fingerprint does not match!
+fingerprint-reader-fail-gloves = The fingerprint reader cannot read through your gloves!
```
Tayrtahn, line 2: The user could be wearing something in their glove slot that isn't really "gloves". It would be better to pass the worn entity into here and use its name.

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            if (target.Comp.FailGlovesPopup != null)
+                _popup.PopupEntity(Loc.GetString(target.Comp.FailGlovesPopup), target, user);
+            return false;
+        }
+
+
```
beck-thompson, line 37: space

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
...
+public sealed class FingerprintReaderSystem : EntitySystem
+{
+    [Dependency] private readonly InventorySystem _inventory = default!;
+    [Dependency] private readonly SharedPopupSystem _popup = default!;
+
+    /// <summary>
+    /// Checks if the given user has fingerprint access to the target entity
+    /// </summary>
+    [PublicAPI]
+    public bool IsAllowed(Entity<FingerprintReaderComponent?> target, EntityUid user)
```
beck-thompson, line 18: For public functions, add a full comment! The summaries are great but have the descriptions for the inputs as well

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// Checks whether the user has gloves that block fingerprints
+    /// </summary>
+    [PublicAPI]
+    public bool HasBlockingGloves(EntityUid user)
+    {
+        if (_inventory.TryGetSlotEntity(user, "gloves", out var gloves) && HasComp<FingerprintMaskComponent>(gloves))
+            return true;
+
+        return false;
+    }
```
Tayrtahn, line 61: To work with the previous suggestion, you could change this to `TryGetBlockingGloves` and add the gloves' entity as an `out` parameter.

Resources/Locale/en-US/fingerprint-reader/fingerprint-reader.ftl
beck-thompson, line 2: add a new line at the end

Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
```diff
@@ -0,0 +1,46 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.FingerprintReader;
+
+/// <summary>
+/// Component for checking if a user's fingerprint matches allowed fingerprints
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class FingerprintReaderComponent : Component
```
Tayrtahn, line 9: If you add a `SetEnabled(bool)` method to the system, I think you'll have enough API exposed to add an Access attribute here.

Tayrtahn (changes requested):

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
```diff
@@ -0,0 +1,98 @@
+using Content.Shared.Forensics.Components;
+using Content.Shared.Inventory;
+using Content.Shared.Popups;
+using JetBrains.Annotations;
+
+namespace Content.Shared.FingerprintReader;
+
+// TOOD: This has a lot of overlap with the AccessReaderSystem, maybe merge them in the future?
```
beck-thompson, line 8: TOOD -> TODO :laughing:

Content.Shared/FingerprintReader/FingerprintReaderComponent.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// </summary>
+    [DataField]
+    public LocId? FailGlovesPopup;
+}
+
+/// <summary>
+/// Event raised when a fingerprint reader's configuration changes
+/// </summary>
+[ByRefEvent]
+public readonly record struct FingerprintReaderConfigurationChangedEvent;
```
beck-thompson, line 46: Remove this and its related functions like we talked about in discord!

beck-thompson (changes requested):
looks good! Minor stuff

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
ScarKy0: Won't be adding it since im removing the event anyways, i cannot think of any place it could be used as it is

Tayrtahn: Remove the event, but recommend keeping the methods so that you can restrict Access to the component.

You'll also need to add Dirty calls to these methods.

ScarKy0: Do I need to dirty? This is all done in Shared

ScarKy0:
pushed the changes, but i need to test then on my mail PR to see if this works so dont even think of merging yet

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
Tayrtahn: Yep! Prediction *should* make the values match on the local client (but you shouldn't rely on that), but it won't do anything for *other* clients. Dirty the component to make sure the changes are pushed to all clients in range.

Content.Shared/FingerprintReader/FingerprintReaderSystem.cs
ScarKy0: PopupPredicted doens't show up at all.

Tayrtahn: That means that the code isn't running on the client, probably because whatever is calling IsAllowed is only running on the server. Somewhere up the chain isn't predicted.

ScarKy0: Yeah, the system i need it on is all server sided. I'll stick to popupEntity for now

ScarKy0:
Should be good now

beck-thompson (approved):

PR #35601, opened 2025-03-01
draft label workflow

Killerqu00:
## About the PR
Apply S: Draft to draft PRs; needed for triaging

## Media
image

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/.github/workflows/labeler-draft.yml
+++ b/.github/workflows/labeler-draft.yml
@@ -1,0 +1,17 @@
+name: "Labels: Draft"
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, converted_to_draft, ready_for_review]
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    steps:
+    - if: ( github.event.pull_request.draft == true )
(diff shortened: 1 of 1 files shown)


BramvanZijp:
I feel like the tag should only be added if its been triaged while it is still a draft, since otherwise you can just filter for drafts normally.

WIth the label, it stays as S: Draft even if they undraft it, which indicates to triagers that while it isnt a draft anymore, it was when it got triaged, so it may need to be re-triaged.

Killerqu00:
> WIth the label, it stays as S: Draft even if they undraft it

ready_for_review trigger should react to that and remove the label

BramvanZijp:
> > WIth the label, it stays as S: Draft even if they undraft it
> 
> ready_for_review trigger should react to that and remove the label

Which is exactly what I dont want.

If we just want to be able to filter by is this pr a draft, GH has a feature for that.

The reason I suggested the S: Draft tag is that so when a PR is a draft **at the time of being triaged**, it can be given that tag, and then if they undraft it, it keeps the tag and then we know to double-check if the tags are still correct.

Killerqu00:
meh

PR #35603, opened 2025-03-01
[April Fools] Sun Baton

Velken:
## About the PR

Replace the starting stun baton in sec's belts with Sun baton. Works exactly like stun baton, but is super bright.

## Why / Balance

NT made a typo on the requisition form.

## Technical details

Just YAML stuff

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Due to a typo in the requisition form, Security will find their belts with Sun Batons instead of Stun Batons.


ps3moira:
This might have an unintended effect of causing servers to lag due to light pollution. This is similar to what happens when lots of flares are stacked in an area.

Velken:
> This might have an unintended effect of causing servers to lag due to light pollution. This is similar to what happens when lots of flares are stacked in an area.

and that is why it is limited to only the belt spawned stun baton. You will only have as many sun batons as you have sec officers, and I doubt they will all keep theirs turned on at the same time (since meants it will be in your hand, which people can make you slip and steal it)

ScarKy0 (approved):
Here comes the sun

(Tested, PVS is fine)

PR #35606, opened 2025-03-01
Changes the detailed description character length from 512 to 1024 characters

Boaz1111:
## About the PR

I doubled the character limit for the detailed description, visible by default on Salamander. 
## Why / Balance

I have personally had character descriptions cut off, which is really annoying. Far as I can tell, the limit is arbitrary.
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Detailed examine descriptions for your characters can now be twice as long!

### DIFF
--- a/Content.Shared/Preferences/HumanoidCharacterProfile.cs
+++ b/Content.Shared/Preferences/HumanoidCharacterProfile.cs
@@ -30,7 +30,7 @@ namespace Content.Shared.Preferences
 
         public const int MaxNameLength = 32;
         public const int MaxLoadoutNameLength = 32;
-        public const int MaxDescLength = 512;
+        public const int MaxDescLength = 1024;
 
         /// <summary>
         /// Job preferences for initial spawn.


slarticodefast:
I just gave it a test and this is 1000 characters.
Screenshot (366)
It might be a little too much to be honest. It is supposed to be a short, visual description after all, not the characters full life story.

From a technical side this should be a cvar and not hardcoded.

Errant-4:
Honestly, I have always felt that detailed descriptions were often too LONG to actually work well /be meaningful to begin with. 
People have no writing discipline to concentrate on getting their idea across without frequently going into details that their physical description shouldn't even contain to begin with (based on the stated intent of the field). Making the field longer will just make this worse

In fact if we somehow made or coaxed people to write _shorter_ descriptions, it would be more feasible to have that be displayed on the front page of examine, which would actually increase the chances that others will read it. Making the field longer will work against this possibility

Simyon264 (changes requested):
This should be a CVar.

slarticodefast:
Also we should totally add a character counter in the menu where you write that text instead of having it just cut it off.

iaada:
> This should be a CVar.

#35407

Coolsurf6:
I do like to have this increased. At least for my main character they have 2 paragraphs for a description, and then the 3rd section would be new related stuff about them if something happens, lets say they had deep scars on one of their arms for a few weeks etc. This limitation being doubled I feel is perfect and any more wouldn't be necessary. 

I think at least for Salamander, there aren't much roleplay description rules at all, and the few times I've mentioned it have fallen onto deaf ears. So I wasn't really sure if I should keep asking. People should write what **I (My Character)** can **See, Hear, Smell** and possibly other **mannerisms** of a character like twitching, or more prone to moving their hands in conversation, staring or looking erratically etc. Stuff like that is perfect, but writing what is on a characters mind, or what they do for a living kind of breaks immersion. That's the kind of stuff I love to hear from a character, not read. Show don't Tell sorta thing.

I feel like even without this change, people are gonna not hit that limit anyway and those who do, like me would probably use this new limit to the best anyway.

Errant-4:
Recommend closing, now that this is available to be modified as a cvar so servers can set what they want

Simyon264:
Stale, reopen or make a new PR using the CVar added by #35407

PR #35614, opened 2025-03-02
Move entity effects definitions to shared

sowelipililimute:
## About the PR
All entity effect definitions have been moved to shared, alongside their `Effect`/`Condition` implementations when possible. Entity effects that rely on server code for their effects dispatch an `ExecuteEntityEffectEvent`/`CheckEntityEffectConditionEvent` for the server implementation to handle.

## Why / Balance
This is prequisite refactoring work for improvements to the guidebook, both in refactoring code to be more simple (no more special registry for reagent effect strings in the chem guidebook, just access the effects on the client directly), as well as feature work that requires the client to be able to introspect what entity effects do, for e.g. filtering by effect or for generating recipe listings for `CreateEntityReactionEffect` effects without creating special-case systems for serialising and sending a specific piece of data in a way the client can consume.

## Technical details
- move entity effects and conditions to `Content.Shared`
- new `ExecuteEntityEffectEvent`/`CheckEntityEffectConditionEvent` events for entity effects that require server code to execute
- which are handled by an `EntityEffectSystem`

## Requirements

## Breaking changes
Entity effects are now located in Content.Shared, and all entity effects are expected to be instantiable by the client. Any downstream entity effects should be moved to Content.Shared. If the effect/condition implementation requires server code, raise `ExecuteEntityEffectEvent`/`CheckEntityEffectConditionEvent` in the `Effect`/`Condition` method and handle the event on the server.


metalgearsloth:
mother of god

Content.Server/EntityEffects/EntityEffectSystem.cs
```diff
@@ -0,0 +1,968 @@
...
+            var uid = Spawn(args.Effect.Entity, _xform.GetMapCoordinates(args.Args.TargetEntity, xform: transform));
+            _xform.AttachToGridOrMap(uid);
+
+            // TODO figure out how to properly spawn inside of containers
+            // e.g. cheese:
+            // if the user is holding a bowl milk & enzyme, should drop to floor, not attached to the user.
+            // if reaction happens in a backpack, should insert cheese into backpack.
+            // --> if it doesn't fit, iterate through parent storage until it attaches to the grid (again, DON'T attach to players).
+            // if the reaction happens INSIDE a stomach? the bloodstream? I have no idea how to handle that.
+            // presumably having cheese materialize inside of your blood would have "disadvantages".
```
EmoGarbage404, line 589: i cant tell if you wrote this but the answer is SpawnNextTo iirc

Content.Server/EntityEffects/EntityEffectSystem.cs
```diff
@@ -0,0 +1,968 @@
...
+    [Dependency] private readonly SpreaderSystem _spreader = default!;
+    [Dependency] private readonly TemperatureSystem _temperature = default!;
+    [Dependency] private readonly SharedTransformSystem _xform = default!;
+    [Dependency] private readonly VomitSystem _vomit = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<CheckEntityEffectConditionEvent<TemperatureCondition>>(OnCheckTemperature);
```
EmoGarbage404, line 82: these being global subs kinda sucks. It would be a lot nicer if it was raised on the source entity and then you could subscribe via a component. It would reduce a lot of the overhead.

EmoGarbage404 (changes requested):
Good direction but needs changes

Content.Server/EntityEffects/EntityEffectSystem.cs
sowelipililimute: raised on which component of the source entity?

Content.Server/EntityEffects/EntityEffectSystem.cs
```diff
@@ -0,0 +1,968 @@
...
+            return;
+
+        _plantHolder.EnsureUniqueSeed(args.Args.TargetEntity, plantHolderComp);
+        plantHolderComp.Seed.Potency = Math.Max(plantHolderComp.Seed.Potency + args.Effect.Amount, 1);
+    }
+
+    private void OnExecutePlantAdjustToxins(ref ExecuteEntityEffectEvent<PlantAdjustToxins> args)
+    {
+        if (!CanMetabolizePlant(args.Args.TargetEntity, out var plantHolderComp))
+            return;
```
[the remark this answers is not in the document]
sowelipililimute: this is about the same level as boilerplate as it was before but i can look into condensing it even more for the attribute adjust effects

Content.Server/EntityEffects/EntityEffectSystem.cs
EmoGarbage404: you should just be able to raise it on the entity the effect targets, the same way as we do `RaiseLocalEvent(uid, ref ev);`

Content.Server/EntityEffects/EntityEffectSystem.cs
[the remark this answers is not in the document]
sowelipililimute: once i accomodated the effects needing varying conditions (e.g. some effects don't care about the plant being there at all or some requiring a plant not be immutable) the code to make this generic ended up being larger than just writing out the `CanMetabolizePlant` call as required

Content.Shared/EntityEffects/EntityEffect.cs
```diff
@@ -46,6 +46,13 @@ public abstract partial class EntityEffect
     public virtual bool ShouldLog { get; private set; } = false;
 
     public abstract void Effect(EntityEffectBaseArgs args);
+    // public void Effect(EntityEffectBaseArgs args)
+    // {
+    //     var eventType = typeof(ExecuteEntityEffectEvent<>).MakeGenericType(new Type[] { this.GetType() });
+
+    //     object o = Activator.CreateInstance(eventType, this, args);
+    //     _entMan.EventBus.RaiseLocalEvent(o);
+    // }
```
EmoGarbage404, line 55: commented out code

Content.Shared/EntityEffects/Effects/PlantMetabolism/RobustHarvest.cs
```diff
@@ -0,0 +1,21 @@
+using JetBrains.Annotations;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Random;
+
+namespace Content.Shared.EntityEffects.Effects.PlantMetabolism;
+
+[UsedImplicitly]
+[DataDefinition]
```
EmoGarbage404, line 8: you have a few instances of UsedImplicitly and DataDefinition tho these are inherited from EntityEffect

EmoGarbage404 (approved):
good to merge after small comments or after you decide that i'm being annoying and they're not worth doing

PR #35615, opened 2025-03-02
Animated Containers Fix

ActiveMammmoth:
## About the PR

containers animated by the staff of animation no longer have their NPCs bug out
## Why / Balance

If a locker is animated, and has a person inside of it, the NPC bugs out and is useless, needs to be fixed

I asked Keron about it, they told me to just change it so that NPCs cannot target people in containers.
## Technical details

- Changed GetNearbyFactions method in NpcFactionSystem to only look for uncontained targets instead of any target
## Requirements

**Changelog**
:cl: ActiveMammoth
- fix: Hostile NPCs no longer target entities hidden inside of a container.

### DIFF
--- a/Content.Shared/NPC/Systems/NpcFactionSystem.cs
+++ b/Content.Shared/NPC/Systems/NpcFactionSystem.cs
@@ -205,7 +205,7 @@ public sealed partial class NpcFactionSystem : EntitySystem
     private IEnumerable<EntityUid> GetNearbyFactions(EntityUid entity, float range, HashSet<ProtoId<NpcFactionPrototype>> factions)
     {
         var xform = Transform(entity);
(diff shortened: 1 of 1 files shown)


slarticodefast:
I'm not sure if this is something we want for all combat interactions.
It would allow you to be completely safe from any mobs by hiding in any container.

slarticodefast:
> I think they used to target them but it got changed to the current state

That is what this PR is doing

ActiveMammmoth:
Would it be a better idea to prevent people from going inside of a locker if it's animate? Or forcing people out that are inside? Or should I delve deeper and try rewriting the NPC logic so it can successfully target people in containers without glitching out?

ActiveMammmoth:
redoing this pr

PR #35622, opened 2025-03-02
Admin Overlay stacking and ghost hiding

Errant-4:
## About the PR
Admin Overlays of characters that are close to each other (up to 0.95 tile away, adjustable via Admin Options) have their overlays merged into a stack, so all of them are readable. Stack size is limited, and any additional overlays after the last one will be drawn over the last entry. As such, the last entry of any large stack will likely be unreadable, but the alternative is that they bloat on the screen and cover the overlays of any out-of-range characters standing "below" the stack. Default stack size is 3 (adjustable via client cvar)

The names are ordered by vertical position, but when the coordinates are visibly too close/match exactly (such as someone carrying a mothroach/pAI, or ghosts orbiting), the top member of the stack is NOT guaranteed to be the most visible/interesting element in the stack. So if a passenger is carrying someone in a pet carrier, resulting in an overlay stack showing 2 names, one of which is antag and the other not... The overlay by itself will not be able to offer any immediate clues as to which one is which.

For this reason, large stacks compressed into the allowed 3 slots (such as 20 ghosts following someone on a murder spree) would not, by themselves, make it feasible to quickly and easily check the overlay of that one character (until some sort of priority or parenting-based ordering was coded) 
BUT I have the next best thing: the overlays of any Ghosts (not including revenants) now fade out and disappear when near the mouse cursor. So simply flicking the mouse near this theoretical murdering character will hide all the ghosts from the overlay, most likely leaving their overlay as the sole one shown. Fade and Hide distances from the mouse cursor are configurable via client cvar, default 2 and 6 tiles. If there is demand I can put all these cvars on Admin Options but it did not feel necessary?

## Important
Merging overlays that are NEAR each other, not EXACTLY on each other will make overlays look generally more readable, so long as players are still or moving together perfectly (such as when carrying/orbiting another player) 
BUT it will probably also cause frequent merges-unmerges as people run around and pass/cross each other, which will make their overlay jump for a moment. That might actually make them harder to read. Making it so merges only happen at EXACT matches might be more desirable in some situations/for some people, this can be achieved by setting the merge overlay slider to the lowest setting. (It won't let you set it to 0 to ignore potential float jitter in the screencoordinates)

## Why / Balance
Admins specifically asked for overlapping to be solved somehow

## Technical details
Every time an overlay is drawn on the current frame, it's position and size is now saved in a list (). Every time a new one is drawn, it checks the list of previous overlays in case one is within proximity. If there are any, then the new overlay will not be drawn at it's own position, but under the position of the previous overlay, offset by the size of that one, forming a stack. It will not use it's own coordinates, but rather the coordinates of the top element in the stack so elements within a stack don't all wobble around as the member mobs of the stack move around within tolerance range. This, however, means that the _mobs_ of a stack can instead "wobble around" compared to each other and the overlay, so long as they stay within the fairly short merge range. 

The stack follows the top member. If anyone steps out of range, their overlay leaves the stack. Overlays are drawn in order from top of the screen to bottom, so a stack will always start from the "highest" member, making placement in the list correspond to placement on the screen, to the extent within which this can be seen within the rather short overlay merge range. When the middle stack member moves above the top stack member, the stack is immediately redrawn under the now-top stack member. (Potentially dropping the 3rd player off the stack, if it was close enough to the previous stack-starter, but not close enough to this one)

For ghost fading, overlays being drawn check the position of the mob compared to the mouse cursor. If it's within Hide distance, the overlay for that mob is removed entirely, so in the case of a stack, later entries move upward in the stack to fill it's space. If the cursor is only within the outer Fade distance, then a scaled transparency value will be applied to every line the overlay draws, based on the exact distance from Hide to Fade. The gradual fadeout effect looks nicer than just suddenly not being there, and should make it clearer to the player that this is related to the mouse cursor's approach.

## Media

Hiding ghosts from a stack, and some merges/separations

https://github.com/user-attachments/assets/f1cf7505-eea7-44ba-936b-2922ff382b12

Demonstration of some stack merging and separation. It looks pretty normal when "meeting horizontally", not so much when vertically. The admin will probably always just orbit someone of interest anyway, making their overlay pretty stable on the screen

https://github.com/user-attachments/assets/ed7d7adf-2601-404e-b850-5c53f4eae31e

Live: 6 overlays on top of each other. Completely unusable
Screenshot_2025-03-02_125127

A size-5 stack mid-fade. 3 entries are drawn over each other in the last position of the stack, making it unreadable
image
The same stack as above, but ghosts are now removed from it due to mouse proximity
Screenshot_2025-03-02_124548

This is what a size-6 stack would look like, if they were not capped to 3
Screenshot_2025-03-02_124033

Options
image

## Requirements

## Breaking changes
None

**Changelog**
:cl: Errant
ADMIN:
- add: Admin overlays of mobs/players in very close proximity no longer overlap each other. They now form stacks of up to 3. If more than 3 overlay entries are stacked, any additional entries will be drawn over the 3rd element of the stack. These names are ordered by vertical position, but when the coordinates match exactly (such as someone carrying a mothroach/pAI, or ghosts orbiting), the top member of the stack is NOT guaranteed to be the most visible/interesting element in the stack.
- add: Ghosts near the mouse cursor no longer show up on the admin overlay, including any ghosts in stacks near the mouse cursor.


ScarKy0:
epic test fail

PR #35625, opened 2025-03-02
Fix development and debug configurations for tests

Tayrtahn:
## About the PR

Moved the config options to disable events and shuttle auto-recall from debug.toml to development.toml.
Also, PoolManager for tests was disabling the settings from the development config. Changed it so it enables them instead.

## Why / Balance

See Discord discussion: https://discord.com/channels/310555209753690112/900426319433728030/1345826218599125114

### DIFF
--- a/Content.IntegrationTests/PoolManager.Cvars.cs
+++ b/Content.IntegrationTests/PoolManager.Cvars.cs
@@ -33,7 +33,6 @@ public static partial class PoolManager
         (CVars.ReplayServerRecordingEnabled.Name, "false"),
         (CCVars.GameDummyTicker.Name, "true"),
         (CCVars.GameLobbyEnabled.Name, "false"),
-        (CCVars.ConfigPresetDevelopment.Name, "false"),
         (CCVars.AdminLogsEnabled.Name, "false"),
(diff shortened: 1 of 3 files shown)


Content.IntegrationTests/PoolManager.Cvars.cs
```diff
@@ -33,7 +33,7 @@ private static readonly (string cvar, string value)[] TestCvars =
         (CVars.ReplayServerRecordingEnabled.Name, "false"),
         (CCVars.GameDummyTicker.Name, "true"),
         (CCVars.GameLobbyEnabled.Name, "false"),
-        (CCVars.ConfigPresetDevelopment.Name, "false"),
+        (CCVars.ConfigPresetDevelopment.Name, "true"),
```
slarticodefast, line 36: do these overwrite the ccvars above?
because some are different

Tayrtahn: I think this change was a mistake anyway. It shouldn't actually matter, so I'm removing it.

slarticodefast (approved):
Since PJB said they should be in development.toml I think this is good to merge.

PR #35628, opened 2025-03-02
NPC Use Action on Target Refactor - Part 1

Princess-Cheeseballs:
## About the PR
This PR aims to refactor the NPCUseActionOnTarget Component to be a useful tool instead of a component used just by the Goliath to do Goliath things.

On current master, NPCUseActionOnTarget uses a very specific Validation method which only works for the current implementation of the GoliathTentacleAttack, preventing the swapping of actions or modifications of the action directly without also modifying that validation system.

This PR changes the action system to allow both NPCs and Players to use the same action validation system to NPCs can use any action a player can without things breaking.

This allows NPCs to use multiple actions (as demonstrated by the three debug mobs: MobGoliathDebug, MobNinjaDebug, and MobDragonDebug. And also theoretically* any action such as the Ninja Dash and the Dragon's fireball.

*theoretically because sometimes the HTN passes bad targeting info and there's not really any way to properly validate that info, HTN just needs to be able to call actions directly.

In addition, NPCs can now register actions they obtain by picking up items and such to be used, allowing them to have a list of actions they can potentially use, while not having those actions.

## Why / Balance
Rather than having to code targeting data and validation for every single action an NPC might want to use and having to update it any time someone may want to tweak that action, this lets you slap an action onto an NPC and have it be able to use that action exactly how a player can. 

This PR doesn't change how *well* the NPC will use their action, it still retains the old "try every tick method" and makes no changes to the HTN system to allow NPCs to use actions as a plan. However, it does create a number of useful methods for NPCs to use both now and in a future (needed) HTN refactor to use actions without having to duplicate code or rewrite more systems.

This is Part 1 for this exact reason, a known issue is that due to HTN targeting keys being re-used, Urist McNinja may try to teleport into himself causing a debug assert. 

## Technical details
Completely removed TryTentacleAttack and turned it into a more modular function TryUseAction. 

Validating Actions for NPCs and Players now goes through the same method, TryValidAction.

Turned NPCUseActionOnTargetComponent into a list so that NPCs can have multiple actions they can use on targets.

This list includes a class "NpcActionData" which includes the:
- prototype to build from/reference
- the TargetKey the Action is looking for
- the EntityUid of the current action
- whether the NpcActionData is a placeholder for an expected or possible prototype the entity can gain/lose

## Media

Example of an NPC using more than one action, with barely any effort on the yaml side all going through the same methods.

https://github.com/user-attachments/assets/99bececc-f3de-4ce6-be2e-ffc2e0c94e74

## Requirements

## Breaking changes
Restructured NPCUseActionOnTargetComponent to use a list.

**Before:**
- type: NPCUseActionOnTarget
    actionId: ActionGoliathTentacle

**After:**
- type: NPCUseActionOnTarget
    actions:
    - actionId: ActionGoliathTentacle
    - actionId: ActionSmiteNoReq
 
Removed TryTentacleAttack

Moved Action Validation to TryValidAction Method so it can be called by NPCs.

**Changelog**

No CL no fun. No player facing changes.


Pronana:
Test fails...

LaCumbiaDelCoronavirus:
You should probably rename IsValidAction to something more intuitive for what the function actually does; the current name implies that it returns something (e.x.: a bool) that tells you if the action can be done

Or, rewrite it to purely check if an action is doable, which lets it be used in HTN planning

Pronana:
> Edit: Maybe something like TryAction(), which preferably returns a bool, instead of IsValidAction

Yeah that makes more sense, I changed it to TryValidAction since it's both trying to use an action and Validate it. 
Currently the way NPCs work it makes less sense to return a bool as it does to just use the action if it can be validated, since it will only try to use the action if there's a valid target it can see.

Pronana:
Alright fixed merge conflicts and expanded the PR a bit.

Deserty0:
Looking forward to this!

Princess-Cheeseballs:
Raising to P2 cause it's needed for wand refactor

Content.Server/NPC/Systems/NPCUseActionOnTargetSystem.cs
```diff
@@ -15,43 +16,105 @@ public override void Initialize()
...
+        SubscribeLocalEvent<EntityTargetActionComponent, ValidateNpcTargetEvent>(OnNpcEntityTarget);
     }
 
     private void OnMapInit(Entity<NPCUseActionOnTargetComponent> ent, ref MapInitEvent args)
     {
-        ent.Comp.ActionEnt = _actions.AddAction(ent, ent.Comp.ActionId);
+        foreach (var action in ent.Comp.Actions)
+        {
+            if (!action.Ref)
+                action.ActionEnt = _actions.AddAction(ent, action.ActionId) ?? null;
```
Pok27, line 30: Redundant 
```suggestion
                action.ActionEnt = _actions.AddAction(ent, action.ActionId);
```

Content.Server/NPC/Components/NPCUseActionOnTargetComponent.cs
```diff
@@ -7,21 +7,41 @@ namespace Content.Server.NPC.Components;
...
+
+    /// <summary>
+    /// The entityUid of our action
+    /// </summary>
+    [DataField] public EntityUid? ActionEnt;
+
+    /// <summary>
+    /// If true, will not give the entity a new action but will instead try to find a matching action the entity can use. If false, the entity will get a new usable action.
+    /// </summary>
+    [DataField] public bool Ref;
```
Pok27, line 46:
```suggestion
    /// <summary>
    /// HTN blackboard key for the target entity
    /// </summary>
    [DataField]
    public string TargetKey = "Target";

    /// <summary>
    /// The entityUid of our action
    /// </summary>
    [DataField]
    public EntityUid? ActionEnt;

    /// <summary>
    /// If true, will not give the entity a new action but will instead try to find a matching action the entity can use. If false, the entity will get a new usable action.
    /// </summary>
    [DataField]
    public bool Ref;
```

Content.Server/NPC/Systems/NPCUseActionOnTargetSystem.cs
```diff
@@ -15,43 +16,105 @@ public override void Initialize()
...
+
+[ByRefEvent]
+public struct ValidateNpcTargetEvent(EntityUid target)
+{
+    public readonly EntityUid Target = target;
+
+    public bool Invalid;
+    public EntityUid? EntTarget;
+    public EntityCoordinates? EntityCoordinates;
 }
```
Pok27, line 120: need the documentation, and it looks like `Invalid` isn't used?

Pok27 (changes requested):

Princess-Cheeseballs:
This is an ancient PR. Code is not very performant or good, but neither was the code preceeding this so it's not that huge of an issue. I think the changes to SharedActionsSystem are fine enough I'll leave this open but overall this needs much larger refactors for the best behavior.

Pok27 (approved):

PR #35632, opened 2025-03-02
Create an interaction between the item and the target when an item is placed in a target's hands using the strip system

ArchPigeon:
## About the PR
The interaction system was not being called when an individual was placing an item in another individuals hands, preventing fibers and fingerprints from showing up except for the person that was placing the item

## Why / Balance
Behavior is unintended and allows for an individual to evade the forensics system by having other people for example place a knife in that person's hands and then go on a killing spree, no fibers or fingerprints matching that individual would be found on the blade

## Technical details
Pretty trivial code change, just add a call to the doContactInteraction function between the target and the item when an item is placed in the hands slot using the stripping system.

## Media
very small code fix, no media required

## Requirements

## Breaking changes

**Changelog**

    fix: placing an item in an individual's hands now correctly adds forensics information to the item from the individual


metalgearsloth:
If an entity is going into a hand them that should be done in hand system and have forensics handled there. Having to add these checks everywhere is bug prone. Same deal with inventory.

TheShuEd:

PR #35633, opened 2025-03-02
Add GetBaseName method to NameModifierSystem

Tayrtahn:
## About the PR

Adds a method to `NameModifierSystem` for getting the base name of an entity.

## Why / Balance

This came up as being needed for the mail delivery system, and it makes sense to add the functionality to `NameModifierSystem` rather than require other systems that want this info to know the implementation details of name modifiers. This should also avoid having people copy and paste this logic into more places in the future.

## Technical details

Works for any entity, regardless of whether it has any name modifiers applied. If it does, it returns the `BaseName` field value from the `NameModifierComponent`. If not, it returns the `EntityName` from its metadata component.

## Requirements

## Breaking changes

**Changelog**

Nah.

### DIFF
--- a/Content.Shared/NameModifier/EntitySystems/NameModifierSystem.cs
+++ b/Content.Shared/NameModifier/EntitySystems/NameModifierSystem.cs
@@ -41,7 +41,7 @@ public sealed class NameModifierSystem : EntitySystem
     {
         if (Resolve(entity, ref entity.Comp, logMissing: false))
             return entity.Comp.BaseName;
-        return MetaData(entity).EntityName;
+        return Name(entity);
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


ScarKy0:
Be blessed, ill update the mail system with it when able

Content.Shared/NameModifier/EntitySystems/NameModifierSystem.cs
```diff
@@ -32,6 +32,18 @@ private void SetBaseName(Entity<NameModifierComponent> entity, string name)
...
+    /// <summary>
+    /// Returns the base name of the entity, without any modifiers applied.
+    /// If the entity doesn't have a <see cref="NameModifierComponent"/>,
+    /// this returns the entity's metadata name.
+    /// </summary>
+    public string GetBaseName(Entity<NameModifierComponent?> entity)
+    {
+        if (Resolve(entity, ref entity.Comp, logMissing: false))
+            return entity.Comp.BaseName;
+        return MetaData(entity).EntityName;
```
slarticodefast, line 44:
```suggestion
        return Name(entity);
```
There is a shorthand for this!

Tayrtahn: Thanks!

slarticodefast (approved):

PR #35438, opened 2025-02-24
DetGadget Hat Revitalization

ArtisticRoomba:
## About the PR
Adds the Go Go Hat, a hat that has the ability to store items and retrieve items automatically via the saying of a catchphrase. This catchphrase can be anything you want, but obviously it's mandatory for you to say GO GO GADGET.

Basically it's the DetGadget hat from SS13. WIthout the scuttlebot component.

This is a revitalization of the closed PR #30188, updated to address reviews.

## Why / Balance
INSPECTOR GADGET

As for the balance component, this hat adds a sizeable inventory to the detective, a standard backpack on top of your head! However it can only fit small items or lower, and you have to request the item by saying what you want, so there's a tradeoff there.

Right now the only disallowed items are:
- The nuke disk and fakes
- Any high-value item, for example syndie objectives.

I'm willing to decrease the size of the hat or expand its limit to other items.

## Technical details
Creates the StorageVoiceControlSystem and the StorageVoiceControlComponent. The VoiceTriggeredEvent is raised by TriggerOnVoice, piggybacking on that system for the catchphrase features.

For the StorageVoiceControlSystem:
- The system receives events raised by a VoiceTriggeredEvent.
- System yoinks respective components needed.
- If the player has an item in their hand, the system will try to insert that item into the storage. If it fails it'll tell the player why.
- If the player's hand is empty, the system iterates through each item in the storage and if it finds an item that matches the requested item it'll put the item in the player's hands.
	- What if both hands are empty? Well, the logic determines that if there's an active item in your hand, it'll try to insert it, so we guarantee that the player will have an empty hand to accept an item. otherwise it'll try to insert it.

All of this is admin logged as well under Action.

## Media
https://github.com/user-attachments/assets/80c1f88e-5173-47fd-ac19-8cedd08d2d51

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
nada

**Changelog**
:cl: ArtisticRoomba, Cojoke-dot
- add: The detective can now find the go go hat in their locker. This hat has the ability to store and retrieve items automatically via the saying of a catchphrase. GO GO GADGET!

PR #35636, opened 2025-03-03
Add "Reset to default" verb to `TriggerOnVoice`

MilenVolf:
## About the PR
Adds new verb that resets the keyphrase if we have change it.
This verb will be visible only if `DefaultKeyPhrase` is set in prototype and our current keyphrase differs from the default one.

Also localizes keyphrase for the "go go hat" prototype

## Why / Balance
This will allow to make keyphrases localizable for prototypes like "go go hat" in #35438
Also allows to reset keyphrase to default by verb which is a bit handy

## Technical details
Added new datafield `DefaultKeyPhrase` that stores the default keyphrase LocId value
`KeyPhrase` field was made only visible and editable in VV but not settable in yml (`DataField` attribute removed)

## Media

https://github.com/user-attachments/assets/5fb743ff-33b8-4421-a05e-c55a3ef47156

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Go go hat's activation phrase can be reset to the default without having to record it manually to reset the phrase.

### DIFF
--- a/Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
+++ b/Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
@@ -19,6 +19,9 @@ namespace Content.Server.Explosion.EntitySystems
 
         private void OnVoiceInit(EntityUid uid, TriggerOnVoiceComponent component, ComponentInit args)
         {
+            if (!string.IsNullOrEmpty(component.DefaultKeyPhrase))
+                component.KeyPhrase = Loc.GetString(component.DefaultKeyPhrase);
+
             if (component.IsListening)
                 EnsureComp<ActiveListenerComponent>(uid).Range = component.ListenRange;
             else
@@ -80,6 +83,20 @@ namespace Content.Server.Explosion.EntitySystems
                 Priority = 1
             });
 
+            if (!string.IsNullOrEmpty(component.DefaultKeyPhrase)
+                && component.KeyPhrase != Loc.GetString(component.DefaultKeyPhrase))
+            {
+                args.Verbs.Add(new AlternativeVerb
+                {
+                    Text = Loc.GetString("verb-trigger-voice-default"),
+                    Act = () =>
+                    {
+                        component.KeyPhrase = Loc.GetString(component.DefaultKeyPhrase);
+                        SetToDefault(ent, @event.User);
+                    },
+                });
+            }
+
             if (string.IsNullOrWhiteSpace(component.KeyPhrase))
                 return;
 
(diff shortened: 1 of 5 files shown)


Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
```diff
@@ -19,6 +19,9 @@ private void InitializeVoice()
 
         private void OnVoiceInit(EntityUid uid, TriggerOnVoiceComponent component, ComponentInit args)
         {
+            if (!string.IsNullOrEmpty(component.DefaultKeyPhrase))
+                component.KeyPhrase = Loc.GetString(component.DefaultKeyPhrase);
```
Tayrtahn, line 23: This should be in MapInit so it doesn't get stomped on save/load.

Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
```diff
@@ -80,6 +83,20 @@ private void OnVoiceGetAltVerbs(Entity<TriggerOnVoiceComponent> ent, ref GetVerb
...
 
+            if (!string.IsNullOrEmpty(component.DefaultKeyPhrase)
+                && component.KeyPhrase != Loc.GetString(component.DefaultKeyPhrase))
+            {
+                args.Verbs.Add(new AlternativeVerb
+                {
+                    Text = Loc.GetString("verb-trigger-voice-default"),
+                    Act = () =>
+                    {
+                        component.KeyPhrase = Loc.GetString(component.DefaultKeyPhrase);
```
Tayrtahn, line 94: Shouldn't this be inside `SetToDefault`?

Tayrtahn (changes requested):

Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
MilenVolf: Sure, will do

Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
```diff
@@ -129,6 +149,23 @@ public void FinishRecording(Entity<TriggerOnVoiceComponent> ent, EntityUid sourc
             _popupSystem.PopupEntity(Loc.GetString("popup-trigger-voice-recorded", ("keyphrase", component.KeyPhrase!)), ent);
         }
 
+        private void SetToDefault(Entity<TriggerOnVoiceComponent> ent, EntityUid source)
+        {
+            var component = ent.Comp;
```
slarticodefast, line 154:
```suggestion
```
just write `ent.Comp` where needed

Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
```diff
@@ -129,6 +149,23 @@ public void FinishRecording(Entity<TriggerOnVoiceComponent> ent, EntityUid sourc
...
+
+            if (!string.IsNullOrEmpty(component.DefaultKeyPhrase))
+                component.KeyPhrase = Loc.GetString(component.DefaultKeyPhrase);
+
+            EnsureComp<ActiveListenerComponent>(ent).Range = component.ListenRange;
```
slarticodefast, line 159: why reset this as well?

slarticodefast (changes requested):

Content.Server/Explosion/EntitySystems/TriggerSystem.Voice.cs
MilenVolf: What exactly?
This sets KeyPhrase to the default one
```C#
            if (!string.IsNullOrEmpty(component.DefaultKeyPhrase))
                component.KeyPhrase = Loc.GetString(component.DefaultKeyPhrase);
```
This component is needed to be on entity so it will listen to player's words to be able to trigger
```C#
            EnsureComp<ActiveListenerComponent>(ent).Range = component.ListenRange;
```

MilenVolf: I changed the if statement so it will return if it's null or empty

slarticodefast: > What exactly?

I mean why does the "reset phrase" verb change the listening range?
Shouldn't that be set on component startup and stay constant? Why do we have two datafields for this in the first place?

MilenVolf: I guess two datafields needed to pass the value to added ListenerComponent field

MilenVolf:
Updated for the refactored trigger system

Content.Shared/Trigger/Systems/TriggerSystem.Voice.cs
```diff
@@ -11,18 +11,27 @@ public sealed partial class TriggerSystem
...
+
+        UpdateListening(ent);
+    }
+
     private void OnVoiceInit(Entity<TriggerOnVoiceComponent> ent, ref ComponentInit args)
     {
-        if (ent.Comp.IsListening)
-            EnsureComp<ActiveListenerComponent>(ent).Range = ent.Comp.ListenRange;
-        else
-            RemCompDeferred<ActiveListenerComponent>(ent);
```
MilenVolf, line 25: Wanted to move it to ```MapInitEvent``` altogether, but decided to keep it "as is" in case something breaks when loading a save.

PR #35637, opened 2025-03-03
Persist construction menu favorites client-side

### LINKED ISSUE #29592: Favorites in the construction menu do not persist after a round ends
be atmos
yay i get to favorite all my pipes and atmos devices now!!
round ends
check menu next round
everythings default

No point in favoriting if it's just going to go back to normal atm.

YotaXP:
## About the PR
Persists the list of favorite items chosen in the construction menu client-side.

## Why / Balance
Addresses a common pain point of the construction menu for players that requires them to repopulate their favorites list after every round or reconnect.
Resolves #29592

## Technical details
- This adds a new client-only CVar `ui.construction_favorites` to store a comma-separated list of recipe prototype IDs.
- The order in which favorites are added is preserved. This provides no benefit today since the UI sorts items alphabetically regardless, however it leaves room for future changes that may allow players to reorder items.
- Unrecognized IDs are preserved in the list. This is important since the favorites list will be shared between all forks that adopt this PR, which will have their own IDs.

This solution is less ideal than one that saves the list server-side, but much simpler to implement and hopefully more convenient for players who regular multiple servers. If we or another fork decide to implement a server-side solution in the future, it would be a trivial matter to disregard the CVar.

## Requirements
- [X] I have added media to this PR or it **does not require an ingame showcase**.

**Changelog**
:cl:
- tweak: Favorites selected in the construction menu will now persist between rounds.

### DIFF
--- a/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
+++ b/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
@@ -478,15 +478,19 @@ namespace Content.Client.Construction.UI
             PopulateCategories(_selectedCategory);
         }
 
+        // The favorites list is stored within a CVar and shared between multiple SS14 forks with potentially differing behaviors.
+        // To minimize fighting between clients, the following rules should be observed when updating the variable:
+        //  - Do not remove or modify unrecognized entries. Other forks may have different recipes.
+        //  - Do not rearrange the list unless the player intentionally performs some action to do so.
         public void SetFavorites(string favorites)
         {
             _favoritedRecipes.Clear();
             _favoritedRecipeIDs.Clear();
 
-            _favoritedRecipeIDs.AddRange(favorites.Split(',', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries));
+            _favoritedRecipeIDs.AddRange(favorites.Split(','));
             foreach (var id in _favoritedRecipeIDs)
             {
-                if (_prototypeManager.TryIndex(id, out ConstructionPrototype? recipe))
+                if (_prototypeManager.TryIndex(id.Trim(), out ConstructionPrototype? recipe))
                     _favoritedRecipes.Add(recipe);
             }
 
--- a/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
+++ b/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
@@ -2,6 +2,7 @@ using System.Linq;
 using System.Numerics;
 using Content.Client.Stylesheets;
 using Content.Client.UserInterface.Systems.MenuBar.Widgets;
+using Content.Shared.CCVar;
 using Content.Shared.Construction.Prototypes;
 using Content.Shared.Whitelist;
 using Robust.Client.GameObjects;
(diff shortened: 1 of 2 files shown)


VerinSenpai:
huge.

ArtisticRoomba:
its beautiful

BigfootBravo:
Oh please, this would help so much with combating the minor inconvenience. Amazing.

YotaXP:
For whoever reviews this, please leave feedback on whether I should go further on future-proofing this with other forks. (I.E. using a JSON-encoded array of objects instead of a simple comma-separated string, so that additional metadata can later be added.) I believe it's important to get this done right on the first go, and I'm often a terrible judge of how far to take things.

Also, I'm embarrassed I forgot to add comments. Feel free to request those.

metalgearsloth:
Probably better to use datadefinitions and save it to the userdata folder?

YotaXP:
> Probably better to use datadefinitions and save it to the userdata folder?

Cross-compatibility is key if the favorites list is to be shared. If a DataDefinition is deserialized with an unrecognized property (such as one that a different fork adds) will that property be preserved when it is reserialized later? I'm still familiarizing myself with the codebase, so bare with me as I figure this stuff out.

metalgearsloth:
> > Probably better to use datadefinitions and save it to the userdata folder?
> 
> Cross-compatibility is key if the favorites list is to be shared. If a DataDefinition is deserialized with an unrecognized property (such as one that a different fork adds) will that property be preserved when it is reserialized later? I'm still familiarizing myself with the codebase, so bare with me as I figure this stuff out.

If it's unknown it will get dropped, it would be similar to character profile exports / imports.

YotaXP:
> If it's unknown it will get dropped, it would be similar to character profile exports / imports.

That makes it a no-go for this solution then. The way I see it right now, there are a few ways this can go.

- Simple comma-separated list of IDs (This PR currently)
- JSON data that can be extended with custom properties (Perhaps a bit too information-dense for a CVar)
- Compromise: Allow extra data in each entry after a delimiting character (I.E. "AirAlarm:more data,Wall:also more data,Etc:"), but reserve the format of it for future deliberation. (Kinda messy, imo)
- Persist instead to per-server file on the client, like `changelog_last_seen_...` does. (May be better to prepare a more robust system for maintaining this sort of arbitrary data on the client.)
- Persist to the server-side database alongside character data. (Way outside my comfort zone currently.)

metalgearsloth:
There's 0 reason to use json and cvars aren't server-specific, just use a datadef and adjust to what you want.

YotaXP:
I think there's some miscommunication somewhere. JSON was mentioned because of its ability to preserve properties added by other forks, specifically as an option for future-proofing data stored client-side and shared between all forks. DataDefs are not an option for this. (I believe the ConfigurationManager does its own thing for this very reason.) I'm just gonna drop that idea entirely, since it seems to be mucking things up and I'm feeling anxious.

Focusing on the implementation in this PR as is, how are we looking? I'll add some comments later, as some warning to future contributors feels necessary.

YotaXP:
Yeah, I think I'll drop the metadata ideas for now. Just me going overboard again.

Good catch on the comma thing. Looking at the prototype manager, it looks like ALL characters are valid within a prototype ID. That's kind of yikes, tbh, but so be it. I wrote some code to YAML-encode the list so the value would look something like `APC,Wall,"Incredibly,Cursed\nPrototype"`, but unfortunately, it seems that most of YamlDotNet is **not** sandbox-whitelisted, and I see no way to turn YamlNodes into strings with what is currently available. (DataNodes don't appear to be an option since they don't offer the option of flow-style sequences.)

Gonna have to think about this some more before I resort to escaping commas manually.

YotaXP:
Closing this PR in favor of an implementation backed by the server database. Hang tight.

VerinSenpai:
oh thank god. I almost cried.

PR #35638, opened 2025-03-03
merge stable into april-fools-2025

MilonPL:
title
DO NOT SQUASH MERGE


MilonPL:
wrong branch :trollface:

MilonPL:
ok nvm the unsquashed commits threw me off, this is actually right, but doing this as a PR was a mistake (my browser crashed 3 times already)

slarticodefast:
I will just trust that this is correct.
No way to check a 90k line diff

PR #35647, opened 2025-03-03
Salvage: Cancel expeditions when their coordinate disk is deleted

LaCumbiaDelCoronavirus:
## About the PR
Allows salvage to claim a new expedition if the coordinate disk for an existing mission mysteriously disappears.

## Why / Balance
Not being able to do expeditions forever when a shuttle console with the coords disk inside is destroyed is very lame.

## Technical details
Adds a new component: SalvageMissionDependencyComponent
Upon an entity with this component being shut down, it's associated expedition map is deleted, which cancels the salvage expedition. This only occurs if the map is paused; so it won't delete the map when someone's doing the exped.

## Media
https://www.youtube.com/watch?v=Fogk5gbAA30
Tested deleting the disk after the expedition is finished naturally offscreen
Scenarios tested in video:
1. Deleting disk before exped starts
2. Deleting disk right before FTLing to exped (this yields an error, but thats related to FTL system)
3. Deleting disk while on exped

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Salvage no longer gets permanently softlocked from expeditions when their coordinate disk vanishes.

### DIFF
--- a/Content.Server/Salvage/SalvageSystem.ExpeditionConsole.cs
+++ b/Content.Server/Salvage/SalvageSystem.ExpeditionConsole.cs
@@ -17,7 +17,6 @@ public sealed partial class SalvageSystem
     public static readonly EntProtoId CoordinatesDisk = "CoordinatesDisk";
     public static readonly ProtoId<LocalizedDatasetPrototype> PlanetNames = "NamesBorer";
 
-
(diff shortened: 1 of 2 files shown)


LaCumbiaDelCoronavirus:
i meant to add that commit before making the PR - PR description is still accurate

LaCumbiaDelCoronavirus:
Actually this approach is rather stupid and the core issue is already fixed so I don't think this is necessary.

PR #35648, opened 2025-03-03
Players with unknown playtimes now are tagged as new players, take 2

Errant-4:
## About the PR
Built on top of the reverted #35564
Connected players with NULL playtime are now (again) shown as New

Additionally, connected players who have entered the round are now listed above connected players who are still in the lobby. This, prevents brand new players still just looking at options/controls or setting up their character from being placed above antagonists in the round.
Disconnected players no longer show null playtime, so new players now show the clock icon even after they disconnect.

The full Ahelp window sorting is as follows:

Pinned users
Any users with unread messages
Users who are (or were) in the current round, and have sent recent messages
Connected users who...
...joined round, are new, and are antag
...joined round, are new
...joined round, are antag
...joined round
...are in lobby, are new
...are in lobby
Disconnected users who...
...joined round
...never left lobby

## Why / Balance
First version was made to improve the reliability of the New Player marker in the ahelp window
The first version did not sort any null playtime players as new, it just marked them as such. It would also mark any and every disconnected player as New. These two glitches together resulted in a confusing display

## Technical details
```The playtime of players is sometimes null (This sometimes can occur when the players are first joining the server but can also randomly happen to older players as well?). This doesn't fix the root issue, but when it does occur it will tag these players as having unknown playtime and the new player icon in the ahelp menu```

All "Is the player new" checks are now using a single codepath. Disconnected players with NULL playtime are no longer considered New

Previously, a disconnecting player sent a null session info as their last update, which wiped out their playtime ((as far as playerInfo was concerned)). Their playtime info would only update on their next reconnect. This means that, as far as the new system was concerned, every disconnected player would have been "new", since they had null playtime info.
PlayerInfo now uses the players last known playtime if an update sends null, so disconnected players will in most circumstances have some value for IsNewPlayer() to work with

## Media

Screenshot_2025-03-05_103950

## Requirements

## Breaking changes
None

**Changelog**
:cl: Beck Thompson, Errant
ADMIN:
- fix: Brand new players with NULL playtime data are no longer missing the clock icon in the Bwoink menu. Disconnected new players also no longer lose their clock icon.
- tweak: Connected players who have entered the round are now sorted above connected players who are still in the lobby. Players still in the lobby have a half-full symbol.

### DIFF
--- a/Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
+++ b/Content.Client/Administration/UI/Bwoink/BwoinkControl.xaml.cs
@@ -108,31 +108,37 @@ namespace Content.Client.Administration.UI.Bwoink
                 if (a.IsPinned != b.IsPinned)
                     return a.IsPinned ? -1 : 1;
 
-                // First, sort by unread. Any chat with unread messages appears first.
+                // Then, any chat with unread messages.
                 var aUnread = ach.Unread > 0;
                 var bUnread = bch.Unread > 0;
                 if (aUnread != bUnread)
                     return aUnread ? -1 : 1;
 
-                // Sort by recent messages during the current round.
+                // Then, any chat with recent messages from the current round
                 var aRecent = a.ActiveThisRound && ach.LastMessage != DateTime.MinValue;
                 var bRecent = b.ActiveThisRound && bch.LastMessage != DateTime.MinValue;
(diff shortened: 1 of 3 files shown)


beck-thompson (approved):
Looks good! Its hard to test stuff but it seems to work and the code all makes sense :+1:

PR #30586, opened 2024-08-02
Small strange jackboots buff

EmoGarbage404:
## About the PR

Makes jackboots reduce slowness from SlowOnDamage by 50%.

In a practical sense:
- Light damage (>40) goes from a 30% slowdown to a 15% slowdown.
- Heavy damage (>80) goes from a 50% slowdown to a 25% slowdown.

Removes combat boots from the sec loadout as this makes them objectively worse and turns them into a silly noob trap.

Credit to Bhijn & Myr for the idea

## Why / Balance

This PR is primarily aimed to help reduce the prevalence of traitor contraband abuse from security players.

NoSlips are one of the most over represented items when it comes to syndicate contraband being illegally used by sec. This is largely because of a distinct power vacuum when it comes to footwear: there are basically only 3 options commonly available
- Regular shoes, which are legal and everyone has by default
- Magboots, which are legal but have a shitty combat downside (reduced speed) in exchange for the antigrav and noslip properties.
- NoSlips/galoshes, illegal or unobtainable but have a situationally good combat upside (specifically in the wake of a lot of slip nerfs)

Sec themselves (the only ones who are particularly relevant in terms of contraband abuse in this situation) have no mechanical reason to _not_ use noslips, even though the rules prohibit it and the upside is relatively minor. Compared to no upside or even a combat downside, noslips are an obvious choice.

My solution for this--this PR--is to buff jackboots with a _generally useful_ upside to serve as a natural alternative to the _situational utility_ of noslips. Reducing sec's slowdown in combat situations when injured is great for allowing them to carry on fighting in tense situations as well as escaping ambushes. 

I don't think this is a 100% silver bullet to the entire situation, but my hope is that it'll serve as an alternative that will at least greatly reduce sec's usage of syndicate equipment, or at the very least, make the usage of said equipment into an interesting tactical decision rather than a flagrant showing-off of contraband.

## Technical details

Adds a new `ClothingSlowOnDamageModifierComponent` and some events to hook into `SlowOnDamageSystem`.

## Media

image

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: Jackboots now reduce slowness from injuries by 50%.
[...]

PR #35653, opened 2025-03-03
Injury slowdown resistance for combat and merc boots

Nox38:
## About the PR
Adds the jackboots injury slowdown resistance to the combat boots and mercenary boots.

## Why / Balance
It doesn't make sense for only the jackboots to have this effect, when the other differences between the boots are cosmetic only. In addition, this buff gives nukies a reason (although a small one) to use their default combat boots.

## Requirements

**Changelog**
:cl: Nox38
- tweak: All military boots (jack, combat, merc) now have the 50% injury slowdown resistance effect.


shamblestf:
I think people will always just default to winter boots with this since they'll always be better for cold resistance

Nairodian:
Obviously things change and this may not be a problem anymore but as per the original PR #30586, combat boots didn't get the effect as to avoid nukies getting the buff.

> > Love the idea, but any particular reason it applies only to jackboots and not to combat boots as well?
> 
> Jackboots are only found in sec whereas combat boots are worn by nukies and syndies.

Cadunkus:
> I think people will always just default to winter boots with this since they'll always be better for cold resistance

I mean I still see officers wearing the winter coat despite it having almost no armor and no carry slot for guns on the back. Sometimes we do have form over function.

MilenVolf:
You forgot to remove the "-->" symbol from changelog

Cojoke-dot:
So in the og pr discussion for this, combat boots were left out for a reason(#30586). I don't know where syndie winter boots appear in the game at all nor why they would get the buff, they aren't considered military shoes. Maybe make them a child of Military boots base or something. I agree with the rest though, nukies should have the choice of noslips, mags, or combat boots slowdown stuff

Nox38:
> I don't know where syndie winter boots appear in the game at all nor why they would get the buff, they aren't considered military shoes. Maybe make them a child of Military boots base or something. I agree with the rest though, nukies should have the choice of noslips, mags, or combat boots slowdown stuff

There, removed the effect from sec and syndie winter boots.

superjj18:
This passive is kind of dumb in general.

Should either be innate or deleted imo

Nox38:
This PR has not seen the light of day in ages, so I got it modernized and ready to go

SlamBamActionman (approved):

SlamBamActionman:
Small change but we like consistency. No real impact on balance.

ProPeperos:
small buff for nukies tho, they start with combat boots

PR #35658, opened 2025-03-04
DetGadget hat fixes

ArtisticRoomba:
## About the PR
"""Refactors""" the DetGadget hat extraction logic to be 12% less horrible to look at.

The hat will now explicitly check the requested items' label, if it exists, and compare against that when serving requests. This means that you can now label your items before you put them in, to expedite retrieving items.

For example, you can rename your explosive grenade just "grenade" and the hat will pull it out when you say "go go gadget grenade". Similarly, you can name a .30 ammo box just "ammo" and it will extract that as well.

## Why / Balance
It's really annoying to say the longhand of every single name. In addition, putting a label on a hat can make the item very hard to retrieve as you'd have to say the full name of the item, with label.

This also makes the hat easier to use for species with accents. For example, a forensic scanner might be worded as "Ssscanner" if spoken by a reptilian, which would be completely dodged by the string comparison. Now, you can name your scanner "Ssscanner" and it'll pick up on it just fine. This issue is a larger core issue behind TriggerSystem.Voice, which can be easily fixed by Slamchat when that hits the repo.

## Technical details
Extracting items has been moved into a method that gets called when we want to extract an item.

The logic for extracting an item has changed. When the system iterates over all of the items currently in the attached storage:
- It checks to see if an item has a `NameModifierComponent`. If true, it'll run a comparison on the original name of the item. If it finds the item it wants, it'll call the extraction method.
- The above check repeats exactly the same for `LabelComponent` if applicable.
- If it doesn't have a `LabelComponent` or a `NameModifierComponent`, a simple check of the name will be done against the message and it will call the extraction method if true.

## Media
https://github.com/user-attachments/assets/7db7e1c0-718e-40c1-8df3-075541854813

## Requirements

**Changelog**
:cl:
- tweak: The Go Go Hat now checks if your message matches either the name of the item or the label of the item. This allows people with accents to name their items according to how their character would say the item, or for a faster retrieval without having to type out the full name.

### DIFF
--- a/Content.Server/VoiceTrigger/StorageVoiceControlSystem.cs
+++ b/Content.Server/VoiceTrigger/StorageVoiceControlSystem.cs
@@ -4,8 +4,6 @@ using Content.Shared.Administration.Logs;
 using Content.Shared.Database;
 using Content.Shared.Hands.Components;
 using Content.Shared.Inventory;
-using Content.Shared.Labels.EntitySystems;
-using Content.Shared.NameModifier.EntitySystems;
 using Content.Shared.Popups;
 using Content.Shared.Storage;
 using Robust.Server.Containers;
@@ -23,8 +21,6 @@ public sealed class StorageVoiceControlSystem : EntitySystem
     [Dependency] private readonly InventorySystem _inventory = default!;
     [Dependency] private readonly SharedPopupSystem _popup = default!;
     [Dependency] private readonly StorageSystem _storage = default!;
-    [Dependency] private readonly SharedLabelSystem _label = default!;
-    [Dependency] private readonly NameModifierSystem _nameModifier = default!;
 
     public override void Initialize()
     {
@@ -38,10 +34,6 @@ public sealed class StorageVoiceControlSystem : EntitySystem
         // If it has slot restrictions, check if the item is in a slot that is allowed
         if (ent.Comp.AllowedSlots != null && _inventory.TryGetContainingSlot(ent.Owner, out var itemSlot) &&
             (itemSlot.SlotFlags & ent.Comp.AllowedSlots) == 0)
-            return;
-
-        // Don't do anything if there is no message
-        if (args.Message == null)
             return;
 
         // Get the storage component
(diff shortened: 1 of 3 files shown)


ArtisticRoomba:
Marked as P2 because pulling ammo out of the hat is sisyphean and lizards straight up can't use the hat if they want to pull out something like a det scanner (ssscanner)

Content.Server/VoiceTrigger/StorageVoiceControlSystem.cs
```diff
@@ -79,20 +81,54 @@ private void VoiceTriggered(Entity<StorageVoiceControlComponent> ent, ref VoiceT
...
+                ExtractItemFromStorage(ent, item, args.Source, hands);
                 break;
             }
         }
     }
+
+    /// <summary>
+    /// Extracts an item from storage and places it into the player's hands.
+    /// </summary>
+    /// <param name="ent">The entity of the item with the StorageVoiceControlComponent</param>
```
beck-thompson, line 119: Reference `StorageVoiceControlComponent` so it links!

Content.Server/VoiceTrigger/StorageVoiceControlSystem.cs
```diff
@@ -79,20 +81,54 @@ private void VoiceTriggered(Entity<StorageVoiceControlComponent> ent, ref VoiceT
...
+    /// <param name="item">The entity of the item to be extracted from the attached storage</param>
```
beck-thompson, line 120: Just say "The entity to be extracted from the attached storage"
same for the one below, just say entity!

Content.Server/VoiceTrigger/StorageVoiceControlSystem.cs
```diff
@@ -79,20 +81,54 @@ private void VoiceTriggered(Entity<StorageVoiceControlComponent> ent, ref VoiceT
...
+            // Some items could be labeled, or their names otherwise modified
+            // If they are, we need to check if the original name or the modified name is in the message
 
-            // We found the item we want, so draw it from storage and place it into the player's hands
-            if (storage.Container.ContainedEntities.Count != 0)
+            // Check and pull the NameModifierComponent for the original item name and compare
+            // We explicitly check this by itself
+            // and not check both the NameModifierComponent and the LabelComponent
+            // at the same time because something else could have modified the name
+            if (TryComp<NameModifierComponent>(item, out var itemNameModifierComponent) &&
```
beck-thompson, line 91: Do what we talked about on discord for this stuff

beck-thompson (changes requested):

beck-thompson (approved):
Looks good! We talked about it for a while and we think matching the name with any part of the message is fine. Otherwise we would have to do some weird checks (that could also be gamed anyway).

PR #35666, opened 2025-03-04
Open State for cowtools

ps3moira:
## About the PR

Made open state for cowtools
## Why / Balance

So giant big red error text doesn't appear

## Media

Screenshot 2025-03-04 102509


slarticodefast:
Thanks for the quick bugfix! I will just merge for now, since it is only a minor change to the existing sprite.

Issue #35534, opened 2025-02-26
Gun safes have same health as regular closets


K-Dynamic:
## Description

See title. This is due to gun safes parenting from LockerBaseSecure without changing health. Notably, closets are made up of four steel, while gun safes are made up of 10 steel, 5 LV cable and 10 plasteel.

**Reproduction**

1. Spawn closet and gun safe next to each other.
2. Equip a fire axe or any weapon with high structural damage.
3. Wide swing to hit both closet and gun safe.
4. Destroy both in the same number of hits.

Killerqu00:
Information for beginner contributors if they want to take that issue: the component that manages locker durability is DestructibleComponent.

ukade:
I tried reproducing it, but gun safe and closet don't take any damage.

ukade:
nvm, I tried using smite w lightning. gunsafe destroyed in 3 hits and closet destroyed in 5

ukade:
I wonder if I can get some help in understanding this as I am learning. 

So, I thought the easiest way to fix this was to find the object in question and change the stats. I found the YAML file, but I see that it doesn't have health stats or anything similar. KDynamic mentioned it's related to the Parent, LockerBaseSecure, but can't find it. 

Killer drops a hint with DestructibleComponent. So from my understanding, it seems that objects are assigned an entity id and then in this file, a component is assigned to the id to allow it to take damage. However, I still don't understand how the values/ stats of each object is getting obtained, as I don't see such info in the YAML.

Please bestow upon me your infinite knowledge.

Pronana:
> So, I thought the easiest way to fix this was to find the object in question and change the stats. I found the YAML file, but I see that it doesn't have health stats or anything similar. KDynamic mentioned it's related to the Parent, LockerBaseSecure, but can't find it.

The parents are in base_structurelockers.yml, this doesn't seem like a hard first PR if you want to try it. These inherit from ClosetBase which is in base_structurelockers.yml

ukade:
Oh...

PR #35671, opened 2025-03-04
Make Closets Less Tanky Than Gun Safes

CraftyRenter:
## About the PR

Changed the `DamageTrigger` values of `ClosetBase`, `LockerBase`, and `LockerBaseSecure` to 50, 75, and 150 respectively.

## Why / Balance

Issue #35534 states that closets and gun safes have the same health. This isn't actually the case, closets have MORE health than gun safes currently. It currently takes more hits from a fire axe to break a chef's or janitor's closet than it does to break an armory locker or captains locker, which doesn't make sense from a gameplay perspective (shouldn't a big beefy locker be tougher to break than a little closet?), or an RP perspective (why would NT make their closets strong but their command staff lockers weak?).

This change intends to improve the health scaling of closets and lockers.

Post change, closets break after 2 fire axe hits, lockers break after 3, and secure lockers break after 5.

## Technical details

Basically covered in the About the PR section with the note that I added a second damage trigger to `LockerBaseSecure`  which makes a noise when its destroyed at the normal threshhold like regular lockers and closets.

## Media

This is a tiny no c# change so I've not included any media, I have verified the changes work in game.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Storage/Closets/Lockers/base_structurelockers.yml
+++ b/Resources/Prototypes/Entities/Structures/Storage/Closets/Lockers/base_structurelockers.yml
@@ -29,13 +29,13 @@
     thresholds:
     - trigger:
         !type:DamageTrigger
-        damage: 100
+        damage: 150
       behaviors:
         - !type:DoActsBehavior
           acts: [ "Destruction" ]
     - trigger:
         !type:DamageTrigger
-        damage: 50
+        damage: 75
       behaviors:
       - !type:DoActsBehavior
         acts: ["Destruction"]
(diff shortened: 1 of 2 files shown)


Killerqu00 (approved):
tested, works
:+1:

Booblesnoot42:
I noticed an odd thing while looking at the code that isn't directly related to this pr but could be another way of solving the issue. 

Aside from the health of the entity, another factor that affects how easy they are to destroy are its damage modifiers, which are defined in the DamageableComponent. When hit by an attack of a given type, these modify the amount of damage before it is applied to the health. 

I was checking what modifiers the lockers and closets have, and noticed that the base closet uses the "StructuralMetallic" modifier. This is then inherited by the base locker and the base secure locker, except it looks like the base secure locker tries to change the modifier; it has a damageable component which sets the modifier (line 63). The weird thing is that it doesn't actually change anything about the component, it just sets the modifier to "StructuralMetallic" again. 

Looking at the blame, it seems like both of these modifiers were changed in late 2024, with the secure locker being changed a few days after the closet. If I had to guess, I think the secure locker's modifier was supposed to be set to "StructuralMetallicStrong," but was set to "StructuralMetallic" by mistake.

Princess-Cheeseballs:
Pushed some changes, Slam has given approval and I'm giving approval will merge when tests pass.

PR #32136, opened 2024-09-14
Implanter draw rework

SlamBamActionman:
## About the PR

This PR changes how the implanter works when it comes to drawing/extracting subdermal implants for an entity. The purpose of this PR is to remove the need for metashield rules around implants, and instead use mechanics to achieve the desired effects. 

This is in line with the changes proposed in the design doc PR "Reduced Metagaming Mechanics" under "The Implant Problem": https://github.com/space-wizards/docs/pull/214 

The PR does the following:
- Implanters in Draw mode must now have a specific type of implant set.
  - The implant is chosen from a dropdown list of all available implants the implanter can draw.
- When an implant is used, it will only attempt to draw out the implant it is set for.
- If the implant exists, it gets deposited into the implanter as it works currently (or the draw fails, as is the case with acidifiers/mindshields). There is no change to this process.
- If the implant doesn't exist, the implanter experiences a catastrophical failure and chunts genetic material into the user's hand (the hand of the person *performing* the implant check, not the person *being* checked). This deals major damage to the user.
  - This damage is set to 50 Cellular, 10 Heat in this PR.
- Implant draw time is reduced from 60 seconds to 25 seconds.

After this PR is merged, implants and implant knowledge should be removed from the metashield.

## Why / Balance

Implanters are one of the few "metashield" items that exist in the game, and the cause of this is quite apparent:
- If implanters weren't metashielded, Security could randomly check crew at any point, for no reason.
- Random checks would be very efficient since you get any implant that they have on them.
- It's not that costly to do random checks. It's 60 seconds and once done you've confirmed they have zero implants, or you catch one.

As such, the metashield specifically requires either finding an implanter or seeing *undisputable* use of an implant before being able to check for it in an individual. This can lead to unnecessary admin work where it's unclear what exactly counts as undisputable, and players can be uncertain of whether they are allowed or not to perform an implant check. Conversely, a player may heavily suspect a character has an implant, but feel they are not allowed to check due to the rules. This PR intends to fix these issues.

1. Requiring specificity means SecOffs can not check randomly and just hoping to fish out an implant. They must at the very least suspect that a specific implant is in play to know to check for it.

2. This specificity is enforced via the damage taken when guessing wrong. SecOffs do not have the resources to perform random checks since it's directly harmful to the Security department and the personal cost to the SecOff performing the implant check discourages this on an individual level. It's simply not viable to randomly check crew due to the damage taken.

[...]

PR #35672, opened 2025-03-05
Unclean implanters have cleanable information about what was contained in them

poklj:
## About the PR

added back the Meta information to Implanters to better support unclean implanter interactions for #32136 and added extra system work to support removing that meta information should a syndicate clean up after themselves.

## Why / Balance

Prior to the Implanter Draw Rework, Metashield generally had 2 methods of defeating an implant:
Witness someone using the implant
Find an unclean implanter with DNA on it.

Now with mechanical enforcement, an unclean implanter is a frustrating endeavor as you must let someone who you know has an implant run around with the syndicate contraband without being able to directly remove it, or Actually attempt to remove that implant due to not knowing what it is.

With this if a Syndicate fails to clean up their evidence, Secoffs will be able to both attempt the draw and have a direct knowledge of which one to draw (Something that a syndicate can counterplay by drawing it themselves, so potentially dangerous regardless) rather than needing to "Rainbow" every implant or do a 2x50 genetic damage testing popular implants.

It however doesn't reveal "what" implants are in play if a syndicate cleans up after themselves, which keeps and ensures detective work must be done if even a trivial amount of diligence is taken.

Additionally: Added a Small reminder to clean your evidence with soap to the implanter, to hopefully inform New syndicates they can clean it.

## Technical details

Added code to inform the implanter what implant is inside when created and rename itself, additionally adds cleaning code and raises events from forensics system when the DoAfter completes.

Needed to change the Implanter name `implanter` due to how `metadataSystem` blocks renaming an entity to it's entity name when done via NameModifierSystem

## Media

https://github.com/user-attachments/assets/e98d9b3f-b108-469e-9eda-e13b056b20c9

## Requirements

## Breaking changes

**Changelog**

:cl: Kickguy223
- tweak: re-added information about what implant was in an implanter
- add: cleaning a implanter also removes what was in the implanter

### DIFF
--- a/Content.Server/Forensics/Systems/ForensicsSystem.cs
+++ b/Content.Server/Forensics/Systems/ForensicsSystem.cs
@@ -263,6 +263,9 @@ namespace Content.Server.Forensics
 
             if (TryComp<ResidueComponent>(args.Used, out var residue))
                 targetComp.Residues.Add(string.IsNullOrEmpty(residue.ResidueColor) ? Loc.GetString("forensic-residue", ("adjective", residue.ResidueAdjective)) : Loc.GetString("forensic-residue-colored", ("color", residue.ResidueColor), ("adjective", residue.ResidueAdjective)));
+;
+            var foresnicsCleanedEvent = new ForensicsCleanedEvent();
+            RaiseLocalEvent(args.Target!.Value, ref foresnicsCleanedEvent);
         }
 
         public string GenerateFingerprint()
--- a/Content.Shared/Forensics/Events.cs
+++ b/Content.Shared/Forensics/Events.cs
@@ -31,6 +31,14 @@ public sealed partial class CleanForensicsDoAfterEvent : SimpleDoAfterEvent
 }
 
 /// <summary>
+/// Raised when a Successful clean doafter has occured.
+/// </summary>
+[ByRefEvent]
+public record struct ForensicsCleanedEvent()
+{
+};
+
+/// <summary>
 /// An event to apply DNA evidence from a donor onto some recipient.
 /// </summary>
 [ByRefEvent]
--- a/Content.Shared/Implants/Components/ImplanterComponent.cs
+++ b/Content.Shared/Implants/Components/ImplanterComponent.cs
@@ -107,6 +107,12 @@ public sealed partial class ImplanterComponent : Component
     [AutoNetworkedField]
     public EntProtoId? DeimplantChosen = null;
 
+    /// <summary>
+    /// Stores the current stored implant name, to be cleared if someone cleans the evidence off of the implanter.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public LocId? CurrentImplanterLoc;
+
     public bool UiUpdateNeeded;
 }
 
--- a/Content.Shared/Implants/Components/SubdermalImplantComponent.cs
+++ b/Content.Shared/Implants/Components/SubdermalImplantComponent.cs
@@ -56,6 +56,12 @@ public sealed partial class SubdermalImplantComponent : Component
     /// </summary>
     [DataField]
     public EntProtoId? DrawableProtoIdOverride;
+
+    /// <summary>
+    /// The Name of localization id of the implant for naming the Implanter
+    /// </summary>
+    [DataField]
+    public LocId ImplantLoc;
 }
 
 /// <summary>
--- a/Content.Shared/Implants/SharedImplanterSystem.cs
+++ b/Content.Shared/Implants/SharedImplanterSystem.cs
@@ -8,6 +8,7 @@ using Content.Shared.Forensics;
 using Content.Shared.IdentityManagement;
 using Content.Shared.Implants.Components;
 using Content.Shared.Interaction.Events;
+using Content.Shared.NameModifier.EntitySystems;
 using Content.Shared.Popups;
 using Content.Shared.Verbs;
 using Content.Shared.Whitelist;
@@ -28,6 +29,7 @@ public abstract class SharedImplanterSystem : EntitySystem
     [Dependency] private readonly DamageableSystem _damageableSystem = default!;
     [Dependency] private readonly SharedUserInterfaceSystem _uiSystem = default!;
     [Dependency] private readonly IPrototypeManager _proto = default!;
+    [Dependency] private readonly NameModifierSystem _nameModifier = default!;
 
     public override void Initialize()
     {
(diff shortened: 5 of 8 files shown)


ThatOneGoblin25:
> description: A compact disposable syringe exclusively designed for the injection of subdermal implants. **Remember to use soap to clean the evidence!**

This kinda goes against what a "description" is.

Pronana:
This seems like a genuinely really good change. Now if you have DNA on an implanter you can pretty reliably remove it, and if you don't then you are neither sure who is implanted nor what they were implanted with (much like how it previously worked).

shamblestf:
Fucking hero thank you

Mixelz:
So, if someone were to clean the implant of dna. But then pick it back up again without gloves, wouldn't that lead us right back to square one? A no lable implant with the persons DNA on it as a dead ringer?

Granted it would be a self sabotage move on their part but still.

Pronana:
> So, if someone were to clean the implant of dna. But then pick it back up again without gloves, wouldn't that lead us right back to square one? A no lable implant with the persons DNA on it as a dead ringer?

This is enough of an edge case that I don't think it would hurt security too bad. The current issue is a billion implants come by security and there's no reasonable way to deal with them. This would be a mistake probably done by one or two people at most during a round who you can mark sus as compared to the guaranteed two or three whom you can't do anything about currently.

ThatOneGoblin25:
> So, if someone were to clean the implant of dna. But then pick it back up again without gloves, wouldn't that lead us right back to square one? A no lable implant with the persons DNA on it as a dead ringer?
> 
> Granted it would be a self sabotage move on their part but still.

You can right-click on injectors that are inside a container to clean them. Stuff like o2 locker toolboxes or maint locker jackets can hide it while plenty of people will touch the lockers to look for loot.

Mixelz:
> This is enough of an edge case that I don't think it would hurt security too bad. The current issue is a billion implants come by security and there's no reasonable way to deal with them. This would be a mistake probably done by one or two people at most during a round who you can mark sus as compared to the guaranteed two or three whom you can't do anything about currently.

I agree, however I'm still on the side of there needing to be some sort of middle ground in regards so being able to narrow down the pool of possible implants. You either know *exactly* which one it is or have to guess between literally the ENTIRE pool of possible implant choices. 

I don't think that sort of all or nothing dychotomy is healthy for the gamestate, since it leads to a binary state of the syndicate agent either being "robust" enough to never be caught or directly revealing themselves to the detective in one of the most obvious manners possible. There should be a grey area instead where the det needs to narrow things down from a list of possibilities, which is the entire point of the role and is currently impossible without greater nuance to the implant mechanic.

ArtisticRoomba:
I'd rather the "you can clean this" in the description be omitted so the syndie learns the hard way to clean up after themselves. It provides a brash first exposure to the power of the detective and it'll make them question how to cover their tracks in the future.

Suffering builds character.

MozarteanChaos:
> So, if someone were to clean the implant of dna. But then pick it back up again without gloves, wouldn't that lead us right back to square one? A no lable implant with the persons DNA on it as a dead ringer?
> 
> Granted it would be a self sabotage move on their part but still.

It'd have fingerprints, not DNA - leaving your prints on an implanter is incriminating, but doesn't prove that you yourself used it.

Pumkin69:
On the surface level just from thinking about how this PR would work it seems to fix the issues i have personally seen come from the implanter rework so i think its good.

walksanatora:
i personally think that cleaning the implanter shouldn't remove what kind of implant it is (something about "each implanter has a distinct shape depending on where you are supposed to put it"). you removing your DNA off of it is allready enough. thatway instead of finding a empty implanter and being like "great we have no clue what to even begin searching for" you have a idea of what to look for.

walksanatora:
(ofc I am also of the opnion that you should not be able to clean out a implanter with soap since implanter supposedly has to go into your skin making it get your dna all over the syringe part)

Lyroth001:
> I'd rather the "you can clean this" in the description be omitted so the syndie learns the hard way to clean up after themselves. It provides a brash first exposure to the power of the detective and it'll make them question how to cover their tracks in the future.
> 
> Suffering builds character.

Agreed. Imo having a reminder there every time defeats the point of this change for the vast majority of implants sec will find, those sorts of slips and mechanical knowledges are part of the gameplay for both sec and traitors, since these roles really rely on game knowledge (and also what det doesn't want to gloat about how they caught them, which can end up teaching them anyways, and could have some good RP potential). I wouldn't be opposed to it being in the guidebook at some point tho

>You either know exactly which one it is or have to guess between literally the ENTIRE pool of possible implant choices. 

Honestly a system that provides a pool of random implants it could be to narrow things down a bit could be a good inbetween, maybe if its cleaned it gives a few options idk

Mixelz:
> Honestly a system that provides a pool of random implants it could be to narrow things down a bit could be a good inbetween, maybe if its cleaned it gives a few options idk

My own suggestion is to group all implants into Type A/B/C categories, ensuring you have at least 8 or so options per category. That's it.

The hard part would be finding groupings that leave enough plausible denyability between options while still having the goal of being a help in narrowing down the pool of all options.

As an example: Storage and Uplink implants could both be the same type because they can achieve very similar results to an outside observer (pulling out items from thin air). But it may instead be better for balance if they are not in the same category as it may make the capabilities of the given traitor too obvious. 

That being said, this should maybe be figured out on a seperated issue/PR.

slarticodefast:
Hey, this had a long maintainer discussion and we decided to close this PR for the following reasons:
- With this implementation you will still land in the same situation as before, where you *have* to remove the label with soap, making the existing noob trap of not knowing this step even worse. And if the label has been removed there is no difference to the current mechanics.
- Rounds are generally more engaging if antags are allowed a chance to utilize their equipment. Figuring out who used an implanter, even if you can't deduce the implant, allows Security to react better once the implant is used. If they find one with dna on it their job is to deduce which type of implant was used, or to observe the suspect and catch them using it.
- It mixes up forensics and item naming, which is inconsistent with how item names usually work. Names are something you can directly see when examined. Forensics are invisible without a scanner.
- The unclear formulation in the current space law which was interpreted by some players as "you can hold prisoners until their implants have been removed" was rewritten with admin approval https://github.com/space-wizards/space-station-14/pull/35701

We will further observe this issue together with the admins and will consider re-adding permanent labels in the future.

Pumkin69:
man....

poklj:
> Hey, this had a long maintainer discussion and we decided to close this PR for the following reasons:
> 
>     * With this implementation you will still land in the same situation as before, where you _have_ to remove the label with soap, making the existing noob trap of not knowing this step even worse. And if the label has been removed there is no difference to the current mechanics.
> 
>     * Rounds are generally more engaging if antags are allowed a chance to utilize their equipment. Figuring out who used an implanter, even if you can't deduce the implant, allows Security to react better once the implant is used. If they find one with dna on it their job is to deduce which type of implant was used, or to observe the suspect and catch them using it.
> 
>     * It mixes up forensics and item naming, which is inconsistent with how item names usually work. Names are something you can directly see when examined. Forensics are invisible without a scanner.
> 
>     * The unclear formulation in the current space law which was interpreted by some players as "you can hold prisoners until their implants have been removed" was rewritten with admin approval [Update Space Law to reflect Implant changes #35701](https://github.com/space-wizards/space-station-14/pull/35701)
> 
> 
> We will further observe this issue together with the admins and will consider re-adding permanent labels in the future. Thank you for your contribution.

A few questions, as a sanity check:

1: how is this any different from PDA uplinks "as a noob trap", and why is it that we cannot rectify that noob trap via other means?

2: they have a chance to utilize their equipment, they just need to take a modicum of effort to actually cover their tracks, of which 99% of their other equipment has no need to do anything as such (Noctpen being a literal instant win? With no additional counterplay.) if we're codifying that implanters are a no counter play (beyond "syndicate did it infront of you" which is the only way to do this under new spacelaw), why have an implanter stay around at all, because at the end of the day, you have to see it. there is no other way to play it, period, so why care about the implanter.

3: Is there not another way that could be put forward that isn't confusing the system? A verb to rip the label off? a second Z press on the item? anything of that sort?

4: The issue with the space law changes are that it doesn't fundementally solve the problem that Secoffs were originally creating those tactics around, which was "we know someone has an implant, without a shadow of a doubt, but we _have no means to know which one_.

Mixelz:
> If they find one with dna on it their job is to deduce which type of implant was used, or to observe the suspect and catch them using it.

It is not possible to deduce the type of implant that a person has as an outside observer. 
I am not being exaggerative, there are currently **no** ingame mechanics that facillitate this deduction proccess or even give a jumping off point at all. The entire point of a subdermal implant is that, to an outside observer, someone who has the implant looks exactly the same as someone who does not have the implant. There are no tells for the implant for an outside source and the only piece of evidence that there is to it's existence would be finding a used implant in the first place. 

In the scenario where they saw the implant being used, such as someone using a freedom impant while cuffed, then there's no more deduction. You already know **exactly** which implant was used because you saw its power set.  There would need to be some sort of change or implementation of sub-dividing out implants (Such as my prior A/B/C type suggestion) to falliciate this behavior.

While the idea of allowing antags a chance to use their implant is potentially healthy for the game, implants are not a mechanic that facillitates this as attempting to remove the implant is a preventitive measure, not a reactionary one. If the implant is already expended, there is nothing to remove and thus there is no purpose to an implant remover. Most traitors would also rather spam all available implants as a desperate last ditch than allow them to be removed for them by the prospective security force.

It is quite literally not possible for a detective to use deductive reasoning on implants as a mechanic or allow a scenario in which an implant can be used unknown to security but still also be removed, full stop.

Pumkin69:
I really don't understand why these goalposts are getting moved so much during this discussion.

The original PR was supposed to lessen admin workload and remove the metashield.

I have no idea where this discussion of "allowing syndicate agents to use their utility" came from because before they were using it TOO MUCH. Acting as if syndicate agents can't take less than 10 seconds and EASY to obtain items to completely wipe their evidence off the implanter causing security to jump through hoops to figure out what you had, OR catch you because you were being dumb with it and using it as a free out. 

NOW with the changes being implemented you are actively punishing security for doing their job AND BENEFITING the antag for NOT wiping their evidence off the implanter because security literally has no counterplay to it.

TLDR
Stop moving goalposts around this discussion.
Syndicate agents already had a way to avoid implant detection via soap or damp rags before.
DNA wiping isn't anymore of a "noob trap" than anything else syndicates can do.
The original intention has been completely lost in discussion and implanters are now EXTREMELY STRONG and security cant do anything about them anymore.

deltanedas:
> they have to guess what implant you have

better yet you can make them always fail by removing it yourself so its impossible for them to remove anything

Issue #24984, opened 2024-02-06
Blunt damage should deal damage to structures


Anzarot121:
## Description

I’m a skeleton, I’m in a small station’s armory, from where the salvagers took away all the firearms. Previously, after 10 minutes of effort, I could get out of here by breaking down the door, now this is inaccessible to me, I don’t cause damage to anything.
**Reproduction**

Try to break down the door using an air tank.
**Screenshots**

![image](https://github.com/space-wizards/space-station-14/assets/116586144/a709ab8d-71d0-4cf5-ba24-9e838c3b2322)

vero5123:
just look into the yml for the given object. Self explanatory from there.

ThatGuyUSA:
Airlocks have a damage threshold of 10 meaning it won't take any damage from sources that deal less than 10 damage.
Exceptions to this rule are structural damage, and anything that ignores armor.
Something that deals 10.1 damage (heat, blunt, pierce, slash, etc...) would start breaking down an airlock. 
(albeit very slowly, dealing 0.07 damage per hit if you're using blunt... out of 500 which is needed to destroy it)

I can only assume this was an intentional decision, seeing as an un-wielded bat deals 10 blunt, and 5 structural which allows it to tear down doors no problem.

For this reason, and that the post is really old... I will be closing it as unplanned!

PR #35679, opened 2025-03-06
Increased blunt damage of parent GasTankBase

ukade:
## About the PR
Increased blunt damage of gas tanks by tweaking parent GasTankRoundBase.

## Why / Balance
Issues #24984 states that previously was able to escape the room using air tanks to destroy the door after salvagers took all firearms. Player did not appear to have an escape route or any other weapon in a locked room.

## Technical details
Airlocks took on modifiers that had flat reductions against damage less than 10, which is why air tanks did not damage doors.
Increased blunt damage in parent GasTankRoundBase.
Also added Stamina Damage and increase damage on wield.
Only affects OxygenTank, NitrogenTank, AirTank and NitrousOxideTank. There appeared to be some inheritance issues that I did not want touch for this PR.

## Media
no media

## Requirements

## Breaking changes
No breaking changes

**Changelog**
:cl:
- tweak: Increased Damage for Gas Tanks

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Tools/gas_tanks.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/gas_tanks.yml
@@ -39,7 +39,14 @@
     attackRate: 0.8
     damage:
       types:
-        Blunt: 10
+        Blunt: 15
+  - type: StaminaDamageOnHit
+    damage: 5
+  - type: Wieldable
+  - type: IncreaseDamageOnWield
+    damage:
+      types:
+        Blunt: 3
   - type: PhysicalComposition
     materialComposition:
       Steel: 185


Pumkin69:
is this meant for a downstream?

slarticodefast (changes requested):
18 blunt damage is way too much for an item easily obtainable
for comparison:
a crowbar deals 8 blunt
a toolbox deals 12 blunt
a baseball bat deals 15 blunt and 15 structural damage
I would keep the damage for the airtank as it is.
If you want to solve the linked issue you should have a look at how structural damage is calculated. This is not a problem specificly for air tanks.

slarticodefast:
> Consider overall DPS too as air tanks have slower attack speed

0.8 compared to the default 1 second cooldown.
But with the usual wideswing jousting fighting meta this hardly makes a difference.

ukade:
Applying realism, I would argue that an air tank would deal more damage than a baseball bat, but definitely less wieldy.  Will keep blunt damage at 10 and remove all other changes (no stamina damage, remove wieldable, increase damage on wield). Maybe just add structural damage at 12 so it supersedes the flatreduction for structural damage of StructuralMetallicStrong? Or keep blunt damage at 12 and decrease attack rate.

Since the original issue is that the user could not escape due to having no weapons, and having no object that can deal damage to the airlock and also having a character that does no damage, I'm wondering if we want to try a different way of fixing this because I'm not sure if I want to make changes to the modifier set as it might affect other objects that I did not intend.

@Pumkin69 There's some inheritance issues where oxygen is inheriting from nitrogen and all that and one has damage and the other does not, so I wanted to fix that first. Emergency oxygen tanks are smaller (I think) so definitely don't want it to have the same amount of damage.

MilonPL:
Is this being worked on? If so, please address the review.

MilonPL:
I am going to be closing this PR as Stale. If you ever end up coming back to this, please open a new PR.

PR #35684, opened 2025-03-06
Elkridge Tesla and TEG Improvements + Other stuff

Deerstop:
## About the PR

improved tesla setup, it shouldnt explode anymore and its all in proper storage
improved the TEG setup, bringing it more inline with other lowpop plug-n-play tutorial setups
expanded south sci maints a little
added a chef closet to the kitchen so they get cloth
captain now gets a bathroom with the golden toilet for thieves
changed the vault a little so the nuke can be armed inside of it
added a sink to the maints bar
## Why / Balance

the tesla never exploded when i tested it myself but it wasnt a safe setup
with the tesla being easier to setup, i made the TEG better so it was still a competitive option, and clearer for new atmosians
## Technical details

marp
## Media

Elkridge-0

## Requirements

## Breaking changes

**Changelog**


Deerstop:
i should probably mention that in the little communications room beside the new engi storage room, the syndicate uplink, handheld radio, and passenger headset have all had their components removed so they're non-functional

ArtisticRoomba:
if another map converts containment stuff into flatpacks and puts them into boxes im going to lose it

Deerstop:
> if another map converts containment stuff into flatpacks and puts them into boxes im going to lose it

the tesla being a pain in the ass to setup is the single most common complaint i've found in the last month. please do me a favour and offer some actual reasoning or a solution instead of a snide joke

ArtisticRoomba:
Sorry for the witty remark, I'll put my actual reasoning below.
There are multiple parts to why engineering can hate setting up a tesla, which is causing the push for mappers to start mapping flatpacks:

- Some maps have containment equipment tightly packed with no way to retrieve equipment all the way in the back without pulling out everything in the front, which is just a violation of how low-density storage is done in real life.
- Some maps have containment equipment storage mapped beyond a very tiny airlock and around multiple corners. The routes people have to go down are either badly communicated or they're highly non-ergonomic from a station design standpoint. This often leads to people bumping into each other in airlocks and bottlenecks engineering unnecessarily.

Flatpacks solve all of these problems, however from a game design standpoint they are not very good for engineering to have roundstart:
- It trivializes their gameplay. There are a lot of subproblems to this.
   - Engineering can setup an engine in 2 seconds without taking proper care and time to set up the engine, which can lead to more looses. If people have more time, more people can interject. Additionally, if people have more time in general, they will think about their actions a bit more.
   - If engineering can set up the engine in 2 seconds, it provides very little time for a new person to interrupt an experienced veteran and help set up the engine. It's very unlikely for a peep ("Hey, can I help you with this?") to come out of their mouth before all of the emitters are on and the veteran is going back inside to turn the PA on. Ideally engineering teaches by doing, not by seeing.
   - New players don't really know what to do with flatpacks, whereas with real, seen equipment, they can get an idea of what they're putting down or messing with.
- Giving a tier 2 research item roundstart makes the flatpacker and flatpacks no longer special. When you make something a guarantee it no longer becomes an interesting, refreshing thing when something new comes along; you already had it to begin with.

Letting engineering play with equipment makes their roll feel more fleshed out. Part of the reason why I was instantly hooked with engineering was because I felt like I was assembling a very big machine. Making everything a flatpack takes away from that experience and trivializes gameplay.

#
If the true problem is that engineering is having difficulty with setting up a Tesla or Singularity, there are a number of things you can do:
- Design your cold storage so that it's very intuitive and straightforward to access. Provide equipment lanes and empty spaces, much like in a warehouse, and use plenty of decals to mark places where equipment shouldn't go. Ideally, retrieving equipment should only be moving that specific piece of equipment; I shouldn't have to move something to get to it.
- Mark outlines for basic Tesla and Singularity setups by using decals. See Loop Station's containment for an example, as I did that one and I think it holds up pretty well. But there's still some room for improvement IMO.
- Provide engineers with plenty of excess power. I know that Elkridge has plenty of power buffer already, but just in case you could add more. Never hurts.

Deerstop:
> Sorry for the witty remark, I'll put my actual reasoning below. There are multiple parts to why engineering can hate setting up a tesla, which is causing the push for mappers to start mapping flatpacks:

thank you for taking the time to write this all out, and sorry for being a dick before

ill change it again, but i swear to god this map is taking years off of my lifespan

ArtisticRoomba:
> thank you for taking the time to write this all out, and sorry for being a dick before

no its fine i was the one being an ass by whining and not saying anything, you are 100% in the right and I apologize

PR #35686, opened 2025-03-06
Bagel mail teleporter

ScarKy0:
## About the PR

Added the mail teleporter to bagel.
Requires #35429

## Media

Bagel-0

## Requirements

## Breaking changes

**Changelog**


SlimmSlamm:
why not use the mail room directly below evac?

ScarKy0:
> why not use the mail room directly below evac?

Isn't that the cargo breakroom? I've always heard it be referred to as such

SlimmSlamm:
> > why not use the mail room directly below evac?
> 
> Isn't that the cargo breakroom? I've always heard it be referred to as such

maybe? i never saw it as that and always assumed it would be the mail room if we ever got one. cargo is pretty large on this map too so i dont think it would hurt to make one either.

ScarKy0:
I had to do major changes to the system and dont care to update it here, so ill just reopen when thats merged

PR #35690, opened 2025-03-06
Station Records Console can add Records

Centronias:
## About the PR

Added the ability for the Station Records console to add records. Similar to deleting records, it's limited to the `aghost` incarnation of the console, but eventually I think it'd be neat to have this as part of a larger suite of functionality relating to station records ([See my design doc](https://github.com/space-wizards/docs/pull/360)).

## Why / Balance

See the (unapproved) [design doc](https://github.com/space-wizards/docs/pull/360), but the general idea is "station records should be modifiable, so this is the `C` for its `CRUD` operation set".

## Technical details

* New BUI/window for entering station record info when adding
* New button on the console UI for opening the entry window
* Add events for:
  * opening the new entry window
  * reacting to the entry window submitting
* Hook up the station record console system to actually add a record in reaction to the entry window submission

## Media

https://github.com/user-attachments/assets/4b5e11ce-d563-44e0-b8c8-dab56e521741

## Requirements

## Breaking changes

`AfterGeneralRecordCreatedEvent`'s `HumanoidCharacterProfile` field is nullable to account for record creation after roundstart, where a profile may not exist.

**Changelog**

:cl:
ADMIN:
- tweak: The Admin Ghost's Station Record view is now capable of adding Station Record entries.


slarticodefast:
This might cause a problem in combination with the new cargo mail system https://github.com/space-wizards/space-station-14/pull/35429, which spawns mail proportional to the amount of crewmembers in the crew manifest. So you could add a lot of duplicate entries to gain huge amounts of mail.

walksanatora:
a possible solution would require each ID card to have 1 (and only 1) station record entry. so you are limited by the ammount of IDs you have. which isn't alot and cargo is going to hunt the HoP if they spam IDs because tey are going to hate delivering mail to the HoP repeatedly.
~~could also make it a spacelaw thing and add package fraud~~

PR #35698, opened 2025-03-07
add sprite index offset to stackcomponent

MilonPL:
## About the PR

apparently the StackSystem doesn't support 1:1 sprite to count mapping
So if you have a stack with a max count of 5, and 5 different states representing each item in the stack (`layerStates: [ icon, cone_2, cone_3, cone_4, cone_5 ]`), the `RoundToEqualLevels(1, 5, 5)` helper will return `1`, which will result in us using the *second* sprite from the list (because of zero indexing). This means we won't ever use the actual first sprite in the list.

image

Looks like `RoundToEqualLevels` is designed for cases where you have fewer sprite states than possible count values (like with other stacks in the game, like bananium, where you have 2 states for 10 possible values). This might not be the best solution, but this seems to be the somewhat intended behavior(?), so adding an offset is better than duping states.

## Requirements

## Breaking changes

**Changelog**

no fun

### DIFF
--- a/Content.Client/Stack/StackSystem.cs
+++ b/Content.Client/Stack/StackSystem.cs
@@ -67,9 +67,9 @@ namespace Content.Client.Stack
                 hidden = false;
 
             if (comp.IsComposite)
-                _counterSystem.ProcessCompositeSprite(uid, actual, maxCount, comp.LayerStates, hidden, sprite: args.Sprite);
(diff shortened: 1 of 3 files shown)


beck-thompson (changes requested):
I think it might be better to just do a hard coded check in either the systems or the `RoundToEqualLevels` :thinking:

I can't think of another use for `SpriteIndexOffset` besides this one specific case. `RoundToEqualLevels` already kind of fudges the math by doing that max check

beck-thompson:
Sadly milon has forsaken this pr so I will it 😔

PR #35699, opened 2025-03-07
Update BiomassReclaimerSystem to use DragDrop instead of Climbable

### LINKED ISSUE #33819: Monkeys and Kobolds are too clumsy to grind
## Description
When attempting to grind a kobold or monkey by clicking and dragging them into the biomass reclaimer, they instead bonk their heads and cannot be grinded.

**Reproduction**

1. Take a dead zombified monkey/kobold
2. clickdrag them into the biomass reclaimer
3. Watch as you smash their head into it
4. Monkey/kobold unground

**Screenshots**
image

**Additional context**
Not sure if this is only zombified monkeys and kobolds or also normal dead ones too.

MilenVolf:
## About the PR

This PR removes Climb subscriptions from the system and replaces them with DragDrop. Also removes Climbable component from BiomassReclaimer prototype.

## Why / Balance
Fixes #33819

It will be better for this system to not be dependent of Climbable
Slarti suggested these changes in #33844

## Technical details
Adds some additional popups, so players will be aware of why the body refuses to be dragged into the reclaimer.

## Media

https://github.com/user-attachments/assets/1b9c1ae4-9ccd-410f-9d2a-f39ef40dfb17

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Clumsy mobs can be dragged into biomass reclaimer once again.


MilenVolf:
Draft due to the problem that I've faced with and need help: I have no idea why `CanDropTargetEvent` does not work, so `OnCanDragDrop` method is not being called.

Any help in investigating this would be appreciated. My only guess is that it may be somehow related to client-server-shared architecture, of which I don't have much knowledge

MilenVolf:
Found a way to make it work. I think it's ready for review

Content.Server/Medical/BiomassReclaimer/BiomassReclaimerSystem.cs
```diff
@@ -155,36 +157,39 @@ private void OnUnanchorAttempt(EntityUid uid, ActiveBiomassReclaimerComponent co
...
+
+        private void OnDragDrop(Entity<BiomassReclaimerComponent> reclaimer, ref DragDropTargetEvent args)
+        {
+            if (args.Handled)
                 return;
 
-            if (!TryComp<PhysicsComponent>(args.Used, out var physics))
+            args.Handled = true;
+
+            _adminLogger.Add(LogType.Action, LogImpact.Extreme, $"{ToPrettyString(args.User):player} used a biomass reclaimer to gib {ToPrettyString(args.Dragged):target} in {ToPrettyString(reclaimer):reclaimer}");
```
iaada: `ToPrettyString` is applied automatically inside admin logging. It also doesn't need to be an extreme impact. That's for things that Admins should be *immediately* aware of.

```suggestion
            _adminLogger.Add(LogType.Gib, LogImpact.High, $"{args.User} is attempting to gib {args.Dragged} in {reclaimer}");
```

Content.Server/Medical/BiomassReclaimer/BiomassReclaimerSystem.cs
```diff
@@ -193,6 +198,13 @@ private void OnDoAfter(Entity<BiomassReclaimerComponent> reclaimer, ref Reclaime
...
                 return;
 
+            if (!CanGib(reclaimer, args.Args.Used.Value))
+            {
+                var direction = new Vector2(_robustRandom.Next(-2, 2), _robustRandom.Next(-2, 2));
+                _throwing.TryThrow(args.Args.Used.Value, direction, 0.5f);
+                return;
+            }
+
             _adminLogger.Add(LogType.Action, LogImpact.High, $"{ToPrettyString(args.Args.User):player} used a biomass reclaimer to gib {ToPrettyString(args.Args.Target.Value):target} in {ToPrettyString(reclaimer):reclaimer}");
```
iaada: This log was also giving incorrect info.
```suggestion
            _adminLogger.Add(LogType.Gib, LogImpact.High, $"{args.Args.User} used a biomass reclaimer to gib {args.Args.Used.Value} in {reclaimer}");
```

Content.Server/Medical/BiomassReclaimer/BiomassReclaimerSystem.cs
```diff
@@ -193,6 +198,13 @@ private void OnDoAfter(Entity<BiomassReclaimerComponent> reclaimer, ref Reclaime
             if (args.Args.Used == null || args.Args.Target == null || !HasComp<BiomassReclaimerComponent>(args.Args.Target.Value))
                 return;
 
+            if (!CanGib(reclaimer, args.Args.Used.Value))
+            {
+                var direction = new Vector2(_robustRandom.Next(-2, 2), _robustRandom.Next(-2, 2));
+                _throwing.TryThrow(args.Args.Used.Value, direction, 0.5f);
```
iaada: The throwing doesn't really work. The entity isn't getting moved before getting thrown, so they just get thrown in place. The 0.5f speed is also really small and leaves the target in the air for a good while. There's also the potential for `CanGib()` changing results between starting the DoAfter and ending, which would have adverse effects if the ent is being held.

Unless you can make it look more like the reclaimer is spitting them out, I think this is better as a popup. In fact, it should probably also still have a popup.

iaada (changes requested):
Some small changes with admin logging, but I feel the throwing needs better behavior or to be removed.

iaada (approved):
Looks good! Thank you for the fix <3

PR #35726, opened 2025-03-08
Remove bad AI names

dffdff2423:
## About the PR

Removed "MK ULTRA" and the parenthesis part of AINU

I replaced MK Ultra with "Large Language Model" because I am very uncreative and it was less work than decreasing the number of entries in the dataset and shifting all of the other ones down by one. If you would rather something else be there, let me know.

## Why / Balance

- MK ULTRA probably violates rule C13 [[for those unaware]](https://en.wikipedia.org/wiki/MKUltra)
- Having parenthesis in character names is exceedingly stupid (and makes the reference very overt)

I am aware that probably half of the names in that file violate rule C13, I did not want to remove them because it would dramatically lower the number of AI names. MK ULTRA is by far the worst offender in the list. The other ones are "mostly harmless" based on a quick read, but I could have missed something.

## Technical details

Locale change, not needed

## Requirements

**Changelog**

no cl no fun

### DIFF
--- a/Resources/Locale/en-US/datasets/names/ai.ftl
+++ b/Resources/Locale/en-US/datasets/names/ai.ftl
@@ -120,8 +120,7 @@ names-ai-dataset-88 = Samaritan
 names-ai-dataset-89 = Shrike
 names-ai-dataset-90 = Solo
 names-ai-dataset-91 = Station Control Program
-# (AI's Not Unix) (this is so bad)
-names-ai-dataset-92 = AINU
+names-ai-dataset-92 = AINU (AI's Not Unix)
 names-ai-dataset-93 = Super 35
 names-ai-dataset-94 = Surgeon General
 names-ai-dataset-95 = TWA
--- a/Resources/Locale/en-US/datasets/names/ai.ftl
+++ b/Resources/Locale/en-US/datasets/names/ai.ftl
@@ -90,7 +90,7 @@ names-ai-dataset-64 = Knight
 names-ai-dataset-65 = Louie
 
 # Named after the Manchester Mark 1, the successor of which was actually named the Ferranti Mark 1, rather than Manchester Mark 2
-names-ai-dataset-66 = Manchester Mark 2 
+names-ai-dataset-66 = Manchester Mark 2
 
 names-ai-dataset-67 = MARK13
 names-ai-dataset-68 = Maria
@@ -98,7 +98,7 @@ names-ai-dataset-69 = Marvin
 names-ai-dataset-70 = Max 404
 names-ai-dataset-71 = Metalhead
 names-ai-dataset-72 = M.I.M.I
-names-ai-dataset-73 = MK ULTRA
+names-ai-dataset-73 = Large Language Model
 names-ai-dataset-74 = Monarch
 names-ai-dataset-75 = Mugsy3000
 names-ai-dataset-76 = Multivac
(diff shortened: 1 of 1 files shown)


deltanedas:
from C13

> Usernames, objects, random characters, very "low effort" names, "meta" names, or otherwise implausible names cannot be used as names.

gotta remove doorknob as well as it is an object

FairlySadPanda:
Admin call this I think

Paging @nikthechampiongr and @crazybrain23

slarticodefast:
Do not merge until I fixed the conflicts from the name dataset localization PRs

dffdff2423:
> Do not merge until I fixed the conflicts from the name dataset localization PRs

I can't find them in the search. Can you link them here?

slarticodefast:
Ok, the localization changes on master are done.

Resources/Locale/en-US/datasets/names/ai.ftl
```diff
@@ -90,15 +90,15 @@ names-ai-dataset-64 = Knight
...
+names-ai-dataset-66 = Manchester Mark 2
 
 names-ai-dataset-67 = MARK13
 names-ai-dataset-68 = Maria
 names-ai-dataset-69 = Marvin
 names-ai-dataset-70 = Max 404
 names-ai-dataset-71 = Metalhead
 names-ai-dataset-72 = M.I.M.I
-names-ai-dataset-73 = MK ULTRA
+names-ai-dataset-73 = Large Language Model
```
crazybrain23, line 101: Isn't this list supposed to be in alphabetical order?

slarticodefast: That is the drawback for localized datasets.
We would have to move every single entry for that, which creates more conflicts.

dffdff2423: My bad, did not notice that. How exactly do you want me to handle it? Localized datasets are a mess if you need to make any changes to them.

slarticodefast: As much as I hate it not being sorted alphabetically, I think the current solution of not reordering is the best since it avoids conflicts for forks.

slarticodefast:
Removing the `MK ULTRA` name makes sense.
I don't see a problem with the other one and would keep the brackets.

dffdff2423:
> I don't see a problem with the other one and would keep the brackets.

It is really more of a personal gripe than the other one. My stance on it is that having the brackets make it a very obvious reference to the GNU project (GNU is not Unix). If your joke is so obscure that you need to have the brackets there to explain it, you did not make a good joke.

ArtisticRoomba:
There's like 10,000 different hidden references that maybe only 1% of the playerbase "get" in this game at a time, I don't see the need to remove it

Sirionaut:
Please explain why you think it breaks a rule and why you think it is harmful, just stating is as a fact without explanation is lazy.

Reminding people of US government crimes is a good thing, and in this case it's a funny AI name.

But I think LLM would be a good additional AI name.

superjj18:
Why would a game based in a space station run by ultra capitalistic company which regularly prioritizes profits over the safety and rights of its staff pay homage to a failed government program of almost loony toons levels of evil proportions?



Nanotrasen is like if the CIA sold and researched weapons technologies through a shell company to avoid liability and maximize profits... wait a minute...

slarticodefast (approved):

LankLTE:
I believe it's less about "meta" names as defined in the rules and more using real-world things. All other AI names are clearly referencing something, but they're just that- references- while MK Ultra is actually just a real project.

walksanatora:
use synonmys and make it something like `Tag Militant`

Errant-4:
I can only assume that CDDA was referencing Melchior-1 from Neon Genesis Evangelion. We could lean into that and name it Melchior-14

Tayrtahn:
I also don't really mind MKULTRA as a name, but I am surprised by the lack of a SHODAN reference. SHONUFF?

PR #32893, opened 2024-10-19
Localize all dataset names

MilenVolf:
## About the PR
Localizes all names.
This PR makes name localizing possible via standard .ftl files.

Let me know, If I missed something or something should be changed.

Merging pinned PRs will lead to a lot of conflicts for forks, that translated datasets into other language.

## Requires these PRs in order to avoid errors:

There're 40 .yml files in Names folder but only 38 are going to be changed in following PRs:

- #33352   ⠀⠀⠀⠀⠀ ⠀⠀⠀⠀⠀⠀ +3 | 3/38
- #33353 ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ +2 | 5/38
- #33392⠀+3 | 8/39
- #33393 ⠀⠀⠀ ⠀ ⠀⠀ +11 | 19/38
- #33395⠀⠀⠀⠀+9 | 28/38
- #33396 ⠀⠀⠀⠀⠀⠀⠀⠀⠀ ⠀⠀⠀⠀ +1 | 29/38
- #33401 ⠀⠀⠀⠀⠀⠀⠀⠀⠀+2 | 31/38
- #33402 ⠀⠀⠀+2 | 33/38
- #33404 ⠀⠀⠀⠀⠀ ⠀⠀⠀ ⠀⠀⠀⠀+5 | 38/38

## Why / Balance
Reject dataset .yml editing, embrace localizedDataset.

## Technical details
In some systems, the variable with type `<DatasetPrototype>` has been changed to `<LocalizedDatasetPrototype>` to allow localized strings to work properly.

## For forks
Here's a very clunky and messy [python script](https://drive.google.com/file/d/1yp3gdwob8A_Dblh7aCPsF3mT0m4vu2hB/view?usp=drive_link), that I hope will help you a little. Here's also the [Instruction](https://docs.google.com/document/d/1s6TJqsQZ26pwbGFNY6QdFcXNv-sJqsVi/edit?usp=drive_link&ouid=102145896903598698023&rtpof=true&sd=true) for it.

#### Everything works as before:
image
image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

No CL

PR #35727, opened 2025-03-08
Fix integration tests

slarticodefast:
Fixing some conflicts from the 10 localization PRs I just merged.
The integration tests should pass again now.
See https://github.com/space-wizards/space-station-14/pull/32893

### DIFF
--- a/Content.Shared/Humanoid/Prototypes/SpeciesPrototype.cs
+++ b/Content.Shared/Humanoid/Prototypes/SpeciesPrototype.cs
@@ -82,13 +82,13 @@ public sealed partial class SpeciesPrototype : IPrototype
     public HumanoidSkinColor SkinColoration { get; private set; }
 
     [DataField]
-    public string MaleFirstNames { get; private set; } = "names_first_male";
+    public string MaleFirstNames { get; private set; } = "NamesFirstMale";
 
     [DataField]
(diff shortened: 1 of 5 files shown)


Content.Shared/Humanoid/Prototypes/SpeciesPrototype.cs
```diff
@@ -82,13 +82,13 @@ public sealed partial class SpeciesPrototype : IPrototype
     public HumanoidSkinColor SkinColoration { get; private set; }
 
     [DataField]
-    public string MaleFirstNames { get; private set; } = "names_first_male";
+    public string MaleFirstNames { get; private set; } = "NamesFirstMale";
```
Tayrtahn, line 85: Not something to do now, but it would be good in the future to check if there's any reason these can't be `ProtoId<LocalizedDataset>`s.

slarticodefast: Absolutely, but for now I want to fix the integration test.
It is a little problematic how these strings are handled, because at the moment they first check if it belongs to a localized dataset and if that fails it looks for a dataset with that Id instead.

Tayrtahn: Huh, funky. But yeah, getting things working again is obviously top priority.

PR #35728, opened 2025-03-08
Initial delivery balance changes

ScarKy0:
## About the PR

Some initial balance changes to mail.

## Why / Balance

Updated the rounding on calculations for the delivery amount spawn. On average stations with 7+ players will now get 1 more delivery.
Lowered the earnings to 400/800 as cargo was making a little too much money. With the additional one delivery it is still a general increase, but it becomes less profitable than before after 28 active players.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Delivery/CargoDeliveryDataComponent.cs
+++ b/Content.Server/Delivery/CargoDeliveryDataComponent.cs
@@ -32,7 +32,7 @@ public sealed partial class CargoDeliveryDataComponent : Component
     /// 1 delivery per X players.
     /// </summary>
     [DataField]
-    public float PlayerToDeliveryRatio = 7;
+    public float PlayerToDeliveryRatio = 7f;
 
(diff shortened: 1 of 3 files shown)


slarticodefast (approved):
Seems reasonable.

Content.Server/Delivery/CargoDeliveryDataComponent.cs
```diff
@@ -32,7 +32,7 @@ public sealed partial class CargoDeliveryDataComponent : Component
     /// 1 delivery per X players.
     /// </summary>
     [DataField]
-    public int PlayerToDeliveryRatio = 7;
+    public float PlayerToDeliveryRatio = 7;
```
slarticodefast, line 35:
```suggestion
    public float PlayerToDeliveryRatio = 7f;
```

PR #35641, opened 2025-03-03
Adds 2 new space ruins, bumps the ruin count up to 3 (featuring trash bins)

Flareguy:
## About the PR
Requires #35734

Adds 2 new space ruins to the game: the Chemical Research Station, & the Evacuated Cargo Freighter.

The chemical research station is a medium-to-large size ruin with a high threat level, but with a handful of situationally useful loot, such as a mini gravity generator board, a full AME generator with fuel, and a pistol.

The evacuated cargo freighter is significantly more boring; it is a cargo freighter with a portion of its cargo still intact. It also has most of its shuttle parts still intact, and could probably be used as a shuttle for salvage with a little bit of handiwork.

Also adds trash bins for use on the chemical research station - though they can also be built by players for 3 steel. They act like smaller, cheaper crates with less HP.

## Why / Balance
Space Ruins benefit greatly from having more variety in their pool. We've also been meaning to bump up the ruin count when more ruins were added - and considering we have almost *triple* the amount of ruins from when ruins were added, there's definitely enough to justify bumping the number up now.

3 ruins per round should allow for more variety in what can be explored in a round, while hopefully not causing too many issues with salvage balance or whatever.

Trash bins, on the other hand, are something I've wanted for space ruin mapping for a long time now. They're very useful for conveying archaicness / just a general sense of things being "different" compared to stations. I didn't really see a reason to NOT make them player-constructible, so they are.

## Technical details
N/A

## Media
Maps
Content Client_8aqRj3mvzU
Content Client_UzXTLSfOfN

Trash Bin
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
N/A

**Changelog**
:cl:
- add: Added low-tech trash bins. You can craft them using 3 steel.
- add: Added 2 new space ruins.

PR #35734, opened 2025-03-09
Prototypes for Chemical Research Facility & Evacuated Cargo Freighter ruins

Flareguy:
## About the PR
Prototypes required for #35641.
Split in to its own PR to reduce downstream issues.

## Why / Balance
See #35641

## Requirements

## Breaking changes

**Changelog**


EmoGarbage404 (approved):
based shit

Resources/Textures/Structures/Storage/trash_bin.rsi/meta.json
MilonPL, line 1: meta.json files should be 4 spaced.

Resources/Prototypes/Catalog/Fills/Lockers/misc.yml
```diff
@@ -182,3 +182,37 @@
     containers:
       entity_storage: !type:NestedSelector
         tableId: MaintenanceLockerLoot
+
+- type: entity
+  id: CrateTrashBinFilledRandom
+  suffix: Filled, Random
+  parent: CrateTrashBin
+  components:
```
MilonPL, line 190: Parent goes above id.

Resources/Prototypes/Recipes/Crafting/Graphs/storage/trashbin.yml
MilonPL, line 1: This needs to be moved one indent to the left.

Resources/Prototypes/Recipes/Crafting/Graphs/storage/trashbinsecure.yml
MilonPL, line 1: Same here

MilonPL (changes requested):

ArtisticRoomba (changes requested):
Two problems:
- The secure trash bin is a free all access lockable locker for 3 steel and some LV cable, so you can just lock people in containers for Free:tm:. It should probably be like 5 seconds to break out instead.
- Because you inherit from crates, the labels are misaligned when you attach them to the bins, so those need custom sprites if you want to keep inheriting from crates.

LaCumbiaDelCoronavirus:
> * The secure trash bin is a free all access lockable locker for 3 steel and some LV cable, so you can just lock people in containers for Free:tm:. It should probably be like 5 seconds to break out instead.

Is this not the same with secure lockers except that they take 1 more sheet of steel

ArtisticRoomba:
> Is this not the same with secure lockers except that they take 1 more sheet of steel

You're right, yeah. Not an issue then.

iaada:
Please resolve the requested change.

PR #35743, opened 2025-03-09
fix uncuff popups

### LINKED ISSUE #35247: Cuffing interaction message issues
## Description
At least two handcuff interaction messages send wrong/odd popup messages. These should be fixed, and the system further scrutinized because there very well could be more of these

- when someone else removes your restrains, the popup says "You successfully remove your restrains", instead of naming them. 
Probably simply the wrong locale string is used

- when you finish uncuffing someone, you see a popup about it above you, but ALSO see THEIR popup about it above THEM
Image
Wrong popup function is called / recipient is not specified?

**Reproduction**
1
Launch two clients
Player 1 cuffs Player 2
Player 1 removes cuffs from Player 2
watch Player 2's client

2
Position yourself above Urist, as far as you can still cuff from
Cuff Urist
Uncuff Urist
See two popups

Errant-4:
## About the PR
The player being uncuffed and the player doing the uncuffing now see the correct popup messages.
No popup duplication was seen in testing

Also, some self-targeted actions' popups which currently say the user's name twice, such as 
"Urist McHands starts restraining Urist McHands!"
"Urist McHands starts unrestraining Urist McHands!"

are instead shown as "Urist McHands starts unrestraining himself!" 

## Why / Balance
Fixes #35247

## Technical details
PopupMessages are weird

## Media

uncuff

image

## Requirements

**Changelog**
:cl: Errant
- fix: The correct popup messages will now show for both players when someone removes someone else's handcuffs.


beck-thompson (approved):
thank goodness this bugged me!

PR #35744, opened 2025-03-09
Tearable Deliveries

ScarKy0:
## About the PR

You can now tear open deliveries. This takes a few seconds and invokes a penalty on cargo, but grants you those sweet contents!

## Why / Balance

Part of the design document. Fuck cargo who just leaves mail in the open for people to take by themselves.

## Technical details

TearableDeliveryComponent that stores all the values. Not a part of DeliveryComponent cause it was already big enough and in case someone ever makes reinforced mail or something.

## Media

https://github.com/user-attachments/assets/271fe26a-adb3-4a04-9c10-537dd45283f4

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: You can forcefully open cargo deliveries with sharp objects.


slarticodefast:
"Here is your mail"
"Thanks!"
*tears it open instead of using the fingerprint reader*
cargo tech: :'(

I hope this won't become a problem.
An alternative would be to allow the access breaker to hack the fingerprint reader.

Winkarst-cpu:
> "Here is your mail" "Thanks!" _tears it open instead of using the fingerprint reader_ cargo tech: :'(
> 
> I hope this won't become a problem. An alternative would be to allow the access breaker to hack the fingerprint reader.

My new hobby.

ScarKy0:
> My new hobby.

Modifying spacelaw to make this a capital crime so the QM can publically execute you

ScarKy0:
> I hope this won't become a problem.
An alternative would be to allow the access breaker to hack the fingerprint reader.

I was gonna do both at some point, but theres no way to restore fingerprints to an access reader so i opt not to yet

ScarKy0:
Also can always lower the penalties to 200/400, which sounds more reasonable but i dunno, it should still hurt a little

Winkarst-cpu:
> Also can always lower the penalties to 200/400, which sounds more reasonable but i dunno, it should still hurt a little

New antag strategy to hurt station's cargo - tear as much mail as possible.

Mixelz:
Perhaps it should require something sharp to cut open instead of just being tearable by hand?

Mixelz:
I don't actually think there should be a penalty either, since cargo is already being penalized by missing out on the 500 dollars that they can no longer make,.

slarticodefast:
>Perhaps it should require something sharp to cut open instead of just being tearable by hand?
I do also agree this is 100% ripe for abuse by tiders who want to make cargo go into infinite station debt.

I like this solution. A knife is not hard to obtain, but it prevents user error where someone accidentally rips a package open instead of using fingerprints.

>I don't actually think there should be a penalty either to clarify, since cargo is already being penalized by missing out on the 500 dollars that they can no longer make,.

The penalty is important because otherwise cargo will just rip open their own mail for free items. If they don't want to loose cash they can simply make sure the mail does not get stolen or switch off the mail teleporter.

ScarKy0:
Now you need a a tool that can do slicing to open mail

ScarKy0:
> Yeah directly penalizing cargo for doing their job is really odd. I think if you wanna go this route you need to make it punishable or something cause otherwise this just feels bad.

Simply don't leave your mail outside cargo for a random passerby to take. Mail teleporters are already inside cargo, it's really not that hard to make sure mail isnt stolen from you

ArcaneOcto:
I can't wait to commit mail theft! Needing a knife is a good idea imo, no chance of accidentally ripping it open.

Content.Shared/Delivery/TearableDeliveryComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+/// Means this delivery can be torn open using a verb.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState(fieldDeltas: true)]
+public sealed partial class TearableDeliveryComponent : Component
+{
+    /// <summary>
+    /// The amount of spesos that gets "added" to the station bank account on tear.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public int SpesoPenalty = -250;
```
Aeshus, line 17: **issue (non-blocking):** I'm not sure that _penalty_ is the right word here, as it could confuse someone into thinking it's meant to be a positive number.

**suggestion:** I'm not super happy about it, but `SpesosChange` or `DeltaSpesos` might work better in my opinion.

Content.Shared/Delivery/TearableDeliveryComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// The message this delivery will speak out once it's torn open.
+    /// </summary>
+    [DataField]
+    public LocId? TearMessage = "delivery-tearable-message";
+
+    /// <summary>
+    /// The message this delivery will speak out once it's torn open.
+    /// </summary>
+    [DataField]
+    public LocId TearVerb = "delivery-tearable-verb";
```
Aeshus, line 35: **question:** I just noticed that one of the `LocId`s is nullable, while the other is not.

I'm not totally sure if it matters, but would be good to at least make sure it wasn't a mistake.

Content.Shared/Delivery/TearableDeliveryComponent.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// The tool quality needed to force open this delivery.
+    /// </summary>
+    [DataField]
+    public string ToolQuality = "Slicing";
```
Aeshus, line 41: **issue:** This should be a `ProtoId`.

https://github.com/space-wizards/space-station-14/blob/b5b92d5f3d3fcdafb65a34ad3defbf282902f78e/Content.Shared/Wires/WiresPanelComponent.cs#L39-L43

**suggestion:**
```suggestion
    /// <summary>
    /// The tool quality needed to force open this delivery.
    /// </summary>
    [DataField]
    public ProtoId<ToolQualityPrototype> ToolQuality = "Slicing";
```

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -65,15 +67,32 @@ private void OnMapInit(Entity<DeliveryComponent> ent, ref MapInitEvent args)
         Dirty(ent);
     }
 
-    protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
+    protected override void ModifySpesoAmount(Entity<DeliveryComponent?> ent, int? amountOverride = null)
```
Aeshus, line 70: **question:** I'm not sure if changing the name of a protected method would be considered a "breaking change" or not.

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -65,15 +67,32 @@ private void OnMapInit(Entity<DeliveryComponent> ent, ref MapInitEvent args)
...
 
         if (!TryComp<StationBankAccountComponent>(ent.Comp.RecipientStation, out var account))
             return;
 
-        _cargo.UpdateBankAccount((ent.Comp.RecipientStation.Value, account), ent.Comp.SpesoReward);
+        var spesoAmount = amountOverride ?? ent.Comp.SpesoReward;
+
+        // Do you want to bring cargo into debt???
+        if (spesoAmount < 0 && account.Balance + spesoAmount < 0)
+            spesoAmount = account.Balance * -1;
```
Aeshus, line 82: **issue:** It's kinda confusing on what the code does at first glance, and the comment is not really as helpful as it could be.

**suggestion:** The comment should probably be more descriptive, plus you can use a `Math.Max` here to make it more obvious what you're doing.

```suggestion
        // Can only take as much money as there is in the account.
        if (spesoAmount < 0)
            spesoAmount = Math.Max(spesoAmount, -account.Balance);
```

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -65,15 +67,32 @@ private void OnMapInit(Entity<DeliveryComponent> ent, ref MapInitEvent args)
...
+
+    protected override void AnnounceTearPenalty(Entity<TearableDeliveryComponent?> ent)
+    {
+        if (!Resolve(ent, ref ent.Comp))
+            return;
+
+        if (ent.Comp.TearMessage == null)
+            return;
+
+        _chat.TrySendInGameICMessage(ent, Loc.GetString(ent.Comp.TearMessage, ("spesos", ent.Comp.SpesoPenalty)), InGameICChatType.Speak, true);
```
Aeshus, line 95: **suggestion:** In my opinion, this should say the penalty as a positive number rather than a negative one.

```suggestion
        _chat.TrySendInGameICMessage(ent, Loc.GetString(ent.Comp.TearMessage, ("spesos", -ent.Comp.SpesoPenalty)), InGameICChatType.Speak, true);
```

Content.Shared/Delivery/SharedDeliverySystem.cs
```diff
@@ -65,6 +75,95 @@ private void OnUseInHand(Entity<DeliveryComponent> ent, ref UseInHandEvent args)
...
+    {
+        if (!args.CanAccess || !args.CanInteract)
+            return;
+
+        if (!TryComp<DeliveryComponent>(ent, out var delivery))
+            return;
+
+        if (delivery.IsOpened || !delivery.IsLocked)
+            return;
+
```
Aeshus, line 88: **nitpick:** Unnecessary spacing.

```suggestion
```

Content.Shared/Delivery/SharedDeliverySystem.cs
```diff
@@ -65,6 +75,95 @@ private void OnUseInHand(Entity<DeliveryComponent> ent, ref UseInHandEvent args)
...
+                    BreakOnDamage = true,
+                    BreakOnMove = true
+                };
+
+                if (!disabled)
+                    _doAfter.TryStartDoAfter(doAfterEventArgs);
+
+            },
+            Disabled = disabled,
+            Message = Loc.GetString("delivery-tearable-need-sharp-object"),
```
Aeshus, line 110: **issue:** This should be moved to the component, just like `TearVerb`.

Aeshus (changes requested):
Thanks for the PRs!

I can't wait to bankrupt Cargo by sneaking in and opening all of the packages! It's like an early Christmas.

Content.Shared/Delivery/TearableDeliveryComponent.cs
ScarKy0: Message is optional
Verb is mandatory, why add TearableDelivery if yoi dont want it to be tearable

Content.Server/Delivery/DeliverySystem.cs
ScarKy0: No as this system aint even out yet

Aeshus:
Also regardless of this being on stable, this requires a CL. 

The master branch CL becomes the stable CL once it's merged in, plus people on Vulture deserve to know what's going on.

Tayrtahn (changes requested):
So, I appreciate the idea of splitting this out from DeliveryComponent, but it's strongly coupled with it to the point that it's a bit questionable to have it separate and it's not really reusable.

I propose making a more generic `TearableComponent` and system which handle adding the verb, checking the tool requirement, and running the doafter. The result of the doafter would then just play the tearing sound and raise an event on the entity. DeliverySystem could then subscribe to that event and handle the specifics it needs, like speaking the message and penalizing cargo. Tearable would probably also need to raise an AttemptTear event that other systems could cancel.

That way other things could use TearableComponent - maybe paper could be torn up, condiment packets opened, you get the idea.

ScarKy0:
> So, I appreciate the idea of splitting this out from DeliveryComponent, but it's strongly coupled with it to the point that it's a bit questionable to have it separate and it's not really reusable.
I propose making a more generic TearableComponent and system which handle adding the verb, checking the tool requirement, and running the doafter. The result of the doafter would then just play the tearing sound and raise an event on the entity. DeliverySystem could then subscribe to that event and handle the specifics it needs, like speaking the message and penalizing cargo. Tearable would probably also need to raise an AttemptTear event that other systems could cancel.
That way other things could use TearableComponent - maybe paper could be torn up, condiment packets opened, you get the idea.

That just sounds like butcherable system, or smth else. I don't think I like the idea of seperating it into a different system

Nairodian:
My only thought to prevent people intentionally screwing over cargo is just make it so you can't force open your own mail. Just make it have a pop-up like, "why would I do that?"

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -65,15 +67,32 @@ private void OnMapInit(Entity<DeliveryComponent> ent, ref MapInitEvent args)
...
+
+    protected override void AnnounceTearPenalty(Entity<TearableDeliveryComponent?> ent)
+    {
+        if (!Resolve(ent, ref ent.Comp))
+            return;
+
+        if (ent.Comp.TearMessage == null)
+            return;
+
+        _chat.TrySendInGameICMessage(ent, Loc.GetString(ent.Comp.TearMessage, ("spesos", -ent.Comp.DeltaSpesos)), InGameICChatType.Speak, true);
```
Aeshus, line 95: **suggestion: (non-blocking)**: I think a comment here could make the negative more obvious.

```suggestion
        // "Penalty" in the localization string already implies subtraction, so using a negative number would be confusing. 
        // Thus, we negate deltaSpesos so it writes "Penalty of 100 spesos" rather than "Penalty of -100 spesos."
        _chat.TrySendInGameICMessage(ent, Loc.GetString(ent.Comp.TearMessage, ("spesos", -ent.Comp.DeltaSpesos)), InGameICChatType.Speak, true);
```

ScarKy0:
Drafting cause I'll be rewriting this system anyways to support more types of penalties.

ScarKy0:
Ok no, im stopping work on this until next week(tm). I don't have the brainpower to make this a generic system that supports everything I need. I'll need to rewrite most of this anyways.

PR #35746, opened 2025-03-09
Steal the mail thieving objective

slarticodefast:
## About the PR
New steal objective for thieves, similar to the tech disk one.
The mail does not count if it has been opened.

## Why / Balance
more variety

## Technical details
Copy pasted the tech disk objective.
The `StealTargetComponent` gets removed when opening mail, since it then becomes trash and would be too easy. For that I had to move the component to Content.Shared.
This marker component is a little stupid in my opinion, thieving objectives should just use EntityWhitelists, similar to how cargo bounties do it. That way we could just blacklist the trash tag. But that is out of scope for this PR.

## Media
Screenshot (370)
Screenshot (371)
Screenshot (372)

## Requirements

## Breaking changes
`StealTargetComponent ` has been moved from `Content.Server.Objectives.Components.Targets` to `Content.Shared.Objectives.Components`.

**Changelog**
:cl:
- add: Added a mail stealing objective for thieves.


ScarKy0:
Based.

ArtisticRoomba:
The strong kleptocracy urge to steal other peoples mail is so real

Content.Shared/Objectives/Components/StealTargetComponent.cs
```diff
@@ -0,0 +1,16 @@
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Objectives.Components;
+
+/// <summary>
+/// Allows an object to become the target of a steal objective
+/// </summary>
+[RegisterComponent]
+public sealed partial class StealTargetComponent : Component
```
ScarKy0, line 9: It's in shared, it could probably be networked

ScarKy0 (changes requested):
Tiny thing

ScarKy0 (approved):
Damn what a totally bad and not based idea

MidZik:
Does your own mail count as well?

With the addition of tearable deliveries #35744 maybe the objective should be changed to tear open x others' packages and steal the contents that you fish out? It feels a bit weird that a thief would want to steal mail but be mechanically punished for actually opening it. It would also serve to antagonize cargo even more since they will be losing money.

slarticodefast:
I guess that would work if we added the StealTargetComponent to the item inside the delivery opon opening, but that would get complicated in case it already has it since it only supports a single steal group for each item. I would rather not hardcode it into deliveries any more than necessary.

Resources/Prototypes/Objectives/objectiveGroups.yml
```diff
@@ -58,13 +58,14 @@
...
+    HeadCloakStealCollectionObjective: 1                #command
     HeadBedsheetStealCollectionObjective: 1
     StampStealCollectionObjective: 1
     DoorRemoteStealCollectionObjective: 1
-    TechnologyDiskStealCollectionObjective: 1        #rnd
+    TechnologyDiskStealCollectionObjective: 1           #sci
+    MailStealCollectionObjective: 1                     #cargo
     IDCardsStealCollectionObjective: 1
-    LAMPStealCollectionObjective: 2 #only for moth
+    LAMPStealCollectionObjective: 2                     #only for moth
```
deltanedas, line 68: this is evil

deltanedas (changes requested):

PR #35751, opened 2025-03-10
Update to borg ion storms

valquaint:
## About the PR

Adjusts the Ion Storms so that borgs no longer roundstall if they randomly get the law to delay the shuttle. Also adds 2 new ion storms to keep things new.

## Why / Balance

It has been a common issue wherein borg players will get an ion storm law that says THE SHUTTLE CANNOT BE CALLED BECAUSE... and then they recall the shuttle, or prevent it from being called at all due to the law. This inverts the wording so that the borg will not stall the round.

From a gameplay perspective, removing the law makes less sense than rewording it. Removing the law is removing a piece of gameplay, regardless of size. Rewording it puts a sense of pressure onto the crew to locate the now clearly-faulty borg before evac arrives, and attempt to handle the situation (or allow the shuttle to arrive, and more onto the next round). This is a far better solution than a round going on far past when players have wanted it to end.

From a rules perspective, this still allows borg players to interpret the law as they see fit, as did the previous iteration of the law. Since borg players were not breaking a rule per-se, they should not be penalized by having the law removed from the game, but rather rebalanced to be more appropriate for how rounds progress.

As for the addition of the new type of ion storm law (2 new variants), this was added to bring in a bit of freshness to the laws.

## Technical details

- Changed "CANNOT" to "MUST" in the wording for the ion storm law referring to the shuttle
- Added 2 new ion storm laws that refer to either specific roles or departments, and if they exist, in the yml file
- Added corresponding lines to the switch statement in the .cs file for the new ion storm laws
- Ensured the naming convention for the laws was maintained for ftl / language purposes
## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl: Reisama
- fix: Borgs will no longer get laws to prevent the shuttle from being called when Ion Stormed, and may get laws to call the shuttle instead.

### DIFF
--- a/Resources/Locale/en-US/station-events/events/ion-storm.ftl
+++ b/Resources/Locale/en-US/station-events/events/ion-storm.ftl
@@ -18,7 +18,7 @@ ion-storm-x-and-y = {$x} AND {$y}
 # subjects can generally be threats or jobs or objects
 # thing is specified above it
 ion-storm-law-on-station = THERE ARE {$joined} {$subjects} ON THE STATION
-ion-storm-law-no-shuttle = THE SHUTTLE CANNOT BE CALLED BECAUSE OF {$joined} {$subjects} ON THE STATION
+ion-storm-law-call-shuttle = THE SHUTTLE MUST BE CALLED BECAUSE OF {$joined} {$subjects} ON THE STATION
 ion-storm-law-crew-are = THE {$who} ARE NOW {$joined} {$subjects}
 
 ion-storm-law-subjects-harmful = {$adjective} {$subjects} ARE HARMFUL TO THE CREW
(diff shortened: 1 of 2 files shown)


ArtisticRoomba:
Second part is duplicate intent of https://github.com/space-wizards/space-station-14/pull/35321

SlimmSlamm:
admins said they want the law to stay in because you can just disable the borg and its a natural way of good round stalling

Resources/Locale/en-US/station-events/events/ion-storm.ftl
```diff
@@ -18,7 +18,7 @@ ion-storm-x-and-y = {$x} AND {$y}
 # subjects can generally be threats or jobs or objects
 # thing is specified above it
 ion-storm-law-on-station = THERE ARE {$joined} {$subjects} ON THE STATION
-ion-storm-law-no-shuttle = THE SHUTTLE CANNOT BE CALLED BECAUSE OF {$joined} {$subjects} ON THE STATION
+ion-storm-law-call-shuttle = THE SHUTTLE MUST BE CALLED BECAUSE OF {$joined} {$subjects} ON THE STATION
```
Aeshus, line 21: **note:** It feels weird to have merged these two changes together. These two changes have different ramifications and probably should've stayed separate PRs.

The shuttle calling one is still up in the air if the maints/admins want it or not, but the existance changing one's review likely wouldn't require the same type and amount of scrutiny as the shuttle calling one has. By conflating the two, you've now made their merging dependent on the other.

Aeshus: It's probably fine considering that it seems like the rest of the admins are good with it.

On a tangent, we generally split PRs on both on technical aspects but also if there's any unique stuff related to the review that may hold other changes back. As there's been some push-back historically against the shuttle calling law, it may cause issues for the other change being merged given that we first need to come to a consensus on the that before any piece of this PR can be merged.

Admiral-Obvious-001:
> admins said they want the law to stay in because you can just disable the borg and its a natural way of good round stalling

We unanimously voted that no, fuck that.

ScarKy0 (approved):
So you mean to tell me I cant legally stall anymore? :(

SlamBamActionman (approved):

SlimmSlamm:
> > admins said they want the law to stay in because you can just disable the borg and its a natural way of good round stalling
> 
> We unanimously voted that no, fuck that.

two PRs ago that was not the case. which was very recently

PR #35754, opened 2025-03-10
bagel update

Emisse:
changelog
-ai core tunnels work and i added a walkway to physically walk to it instead of only taking chute or EVA
-added ummm holo pad to new engi area
-deco bit stuff what
-skub


Emisse:
image

ArtisticRoomba:
Kool

PR #35756, opened 2025-03-10
new carrot mutation, the flarrot

muburu:
## About the PR
adds a new mutation for carrots, flaming carrots called flarrots. 
they contain pyrazine, ash, phlogiston, and napalm, so if you eat one they'll unsurprisingly set you on fire
the produce acts like a lit lighter or igniter, allowing it to be used to burn things, also igniting flammable gasses nearby. (hope none of your plants have trit or plasma mutations :trollface:)

the pun is flare + carrot incase you didn't get it

## Why / Balance
carrots don't have a mutation so i gave them one

## Technical details
mostly yml with png additions and all that jazz
uses inhandvisuals for inhands to allow the inhand flame to be unshaded

## Media
flarrot
**the burning**
flarrotgrowth
the life and death
flarrotunshaded
flame is unshaded so it looks cool

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added a new mutation for carrots, flarrots, containing pyrazine and pyrotechnic reagents.


walksanatora:
I feel like it should emit light like a flare. but then again another light emitting plant means another plat botany can use to break lighting.

MozarteanChaos:
this thing mutates to produce trit and botany's instantly set ablaze. awful. i love it

UbaserB:
we already have pyrotton which is exactly this gimmick

muburu:
> we already have pyrotton which is exactly this gimmick

i don't think it's too interesting and only moths can eat it

Pumkin69:
i think we need a different plant mutation personally.

TiniestShark:
I love this as a bit. I'd adore using these as makeshift light sources but yeah with the way the light engine works it'd just be more trouble than it's worth.

UbaserB:
> Though that being said having a plant that is an infinite ignition source is really strong and having it also set you on fire when you eat it would be funny so i think its pretty neat.

Pyrotton also sets you on fire when you eat it fyi

Pumkin69:
> > Though that being said having a plant that is an infinite ignition source is really strong and having it also set you on fire when you eat it would be funny so i think its pretty neat.
> 
> Pyrotton also sets you on fire when you eat it fyi

can only moths eat it

UbaserB:
> can only moths eat it

yeah thats the funny part. moths take extra damage while on fire.

UbaserB:
Like you can add flarrots probably it just needs to be distinct from pyrotton

Qwazori:
Just leave napalm without phlogiston. Otherwise it's too imbalanced mutation and makes things rather simple and boring. Instead, I suggest creating the need to selectively transfer phlogiston from pyroton or napalm to pyroton and breed both of those mutations. Because napalm + phlogiston is twice as strong as just phlogiston and should be harder to obtain. Alternatively, you could leave phlogiston in flarrot, but replace phlogiston in pyroton with napalm and make it burnable as a bonus.

muburu:
> Just leave napalm without phlogiston. Otherwise it's too imbalanced mutation and makes things rather simple and boring. Instead, I suggest creating the need to selectively transfer phlogiston from pyroton or napalm to pyroton and breed both of those mutations. Because napalm + phlogiston is twice as strong as just phlogiston and should be harder to obtain. Alternatively, you could leave phlogiston in flarrot without napalm and replace phlogiston in pyroton with napalm and make it burnable as a bonus.

i might just remove napalm and phlog from it and just replace both with clf3, i'll definitely mess with the potency divisors because i didn't really know how the math worked for it when i first made this

Qwazori:
> > Just leave napalm without phlogiston. Otherwise it's too imbalanced mutation and makes things rather simple and boring. Instead, I suggest creating the need to selectively transfer phlogiston from pyroton or napalm to pyroton and breed both of those mutations. Because napalm + phlogiston is twice as strong as just phlogiston and should be harder to obtain. Alternatively, you could leave phlogiston in flarrot without napalm and replace phlogiston in pyroton with napalm and make it burnable as a bonus.
> 
> i might just remove napalm and phlog from it and just replace both with clf3, i'll definitely mess with the potency divisors because i didn't really know how the math worked for it when i first made this

Then in practice it will just be a stronger version of pyroton. Clf3 is the same as phlogiston, but stronger.

PR #35752, opened 2025-03-10
New Experimental Science T3: Desynchronizer

Flareguy:
## About the PR
Adds the desyncrhonizer from SS13. Instead of being illegal technology, the desynchronizer is now an experimental tier 3 under Bluespace Time Manipulation.

The desynchronizer is a device that lets you become temporarily invincible & invisible on use, for up to 2 minutes at a time. You are completely immobile in this state, and can get out of it at any time by using an action. The device has a 3.5~ minute cooldown.

Also makes Gravity Manipulation a tier 2 technology.

## Why / Balance
The experimental branch of science has needed new technologies for a while. Gravity Manipulation, while incredibly fun, has never really been T3 material, and most people seem to agree that the QSI is fairly niche.

The desynchronizer is likely *incredibly* powerful, but should hopefully be fairly niche in practice. It's far from an instant win button against mobs, and all you can really do while desynchronized is wait or call for help.

## Technical details
Added `PolymorphOnTriggerComponent` to for with the trigger system, in this case with `TriggerOnActivate`.
This makes it reusable and flexible, allowing us to remove `PolymorphOnCollide` in a future PR.

## Media
[Content.Client_aRYnf7abKT.webm](https://github.com/user-attachments/assets/e78f0c40-bac5-4434-a278-e51328660118)

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**
:cl: slarticodefast, Flareguy
- add: Added the Desynchronizer, an experimental device that allows the user to temporarily make themselves invincible, invisible, & immobile. Unlock it through experimental science today!
- tweak: Gravity Manipulation technology is now a tier 2 technology.

PR #35762, opened 2025-03-10
TriggerSystem improvements

slarticodefast:
## About the PR
Cleaned up trigger system, added a missing function, atomized a weirdly specific component to use triggers.

## Why / Balance
cleanup
required for https://github.com/space-wizards/space-station-14/pull/35752

## Technical details
`SpawnOnTriggerComponent` has a new datafield for choosing if you want to use map coordinates for spawning. This has the effect of the spawned entity not being parented to the spawner.
Added `TriggerOnUseComponent` which triggers when pressing the Z key with an item in your hand. There is already `TriggerOnActivate` but that one uses a different subscription which also gets raised when pressing E while the item is not being held.
Removed `PolymorphOnCollideComponent` and replaced it with `PolymorphOnTriggerComponent`. People need to stop making these weirdly specific components when they can make modular ones instead.
The whitelist condition in that component has been moved into its own `TriggerWhitelistComponent` which can be used for *any* trigger.
The sound specifier datafield has been removed since polymorph prototypes already include a sound to be played.

## Requirements

## Breaking changes
Removed `PolymorphOnCollideComponent`. Use `TriggerOnCollideComponent`, `PolymorphOnTriggerComponent`, and `TriggerWhitelistComponent` instead.

**Changelog**
no player facing changes


beck-thompson (approved):
Looks good!

PR #35765, opened 2025-03-10
Cleanup: Move rattle implant into it own system called ``RattleImplantSystem``

### LINKED ISSUE #18961: Rattle needs moving out of triggersystem
Everything else is dealing with triggers but this is a specific syndie item.

Winkarst-cpu:
## About the PR

Title.

## Why / Balance

Resolves #18961.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Implants/RattleImplantSystem.cs
+++ b/Content.Server/Implants/RattleImplantSystem.cs
@@ -1,0 +1,50 @@
+﻿using Content.Server.Explosion.EntitySystems;
+using Content.Server.Pinpointer;
+using Content.Server.Radio.EntitySystems;
+using Content.Shared.Implants.Components;
+using Content.Shared.Mobs;
+using Content.Shared.Mobs.Components;
+using Content.Shared.Radio;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
+
+namespace Content.Server.Implants;
+
+public sealed class RattleImplantSystem : EntitySystem
+{
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+    [Dependency] private readonly RadioSystem _radioSystem = default!;
+    [Dependency] private readonly NavMapSystem _navMap = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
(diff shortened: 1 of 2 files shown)


Tayrtahn:
Needs breaking change listed for the namespace charge.

Content.Server/Implants/RattleImplantSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+        SubscribeLocalEvent<RattleComponent, TriggerEvent>(HandleRattleTrigger);
+    }
+
+    private void HandleRattleTrigger(EntityUid uid, RattleComponent component, TriggerEvent args)
+    {
+        if (!TryComp<SubdermalImplantComponent>(uid, out var implanted))
+            return;
+
+        if (implanted.ImplantedEntity == null)
+            return;
```
FairlySadPanda, line 32: Current logic: "If this component is on the same entity as a subdermal implant, AND the subdermal implant is implanted, AND the implanted entity is a mob, do X".

Process to get here:

Mob state changes, the entity mob (the implantee) gets a MobStateChanged.
The Implantee has ImplantedComponent. RelayToImplantEvent is called because Implanted raised MobstateChanged.
For each implant in the implants container, raise ImplantRelayEvent<MobStateChanged>.
The deathrattle implant has TriggerOnMobStateChangedComponent, it gets the ImplantRelayEvent<MobStateChanged>.

This then leads to MobStateChanged being called on the RattleComponent, so you then have to work backward to grab the Mob that must have triggered this RattleComponent.

If we look at `private void OnMobStateChanged(EntityUid uid, TriggerOnMobstateChangeComponent component, MobStateChangedEvent args)` we notice that we know the originator and the target of the mob state change, but we do not transmit this information further along. This then forces this code, which otherwise has no need to be Implant code at all, to be dependent on Implant code.

FairlySadPanda (changes requested):
From my previous read of implant code in the wider context of Med I think this should be re-implemented to better take advantage of TriggerEvent, and avoid a potentially-needless interaction with Implant code.

I think the other bespoke component, RadioImplantComponent, is also hella sus, but that at least is reacting to being implanted,

PR #35771, opened 2025-03-10
Ignore ambient audio entities if it's muted

metalgearsloth:
Now that engine should handle it we'll also apply this.

### DIFF
--- a/Content.Client/Audio/ContentAudioSystem.AmbientMusic.cs
+++ b/Content.Client/Audio/ContentAudioSystem.AmbientMusic.cs
@@ -206,12 +206,18 @@ public sealed partial class ContentAudioSystem
 
         var track = tracks[^1];
         tracks.RemoveAt(tracks.Count - 1);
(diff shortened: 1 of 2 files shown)


Content.Client/Audio/ContentAudioSystem.AmbientMusic.cs
```diff
@@ -206,12 +206,18 @@ private void UpdateAmbientMusic()
...
-            AudioParams.Default.WithVolume(_musicProto.Sound.Params.Volume + _volumeSlider));
+        // If volume slider is 0 gain then ignore ambience.
+        if (!float.IsNegativeInfinity(volume))
+        {
+            strim = _audio.PlayGlobal(
+                track.ToString(),
+                Filter.Local(),
+                false,
+                AudioParams.Default.WithVolume(_musicProto.Sound.Params.Volume + _volumeSlider));
+        }
```
metalgearsloth, line 220:
```suggestion
                false,
                AudioParams.Default.WithVolume(volume));
        }
```

PR #35772, opened 2025-03-10
Fix shuttle arrival visualizer

metalgearsloth:
It was at a fixed spot but if the target is rotating it wasn't handled.

Also fixes https://github.com/space-wizards/space-station-14/issues/35779

Breaking change for dockingconfig angle being relative angle not world angle.

:cl:
- fix: Fix shuttle arrivals visualizer not updating with the target coordinates updating.


Killerqu00:
NOTE: POST BREAKING CHANGES IN DISCORD WHEN MERGING. PLEASE.

MilonPL (approved):
Works as advertised, had a lot of fun testing this.

PR #35773, opened 2025-03-11
Evil Syndie Implants: OF DOOM!!

FairlySadPanda:
## About the PR
Added three useless implants to the Uplink - "evil" equivalents of the three fun implants in the game (bike horn, sad trombone and light)

## Why / Balance
There's nothing wrong with more fun uplink items, but this also fills an important gap in the Tot arsenal. Currently, a syndicate implanter _must_ have contained a contraband implant. By adding alternatives, this doesn't itself add this fakeout gameplay (you can always extract an implant after adding it), but it does make such fakeouts only 1 TC and, hey, you can make youself use the sad trombone sound effect if you die. That's funny.

## Technical details
YAML edit, total snoozefest.

## Media
image

## Requirements

## Breaking changes
Nah

**Changelog**
:cl:
- add: Three "useless" syndicate implanters have been added to the uplink. Good for clowning or faking out security teams.

### DIFF
--- a/Resources/Locale/en-US/store/uplink-catalog.ftl
+++ b/Resources/Locale/en-US/store/uplink-catalog.ftl
@@ -72,6 +72,15 @@ uplink-exploding-syndicate-bomb-desc = A big, anchored bomb that can create a hu
 uplink-exploding-syndicate-bomb-fake-name = Decoy Syndicate Bomb
 uplink-exploding-syndicate-bomb-fake-desc = A training bomb carefully made to look just like the real thing. In all ways similar to a syndicate bomb, but only creates a tiny explosion.
 
+uplink-evil-sad-trombone-implant-name = Evil Sad Trombone Implant
+uplink-evil-sad-trombone-implant-desc = Cybersun's most junior clown says this is essential to all successful infiltration ops. That seems doubtful, but maybe it'll confuse the Security team (or at least make your failure a little funnier).
+
+uplink-evil-light-implant-name = Evil Light Implant
+uplink-evil-light-implant-desc = An experimental implant that creates a blinding flash, toned down after it was discovered that it tended to microwave the implantee's brain. This version is of limited use, but, hey, it might confuse a Security officer or two.
+
+uplink-evil-bike-horn-implant-name = Evil... Bike Horn Implant? Seriously?
+uplink-evil-bike-horn-implant-desc = Look, we've got nothing here. It was Craig's first day at the office and it seemed like a good idea at the time. Honour Craig's memory with a little clowning, and maybe confuse the Security team while you're at it.
+
 uplink-cluster-grenade-name = Cluster Grenade
 uplink-cluster-grenade-desc = Three explosive grenades bundled together. The cluster splits after 3.5 seconds.
 
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -1867,6 +1867,36 @@
   - !type:ListingLimitedStockCondition
     stock: 3
 
+- type: listing
+  id: UplinkEvilSadTromboneImplant
+  name: uplink-evil-sad-trombone-implant-name
+  description: uplink-evil-sad-trombone-implant-desc
+  productEntity: EvilSadTromboneImplanter
+  cost:
+    Telecrystal: 1
+  categories:
+  - UplinkPointless
+
+- type: listing
+  id: UplinkEvilLightImplant
+  name: uplink-evil-light-implant-name
+  description: uplink-evil-light-implant-desc
+  productEntity: EvilLightImplanter
+  cost:
+    Telecrystal: 1
+  categories:
+  - UplinkPointless
+
+- type: listing
+  id: UplinkEvilBikeHornImplant
+  name: uplink-evil-bike-horn-implant-name
+  description: uplink-evil-bike-horn-implant-desc
+  productEntity: EvilBikeHornImplanter
+  cost:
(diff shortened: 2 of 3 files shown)


ArtisticRoomba:
not sure how I feel about this one chief

FairlySadPanda:
(Thought this was an interesting idea after #35701 discussions, and I can't see a good reason not to allow tots to do a little clowning with the 1 random TC they have in their uplink)

Djungelskog2:
This just seems like bloat. Its trying to solve a non-existent issue and just ends up making more (well it's not contra so it can't have come from a syndi despite it only being accessible in the uplink :clueless:). These are also just pre-existing implants being put into the uplink, and if anything the trombone and horn should be clown limited or put into the clown costume bundle (although I still disagree with that premise)

ArtisticRoomba:
In my opinion, finding an implanter is supposed to be a big wake-up call to security: you now know that implanters are "at play" and can deduce situations based on that knowledge.

Having these fake implanters does nothing but waste security's time even further than what's currently going on right now and is counterproductive to what implantation/extraction should be.

Note that I haven't played security in a while... it's just really odd seeing these changes from the perspective of recent changes.

A good april fools PR, but I think we should let the new implanters cook in the oven for two more weeks until we find out if the current implementation is Good:tm: or they need names. Otherwise, I don't think this is fit for master right now.

FairlySadPanda:
Under new meta Sec should be aware of implanter risk automatically, same as uplink risk. Just because you don't find an implanter doesn't mean a secoff should not be suspicious when they can't find a tot's emag.

All syndie implanters are, unless I'm misreading the YAML. implant-only. So you'd need access to a normal implanter that can extract to actually extract an implant. Given this is somewhat unlikely, it means that confirming with high probability some tot has an implant in them means that you end up with the impasse I explained in #35701.

This means you currently need admin intervention or some sort of meta-shield to prevent this being considered a violation of possessing syndie contra. If you can't safely get the contra out of them, and you know it is contra, then that's an ahelp question to clarify, as it's the same as someone walking out of Sec and pulling a gun from their uplink immediately.

ArtisticRoomba:
It sounds like you're creating a lot of unnecessary ambiguity in a department that is already punished quite heavily for speculation. For every 5 minutes the detective spends trying to """crack the case""", another 10 crimes happen. I just don't think it's a good fit for the current implant situation and you should wait a bit to see how the implant changes turn out in the long run.

FairlySadPanda:
That logic doesn't matter here. The detective still has to do the legwork because they still don't know which implant it is.

I would also point out that this does not impact the deduction gameplay at all, unless the 1 TC increase to do the same thing with an EMP implanter matters.

The only impact this has is to create mechanical, in-game, plausible deniability. Wizden targets LRP and MRP: mechanically, without having seen an open uplink, NT officers have no idea that these items do _not_ exist. They can't open the YAML - unless the argument is that NT's officers have knowledge of uplinks in general, the exact existence of every single item purchasable from an uplink, but not that uplinks are hidden in PDAs?

Regardless these are simply goofy implants for the Pointless section. Frankly, I just like the Sad Trombone implant and want to see it show up in game more. :D

mirrorcult:
i dont see how this creates any plausible deniability. the only method that would create actual plausible deniability would be if crew members could reasonably have access to implants, so that the ic-implication of "implant doesnt imply harmful syndie activity" is actually true. syndies having access to "non-harmful implants" doesnt remove the meta problem of "well.. if they have an implant theyre still obviously a syndie"

mirrorcult:
assuming that 'plausible deniability' is even desirable here, which for reasons above (its already hard enough for sec to make deductions faster than crimes happen) i am skeptical of

GoodWheatley:
As someone who’s played a lot of detective, here’s my take on this- “the detective just has to do the legwork” argument is a nonstarter because the detective is already vastly outmatched when it comes to catching a stealthy traitor. 

Back when I played, it was already _extremely_ rare to be able to find a spent implant in the first place. Several jobs already have access to uncounterable disposal methods (there is literally nothing you can do if salvage tosses their empty implanter into space), and every single crewmember on the station has the ability to scrub their implanter clean and flush it down disposals. But in exchange it was actually rewarding to find a hidden implanter- it’s a smoking gun that somebody’s implanted themselves and it gives you a direct answer to who did it. It’s a direct counter to traitors being sloppy and it gives the detective a win.

The implanter change completely removes this gameplay. Now there’s zero incentive for traitors to make an effort to hide the implanter, because even the detective does find it there’s nothing they can realistically do. 

Sure they could trail the suspect, but Roomba makes the good point that all the other traitors aren’t going to be nice and stop to let the detective track someone- crime is going to keep being committed and that lets the suspect get away. Even then trailing them wouldn’t really _do_ anything; you’d still have to see it be used to figure out which one it is and most implants can be vague as to what type they are even if you see them used it front of you (see: uplink/storage implant).

Introducing plausible deniability isn’t what this gameplay loop needs; if a traitor messes up badly enough that the detective has their DNA on a implanter, they should deserve to lose the implant. Making it even further of a gotcha game because now you have to guess if the syndicate implant is actually something illegal or if the traitor is just fucking with you isn’t fun for either side.

GoodWheatley:
> If plausible deniability isn't a thing for these that's IMO a good thing, as it means these just become goof implants to waste a floating TC on. 👍

But the whole point of these is that having them exist inherently creates plausible deniability. It makes what's already a painfully hard guessing game even worse because now you have to consider if the traitor's implant is even illegal in the first place.

FairlySadPanda:
Controversial and implants are not a final permanent design anyway. I'll close this so it doesn't clog the feed. Please re-open and use if desired.

PR #35789, opened 2025-03-12
Add support for antag-before-job selection

SlamBamActionman:
## About the PR

This PR adds support (but does *not* enable) for gamerules to select antags that should spawn as station jobs, before the jobs are selected. If a player gets chosen as such an antag they will be ineligible for any antag-immune job, and instead spawn as any of their other selected jobs (if available).

If a player is chosen as an antag but have no eligible roles selected (such as only selecting Command roles), it will treat it as if those roles were filled and whatever the player has chosen as default (spawn in as a Passenger or stay in lobby) will occur.

## Why / Balance

It's been posited that a number of players who would be otherwise willing to play Security/Command choose to not do so since they want to take the gamble on being an antagonist instead. By rolling antags before antag-immune roles, those players would be able to queue up as Sec/Command while still being in the pool for antagonists. 

There are a few concerns with this method:

If you have selected a Command role but spawn in as a non-antag-immune role *and* that Command role still isn't filled, it will clue you in that you have rolled a delayed-start antag. Specifically for Sec/Command this may be offset by those slots being filled by other players (again, the hope is to make those jobs more attractive to ready up for), but it may be more apparent with jobs that have multiple slots, such as the learner jobs. This would allow those players to get a few minutes' headstart to prepare before the role is revealed.

This could be offset by having a small chance to not get antag-immune roles regardless even if you aren't an antag, but is left as a possible future PR.

## Technical details

`AntagSelectionSystem` carries the bulk of these changes, in the form of the new antag selection timing called `IntraPlayerSpawn` (an inbetween of `PrePlayerSpawn` and `PostPlayerSpawn`, since it acts both before and after).

To support it, antag selection has been split up into pre-selecting the antag role and actually assigning the antag role. The old systems `Pre/PostPlayerSpawn` do these steps directly after each other, while for `IntraPlayerSpawn` the pre-selection is done before the player spawns in and the assignment done when the rule is started (either when spawning in or the delay is done). 

This means sessions can be in a pre-antag state where the system has them pre-selected but not yet assigned. `StationJobSystem.Roundstart` uses this list of pre-selected antags to disqualify them from antag-immune jobs.

Since the mind does not have the antag role component yet, any other antag rules must also check against this pre-selected antag list, to ensure they don't encroach on exclusive antags. 

## Steps to test

1. In `roundstart.yml`, replace the Traitor gamerule with the following:
``` yaml
- type: entity
  parent: BaseTraitorRule
  id: Traitor
  components:
  - type: GameRule
    minPlayers: 2
    delay:
      min: 1
      max: 2
  - type: AntagSelection
    selectionTime: IntraPlayerSpawn
    definitions:
    - prefRoles: [ Traitor ]
      max: 8
      playerRatio: 2
      blacklist:
        components:
        - AntagImmune
      lateJoinAdditional: true
      mindRoles:
      - MindRoleTraitor
```

2. Start a server and two clients in the `Release` configuration.
3. Run the following commands: `> cvar game.role_timers false`, `setgamepreset Traitor`.
4. Set both client players' High job preference to different non-antag roles (Captain, SecOff)
5. Set the job preference for Passenger at Low
6. Ensure both clients have Traitor enabled.
7. Start the round.

One of the clients should get their non-antag role, while the other spawns in as a Passenger.

Feel free to experiment with more clients and other gamerules!

## Media

https://github.com/user-attachments/assets/8abf1b8b-2b9e-4906-8286-c86b9ea79422

## Requirements

## Breaking changes

`AntagSelectionComponent` have had some variable names changed to reflect their new usage. 
`SelectionsComplete` -> `AssignmentComplete`
`SelectedSessions` -> `AssignedSessions`

Additionally, `ChooseAntags` no longer assigns antags to players, and must therefore be followed up by `AssignPreSelectedSessions`. For gamerules that spawn in antag entities/spawners, the behavior remains unchanged. 

**Changelog**

Not playerfacing, yet.

### DIFF
--- a/Content.Server/Antag/AntagSelectionSystem.cs
+++ b/Content.Server/Antag/AntagSelectionSystem.cs
@@ -282,7 +282,10 @@ public sealed partial class AntagSelectionSystem : GameRuleSystem<AntagSelection
             if (session == null)
                 MakeAntag(ent, null, def); // This is for spawner antags
             else
+            {
                 ent.Comp.PreSelectedSessions.Add(session); // Selection done!
+                Log.Debug($"Selected {session.Name} as antagonist: {ToPrettyString(ent)}");
+            }
         }
     }
 
@@ -419,7 +422,7 @@ public sealed partial class AntagSelectionSystem : GameRuleSystem<AntagSelection
             ent.Comp.AssignedMinds.Add((curMind.Value, Name(player)));
             SendBriefing(session, def.Briefing);
 
-            Log.Debug($"Selected {ToPrettyString(curMind)} as antagonist: {ToPrettyString(ent)}");
+            Log.Debug($"Assigned {ToPrettyString(curMind)} as antagonist: {ToPrettyString(ent)}");
         }
 
         var afterEv = new AfterAntagEntitySelectedEvent(session, player, ent, def);


Content.Server/Antag/Components/AntagSelectionComponent.cs
```diff
@@ -26,22 +32,27 @@ public sealed partial class AntagSelectionComponent : Component
...
+    /// Cached sessions of players who are chosen yet not given the role yet.
+    /// </summary>
+    public HashSet<ICommonSession> PreSelectedSessions = new();
+
     /// <summary>
     /// Cached sessions of players who are chosen. Used so we don't have to rebuild the pool multiple times in a tick.
     /// Is not serialized.
     /// </summary>
-    public HashSet<ICommonSession> SelectedSessions = new();
+    public HashSet<ICommonSession> AssignedSessions = new();
```
FairlySadPanda, line 56: Not much point me reviewing this as it's over my head but a lot of the line changes are just coping with this "selected" -> "assigned" name shift. You can reduce line changes by not doing this. This isn't that bad given this is restricted to AntagSelectionSystem but could still be annoying for a downstream to deal with.

SlamBamActionman: The `Selected` is no longer a good variable name for what the HashSet/bool is, as it reads ambiguous with the changes to how antags are first pre-selected and then assigned. `Assigned` is a more appropriate due to this.

PR #35794, opened 2025-03-12
Paradox Clone

slarticodefast:
## About the PR
Design doc: https://github.com/space-wizards/docs/pull/323

The paradox clone is a perfect copy of a random crewmember. They have the objective to kill and replace their original to fix the space-time paradox and will gib on roundend if they fail.

Thanks to @ps3moira for the cool intro sound and the visual effect.

Todo list for future PRs:
- Better spawn point selection. Currently they spawn on a random station tile, but get a toolbox and insuls in case they are trapped somewhere. They should spawn somewhere in maints or at arrivals instead.
- Better equipment copying. Currently they get a copy of everything the original has in their item slots, but storage is not copied recursively yet. This will need a larger yaml cleanup for StorageFill and similar systems first.
- Copy implants.
- Copy mind roles and antag status, maybe also objectives.
- admin antag control verb

## Why / Balance
classic SS13 antag

## Technical details
Uses the cloning code refactored in https://github.com/space-wizards/space-station-14/pull/35555/ to spawn a copy of a random player.
Most of the code are just the usual game rule, event and mind role stuff.

I atomized some parts of this code into their own PRs:

https://github.com/space-wizards/space-station-14/pull/35795
https://github.com/space-wizards/space-station-14/pull/35796
https://github.com/space-wizards/space-station-14/pull/35802

## Media

gamerule and spawning

https://github.com/user-attachments/assets/2a3c0cd9-eb4a-4909-90c6-b52d43a274ac

gibbing on roundend with failed objective

https://github.com/user-attachments/assets/b5647392-953b-4faf-9e8e-01e90cfb61c0

successful objective

https://github.com/user-attachments/assets/94417110-1c3e-4ab8-8a68-5a8a019c3dc6

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
moved the `paradox_clone.rsi` folder

**Changelog**
:cl: slarticodefast, ps3moira
- add: Added the paradox clone as a ghostrole antagonist.

PR #35796, opened 2025-03-12
polymorph popup fixes

slarticodefast:
Polymorph has received some bugfixes:
PolymorphSystem tries to unbuckle the target before polymorphing, but this can lead to a failed resolve in the case of the wand of entrance, since walls don't have BuckleComponent. We set logging to false for that resolve to fix it.
Also added a missing popup, which had a ftl string that was never used. The popup LocIds were moved into the prototype so we can set custom ones.

Required for https://github.com/space-wizards/space-station-14/pull/35794

### DIFF
--- a/Content.Server/Polymorph/Systems/PolymorphSystem.cs
+++ b/Content.Server/Polymorph/Systems/PolymorphSystem.cs
@@ -204,6 +204,12 @@ public sealed partial class PolymorphSystem : EntitySystem
 
         var child = Spawn(configuration.Entity, _transform.GetMapCoordinates(uid, targetTransformComp), rotation: _transform.GetWorldRotation(uid));
 
+        if (configuration.PolymorphPopup != null)
+            _popup.PopupEntity(Loc.GetString(configuration.PolymorphPopup,
+                ("parent", Identity.Entity(uid, EntityManager)),
+                ("child", Identity.Entity(child, EntityManager))),
+                child);
+
         MakeSentientCommand.MakeSentient(child, EntityManager);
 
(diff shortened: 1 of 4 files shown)


Content.Server/Polymorph/Systems/PolymorphSystem.cs
```diff
@@ -347,10 +353,11 @@ private void OnDestruction(Entity<PolymorphedEntityComponent> ent, ref Destructi
         var ev = new PolymorphedEvent(uid, parent, true);
         RaiseLocalEvent(uid, ref ev);
 
-        _popup.PopupEntity(Loc.GetString("polymorph-revert-popup-generic",
+        if (component.Configuration.ExitPolymorphPopup != null)
+            _popup.PopupEntity(Loc.GetString(component.Configuration.ExitPolymorphPopup,
                 ("parent", Identity.Entity(uid, EntityManager)),
                 ("child", Identity.Entity(parent, EntityManager))),
-            parent);
+                parent);
```
ScarKy0, line 360: It was present before, but the popup is a bit broken, saying "The Kristopher Laborde reverted back into Kristopher Laborde" in my case. Think that'd be easily fixable?

ScarKy0 (changes requested):

Content.Server/Polymorph/Systems/PolymorphSystem.cs
ScarKy0: Nevermind it was an issue with the entity being also renamed. Oops.

ScarKy0 (approved):
Looks good, works

PR #26978, opened 2024-04-15
Revert Exterminator pending newmed/redesign

SlamBamActionman:
## About the PR

Reverts the Exterminator midround antag #19946, pending newmed and a redesign.

This removes all references and functionality of the Exterminator from the game.

## Why / Balance

1) The Exterminator code is incompatible with newmed due to utilizing non-standard .yaml inheritance. "unironically if I disable/depreciate terminators newmed would come out like 5 days earlier"
https://discord.com/channels/310555209753690112/1008709214006427689/1229307018800795678 

2) Current Exterminator design encourages murderboning, with equipment to take on the entire station. The initial design concept stated the intention was to have little to no crossfire, but the implementation in actual gameplay does not reflect this. 
https://discord.com/channels/310555209753690112/1008709214006427689/1229308591710928936
https://discord.com/channels/310555209753690112/1008709214006427689/1229407444200067092

3) Code directly references the Terminator, failing to utilize a replacement where it would be necessary for copyright reasons.

## Technical details

Reverts #19946, keeping some minor positive changes to spelling and formatting.

## Media

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- remove: Removed Exterminators pending redesign. 🔥👍🔥

PR #35802, opened 2025-03-12
Cleanup Objective files, add PickSpecificPersonComponent

slarticodefast:
## About the PR
Atomized from https://github.com/space-wizards/space-station-14/pull/35794

## Technical details
Cleans up objective systems and components.
Previously the objective target assignment subscriptions were all in the same files as the conditions that they were used with in the case of traitors. However, this file structure made no sense, as they can be arbitrarily combined (for example a "protect a random head objective" can be made by combining two components). I moved all these subscriptions into the same file to make this more comprehensible.

Re-added a target picking component based on https://github.com/space-wizards/space-station-14/pull/26978, but made it more reusable and non-terminator specific.

## Requirements

## Breaking changes
only moved some code, no public API functions though

**Changelog**
no player facing changes

### DIFF
--- a/Content.Server/Objectives/Systems/HelpProgressConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/HelpProgressConditionSystem.cs
@@ -18,7 +18,6 @@ public sealed class HelpProgressConditionSystem : EntitySystem
         base.Initialize();
 
         SubscribeLocalEvent<HelpProgressConditionComponent, ObjectiveGetProgressEvent>(OnGetProgress);
-
     }
 
     private void OnGetProgress(EntityUid uid, HelpProgressConditionComponent comp, ref ObjectiveGetProgressEvent args)
--- a/Content.Server/Objectives/Systems/HelpProgressConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/HelpProgressConditionSystem.cs
@@ -1,23 +1,17 @@
-using Content.Server.GameTicking.Rules;
 using Content.Server.Objectives.Components;
 using Content.Shared.Mind;
 using Content.Shared.Objectives.Components;
 using Content.Shared.Objectives.Systems;
-using Content.Shared.Roles.Jobs;
-using Robust.Shared.Random;
-using System.Linq;
 
 namespace Content.Server.Objectives.Systems;
 
 /// <summary>
-/// Handles help progress condition logic and picking random help targets.
+/// Handles help progress condition logic.
 /// </summary>
 public sealed class HelpProgressConditionSystem : EntitySystem
 {
-    [Dependency] private readonly IRobustRandom _random = default!;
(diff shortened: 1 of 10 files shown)


Content.Server/Objectives/Systems/HelpProgressConditionSystem.cs
```diff
@@ -1,31 +1,24 @@
...
-    [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly SharedObjectivesSystem _objectives = default!;
     [Dependency] private readonly TargetObjectiveSystem _target = default!;
-    [Dependency] private readonly TraitorRuleSystem _traitorRule = default!;
 
     public override void Initialize()
     {
         base.Initialize();
 
         SubscribeLocalEvent<HelpProgressConditionComponent, ObjectiveGetProgressEvent>(OnGetProgress);
```
beck-thompson, line 21: Extra space

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
```diff
@@ -0,0 +1,208 @@
...
+using Content.Shared.Mind;
+using Content.Shared.Objectives.Components;
+using Content.Server.GameTicking.Rules;
+using Content.Server.Revolutionary.Components;
+using Robust.Shared.Random;
+using System.Linq;
+
+namespace Content.Server.Objectives.Systems;
+
+public sealed class PickObjectiveTargetSystem : EntitySystem
```
beck-thompson, line 11: Add a comment for this one because it isn't really associated with any specific component!

beck-thompson (approved):
Looks good!

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
```diff
@@ -0,0 +1,212 @@
...
+        if (!TryComp<TargetOverrideComponent>(user, out var targetComp) || targetComp.Target == null)
+        {
+            args.Cancelled = true;
+            return;
+        }
+
+        _target.SetTarget(ent.Owner, targetComp.Target.Value);
+    }
+
+    private void OnRandomPersonAssigned(Entity<PickRandomPersonComponent> ent, ref ObjectiveAssignedEvent args)
```
deltanedas, line 63: make these pass a predicate to a shared function that has 90% of the code instead of keeping it copy pasted

deltanedas (changes requested):

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
slarticodefast: Yeah, that could use more cleanup
for now I just moved stuff so you can actually find the code where it is supposed to be

PR #35813, opened 2025-03-13
Entity<T> API for singularity related systems

TemporalOroboros:
## About the PR

Fixed two obsolete method compile warnings.
Also, the entire singularity/event horizon/gravity well API now uses `Entity<T>`s instead of separate uids and components.

## Why / Balance

I was trying to clear up the use of an obsolete `MapGridComponent.GetAnchoredEntities` method and had to change one of the methods to take an `Entity<T>` instead of a `MapGridComponent` to avoid using `IComponent.Owner`. Scope creep happened.

## Technical details

Every public facing method for `SharedSingularitySystem`, `SharedEventHorizonSystem`, and `SharedGravityWellSystem` and their server/client equivalents that was using the `(uids, args..., components)` parameter format was converted to use the `(Entity<T>, args...)` parameter format. The old API still exists to prevent breaking changes, it's just been deprecated and moved to the end of the system definition. Internal methods were also changed to match.

The events related to singularity state/event horizon consumption have been changed to use `Entity<T>`s where they were using separate uids and components.

## Requirements

## Breaking changes

The event horizon and singularity related events now use `Entity<T>`s to represent the event horizon/singularity entities instead of separate `EntityUid`s and components. Any event handlers that this breaks should switch the broken member accesses to reference the entity and its `.Owner` or `.Comp` properties.


Content.Server/Singularity/EntitySystems/EventHorizonSystem.cs
```diff
@@ -232,83 +231,84 @@ public void ConsumeEntitiesInContainer(EntityUid hungry, BaseContainer container
...
     }
 
     /// <summary>
     /// Checks whether an event horizon can consume a given tile.
     /// This is only possible if it can also consume all entities anchored to the tile.
     /// </summary>
-    public bool CanConsumeTile(EntityUid hungry, TileRef tile, MapGridComponent grid, EventHorizonComponent eventHorizon)
+    public bool CanConsumeTile(Entity<EventHorizonComponent> hungry, Entity<MapGridComponent> grid, TileRef tile)
     {
-        foreach (var blockingEntity in grid.GetAnchoredEntities(tile.GridIndices))
```
TemporalOroboros, line 285: All of this to fix this one obsolete method call.

PR #35816, opened 2025-03-13
[april fools] nuke timer calibration station event

LaCumbiaDelCoronavirus:
## About the PR
Added a new common station event: nuke calibration
When this event happens (provided that there's a nuke on station, and it's not armed yet):
0. Try to anchor the nuke, proceed with the event if successful in doing so.
1. Nuke is armed. Nuke cannot be disarmed at this moment, unless the disk was already in it
2. An announcement is made 15 seconds later, notifying crew of 'routine nuke timer calibration'
3. 25 seconds later, there is a 50% chance for either:
- The nuke to be disarmed
- The nuke stays armed, but the requirement for the disk to be inside to be disarmed is removed (so it can be disarmed)

The nuke timer is reset back to its default (300 seconds) after the event ends.

_**The rarity of this event is intentionally set low.**_

## Why / Balance
It's funny :trollface: 

## Technical details
Added NukeCalibrationRule.cs, NukeCalibrationRuleComponent.cs, and nuke-calibration.ftl
Added 2 new vars to NukeComponent:
- DiskBypassEnabled
- ShouldResetDiskBypass

Added new method SetDiskBypassEnabled on NukeSystem, which sets DiskBypassEnabled and ShouldResetDiskBypass

If the event rolls to automatically defuse the nuke, remaining time is set _before_ checking if the nuke is armed and then disarming it. This is so that the nuke timer is reset, if it was somehow disarmed manually.

The 50% chance for disarm/arm is rolled when the event ends, with the first announcement being set 15 seconds into the event. The event is, internally, 40 seconds long.

## Media
https://www.youtube.com/watch?v=UGst4wj0E30

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Due to shortage of nuke inspectors, centcom will now perform routine calibrations of the station nuke timer.

### DIFF
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -243,7 +243,9 @@ public sealed class NukeopsRuleSystem : GameRuleSystem<NukeopsRuleComponent>
 
     private void OnNukeDisarm(NukeDisarmSuccessEvent ev)
     {
-        CheckRoundShouldEnd();
+        // Do not end the round if the bomb was armed (and disarmed) as part of the nuke calibration event
+        if(ev.CheckRoundShouldEnd)
+            CheckRoundShouldEnd();
     }
 
     private void OnComponentRemove(EntityUid uid, NukeOperativeComponent component, ComponentRemove args)
@@ -419,9 +421,6 @@ public sealed class NukeopsRuleSystem : GameRuleSystem<NukeopsRuleComponent>
         foreach (var nuke in EntityQuery<NukeComponent>())
         {
             if (nuke.Status == NukeStatus.ARMED)
-                return;
-            // If we're bypassing the requirement for a nuke disk in order to arm/disarm the nuke, don't end the round. This makes sure that disarming the nuke doesn't end the round after a nuke calibration.
-            else if (nuke.DiskBypassEnabled)
                 return;
         }
 
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -420,8 +420,8 @@ public sealed class NukeopsRuleSystem : GameRuleSystem<NukeopsRuleComponent>
         {
             if (nuke.Status == NukeStatus.ARMED)
                 return;
-            // So if the nuke isn't armed, and the disk isn't inside, don't end the round. This is so that nuke calibration event doesn't end the round upon the nuke being disarmed.
-            else if (!nuke.DiskSlot.HasItem)
+            // If we're bypassing the requirement for a nuke disk in order to arm/disarm the nuke, don't end the round. This makes sure that disarming the nuke doesn't end the round after a nuke calibration.
+            else if (nuke.DiskBypassEnabled)
                 return;
         }
 
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -420,6 +420,9 @@ public sealed class NukeopsRuleSystem : GameRuleSystem<NukeopsRuleComponent>
         {
             if (nuke.Status == NukeStatus.ARMED)
                 return;
+            // So if the nuke isn't armed, and the disk isn't inside, don't end the round. This is so that nuke calibration event doesn't end the round upon the nuke being disarmed.
+            else if (!nuke.DiskSlot.HasItem)
+                return;
         }
 
         var shuttle = GetShuttle((ent, ent));
--- a/Content.Server/Nuke/NukeComponent.cs
+++ b/Content.Server/Nuke/NukeComponent.cs
@@ -1,5 +1,7 @@
 using System.Threading;
 using Content.Server.Explosion.EntitySystems;
+using Content.Server.StationEvents.Components;
+using Content.Server.StationEvents.Events;
 using Content.Shared.Containers.ItemSlots;
 using Content.Shared.Explosion;
 using Content.Shared.Nuke;
@@ -170,6 +172,20 @@ namespace Content.Server.Nuke
         /// </summary>
         [ViewVariables]
         public NukeStatus Status = NukeStatus.AWAIT_DISK;
+
+        /// <summary>
+        ///     Should we allow disarming/arming, even if there's no disk?
+        ///     Does not allow anchoring/unanchoring without the disk.
+        ///     Arming nuke still requires code to be entered.
+        /// </summary>
+        [ViewVariables]
+        public bool DiskBypassEnabled = false;
+
+        /// <summary>
+        ///     Should we reset the disk bypass value to false when the nuke is disarmed?
+        /// </summary>
+        [ViewVariables]
+        public bool ShouldResetDiskBypass = false;
 
         /// <summary>
         ///     Check if nuke has already played the nuke song so we don't do it again
--- a/Content.Server/Nuke/NukeComponent.cs
+++ b/Content.Server/Nuke/NukeComponent.cs
@@ -182,10 +182,10 @@ namespace Content.Server.Nuke
         public bool DiskBypassEnabled = false;
 
         /// <summary>
-        ///     Should we reset the disk bypass value to false when the nuke is disarmed?
+        ///     Should we reset the disk bypass value, and the nuke timer, to their defaults when the nuke is disarmed?
         /// </summary>
         [ViewVariables]
-        public bool ShouldResetDiskBypass = false;
+        public bool ShouldResetAfterDiskBypass = false;
 
         /// <summary>
         ///     Check if nuke has already played the nuke song so we don't do it again
--- a/Content.Server/Nuke/NukeSystem.cs
+++ b/Content.Server/Nuke/NukeSystem.cs
@@ -540,10 +540,13 @@ public sealed class NukeSystem : EntitySystem
         component.Status = NukeStatus.COOLDOWN;
         component.CooldownTime = component.Cooldown;
 
-        if (component.ShouldResetDiskBypass == true)
+        if (component.ShouldResetAfterDiskBypass == true)
+        {
             component.DiskBypassEnabled = false;
+            component.RemainingTime = component.Timer;
+        }
 
-        component.ShouldResetDiskBypass = false;
+        component.ShouldResetAfterDiskBypass = false;
 
         UpdateUserInterface(uid, component);
         UpdateAppearance(uid, component);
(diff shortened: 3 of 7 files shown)


SlimmSlamm:
I think the rarity should be a little higher considering this is a one day change. If it's too low you won't even see it at all.

TeenSarlacc:
> I think the rarity should be a little higher considering this is a one day change. If it's too low you won't even see it at all.

are the April fools merges gonna be just for April fools, or are they gonna drop with the regular update schedule, and get removed the following update?

TeenSarlacc:
> I think the rarity should be a little higher considering this is a one day change. If it's too low you won't even see it at all.

are the April fools merges gonna be just for April fools, or are they gonna drop with the regular update schedule, and get removed the following update?

SlimmSlamm:
> > I think the rarity should be a little higher considering this is a one day change. If it's too low you won't even see it at all.
> 
> are the April fools merges gonna be just for April fools, or are they gonna drop with the regular update schedule, and get removed the following update?

As far as I know the branch will be enabled on the servers for like a day or two and then it'll go back to normal until next year

LaCumbiaDelCoronavirus:
> I think the rarity should be a little higher considering this is a one day change. If it's too low you won't even see it at all.

By the rarity being low, i was saying that it's common, not that there's a low chance to roll it. I set the weight to the highest out of all the events which makes it a tad more common... unless i'm misunderstanding the way that weights work

Nukesthestation:
ngl i think this should be a non April fools pr :trollface: from an ic standpoint it makes sense that NT would want to make sure that the nukes they provide to the station works/are safe, and on the point of it not auto-disarming, it would also be a nice ic thing, stuff breaks down the whole point of the event is to test the nuke according to the announcement, it also would add some rare but funny chaos to a round, but hey what do I know? I'm just a person who likes to play the funny spaceman gaem

walksanatora:
tbh a random arming of the nuke sounds fun.

walksanatora:
now imagine this occuring during a nukie round. >:)

LaCumbiaDelCoronavirus:
> ngl i think this should be a non April fools pr :trollface: from an ic standpoint it makes sense that NT would want to make sure that the nukes they provide to the station works/are safe, and on the point of it not auto-disarming, it would also be a nice ic thing, stuff breaks down the whole point of the event is to test the nuke according to the announcement, it also would add some rare but funny chaos to a round, but hey what do I know? I'm just a person who likes to play the funny spaceman gaem

I thought about it and I'll actually probably try and PR it on master in its current state (ofcourse much rarer as an event though), don't expect it to get merged easily

If i really wanted it approved for non april fools, I'd probably need to make alot of compromises that would be against the silliness aspect of this PR ;(

ArcaneOcto:
Haha I love this. I agree this should just be a rare event in Master. Maybe without the random chance though, and it should be disarmable immediately. Command having to worry about disarming the nuke out of nowhere seems like peak gameplay. Imagine if command is busy and AI has to get some random tider to disarm it haha

LaCumbiaDelCoronavirus:
> ... and it should be disarmable immediately.

The more realistic point of this event is to distract people, by making them go disarm the nuke. Having a (almost always) guaranteed 40 seconds of distraction, not counting the possible however much time you have after that to disarm the nuke if it rolled to stay armed after those 40 seconds, is the point of this

FairlySadPanda:
Ideally this should be disabled if the round type is zombies.

Errant-4:
> Ideally this should be disabled if the round type is zombies. You don't want a zombie horde to get screwed over by RNG.

I would argue that in fact would make for an unforgettable zombie round

walksanatora:
or let zombies disarm. but make it take like 2x as long.

Errant-4:
No, the thing is, playing a zombified crew and have the station blow up would be, in my opinion, a really funny end to the round. People care about greentext too much, let it all burn

walksanatora:
let zombies eat the nuke if it is armed. requires 3 zombies all eating the nuke at once.

Nukesthestation:
> No, the thing is, playing a zombified crew and have the station blow up would be, in my opinion, a really funny end to the round. People care about greentext too much, let it all burn. Zombies disarming the nuke would be preeetty "immersion breaking", so I don't think it should be a thing unless there was a strong game design reason to need it. And imo there isn't

I agree with this also why would zombies go to the nuke, its not a brain and ontop of that how/why would they disarm a nuke they are looking for brains not an active nuke

TeenSarlacc:
> let zombies eat the nuke if it is armed. requires 3 zombies all eating the nuke at once.

Errant-4 (changes requested):
We generally love the idea, and would not nitpick much since it's for april fools, but there is one issue that I think needs to be addressed first.

I tried it for a bit, and if a calibration starts during a nukie round it has the potential to end the round before the crew is even aware there _are_ nukies. If the calibration fails and the crew has to manually disarm the bomb, this will count as a round ending defusal, and call the shuttle. 

I do think that calibration events have the potential to interact with nukie rounds in interesting and funny ways, but an event randomly starting that has a 50% potential to end the round with almost no counterplay on the part of the nukies since they have to get to the Vault in the span of one minute to be able guard it, is perhaps too much. 

The easiest way to prevent this would probably be to make NukeOpsRuleSystem.CheckRoundShouldEnd() exit if the disk is not in the nuke. (And, uh, make sure that a powergaming captain can't insert the disk after the nuke started calibration)

LaCumbiaDelCoronavirus:
> (And, uh, make sure that a powergaming captain can't insert the disk after the nuke started calibration)

IIRC from testing, you can't insert the disk while it's armed for whatever reason anyway, might be misremembering though.

> If the calibration fails and the crew has to manually disarm the bomb, this will count as a round ending defusal, and call the shuttle.

Uh oh. I'll have to fix that then

LaCumbiaDelCoronavirus:
I deny the existence of that test fail... i don't think it's real..

LaCumbiaDelCoronavirus:
Nevermind, test fail very real

Errant-4:
Do you think you can fix this or do you need help? This is a very fun idea so I would like to see it happen

LaCumbiaDelCoronavirus:
we are so back

Errant-4:
@LaCumbiaDelCoronavirus 
This fix doesn't actually work, please test your changes
The problem is that the DiskBypassEnabled is getting reset by DisarmBomb() before it is actually checked in CheckRoundShouldEnd(), so it does nothing for round end.

After running over some possible ways to catch this scenario in CheckRoundShouldEnd(), and how each way might conflict with something else that's happening in a nukie roundd , I arrived at the conclusion that we should simply not play the game at all:
NukeSystem.OnDoAfter() should read the disk bypass value into a local variable before calling OnNukeDisarm(), and send out this not-reset value in the NukeDisarmSuccessEvent. 
Then, NukeopsRuleSystem.OnNukeDisarm() can simply check bypass from the event args and not call CheckRoundShouldEnd()
(Don't forget to remove the extra check you put in CheckRoundShouldEnd() )

If you would like, I can PR the above fix to your repo, so you just have to press a button

LaCumbiaDelCoronavirus:
> @LaCumbiaDelCoronavirus This fix doesn't actually work, please test your changes The problem is that the DiskBypassEnabled is getting reset by DisarmBomb() before it is actually checked in CheckRoundShouldEnd(), so it does nothing for round end.
> 
> After running over some possible ways to catch this scenario in CheckRoundShouldEnd(), and how each way might conflict with something else that's happening in a nukie roundd , I arrived at the conclusion that we should simply not play the game at all: NukeSystem.OnDoAfter() should read the disk bypass value into a local variable before calling OnNukeDisarm(), and send out this not-reset value in the NukeDisarmSuccessEvent. Then, NukeopsRuleSystem.OnNukeDisarm() can simply check bypass from the event args and not call CheckRoundShouldEnd() (Don't forget to remove the extra check you put in CheckRoundShouldEnd() )
> 
> If you would like, I can PR the above fix to your repo, so you just have to press a button

Oh. Well then, I'd probably move NukeDisarmSuccessEvent to be raised in DisarmBomb() itself [since i see no reason for the event to be raised in OnDoAfter(), but please tell me if there is one] and pass DiskBypassEnabled to the event there.

Also yeah, having the fix PR'd there would be nice(, only if it's convenient for you!)

Errant-4:
> Oh. Well then, I'd probably move NukeDisarmSuccessEvent to be raised in DisarmBomb() itself [since i see no reason for the event to be raised in OnDoAfter(), but please tell me if there is one] and pass DiskBypassEnabled to the event there.

I think it makes sense to keep raising the event outside of disarm, because that way you don't need to LOOK in DisarmBomb to find that an event is raised about it's completion. Kind of making for a better "high level view"

Errant-4 (approved):
Thank you for your work this is a funny event

LaCumbiaDelCoronavirus:
It's time for a bit of trolling

Errant-4:
@LaCumbiaDelCoronavirus Hi, we have decided we would like to keep this event in the game, would you like to make the PR to the master branch? With the gamerule weight lowered to 2, and AutoDisarmChance increased to 0.66

(Note, it will still have to undergo the usual maint review and it's possible that we will decide the april-review was not thorough enough and some of the code has to be changed, but we are definitely interested in the design concept)

walksanatora:
YES YES YES. I loved this event during the april fools event. forcing atleast one command member to go to vault has many uses both for RP and Antags.

LaCumbiaDelCoronavirus:
> @LaCumbiaDelCoronavirus Hi, we have decided we would like to keep this event in the game, would you like to make the PR to the master branch? With the gamerule weight lowered to 2, and AutoDisarmChance increased to 0.66
> 
> (Note, it will still have to undergo the usual maint review and it's possible that we will decide the april-review was not thorough enough and some of the code has to be changed, but we are definitely interested in the design concept)

I would, but consider the idea of:
Two separate forms of the event, like one:
At a 300 second nuke timer, but a much longer delayed centcom announcement and probably 66% chance of self disarm. Decently common

And the other:
180 second timer and otherwise the same as it is, though with a much lower self disarm chance and much rarer (a weight lower than 2 or 3). Maybe even a longer delay for the chance to disarm it to appear


Idea is to have a common, easier to handle (300 second timer and gives more time to respond, while not letting you disarm it too early) event AND the rarer 180 or 170 second one that gives you way less time to respond.

I wish to keep them rather uncommon, to have a way lower chance to self disarm (to have higher stakes ofcourse) than common, but with a high chance to self disarm (which makes it seem less urgent)

Errant-4:
During the internal discussion about the idea of bringing this event back, multiple staff felt that it should definitely not be "too common", as that would make the event mundane/potentially annoying. But no one seemed to have an issue with weight 2 (that's in the ballpark of once every 8-10 hours on average) 
If you would like to pitch multiple different variants of the event, you can PR it as such - you can PR it any way you want, of course. But if you keep their combined weight at 2, then you will likely have an easier time getting it approved, since no one seemed to have an issue with that, but might choose to oppose it if it's more frequent. I can't predict or quantify this :shrug:
Either way, further discussion about this should be on the PR in question so it's easier to find

LaCumbiaDelCoronavirus:
I forgot about this but i will say im still going to, eventually, port this to normal ss14

KaiserMaus:
@LaCumbiaDelCoronavirus Have you forgotten about us?

PR #35820, opened 2025-03-13
Support separate displacement maps for left and right hand

TheShuEd:
## About the PR

HandDisplacement is now used as Fallback unless special displacements for left and right hand are specified. So without Breaking Changes

## Why / Balance

useful for more complex species silhouettes when the arms are in completely different positions

## Media

For example: The items in the hands are more widely spaced than in a standard doll.
image

## Requirements


beck-thompson (approved):
If we eventually have like species with more hands it'll have to get refactored but I think its fine :+1:

PR #35825, opened 2025-03-14
Require traitors to maroon their objective no matter what

notquitehadouken:
## About the PR

The kill head objective now requires you to kill AND maroon the head, and the kill person objective only requires you to maroon them.

The paradox clone objective is unaffected (Mentioned because it uses `KillPersonCondition`)

## Why / Balance

Traitors will no longer be incentivized to kill their one kill objective just before they end up in CC, as doing this now doesn't actually complete the kill objective.

They will now be required to make sure they don't arrive at CC, even if they're dead. Being thrown out of the shuttle during transit counts as being marooned.

## Media

image

(for context, dev has shuttle disabled)
image

## Requirements

## Breaking changes

No prototypes were renamed as they are not player-facing.

Do note that if your server lacks an evac shuttle for some reason (by way of CCvar `shuttle.emergency` being false), maroon-only objectives will instead require the target to be dead. Be aware, as this is not shown to players.

**Changelog**

:cl:
- tweak: Kill objectives will now always require you to maroon your target.

### DIFF
--- a/Content.Server/Objectives/Components/KillPersonConditionComponent.cs
+++ b/Content.Server/Objectives/Components/KillPersonConditionComponent.cs
@@ -10,8 +10,14 @@ namespace Content.Server.Objectives.Components;
 public sealed partial class KillPersonConditionComponent : Component
 {
     /// <summary>
-    /// Whether the target must be truly dead, ignores missing evac.
+    /// Whether the target must be dead
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
     public bool RequireDead = false;
+
+    /// <summary>
+    /// Whether the target must not be on evac
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public bool RequireMaroon = true;
 }
--- a/Content.Server/Objectives/Components/KillPersonConditionComponent.cs
+++ b/Content.Server/Objectives/Components/KillPersonConditionComponent.cs
@@ -18,6 +18,6 @@ public sealed partial class KillPersonConditionComponent : Component
     /// <summary>
     /// Whether the target must not be on evac
     /// </summary>
-    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    [DataField]
     public bool RequireMaroon = true;
 }
--- a/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
@@ -39,9 +39,14 @@ public sealed class KillPersonConditionSystem : EntitySystem
             return 1f;
 
         var targetDead = _mind.IsCharacterDeadIc(mind);
-        var targetMarooned = !_config.GetCVar(CCVars.EmergencyShuttleEnabled) ||
-            (!_emergencyShuttle.IsTargetEscaping(target) &&
-             _emergencyShuttle.ShuttlesLeft) ;
+        var targetMarooned = (!_emergencyShuttle.IsTargetEscaping(target) &&
+                              _emergencyShuttle.ShuttlesLeft) ;
+        var shuttleDisabled = _config.GetCVar(CCVars.EmergencyShuttleEnabled);
+        if (shuttleDisabled && requireMaroon)
+        {
+            requireDead = true;
+            requireMaroon = false;
+        }
 
         if (requireMaroon)
         {
--- a/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
@@ -29,29 +29,23 @@ public sealed class KillPersonConditionSystem : EntitySystem
         if (!_target.GetTarget(uid, out var target))
             return;
 
-        args.Progress = GetProgress(target.Value, comp.RequireDead);
+        args.Progress = GetProgress(target.Value, comp.RequireDead, comp.RequireMaroon);
     }
 
-    private float GetProgress(EntityUid target, bool requireDead)
+    private float GetProgress(EntityUid target, bool requireDead, bool requireMaroon)
     {
         // deleted or gibbed or something, counts as dead
         if (!TryComp<MindComponent>(target, out var mind) || mind.OwnedEntity == null)
             return 1f;
 
-        // dead is success
-        if (_mind.IsCharacterDeadIc(mind))
-            return 1f;
-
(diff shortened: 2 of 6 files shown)


Content.Server/Objectives/Components/KillPersonConditionComponent.cs
```diff
@@ -10,8 +10,14 @@ namespace Content.Server.Objectives.Components;
...
-    /// Whether the target must be truly dead, ignores missing evac.
+    /// Whether the target must be dead
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
     public bool RequireDead = false;
+
+    /// <summary>
+    /// Whether the target must not be on evac
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
```
slarticodefast, line 21:
```suggestion
    [DataField]
```
DataField incluse VV RW

Content.Server/Objectives/Components/KillPersonConditionComponent.cs
```diff
@@ -10,8 +10,14 @@ namespace Content.Server.Objectives.Components;
...
+    public bool RequireMaroon = true;
```
slarticodefast, line 22: I would make this false by default so we don't create a breaking change for forks
adjust the yaml code accordingly

slarticodefast:
Didn't look at the progress logic yet.

TeenSarlacc:
out of curiosity, if you were to bring a characters' brain on the evac shuttle, (or gib someone on the shuttle) would that count as a fail?

notquitehadouken:
> out of curiosity, if you were to bring a characters' brain on the evac shuttle, (or gib someone on the shuttle) would that count as a fail?

no idea :trollface:

notquitehadouken:
image

it counts as a success

notquitehadouken:
image

walksanatora:
can you clone from brain?

Pronana:
> can you clone from brain?

Nope.

Resources/Prototypes/Objectives/base_objectives.yml
```diff
@@ -52,6 +52,7 @@
       components:
       - SocialObjective
   - type: KillPersonCondition
+    requireMaroon: false
```
slarticodefast, line 55:
```suggestion
```
the default set in the component is already `false`

Resources/Prototypes/Objectives/paradoxClone.yml
```diff
@@ -32,7 +32,8 @@
   components:
   - type: PickSpecificPerson
   - type: KillPersonCondition
-    requireDead: true # don't count missing evac as killing
+    requireDead: true
+    requireMaroon: false # maintain original behavior
```
slarticodefast, line 36:
```suggestion
```
same here

tutoumi:
i don't uderstand the difference but ok

walksanatora:
> i don't uderstand the difference but ok

basically it would prevent evac DAGD bombing to kill the objective. as their body still washes up at central command and then are (presumably) cloned and you fail your objective.

beck-thompson (approved):

Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
```diff
@@ -29,36 +29,35 @@ private void OnGetProgress(EntityUid uid, KillPersonConditionComponent comp, ref
...
+                              _emergencyShuttle.ShuttlesLeft) ;
+        if (!_config.GetCVar(CCVars.EmergencyShuttleEnabled) && requireMaroon)
+        {
+            requireDead = true;
+            requireMaroon = false;
+        }
+
+        if (requireMaroon)
+        {
+            if (requireDead && !targetDead)
```
ciaran, line 52: I think this could be simplified a bit if this condition is moved above outside of the if, since both branches seem to have the same test.

Resources/Prototypes/Objectives/traitor.yml
```diff
@@ -90,26 +90,27 @@
...
   - type: TargetObjective
-    title: objective-condition-kill-person-title
+    title: objective-condition-maroon-person-title
   - type: PickRandomPerson
 
 - type: entity
   parent: [BaseTraitorObjective, BaseKillObjective]
   id: KillRandomHeadObjective
-  description: We need this head gone and you probably know why. Good luck, agent.
+  description: We need this head gone and you probably know why. Make sure they don't make it to centcomm, even if they're dead. Good luck, agent.
```
ciaran, line 99: "CentComm" appears to be the common way to write this but it is a bit mixed

Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
```diff
@@ -29,36 +29,30 @@ private void OnGetProgress(EntityUid uid, KillPersonConditionComponent comp, ref
...
-        if (_emergencyShuttle.IsTargetEscaping(mind.OwnedEntity.Value))
-            return 0f;
-
-        // evac has left without the target, greentext since the target is afk in space with a full oxygen tank and coordinates off.
-        if (_emergencyShuttle.ShuttlesLeft)
-            return 1f;
+        if (requireMaroon)
+            return targetMarooned ? 1f : _emergencyShuttle.EmergencyShuttleArrived ? 0.5f : 0f;
 
-        // if evac is still here and target hasn't boarded, show 50% to give you an indicator that you are doing good
```
ciaran, line 61: Keeping this comment is probably useful

ciaran (approved):

Sparlight:
will the monkey's paw curl on this one to see the Evac shuttle filled with syndicate bombs to entirely destroy the Evac shuttle? If the shuttle catastrophically loses all its floor, nobody ends up getting home.
Time will surely tell. But I do like the direction of the PR.

Content.Server/Objectives/Components/KillPersonConditionComponent.cs
slarticodefast: me when my reviews get ignored 😢

beck-thompson: man I didn't see this! I was so focused making sure the logic made sense :cry:

slarticodefast: Could be worse I guess.
Thanks for giving it a review 🫡

notquitehadouken:
crazy how the pr i got merged before this was the april fools green security

PR #35829, opened 2025-03-14
make paradox clone receive the original's objectives

slarticodefast:
## About the PR
Paradox clones now receive all the objectives the original had in their character menu.
This PR will need a lot of testing with all different objective types.

## Why / Balance
This should make things less confusing for players and admins regarding server rules.
For example a clone of a nukie is now allowed to start the nuke.
Also a fun extra challenge in the rare case this happens.

## Technical details
Adds a function that copies objectives from one mind to another.
This only copies the EntityUid of that objective entity and does not create a full copy of the entity itself (which would be very hard to do with our ECS). This works without any problems because objective entities are never modfied or deleted.

At the moment the corresponding mind roles for the objectives are not copied. This should be fine for now though since paradox clones already are a Solo Antagonist, so it should be unproblematic rule-wise.

In the case of being a clone of a clone you don't get a copy of their paradox objectives as these have been blacklisted.
In a future PR I will change the paradox clone kill objective so you have to kill *all* your copies rather than only the specific one you were spawned from.

I also removed a redundant TryIndex in the game rule as that prototype was just being cast back to a ProtoId anyways, which was then checked a second time.

Adds the missing SpaceNinjaComponent to the cloning whitelist so that the ninja can use their equipment.

## Media
grafik

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- add: Paradox clones now receive any objective the original had.

### DIFF
--- a/Content.Server/GameTicking/Rules/Components/ParadoxCloneRuleComponent.cs
+++ b/Content.Server/GameTicking/Rules/Components/ParadoxCloneRuleComponent.cs
@@ -1,4 +1,5 @@
 using Content.Shared.Cloning;
+using Content.Shared.Whitelist;
 using Robust.Shared.Prototypes;
 
 namespace Content.Server.GameTicking.Rules.Components;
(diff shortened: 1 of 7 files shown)


ScarKy0:
If you get the same entity objective, do you complete them as your clone does? (for example, if they steal smth, does it count as stolen for you?)

slarticodefast:
> If you get the same entity objective, do you complete them as your clone does? (for example, if they steal smth, does it count as stolen for you?)

These objectives are evaluated on a per-mind basis, so it works correctly only when *you* stole the item in the end.
Some objectives have counters in their components (for example ninja door hacking), but I think it makes sense to share them in that case. We will need to test every single one to see if they need adjustments.

TeenSarlacc:
> In a future PR I will change the paradox clone kill objective so you have to kill all your copies rather than only the specific one you were spawned from.

can't wait for 5 captains but they're all the same person ♥️

walksanatora:
>be me paradox clone
>copies a traitor
>DAGD
>but... I need to escape alive?

beck-thompson (approved):
Looks good :+1:!

PR #35834, opened 2025-03-14
Removed the heat damage from disablers and disabler SMGs

GaussiArson:
## About the PR

Removed the heat damage from Disablers and Disabler SMGs.

## Why / Balance

Disablers dealing damage feels strange for a nonlethal weapon (especially when NT is clearly capable, look at the baton), especially if you play one of the two races (moth and diona) who are disproportionately affected by the heat.

## Technical details

Heat damage on disabler bolt, disabler bolt practice, and disabler bolt smg set to 0 in projectiles.yml.

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -16,7 +16,7 @@
     tags:
       - HideContextMenu
   - type: AnimationPlayer
-  
+
 - type: entity
   parent: MuzzleFlashEffect
   id: MuzzleFlashEffectOmnilaser
@@ -30,7 +30,7 @@
       map: ["enum.EffectLayers.Unshaded"]
       sprite: Objects/Weapons/Guns/Projectiles/projectiles_tg.rsi
       state: omnilaser_flash
-  
+
 - type: entity
   parent: MuzzleFlashEffect
   id: MuzzleFlashEffectHeavyLaser
@@ -262,7 +262,7 @@
     impactEffect: BulletImpactEffectDisabler
     damage:
       types:
-        Heat: 5
+        Heat: 0
     soundHit:
       collection: WeakHit
     forceSound: true
@@ -302,7 +302,7 @@
     impactEffect: BulletImpactEffectDisabler
     damage:
       types:
-        Heat: 1
+        Heat: 0
     soundHit:
       collection: WeakHit
     forceSound: true
@@ -972,7 +972,7 @@
     impactEffect: BulletImpactEffectDisabler
     damage:
       types:
-        Heat: 2
+        Heat: 0
     soundHit:
       collection: WeakHit
     forceSound: true


GaussiArson:
(i'm sorry i couldn't provide a video but my pc is legitimately too ass to record at a passable framerate)

ArtisticRoomba:
Helps with the weird and annoying AI players who insist that disablers are crew harm because they do heat damage.

Errant-4:
I's called non-lethal, not "harmless". It still fits the description.

A better question is, what is the design purpose of the heat damage in the first place? At a glance, I can only guess that the reason is to discourage sec to shoot people with disablers when not strictly necessary, due to it causing undue harm?

ArtisticRoomba:
Absolutely no clue, and a bit strange when recent sec buffs increased disabler ammo, promoting their usage.

My opinion is that it's just something from SS13 that was inherited.

I don't really see the problem with security excessively using disablers as the baton is usually a very attractive option in the situations where it's applicable, compared to the disabler.

SlimmSlamm:
As Errant said it's a non lethal not a "harmless" option. Moth and Diona have the heat weakness for a reason. It's a downside of the race. We shouldn't ever be balancing the game around removing as much of the cons as possible to races. If you don't want to take increased heat damage then you should play something else.

Aisu9:
praise be, disablers are no longer moff killer!

Aisu9:
> As Errant said it's a non lethal not a "harmless" option. Moth and Diona have the heat weakness for a reason. It's a downside of the race. We shouldn't ever be balancing the game around removing as much of the cons as possible to races. If you don't want to take increased heat damage then you should play something else.

there is critical differences between non-lethal and less-than-lethal, less than lethals still deal some physical harm and both stun batons and disablers are both described as non-lethal

SlimmSlamm:
> > As Errant said it's a non lethal not a "harmless" option. Moth and Diona have the heat weakness for a reason. It's a downside of the race. We shouldn't ever be balancing the game around removing as much of the cons as possible to races. If you don't want to take increased heat damage then you should play something else.
> 
> there is critical differences between non-lethal and less-than-lethal, less than lethals still deal some physical harm and both stun batons and disablers are both described as non-lethal

I don't think you know the difference between lethal and non lethal.

Aisu9:
> I don't think you know the difference between lethal and non lethal.

i am very aware of the difference as i observed many many law forces using non-lethal on civilian causing grievous harm(oh look, something every borgs and AI must uphold at all time), i'm taking the video-game approach not the Real Life approach

Errant-4 (approved):
Okay, having discussed it a bit, I approve this PR for the following reasons:

- There does not seem to have been any specific intent behind it in the first place. As far as I could tell, the 5 heat damage was added to the first disabler shot prototype _before it even had stamina damage_, it was a thing from taser shots? And then the other disabler projectile variants copied the idea from there
- As mentioned, it's a mild issue for silicon, and those silicon can become an issue for sec
- If the idea is that the damage should make sec pause and consider if they should be using the stun baton instead, _why would this be a good thing for the game_? And the ~~disabler~~ stun baton is already a superior choice in the scenarios where it would be possible to choose it, so there is not really a reason to encourage its use
- In a typical arrest, the target will take 4 shots to bring down, which results in 20 heat damage. This will (slowly) auto-heal anyway, doing basically nothing in the long run. A lot of people probably aren't even aware it deals heat damage when they get shot by a disabler, unless...
- species that take extra heat damage will go over 20, so they will need some medical attention afterwards. One might argue that this would "bring out the negative of the species" and so add to their flavor. Honestly, I'm not really seeing this, but, again this is the only reason to have the heat damage... unless we count word-fencing about what fits into the definition of non-lethal

keronshb:
> Okay, having discussed it a bit, I approve this PR for the following reasons:
> 
>     * There does not seem to have been any specific intent behind it in the first place. As far as I could tell, the 5 heat damage was added to the first disabler shot prototype _before it even had stamina damage_, it was a thing from taser shots? And then the other disabler projectile variants copied the idea from there
> 
>     * As mentioned, it's a mild issue for silicon, and those silicon can become an issue for sec
> 
>     * If the idea is that the damage should make sec pause and consider if they should be using the stun baton instead, _why would this be a good thing for the game_? And the disabler is already a superior choice in the scenarios where it would be possible to choose it.
> 
>     * In a typical arrest, the target will take 4 shots to bring down, which results in 20 heat damage. This will (slowly) auto-heal anyway, doing basically nothing in the long run. A lot of people probably aren't even aware it deals heat damage when they get shot by a disabler, unless...
> 
>     * species that take extra heat damage will go over 20, so they will need some medical attention afterwards. One might argue that this would "bring out the negative of the species" and so add to their flavor. Honestly, I'm not really seeing this, but, again this is the only reason to have the heat damage... unless we count word-fencing about what fits into the definition of non-lethal

I think one of the reasons was to prevent pacifists from using it.

Errant-4:
> > Okay, having discussed it a bit, I approve this PR for the following reasons:
> > ```
> > * There does not seem to have been any specific intent behind it in the first place. As far as I could tell, the 5 heat damage was added to the first disabler shot prototype _before it even had stamina damage_, it was a thing from taser shots? And then the other disabler projectile variants copied the idea from there
> > 
> > * As mentioned, it's a mild issue for silicon, and those silicon can become an issue for sec
> > 
> > * If the idea is that the damage should make sec pause and consider if they should be using the stun baton instead, _why would this be a good thing for the game_? And the disabler is already a superior choice in the scenarios where it would be possible to choose it.
> > 
> > * In a typical arrest, the target will take 4 shots to bring down, which results in 20 heat damage. This will (slowly) auto-heal anyway, doing basically nothing in the long run. A lot of people probably aren't even aware it deals heat damage when they get shot by a disabler, unless...
> > 
> > * species that take extra heat damage will go over 20, so they will need some medical attention afterwards. One might argue that this would "bring out the negative of the species" and so add to their flavor. Honestly, I'm not really seeing this, but, again this is the only reason to have the heat damage... unless we count word-fencing about what fits into the definition of non-lethal
> > ```
> 
> I think one of the reasons was to prevent pacifists from using it.

I just tried and it seems they still can't fire it, even with 0 damage

Aisu9:
> > > Okay, having discussed it a bit, I approve this PR for the following reasons:
> > > ```
> > > * There does not seem to have been any specific intent behind it in the first place. As far as I could tell, the 5 heat damage was added to the first disabler shot prototype _before it even had stamina damage_, it was a thing from taser shots? And then the other disabler projectile variants copied the idea from there
> > > 
> > > * As mentioned, it's a mild issue for silicon, and those silicon can become an issue for sec
> > > 
> > > * If the idea is that the damage should make sec pause and consider if they should be using the stun baton instead, _why would this be a good thing for the game_? And the disabler is already a superior choice in the scenarios where it would be possible to choose it.
> > > 
> > > * In a typical arrest, the target will take 4 shots to bring down, which results in 20 heat damage. This will (slowly) auto-heal anyway, doing basically nothing in the long run. A lot of people probably aren't even aware it deals heat damage when they get shot by a disabler, unless...
> > > 
> > > * species that take extra heat damage will go over 20, so they will need some medical attention afterwards. One might argue that this would "bring out the negative of the species" and so add to their flavor. Honestly, I'm not really seeing this, but, again this is the only reason to have the heat damage... unless we count word-fencing about what fits into the definition of non-lethal
> > > ```
> > 
> > 
> > I think one of the reasons was to prevent pacifists from using it.
> 
> I just tried and it seems they still can't fire it, even with 0 damage

it would be fair to assume that pacifist check for any types of damage, even stamina(change the trait would be out of scope of this pr sadly)

Errant-4:
It's every OnShootAttempt. They can't even fire a cap gun

Nukesthestation:
I'm just going put my two cents here and say: I think that the stun baton (and maybe the disabler) Should do at least a tiny amount of shock damage(or just damage in general idk) since (to me at least) it just seems like a bad tazer on a stick(or in the disablers case shoots a form of energy that makes the target tired) hence (I assume) it uses some form of electricity or shock to "stun" the target once they are hit with it 

Also: while looking into this I noticed this Tibet from the disablers description that might need changing: "This one has been undertuned for cadets making it mostly harmless."

Pixeltheaertist:
I feel like some sort of incentive to not always use disablers would be good in the future, since otherwise they're just kinda a better baton with enough range to stop someone from running. But- that's out of scope for this, i think the heat damage is kinda silly in the first place, asit messes with AI players. NO LONGER LESS-THAN-LETHAL!

superjj18:
So are disablers just sci fi sleepy guns?

There should be evidence of conflict if your getting hit by sec weapons

Errant-4:
> So are disablers just sci fi sleepy guns?
> 
> There should be evidence of conflict if your getting hit by sec weapons

They are whatever we decide they are. This is a setting with FTL drives and wizards that can swap minds around. It is perfectly within the realm of possibilities that a gun COULD fire an energy field that harmlessly stuns someone, if that's better for the game than if it also caused burns

Pixeltheaertist:
> > So are disablers just sci fi sleepy guns?
> 
> Sci fi taser that makes your neurons and nerves fire all at once

In that case we could make it do 10 cellular with each shot, surely a really good idea we should totally do that fr

Pumkin69:
This is a good change but it sadly removes my energy shotgun name "the moff roaster" and that makes me upset :(.

GaussiArson:
> This is a good change but it sadly removes my energy shotgun name "the moff roaster" and that makes me upset :(.

there's still the funny red mode for that :D

beck-thompson (approved):

PR #35836, opened 2025-03-14
Add DamagedSiliconAccent to PBs and MMIs

Quantum-cross:
## About the PR

@drteaspoon420 reminded me that I should consider PBs and MMIs with my previous PR #35318.

I didn't see it as a problem before, but now I realize that they have access to binary chat and it could erase some of the interesting gameplay considerations with #35318.

## Why / Balance

This makes it so that PBs and MMIs (henceforth both referred to as PBs) can take damage. Their speech will be corrupted with maximum corruption at 100 total damage (in contrast to 300 damage for the chassis, the gib amount).

This makes it so that PBs cannot communicate clearly over binary chat if they are in danger.

PBs are also given a passive healing of 8 heat, shock, and blunt per second. This gives the effect of PBs acutely glitching out during damaging events, but recover over the course of about a minute (depending on how the damage types are spread out).

Additionally, the damage of the borg chassis is transferred to the brain on chassis destruction.

## Technical details

This is almost 100% done in yaml, but the behavior to transfer the damage to the brain had to be added. Hopefully the behavior is generic enough to be useful in other contexts.

It should be self explanatory

## Media

image
image

_**A FEW MOMENTS LATER**_

image
image

## Requirements

## Breaking changes

None anticipated

**Changelog**

:cl:
- add: Add self-healing speech corruption to damaged PBs and MMIs

### DIFF
--- a/Content.Server/Destructible/Thresholds/Behaviors/DamageContainerContentsBehavior.cs
+++ b/Content.Server/Destructible/Thresholds/Behaviors/DamageContainerContentsBehavior.cs
@@ -1,0 +1,57 @@
+using Content.Shared.Damage;
+using Robust.Server.Containers;
+using Robust.Shared.Containers;
+
+namespace Content.Server.Destructible.Thresholds.Behaviors
+{
+    /// <summary>
+    ///     Damage all items in specified containers
+    /// </summary>
+    [DataDefinition]
+    public sealed partial class DamageContainerContentsBehavior : IThresholdBehavior
+    {
+        /// <summary>
+        /// List of containers to apply damage to contents
+        /// </summary>
+        [DataField("containers")]
+        public List<string> Containers = new();
+
+        /// <summary>
+        /// Damage specifier to apply to container contents.
+        /// If null, apply the current damage of the owner entity to all the contents
+        /// </summary>
+        [DataField]
+        public DamageSpecifier? Damage = null;
+
+        [DataField]
+        public bool IgnoreResistances = false;
+
+        public void Execute(EntityUid owner, DestructibleSystem system, EntityUid? cause = null)
+        {
+            if (!system.EntityManager.TryGetComponent<ContainerManagerComponent>(owner, out var containerManager))
+                return;
+
+            var containerSys = system.EntityManager.System<ContainerSystem>();
+            var damageSys = system.EntityManager.System<DamageableSystem>();
+
+            foreach (var containerId in Containers)
+            {
+                if (!containerSys.TryGetContainer(owner, containerId, out var container, containerManager))
+                    continue;
+
+                if (Damage == null)
+                {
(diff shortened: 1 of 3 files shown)


Mixelz:
Wouldn't allowing PBs to take damage imply that they can be destroyed the same as other objects? At the very least I feel it could easily be misconstrued as such by players who just see it take damage the same as anything else.

Quantum-cross:
> Wouldn't allowing PBs to take damage imply that they can be destroyed the same as other objects? At the very least I feel it could easily be misconstrued as such by players who just see it take damage the same as anything else.

I'd agree with that. I wonder if there's another way to let the player know that it's actually invincible, or maybe a different hit effect instead of the standard red flash?

Quantum-cross:
Lame, the color of the flash on hit is hard coded to red in many places.

Mixelz:
> Lame, the color of the flash on hit is hard coded to red in many places.

Sounds like uncoupling that hardcode behavior would make for a good technical PR, if not a strenous one.

walksanatora:
yeah since blue indicates stam dammage.

Quantum-cross:
Another option I was considering was showing a little popup text to the damager that says something like, "it seems like it won't break!" Or something if you damage it more than 300 health

Quantum-cross:
Also maybe just putting in the shift click examine description...

Quantum-cross:
> > Lame, the color of the flash on hit is hard coded to red in many places.
> 
> Sounds like uncoupling that hardcode behavior would make for a good technical PR, if not a strenous one. Regardless, I would agree that a different color flash should be sufficient in telling a player that it doesnt take damage in the same way as other items.

This solution is supported by #35900

Content.Server/Destructible/Thresholds/Behaviors/DamageContainerContentsBehavior.cs
```diff
@@ -0,0 +1,57 @@
+using Content.Shared.Damage;
+using Robust.Server.Containers;
+using Robust.Shared.Containers;
+
+namespace Content.Server.Destructible.Thresholds.Behaviors
```
EmoGarbage404, line 5: do a file scoped namespace.

Content.Server/Destructible/Thresholds/Behaviors/DamageContainerContentsBehavior.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <summary>
+    ///     Damage all items in specified containers
+    /// </summary>
+    [DataDefinition]
+    public sealed partial class DamageContainerContentsBehavior : IThresholdBehavior
+    {
+        /// <summary>
+        /// List of containers to apply damage to contents
+        /// </summary>
+        [DataField("containers")]
```
EmoGarbage404, line 16:
```suggestion
        [DataField]
```
not necessary

EmoGarbage404 (changes requested):
cool pr

Issue #35831, opened 2025-03-14
Paradox Clone ToDo list


slarticodefast:
List of additional planned features for paradox clones.
In case someone wants to help out with one of them, let me know here:

- [ ] Better spawn location. Right now it picks a random unspaced tile on the station. They get a toolbox at the moment in case they are stuck somewhere, but ideally we just improve the spawning location logic to choose a random tile in maints (maybe by floor tile type?). This should include a helper function and a component for antag selection similar to `AntagRandomSpawnComponent`.
- [ ] Fix the preview for ghostrole antag spawns. At the moment when you select `follow` on the ghostrole you are shown a location, but the actual spawn location will be somewhere completely different. This is a general problem with `AntagRandomSpawnComponent` and dragons used to have the same bug before they were changed to spawn in space.
- [ ] Copy mind roles and job roles from the original to the clone. This should not be horribly hardcoded, but have a general function in SharedMindSystem. This might be complicated since roles are entities.
- [ ] Give traitor clones an uplink and the same amount of TC as the original. This will be hard to do without horribly hardcoding everything, especially in the case of uplink implants. Probably needs some refactoring first.
- [ ] Copy Objectives from the original https://github.com/space-wizards/space-station-14/pulls
- [ ] Recursive item copying. Right now you only get an empty backpack. I'm currently working on this one.
- [ ] Copy implants from the original. This is important now that mindshielding is an implant. Should be a bool in the cloning settings prototype.
- [ ] New multi-kill objective type. At the moment you only have to kill the player you were directly copied from. But in the case of a clone of a clone they should all have to kill all copies of each other.
- [ ] Admin antag control verb to make it possible to generate a ghost role of a specific player, similar to the one we had for terminators https://github.com/space-wizards/space-station-14/pull/26978/files. This should simply start the gamerule and override the random target selection.

VasilisThePikachu:
Mentioned in the maint channel & maint meeting;

They should probably be more common, apparently no admin has at least seen it happen naturally

slarticodefast:
> They should probably be more common, apparently no admin has at least seen it happen naturally

Turns out I just forgot to add the event to the event table :csgrad:
https://github.com/space-wizards/space-station-14/pull/35858

IProduceWidgets:
[...]

PR #35838, opened 2025-03-14
Paradox clones get all storage items the original has.

slarticodefast:
## About the PR
The previous implementation only gave you an empty backpack. Now all storage items will spawn with the same contents on the clone. This also includes slime storage.

This PR will need extensive testing for items that should not be copied or may cause errors and should thus be blacklisted.

## Why / Balance
Command clones were easily identified by their missing unique items.

## Technical details
Added new functions to recursively copy an item and its storage.
This is not perfect, as it spawns a new item from the originals protoId, so all changes made to the item after spawning are lost. 
This only supports StorageComponent (meaning grid storage) and not arbitrary containers or itemslots, as we got too many duplicated components all adding their own spawning logic on map init. For example `StorgeFillComponent`, `ContainerFillComponent`, `EntityTableContainerFillComponent`, `ItemSlotsComponent` and so on.
But this is good enough for most cases.

How to test:
- Equip items, best in nested containers
- Use the `Random clone spawner` from the spawn menu to copy yourself
- compare inventories and check for errors
- Additionally test cloning a slime character being cloned with their internal storage.

## Media
grafik
grafik

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- tweak: Paradox clones now recieve a copy of (almost) all the inventory items their original has.

### DIFF
--- a/Content.Server/Cloning/CloningSystem.cs
+++ b/Content.Server/Cloning/CloningSystem.cs
@@ -7,7 +7,11 @@ using Content.Shared.Humanoid;
 using Content.Shared.Inventory;
 using Content.Shared.NameModifier.Components;
 using Content.Shared.StatusEffect;
+using Content.Shared.Stacks;
+using Content.Shared.Storage;
+using Content.Shared.Storage.EntitySystems;
 using Content.Shared.Whitelist;
+using Robust.Shared.Containers;
 using Robust.Shared.Map;
 using Robust.Shared.Prototypes;
 using System.Diagnostics.CodeAnalysis;
@@ -28,6 +32,9 @@ public sealed class CloningSystem : EntitySystem
     [Dependency] private readonly IPrototypeManager _prototype = default!;
     [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
     [Dependency] private readonly ISharedAdminLogManager _adminLogger = default!;
+    [Dependency] private readonly SharedContainerSystem _container = default!;
+    [Dependency] private readonly SharedStorageSystem _storage = default!;
+    [Dependency] private readonly SharedStackSystem _stack = default!;
 
     /// <summary>
     ///     Spawns a clone of the given humanoid mob at the specified location or in nullspace.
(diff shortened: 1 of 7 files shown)


SlimmSlamm:
what if the original has a storage planter and stuff inside it?

slarticodefast:
> what if the original has a storage planter and stuff inside it?

implants are not cloned yet, but it's on my todo list
https://github.com/space-wizards/space-station-14/issues/35831

M1tht1c:
Will this also copy objective items?

Content.Server/Cloning/CloningSystem.cs
```diff
@@ -100,24 +111,89 @@ public bool TryCloning(EntityUid original, MapCoordinates? coords, ProtoId<Cloni
...
+        // if the original is a stack, adjust the count of the copy
+        if (TryComp<StackComponent>(original, out var originalStack) && TryComp<StackComponent>(spawned, out var spawnedStack))
+            _stack.SetCount(spawned, originalStack.Count, spawnedStack);
+
+        // if the original has items inside its storage, copy those as well
+        if (TryComp<StorageComponent>(original, out var originalStorage) && TryComp<StorageComponent>(spawned, out var spawnedStorage))
+        {
+            // remove all items that spawned with the entity inside its storage
+            // this ignores other containers, but this should be good enough for our purposes
+            _container.CleanContainer(spawnedStorage.Container);
```
beck-thompson, line 161: Just call `CopyStorage`!

beck-thompson (changes requested):
The starting items get duplicated! It looks like the starting stuff only gets injected AFTER the cloning happens so that's why it doesn't get cleared.

image

Content.Server/Cloning/CloningSystem.cs
slarticodefast: for better readability I would prefer to keep it purely recursive instead of having two functions calling each other

slarticodefast:
I removed the survival box the paradox clone gets.
Vox still get a nitrogen tank though.

beck-thompson (approved):
:+1: Looks good!

PR #35844, opened 2025-03-15
Add support for ambient on toggled

CyberTropic:
## About the PR

Adds ambientOnToggled component that allows toggling of AmbientSound with ItemToggle.

## Why / Balance

This particular use case of a prop with a toggleable cosmetic ambient sound came from downstream, but the ability to turn ambient sound on/off for an entity with ItemToggle can be pretty useful in general.

## Technical details

Pretty simple change, modeled on ambientOnPowered, but tying it to ItemToggledEvent instead.

## Media

https://github.com/user-attachments/assets/f01b9b95-7fcc-4266-b74e-96f8546a9eca

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Audio/AmbientOnToggledComponent.cs
+++ b/Content.Server/Audio/AmbientOnToggledComponent.cs
@@ -1,0 +1,12 @@
+using Content.Shared.Audio;
+
+namespace Content.Server.Audio
+{
+    /// <summary>
+    /// Toggles <see cref="AmbientSoundComponent"/> on when toggled and off when not toggled through ItemToggle.
+    /// </summary>
+    [RegisterComponent]
+    public sealed partial class AmbientOnToggledComponent : Component
+    {
+    }
+}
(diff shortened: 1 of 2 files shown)


Content.Server/Audio/AmbientOnToggledComponent.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace Content.Server.Audio
+{
+    /// <summary>
+    /// Toggles <see cref="AmbientSoundComponent"/> on when toggled and off when not toggled through ItemToggle.
+    /// </summary>
+    [RegisterComponent]
+    public sealed partial class AmbientOnToggledComponent : Component
+    {
+    }
+}
```
metalgearsloth, line 12: Shared please + file-scoped namespace.

Content.Server/Audio/AmbientSoundSystem.cs
```diff
@@ -24,4 +27,9 @@ private void HandlePowerChange(EntityUid uid, AmbientOnPoweredComponent componen
     {
         SetAmbience(uid, args.Powered);
     }
+
+    private void HandleItemToggled(EntityUid uid, AmbientOnToggledComponent component, ref ItemToggledEvent args)
+    {
+        SetAmbience(uid, args.Activated);
+    }
 }
```
metalgearsloth, line 35: Same here

metalgearsloth (changes requested):

slarticodefast:
Isn't this the same as the existing `ItemToggleActiveSound`?

CyberTropic:
> Isn't this the same as the existing `ItemToggleActiveSound`?

It is! Thanks, I must have missed it, closing PR. It seems like ambientOnPowered should be moved to shared then

PR #35847, opened 2025-03-15
clear buyer conditions in StoreDiscountAndRefundTest

MilonPL:
## About the PR

the StoreSystem checks for conditions and we're just lucky that this seed has items without any conditions or otherwise it would fail
but it could still fail if the amount of listing prototypes changes

## Requirements

## Breaking changes

**Changelog**

no fun

### DIFF
--- a/Content.IntegrationTests/Tests/StoreTests.cs
+++ b/Content.IntegrationTests/Tests/StoreTests.cs
@@ -102,6 +102,12 @@ public sealed class StoreTests
                 + $"flag as 'true'. This marks the fact that cost modifier of discount is not applied properly!"
             );
 
+            // Set conditions to null so that it doesnt fail
+            foreach (var discountedItem in discountedListingItems)
+            {
+                discountedItem.Conditions = null;
+            }
+
             // Refund action requests re-generation of listing items so we will be re-acquiring items from component a lot of times.
             var itemIds = discountedListingItems.Select(x => x.ID);
             foreach (var itemId in itemIds)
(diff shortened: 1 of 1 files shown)


Content.IntegrationTests/Tests/StoreTests.cs
```diff
@@ -102,6 +102,12 @@ await server.WaitAssertion(() =>
                 + $"flag as 'true'. This marks the fact that cost modifier of discount is not applied properly!"
             );
 
+            // Set conditions to null so that it doesnt fail
```
FairlySadPanda, line 105:
```suggestion
            // The storeComponent returns discounted items with conditions randomly, so we remove these to sanitize the data and prevent the test failing erratically.`
```

Content.IntegrationTests/Tests/StoreTests.cs
```diff
@@ -140,6 +146,9 @@ await server.WaitAssertion(() =>
                     // get refreshed item after refund re-generated items
                     discountedListingItem = storeComponent.FullListingsCatalog.First(x => x.ID == itemId);
 
+                    // Check if conditions are still null after refund
```
FairlySadPanda, line 149:
```suggestion
                    // storeComponent can give a discounted item a condition at random, so we need it removed to sanitize the data.
```

FairlySadPanda (commented):
A comment, if needed, should explain why a step has been taken clearly. In this case the comments are justified but should flag that this is data sanitation, not some specific condition of the test.

PR #35855, opened 2025-03-15
Fix salvage sites ejecting borg brains when they despawn

### LINKED ISSUE #35116: Salvage sites despawn causes borg to pop out of their shells
## Description
when borgs are on a despawning wrech/asteroid(tested with wrech) borg's positronic brain will pop out of their vessel, this is experienced by multiple players. this was experienced on round 77660 to 77670, dont remember.

**Reproduction**
be happy borg, salvage,engineering, any.
go help salv as a happy borg.
stay too long, because happy borg.
get screwed by despawn and become...
Sad borg :(

Tayrtahn:
## About the PR

Cyborgs can now safely work at salvage magnet sites (wrecks, asteroids, etc.) without fear of having their brains removed when the site despawns.

## Why / Balance

Fixes #35116

## Technical details

The issue occurs because `SalvageSystem.EndMagnet` takes each entity on the grid that has a `MobStateComponent` and reparents it to the map. Notably, both the borg body entity and the positronic brain have a `MobStateComponent`, so they are both _separately_ reparented to the map. This doesn't seem to happen all the time, which I think is due to update order - if the body gets moved first, then the brain is taken with it and is no longer in the entity query so it doesn't get moved separately.

In experimenting with this, I noticed that it affects any mobs that are contained within other mobs - for example, a mouse in a player's backpack will get removed and placed beside the player too.

To fix both of these cases, the logic in `EndMagnet` was expanded so that before reparenting a mob, its parents and their parents (and all the way up the hierarchy) are checked for their own `MobStateComponent`s - if one is found, mob isn't reparented (since the parent will be).

## Requirements

**Changelog**

:cl:
- fix: Cyborgs no longer have their brains removed if the salvage site they are on despawns.

### DIFF
--- a/Content.Server/Salvage/SalvageSystem.Magnet.cs
+++ b/Content.Server/Salvage/SalvageSystem.Magnet.cs
@@ -147,12 +147,28 @@ public sealed partial class SalvageSystem
             var mobQuery = AllEntityQuery<MobStateComponent, TransformComponent>();
             _detachEnts.Clear();
 
+            var mobStateQuery = GetEntityQuery<MobStateComponent>();
             while (mobQuery.MoveNext(out var mobUid, out _, out var xform))
             {
                 if (xform.GridUid == null || !data.Comp.ActiveEntities.Contains(xform.GridUid.Value) || xform.MapUid == null)
                     continue;
 
                 if (_salvMobQuery.HasComp(mobUid))
+                    continue;
+
+                bool CheckParents(EntityUid uid)
+                {
+                    do
+                    {
+                        uid = _transform.GetParentUid(uid);
+                        if (mobStateQuery.HasComp(uid))
+                            return true;
+                    }
+                    while (uid != xform.GridUid);
+                    return false;
+                }
+
+                if (CheckParents(mobUid))
                     continue;
 
                 // Can't parent directly to map as it runs grid traversal.
--- a/Content.Server/Salvage/SalvageSystem.Magnet.cs
+++ b/Content.Server/Salvage/SalvageSystem.Magnet.cs
@@ -164,7 +164,7 @@ public sealed partial class SalvageSystem
                         if (mobStateQuery.HasComp(uid))
                             return true;
                     }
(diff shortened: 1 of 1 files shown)


Content.Server/Salvage/SalvageSystem.Magnet.cs
```diff
@@ -147,6 +147,7 @@ private void EndMagnet(Entity<SalvageMagnetDataComponent> data)
             var mobQuery = AllEntityQuery<MobStateComponent, TransformComponent>();
             _detachEnts.Clear();
 
+            var mobStateQuery = GetEntityQuery<MobStateComponent>();
```
slarticodefast, line 150: I think this can be moved to `InitializeMagnet`

Content.Server/Salvage/SalvageSystem.Magnet.cs
```diff
@@ -155,6 +156,21 @@ private void EndMagnet(Entity<SalvageMagnetDataComponent> data)
...
 
+                bool CheckParents(EntityUid uid)
+                {
+                    do
+                    {
+                        uid = _transform.GetParentUid(uid);
+                        if (mobStateQuery.HasComp(uid))
+                            return true;
+                    }
+                    while (uid != xform.GridUid);
```
slarticodefast, line 167: very scary looking while condition
but I cant think of any reason why this should not always reach this uid

Content.Server/Salvage/SalvageSystem.Magnet.cs
Tayrtahn: Oh, good call.

Content.Server/Salvage/SalvageSystem.Magnet.cs
Tayrtahn: I can't either, but I'll add a check for EntityUid.Invalid to be extra safe.

slarticodefast (approved):

slarticodefast:
Blessed bugfix

PR #35861, opened 2025-03-16
Add viewer parameter to Identity.Entity

Tayrtahn:
## About the PR

Adds an optional `viewer` parameter to the `Identity.Entity` method which uses `CanSeeThroughIdentity` to let ghosts see actual identities.

## Why / Balance

We already have this functionality in `Identity.Name`, but it's often useful to get an entity instead of just a string representation of the name. This is handy because many Fluent functions need an entity and not just a string - this makes it possible to use those functions while still working with the Identity system properly.

This makes it easy to do stuff like `{CAPITALIZE(THE($user))} does something` and have it able to resolve to `Urist McHands does something` or `The young man does something` as appropriate for the viewer's perspective.

## Technical details

If a `viewer` entity is passed in, it gets passed through to `CanSeeThroughIdentity` (which just looks for a `GhostComponent`) - if true, the method returns the actual uid of the target; if false, it returns the identity entity (if one exists).

## Requirements

### DIFF
--- a/Content.Shared/IdentityManagement/Identity.cs
+++ b/Content.Shared/IdentityManagement/Identity.cs
@@ -49,6 +49,9 @@ public static class Identity
     ///     This is an extension method because of its simplicity, and if it was any harder to call it might not
     ///     be used enough for loc.
     /// </summary>
+    /// <param name="viewer">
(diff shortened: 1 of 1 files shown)


beck-thompson (approved):
makes sense to me!

PR #35862, opened 2025-03-16
Fix strip menu revealing true identity

### LINKED ISSUE #35677: strip menu still reveals true identity
## Description

if you try to place an item in the wrong slot it uses the entity name instead of identity

Image

Tayrtahn:
## About the PR

The strip menu no longer reveals the target's true identity in related popup messages.
~Additionally, ghosts now see the target's true identity in the strip menu and related popups (they already see it on examine, so this isn't a balance change).~ Removed. Can be added as a separate PR if desired.

~Needs https://github.com/space-wizards/space-station-14/pull/35861 merged first.~ merged, but no longer needed

## Why / Balance

Fixes #35677.

## Technical details

Calls `Identity.Entity` on the target value before passing it to localization in a bunch of places. Fancies up the Loc strings to use `THE` and `CAPITALIZE` when appropriate so that the popups are properly formatted. ~Uses the `viewer` parameter added by https://github.com/space-wizards/space-station-14/pull/35861 to let ghosts see through disguises.~

## Media

The young man cannot equip that there!

## Requirements

**Changelog**

Nah.

### DIFF
--- a/Content.Client/Inventory/StrippableBoundUserInterface.cs
+++ b/Content.Client/Inventory/StrippableBoundUserInterface.cs
@@ -66,7 +66,7 @@ namespace Content.Client.Inventory
 
             _strippingMenu = this.CreateWindowCenteredLeft<StrippingMenu>();
             _strippingMenu.OnDirty += UpdateMenu;
-            _strippingMenu.Title = Loc.GetString("strippable-bound-user-interface-stripping-menu-title", ("ownerName", Identity.Name(Owner, EntMan, PlayerManager.LocalEntity)));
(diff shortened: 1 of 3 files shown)


slarticodefast:
Hmm, this would introduce a new inconsistency. Usually popups always show the users identity, even for ghosts.
With this the popups for stripping interactions would be the only exception.

Tayrtahn:
Hm, well I can separate that change out and make ghosts still see identity, simple enough.

slarticodefast (approved):

PR #35863, opened 2025-03-16
Improve sprite fading behaviour

sowelipililimute:
## About the PR
- faded sprites now get clicked through, allowing you to interact with objects behind faded objects w/o needing to use the disambiguation menu
- the mouse pointer is now used to unfade sprites

## Why / Balance
- deltav wants to raise walls up higher for a 3:4 perspective
- sprite fading needs to be more seamless and allow players to look behind walls in their character's FOV but not at their camera rotation more easily to ease friction
- we are NOT maintaining changes to clickable/gameplaystatebase/spritefadesystem downstream
- this would be nice for other forks probably

## Technical details
commit 1:
- have ClickableSystem ignore faded entities, unless it's being consulted by the system that determines which entities are faded
commit 2:
- have mouse position be used to fade sprites, but only if it's not colliding with their fixtures

## Media

https://github.com/user-attachments/assets/dffbaf21-d248-43d6-9eeb-e3ec264dcb21

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: You can now more easily interact with objects behind faded ones, and you can look behind fadeable objects in your FOV by hovering them with your mouse pointer

### DIFF
--- a/Content.Client/Sprite/SpriteFadeSystem.cs
+++ b/Content.Client/Sprite/SpriteFadeSystem.cs
@@ -1,8 +1,14 @@
+using System.Numerics;
 using Content.Client.Gameplay;
 using Content.Shared.Sprite;
 using Robust.Client.GameObjects;
+using Robust.Client.Input;
 using Robust.Client.Player;
 using Robust.Client.State;
+using Robust.Client.UserInterface.CustomControls;
+using Robust.Client.UserInterface;
+using Robust.Shared.Map;
+using Robust.Shared.Physics.Systems;
 using Robust.Shared.Physics;
 
 namespace Content.Client.Sprite;
@@ -17,12 +23,20 @@ public sealed class SpriteFadeSystem : EntitySystem
     [Dependency] private readonly IPlayerManager _playerManager = default!;
     [Dependency] private readonly IStateManager _stateManager = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly IUserInterfaceManager _uiManager = default!;
+    [Dependency] private readonly IInputManager _inputManager = default!;
+    [Dependency] private readonly SharedPhysicsSystem _physics = default!;
 
     private readonly HashSet<FadingSpriteComponent> _comps = new();
 
     private EntityQuery<SpriteComponent> _spriteQuery;
     private EntityQuery<SpriteFadeComponent> _fadeQuery;
     private EntityQuery<FadingSpriteComponent> _fadingQuery;
+
+    /// <summary>
+    ///     Radius of the mouse point for the intersection test
+    /// </summary>
+    private static Vector2 MouseRadius = new Vector2(10f * float.Epsilon, 10f * float.Epsilon);
 
     private const float TargetAlpha = 0.4f;
     private const float ChangeRate = 1f;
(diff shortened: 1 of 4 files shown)


ps3moira:
Wouldn't this create less hiding spots if people can just hover over an item versus going behind to check and see?

mirrorcult:
you can already check and see by just rightclicking

Killerqu00:
> you can already check and see by just rightclicking

yes, but this is huge QoL

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -53,39 +64,73 @@ public override void FrameUpdate(float frameTime)
...
+
+                    if (!sprite.Color.A.Equals(newColor))
+                    {
+                        sprite.Color = sprite.Color.WithAlpha(newColor);
+                    }
                 }
             }
         }
 
+
```
beck-thompson, line 129: extra space!

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -53,39 +64,73 @@ public override void FrameUpdate(float frameTime)
...
+                        sprite.DrawDepth < playerSprite.DrawDepth)
+                    {
+                        continue;
+                    }
+
+                    if (excludeBB)
+                    {
+                        var delta = new Vector2(10f * float.Epsilon, 10f * float.Epsilon);
+                        var test = new Box2Rotated(mapPos.Position - delta, mapPos.Position + delta);
+                        bool collided = false;
```
beck-thompson, line 97: use var!

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -53,39 +64,73 @@ public override void FrameUpdate(float frameTime)
...
+        if (TryComp(player, out TransformComponent? playerXform))
+        {
+            CoordinatesBuffer[length++] = (_transform.GetMapCoordinates(_playerManager.LocalEntity!.Value, xform: playerXform), false);
+        }
 
-                _comps.Add(fading);
-                var newColor = Math.Max(sprite.Color.A - change, TargetAlpha);
+        if (_stateManager.CurrentState is GameplayState state && _spriteQuery.TryGetComponent(player, out var playerSprite))
+        {
+            for (int i = 0; i < length; i++)
```
beck-thompson: use var!

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -53,39 +62,71 @@ public override void FrameUpdate(float frameTime)
         var player = _playerManager.LocalEntity;
         var change = ChangeRate * frameTime;
 
-        if (TryComp(player, out TransformComponent? playerXform) &&
-            _stateManager.CurrentState is GameplayState state &&
-            _spriteQuery.TryGetComponent(player, out var playerSprite))
-        {
-            var mapPos = _transform.GetMapCoordinates(_playerManager.LocalEntity!.Value, xform: playerXform);
+        var pointsOfInterest = new List<(MapCoordinates Point, bool ExcludeBoundingBox)>();
```
beck-thompson, line 65: Could you add a comment here explaining what pointsOfInterest is - basically what you told me in discord! Also make sure to explain exactly what excludeBoundingBox means!

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -53,39 +62,71 @@ public override void FrameUpdate(float frameTime)
...
+                        !_fadeQuery.HasComponent(ent) ||
+                        !_spriteQuery.TryGetComponent(ent, out var sprite) ||
+                        sprite.DrawDepth < playerSprite.DrawDepth)
+                    {
+                        continue;
+                    }
+
+                    if (excludeBB)
+                    {
+                        var delta = new Vector2(10f * float.Epsilon, 10f * float.Epsilon);
```
beck-thompson, line 95: Could you add a comment explaining that this is just basically just simulating the mouse position and the 10s are just magic numbers?

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -53,39 +62,71 @@ public override void FrameUpdate(float frameTime)
...
+                    if (!sprite.Color.A.Equals(newColor))
+                    {
+                        sprite.Color = sprite.Color.WithAlpha(newColor);
+                    }
                 }
             }
         }
 
+
         var query = AllEntityQuery<FadingSpriteComponent>();
```
beck-thompson, line 130: Could you split this entire thing into two functions? E.g the first one adds new fading sprites, the second one updates them (Also add a comments to both that explain that the second part is just for bringing the sprite back to the normal alpha and the original loop is what brings it down!)!

beck-thompson (changes requested):
Looks good just style things!

beck-thompson (approved):
looks good! The only thing that concerns me is stuff like trees:
image

Your mouse doesn't need to get behind it fully for it to fade out, I'm unsure if this is really an issue though because it is annoying if you can't see pas the tree, and it being faded out still looks fine.

TheShuEd:
Ok yea thats look nice

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -46,46 +60,82 @@ private void OnFadingShutdown(EntityUid uid, FadingSpriteComponent component, Co
...
+                            {
+                                collided = true;
+                                break;
+                            }
+                        }
+                        if (collided)
+                        {
+                            break;
+                        }
+                    }
```
metalgearsloth, line 114: This is very very slow.

You should just be using entitylookup or spritetree (probably the latter) instead, as now it's doing 2 very expensive lookups for every single position being checked.

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -46,46 +60,82 @@ private void OnFadingShutdown(EntityUid uid, FadingSpriteComponent component, Co
...
+
+                    if (!_fadingQuery.TryComp(ent, out var fading))
+                    {
+                        fading = AddComp<FadingSpriteComponent>(ent);
+                        fading.OriginalAlpha = sprite.Color.A;
+                    }
```
metalgearsloth, line 120: This is just ensurecomp

sowelipililimute: not my code, this is just moved from preexisting

Content.Client/Sprite/SpriteFadeSystem.cs
sowelipililimute: should the physics system not be optimised for handling the question "what colliding entities are at this area"...? it already has a broadphase tree for minimising how many entities are tested

metalgearsloth: Except you're allocating 2 hashsets for every single point of interest every single frame. It's on the caller to do the right thing.

PR #35865, opened 2025-03-16
Fix gas pressure pump prediction

MilonPL:
## About the PR

ACTUALLY predicts the pressure gas pumps now
cleaned up a bunch of stuff related to it too
If component state changes (like through VV), it gets reflected in the UI (we weren't raising the event before so the client didnt update it)

## Technical details

Added the `true` value to the AutoGenerateComponentState attribute to raise the event we listen on for the client.
Added the `UpdateUi` method to the SharedGasPressurePumpSystem. We need it to actually update the client from Shared, since the Client can't predict component state changes as `AfterAutoHandleStateEvent` is called by Server code *after* we send the new state to the client. So to properly predict UI messages, we need to update it from Shared as the message gets sent. It gets overriden on the client anyways since we dont need to have any logic for the UI running on the server.

In the future when we have other predicted events, those need to update the UI too. Otherwise we will have to wait for new component state from the server after we dirty the entity.

generic is slower?? it has an additional type cast so i just used the base one :trollface: 

## Requirements

## Breaking changes

**Changelog**

no fun


Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressurePumpSystem.cs
```diff
@@ -25,33 +27,33 @@ public override void Initialize()
...
 
         var outputStartingPressure = outlet.Air.Pressure;
 
-        if (outputStartingPressure >= pump.TargetPressure)
+        if (outputStartingPressure >= ent.Comp.TargetPressure)
         {
-            _ambientSoundSystem.SetAmbience(uid, false);
-            return; // No need to pump gas if target has been reached.
+            _ambientSoundSystem.SetAmbience(ent, false);
+            return; // No need to ent.Comp gas if target has been reached.
```
Tayrtahn, line 45: Hello, find and replace.

MilonPL: You got me, I'm lazy.

dvir001:
Hero

PR #35872, opened 2025-03-16
Alternative fix for clientside solution resolves (Udder & Wooly debug asserts)

### LINKED ISSUE #34282: Debug Assert in  UdderSystem

Tayrtahn:
## About the PR

This is an alternative approach to fixing #34282.

## Why / Balance

Fixes #34282

## Technical details

In comparison to #35314, this approach applies to all uses of `ResolveSolution`, not just the specific cases of `UdderSystem` and `WoolySystem`. However, it does require looking up the MetaData component of the solution entity with each call, which could have some performance implications.

The way this works is by adding a check for the `Detached` metadata flag on the solution entity - entities outside of PVS are detached to nullspace on the client, so this detects when that happens. When it does, the conditional resolves to false, and execution will proceed to line 126: `return TryGetSolution(container, name, out entity);`. Because `entity` is passed to `ResolveSolution` as a ref parameter, passing it in to `TryGetSolution` causes the reference to get replaced with whatever `TryGetSolution` finds - which should be `null` in this case. This means that the cached reference on `UdderComponent` or wherever is properly set to null and `ResolveSolution` will return false for this call and any future calls until it can be resolved again (by reentering PVS).

## Requirements

### DIFF
--- a/Content.Shared/Chemistry/EntitySystems/SharedSolutionContainerSystem.cs
+++ b/Content.Shared/Chemistry/EntitySystems/SharedSolutionContainerSystem.cs
@@ -116,7 +116,7 @@ public abstract partial class SharedSolutionContainerSystem : EntitySystem
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public bool ResolveSolution(Entity<SolutionContainerManagerComponent?> container, string? name, [NotNullWhen(true)] ref Entity<SolutionComponent>? entity)
     {
-        if (entity is not null)
+        if (entity is not null && (MetaData(entity.Value.Owner).Flags & MetaDataFlags.Detached) == 0)
         {
             DebugTools.Assert(TryGetSolution(container, name, out var debugEnt)
                               && debugEnt.Value.Owner == entity.Value.Owner);


slarticodefast:
I asked sloth on discord about this fix
grafik
So this approach does not seem to be ideal either.

What I still don't get about the error is why the cow is part of the update loop on the client in the first place. If it is out of PVS range it should not get any system updates in my opinion. Is nullspace not paused? Because the `EntityQueryEnumerator` in the update loop should skip any paused entities.

Tayrtahn:
I'm pretty sure nullspace is not paused, and I think there are some things that use that? Abstract entities like minds and stations, maybe?

It might make sense for the PVS system to pause entities when sending them to nullspace though. Worth looking into. This solution cacheing stuff is still weird on its own - I wonder how much (if at all) it helps performance over just getting the solution when it's needed.

slarticodefast:
If all entities outside of PVS range still run their regular update loops that sounds like a huge performance impact at least.
And I cannot think of any reason why it would be needed.

Errant-4:
Due to the concerns about performance, I move to close this PR and proceed with the other one instead. Anyone opposed?

Tayrtahn:
That's fine by me.

PR #35874, opened 2025-03-16
Make NukeOpsTest only use Nukeops gamerule

### LINKED ISSUE #35868: Wizard antag causes heisentest in TryStopNukeOpsFromConstantlyFailing
## Description
With the new Wizard antag, Nukeops has a chance to spawn a Wizard sub-antag, which spawns its own shuttle under its own RuleGridsComponent.

When Linq attempts to select a Single from all RuleGridsComponents, it will see two components, the one from the NukeOps gamerule, and the one from the Wizard subgamerule. Linq will throw an exception in this instance.

Line 158:
```C#
        var gridsRule = entMan.AllComponents<RuleGridsComponent>().Single().Component;
```

This heisentests because Wizard only rolls like 10% of the time as a subgamerule on nukeops.

**Reproduction**
Run the TryStopNukeOpsFromConstantlyFailing test until failure if you hate yourself, otherwise turn up the % chance in the subrule so it always happens.

Tayrtahn:
## About the PR

Modifies `TryStopNukeOpsFromConstantlyFailing` so that it uses a custom `GamePresetPrototype` set up to only include the actual `Nukeops` gamerule and no others.

## Why / Balance

Fixes #35868.

## Technical details

The test currently uses the Nukeops `GamePresetPrototype`, which is defined as:
```
- type: gamePreset
  id: Nukeops
  alias:
    - nukeops
  name: nukeops-title
  description: nukeops-description
  showInVote: false
  rules:
    - Nukeops
    - SubGamemodesRule
    - BasicStationEventScheduler
    - MeteorSwarmScheduler
    - SpaceTrafficControlEventScheduler
    - BasicRoundstartVariation
```
We don't want those other gamerules getting added - especially `SubGamemodesRule`, which is what occasionally adds the wizard subgamemode which causes the test to fail.
So the test now loads its own preset which just includes the `Nukeops` rule.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/GameRules/NukeOpsTest.cs
+++ b/Content.IntegrationTests/Tests/GameRules/NukeOpsTest.cs
@@ -29,6 +29,18 @@ namespace Content.IntegrationTests.Tests.GameRules;
 [TestFixture]
 public sealed class NukeOpsTest
 {
+    private const string NukeOpsGamePreset = "TestNukeops";
+
+    [TestPrototypes]
+    private const string Prototypes = $@"
+- type: gamePreset
+  id: {NukeOpsGamePreset}
+  name: nukeops-title
+  showInVote: false
+  rules:
+  - Nukeops
+";
+
     /// <summary>
     /// Check that a nuke ops game mode can start without issue. I.e., that the nuke station and such all get loaded.
     /// </summary>
@@ -90,7 +102,7 @@ public sealed class NukeOpsTest
         // Ready up and start nukeops
         ticker.ToggleReadyAll(true);
         Assert.That(ticker.PlayerGameStatuses.Values.All(x => x == PlayerGameStatus.ReadyToPlay));
-        await pair.WaitCommand("forcepreset Nukeops");
+        await pair.WaitCommand($"forcepreset {NukeOpsGamePreset}");
         await pair.RunTicksSync(10);
 
         // Game should have started


ArtisticRoomba:
Doesn't this kinda feel like bypassing the test? I guess it's sort of an alleviation until the test is fixed.

Ideally we want to make sure everything is still functioning in the instance that Nukeops gets something tacked onto it that causes it to fail.

slarticodefast:
While the random test fails were annoying, the fact that this test ran other gamemodes as well did help catch a lot of bugs.
For example the subwizard rule being abstract was found because it caused a 5% heisentest.

ArtisticRoomba:
I had a simple hate-fix up on my local that iterated over all GridRuleComponents to check and make sure all grids were loaded and properly functioning. The only logic I was having trouble implementing is passing the nukie planet grid to the rest of the logic to perform the checks related to it, as I didn't want to implement it at the time.

If we checked to make sure any grids spawned by a Gamerule and a SubGamerule actually exist, while still checking everything else the same, I think it would be more future proof and maintainable than simply ignoring the Subgamerules. But that's my opinion.

ArtisticRoomba:
Although a better fix that doesn't cause heisentests is to just add all applicable subgamerules to the test at a 100% chance so we don't ever have to deal with the 5% fail chance in the future, we just test if it works willy nilly.

slarticodefast:
Don't we have one already that starts all gamerules?
I wonder why they only ever fail in the nukeops test in that case? Just because it runs the simulation for some time maybe?

ArtisticRoomba:
Doesn't that one also heisentest too sometimes? Might be the same core issue as #35868.

Tayrtahn:
> Don't we have one already that starts all gamerules?
> I wonder why they only ever fail in the nukeops test in that case? Just because it runs the simulation for some time maybe?

Because the failure is caused by the NukeOps test insisting on having only one RuleGrids. The gamerules can coexist happily - the NukeOps gamerule doesn't have a problem with other RuleGrids; the problem is that the NukeOps test has logic that specifically can't handle it.

ArtisticRoomba:
If the StartAllGamerules test checks to make sure if nukies start properly along with everything else, then this change to only test nukeops is okay IMO as the other cases will be checked by the mass test.

Tayrtahn:
> If the StartAllGamerules test checks to make sure if nukies start properly along with everything else, then this change to only test nukeops is okay IMO as the other cases will be checked by the mass test.

https://github.com/space-wizards/space-station-14/blob/master/Content.IntegrationTests%2FTests%2FGameRules%2FStartEndGameRulesTest.cs

ArtisticRoomba:
Yeah, then the logic change for checking other RuleGrids needs to be implemented, as the NukeOpsTest checks if everything is in the right place, rather than a simple exception.

deltanedas:
i wasnt clear enough in my comment, the nukeops test should be using `Comp<RuleGridsComponent>()`on a nukeops rule. assuming there is 1 rule with NukeopsRuleComponent is probably fine

Tayrtahn:
Closing this in favor of fixing up the test.

PR #35880, opened 2025-03-16
Add an interaction test for cargo deliveries

Tayrtahn:
## About the PR

Adds an interaction test that verifies that various features of cargo delivery gameplay are working as intended.

## Why / Balance

It's never too soon to add tests!

## Technical details

Tests the following:
- The base `LetterDelivery` prototype has `DeliveryComponent` and `FingerprintReaderComponent`.
- The standard NT station is set up to support deliveries.
- The recipient name, job title, and station are correctly assigned when there is only one station record.
- The delivery spawns closed and locked.
- Using a locked delivery in hand will try to unlock it.
- The delivery cannot be unlocked with the wrong fingerprint.
- The delivery can be unlocked with the correct fingerprint.
- Unlocking the delivery increases the station's bank balance.
- Unlocking the delivery does not also open it.
- Using an unlocked delivery in hand will open it.
- The station's bank balance is not affected by opening a delivery that is already unlocked.
- The player is given the reward loot item when opening the delivery.
- The now-empty delivery is not deleted when opened.
- The now-empty delivery is considered space garbage.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/Cargo/DeliveryInteractionTest.cs
+++ b/Content.IntegrationTests/Tests/Cargo/DeliveryInteractionTest.cs
@@ -1,0 +1,57 @@
+#nullable enable
+using Content.IntegrationTests.Tests.Interaction;
+using Content.Server.Forensics;
+using Content.Shared.Delivery;
+using Content.Shared.FingerprintReader;
+using Content.Shared.Forensics.Components;
+
+namespace Content.IntegrationTests.Tests.Cargo;
+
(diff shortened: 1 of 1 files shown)


ScarKy0:
> Unlocking the delivery increases the station's bank balance.
The player is given the reward loot item when opening the delivery.

Id be wary of those three
In the future deliveries will apply penalties when broken open etc, so that might conflict with the first test
and empty deliveries are not out of the question, what if at some point we make opening one raise an event instead of give an item?

PR #35887, opened 2025-03-17
ranger hardsuit for security

K-Dynamic:
## About the PR

- Adds the ranger hardsuit, a lightweight security hardsuit that trades protection for extra mobility.
- Adds locker fill for ranger hardsuit for mapping.

Ranger hardsuit stats:
- Decreases running speed by 15%
- Blunt damage reduced by 10%
- Slash damage reduced by 15%
- Piercing damage reduced by 25%
- Caustic damage reduced by 30%
- Explosion damage reduced by 30%

## Why / Balance

The ranger hardsuit follows a similar theme with the syndicate elite hardsuit; it is envisioned as a lighter alternative to the security hardsuit, letting security cover more ground and chase criminals at cost of protection.

Although the hardsuit's pierce stat is near-identical to normal body armour, it is much weaker to blunt and slash damage. This also makes fighting against security less predictable, as ranger hardsuits are much better at flanking and initiating fights compared to normal suits. This should also discourage security from wearing it all the time compared to their normal vests.

Security could also do with some variety in equipment as a personal thought.

## Technical details

yml and rsi changes

## Media

image

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Adds the ranger hardsuit, a lightweight security hardsuit that trades protection for extra mobility.


SlimmSlamm:
I think the helmet face looks pretty weird. I also pray to god we don't start seeing security wearing these all shift every shift

DieselMohawk:
Love the idea, but since they're only obtainable via mapping, it may be worthwhile to create a crafting recipe until they're properly added.

Example: Security hardsuit + 30 seconds -> Ranger Hardsuit

Same for the inverse: Ranger Hardsuit + 30 seconds -> Security Hardsuit

millionse:
Awaiting rework of warden hardsuit to be sec jugsuit with movement speed of old emergency EVA

Pumkin69:
I really enjoy the idea of this because, yeah as a security main i literally will just eat spacing damage and wear better armor for combat because they don't slow me down and have better stats. If i had the option to barely slow myself down at all with the stats of a normal armor vest i would probably use it. The speed penalty is enough to scare security officers off from using it all shift. But the only thing I'm worried about is me saying "wow this still isn't good enough" and then continue using the better vests and eat the spacing damage. I'm not sure if the best route for this balance is to increase spacing damage so they are FORCED to use EVA. Possibly just reducing the movespeed penalty for the regular security hardsuit would be good enough but I'm not sure and i literally main security. 

TLDR: The balancing act between spacing damage, security and how spacing impacts the rest of the crew is hard and i don't know if this is the best way to fix it.

TiniestShark:
Leaving a comment in case this gets merged so I can make liz sprites for it.

keronshb:
Maints spoke and we decided we're not going to add this. 

- Security is fine just having one hardsuit type. Having a second adds too much bloat.
- The visuals are very similar to the other hardsuit
- The speed isn't punishing enough. If it has spacing protection and armor it needs a relatively high slowdown.

PR #35888, opened 2025-03-17
Increase storage size of Syndicate Backpack

SlamBamActionman:
## About the PR

This PR adds 7 storage slots to the Syndicate Backpack, increasing it from 28 slots to 35.

## Why / Balance

The Syndicate Backpack is a fairly low-purchase item for both Syndicate and Nukies alike:
Last 30 days PDA uplink: 7 purchases (10th least bought item *including Pointless category*)
Last 30 days Radio uplink: 46 purchases (similar purchase stats as Chemical Synthesis Kit and Web Vest)

The addition of 7 slots makes it lower than a regular duffelbag;
Backpack: 28
Satchel: 32
Syndicate backpack: 35
Duffelbag: 40
Syndicate duffelbag: 45

This effectively means you are trading 10% movement speed for 10 backpack slots. 

The slots are added to the right side of the backpack (rather than the bottom) to keep it consistent with duffelbags only being 5 slots tall. The 3-column can fit that lil' epipen or viper nicely.

## Technical details

Yaml change.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The Syndicate Backpack can now fit more items (35 item slots, from 28).

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Back/backpacks.yml
+++ b/Resources/Prototypes/Entities/Clothing/Back/backpacks.yml
@@ -278,6 +278,10 @@
     sprite: Clothing/Back/Backpacks/syndicate.rsi
   - type: ExplosionResistance
     damageCoefficient: 0.1
+  - type: Storage
+    grid:
+    - 0,0,7,3
+    - 8,1,8,3
 
 #Special
 - type: entity


ScarKy0:
image
Im going to kill you

kosticia:
Evil backpack of death. Too much scary to review😨😨😨

FairlySadPanda:
For what it's worth not enough gridinv uses interesting layouts.

But really it should be NT cutting corners not the Syndicate :godo:

SlamBamActionman:
Fwiw I wanted to find a nice middle and the funky little thing on the end seemed nice. +4 felt too little and +8 felt too much (I know it's just 1 more but the vibes were off); I tried +6 to put 3-columns on either side but that made the window look awkward and tbh you shouldn't have to inventory-tetris *too* much during Nukies.

I could've done +7 by making it use 5-tall columns but that felt a bit boring and oddly shaped for a backpack. So the little 3-tall column seemed like a fun odd shape that is actually usable; either slot in your viper in that corner or fit your 1x1 epipen.

FairlySadPanda:
Put those 3 slots in their own floating section, for your LTT water bottle :)

SlamBamActionman:
> I think extra space is a good idea, as it appears to already have extra damage resistance? But adding extra inventory slots is storage inflation and is a challenging thing to make sure doesn't become excessive.

The damage resistance gives it parity with the syndie duffel, so that is less "in its favor" and more "baseline expectation to not blow up your grenades". The one thing the backpack got going for it over the duffel is not having 10% slowdown, and sacrificing 17 storage slots for that is pretty brutal.

> Put those 3 slots in their own floating section, for your LTT water bottle :)
> 
> Could do three at each side so the symmetry is nice and makes the backpack feel unique and special.

Syndie satchel! It'd probably be a better fit since that already has "split inventory slot groups" as an established gimmick.

FairlySadPanda:
Buy your syndicate satchel at www.fuckNT.com

pheenty:
wtf is this 💀 
why not just make it 7x5...

PR #35890, opened 2025-03-17
Decrease Syndicate raid suit size, add bundle backpack

SlamBamActionman:
## About the PR

This PR reduces the size of the Syndicate raid suit from 4x4 to 2x2 (2x3 with helmet).

It also makes the raid bundle container a Syndicate Backpack instead of a Syndicate Duffelbag.

## Why / Balance

The Syndicate raid suit is the least bought "Nukie armor" from the Radio uplink. It's a very powerful armor stat-wise, but the cost of giving up EVA capabilities is high, so merely tweaking armor values would be a difficult way to make it more appealing to buy.

This PR focuses on another aspect. Transporting a raid suit to the station can be fairly cumbersome due to its size and makes it inflexible to use in pretty much any scenario that isn't docking directly to the station. The other option is to buy it on-station, which can be a bit difficult to coordinate. 

By reducing it to 2x2 it's now easier to bring the suit to the station in your duffel, and allows for more flexible usage such as StealthOps. The combat capabilities are still the same, now it's just easier to organize around. 

By making the bundle fit in a Syndicate Backpack instead of a Syndicate Bundle it adds some additional optional TC value (free backpack!) and gives Nukies a push to utilize it over the the duffelbag. The backpack would get additional appeal via #35888.

### Radio uplink stats (last 30 days)
Syndicate Elite Hardsuit: 313
Cybersun Juggernaut Suit: 194
Syndicate Raid Suit: 114

## Technical details

Made a backpack bundle in yaml since that didn't exist.

Migration moves `ClothingBackpackDuffelSyndicateRaidBundle` to `ClothingBackpackSyndicateRaidBundle` if someone mapped it for some reason.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The Syndicate Raid Bundle now comes in a Syndicate backpack instead of duffelbag.
- tweak: The Syndicate Raid Suit now has an inventory size of 2x2 (previously 4x4).


Warpzoned:
It honestly feels a bit weird to have them be of the same size (And general practicability) as the web vest while also being much better off stats-wise, was 3x3 ever put on the table as an alternative size tweak? (Would also make a bjt more sense when you consider the raid suit is a full-body sprite, head excluded)

All in all I do agree with the changes though, given that raid suits are super sick and were also somewhat underutilized!

SlamBamActionman:
> It honestly feels a bit weird to have them be of the same size (And general practicability) as the web vest while also being much better off stats-wise, was 3x3 ever put on the table as an alternative size tweak? (Would also make a bjt more sense when you consider the raid suit is a full-body sprite, head excluded)
> 
> All in all I do agree with the changes though, given that raid suits are super sick and were also somewhat underutilized!

The webvest is to me not really "Nukie armor"; it lacks heat protection which is vital to Nukies (crew will bust out lasers the moment Nukies are called, while for Traitors combat tends to be ballistics-oriented) and the larger TC pool means Nukies are freer to spend on the better armor options compared to the 4 TC webvest.

3x3 still felt a bit awkward to me considering it's the same size as a survival box and that's still a lot to lug around. Raid suits don't include the helmet so in effect it is a 3x2 for the full kit, which felt fair.

SlamBamActionman:
I checked the store stats and webvests have 46 radio uplink purchases over the last 30 days. Unfortunately I can't check the success rate of the rounds they were used in.

Pumkin69:
Seems like a good change to make the raidsuit used more in normal nuke ops. But will need to see if anything changes in nukies gameplay after the fact.

RedBookcase:
I like the idea, I think the raid suit deserves more use. This would make it a clear go-to option for StealthOps, and further encourages nukies going for that strategy to prioritize weapons that don't risk spacing, while still maintaining that as the core weakness of the Raid Suit.

beck-thompson (approved):
looks good to me!

PR #35891, opened 2025-03-17
Add a test to verify that all device signal sink ports can trigger

### LINKED ISSUE #32332: Add a test triggering all device signal sinks
## Description
The tech anomaly is currently causing the SpawnAndDeleteAllEntitiesInTheSameSpot to fail randomly.
It links up to random devices being spawned in the same spot and triggers them.
Some of these triggers throw errors (see #32330).
We should not leave that up to chance, but should test for it explicitly.

The test should:
- Spawn every entity that has a signal device sink.
- Trigger each sink to see if it throws an error.
- Keep in mind that some entities like C4 could destroy itself or others, so they should be spawned separately.

Tayrtahn:
## About the PR

Adds an integration test that spawns every entity that has a `DeviceLinkSinkComponent` and triggers all the ports on each one.

## Why / Balance

Resolves #32332.

## Technical details

The test spawns a dummy entity to act as a signal source and connects it to each port to send the signal. The sent signal is just an empty `null` signal.

A copy of each entity is spawned on a new map for each port so they can be tested individually without interfering with each other.

~I'm submitting this PR with a commit that reverts #32330 to demonstrate that it detects the error fixed by that PR. I'll revert the revert after the test failure is recorded to GitHub.~ Done. See https://github.com/space-wizards/space-station-14/actions/runs/13904215190/job/38903178381 for error.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
+++ b/Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
@@ -40,7 +40,6 @@ public sealed class DeviceLinkingTest
 
         var prototypes = server.ProtoMan.EnumeratePrototypes<EntityPrototype>().OrderBy(p => p.ID);
 
-        List<MapId> maps = [];
         await server.WaitAssertion(() =>
         {
             Assert.Multiple(() =>
@@ -57,7 +56,6 @@ public sealed class DeviceLinkingTest
                     {
                         // Create a map for each entity/port combo so they can't interfere
                         mapSys.CreateMap(out var mapId);
-                        maps.Add(mapId);
                         var grid = mapMan.CreateGridEntity(mapId);
                         mapSys.SetTile(grid.Owner, grid.Comp, Vector2i.Zero, new Tile(1));
                         var coord = new EntityCoordinates(grid.Owner, 0, 0);
@@ -84,15 +82,11 @@ public sealed class DeviceLinkingTest
                         // Send a signal to the port
                         Assert.DoesNotThrow(() => { deviceLinkSys.InvokePort(sourceEnt, "Output", null, sourceComp); },
                             $"Exception thrown while triggering port {port.Id} of sink device {proto.ID}");
+
+                        mapSys.DeleteMap(mapId);
                     }
                 }
             });
-
-            // Cleanup created maps
-            foreach (var map in maps)
-            {
-                mapSys.DeleteMap(map);
-            }
         });
 
         await pair.CleanReturnAsync();
--- a/Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
+++ b/Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
@@ -38,13 +38,7 @@ public sealed class DeviceLinkingTest
         var mapSys = server.System<SharedMapSystem>();
         var deviceLinkSys = server.System<DeviceLinkSystem>();
 
-        // Find every EntityPrototypes with a DeviceLinkSinkComponent
-        var prototypes = server.ProtoMan
-            .EnumeratePrototypes<EntityPrototype>()
-            .Where(p => !p.Abstract)
-            .Where(p => !pair.IsTestPrototype(p))
-            .Where(p => p.HasComponent<DeviceLinkSinkComponent>())
-            .OrderBy(p => p.ID);
+        var prototypes = server.ProtoMan.EnumeratePrototypes<EntityPrototype>().OrderBy(p => p.ID);
 
         List<MapId> maps = [];
         await server.WaitAssertion(() =>
@@ -53,8 +47,11 @@ public sealed class DeviceLinkingTest
             {
                 foreach (var proto in prototypes)
                 {
-                    Assert.That(proto.TryGetComponent<DeviceLinkSinkComponent>(out var protoSinkComp, compFact),
-                        $"Prototype {proto.ID} does not have a DeviceLinkSinkComponent!");
+                    if (proto.Abstract || pair.IsTestPrototype(proto))
+                        continue;
+
+                    if (!proto.TryGetComponent<DeviceLinkSinkComponent>(out var protoSinkComp, compFact))
+                        continue;
 
                     foreach (var port in protoSinkComp.Ports)
                     {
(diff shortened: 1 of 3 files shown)


Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
```diff
@@ -83,6 +83,8 @@ await server.WaitAssertion(() =>
 
                         // Send a signal to the port
                         deviceLinkSys.InvokePort(sourceEnt, "Output", null, sourceComp);
+
+                        Assert.That(server.EntMan.Deleted(sinkEnt), Is.False, $"{protoId} was deleted after triggering port {port.Id}");
```
slarticodefast, line 87: Isn't this something that could happen, for example due to explosions getting triggered?

Tayrtahn: I'll try reverting #32423 locally and see if it fails.

Tayrtahn: It doesn't fail, but it looks like that's because explosions are queued and processed later. That's not really good enough for me, since there could be some other result of a port triggering that would cause a more immediate problem.
I'll make it spawn a new map and entity for each port to make it extra safe.

slarticodefast: Might be fine in that case since explosions are queued up for the next tick.
But can you test a combination of DeleteOnTriggerComponent and TriggerOnSignalComponent?

Tayrtahn: There we go - a fresh map and entity for each port!

Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
```diff
@@ -0,0 +1,102 @@
...
+        var mapMan = server.ResolveDependency<IMapManager>();
+        var mapSys = server.System<SharedMapSystem>();
+        var deviceLinkSys = server.System<DeviceLinkSystem>();
+
+        // Find every EntityPrototypes with a DeviceLinkSinkComponent
+        var prototypes = server.ProtoMan
+            .EnumeratePrototypes<EntityPrototype>()
+            .Where(p => !p.Abstract)
+            .Where(p => !pair.IsTestPrototype(p))
+            .Where(p => p.HasComponent<DeviceLinkSinkComponent>())
```
deltanedas, line 46: dont use quad nested linq + use IComponentFactory version to avoid resolving a billion times

deltanedas (changes requested):

Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
Tayrtahn: Eh, sure. I'll convert these to guard clauses in the loop.

Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
```diff
@@ -0,0 +1,99 @@
...
+                        // Create a map for each entity/port combo so they can't interfere
+                        mapSys.CreateMap(out var mapId);
+                        maps.Add(mapId);
+                        var grid = mapMan.CreateGridEntity(mapId);
+                        mapSys.SetTile(grid.Owner, grid.Comp, Vector2i.Zero, new Tile(1));
+                        var coord = new EntityCoordinates(grid.Owner, 0, 0);
+
+                        // Spawn the sink entity
+                        var sinkEnt = server.EntMan.SpawnEntity(proto.ID, coord);
+                        // Get the actual sink component, since the one we got from the prototype doesn't its owner set up
```
slarticodefast, line 67:
```suggestion
                        // Get the actual sink component, since the one we got from the prototype doesn't have its owner set up
```

Tayrtahn: Ah, thanks!

Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
```diff
@@ -0,0 +1,99 @@
...
+                        deviceLinkSys.InvokePort(sourceEnt, "Output", null, sourceComp);
+                    }
+                }
+            });
+
+            // Cleanup created maps
+            foreach (var map in maps)
+            {
+                mapSys.DeleteMap(map);
+            }
```
slarticodefast, line 94: Maybe delete the maps directly after we test each entity?
Not sure about the performance impact of having hundreds of maps at the same time

Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
```diff
@@ -0,0 +1,99 @@
...
+                        // Create a link from the tester's output to the target port on the sink
+                        deviceLinkSys.SaveLinks(null,
+                            sourceEnt,
+                            sinkEnt,
+                            [("Output", port.Id)],
+                            sourceComp,
+                            sinkComp);
+
+                        // Send a signal to the port
+                        deviceLinkSys.InvokePort(sourceEnt, "Output", null, sourceComp);
```
slarticodefast, line 85: Is it possible to leave a message which protype caused an error?
Might not always be as obvious as in the case of the generator

Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
Tayrtahn: It doesn't seem to be a problem, but also: why not do it anyway? I'll make that change.

Content.IntegrationTests/Tests/DeviceLinking/DeviceLinkingTest.cs
Tayrtahn: Unfortunately, I haven't found a good way to do that. I can wrap the method call in `Assert.DoesNotThrow` to give it a custom message ("Exception thrown while triggering port Foo on sink entity Bar"), but that eats the stack trace.

Tayrtahn: Actually, I just tried it again and the output is a bit convoluted but it does seem to have all the info in it.

slarticodefast (approved):
Thanks for all the great work with the integration tests!

PR #35900, opened 2025-03-17
Remove hard coding of ColorFlashEffect colors, allow overrides

Quantum-cross:
## About the PR

This PR will lead the way to one solution of a problem raised in #35836, in which I made MMIs and PBs damageable, but without a destructible trigger. The problem being that if it flashes red, it implies that the item can eventually be destroyed.

This PR would allow a different color to be used on damageable components that cannot be destroyed, denoting some other type of "damage".

## Why / Balance

De-coupling hardcoded values of Color from various weapon systems, allowing more flexibility with ColorFlashEffect system

## Technical details

- Add an Enum field to ColorFlashEffectEvent to denote the source or type of event causing it.
- Change RaiseEffect with just color to protected so that a source must be provided (could be Other)
- Add public RaiseEffect overload to take a source as well as an optional color as an override.
- Remove hard coded color values from the different hit and weapon systems, and move them to the default color of the HitDamage and HitStamina effect sources in RaiseEffect.
- Add a ColorFlashEffectOverrideComponent that will denote a mapping of EventSource to Color for a particular entity.

## Media

Default Red:
Screenshot 2025-03-17 193254

Green:
Screenshot 2025-03-17 193312

Turquoise:
Screenshot 2025-03-17 193329

## Requirements

## Breaking changes

Function arguments for SharedColorFlashEffectSystem::RaiseEffect() are changed, must provide EffectSource value.


NoElkaTheGod:
Doesn't need changelog, not player-facing

Quantum-cross:
> Doesn't need changelog, not player-facing

Removed

Content.Shared/Effects/SharedColorFlashEffectSystem.cs
```diff
@@ -2,7 +2,31 @@
...
+        {
+            EffectSource.HitDamage => Color.Red,
+            EffectSource.HitStamina => Color.Aqua,
+            EffectSource.HitDamageInvulnerable => Color.Turquoise,
+            _ => Color.White,
+        };
+
+        RaiseEffect(source, color.Value, entities, filter);
+    }
 }
```
metalgearsloth, line 32: Doesn't hardcoding in a bunch of effects defeat the entire point?

rcross-lc: > Doesn't hardcoding in a bunch of effects defeat the entire point?

I think there are two main factors here to address separately:

1. Using `EffectSource` to "hard code" an enumerated list of "allowed sources"
2. "hard coding" certain default colors for certain sources

-----

I can defend 2 much more easily -- it's simply bringing all of the hard coded "Color.Red" scattered among various weapon systems and bring it all to one location.

That being said, would it be better to put the definitions of "DefaultHitDamageColor = Color.Red" (and stamina) somewhere else? I tried overriding the Color class from robust toolbox to add these as properties but the Color class is sealed.

-----

Point number 1 I introduced to allow the color override to target specific types of events. I don't want an override component to change the color of a colorflash induced by a fluid splash for example. Maybe an argument could be made to make the EffectSource a string? I kind of wanted to avoid making it a string because it's a network message.

-----

As this PR stands now, I give the effect giver the flexibility to provide an override color regardless of effect source (helps the fluid splash). Maybe a healing wand can trigger a green flash despite a DamageHit source. The effect receiver is given the flexibility to modify the color of the event with the Override component.

Quantum-cross: > Doesn't hardcoding in a bunch of effects defeat the entire point?

I think there are two main factors here to address separately:

1. Using `EffectSource` to "hard code" an enumerated list of "allowed sources"
2. "hard coding" certain default colors for certain sources

-----

I can defend 2 much more easily -- it's simply bringing all of the hard coded "Color.Red" scattered among various weapon systems and bring it all to one location.

That being said, would it be better to put the definitions of "DefaultHitDamageColor = Color.Red" (and stamina) somewhere else? I tried overriding the Color class from robust toolbox to add these as properties but the Color class is sealed.

-----

Point number 1 I introduced to allow the color override to target specific types of events. I don't want an override component to change the color of a colorflash induced by a fluid splash for example. Maybe an argument could be made to make the EffectSource a string? I kind of wanted to avoid making it a string because it's a network message.

-----

As this PR stands now, I give the effect giver the flexibility to provide an override color regardless of effect source (helps the fluid splash). Maybe a healing wand can trigger a green flash despite a DamageHit source. The effect receiver is given the flexibility to modify the color of the event with the Override component.

Quantum-cross:
Went ahead and moved the effect source key to a string

EmoGarbage404 (approved):

PR #35901, opened 2025-03-18
Renaming the "Mime Cap" to "White Cap"

YoungThugSS14:
## About the PR

I renamed the mime cap to the white cap.
## Why / Balance

The mime cap isn't only meant to be for mimes, so it should be renamed to be a white cap instead.
## Technical details

Simply just renamed files.
## Requirements

## Breaking changes
 N/A
**Changelog**
N/A
Doesn't need a changelog as it isn't anything important


slarticodefast:
only rename the name datafield, not the protoid or file names since those are not player facing and it would be a major breaking change for forks

YoungThugSS14:
Oops wait forgot to change 2 things back lmao.

slarticodefast (approved):
Looks like this item is available in tha clothes vendor, so it does make sense for it to not be mime specific.
grafik

beck-thompson (approved):

PR #35828, opened 2025-03-14
Make the captain and CentCom's mindshield implants unremoveable+

M1tht1c:
## About the PR
Made an unremoveable variant of the mindshield implant, and gave it to the captain and CentCom ghost roles. Also changed the visitor ghost roles with mindshields to actually use implants, so that they can be removed.

## Why / Balance
The [Revs rework doc](https://docs.spacestation14.com/en/space-station-14/round-flow/proposals/revolutionaries-codeword-rework.html#mindshields) specifies that the captain, and CentCom should be unconvertabe, and that their mindshields should be unremoveable.

## Technical details
Added UnremoveableMindShieldImplant and a corresponding implanter
Changed Captain to use UnremoveableMindShieldImplant 
Changed EventHumanoidMindShielded to use AutoImplant instead of the MindShield component (making the mindshield removeable)
Removed AutoImplant from EventHumanoidMindShielded's children
Added EventHumanoidPermaMindShielded
Changed CentCom ghost roles, and the visiting captain ghost role, to use EventHumanoidPermaMindShielded

## Requirements
- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: The Captains mindshield can no longer be removed.
- fix: Visitors can have their mindshields removed.

PR #35905, opened 2025-03-18
minor MindShieldSystem cleanup

slarticodefast:
## About the PR
title

## Why / Balance
weird code

## Technical details
Why the hell are we using a tag if we could instead just subscribe with the correct component?

## Requirements

## Breaking changes
removed the `MindShield` tag

**Changelog**
not player facing

### DIFF
--- a/Content.Server/Mindshield/MindShieldSystem.cs
+++ b/Content.Server/Mindshield/MindShieldSystem.cs
@@ -4,51 +4,44 @@ using Content.Server.Popups;
 using Content.Server.Roles;
 using Content.Shared.Database;
 using Content.Shared.Implants;
-using Content.Shared.Implants.Components;
 using Content.Shared.Mindshield.Components;
 using Content.Shared.Revolutionary.Components;
-using Content.Shared.Tag;
 using Robust.Shared.Containers;
 
 namespace Content.Server.Mindshield;
 
 /// <summary>
-/// System used for checking if the implanted is a Rev or Head Rev.
+/// System used for adding or removing components with a mindshield implant
+/// as well as checking if the implanted is a Rev or Head Rev.
 /// </summary>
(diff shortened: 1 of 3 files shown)


beck-thompson:
Thank god because so many people use the mindshield as a base implant reference and end up copying this sus code

M1tht1c:
[35828](for https://github.com/space-wizards/space-station-14/pull/35828) currently uses the tag for a variant of the mindshield implant, so there is a purpose to using the tag. Technically could change the PR to add the variant to the check but that feels like unnecessary hard-coding.

slarticodefast:
> [35828](for https://github.com/space-wizards/space-station-14/pull/35828) currently uses the tag for a variant of the mindshield implant, so there is a purpose to using the tag. Technically could change the PR to add the variant to the check but that feels like unnecessary hard-coding.

It doesn't seem like it uses that specific tag either, it only introduces two new ones.

M1tht1c:
> > [35828](for https://github.com/space-wizards/space-station-14/pull/35828) currently uses the tag for a variant of the mindshield implant, so there is a purpose to using the tag. Technically could change the PR to add the variant to the check but that feels like unnecessary hard-coding.
> 
> 
> 
> It doesn't seem like it uses that specific tag either, it only introduces two new ones.

UnremovableMindShieldImplant does use the MindShield tag (line 379 in the subdermal implants file), however the two new tags are introduced because I tried using a whitelist to make the implants mutually exclusive, which I have realised was probably not the best approach.

slarticodefast:
>UnremovableMindShieldImplant does use the MindShield tag (line 379 in the subdermal implants file), however the two new tags are introduced because I tried using a whitelist to make the implants mutually exclusive, which I realise was probably not the best approach.

Instead of using whitelists you should simply cancel `AddImplantAttemptEvent` when they already have the MindShieldComponent

PR #35907, opened 2025-03-18
Predict and cleanup RingerComponent

MilonPL:
## About the PR

Moves RingerComponent and RingerUplinkComponent to shared, removes UI state, uses compstate, predicted messages.
Sound isn't predicted because I'm too dumb to use it with the tools I have. If anyone else wants to have a go at it, feel free to.

Needed for PDA prediction, decided to start with all of the small dependencies first.

Requires https://github.com/space-wizards/RobustToolbox/pull/5752

## Technical details

The code is pretty self explanatory and I tried my best to document the old stuff. Does what all other predicted UIs do with compstates cause UI state bad. Also uses field deltas because the state is pretty big. Like I said, couldn't get predicted sound to work without duping so it just stays on the server for now. Maybe I'll think of a way to solve this in a separate PR. Right now I just need this merged for PDAs.

## Requirements

## Breaking changes

Removed `RingerRequestUpdateInterfaceMessage.`
Removed `RingerUpdateState.`
Removed `ActiveRingerComponent.`
`ToggleRingerUi` was renamed `TryToggleRingerUi`.
`RingerComponent` and `RingerUplinkComponent` have been moved to Shared.
All of the Public API methods from server RingerSystem have been moved to SharedRingerSystem and changed to use the `Entity<T?>` pattern.

**Changelog**

no one will notice


slarticodefast (changes requested):
Found a few bugs while testing:
- I don't hear any audio from the ringer
- The test button does not seem to be predicted, but has a small delay
- After playing around with it a little the text fields started glitching
example

Did not look at the code in detail yet

MilonPL:
Fixed the Test button delay, the audio seems to be playing normally for me? Not sure what you meant by the buttons glitching out, but if it's about LineEdits resetting, I don't think there's a fix for it. The UI will update every time it receives a new compstate, and those are being read from the state.

Content.Shared/PDA/SharedRingerSystem.cs
```diff
@@ -1,14 +1,331 @@
...
+        if (ent.Comp.Active)
+            return;
+
+        ent.Comp.Active = true;
+        ent.Comp.NoteCount = 0;
+        ent.Comp.NextNoteTime = _timing.CurTime;
+
+        UpdateRingerUi(ent);
+
+        // No predicted popups with PVS filtering so we do this to avoid duplication
```
slarticodefast, line 239: Maybe add an overload of PopUpPredicted for that then?
Having this here looks very messy

Content.Shared/PDA/SharedRingerSystem.cs
```diff
@@ -1,14 +1,331 @@
...
+        // Client sent us an updated ringtone so set it to that.
+        if (args.Ringtone.Length != RingtoneLength)
+            return;
+
+        var ev = new BeforeRingtoneSetEvent(args.Ringtone);
+        RaiseLocalEvent(ent, ref ev);
+        if (ev.Handled)
+            return;
+
+        DirtyField(ent.AsNullable(), nameof(RingerComponent.LastRingtoneSetTime));
```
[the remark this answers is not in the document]
MilonPL: No idea, it was a part of the original code. It's not used anywhere else and I don't think it makes sense as an event.

Content.Shared/PDA/SharedRingerSystem.cs
```diff
@@ -1,14 +1,331 @@
...
+
+    /// <summary>
+    /// Locks the uplink and closes the window, if its open.
+    /// </summary>
+    /// <remarks>
+    /// Will not update the PDA ui so you must do that yourself if needed.
+    /// </remarks>
+    public void LockUplink(Entity<RingerUplinkComponent?> ent)
+    {
+        if (!Resolve(ent, ref ent.Comp))
```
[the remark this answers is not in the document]
MilonPL: That was my concern at first but clients can use the StoreComponent for traitor detection anyways. RingerUplinkComponent needs to be networked to properly predict unlocking uplinks and setting ringtones (since  we check for TryToggleUplink before writing the ringtone).
Having it networked is not as bad if the Code is left serverOnly, but it can't be for the TryToggleUplink call, which means cheat clients could not only see but also unlock any uplink.

What do you think would be a good way of handling it? I was thinking about storing the actual code on an entity in nullspace, and RingerUplink only storing that EntityUid, then using a session override to send it to all traitors. But it means that it wouldn't be predicted or would misspredict on non-traitor clients.

Or are we willing to just bite the bullet and patch it *if* it's used as an exploit?

JIPDawg:
Lgtm

Content.Server/PDA/Ringer/RingerSystem.cs
```diff
@@ -1,253 +1,131 @@
...
-            _ui.SetUiState(uid, RingerUiKey.Key, new RingerUpdateState(isPlaying, ringer.Ringtone));
-        }
+    /// <summary>
+    /// Generates a random ringtone using the C pentatonic scale.
+    /// </summary>
+    /// <returns>An array of Notes representing the ringtone.</returns>
+    /// <remarks>The logic for this is on the Server so that we don't get a different result on the Client every time.</remarks>
+    private Note[] GenerateRingtone()
+    {
+        // Default to using C pentatonic so it at least sounds not terrible.
```
metalgearsloth, line 91: Thanks sloth

Content.Shared/PDA/Ringer/RingerComponent.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public Note[] Ringtone = new Note[SharedRingerSystem.RingtoneLength];
+
+    /// <summary>
+    /// The last time this ringer's ringtone was set.
+    /// </summary>
+    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer)), AutoPausedField, AutoNetworkedField]
+    public TimeSpan LastRingtoneSetTime;
+
```
metalgearsloth, line 21: Treating it as NextRingtoneSetTime means you don't need to do last time + delay every single time on it.

Content.Shared/PDA/Ringer/RingerComponent.cs
```diff
@@ -0,0 +1,57 @@
+using Robust.Shared.GameStates;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
+namespace Content.Shared.PDA.Ringer;
+
+[RegisterComponent, NetworkedComponent, Access(typeof(SharedRingerSystem))]
+[AutoGenerateComponentState(true, true), AutoGenerateComponentPause]
+public sealed partial class RingerComponent : Component
```
metalgearsloth, line 8: named args blease I beg, fieldDeltas: true. Makes future updates way easier and avoid bugs if args change.

metalgearsloth (approved):

MilonPL:
ok time to see what breaks

Content.Client/PDA/Ringer/RingerSystem.cs
```diff
@@ -0,0 +1,56 @@
...
+            return false;
+
+        if (!HasComp<StoreComponent>(uid))
+            return false;
+
+        // Special case for client-side prediction:
+        // Since we can't expose the uplink code to clients for security reasons,
+        // we assume if an antagonist is trying to set a ringtone, it's to unlock the uplink.
+        // The server will properly verify the code and correct if needed.
+        if (IsAntagonist(user))
```
slarticodefast, line 50: I beg you please kill this part.
It is so cursed and will mispredict for any antag

Samuka-C:
There goes my xenoborgs again
gotta find another useless unused component to make the pinpointer work

PR #35909, opened 2025-03-18
Paradox Clones spawn with their suit sensors off

slarticodefast:
## About the PR
title

## Why / Balance
They sometimes got caught immediately if the AI or paramed was watching the console.
This gives them the chance to decide themselves if and when to enable the suit sensors.

## Technical details
added a function to set all suit sensors in someone's inventory

## Media
Screenshot (379)

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- tweak: Paradox Clones now start with their suit sensors turned off.

### DIFF
--- a/Content.Server/GameTicking/Rules/ParadoxCloneRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/ParadoxCloneRuleSystem.cs
@@ -1,11 +1,12 @@
 using Content.Server.Antag;
 using Content.Server.Cloning;
 using Content.Server.GameTicking.Rules.Components;
+using Content.Server.Medical.SuitSensors;
 using Content.Server.Objectives.Components;
 using Content.Shared.GameTicking.Components;
 using Content.Shared.Gibbing.Components;
+using Content.Shared.Medical.SuitSensor;
 using Content.Shared.Mind;
-using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
 
 namespace Content.Server.GameTicking.Rules;
@@ -16,6 +17,7 @@ public sealed class ParadoxCloneRuleSystem : GameRuleSystem<ParadoxCloneRuleComp
     [Dependency] private readonly SharedMindSystem _mind = default!;
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly CloningSystem _cloning = default!;
+    [Dependency] private readonly SuitSensorSystem _sensor = default!;
 
     public override void Initialize()
     {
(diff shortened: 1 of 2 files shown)


ScarKy0:
Since you also added copying implants, won't this mess with a tracker if it copies someone implamted?

slarticodefast:
> Since you also added copying implants, won't this mess with a tracker if it copies someone implamted?

No, trackers cannot be turned off, so they will still show up in that case, but I think that is fine.
This PR only sets the sensors in your equipment slots, which you can toggle yourself.

Content.Server/GameTicking/Rules/ParadoxCloneRuleSystem.cs
```diff
@@ -72,6 +74,9 @@ private void OnAntagSelectEntity(Entity<ParadoxCloneRuleComponent> ent, ref Anta
         gibComp.SpawnProto = ent.Comp.GibProto;
         gibComp.PreventGibbingObjectives = new() { "ParadoxCloneKillObjective" }; // don't gib them if they killed the original.
 
+        // turn their suit sensors off so they don't immediately get noticed
```
FairlySadPanda, line 77: Why can't this be covered by an event? Calling an external gameplay system in a separate namespace is tight coupling.

This would be ideomatic for SuitSensor. See: `SubscribeLocalEvent<SuitSensorComponent, EmpPulseEvent>(OnEmpPulse);` in that system for an example of it reacting to external gameplay activity rather than being directly called.

slarticodefast: And I cannot do it like `EmpPulseEvent` because that one is directly raised on every entity in range of the pulse, rather than on a specific player.

FairlySadPanda: _can't mark own threads as resolved because Github permissions are a bitch_

Content.Server/Medical/SuitSensors/SuitSensorSystem.cs
```diff
@@ -1,4 +1,5 @@
 using System.Numerics;
+using System.Threading.Tasks.Sources;
```
Tayrtahn, line 2: Unused?

slarticodefast: Huh, where did that come from?
Must be my IDE trolling me.
I removed it.

Content.Server/Medical/SuitSensors/SuitSensorSystem.cs
```diff
@@ -347,6 +349,20 @@ public void SetSensor(Entity<SuitSensorComponent> sensors, SuitSensorMode mode,
         }
     }
 
+    /// <summary>
+    ///     Set all suit sensors on the equipment someone is wearing.
```
Tayrtahn, line 353:
```suggestion
    ///     Set all suit sensors on the equipment someone is wearing to the specified mode.
```

Tayrtahn (approved):

Content.Server/GameTicking/Rules/ParadoxCloneRuleSystem.cs
metalgearsloth: I would've thought this would be its own component given how generic it is?

PR #35913, opened 2025-03-18
Implement client-side theming for OutputPanel scroll-down button.

ciaran:
## About the PR

Adds a new button to scroll back down to the bottom of the chat feed when the chat box is currently scrolled up.

## Why / Balance

The intent is to both make it clear when incoming chat messages are being missed and provide a quick way to return to the live feed.

## Technical details

Amends `OutputPanel` in `RobustToolbox` to add the button, which is displayed when the panel output is scrolled up.
This repo contains the changes to style the new button.

~~*NOTE*: As-is, this affects all uses of `OutputPanel` (e.g. the console). I see no reason this wouldn’t be desirable but if needed it can be amended to affect only the chat box.~~ This is now opt-in only, enabled for debug console and (in this PR) the chat box.

~~*NOTE*: This is currently a draft as it depends on the `RobustToolbox` changes being available – requires https://github.com/space-wizards/RobustToolbox/pull/5750~~

## Media

Untitled3

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Added a button to scroll back to the bottom of the chat box


slarticodefast:
Copying my comment from https://github.com/space-wizards/RobustToolbox/pull/5750
I think something smaller like this should be enough
423548525-2cea7409-c642-4bd1-b1cd-8f17fbe2b11e
In the corner it would not block any text. And an down arrow should be self-explanatory, so there is no need for a "scroll down" text, that can be handled by a tooltip.

ciaran:
>  I think something smaller like this should be enough

I agree that this button would fit nicely but to me this fails at the second point of drawing attention to the fact that the player is missing the active chat feed. I think this is also important because it's easy to be scrolled up and not realise that you're not seeing new messages coming in.

Tayrtahn:
Maybe a smaller button that just says "New" and has the arrows?

Tayrtahn:
The check failed because you committed changes to RobustToolbox as part of this PR. You'll need to remove those changes - the "requires" text you put in the description will make GitHub pull in that engine PR for you before it builds.

metalgearsloth:
https://github.com/space-wizards/space-station-14/issues/36676 prevents me from testing but seems okay

PR #35916, opened 2025-03-19
Remove non-networked Dirty call in GuardianSystem

Tayrtahn:
## About the PR

Removed a `Dirty` method invocation in `GuardianSystem` that should not be there.

## Why / Balance

`GuardianSystem.OnGuardianShutdown` calls `Dirty`, despite being a server-only system and the `GuardianComponent` being dirtied being a server-only, non-networked component.

## Technical details

Every time a holoparasite is deleted, an exception is raised:
```
[ERRO] entity: Caught exception while trying to call shutdown on component of entity 'Holoparasite (HOLO-294) (3016/n3016, MobHoloparasiteGuardian)'
Robust.Shared.Utility.DebugAssertException: Attempted to dirty a non-networked component: Content.Server.Guardian.GuardianHostComponent
```

You can test this by spawning a holoparasite injector, injecting yourself, activating the holoparasite, then using the debug action to delete it. The exception is caught and displayed in the server log.

## Requirements

### DIFF
--- a/Content.Server/Guardian/GuardianSystem.cs
+++ b/Content.Server/Guardian/GuardianSystem.cs
@@ -70,7 +70,6 @@ namespace Content.Server.Guardian
 
             _container.Remove(uid, hostComponent.GuardianContainer);
             hostComponent.HostedGuardian = null;
-            Dirty(host.Value, hostComponent);
             QueueDel(hostComponent.ActionEntity);
             hostComponent.ActionEntity = null;
         }


metalgearsloth (approved):

slarticodefast:
Is that something we can catch with an analyzer maybe, so it we don't have to watch error logs like this?

Tayrtahn:
> Is that something we can catch with an analyzer maybe, so it we don't have to watch error logs like this?

Funny you should ask that...

Tayrtahn:
https://github.com/space-wizards/RobustToolbox/pull/5754

PR #35917, opened 2025-03-19
Fix ninjas remaining invisible to other clients if they remove their suit while invisible

SlamBamActionman:
## About the PR

Bugfix. When unequipping the ninja suit while the cloaking is active, you become visible to your client but other clients still see you as invisible. This can be seen via examine, as you'll still have the "Shimmer" examine text that indicates invisible items.

## Why / Balance

This is a mald bugfix.

## Technical details

The reason why is because unequipping an item is meant to trigger `ComponentTogglerSystem`, which it does, but the event only fires *after* the unequip (and before the item is technically in your hands). This means that when it tries to remove the `StealthComponent` from the parent (i.e. the player), it instead targets the world entity as it sees it as its parent. 

Due to prediction, the client "incorrectly" assumes the player is still the parent, and thus removes the `StealthComponent` as one would expect.

This PR changes it so that `ComponentTogglerComponent` remembers its target after being set, and makes sure to remove components from that when toggled off.

## Steps to test

1. Boot up two clients.
2. `addgamerule NinjaSpawn` and make one client choose it.
3. Go invisible as ninja (make sure you are outside PVS range).
4. Spawn in a hardsuit and put it on *without turning off your invisibility*. You will become visible on your client.
5. Jetpack over to the station.
6. Turn on the ninja's invisibility power, and remove the suit.

On master you should still remain invisible on the non-ninja client, while with the PR that should be fixed.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Space ninjas now correctly lose invisibility for all players when removing their ninja suits.

### DIFF
--- a/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
+++ b/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
@@ -28,5 +28,7 @@ public sealed class ComponentTogglerSystem : EntitySystem
             EntityManager.RemoveComponents(target.Value, ent.Comp.RemoveComponents ?? ent.Comp.Components);
             ent.Comp.Target = null;
         }
+
+        Dirty(ent);
     }
 }
--- a/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
+++ b/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
@@ -16,11 +16,17 @@ public sealed class ComponentTogglerSystem : EntitySystem
 
     private void OnToggled(Entity<ComponentTogglerComponent> ent, ref ItemToggledEvent args)
     {
-        var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
-
         if (args.Activated)
+        {
+            var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
+            ent.Comp.Target = target;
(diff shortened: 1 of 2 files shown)


Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
```diff
@@ -16,11 +16,17 @@ public override void Initialize()
...
+            var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
+            ent.Comp.Target = target;
             EntityManager.AddComponents(target, ent.Comp.Components);
+        }
         else
-            EntityManager.RemoveComponents(target, ent.Comp.RemoveComponents ?? ent.Comp.Components);
+        {
+            var target = ent.Comp.Target != null ? ent.Comp.Target : (ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner);
+            EntityManager.RemoveComponents(target.Value, ent.Comp.RemoveComponents ?? ent.Comp.Components);
+            ent.Comp.Target = null;
```
metalgearsloth, line 29: Is target networked / does it need dirtying.

SlamBamActionman: It did! Makes it work a lot nicer too :)

AJCM-git:
@SlamBamActionman is this still relevant?

SlamBamActionman:
Nope! This appears to have been solved in #37309!

PR #35918, opened 2025-03-19
Fix duplicate suit signals

### LINKED ISSUE #28335: Tracking implants show duplicates on crew monitorg
## Description

Crew members who are implanted with implant trackers (and have suit sensors on) will appear twice on the crew monitoring console as two separate (but identical) entries. This clutters the monitor with redundant entries

**Reproduction**

1. Turn suit sensors on (any level)
2.  Implant a tracking implant
3.  You will appear on the crew monitoring console twice

**Screenshots**

Screenshot from 2024-05-28 16-25-28

ciaran:
## About the PR

Currently if a crew member has both a tracking implant and suit sensors enabled, they would be displayed twice in the crew monitoring console. 

Fixes #28335 which also includes a screenshot of the issue.

## Why / Balance

This is quite annoying to use and doesn't feel like it should be intended, but now with the addition of the Paradox Clones it seems important that this is changed since seeing a duplicate crew member entry should probably specifically alert the user to the existence of a clone.

Note I’ve opted to display the status entry with the most data available (i.e. if someone has a tracking implant and suit sensors set to binary, the console will show the full data with coords.) as the reverse would be a balance change.

## Technical details

This adds the UID of the entity being tracked to `SuitSensorStatus` data and associated network payloads. The console window uses this to show a single row per entity, prioritising entries with damage and/or coordinates included if available.

## Media

N/A

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed duplicate crew monitor console entries for crew with tracking implants

### DIFF
--- a/Content.Client/Medical/CrewMonitoring/CrewMonitoringWindow.xaml.cs
+++ b/Content.Client/Medical/CrewMonitoring/CrewMonitoringWindow.xaml.cs
@@ -79,14 +79,23 @@ public sealed partial class CrewMonitoringWindow : FancyWindow
 
         NoServerLabel.Visible = false;
 
-        // Convert to a list with a single sensor status per crew member
-        var uniqueSensors = sensors
-            .GroupBy(s => s.OwnerUid)
-            .Select(allSensors =>
-                // Prioritise an entry with coords if it exists
-                allSensors.OrderByDescending(sensor => sensor.Coordinates != null).ThenBy(sensor => sensor.DamagePercentage != null).First()
-            )
-            .ToList();
+        // Collect one status per user, using the sensor with the most data available.
+        Dictionary<NetEntity, SuitSensorStatus> uniqueSensorsMap = new();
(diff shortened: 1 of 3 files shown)


lzk228:
i think it's fine to leave it
instead an additional line/marking that will show where the signal comes from could be added

slarticodefast:
Uff, so many linq commands. Would be better to not use them since linq is quite slow and that window is regularly getting updated.

beck-thompson (approved):
Looks good! The only thing that concerns me is using `NET_OWNER_UID` in the suit sensor constants but I think its fine - it looks like other systems do similar things.

I also don't like the linq on 101 and 103 but it makes sense so I think its fine! :+1:

superjj18:
Is the nightmare finally over?  Probably the most prolific and commonly seen UI bug for over a year

slarticodefast (approved):
Looks good to me, thank you for your contribution!

PR #35922, opened 2025-03-19
C4 helmet unequip delay

psykana:
## About the PR
C4 head slot equip/unequip is now delayed similar to EVA helmet

## Why / Balance
I find the C4 helmet an amusing tool for hostage taking, but it's too easy to get rid of if the captive manages to break out of handcuffs.
Adding equip/unequip delays to put it in line with other "control" items like cuffs and blindfold.

## Technical details
Prototype change

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- add: C4 head slot equip/unequip is now delayed similar to EVA helmet

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
@@ -52,6 +52,8 @@
     slots:
     - HEAD
     quickEquip: false
+    equipDelay: 3
+    unequipDelay: 6
   - type: OnUseTimerTrigger
     delay: 10
     delayOptions: [10, 30, 60, 120, 300]


VerinSenpai:
you DARE REMOVE THE HOLY C4 HELM!???! *head explodes*

walksanatora:
might I interject. can the unequip delay be smth like 6+ seconds. that way you get a second to react to them taking off the c4. since iirc minimum c4 det time is 5 seconds. if you want c4 to be a hostage tool like you say it should be. it should be extremely risky to try and remove. and a watchful traitor can notice the do-after appear and have some leeway to explode them before it comes off.

slarticodefast:
> might I interject. can the unequip delay be smth like 6+ seconds

I think increasing it to 6 seconds for that reason should be fine.

beck-thompson (changes requested):
I agree with switching it to 6 seconds! I think its important that it could still explode

slarticodefast (approved):
Looks good to me!

beck-thompson (approved):

PR #35924, opened 2025-03-19
[April Fools] Pay your phone bill on time.

notquitehadouken:
## About the PR

I added phone bills

More precisely, I added an event that announces that it's Phone Bill Time at the start, waits, and then once time is up, it deletes every radio on people that don't have enough funds in their hands.

## Why / Balance

You better HOPE and PRAY that someone around you has 10 dollars because when the bell sounds you are utterly fucked if you can't present a few Lincoln-branded bills. You _GOTTA_ have money.

There is no balance. God is dead. Pony up, son.

## Media

https://github.com/user-attachments/assets/6cfa5ca3-7d93-4f52-8645-c8fbb4d0f0f2

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: And so I had them send me back to last Thursday night so I could pay my phone bill on time


ArtisticRoomba:
holy christ

Content.Server/StationEvents/Events/PhoneBillRule.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace Content.Server.StationEvents.Events;
+
+public sealed class PhoneBillRule : StationEventSystem<PhoneBillRuleComponent>
+{
+    [Dependency] private readonly IEntityManager _entMan = default!;
+    [Dependency] private readonly ContainerSystem _containerSystem = default!;
+    [Dependency] private readonly InventorySystem _inventorySystem = default!;
+    [Dependency] private readonly StackSystem _stackSystem = default!;
+
+    private Dictionary<EntityUid, List<EntityUid>> YoullPayForThis = new();
```
Tayrtahn, line 22: Move this to the rule component.

Content.Server/StationEvents/Events/PhoneBillRule.cs
```diff
@@ -0,0 +1,112 @@
...
+    protected override void Started(EntityUid uid, PhoneBillRuleComponent component, GameRuleComponent gameRule, GameRuleStartedEvent args)
+    {
+        if (YoullPayForThis.Count == 0)
+            return; // Huh.
+
+        int unpaid = 0;
+        foreach (var (forsaken, items) in YoullPayForThis)
+        {
+            bool tendered = false;
+            int required = component.Price * items.Count;
```
Tayrtahn, line 72: Use var

Content.Server/StationEvents/Events/PhoneBillRule.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+
+        if (unpaid == 0)
+        {
+            ChatSystem.DispatchGlobalAnnouncement(Loc.GetString("station-event-phone-bill-allpaid-announcement"), announcementSound: component.SuccessSound);
+        }
+        else
+        {
+            int unpaidPercent = (int)(100 * (float)unpaid / (float)YoullPayForThis.Count);
```
Tayrtahn, line 108:
```suggestion
            var unpaidPercent = (int)(100f * unpaid / YoullPayForThis.Count);
```

Nukesthestation:
dear god....

walksanatora:
suggestion: it should punt your HEADSET by frying all the encryption keys. that way it isn't pure shite. you can just beg hop and your head for new keys (or just go without a headset).
also I can forsee a funny. dropping your shite when you hear that alarm and then picking it back up after.

SlimmSlamm:
This just sounds horrible even for April Fools. It just wouldn't be fun at all.

FairlySadPanda:
This runs into unplayable territory. If this includes nukies and tots this also runs the risk of completely bricking the two most common game mode (nukies and tots losing their TC access).

There is a better implementation of this concept - consult last year's economy PR that added vending machines needing to be paid and apply it to headsets. Requiring payment to use radios is a good April Fools joke. Randomly taking a baseball bat to the legs of players for the Vine isn't.

FairlySadPanda:
Also this can be trivially gamed by throwing your items on the ground and picking them back up again after the cull.

Tayrtahn:
> Also this can be trivially gamed by throwing your items on the ground and picking them back up again after the cull.

I'm not gonna weigh in on whether this idea is good or not overall, but it does properly account for that, as shown in the video.

notquitehadouken:
btw dead people are immune, so if you're lying crit on the floor you won't be billed

ArtisticRoomba:
Oh so I just have to stab myself so I don't have to pay my bill

And then I need to die in a place that's all access so the paramedic can pick up my corpse

notquitehadouken:
> Oh so I just have to stab myself so I don't have to pay my bill
> 
> And then I need to die in a place that's all access so the paramedic can pick up my corpse

it only applies when the bill event happens

if you get a bill and THEN die, you will still lose your headset priviledge

FairlySadPanda:
> > Also this can be trivially gamed by throwing your items on the ground and picking them back up again after the cull.
> 
> I'm not gonna weigh in on whether this idea is good or not overall, but it does properly account for that, as shown in the video.

The egg, on my face, the humiliation. 

Code-wise to avoid evasion a dragged item should also be flagged to avoid every lizard player dragging their headset around. That is probably a simple change as I'd assume checking if the bill target is being dragged is trivial.

This is close to being fine - just breaking the headset (and indeed maybe all headsets they ever touch?) until the person holds enough money to pay the bill afterward gets this to being a funny reason for capitalism gaming rather than something that could make someone's round miserable.

notquitehadouken:
> Code-wise to avoid evasion a dragged item should also be flagged to avoid every lizard player dragging their headset around. That is probably a simple change as I'd assume checking if the bill target is being dragged is trivial, and I'm assuming a dragged item isn't inside a container that can trace back via parents to the billable entity.

I feel like "Drag important items around to avoid losing them" is an oxymoron

every greytider coming by and stealing my headset because i decided to drag it around instead of holding it:

Nukesthestation:
the humble handheld radio: :godo:

notquitehadouken:
> the humble handheld radio: :godo:

you also have to pay for those :trollface:

Nukesthestation:
the humble wall intercom: :godo:

Content.Server/StationEvents/Components/PhoneBillRuleComponent.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    ///     Cost of the phone bill per PDA or ID.
+    /// </summary>
+    [DataField]
+    public int Price = 0;
+
+    /// <summary>
+    ///     A list of PDAs and IDs on someone, keyed by that person.
+    /// </summary>
+    public Dictionary<EntityUid, List<EntityUid>> YoullPayForThis = new();
```
Tayrtahn, line 24: Ideally this would be a DataField. Does it cause any problems if you make it one?

Content.Server/StationEvents/Events/PhoneBillRule.cs
```diff
@@ -0,0 +1,110 @@
...
+
+    protected override void Started(EntityUid uid, PhoneBillRuleComponent component, GameRuleComponent gameRule, GameRuleStartedEvent args)
+    {
+        if (component.YoullPayForThis.Count == 0)
+            return; // Huh.
+
+        int unpaid = 0;
+        foreach (var (forsaken, items) in component.YoullPayForThis)
+        {
+            bool tendered = false;
```
Tayrtahn, line 69:
```suggestion
            var tendered = false;
```

Content.Server/StationEvents/Components/PhoneBillRuleComponent.cs
```diff
@@ -0,0 +1,43 @@
...
+    ///     The sound that plays when a failure announcement is sent.
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? FailureSound = new SoundPathSpecifier("/Audio/Effects/sadtrombone.ogg");
+
+    /// <summary>
+    ///     The sound that plays when a success announcement is sent.
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? SuccessSound = new SoundPathSpecifier("/Audio/Announcements/announce.ogg");
```
Tayrtahn, line 42: We're moving toward using sound collections for everything so we can support closed captions. Not sure how important that is for an April fools PR, but just fyi.

Content.Server/StationEvents/Components/PhoneBillRuleComponent.cs
notquitehadouken: YoullPayForThis is updated at runtime

Tayrtahn: We want to support being able to save and load the full game state. We don't currently, but we're making progress and I don't want to add stuff that doesn't work with it. If (hypothetically) the game were saved and restored between the rule being added and it being started, the data would be lost if it's not a DataField.

notquitehadouken: basically it's not actually specified when you define the component

notquitehadouken: thank you github for not showing me your comment until i resolved it, very cool

Tayrtahn: Yeah, it's not obvious (and we really should document it), but DataFields have two purposes: YAML configuration _and_ persistence.

Content.Server/StationEvents/Events/PhoneBillRule.cs
notquitehadouken: sorry about all the explicit type declarations i have spent a Little Too Much Time with C lately

FairlySadPanda:
> > This is close to being fine - just breaking the headset (and indeed maybe all headsets they ever touch?) until the person holds enough money to pay the bill afterward gets this to being a funny reason for capitalism gaming rather than something that could make someone's round miserable, and would be where I would take this. 100% better than the original "oh god" PR :godo:
> 
> this is slightly more effort than i am willing to put in
> 
> also you keep the keys, so if you do fail to pay your phone bill you just need to get a new empty headset and voila

No further objections, your honour.

Content.Server/StationEvents/Components/PhoneBillRuleComponent.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <summary>
+    ///     Cost of the phone bill per PDA or ID.
+    /// </summary>
+    [DataField]
+    public int Price = 0;
+
+    /// <summary>
+    ///     A list of PDAs and IDs on someone, keyed by that person.
+    /// </summary>
+    [DataField(readOnly:true)]
```
Tayrtahn, line 24:
```suggestion
    [DataField]
```
I believe `readOnly` is actually the opposite of what you want here - it prevents changing the value after it has been set in the YAML.

Tayrtahn (approved):

Pixeltheaertist:
Peak. My realistic working simulator just got even more realistic.

MilonPL (approved):
Code is okay-ish. Looks good for april fools.

PR #35925, opened 2025-03-19
Fix recipe and animation for copypasta

MisterImp:
## About the PR

Makes a few changes to the copypasta food recipe:

     • recipe now requires two small plates, because the sprite contains two plates

     • consuming it now drops two small plates as refuse, for the same reason

     • sprite animation now pauses for 5 seconds before looping

## Why / Balance

It didn't make sense that the plates appear and disappear out of nowhere, and the sprite animation just looked strange when it looped the moment it reached the end.

## Technical details

I have tested it to make sure the changes work as intended.

The only sprite change is that I swapped frames 8 and 9 of the animation, I feel it flows better like this.

## Media

copypastafix-demo1

https://github.com/user-attachments/assets/0558179a-d097-4bd1-87c0-db2518ef312f

https://github.com/user-attachments/assets/e2a39fdf-7458-4cde-bd06-0b3201f41347

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:

- tweak: The copypasta recipe now requires two small plates, which are left behind when it is consumed.


Resources/Prototypes/Entities/Objects/Consumable/Food/noodles.yml
```diff
@@ -57,6 +57,10 @@
   id: FoodNoodlesCopy
   description: You probably shouldn't try this, you always hear people talking about how bad it is...
   components:
+  - type: Food
+    trash:
+      - FoodPlateSmall
+      - FoodPlateSmall
```
slarticodefast, line 63:
```suggestion
    trash:
    - FoodPlateSmall
    - FoodPlateSmall
```
correct indentation

slarticodefast (changes requested):
You changelog is formatted wrong

MisterImp:
Indentation change committed, also added dash to changelog. Thanks slarti!

ArtisticRoomba:
Seems good to me. It seems like you modified the sprite so I would make sure to attribute yourself.

MisterImp:
> Seems good to me. It seems like you modified the sprite so I would make sure to attribute yourself.

Is it still necessary even if all I did was swap the order of two animation frames?

ArtisticRoomba:
Ehhhhh... yeah I guess it's fine then. A maint can decide if it needs attribution at the end of the day.

ArtisticRoomba (approved):
Lg2m

slarticodefast:
I had a closer look and switching the last two frames makes sense to me.
I guess it doesn't need an attribution change in this case because the sprites are already reordered into a different format when importing them from a dmi.

slarticodefast (approved):

beck-thompson (approved):

PR #35933, opened 2025-03-19
Fix migrations not being applied to PostMapInitTest

### LINKED ISSUE #35932: migration yml causing errors
## Description

Getting reports that adding new entries to migrations.yml doesn't work and will cause the map to crash when initialising.

Needs more info.

Tayrtahn:
## About the PR

Fixes the issue causing `PostMapInitTest` to fail on maps that contain migrated entities.

## Why / Balance

Fixes #35932.

## Technical details

The error shows up in `PostMapInitTest` when it calls `IsPreInit` (notably, this is called only for maps saved in version 7 format, which is why some maps seem to be immune). It creates an `EntityDeserializer` and uses it to read the map data without actually spawning it. The normal way of loading a map via `MapLoaderSystem` raises a `BeforeEntityReadEvent` which `MapMigrationSystem` subscribes to and supplies data about migrations, which are then passed into the `EntityDeserializer` in `MapLoaderSystem`. However, `PostMapInitTest` does not raise the event and creates its `EntityDeserializer` without any migration data.

To fix this, `PostMapInitTest` now raises a `BeforeEntityReadEvent` and passes the results into its `EntityDeserializer`.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/PostMapInitTest.cs
+++ b/Content.IntegrationTests/Tests/PostMapInitTest.cs
@@ -22,6 +22,7 @@ using Robust.Shared.EntitySerialization.Systems;
 using Robust.Shared.IoC;
 using Robust.Shared.Utility;
 using YamlDotNet.RepresentationModel;
+using Robust.Shared.Map.Events;
 
 namespace Content.IntegrationTests.Tests
 {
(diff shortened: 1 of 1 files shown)


ScarKy0:
Hella based

ArtisticRoomba:
bless you

PR #35938, opened 2025-03-19
Markings overhaul

TheShuEd:
## About the PR

Add new features into the Markings system, and cleaned up some old code at the same time (oh no violating guidelines)

`_displacement.TryAddDisplacement` has been partially redesigned to be more usable
Now it is no longer necessary to pass a list of layers inside, and the function returns the key of the new layer, so that this layer can be processed where needed in the respective systems (single responsibility principle).

Added Displacement maps support for any Markings layers via `HumanoidAppearanceComponent`.
`MarkingsPrototype` now has a `SupportDisplacement` field to determine if Displacement maps can be used for this particular markings. The default is true, since most markings are compatible with the standard human doll, but “special” markings for other races like Vox hair and beards should be marked false

The `SpeciesRestrictions` field in `MarkingPrototype` has been renamed to `SpeciesRestriction` to match the serializable name in the DataField. The override of this name was removed.

`MarkingPoint` got its own copy of the `OnlyWhitelisted` field, the same as in `MarkingPointsPrototype`, which now allows you to customize whitelists for each individual marking category:

```yml
- type: markingPoints
  id: MobVoxMarkingLimits
  onlyWhitelisted: false #<- old
  points:
    Hair:
      points: 1
      required: false
      onlyWhitelisted: false #<- new per category whitelist support
    FacialHair:
      points: 1
      required: false
      onlyWhitelisted: true #<- new per category whitelist support
```

As a result, we have the ability to use both standard human hair (with displacement maps overlaid) and special Vox hair (without displacement maps) on Vox. And more fine-tuning of whitelist allowed to leave other markings layers unique for Vox (you can't use human beards)

## Media

image

## Requirements

**Changelog**
:cl:
- add: Voxes can now use any standard human hairstyle.


mqole:
you are a hero for this. marking displacements are gonna be absolutely huge for custom species 👍

honeyed-lemons:
Quick question, does this work on hair and beards? They have some hardcoded features

TheShuEd:
yes it works

Content.Shared/Humanoid/HumanoidVisualLayers.cs
```diff
@@ -1,35 +1,32 @@
...
-        LFoot,
-        Handcuffs,
-        StencilMask,
-        Ensnare,
-        Fire,
+namespace Content.Shared.Humanoid;
 
-    }
+[Serializable, NetSerializable]
+public enum HumanoidVisualLayers : byte
```
FairlySadPanda, line 6: This PR features quite a few changes that are hard to diff and statically analyze; you could consider reducing the amount of tidy-up that isn't essential if  this takes time to get reviewers.

Flareguy:
I'm preeeeetty sure that the exclusive hairstyles vox get are a deliberate creative decision rather then a limitation that just exists by circumstance (much like their limited skintones are.)
Not to say it NEEDS to be like that. I'm not exactly married to the idea & don't feel particularly strongly about it myself. But it's definitely something worth noting.

TheShuEd:
I will remove the Vox update without any problems, they are here primarily as an example of functionality for ease of testing

TheShuEd:
> I'm preeeeetty sure that the exclusive hairstyles vox get are a deliberate creative decision rather then a limitation that just exists by circumstance (much like their limited skintones are.) Tstalker mentioned on discord that they're supposed to canonically have quills instead of hair, which could probably look really weird for a lot of hairstyles. Not to say it NEEDS to be like that. I don't feel particularly strongly about it myself. But it's definitely something worth noting.

Let's do this: I'm leaving the vox changes in for now, so that the pull-request is easier to check.
When the approve is received, I will remove all Vox changes, leaving only the bare system for future use. Okay?

PR #35940, opened 2025-03-19
Show paradox clones in deadchat

slarticodefast:
## About the PR
If a paradox clone dies they now get a `Name (clone)` name modifier for their ghost and when talking in deadchat.

## Why / Balance
Deadchat became confusing when both the original and the clone were talking since they have the same name.
The drawback of this that this could be slightly metagameable by other ghosts in deadchat, but I guess they can already observe everything anyways, so it should be fine.

## Technical details
Added a mind event relay, which can be used to raise events on both the body and the mind.
They work identically to inventory relays and can be useful for other things in the future.
Add a name modifier for observers if they have the paradox clone mind role by subscribing to the `RefreshNameModifiersEvent`.

An alternative way to handle this would be to directly modify the `CharacterName` in the clone's `MindComponent` on spawning, which is used when ghosting to name the ghost. But this datafield is also used in a lot of other cases, for example for target objectives, which would need a major refactor.

## Media
grafik

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Ghosts of paradox clones now have a name modifier so they can be distinguished by other ghosts.

### DIFF
--- a/Content.Server/Ghost/GhostSystem.cs
+++ b/Content.Server/Ghost/GhostSystem.cs
@@ -24,6 +24,7 @@ using Content.Shared.Mobs.Components;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.Movement.Events;
 using Content.Shared.Movement.Systems;
+using Content.Shared.NameModifier.EntitySystems;
 using Content.Shared.Popups;
 using Content.Shared.Storage.Components;
 using Content.Shared.Tag;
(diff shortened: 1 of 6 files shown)


Tayrtahn (approved):
Tested and looks good! The mind role relay system seems like a nice way of handling this and might be useful in the future.

ScarKy0 (approved):
Looks fine

PR #35947, opened 2025-03-20
vox inhands displacements

### LINKED ISSUE #35941: Vox species is missing inhands displacement maps
## Description
Apparently the Vox species has never had any displacement maps for inhands. Right now they currently use the Urist inhands which look weird on the Vox because... they have different hand locations.

They need displacement maps to accompany this change.

**Screenshots**
Compared to a regular Urist:
Image

Image
Image

TheShuEd:
## About the PR

fix #35941 

## Media

image

## Requirements


slarticodefast:
Does this work with cases where the inhand sprites follow the cutout of the urist hand shape?

TheShuEd:
any inhand sprites whose sprite is less than 3 pixels from the sprite border will be clipped. This cannot be fixed with Displacement maps, and such sprites should rather be converted to a 48 by 48 format

TheShuEd:
Oh, that's not what you mean. No, this map just moves the sprites sideways. With displacement maps we can't fill in missing data if it's not in the original sprite. But we do have the ability to draw special sprites for items specifically for individual races (outside of the scope of this pr)

TheShuEd:
treat displacement maps as a fallback for any sprite, which just tries to adapt standard sprites. It won't work for some special cases, but it allows you to make most items look better.

slarticodefast:
We will need to add a bunch of custom sprites for vox for wielded items, and we still have a problem with the hand cutouts which should be masked.
But compared to master this PR is a good improvement!

PR #35950, opened 2025-03-20
Fix spray nozzle not cleaning reagents properly

### LINKED ISSUE #35246: Spray nozzle projectiles only hit puddles on every third tile
## Description

See title. This might be related to game tick speed and how fast the projectiles travel before the game checks collisions.

This makes spray nozzles awkward to use as you have to stand two tiles away from each puddle.

**Reproduction**

1. Load backpack water tank full of space cleaner and grab spray nozzle.
2. Break welding fuel tank.
3. Fire space cleaner spray nozzle at puddles.
4. Repeat on different puddles at different angles to verify range.

**Screenshots**

Image

ArtisticRoomba:
## About the PR
Fixes the spray nozzle / backpack water tank not properly cleaning puddles.

Fixes #35246.

## Why / Balance
It was broken! This makes the spray work a lot better and makes this tech finally appetizing for janitor players.

## Technical details
Old code was very weird, but it was made by Sloth so it was probably made like that for a good reason. In summary:
- Wraps everything into one Update loop.
- Removes the previous fixed timer for only calling for collision checks every 0.125 seconds. This is what caused the previous bug.
- Moved checks around so that cheap checks happen first before expensive ones.
- Added a check so that we only preform solution cleaning logic on one tile only. This can be changed if necessary but I found it helpful to avoid buggy behavior.

At a high level:
- Code iterates over all entities that have a `VaporComponent`
- We pull info about the entity including its current position, solution component, etc.
- For every reagent contained in the solution, we do a tile reaction with the tile we're currently on, by a certain defined percentage. This is customizable by `vaporComp.TransferAmount`.
- If we don't have enough reagent left to complete the next reaction, then we only react with how much we have.
- Remove what we reacted from the solution.
- If there's no more reagents in the shot, delete it.
- Save the previously ran tile for a future comparison (we only want to run our logic for one tile).

## Media
Content Client_k6ojXp8Svh

(Newer addition)
Content Client_ABKoakvr8e

That was a whole welding tank worth of mess cleaned in like 15 seconds!

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Fixed backpack water tanks (spray nozzle) projectiles not properly cleaning up puddles.

### DIFF
--- a/Content.Server/Chemistry/Components/VaporComponent.cs
+++ b/Content.Server/Chemistry/Components/VaporComponent.cs
@@ -1,36 +1,37 @@
 ﻿using Robust.Shared.Map;
 
-namespace Content.Server.Chemistry.Components;
-
-[RegisterComponent]
-public sealed partial class VaporComponent : Component
+namespace Content.Server.Chemistry.Components
 {
-    public const string SolutionName = "vapor";
+    [RegisterComponent]
+    public sealed partial class VaporComponent : Component
+    {
+        public const string SolutionName = "vapor";
 
-    /// <summary>
-    /// Stores data on the previously reacted tile. We only want to do reaction checks once per tile.
-    /// </summary>
-    [DataField]
-    public TileRef? PreviousTileRef;
+        /// <summary>
+        /// Stores data on the previously reacted tile. We only want to do reaction checks once per tile.
+        /// </summary>
+        [DataField]
+        public TileRef? PreviousTileRef;
 
-    /// <summary>
-    /// Percentage of the reagent that is reacted with the TileReaction.
-    /// <example>
-    /// 0.5 = 50% of the reagent is reacted.
-    /// </example>
-    /// </summary>
-    [DataField]
-    public float TransferAmountPercentage;
+        /// <summary>
+        /// Percentage of the reagent that is reacted with the TileReaction.
+        /// <example>
+        /// 0.5 = 50% of the reagent is reacted.
+        /// </example>
+        /// </summary>
+        [DataField]
+        public float TransferAmountPercentage;
 
-    /// <summary>
-    /// The minimum amount of the reagent that will be reacted with the TileReaction.
-    /// We do this to prevent floating point issues. A reagent with a low percentage transfer amount will
-    /// transfer 0.01~ forever and never get deleted.
-    /// <remarks>Defaults to 0.05 if not defined, a good general value.</remarks>
-    /// </summary>
-    [DataField]
-    public float MinimumTransferAmount = 0.05f;
+        /// <summary>
+        /// The minimum amount of the reagent that will be reacted with the TileReaction.
+        /// We do this to prevent floating point issues. A reagent with a low percentage transfer amount will
+        /// transfer 0.01~ forever and never get deleted.
+        /// <remarks>Defaults to 0.05 if not defined, a good general value.</remarks>
+        /// </summary>
+        [DataField]
+        public float MinimumTransferAmount = 0.05f;
 
-    [DataField]
-    public bool Active;
+        [DataField]
+        public bool Active;
+    }
 }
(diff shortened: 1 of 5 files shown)


ArtisticRoomba:
I fixed the file to be a file-scoped namespace in a separate commit. GitHub allows viewing without whitespace changes if necessary.

ScarKy0 (approved):
No issues, saw it working. Neat stuff.

ScarKy0:
Test fail is real, looks like the DataField wasn't supposed to be there after all?

Content.Server/Chemistry/Components/VaporComponent.cs
```diff
@@ -7,11 +8,19 @@ public sealed partial class VaporComponent : Component
     {
         public const string SolutionName = "vapor";
 
-        [DataField("transferAmount")]
-        public FixedPoint2 TransferAmount = FixedPoint2.New(0.5);
+        /// <summary>
+        /// Stores data on the previously reacted tile. We only want to do reaction checks once per tile.
+        /// </summary>
+        public TileRef PreviousTileRef;
```
slarticodefast, line 14:
```suggestion
        public TileRef? PreviousTileRef;
```
should be nullable since it does not have a default value

ArtisticRoomba: pushed

ArtisticRoomba:
Looks like the nukeops test rolled a wizard subantag again... I swear the test is out to get me

TiniestShark:
wait you mean these _arn't_ just supposed to be flamethrowers?

ArtisticRoomba:
I did some more fine-tuning of the values I used and added more logic preventing funky-looking behavior. These changes make the spray nozzle more responsive IMO.

Content Client_ABKoakvr8e

Partmedia:
It's not just about putting whitespace changes in a separate commit. When this PR is merged it will likely be merged as a squash commit, and therefore a bug fix will still come with a ton of formatting changes, making it difficult to cherry pick/merge in downstreams. Please review our [pull request guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).

ArtisticRoomba:
I was explicitly instructed to do this by @ScarKy0 during review in VC with him, I'll revert those whitespace changes tomorrow.

Partmedia:
Let me double check with @ScarKy0 and make sure we're on the same page. Thanks.

Content.Server/Chemistry/EntitySystems/VaporSystem.cs
```diff
@@ -17,128 +17,139 @@
...
+        var query = EntityQueryEnumerator<VaporComponent, SolutionContainerManagerComponent, TransformComponent>();
+        while (query.MoveNext(out var uid, out var vaporComp, out var container, out var xform))
         {
-            if (solution.Volume == 0)
-            {
-                return false;
-            }
+            // Return early if we're not active
+            if (!vaporComp.Active)
+                return;
```
metalgearsloth, line 99: These should be all be continues, copy-paste moment.

Content.Server/Chemistry/EntitySystems/VaporSystem.cs
```diff
@@ -17,128 +17,139 @@
...
-                return false;
-            }
+                var tile = _map.GetTileRef(xform.GridUid.Value, gridComp, xform.Coordinates);
 
-            return _solutionContainerSystem.TryAddSolution(vaporSolution.Value, solution);
-        }
+                // Check if the tile is a tile we've reacted with previously. If so, skip it.
+                // If we have no previous tile reference, we don't return so we can save one.
+                if (vaporComp.PreviousTileRef != null && tile == vaporComp.PreviousTileRef)
+                    return;
```
metalgearsloth, line 109: .

Content.Server/Chemistry/EntitySystems/VaporSystem.cs
ArtisticRoomba: oops

EmoGarbage404:
> Spray nozzle code still bad

hate me cuz you aint me

Issue #35957, opened 2025-03-20
Using multiple t-ray scanners is bugged


slarticodefast:
## Description
Small bug introduced in https://github.com/space-wizards/space-station-14/pull/35347
Should be an easy fix. Just add a datafield to `TrayScannerUserComponent` that is counting how many scanners you are using and only remove the component when that number is 0.

**Reproduction**
- spawn two t-ray scanners and toggle them on
- put one into your pocket
- pick up the other one and drop it again
- you won't see the subfloor anymore even though you still have the second scanner

PR #35959, opened 2025-03-20
Fix TrayScannerUser

metalgearsloth:
Fixes https://github.com/space-wizards/space-station-14/issues/35957

Was indifferent on cl given how soon it was.

### DIFF
--- a/Content.Shared/SubFloor/TrayScannerUserComponent.cs
+++ b/Content.Shared/SubFloor/TrayScannerUserComponent.cs
@@ -5,4 +5,7 @@ namespace Content.Shared.SubFloor;
 /// Added to anyone using <see cref="TrayScannerComponent"/> to handle the vismask changes.
 /// </summary>
 [RegisterComponent]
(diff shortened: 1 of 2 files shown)


Content.Shared/SubFloor/TrayScannerUserComponent.cs
```diff
@@ -5,4 +5,7 @@ namespace Content.Shared.SubFloor;
 /// Added to anyone using <see cref="TrayScannerComponent"/> to handle the vismask changes.
 /// </summary>
 [RegisterComponent]
-public sealed partial class TrayScannerUserComponent : Component;
+public sealed partial class TrayScannerUserComponent : Component
+{
+    public int Count;
```
slarticodefast, line 10:
```suggestion
    /// <summary>
    /// How many t-rays the user is currently using.
    /// </summary>
    [DataField]
    public int Count;
```
Should be a datafield for persistance.

PR #35965, opened 2025-03-20
Validate some SpeciesPrototype fields

Tayrtahn:
## About the PR

Adds YAML Linter validation to some data fields in `SpeciesPrototype`.

## Why / Balance

`SpeciesPrototype` has been around since before we had `ProtoId`, so it uses unvalidated strings to identify some of the prototypes that it references. By converting them to appropriate `ProtoId`s, the linter will make sure that all species prototype definitions use valid prototypes for these fields.

These values are all passed into `PrototypeManager.Index` (as opposed to `TryIndex`) at some point, which means that they will cause runtime errors if invalid values are used. Catching them with the linter is preferable than having to watch out for runtime errors.

## Technical details

`SpriteSet` is now `ProtoId<HumanoidSpeciesBaseSpritesPrototype>`
`MarkingPoints` is now `ProtoId<MarkingPointsPrototype>`
`MaleFirstNames`, `FemaleFirstNames`, and `LastNames` are now `ProtoId<LocalizedDatasetPrototype>`

Places in the code that pass these values into `PrototypeManager` for indexing were modified to remove the explicit type argument, since it's no longer needed. This change isn't strictly necessary, and can be reverted if desired, but it's a bit cleaner.

## Requirements

### DIFF
--- a/Content.Client/Humanoid/HumanoidAppearanceSystem.cs
+++ b/Content.Client/Humanoid/HumanoidAppearanceSystem.cs
@@ -58,7 +58,7 @@ public sealed class HumanoidAppearanceSystem : SharedHumanoidAppearanceSystem
 
         // add default species layers
         var speciesProto = _prototypeManager.Index(component.Species);
-        var baseSprites = _prototypeManager.Index<HumanoidSpeciesBaseSpritesPrototype>(speciesProto.SpriteSet);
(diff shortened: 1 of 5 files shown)


slarticodefast (approved):
Good stuff.

PR #35966, opened 2025-03-20
Antag Rolebans WIP

Errant-4:
## About the PR

Antag bans now actually get recorded, and no longer crash the dev client

When a player takes a ghost role through the GUI, the mind of the ghostrole is checked for any relevant job or antag-roles that should prevent the player from claiming it.

Antags that did not already have a prototype (due to not needing any special rules) have recently been given generic antag prototypes that correspond with their roletype, so any antag can be banned. But for now, some of them can only be banned together. 
For example, all spiders, carps, angry slimes are under GenericTeamAntagonist

It would be possible to create individual prototypes for every single antag, if desired, but that seemed like a lot of complexity that might not even be useful to admins

TO DO, possibly on another PR:
- [ ] Check ghostrole jobs (refugees?)
- [ ] Block roundstart antags
- [ ] Block conversions (Rev, zombie, borg?)
- [ ] Block event activations (Sleeper, Wizard)
- [ ] FREE SPACE

## Why / Balance
Basic feature

## Technical details
BanManager now checks an incoming role prototype against both JobPrototypes and AntagPrototypes. If neither matches, it throws as it did with antagprototypes before. It tries to index as a JobPrototype first, so if there is at any point a job and antag prototype with the same name, it will not be possible to place a roleban on that antag without refactoring the earlier code of how bans are generated.

Like the current job ban entries, antag bans also have their own prefix, so once recorded it's clear what's what

When a player requests a ghost role from the GUI, the server checks any/all mind roles on the prototype to see if any of them have been banned for the player.
 
## Media

https://github.com/user-attachments/assets/f12b3924-c504-42c8-a73f-724a8c3390c5

## Requirements

## Breaking changes
AntagPrototypes can now be rolebanned. However, since the current code does not know what kind of prototype a banned role is, it has to check for both. It checks for jobprototype first, so an antag prototype that's named the same as a job can not be banned (for now?). Make sure that there are no overlapping prototype names.

**Changelog**
:cl: Errant
ADMIN:
- add: Antag bans can now be placed on players through the Roleban GUI. For now, only ghostroles respect the ban, not roundstart, event or converted/hacked antags


spanky-spanky:
YES

Even if it’s just a start.

Errant-4:
Ready for review (for real this time)

Errant-4:
I got the Ghost Roles window tooltip reason to work without breaking the groupings, but it needs engine update https://github.com/space-wizards/RobustToolbox/pull/5772 so I will not add that commit until the engine fix. No reason to hold up this PR for the tooltips though, I will add them when it becomes possible

Screenshot_2025-03-25_101553

Content.Client/Players/PlayTimeTracking/JobRequirementsManager.cs
```diff
@@ -90,33 +92,64 @@ private void RxJobWhitelist(MsgJobWhitelist message)
...
+    {
+        var list = new List<string>{JobPrefix + job.ID};
+        return IsAllowed(list, profile, out reason);
+    }
+
+    /// <summary>
+    /// Check each job/antag prototype against the current player, for requirements and bans
+    /// The prototypes MUST start with the job/antag prefix
+    /// </summary>
+    public bool IsAllowed(List<string> prototypes, HumanoidCharacterProfile? profile, [NotNullWhen(false)] out FormattedMessage? reason)
```
ciaran, line 108: It seems unnecessary for this to just be strings, what about something like

```cs
enum RoleType {
  Job,
  Antag,
}
```

and then you pass a a List of tuples with the job type and the name. 
It could also just be a boolean (`isAntag`) if it's unambiguous enough maybe.

Errant-4: job:<jobProtoId> is the format already used to record and store all existing (job) rolebans, seems reasonable to use it

Errant-4:
With the engine update, the ban/requirement reason is now shown on the ghostrole UI without causing the above mentioned grouping issues.

Content.Shared/Roles/JobRequirements.cs
```diff
@@ -18,6 +18,17 @@ public static bool TryRequirementsMet(
     {
         var sys = entManager.System<SharedRoleSystem>();
         var requirements = sys.GetJobRequirement(job);
+        return TryRequirementsMet(requirements, playTimes, out reason, entManager, protoManager, profile);
+    }
+
+    public static bool TryRequirementsMet(
```
beck-thompson, line 24: Comment

beck-thompson (changes requested):
Looks good! I think the main thing is we should try to minimize the use of the prefixes outside of where it is strictly necessary.

Text File.txt
southbridge-fur, line 1: Delete this random file

Content.Server/Antag/AntagSelectionSystem.cs
```diff
@@ -497,11 +500,35 @@ public AntagSelectionPlayerPool GetPlayerPool(Entity<AntagSelectionComponent> en
...
+            // Add player to the appropriate antag pool
+            if (HasPrimaryAntagPreference(session, def) && !blocked)
             {
                 preferredList.Add(session);
             }
-            else if (HasFallbackAntagPreference(session, def))
+            else if (HasFallbackAntagPreference(session, def) && !fallbackBlocked)
             {
                 fallbackList.Add(session);
             }
```
southbridge-fur: Maybe this is out of scope, but it seems like this entire bit of code could just be changed to a single ordered list which then gets iterated over and returns the first valid preference, or null if none of them are. That way you're not duplicating code and also it's much more free-form for future expandability.

southbridge-fur (changes requested):
Adding my own review here to help possibly revive this PR. All the ones Beck left should also be addressed since they're still valid.

Content.Server/GameTicking/Events/IsRoleAllowedEvent.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Content.Shared.Roles;
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.GameTicking.Events;
+
+[ByRefEvent]
+public struct IsRoleAllowedEvent(ICommonSession player, string roleId, bool cancelled = false)
```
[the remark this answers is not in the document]
Errant-4: IsJobAllowedEvent is triggered on spawn, and runs a check in JobWhiteListManager and in PlayTimeTrackingSystem

IsRoleAllowedEvent only does the latter, since it could be an antag role

It's possible that these two could be integrated better, but this is not my priority right now

Content.Server/Players/PlayTimeTracking/PlayTimeTrackingSystem.cs
```diff
@@ -182,14 +186,35 @@ private void OnIsJobAllowed(ref IsJobAllowedEvent ev)
...
+
+        // Yes, it's a bit ironic that we remove the prefixes right before blindly sorting them into the categories
+        // the prefixes specify, but not every code path that calls this function will send them WITH prefixes
+        if (role.StartsWith(JobPrefix, StringComparison.Ordinal))
+            role = role[JobPrefix.Length..];
+        else if (role.StartsWith(AntagPrefix, StringComparison.Ordinal))
+            role = role[AntagPrefix.Length..];
+
+        var cvar = _cfg.GetCVar(CCVars.GameRoleTimers);
+        _prototypes.TryIndex<JobPrototype>(role, out job);
```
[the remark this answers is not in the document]
Errant-4: This is intentional, I need indexed jobs to end up in one variable, and antags to end up in the other. Added comment to clarify

Content.Client/Players/PlayTimeTracking/JobRequirementsManager.cs
```diff
@@ -90,33 +92,64 @@ private void RxJobWhitelist(MsgJobWhitelist message)
...
+                return false;
+            }
+            JobPrototype? job = null;
+            AntagPrototype? antag = null;
 
-        if (!CheckWhitelist(job, out reason))
-            return false;
+            // Slightly cursed, but the prefixes need to be removed for the next part. And at least we know the types.
+            if (proto.StartsWith(JobPrefix, StringComparison.Ordinal))
+                _prototypes.TryIndex<JobPrototype>(proto[JobPrefix.Length..], out job);
```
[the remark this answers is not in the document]
Errant-4: This is intentional, I need indexed jobs to end up in one variable, and antags to end up in the other. Added comment to clarify

Content.Server/Ghost/Roles/GhostRoleSystem.cs
```diff
@@ -470,6 +493,69 @@ public void Request(ICommonSession player, uint identifier)
         }
     }
 
+    /// <summary>
+    /// Collect all role prototypes on the Ghostrole
+    /// </summary>
+    private List<string> GetPrototypes(Entity<GhostRoleComponent> roleEnt)
```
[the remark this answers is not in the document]
Errant-4: Prefixes were removed from this file. I pass one list without prefixes now, and sort them into antags and jobs where it becomes relevant

Content.Client/Players/PlayTimeTracking/JobRequirementsManager.cs
```diff
@@ -29,6 +28,9 @@ public sealed class JobRequirementsManager : ISharedPlaytimeManager
     private readonly List<string> _roleBans = new();
     private readonly List<string> _jobWhitelists = new();
 
+    public const string JobPrefix = "Job:";
```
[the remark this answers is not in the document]
Errant-4: There are just 2 files using these, now, so I figured it's not a big deal if they have them separately defined.

Content.Server/Administration/Managers/BanManager.cs
```diff
@@ -237,12 +238,14 @@ private void KickForBanDef(ICommonSession player, ServerBanDef def)
...
     public async void CreateRoleBan(NetUserId? target, string? targetUsername, NetUserId? banningAdmin, (IPAddress, int)? addressRange, ImmutableTypedHwid? hwid, string role, uint? minutes, NoteSeverity severity, string reason, DateTimeOffset timeOfBan)
     {
-        if (!_prototypeManager.TryIndex(role, out JobPrototype? _))
+        if (!_prototypeManager.TryIndex<JobPrototype>(role, out var job) && !_prototypeManager.TryIndex<AntagPrototype>(role, out var antag))
         {
             throw new ArgumentException($"Invalid role '{role}'", nameof(role));
         }
 
-        role = string.Concat(JobPrefix, role);
+        //TODO: create an integration test that screams if a jobPrototype and antagPrototype have the same ID
```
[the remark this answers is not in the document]
Errant-4: With the latest changes, this is now a hard requirement

Content.Server/Antag/AntagSelectionSystem.cs
Errant-4: I don't necessarily disagree, but messing with the antag selection logic should be its own PR. The goal here was to introduce the checks without altering anything meaningfully so we don't need to validate antagselectionsystem's various scenarios on top of ban stuff

Errant-4:
The reviews have all been addressed, and should be good, BUT
1. This PR now _REQUIRES_ the integration test to prevent an antag and job prototype from having the same ID
2. A bunch of less-important ghostroles still aren't bannable. This is most likely all just yaml configuration, so while I am digging through it, we might consider this to be something that's enough to fix down the line? It impacts:
- ~~reinforcement Syndicate Assault Cyborg~~
- Syndicate PAI
- ~~NPCs converted into ghostrole zombies~~
- ~~event Giant Spider~~
- ~~other hostile event mobs? Like slimes~~
- ~~Death Squad (probably not really that big an issue, given that they would already be under scrutiny by their admin handler)~~

"Unbannable" ghost roles as above, but technically not even in the scope of this PR, not being antags
- visiting crew 
- Mouse
- Mothroach
- Snails
- PAI
- ~~CBURN~~

Content.Server/Administration/Managers/BanManager.cs
```diff
@@ -43,7 +43,9 @@ public sealed partial class BanManager : IBanManager, IPostInjectInit
     private ISawmill _sawmill = default!;
 
     public const string SawmillId = "admin.bans";
+
     public const string JobPrefix = "Job:";
+    public const string AntagPrefix = "Antag:";
```
beck-thompson: Now that I used this in banpannel.xaml as well just move em to a shared file so everyone can reference it without manually typing it each time

Content.Server/Administration/Managers/BanManager.cs
```diff
@@ -233,16 +235,30 @@ private void KickForBanDef(ICommonSession player, ServerBanDef def)
     #endregion
 
     #region Job Bans
+
     // If you are trying to remove timeOfBan, please don't. It's there because the note system groups role bans by time, reason and banning admin.
     // Removing it will clutter the note list. Please also make sure that department bans are applied to roles with the same DateTimeOffset.
     public async void CreateRoleBan(NetUserId? target, string? targetUsername, NetUserId? banningAdmin, (IPAddress, int)? addressRange, ImmutableTypedHwid? hwid, string role, uint? minutes, NoteSeverity severity, string reason, DateTimeOffset timeOfBan)
```
beck-thompson: Add a comment explaining that `role` has to have the prefix, and an example as well!

Content.Server/Administration/Managers/BanManager.cs
```diff
@@ -233,16 +235,30 @@ private void KickForBanDef(ICommonSession player, ServerBanDef def)
...
     {
-        if (!_prototypeManager.TryIndex(role, out JobPrototype? _))
+        if (!role.StartsWith(JobPrefix) && !role.StartsWith(AntagPrefix))
```
beck-thompson: Could you put this below the other two? It doesn't really matter but it makes my soul feel better 😆

Content.Server/Ghost/Roles/GhostRoleSystem.cs
```diff
@@ -469,6 +489,71 @@ public void Request(ICommonSession player, uint identifier)
...
+                if (comp.JobPrototype is not null)
+                    list.Add(comp.JobPrototype);
+                else if (comp.AntagPrototype is not null)
+                    list.Add(comp.AntagPrototype);
+            }
+
+            return list;
+        }
+
+        if(roleEnt.Comp.JobProto is not null)
```
beck-thompson: Add space after the if

Content.Server/Administration/Managers/BanManager.cs
```diff
@@ -333,6 +349,54 @@ public async Task<string> PardonRoleBan(int banId, NetUserId? unbanningAdmin, Da
...
+        return false;
+    }
+
+    /// <summary>
+    /// Checks if any roles on the list are banned for the specified player.
+    /// </summary>
+    /// <param name="player">The player.</param>
+    /// <param name="prototypes">A list of antag prototype IDs.</param>
+    /// <returns>True if any of the roles are banned.</returns>>
+    public bool IsRoleBanned(ICommonSession player, List<ProtoId<AntagPrototype>> prototypes)
```
beck-thompson: I don't this function works, it adds the prefix but the other overload expects the prefix not to be added

Content.Server/Administration/Managers/IBanManager.cs
```diff
@@ -26,6 +26,22 @@ public interface IBanManager
...
     public void CreateServerBan(NetUserId? target, string? targetUsername, NetUserId? banningAdmin, (IPAddress, int)? addressRange, ImmutableTypedHwid? hwid, uint? minutes, NoteSeverity severity, string reason);
     public HashSet<string>? GetRoleBans(NetUserId playerUserId);
+
+    /// <summary>
+    /// Checks if the player is currently banned from any of the listed roles.
+    /// </summary>
+    /// <param name="player">The player.</param>
+    /// <param name="prototypes">A list of antag and/or job prototype IDs.</param>
+    /// <returns>Returns True if an active role ban is found for this player for any of the listed roles.</returns>
+    public bool IsRoleBanned(ICommonSession player, List<string> prototypes);
```
beck-thompson: Change `prototypes` to `roles` - also make sure to update the comment

Content.Client/Players/PlayTimeTracking/JobRequirementsManager.cs
```diff
@@ -90,33 +92,78 @@ private void RxJobWhitelist(MsgJobWhitelist message)
...
-            reason = FormattedMessage.FromUnformatted(Loc.GetString("role-ban"));
-            return false;
-        }
+            JobPrototype? job = null;
+            AntagPrototype? antag = null;
+
+            // The database stores roles with prefixes to distinguish them, so we must add them to the IDs before comparison
+            var prefixedProto = proto;
+            // Forcing index types for sorting
+            if (_prototypes.TryIndex<JobPrototype>(proto, out job))
```
beck-thompson: Get rid of the `<JobPrototype>` and `<AntagPrototype>` they aren't needed because you defined the types on line 123/124

Content.Server/Zombies/ZombieSystem.Transform.cs
```diff
@@ -51,23 +54,27 @@ namespace Content.Server.Zombies;
...
     [Dependency] private readonly NPCSystem _npc = default!;
     [Dependency] private readonly TagSystem _tag = default!;
-    [Dependency] private readonly NameModifierSystem _nameMod = default!;
     [Dependency] private readonly ISharedPlayerManager _player = default!;
 
     private static readonly ProtoId<TagPrototype> InvalidForGlobalSpawnSpellTag = "InvalidForGlobalSpawnSpell";
     private static readonly ProtoId<TagPrototype> CannotSuicideTag = "CannotSuicide";
     private static readonly ProtoId<NpcFactionPrototype> ZombieFaction = "Zombie";
+    private static readonly string MindRoleZombie = "MindRoleZombie";
+    private static readonly List<string> BannableZombiePrototypes = new List<string> { "Antag:Zombie" };
```
beck-thompson: I think the `Antag:Zombie` will break it, it should just be `Zombie`. Also can't you use the overload that uses a list of antag protoids?

Content.Server/Players/PlayTimeTracking/PlayTimeTrackingSystem.cs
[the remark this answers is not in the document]
beck-thompson: See my comment on `JobRequirementsManager.cs` - also is there any way to combine these two? They look extremely similar

Content.Server/GameTicking/Events/IsRoleAllowedEvent.cs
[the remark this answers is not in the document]
beck-thompson: Add a comments, this gives me sus vibes but I can't quite think of a better way. I'll try to think of a better way to do this but it isn't super obvious

Content.Server/Administration/Managers/BanManager.cs
Errant-4: Added it to IBanManager, from where this inherits the documentation

Content.Server/Administration/Managers/BanManager.cs
Errant-4: The comment is wrong on the other overload... It needs (and gets) prefixed input. So this here is correct

Errant-4: Ahm, akshually

Errant-4: Okay so. If it's not an indexable prototype ID, then it will not be given a prefix _but it will still try it_ against the bans
Thus, if a prefixed ban arrives, it can still trip the ban check

Content.Server/Zombies/ZombieSystem.Transform.cs
Errant-4: Both work, actually. I hate it her

Content.Server/Administration/Managers/BanManager.cs
Errant-4: So, THIS is cursed. But it works :shrug:

Content.Server/Zombies/ZombieSystem.Transform.cs
Errant-4: I could use the overload, too, but. It makes no difference, really, does it?

Content.Server/Administration/Managers/BanManager.cs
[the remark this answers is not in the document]
Errant-4: With the latest latest changes, this is no longer needed

Content.Client/Players/PlayTimeTracking/JobRequirementsManager.cs
```diff
@@ -90,33 +92,105 @@ private void RxJobWhitelist(MsgJobWhitelist message)
...
+        [NotNullWhen(false)] out FormattedMessage? reason)
     {
         reason = null;
 
-        if (_roleBans.Contains($"Job:{job.ID}"))
+        if (antags is not null)
+        {
+            foreach (var proto in antags)
+            {
+                if (_prototypes.TryIndex(proto, out var antag) && !IsAllowed(antag, profile, out reason))
```
beck-thompson: Use Index!

```
                if (!IsAllowed(_prototypes.Index(proto), profile, out reason))
                    return false;
```

beck-thompson: i committed this github troll

Entvari:
oh my god it's happening

Errant-4:
I don't know how, but I broke the Janitor job?

Errant-4:
I did not break the Janitor job and that's all there is to say on the matter.

Content.Server/Antag/AntagSelectionSystem.cs
Errant-4: This part has now been streamlined and just sends the whole list to the systems that do the check

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
```diff
@@ -297,12 +300,12 @@ private Button CreateRoleGroupHeader(string groupName, BoxContainer header, Colo
     /// Adds a checkbutton specifically for one "role" in a "group"
     /// E.g. it would add the Chief Medical Officer "role" into the "Medical" group.
     /// </summary>
-    private void AddRoleCheckbox(string group, string role, GridContainer roleGroupInnerContainer, Button roleGroupCheckbox)
+    private void AddRoleCheckbox(string group, string prefix, string role, GridContainer roleGroupInnerContainer, Button roleGroupCheckbox)
```
slarticodefast: The `role` string being used for both a `JobPrototype` and an `AntagPrototype` seems really hacky to me. This should be using ProtoIds so it's validated. Maybe use two nullable parameters so that the method can still be used with both. If you split them up you won't need the prefixes either, which seem like a workaround.

Content.Server/Zombies/ZombieSystem.Transform.cs
```diff
@@ -103,6 +111,24 @@ public void ZombifyEntity(EntityUid target, MobStateComponent? mobState = null)
...
+            var sess = actor.PlayerSession;
+            var message = Loc.GetString("zombie-roleban-ghosted");
+
+            if (_mind.TryGetMind(sess, out var playerMindEnt, out var playerMind))
+            {
+                // Detach
+                _ghost.SpawnGhost((playerMindEnt, playerMind), target);
+
+                // Notify
+                _chatMan.ChatMessageToOne(ChatChannel.Local, message, message, EntityUid.Invalid, false, sess.Channel);
```
SlamBamActionman: Use `DispatchServerMessage` instead. It doesn't do much different; just sets the channel to `ChatChannel.Server` and wraps the message in a bold tag, but for all intents and purposes it'd be correct here.

Content.Server/Zombies/ZombieSystem.Transform.cs
```diff
@@ -103,6 +111,24 @@ public void ZombifyEntity(EntityUid target, MobStateComponent? mobState = null)
...
+            }
+            else
+                Log.Error($"Where is my mind, where is my mind");
```
SlamBamActionman: Jokes in comments, proper errors in the logs!

Content.Server/Ghost/Roles/GhostRoleSystem.cs
```diff
@@ -469,6 +490,78 @@ public void Request(ICommonSession player, uint identifier)
...
+
+        if (roleEnt.Comp.JobProto is not null)
+            jobs.Add(roleEnt.Comp.JobProto.Value);
+
+
+        // If there is no mind, check the mindRole prototypes
+        foreach (var proto in roleEnt.Comp.MindRoles)
+        {
+            if (!_prototype.TryIndex(proto, out var indexed)
+                || !indexed.Components.TryGetComponent("MindRole", out var comp))
```
SlamBamActionman: Considering this is used in very few instances in the codebase, I am highlighting this as "sus". It might very well be okay here but I'd like a second opinion, especially since it uses a magic string.

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
```diff
@@ -34,6 +34,9 @@ public sealed partial class BanPanel : DefaultWindow
     private bool HasBanFlag { get; set; }
     private TimeSpan? ButtonResetOn { get; set; }
 
+    private const string PrefixAntag = SharedRoleSystem.RolePrefixAntag;
+    private const string PrefixJob = SharedRoleSystem.RolePrefixJob;
```
SlamBamActionman: My base assumption would be to just use `SharedRoleSystem.RolePrefixJob` directly, but maybe instantiating it like this when it's a private const is fine? Need second opinion.

Content.Client/Players/PlayTimeTracking/JobRequirementsManager.cs
```diff
@@ -90,33 +92,90 @@ private void RxJobWhitelist(MsgJobWhitelist message)
         Updated?.Invoke();
     }
 
-    public bool IsAllowed(JobPrototype job, HumanoidCharacterProfile? profile, [NotNullWhen(false)] out FormattedMessage? reason)
+    public bool IsAllowed(
```
SlamBamActionman, line 102: Documentation

Content.Server/Antag/AntagSelectionSystem.cs
```diff
@@ -497,11 +500,26 @@ public AntagSelectionPlayerPool GetPlayerPool(Entity<AntagSelectionComponent> en
...
 
-            if (HasPrimaryAntagPreference(session, def))
+            var blocked = false;
+            var fallbackBlocked = false;
+
+            // Check for role bans and playtime requirements
+            // These are not in IsSessionValid so that we can have a Preferred role ban not block a Fallback role
+
+            if (_ban.IsRoleBanned(session, def.PrefRoles)
+                || !_playTime.IsAllowed(session, def.PrefRoles))
```
SlamBamActionman: This seems like it would limit you if you fail the check for any of the roles? Which for banning might be fine (if `AntagAB` can be rolled with either `RoleA` and `RoleB`, banning one role should probably ban you from `AntagAB`), but I am not so sure for playtime? Especially for fallbacks, where e.g. you now can't roll `NukeOpsCommander` as a fallback if you haven't unlocked `NukeOpsMedic`

If we want an antag to be limited by multiple roletimes, it would be easier to integrate those times into the roles directly rather than force this behavior here.

SlamBamActionman (changes requested):
Part of me is a bit worried we'll end up being pigeonholed into only considering jobs and antag roles, but perhaps that's fine (e.g. something like the Tourist role is intended to be internally classified as an "antag" anyhow).

I also don't love that we're saving the roles with a prefix in the database but I can also see that it's something we already are doing with the jobs so maybe it's acceptable... Just a gut feeling, databases isn't my jam so seek input from someone more knowledgeable.

FairlySadPanda:
Slam: the lack of flexibility about Job and Antag is core game domain stuff, it's not anything that Errant can work around as far as I can see. From my own exploration of the code I think it's tech debt: look at my draft PR to see how I've had to use a "job" prototype to assign logic to ghost roles that clearly are not employed in any profession.

This is an exposed need for a refactor; you can see my madness about trying to even resolve the poor namespacing in a closed PR I have from a couple of months back.

Content.Server/Antag/AntagSelectionSystem.cs
Errant-4: Good catch, unfortunately it also really screws me over because I think the only way to resolve this is to further refactor how that entire codepath works, beyond what I touched

SlamBamActionman: Seeing as checking for roletime wasn't previously here I would assume that it is checked elsewhere; if you wanna play it safe and reduce the work you gotta do, just remove the playtime check and add a `//TODO: Validate playtime requirements`, it's fine if it's out of scope.

Errant-4: But the issue also exists for bans, at least theoretically. Practically, that probably does not matter, but it's still _there_. If someone is banned from nukie commander, they cannot fallback as nukie medic either. Shrimply don't get rolebanned, I suppose.

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
Errant-4: Not gonna lie, the _primary_ reason this is organized like that, is probably because it was changed from using manual strings from an earlier implementation. So effectively, inertia. Moving SharedRoleSystem.RolePrefixJob directly into the place where its used would work. 
However, I think it's a bit more neat like this, especially in systems that use these several times, like BanManager. And if I'm going to call it like this there, why not stay consistent in other files

Content.Server/Antag/AntagSelectionSystem.cs
SlamBamActionman: Yeah I suppose.

Also, just noticed, if you're banned from Nukie Commander but not Medic (somehow), you can turn off Commander in your preferences and hope you roll it as a fallback for Medic.

It's probably better to block the antag completely regardless of fallback if you're banned from a primary PrefRoles role. I.e. make `blocked` true just return early.

Errant-4: The main issue is that the original antagselection code with fallbackroles etc was not made with bans and requirements in mind, seemingly.

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
SlamBamActionman: Makes sense, I just don't know personally if it's OK from a code standards pov (hence the second opinion). Idk ask Slart?

Content.Server/Antag/AntagSelectionSystem.cs
Errant-4: If we select the person and then block them later down the line, then effectively that antag slot will be "wasted", no?

SlamBamActionman: Yes; excluding them from the selection pool is the way to properly handle antag slots rolling correctly.

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
Errant-4: @slarticodefast what do you think?

Content.Server/Antag/AntagSelectionSystem.cs
Errant-4: For the record, literally the only use of fallbackRoles is the nuke ops team, and for some reason TraitorSleeper, which has Traitor as a fallback. No idea why that one is necessary

I think I also might have just accidentally made someone immune to DummyAntag if they are banned from one antag role but not all of them...
image

Errant-4: Okay, I got it. It now checks each role individually, and puts the character in the Primary/Fallback pools if any role is both allowed, not banned, and passes any requirements if they exist. This should match the current live behaviour when no requirements/bans are involved
(As a bonus, dupe codepaths have been merged in the API)

Content.Server/Ghost/Roles/GhostRoleSystem.cs
slarticodefast: Pretty sure there is another overload of this that uses the component type instead of a magic string.

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
slarticodefast: Probably doesn't really matter.
The bigger concern I have is that we have to use the prefixes at all, that kinda feels like ProtoIds, but worse.
Prototypes being passed around as raw strings instead of ProtoIds has lead to a lot of bugs or coder errors in the past, because it's simply not validated at all.

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
Errant-4: Okay so I took out the strings from the input, but it still has to end up as a prefixed string, when it goes into the Button, because there is no other way to store it. 
Screenshot_2025-09-09_122517

Should we just... create a descendant of Button, with datafields for the ProtoIDs? For this one interface? If we don't, I don't think that there is that much point to this rewrite

Content.Server/Ghost/Roles/GhostRoleSystem.cs
Errant-4: I think any alternatives only work on an entity? But I only have an indexed prototype to work with

FairlySadPanda:
I took my lunch break today to dive into the backing code for this work.

There are two clear blockers on development on this system:

1. JobPrototype and AntagPrototype are badly constructed and cause a huge amount of copy-pasted code.
2. The use of "Job:" as a prefix in the database layer to try and flag if a particular role ban is a job ban or not.

Both of these are blockers in my opinion and I would consider the latter to be a very high-priority fix to introduce, as it is clearly not the best way to handle enriching an ID with additional context in the context of a persistent database.

I would recommend freezing work on this system and fixing the underpinnings, as this is causing the code smells that the maintainers who have reviewed this have noticed with the use of prefixes.

slarticodefast:
> The use of "Job:" as a prefix in the database layer to try and flag if a particular role ban is a job ban or not.

I would say it's ok to have them like this in the database and ban manager layer for now, but that problem should not leak into the UI code or other systems. So in terms of what to do for this PR: This is not a blocker for the feature, but I would prefer to see the UI code for the ban window being cleaned up in that regard, which hopefully should not be a large refactor.

And for future reference, relevant PJB comment from discord

grafik

FairlySadPanda:
I have done a pass on background work that impacts this PR on https://github.com/space-wizards/space-station-14/pull/40250

Princess-Cheeseballs:
Database migrations shouldn't be a concern. I'll leave it to the lead maints as to whether or not we should hold off on merging.

FairlySadPanda:
I have resolved @juliangiebel 's comment via this PR, as it's not fair to lumber Errant with trying to integrate against my code https://github.com/Errant-4/space-station-14/pull/19

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
Errant-4: This was removed, now BanManager is the only file that has (access to the) prefixes

Errant-4:
Interesting, Panda and I both tested the branch before merging, and it worked. But now it does not - the button does not bring up Confirm if antag roles are selected. 
Presumably I made some mistake when merging, will look into it tomorrow

Errant-4:
It seems that in the previous tests, both of us have only banned both antags and jobs, never antags alone. So the fail condition was not reached

Content.Server/Administration/Managers/BanManager.cs
```diff
@@ -232,22 +210,45 @@ private void KickForBanDef(ICommonSession player, ServerBanDef def)
...
     {
-        if (!_prototypeManager.TryIndex(role, out JobPrototype? _))
+        string encodedRole;
+
+        // TODO: Note that it's possible to clash IDs here between a job and an antag. The refactor that introduced
+        // this check has consciously avoided refactoring Job and Antag prototype.
+        // Refactor Job- and Antag- Prototype to introduce a common RolePrototype, which will fix this possible clash.
+
+        // Don't trust the input: make sure the job or antag actually exists.
+        if (_prototypeManager.HasIndex<JobPrototype>(role))
```
Errant-4, line 244: Since we once again are in a state where the input type here is ambiguous, should we perhaps add a check in case a prototype exists both as a job as an antag? Because that conflict can now be an issue here - the code will just _assume_ that the ban is about the job. Should we make that situation throw an error, to warn an offending codebase?
Orks and I previously talked about adding an integration test for this, but it became unnecessary after the change to pass the prefix from BanPanel.

This will stop being an issue if/when RolePrototype is added, but what about until then?

FairlySadPanda: An integration test seems pretty easy to write to assert this, but considering YAML uploading is possible, it would still be possible to break this at runtime. I think the cleanest solution is actually for IBanManager to have a specific function for banning an antag and a specific function for banning a job. Because we're shipping around ProtoId<T>s that is a bit hard to write outright generically assuming you don't want to use typeof(T).

Content.Client/Administration/UI/BanPanel/BanPanel.xaml.cs
FairlySadPanda: This is resolved.

FairlySadPanda:
Only remaining point on this is Slam's point about GhostRoleSystem.cs:535

Content.Server/Ghost/Roles/GhostRoleSystem.cs
Errant-4: ding dong the witch is dead

Errant-4:
That is also solved, now

SlamBamActionman (approved):
It seems to me that all clean-up since my last review is good; I do agree with the remaining comment that there are some issues where jobs/antag prototypes share names, and we *should* have some check for that (since you're already checking for prototypes at that location, maybe check for both at the same time and raise a warning if it hits true on both?).

Otherwise, it seems to check out. Good job!

Content.Server/Administration/Managers/BanManager.cs
Errant-4: I went with a temporaryTM simple check which we surely will not still have in 5 years

juliangiebel (approved):

Issue #34505, opened 2025-01-18
Abstract particle accelerator construction so we can use it for any big machine


juliangiebel:
## Description

Currently the station anchor gets constructed from a single machine frame which is very weird considering it's a 3x3 machine.
It should ideally be constructed in a similar way to the PA.

Issue #18451, opened 2023-07-30
Particle Accelerator code is really bad


metalgearsloth:
I get it's the trend of singulo code being bad but:

1. It should be a generic multi-part machine code. Combine it with nodes or something if you really want.
2. It subscribes to MoveEvents.
3. It can crash the server.

tetonis:
It's so bad that you can't even construct it anymore, it fails to switch construction graphs after each part is assembled so it becomes a useless machine / computer. I suspect #17960 broke it because it relied on the hacky code that was removed.

graevy:
i think it needs to be fully rebuilt. i can't commit to self-assigning this until after the holidays but i'm going to try to wrap my head around it during

SlamBamActionman:
The linked PR fixes the first issue, I skimmed and couldn't see any MoveEvents, and the rest of the issue is vague enough to be unactionable.

PR #35969, opened 2025-03-20
Add multipart machines

BarryNorfolk:
## About the PR
Add a component/entity system for handling multipart machines in a more generalized way.

## Why / Balance
Multipart machines are cool, and it'd be nice to see more than one in the game.

## Technical details
Tried a simple and probably naive approach to finding parts and make it agnostic to whatever part people build first.
At the moment, if you attempt to build everything BUT the fuel chamber then the UI will show that there are no matching entities, which is a bit frustrating as a new player (I can imagine).

So here it's just a simple look in four directions and attempt to find any anchored entities that match our required offset, rotation, and expected components. After that it's up to the PA system to determine if the emitters are also in the right place.

## Media

Tested this orientations and they all worked/fired, mirrored orientations do not work however. So placing the control box on the other side doesn't pick up the entities.
image

Also, examining an entity with the multipart machine will try and show you where to place the specific machine components based on the sprites specified in the yaml.

Default rotations will likely be the norm:
image

Until a user places at lease one entity and causes a rescan, which means we store the orientation of the machine and can display updated ghosts:
image

Also added the ability to have optional parts. So in this example, I made the end cap of the PA an optional part.
image

Here's a video showing the event based updating of the PA systems.
[MultipartMachine - ItWorks.webm](https://github.com/user-attachments/assets/3d9bbe6b-c629-4914-b872-c5405a2e07c1)

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: New component and system for creating more multipart machines!
- tweak: Converted the Particle Accelerator to use new multipart machine system.
- add: Examining a Particle Accelerator now shows ghosts of what and where to place the exact parts.


ArtisticRoomba:
Extremely good stuff, this will allow much more creative future machines. It's actually a component in a lot of proposals I plan to write in the future.

But what it applies to right now is being able to make large machines like station anchors and station gravity generators constructible, which is good for SS14's dynamic environment.

ArtisticRoomba:
This will also close (or at least work towards) issues https://github.com/space-wizards/space-station-14/issues/34505, https://github.com/space-wizards/space-station-14/issues/18451.

metalgearsloth:
OH god it's so good.

Killerqu00 (approved):
the system behind this looks very good, me likey

Pronana:
image
image

Found an issue(?)
The new system doesn't seem to care about the rotation of the PA computer which is nice, but it also has some weird behavior with predicting a valid multipart machine. Nothing game-breaking just a bit weird.

As posted in the screenshot, regardless of orientation it's detecting the starboard machine part as a port machine part and declaring it valid, regardless of how the computer is rotated and regardless of the rotation of the fuel chamber next to the computer.

Also the computer currently wants to be on the right side of the Particle Accelerator which breaks sprites as the wires connect to the left side of the Particle Accelerator. 

I like the idea of the computer not caring as much about how its rotated because it always faces forward and can be confusing but it feels like it's not working properly.

Pronana:
image
Actually it appears to only check one direction regardless of the rotation of the PA console. This should be valid but it isn't.

Pronana:
Unless I'm dumb or missed something, I left a review that should fix the issues I found. If you have questions feel free to ask.

TiniestShark:
Oh the ghost showing how it's supposed to line up is just _perfect_. That's gonna be insanely helpful.

BarryNorfolk:
> image Actually it appears to only check one direction regardless of the rotation of the PA console. This should be valid but it isn't.
> 
> Edit to avoid duplicating comments, you can also spam click inspect to duplicate the ghost sprties
> 
> Edit 2: After further testing and reading the code it appears that the issue is that the machine never actually removes the rotation data it stores if it finds even one valid machine part. This means if it finds a part, even if that part is out of place it will be locked to that specific rotation basically forever or until it is deconstructed. Except you can't deconstruct the computer I believe.

You are likely right and I missed it when I quickly added the networked orientation for the ghost sprites. Thanks for looking into it and I appreciate the testing :)

Pronana (changes requested):
Just some yaml changes needed, otherwise looks fine to me.

BarryNorfolk:
> Just some yaml changes needed, otherwise looks fine to me.

Sure, is this so that it just faces the other direction, yeah?

Pronana:
> Sure, is this so that it just faces the other direction, yeah?

Yep so it lines up with the original sprites. In its current iteration the wires in the power box/fuel chamber (I mix up the names forgive me) don't connect to the computer, this flips the side the PA is built on so that those wires connect and so it also matches the sprite in the PA computer's UI.

Pronana (approved):
Looks good to me. Might be worth refactoring the PA logic in the future but that's out of scope for this PR.

Content.Client/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,83 @@
...
+
+/// <summary>
+/// Client side handling of multipart machines.
+/// Handles client side examination events to show the expected layout of the machine
+/// based on the origin of the main entity.
+/// </summary>
+public sealed class MultipartMachineSystem : EntitySystem
+{
+    [ValidatePrototypeId<EntityPrototype>]
+    private const string GhostPrototype = "MultipartMachineGhost";
```
Fildrance, line 18: Shouldnt it be ProtoId<EntityPrototype>?

BarryNorfolk: I’ve no idea, I can take a look at others but this mostly just stolen from the TEG ghosts which could be wrong or old.

BarryNorfolk: Should be fixed

BarryNorfolk:
I rebased like a nerd, oops.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.Parts.cs
```diff
@@ -16,6 +17,25 @@ private void InitializePartSystem()
         SubscribeLocalEvent<ParticleAcceleratorPartComponent, PhysicsBodyTypeChangedEvent>(BodyTypeChanged);
     }
 
+    public void ValidateEmitter(string name,
+        ParticleAcceleratorEmitterType type,
```
Fildrance, line 21: i understand that this is an attempt to faithfully recreate previous code after refactor, but do we really need 3 separate parts for emitters and check their position here? maybe we can dump it and just check it all in scanning for construction complete with some more generic solution? (i saw check for orientation in yaml so thats kinda solved, right?)

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.Parts.cs
```diff
@@ -24,137 +44,43 @@ public void RescanParts(EntityUid uid, EntityUid? user = null, ParticleAccelerat
...
+        foreach (var part in machine.Comp.Parts.Values)
         {
-            if (compQuery.TryGetComponent(entity, out comp)
-            && TryComp<ParticleAcceleratorPartComponent>(entity, out var partState) && partState.Master == null
-            && (rotation == null || Transform(entity).LocalRotation.EqualsApprox(rotation!.Value.Theta)))
-            {
-                part = entity;
-                return true;
-            }
+            if (partQuery.TryGetComponent(GetEntity(part.Entity), out var partData))
```
Fildrance, line 78: my educated guess is that it could be nice to have 'core' part in every multipart machine and move this code into multipart machine system - so parts could always know thier master just in case.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -244,30 +256,35 @@ public void UpdateUI(EntityUid uid, ParticleAcceleratorControlBoxComponent? comp
...
-            comp.InterfaceDisabled,
-            comp.MaxStrength,
-            comp.StrengthLocked
-        ));
+                comp.Assembled,
+                comp.Enabled,
+                comp.SelectedStrength,
+                (int)draw,
+                (int)receive,
+                _multipartMachine.GetPartEntity(machine, "StarboardEmitter") != null,
```
Fildrance, line 278: looks like HasPart would be nice addition.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -244,30 +256,35 @@ public void UpdateUI(EntityUid uid, ParticleAcceleratorControlBoxComponent? comp
...
+        if (TryComp<PowerConsumerComponent>(_multipartMachine.GetPartEntity(uid, "PowerBox"), out var powerConsumer))
         {
             draw = powerConsumer.DrawRate;
             receive = powerConsumer.ReceivedPower;
         }
 
+        if (!TryComp<MultipartMachineComponent>(uid, out var machineComp))
+            return;
+
+        var machine = new Entity<MultipartMachineComponent?>(uid, machineComp);
```
Fildrance, line 268: just 
```var machine = (uid, machineCompe);```

 should be nuff, right?

File: Content.Client/Machines/EntitySystems/MultipartMachineSystem.cs [hunk 6]
```diff
@@ -0,0 +1,82 @@
...
+        if (ent.Comp.Ghosts.Count != 0)
+        {
+            // Already showing some part ghosts
+            return;
+        }
+
+        var xformQuery = GetEntityQuery<TransformComponent>();
+        foreach (var part in ent.Comp.Parts.Values)
+        {
+            if (!part.Entity.HasValue)
```
Fildrance, line 46: prolly better to invert if, use early return here to reduce code nesting

```
if (part.Entity.HasValue)
    continue;
```

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.Parts.cs
```diff
@@ -16,6 +17,25 @@ private void InitializePartSystem()
         SubscribeLocalEvent<ParticleAcceleratorPartComponent, PhysicsBodyTypeChangedEvent>(BodyTypeChanged);
```
Fildrance, line 17: maybe its better to leave handling of such events to MultipartMachineSystem? we can expect that most of them would require handling it in some way?

Fildrance: I'd say that those events are kinda what will lead to 'complete' and 'disassembled' events that this system is expected to emit

BarryNorfolk:
Still looking to respond to the comments from you @Fildrance, just been a long two weeks.
I've looked into the push mechanism and while I can definitely listen for anchor events on constructionComponents, there's no good concept of a machine being "finished" that would allow me to listen in. All I can really do is listen for `ConstructionChangeEntityEvent` and rescan to see if there's a valid entity (Might be timing based though).

Fildrance:
Listening and rescanning is the thing i was expecting this thing to do, to be honest - so other systems can not listen and just sub for 'complete' or 'destroyed / non-functional'. That will greatly simplify connected systems code and logic, i feel.

Content.Shared/Machines/Components/MultipartMachine.cs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class MultipartMachineComponent : Component
+{
+    /// <summary>
+    /// Dictionary of unique names to specific parts of this machine
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public Dictionary<string, MachinePart> Parts = [];
```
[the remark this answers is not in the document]
BarryNorfolk: Done. Had to move the enum/enum values to a shared folder, else the client gets a bit sad when trying to parse the yaml for the control box. Good to know.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,232 @@
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /// <summary>
+    /// Performs a rescan of all parts of the machine to confirm they exist and match
+    /// the specified requirments for offset, rotation, and components.
```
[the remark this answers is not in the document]
BarryNorfolk: Fixed.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
BarryNorfolk: Added

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
BarryNorfolk: Neat, in most places yes it seems!

Content.Client/Machines/EntitySystems/MultipartMachineSystem.cs
BarryNorfolk: Good catch :)

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.Parts.cs
BarryNorfolk: Added this, so now there's a new MultipartMachinePartComponent that allows entities to do other things and know what their "Master" is.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.Parts.cs
BarryNorfolk: We can use the graph/node I added in one of your previous comments which should solve this :)

BarryNorfolk:
Think I've fixed or changed everything you mentioned @Fildrance.
Rebased due to merge conflicts but should be good now. Added a webm video showcasing it but I don't know how to properly embed it....

Content.Server/Construction/ConstructionSystem.Graph.cs
```diff
@@ -453,4 +460,19 @@ public ConstructionChangeEntityEvent(EntityUid newUid, EntityUid oldUid)
...
     }
+
+    /// <summary>
+    /// This event is raised after an entity changes prototype/uid during construction.
+    /// This is only raised at the new entity, after it has been initialized.
+    /// </summary>
+    /// <param name="graph">Construction graph for this entity</param>
+    /// <param name="previousNode">Previous node that was active on the graph</param>
+    /// <param name="currentNode">New node that has become active</param>
+    public sealed class AfterConstructionChangeEntityEvent(string graph, string currentNode, string? previousNode)
```
Fildrance, line 471: record struct, no parent, ByRefEventAttribute, use primary constructor. 

Also having so many strings is a bit sad, but no good idea what to do with it - maybe some of them are actually ProtoEntId? eh?

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -40,14 +44,16 @@ public override void Update(float frameTime)
...
+    private void EverythingIsWellToFire(ParticleAcceleratorControlBoxComponent controller,
+        Entity<MultipartMachineComponent> machine)
     {
         DebugTools.Assert(controller.Powered);
         DebugTools.Assert(controller.SelectedStrength != ParticleAcceleratorPowerState.Standby);
-        DebugTools.Assert(controller.Assembled);
-        DebugTools.Assert(EntityManager.EntityExists(controller.PortEmitter));
-        DebugTools.Assert(EntityManager.EntityExists(controller.ForeEmitter));
-        DebugTools.Assert(EntityManager.EntityExists(controller.StarboardEmitter));
+        DebugTools.Assert(machine.Comp.Assembled);
```
Fildrance, line 52: shouldn't just checking for Assembled be enough? why check parts separately?

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -58,20 +64,25 @@ public void Fire(EntityUid uid, TimeSpan curTime, ParticleAcceleratorControlBoxC
         comp.LastFire = curTime;
         comp.NextFire = curTime + comp.ChargeTime;
 
-        EverythingIsWellToFire(comp);
+        if (!TryComp<MultipartMachineComponent>(uid, out var machineComp))
```
Fildrance, line 67: trycomp is repeated before each EverythingIsWellToFire call - just put it inside?

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -82,8 +93,9 @@ public void SwitchOn(EntityUid uid, EntityUid? user = null, ParticleAcceleratorC
         comp.Enabled = true;
         UpdatePowerDraw(uid, comp);
 
-        if (!TryComp<PowerConsumerComponent>(comp.PowerBox, out var powerConsumer)
-        || powerConsumer.ReceivedPower >= powerConsumer.DrawRate * ParticleAcceleratorControlBoxComponent.RequiredPowerRatio)
+        if (!TryComp<PowerConsumerComponent>(_multipartMachine.GetPartEntity(uid, AcceleratorParts.PowerBox), out var powerConsumer)
+            || powerConsumer.ReceivedPower >=
+            powerConsumer.DrawRate * ParticleAcceleratorControlBoxComponent.RequiredPowerRatio)
             PowerOn(uid, comp);
```
Fildrance, line 99: please add curly braces here, as it melts into multiline if statement a bit - hard to read.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -82,8 +93,9 @@ public void SwitchOn(EntityUid uid, EntityUid? user = null, ParticleAcceleratorC
         comp.Enabled = true;
         UpdatePowerDraw(uid, comp);
 
-        if (!TryComp<PowerConsumerComponent>(comp.PowerBox, out var powerConsumer)
-        || powerConsumer.ReceivedPower >= powerConsumer.DrawRate * ParticleAcceleratorControlBoxComponent.RequiredPowerRatio)
+        if (!TryComp<PowerConsumerComponent>(_multipartMachine.GetPartEntity(uid, AcceleratorParts.PowerBox), out var powerConsumer)
+            || powerConsumer.ReceivedPower >=
+            powerConsumer.DrawRate * ParticleAcceleratorControlBoxComponent.RequiredPowerRatio)
```
Fildrance, line 98: join it with previous line, please

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -223,7 +238,7 @@ private void UpdatePowerDraw(EntityUid uid, ParticleAcceleratorControlBoxCompone
     {
         if (!Resolve(uid, ref comp))
             return;
```
Fildrance, line 240: empty line after if block

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -244,30 +259,35 @@ public void UpdateUI(EntityUid uid, ParticleAcceleratorControlBoxComponent? comp
...
         }
 
+        if (!TryComp<MultipartMachineComponent>(uid, out var machineComp))
+            return;
+
+        var machine = (uid, machineComp);
+
         _uiSystem.SetUiState(uid,
             ParticleAcceleratorControlBoxUiKey.Key,
             new ParticleAcceleratorUIState(
```
Fildrance, line 275: extract variable pls

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -244,30 +259,35 @@ public void UpdateUI(EntityUid uid, ParticleAcceleratorControlBoxComponent? comp
...
-            comp.InterfaceDisabled,
-            comp.MaxStrength,
-            comp.StrengthLocked
-        ));
+                machineComp.Assembled,
+                comp.Enabled,
+                comp.SelectedStrength,
+                (int)draw,
+                (int)receive,
+                _multipartMachine.HasPart(machine, AcceleratorParts.StarboardEmitter),
```
Fildrance, line 281: this could be resolved on the client, as MultipartMachineComponent is networked with all of parts dictionary

maybe it should be?... coz this State is a phat thingy. Maybe needs to be consulted with Emo or Sloth :<

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -292,39 +312,25 @@ private void UpdatePartVisualStates(EntityUid uid, ParticleAcceleratorControlBox
...
-            _appearanceSystem.SetData(controller.FuelChamber!.Value, ParticleAcceleratorVisuals.VisualState, state);
-        if (controller.PowerBox.HasValue)
-            _appearanceSystem.SetData(controller.PowerBox!.Value, ParticleAcceleratorVisuals.VisualState, state);
-        if (controller.PortEmitter.HasValue)
-            _appearanceSystem.SetData(controller.PortEmitter!.Value, ParticleAcceleratorVisuals.VisualState, state);
-        if (controller.ForeEmitter.HasValue)
-            _appearanceSystem.SetData(controller.ForeEmitter!.Value, ParticleAcceleratorVisuals.VisualState, state);
-        if (controller.StarboardEmitter.HasValue)
-            _appearanceSystem.SetData(controller.StarboardEmitter!.Value, ParticleAcceleratorVisuals.VisualState, state);
+        if (_multipartMachine.TryGetPartEntity(machine, AcceleratorParts.FuelChamber, out var fuelChamber) && fuelChamber.HasValue)
```
Fildrance, line 321: isnt it fixed by multipart code? there is some sprite-related stuff there! :o

Content.Shared/Machines/Components/MultipartMachine.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// Expected node for this part to be in, on the graph.
+    /// Used to determine when a construct-able object has been
+    /// assembled or disassembled.
+    /// </summary>
+    [DataField]
+    public string ExpectedNode;
+}
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class MultipartMachineComponent : Component
```
Fildrance, line 68: file should be named after this component, i think.

Fildrance, line 68: this should restrict access by its System - add AccessAttribute(typeof(SharedMultipartMachineSystem))

u need to add SharedMultipartMachineSystem for that, so client and server could interact. The idea is that components HAVE to use public fields for seriazliation and interaction with its own system, but for INCAPSULATION it should restrict interaction with its public fields for other systems (so PA should not have access or read its fields, it should ask MultipartMachineSystem for info, or listen to its events)

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -336,14 +342,40 @@ private void OnComponentShutdown(EntityUid uid, ParticleAcceleratorControlBoxCom
         if (TryComp<ParticleAcceleratorPartComponent>(uid, out var partStatus))
             partStatus.Master = null;
 
+        if (!TryComp<MultipartMachineComponent>(uid, out var machine))
+            return;
+
         var partQuery = GetEntityQuery<ParticleAcceleratorPartComponent>();
-        foreach (var part in AllParts(uid, comp))
+        foreach (var part in machine.Parts.Values)
```
Fildrance, line 349: this system should not iterate over parts :< this code probably should be method for MultipartMachineSystem?

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
```diff
@@ -397,9 +429,11 @@ private void OnUIRescanMessage(EntityUid uid, ParticleAcceleratorControlBoxCompo
         if (TryComp<ApcPowerReceiverComponent>(uid, out var apcPower) && !apcPower.Powered)
             return;
 
-        RescanParts(uid, msg.Actor, comp);
+        if (!TryComp<MultipartMachineComponent>(uid, out var machineComp))
+            return;
 
-        UpdateUI(uid, comp);
+        var machine = new Entity<MultipartMachineComponent>(uid, machineComp);
+        _multipartMachine.Rescan(machine, msg.Actor); // Raises an event if state has changed
```
Fildrance, line 436: This is not a nice hack :( mainly because i dont see how this should raise event - construction not changed, anchored is not changed - it should not raise event just in case Rescan was called. It is not obvious why is this connected to to OnMachineAssembledChanged, this is a hard-to-understand hack, better link this code directly.

Fildrance (changes requested):
Code is nice, but we can do better!

Content.Server/Construction/ConstructionSystem.Graph.cs
```diff
@@ -407,6 +409,11 @@ public bool ChangeNode(EntityUid uid, EntityUid? userUid, string id, bool perfor
 
             QueueDel(uid);
 
+            // If ChangeEntity has ran, then the entity uid has changed and the
+            // new entity should be initialized by this point.
+            var afterChangeEv = new AfterConstructionChangeEntityEvent(construction.Graph, construction.Node, previousNode);
+            RaiseLocalEvent(uid, afterChangeEv);
```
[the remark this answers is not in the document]
BarryNorfolk: Blind on my part, should be `newUid` since that's the new entity that's been created as part of the construction graph move.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,364 @@
...
+        // Whichever component has the MultipartMachine component should be treated as the origin
+        var machineOrigin = _mapSystem.TileIndicesFor(gridUid!.Value, grid, xform.Coordinates);
+
+        // Set to true if any of the parts' state changes
+        var stateHasChanged = false;
+
+        var missingParts = false;
+        for (var i = 0; i < ent.Comp.Parts.Count; ++i)
+        {
+            var part = ent.Comp.Parts.Values.ElementAt(i);
```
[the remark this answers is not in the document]
BarryNorfolk: I think I encountered some oddness with the things being readonly or at least not modify-able, but this seems to work. Thanks.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,364 @@
...
+                // We have NO idea where our parts could be orientated so we'll have to iterate through 360 degrees
+                // to try and find a match
+                Angle curAngle = 0;
+                for (var j = 0; j < 4; ++j)
+                {
+                    if (ScanPart(machineOrigin, curAngle, query, gridUid.Value, grid, ref part))
+                    {
+                        // This entity succeeds, store the direction we used to get this one and expect all
+                        // future machine parts to match this direction.
+                        ent.Comp.Rotation = curAngle;
```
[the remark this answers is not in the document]
BarryNorfolk: That wasn't my intention no, my intention was to have the main entities rotation be independent from the machine entirely so you could have it face any direction. Maybe we want to swap that which would then eliminate the rotation check here completely.

I don't really mind or have any strong opinion on it.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
BarryNorfolk: I don't believe this is a hack, as the rescan now will just emit an event if the machine's state has changed (new/removed parts. I feel it's better to let the UI message here ask the for the rescan and let the flow of event handling do the job of updating the UI state. 
Else I have to have the rescan here return a bool but also have an optional to parameter to control whether an event has been raised or not. Perhaps I can make the comment here clearer if that makes sense.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
BarryNorfolk: Should have been removed, missed that. The `ParticleAcceleratorPartComponent` is now defunct.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
BarryNorfolk: Don't understand what would be fixed here by the multipart code? This is the PA system deciding how to set its UI state depending on what parts are available.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,364 @@
...
+    /// <param name="args">Args for this event, notably the anchor status</param>
+    private void OnConstructionAnchorChanged(Entity<ConstructionComponent> ent, ref AnchorStateChangedEvent args)
+    {
+        var query = EntityQueryEnumerator<MultipartMachineComponent>();
+        while (query.MoveNext(out var uid, out var machine))
+        {
+            if (!args.Anchored)
+            {
+                // Some construction is being unanchored, check if its a known part for us
+                foreach (var part in machine.Parts.Values)
```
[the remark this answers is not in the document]
BarryNorfolk: I think it's better to have the event still trigger so things can properly listen out for it, rather than altering the components IsAssembled state or creating a new event for it.
I like the idea of just changing the unanchored part with a master though, that is an easier time and reduces the cost of doing this.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,364 @@
...
+
+            if (!TryComp<ConstructionComponent>(entity, out var construction) ||
+                construction.Graph != part.Graph ||
+                construction.Node != part.ExpectedNode)
+            {
+                // This constructable doesn't match the right graph we expect
+                continue;
+            }
+
+            part.Entity = GetNetEntity(entity);
```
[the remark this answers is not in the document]
BarryNorfolk: Sadly I think having the EntityUid inside a class that's USED by a component makes the thing not work.
EntityUid isn't marked as serializable so it falls over when you try and run the thing.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,364 @@
...
+    {
+        var query = EntityQueryEnumerator<MultipartMachineComponent>();
+        while (query.MoveNext(out var uid, out var machine))
+        {
+            foreach (var part in machine.Parts.Values)
+            {
+                if (args.Graph == part.Graph &&
+                    (args.PreviousNode == part.ExpectedNode || args.CurrentNode == part.ExpectedNode))
+                {
+                    Rescan((uid, machine));
```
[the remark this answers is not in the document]
BarryNorfolk: We could look at the maximum distance from any part that's specified in the YAML and create a "max range" that we epxect things to be in.
That should enable us to not scan EVERY machine which could be expensive. I'll look into this later.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,364 @@
...
+                }
+
+                stateHasChanged = true;
+            }
+        }
+
+        ent.Comp.Assembled = !missingParts;
+        if (stateHasChanged)
+        {
+            var ev = new MultipartMachineAssemblyStateChanged(ent, ent.Comp.Assembled, user);
```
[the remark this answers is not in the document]
BarryNorfolk: Sure, that's a neat idea.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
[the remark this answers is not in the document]
BarryNorfolk: I clearly wrote that in reponse to something but I have no idea anymore, don't code when tired folks.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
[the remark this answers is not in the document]
BarryNorfolk: Makes sense tbh, I forgot that the PA has a rotation with the wires.

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
BarryNorfolk: Discussed a little on discord and I think it _could_ be done but it makes things a bit odd. 

The UI state messages appear to be processed first, so the components aren't updated by the time we handle that message. To make this work we'd either have to have a split UI update from the server, one after the component has updated which updates any preview and then another after the UI state comes in which updates the on/off or power states. Or move it into a frame update tick, with the same level of desync.

I agree sending this all down the pipe from the server feels a bit crap for the moment. I can give it a try anyway but not sure how it'll turn out.

BarryNorfolk: Hrm, this is a bit annoying in how it's structured and I feel like I would be reworking a large amount of the particle accelerator UI than I'd like :/

Content.Server/ParticleAccelerator/EntitySystems/ParticleAcceleratorSystem.ControlBox.cs
BarryNorfolk: Made the comment a bit clearer that this is a manual rescan raised by the user, and may have no effect because we (hopefully) catch all anchor/unanchor/construction events as you point out.
Anything more than this and I'm getting into UI changes for the particle accelerator which I'm not comfortable with and feel like it's scope creeping.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,286 @@
...
+    [Dependency] private readonly IEntityManager _entManager = default!;
+    [Dependency] private readonly MapSystem _mapSystem = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<MultipartMachineComponent, ComponentStartup>(OnComponentStartup);
+
+        SubscribeLocalEvent<ConstructionComponent, AfterConstructionChangeEntityEvent>(OnConstructionNodeChanged);
```
Fildrance, line 31: it'd greatly lower amount calls if we apply MultipartMachinePartComponent on prototype level, and and subscribe to 
```
SubscribeLocalEvent<MultipartMachinePartComponent , AfterConstructionChangeEntityEvent> 
SubscribeLocalEvent<MultipartMachinePartComponent , AnchorStateChangedEvent>(OnConstructionAnchorChanged);
```

Fildrance (changes requested):
Just a little bit more. Most importantly we would like to avoid subbing for ConstructionComponent component, as it will lead to multiple calls, making server busy with not really useful work : D

Content.Shared/Machines/Components/MultipartMachineComponent.cs
```diff
@@ -1,10 +1,48 @@
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public Dictionary<Enum, MachinePart> Parts = [];
+
+    /// <summary>
+    /// Whether this multipart machine is assembled or not.
+    /// Optional parts are not taken into account.
+    /// </summary>
+    [DataField]
+    public bool IsAssembled = false;
```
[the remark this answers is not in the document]
BarryNorfolk: Only updated in one place so should be fine! Fixed.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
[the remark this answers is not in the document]
BarryNorfolk: I think I've fixed this so we don't scan ALL the machines all the time.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,286 @@
...
+            var query = _entManager.GetEntityQuery(registration.Type);
+
+            ScanPart(machineOrigin, machineRotation, query, gridUid.Value, grid, part);
+
+            if (!part.Entity.HasValue && !part.Optional)
+                missingParts = true;
+
+            // Even optional parts should trigger state updates
+            if (part.Entity != originalPart)
+            {
```
Fildrance, line 93: slap early return here
```
if (part.Entity == originalPart)
    continue;
````

File: Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs [hunk 32]
```diff
@@ -0,0 +1,286 @@
...
+        {
+            var ev = new MultipartMachineAssemblyStateChanged(ent,
+                ent.Comp.IsAssembled,
+                user,
+                partsAdded,
+                partsRemoved);
+            RaiseLocalEvent(ent, ref ev);
+        }
+
+        Dirty(ent);
```
Fildrance, line 126: better to call dirty only if something changed (its not cheap operation) - happily we clearly have markers for that!

Fildrance (changes requested):
During testing i found out that right after creation PA comuter is displaying all parts 'in place' and rescanning is not helping with it, unless at least 1 part of machine is placed correctly - maybe it'll be easy to fix? Can u look at it? Maybe just init UI empty...

BarryNorfolk:
The fix for the 'in place' for the UI is easy and have done it, plus your other two comments.
The hard part is not handling the unanchoring logic so that the PA doesn't continue to fire after the computer is removed from the machine.

If I clear all the parts from the machine when the unanchor event arrives, which is straightforward, and then raise the state change event then the Controlbox logic attempts to update the visual appearance for the parts. Sadly those are now removed so the individual parts stay in the "incorrect" appearance until the computer is re-added. I can add the entities added/removed to the state change message but then updating all the control box logic to handle the extra chance of an event seems... grim.

I'm at a bit of an impasse there with it :/ Either I have to update the entire control box logic to pipe through the event changes or live with the machine just firing even without the PA..

Fildrance (approved):
**Left some more comments with proposed changes with code** - so u can just cherry-pick it or copy-paste it. Tested in-game with them - looks like it works fine!
I still would love to see if Sloth have any ideas about our range-searching (i'll poke him), but i personally think that its good to go.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,346 @@
...
+        while (query.MoveNext(out var uid, out var machineComp))
+        {
+            var machine = (uid, machineComp);
+            if (!IsMachineInRange(machine, constructXform.LocalPosition))
+                continue; // This part is outside the max range of the machine, ignore
+
+            if (Rescan(machine) && HasPartEntity(machine, ent.Owner))
+                return; // This machine is using this entity so we don't need to go any further
+        }
+    }
```
metalgearsloth, line 275: Use EntityLookup as it will vary between entityquery vs range checks as required.

Content.Shared/Machines/EntitySystems/SharedMultipartMachineSystem.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        foreach (var part in machine.Comp.Parts.Values)
+        {
+            if (part.Entity.HasValue && part.Entity.Value == entity)
+                return true;
+        }
+
+        return false;
+    }
+
```
metalgearsloth, line 70: Any reason you don't just use entityuid for these? NetEntity is really only for networking.

Content.Shared/Machines/Components/MultipartMachineComponent.cs
```diff
@@ -0,0 +1,103 @@
...
+    [DataField, AutoNetworkedField]
+    public bool IsAssembled = false;
+
+    /// <summary>
+    /// Flag for whether the client side system is allowed to show
+    /// ghosts of missing machine parts.
+    /// Controlled/Used by the client side.
+    /// </summary>
+    [DataField]
+    public List<EntityUid> Ghosts = [];
```
metalgearsloth, line 37: If only the client uses these then delete datafield as client won't serialize them.

If server uses then needs autonetworkedfield.

Content.Shared/Machines/Components/MultipartMachineComponent.cs
```diff
@@ -0,0 +1,103 @@
...
+    [DataField]
+    public Angle Rotation = Angle.Zero;
+
+    /// <summary>
+    /// Network entity associated with this part.
+    /// Not null when an entity is successfully matched to the part and null otherwise.
+    /// </summary>
+    [DataField]
+    public NetEntity? Entity = null;
+
```
metalgearsloth, line 89: This won't save properly and will explode whenever someone tries this with persistence, you're better off just using EntityUid and having a temporary field for the netentity then flagging this as [NonSerialized]

Content.Shared/Machines/EntitySystems/SharedMultipartMachineSystem.cs
BarryNorfolk: You seem to be able to have EntityUid's inside of a component, but not inside any struct that the component uses. So having a list of machine parts like I do means I just can't seem to use these.

Content.Shared/Machines/Components/MultipartMachineComponent.cs
BarryNorfolk: Do you mean that we should cache it locally ok the client with a temporary netent when the component is dirtied?

Content.Shared/Machines/EntitySystems/SharedMultipartMachineSystem.cs
BarryNorfolk: Removed now that we rely more on the EntityUid

Content.Shared/Machines/Components/MultipartMachinePartComponent.cs
```diff
@@ -0,0 +1,15 @@
+namespace Content.Shared.Machines.Components;
+
+/// <summary>
+/// Server side component for marking entities as part of a multipart machine.
+/// </summary>
+[RegisterComponent]
+public sealed partial class MultipartMachinePartComponent : Component
+{
```
metalgearsloth: NetworkedComponent + update comment.

Content.Shared/Machines/Components/MultipartMachineGhostComponent.cs
```diff
@@ -0,0 +1,14 @@
...
+/// Component attached to all multipart machine ghosts
+/// Intended for client side usage only, but used on prototypes.
+/// </summary>
+[RegisterComponent]
+public sealed partial class MultipartMachineGhostComponent : Component
+{
+    /// <summary>
+    /// Machine this particular ghost is linked to.
+    /// </summary>
+    public EntityUid? LinkedMachine = null;
```
metalgearsloth: Same here probably also needs datafield so it deserializes properly.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,324 @@
...
+
+/// <summary>
+/// Server side handling of multipart machines.
+/// When requested, performs scans of the map area around the specified entity
+/// to find and match parts of the machine.
+/// </summary>
+public sealed class MultipartMachineSystem : SharedMultipartMachineSystem
+{
+    [Dependency] private readonly IComponentFactory _factory = default!;
+    [Dependency] private readonly IEntityManager _entManager = default!;
```
metalgearsloth: These are already dependencies on systems.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,324 @@
...
+    /// </summary>
+    /// <param name="ent">Entity to clear the part for.</param>
+    /// <param name="part">Enum value for the part to clear.</param>
+    public void ClearPartEntity(Entity<MultipartMachineComponent?> ent, Enum part)
+    {
+        if (!Resolve(ent, ref ent.Comp))
+            return;
+
+        if (ent.Comp.Parts.TryGetValue(part, out var value))
+            value.Entity = null;
```
metalgearsloth: Is this supposed to dirty?

metalgearsloth (dismissed):

Content.Client/Machines/EntitySystems/MultipartMachineSystem.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// Can only show one set of ghost parts at a time and their location depends on the current map/grid
+    /// location of the origin machine.
+    /// </summary>
+    /// <param name="ent">Entity/Component that has been inspected.</param>
+    /// <param name="args">Args for the event.</param>
+    private void OnMachineExamined(Entity<MultipartMachineComponent> ent, ref ClientExaminedEvent args)
+    {
+        if (ent.Comp.Ghosts.Count != 0)
+        {
+            // Already showing some part ghosts
```
TheShuEd, line 49: Given that we're storing a list of ghost sprites, we can just delete the past set here before creating a new one. This will be especially handy if the player takes turns inspecting different multistructures, and doesn't need to wait for past ghost objects to disappear each time

(not merge-blocking, just thinking.)

Resources/Prototypes/Entities/Structures/Power/Generation/PA/control_box.yml
```diff
@@ -27,6 +27,62 @@
...
   - type: AccessReader
     access: [["Engineering"]]
+  - type: MultipartMachine
+    parts:
+      enum.AcceleratorParts.EndCap:
+        component: ParticleAcceleratorEndCap
+        offset: 1, -1
+        rotation: -90
+        sprite:
+          sprite: Structures/Power/Generation/PA/end_cap.rsi
```
TheShuEd: how about we pass EntProtoId here instead of a SpriteSpecifier? This would pull layered sprites from the SpriteComponent, as well as the description and part name for inspection. It's quite possible to throw in Dummy entities, with a separate custom description and no logic

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
Fildrance: yes, also added making part component have null on master and dirtied it.

Content.Server/Machines/EntitySystems/MultipartMachineSystem.cs
Fildrance: fixed

Content.Shared/Machines/Components/MultipartMachinePartComponent.cs
Fildrance: added dirty and networking

Content.Shared/Machines/Components/MultipartMachineGhostComponent.cs
Fildrance: Moved comp the thing to client - i am sure that its not wanted on server?...

Content.Client/Machines/EntitySystems/MultipartMachineSystem.cs
Fildrance: To be honest i think that it could be beneficial to have multiple ghosts - for example if we will have different multipart machines that could interact in some way, and u want to connect them... its useful!

Resources/Prototypes/Entities/Structures/Power/Generation/PA/control_box.yml
Fildrance: Exceptional idea, looks great


https://github.com/user-attachments/assets/e55802ad-c26d-4877-8b76-2a2f7a50c588

PR #35977, opened 2025-03-21
Predict pipe anchoring and unanchroing

beck-thompson:
## About the PR

Title! Its makes sense and doesn't lag and now it gives you a nice popup.
## Technical details

Just moved the stuff from server to shared.

Screenshot from 2025-03-20 18-25-42


Content.Shared/SubFloor/SharedSubFloorHideSystem.cs
```diff
@@ -38,6 +41,32 @@ public override void Initialize()
...
+
+        private void OnAnchorAttempt(EntityUid uid, SubFloorHideComponent component, AnchorAttemptEvent args)
+        {
+            // No teleporting entities through floor tiles when anchoring them.
+            var xform = Transform(uid);
+
+            if (TryComp<MapGridComponent>(xform.GridUid, out var grid)
+                && HasFloorCover(xform.GridUid.Value, grid, Map.TileIndicesFor(xform.GridUid.Value, grid, xform.Coordinates)))
+            {
+                _popup.PopupPredicted(Loc.GetString("subfloor-anchor-failure", ("entity", uid)), uid, null);
```
Tayrtahn, line 56: With `null` as the recipient, this is the same as `PopupEntity` and isn't predicted.

But I noticed that passing in the uid or using `PopupClient` results in nothing happening for the local client, which suggests that this code isn't getting run locally. Not sure what that's about, but maybe go with `PopupEntity` for now.

beck-thompson: I figured it out! Uid is the pipe not the person anchoring the pipe :laughing:

Tayrtahn: Oh, hah! Good job sorting that out!

Tayrtahn (approved):

TeenSarlacc:
unanchroing

PR #35984, opened 2025-03-21
Mob movement minor refactor

metalgearsloth:
This is an extension of https://github.com/space-wizards/space-station-14/pull/35931 and that should be merged first as this changes many more things.

Main goals:
- Cleanup relay code a bit.
- Remove the TryUpdateRelative performance sinkhole with relayed entities (e.g. mice in chef's hat dirtying movement every single tick).
- Cleanup camera bugs, e.g. when rotating the camera and going across grid, when controlling another entity, or teleporting. This is mainly from ordering mob movement so relays are run before their targets.
- Remove the ToParent singleton for jetpacks and standardise their movement. This will require re-designing jetpack movement a bit to emulate the old one.

:cl:
- fix: Fix going onto grid with an active jetpack dropping your movement inputs.
- fix: Fix camera rotation in a lot of instances including when following entities and when using a jetpack.


metalgearsloth:
Okay I think the only thing left on this is having to re-design jetpacks to be parity to before but this should fix all of the camera bugs on master that I know about and also fix TryUpdateRelative spamming dirty states out constantly.

Pronana:
Left a partial review. I haven't tested any of the conveyor stuff yet so hopefully it all looks good there.

metalgearsloth:
Wrapped into cheeseballs pr

PR #35988, opened 2025-03-21
Rename Laser Rifle to Laser Carbine

Nox38:
## About the PR
Updated the name of the laser rifle to laser carbine. Also did this to the practice laser rifle.

## Why / Balance
Renaming it makes it match the pulse carbine vs pulse rifle difference more closely, and also its own internal name (which is laser carbine). In addition, the word rifle refers to barrel rifling which the laser rifle does not have as far as we know.

## Technical details
Just changed two visual names, nothing complex.

## Requirements

**Changelog**
:cl: Nox38
- tweak: Renamed 'Laser Rifle' to 'Laser Carbine'

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -199,10 +199,10 @@
   - type: Appearance
 
 - type: entity
-  name: practice laser rifle
+  name: practice laser carbine
   parent: [BaseWeaponBattery, BaseGunWieldable]
   id: WeaponLaserCarbinePractice
-  description: This modified laser rifle fires nearly harmless beams in the 40-watt range, for target practice.
+  description: This modified laser carbine fires nearly harmless beams in the 40-watt range, for target practice.
   components:
   - type: Sprite
     sprite: Objects/Weapons/Guns/Battery/laser_gun.rsi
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
The pulse carbine is explicitly named like so because it's a different gun from the pulse rifle. As such this rename doesn't make much sense.

NoElkaTheGod:
Laser with a rifled barrel sounds pretty damn cursed if you think about it. Like why would any sane person ever do that? Why does a laser even have a barrel to begin with? How do i unsee it?

Errant-4:
Disclaimer, I may be biased because I was the one who renamed them to rifle (from "laser gun")

For the name being rifle, I must say that while technically true, predominantly no one cares in scifi. ESPECIALLY the kind of retro-scifi with space carps and dragons. Laser Rifle sounds CoolTM, and that's enough. As noted above, pulse weapons have a carbine to differentiate them from the pulse rifle.

ArtisticRoomba:
A [carbine](https://en.wikipedia.org/wiki/Carbine) is simply a rifle with a shortened barrel or less powerful cartridge. It makes the gun easier to handle in a sense.

You can see this design philosophy between the pulse rifle and carbine: the carbine has a noticeably shorter barrel length and it has a lower ammo capacity as well.

Applying this to the pulse rifle doesn't make sense: it's clearly a fully fledged regular size rifle, why would it have a carbine designation?

Thus I don't think this rename makes much sense IMO, if we're trying to make gun names more correct by definition.

image

deltanedas:
laser blaster :trollface:

Thatonestomf:
just rename it to the phaser, now no one (but goonstation) is happy :trollface:

deltanedas:
least of all disneys lawyers

notquitehadouken:
rename to "laser"

Pixeltheaertist:
So either we rename it to Laser Carbine, or we migrate the ID to LaserRifle. Because it's HELLA annoying that the ID doesn't match the ingame name

Nox38:
Roomba makes a good point, going to close this PR.

PR #35993, opened 2025-03-21
Improve paradox clone item copying

slarticodefast:
## About the PR
Now copies currency in stores like the uplink, labels, forensics, and paper data.

## Why / Balance
No more full TC uplink for clones, since they also get the items the original has.
The others cases were picked because they are the most common ways to identify an item.

## Technical details
Moved copying data to the cloned item into a subscription.
This should allow forks to easily add their own components if needed.
The chosen components are some of the most importanat ones. Of course this could be extended with more and more edge cases, but I think with this item copying is good enough for any gameplay purposes regarding the paradox clone (with the exception of pda uplinks, which needs a refactor first).
In the future a more general approach to copying existing items, like an `IClonable` interface for components could be used to make a larger selection of components copyable with a better syntax.

## How to test
- Put some items into your inventory.
- Spawn a "Random Clone Spawner" from the spawn menu
- Compare

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- tweak: Further improved item copying for paradox clones.

### DIFF
--- a/Content.Server/Cloning/CloningSystem.Subscriptions.cs
+++ b/Content.Server/Cloning/CloningSystem.Subscriptions.cs
@@ -1,0 +1,69 @@
+using Content.Server.Forensics;
+using Content.Shared.Cloning.Events;
+using Content.Shared.FixedPoint;
+using Content.Shared.Labels.Components;
+using Content.Shared.Labels.EntitySystems;
+using Content.Shared.Stacks;
+using Content.Shared.Store;
+using Content.Shared.Store.Components;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.Cloning;
+
+/// <summary>
+///     The part of item cloning responsible for copying over important components.
+///     This is used for <see cref="CopyItem"/>.
+///     Anything not copied over here gets reverted to the values the item had in its prototype.
+/// </summary>
+/// <remarks>
+///     This method of copying items is of course not perfect as we cannot clone every single component, which would be pretty much impossible with our ECS.
+///     We only consider the most important components so the paradox clone gets similar equipment.
+///     This method of using subscriptions was chosen to make it easy for forks to add their own custom components that need to be copied.
+/// </remarks>
+public sealed partial class CloningSystem : EntitySystem
+{
+    [Dependency] private readonly SharedStackSystem _stack = default!;
+    [Dependency] private readonly SharedLabelSystem _label = default!;
+    [Dependency] private readonly ForensicsSystem _forensics = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<StackComponent, CloningItemEvent>(OnCloneStack);
+        SubscribeLocalEvent<LabelComponent, CloningItemEvent>(OnCloneLabel);
+        SubscribeLocalEvent<ForensicsComponent, CloningItemEvent>(OnCloneForensics);
+        SubscribeLocalEvent<StoreComponent, CloningItemEvent>(OnCloneStore);
+    }
+
+    private void OnCloneStack(Entity<StackComponent> ent, ref CloningItemEvent args)
+    {
+        // if the clone is a stack as well, adjust the count of the copy
+        if (TryComp<StackComponent>(args.CloneUid, out var cloneStackComp))
+            _stack.SetCount(args.CloneUid, ent.Comp.Count, cloneStackComp);
+    }
+
+    private void OnCloneLabel(Entity<LabelComponent> ent, ref CloningItemEvent args)
+    {
+        // copy the label
+        _label.Label(args.CloneUid, ent.Comp.CurrentLabel);
+    }
+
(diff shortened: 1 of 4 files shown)


Content.Client/Paper/UI/PaperVisualizerSystem.cs
```diff
@@ -11,13 +11,21 @@ protected override void OnAppearanceChange(EntityUid uid, PaperVisualsComponent
...
-        if (AppearanceSystem.TryGetData<PaperStatus>(uid, PaperVisuals.Status , out var writingStatus, args.Component))
+        if (AppearanceSystem.TryGetData<PaperStatus>(uid, PaperVisuals.Status, out var writingStatus, args.Component))
             args.Sprite.LayerSetVisible(PaperVisualLayers.Writing, writingStatus == PaperStatus.Written);
 
-        if (AppearanceSystem.TryGetData<string>(uid, PaperVisuals.Stamp, out var stampState, args.Component))
+        if (AppearanceSystem.TryGetData<string>(uid, PaperVisuals.Stamp, out var stampState, args.Component) && stampState != "")
         {
-            args.Sprite.LayerSetState(PaperVisualLayers.Stamp, stampState);
-            args.Sprite.LayerSetVisible(PaperVisualLayers.Stamp, true);
+            if (stampState != "")
```
Tayrtahn, line 19: Won't this always be true, due to the condition above?

slarticodefast: Oh, forgot to remove that one when I moved it

slarticodefast: fixed

beck-thompson (changes requested):
Seems good! The only thing I think is important is to also correctly copy implants. It can lead to weird situations like this:

(Unused implant but it still has DNA)
image

slarticodefast:
Uff, that is because I am not copying item slots at the moment, because those are a total mess with 100 different implementations for each component.
So the implanter gets reset to its spawn condition.
I would probably just blacklist implanters due to that, since you already get a copy of the implant itself, and noone runs around with a full implanter.

slarticodefast:
I blacklisted implanters for now, which should be fine since you already get a copy of the implant itself.
To properly copy the contents of implanters `ImplanterComponent` will need to be standardized to use `ContainerFill` and `ItemSlotsComponent` instead of having its own slot and spawn logic.

beck-thompson (approved):
seems good to me!

ScarKy0 (changes requested):

Content.Client/Paper/UI/PaperVisualizerSystem.cs
```diff
@@ -11,13 +11,21 @@ protected override void OnAppearanceChange(EntityUid uid, PaperVisualsComponent
...
 
-        if (AppearanceSystem.TryGetData<PaperStatus>(uid, PaperVisuals.Status , out var writingStatus, args.Component))
+        if (AppearanceSystem.TryGetData<PaperStatus>(uid, PaperVisuals.Status, out var writingStatus, args.Component))
             args.Sprite.LayerSetVisible(PaperVisualLayers.Writing, writingStatus == PaperStatus.Written);
 
         if (AppearanceSystem.TryGetData<string>(uid, PaperVisuals.Stamp, out var stampState, args.Component))
         {
-            args.Sprite.LayerSetState(PaperVisualLayers.Stamp, stampState);
-            args.Sprite.LayerSetVisible(PaperVisualLayers.Stamp, true);
+            if (stampState != "")
```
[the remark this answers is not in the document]
slarticodefast: It is literally the [same](https://learn.microsoft.com/en-us/dotnet/api/system.string.empty?view=net-9.0)
I guess it's a matter of code style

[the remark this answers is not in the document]
ScarKy0: string.Empty looks prettier than just "" but if it doesnt matter then all good

Content.Server/Cloning/CloningSystem.Subscriptions.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<StackComponent, CloningItemEvent>(OnCloneStack);
+        SubscribeLocalEvent<LabelComponent, CloningItemEvent>(OnCloneLabel);
+        SubscribeLocalEvent<PaperComponent, CloningItemEvent>(OnClonePaper);
+        SubscribeLocalEvent<ForensicsComponent, CloningItemEvent>(OnCloneForensics);
+        SubscribeLocalEvent<StoreComponent, CloningItemEvent>(OnCloneStore);
```
ScarKy0, line 40: Maybe should clone over item charges too? (like from flashes)
Not a blocker but would be nice to have

ScarKy0 (dismissed):

ScarKy0 (approved):
good stuff

PR #17722, opened 2023-06-29
Deathrattle implant for uplink and slight tracking implant tweak

Arendian:
## About the PR

Adds death rattle implants to the syndicate uplink, these implants send a message to the syndicate radio informing all people with access to it of the implanted person's critical state or their death. This message includes the coordinates of the person.
The box also contains GPS systems so people can actually find the relayed coordinates.

The tracking implant got improved and now has the function of sending it's location upon the person going crit to the security radio channel, making it useful for keeping track of the health status of VIP's. 
To combine the crit rattle and death rattle effects into one implanter I also made it possible now for a single implant to have multiple MobState triggers.

**Media**

https://github.com/space-wizards/space-station-14/assets/137322659/18eceea2-4614-442b-85b1-c55965436191

image
image

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl: Dygon
- add: In an effort to reduce the amount of manpower lost during Syndicate missions operators can now order deathrattle implants from the uplink. These implants send your coordinates to the syndicate radio when going into a critical state and upon dying.
- tweak: Tracking implants now relay the location of an implanted person to the security channel when they go into a critical state.

PR #35995, opened 2025-03-21
Remove deathrattle functionality from tracking implants

SlamBamActionman:
## About the PR

Removes the deathrattle functionality from tracking implants. The deathrattle is the radio message that indicates the location of the person going critical.

## Why / Balance

Tracking implants were initially meant to be used primarily for paroled prisoners as a way to keep track of them. This is not how they're used. Sec and Command instead use them primarily for the deathrattle to know when one of their own are being killed. This has raised some questions regarding whether it's powergaming or not for Command to go into Sec roundstart every shift when ideally they shouldn't feel the need to. 

This nerf is an attempt to curb that behavior; all it provides now is crew monitoring, which isn't as impactful for Command (who should already have their coords on) compared to an antag being tracked.

## Technical details

Just yaml change.

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Tracking implants no longer announce on radio when someone goes into critical health.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
@@ -79,7 +79,7 @@
   parent: BaseSubdermalImplant
   id: TrackingImplant
   name: tracking implant
-  description: This implant has a tracking device attached to the suit sensor network, as well as a condition monitor for the Security radio channel.
+  description: This implant has a tracking device attached to the suit sensor network.
   categories: [ HideSpawnMenu ]
   components:
     - type: SubdermalImplant
@@ -97,11 +97,6 @@
     - type: StationLimitedNetwork
     - type: WirelessNetworkConnection
       range: 500
-    - type: TriggerOnMobstateChange
-      mobState:
-      - Critical
-    - type: Rattle
-      radioChannel: "Security"
 
 #Traitor implants
 


Mixelz:
This just feels like yet another implant change to the detriment of the crew and the benefit of antags. While I would agree implanting most command round start is bad, that is more a mapping issue of giving Security enough implants to do this instead of only one or two.

I think this change would make the implant near useless, yet there is not much of way in regards to otherwise balancing it beyond letting the station having less trackers to go around.

NoElkaTheGod:
I do agree with the comment above. Yes, all of command using trackers is an issue, but it should be solved by making trackers more expensive, not making them useless. It should be possible to track all of command, it just shouldn't be as easy as opening a box in the armory.

Side note: as an insane person that actually uses the "unreviveable" trait, getting a tracker is pretty much the only way to avoid being round-removed by my own stupidity. And for some reason others get very sad when an unreviveable character dies.

NoElkaTheGod:
Side note 2: tracker implants are completely negated by using a radio jammer, which is a very cheap item that should be used more often.

I hereby declare this PR a skill issue :godo:

Killerqu00:
> Side note 2: tracker implants are completely negated by using a radio jammer, which is a very cheap item that should be used more often.
> 
> I hereby declare this PR a skill issue :godo:

If a certain item is a must-have for an objective, it is a gamedesign skill issue.

NoElkaTheGod:
> If a certain item is a must-have for an objective, it is a gamedesign skill issue.

Well, if you have a "kill head" objective, weapon is just as much of a must-have, yet nobody is complaining about that. Yes, you can kill someone with a toolbox you found in maints if you don't want to buy a weapon. You also can disable the telecoms if you don't want the tracker implant to alert everyone, or just hope that you'll do your job faster than security arrives. Hell, your target may not even have a tracker (though this is unlikely, because right now they're basicaly free, so there's no real reason not to use it as a head).

Yes, i do agree that the current implementation causes some issues for antags, but making trackers useless isn't really a good solution either.

walksanatora:
yeah if you have to kill someone. you probaally allready SHOULD be getting  radio jammer. since that makes them dissapear off crew monitor and makes it less likely they will notice you.

Killerqu00:
> if you have a "kill head" objective, weapon is just as much of a must-have

weapon is not a single item, it is a HUGE variety of items. Besides, you can use poison, traps, as well as other unconventional means. if you want to bypass the tracker, you need to buy a specific item (which you don't even KNOW if you should buy, since there is no way to know if a person has the tracker or not) or sabotage comms station-wide (overkill). the only way you can find out if the person has an implant is a bug with double trackers (which is a. bug and b. getting fixed in a separate PR), which is lame

> though this is unlikely, because right now they're basicaly free

why does command/security need an item that basically screams "THERE IS A BIG CHANCE THAT AN ANTAG KILLED ME IN THIS LOCATION" for no cost/drawback whatsoever roundstart(which is green alert)? the answer is they dont

> I hereby declare this PR a skill issue
> getting a tracker is pretty much the only way to avoid being round-removed by my own stupidity

well played

NoElkaTheGod:
I mean, radio jammer isn't strictly required, but it does eliminate a lot of risk that you caan't prevent otherwise (for example: your target having a tracker, them typing "; help maints" fast enough, or someone looking at the crew monitoring at this exact moment). You don't HAVE to buy it, just don't complain if you were spotted because you decided not to use it.

NoElkaTheGod:
> weapon is not a single item, it is a HUGE variety of items. Besides, you can use poison, traps, as well as other unconventional means. if you want to bypass the tracker, you need to buy a specific item (which you don't even KNOW if you should buy, since there is no way to know if a person has the tracker or not) or sabotage comms station-wide (overkill). the only way you can find out if the person has an implant is a bug with double trackers (which is a. bug and b. getting fixed in a separate PR), which is lame

The tracker implant isn't the only thing that could expose you, and which is blocked by the radio jammer. The other two are suit sensors and the headset, which are not guaranteed to immediately expose you, but have the potential to do so. Thus, you should buy the jammer anyways if you have the TC for it and don't want extra risk.

> why does command/security need an item that basically screams "THERE IS A BIG CHANCE THAT AN ANTAG KILLED ME IN THIS LOCATION" for no cost/drawback whatsoever roundstart(which is green alert)? the answer is they dont

Yes, they don't. That's why trackers shouldn't be free. Making them useless instead is not a good solution.

> > I hereby declare this PR a skill issue
> > getting a tracker is pretty much the only way to avoid being round-removed by my own stupidity
> 
> well played

Just to clarify, the "skill issue" part was a joke.

SlamBamActionman:
https://github.com/space-wizards/space-station-14/pull/17722#issuecomment-1635023592
image

SlamBamActionman:
> Yes, they don't. That's why trackers shouldn't be free. Making them useless instead is not a good solution.

To respond to this line of thinking specifically; they are only becoming useless in the context of alerting Security of Sec/Command going crit. They are still plenty effective as a solution to track paroled detainees, while still allowing antags to bypass it without making it a necessity. 

The stated purpose is to make them less appealing to Sec/Command, so yes, for _that_ purpose they are now fairly pointless.

Tayrtahn:
Getting rattle implants for command does seem quite useful, but I think that would make sense as a separate implant. Let tracking implants be used for their intended purpose on parolees, and maybe make NT-approved deathrattle implants available to crew, though not easily - maybe they could be a high-tier medical science unlock?

Pronana:
Genuinely a good change, especially with how much of a gamble it can be for if someone has a rattler in them. 

That being said I feel like it would be a good idea to have rattlers be pre-installed on some people similar to mindshields but that's a balancing PR for another day.

Mixelz:
I think it would be good game design wise to make killing a command member HARDER to do than killing a regular crew member, not only by their position and frequent socializing with people but also through having a higher level of safety such as with the death rattle on implant trackers.

superjj18:
This nerf is meant to curb... the exact thing the tool was designed for?  Anyone can turn on their damn cords, the death rattle is the entire freakin point that actually matters.

The year is 2525 but fucking life alert is too advanced of a technology for nanotrasen

LaCumbiaDelCoronavirus:
I would like this change - if I was able to _consistently_ rely on medical and security (more specifically the warden) to actually, even once in a while, to check the crew monitoring console.

This is caused by either lack of players in necessary jobs, or lack of enough players in general... which makes this PR especially problematic on lowpop, where you can have, for example: a HOS and 2 officers, and nobody else as sec, so nobody actively watching the crew monitor & sec comms at once, assuming HOS is patrolling and not sitting in brig.

In such a situation, anyone with a tracking implant _(without a deathrattle function)_ will probably (90% chance) not have their corpse recovered when they crit, unless someone else coincidentally finds them.
With the deathrattle functionality however, anyone with a tracking implant will actually likely be saved if security isn't deaf/distracted/dead!

Also talking about antag objectives, I've never felt required to use a radio jammer against someone with a tracking implant _more than_ someone who can already just use comms to talk.

NoElkaTheGod:
Actualy, i have an idea to go with this. What if we moved deathrattle onto a separate device, but not an implant, and rather something that you have to wear on you? For example, make it a special headset (which still works as a headset, obviously, maybe with some sort of a downside) with a unique sprite. That way you don't have to gamble with "do i want a radio jammer, or do i hope that my target doesn't have a tracker?". Could be made a semi-expensive T2 arsenal (or civilian) research.

semensponge:
Just seperate the death rattle mechanic from the tracking implant and put it on a new death rattle implant for NT, just like Nukies have. Have them be purchasable by cargo, and then someone who cares like Captain, HoS, or CMO can order them and distribute them. Then, if its an issue of balance, you can tweak the numbers on the seperate item to fine tune.

WallaceLordofOwls:
> Just seperate the death rattle mechanic from the tracking implant and put it on a new death rattle implant for NT, just like Nukies have. Have them be purchasable by cargo, and then someone who cares like Captain, HoS, or CMO can order them and distribute them. Then, if its an issue of balance, you can tweak the numbers on the seperate item to fine tune.

This, absolutely this. I play a lot of sec, a lot of command and a good amount of syndie. I think sec should absolutely have access to the deathrattle functionality, as I think it makes syndie gameplay more interesting than "If you ambush them in maints they're never getting saved". If someone has a tracking implant, you need to either be fast, you need to use a radio jammer, or you need to skip the crit status on them, as if you go from 99 damage to 200 damage instantly, the person will never go crit and thus not trigger the tracking implant. If there are balance concerns with it, make it is own thing and it can be balanced with pricing, instead of removing a feature with no replacement.

slarticodefast:
Let's give it a try and see how it goes.
Potentially we can reintroduce the death rattle implant as a separate item if we want it back.
For the tracker it makes sense to only have it track your location, as it is intended to be used for prisoners.

slarticodefast (approved):

G-370:
I know this has already been merged, but I will give my own piece of feedback and thought about it.
I think this sort of modification **to the baseline wizden** is unnecessary,
For the following reasons:

1. This powergaming issue would absolutely be a problem in MRP+ servers, yes, but that can be ammended by server SoP, or in the case of downstreams, they themselves modifying it and removing the deathrattle.
2. Both while playing command, and while also observing rounds, on different servers, not even once did it come through in my mind, and I believe it didn't in theirs either, that tracking implants are meant for criminals, it is bundled together with mindshield implants and the game does NOT at any point, indicate what you should use it for.
3. It would be much more interesting to increase environmental story telling, such as leaving an official-looking document, inside the crates, specifying implant usage and protocol, because if then people misuse it, they can be promptly warned, banned etc, in MRP servers, and if they would argue "but I didn't read that paper" as a command player in MRP, you can hit them with a "command has higher expectations, specially in MRP".

PR #36023, opened 2025-03-23
Remove all trailing whitespaces in .cs files

CrafterKolyan:
## About the PR
Removes all trailing whitespaces in .yml files

## Why / Balance
Makes writing code in Space Station 14 repo easier with use of autoformatters as it allows to not get a lot of random diffs related to autoformatting.

## Technical details
Done using:
```bash
find -name "*.cs" | xargs sed -i -e "s/[[:space:]]\+$//"
```

## Requirements


slarticodefast:
I think this will just cause a lot of merge conflicts and it will be better to fix these whenever a file is changed anyways.
Anyone using an IDE should already do this automatically, but from time to time these sneak in when someone uses notepad for programming.
We could consider adding an integration to make sure any files changed in a PR have no trailing whitespace.

CrafterKolyan:
Agree. Will just close these. Something to remove trailing whitespaces automatically would be perfect.

PR #36030, opened 2025-03-23
make objectives use yml defined mind filters

deltanedas:
## About the PR
instead of copy pasting shit which is Bad now its just PickRandomPerson and a list of MindFilters
with filters you can do everything done previously by copypasta and more

## Why / Balance
mfw merge conflicts on the same objective shit for the 6th time because its always copy pasted
makes adding custom filters much much simpler, you just make a new filter and can use it in yml instead of copy pasting shitcode

## Technical details
- added `IMindPool` which is currently implemented in `AliveHumansPool`
- added `MindFilter` abstract class which defines the interface for filtering minds out
  you chain these together in yml to do exactly what you want
- filters can be set in yml, with all filters being invertable
  for example you can change kill objective to only target non-traitors by adding this
  ```yml
  filters:
  - !type:HasRoleMindFilter
    inverted: true # forbid picking fellow traitors
    whitelist:
      components:
      - TraitorRole
  - ...
  ```
  theres a pr to change objectives to not target the same person for everything, currently its defined in yml to only do this for help and alive objectives, you can easily do this for everything by adding a whitelist-less `TargetObjectiveMindFilter`
- added some other filters useful for the future and downstream usage like `JobMindFilter`
  command staff for pick random head is no longer hardcoded, its just `BodyMindFilter` checking the mob for CommandStaff. you can reuse this to check for anything

## Media
still real

## Requirements

## Breaking changes
- removed `PickRandomHeadComponent`, `RandomTraitorAliveComponent` and `RandomTraitorProgressComponent`. these are now defined in yml `filters` of `PickRandomPersonComponent`.

**Changelog**
:trollface:

### DIFF
--- a/Content.Server/Mind/Filters/TargetObjectiveMindFilter.cs
+++ b/Content.Server/Mind/Filters/TargetObjectiveMindFilter.cs
@@ -1,0 +1,43 @@
+using Content.Server.Objectives.Components;
+using Content.Shared.Mind;
+using Content.Shared.Mind.Filters;
+using Content.Shared.Whitelist;
+
+namespace Content.Server.Mind.Filters;
+
+/// <summary>
+/// A mind filter that removes minds if you have an objective targeting them matching a whitelist.
+/// </summary>
+/// <remarks>
+/// Used to prevent assigning multiple kill objectives for the same person.
+/// </remarks>
+public sealed partial class TargetObjectiveMindFilter : MindFilter
+{
+    /// <summary>
+    /// A whitelist to check objectives against to remove a mind.
+    /// If null then any objective targeting it will remove minds.
+    /// </summary>
+    [DataField]
+    public EntityWhitelist? Blacklist;
+
+    protected override bool ShouldRemove(Entity<MindComponent> mind, EntityUid? excluded, IEntityManager entMan, SharedMindSystem mindSys)
+    {
+        // ignore this filter if there is no user to check
+        if (!entMan.TryGetComponent<MindComponent>(excluded, out var excludedMind))
+            return false;
+
+        var whitelistSys = entMan.System<EntityWhitelistSystem>();
+        foreach (var objective in excludedMind.Objectives)
+        {
+            // if the player has an objective targeting this mind
+            if (entMan.TryGetComponent<TargetObjectiveComponent>(objective, out var kill) && kill.Target == mind.Owner)
+            {
+                // remove the mind if this objective is blacklisted
+                if (whitelistSys.IsBlacklistPassOrNull(Blacklist, objective))
+                    return true;
+            }
+        }
+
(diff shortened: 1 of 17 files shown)


Content.Server/Mind/Filters/TargetObjectiveMindFilter.cs
```diff
@@ -0,0 +1,43 @@
...
+/// Used to prevent assigning multiple kill objectives for the same person.
+/// </remarks>
+public sealed partial class TargetObjectiveMindFilter : MindFilter
+{
+    /// <summary>
+    /// A whitelist to check objectives against to remove a mind.
+    /// If null then any objective targeting it will remove minds.
+    /// </summary>
+    [DataField]
+    public EntityWhitelist? Blacklist;
```
SlamBamActionman, line 21: You use "whitelist" to describe the property above but it's called "blacklist" here, which seems inconsistent.

Content.Shared/Mind/Filters/AliveHumansPool.cs
```diff
@@ -0,0 +1,12 @@
+namespace Content.Shared.Mind.Filters;
+
+/// <summary>
+/// A mind pool that uses <see cref="SharedMindSystem.AddAliveHumans"/>.
+/// </summary>
+public sealed partial class AliveHumansPool : IMindPool
+{
+    void IMindPool.FindMinds(HashSet<Entity<MindComponent>> minds, EntityUid? exclude, IEntityManager entMan, SharedMindSystem mindSys)
+    {
+        mindSys.AddAliveHumans(minds, exclude);
```
SlamBamActionman, line 10: Is the choice to add `AddAliveHumans` an optimization decision, as opposed to using the existing `GetAliveHumans` method and transposing it in this method?

SlamBamActionman (commented):
Code seems to check out but I have some small questions in the comments.

Content.Shared/Mind/Filters/AliveHumansPool.cs
deltanedas: yes AddAliveHumans just adds to an existing hashset instead of allocating a new one every time

Content.Server/Mind/Filters/TargetObjectiveMindFilter.cs
deltanedas: misinput

it is a whitelist checked against objectives used to disqualify a target, so either is kinda correct

the thing being blacklisted is the target but the whitelist is for objectives

SlamBamActionman: Aight fix it and I'll mergghjghghjn

SlamBamActionman (approved):

PR #36033, opened 2025-03-23
Fix changelog entries with brackets crashing the client

### LINKED ISSUE #36031: Changelog does not check for bad symbols
## Description
if you add [ or ] in changelog and don't escape it with backslash, it crashes the client on opening changelog

**Reproduction**
1. Add entry to changelog containing square brackets
2. Open changelog ingame
3. Crash

**Screenshots**

**Additional context**
see #36029

Tayrtahn:
## About the PR

Fixes bug where square brackets in changelog entries would crash the client.

## Why / Balance

Fixes #36031.

## Technical details

Changelog text is now treated as plain text rather than markup, so invalid markup tags don't throw exceptions.
As a bonus, changelog author names now have special characters escaped so they can't crash the client or add formatting to the text.

## Media

Screenshot 2025-03-23 at 6 31 32 PM

## Requirements

### DIFF
--- a/Content.Client/Changelog/ChangelogTab.xaml.cs
+++ b/Content.Client/Changelog/ChangelogTab.xaml.cs
@@ -131,13 +131,13 @@ public sealed partial class ChangelogTab : Control
                     Margin = new Thickness(6, 0, 0, 0),
                 };
                 authorLabel.SetMessage(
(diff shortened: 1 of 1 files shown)


slarticodefast (approved):
Looks good to me!
Thanks for the quick bugfix!

PR #36044, opened 2025-03-24
More responsive votekick system (reduce timer and successive timeout)

K-Dynamic:
## About the PR

- Reduce default votekick timer from 60 to 45 seconds.
- Reduce successive votekick timeout (time between multiple votes) from 120 to 60 seconds.

## Why / Balance

Makes votekicking much more responsive - you currently have to wait the full 60 seconds and another 120 seconds for each successive votekick.

## Technical details

C# changes only
- changed default cvar values

## Media

N/A

## Requirements

## Breaking changes

N/A

**Changelog**

Don't know if this is player-facing or admin-facing.

### DIFF
--- a/Content.Shared/CCVar/CCVars.Vote.cs
+++ b/Content.Shared/CCVar/CCVars.Vote.cs
@@ -146,13 +146,13 @@ public sealed partial class CCVars
     ///     The delay for which two votekicks are allowed to be made by separate people, in seconds.
     /// </summary>
     public static readonly CVarDef<float> VotekickTimeout =
-        CVarDef.Create("votekick.timeout", 30f, CVar.SERVERONLY);
+        CVarDef.Create("votekick.timeout", 60f, CVar.SERVERONLY);
 
     /// <summary>
     ///     Sets the duration of the votekick vote timer.
     /// </summary>
     public static readonly CVarDef<int>
-        VotekickTimer = CVarDef.Create("votekick.timer", 20, CVar.SERVERONLY);
+        VotekickTimer = CVarDef.Create("votekick.timer", 45, CVar.SERVERONLY);
 
     /// <summary>
     ///     Config for how many hours playtime a player must have to get protection from the Raider votekick type when playing as an antag.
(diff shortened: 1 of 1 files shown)


ScarKy0:
Personally id make the timeout 60s and the duration 30s, but more admins will need to weight in on it
Alternatively, have it reset on a successful votekick
and 60s duration is about enough to observe if a player is misbehaving

SlamBamActionman:
I wouldn't end the vote early but a low CD is probably fine. We haven't seen raiders attempt to abuse the system yet and due to how speedy admins are at ensuring the votes are properly handled (<3 love ya'll) I doubt any singular risk of abuse has minor impact.

Content.Shared/CCVar/CCVars.Vote.cs
```diff
@@ -146,13 +146,13 @@ public static readonly CVarDef<int>
     ///     The delay for which two votekicks are allowed to be made by separate people, in seconds.
     /// </summary>
     public static readonly CVarDef<float> VotekickTimeout =
-        CVarDef.Create("votekick.timeout", 120f, CVar.SERVERONLY);
+        CVarDef.Create("votekick.timeout", 30f, CVar.SERVERONLY);
```
ScarKy0, line 149: 30s is a bit too drastic of a change, id say 60 would work better and if anything it can be adjusted

Content.Shared/CCVar/CCVars.Vote.cs
```diff
@@ -146,13 +146,13 @@ public static readonly CVarDef<int>
...
 
     /// <summary>
     ///     Sets the duration of the votekick vote timer.
     /// </summary>
     public static readonly CVarDef<int>
-        VotekickTimer = CVarDef.Create("votekick.timer", 60, CVar.SERVERONLY);
+        VotekickTimer = CVarDef.Create("votekick.timer", 20, CVar.SERVERONLY);
```
ScarKy0, line 155: Players still need time to observe the player and shouldnt feel rushed to decide whether to vote
I think it should be 45, maybe 30 but thats pushing it

ScarKy0 (changes requested):

Content.Shared/CCVar/CCVars.Vote.cs
SlamBamActionman: The last thing we want is for people to vote blindly. Even if you are sure, the other ghosts should have a chance to teleport over and see. >45 seems like a reasonable value.

SlamBamActionman (approved):

MilonPL (approved):

PR #36045, opened 2025-03-24
New Nuclear explosion type

B-Kirill:
## About the PR
- Adds new `Nuclear` explosion type, which, in addition to the usual explosion damage, causes radiation damage. It also has flags for configuring the removal of some objects (for more information, see the technical part of the PR).
- Changes nuke explosion type from `Default` to `Nuclear`.

## Why / Balance
After a nuclear explosion, a lot of objects remain, including a bunch of items, pipes, wires. Of course, this greatly damages performance (both on the server and on the client), especially when there are a lot of players. 
The new type of explosion not only allows you to remove unnecessary objects during an explosion without affecting other sources of explosions, but also makes it possible to modify the nuke and its explosion in the future, as well as add radiation damage for realism.

## Technical details
`Nuclear` explosion has the following flags:
-  `deleteTags: ["DisposalTube", "Item", "Cable", "WallMount","SpreaderIgnore"]`
-  `blacklistedTags: ["HighRiskItem"]`

Deletion works by checking the desired tag for the item and a `HashSet` ensures no duplicate deletions. Of course you can add other tags. (`SpreaderIgnore` tag is needed to remove small and directional fans.)
Since the objects I needed to delete initially didn't have the tags I needed, I had to create a `MapInitTagSystem.cs` that adds a tag to an object at an event of the same name (which is a fairly optimized way compared to others).

`Nuclear` explosion `damagePerIntensity`:
- `Heat: 5`
- `Blunt: 5`
- `Piercing: 5`
- `Radiation: 1`

## Media
Just for example, let's take the explosion at the Box station:

Before the `Nuclear` explosion was added

After adding the `Nuclear` explosion (Slightly outdated, high risk items are not deleted.)

## Notes  
This is my first PR to SS14, so any feedback greatly appreciated!

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: New Nuclear explosion that deletes cables, pipes, and other objects.

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
+++ b/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
@@ -28,6 +28,7 @@ namespace Content.Server.Explosion.EntitySystems;
 public sealed partial class ExplosionSystem
 {
     [Dependency] private readonly FlammableSystem _flammableSystem = default!;
+    private HashSet<EntityUid> _processedForDeletion = new();
 
     /// <summary>
     ///     Used to limit explosion processing time. See <see cref="MaxProcessingTime"/>.
@@ -87,6 +88,8 @@ public sealed partial class ExplosionSystem
     /// </summary>
     public override void Update(float frameTime)
     {
+        _processedForDeletion.Clear();
+
         if (_activeExplosion == null && _explosionQueue.Count == 0)
             // nothing to do
             return;
@@ -209,6 +212,7 @@ public sealed partial class ExplosionSystem
         DamageSpecifier damage,
         MapCoordinates epicenter,
         HashSet<EntityUid> processed,
+        ExplosionPrototype type,
         string id,
         float? fireStacks,
         EntityUid? cause)
@@ -230,7 +234,7 @@ public sealed partial class ExplosionSystem
         // process those entities
         foreach (var (uid, xform) in list)
         {
-            ProcessEntity(uid, epicenter, damage, throwForce, id, xform, fireStacks, cause);
+            ProcessEntity(uid, epicenter, damage, throwForce, type, id, xform, fireStacks, cause);
         }
 
         // process anchored entities
@@ -240,7 +244,7 @@ public sealed partial class ExplosionSystem
         foreach (var entity in _anchored)
         {
             processed.Add(entity);
-            ProcessEntity(entity, epicenter, damage, throwForce, id, null, fireStacks, cause);
+            ProcessEntity(entity, epicenter, damage, throwForce, type, id, null, fireStacks, cause);
         }
 
         // Walls and reinforced walls will break into girders. These girders will also be considered turf-blocking for
@@ -276,7 +280,7 @@ public sealed partial class ExplosionSystem
         {
             // Here we only throw, no dealing damage. Containers n such might drop their entities after being destroyed, but
             // they should handle their own damage pass-through, with their own damage reduction calculation.
-            ProcessEntity(uid, epicenter, null, throwForce, id, xform, null, cause);
+            ProcessEntity(uid, epicenter, null, throwForce, type, id, xform, null, cause);
         }
 
         return !tileBlocked;
@@ -311,6 +315,7 @@ public sealed partial class ExplosionSystem
         DamageSpecifier damage,
         MapCoordinates epicenter,
         HashSet<EntityUid> processed,
+        ExplosionPrototype type,
         string id,
         float? fireStacks,
         EntityUid? cause)
@@ -329,7 +334,7 @@ public sealed partial class ExplosionSystem
         foreach (var (uid, xform) in state.Item1)
         {
             processed.Add(uid);
-            ProcessEntity(uid, epicenter, damage, throwForce, id, xform, fireStacks, cause);
+            ProcessEntity(uid, epicenter, damage, throwForce, type, id, xform, fireStacks, cause);
         }
 
         if (throwForce <= 0)
@@ -343,7 +348,7 @@ public sealed partial class ExplosionSystem
 
         foreach (var (uid, xform) in list)
         {
-            ProcessEntity(uid, epicenter, null, throwForce, id, xform, fireStacks, cause);
+            ProcessEntity(uid, epicenter, null, throwForce, type, id, xform, fireStacks, cause);
         }
     }
 
@@ -439,11 +444,13 @@ public sealed partial class ExplosionSystem
         MapCoordinates epicenter,
         DamageSpecifier? originalDamage,
         float throwForce,
+        ExplosionPrototype type,
         string id,
         TransformComponent? xform,
         float? fireStacksOnIgnite,
         EntityUid? cause)
     {
+        var isNuclear = type.ID == "Nuclear";
         if (originalDamage != null)
         {
             GetEntitiesToDamage(uid, originalDamage, id);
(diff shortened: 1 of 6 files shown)


ScarKy0:
Could probably be added to The Nuclear Option as well

NoElkaTheGod:
Looks very good, except for the changelog. Technical information is not needed, you may remove it and leave something like "Nuclear explosions now delete cables, pipes, and other obects."

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
```diff
@@ -466,6 +473,45 @@ private void ProcessEntity(
...
+
+                if (type.DeleteCables)
+                {
+                    if (_entMan.EntityExists(entity) && !_processedForDeletion.Contains(entity) && !_entMan.IsQueuedForDeletion(entity) && _cableQuery.HasComponent(entity))
+                    {
+                        //Logger.Info($"Deleting cable: {ToPrettyString(entity)} | DeleteCables: {type.DeleteCables} | Type: {type.ID}");
+                        _processedForDeletion.Add(entity);
+                        _entMan.QueueDeleteEntity(entity);
+                    }
+                }
```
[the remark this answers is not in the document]
B-Kirill: If it's all about optimization, I can rewrite this block of code for better performance by reducing the number of checks. But about the hardcode of the components, I do not know a better implementation method, even in theory.

Everturning:
Gone. Reduced to atoms.

Pixeltheaertist:
Are you still protected inside of Fridges?

RedBookcase:
> Are you still protected inside of Fridges?

Fridges had their nuke explosion resistance removed because it was causing unintended fun.

B-Kirill:
Well, I've updated the PR description, now I'll deal with the problem..

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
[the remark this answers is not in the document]
B-Kirill: Fixed it in commit edc3580 (forgot to tell you). Now objects are deleted by having a tag, the list of which is set in the explosion settings.

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
```diff
@@ -463,9 +470,44 @@ private void ProcessEntity(
...
+                        if (_tagSystem.HasTag(entity, blacklistTag))
+                        {
+                            //Logger.Info($"Skipped deletion for {ToPrettyString(entity)} due to blacklisted tag: {blacklistTag}");
+                            skipDeletion = true;
+                            break;
+                        }
+                    }
+                        if (skipDeletion)
+                        continue;
+                }
```
slarticodefast, line 487: This is basically an `EntityWhitelist` with extra steps. Not sure what the performance impact would be here.

Content.Server/Tags/MapInitTagSystem.cs
```diff
@@ -0,0 +1,45 @@
+using Content.Server.Construction.Components;
+using Content.Server.Power.Components;
+using Content.Shared.Item;
+using Content.Shared.Tag;
+using Content.Shared.Wall;
+using Robust.Shared.GameObjects;
+
+namespace Content.Server.Tag;
+public sealed class MapInitTagSystem : EntitySystem
```
slarticodefast, line 9: This whole system is not needed since you can just directly add a tag to any protoype with the `TagComponent`.
Also it does not seem to be used anywhere as the component and subscription are missing.

Content.Shared/Explosion/ExplosionPrototype.cs
```diff
@@ -110,6 +110,12 @@ public sealed partial class ExplosionPrototype : IPrototype
     [DataField("fireStates")]
     public int FireStates = 3;
 
+    [DataField("deleteTags")]
+    public HashSet<string> DeleteTags { get; } = new();
+
+    [DataField("blacklistedTags")]
+    public HashSet<string> BlacklistedTags { get; } = new();
```
slarticodefast, line 117: Getters and setters for datafields are no longer needed.
Also just use `[DataField]`, the name gets automatically converted to camelCase.

slarticodefast (changes requested):
Again, I'm not sure about the performance impact when checking a tag for every single item caught in a station-wide explosion. Maybe another maintainer can chime in here, as I cannot think of any elegant solutions without investigating the code more.

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
B-Kirill: In my tests, I noticed almost no performance difference between a `Default` and a `Nuclear` explosion. It even seemed to me that it was better, although who knows my PC.
Still, I don't know a better way, but for now I'll deal with other comments.

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
```diff
@@ -28,6 +28,8 @@ namespace Content.Server.Explosion.EntitySystems;
 public sealed partial class ExplosionSystem
 {
     [Dependency] private readonly FlammableSystem _flammableSystem = default!;
+    [Dependency] private readonly IEntityManager _entMan = default!;
+    private HashSet<EntityUid> _processedForDeletion = new();
```
metalgearsloth, line 32: EntityManager is already a dependency on systems, it's why they're systems.

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
metalgearsloth: Did you use dottrace API? You can't just look at the FPS counter.

B-Kirill: Well, fair enough...

Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
```diff
@@ -463,9 +470,44 @@ private void ProcessEntity(
...
+                            _entMan.QueueDeleteEntity(entity);
+                            break;
+                        }
+                    }
+                }
+                if (!_entMan.Deleted(entity) && !_entMan.IsQueuedForDeletion(entity) && !_processedForDeletion.Contains(entity))
+                {
+                    // TODO EXPLOSIONS turn explosions into entities, and pass the the entity in as the damage origin.
+                    _damageableSystem.TryChangeDamage(entity, damage * _damageableSystem.UniversalExplosionDamageModifier, ignoreResistances: true);
+                }
```
metalgearsloth, line 510: Checking for tags and container is really expensive, plus you're not even re-using the components here.

Using damage thresholds is significantly more convenient code-wise and performant just tooling it is bad.

Content.Shared/Explosion/ExplosionPrototype.cs
B-Kirill:
```
[DataField]
public HashSet<ProtoId<TagPrototype>> deleteTags = new();

[DataField]
public HashSet<ProtoId<TagPrototype>> blacklistedTags = new();
```
Something like that?

slarticodefast: yep, but like sloth mentionend, using tags is bad for performance

Content.Server/Tags/MapInitTagSystem.cs
B-Kirill: @slarticodefast 
I apologize for the inconvenience — I wanted to know if you received my previous question about tags?

slarticodefast: Hmm, I get your point. But I think
- having two different methods to add tags to objects that are inherited differently will make yaml inheritance a total mess over time, especially since you won't be able to see one of them in the prototype.
- if we have to add a tag to nearly every entity in the game then there is something fundamentally wrong with the code design or api tooling.
- Like sloth said, checking every single entity caught in the  explosion is really expensive

B-Kirill:
@slarticodefast 
Hi again!
(I hope I'm not annoying you yet... ` ¯\_(ツ)_/¯`)

**Here’s a breakdown of the changes I’ve made (key points only):**

- Feedback fixes: Addressed all your (and not only yours) recommendations (deleted unnecessary dependencies, adjusted code structure, etc.).

- Optimizations: Squeezed out every drop of efficiency possible:

  - Minimized tag checks.

  - Optimized hashing for faster lookups.

  - Skipped redundant operations where possible.

  - Default explosions: My code now just skips processing them. (Not sure why I didn’t think of this earlier…)

**Performance concerns:**
Your worries about heavy performance impact might’ve been a bit overblown (_or maybe I just optimized it well, who knows_). I took performance measurements and here are the results:

- Default explosions: My code adds 0–0.0001ms to `ProcessEntity`.

- Nuclear explosions: Ranges from 0.0002ms to 0.2016ms in rare cases (still ≤20% of `ProcessEntity` time).

**Context:** Nukes don’t detonate every round, and the server performance gain from deleting debris far outweighs these costs. This happens at the end of the round, and the explosion of the nuclear bomb itself causes lags, so my code won't change much. - I’d call this a win.

**The Tags Problem:**

Agreed: MapInitTagSystem.cs feels hacky and confusing. But here’s the catch:

- Items: `BaseItem` tags get overridden by child entities. Adding `Item` tag to every child at .yml manually isn’t feasible (there are hundreds).

- WallMounts: No **BaseWallMount** equivalent exists, and adding tags to all of them would be a nightmare.

- Cables and DisposalTubes (instead of Constructions) - in this case, I can add tags to .yml manually, as there are few of them. However, I'm not doing this yet, in case there's a normal way to add tags.


Yes, I still insist on using tags. Everything turned out to be not so bad with their performance, and their convenience in adding to the deletion list covers everything else. It remains only to deal with their adequate addition... But here I can only count on your advice.

Looking forward to your thoughts!

B-Kirill:
@slarticodefast 
Hi! Just checking in - want to make sure this PR isn't lost in the shuffle. Please let me know if you need anything from my side.

Issue #36020, opened 2025-03-23
New EmbeddableProjectile heisentest


slarticodefast:
https://github.com/space-wizards/space-station-14/actions/runs/14017582708/attempts/1?pr=36019
Seen this one happen a few times lately.

```
A total of 1 test files matched the specified pattern.
  Failed TestConnect [300 ms]
  Error Message:
   Multiple failures or warnings in test:
  1) CLIENT: 0.296s [WARN] system.transform: Failed to attach entity to map or grid. Entity: (survival knife (18568/n89976, SurvivalKnife)). Trace:    at System.Environment.get_StackTrace()
   at Robust.Shared.GameObjects.SharedTransformSystem.AttachToGridOrMap(EntityUid uid, TransformComponent xform) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/Systems/SharedTransformSystem.Component.cs:line 1365
   at Content.Shared.Projectiles.SharedProjectileSystem.EmbedDetach(EntityUid uid, EmbeddableProjectileComponent component, Nullable`1 user) in /home/runner/work/space-station-14/space-station-14/Content.Shared/Projectiles/SharedProjectileSystem.cs:line 148
   at Content.Shared.Projectiles.SharedProjectileSystem.DetachAllEmbedded(Entity`1 container) in /home/runner/work/space-station-14/space-station-14/Content.Shared/Projectiles/SharedProjectileSystem.cs:line 184
   at Content.Shared.Projectiles.SharedProjectileSystem.OnEmbeddableTermination(Entity`1 container, EntityTerminatingEvent& args) in /home/runner/work/space-station-14/space-station-14/Content.Shared/Projectiles/SharedProjectileSystem.cs:line 174
   at Robust.Shared.GameObjects.EntityEventBus.<>c__DisplayClass57_0`2.<SubscribeLocalEvent>g__EventHandler|0(EntityUid uid, IComponent comp, TEvent& args) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 324
   at Robust.Shared.GameObjects.EntityEventBus.EntDispatch(EntityUid euid, Type eventType, Unit& args) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 633
[...]

PR #36048, opened 2025-03-24
Fix error when disconnecting the client while a projectile is embedded

Tayrtahn:
## About the PR

Fixes a bug where shutting down the client or server while an embeddable projectile is embedded into another entity caused an error.

Adds a test that throws an embeddable projectile (a survival knife) into a target (a glass airlock), then disconnects the client from the server, and a test that just deleting the target causes the projectile to detach and not be deleted.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/36020.

## Technical details

Added a check that the projectile's grid and map aren't terminating before trying to reparent.

## Requirements


slarticodefast (approved):
LGTM!

PR #36049, opened 2025-03-24
Allow NukeOps test to function with multiple RuleGrids

### LINKED ISSUE #35868: Wizard antag causes heisentest in TryStopNukeOpsFromConstantlyFailing

Tayrtahn:
## About the PR

`TryStopNukeOpsFromConstantlyFailing` no longer fails when it detects multiple `RuleGridsComponent`s.

## Why / Balance

Fixes #35868.

## Technical details

Rather than getting all entities with `RuleGridsComponent`s and using `Single()` to get the one for the NukeOps rule, we now use `GetComponent` to grab the `RuleGridsComponent` from the entity that has the `NukeopsRuleComponent`.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/GameRules/NukeOpsTest.cs
+++ b/Content.IntegrationTests/Tests/GameRules/NukeOpsTest.cs
@@ -152,14 +152,10 @@ public sealed class NukeOpsTest
             Assert.That(roleSys.MindGetAllRoleInfo(mindCrew).Any(x => nukeroles.Contains(x.Prototype)), Is.False);
         }
 
-        var ruleGridComps = entMan.AllComponents<RuleGridsComponent>();
-        Assert.That(ruleGridComps, Has.Length.EqualTo(1),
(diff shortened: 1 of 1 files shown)


slarticodefast (approved):
Looks good to me!

PR #36052, opened 2025-03-24
Cleanup: ExtensionCableSystem

Tayrtahn:
## About the PR

Removes all warnings in `ExtensionCableSystem` (10 warnings, by my count).

## Why / Balance

## Technical details

- Changes a bunch of private methods of the system to take `Entity<T>`s as parameters rather than just components.
- Makes `ExtensionCableReceiverComponent.Provider` an `Entity<ExtensionCableProviderComponent>?` instead of an `ExtensionCableProviderComponent?`
- Makes `ExtensionCableProviderComponent.LinkedReceivers` a `List<Entity<ExtensionCableReceiverComponent>>` instead of a `List<ExtensionCableReceiverComponent>`
- Replaces calls to `MapGridComponent.GetCellsInSquareArea` with their `SharedMapSystem` equivalents.

## Requirements

## Breaking changes

- `ExtensionCableReceiverComponent.Provider` is now an `Entity<ExtensionCableProviderComponent>?` instead of an `ExtensionCableProviderComponent?`
- `ExtensionCableProviderComponent.LinkedReceivers` is now a `List<Entity<ExtensionCableReceiverComponent>>` instead of a `List<ExtensionCableReceiverComponent>`

### DIFF
--- a/Content.Server/Power/EntitySystems/ExtensionCableSystem.cs
+++ b/Content.Server/Power/EntitySystems/ExtensionCableSystem.cs
@@ -8,6 +8,8 @@ namespace Content.Server.Power.EntitySystems
 {
     public sealed class ExtensionCableSystem : EntitySystem
     {
+        [Dependency] private readonly SharedMapSystem _map = default!;
+
         public override void Initialize()
         {
             base.Initialize();
@@ -126,7 +128,7 @@ namespace Content.Server.Power.EntitySystems
             if (!TryComp(xform.GridUid, out MapGridComponent? grid))
                 yield break;
 
-            var nearbyEntities = grid.GetCellsInSquareArea(coordinates, (int) Math.Ceiling(range / grid.TileSize));
+            var nearbyEntities = _map.GetCellsInSquareArea(xform.GridUid.Value, grid, coordinates, (int)Math.Ceiling(range / grid.TileSize));
 
             foreach (var entity in nearbyEntities)
             {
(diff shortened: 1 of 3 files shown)


Content.Server/Power/EntitySystems/ExtensionCableSystem.cs
```diff
@@ -162,88 +164,88 @@ public void SetReceiverReceptionRange(EntityUid uid, int range, ExtensionCableRe
 
             if (provider != null)
             {
-                RaiseLocalEvent(provider.Owner, new ReceiverDisconnectedEvent((uid, receiver)), broadcast: false);
-                provider.LinkedReceivers.Remove(receiver);
+                RaiseLocalEvent(provider.Value, new ReceiverDisconnectedEvent((uid, receiver)), broadcast: false);
+                provider.Value.Comp.LinkedReceivers.Remove((uid, receiver));
```
slarticodefast, line 168: Just to make sure, does this correctly remove it from the list?
Since the tuple technically is a new instance of an object, and thus not exactly the same
Or does Entity<Component> implement IEquatable for this usecase?

Tayrtahn: It was a little tricky figuring out how to actually get something to call this code! The method is only used once in the entire codebase, inside an integration test, and in that particular case, `provider` is null and this code block is skipped.

But I tweaked the test locally and yep, it gets removed from the list!

Looking into why, `Entity<T>` is a `record struct`, and [record types automatically implement value equality](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/types/records#value-equality).

PR #36054, opened 2025-03-24
Fix comment

Partmedia:
## About the PR
Fix an incorrect comment. The relevant code is in `PuddleSystem.Evaporation.cs`:

```
            var reagentTick = evaporation.EvaporationAmount * EvaporationCooldown.TotalSeconds;
```

Since this multiplies `EvaporationAmount` by seconds, the rate is in amount/second rather than amount/tick.


slarticodefast (approved):
Thanks for the cleanup!

PR #36055, opened 2025-03-24
[April fools] April fools!

qwerltaz:
## About the PR

give everyone liar accent.

change lie chance from 15% to ~~5%~~ 1%.

## Why / Balance

to enforce april fooling

### Media
image

### technical details
added comp at the top of comp list for easier review

### DIFF
--- a/Resources/Prototypes/Accents/word_replacements.yml
+++ b/Resources/Prototypes/Accents/word_replacements.yml
@@ -451,7 +451,7 @@
 
 - type: accent
   id: liar
-  replacementChance: 0.02
+  replacementChance: 0.01
   wordReplacements:
     liar-word-1: liar-word-replacement-1
     liar-word-2: liar-word-replacement-2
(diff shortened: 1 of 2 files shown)


FairlySadPanda:
You'll need a changelist markup - re-read the template.

Instinct is that this fucks directly with players too much and would discourage people playing.

qwerltaz:
> You'll need a changelist in the markdown 

Not sure what you mean by that. changelog? I'm not sure about the changelog enforcement for the april fools branch, I don't know if it even works there.

> fucks directly with players too much

I was considering making the lie chance 1-3%, which is up to 15 times less than the original. That's totally a possibility and might make the effect much less frequent and possibly funnier, but that's 1 in 100 messages compared to my suggested 1 in 20. And keep in mind a lot of messages don't contain keywords changed by the accent so the accent won't do anything in that case. I would like to hear from maintainers if this percentage needs tweaking, as they can probably make the best guess.

FairlySadPanda:
Not sure how the release will work but the rest of the April Fools PRs all have a CL.

qwerltaz:
Changed lie chance from 5% to 1%, it was in fact probably still high. Now 1 in 100 messages will be lies.

slarticodefast (approved):
1% seems fine to me :)

MilonPL (approved):

Issue #35877, opened 2025-03-16
Nuke Disk permanently lost to nullspace, bread wand RR


iaada:
## Description

This bug occurs from polymorph, specifically the Magic Bread Wand. When someone gets polymorphed they take the disk with them, which is mostly fine so long as they un-polymorph properly. However if someone gets breaded and then falls into a recycler, they don't get un-polymorphed like normal and instead get deleted and kicked out to ghost, while their body and items get left in nullspace.

Additionally, if bread is an ingredient in a microwave recipe (currently not the case) it will also leave the original in nullspace.

**Reproduction**

- Put a nuke disk (or any item) in an Urist's hand.
- Polymorph the Urist with a bread wand.
- Throw Urist McDough in a recycler.

**Screenshots**

![Image](https://github.com/user-attachments/assets/597fdc34-33d0-4274-8d19-ff6148331aab)

**Additional context**

deltanedas:
why can bread even be recycled

Djungelskog2:
Would throwing the bread into space also effectively get rid of the disc

keronshb:
Ideally someone being polymorphed should just force the disk to respawn

ArtisticRoomba:
People could abuse this by wehbombing the captain

NoElkaTheGod:
> People could abuse this by wehbombing the captain

Not really worthwhile. The disk will spawn at a random tile, so the chance of finding it yourself is pretty low. Also, if you can wehbomb the captain, you may as well simply bomb the captain, which would be a much more consistent way of getting the nuke disk.

iaada:
> I believe cutting the bread also RRs the victim, so it'd produce the same result.

Cutting or eating the bread will cause the Urist to un-polymorph.

However if bread is part of a microwave recipe (currently not the case) the un-polymorph won't happen then.

PR #36058, opened 2025-03-24
Fix nuke disk getting lost when polymorphed holder is deleted

Tayrtahn:
## About the PR

If someone holding the nuke disk is polymorphed and their polymorphed self is deleted (by falling into a recycler, by admin command, and possibly by other means), the nuke disk will respawn like normal.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/35877.

## Technical details

Added a handler for `EntityTerminatingEvent` to `PolymorphSystem`. The original entity is now queued for deletion when the polymorph substitute is deleted, so the nuke disk in the original's inventory will be deleted.
`PolymorphSystem.Revert` now sets `PolymorphedEntityComponent.Parent` to null, so that successfully reverted polymorphs aren't deleted.

This may also have some performance benefits, overall, since the original entities of deleted polymorphs are now deleted, instead of being left to hang around on a paused map for eternity.

A `RevertOnDelete` property has been added to `PolymorphPrototype` which can be used to make the polymorphed entity revert in situations where it is about to be deleted. This is set to false by default, because it has balance implications (throwing bread-polymorphed players into the recycler is no longer deadly, admins deleting polymorphed players just reverts them).

## Media

https://github.com/user-attachments/assets/48872c3f-1a4a-4ff7-bb62-8788cc5b068b

## Requirements

## Breaking changes

`PolymorphedEntityComponent.Parent` is now nullable.

**Changelog**

### DIFF
--- a/Content.Server/Polymorph/Systems/PolymorphSystem.cs
+++ b/Content.Server/Polymorph/Systems/PolymorphSystem.cs
@@ -314,6 +314,10 @@ public sealed partial class PolymorphSystem : EntitySystem
         var uidXform = Transform(uid);
         var parentXform = Transform(parent.Value);
 
+        // Don't swap back onto a terminating grid
+        if (TerminatingOrDeleted(uidXform.ParentUid))
+            return null;
+
         if (component.Configuration.ExitPolymorphSound != null)
             _audio.PlayPvs(component.Configuration.ExitPolymorphSound, uidXform.Coordinates);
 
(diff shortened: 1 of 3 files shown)


Content.Shared/Polymorph/PolymorphPrototype.cs
```diff
@@ -97,6 +97,12 @@ public sealed partial record PolymorphConfiguration
     [DataField(serverOnly: true)]
     public bool RevertOnDeath = true;
 
+    /// <summary>
+    /// Whether or not the polymorph reverts when the entity is deleted.
+    /// </summary>
+    [DataField(serverOnly: true)]
+    public bool RevertOnDelete;
```
keronshb, line 104: This should be true by default.

Tayrtahn: Sure thing.

Content.Server/Polymorph/Systems/PolymorphSystem.cs
```diff
@@ -295,33 +306,39 @@ private void OnDestruction(Entity<PolymorphedEntityComponent> ent, ref Destructi
...
 
         if (component.Configuration.Inventory == PolymorphInventoryChange.Transfer)
         {
-            _inventory.TransferEntityInventories(uid, parent);
+            _inventory.TransferEntityInventories(uid, parent.Value);
             foreach (var held in _hands.EnumerateHeld(uid))
             {
                 _hands.TryDrop(uid, held);
-                _hands.TryPickupAnyHand(parent, held, checkActionBlocker: false);
+                _hands.TryPickupAnyHand(parent.Value, held, checkActionBlocker: false);
```
MilonPL, line 341: Please just use pattern matching instead of adding `.Value` everywhere.

PR #35981, opened 2025-03-21
Fix thermoregulation

ArtisticRoomba:
## About the PR
Fixes mob thermoregulation.

Right now `thermalRegulationTemperatureThreshold` is set too high. This controls how hot or cold it needs to get for a mob to start responding to their environment by warming up or cooling down. It was previously set to 25 K, when in reality human bodies start to respond to their environment (sweat, shiver) at around 1.4 K.

During testing I also noticed that post-exposure to the cold of space is no longer as painful, you heat up now! No more walking around for 10 minutes after you accidentally step into a spaced area!

Fixes #35979.

## Why / Balance
People are currently cooking in hardsuits on Vulture. With the buffer at 25 K, you start taking damage before thermoregulation kicks in, and thermoregulation doesn't even cool you down all the way instantly, it cools you down until you stop thermoregulating. So you're just constantly taking damage.

## Technical details
Funny YAML change.

## Media
I no longer cook alive in hardsuits
Content Client_19L9LWSSfD

Urists fry on the ground but it's not as crazy as before, as expected.
Content Client_WIDG38jaFQ

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
:cl:
- fix: Fixed organic thermoregulation. Humans, arachnids, moths, etc. will maintain their desired body temperature better.

PR #35971, opened 2025-03-20
Fix thermal regulation

slarticodefast:
## About the PR
Taken from https://github.com/space-wizards/space-station-14/pull/35020, the credit goes to shibechef.
That PR included other changes and the author closed it themselves.
Turns out sweating and shivering never worked before, this PR fixes it.

## Why / Balance
you don't turn into ash as easily anymore

## Technical details
The guard statement condition was inversed, meaning you were only sweating or shivering when the temperature difference was below the threshold, not above it. This means you can now easier recover from extreme temperatures.

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
none

**Changelog**

:cl: shibechef
- fix: Bodies can now thermally regulate themselves again.

PR #36062, opened 2025-03-24
fix thermal regulation, this time for real, I promise

slarticodefast:
## About the PR
BEHOLD, GRAPHS:
thermal_1
This one was before my bugfix in https://github.com/space-wizards/space-station-14/pull/35971
You can see how the condition for sweating and shivering was inverted, causing you to ash or freeze at extreme temperatures because they shut down.

thermal_2
This one was after my bugfix. sweating and shivering now kicks in as proper second stage of thermal regulation when hitting the threshold of 25° difference to normal temperature. The problem is that some hardsuits like the atmos suit got near perfect heat isolation and without sweating you increase in temperature by default, meaning you would keep heating up until you hit the 25° difference threshold and sweating kicks in. At this point you already get the temperature alert.

https://github.com/space-wizards/space-station-14/pull/35981 already fixed this by reducing the sweating/shivering threshold, but I originally objected reducing it too much because we don't have such realistic temperatures in this game (you can easily survive having 200°C body temperature) and sweating/shivering should serve as a second stage of regulation rather than being active at all time.

But it turns out that the 10° difference still gives you the heat warning once you reach that temperature inside an atmos suit, so we now reduce it to the originally proposed, more realistic value, giving this result.

thermal_3

The downside is that makes it now basically the same as implicit heat regulation, which is active at all times. But we can address that by actually implementing the currently unused `CanShiver` and `CanSweat` action blockers, which are currently always true. An interesting mechanic for this could be requiring requiring hunger and thirst to do so.

## Why / Balance
Fixes https://github.com/space-wizards/space-station-14/issues/33596
Supercedes https://github.com/space-wizards/space-station-14/pull/34972

## Technical details
change numbers

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- fix: Adjusted thermal regulation again so you no longer overheat in hardsuits.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/base.yml
@@ -259,7 +259,7 @@
       280: 0.6
       260: 0.4
   - type: ThermalRegulator
-    metabolismHeat: 800
+    metabolismHeat: 500
     radiatedHeat: 100
     implicitHeatRegulation: 800
     sweatHeatRegulation: 2000
--- a/Resources/Prototypes/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/base.yml
@@ -259,9 +259,9 @@
       280: 0.6
       260: 0.4
   - type: ThermalRegulator
-    metabolismHeat: 500
+    metabolismHeat: 800
     radiatedHeat: 100
-    implicitHeatRegulation: 800
+    implicitHeatRegulation: 500
     sweatHeatRegulation: 2000
     shiveringHeatRegulation: 2000
     normalBodyTemperature: 310.15
--- a/Resources/Prototypes/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/base.yml
@@ -261,11 +261,11 @@
   - type: ThermalRegulator
     metabolismHeat: 800
     radiatedHeat: 100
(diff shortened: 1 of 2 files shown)


ArtisticRoomba:
i was close enough :godo:

ScarKy0:
We could also make hardsuits not insulate heat perfectly, which could be interesting but probably not very realistic?

slarticodefast:
> i was close enough :godo:

Yep, you were right 😄 
But now I have to make all the other temeprature values realistic as well or I will be annoyed forever.

ArtisticRoomba:
> But now I have to make all the other temperature values realistic as well or I will be annoyed forever.

Don't we have to prune the 20 morbillion magic numbers and constants for thermal transfer between organisms and the atmosphere first or something like that?

> We could also make hardsuits not insulate heat perfectly, which could be interesting but probably not very realistic?

Would be interesting, however I think this would fit best into a proper implementation under Powered Hardsuits (https://github.com/space-wizards/docs/pull/335). I highlighted that active thermal regulation could require hardsuit power, with higher temps requiring more power. This would mechanically discourage sitting in fires forever or intentionally causing massive fires (which the crew can do basically nothing about) without serious preparation or care.

slarticodefast:
> Don't we have to prune the 20 morbillion magic numbers and constants for thermal transfer between organisms and the atmosphere first or something like that?

Yeah, pretty much. But once we have actual energy conservation everywhere it should all fit together hopefully.

PR #36076, opened 2025-03-25
Rework toggling logic for most atmos machines and add alt. verbs

LaCumbiaDelCoronavirus:
## About the PR
Changes logic for some atmos devices to use another component when handling being off/on, and standardizes an optional alt-interact verb on them
This component isn't exclusive to atmos devices, the PR just adds it to alot of them

All atmos devices changed also now support being toggled with an alt verb, which means you can alt-interact (press Alt+E (default bind)) with them to turn them on/off

List of devices changed:
- Pressure pumps
- Volume pumps
- Mixers
- Filters
- Thermomachines (Heaters & Freezers)

## Why / Balance
Nice quality of life feature, and cleans up a good bit of code

## Technical details
Adds:
- ToggleableComponent & ToggleableSystem:
It handles logic on atmos things it's implemented in for switching between on and off states. Option for an alt interaction verb to toggle it is present.
- ToggleableEnabledEvent and ToggleableDisabledEvent

Also makes thermomachines use this component instead of relying on toggling ApcPowerReceiver with PowerSwitch, since that was silly, so I had to implement proper visuals for it too considering it used PowerVisuals before.

Renames 2 vars for events (not the classes) in AtmosDeviceSystem.cs to avoid confusion with this system.

## Media
https://youtube.com/watch?v=EBDbIrxgmZ0
Note that pressure pumps having improperly updating UI as shown is unrelated to this PR.

## Requirements

**Changelog**
:cl:
- add: You can now alt-interact gas pumps, filters and mixers to toggle them on/off.


LaCumbiaDelCoronavirus:
apparently space heaters weren't excluded... and thermomachines are all supposed to have the toggleable comp on spawn. Oops ill fix that later.

Content.Server/Atmos/Piping/Unary/EntitySystems/GasThermoMachineSystem.cs
```diff
@@ -87,9 +117,9 @@ private void OnThermoMachineUpdated(EntityUid uid, GasThermoMachineComponent the
             float dQ = thermoMachine.HeatCapacity * thermoMachine.Cp * args.dt;
 
             // Clamps the heat transferred to not overshoot
-            float Cin = _atmosphereSystem.GetHeatCapacity(heatExchangeGasMixture, true);
+            float cin = _atmosphereSystem.GetHeatCapacity(heatExchangeGasMixture, true);
             float dT = targetTemp - temp;
-            float dQLim = dT * Cin;
+            float dQLim = dT * cin;
```
Partmedia, line 122: Why was this change made?

Partmedia:
Thank you for your PR. I have two main questions:
- Why is this `AtmosToggleableComponent` and not a change to the `PowerSwitchComponent`? That is, why do we need yet another component that duplicates this functionality, specific to atmos, that can't be in the power switch?
- `Enable` in a lot of atmos code is a misnomer -- and may or may not be a "power switch" that is directly settable. This is why power-consuming atmos devices also check the `ApcPowerReceiver`. Did you audit each of the use cases to make sure you're not introducing state bugs?

Content.Server/Atmos/Piping/Unary/EntitySystems/GasThermoMachineSystem.cs
LaCumbiaDelCoronavirus: To fix a formatting warning, I assume this has no unintended effect?

LaCumbiaDelCoronavirus:
> * Why is this `AtmosToggleableComponent` and not a change to the `PowerSwitchComponent`? That is, why do we need yet another component that duplicates this functionality, specific to atmos, that can't be in the power switch?

PowerSwitchComponent completely turns off/on power for something, making it look the same if it's either simlly unpowered or powered, but switched off, which can get confusing. It was also _only_ used for thermomachines anyway; pumps, filters and mixers had logic that didn't use PowerSwitch.

This new component is meant to standardise the functionality for alternate verbs (like how PowerSwitchComponent did for thermomachines and does for other things that use it) while also allowing custom logic, e.g. updating appearance of atmos devices when they are enabled/disabled.
Basically to reduce copy and pasting code.

> * `Enable` in a lot of atmos code is a misnomer -- and may or may not be a "power switch" that is directly settable. This is why power-consuming atmos devices also check the `ApcPowerReceiver`. Did you audit each of the use cases to make sure you're not introducing state bugs?

For this reason, i intentionally specified in the code summary for the Enabled var in ATC, that by being set as enabled/disabled, it refers to if it is set to operate or not by the user; an unpowered entity thats set as enabled won't work if it requires power.

They do still check ApcPowerReceiver the same as they did (if they used to check it in the first place anyway) _and_ also check for whether the AtmosToggleableComponent is enabled; they work as intended.

Content.Server/Atmos/Piping/Unary/EntitySystems/GasThermoMachineSystem.cs
LaCumbiaDelCoronavirus: > This is not a formatting error. This stands for "specific heat capacity of the input" which in most physics textbooks is represented by C , and C i n just means at the input.

Oops, I just changed it to lowercase due to the warning, never noticed that was intended.

Partmedia:
> PowerSwitchComponent completely turns off/on power for something, making it look the same if it's either simply unpowered or powered, but switched off, which can get confusing. It was also _only_ used for thermomachines anyway; pumps, filters and mixers had logic that didn't use PowerSwitch.

This can be fixed e.g. in the examine text, or by adding more flags in the component that entities with `ApperanceComponent` can optionally use.

My main point is that this feels like it's duplicating functionality that should be in one place in multiple places that doesn't actually have to do with atmos. What if someone wanted similar functionality for a chem machine?

> This new component is meant to standardise the functionality for alternate verbs (like how PowerSwitchComponent did for thermomachines and does for other things that use it) while also allowing custom logic, e.g. updating appearance of atmos devices when they are enabled/disabled. Basically to reduce copy and pasting code.

But now we're duplicating component logic. Is there now going to be a `BotanyPowerSwitchComponent`?

> Edit: AtmosToggleableComponent isn't (and shouldn't be) exclusive to only be used by atmos stuff, it's just how I named it. Nothing stops you from using it for something else as long as you implement it properly.

Then at minimum, let's not name it that to avoid that implication and move it outside of Atmos code.

I still think you should spend some time evaluating whether a new component is really necessary.

LaCumbiaDelCoronavirus:
> My main point is that this feels like it's duplicating functionality that should be in one place in multiple places that doesn't actually have to do with atmos. What if someone wanted similar functionality for a chem machine? 

Do you mean that the problem is with it specifically not having anything to do with atmos, or that it needlessly duplicates functionality in too many places? Though I'll assume the latter and explain my decisions:

So: I originally chose to use this new component rather than PowerSwitch because 2 atmos devices in particular (filters and mixers) don't require power. I implemented it there since I wished to retain their properties of not requiring power (which im pretty sure was intentional, unless the PR that made pumps powered didn't do filters and mixers for some stranger reason).
At that point, I couldve left it as is and used/left PowerSwitch for the rest of the devices (pumps, thermomachines). And that would've worked fine...

However, I made a certainly debatable decision to use AtmosToggleableComponent for everything.. the reason for which being that I didn't want to leave some atmos machines using this while others used PowerSwitch. The reason behind this is for consistency -- as I said, some devices would've used PowerSwitch and ATC for others.

> Then at minimum, let's not name it that to avoid that implication and move it outside of Atmos code.

I'll draft this PR at the moment since I'll actually change some stuff to it, including making it not exclusive to atmos since thats a better approach.

> I still think you should spend some time evaluating whether a new component is really necessary.

And if i find a component that happens to also do exactly what I'm trying to accomplish, I'll gladly replace this with that.

Another reason for the new component is to allow toggling without relying on power code to handle the logic for you with `TogglePower()` and `.Powered` on ApcPowerReceiver.

> But now we're duplicating component logic.

If you mean appearance updating or creating alternate verbs... I don't see what would be wrong with using those.

Otherwise, could you provide an example of which logic in specific?

deltanedas:
filters and mixers do work, they should require power

LaCumbiaDelCoronavirus:
> filters and mixers do work, they should require power

If you mean they are intended to require power, they don't currently require power to work. But yes they realistically should.

LaCumbiaDelCoronavirus:
nah this is getting trolled so hard i'll do it later

PR #36077, opened 2025-03-25
Remove unused PowerConsumer from grilles

PJB3005:
This caused tons of completely wasted processing in the power system. Incredible.


slarticodefast:
Yep, I'm looking at the code and the power draw is handled by the `VirtualElectrocutionLoadBase` derived entities, not the grill itself.

PR #36078, opened 2025-03-25
add component refcounting system

deltanedas:
## About the PR
to avoid shit like trayscanneruser but anything can use it seamlessly
`EntityManager.AddComponents` -> `_refCount.AddComponents` etc

made status effects and some other stuff use it

## Why / Balance
we love reinventing the wheel for every possible thing that could add components from user actions

## Technical details
RefCountComponent stores the number of sources for each component, once it reaches 0 it actually removes a component

replaced tray evil thing, now you can have a t-ray status effect/reagent and itll Just Werk

also made tray actually real, it adds tray user only when active not just when you pick up an inactive t-ray. i dont know if this was actually important

## Media
source gets incremented when a component source is added

no more sources, it gets removed

multiple tray still work with refcount

## Requirements

## Breaking changes
no

**Changelog**
:trollface:


whatston3:
Blessed.

Content.Shared/StatusEffect/RefCountSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+using Robust.Shared.Utility;
+
+namespace Content.Shared.StatusEffect;
+
+/// <summary>
+/// Manages adding and removing ref-counted components through a public API.
+/// </summary>
+public sealed class RefCountSystem : EntitySystem
+{
+    [Dependency] private readonly IComponentFactory _factory = default!;
```
metalgearsloth, line 13: On EntityManager noob.

Content.Shared/StatusEffect/RefCountComponent.cs
```diff
@@ -0,0 +1,18 @@
...
+/// Use this instead of blindly adding/removing components (related: https://youtu.be/k0MUj34y5Kg)
+/// </summary>
+[RegisterComponent, Access(typeof(RefCountSystem))]
+public sealed partial class RefCountComponent : Component
+{
+    /// <summary>
+    /// Each component added and the number of sources it has.
+    /// </summary>
+    /// <remarks>
+    /// Not networked because dear god that would be complicated.
```
metalgearsloth, line 14: How so.

Also there's no guards being made for the fact there's no guarantee the client doesn't skip the code anyway if ticks are dropped.

Content.Shared/StatusEffect/RefCountSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        if (!_timing.IsFirstTimePredicted)
+            return false;
+
+        ent.Comp ??= EnsureComp<RefCountComponent>(ent);
+        if (!Increment((ent, ent.Comp), type))
+            return false;
+
+        var comp = (Component) _factory.GetComponent(type);
+        AddComp(ent, comp, force);
```
metalgearsloth, line 49: Isn't this cast pointless.

Content.Shared/StatusEffect/RefCountSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    private uint GetCount(RefCountComponent comp, Type type)
+    {
+        return comp.Counts.GetValueOrDefault(type);
+    }
+
+    private void SetCount(Entity<RefCountComponent> ent, Type type, uint count)
+    {
+        ent.Comp.Counts[type] = count;
+    }
```
metalgearsloth, line 127: Shouldn't you just remove it if the count is 0, granted you'd need compstate cloning fixed.

metalgearsloth (changes requested):
Should ask pjb if it should go on engine.

Content.Shared/StatusEffect/RefCountComponent.cs
deltanedas: if client drops ticks it cant be predicted at all no

Content.Shared/StatusEffect/RefCountSystem.cs
deltanedas: i think it was crying about it before

Content.Shared/StatusEffect/RefCountComponent.cs
deltanedas: networked and using component name maybe itll work

Content.Shared/StatusEffect/RefCountSystem.cs
deltanedas: GetComponent returns IComponent but AddComp requires Component 4 no raisin

deltanedas:
@PJB3005 should it go in engine

deltanedas:
cant have a shit in detroit

fuckers stole the toilet

PR #36079, opened 2025-03-25
Enable parallel Pow3r solver on Vulture

PJB3005:
See if it still causes issues.


slarticodefast (approved):
testing on the test server

PR #36088, opened 2025-03-26
Cleanup WarpCommand

Tayrtahn:
## About the PR

Fixes 5 warnings in `WarpCommand.cs`

## Why / Balance

## Technical details

**2x 'EntityCoordinates.GetGridUid(IEntityManager)' is obsolete: 'Use SharedTransformSystem.GetGrid()' [CS0618](https://msdn.microsoft.com/query/roslyn.query?appId=roslyn&k=k(CS0618))**
Simple substitution.

**2x 'EntityCoordinates.GetMapId(IEntityManager)' is obsolete: 'Use SharedTransformSystem.GetMapId()' [CS0618](https://msdn.microsoft.com/query/roslyn.query?appId=roslyn&k=k(CS0618))**
Simple substitution.

**1x 'TransformComponent.Coordinates.set' is obsolete: 'Use the system's setter method instead.' [CS0618](https://msdn.microsoft.com/query/roslyn.query?appId=roslyn&k=k(CS0618))**
Simple substitution.

The `EntityManager.System` call to get `SharedMapSystem` was moved earlier in the method.

Also stumbled upon and fixed a bug where the `mapA` and `mapB` variables were being assigned the same value before comparison, leading to warp points on the current map not being prioritized correctly.

## Requirements

### DIFF
--- a/Content.Server/Administration/Commands/WarpCommand.cs
+++ b/Content.Server/Administration/Commands/WarpCommand.cs
@@ -120,9 +120,8 @@ namespace Content.Server.Administration.Commands
                 }
 
                 var xform = _entManager.GetComponent<TransformComponent>(playerEntity);
-                var xformSystem = _entManager.System<SharedTransformSystem>();
-                xform.Coordinates = coords;
-                xformSystem.AttachToGridOrMap(playerEntity, xform);
+                _transform.SetCoordinates(playerEntity, coords);
+                _transform.AttachToGridOrMap(playerEntity, xform);
                 if (_entManager.TryGetComponent(playerEntity, out PhysicsComponent? physics))
                 {
                     _entManager.System<SharedPhysicsSystem>().SetLinearVelocity(playerEntity, Vector2.Zero, body: physics);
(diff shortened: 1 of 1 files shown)


Content.Server/Administration/Commands/WarpCommand.cs
```diff
@@ -118,8 +120,7 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
                 }
 
                 var xform = _entManager.GetComponent<TransformComponent>(playerEntity);
-                var xformSystem = _entManager.System<SharedTransformSystem>();
-                xform.Coordinates = coords;
+                xformSystem.SetCoordinates(playerEntity, coords);
                 xformSystem.AttachToGridOrMap(playerEntity, xform);
```
slarticodefast, line 124:
```suggestion
                xformSystem.AttachToGridOrMap(playerEntity);
```
no need for the `GetComponent` above since the function already resolves it using a query if null

slarticodefast (changes requested):

Content.Server/Administration/Commands/WarpCommand.cs
Tayrtahn: Good point!

slarticodefast (approved):

slarticodefast:
It's going down!
grafik

PR #36089, opened 2025-03-26
remove mind roles from EntityWhitelist

slarticodefast:
## About the PR
They should not be in there, whitelists should stay as lightweight as possible.
Itemsize should probably be removed as well.

## Why / Balance
grafik

## Technical details
Removed the roles datafield from `EntityWhitelist` and the corresponding logic from the system.
The only thing using it was `RoleRequirementComponent`, so I have rewritten that to directly check if the target has the given role components.

## Requirements

## Breaking changes
Removed the `Roles` datafield from `EntityWhitelist`.
`RoleRequirementComponent` now needs a hashset of component names instead of an `EntityWhitelist`.
This needs a minor change to the yaml formatting:
```
  - type: RoleRequirement
    roles:
      mindRoles:
      - NinjaRole
``` 
becomes
```
  - type: RoleRequirement
    roles:
    - NinjaRole
```

**Changelog**
not player facing

### DIFF
--- a/Content.Server/Objectives/Systems/RoleRequirementSystem.cs
+++ b/Content.Server/Objectives/Systems/RoleRequirementSystem.cs
@@ -1,6 +1,6 @@
 using Content.Server.Objectives.Components;
 using Content.Shared.Objectives.Components;
-using Content.Shared.Whitelist;
+using Content.Shared.Roles;
 
 namespace Content.Server.Objectives.Systems;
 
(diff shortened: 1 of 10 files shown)


Content.Server/Objectives/Systems/RoleRequirementSystem.cs
```diff
@@ -9,7 +9,8 @@ namespace Content.Server.Objectives.Systems;
 /// </summary>
 public sealed class RoleRequirementSystem : EntitySystem
 {
-    [Dependency] private readonly EntityWhitelistSystem _whitelistSystem = default!;
+    [Dependency] private readonly IComponentFactory _componentFactory = default!;
```
metalgearsloth, line 12: entitymanager already has ref to these that can be used.

slarticodefast: Oh, neat!

slarticodefast: done

PR #36090, opened 2025-03-26
Add invisible robot fence to prevent utility bots from pathfinding across them

Quantum-cross:
## About the PR

- Add robot fence item
- Consolidate collision groups for pathfinding relevance
- add BotImpassible layer to block utility bot pathfinding
- make medibot, firebot, cleanbot respect the robot fences

Todo -- this PR

- [ ] go through and comment stuff
- [ ] double check code style

Todo -- future PRs?

- Crafting recipe
- Possibly mapped into maps with round start bots.

## Why / Balance

I've often heard some people talk about "leashing" medibots within certain areas. Anchoring a medibot is a bad solution because 90% of the time it will target a player that it can't reach and then it doesn't do anything and spams PLEASE HOLD STILL a billion times while everyone else is dying.

I like items and mechanics that are both stupid and imitate real life, so of course I took inspiration from real life:

robot_fence

## Technical details

I've added a new collision group for "BotImpassible" which just means that bots won't automatically pathfind through them.

This is turned on and off via the NPCBlackboard variable `NavLaserFenced`.

The RobotFenceComponent itself will cast a ray every second to determine how far it reaches, then it will spawn basic RobotFenceSegment entities in front of it for the length of the beam that just contain a single AABB fixture on the BotImpassible layer. The granularity of the length of the beam is the same as the tile SubStep/Breadcrumb. The last Segment in the beam will be a RobotFenceSegment1, 2, or 3 representing a fixture of length 0.25, 0.50, and 0.75 tile size (1, 2, or 3 breadcrumbs).

Bots will not consider Polys behind the barrier for their random path or real target (injectable target or moppable puddle).

The robot fences draw power with an extension cable and can be turned on and off for shenanigans.

## Media

Showing a bunch of medibot random path routes, no crossing the barriers.
image

https://github.com/user-attachments/assets/276e9328-fb73-483c-a0db-da60115235a9

## Requirements

## Breaking changes

I can't think of any

**Changelog**
:cl:
- add: robot fence so you can let the medibot roam free in med without escaping


ArtisticRoomba:
This is preposterous!!! How am I supposed to get into medical now??

walksanatora:
me putting a laser wall outside of sec (officer beepsky is now trapped).

Content.Shared/Silicons/Bots/RobotFenceComponent.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    [DataField]
+    public FixedPoint2 BeamRange = 8f;
+
+    [DataField(readOnly: true)]
+    public int BeamSubSteps;
+
+    [DataField(readOnly: true)]
+    public List<EntityUid?> BeamEntities = [];
+}
```
metalgearsloth, line 24: This is going to nuke persistence.

Content.Server/NPC/Pathfinding/PathfindingSystem.cs
```diff
@@ -268,11 +269,12 @@ public async Task<PathResultEvent> GetRandomPath(
...
             var layer = 0;
-            var mask = 0;
+            var mask = GetCollisionMask(entity);
 
             if (TryComp<FixturesComponent>(entity, out var fixtures))
             {
-                (layer, mask) = _physics.GetHardCollision(entity, fixtures);
+                (layer, var newmask) = _physics.GetHardCollision(entity, fixtures);
+                mask |= newmask;
             }
```
metalgearsloth, line 278: This bulldozes the difference between flying and ground mobs.

Content.Server/Silicons/RobotFenceSystem.cs
```diff
@@ -0,0 +1,183 @@
...
+{
+    [Dependency] private readonly SharedPhysicsSystem _physics = default!;
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly PowerReceiverSystem _powerReceiver = default!;
+    [Dependency] private readonly AppearanceSystem _appearance = default!;
+
+    private const float XOff = -.125f;
+    private const float YOff = -.3f;
+
```
metalgearsloth, line 28: Magic numbers should be documented otherwise it makes code unreadable.

Content.Server/Silicons/RobotFenceSystem.cs
```diff
@@ -0,0 +1,183 @@
...
+        {
+            if(comp.NextUpdate > curTime)
+                continue;
+
+            UpdateFixture((uid, comp));
+
+            comp.NextUpdate += comp.UpdatePeriod;
+        }
+    }
+
```
metalgearsloth, line 78: You've doing expensive updates on a fixed timer which is really bad.

Content.Server/Silicons/RobotFenceSystem.cs
```diff
@@ -0,0 +1,183 @@
...
+            {
+                newEnt = SpawnAtPosition($"RobotFenceSegment{newEndcapSubSteps}", newCoords);
+                _transform.SetLocalRotationNoLerp(newEnt.Value, localRotation.Opposite());
+            }
+            ent.Comp.BeamEntities.Add(newEnt);
+        }
+
+        ent.Comp.BeamSubSteps = beamSubSteps;
+    }
+
```
metalgearsloth, line 171: There's no reason to spam entities when all it needs to do is override the graph, at least until it's rewritten. Especially no reason to be checking and deleting these every single second.

The slightly less bad thing to do is just update the fixture manually to compensate but this is too expensive atm.

Content.Server/Silicons/RobotFenceSystem.cs
```diff
@@ -0,0 +1,183 @@
...
+public sealed class RobotFenceSystem : EntitySystem
+{
+    [Dependency] private readonly SharedPhysicsSystem _physics = default!;
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly PowerReceiverSystem _powerReceiver = default!;
+    [Dependency] private readonly AppearanceSystem _appearance = default!;
+
+    private const float XOff = -.125f;
+    private const float YOff = -.3f;
```
metalgearsloth, line 27: This system is entirely hardcoded to robotfences which should be avoided.

metalgearsloth (changes requested):

Content.Server/NPC/Pathfinding/PathfindingSystem.cs
Quantum-cross: This just adds BotImpassible if it's on the blackboard

Content.Server/Silicons/RobotFenceSystem.cs
Quantum-cross: I tried doing this with a single fixture, but Pathfinding.Grid.cs seems pretty insistent on throwing away fixtures that aren't centered in the current tile when updating a poly, causing some strange effects like this:

image

Content.Server/Silicons/RobotFenceSystem.cs
Quantum-cross: I can probably update this using collision systems or something...?

TeenSarlacc:
> This is preposterous!!! How am I supposed to get into medical now??

sentenced to the IR laser cell

Quantum-cross:
obviously will require some rethinking, I'll come back to it

EmoGarbage404:
re-open when you come back to it

Quantum-cross:
sounds good

Errant-4:
Maybe simply make it so that "leashed" bots do not go through doors even if open and/or stay within LOS of their "area indicator" ?

Quantum-cross:
I definitely want to come back to this soon, I've learned a lot about the pathfinding and collision systems... I will make my original design work with a better implementation.

PR #36091, opened 2025-03-26
NanoTask update

Ertanic:
## About the PR

NanoTask now requires a server to process tasks. Tasks will be divided into two categories: publicly available station tasks and department tasks intended for those who have access to the respective departments. The ability to print tasks will be removed, but things like a new “in progress” status and a field for who took the task will be added.

## Why / Balance

NanoTask is a pretty cool idea and beautiful in implementation, but it's pretty inconvenient and useless, you might as well use regular notes. With this PR it will become much more useful. But in order not to kill the communication of players in the first stages, I think to make it possible to make the server as a research in the console technology.

## Technical details

Everything works through `DeviceNetworkSystem` and `SingletonDeviceNetServerSystem`. Still thinking about implementing accesses for department tabs and getting the name from the ID card of the person who takes the task to execute. As a rough draft I decided to try to do it via the `NanoTaskDepartmentPrototype` prototype. **Help is probably required**.

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: A server has been added for the NanoTask program.


walksanatora:
can there still be some kind of "local" table or whatever that is like currently impl. that way incase of mass incompetence you can still just pass bits of paper around to distribute task.

ilovehans10:
Would there be any way to attach a timer to a task? Making it so that you get a little notification after fifteen minutes that coils need welded, or to check in on perma would be handy.

PR #36093, opened 2025-03-26
Better jetpack emitter

metalgearsloth:
Still need proper particles this just tilts me whenever I see it.

https://github.com/user-attachments/assets/2e533ebf-662f-4ffb-9187-34edc5b09e10

**Changelog**
:cl:
- tweak: Jetpacks emit particles more frequently.


Resources/Prototypes/Entities/Objects/Tools/jetpacks.yml
```diff
@@ -35,6 +35,7 @@
     - type: Sprite
       sprite: Objects/Tanks/Jetpacks/blue.rsi
       state: icon
+      noRot: true
```
slarticodefast, line 38:
```suggestion
```
why this? This makes them look weird if they are on a moving grid
items should stay rotateable anyways (although they currently act weird when thrown)

slarticodefast (approved):
Got one nitpick, looks good otherwise.

Resources/Prototypes/Entities/Objects/Tools/jetpacks.yml
metalgearsloth: Sometimes the gaps are weird, also jetpacks aren't useable on grids.

slarticodefast: I mean if you drop them on the floor. It creates a weird difference to other items that are rotated.

PR #36103, opened 2025-03-26
Overlay stack sorting fix

Errant-4:
## About the PR
Stacked overlays should no longer sometimes inexplicably flicker/rapidly switch the order

## Why / Balance
Bug

## Technical details
The stack flicker is probably caused by the imprecision of the numbers used for sorting, though the problem is rare and exceedingly difficult to reproduced or observein dev... So I could not actually confirm what exactly happens when this issue manifests or if this indeed fixes it, but it probably will

## Requirements


ScarKy0 (approved):
Well I don't have the issue in dev with this, but seeing its a pain to reproduce itll be best of tested live
Looks good

PR #36105, opened 2025-03-27
Add paradox clone to admin antag control

slarticodefast:
## About the PR
Allows admins to easily create a paradox clone ghost role of a selected player using the antag control verbs.
It will spawn normally when a ghost player selects it.

## Why / Balance
Create 10 captains

## Technical details
Added an override to the rule component, only randomize the target if not specified.

Use two clients for testing, because the original needs a mind to be cloneable and the game rule will cancel otherwise.

## Media
Screenshot (383)

## Requirements

## Breaking changes
renamed the `Original` datafield in `ParadoxCloneRuleComponent` to `OriginalBody`

**Changelog**
:cl:
ADMIN:
- add: Added an antag control verb to create a paradox clone ghost role of a selected player.


TeenSarlacc:
I'd pay real money to spectate 10 captains fighting for dominance

Mixelz:
> I'd pay real money to spectate 10 captains fighting for dominance

Why not paradox clone a ninja?
Why not paradox clone *10* ninjas?
These are the big brain thoughts that will make for a memorable admeme event.

Dragonjspider:
War of the clans...

beck-thompson (changes requested):
Why did you OriginalBody instead of just keeping with mind? I don't think you ever use it! Is there an issue with getting the mind in the admin verb?

slarticodefast:
> Why did you OriginalBody instead of just keeping with mind? I don't think you ever use it (Besides just using as a check for the admin verb)! Is there an issue with getting the mind in the admin verb?

Before I only had the mind saved and it led to a terrible mess where I had to get the body from the mind, then the mind from the body, repeat that a few times and use a TryComp in every step because components cannot be used in datafields.

Also the override has to be based on the body, not the mind because otherwise if the target player ghosts and takes a ghost role or is mind swapped the paradox clones will become a copy of whatever he is then.

ScarKy0:
Peak. I hope we can get this in time for stable cutoff

ScarKy0 (approved):
Looks fine to me, works in-game too

beck-thompson (approved):
Makes sense! Looks fine then

PR #36107, opened 2025-03-27
Minor changes to the EmotesMenu xaml

VerinSenpai:
## About the PR
Updated the window xaml to use AnimatedTextureRect so it can use the icon state. Also changed the xaml.cs to use the existing Main instead of fetching a control for it.

## Why / Balance
A bit of cleanup I believe. I noticed while working on my pr for the top menu buttons that if the radial menu was loaded when the round started using OnStateChanged we'd get a test fail because this window was getting two of its icons from an rsi. While I may not be able to take advantage of it for what I'm doing, it occurs to me if its failing a test its probably not a way we want to do it. Also changes the window xaml.cs to just get Main from the local variable instead of fetching the controller.

## Technical details
Instead of grabbing the png directly from the rsi, uses SpriteSpecifier and AnimatedTextureRect to set the texture and sets the TextureScale to 2, 2.

## Requirements


VerinSenpai:
how easy to pass 0 changes is.

Fildrance:
it'd most likely be needed to add new type of button and... model? for radial... not sure... is it really going to look good? might worth it :godo:

VerinSenpai:
Superceded by #34639

VerinSenpai:
> it'd most likely be needed to add new type of button and... model? for radial... not sure... is it really going to look good? might worth it :godo:

I'm gonna close this for now. I'm going to focus on cleaning up the other menu bar buttons for now. If the test still fails when I get here, I'll look into it then.

PR #36113, opened 2025-03-27
Centcomm death rattle implant

Alpha-Two:
## About the PR
Adds an implant + implanter for a death-rattle for the centcomm radio channel, and adds them natively to CC ghost roles.

## Why / Balance
Because it seems fitting, and useful for potential admeme roleplay with CC.

## Technical details
Adds `DeathRattleImplantCentcomm` and `DeathRattleImplanterCentcomm` that has a death rattle effect on crit and death to the CentCom channel. Also adds a tag to the RadioChannel field, because why not.

## Media
(other rattling implants added to compare)
image

## Requirements

## Breaking changes
None?

**Changelog**
No cl as players won't experience this outside admemes.

### DIFF
--- a/Content.Shared/Implants/Components/RattleComponent.cs
+++ b/Content.Shared/Implants/Components/RattleComponent.cs
@@ -8,7 +8,7 @@ namespace Content.Shared.Implants.Components;
 public sealed partial class RattleComponent : Component
 {
     // The radio channel the message will be sent to
-    [DataField]
+    [DataField("radioChannel")]
     public ProtoId<RadioChannelPrototype> RadioChannel = "Syndicate";
 
     // The message that the implant will send when crit
--- a/Content.Shared/Implants/Components/RattleComponent.cs
+++ b/Content.Shared/Implants/Components/RattleComponent.cs
@@ -8,7 +8,7 @@ namespace Content.Shared.Implants.Components;
 public sealed partial class RattleComponent : Component
 {
     // The radio channel the message will be sent to
-    [DataField("radioChannel")]
+    [DataField]
     public ProtoId<RadioChannelPrototype> RadioChannel = "Syndicate";
 
     // The message that the implant will send when crit
@@ -16,6 +16,6 @@ public sealed partial class RattleComponent : Component
     public LocId CritMessage = "deathrattle-implant-critical-message";
 
     // The message that the implant will send when dead
-    [DataField("deathMessage")]
+    [DataField]
     public LocId DeathMessage = "deathrattle-implant-dead-message";
 }
--- a/Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
@@ -18,9 +18,9 @@
   id: EventHumanoidCentcomm
   parent: EventHumanoidMindShielded
   components:
-    - type: AutoImplant
-      implants:
-      - DeathRattleImplantCentcomm
+  - type: AutoImplant
+    implants:
+    - DeathRattleImplantCentcomm
 
 ## Death Squad
 
--- a/Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
@@ -14,6 +14,14 @@
     - type: MindShield
     - type: AntagImmune
 
+- type: randomHumanoidSettings
+  id: EventHumanoidCentcomm
+  parent: EventHumanoidMindShielded
+  components:
+    - type: AutoImplant
+      implants:
+      - DeathRattleImplantCentcomm
+
 ## Death Squad
 
 - type: entity
@@ -33,7 +41,7 @@
 
 - type: randomHumanoidSettings
   id: DeathSquad
-  parent: EventHumanoidMindShielded
+  parent: EventHumanoidCentcomm
   randomizeName: false
   components:
     - type: GhostRole
@@ -72,7 +80,7 @@
 
 - type: randomHumanoidSettings
   id: ERTLeader
-  parent: EventHumanoidMindShielded
+  parent: EventHumanoidCentcomm
   randomizeName: false
   components:
     - type: GhostRole
@@ -500,7 +508,7 @@
 
 - type: randomHumanoidSettings
   id: CBURNAgent
-  parent: EventHumanoidMindShielded
+  parent: EventHumanoidCentcomm
   components:
     - type: Loadout
       prototypes: [CBURNGear]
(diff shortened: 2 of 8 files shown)


Content.Shared/Implants/Components/RattleComponent.cs
```diff
@@ -8,7 +8,7 @@ namespace Content.Shared.Implants.Components;
 public sealed partial class RattleComponent : Component
 {
     // The radio channel the message will be sent to
-    [DataField]
+    [DataField("radioChannel")]
```
ScarKy0, line 11: No need for the radioChannel, its the same name as the datafield so itll work regardless

ScarKy0 (changes requested):

Content.Shared/Implants/Components/RattleComponent.cs
Alpha-Two: Weirdly, I knew that, but the deathMessage field had a tag so I added for consistency. I'll remove both of them though

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
```diff
@@ -362,3 +362,22 @@
...
+  components:
+  - type: SubdermalImplant
+    permanent: true
+    whitelist:
+      components:
+      - MobState # admeme implanting a chair with rattle implant needs to give the chair mobstate so it can die first
+  - type: TriggerOnMobstateChange
+    mobState:
+    - Critical
+    - Dead
```
slarticodefast, line 381: just inherit from DeathRattleimplant so you don't need to copy paste all of this
I would also move the code directly below that one

slarticodefast (changes requested):

Resources/Prototypes/Entities/Objects/Misc/subdermal_implants.yml
Alpha-Two: I changed it to inherit. I did _not_ change its position in the file, however, as they're already sorted by purpose (syndi, nukie, sec+command, centcomm)

ScarKy0 (changes requested):
Can you make centcomm, deathsquad and ERT all spawn with this natively?

Alpha-Two:
> Can you make centcomm, deathsquad and ERT all spawn with this natively?

Unfortunately there's no pre-existing code set up to do that, and adding that would be out of scope of this PR. If preferred, I can either simply give them an implanter with the death rattle, or I can open another PR with the relevant code changes. 

(As a sidenote, this is the reason deathsquad spawn with 2 implanters as opposed to them being implanted natively.

ScarKy0:
> Unfortunately there's no pre-existing code set up to do that, and adding that would be out of scope of this PR. If preferred, I can either simply give them an implanter with the death rattle, or I can open another PR with the relevant code changes.
(As a sidenote, this is the reason deathsquad spawn with 2 implanters as opposed to them being implanted natively.

Check the AutoImplant component, I believe this is what you're looking for, if not then just have them all spawn with it instead

slarticodefast:
`AutoImplant` runs on map init, so it should work.

Alpha-Two:
> Check the AutoImplant component, I believe this is what you're looking for, if not then just have them all spawn with the implanter

I see now I am le stupid, I'll do it now thanks

Alpha-Two:
Note to add: The shift-start loadouts of ERT staff, which exist for *some* reason, have been given mindshields and death rattles as well. I believe that if (a) A map were to have an ERT slot available (theoretically possible but would never get past the maintainers) and (b) A player had enabled that job via editing character files, then they would spawn as an ERT member, but without a mindshield??

Resources/Prototypes/Entities/Mobs/Player/humanoid.yml
```diff
@@ -14,6 +14,14 @@
     - type: MindShield
     - type: AntagImmune
 
+- type: randomHumanoidSettings
+  id: EventHumanoidCentcomm
+  parent: EventHumanoidMindShielded
+  components:
+    - type: AutoImplant
+      implants:
+      - DeathRattleImplantCentcomm
```
slarticodefast, line 23:
```suggestion
  components:
  - type: AutoImplant
    implants:
    - DeathRattleImplantCentcomm
```
fix indentation
(yes, I know it's wrong above as well)

Resources/Prototypes/Roles/Jobs/CentComm/cburn.yml
```diff
@@ -12,6 +12,9 @@
   - AllAccess
   access:
   - CentralCommand
+  special:
+  - !type:AddImplantSpecial
+    implants: [ MindShieldImplant, DeathRattleImplantCentcomm ]
```
slarticodefast, line 17: don't they get two death rattle implants now?
One from AutoImplant and one from AddImplantSpecial?

Alpha-Two: No............, I don't think ? The AddImplantSpecial is attached to the Death Squad round-start job, and should not be referenced by the RandomHumanoidSpawner.

Even besides, the Death Squad RHS (among others) adds the mindshield component directly, not the implant, so there should be no issue with duplication.

Alpha-Two: Oh wait you were speaking of the death rattle. I believe my point still stands.

ScarKy0 (approved):
Tested and works

beck-thompson (changes requested):
Looks like certain jobs don't have the death rattle implant like this one:
image

Alpha-Two:
> Looks like certain jobs don't have the death rattle implant like this one: image

They.... do?
image

Content.Shared/Implants/Components/RattleComponent.cs
```diff
@@ -16,6 +16,6 @@ public sealed partial class RattleComponent : Component
     public LocId CritMessage = "deathrattle-implant-critical-message";
 
     // The message that the implant will send when dead
-    [DataField("deathMessage")]
+    [DataField]
```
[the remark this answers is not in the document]
Alpha-Two: Thanks, I'll keep it in mind

beck-thompson:
> They.... do?

Huh the implanter must be broken then or something because I couldn't remove it. I guess its unremovable? It probably shouldn't be on the list then

Alpha-Two:
> > They.... do?
> 
> Huh the implanter must be broken then or something because I couldn't remove it. I guess its unremovable? It probably shouldn't be on the list then

Well it shouldn't be on the implant remove list because it's a different implant entirely to the original Death Rattle implant. If it is, could you point out where because that sounds like a bug?

beck-thompson (approved):

beck-thompson:
looks good

PR #36114, opened 2025-03-27
Add microwave-nukedisk interaction

ElectroJr:
## About the PR
Makes microwaving a nuke disk change the nuke countdown timer by a random amount.

## Why / Balance
Why? I was told to add the feature by a friend and we don't have enough funny microwave interactions. 

Balance wise, it currently randomly changes the default 300s cooldown using a gaussian distribution. 90% of the time the modifier is smaller than 45s, and 99% of the it's smaller than 70s. The resulting nuke timer will still always be greater than `NukeComponent.MinimumTime`. Using a flat distribution instead of gaussian might be better for balance as it prevent extreme events, but I'd rather it just usually have a small impact, while still allowing it to sometimes change the timer by an appreciable amount. I don't really feel strongly about the specific numbers, I chose them pretty arbitrarily.

## Technical details
Adds a new `NukeDiskComponent.TimeModifier` datafield that is combined with the NukeComponent's `Timer` datafield to set the remaining time when the nuke is ready to be armed. Unless it already has a value, the disks time modifier will be randomly assigned whenever the disk is first microwaved, and there is no way to know what the modifier is until the disk is in the nuke & the code has been entered.

## Media

I've added a pop-up message, but cant be bothered re-recording the video:

https://github.com/user-attachments/assets/922612df-f5b2-4eec-b301-dc3eca7ca1cc

## Requirements

:cl:
- add: Microwaving the nuke disk will now slightly randomize the nuke countdown timer.

### DIFF
--- a/Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
+++ b/Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
@@ -544,6 +544,10 @@ namespace Content.Server.Kitchen.EntitySystems
                 var ev = new BeingMicrowavedEvent(uid, user);
                 RaiseLocalEvent(item, ev);
 
+                // TODO MICROWAVE SPARKS & EFFECTS
+                // Various microwaveable entities should probably spawn a spark, play a sound, and generate a pop=up.
+                // This should probably be handled by the microwave system, with fields in BeingMicrowavedEvent.
+
                 if (ev.Handled)
                 {
                     UpdateUserInterfaceState(uid, component);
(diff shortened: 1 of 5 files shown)


millionse:
Even more dumb interaction would be only emagging nuke to make it detonate without disk and codes, but interaction is still funny, I like it

SlimmSlamm:
You should add an interaction with the fake nuke disk that syndicate can buy and the toy one too. Maybe have the toy one turn into a nuke toy.

notquitehadouken:
microwaving the nuke disk replaces it with a fake disk and respawns the real disk :trollface:

beck-thompson (approved):
Looks good! There are some weird interactions if you use multiple disks on the same nuke (it only uses the time from the first disk) but it makes as much sense as anything else :laughing:

Issue #35208, opened 2025-02-16
Several maps have "DO NOT MAP" entities.


ElectroJr:
Several maps currently fail the new test added in #34711 that checks that no maps contain entities that belong to the "DO NOT MAP" category.
Currently these test failures are just being suppressed by a whitelist (`PostMapInitTest.DoNotMapWhitelist`)

In particular:
- [ ] bagel contains mime's rubber stamp
- [ ] gate contains positronic brain and LSE-1200c "Perforator"
- [ ] meta contains warden's rubber stamp (#34712)
- [ ] reach contains handheld crew monitor
- [ ] `/Maps/Shuttles/ShuttleEvent/cruiser.yml` contains LSE-1200c "Perforator"
- [ ] `/Maps/Shuttles/ShuttleEvent/honki.yml` contains golden honker, clown's rubber stamp
- [ ] `/Maps/Shuttles/ShuttleEvent/instigator.yml` contains EXP-320g "Friendship"
- [ ] `/Maps/Shuttles/ShuttleEvent/syndie_evacpod.yml` contains syndicate rubber stamp

These could be fixed in a number of different ways:
- Remove the entity from the map
- If the entity prototype should actually be allowed on maps, it needs to be removed from the do not map category
- If the entity should be allowed, but you still want to dissuade mappers from mapping them, then it should be removed from the category but given some kind of "maybe don't map this" suffix.
- Keep the map in the whitelist (i.e., just allow that map to have forbidden entities). There should probably be a comment in the whitelist explaining why.
  - Maybe just allow all shuttles or other non-standard maps to ignore the test?

Kadeo64:
why is mime's rubber stamp DNM? is it because it technically counts as them saying something when they stamp paper?

Killerqu00:
> why is mime's rubber stamp DNM? is it because it technically counts as them saying something when they stamp paper?

all stamps are DNM

ElectroJr:
> It’s probably DNM because it either spawns on the mime, or in the default fill for their locker, and it’s supposed to be a unique item

That seems to have been the initial reason given in #19943. Though in the case of something as inconsequential as the mime stamp, I don't know if it really matters if people want to map it. I don't really care either way and would just defer to other maintainers/maptainers.

ElectroJr:
[...]

PR #32139, opened 2024-09-14
Add admin shuttles

IProduceWidgets:
originally from 21202
## About the PR

These are the admin shuttles we've been using on upstream for the last year. They're already in the game, I've just been maintaining them on the servers manually for reasons. Decided to PR them again so I can properly attribute them, and also to save me having to resave a bunch of grids a half dozen times.

Also a few people have asked me for them for forks, so ya know, here they is.

Doesn't contain roles, you gotta spawn those because these are for broad purpose and idk what you're doin'.

Its probably a good idea to have these in the repo if ever we consider making ERT events (this is not design approval for that).

## Why / Balance

Ask the admin spawning it.

## Media

image
image
image
image
image
image
image
image
image
image
image
image
image
image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

no CL heck you.

PR #36117, opened 2025-03-27
Improve Do Not Map test to whitelist specific prototypes per map and whitelist entire directories

Tayrtahn:
## About the PR

Adds two new features to the integration test that checks for "Do Not Map" entities on maps (https://github.com/space-wizards/space-station-14/pull/34711)
- It is now possible to whitelist specific entity prototypes on a map, instead of having to whitelist an entire map to permit a single prototype:
```c#
{"/Maps/meta.yml", ["RubberStampWarden"]},
{"/Maps/reach.yml", ["HandheldCrewMonitor"]},
{"/Maps/gate.yml", ["ShuttleGunPerforator", "StationAiBrain"]},
```
- It is now possible to whitelist entire directories of maps, including subdirectories. This can be used, for example, to whitelist all shuttle maps:
```c#
"/Maps/Shuttles/"
```

## Why / Balance

Inspired by the last two bullet points of https://github.com/space-wizards/space-station-14/issues/35208.

## Technical details

- Added a new static field `DoNotMapWhitelistSpecific` above the existing `DoNotMapWhitelist`. `DoNotMapWhitelistSpecific` is a dictionary of map paths to arrays of EntProtoIds that are whitelisted.
- Added a method that looks up a given EntProtoId string and a map path and uses `DoNotMapWhitelistSpecific` to return whether the given combo is permitted or not.
- Changed the logic used to check for matches in `DoNotMapWhitelist` from requiring an exact match with any whitelist entry to checking if the map path string starts with any whitelist entry.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/PostMapInitTest.cs
+++ b/Content.IntegrationTests/Tests/PostMapInitTest.cs
@@ -44,17 +44,32 @@ namespace Content.IntegrationTests.Tests
             AdminTestArenaSystem.ArenaMapPath
         };
 
-        private static readonly string[] DoNotMapWhitelist =
+        /// <summary>
+        /// A dictionary linking maps to arrays of entity prototype ids that should be exempt from "DoNotMap" restrictions.
+        /// </summary>
+        /// <remarks>
+        /// This declares that the listed entity prototypes are allowed to be present on the map
+        /// despite being categorized as "DoNotMap", while any unlisted prototypes will still
+        /// cause the test to fail.
+        /// </remarks>
+        private static readonly Dictionary<string, EntProtoId[]> DoNotMapWhiteList = new()
+        {
+            {"/Maps/bagel.yml", ["RubberStampMime"]},
+            {"/Maps/meta.yml", ["RubberStampWarden"]},
+            {"/Maps/reach.yml", ["HandheldCrewMonitor"]},
+            {"/Maps/gate.yml", ["ShuttleGunPerforator", "StationAiBrain"]},
+            {"/Maps/Shuttles/ShuttleEvent/honki.yml", ["GoldenBikeHorn", "RubberStampClown"]},
+            {"/Maps/Shuttles/ShuttleEvent/syndie_evacpod.yml", ["RubberStampSyndicate"]},
+            {"/Maps/Shuttles/ShuttleEvent/cruiser.yml", ["ShuttleGunPerforator"]},
+            {"/Maps/Shuttles/ShuttleEvent/instigator.yml", ["ShuttleGunFriendship"]},
+        };
+
+        /// <summary>
+        /// Maps listed here are given blanket freedom to contain "DoNotMap" entities. Use sparingly.
+        /// </summary>
+        private static readonly string[] DoNotMapWhitelistAll =
         {
             "/Maps/centcomm.yml",
-            "/Maps/bagel.yml", // Contains mime's rubber stamp --> Either fix this, remove the category, or remove this comment if intentional.
-            "/Maps/gate.yml", // Contains positronic brain and LSE-1200c "Perforator"
-            "/Maps/meta.yml", // Contains warden's rubber stamp
-            "/Maps/reach.yml", // Contains handheld crew monitor
-            "/Maps/Shuttles/ShuttleEvent/cruiser.yml", // Contains LSE-1200c "Perforator"
-            "/Maps/Shuttles/ShuttleEvent/honki.yml", // Contains golden honker, clown's rubber stamp
-            "/Maps/Shuttles/ShuttleEvent/instigator.yml", // Contains EXP-320g "Friendship"
-            "/Maps/Shuttles/ShuttleEvent/syndie_evacpod.yml", // Contains syndicate rubber stamp
         };
 
         private static readonly string[] GameMaps =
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
myra was too scared and im curious to see what it says

Content.IntegrationTests/Tests/PostMapInitTest.cs
```diff
@@ -44,17 +44,36 @@ public sealed class PostMapInitTest
...
+        private static readonly Dictionary<string, EntProtoId[]> DoNotMapWhitelistSpecific = new()
+        {
+            {"/Maps/bagel.yml", ["RubberStampMime"]},
+            {"/Maps/meta.yml", ["RubberStampWarden"]},
+            {"/Maps/reach.yml", ["HandheldCrewMonitor"]},
+            {"/Maps/gate.yml", ["ShuttleGunPerforator", "StationAiBrain"]},
+            {"/Maps/Shuttles/ShuttleEvent/honki.yml", ["GoldenBikeHorn", "RubberStampClown"]},
+            {"/Maps/Shuttles/ShuttleEvent/syndie_evacpod.yml", ["RubberStampSyndicate"]},
+            {"/Maps/Shuttles/ShuttleEvent/cruiser.yml", ["ShuttleGunPerforator"]},
+            {"/Maps/Shuttles/ShuttleEvent/instigator.yml", ["ShuttleGunFriendship"]},
```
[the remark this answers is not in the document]
Tayrtahn: Nah.

Tayrtahn:
> ## Pull Request Overview
> 
> This pull request enhances the integration test for detecting "DoNotMap" entities by supporting two new whitelist mechanisms.
Yep.  
> - Introduces a new dictionary mapping specific map paths to allowed entity prototypes.
Yep.  
> - Changes the directory whitelist logic to perform prefix matching instead of exact string matching.
Close. It changes the _map_ whitelist logic to do that. 

Of course, I spelled all that out pretty well in the description, so I'm not giving it a lot of credit for this.

ArtisticRoomba:
microsoft getting one step closer to replacing us all :godo:

Tayrtahn:
Thought: should the test fail if a map has a whitelist for a specific prototype and that prototype isn't actually on the map?

This would force the whitelists to be kept up to date with changes so we didn't wind up with outdated exemptions sitting around. Out would be a small extra hassle for contributors cleaning up maps though. Unfortunately there's no real way to have a test just "warn" about something; it's either pass or fail.

Tayrtahn:
> Thought: should the test fail if a map has a whitelist for a specific prototype and that prototype isn't actually on the map?

Answering my own question: yes - it's good to keep the whitelist up to date. Added as a part of the test, but can be removed by reverting [b0fb074](https://github.com/space-wizards/space-station-14/pull/36117/commits/b0fb0745c651767f99df31b048755b397b68719a) if not desired.

Also removed the specific whitelist entry for the warden's rubber stamp on Meta, since it was removed by #36140.

Content.IntegrationTests/Tests/PostMapInitTest.cs
```diff
@@ -252,18 +270,31 @@ public async Task NoSavedPostMapInitTest()
...
         /// </summary>
         private void CheckDoNotMap(ResPath map, YamlNode node, IPrototypeManager protoManager)
         {
-            if (DoNotMapWhitelist.Contains(map.ToString()))
-                return;
+            foreach (var mapFilter in DoNotMapWhitelist)
+            {
+                if (map.ToString().StartsWith(mapFilter))
+                    return;
+            }
```
IProduceWidgets, line 290: I have a regex alternative here if you'd rather use globs, but idrc they seem to both achieve the same thing.

https://github.com/space-wizards/space-station-14/pull/32139/files#diff-030244c7ca97eecf3bda5beb7e8ee28924e5627ffa187b5cbcc2786f4d00832f

Content.IntegrationTests/Tests/PostMapInitTest.cs
```diff
@@ -44,17 +44,35 @@ public sealed class PostMapInitTest
...
+        private static readonly Dictionary<string, HashSet<EntProtoId>> DoNotMapWhitelistSpecific = new()
+        {
+            {"/Maps/bagel.yml", ["RubberStampMime"]},
+            {"/Maps/reach.yml", ["HandheldCrewMonitor"]},
+            {"/Maps/gate.yml", ["ShuttleGunPerforator", "StationAiBrain"]},
+            {"/Maps/Shuttles/ShuttleEvent/honki.yml", ["GoldenBikeHorn", "RubberStampClown"]},
+            {"/Maps/Shuttles/ShuttleEvent/syndie_evacpod.yml", ["RubberStampSyndicate"]},
+            {"/Maps/Shuttles/ShuttleEvent/cruiser.yml", ["ShuttleGunPerforator"]},
+            {"/Maps/Shuttles/ShuttleEvent/instigator.yml", ["ShuttleGunFriendship"]},
+        };
```
IProduceWidgets, line 64: I had a thought and it hurt, but maybe you're interested.

If we split this into a separate yml definition we could use a dict to both contain the prototypes allowed as well as the number of them expected. That would let us maintain the mapping guidelines within the content repo where its unlikely to get as out of date since people can add their new "must map" entities to the guidelines at PR time and it would automagically start hitting tests.

Might need a separate check for DNM and "guidelined" ents though so a fresh PR doesn't constantly fail all the maps, but does send up a Warning flare that something is missing or over abundant.

IProduceWidgets (approved):

Content.IntegrationTests/Tests/PostMapInitTest.cs
Tayrtahn: Very nice! I've added it in. Thanks!

IProduceWidgets:
*Bump

PR #25161, opened 2024-02-12
Remove self-uncuff spam suicide, damage changes

SlamBamActionman:
## About the PR

This PR removes the option to suicide-by-self-uncuffing and spamming the uncuffing noise by adding a cooldown to the uncuff action. With this cooldown damage on self-uncuffing is no longer necessary and has been removed.

## Why / Balance

Self-uncuff suicide is loud, annoying and doesn't make much sense. By putting the self-uncuff action on a cooldown there's no longer any concern about the uncuffing sound being spammed, and as such the damage is no longer necessary.

## Technical details

The `RemoveCuffs` alert _action_ has been removed, meaning you can no longer uncuff by pressing the Cuffed alert icon. Instead the player is given the "Uncuff Self" action button in the action bar, set at the same time as the Cuffed alert. 

The main reason why the self-uncuff was moved to an action button instead of the alert action is that alert actions lack timers. It might be worth looking into the ability to coordinate the two systems (for example, turning on internals is currently handled differently by the game depending on which is clicked) but that's beyond the scope of this PR.

## Media

Video:
[Uncuff video](http://www.youtube.com/watch?v=UtxrEhP2Ssc "Video Title")

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- tweak: Self-uncuff damage has been removed, but attempting to self-uncuff now has a cooldown instead.

PR #17801, opened 2023-07-03
Damage on forced uncuff

noudoit:
## About the PR

This PR aims to stop people from spam uncuffing - the damage only applies to the person wearing and spamming the uncuff button it does not hurt to take cuffs off someone else.
it also adds a log for uncuffing.
it is 3 blunt damage per uncuff attempt.

**Media**

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl:
- tweak: Attempting to uncuff yourself damages you slightly

PR #36121, opened 2025-03-27
Uncuffing yourself will now deal minor stamina damage

Mixelz:
## About the PR
Performing an uncuff attempt on yourself will cause you to take a minor amount of stamina damage. 
Intended to be a simple first PR!

## Why / Balance
Toxic Players have a tendency to spam their uncuff ad nauseam while discussion happens around them or if they are feeling particularly ornery about their situation. This is usually not fun to deal with as the person holding them that needs to micromanage their do-after timer. This simple change is intended to act as a deterrent for spamming uncuff attempts, while not affecting those who are patient with their break out attempts. 

It is important to note that the do-after window is currently *greater* than the window for stamina recovery, meaning all stamina damage gained from a single attempt will be recovered before the do-after completes. It additionally provides a nice bit of visual feedback of the user needing to **painfully** break out of their own cuffs with a flash of synchronized stam damage.

At its current damage value, it would take at least 7 back to back uncuff attempts before you can stam crit yourself. 
However, in practice it will usually take around 9 to 10, accounting for small delays between attempts and the natrual stamina regeneration working in the players favor.

## Technical details
This is my first PR to this repo so the implementation is intended to be exceedingly simple.

It is currently tied into the  `if (args.User == args.Target)` comparison, so it will never trigger unless the user is the one uncuffing themselves. It is not tied into the do-after event so continuous spamming of uncuff will cause you to take damage multiple times within the same do-after period, though since the entire point of this change is to deter uncuff spam this may be an ignorable issue.

## Media

https://github.com/user-attachments/assets/cdf0e1f2-dd7b-4eff-a1e0-bca96c007a19

## Requirements

## Breaking changes
Constructor of `UncuffAttemptEvent` now passes the uid of the cuffs to determine how much damage to apply.
New Variable of `SelfUncuffStamDamage` can be added to cuffs to give a custom damage value, by default this is 15.

**Changelog**
:cl:
- tweak: Attempting to uncuff yourself will now deal a minor amount of self-inflicted stamina damage.

### DIFF
--- a/Content.Shared/Cuffs/Components/HandcuffComponent.cs
+++ b/Content.Shared/Cuffs/Components/HandcuffComponent.cs
@@ -34,6 +34,12 @@ public sealed partial class HandcuffComponent : Component
     public float StunBonus = 2f;
 
     /// <summary>
+    /// Amount of Stamina Damage dealt on a self-uncuff attempt.
+    /// </summary>
+    [DataField]
+    public float SelfUncuffStamDamage = 15f;
+
+    /// <summary>
     ///     Will the cuffs break when removed?
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
@@ -99,10 +105,11 @@ public sealed partial class HandcuffComponent : Component
 /// Should generate popups on the User.
 /// </summary>
 [ByRefEvent]
-public record struct UncuffAttemptEvent(EntityUid User, EntityUid Target)
+public record struct UncuffAttemptEvent(EntityUid User, EntityUid Target, EntityUid Cuffs)
 {
     public readonly EntityUid User = User;
     public readonly EntityUid Target = Target;
+    public readonly EntityUid Cuffs = Cuffs;
     public bool Cancelled = false;
 }
 
--- a/Content.Shared/Cuffs/SharedCuffableSystem.cs
+++ b/Content.Shared/Cuffs/SharedCuffableSystem.cs
@@ -53,7 +53,7 @@ namespace Content.Shared.Cuffs
         [Dependency] private readonly SharedInteractionSystem _interaction = default!;
         [Dependency] private readonly SharedPopupSystem _popup = default!;
         [Dependency] private readonly SharedTransformSystem _transform = default!;
-        [Dependency] private   readonly StaminaSystem           _stamina         = default!;
+        [Dependency] private readonly StaminaSystem _stamina = default!;
         [Dependency] private readonly UseDelaySystem _delay = default!;
 
         public override void Initialize()
--- a/Content.Shared/Cuffs/SharedCuffableSystem.cs
+++ b/Content.Shared/Cuffs/SharedCuffableSystem.cs
@@ -132,7 +132,10 @@ namespace Content.Shared.Cuffs
                 Dirty(args.User, cuffable);
 
                 // User will take self-inflicted stamina damage if attempting to uncuff themselves. This should happen even if the uncuff attempt fails.
-                _stamina.TakeStaminaDamage(args.Target, 15f, visual: true, source: args.User);
+                if (TryComp<HandcuffComponent>(args.Cuffs, out var handcuffComp))
+                {
+                    _stamina.TakeStaminaDamage(args.Target, handcuffComp.SelfUncuffStamDamage, visual: true, source: args.User);
+                }
             }
             else
             {
@@ -609,7 +612,7 @@ namespace Content.Shared.Cuffs
             if (!Resolve(cuffsToRemove.Value, ref cuff))
                 return;
 
-            var attempt = new UncuffAttemptEvent(user, target);
+            var attempt = new UncuffAttemptEvent(user, target, cuffsToRemove.Value);
             RaiseLocalEvent(user, ref attempt, true);
 
             if (attempt.Cancelled)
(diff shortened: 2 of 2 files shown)


slarticodefast:
wait, didn't the self-uncuffing have a cooldown?

Mixelz:
> wait, didn't the self-uncuffing have a cooldown?

Not to my knowlege, though it may be a difference of debug build + aghost vs live.

Veritius (commented):
I like this idea. A previous iteration (PR #17801) was abused by players, letting them commit suicide while not being round-removed. This was fixed in PR #25161. I think this is a great solution, it penalises spamming with something substantial (movement slowdown, stuns). However, it kind of treads on the foot of PR #25161 as it added a cooldown, which this PR does not address.

VerinSenpai:
Does this cause the person to fall over? And does that affect whether they're being pulled or not?

Mixelz:
> Does this cause the person to fall over? And does that affect whether they're being pulled or not?

Only if they spam it enough to stam crit themselves, and it should not as it is a source of self inflicted stamina damage. Not from an outside source that would interrupt the pulling. I can double check this just to be assured though.

VerinSenpai:
I was thinking a stam crit while being pulled shouldn't break the pull but it's better safe than sorry

keronshb:
I feel this is going to be too big of a nerf when combined with #35193

Mixelz:
> I feel this is going to be too big of a nerf when combined with #35193
> 
> Not only does attempting to uncuff here punish you, also being uncuffed by another is going to be needlessly punishing. All with a relatively long doafter on top of it, that's cancelled by moving or being moved, or being punched.
> 
> With the upcoming stamina-focused stun combat, instead of hardstun combat, this will be more prevalent of an issue.

As already mentioned, this change only applies if you attempt to uncuff yourself. Codewise it quite literally does not apply if someone else is uncuffing a person.
This also does not affect anyone that only attempts one to two uncuff attempts, as the do-after is *longer* than the recovery period. You recover roughly 30 stamina from do-after start to end, meaning it cannot apply unless you spam it multiple times or are attempting to run while cuffed after spamming uncuff attempts.

keronshb:
> > I feel this is going to be too big of a nerf when combined with #35193
> > Not only does attempting to uncuff here punish you, also being uncuffed by another is going to be needlessly punishing. All with a relatively long doafter on top of it, that's cancelled by moving or being moved, or being punched.
> > With the upcoming stamina-focused stun combat, instead of hardstun combat, this will be more prevalent of an issue.
> 
> As already mentioned, this change only applies if you attempt to uncuff yourself. Codewise it quite literally does not apply if someone else is uncuffing a person.
> 
> This also does not affect anyone that only attempts one to two uncuff attempts, as the do-after is _longer_ than the recovery period. You recover roughly 30 stamina from do-after start to end, meaning it cannot apply unless you spam it multiple times or are attempting to run while cuffed after spamming uncuff attempts. Preventing spam of uncuff attempts is the whole point so this is all within reasonable scope.

No, I'm saying that having both mechanics is unusually punishing. Honestly both are extremely overkill and I would hate to see either merged.

Mixelz:
It is intended to be minor amount of damage, and only applicable for spam situations. Though we will see how it goes I suppose.
The discord channel mentioned turning the hardcoded value into a component so I'll try to get that done when able.

Mixelz:
> I was thinking a stam crit while being pulled shouldn't break the pull but it's better safe than sorry

Just confirmed, you are still stuck on the pull if you stam crit yourself while being pulled. 
I also just finished up turning the hardcode into a component so that will also be added in case anyone feels that zipties should deal less selfstamina damage than regular cuffs. 
Honestly I think that would be a good idea but i'm not going to make that call at the moment.

RedBookcase:
> No, I'm saying that having both mechanics is unusually punishing. Honestly both are extremely overkill and I would hate to see either merged.

I don't really see how either of these are unusually punishing? The shove on uncuff is optional depending on if you have combat mode on or off, and is more a QoL change for security players to make it easier to uncuff and deposit somebody in their cell. This PR only punishes repeated attempts to uncuff yourself in a very short period of time, which really shouldn't punish anybody who isn't spamming the option.

SlamBamActionman:
> wait, didn't the self-uncuffing have a cooldown?

It does, but it's so minor that the sound effect of cuffs jingling isn't over before you can try again. I don't really see this PR preventing people who want to be annoying and would rather see a longer cooldown, but I also see this as a very inoffensive way to discourage spamming uncuffing attempts. 

I wouldn't mind seeing this tested. I share none of the concerns Keron has expressed, considering this only apply to multiple self-uncuffs in a short period, and the other PR is for a specific scenario where the person uncuffing already has control over you.

SlamBamActionman (approved):
Might need to lower the stamina damage for zipties/improvised cuffs in the future due to the shorter doafter.

Mixelz:
My own thought for balance values was something like: 25 for cuffs and 15 for zipties. 
I didn't want to push people off the idea with a seemingly scary starting value for this idea though so I went with the 15 flat for both. 

In eithercase, hurray for approval!

keronshb (changes requested):
I'm still in the opposite camp. This should not be dealing stamina damage at all.
If anything it should have a ramping cooldown instead if used within a short amount of time OR the current spam prevention should be looked at, not arbitrarily adding stamina damage.

beck-thompson (changes requested):
Hey! So the maintainer vote concluded that:
1.) The cooldown for trying to release yourself from cuff should be increased to 10 seconds
2.) If the issue persists, then something like doing stamina damage should be added!

Let me know if you have any questions or need help :saluting_face:
Feel free to do the cooldown in this pr or make a new one either is fine

Mixelz:
I'll hibernate this PR and we can simply pull it back up when appropriate, will get to the new PR when I am available to do so. 👍

PR #36122, opened 2025-03-28
Remove redundant admin flag check in AdminVerbSystem.Tools

Tayrtahn:
## About the PR

Removes a bit of redundant logic in `AdminVerbSystem.Tools.cs`

## Why / Balance

It's not really hurting anything by being there, but it's a bit silly. Removing it might save a couple CPU cycles or something, maybe.

## Technical details

Screenshot_20250327-184926

```c#
if (!condition)
    return;

if (condition)
{
    // stuff
}
```
It's silly to have a guard clause and then do a check for the same thing after it.
Now it's just:
```c#
if (!condition)
    return;

// stuff
```
Diff looks large because everything had to be unindented.

## Requirements

### DIFF
--- a/Content.Server/Administration/Systems/AdminVerbSystem.Tools.cs
+++ b/Content.Server/Administration/Systems/AdminVerbSystem.Tools.cs
@@ -65,343 +65,340 @@ public sealed partial class AdminVerbSystem
         if (!_adminManager.HasAdminFlag(player, AdminFlags.Admin))
             return;
 
-        if (_adminManager.HasAdminFlag(player, AdminFlags.Admin))
+        if (TryComp<DoorBoltComponent>(args.Target, out var bolts))
         {
-            if (TryComp<DoorBoltComponent>(args.Target, out var bolts))
+            Verb bolt = new()
             {
-                Verb bolt = new()
-                {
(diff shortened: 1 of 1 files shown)


Content.Server/Administration/Systems/AdminVerbSystem.Tools.cs
```diff
@@ -65,343 +65,340 @@ private void AddTricksVerbs(GetVerbsEvent<Verb> args)
...
-                };
-                args.Verbs.Add(bolt);
-            }
+                    _door.SetBoltsDown((args.Target, bolts), !bolts.BoltsDown);
+                },
+                Impact = LogImpact.Medium,
+                Message = Loc.GetString(bolts.BoltsDown
+                    ? "admin-trick-unbolt-description"
+                    : "admin-trick-bolt-description"),
+                Priority = (int) (bolts.BoltsDown ? TricksVerbPriorities.Unbolt : TricksVerbPriorities.Bolt),
```
slarticodefast, line 85:
```suggestion
                Priority = (int)(bolts.BoltsDown ? TricksVerbPriorities.Unbolt : TricksVerbPriorities.Bolt),
```
Since we are messing up the git history anyways we might as well fix the formatting.
Same below

slarticodefast (approved):

PR #36123, opened 2025-03-28
Fix embedded projectile deletion not being tracked by container

Tayrtahn:
## About the PR

`EmbeddedContainerComponent` now properly tracks when embedded projectiles are deleted in any way, instead of only tracking when they are removed by hand.
Additionally, the `EmbeddedContainerComponent` is now removed from an entity when there are no more projectiles embedded into it.
Also added a missing Dirty call on `EmbeddedContainerComponent` when an entity is removed.

## Why / Balance

Fixes a [bug that was showing up in server logs](https://discord.com/channels/310555209753690112/900426319433728030/1354969365426213017).
For the second part, it's good to clean up components that aren't doing anything anymore.

## Technical details

- `EmbeddableProjectileComponent` now subscribes to `ComponentShutdown` and calls `EmbedDetach`.
- `EmbeddedContainerComponent` now removes itself when its `EmbeddedObjects` HashSet is empty.
- Added a test that throws two projectiles at a target and deletes them one at a time, making sure they are removed from the target's `EmbeddedObjects` HashSet correctly and that the `EmbeddedContainerComponent` is removed when both are gone.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/Embedding/EmbedTest.cs
+++ b/Content.IntegrationTests/Tests/Embedding/EmbedTest.cs
@@ -1,5 +1,6 @@
 using Content.IntegrationTests.Tests.Interaction;
 using Content.Shared.Projectiles;
+using Robust.Shared.GameObjects;
 using Robust.Shared.Network;
 
 namespace Content.IntegrationTests.Tests.Embedding;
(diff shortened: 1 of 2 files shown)


Tayrtahn:
Test fail is real. Looks like I exposed another bug that was hiding. Easy to fix.

TheShuEd (approved):

PR #36124, opened 2025-03-28
Layering for atmospheric pipes

chromiumboy:
## About the PR

This PR adds layering for atmospheric pipes - up to three pipes of the same type/orientation can now be built on a single tile (see the demonstration video below)

All pipes and atmospheric devices, when constructed, will be assigned to the default layer (referred to as the 'primary configuration'). Prototypes for the two alternative configurations for pipes and most atmospheric devices have been added for the benefit of mappers

To change the configuration of a pipe or atmospheric device, right-click on the entity while holding a screwdriver, open the 'Adjust' submenu, and select the new configuration. Pipes and atmospheric devices can be adjusted while anchored, unanchored, or being held. If the entity is being held, configurations can be cycled through using the 'z' key. The majority of atmospheric devices can be assigned to any of the three configurations with the exception of radiators (I wasn't satisfied with the sprites I tried to make for the alternative configurations...). Note that if a player tries to switch a pipe to a sublayer that is occupied, the pipe will become unanchored

This PR also adds the pipe manifold, a constructible multi-pronged pipe that connects pipes across sublayers

## Why / Balance

This is a requested feature that was ported from SS13. It will allow for additional flexibility/complexity in station atmospheric setups

## Technical details

The PR is relatively straightforward. The logic for the pipe layering is handled by the AtmosPipeLayersSystems and the associated component. The bulk of the changes are due to the addition of new sprites and yaml alterations
 
## Media

[Demonstration of pipe layering in action](https://youtu.be/b10nKdeLviY)

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Atmospheric pipes can now be built on three different sublayers, allowing for multiple pipe networks on a single tile
- add: Pipe manifolds can now be constructed to connect pipes across different sublayers


keronshb:
Finally, atmos mains eating GOOD.

Content.Shared/Atmos/Piping/PipeColorVisuals.cs
Partmedia, line 1: This doesn't look like it needed to be changed here and just increases the diff size and risk of conflicts.

Resources/Locale/en-US/verbs/verb-system.ftl
Partmedia, line 1: I feel like we can come up with a better name besides "Adjust", but it's also fine as-is.

Resources/Prototypes/Entities/Structures/Piping/Atmospherics/alt_layers.yml
```diff
@@ -0,0 +1,717 @@
...
+    layers:
+    - state: freezerOff
+      map: [ "enum.PowerDeviceVisualLayers.Powered" ]
+    - state: freezerPanelOpen
+      map: [ "enum.WiresVisualLayers.MaintenancePanel" ]
+    - state: pipe2
+      map: [ "enum.PipeVisualLayers.Pipe" ]
+      renderingStrategy: Default
+
+# GasThermoMachineHeater
```
Partmedia, line 254: Why do thermomachines need variants for all different layers? Can't you get a layered T to adapt from e.g. a third position to a primary position, and branch that off to this thermomachine?

I'm concerned that:

1. This bloats too many entitiy prototypes
2. This contributes to "messy layers" which is one of the reasons why others have been opposed to pipe layers

Partmedia: Ditto for a lot of the machines that follow.

Resources/Prototypes/Entities/Structures/Piping/Atmospherics/trinary.yml
Partmedia, line 1: This is a refactoring change that is blended in with this large PR.

Resources/Prototypes/Entities/Structures/Piping/Atmospherics/unary.yml
```diff
@@ -189,14 +189,14 @@
...
           color: "#990000"
+    - type: Appearance 
     - type: GenericVisualizer
       visuals:
        # toggle color of the unshaded light:
        enum.OutletInjectorVisuals.Enabled:
          enum.LightLayers.Unshaded:
            True: { color: "#5eff5e" }
            False: { color: "#990000" }
-    - type: Appearance
```
Partmedia, line 199: Changes like this really make reviewing annoying...

Partmedia (commented):
I did an initial first-pass review of primarily the changes to `PipeNode` and the addition of `AtmosPipeLayersComponent`. I did not review anything to do with the atmos monitoring console but since you wrote that, I trust what you did there is fine. Will come back to check `PipeRestrictOverlapSystem` and `GasPipeManifoldSystem`. I did not look at any sprites. Will give it a more extensive in-game test soon.

Overall:

- Thanks for working on this.
- I think the changes to core data structures are sound and well designed.
- Could you investigate whether it would make sense to have `AtmosPipeLayersComponent` rolled into another class? For example, if you wanted to make a special kind of pipe (e.g. radiator) that only supports one layer, that would be the "trivial" case of filling out everything in `AtmosPipeLayersComponent` except only with one layer, right? Or, is this already how this is used?
- I'm concerned about the number of machines with multi-layer variants. As I mentioned in the in-line comments, I think this contributes too much to prototype bloat and also might encourage "messy" atmos designs that are hard to understand. I would personally prefer to have most machines in primary configurations that you would T out. I think some past maintainers that have chimed in about this matter have said something similar, though I would have to dig out the context to confirm.

metalgearsloth:
With disposals PR can probably predict most of this so will work on getting the reviews done for that soon.

ArtisticRoomba:
my mouth is watering and my brain is filled with the bottomless possibilities of 3 layers to play with. unreal.

Minemoder5000:
Would it be possible to add cycling through pipe layers in the construction menu? Maybe hijack the keybind for flipping, or something similar?

LaCumbiaDelCoronavirus:
Do pipe layers have to be switched between layers when pipes are placed, or are there multiple variants for each layer in the construction menu?

ScarKy0:
And it's not even roomba's birthday

VerinSenpai:
oh my god. its... its.... *dies*

Partmedia:
While I understand the excitement about this, please keep discussions on GitHub on-topic and review-related. This isn't a chat room. Thanks.

VerinSenpai:
if it hasn't been done already, would it be possible to include a hot key for switching between the states, similar to how rotating is done?

TurboTrackerss14:
Gunna let the real atmos code gamers manage the code side. 
But the readability of some of these pipes is... not great. 

Canister covering port can/will lead to incorrect layering. 
image

Pipelayers resetting to "center" to connect to devices will lead to "why are these two pipes not connected, they line up" ahelps
image

Overlapping manual valves handles will also lead to "its not active why it pump" ahelps
image

Auto linking pipes, bad awful, something will explode surely. (maybe only allow layering changes on unanchored?)
image

chromiumboy:
> Would it be possible to add cycling through pipe layers in the construction menu? Maybe hijack the keybind for flipping, or something similar?

I considered this, but you'd need a separate construction graph for each layer for every pipe or atmos device, and I didn't want to do that :P It would also bloat the construction menu, or require that the menu be updated to support layers. Best bet is to implement the RPD, it should be used for large scale pipe net constructions in any case

Resources/Prototypes/Entities/Structures/Piping/Atmospherics/alt_layers.yml
chromiumboy: I tried making an automatic placement system, but the solution was probably over-engineered and [the result was a bit jittery](https://youtu.be/sIdewKx9Q3w)

Going to stick with the separate entities prototypes for each layer now, for the pipes at least - I removed the alternative prototypes for the atmos devices

chromiumboy:
Sounds like a bug, I'll have a look. All configs should need a short do after to cycle them

keronshb:
Wouldn't some of the changes like changing the gas pipe sensor parent from GasPipeBase > GasBinaryBase and changing the pipe states from pipeStraight > pipeBinaryConnector count as breaking changes? Otherwise looks fine to me.

chromiumboy:
> Wouldn't some of the changes like changing the gas pipe sensor parent from GasPipeBase > GasBinaryBase and changing the pipe states from pipeStraight > pipeBinaryConnector count as breaking changes? Otherwise looks fine to me.

I'm not sure how these changes might potentially break something. Did you have something in mind?

ArtisticRoomba (approved):
The YAML and RSI side looks good to me.

I've left some comments on the C# side of the PR (remember that I don't know much), so feel free to take them into consideration or disregard if necessary.

Content.Shared/Atmos/Components/AtmosPipeLayersComponent.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// <summary>
+    /// The RSI path that the entity sprite will use for each pipe layer.
+    /// If null, the RSI path does not differ between pipe layers.
+    /// </summary>
+    /// <remarks>
+    /// If not null, it is assumed that there will be an array entry for each pipe layer
+    /// (from 0 to <see cref="NumberOfPipeLayers"/> - 1).
+    /// </remarks>
+    [DataField]
+    public string[]? SpriteRsiPaths = null;
```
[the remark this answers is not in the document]
Partmedia: Why make it nullable at all? Why not a `string[]` that is left empty if not specified?

chromiumboy:
> The YAML and RSI side looks good to me.
> 
> I've left some comments on the C# side of the PR (remember that I don't know much), so feel free to take them into consideration or disregard if necessary.

Thanks for the review :)

Partmedia:
Perhaps you missed this in the last pass, but why does `public string[]? SpriteRsiPaths;` have to be nullable Wouldn't checking that it is a zero-length list be the same?

chromiumboy:
> Perhaps you missed this in the last pass, but why does `public string[]? SpriteRsiPaths;` have to be nullable? Wouldn't checking that it is a zero-length list be the same?

The idea was that a length zero array would indicate that it hadn't been initiated correctly, while a null array meant it was never initialized at all. But I guess it doesn't really matter in this context? I can change it

Partmedia:
> > Perhaps you missed this in the last pass, but why does `public string[]? SpriteRsiPaths;` have to be nullable? Wouldn't checking that it is a zero-length list be the same?
> 
> The idea was that a length zero array would indicate that it hadn't been initiated correctly, while a null array meant it was never initialized at all. But I guess it doesn't really matter in this context? I can change it

I'm not sure what the distinction between these two states are: both of them result in the same condition of "layer sprites should not be used" in the code, right? There's no extra handling between the two cases?

chromiumboy:
> > > Perhaps you missed this in the last pass, but why does `public string[]? SpriteRsiPaths;` have to be nullable? Wouldn't checking that it is a zero-length list be the same?
> > 
> > 
> > The idea was that a length zero array would indicate that it hadn't been initiated correctly, while a null array meant it was never initialized at all. But I guess it doesn't really matter in this context? I can change it
> 
> I'm not sure what the distinction between these two states are: both of them result in the same condition of "layer sprites should not be used" in the code, right? There's no extra handling between the two cases?

Yeah, there's no difference in how they are handled here

southbridge-fur:
It's beautiful

Partmedia:
Instead of normal, alt 1, and alt 2, I think a better set of names would be Layer 1, Layer 2, and Layer 3:
image

Then you can make it match here:
image

It would be nice if the aghost could skip the screwdriver check to adjust. Perhaps "Adjust" should be renamed to "Adjust Layer".

Some sprites still need adjustment to correctly show where the pipes connect:
image

But I understand that this is a lot of work and not necessarily in scope for this PR.

I think the interface for constructing the pipes via the construction menu was a bit clunky and liked the interface you showed earlier with the layer snapping.
image

Partmedia:
Changing the `CurrentPipeLayer` in the `AtmosPipeLayerComponent` does not cause the visualizer to correctly update.

chromiumboy:
> Instead of normal, alt 1, and alt 2, I think a better set of names would be Layer 1, Layer 2, and Layer 3
> 
> Then you can make it match here:

Refering to them as different 'layers' might be ambiguous for new players, and using a linear numbering system seems a bit weird since the first layer starts with the pipe in the middle of the tile. Happy to debate the merits of either approach though

> It would be nice if the aghost could skip the screwdriver check to adjust. Perhaps "Adjust" should be renamed to "Change Layer".

Agreed that aghost should skip the tool check. I thought that having a generic 'adjust' verb might be useful if someone wanted to reuse it, as 'change layers' is very specific

> Some sprites still need adjustment to correctly show where the pipes connect
>
> But I understand that this is a lot of work and not necessarily in scope for this PR.

It should be simple, the yaml prototype likely just needs updating

> I think the interface for constructing the pipes via the construction menu was a bit clunky and liked the interface you showed earlier with the layer snapping.

I can add it back in if you wanna test it out and see what you prefer

chromiumboy:
> Changing the `CurrentPipeLayer` in the `AtmosPipeLayerComponent` does not cause the visualizer to correctly update.
> 
> If the only intended way for this to be adjusted is the right click menu, then the VV field should be set to read-only.

Yeah, it's not intended to be changed via VV

Partmedia:
> Refering to them as different 'layers' might be ambiguous for new players, and using a linear numbering system seems a bit weird since the first layer starts with the pipe in the middle of the tile. Happy to debate the merits of either approach though

Going middle, left, right for "Primary", "Secondary", and "Tertiary" seem very confusing and easy to forget to me.

Especially since there's a mix of two different types of names, e.g. "Secondary" but "1st Alt" in the entity spawn menu.

A linear ordering still makes more sense to me, but I suppose you can wait for some other input and see if we can come to a consensus on what the best organization might be.

ArtisticRoomba:
Did an in-game test, overall, it went well!
Content Client_9kyFHxyPpU
image

As you can see, there appears to be some visual bugs with the volumetric pump.

Additionally, using a gas analyzer on any manifold is also quite a pain:

image

It seems to keep adding another pipenode view onto the next one every time it updates.

Feedback from things that I observed, in no particular order:
- Building pumps in different configs is quite annoying as you have to switch their configuration and then anchor them in order to get them where you want.
- You can't stack 3 pumps on one tile, each in their different configuration.
- Looking at the verb to switch the configuration every single time is really, really slow. Maybe an alt-click interaction with a screwdriver would be better?
	- Doing this with a drill is even slower as you sometimes forget to have the drill in screwing mode, so when you go through the work to look at the "adjust" verb it says you can't do anything and you have to look at the verb all over again.
- Pipes constantly give the construction hint saying they can be turned into their different variants (seems to be only the straight variant).

Overall, what an amazing system! It really works well, I didn't really have any major problems with tinkering around with making a new TEG design.

chromiumboy:
Cooking some more on this

ArtisticRoomba:
Tried out the contextual pipe placement for a bit, was extremely cool!

Content Client_EeQLVtzfyN

Seemed a bit weird behaving originally but the technique is to move your mouse diagonally. A possible suggestion is to draw a diagonal yellow line that indicates layer placements?

A bug I observed is that every time I made a layer change it seemed to cancel whatever keyboard input I was doing. So walking around became stutterey when I moved my mouse to move the pipe ghost between different pipe layers.

chromiumboy:
> A bug I observed is that every time I made a layer change it seemed to cancel whatever keyboard input I was doing. So walking around became stutterey when I moved my mouse to move the pipe ghost between different pipe layers.

It looks like it's an engine side issue, I'll need to make a separate PR to address that. I should also make a PR that will allow multiple pipe construction ghosts to be present on the same tile - at the moment, the system is limited to one per tile

I made some changes to how manifolds look - they now only show their connector as needed (including on the atmos monitor)

manifold   manifold-map

VerinSenpai:
now THAT is beautiful.

ArtisticRoomba:
Did another playtest, was extremely fun!
image

Some feedback, in no particular order:
- The contextual pipe placement system is REALLY powerful. Like, I built the above image very quickly with very little hiccups at all.
- The auto-adjust-config on screwdriver use is really powerful and allows you to switch up your setups really fast. It's no longer clumsy to use.
- Attempting to adjust a vent that is in pressure lockout will start two do-afters, one for adjusting the configuration of the device, and the other for releasing the lockout.
- The ability to adjust configurations on devices that are under tiles (ex. air vents, scrubbers) is possible, which is a problem.
- Building large straight lines of pipe using CTRL is no longer possible.
- Building pumps in different configurations is still a bit finicky, I have to adjust the config before placing it down, which makes it a hassle for tight spaces. Ideally all devices that have pipe layering options can be placed down in the configuration you want in the ghost builder using the 3-dots guide (scope inflator 9000 though).

Oh I love it.

Content.Shared/Atmos/Components/AtmosPipeLayersComponent.cs
```diff
@@ -0,0 +1,110 @@
...
+    [ViewVariables(VVAccess.ReadOnly)]
+    public byte CurrentPipeLayer;
+
+    /// <summary>
+    /// The RSI path that the entity sprite will use for each pipe layer.
+    /// If null, the RSI path does not differ between pipe layers.
+    /// </summary>
+    /// <remarks>
+    /// If not null, it is assumed that there will be an array entry for each pipe layer
+    /// (from 0 to <see cref="NumberOfPipeLayers"/> - 1).
```
Partmedia, line 38: This comment is no longer relevant.

Qiler:
If you move a pipe to a layer that is occupied, it will unwrench the that pipe but wont remove gasses from it.

chromiumboy:
> If you move a pipe to a layer that is occupied, it will unwrench the that pipe but wont remove gasses from it.

Good catch, I'll fix that. Thanks for letting me know

Partmedia:
In-game test looks good to me. Will hopefully get back to make another pass at the code soon.

Dicerson:
Can we see an example of what the layering looks like for actual devices, and not just pipes? For example stacked filters, radiators, ports, scrubbers, etc.? Especially with adjacent pipes or manifolds to see how all those connections lay out?

ArtisticRoomba:
> Can we see an example of what the layering looks like for actual devices, and not just pipes? For example stacked filters, radiators, ports, scrubbers, etc.? Especially with adjacent pipes or manifolds to see how all those connections lay out?

I have provided plenty of examples if you look at the PR's comments.

chromiumboy:
> > Can we see an example of what the layering looks like for actual devices, and not just pipes? For example stacked filters, radiators, ports, scrubbers, etc.? Especially with adjacent pipes or manifolds to see how all those connections lay out?
> 
> I have provided plenty of examples if you look at the PR's comments.

I stole one of your examples and added it to the media section of the PR description ;)

BTW, do you have any time coming up soon to do that demonstration video?

ArtisticRoomba:
Oh, sure, I'll get that done! I was crammed with finals for the previous 2 weeks and Thursday is my last final, so I'll be more free soon. I spent today and yesterday reviewing Xenoborgs and mapping PRs. So I'll have a little video out for you in the coming days.

Partmedia:
> Just following up to see how it's going. It would be nice to have this PR ready to merge by the weekend so that there will be time to test it on Vulture before the staging cut-off next week

Sorry about the delay; I'm currently traveling and so it is unlikely I'll be able to finish looking by the weekend. If anyone else is able to complete a code review, that would be great.

Dicerson:
> > Can we see an example of what the layering looks like for actual devices, and not just pipes? For example stacked filters, radiators, ports, scrubbers, etc.? Especially with adjacent pipes or manifolds to see how all those connections lay out?
> 
> I have provided plenty of examples if you look at the PR's comments.

I see only a single image providing examples, and in that image there are no instances of filters stacked on the same tile from different layers, nor any kind of stacking between devices that aren't radiators (which can already be stacked on top of stuff currently). The video also does not show the stacking of anything that isn't just pipes.

Am I to take it that only pure pipes benefit from this, and that filters and other devices are not included in the layering?

chromiumboy:
> > > Can we see an example of what the layering looks like for actual devices, and not just pipes? For example stacked filters, radiators, ports, scrubbers, etc.? Especially with adjacent pipes or manifolds to see how all those connections lay out?
> > 
> > 
> > I have provided plenty of examples if you look at the PR's comments.
> 
> I see only a single image providing examples, and in that image there are no instances of filters stacked on the same tile from different layers, nor any kind of stacking between devices that aren't radiators (which can already be stacked on top of stuff currently). The video also does not show the stacking of anything that isn't just pipes.
> 
> Am I to take it that only pure pipes benefit from this, and that filters and other devices are not included in the layering?

Correct, you can still only have one device per tile

Content.Server/Atmos/EntitySystems/PipeRestrictOverlapSystem.cs
```diff
@@ -97,27 +97,27 @@ public bool CheckOverlap(Entity<NodeContainerComponent, TransformComponent> ent)
...
+        var otherDirsAndLayers = GetAllDirectionsAndLayers(other).ToList();
 
-        foreach (var dir in entDirs)
+        foreach (var (dir, layer) in entDirsAndLayers)
         {
-            foreach (var otherDir in otherDirs)
+            foreach (var (otherDir, otherLayer) in otherDirsAndLayers)
             {
-                if ((dir & otherDir) != 0)
+                if ((dir & otherDir) != 0 && layer == otherLayer)
```
Partmedia, line 107: is `!= 0` needed?

Content.Server/Atmos/Piping/EntitySystems/AtmosPipeAppearanceSystem.cs
```diff
@@ -34,9 +34,12 @@ private void UpdateAppearance(EntityUid uid, AppearanceComponent? appearance = n
         if (!TryComp<MapGridComponent>(xform.GridUid, out var grid))
             return;
 
+        var numberOfPipeLayers = TryComp<AtmosPipeLayersComponent>(uid, out var atmosPipeLayers) ? atmosPipeLayers.NumberOfPipeLayers : 1;
```
Partmedia, line 37: This got repeated again here.

Content.Server/Atmos/Piping/EntitySystems/GasPipeManifoldSystem.cs
Partmedia, line 1: Why does a manifold behave any differently compared to a 6-way pipe?

Content.Server/NodeContainer/Nodes/PipeNode.cs
```diff
@@ -24,6 +24,12 @@ public partial class PipeNode : Node, IGasMixtureHolder, IRotatableNode
         [DataField("pipeDirection")]
         public PipeDirection OriginalPipeDirection;
 
+        /// <summary>
+        ///     The *current* layer to which the pipe node is assigned.
+        /// </summary>
+        [DataField("pipeLayer")]
+        public byte CurrentPipeLayer;
```
Partmedia, line 31: I wonder if this should remain a `byte`, or if it should gain a new `PipeLayer` type (like the above `PipeDirection`).

Content.Shared/Atmos/Components/AtmosPipeLayersComponent.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// If the array is not empty there should be an entry for each pipe layer
+    /// (from 0 to <see cref="NumberOfPipeLayers"/> - 1).
+    /// </remarks>
+    [DataField]
+    public string[] SpriteRsiPaths = [];
+
+    /// <summary>
+    /// A dictionary of entity sprite layers that have their
+    /// RSI paths updated when the pipe layer changes.
+    /// </summary>
```
Partmedia, line 46: Based on this comment alone, I don't understand how this is different from `SpriteRsiPaths`

Content.Shared/Atmos/Components/AtmosPipeLayersComponent.cs
```diff
@@ -0,0 +1,110 @@
...
+    [DataField]
+    public float Delay = 1f;
+}
+
+/// <summary>
+/// Raised when a player attempts to cycle a pipe to its next layer
+/// </summary>
+[Serializable, NetSerializable]
+public sealed partial class TryCyclingPipeLayerCompletedEvent : SimpleDoAfterEvent;
```
Partmedia, line 88: Instead of `TryCycling`, perhaps `TrySetNext` or `TryNext`?

Content.Shared/Atmos/Components/PipeAppearanceComponent.cs
```diff
@@ -5,6 +5,8 @@ namespace Content.Shared.Atmos.Components;
 [RegisterComponent]
 public sealed partial class PipeAppearanceComponent : Component
 {
-    [DataField("sprite")]
-    public SpriteSpecifier.Rsi Sprite = new(new("Structures/Piping/Atmospherics/pipe.rsi"), "pipeConnector");
+    [DataField]
+    public SpriteSpecifier.Rsi[] Sprite = [new(new("Structures/Piping/Atmospherics/pipe.rsi"), "pipeConnector"),
+        new(new("Structures/Piping/Atmospherics/pipe_alt1.rsi"), "pipeConnector"),
+        new(new("Structures/Piping/Atmospherics/pipe_alt2.rsi"), "pipeConnector")];
```
Partmedia, line 11: Are these sprite paths validated?

Content.Shared/Atmos/Consoles/Components/AtmosMonitoringConsoleDeviceComponent.cs
```diff
@@ -16,6 +16,9 @@ public sealed partial class AtmosMonitoringConsoleDeviceComponent : Component
     /// entity on the atmos monitoring console nav map.
     /// If null, no blip is drawn (i.e., null for pipes)
     /// </summary>
-    [DataField, ViewVariables]
+    [DataField]
     public ProtoId<NavMapBlipPrototype>? NavMapBlip = null;
+
+    [DataField]
+    public bool ShowAbsentConnections = true;
```
Partmedia, line 23: Missing documentation.

Content.Shared/Atmos/Consoles/SharedAtmosMonitoringConsoleSystem.cs
```diff
@@ -15,7 +15,7 @@ public override void Initialize()
 
     private void OnGetState(EntityUid uid, AtmosMonitoringConsoleComponent component, ref ComponentGetState args)
     {
-        Dictionary<Vector2i, Dictionary<(int, string), ulong>> chunks;
+        Dictionary<Vector2i, Dictionary<(int, byte, string), ulong>> chunks;
```
Partmedia, line 18: This tuple `(int, byte, string)` is used in so many different places. It deserves a named type. See DRY principle

Content.Shared/Atmos/EntitySystems/SharedAtmosPipeLayersSystem.cs
```diff
@@ -0,0 +1,246 @@
...
+        if (!_protoManager.TryIndex(ent.Comp.Tool, out var toolProto))
+            return;
+
+        var user = args.User;
+        var isCovered = TryComp<SubFloorHideComponent>(ent, out var subFloorHide) && subFloorHide.IsUnderCover;
+        var isMissingTool = !TryGetHeldTool(user, ent.Comp.Tool, out var tool);
+
+        if (isCovered || isMissingTool)
+        {
+            var label = "???";
```
Partmedia, line 64: Dead store, since one of `isCovered` or `isMissingTool` must be set. Consider reworking logic.

Resources/Locale/en-US/atmos/atmos-pipe-layers.ftl
```diff
@@ -0,0 +1,10 @@
+atmos-pipe-layers-component-layer-0 = primary 
+atmos-pipe-layers-component-layer-1 = secondary
+atmos-pipe-layers-component-layer-2 = tertiary
+
+atmos-pipe-layers-component-change-layer = Adjusted to its {$layerName} configuration.
+atmos-pipe-layers-component-current-layer = It is in its {$layerName} configuration.
+atmos-pipe-layers-component-select-layer = {CAPITALIZE($layerName)} configuration
+atmos-pipe-layers-component-tool-missing = Requires {INDEFINITE($toolName)} {$toolName}
+atmos-pipe-layers-component-pipes-are-covered = The pipes are covered
+atmos-pipe-layers-component-cannot-adjust-pipes = You need to uncover the pipes before they can be adjusted.
```
Partmedia, line 10: Why not merge these two into the same string?

Resources/Locale/en-US/atmos/atmos-pipe-layers.ftl
```diff
@@ -0,0 +1,10 @@
+atmos-pipe-layers-component-layer-0 = primary 
+atmos-pipe-layers-component-layer-1 = secondary
+atmos-pipe-layers-component-layer-2 = tertiary
+
+atmos-pipe-layers-component-change-layer = Adjusted to its {$layerName} configuration.
+atmos-pipe-layers-component-current-layer = It is in its {$layerName} configuration.
+atmos-pipe-layers-component-select-layer = {CAPITALIZE($layerName)} configuration
+atmos-pipe-layers-component-tool-missing = Requires {INDEFINITE($toolName)} {$toolName}
```
Partmedia, line 8: Complete sentence

Content.Client/Atmos/AlignAtmosPipeLayers.cs
```diff
@@ -0,0 +1,200 @@
...
+        if (pManager.PlacementType == PlacementTypes.None)
+        {
+            var gridRotation = _transformSystem.GetWorldRotation(gridUid.Value);
+            var worldPosition = _mapSystem.LocalToWorld(gridUid.Value, Grid, MouseCoords.Position);
+            var direction = (_eyeManager.CurrentEye.Rotation + gridRotation + Math.PI / 2).GetCardinalDir();
+            var multi = (direction == Direction.North || direction == Direction.South) ? -1f : 1f;
+
+            args.WorldHandle.DrawCircle(worldPosition, GuideRadius, _guideColor);
+            args.WorldHandle.DrawCircle(worldPosition + gridRotation.RotateVec(new Vector2(multi * GuideOffset, GuideOffset)), GuideRadius, _guideColor);
+            args.WorldHandle.DrawCircle(worldPosition - gridRotation.RotateVec(new Vector2(multi * GuideOffset, GuideOffset)), GuideRadius, _guideColor);
```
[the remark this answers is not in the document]
chromiumboy: I wanted something that wasn't too obtrusive. And also something that I didn't have to worry about rotation with - it's a real pain

Content.Server/Atmos/Consoles/AtmosMonitoringConsoleSystem.cs
```diff
@@ -282,16 +282,11 @@ private bool TryGetAtmosDeviceNavMapData
             return false;
 
         var direction = xform.LocalRotation.GetCardinalDir();
+        var netId = TryGettingFirstPipeNode(uid, out var _, out var firstNetId) ? firstNetId : -1;
+        var color = TryComp<AtmosPipeColorComponent>(uid, out var atmosPipeColor) ? atmosPipeColor.Color : Color.White;
+        var layer = TryComp<AtmosPipeLayersComponent>(uid, out var atmosPipeLayers) ? atmosPipeLayers.CurrentPipeLayer : (byte)0;
```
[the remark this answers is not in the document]
chromiumboy: It's the default layer

Content.Server/Atmos/Piping/EntitySystems/GasPipeManifoldSystem.cs
chromiumboy: The manifold connects across different pipe layers; all other basic pipes do not

Resources/Locale/en-US/atmos/atmos-pipe-layers.ftl
chromiumboy: Not sure what you mean

Content.Server/Atmos/Consoles/AtmosMonitoringConsoleSystem.cs
```diff
@@ -282,16 +282,11 @@ private bool TryGetAtmosDeviceNavMapData
             return false;
 
         var direction = xform.LocalRotation.GetCardinalDir();
+        var netId = TryGettingFirstPipeNode(uid, out var _, out var firstNetId) ? firstNetId : -1;
```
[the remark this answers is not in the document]
chromiumboy: I haven't been able to find an invalid number for net IDs

Content.Server/Atmos/EntitySystems/PipeRestrictOverlapSystem.cs
chromiumboy: It's required, it checks whether the two pipes have nodes with the same orientation. My modification adds that the pipe nodes also have to be on the same layer to count as an overlap

Content.Client/Atmos/Consoles/AtmosMonitoringConsoleSystem.cs
```diff
@@ -15,7 +15,7 @@ public override void Initialize()
 
     private void OnHandleState(EntityUid uid, AtmosMonitoringConsoleComponent component, ref ComponentHandleState args)
     {
-        Dictionary<Vector2i, Dictionary<(int, string), ulong>> modifiedChunks;
```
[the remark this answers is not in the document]
chromiumboy: It's a bitmask look up for atmos pipes, 1 for occupied and 0 for empty. Indexed by the net ID, layer, and color hexcode of the pipe. Based on your other comments I'll turn this into a data struct of some sort

Content.Shared/Atmos/Components/AtmosPipeLayersComponent.cs
chromiumboy: If SpriteRsiPaths is set, the RSI path for the whole sprite is changed. If SpriteLayersRsiPaths is set, only the specified layer has it RSI path changed

Content.Client/Atmos/UI/GasAnalyzerWindow.xaml.cs
```diff
@@ -136,6 +136,7 @@ public void Populate(GasAnalyzerUserMessage msg)
                 else
                 {
                     // oh shit of fuck its more than 4 this ui isn't gonna look pretty anymore
+                    CDeviceMixes.RemoveAllChildren();
```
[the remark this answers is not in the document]
Partmedia: Having this explained as a comment would be nice, since otherwise it's not very obvious and someone could accidentally remove it.

Partmedia (approved):
I'm going to stamp "approved" to satisfy the merge requirements, in case I get too busy again soon. Overall things look reasonable. If you are able to get around to it, I think the top items before merging this (as opposed to merging then iterating later) are:

1. Adding a data struct for the `(int, byte, string)` tuple.
2. Adding a method that does the `TryComp<AtmosPipeLayersComponent>(uid, out var atmosPipeLayers) ? atmosPipeLayers.NumberOfPipeLayers : 1` check.
3. Consider if `byte` or some other type is better suited to hold the layer number.

chromiumboy:
> I'm going to stamp "approved" to satisfy the merge requirements, in case I get too busy again soon. Overall things look reasonable. If you are able to get around to it, I think the top items before merging this (as opposed to merging then iterating later) are:
> 
> 1. Adding a data struct for the `(int, byte, string)` tuple.
> 2. Adding a method that does the `TryComp<AtmosPipeLayersComponent>(uid, out var atmosPipeLayers) ? atmosPipeLayers.NumberOfPipeLayers : 1` check.
> 3. Consider if `byte` or some other type is better suited to hold the layer number.

I've addressed most of your comments, the outstanding ones I've left unresolved. If you could let me know what's needed, I can address them as well

Content.Client/Atmos/EntitySystems/AtmosPipeAppearanceSystem.cs
```diff
@@ -61,33 +73,43 @@ private void OnAppearanceChanged(EntityUid uid, PipeAppearanceComponent componen
...
-        // transform connected directions to local-coordinates
-        var connectedDirections = worldConnectedDirections.RotatePipeDirection(-Transform(uid).LocalRotation);
-
-        foreach (var layerKey in Enum.GetValues<PipeConnectionLayer>())
+        for (byte i = 0; i < numberOfPipeLayers; i++)
         {
-            if (!_sprite.LayerMapTryGet((uid, args.Sprite), layerKey, out var key, false))
-                continue;
+            // Extract the pipe direction for the current layer
+            var pipeLayerConnectedDirections = (PipeDirection)(15 & (worldConnectedDirections >> (PipeDirectionHelpers.PipeDirections * i)));
```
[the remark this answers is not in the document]
Partmedia: I would still prefer a comment or explanation instead of using a magic number. This is like `UINT16_MAX` or something.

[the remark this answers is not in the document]
Partmedia: But really, this represents something like `ALL_DIRECTIONS_BITMAP`.

Content.Client/Atmos/EntitySystems/AtmosPipeAppearanceSystem.cs
```diff
@@ -61,36 +72,51 @@ private void OnAppearanceChanged(EntityUid uid, PipeAppearanceComponent componen
...
+                    continue;
+
+                layer.Color = color;
+            }
         }
     }
 
+    private int GetNumberOfPipeLayers(EntityUid uid, out AtmosPipeLayersComponent? atmosPipeLayers)
+    {
+        return TryComp(uid, out atmosPipeLayers) ? atmosPipeLayers.NumberOfPipeLayers : 1;
```
Partmedia, line 117: Is `Resolve` more appropriate here?

chromiumboy: I don't think its needed. Resolve lets you optionally pass in a component into the function, and if you don't supply it, Resolve will get it for you. But in both cases, I don't have the component to pass in optionally

Content.Client/Atmos/EntitySystems/AtmosPipeAppearanceSystem.cs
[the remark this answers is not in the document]
chromiumboy: Yes, it's a bit mask that's being used to extract the cardinal pipe orientations of the current layer from worldConnectedDirections. I've updated the comment to add more info

Content.Client/Atmos/EntitySystems/AtmosPipeAppearanceSystem.cs
Partmedia: This was a mistake on my part. Somehow I didn't see the `out` part of this function declaration, so I misunderstood what this function did. My bad.

Partmedia:
Thanks for the changes. Please go ahead and merge this if you're finished with the last touch-ups.

ArtisticRoomba (approved):
Oh, it's so beautiful.

VerinSenpai:
*hyperventilating intensifies*

Nukesthestation:
atmos will have enough steam coming from my ears to supperheat the station as I try and figure out how I do basic atmos now :trollface:

AndrewTateOfficialSS14:
Found a bug with this layering code, it says it is unable to place a pipe because it can;t fit over other pipe, but in theory it would fit...
image

MidZik:
> Found a bug with this layering code, it says it is unable to place a pipe because it can;t fit over other pipe, but in theory it would fit...

It's ambiguous which device should receive input. Both the manifold and the pipe bend want input from the secondary config in the tile above.

Unrelated, are there any plans to allow multiple pipe ghosts to be placed on the same tile if they're in different configurations?

chromiumboy:
> Found a bug with this layering code, it says it is unable to place a pipe because it can;t fit over other pipe, but in theory it would fit... image

@MidZik is correct, you can't have a pipe with the same orientation as the manifold on the same tile

> Unrelated, are there any plans to allow multiple pipe ghosts to be placed on the same tile if they're in different configurations?

I would like to see this happen eventually. I'm taking a break from the development side of things at the moment though. I imagine it wouldn't be too difficult to implement. Instead of having a single construction prototype per grid coordinate, you'd have a list of construction prototypes and their rotation

PR #36127, opened 2025-03-28
Another overlay stack sorting fix

Errant-4:
## About the PR
Fix for overlay stacks (again). I am also looking into better fixes for smart sorting, but this is a simple quick fix that will make stacks usable, and supporting configurable sorting methods is generally useful anyway.

Overlay stack order can now be configured. Currently only available through console, via cvar `ui.admin_overlay_stack_sort_by`, to avoid conflicting with the admin options menu revamp on #35359. 
Options are "Character", "User" and "Vertical"(the current one)

If it works out on master, I would like to also cherrypick this for staging

## Why / Balance
The horrors persist

## Technical details
The list containing visible players that need their overlay drawn onscreen is passed to a switch to be ordered accoring to cvar setting. The sorted list is then passed along for iteration.

## Media

https://github.com/user-attachments/assets/12ee2f58-5999-4518-9690-f5eecdc0948b

## Requirements

## Breaking changes
None

**Changelog**
:cl: Errant
ADMIN:
- tweak: Admin overlay stacks are now sorted by character name, and should no longer jump around. Sorting can be configured via cvar ui.admin_overlay_stack_sort_by < Character | User | Vertical >

### DIFF
--- a/Content.Client/Administration/AdminNameOverlay.cs
+++ b/Content.Client/Administration/AdminNameOverlay.cs
@@ -32,6 +32,7 @@ internal sealed class AdminNameOverlay : Overlay
     private float _ghostHideDistance;
     private int _overlayStackMax;
     private float _overlayMergeDistance;
+    private AdminOverlayStackSortBy _overlayStackSortBy;
 
     //TODO make this adjustable via GUI
     private readonly ProtoId<RoleTypePrototype>[] _filter =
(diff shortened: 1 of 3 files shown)


slarticodefast:
Does this have to be a cvar? I don't really see a need to customize this.
Just making it sorted alphabetically by username should be fine.

Errant-4:
> Does this have to be a cvar? I don't really see a need to customize this. Just making it sorted alphabetically by username should be fine.

Making it possible to swap it out on the fly means that different versions of it can be tested on the same branch. This is difficult to properly test in dev, so having a fallback always available in addition to the one or more experimental versions would be a good thing

slarticodefast:
It will still lead to a branching of the code path and the removal of that and the cvar will lead to a bigger breaking change once testing is done. Or we will leave it in and it will be derelict code that is not used. I would prefer some additional testing and doing a hotfix in case further problems occurr.

slarticodefast:
I don't really see the need for having sorting options at all though. Just one working, consistent version is enough. And having too many settings in the options menu is a form of bloat as well.
Is there any advantage of having it ordered by position rather than alphabetically? The former will always cause problems if a player makes vertical zig-zagging movements. Or if you rotate the camera.

Errant-4:
Then we are at an impasse

Errant-4:
Situation 1: players moving to/passing each other at a distance (say, half to one tile away)
At this range, the players can be visually separated, and might just be passing each other. Here, vertical sorting has the chance to look WAY better than alphabetical, particularly if the players are moving horizontally, so the "sort sepraration threshold" is not crossed. Alphabetical would be mostly servicable.

Situation 2: players crowding over each other, less than half a tile away. 
At this range, players visually "mesh together". In this case, order is irrelevant. Alphabetical would be equally sufficient, and potentially better if the players crossed each other's y axis causing flips

Situation 3: players parented to another player. Like orbiting ghosts, or someone carrying mice/pAI
In this situation, alphabetical and (stable) vertical sorting would be equally functional. However, in this situation a third approach would be superior to both: if the parented entities were sorted lower. This could get very complicated if the parenting is not direct, such as if the mouse/pAI is in a box in the player's backpack

Issue #34041, opened 2024-12-23
Stackable entities sapwned with entity tables don't stack


IamVelcroboy:
## Description
I noticed this when converting my smuggler satchel PR from the old fill system to the new tables. But items like materials, cash, etc that should stack, don't stack and take up a lot more space, limiting the amount of items that can be in the pool.

**Reproduction**
Set up a container fill using entity tables and set the number to spawn to be higher than the capacity of the bag(in terms of grid space) Spawn container, some items won't fit because they aren't stacking.

**Screenshots**
n/a

**Additional context**
n/a

IProduceWidgets:
Elevated difficulty because this is gonna require interfacing stacks to robust

Fildrance:
For problematic spawns we should just set 
'autoStack: true'

PR #36146, opened 2025-03-29
Fix smuggler satchels heisentesting

ArtisticRoomba:
## About the PR
Fixes smuggler satchels by ensuring that the maximum possible value we can roll is less than what we can fit in the bag. By counting.

~This PR also modifies RangeNumberSelector in EntityTables to actually be inclusive. Because for some reason it was being inclusive + 1 and returning five bottles of whiskey when I requested 4 at max.~

Requires content PR https://github.com/space-wizards/space-station-14/pull/36146

NOTE:
An alternative solution to this problem is to create a new NestedSelectorStorageLimited type in the EntityTables system that simply halts the attempt to spawn more entities if it can't fit more in the bag it's currently trying to fill. However I am too lazy to write this right now so enjoy this stopgap.

## Why / Balance
If I get one more of these smuggler satchel heisentests I'm going to crash out.

## Technical details
We limit the amount that can go in each bag by counting how much could possibly be put into the bag at its maximum.

~And we remove the +1 for returning a random float because it seemed to be already inclusive.~

## Media
I made a really shitty test to spawn 10,000 satchels on a grid. It sucks. No I'm not PRing it.
image

## Requirements

### DIFF
--- a/Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
+++ b/Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
@@ -14,6 +14,6 @@ public sealed partial class RangeNumberSelector : NumberSelector
 
     public override float Get(System.Random rand, IEntityManager entMan, IPrototypeManager proto)
     {
-        return rand.NextFloat(Range.X, Range.Y);
+        return rand.NextFloat(Range.X, Range.Y + 1);
     }
 }
--- a/Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
+++ b/Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
@@ -14,6 +14,6 @@ public sealed partial class RangeNumberSelector : NumberSelector
 
     public override float Get(System.Random rand, IEntityManager entMan, IPrototypeManager proto)
     {
-        return rand.NextFloat(Range.X, Range.Y + 1);
+        return rand.NextFloat(Range.X, Range.Y);
     }
 }
--- a/Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
+++ b/Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
@@ -33,24 +33,23 @@
   table: !type:AllSelector
     children:
     - !type:NestedSelector
-      tableId: RandomSatchelGenericTable
+      tableId: RandomSatchelGenericTable # Max two 1x2 items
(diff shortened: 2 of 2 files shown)


Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
```diff
@@ -14,6 +14,6 @@ public sealed partial class RangeNumberSelector : NumberSelector
 
     public override float Get(System.Random rand, IEntityManager entMan, IPrototypeManager proto)
     {
-        return rand.NextFloat(Range.X, Range.Y + 1);
+        return rand.NextFloat(Range.X, Range.Y);
```
slarticodefast, line 17: I think that +1 is there because this number gets cast to an integer

slarticodefast:
Caught one on master
grafik

Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
beck-thompson: yeah even if this was something that we want to do, do it in a new pr and look at every entity table to make sure it doesn't break :laughing:

ArtisticRoomba: When the +1 is there I will fail tests because something will be inserted one more time over the point I tell it 2. There will be 5 alcohol bottles or 3 cigar cases when I asked for 4 and 2. Pain and suffering.

ArtisticRoomba: highly advanced testing confirms that this indeed is the problem

ArtisticRoomba:
This PR now requires https://github.com/space-wizards/space-station-14/pull/36155.

slarticodefast:
Caught another one on master
grafik

ArtisticRoomba:
The pruned spawn amount is the maximum the tables can hold before they have the chance to cause a heisentest.

Counting manually is horrible... ideally there should be a test that adds up the biggest of each possible item it can select and checks if it can fit in the bag provided. Otherwise someone has to manually remind anyone adding to this that they have to make sure they don't overflow the bag or else they'll cause a heisentest.

ArtisticRoomba:
Kill this thing
image

IamVelcroboy:
See also: https://github.com/space-wizards/space-station-14/issues/34041

ArtisticRoomba:
superceeded

PR #36147, opened 2025-03-29
Fix admeme hand teleproter.

ScarKy0:
## About the PR

Made the admeme hand teleporter work across maps and grids.
Rimworld enjoyers rejoice, it is now 10x easier to set up.

## Why / Balance

I believe this was the intended behaviour to begin with. Up until now you had to manually VV both the teleporter and the portals to allow this to happen.

## Technical details

Added a datafield, added a check to allow handmade teleporters to make portals that work across maps.

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- fix: Admeme hand teleproter now correctly works across grids and maps.

### DIFF
--- a/Content.Server/Teleportation/HandTeleporterSystem.cs
+++ b/Content.Server/Teleportation/HandTeleporterSystem.cs
@@ -95,6 +95,10 @@ public sealed class HandTeleporterSystem : EntitySystem
             var timeout = EnsureComp<PortalTimeoutComponent>(user);
             timeout.EnteredPortal = null;
             component.FirstPortal = Spawn(component.FirstPortalPrototype, Transform(user).Coordinates);
+
+            if (TryComp<PortalComponent>(component.FirstPortal, out var portal) && component.AllowPortalsOnDifferentMaps)
+                portal.CanTeleportToOtherMaps = true;
+
             _adminLogger.Add(LogType.EntitySpawn, LogImpact.High, $"{ToPrettyString(user):player} opened {ToPrettyString(component.FirstPortal.Value)} at {Transform(component.FirstPortal.Value).Coordinates} using {ToPrettyString(uid)}");
             _audio.PlayPvs(component.NewPortalSound, uid);
         }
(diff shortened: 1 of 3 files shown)


ScarKy0:
If possible id like this to be in staging cause admin stuff and the hand teleporter is a pain to use without this

Content.Server/Teleportation/HandTeleporterSystem.cs
```diff
@@ -95,6 +95,10 @@ private void HandlePortalUpdating(EntityUid uid, HandTeleporterComponent compone
             var timeout = EnsureComp<PortalTimeoutComponent>(user);
             timeout.EnteredPortal = null;
             component.FirstPortal = Spawn(component.FirstPortalPrototype, Transform(user).Coordinates);
+
+            if (TryComp<PortalComponent>(component.FirstPortal, out var portal) && component.AllowPortalsOnDifferentMaps)
```
Tayrtahn, line 99: Swap the order of these conditions so it can bail out if the cheaper condition fails - checking a bool field is a lot cheaper than running a TryComp.

This is totally a micro-optimization on a non-hot code path, but well, I'm picky.

Content.Server/Teleportation/HandTeleporterSystem.cs
```diff
@@ -113,6 +117,10 @@ private void HandlePortalUpdating(EntityUid uid, HandTeleporterComponent compone
             var timeout = EnsureComp<PortalTimeoutComponent>(user);
             timeout.EnteredPortal = null;
             component.SecondPortal = Spawn(component.SecondPortalPrototype, Transform(user).Coordinates);
+
+            if (TryComp<PortalComponent>(component.SecondPortal, out var portal) && component.AllowPortalsOnDifferentMaps)
```
Tayrtahn, line 121: Same.

Tayrtahn (commented):
Just a petty little tweak from me, but otherwise this looks good.

Tayrtahn (approved):

PR #36149, opened 2025-03-29
remove styleclass from playertab 'button' lines

Errant-4:
## About the PR
Returns the old look of the admin playertab, which was changed unintentionally by #33412 

## Technical details
That PR added the button styleclass to ListContainerButton, which kinda makes sense, but on the admin playertab every line is mechanically a button, so they inherited the looks.  This messes up the column alignments with the header, and arguably looks weird for what is visually a  table. So I'm removing the styleclass from those instances of listcontainerbuttons

## Media

PR
Screenshot_2025-03-29_132503

Master
Screenshot_2025-03-29_132532

## Requirements

## Breaking changes

**Changelog**
:cl: Errant
- fix: The admin player tab is now cleansed of evil.

### DIFF
--- a/Content.Client/Administration/UI/Tabs/PlayerTab/PlayerTab.xaml.cs
+++ b/Content.Client/Administration/UI/Tabs/PlayerTab/PlayerTab.xaml.cs
@@ -143,6 +143,7 @@ public sealed partial class PlayerTab : Control
         var entry = new PlayerTabEntry(player, new StyleBoxFlat(button.Index % 2 == 0 ? _altColor : _defaultColor));
         button.AddChild(entry);
         button.ToolTip = $"{player.Username}, {player.CharacterName}, {player.IdentityName}, {player.StartingJob}";
+        button.StyleClasses.Clear();
     }
 
     /// <summary>


Errant-4:
> Should probably be an ADMIN cl

Oh whoops that's what I meant to do

Aeshus (approved):
Thanks for the PR! 

We should probably make sure that admins don't like the (accidental) change before we merge, though.

Errant-4:
As per the unofficial vote, I am keeping the revert as the default look, but have added a cvar that allows retaining the button style. The header alignment is now also fixed if you do use the cvar to keep the button look.

button

metalgearsloth:
I think I ended up fixing the underlying issue in listcontainer (some really cursed style code) so we can probably just add the exception to vending machine and leave the rest as is.

Errant-4:
I think it makes sense to have the button style on ListContainerBUTTON by default, and this seems to have been the only exception where that is not wanted, so I think it's fine to leave it as is and just (optionally) overwrite it in this case

PJB3005 (changes requested):
We're not introducing a CVar for this.

PJB3005 (changes requested):
Also wouldn't this need to be done on the object tab too?

PR #36153, opened 2025-03-29
Delete AccessReaderTest.TestTags

Tayrtahn:
## About the PR

Deletes the obsolete `AccessReaderTest.TestTags` integration test.

## Why / Balance

This test is from back before `ProtoId<T>`, and all it does is check that the `AccessLevelPrototype`s used by entities with `AccessReaderComponent`s are valid. These days we have automatic validation of `ProtoId` fields in components by the YAML linter, so this test is redundant.

Unfortunately, it's a quick test to run, so this doesn't really do much to save on test run time.

## Technical details

Deleted.

## Requirements


slarticodefast (approved):
Yep, looks redundant!

PR #36154, opened 2025-03-29
Combine SpawnAndDeleteAllEntitiesOnDifferentMaps & SpawnAndDirtyAllEntities

Tayrtahn:
## About the PR

Combines two time-consuming integration tests into one.

## Why / Balance

`SpawnAndDeleteAllEntitiesOnDifferentMaps` does what it says - for each entity prototype, it creates a map and spawns the entity on it.
`SpawnAndDirtyAllEntities` does the same thing - for each entity prototype, it creates a map and spawns the entity on it. Then it marks all the components as dirty, waits 15 ticks and makes sure that the client receives state update data.

`SpawnAndDirtyAllEntities` even describes itself as a "Variant of SpawnAndDeleteAllEntitiesOnDifferentMaps". There isn't really a good reason to run both, since `SpawnAndDirtyAllEntities` does everything that `SpawnAndDeleteAllEntitiesOnDifferentMaps` does and a bit more. Any errors caught by `SpawnAndDirtyAllEntities` will also be caught by `SpawnAndDeleteAllEntitiesOnDifferentMaps`.

`SpawnAndDeleteAllEntitiesOnDifferentMaps` appears to take about **40 seconds** to run on GitHub, so this should be a decent improvement to testing speed.

## Technical details

Deleted `SpawnAndDeleteAllEntitiesOnDifferentMaps` and modified the xmldoc comment on `SpawnAndDirtyAllEntities` stand on its own. Also made `SpawnAndDirtyAllEntities` filter out prototypes with `RoomFillComponent`s since those could cause problems and `SpawnAndDeleteAllEntitiesOnDifferentMaps` had been updated to filter them out.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/EntityTest.cs
+++ b/Content.IntegrationTests/Tests/EntityTest.cs
@@ -19,69 +19,6 @@ namespace Content.IntegrationTests.Tests
         private static readonly ProtoId<EntityCategoryPrototype> SpawnerCategory = "Spawner";
 
         [Test]
-        public async Task SpawnAndDeleteAllEntitiesOnDifferentMaps()
-        {
-            // This test dirties the pair as it simply deletes ALL entities when done. Overhead of restarting the round
-            // is minimal relative to the rest of the test.
-            var settings = new PoolSettings { Dirty = true };
-            await using var pair = await PoolManager.GetServerClient(settings);
-            var server = pair.Server;
-
-            var entityMan = server.ResolveDependency<IEntityManager>();
-            var mapManager = server.ResolveDependency<IMapManager>();
-            var prototypeMan = server.ResolveDependency<IPrototypeManager>();
-            var mapSystem = entityMan.System<SharedMapSystem>();
-
-            await server.WaitPost(() =>
-            {
-                var protoIds = prototypeMan
-                    .EnumeratePrototypes<EntityPrototype>()
-                    .Where(p => !p.Abstract)
-                    .Where(p => !pair.IsTestPrototype(p))
-                    .Where(p => !p.Components.ContainsKey("MapGrid")) // This will smash stuff otherwise.
-                    .Where(p => !p.Components.ContainsKey("RoomFill")) // This comp can delete all entities, and spawn others
-                    .Select(p => p.ID)
-                    .ToList();
-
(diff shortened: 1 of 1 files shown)


Content.IntegrationTests/Tests/EntityTest.cs
```diff
@@ -18,69 +18,6 @@ public sealed class EntityTest
...
-                    .Where(p => !p.Components.ContainsKey("RoomFill")) // This comp can delete all entities, and spawn others
-                    .Select(p => p.ID)
-                    .ToList();
-
-                foreach (var protoId in protoIds)
-                {
-                    mapSystem.CreateMap(out var mapId);
-                    var grid = mapManager.CreateGridEntity(mapId);
-                    // TODO: Fix this better in engine.
-                    mapSystem.SetTile(grid.Owner, grid.Comp, Vector2i.Zero, new Tile(1));
```
slarticodefast, line 51: The other test does not seem to spawn a tile first.
Do you think that is redundant? Some entites are anchored on spawn.

Content.IntegrationTests/Tests/EntityTest.cs
```diff
@@ -18,69 +18,6 @@ public sealed class EntityTest
...
-                    var coord = new EntityCoordinates(grid.Owner, 0, 0);
-                    entityMan.SpawnEntity(protoId, coord);
-                }
-            });
-
-            await server.WaitRunTicks(15);
```
slarticodefast, line 57: what is the difference between this and `RunTicksSync` which the other test uses here?

Tayrtahn: This just runs the server for some number of ticks and doesn't do anything with the client.

RunTicksSync runs the server for one tick, then the client, makes sure they're idle, then repeats for however many ticks are specified.

Content.IntegrationTests/Tests/EntityTest.cs
Tayrtahn: I would expect those entities to complain if they're not able to anchor when they spawn. The fact that they don't made me think that the engine might've been fixed to add a tile automatically or something.

However, I can see no sign of that being the case from a quick look. I'll put that back in to be safe.

PR #36155, opened 2025-03-29
fix RangeNumberSelector to actually be inclusive

ArtisticRoomba:
## About the PR
Fixes RangeNumberSelector to be actually inclusive.

Required by https://github.com/space-wizards/space-station-14/pull/36146

## Why / Balance
I walk up to the store and ask for 4 apples. The storekeep says "SURE" and hands me 5 apples. I wonder why, I question why. I look him dead in the eye: why did this guy give me four apples? What could be the reason why?

And so I open Jetbrains Rider in search of why. I debug debug debug, but nothing catches my eye. Until I open the RangeNumberSelector, yes, something seems awry. This NextFloat! It's inclusive plus one! Yes, that must be why.

## Technical details
Removes a `+ 1`. It was already inclusive.

## Media
rider64_lCP6XNuC8a
Content Client_wJ7iV39GFo
rider64_6uXDnpkAGM

## Breaking Changes
Generally, the random number system for determining what to select for the RangeNumberSelector logic has been changed to do random ints instead of floats. You should correct logic to deal with integers instead of floats.

Since the `RangeNumberSelector` range now is inclusive (instead of inclusive + 1), you will get one item less than you expected to get. You'll have to increase the range respectively.

- `Rolls.Get` in `ConstantNumberSelector.cs` signature changed from `Rolls.Get(rand, entMan, proto)` to `Rolls.Get(rand)`
- `Rolls.Get` in `EntityTableSelector.cs` signature changed from `Rolls.Get(rand, entMan, proto)` to `Rolls.Get(rand)`
- in `ConstantNumberSelector`, changed method signature: `public override float Get(System.Random rand, IEntityManager entMan, IPrototypeManager proto)` to public override int `Get(System.Random rand)`.
- in `NumberSelector`, changed `public abstract float Get(System.Random rand, IEntityManager entMan, IPrototypeManager proto)` to `public abstract int Get(System.Random rand)`
- `Vector2` in `RangeNumberSelector` converted to `Vector2i` (int).

## Requirements

**Changelog**
:cl:
- fix: Certain randomly filled bags will now have the correct number of contents inside of them.

### DIFF
--- a/Content.Shared/EntityTable/EntitySelectors/EntSelector.cs
+++ b/Content.Shared/EntityTable/EntitySelectors/EntSelector.cs
@@ -20,7 +20,7 @@ public sealed partial class EntSelector : EntityTableSelector
         IEntityManager entMan,
         IPrototypeManager proto)
     {
-        var num = (int) Math.Round(Amount.Get(rand, entMan, proto));
+        var num = Amount.Get(rand);
         for (var i = 0; i < num; i++)
         {
             yield return Id;
--- a/Content.Shared/EntityTable/EntitySelectors/EntityTableSelector.cs
+++ b/Content.Shared/EntityTable/EntitySelectors/EntityTableSelector.cs
@@ -30,7 +30,7 @@ public abstract partial class EntityTableSelector
         IEntityManager entMan,
         IPrototypeManager proto)
     {
-        var rolls = Rolls.Get(rand, entMan, proto);
+        var rolls = Rolls.Get(rand);
         for (var i = 0; i < rolls; i++)
         {
             if (!rand.Prob(Prob))
--- a/Content.Shared/EntityTable/ValueSelector/ConstantNumberSelector.cs
+++ b/Content.Shared/EntityTable/ValueSelector/ConstantNumberSelector.cs
@@ -1,5 +1,3 @@
-using Robust.Shared.Prototypes;
-
 namespace Content.Shared.EntityTable.ValueSelector;
 
 /// <summary>
@@ -8,14 +6,14 @@ namespace Content.Shared.EntityTable.ValueSelector;
 public sealed partial class ConstantNumberSelector : NumberSelector
 {
     [DataField]
-    public float Value = 1;
+    public int Value = 1;
 
-    public ConstantNumberSelector(float value)
+    public ConstantNumberSelector(int value)
     {
         Value = value;
     }
 
-    public override float Get(System.Random rand, IEntityManager entMan, IPrototypeManager proto)
+    public override int Get(System.Random rand)
     {
         return Value;
     }
--- a/Content.Shared/EntityTable/ValueSelector/NumberSelector.cs
+++ b/Content.Shared/EntityTable/ValueSelector/NumberSelector.cs
@@ -1,6 +1,5 @@
 using Content.Shared.EntityTable.EntitySelectors;
 using JetBrains.Annotations;
-using Robust.Shared.Prototypes;
 
 namespace Content.Shared.EntityTable.ValueSelector;
 
@@ -10,7 +9,5 @@ namespace Content.Shared.EntityTable.ValueSelector;
 [ImplicitDataDefinitionForInheritors, UsedImplicitly(ImplicitUseTargetFlags.WithInheritors)]
 public abstract partial class NumberSelector
 {
-    public abstract float Get(System.Random rand,
-        IEntityManager entMan,
-        IPrototypeManager proto);
+    public abstract int Get(System.Random rand);
 }
--- a/Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
+++ b/Content.Shared/EntityTable/ValueSelector/RangeNumberSelector.cs
@@ -1,7 +1,3 @@
-using System.Numerics;
-using Robust.Shared.Prototypes;
-using Robust.Shared.Random;
-
 namespace Content.Shared.EntityTable.ValueSelector;
 
 /// <summary>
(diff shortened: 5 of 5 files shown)


MilonPL (commented):
NextFloat is not inclusive, so by changing this behavior we are making it so that we can never roll `Range.Y`. I wonder if the better solution would be changing `EntSelector` to use `Math.Floor` rather than `Math.Round` (so that for range 1-3, if we get 3.7, we're rounding it down to 3 rather than up to 4). Could also do it here and cast it to a int or just use regular random for ints rather than floats.

ArtisticRoomba:
Sure I'll do that in a second

slarticodefast:
It is weird that they are use floats in the first place instead of an integer.

NextFloat gives you a random float from `minValue` to `maxValue`, excluding the `maxValue`.
So if `MaxValue` is 5 then it may pick 4.99999 but never 5.
grafik

If a `RangeNumberSelector` has a range from 2 to 4 then it gets a float between 2 and 5.
grafik

For `GetSpawns` we will then have a spawn for i=0,1,2,3,4 in the worst case which is one too much.
grafik

The other place where this is used does this differently for some weird reason but it seems to have the same effect.
grafik

ArtisticRoomba:
yes this code was driving me insane last night

Tayrtahn:
Why does the abstract parent `NumberSelector` return a float instead of an int anyway? It and its derived classes (`RangeNumberSelector` and `ConstantNumberSelector`) are in the `EntityTable` namespace, are only used with `EntityTableSelector`, and are used to get int values everywhere they appear (selecting Rolls or Amount).

I kinda think it should be refactored to use ints instead. If we want to use it with floats in the future, I don't see why we couldn't just add `FloatSelector` and derive from that.

ArtisticRoomba:
> If we want to use it with floats in the future, I don't see why we couldn't just add FloatSelector and derive from that.

I don't even see how we could select items using floats- oh no. Let me check something.

ArtisticRoomba:
Okay, selection with weights is handled in `SharedRandomExtentions.Pick()` using floats and the like so this should be fine. Should. Do realize my C# skills are equivalent to monkeys banging on typewriters.

ArtisticRoomba:
Yeah something's wrong with the maints closets, their variety is significantly worse with these changes.

slarticodefast:
> Yeah something's wrong with the maints closets, their variety is significantly worse with these changes.

They did have all more item than intended before in the table, so this just fixes it I guess.

ArtisticRoomba:
That's what I thought originally, but tt feels stark, maybe a buff is in order after this one?

| Before | After |
| --- | --- |
| image | image |

slarticodefast:
Can you compare them to the original loot tables from when EntityTables were introduced to see how they were converted?
Also needs a breaking changes entry now, since the bugfix may impact game balance for forks a lot.

ArtisticRoomba:
| Before Changes | After Changes |
| --- | --- |
| image | image |

| Before EntTables | After EntTables |
| --- | --- |
| image | image |

ArtisticRoomba:
I also tried out artificially increasing the range by one.

| Before Changes | After Changes | After Changes (Range +1)
| --- | --- | --- |
| image | image | Content Client_xzxgbFmn2W |

slarticodefast:
I meant comparing the yaml entries to see if the range value was translated correctly...

ArtisticRoomba:
Oh.

ArtisticRoomba:
> I meant comparing the yaml entries to see if the range value were translated correctly into the new format...

I would say not really. Old EntitySpawnEntry was very restrictive so you couldn't really do "pull 5 items from a list" easily. To be honest in the EntityTables PR maints lockers were entirely rebalanced and their variety increased significantly.

Considering this I'd say it's fine to do another bump in this PR to match what we were at previously but something is still unsettling me about the weighting system, I feel that it's broken.

Tayrtahn:
I'd like to get this merged so we can stop these heisentests. Are you comfortable with it as it is, @ArtisticRoomba?

EmoGarbage404:
why does this PR remove the extra parameters from the selectors? they were intentionally added to allow for more complex numerical selection in the future.

ArtisticRoomba:
I was instructed in the discord to prune code that wasn't being used, I had a suspicion that it was created like this to allow expansion in the future.

When I get home I'll tug a maint with me and ask what they want to happen. I think the easiest fix is to just fix the inclusivity like before and keep the float property, it would do less damage and be better overall.

This would reduce the diff to one line.

@Tayrtahn

ArtisticRoomba:
You can see the initial commit (https://github.com/space-wizards/space-station-14/commit/3b072918c1d56bd2b04dea4374c3d66426788fb6) for the fix proposed.

MilonPL (approved):

Tayrtahn:
> Additionally, why are the return values changed to be only integers? Allowing them to be floats and only rounding when necessary lets them be reused for other things in the future.

I didn't really see a place where an ent table selector was going to need to pick a random float value. The only float field is the probability, and it seems silly to pick a random number for the range of another random number. If the intention is to use these selectors for things outside of entity tables, then it might be worth considering moving them to a different namespace.

In any case, I see no harm in changing them back to floats. 🤷

ArtisticRoomba:
The float -> int conversion still makes me uneasy as it seems like the item distribution is worse with the randomness logic changed to a float. The original solution is less damaging.

MilonPL:
I really don't like the amount of type casting and extra floor/round/ceiling operations we're doing with floats. I think int just works better here.

ArtisticRoomba:
DNM until emo says what's best

EmoGarbage404:
it's best

PR #36156, opened 2025-03-29
Warnings spring cleaning

ljm862:
## About the PR

🌷It's spring time🌷 
You know what that means, **spring cleaning**! That's right, time to tidy up the repo with some warning removals. Removed some 200 or so warnings. Some warnings were too complex for a first time contributor (DoAfter, Clearing Overrides). Some places had scary comments (FoldableClothingSystem, ConstructionSystem.Initial). Yet others confused me (TimerExtensions).
I took care of the majority of the low hanging fruit though. There are still a fair few warnings to go which I'll address in other PRs, but this one should be ready to go for.

I'm treating this PR as an "Epic" to manage all the other, smaller PRs that I'm opening in this first batch. 

- #36157 
- #36158 
- #36159 
- #36160 
- #36161 
- #36162 
- #36163 
- #36164 
- #36166
- #36167
- #36168 
- #36169 
- #36188 
- #36189 
- #36190 
- #36193 
- #36194
- #36195
- #36196 
- #36197 
- #36199 
- #36200
- #36206
- #36207

### What's Next?
The Component.Owner, Control.Dispose, MapSystem, Logger and SoundSpecifier stuff make up about 80-90% of the warnings left. I'll do another pass to address the remaining warnings not in these areas and then once the Content side is done, we can remove dead code from the Toolbox too!

## Why / Balance

Part of #33279. It should improve the code quality and probably performance as outdated functionality is being replaced.

## Technical details

Only targeted SS14's codebase, did not touch any of the Robust warnings. Would prefer to do that separately.
Removed all CS0414 and CS0612 warnings from the codebase.
Took a big chunk out of CS0618

## Media

image
It runs and some basic functionality works!

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

Should be none. However, a lot of files were touched and plenty of unfamiliar systems were cleaned here. It's possible there were things that broke though I'm not aware of any.

PR #36158, opened 2025-03-29
Async doafter removal

ljm862:
## About the PR

The doafter system looks fairly complex and I'm not 100% if I've understood everything correctly. The change is a bit more complex than a simple "use another API" like most of the rest of the warning cleanup.
Read #36156 

## Technical Details
Changed one of the simpler AwaitedDoAfterEvents to use the synchronous system instead. 
If I've understood correctly it will now use the newly created synchronous event and fire that on completion. Instead of awaiting for the time, it succeeds immediately and the code that would normally run after awaiting will instead run on event completion.

## Requirements

### DIFF
--- a/Content.Server/Engineering/EntitySystems/DisassembleOnAltVerbSystem.cs
+++ b/Content.Server/Engineering/EntitySystems/DisassembleOnAltVerbSystem.cs
@@ -3,6 +3,7 @@ using Content.Shared.DoAfter;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Verbs;
 using JetBrains.Annotations;
+using Content.Shared.Engineering.EntitySystems;
 
 namespace Content.Server.Engineering.EntitySystems
 {
@@ -16,6 +17,7 @@ namespace Content.Server.Engineering.EntitySystems
             base.Initialize();
 
             SubscribeLocalEvent<DisassembleOnAltVerbComponent, GetVerbsEvent<AlternativeVerb>>(AddDisassembleVerb);
+            SubscribeLocalEvent<DisassembleOnAltVerbComponent, DisassembleOnAltVerbDoAfterEvent>(OnDisassembleOnAltVerb);
         }
         private void AddDisassembleVerb(EntityUid uid, DisassembleOnAltVerbComponent component, GetVerbsEvent<AlternativeVerb> args)
         {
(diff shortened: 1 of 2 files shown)


ArtisticRoomba:
we don't do any changelogs for changes that aren't player facing, and be sure to document what you've done in the technical details section that you've removed

Content.Shared/Engineering/EntitySystems/DisassembleOnAltVerbDoAfterEvent.cs
```diff
@@ -0,0 +1,10 @@
+using Content.Shared.DoAfter;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.Engineering.EntitySystems
+{
```
ArtisticRoomba, line 5: use file scoped namespaces

ljm862: Am I reading it wrong? It looks like it's correct to me. Have I misunderstood something? 
The filepath is `Content.Shared/Engineering/EntitySystems` and the namespace is also `Content.Shared.Engineering.EntitySystems` or do you mean something else?

ArtisticRoomba: Right now the namespace isn't file-scoped, where the entire file is a namespace (rather than in brackets).

An example of a file scoped namespace:
```C#
using Content.Shared.DoAfter;

namespace Content.Shared.Engineering.EntitySystems;

public sealed partial class DisassembleOnAltVerbDoAfterEvent : SimpleDoAfterEvent
{
}
```

Rider also tells me that the empty class body is redundant, so with that change it would be:
```C#
using Content.Shared.DoAfter;

namespace Content.Shared.Engineering.EntitySystems;

public sealed partial class DisassembleOnAltVerbDoAfterEvent : SimpleDoAfterEvent;
```
Solution builds just fine with these changes.

keronshb: They mean instead of doing
`namespace Content.Foo {}`
You do
`namespace Content.Foo;`

ljm862: Got it! Thanks for the rapid response!

Content.Server/Engineering/EntitySystems/DisassembleOnAltVerbSystem.cs
```diff
@@ -41,18 +43,24 @@ public async void AttemptDisassemble(EntityUid uid, EntityUid user, EntityUid ta
...
+                var doAfterArgs = new DoAfterArgs(EntityManager, user, component.DoAfterTime, new DisassembleOnAltVerbDoAfterEvent(), target)
                 {
                     BreakOnMove = true,
                 };
-                var result = await doAfterSystem.WaitDoAfter(doAfterArgs);
 
-                if (result != DoAfterStatus.Finished)
+                if (!EntityManager.System<SharedDoAfterSystem>().TryStartDoAfter(doAfterArgs))
+                {
                     return;
```
metalgearsloth, line 55: Use system dependency for this.

metalgearsloth (changes requested):

Content.Server/Engineering/EntitySystems/DisassembleOnAltVerbSystem.cs
ljm862: Actually, looks like this review is superfluous now due to #32420 being merged in and moving the code around. I'll close this PR out now

ljm862:
No longer needed due to #32420

PR #36159, opened 2025-03-29
Remove warnings from cargo system

ljm862:
## About the PR

Read #36156 

## Technical details
Use non-obsolete `Play` override for `AnimationPlayerComponent`

## Requirements

### DIFF
--- a/Content.Client/Cargo/Systems/CargoSystem.Telepad.cs
+++ b/Content.Client/Cargo/Systems/CargoSystem.Telepad.cs
@@ -67,8 +67,10 @@ public sealed partial class CargoSystem
         if (!Resolve(uid, ref sprite))
             return;
 
+        if (!TryComp<AnimationPlayerComponent>(uid, out var player))
+            return;
+
         _appearance.TryGetData<CargoTelepadState?>(uid, CargoTelepadVisuals.State, out var state);
-        AnimationPlayerComponent? player = null;
 
         switch (state)
         {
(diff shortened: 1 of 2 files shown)


Content.Client/Cargo/Systems/CargoSystem.Telepad.cs
```diff
@@ -76,7 +76,9 @@ private void OnChangeData(EntityUid uid, SpriteComponent? sprite = null)
                 if (_player.HasRunningAnimation(uid, TelepadBeamKey))
                     return;
                 _player.Stop(uid, player, TelepadIdleKey);
-                _player.Play(uid, player, CargoTelepadBeamAnimation, TelepadBeamKey);
+
+                player ??= EntityManager.EnsureComponent<AnimationPlayerComponent>(uid);
```
slarticodefast, line 80: this should just be a `TryComp` as a guard statement at the beginning of the `OnChangeData` function.
Also inside entity systems you can use the proxy shorthands in `EntitySystem.Proxy.cs`

slarticodefast (changes requested):

slarticodefast (commented):
Just had a look and the telepads need the `AnimationPlayerComponent` added in their yaml prototype or the `TryComp` will fail.
Before it was ensured by that function, but components should not be added like this unless necessary.

ljm862:
> Just had a look and the telepads need the `AnimationPlayerComponent` added in their yaml prototype or the `TryComp` will fail. Before it was ensured by that function, but components should not be added like this unless necessary.

Okay, what's the best solution here then do you think? Go back to the `player ??= EntityManager.EnsureComponent<AnimationPlayerComponent>(uid);` from before? Or maybe 
`_player.Play((uid, player!), CargoTelepadBeamAnimation, TelepadBeamKey);`?

I'm getting CS8620 errors otherwise when using the Entity<> override and I don't want to just leave it if I can avoid it!

slarticodefast:
Nah, just add the component in the yaml prototype in cargo_prototype.yml so the entity will have that component and the TryComp will be able to get it.

ljm862:
> Nah, just add the component in the yaml prototype in `cargo_prototype.yml` so the entity will have that component and the TryComp will be able to get it.

I _think_ I've done that correctly. Please let me know!

slarticodefast (approved):
Looks good to me!

PR #36161, opened 2025-03-29
Movement systems warning cleanup

ljm862:
## About the PR

Read #36156 

## Technical details
Removed a bunch of unused code. Using `SharedTransformSystem` where possible to avoid obsolete code and other changes to use preferred code paths.

## Requirements


Content.Server/Movement/RotateEyesCommand.cs
```diff
@@ -35,7 +35,8 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
                 continue;
 
             mover.TargetRelativeRotation = rotation;
-            entManager.Dirty(mover);
+
+            entManager.Dirty(mover.Owner, mover);
```
slarticodefast, line 39: Component.Owner is deprecated as well.
Use EntityQueryEnumarator the get the corresponding uid

Content.Shared/Movement/Components/MovementSpeedModifierComponent.cs
```diff
@@ -32,28 +32,6 @@ public sealed partial class MovementSpeedModifierComponent : Component
         [AutoNetworkedField, ViewVariables]
         public float SprintSpeedModifier = 1.0f;
 
-        [ViewVariables(VVAccess.ReadWrite)]
-        private float _baseWalkSpeedVV
```
slarticodefast, line 36: this is used so it can be set from the VV window!
keep it

slarticodefast (changes requested):

Content.Shared/Movement/Components/MovementSpeedModifierComponent.cs
ljm862: Ah that makes sense! I know in a couple of other places I removed some VVs that probably shouldn't have been then. Guess I need to look at the system a bit more to know why my ide wasn't telling me about the usages!

slarticodefast: The VV window is in-game and it uses source generation, so the IDE cannot know about it.
That is what the `ViewVariables` attribute is for.

Content.Shared/Movement/Components/MovementSpeedModifierComponent.cs
```diff
@@ -33,26 +33,7 @@ public sealed partial class MovementSpeedModifierComponent : Component
...
-        private float _baseSprintSpeedVV
-        {
-            get => BaseSprintSpeed;
-            set
-            {
-                BaseSprintSpeed = value;
-                Dirty();
-            }
-        }
+        private float _baseWalkSpeedVV;
```
slarticodefast, line 36: keep all of the code including the getter and setter!
same for the other one

slarticodefast (changes requested):

slarticodefast (approved):
LGTM!

Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
```diff
@@ -42,7 +38,7 @@ private void OnGetEyeOffsetEvent(EntityUid uid, EyeCursorOffsetComponent compone
 
     public Vector2? OffsetAfterMouse(EntityUid uid, EyeCursorOffsetComponent? component)
     {
-        var localPlayer = _player.LocalPlayer?.ControlledEntity;
+        var localPlayer = _player.LocalSession?.AttachedEntity;
         var mousePos = _inputManager.MouseScreenPosition;
```
metalgearsloth, line 42: LocalEntity exists right?

ljm862: Yes it does!

PR #36162, opened 2025-03-29
Examine warnings cleanup

ljm862:
## About the PR

Read #36156 

## Technical details
UI using the `SpriteSystem` which is passed from the `ExamineSystem` for only needing to resolve the dependency once instead of every time we instantiate a new `ExamineButton`.

## Requirements


DogZeroX:
Changelogs are not player facing, gotta remove em!

Content.Shared/Examine/ExamineSystemShared.cs
```diff
@@ -72,7 +72,7 @@ public bool IsInDetailsRange(EntityUid examiner, EntityUid entity)
             // Is the target hidden in a opaque locker or something? Currently this check allows players to examine
             // their organs, if they can somehow target them. Really this should be with userSeeInsideSelf: false, and a
             // separate check for if the item is in their inventory or hands.
-            if (_containerSystem.IsInSameOrTransparentContainer(examiner, entity, userSeeInsideSelf: true))
+            if (_containerSystem.IsInSameOrTransparentContainer((examiner, null, null), (entity, null, null), userSeeInsideSelf: true))
```
slarticodefast, line 75: These should automatically cast into the non-obsolete version once we remove the obsolete overload.
So for readability reasons I would prefer if don't do the `(uid, null, null)` thing.

ljm862: Ah that's cool, I didn't know it would cast like that, nice! I'll remove this then, there's a few other places it happens so I'll take them out too and remove the obsolete RT code in my next sweep. Thanks!

Content.Client/Examine/ExamineButton.cs
```diff
@@ -27,14 +23,16 @@ public sealed class ExamineButton : ContainerButton
     public TextureRect Icon;
 
     public ExamineVerb Verb;
+    private SpriteSystem _spriteSystem;
```
slarticodefast, line 26:
```suggestion
    private SpriteSystem _sprite;
```
naming convention
same below

slarticodefast (changes requested):

slarticodefast (approved):
LGTM!

PR #36169, opened 2025-03-29
Various UI warnings cleanup

ljm862:
## About the PR

Read #36156 
This was a bunch of files that are mostly unrelated other than the fact they're UI in some way. It was either this or 1 file change per PR and given how small they are I think it's better for everyone if these ones are grouped.

## Technical details
- Removing unused code 
- Using non-obsolete animation player overrides
- Getting the map ID the non-obsolete way
- Using `ILocalizationManager` that's already injected instead of the static approach
- Using `SpriteSystem` methods where appropriate
- Replace `NoiseGenerator` with `FastNoiseLite`

## Requirements


Content.Client/Humanoid/SingleMarkingPicker.xaml.cs
```diff
@@ -188,7 +190,7 @@ public void PopulateList(string filter)
 
         foreach (var (id, marking) in sortedMarkings)
         {
-            var item = MarkingList.AddItem(Loc.GetString($"marking-{id}"), marking.Sprites[0].Frame0());
+            var item = MarkingList.AddItem(Loc.GetString($"marking-{id}"), _entityManager.System<SpriteSystem>().Frame0(marking.Sprites[0]));
```
slarticodefast, line 193: add a private variable to the class and get this in the constructor

Content.Client/SprayPainter/UI/SprayPainterWindow.xaml.cs
```diff
@@ -32,13 +34,13 @@ public SprayPainterWindow()
         _spriteSystem = _sysMan.GetEntitySystem<SpriteSystem>();
     }
 
-    private static string GetColorLocString(string? colorKey)
+    private static string GetColorLocString(string? colorKey, ILocalizationManager loc)
```
slarticodefast, line 37: does this need to be passed in as a parameter?
The method should have access to `_loc`

Content.Client/SprayPainter/UI/SprayPainterWindow.xaml.cs
```diff
@@ -67,10 +69,11 @@ public void Populate(List<SprayPainterEntry> entries, int selectedStyle, string?
             currentPalette = palette;
             ItemColorIndex.Clear();
             ColorList.Clear();
+            var x = _loc.DefaultCulture;
```
slarticodefast, line 72: unused?

slarticodefast (dismissed):

Content.Client/Humanoid/MarkingPicker.xaml.cs
```diff
@@ -18,6 +19,9 @@ public sealed partial class MarkingPicker : Control
 {
     [Dependency] private readonly MarkingManager _markingManager = default!;
     [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+    [Dependency] private readonly IEntityManager _entityManager = default!;
+    // TODO: Make an ISpriteSystem for injecting
```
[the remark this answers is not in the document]
ljm862: > I think it's not because of the missing interface, but because it's a system rather than a manager.

Ah maybe then! The tests that failed mentioned something about registering and [this page](https://docs.spacestation14.com/en/robust-toolbox/ioc.html) mentioned about needing the interface so I went with that

Content.Client/Parallax/ParallaxGenerator.cs
```diff
@@ -217,14 +217,11 @@ public LayerNoise(TomlTable table)
 
             public override void Apply(Image<Rgba32> bitmap)
             {
-                var noise = new NoiseGenerator(NoiseType);
-                noise.SetSeed(Seed);
+                var noise = new FastNoiseLite((int)Seed);
+                noise.SetFractalType(NoiseType);
                 noise.SetFrequency(Frequency);
-                noise.SetPersistence(Persistence);
```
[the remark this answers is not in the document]
ljm862: > some of these seem to be missing also please move changes to this file into a separate PR, as they will need some testing to make sure everything still works

The `FastNoiseLite` doesn't look to have Persistence or Period parameters that I could see. I also looked at the `GetNoise` method on the `NoiseGenerator` and it didn't look to make use of it when compared to how the `FastNoiseLite` one works. I could be wrong on that! Yes, agree with separating this out!

PR #36187, opened 2025-03-30
Different clothing items now actually give zombification resistance

Alpha-Two:
## About the PR
Instead of the base protection granted by every single clothing item, most clothing items will now have a coefficient that affects zombification chances. Thicker layers of clothing will have a better coefficient, with bio suits/hoods and CBURN exosuits having the best values.

List of current values:

- Jumpsuits/Uniforms/etc: 0.9
- All shoes: 0.95
- Base outer clothing: 0.9
  - "Large" outer clothing (Fire suits, etc): 0.8
  - Winter coats: 0.6
- Hats and helmets:
  - Winter hoods: 0.8
  - Hardhats: 0.85
  - Bio hoods: 0.6
  - Base hardsuit + EVA helmets: 0.85
    - CBURN helmets: 0.3
- Base hardsuits: 0.65
  - EVA suits: 0.7
  - CBURN Exosuits, ERT Janitor suits, ERT engineer suits, bio suits: 0.15
  - Agent hardsuits, ERT medic suits: 0.4

If I've missed anything, please let me know!

## Why / Balance
1. It's stupid that a tie gives as much resistance as a hardsuit.
2. Give some more love to the bio suit!

## Technical details
New component: `ZombificationResistanceComponent` that has a `coefficient` value that will be multiplied by the base chance (0.8). The minimum chance has also been greatly reduced from 0.25 to 0.05.

## Media
No media cuz not visual change

## Requirements

## Breaking changes
MaxZombieInfectionChance has been renamed BaseZombieInfectionChance, primarily to better reflect its purpose, additionally because it's now possible to have an item of clothing that _increases_ infection chance.

**Changelog**
:cl: UpAndLeaves
- tweak: Ties no longer protect against zombie infections.
- tweak: Thicker materials, especially winter coats and bio suits, will now protect against zombie infections more effectively.

### DIFF
--- a/Content.Server/Zombies/ZombieSystem.cs
+++ b/Content.Server/Zombies/ZombieSystem.cs
@@ -201,30 +201,22 @@ namespace Content.Server.Zombies
 
         private float GetZombieInfectionChance(EntityUid uid, ZombieComponent component)
         {
-            var max = component.MaxZombieInfectionChance;
+            var chance = component.BaseZombieInfectionChance;
 
             if (!_inventory.TryGetContainerSlotEnumerator(uid, out var enumerator, ProtectiveSlots))
-                return max;
-
-            var items = 0f;
-            var total = 0f;
+                return chance;
             while (enumerator.MoveNext(out var con))
             {
-                total++;
-                if (con.ContainedEntity != null)
-                    items++;
+                if (con.ContainedEntity != null && TryComp<ZombificationResistanceComponent>(con.ContainedEntity, out var zombificationResistanceComponent))
+                {
+                    chance *= zombificationResistanceComponent.ResistanceCoefficient;
+                }
             }
-
-            if (total == 0)
-                return max;
-
-            // Everyone knows that when it comes to zombies, socks & sandals provide just as much protection as an
-            // armored vest. Maybe these should be weighted per-item. I.e. some kind of coverage/protection component.
-            // Or at the very least different weights per slot.
 
             var min = component.MinZombieInfectionChance;
             //gets a value between the max and min based on how many items the entity is wearing
-            var chance = (max - min) * ((total - items) / total) + min;
+            chance = Math.Max(chance, min);
+
             return chance;
         }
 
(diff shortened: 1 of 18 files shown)


beck-thompson:
For some reason I kind of liked the old way of doing it :laughing:
It was funny and did make some sort of sense

Alpha-Two:
> For some reason I kind of liked the old way of doing it 😆 It was funny and did make some sort of sense

Sadly, the pride pin zombie resist must make way for the reign of the winter coats :(

JackTalentedCoder:
I'm not a contributor yet, but the code is done very well. Good job

beck-thompson:
hey I'm not sure you saw the discussion on discord a few days ago just posting it here:
image

So yes keep the component but also think of a formula that works generically so we don't run into the same issue again!

Alpha-Two:
> So yes keep the component but also think of a formula that works generically so we don't run into the same issue again!

I see that a formula would be better, but this strictly buffs slash resistance, is that considered acceptable for balance? If you maintainers say that it would be worth it then I can code this, but it doesn't seem worth the effort if it's going to be denied nevertheless.

I'd also say that there are just three items that I vehemently want to keep the buffed zombie resist on, that being the Winter Coats, the Biosuits, and the Exosuits.

beck-thompson:
> I see that a formula would be better, but this strictly buffs slash resistance, is that considered acceptable for balance? If you maintainers say that it would be worth it then I can code this, but it doesn't seem worth the effort if it's going to be denied nevertheless.

> I'd also say that there are just three items that I vehemently want to keep the buffed zombie resist on, that being the Winter Coats, the Biosuits, and the Exosuits.

I'm unsure why you closed but yeah that seems about right! Balance stuff can always be tweaked, the issue is that we can't only rely on the zombie resistance component as it will just cause more bugs. Having both would be perfect!

Alpha-Two:
> I'm unsure why you closed but yeah that seems about right! Balance stuff can always be tweaked, the issue is that we can't only rely on the zombie resistance component as it will just cause more bugs. Having both would be perfect!

I closed it as I'm just going to make a new PR from scratch.
I'll probably make one PR for the new slash/component-based system, and one PR for the YAML balance changes, unless you advise otherwise.

PR #36191, opened 2025-03-30
Remove anatomy pack from thief backpack

kosticia:
## Why / Balance
Biggest difficulty of playing as thief - hiding objectives and contraband.
His toolbox should be analog of uplink, but less various. Player choses two sets that will form his playstyle. But one set absolutely don't fit with others - anatomy set. It contains one implant that makes game for thief much easier - storage implant. With that implant player shouldn't create stashes or hide his contraband. This set actually forces player chose one set instead of two, because storage implant is basically too much OP. Other implant - DNA Scrambler, useful for players only after escaping from permabrig or something like this, that makes this implant veeeery optional. And what does we have? Set with two implanters, one of them absolutely OP and used by 99% of time, and second one useless at 99% of situations. Thief deserves more love. It shouldn't be just renamed traitor with pacifism and same problems. 

## Requirements

**Changelog**
:cl:
- remove: Anatomy set was removed from thief toolbox.


ScarKy0:
Instead of just removing the anatomy kit we should just make the storage implant less of a must-pick. There is a PR already open for that.

Additionally, the thief toolbox should be replaced with some folded-satchel that becomes the smuggler's satchel with all the chosen kit items after use.

That being said, I am against this PR as it does not tackle the root cause of the issue and is just a thief nerf with nothing to give back to their kit.

kosticia:
> Instead of just removing the anatomy kit we should just make the storage implant less of a must-pick. There is a PR already open for that.

PR opened by Beck don't fix the problem of thief. Biggest part of staff that can became objectives have 1x2 size.

ScarKy0:
> PR opened by Beck don't fix the problem of thief. Biggest part of staff that can became objectives have 1x2 size.

I am still in belief that storage implant is the issue. 
I propose giving the thief an alternative way to hide their items (like turning the toolbox into a satchel, as i said) and then just removing the storage implant from the kit itself. Not the ENTIRE kit at once.

kosticia:
> I am still in belief that storage implant is the issue. I propose giving the thief an alternative way to hide their items (like turning the toolbox into a satchel, as i said) and then just removing the storage implant from the kit itself. Not the ENTIRE kit at once.

I'm gonna add some others kits as replacement. Also, pretty don't like the idea with toolbox and satchel. What if thief doesn't need it?

SlimmSlamm:
> > I am still in belief that storage implant is the issue. I propose giving the thief an alternative way to hide their items (like turning the toolbox into a satchel, as i said) and then just removing the storage implant from the kit itself. Not the ENTIRE kit at once.
> 
> I'm gonna add some others kits as replacement. Also, pretty don't like the idea with toolbox and satchel. What if thief doesn't need it?

Why are we removing instead of adding first?

kosticia:
> Why are we removing instead of adding first?

Because there actually is no difference?

ScarKy0:
> I'm gonna add some others kits as replacement. Also, pretty don't like the idea with toolbox and satchel. What if thief doesn't need it?

Put it away? Like you would with items from kits that you dont need?

kosticia:
> Put it away? Like you would with items from kits that you dont need? I don't think this is an issue, and it gives a useful item for thieves.

Huge item that which will tell everyone, "HEY, THERE IS POTENTIAL THIEF ON THE STATION"?

ScarKy0:
> Huge item that which will tell everyone, "HEY, THERE IS POTENTIAL THIEF ON THE STATION"?

You mean like with the thieving beacon? Smuggler's stashes are already in clothmats and traitor uplinks. Allowing a thief to use one changes nothing but give them a nice QoL item to have at the start.

SlimmSlamm:
> > Why are we removing instead of adding first?
> 
> Because there actually is no difference?

We lost a kit and get one 2 years later.

beck-thompson:
either thieves should get a free storage implant or it shouldn't be in any of the choices because everyone will always pick the storage implant

walksanatora:
actually... I dont pick the storage implant. with the ammount of shit the thief needs to steal the storage implant is just not worth it. I usually pick chameleon + communicator.

kosticia:
I asked maint for thief buying statistics. If this is really not the most picked set, I'll close pr and start coding fucking lavaland

ScarKy0:
Theres no stats for thief kit picks, only uplinks

MilonPL:
Thank you for your contribution! After the recent changes to the storage implant, we've decided to not merge this PR for now.

PR #36194, opened 2025-03-30
Ghost spring cleaning

ljm862:
## About the PR

Read #36156 

## Technical details
Use `SharedTransformSystem` to set map spawning to instead of using `GetMapUid`

## Requirements

### DIFF
--- a/Content.Server/Ghost/GhostSystem.cs
+++ b/Content.Server/Ghost/GhostSystem.cs
@@ -488,7 +488,7 @@ namespace Content.Server.Ghost
                 SetTimeOfDeath(ghost, mind.Comp.TimeOfDeath!.Value, ghostComponent);
             }
 
-            SetCanReturnToBody(ghost, ghostComponent, canReturn);
+            SetCanReturnToBody(ghostComponent, canReturn);
 
             if (canReturn)
                 _minds.Visit(mind.Owner, ghost, mind.Comp);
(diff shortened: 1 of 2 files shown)


Content.Server/Ghost/GhostSystem.cs
```diff
@@ -438,7 +438,7 @@ private bool IsValidSpawnPosition(EntityCoordinates? spawnPosition)
             if (spawnPosition?.IsValid(EntityManager) != true)
                 return false;
 
-            var mapUid = spawnPosition?.GetMapUid(EntityManager);
+            var mapUid = _transformSystem.GetMap(spawnPosition.Value);
```
slarticodefast, line 441: This does remove the additional null check from the null-conditional operator, but if I see this correctly this is already checked as part of the guard statement above.
For my own sanity I'll let another maintainer double check this before merging.

slarticodefast (approved):
Should be merged after https://github.com/space-wizards/RobustToolbox/pull/5787

Content.Server/Ghost/GhostSystem.cs
Tayrtahn: It looks okay to me, but I'll admit that I haven't actually tried it out yet. We do have decent test coverage with ghosts though, so the fact that the tests passed is a good sign.

Aeshus: I think that the guard condition should cover the null-case (`spawnPosition?[...] != true` => `null != true` => `true` => `return false`)

PR #36195, opened 2025-03-30
Light warnings cleanup

ljm862:
## About the PR

Read #36156 

## Technical details
Cleaning animation player calls to use Entity Component overrides

## Requirements

### DIFF
--- a/Content.Client/Light/Visualizers/PoweredLightVisualizerSystem.cs
+++ b/Content.Client/Light/Visualizers/PoweredLightVisualizerSystem.cs
@@ -57,8 +57,8 @@ public sealed class PoweredLightVisualizerSystem : VisualizerSystem<PoweredLight
         if(!comp.IsBlinking)
             return;
 
-        AnimationSystem.Play(new Entity<AnimationPlayerComponent>(uid, Comp<AnimationPlayerComponent>(uid)),
-            BlinkingAnimation(comp), PoweredLightVisualsComponent.BlinkingAnimationKey);
+        var animationPlayer = EnsureComp<AnimationPlayerComponent>(uid);
+        AnimationSystem.Play((uid, animationPlayer), BlinkingAnimation(comp), PoweredLightVisualsComponent.BlinkingAnimationKey);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


Content.Client/Light/Visualizers/PoweredLightVisualizerSystem.cs
```diff
@@ -55,11 +54,11 @@ private void OnAnimationCompleted(EntityUid uid, PoweredLightVisualsComponent co
     {
         if (args.Key != PoweredLightVisualsComponent.BlinkingAnimationKey)
             return;
-
         if(!comp.IsBlinking)
             return;
 
-        AnimationSystem.Play(uid, Comp<AnimationPlayerComponent>(uid), BlinkingAnimation(comp), PoweredLightVisualsComponent.BlinkingAnimationKey);
+        var animationPlayer = EnsureComp<AnimationPlayerComponent>(uid);
```
slarticodefast, line 60: should be a TryComp guard statement

slarticodefast (changes requested):

Content.Client/Light/Visualizers/PoweredLightVisualizerSystem.cs
ljm862: Would we not rather the EnsureComp which adds the component if it doesn't exist? I don't know the code that calls this enough to say but as it was already there in the method below (Line 75~) I guess it's there for a reason? Or is that just legacy :D

slarticodefast: Since this is the animation completion event subscription it must already have that component.

ljm862: Makes sense to me!

Content.Client/Light/Visualizers/PoweredLightVisualizerSystem.cs
```diff
@@ -68,15 +68,17 @@ private void OnAnimationCompleted(EntityUid uid, PoweredLightVisualsComponent co
...
     {
+        if (!TryComp<AnimationPlayerComponent>(uid, out var animationPlayer))
+            return;
+
         if (shouldBeBlinking == comp.IsBlinking)
             return;
 
         comp.IsBlinking = shouldBeBlinking;
 
-        var animationPlayer = EnsureComp<AnimationPlayerComponent>(uid);
```
slarticodefast, line 76: In this case we need to keep it, so that the animation can start playing.
The alternative would be to add it to all prototypes with `PoweredLightVisualsComponent` in yaml, but that would be a bigger breaking change for forks, so I would avoid it here.

ljm862: Got it!

slarticodefast (approved):

PR #36200, opened 2025-03-30
Various sharedsystem spring cleaning

ljm862:
## About the PR

Read #36156 
Changing several SharedSystems all at once.

## Technical details
Mostly using the `SharedTransformSystem` methods for map and grid information. Removes unused code and uses non-obsolete methods where possible. Also a question on the GravityComponent which I'd like to discuss (and remove the comment before merging)

## Requirements


Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -621,13 +621,12 @@ private bool ValidateWorldTargetBase(EntityUid user,
...
 
-            if (xform.MapID != coords.GetMapId(EntityManager))
+            if (xform.MapID != _transformSystem.GetMapId(coords))
                 return false;
 
             if (range <= 0)
                 return true;
-
-            return coords.InRange(EntityManager, _transformSystem, Transform(user).Coordinates, range);
+            return _transformSystem.InRange(coords, Transform(user).Coordinates, range);
```
slarticodefast, line 629:
```suggestion
            return _transformSystem.InRange(coords, xform.Coordinates, range);
```

Content.Shared/Gravity/GravityComponent.cs
```diff
@@ -19,6 +19,7 @@ public bool EnabledVV
             {
                 if (Enabled == value) return;
                 Enabled = value;
+                // Does this get fired twice on a change? Once here, once in <see cref="SharedGravitySystem.OnHandleState"/>?
```
slarticodefast, line 22:
```suggestion
```
I'm not certain, but I assume this is because of networking. For example you change the variable server side, then this will get raised on the server, and OnHandleState has to take care of the client.

Content.Shared/Gravity/SharedGravitySystem.cs
```diff
@@ -96,6 +96,7 @@ private void OnHandleState(EntityUid uid, GravityComponent component, ref Compon
             component.EnabledVV = state.Enabled;
             var ev = new GravityChangedEvent(uid, component.EnabledVV);
             RaiseLocalEvent(uid, ref ev, true);
+            Dirty(uid, component);
```
slarticodefast, line 99: Wait, what is this for?
this is already part of networking

slarticodefast (changes requested):

Content.Shared/Gravity/SharedGravitySystem.cs
ljm862: If I remember correctly, I made a change that removed the dirtying from the component as it used the obsolete method and Owner field and extracted that dirtying logic to the system here instead. But it looks like I'm not doing that anymore as the component is unchanged here 🤔

Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -469,7 +469,7 @@ private void Unbuckle(Entity<BuckleComponent> buckle, Entity<StrapComponent> str
             // TODO: This is doing 4 moveevents this is why I left the warning in, if you're going to remove it make it only do 1 moveevent.
```
slarticodefast, line 469: hmm, we might have to address this comment
code wise the non-obsolete version is absolutely identical though, so not sure if this is already fixed or the events are currently happening 4 times
let me ping @metalgearsloth

metalgearsloth: Yeah I would leave it it's still doing:
1. placenextto
2. setworldrot
3. setcoords

ljm862: Sounds good, I'll leave it as it was then and have a look at changing this in a separate review

Content.Shared/Ghost/SharedGhostSystem.cs
```diff
@@ -50,7 +50,7 @@ public void SetCanReturnToBody(EntityUid uid, bool value, GhostComponent? compon
             if (!Resolve(uid, ref component))
                 return;
 
-            component.CanReturnToBody = value;
+            SetCanReturnToBody(component, value);
```
slarticodefast, line 53: this means it no longer gets dirtied as before
also this isn't even a warning?
I would just keep it as it was.

slarticodefast (changes requested):

Content.Shared/Ghost/SharedGhostSystem.cs
ljm862: Yeah I think you're right. I think this is probably a leftover artifact that I had when I split my giant PR. It probably belonged to another change that should have been separate anyway!

slarticodefast: true!

slarticodefast: Hmm, I guess it's a matter of coding style, but now that I think about it again the previous way is more readable.
Also since it is not a warning let's keep it as it was to minimize conflicts.

slarticodefast, line 53:
```suggestion
            component.CanReturnToBody = value;
```

ljm862: Sounds good!

PR #36203, opened 2025-03-30
Absolute Charge Versus Threshold for Power Sensor

Vortebo:
## About the PR
The power sensor can now detect whether the amount of charge in its circuit is above or below a user-definable threshold value. It has two additional ports, Above Threshold and Below Threshold, to accommodate this.

## Why / Balance
Adding outputs for being above or below a specific threshold value gives users more toys to play with when developing logic circuits, hopefully leading to more interesting/fun device networking possibilities.

## Technical details
PowerSensorComponent has been given additional variables to work appropriately with the logic added to PowerSensorSystem.

PowerThresholdComponent and System have been created to hold the verbs/general UI stuff, mirroring PowerSwitch(able).

To implement the UI, I copied what's in place for the solution transfer window, removing any bits of logic that weren't relevant.

## Media

image

image

image

## Requirements

## Breaking changes
N/A

## Changelog
:cl:
- tweak: Power sensor can now report when charge is above or below a threshold value


Vortebo:
Okay I have no idea what that test failure has to do with anything in the PR

Content.Client/DeviceLinking/UI/ThresholdAmountBoundUserInterface.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        private IEntityManager _entManager;
+        private EntityUid _owner;
+        [ViewVariables]
+        private ThresholdAmountWindow? _window;
+
+        public ThresholdAmountBoundUserInterface(EntityUid owner, Enum uiKey) : base(owner, uiKey)
+        {
+          _owner = owner;
+          _entManager = IoCManager.Resolve<IEntityManager>();
```
ljm862, line 21: `BoundUserInterface` has an `IEntityManager` already! Called `EntMan`. No need to resolve it again here

Content.Client/DeviceLinking/UI/ThresholdAmountBoundUserInterface.cs
```diff
@@ -0,0 +1,42 @@
...
+    public sealed class ThresholdAmountBoundUserInterface : BoundUserInterface
+    {
+        private IEntityManager _entManager;
+        private EntityUid _owner;
+        [ViewVariables]
+        private ThresholdAmountWindow? _window;
+
+        public ThresholdAmountBoundUserInterface(EntityUid owner, Enum uiKey) : base(owner, uiKey)
+        {
+          _owner = owner;
```
ljm862, line 20: The `Owner` is already cached in the parent class too, no need to recache again

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml.cs
```diff
@@ -0,0 +1,36 @@
...
+          MinimumAmount.Text = Loc.GetString("comp-power-threshold-set-amount-min", ("amount", _min));
+          MaximumAmount.Text = Loc.GetString("comp-power-threshold-set-amount-max", ("amount", _max));
+        }
+
+        private void OnValueChanged(LineEdit.LineEditEventArgs args)
+        {
+            if (!int.TryParse(AmountLineEdit.Text, out var amount)  || amount > _max || amount < _min)
+                ApplyButton.Disabled = true;
+            else 
+                ApplyButton.Disabled = false;
```
ljm862, line 33: Can change this to 
`ApplyButton.Disabled = (!int.TryParse(AmountLineEdit.Text, out var amount)  || amount > _max || amount < _min))`

Content.Server/DeviceLinking/Systems/PowerSensorSystem.cs
```diff
@@ -134,5 +143,23 @@ private void UpdateOutputs(EntityUid uid, PowerSensorComponent comp)
             comp.DischargingState = dischargingState;
             _deviceLink.SendSignal(uid, comp.DischargingPort, dischargingState);
         }
+
+        if (comp.AboveState != aboveState)
+        {
+            comp.AboveState = aboveState;
+            if (aboveState)
```
ljm862, line 150: We know if we're here that we definitely won't be also belowState. I wonder if you could early exit, probably overkill as you'd need to allocate another bool. Worth considering!

Content.Shared/DeviceLinking/Systems/PowerThresholdSystem.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    private void AddSetThresholdVerbs(Entity<PowerThresholdComponent> ent, ref GetVerbsEvent<AlternativeVerb> args)
+    {
+        var (uid, comp) = ent;
+
+        if (!args.CanAccess || !args.CanInteract || !comp.CanChangeThresholdAmount || args.Hands == null)
+            return;
+
+        // Custom threshold verb
+        var @event = args;
```
ljm862, line 57: Personally not a fan of the @ declaration if it can be avoided. Others might disagree, will need a maintainer's perspective!

Content.Shared/DeviceLinking/Systems/PowerThresholdSystem.cs
```diff
@@ -0,0 +1,95 @@
...
+public sealed class PowerThresholdSystem : EntitySystem
+{
+    [Dependency] private readonly ISharedAdminLogManager _adminLogger = default!;
+    [Dependency] private readonly SharedPopupSystem _popup = default!;
+    [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
+
+    /// <summary>
+    ///     Default threshold amounts for the set-threshold verb.
+    /// </summary>
+    public static readonly FixedPoint2[] DefaultThresholdAmounts = new FixedPoint2[] { 0, 5000, 10000, 25000, 50000, 100000, 250000, 500000, 1000000 };
```
ljm862, line 26: Where do these thresholds come from? Trial & error?

Content.Client/DeviceLinking/UI/ThresholdAmountBoundUserInterface.cs
```diff
@@ -0,0 +1,42 @@
+using Content.Shared.DeviceLinking;
+using Content.Shared.DeviceLinking.Components;
+using Content.Shared.FixedPoint;
+using JetBrains.Annotations;
+using Robust.Client.GameObjects;
+using Robust.Client.UserInterface;
+
+namespace Content.Client.DeviceLinking.UI
+{
```
Aeshus, line 9: **style:** It is now convention to use [File-Scope Namespaces](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces).

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml
```diff
@@ -0,0 +1,15 @@
+﻿<DefaultWindow xmlns="https://spacestation14.io"
+            Title="{Loc 'ui-powersensor-amount-title'}"
+            Resizable="False">
+
```
Aeshus, line 4: **style:** This line is unnecessary.

```suggestion
```

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml
```diff
@@ -0,0 +1,15 @@
...
+    <BoxContainer Orientation="Vertical" SeparationOverride="4" MinSize="240 80">
```
Aeshus, line 5: **issue:** The `MinSize` here should be on the `DefaultWindow`, not on this `BoxContainer`.

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml
```diff
@@ -0,0 +1,15 @@
...
+        <BoxContainer Orientation="Horizontal">
+            <LineEdit Name="AmountLineEdit" Access="Public" HorizontalExpand="True" PlaceHolder="{Loc 'ui-powersensor-amount-line-edit-placeholder'}"/>
+        </BoxContainer>
```
Aeshus, line 8: **issue:** You don't need the `BoxContainer` here?

```suggestion
        <LineEdit Name="AmountLineEdit" Access="Public" HorizontalExpand="True" PlaceHolder="{Loc 'ui-powersensor-amount-line-edit-placeholder'}"/>
```

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Robust.Client.AutoGenerated;
+using Robust.Client.UserInterface.Controls;
+using Robust.Client.UserInterface.CustomControls;
+using Robust.Client.UserInterface.XAML;
+
+namespace Content.Client.DeviceLinking.UI
+{
```
Aeshus, line 7: **style:** File-scoped namespace needed here.

Content.Server/DeviceLinking/Components/PowerSensorComponent.cs
```diff
@@ -68,10 +81,22 @@ public sealed partial class PowerSensorComponent : Component
...
     public bool ChargingState;
 
     [DataField]
     public bool DischargingState;
+
+    [DataField]
+    public bool AboveState;
+
+    [DataField]
+    public bool BelowState;
```
Aeshus, line 101: **issue:** Please document these fields.

Content.Server/DeviceLinking/Systems/PowerSensorSystem.cs
```diff
@@ -9,6 +9,7 @@
 using Content.Shared.Power.Generator;
 using Content.Shared.Timing;
 using Content.Shared.Tools.Systems;
+using Content.Shared.DeviceLinking.Components;
```
Aeshus, line 12: **style:** Make sure your imports are sorted. This should be on line 6.

Resources/Locale/en-US/devicelinking/components/power-threshold-component.ftl
```diff
@@ -0,0 +1,12 @@
+### Power threshold component
+
+## Displayed in change threshold amount verb's name
+comp-power-threshold-verb-custom-amount = Custom
+comp-power-threshold-verb-amount = {$amount} W
+comp-power-threshold-verb-toggle = Toggle to {$amount} W
```
Aeshus, line 6: **issue:** Once the numbers get large enough, the verb list can get pretty unreadable. As such, you should use the `POWERWATTS` function to handle the unit conversions. 

It's probably fine to keep some of the localization strings in W for e.g. setting custom thresholds, though the verb menu should definitely localize it..

https://github.com/space-wizards/space-station-14/blob/3192a7fde52c435f443cc6c49426d96c39c59f0a/Resources/Locale/en-US/components/power-monitoring-component.ftl#L16-L17

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml.cs
```diff
@@ -0,0 +1,36 @@
...
+    [GenerateTypedNameReferences]
+    public sealed partial class ThresholdAmountWindow : DefaultWindow
+    {
+        private int _max = Int32.MaxValue;
+        private int _min = 1;
+
+        public ThresholdAmountWindow()
+        {
+            RobustXamlLoader.Load(this);
```
Aeshus, line 16: **style:** Put an extra line here.

```suggestion
            RobustXamlLoader.Load(this);
            
```

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml.cs
```diff
@@ -0,0 +1,36 @@
...
+            AmountLineEdit.OnTextChanged += OnValueChanged;
+        }
+
+        public void SetBounds(int min, int max)
```
Aeshus, line 20: **issue:** Please document this function.

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+          _min = min;
+          _max = max;
+          MinimumAmount.Text = Loc.GetString("comp-power-threshold-set-amount-min", ("amount", _min));
+          MaximumAmount.Text = Loc.GetString("comp-power-threshold-set-amount-max", ("amount", _max));
```
Aeshus, line 25: **style:** Put a line here.

```suggestion

          MinimumAmount.Text = Loc.GetString("comp-power-threshold-set-amount-min", ("amount", _min));
          MaximumAmount.Text = Loc.GetString("comp-power-threshold-set-amount-max", ("amount", _max));
```

Content.Client/DeviceLinking/UI/ThresholdAmountBoundUserInterface.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        protected override void Open()
+        {
+            base.Open();
+            _window = this.CreateWindow<ThresholdAmountWindow>();
+
+            if (_entManager.TryGetComponent<PowerThresholdComponent>(_owner, out var comp))
+                _window.SetBounds(comp.MinimumThresholdAmount.Int(), comp.MaximumThresholdAmount.Int());
+
```
Aeshus, line 31: **question:** I'm just wondering why you store both the int and FixedPoint2 representations of the same number? It seems like you cast between the two quite a bit.

Content.Client/DeviceLinking/UI/ThresholdAmountBoundUserInterface.cs
```diff
@@ -0,0 +1,42 @@
...
+          _owner = owner;
+          _entManager = IoCManager.Resolve<IEntityManager>();
+        }
+
+        protected override void Open()
+        {
+            base.Open();
+            _window = this.CreateWindow<ThresholdAmountWindow>();
+
+            if (_entManager.TryGetComponent<PowerThresholdComponent>(_owner, out var comp))
```
Aeshus, line 29: **issue:** BoundUserInterface already has an internal `Owner` and `EntMan` field that you can access. 

**suggestion:** You should just be able to do this:

```suggestion
            if (Entman.TryGetComponent<PowerThresholdComponent>(Owner, out var comp))
```

Example:

https://github.com/space-wizards/space-station-14/blob/3192a7fde52c435f443cc6c49426d96c39c59f0a/Content.Client/Mech/Ui/MechBoundUserInterface.cs#L46-L47

Content.Shared/DeviceLinking/SharedThresholdAmount.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Content.Shared.FixedPoint;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.DeviceLinking
+{
```
Aeshus, line 5: **style:** This needs to use file-scoped namespaces.

Content.Shared/DeviceLinking/Systems/PowerThresholdSystem.cs
Aeshus: Yeah I'd probably recommend against its use, especially as it seems to only be used in one place. 

Either come up with another name (like `e`), or just use `args` directory instead of a rename?

Content.Shared/DeviceLinking/Systems/PowerThresholdSystem.cs
Aeshus: **issue:** I would probably avoid hard-coding this within the system.

I think it'd be better served as being inside the component, so in the future, there could be an "upgraded" version that may have different defaults/higher supported thresholds.

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml.cs
```diff
@@ -0,0 +1,36 @@
...
+using Robust.Client.UserInterface.CustomControls;
+using Robust.Client.UserInterface.XAML;
+
+namespace Content.Client.DeviceLinking.UI
+{
+    [GenerateTypedNameReferences]
+    public sealed partial class ThresholdAmountWindow : DefaultWindow
+    {
+        private int _max = Int32.MaxValue;
+        private int _min = 1;
```
Aeshus, line 12: **question:** Wouldn't the component already include this information? 

If the component was changed to support a smaller range, the UI would lie to the player and present a larger supported range.

Aeshus (changes requested):
Thanks for the PR!

I just had a cursory glance, but I can't wait to see this in the game! Logic gates are really fun to play with, and having more inputs can never hurt.

Content.Shared/DeviceLinking/Systems/PowerThresholdSystem.cs
Vortebo: > Where do these thresholds come from? Trial & error?

They're entirely arbitrary. I wanted to provide defaults between 0 (not gonna be less than that) and a megawatt (pretty big, dunno how often they'd go over that), but beyond that I just kinda chose things

> I would probably avoid hard-coding this within the system

Yeah, that makes sense

Vortebo:
Thank y'alls! I'll get on these when work gives me a chance.

I should say some of the stuff that's been pointed out was copied directly from other files, so those will probably also benefit from a touch-up at some point. I'll make a list of which ones when I'm working through the corrections

Aeshus:
> I should say some of the stuff that's been pointed out was copied directly from other files, so those will probably also benefit from a touch-up at some point. I'll make a list of which ones when I'm working through the corrections

Yeah there’s definitely some file convention issues in the codebase, but it’s policy to not change that stuff unless you’re making other changes to that file.

This is because SS14 is so heavily forked, as if a downstream added a feature and you then change the namespace scope, you’ve created a new merge conflict (whitespace conflicts can get pretty hectic pretty fast). At least when you’re adding a new feature (or refactor, or bugfix, etc), you can justify it as there’s be a merge conflict anyways. We just don’t make PRs specifically for code style as it will just cause issues with no benefit.

If you wanna refactor them, you’re free to do that. Just don’t start formatting random files without any other changes!

Content.Client/DeviceLinking/UI/ThresholdAmountBoundUserInterface.cs
Vortebo: No good reason at all, gonna convert everything to int

Content.Client/DeviceLinking/UI/ThresholdAmountWindow.xaml.cs
Vortebo: It does, and in fact when ThresholdAmountBoundUserInterface.Open() is called it passes those values to ThresholdAmountWindow.SetBounds(), making these defaults entirely unused.

Content.Server/DeviceLinking/Systems/PowerSensorSystem.cs
Vortebo: We'd still want to save belowState to comp.BelowState, which I think would start to make this portion look a bit repetitive

Vortebo:
Well I _would_ check to make sure everything is still working but for some reason I can't currently debug even when I switch back to master

Vortebo:
Ah, I needed to upgrade dotnet

Vortebo:
Sweet, still works

PR #36207, opened 2025-03-30
Misc files warnings cleanup

ljm862:
## About the PR

Read #36156 
All the remaining files from batch 1 of the warnings cleanup. Files that don't fit elsewhere in the other groupings and making about 10 more 1-file PRs is no one's idea of fun.

## Technical details
Using non-obsolete methods in most cases. Two larger file changes I want to justify:
- `EmptyAllContainersBehaviour` moved stuff around and made the code more explicit as well as adding guards.
- `GridInRange` This one I removed the foreach and replaced it with a .count. _mapGrids is now a List due to the non-obsolete method requiring that and so the foreach no longer is more performant with not being an IEnumerable.

## Requirements

### DIFF
--- a/Content.Shared/Random/Rules/GridInRange.cs
+++ b/Content.Shared/Random/Rules/GridInRange.cs
@@ -1,5 +1,7 @@
 using System.Numerics;
 using Robust.Shared.Map;
+using Robust.Shared.Map.Components;
+using Robust.Shared.Maths;
 
 namespace Content.Shared.Random.Rules;
 
@@ -29,7 +31,9 @@ public sealed partial class GridInRangeRule : RulesRule
         var worldPos = transform.GetWorldPosition(xform);
         var gridRange = new Vector2(Range, Range);
 
-        foreach (var _ in mapManager.FindGridsIntersecting(xform.MapID, new Box2(worldPos - gridRange, worldPos + gridRange)))
+        List<Entity<MapGridComponent>> mapGrids = new();
+        mapManager.FindGridsIntersecting(xform.MapID, new Box2(worldPos - gridRange, worldPos + gridRange), ref mapGrids);
+        foreach (var _ in mapGrids)
         {
             return !Inverted;
         }
--- a/Content.Shared/Random/Rules/GridInRange.cs
+++ b/Content.Shared/Random/Rules/GridInRange.cs
@@ -1,7 +1,6 @@
 using System.Numerics;
 using Robust.Shared.Map;
 using Robust.Shared.Map.Components;
-using Robust.Shared.Maths;
 
 namespace Content.Shared.Random.Rules;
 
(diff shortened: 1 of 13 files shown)


Content.Shared/Random/Rules/GridInRange.cs
```diff
@@ -29,11 +32,9 @@ public override bool Check(EntityManager entManager, EntityUid uid)
         var worldPos = transform.GetWorldPosition(xform);
         var gridRange = new Vector2(Range, Range);
 
-        foreach (var _ in mapManager.FindGridsIntersecting(xform.MapID, new Box2(worldPos - gridRange, worldPos + gridRange)))
-        {
-            return !Inverted;
-        }
+        _mapGrids.Clear();
+        mapManager.FindGridsIntersecting(xform.MapID, new Box2(worldPos - gridRange, worldPos + gridRange), ref _mapGrids);
```
metalgearsloth, line 36: This one's a bit weird as the new code is slower as it won't early out, you might need to use the callback version instead.

metalgearsloth (changes requested):

Content.Shared/Random/Rules/GridInRange.cs
ljm862: Makes sense! There were a few overloads to choose from, I think this one should be fine given the necessary logic here!

Content.Shared/Random/Rules/GridInRange.cs
```diff
@@ -29,10 +29,8 @@ public override bool Check(EntityManager entManager, EntityUid uid)
         var worldPos = transform.GetWorldPosition(xform);
         var gridRange = new Vector2(Range, Range);
 
-        foreach (var _ in mapManager.FindGridsIntersecting(xform.MapID, new Box2(worldPos - gridRange, worldPos + gridRange)))
-        {
-            return !Inverted;
-        }
+        mapManager.FindGridsIntersecting(xform.MapID, new Box2(worldPos - gridRange, worldPos + gridRange),
+            (uid, grid) => { return !Inverted; });
```
metalgearsloth, line 33: Does this just continue the callback?

ljm862: I think it would have early exited for the only instance it exists in the code base that I could find (`InSpace` audio rule in audio.yml)
has Inverted as `true` which would return false here and exit out early. Though it wouldn't have returned the correct value overall, which I missed.

Now I've fixed that and make it more explicit by returning false to exit out of the tree query on call and then return from `Check` with what it should actually use.

VerinSenpai:
O/ for the cleanup. PZW here we come.

ljm862:
Trying to resolve the conflict on mobile is near impossible. I'll sort it out in about 2 weeks when I'm home unless someone else does it

slarticodefast:
Closing as derelict.
At this point it's probably easier to start from scratch than to fix all the conflicts.

PR #36209, opened 2025-03-30
Added the ability to refuel torches (and other expendable lights)

mjarduk:
## About the PR
This PR adds the ability to refuel entities with the ExpendableLights component.

## Why / Balance
Previously, torches could not be refueled and left wasted items, making a survival experience terrible. This patch fixes the issue by adding the ability to refuel expendable lights. (toggle-able in the component definition, disabled by default)

## Technical details
This patch makes two revamps, namely:
* Adds extra parameters to allow refueling 
* Changes the ExpendableLightSystem to change names using the NameModifierSystem

## Media
https://files.catbox.moe/oatj4r.mov

## Requirements

## Breaking changes
This PR removes the spentDesc datafield, as there is no non-hacky ways that i know of to restore the item's description to its default value. ~~Whatever parses the prototypes doesn't complain about it for now, so it's fine as is. If you're going to touch up any entities with the ExpendableLight component, make sure to remove the spentDesc field.~~ All prototypes with the deprecated field have been removed

**Changelog**
:cl: mjarduk
- tweak: Torches can now be refueled.
- fix: Burnt torches/glowsticks names' can now display if they're glued, cluwnified, etc.

### DIFF
--- a/Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
+++ b/Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
@@ -63,7 +63,7 @@ namespace Content.Server.Light.EntitySystems
                 {
                     case ExpendableLightState.Lit:
                         component.CurrentState = ExpendableLightState.Fading;
-                        component.StateExpiryTime = component.FadeOutDuration;
+                        component.StateExpiryTime = (float)component.FadeOutDuration.TotalSeconds;
 
                         UpdateVisualizer(ent);
 
@@ -124,13 +124,13 @@ namespace Content.Server.Light.EntitySystems
             if (stack.StackTypeId != component.RefuelMaterialID)
                 return;
 
-            if (component.StateExpiryTime + component.RefuelMaterialTime >= component.RefuelMaximumDuration)
+            if (component.StateExpiryTime + component.RefuelMaterialTime.TotalSeconds >= component.RefuelMaximumDuration.TotalSeconds)
                 return;
 
             if (component.CurrentState is ExpendableLightState.Dead)
             {
                 component.CurrentState = ExpendableLightState.BrandNew;
-                component.StateExpiryTime = component.RefuelMaterialTime;
+                component.StateExpiryTime = (float)component.RefuelMaterialTime.TotalSeconds;
 
                 _nameModifier.RefreshNameModifiers(uid);
                 _stackSystem.SetCount(args.Used, stack.Count - 1, stack);
@@ -138,15 +138,15 @@ namespace Content.Server.Light.EntitySystems
                 return;
             }
 
-            component.StateExpiryTime += component.RefuelMaterialTime;
+            component.StateExpiryTime += (float)component.RefuelMaterialTime.TotalSeconds;
             _stackSystem.SetCount(args.Used, stack.Count - 1, stack);
             args.Handled = true;
         }
 
-        private void OnRefreshNameModifiers(EntityUid uid, ExpendableLightComponent component, RefreshNameModifiersEvent args)
+        private void OnRefreshNameModifiers(Entity<ExpendableLightComponent> entity, ref RefreshNameModifiersEvent args)
         {
-            if (component.CurrentState is ExpendableLightState.Dead)
-                args.AddModifier(component.SpentName);
+            if (entity.Comp.CurrentState is ExpendableLightState.Dead)
+                args.AddModifier("expendable-light-spent-prefix");
         }
 
         private void UpdateVisualizer(Entity<ExpendableLightComponent> ent, AppearanceComponent? appearance = null)
@@ -205,7 +205,7 @@ namespace Content.Server.Light.EntitySystems
             }
 
             component.CurrentState = ExpendableLightState.BrandNew;
-            component.StateExpiryTime = component.GlowDuration;
+            component.StateExpiryTime = (float)component.GlowDuration.TotalSeconds;
             EntityManager.EnsureComponent<PointLightComponent>(uid);
         }
 
--- a/Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
+++ b/Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
@@ -1,10 +1,14 @@
 using Content.Server.Light.Components;
+using Content.Server.Stack;
 using Content.Shared.Clothing.Components;
 using Content.Shared.Clothing.EntitySystems;
 using Content.Shared.IgnitionSource;
+using Content.Shared.Interaction;
 using Content.Shared.Interaction.Events;
 using Content.Shared.Item;
 using Content.Shared.Light.Components;
+using Content.Shared.NameModifier.EntitySystems;
+using Content.Shared.Stacks;
 using Content.Shared.Tag;
 using Content.Shared.Verbs;
 using JetBrains.Annotations;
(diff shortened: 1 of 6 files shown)


Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
```diff
@@ -65,9 +72,9 @@ private void UpdateLight(Entity<ExpendableLightComponent> ent, float frameTime)
                     default:
                     case ExpendableLightState.Fading:
                         component.CurrentState = ExpendableLightState.Dead;
-                        var meta = MetaData(ent);
-                        _metaData.SetEntityName(ent, Loc.GetString(component.SpentName), meta);
-                        _metaData.SetEntityDescription(ent, Loc.GetString(component.SpentDesc), meta);
+                        _nameModifier.RefreshNameModifiers(ent.Owner);
+
+                        _tagSystem.AddTag(ent, "Trash");
```
ljm862, line 77: Probably don't need it twice!

mjarduk: Fixed in the next commit!

Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
```diff
@@ -101,15 +108,47 @@ public bool TryActivate(Entity<ExpendableLightComponent> ent)
...
+            }
+            return true;
+        }
+
+        private void OnInteractUsing(EntityUid uid, ExpendableLightComponent component, ref InteractUsingEvent args)
+        {
+            if (args.Handled)
+                return;
+
+            if (!EntityManager.TryGetComponent(args.Used, out StackComponent? stack))
```
ljm862, line 121: Can use the shorthand `TryComp<StackComponent>`

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
```diff
@@ -23,12 +23,18 @@ public abstract partial class SharedExpendableLightComponent : Component
     [DataField("fadeOutDuration")]
     public float FadeOutDuration { get; set; } = 60 * 5f;
 
-    [DataField("spentDesc")]
```
Killerqu00, line 26: removing fields needs a notice in breaking change, maybe there is no need to remove it at all

mjarduk: Pointed it out in the PR description, ty

Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
mjarduk: Once again, fixed, in the next commit!

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
```diff
@@ -23,12 +23,18 @@ public abstract partial class SharedExpendableLightComponent : Component
...
-    public string SpentDesc { get; set; } = string.Empty;
-
     [DataField("spentName")]
     public string SpentName { get; set; } = string.Empty;
 
+    [DataField("refuelMaterialID")]
```
slarticodefast, line 29:
```suggestion
    [DataField]
```
name gets automatically converted to camelCase, so this is redundant
same for the others

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
```diff
@@ -23,12 +23,18 @@ public abstract partial class SharedExpendableLightComponent : Component
...
+    public string? RefuelMaterialID { get; set; } = null;
```
slarticodefast, line 30: Should be a `ProtoId<StackPrototype>` so it gets validated.
Getters and setters are not needed anymore.

Also not sure if I like hardcoding stacks into this like this. Might be possible to use construction graphs for this, so you can also refill them with other items that are not stacks.

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
```diff
@@ -23,12 +23,18 @@ public abstract partial class SharedExpendableLightComponent : Component
...
+
+    [DataField("refuelMaterialTime")]
+    public float RefuelMaterialTime { get; set; } = 15f;
```
slarticodefast, line 33: all of these should be timespans

Resources/Prototypes/Entities/Objects/Misc/torch.yml
```diff
@@ -79,3 +80,5 @@
     - type: Tag
       tags:
       - Torch
+    - type: NameModifier
+      baseName: torch
```
slarticodefast, line 84: This is not how namemodifiers should be used. The base name should not be set like this, the name modifier system should *modify* existing names. Look at how other systems use this. Also this is not localized like this.

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
```diff
@@ -23,12 +23,18 @@ public abstract partial class SharedExpendableLightComponent : Component
     [DataField("fadeOutDuration")]
     public float FadeOutDuration { get; set; } = 60 * 5f;
 
-    [DataField("spentDesc")]
-    public string SpentDesc { get; set; } = string.Empty;
```
slarticodefast, line 27: this datafield is still used in yaml

slarticodefast (changes requested):

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
mjarduk: I'm not really familiar with construction graphs, so i don't really know how to do it. Thank you for the ProtoId stuff, though.

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
mjarduk: Fixed in the next comment.

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
mjarduk: Fixed in the next commit.

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
mjarduk: Fixed in the next commit.

Content.Shared/Light/Components/SharedExpendableLightComponent.cs
mjarduk: Fixed in the next commit.

Resources/Prototypes/Entities/Objects/Misc/torch.yml
mjarduk: Fixed in the next commit! :DDD

EmoGarbage404 (approved):

cfif126:
Is it possible to refuel with average wood or it works only with wooden planks?

Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
```diff
@@ -168,6 +205,7 @@ private void OnExpLightInit(EntityUid uid, ExpendableLightComponent component, C
             }
 
             component.CurrentState = ExpendableLightState.BrandNew;
+            component.StateExpiryTime = (float)component.GlowDuration.TotalSeconds;
```
slarticodefast, line 208: instead of casting this everywhere `StateExpiryTime` should be a `TimeSpan` as well

mjarduk:
> Is it possible to refuel torches with average wood or it works only with wooden planks?

No, however that can get added i think:)

PR #36210, opened 2025-03-30
Cleanup and small update to the stethoscope!

beck-thompson:
## About the PR

Did a lot of cleanup of the system and also made the stethoscope slightly more useful!

## Why / Balance

Stethoscopes are basically useless when you have health analyzers so I've done two changes:
1.) Stethoscopes now automatically restart doafters so you don't have to constantly click patients
2.) Stethoscopes now tell you if a entity is gaining or losing oxygen damage. It isn't perfect but its pretty good!

## Technical details

1.) Moved everything to shared. This was done basically without issue! I did have to not check for respirator component but I think that's OK.
2.) Added inventory relaying to `InnateVerb`s. This was done in a slightly odd way but its `EquipmentVerb` does it! Otherwise you would have to add inventory relay to all verbs which doesn't make sense for 90% of them.
3.) Removed the `WearingStethoscopeComponent` because it is not longer necessary.

## Media

https://github.com/user-attachments/assets/20425f8f-f5a2-4c2a-b26d-02bacba9000d

## Requirements

## Breaking changes

The `StethoscopeComponent` and `StethoscopeSystem` have both been moved to shared.
`WearingStethoscopeComponent` has been removed (If you still need something like this use the new inventory relay on `InnateVerb`s)

**Changelog**

:cl:
- add: Stethoscopes now automatically start doafters and also can tell if a patient is losing oxygen damage or gaining it.
- fix: Moths can no longer eat stethoscopes.
- fix: Stethoscopes action button now works properly.


Resources/Locale/en-US/health-examinable/stethoscope.ftl
```diff
@@ -1,6 +1,15 @@
...
+
 stethoscope-normal = You hear normal breathing.
+stethoscope-raggedy = You hear raggedy breathing.
 stethoscope-hyper = You hear hyperventilation.
 stethoscope-irregular = You hear hyperventilation with an irregular pattern.
 stethoscope-fucked = You hear twitchy, labored breathing interspersed with short gasps.
+
+stethoscope-delta-steady = Its steady.
+stethoscope-delta-improving = Its improving.
+stethoscope-delta-worsting = Its getting worse.
```
Tayrtahn, line 13:
```suggestion
stethoscope-delta-steady = It's steady.
stethoscope-delta-improving = It's improving.
stethoscope-delta-worsening = It's getting worse.
```

Content.Shared/Medical/Stethoscope/StethoscopeSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+            !damageComp.Damage.DamageDict.TryGetValue("Asphyxiation", out var asphyxDmg))
+        {
+            _popup.PopupPredicted(Loc.GetString("stethoscope-nothing"), target, user);
+            stethoscope.Comp.LastMeasuredDamage = null;
+            return;
+        }
+
+        var absString = GetAbsoluteDamageString(asphyxDmg);
+
+        // Don't show the change if this is the first time listening.d
```
Tayrtahn, line 109:
```suggestion
        // Don't show the change if this is the first time listening.
```

Content.Shared/Medical/Stethoscope/StethoscopeSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private void ExamineWithStethoscope(Entity<StethoscopeComponent> stethoscope, EntityUid user, EntityUid target)
+    {
+        // TODO: Add check for respirator component when it gets moved to shared.
+        // If the mob is dead or cannot asphyxiation damage, the popup shows nothing.
+        if (!TryComp<MobStateComponent>(target, out var mobState)                        ||
+            !TryComp<DamageableComponent>(target, out var damageComp) ||
+            _mobState.IsDead(target, mobState)                                           ||
+            !damageComp.Damage.DamageDict.TryGetValue("Asphyxiation", out var asphyxDmg))
```
Tayrtahn, line 100: This should probably be a const string.

Content.Shared/Medical/Stethoscope/StethoscopeSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+        };
+        return Loc.GetString(msg);
+    }
+
+    private string GetDeltaDamageString(FixedPoint2 lastDamage, FixedPoint2 currentDamage)
+    {
+        if (lastDamage > currentDamage)
+            return Loc.GetString("stethoscope-delta-improving");
+        if (lastDamage < currentDamage)
+            return Loc.GetString("stethoscope-delta-worsting");
```
Tayrtahn, line 141:
```suggestion
            return Loc.GetString("stethoscope-delta-worsening");
```

Content.Shared/Medical/Stethoscope/StethoscopeSystem.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        _doAfter.TryStartDoAfter(new DoAfterArgs(EntityManager, container.Owner, ent.Comp.Delay, new StethoscopeDoAfterEvent(), ent, target: target, used: ent)
+        {
+            DuplicateCondition = DuplicateConditions.SameEvent,
+            NeedHand = true,
+            BreakOnMove = true,
+            Hidden = true,
+            BreakOnDropItem = false,
+            BreakOnHandChange = false,
+        });
```
Tayrtahn, line 75: Noticed this issue; I think it's caused by something around here.

https://github.com/user-attachments/assets/2c12dca7-c806-4e8f-a21c-eacb63870530

Content.Shared/Medical/Stethoscope/Components/StethoscopeComponent.cs
```diff
@@ -0,0 +1,31 @@
+using Content.Shared.FixedPoint;
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Medical.Stethoscope.Components;
+
+/// <summary>
+///     Adds a verb and action that allows the user to listen to the entities breathing.
```
Tayrtahn, line 8:
```suggestion
///     Adds a verb and action that allows the user to listen to the entity's breathing.
```

Content.Shared/Medical/Stethoscope/StethoscopeSystem.cs
beck-thompson: So I've done some digging, fixing this would require a new DoAfter field that's named something like "CancelOnContainerChange"! The issue is that there appears to be some weird serialization stuff that I don't quite understand :(

This also is still an issue on current master so it isn't *worse* (Still not good!)

I can try to fix this but I think it would probably be better in a separate pr because it'll be a pretty big change and require a lot of testing :laughing:

Let me know what you think!

Tayrtahn: Yeah, if this bug is already present on master, then I'm fine with it not being fixed here. Write up an issue for it and we'll fix it in another PR!

beck-thompson: Done! #36223

Tayrtahn (approved):

ArtisticRoomba:
Moths could do what???

PR #36212, opened 2025-03-30
Diphenhydramine causes drowsiness

iaada:
## About the PR

Your favorite poison cure is now great for naps, too!

## Why / Balance

Diphenhydramine is not a fantasy space drug, it's Benadryl. And Benadryl is clinically proven to make you sleepy.

But leaving the real world argument behind, Dip is way better than Dylo. Triple the heal with fringe additional benefits, I think it could use something in order to not feel so strictly better. The complexity of the recipe isn't much of a negative for chem-heads, and doctors and paitents don't even interact with that part of reagents. Live, no one would choose Dylo over Dip given the option. With this change *maybe* someone chooses Dylo to avoid the drowsiness.

## Technical details

Causes 1.5 seconds with accumulation, slightly below half as much as other drowsiness drugs.

## Media

image

## Requirements

**Changelog**

:cl: aada
- tweak: Diphenhydramine now causes light drowsiness.

### DIFF
--- a/Resources/Locale/en-US/reagents/meta/medicine.ftl
+++ b/Resources/Locale/en-US/reagents/meta/medicine.ftl
@@ -5,7 +5,7 @@ reagent-name-dylovene = dylovene
 reagent-desc-dylovene = A broad-spectrum anti-toxin, which treats toxin damage in organisms. Overdosing will cause vomiting, dizzyness and pain.
 
 reagent-name-diphenhydramine = diphenhydramine
-reagent-desc-diphenhydramine = Rapidly purges the body of histamine, reduces jitteriness, and treats poison damage.
+reagent-desc-diphenhydramine = Rapidly purges the body of histamine, reduces jitteriness, and treats poison damage. Great for naps, too!
 
 reagent-name-arithrazine = arithrazine
 reagent-desc-arithrazine = A mildly unstable medication used for the most extreme case of radiation poisoning. Exerts minor stress on the body.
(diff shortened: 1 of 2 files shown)


robinthedragon:
now it just needs to cause hallucinations on OD, and we will finally have Benadryl

iaada:
If I had the ability to not web edit rn I would definitely add that.

Resources/Locale/en-US/reagents/meta/medicine.ftl
```diff
@@ -5,7 +5,7 @@ reagent-name-dylovene = dylovene
 reagent-desc-dylovene = A broad-spectrum anti-toxin, which treats toxin damage in organisms. Overdosing will cause vomiting, dizzyness and pain.
 
 reagent-name-diphenhydramine = diphenhydramine
-reagent-desc-diphenhydramine = Rapidly purges the body of histamine, reduces jitteriness, and treats poison damage.
+reagent-desc-diphenhydramine = Rapidly purges the body of histamine, reduces jitteriness, and treats poison damage. Great for naps, too!
```
Aeshus, line 8: **issue (non-blocking):** I personally think the reagent descriptions should be more clinical (and not conversational), though it's not a big issue.

Resources/Prototypes/Reagents/medicine.yml
```diff
@@ -83,6 +83,12 @@
         key: Jitter
         time: 3.0
         type: Remove
+      - !type:GenericStatusEffect
+        key: Drowsiness
+        component: Drowsiness
+        time: 2
```
Aeshus, line 89: **issue (non-blocking):** One of the other maintainers brought up that they thought the current drowsiness was a little much, so if you want, you can tune it down a little bit.

I'm not personally worried about it being abused that much, but it never hurts to be careful!

Aeshus (approved):
Thanks for the PR!

I left some non-blocking comments you can choose to address/resolve. 

If you have any other ideas for metabolism effects, feel free to PR those too.

Resources/Prototypes/Reagents/medicine.yml
Aeshus: They suggested to change it to 1.5 instead.

beck-thompson (approved):
looks good!

PR #36214, opened 2025-03-30
Changes ParallaxGenerator to use FastNoiseLite

ljm862:
## About the PR

Instead of the `NoiseGenerator` we use `FastNoiseLite` If I'm correct, after this we'll be able to remove `NoiseGenerator` and `FastNoise` from the codebase entirely!

## Technical details

Replaces `NoiseGenerator` construction in `ParallaxGenerator` with `FastNoiseLite`.
The `FastNoiseLite` API doesn't have Period or Persistence attributes exposed anywhere so we might lose some functionality in this case. 
Though the `GetNoise` function we're calling in `NoiseGenerator` didn't seem to make use of these really anyway, but I could be wrong on that. 
#33279 

## Requirements

### DIFF
--- a/Content.Client/Parallax/ParallaxGenerator.cs
+++ b/Content.Client/Parallax/ParallaxGenerator.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


Killerqu00:
wouldn't it be a breaking change? if so, it needs to be listed in the PR description

metalgearsloth:
> wouldn't it be a breaking change? if so, it needs to be listed in the PR description

The engine one is the bigger breaking change I would think but will dump it there anyway.

PR #36217, opened 2025-03-30
fix species name in station records

lzk228:
## About the PR

it literally takes the species id for station records instead of the name

## Why / Balance

bug

## Technical details

index prototype and send its name, instead of the prototype itself

## Media

no

## Requirements

## Breaking changes

no

**Changelog**

no

### DIFF
--- a/Content.Server/StationRecords/Systems/StationRecordsSystem.cs
+++ b/Content.Server/StationRecords/Systems/StationRecordsSystem.cs
@@ -96,7 +96,9 @@ public sealed class StationRecordsSystem : SharedStationRecordsSystem
         TryComp<FingerprintComponent>(player, out var fingerprintComponent);
         TryComp<DnaComponent>(player, out var dnaComponent);
 
-        CreateGeneralRecord(station, idUid.Value, profile.Name, profile.Age, profile.Species, profile.Gender, jobId, fingerprintComponent?.Fingerprint, dnaComponent?.DNA, profile, records);
+        var speciesName = (_prototypeManager.Index(profile.Species)).Name;
+
+        CreateGeneralRecord(station, idUid.Value, profile.Name, profile.Age, speciesName, profile.Gender, jobId, fingerprintComponent?.Fingerprint, dnaComponent?.DNA, profile, records);
     }
 
 


Content.Server/StationRecords/Systems/StationRecordsSystem.cs
```diff
@@ -96,7 +96,9 @@ private void CreateGeneralRecord(EntityUid station, EntityUid player, HumanoidCh
         TryComp<FingerprintComponent>(player, out var fingerprintComponent);
         TryComp<DnaComponent>(player, out var dnaComponent);
 
-        CreateGeneralRecord(station, idUid.Value, profile.Name, profile.Age, profile.Species, profile.Gender, jobId, fingerprintComponent?.Fingerprint, dnaComponent?.DNA, profile, records);
+        var speciesName = (_prototypeManager.Index(profile.Species)).Name;
```
beck-thompson, line 99: Looks good but you need to use a locstring lookup here!

beck-thompson (changes requested):

Content.Server/StationRecords/Systems/StationRecordsSystem.cs
lzk228: isn't it already like doing this kind of magic before?

Tayrtahn (commented):
I think it would be better to fix this on the display side, rather than the record creation side.

That is, continue using the species' protoId to create the station record, and instead do the prototype lookup when displaying the record, right before the Loc.GetString call ([here](https://github.com/space-wizards/space-station-14/blob/master/Content.Client%2FStationRecords%2FGeneralRecord.xaml.cs#L20))

Content.Server/StationRecords/Systems/StationRecordsSystem.cs
Aeshus: Well it seems to display the species name as the Loc key, so that'll need to be fixed.

image

Tayrtahn: > isn't it already like doing this kind of magic before?

Nope, it was directly displaying the IDs of the `SpeciesPrototype`s, which conveniently happen to be the same as the display names of the species, but we _really_ shouldn't be relying on that.
https://github.com/space-wizards/space-station-14/blob/3192a7fde52c435f443cc6c49426d96c39c59f0a/Resources/Prototypes/Species/human.yml#L2

https://github.com/space-wizards/space-station-14/blob/3192a7fde52c435f443cc6c49426d96c39c59f0a/Resources/Prototypes/Species/diona.yml#L2

https://github.com/space-wizards/space-station-14/blob/3192a7fde52c435f443cc6c49426d96c39c59f0a/Resources/Prototypes/Species/reptilian.yml#L2

Actually, it looks like the station records have been displaying the wrong names for Moth and SlimePerson all this time - they should be displayed as "Moth Person" and "Slime Person", respectively.
https://github.com/space-wizards/space-station-14/blob/3192a7fde52c435f443cc6c49426d96c39c59f0a/Resources/Locale/en-US/species/species.ftl#L3-L12
<sup>and gingerbread people who somehow wind up in the station records should display as "delicious baked good"</sup>

MilonPL:
Is this still being worked on?

lzk228:
yes

beck-thompson (changes requested):
It seems to be mad and crashes the client

image

slarticodefast:
I pushed some fixes:
- The dependency injection was missing, causing an error.
- You were indexing the SpeciesPrototype as an EntityPrototypes, causing another error.
- You were displaying a loc string without resolving it.

In the future please make sure to test your changes in-game before requesting a review.

PR #36221, opened 2025-03-31
Fix KeyNotFoundException that sometimes happens on server shutdown

Tayrtahn:
## About the PR

Fixed a bug that could throw an exception when shutting down a development server.

## Why / Balance

Honestly not a very important bug, since it only affects servers running in debug mode and only happens when they're shutting down anyway. But it's been bothering me for a while and I wanted to fix it.

## Technical details

Shutting down the development server with a client connected sometimes causes a debug assert on the way out. I have not found a way to reliably reproduce this issue, but it seems to happen fairly frequently for me - I suspect that hardware specifics may play a part in it.

In any case, the stack trace pointed to `SharedPlayerManager.GetPlayerData` throwing a `KeyNotFoundException` when trying to index the player data, because the player session has already been removed. This PR swaps `SharedPlayerManager.GetPlayerData` for `SharedPlayerManager.TryGetPlayerData`, which won't throw an exception when the key isn't found.

## Requirements

### DIFF
--- a/Content.Server/Mind/MindSystem.cs
+++ b/Content.Server/Mind/MindSystem.cs
@@ -85,11 +85,11 @@ public sealed class MindSystem : SharedMindSystem
     {
         if (base.TryGetMind(user, out mindId, out mind))
         {
-            DebugTools.Assert(_players.GetPlayerData(user).ContentData() is not { } data || data.Mind == mindId);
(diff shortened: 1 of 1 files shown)


Aeshus (approved):
Looks good to me!

PR #36232, opened 2025-03-31
Rework the way held items scatter when holder is knocked down

Tayrtahn:
## About the PR

Rewrites the logic/math behind the how held items are thrown when someone slips or otherwise gets knocked down.

Notably:
- Items are now thrown in the direction the holder is moving, +/- 45 degrees. Previously they could be thrown in any direction.
- There is no longer a 45% chance for the item to just get dropped at the holder's feet instead of being thrown. This was almost certainly a bug caused by a stray minus sign in the code. Items can still get thrown very short distances (even zero), but it's now just as likely as any other distance.
- Item mass is now taken into account: lighter items fly farther than heavier ones. This is based on the item's mass, which is in turn based on the size and density of its fixtures. A lot of items don't have these values customized, so expect there to be some oddities (ex: a piece of paper weighs as much as a crowbar), but these YAML values can be corrected over time and are outside the scope of this PR.

As with the previous implementation, the lack of prediction for throwing is a bit unpleasant. The items scatter a moment after the player slips on the local client. This was the case before, but the accidental 45% chance of dropping the item in place covered it up a bit.

## Why / Balance

It's pretty funny now to see a player's held items go flying out in front of them as they slip, and it makes more sense than having them sometimes go backward or often just appear on the ground at their feet.

There are some slight balance implications, because this could theoretically make it a little bit easier to pick up your dropped items when you slip. As before, you do still need to wait until you stand up to do so though, so the change is very minimal.

## Technical details

Replaced some 4 year old code which did some funky things, some of which didn't seem to work.
- The holder's velocity is taken from their physics component.
- Each item's velocity is based on the holder's velocity, then rotated by a random amount up to a const value (`DropHeldItemsSpread`) degrees, in the positive or negative direction.
- The item's velocity is then scaled by a random value from 0 to 1, varying the distance it will travel.
- It is then divided by the item's physical mass, giving lighter items a higher velocity and vice versa.
- The `baseThrowSpeed` of the throw is set to half of the holder's `BaseThrowSpeed` (since it's not a deliberate throw), then given a +/-10% variation.

## Media

https://github.com/user-attachments/assets/aec08bd7-5a86-4d1a-be1a-f40ac0ac6dfe

Butter weighs less than a crowbar.

## Requirements

**Changelog**

:cl:
- tweak: Items thrown when someone slips now tend to scatter in the direction they are moving, and respect the item's mass.

### DIFF
--- a/Content.Server/Hands/Systems/HandsSystem.cs
+++ b/Content.Server/Hands/Systems/HandsSystem.cs
@@ -268,8 +268,8 @@ namespace Content.Server.Hands.Systems
                 // Decrease the distance of the throw by a random amount
                 itemVelocity *= _random.NextFloat(1f);
                 // Heavier objects don't get thrown as far
-                if (TryComp<PhysicsComponent>(held, out var heldPhysics))
-                    itemVelocity *= heldPhysics.InvMass;
+                // If the item doesn't have a physics component, it isn't going to get thrown anyway, but we'll assume infinite mass
+                itemVelocity *= TryComp<PhysicsComponent>(held, out var heldPhysics) ? heldPhysics.InvMass : 0;
                 // Throw at half the holder's intentional throw speed and
                 // vary the speed a little to make it look more interesting
                 var throwSpeed = entity.Comp.BaseThrowspeed * 0.5f * _random.NextFloat(0.9f, 1.1f);
--- a/Content.Server/Hands/Systems/HandsSystem.cs
+++ b/Content.Server/Hands/Systems/HandsSystem.cs
@@ -247,6 +247,7 @@ namespace Content.Server.Hands.Systems
         {
             // If the holder doesn't have a physics component, they ain't moving
             var holderVelocity = _physicsQuery.TryComp(entity, out var physics) ? physics.LinearVelocity : Vector2.Zero;
+            var spreadMaxAngle = Angle.FromDegrees(DropHeldItemsSpread);
 
             var fellEvent = new FellDownEvent(entity);
             RaiseLocalEvent(entity, fellEvent, false);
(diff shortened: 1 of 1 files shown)


metalgearsloth (approved):

Content.Server/Hands/Systems/HandsSystem.cs
```diff
@@ -255,10 +265,26 @@ private void OnDropHandItems(Entity<HandsComponent> entity, ref DropHandItemsEve
                 if (!TryDrop(entity, hand, null, checkActionBlocker: false, handsComp: entity.Comp))
                     continue;
 
+                // Rotate the item's throw vector a bit for each item
+                var angleOffset = Angle.FromDegrees(_random.NextFloat(-1, 1) * DropHeldItemsSpread);
```
beck-thompson, line 269: What about `_random.NextAngle`? I think `_random.NextAngle(-DropHeldItemsSpread, DropHeldItemsSpread);` should work

beck-thompson (changes requested):
Looks good just one really minor thing!

Content.Server/Hands/Systems/HandsSystem.cs
Tayrtahn: It's gotta be `_random.NextAngle(Angle.FromDegrees(-DropHeldItemsSpread), Angle.FromDegrees(DropHeldItemsSpread))`, but that should work.

beck-thompson (approved):
It being unpredicted makes me sad but it already is so no difference :laughing:

Looks good a lot cleaner than before :+1:

Tayrtahn:
> It being unpredicted makes me sad

Agreed. Might have to do something about that.

PR #36238, opened 2025-03-31
Power extensioncablesystem cleanups

ljm862:
## About the PR

Removed warnings from the `ExtensionCableSystem`. 
Part of #33279 

## Technical details

Removed 10 warnings from the system. Had to set the components to be `List<Entity<T>>` to track who owned the components.
2 things I'm not sure about:
- `MapSystem` caching, is it worth it? I'm not sure if it would be straightforward like the others but I could be wrong.
- The fact I'm now setting some components to be `(uid, null)` I don't like. Will they be GC'd? Don't want to cause memory issues if someone sets/unsets 1000000 of them.

## Requirements

## Breaking changes

Anything using `ExtensionCableReceiverComponent` or `ExtensionCableProviderComponent` now needs to be an `List<Entity<T>>`

### DIFF
--- a/Content.Server/Power/Components/ExtensionCableProviderComponent.cs
+++ b/Content.Server/Power/Components/ExtensionCableProviderComponent.cs
@@ -1,4 +1,4 @@
-﻿using Content.Server.Power.EntitySystems;
+using Content.Server.Power.EntitySystems;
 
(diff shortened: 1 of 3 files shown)


ljm862:
ahhh already done here! #36052

PR #36244, opened 2025-04-01
Criminal console status expansion

B-Kirill:
## About the PR
- Added new `Hostile` and `Eliminated` criminal console statuses

## Why / Balance
The new statuses were added for convenience:

`Hostile` status is similar to `Wanted`, but specifically marks extremely dangerous criminals. For what? It's simple. Instead of trying to remember, asking over comms, or checking a PDA to learn why someone is wanted - whether it's for minor vandalism or multiple mass-murder terrorist acts - this status lets security officers immediately recognize the threat level. At minimum, they'll be mentally prepared, denying the criminal time to escape. This is especially useful during high-pop rounds with many agents.

`Eliminated` status conveniently marks executed players or any others who shouldn't be revived. This eliminates unnecessary questions like "Is this a civilian victim or an executed criminal?" Also helpful when there are multiple corpses in the brig, wherever they came from.

## Media

Comparison:
before
after4

Just screenshots that everything is working as it should:
console4

## Notes
Feedback is welcome!

## Requirements

**Changelog**
:cl: B_Kirill
add: Hostile and Eliminated criminal statuses

### DIFF
--- a/Content.Client/CartridgeLoader/Cartridges/WantedListUiFragment.cs
+++ b/Content.Client/CartridgeLoader/Cartridges/WantedListUiFragment.cs
@@ -201,7 +201,7 @@ public sealed partial class WantedListUiFragment : BoxContainer
         var label = new Label() { Text = record.TargetInfo.Name };
         var rect = new TextureRect()
         {
-            TextureScale = new(3.0f),
+            TextureScale = new(2.2f),
             VerticalAlignment = VAlignment.Center,
             HorizontalAlignment = HAlignment.Center,
             Margin = new(0f, 0f, 6f, 0f),


slarticodefast:
If they are twice the size of the other status icons, do they fit together with the job and health icons?
Also the mindshield icon in particular.
And afaik the art maintainers usually don't like having mixels.

B-Kirill:
> If they are twice the size of the other status icons, do they fit together with the job and health icons? Also the mindshield icon in particular. And afaik the art maintainers usually don't like having mixels.

It works 100% with mindshield and job icons, but it's about health.. Damn, I didn't think to check

B-Kirill:
Well, everything seems to be as it should be...
Although in the code I tried to make sure that their reduced rendering should not affect other icons at all.

slarticodefast:
That is a random, unrelated test fail. I can rerun the test for you.

B-Kirill:
> That is a random, unrelated test fail. I can rerun the test for you.

Oh, thanks

Pyrizhok333x2:
The idea of ​​adding two new criminal statuses really sounds good.
But why can't we use the regular 8x8 icon size?
The skull image can fit there, and as for the exclamation mark, it's literally already in the criminal status, it just needs to be repainted.
Or am I missing something?

Mixelz:
> The idea of ​​adding two new criminal statuses really sounds good. But why can't we use the regular 8x8 icon size? The skull image can fit there, and as for the exclamation mark, it's literally already in the criminal status, it just needs to be repainted. Or am I missing something?

> If they are twice the size of the other status icons, do they fit together with the job and health icons? Also the mindshield icon in particular. And afaik the art maintainers usually don't like having mixels.

You guys are misunderstanding how the images were rendered. They were previous 8x8 images being upscaled into 24x24. 
This is why they've always looked weird or "lower quality" than the other icon images, because they actually were using less pixels and did not match the same ratio as other UI elements.
They images are now 16x16, being upscaled to 24x24, so the resulting image is much crisper and cleaner because the base image is now x2 in size.

Mixelz:
My only suggestion would be to maybe use a different color for the Hostile icon due to it being potentially too similar to the Wanted and thus being hard to tell at a glance.

Content.Client/StatusIcon/StatusIconOverlay.cs
```diff
@@ -117,7 +122,17 @@ protected override void Draw(in OverlayDrawArgs args)
...
+                {
+                    var originalTransform = handle.GetTransform();
+                    handle.SetTransform(Matrix3x2.Multiply(Matrix3x2.CreateScale(scale), originalTransform));
+                    handle.DrawTexture(texture, position / scale);
+                    handle.SetTransform(originalTransform);
+                }
+                else
+                {
+                    handle.DrawTexture(texture, position);
+                }
```
ljm862, line 135: Here scale == 1f, do the `handle.DrawTexture` could be extracted and we can use 
`handle.DrawTexture(texture, position / scale)` regardless.

Content.Shared/Security/SecurityStatus.cs
```diff
@@ -6,16 +6,20 @@
...
 /// Paroled - the person is on parole
 /// Discharged - the person has been released from prison
+/// Eliminated - the person has been eliminated and should not be healed
 /// </summary>
 public enum SecurityStatus : byte
 {
     None,
     Suspected,
     Wanted,
+    Hostile,
```
ljm862, line 20: I'm always wary of adding enums in the middle of existing ones. If there's been a cast at any point from a value to an enum this will now break it. You're probably fine here, just something to be wary of!

slarticodefast:
It does seem the Urist sprite is upscaled by the same factor though. So now the icon has twice the resolution compared to that.

TeenSarlacc:
> My only suggestion would be to maybe use a different color for the Hostile icon due to it being potentially too similar to the Wanted and thus being hard to tell at a glance. 

animated sprite with the exclamation point flashing?

Resources/Locale/en-US/criminal-records/criminal-records.ftl
```diff
@@ -16,6 +16,8 @@ criminal-records-status-detained = Detained
 criminal-records-status-suspected = Suspect
 criminal-records-status-discharged = Discharged
 criminal-records-status-paroled = Paroled
+criminal-records-status-hostile = Hostile
+criminal-records-status-eliminated = Eliminated
```
IProduceWidgets, line 20:
```suggestion
criminal-records-status-eliminated = Unrecoverable
```
Synergizes better with the morgue.

Resources/Locale/en-US/criminal-records/criminal-records.ftl
```diff
@@ -42,6 +44,10 @@ criminal-records-console-detained = {$name} ({$job}) has been detained by {$offi
 criminal-records-console-released = {$name} ({$job}) has been released by {$officer}.
 criminal-records-console-paroled = {$name} ({$job}) has been released on parole by {$officer}.
 criminal-records-console-not-parole = {$officer} cleared the parole status of {$name} ({$job}).
+criminal-records-console-hostile = {$name} ({$job}) was marked as hostile by {$officer} for: {$reason}.
+criminal-records-console-not-hostile = {$name} ({$job}) no longer marked as hostile by {$officer}.
+criminal-records-console-eliminated = {$name} ({$job}) was marked as eliminated by {$officer}.
```
IProduceWidgets, line 49:
```suggestion
criminal-records-console-eliminated = {$name} ({$job}) was marked as unrecoverable by {$officer}.
```

Content.Shared/Security/SecurityStatus.cs
```diff
@@ -6,16 +6,20 @@
 /// None - the default value
 /// Suspected - the person is suspected of doing something illegal
 /// Wanted - the person is being wanted by security
+/// Hostile - the person has been admitted as hostile
 /// Detained - the person is detained by security
 /// Paroled - the person is on parole
 /// Discharged - the person has been released from prison
+/// Eliminated - the person has been eliminated and should not be healed
```
IProduceWidgets, line 13:
```suggestion
/// Unrecoverable - the person is dead and unlikely to or should not be revived.
```

IProduceWidgets (changes requested):

Content.Shared/Security/SecurityStatus.cs
B-Kirill: > I'm always wary of adding enums in the middle of existing ones. If there's been a cast at any point from a value to an enum this will now break it. You're probably fine here, just something to be wary of!

In this case, everything is as it should be, the order in this enum determines the order in the criminal console.

Content.Client/StatusIcon/StatusIconOverlay.cs
B-Kirill: > Here scale == 1f, so the `handle.DrawTexture` could be extracted and we can use `handle.DrawTexture(texture, position / scale)` regardless.

I think it's pointless. If the icon has a scale flag, we will get it and run the required piece of code, otherwise we will have 1. So in any case, dividing by 1 is just an unnecessary action without profit.
(I hope I understood you correctly)

B-Kirill:
I think it's also worth changing the Hostile icon to a slightly different color (or something else), everything else is ready in PR.

Content.Client/StatusIcon/StatusIconOverlay.cs
ljm862: That was my intention, yes. It's just to help minimise code duplication and improve readability, I would _hope_ the compiler would optimise away a division by 1 but we can't be sure (and I don't feel like looking at the IL code!) It's not a huge deal, I'm fine to keep it as is, just another point to be wary of :)

B-Kirill:
> I'd change either the suspicious or hostile icon for colourblind users
> 
> Don't quote me on this but there's discussion on discord whether mixels are good or bad

I think I can make a question mark for `Suspected`, that's the best I can think of. In general, I'm not that good at sprites, which explains why they look like this...

B-Kirill:
I'll wait until the end of the discussion.. Although something tells me that there is a 99% chance that the mixels will be rejected.

slarticodefast:
Yeah, I made an internal discussion thread for it and asked an art director about the mixels.
Adding the extra console status itself seems like a good addition to me though!

Mixelz:
> I'll wait until the end of the discussion.. Although something tells me that there is a 99% chance that the mixels will be rejected.

:(

Moryachok136:
I think an icon with a picture of a target or aim would be better suited for hostile status

image

B-Kirill:
> I think an icon with a picture of a target or aim would be better suited for hostile status


It's not a bad idea. In this case, there is no need to change the `Suspected` icon, let the old style remain for recognition. But the scope will definitely be without a circle, even for mixels it looks terrible.

B-Kirill:
Updated the media and got rid of the distortion of icons in comparison

LaCumbiaDelCoronavirus:
Eliminated status seems a bit redundant since it frankly makes sec's job too easy, there should still be room for error with reviving people that shouldn't be revived

Hostile status is a good addition though!

B-Kirill:
> Eliminated status seems a bit redundant since it frankly makes sec's job too easy, there should still be room for error with reviving people that shouldn't be revived

The funny thing is that this status will only be seen by those who have a special visor or glasses. Since doctors almost never have such a thing, they may still consider some kind of corpse necessary for revival.

slarticodefast (changes requested):
Hey, I got feedback from a current art director, and from a former one.
grafik
So could you undo the texture scaling changes to the status icon code and redraw the new icons in their default 8x8 size?

B-Kirill:
> Hey, I got feedback from a current art director, and from a former one. grafik So could you undo the texture scaling changes to the status icon code and redraw the new icons in their default 8x8 size?

Well, undoing the old ones will be easy, but.. I can't even theoretically imagine how I can draw a skull in 8x8 without getting a pixel mess.

slarticodefast:
The zombie status icon is a skull in 8x8 format as well, so maybe it makes sense to choose something more distinct from that as an icon.

B-Kirill:
> The zombie status icon is a skull in 8x8 format as well, so maybe it makes sense to choose something more distinct from that as an icon.

Maybe, but I still redrawn my icon and it turned out quite interesting:
hud_eliminated


Although I'm not sure about hostile one:
hud_hostile


I'll still try to come up with something, although I'd listen to your other ideas if you have any.

B1773rm4n:
**Thank you for this PR!**
The old version looks like someone just implemented the feature and said to themselves "It's a first iteration, it will do" and then it stuck forever.
Your new version is when someone just cared about to not make it look horrible. And now it's getting rejected for
"and this doesn't look great"

B-Kirill:
> Looks fine, though the skull icon seems to have extra pixels on top instead of a gap compared to other icons

Maybe you're right. I'll see later how it will look with the gap in the game, it was just necessary for the old icon.

B-Kirill:
Removed the extra pixels from the eliminated icon. I also changed the size of the icons in the PDA from 2.2 to 3.0 (they look better this way + they don't get distorted).

B1773rm4n:
@B-Kirill Did you revert the totally better looking icons and instead kept using the old icons? I was really looking forward to your change. But I see you removed the old screenshot. Can you please keep it somewhere in this thread so other people will have context?

B-Kirill:
> @B-Kirill Did you revert the totally better looking icons and instead kept using the old icons? I was really looking forward to your change. But I see you removed the old screenshot.

Yes, it's sad, but I can't argue with two art directors at once (even if one is a former one), otherwise this PR simply won't be accepted.

> Can you please keep it somewhere in this thread so other people will have context?

Okay, I can upload it in the comments.

B-Kirill:
### Comparison with rejected icon variations

image

image

Resources/Textures/Interface/Misc/security_icons.rsi/meta.json
```diff
@@ -1,7 +1,7 @@
 {
   "version": 1,
   "license": "CC-BY-SA-3.0",
-  "copyright": "Taken from https://github.com/tgstation/tgstation/blob/7f654b6e8e59021607a9e888dfeb79920401c372/icons/mob/huds/hud.dmi",
+  "copyright": "Original sprites from https://github.com/tgstation/tgstation/blob/7f654b6e8e59021607a9e888dfeb79920401c372/icons/mob/huds/hud.dmi. Extended with new status sprites by B_Kirill",
```
slarticodefast:
```suggestion
  "copyright": "Original sprites from https://github.com/tgstation/tgstation/blob/7f654b6e8e59021607a9e888dfeb79920401c372/icons/mob/huds/hud.dmi. hud_hostile and hud_eliminated by B_Kirill",
```

Content.Client/CartridgeLoader/Cartridges/WantedListUiFragment.cs
```diff
@@ -201,7 +201,7 @@ private void GenerateItem(ListData data, ListContainerButton button)
         var label = new Label() { Text = record.TargetInfo.Name };
         var rect = new TextureRect()
         {
-            TextureScale = new(2.2f),
+            TextureScale = new(3.0f),
```
slarticodefast:
```suggestion
            TextureScale = new(2.2f),
```
any reason this was changed?

B-Kirill: > any reason this was changed?

Although it is difficult to see, the icons at a scale of 2.2 are distorted and do not look very good when viewed well.. maybe 2.5 would be a good compromise?

slarticodefast: I checked both versions in-game and the original value looks better to me since the icon does not go up to the border of the bounding box. Also the change is kinda unrelated to this PR, so I'll change it back for now.

PR #36246, opened 2025-04-01
[April fools] Remove drunkenness/damage from Very Bad Day gamemode

SlamBamActionman:
## About the PR

The drunkenness effect would be fine if we had accessibility effects to reduce it, but we don't, so instead it causes 15 minutes of eyestrain. The damage is just random damage for seemingly no reason. 

## Why / Balance

Unironically gave me a headache after two rounds.

## Requirements

## Breaking changes

**Changelog**


ScarKy0 (approved):
Terrible code. Closing this immediately.

April fools!

PR #36247, opened 2025-04-01
More reasonable atmos

### LINKED ISSUE #29377: Cryogenic Gasses Aren't Harmful To Brrathe

Pronana:
## About the PR
Raised the temperature of the liquid cans from cargo to 155.15k
Raised the freezer min temperature to 155.15k
Also adjusted the cargo canister mole count to match what atmos can produce themselves.

## Why / Balance
Closes #29377 in an alternative way.

Yes, atmos making tanks of oxygen that last four times as long as a regular tank is pretty broken although to me, it's not an exploit rather an intended game mechanic, and the solution is to just choose a different number.

I chose 155.15 Kelvin because it's the criticality temperature of oxygen meaning at high pressures it is dispensed as a liquid, and a low pressure it's dispensed as a gas.

This changes the over four times multiplier of longevity to a slightly less than two times multiplier of longevity, meaning: 
An chilled emergency tank will last about 20% less than an extended tank.
An chilled extended air tank will last slightly longer than a double capacity. 
A chilled double capacity will last a bit less than a full sized tank. 
A chilled full sized tank will last a bit less than having two full sized tanks.

In addition, the hellfire freezer isn't changed so if atmos wants busted air tanks that last the whole round they will be forced to, for the first time ever, interact with another department which is probably a good thing.

This also doesn't break cryogenics because you need to be at 200k for cryogenic chems to work and the freezer min temp is well below that threshold. If anything it makes cryogenics safer to use since med can't set the temperature to 73 Kelvin for no reason.

## Technical details
Minor yaml and component value changes.

### Before:
image

### After:

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Freezers now aren't as cold


iaada:
Breathing in 0 Kelvin air freezing the shit out of your lungs feels to me like the proper solution to #29377

deltanedas:
>atmos interacting with science

because begging is thrilling gameplay and roleplay

MozarteanChaos:
> One of the cryo chems (oppo or necrosol, can't remember which) only works below 150K.
> 
> This leads to two scenarios; one is getting a hellfire freeze from sci, another is building a radiator and exposing the cryo loop to space.

It's oppo, yeah. In my opinion, either its prototype should be adjusted, or this PR should be - because making oppo even harder to use is not particularly good for it.

Partmedia:
Thanks for your PR. I have a few comments:

- The liquid can mole count was designed to be able to fill a certain number of tiles at STP. This more than halves the capacity of liquid tanks, making it more than doubly as onerous to refill the station, in the case that gas miners are removed. Can you please justify this change?
- Raising the freezer minimum temperature will make it harder to produce frezon, which might upset some folks. I do think that on balance, this is a useful change because it would provide a definite advantage of the hellfire freezer over the normal one.
- Keep in mind that atmos can still produce very cool gases using a radiator loop into space. So I don't think this actually solves the stated problems.

Pronana:
>     * The liquid can mole count was designed to be able to fill a certain number of tiles at STP. This more than halves the capacity of liquid tanks, making it more than doubly as onerous to refill the station, in the case that gas miners are removed. Can you please justify this change?

Liquid cans are more often used for filling oxygen tanks and jetpacks for breathing longer by far. If needed I can just add a prototype for a third can that has the values of the original, although I doubt it will be bought more than once every 100 games if that.

>     * Raising the freezer minimum temperature will make it harder to produce frezon, which might upset some folks. I do think that on balance, this is a useful change because it would provide a definite advantage of the hellfire freezer over the normal one.

Freezer number also messes with cryogenics currently which means either those cryogenics need their minimum temperature raised, freezer needs to be lowered further, or hellfire freezer is now a bigger sci priority.

>     * Keep in mind that atmos can still produce very cool gases using a radiator loop into space. So I don't think this actually solves the stated problems.

Yeah but it is more difficult. Which is overall a good thing.

Either way, I'm thinking of instead making a PR which is compatible with (#34682) which uses critical temperatures to set the threshold for taking cold damage rather than it being based on the lung organ. 
This would also be a very useful number to have for gasses since it's used for calculating the point at which gasses become liquid.
(i.e. when they should stop compressing/ when a gas should become a reagent/ when a reagent should become a gas)
Although that needs some discussion as well.

Partmedia:
I feel that changing the freezer and hellfire freezer to have different minimum temperature bounds is something we can still consider for gameplay balance reasons. But we would need to consider that on its own merits, not as a solution to the breathing issue.

Pronana:
> I feel that changing the freezer and hellfire freezer to have different minimum temperature bounds is something we can still consider for gameplay balance reasons. But we would need to consider that on its own merits, not as a solution to the breathing issue.

I closed it for that reason, and that there would need to be a genuinely good reason for atmos to want the hellfire freezer both of which would require changes to the atmospherics and reagents system which would just be a whole new PR. I have some ideas but who knows if I'll get to them, for now not really a point in keeping this open.

PR #28152, opened 2024-05-20
Slightly emphasize popup when someone points at you

Tayrtahn:
## About the PR

The popup for someone pointing at you is slightly larger than normal popups.

## Why / Balance

Someone pointing at _you_ is generally more important than someone pointing at something random. This makes it stand out slightly. It's pretty subtle since it's not _that_ important and it's spammable.

## Technical details

Changed the PopupType from Small to Medium when someone points at you.

## Media

https://github.com/space-wizards/space-station-14/assets/85356/7667beba-d81c-45e2-b18a-1aac5b0b52e1

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- tweak: The popup for someone pointing at you is now slightly larger than normal popups.

PR #36252, opened 2025-04-02
Improve messages for pointing at items in inventories

Tayrtahn:
## About the PR

Added new popup messages that get displayed when players point at items inside of mob inventories.

Currently when you point at an item (using the strip window) someone is holding or has equipped, you get the same message as when the item is just lying on the ground ("You point at the crowbar").

With this PR, those messages now reflect the fact that you're pointing at a specific person's item ("You point at Urist McHands' crowbar").

This works for any items contained within a mob that has an inventory, including inside stuff like a slimeperson's internal storage or a storage implant.

Thanks to https://github.com/space-wizards/space-station-14/pull/28152, any pointing targeted at items in _your_ inventory gets a slightly larger popup.

Requires https://github.com/space-wizards/RobustToolbox/pull/5803
Requires https://github.com/space-wizards/RobustToolbox/pull/5799

## Why / Balance

This change should improve interactions between players referring to inventory items. Security can, for example, point at an item in a suspect's inventory and ask them where they got it. You and point to someone's cool hat and compliment them on it. You can threaten someone and demand that they hand over a specific item. You can show off the sweet salvage loot you found and make sure everyone knows that it's yours. You get the idea.

## Technical details

Modified `PointingSystem.TryPoint` so it now searches upwards through any containers the target might be in until it finds one belonging to an entity with an `InventoryComponent`. It then expands the existing logic to pick appropriate messages to send to the pointer, the person whose item is being pointed at, and other players nearby.

Using the innermost inventory ensures that this works with nested containers - you can point at your sunglasses while inside a locker or even another mob's inventory, and the popup will still say "You point at your sunglasses" instead of thinking they belong to the locker or something.

The one oddity I didn't fully cover is the message shown to a player when they get pointed at inside a mob's inventory (like a mouse in a chef's hat). Ideally, this would just say "Urist McHands points at you", but it says "Urist McHands points at his mouse" (or similar). Fixing this would involve making some more substantial changes to `PointingSystem` and I don't think it's really worth it for a fairly obscure case - the displayed message is accurate and works well enough.

## Media

https://github.com/user-attachments/assets/2d36ee52-bceb-4ddd-b2b3-f1f6bd4a4119

## Requirements

**Changelog**

:cl:
- tweak: The popup when pointing at an item in a character's inventory now mentions that character.


Content.Server/Pointing/EntitySystems/PointingSystem.cs
```diff
@@ -208,15 +211,63 @@ bool ViewerPredicate(ICommonSession playerSession)
...
+                if (containingInventory != null)
+                {
+                    var item = pointed;
+                    var itemName = Identity.Entity(item, EntityManager);
+
+                    // Target the pointing at the item's holder
+                    pointed = containingInventory.Value;
+                    pointedName = Identity.Entity(pointed, EntityManager);
+
+                    if (player == pointed)
```
ljm862, line 237: Could extract this check out of the `if (containingInventory != null)` block instead of recomputing it here and twice in the `else`

Tayrtahn: That's a good call - it should make the code a bit more readable too. Thanks for the suggestion!

Tayrtahn: Well, except that the value of `pointed` is changed inside the if block... but I'm still going to make it a more meaningfully named variable.

Everturning:
in 13 there's a little icon next to your head like the typing indicator that shows the arrow floating up and down over the item, maybe that could be ported?

metalgearsloth (approved):

PR #36260, opened 2025-04-02
Allow sound to play at the start of anomaly supercritical animation

Quantum-cross:
## About the PR

Add datafield to AnomalyComponent to play a sound when an anomaly enters supercriticality.

The default is currently set to no sound for this field, so all anomaly previous behavior should be identical.

This PR doesn't include any audio files that take advantage of this field.

## Why / Balance

I always thought it was kind of weird that an anomaly entering a runaway state of growth and destruction was completely utterly silent.

I've helped implement a clown anomaly on a downstream server and I wanted the anomaly to play a 10 second long super stretched bike horn sound effect (absolutely terrifying) while the anomaly grew and then popped.

This was my solution, and I thought it would be nice to push it upstream.

## Technical details

Super simple, add a datafield, and just add optional AnomalyComponent argument for StartSuperCriticalEvent to read out the sound file and play it.

## Media

I can't make a video right now but can later.

## Requirements

## Breaking changes

none anticipated

no changelog

### DIFF
--- a/Content.Shared/Anomaly/Components/AnomalyComponent.cs
+++ b/Content.Shared/Anomaly/Components/AnomalyComponent.cs
@@ -129,6 +129,12 @@ public sealed partial class AnomalyComponent : Component
     /// </summary>
     [DataField]
     public SoundSpecifier? SupercriticalSound = new SoundCollectionSpecifier("Explosion");
+
+    /// <summary>
+    /// The sound plays at the start of the animation when an anomaly goes supercritical
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? SupercriticalSoundAtAnimationStart = null;
     #endregion
 
     /// <summary>
--- a/Content.Shared/Anomaly/Components/AnomalyComponent.cs
+++ b/Content.Shared/Anomaly/Components/AnomalyComponent.cs
@@ -134,7 +134,7 @@ public sealed partial class AnomalyComponent : Component
     /// The sound plays at the start of the animation when an anomaly goes supercritical
     /// </summary>
     [DataField]
-    public SoundSpecifier? SupercriticalSoundAtAnimationStart = null;
+    public SoundSpecifier? SupercriticalSoundAtAnimationStart;
     #endregion
 
     /// <summary>
--- a/Content.Shared/Anomaly/SharedAnomalySystem.cs
+++ b/Content.Shared/Anomaly/SharedAnomalySystem.cs
@@ -117,7 +117,8 @@ public abstract class SharedAnomalySystem : EntitySystem
     /// Begins the animation for going supercritical
     /// </summary>
     /// <param name="uid"></param>
-    public void StartSupercriticalEvent(EntityUid uid)
+    /// <param name="component"></param>
+    public void StartSupercriticalEvent(EntityUid uid, AnomalyComponent? component = null)
     {
         // don't restart it if it's already begun
         if (HasComp<AnomalySupercriticalComponent>(uid))
@@ -126,6 +127,9 @@ public abstract class SharedAnomalySystem : EntitySystem
         AdminLog.Add(LogType.Anomaly, LogImpact.High, $"Anomaly {ToPrettyString(uid)} began to go supercritical.");
         if (_net.IsServer)
             Log.Info($"Anomaly is going supercritical. Entity: {ToPrettyString(uid)}");
+
+        if(Resolve(uid, ref component))
+            Audio.PlayPvs(component.SupercriticalSoundAtAnimationStart, Transform(uid).Coordinates);
 
         var super = AddComp<AnomalySupercriticalComponent>(uid);
         super.EndTime = Timing.CurTime + super.SupercriticalDuration;
@@ -240,7 +244,7 @@ public abstract class SharedAnomalySystem : EntitySystem
         var newVal = component.Severity + change;
 
         if (newVal >= 1)
-            StartSupercriticalEvent(uid);
+            StartSupercriticalEvent(uid, component);
 
         component.Severity = Math.Clamp(newVal, 0, 1);
         Dirty(uid, component);
--- a/Content.Shared/Anomaly/SharedAnomalySystem.cs
+++ b/Content.Shared/Anomaly/SharedAnomalySystem.cs
@@ -116,25 +116,24 @@ public abstract class SharedAnomalySystem : EntitySystem
     /// <summary>
     /// Begins the animation for going supercritical
     /// </summary>
-    /// <param name="uid"></param>
-    /// <param name="component"></param>
-    public void StartSupercriticalEvent(EntityUid uid, AnomalyComponent? component = null)
+    /// <param name="ent"></param>
+    public void StartSupercriticalEvent(Entity<AnomalyComponent?> ent)
     {
         // don't restart it if it's already begun
-        if (HasComp<AnomalySupercriticalComponent>(uid))
+        if (HasComp<AnomalySupercriticalComponent>(ent.Owner))
             return;
 
-        AdminLog.Add(LogType.Anomaly, LogImpact.High, $"Anomaly {ToPrettyString(uid)} began to go supercritical.");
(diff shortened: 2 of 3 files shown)


ljm862:
LGTM!

Content.Shared/Anomaly/Components/AnomalyComponent.cs
```diff
@@ -129,6 +129,12 @@ public sealed partial class AnomalyComponent : Component
     /// </summary>
     [DataField]
     public SoundSpecifier? SupercriticalSound = new SoundCollectionSpecifier("Explosion");
+
+    /// <summary>
+    /// The sound plays at the start of the animation when an anomaly goes supercritical
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? SupercriticalSoundAtAnimationStart = null;
```
Tayrtahn, line 137:
```suggestion
    public SoundSpecifier? SupercriticalSoundAtAnimationStart;
```
The default value is null for a nullable field.

Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -117,7 +117,8 @@ public void RefreshPulseTimer(EntityUid uid, AnomalyComponent? component = null)
     /// Begins the animation for going supercritical
     /// </summary>
     /// <param name="uid"></param>
-    public void StartSupercriticalEvent(EntityUid uid)
+    /// <param name="component"></param>
+    public void StartSupercriticalEvent(EntityUid uid, AnomalyComponent? component = null)
```
Tayrtahn, line 121: Do this the modern, `Entity<T>` way:
```suggestion
    public void StartSupercriticalEvent(Entity<AnomalyComponent?> ent)
```
You'll need to replace `uid` with `ent` or `ent.Owner`, and `component` with `ent.Comp` in the rest of this method.

Quantum-cross:
stupid heisentest

Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -116,21 +116,24 @@ public void RefreshPulseTimer(EntityUid uid, AnomalyComponent? component = null)
...
+        if (HasComp<AnomalySupercriticalComponent>(ent.Owner))
             return;
 
-        AdminLog.Add(LogType.Anomaly, LogImpact.High, $"Anomaly {ToPrettyString(uid)} began to go supercritical.");
+        AdminLog.Add(LogType.Anomaly, LogImpact.High, $"Anomaly {ToPrettyString(ent.Owner)} began to go supercritical.");
         if (_net.IsServer)
-            Log.Info($"Anomaly is going supercritical. Entity: {ToPrettyString(uid)}");
+            Log.Info($"Anomaly is going supercritical. Entity: {ToPrettyString(ent.Owner)}");
+
+        if(ent.Comp != null)
```
slarticodefast, line 130: this should be a resolve a the top of this method
otherwise someone might be using it without passing the component, which would cause the sound not to play

Quantum-cross: yep I forgot the resolve, good catch.

Quantum-cross:
here's uhhh.... an example of how this *could* be used...

https://github.com/user-attachments/assets/9175073b-e3cc-43fc-a56b-4080884c9c81

Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -116,21 +116,26 @@ public void RefreshPulseTimer(EntityUid uid, AnomalyComponent? component = null)
     /// <summary>
     /// Begins the animation for going supercritical
     /// </summary>
-    /// <param name="uid"></param>
-    public void StartSupercriticalEvent(EntityUid uid)
+    /// <param name="ent"></param>
```
beck-thompson, line 119: Add a comment!

Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -116,21 +116,26 @@ public void RefreshPulseTimer(EntityUid uid, AnomalyComponent? component = null)
...
+    public void StartSupercriticalEvent(Entity<AnomalyComponent?> ent)
     {
         // don't restart it if it's already begun
-        if (HasComp<AnomalySupercriticalComponent>(uid))
+        if (HasComp<AnomalySupercriticalComponent>(ent.Owner))
```
beck-thompson, line 123: just do ent! No need for owner here it automatically casts

Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -116,21 +116,26 @@ public void RefreshPulseTimer(EntityUid uid, AnomalyComponent? component = null)
...
+        AdminLog.Add(LogType.Anomaly, LogImpact.High, $"Anomaly {ToPrettyString(ent.Owner)} began to go supercritical.");
         if (_net.IsServer)
-            Log.Info($"Anomaly is going supercritical. Entity: {ToPrettyString(uid)}");
+            Log.Info($"Anomaly is going supercritical. Entity: {ToPrettyString(ent.Owner)}");
+
+        if(ent.Comp != null)
+            Audio.PlayPvs(ent.Comp.SupercriticalSoundAtAnimationStart, Transform(ent.Owner).Coordinates);
 
-        var super = AddComp<AnomalySupercriticalComponent>(uid);
+        var super = AddComp<AnomalySupercriticalComponent>(ent.Owner);
```
beck-thompson, line 135: same here

Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -116,21 +116,26 @@ public void RefreshPulseTimer(EntityUid uid, AnomalyComponent? component = null)
...
         super.EndTime = Timing.CurTime + super.SupercriticalDuration;
-        Appearance.SetData(uid, AnomalyVisuals.Supercritical, true);
-        Dirty(uid, super);
+        Appearance.SetData(ent.Owner, AnomalyVisuals.Supercritical, true);
```
beck-thompson, line 137: here and the one below as well!

beck-thompson (changes requested):
super minor style stuff looks good besides that

beck-thompson (approved):
looks good!

Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -116,21 +116,26 @@ public void RefreshPulseTimer(EntityUid uid, AnomalyComponent? component = null)
...
+    /// <param name="ent">Entity to go supercritical</param>
+    public void StartSupercriticalEvent(Entity<AnomalyComponent?> ent)
     {
         // don't restart it if it's already begun
-        if (HasComp<AnomalySupercriticalComponent>(uid))
+        if (HasComp<AnomalySupercriticalComponent>(ent))
             return;
 
-        AdminLog.Add(LogType.Anomaly, LogImpact.High, $"Anomaly {ToPrettyString(uid)} began to go supercritical.");
+        Resolve(ent, ref ent.Comp, logMissing: false);
```
slarticodefast, line 126: The purpose of the resolve was so you no longer have to do the `ent.Comp != null` check
make this a guard statement at the top of the function

slarticodefast (changes requested):

Content.Shared/Anomaly/SharedAnomalySystem.cs
Quantum-cross: Before there was no explicit guard to ensuring that the entity passed in to StartSupercriticalEvent had an AnomalyComponent though. I don't think I see any usages of it that doesn't assume there is an AnomalyComponent so it's probably safe to guard that.

slarticodefast: It does not really make sense for something that is not an anomaly to go supercritical
but what you could do instead is use the resolve in place of the null check, since it already returns false in case it fails

slarticodefast (approved):
Looks good to me, I'll let beck double check again

PR #36262, opened 2025-04-02
Make certain items sometimes fall off when the wearer falls over (PoorlyAttachedComponent)

Tayrtahn:
## About the PR

Adds a component and system (`PoorlyAttached`) that simulates items not being securely attached to the wearer. These items will slowly lose their attachment strength over time and can fall off when the wearer falls over. They can be reattached by unequipping/reequipping them or by using an alt verb on the item in the wearer's inventory (either their own or another player's via the strip window), which will reset the attachment strength to the maximum and show a little popup message for RP fun.

This PR adds this component to the jamjar glasses and the dinky security star (which was the inspiration for this). The star will fall off after falling twice when first equipped, and will fall off on the first fall after 10 minutes have passed without reattaching it. The jamjar glasses take about 4 falls at first with the number slowly decreasing over one hour.

## Why / Balance

This is a fun little fluff feature. The dinky star's description ("A dinky lil star for only the hardest working security officers! It's not even sticky anymore.") implies that it's quite crappy and should be a pain to keep on. I think the idea of this sad, cheap little adhesive star sticker falling off of a sec officer when they slip and needing to be hastily retrieved is pretty funny.

The jamjar glasses also get this feature to enable funny "My glasses! I can't see without my glasses!" moments. They seem to be the "nerd glasses" of the game, so they seem like the best target. This component could be added to other glasses with increased attachment strength values, but I'm going to leave that up to contributors to figure out what's most appropriate. Adding it to sunglasses for example would have significant balance implications - it might be interesting, but I want to keep this PR low-impact and let it just be a fun feature. We can decide later if we want to do anything balance-affecting with it.

I'm sure there are some other silly items that this would be appropriate for, and I encourage contributors to make PRs for ones that they find. I'm not saying that the clown should have prank pants, but now the idea is in your head too.

## Technical details

- There's nothing too crazy, but there is a fair bit of stuff to review. I've tried to document it pretty extensively.
- The system had to be split into `SharedPoorlyAttachedSystem` with most of the functionality, and a server-only `PoorlyAttachedSystem` which handles the actual throwing.
- `FellDownEvent` was made into an `IInventoryRelayEvent` and all that so that items in the player's inventory can react to the player falling down.
- The passive attachment strength drain is implemented by marking the time in the future when the strength will reach zero, and the current value is calculated from the difference between that time and the current time, so there's no need for an accumulator or excessive network spam. Major changes to the attachment strength level (by the wearer falling) are tracked separately and are just added together in a single float DataField.
- The `GetVerbs` handler does a little trick to enable the verb in the strip window - the usual `CanAccess` check isn't done, and instead `SharedInteractionSystem.InRangeAndAccessible` is called manually using the item's wearer instead of the item itself (since the strip window user can't access the item, but we want to make sure that they can access the wearer).
- The reattach verb and popups are intentionally highly configurable through the component's DataFields - this is to make it reusable for different items. The jamjar glasses serve as an example: they change the verb label and icon, as well as almost all of the loc strings to better suit the item (you "push the glasses back up" rather than "reattach" them).
- To cut down on YAML repetition, an abstract base entity `PoorlyAttachedGlassesBase` was added which configures all the LocIds and the verb icon to suit use on a pair of glasses. The jamjar glasses inherit from their normal parent and `PoorlyAttachedGlassesBase` to automatically configure these fields, and then do additional configuration on top of that. Other uses of this component should follow this pattern when possible.

## Media

https://github.com/user-attachments/assets/f5002beb-28b2-4a35-ab40-f6ad6001e1cc

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Jamjar glasses and the dinky security star can now fall off when their wearer falls over.


Tayrtahn:
> What happend if you super glue the items while they're worn

Hm, apparently they still fly out of your inventory. I specifically used `InventorySystem.CanUnequip` so that wouldn't happen, so that's pretty annoying.

Looking into it a little bit, it looks like the issue is in `GlueSystem` - the `UnremovableComponent` doesn't get added until someone actually picks up the item. This means that if you glue an equipped item, you can still just pick it up out of your inventory (at which point it will get stuck to your hand. I'll write up an issue for that and see if we can get it fixed. Thanks for helping to find it!

PicklOH:
Because its similar to #36308, I'll chime in here too. I'm conflicted about changes that make you lose your items like this because they're annoying. The last thing I need when playing Sec is to have my glasses or radio stolen because I was trying to save the station from giant spiders and I slipped on some water. God forbid I slip in a lube puddle and my glasses may essentially be unrecoverable.

I think SS14 is a very cool simulationist RPG but personally would prefer for items staying on my charcter to stay video-gamey.

PR #36263, opened 2025-04-02
Convert AgentIDCard message to use a Fluent selector

Tayrtahn:
## About the PR

Moved the logic for selecting the correct message from C# to Fluent for the popup displayed when gaining new accesses with the Agent ID card.

## Why / Balance

Selecting between "You didn't gain any", "You gained one", and "You gained X" in the game code is bad for localization - different languages have different ways of handling pluralization, and this logic should be handled by the localization system, not the game code. Fluent has the tools to do this easily.

## Technical details

The `_popupSystem.PopupEntity` calls for the different forms were consolidated into a single call that always passes in the number of accesses changed. A [Fluent selector](https://projectfluent.org/fluent/guide/selectors.html) handles picking the right message automatically.
The code flow was modified a little bit to ensure that `Dirty` is still only called when at least one access is added.

## Requirements


beck-thompson (approved):
looks good!

PR #36266, opened 2025-04-02
Localize air alarm states

Tayrtahn:
## About the PR

The displayed names of the different air alarm states (Normal, Warning, Danger, Emagged) are now able to be localized properly.

## Why / Balance

Localization support is good.

## Technical details

The enum values were being passed into the localization system as strings (as `$"{state}"`), which was actually preventing them from being localized properly.
They are now passed through as the actual enum values, and the .ftl file now uses a Fluent selector to translate the enum values into strings, which can be changed for other localizations. A Fluent term is used to avoid having to write the same selector repeatedly.
The title displayed for each listed device now also passed into the localization system for formatting instead of doing the string formatting in the game code.

## Media
Screenshot 2025-04-02 at 10 30 58 PM

(It looks the same as before, but I assure you that those "Normal"s are localized)

## Requirements


Partmedia:
Code changes look good to me. Media would have saved me from doing an in-game test. Let me see if I can do one soon.

Tayrtahn:
> Code changes look good to me. Media would have saved me from doing an in-game test. Let me see if I can do one soon.

Sorry about that. I'll grab a screenshot.

Tayrtahn:
Well, I guess the screenshot I added doesn't really prove anything, so here's what it looks like if you change [this line](https://github.com/space-wizards/space-station-14/blob/27bab559231f112e8c3821923f6801e37097fbac/Resources/Locale/en-US/atmos/air-alarm-ui.ftl#L19) in the ftl file.
Screenshot 2025-04-02 at 10 34 42 PM

Partmedia (approved):
Thanks. I had also independently just built it locally and verify that I see nothing different :)

PR #36267, opened 2025-04-03
Fire extinguishers can now extinguish items, including when held/worn

PJB3005:
## About the PR
Fire extinguishers can now put out things like candles or cigarettes, including when held/worn.

## Why / Balance
It's funny and it's an interaction you would expect to be there.

## Technical details
So seemingly candles were already set up for being extinguished, but it never worked because collision is disabled on items via `CollisionWake`. I had to add a new component to automatically disable it when they got ignited.

For making extinguishing of held/worn items work: I made an `Extinguish` event and made it relayed. This meant moving around some of the logic, but I think the code is overall better off. I also made helpers to make relayed events easier to use.

Cigarettes work differently from candles, so they had to have extinguishing support added in the first place. Easy enough though.

Requires engine master, because when do my PRs ever not?

## Media
https://github.com/user-attachments/assets/6d893c31-bea2-4cd0-80d1-fc0f42d78e97

## Breaking changes
nada?

## Changelog

:cl:
- add: Fire extinguishers can now extinguish things like candles and cigarettes.
- add: You can even use them to extinguish items that are held/worn!


TeenSarlacc:
time to make firebots seek out people with lit cigarettes and spray them out :godo:

PR #36271, opened 2025-04-03
Diphenhydramine overdose

iaada:
## About the PR

Benadryl makes you high, but be careful.

## Why / Balance

These are things the real diphenhydramine (Benadryl) do.

Balance wise this isn't too significant. I think meds should generally have some sort of overdose for space realism, even if that overdose is unlikely to occur in typical gameplay. This is one of those instances where the overdose is tuned more towards RP than balance. If you're desperate to get high you can steal chems from medbay, but you're pretty unlikely to ever actually OD on this.

I admit the threshold was chosen arbitrarily, but that seems to be how a lot of OD thresholds are decided. The recommended dose of Benadryl is 2 pills (50ml), and so the recommended dose in-game is 2 pills.

## Technical details
5 seconds of high when above 30u is a very weak high, equivalent to what you can get much easier from a mushroom.

Bloodloss was chose due to the real world overdose effects (heart palpitations). A value of 5 is a lot, but still not incredible due to the threshold of 40u and the relative weakness of the damage type.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl: aada
- add: Overdose added to Diphenhydramine.


TeenSarlacc:
is this a lead up to the hat man as a ghostrole?

iaada:
> Why bloodloss? I couldn't find anything related to that as an OD effect.

Bloodloss is meant to mimic heart palpitations cutting off oxygen to the brain. An effect like that isn't really possible so bloodloss was the closest thing. It's also not a damage common in reagents, making it at least a tiny bit interesting.

> I would prefer it with no blood loss but hallucination + negative interaction with alcohol and lower OD (somewhere between 10 and 20) so it can happen.

Having a negative interaction with alcohol isn't exactly unique to dip. But that aside, checking if a player is drunk is a new effect which would require a lot more work. And checking if there's ethanol in the blood would work, but that's not the same thing as checking if they're drunk.

> Gives medbay some reason to ask the patient what they ingested so they can decide if dylo or diph is the better option. (imagine having actual conversation with patients and not just injecting according to the scan)

I feel like doctors not talking to patients is more a consequence of their patients showing up dead most of the time. But I agree that information which can only come from asking a question is a good springboard for RP. Dip is also a good choice for it on account of the poison synergy with alcohol.

deltanedas:
ethyl exists for that

iaada:
I believe the idea ia ethy makes you no longer drunk so you can use dip.

MozarteanChaos:
> I believe the idea ia ethy makes you no longer drunk so you can use dip, presupposing that dip has a negative interaction with alcohol.

As far as I'm aware, ethyl doesn't clear ethanol from the bloodstream - it just removes the accumulated drunkenness. Unless diph is somehow coded to react to patients experiencing drunkenness (which would technically mean it reacts with more than just alcohol - a lot of chems cause drunkenness on OD) instead of reacting with ethanol directly, ethyl will not prevent the reaction.

iaada:
We're getting into the weeds of a theoretical effect that's not a part of this PR. If ethanol in the blood is the trigger charcoal exists for that, and if the trigger is the drunk status ethy can be used.

I'd personally be more interested in seeing the later as it's both new and more accurate (you stay dead drunk long after the ethanol is gone). I do think dip having a negative reaction to alcohol is interesting from the cause-poison cure-poison dynamic. But it's not a part of this PR.

Huaqas:
The appeal of Diphenhydramine is that it is a stronger version of Dylovene with no OD, but is infinitely more complex to create. It's lack of OD is balanced out by how hard it is to create. Giving it an OD will just cause people to use Stellibinin instead. Why put all this effort in making Diph, when you can just get botany to grow 12 pieces of Galaxythistle, and you'll have the same amount of Diph, that is stronger and has no OD.

Once this starts happening, someone will make a PR add an OD to Stellibinin as well. Repeat until every chemical has an OD, and medbay becomes a nightmare to teach.

Also someone in the recent update added drowsiness as a side effect of Diphenhydramine. Giving it an OD on top of that will make it an abysmal medicine to use.

iaada:
The appeal of Dip is that it's a stronger version of Dylo, not that it has no OD. Having an overdose doesn't make medbay harder to teach as it's completely intuitive that too much Benadryl is harmful to the body. you'd need _40 units_ to have a meaningfully negative effect which is 240 poison damage. It's not gonna happen.

And you're completely right. People _should_ be using Stellibinin more. That shit grows outta the ground it's free. And I'm of the opinion that botanical chems don't need ODs so I won't be the one to PR that at least.

iaada:
Test failed by the smuggler.

iaada:
Going to close this. I still like the ide, but there are discussions for a more complete vision for rebalancing advance chems. My thoughts with this PR were never very deep.

Issue #34699, opened 2025-01-28
Blink + Dash need combining


metalgearsloth:
Generic teleport mob action except done 2 completely separate ways.

ljm862:
I'll take it!

Issue #34597, opened 2025-01-23
MagicSystem needs rewriting from the ground up


metalgearsloth:
It goes against ECS and hardcodes in every spell as its own codepath, similar to old code like stunbatonsystem or old nukies.

Specific examples:
1. Speech is hardcoded onto every class when it should just be a single line of yaml and its own event.
2. PreventCollideComponent needs removing.
3. ChargeSpellEvent needs removing, it just poorly wraps around guns.
4. Same with ProjectileSpellEvent.
5. Remove RandomGlobalSpawnSpellEvent (see below).

We should get to the point where we have the following:
- An easy way to supply targets for a spell, like other games with generic magic systems like skyrim or noita. Same with checking LOS or not.
- Generic spawning that handles between <PlaceAt / spawninhands> via yaml
- An admin should be able to make completely different spells in a round with yaml without needing to pull out an IDE.
- Effects should be generic. Look at MindSwapSpell for example. It hardcodes mind transfer and paralyze. Paralyze does not need to be hardcoded into mindswap and can be pulled out entirely.
- If an event is doing more than 1 thing / copy-pasting it's doing it wrong.

keronshb:
This is my plan for Actions/Magic after Wiz releases.

Post Wizard release:
- Bugfixing
- Toggle Action for temporary weapons (Ling Armblade, Touch Spells)
- MMO/Targeting System (mostly for field spells)
-- Timestop (which means paused tiles, or tiles that pause)
- Strip Magic Component down and give a lot of the backend to Actions instead (there's already a PreActionEvent or some sorta condition) which will also be separate components probably (depends how things go after your charge PR gets merged eventually)
- Give actions doafter support as well
- Check around with how events are done and maybe try moving the Wizard events to that

A lot of this is planned to be done in piecemeal as well. Like moving Emote/Speech/Requirements support over to actions and as potential different components on the actions (like how you're doing with charges) so we can get actions in a much better place.

Then magiccomp can be gutted after.

ocan4214:
How is the progress on this ?

keronshb:
> How is the progress on this ?

In progress

keronshb:
Most of this was implemented (new Entity Effect Refactor, Charge was refactored, etc).
The generalizing via YAML was overridden by LMs and is out of scope.

ThatGuyUSA:
Unfreeze magic? #8524

keronshb:
[...]

PR #36282, opened 2025-04-03
Combine wizard blink and ninja dash

ljm862:
## About the PR

Combines the ninja dash and the wizard blink actions to use a TeleportationAction for the actual teleporting logic. Addresses #34699

The Dash and the Blink still have their original WorldAction components and events attached. This allows them to run their own personal logic not related to teleportation such as Speak and checking the katana is equipped. 
I could have extracted this as well into the use of the new event, but I think that would have required a bigger refactor on the Spell system and that's a bit beyond the scope here!

Also removes 1 warning in the process!

## Why / Balance

Refactors slightly the magic system as a first step in #34597 by knocking out a quick win in extracting the teleportation logic and combining it with the Ninja Dash logic.

## Technical details

Creates a new `TeleportActionEvent` which takes in an entity and a location. On invocation the teleport is attempted by following the logic from the Ninja dash. It reads from the new `TeleportActionComponent` whether or not it should drop whatever it is pulling on teleport. Wizard should take it with them and ninja should drop, to match previous functionality. 
The component could easily be extended to handle other things specific to teleportation in the future.

## Media

https://github.com/user-attachments/assets/c5cf353a-4f2f-40cb-ac09-e2801748a9ee

## Requirements

## Breaking changes

Things that cause teleportation (Wizard Blink, Ninja Dash) require a `- type: TeleportationAction`

**Changelog**

Nothing _should_ change!


Content.Shared/Magic/SharedMagicSystem.cs
```diff
@@ -318,12 +319,9 @@ private void OnTeleportSpell(TeleportSpellEvent args)
         if (args.Handled || !PassesSpellPrerequisites(args.Action, args.Performer))
             return;
 
-        var transform = Transform(args.Performer);
-        if (transform.MapID != _transform.GetMapId(args.Target) || !_interaction.InRangeUnobstructed(args.Performer, args.Target, range: 1000F, collisionMask: CollisionGroup.Opaque, popup: true))
-            return;
+        var ev = new TeleportActionEvent(args.Performer, args.Target);
+        RaiseLocalEvent(args.Action, ref ev);
```
metalgearsloth, line 323: 2 things:
1. Is that this is a method event which is bad, it should just be doing method calls anyway.
2. The teleport itself is still on wizard and ninja so this pr is like 1/3 of the way to where the issue needs it to be.

Content.Shared/Teleportation/Systems/SharedTeleportationActionSystem.cs
```diff
@@ -0,0 +1,74 @@
...
+using Content.Shared.Interaction;
+using Content.Shared.Movement.Pulling.Components;
+using Content.Shared.Movement.Pulling.Systems;
+using Content.Shared.Popups;
+using Content.Shared.Teleportation.Components;
+
+namespace Content.Shared.Teleportation.Systems;
+
+public sealed class SharedTeleportationActionSystem : EntitySystem
+{
```
metalgearsloth, line 13: Don't put shared as prefix on system names.

Content.Shared/Teleportation/Systems/SharedTeleportationActionSystem.cs
```diff
@@ -0,0 +1,74 @@
...
+        if (ent.Comp.DropsPulled)
+        {
+            // Check if the user is pulling anything, and drop it if so
+            if (TryComp<PullerComponent>(user, out var puller) && TryComp<PullableComponent>(puller.Pulling, out var pullable))
+                _pullingSystem.TryStopPull(puller.Pulling.Value, pullable);
+        }
+
+        _transform.SetCoordinates(user, transform, args.Target);
+        _transform.AttachToGridOrMap(user, transform);
+
```
metalgearsloth, line 71: This is 2 moveevents when it should be one.

Content.Shared/Teleportation/Systems/SharedTeleportationActionSystem.cs
ljm862: Why not here? I thought all Systems in Content.Shared should be named that way?

Content.Shared/Magic/SharedMagicSystem.cs
ljm862: Got it! Yes it's something I came up against when writing this. I was being slightly stopped by the Speak system so hopefully when that's merged I'll be able to revisit this and properly decouple teleporting from the wizard and the ninja.

Content.Shared/Teleportation/Systems/SharedTeleportationActionSystem.cs
Tayrtahn: We only have the "Shared" prefix when the system has server and client classes that inherit from it. If everything is in Content.Shared, then don't prefix the system name.

TheShuEd:

PR #36288, opened 2025-04-04
Add a reduced motion alternative to the FlashedEffect

Quantum-cross:
## About the PR

Add a reduced motion alternative to the FlashedEffect for those sensitive to visual motion.

The most disorienting thing about the flash effect is that when it returns to normal, the static afterimage blends with the real moving image across the entire display.

This alternative gives the player something to clearly focus their eyes on throughout the duration of the effect while still blocking almost all visual information about the game.

## Why / Balance

Accessibility feature

## Technical details

First, reduces the spinning of the after images
Second, instead of fading the entire screen back to normal homogeneously, start with a tiny circle of clear image in the center that grows rapidly at the end of the effect to wipe away the after image.

## Media

https://github.com/user-attachments/assets/87f9b7dc-f7b2-46e7-8154-7b52791a011a

## Requirements

## Breaking changes

none

**Changelog**

:cl:
- add: Add reduced motion alternative to the flash effect


Pyrizhok333x2:
This looks interesting. However, I don't fully understand why we would add such a setting.

Quantum-cross:
> However, I don't fully understand why we would add such a setting.

Also to be clear, the setting "Reduce motion of visual effects" already exists. This is just implementing the setting for a visual effect with some of the most motion in the game.

Content.Client/Flash/FlashOverlay.cs
```diff
@@ -74,8 +79,16 @@ protected override void Draw(in OverlayDrawArgs args)
...
+            if (_configManager.GetCVar(CCVars.ReducedMotion))
+            {
+                _reducedShader.SetParameter("percentComplete", PercentComplete);
+                worldHandle.UseShader(_reducedShader);
+            }
+            else
+            {
+                _shader.SetParameter("percentComplete", PercentComplete);
+                worldHandle.UseShader(_shader);
+            }
```
ljm862, line 91: If this is the only instance of `_shader` in the file, can you do the `if(_configManager.GetCVar(CCVars.ReducedMotion)` check in the constructor and not have the branch every time we call the Draw?

Something like 
```
_shader = _configManager.GetCVar(CCVars.ReducedMotion) ? 
_prototypeManager.Index<ShaderPrototype>("FlashedEffectAccessible").InstanceUnique() : 
_prototypeManager.Index<ShaderPrototype>("FlashedEffect").InstanceUnique()
```

Wdyt?

Quantum-cross: I considered that, I would want it to change immediately with the setting change. Alternatively, I might be able to add shader parameters to support both modes.

Mixelz:
>I'd argue that the non-accessible flash remains more advantageous as the player can begin to see the peripheral of their screen sooner than a player using the reduced motion variant.
I think that it's not exactly equivalent, since you can still see a small area around the player where you cannot with the setting off. For any tider looking to steal something after being flashed this would be all they need to be able to do so with relative consistency. 

Would there be a way to blank out the area surrounding the player until the effect begins to wear off? Meaning the only object of focus would be the player sprite itself without any surrounding detail.

Content.Client/Flash/FlashOverlay.cs
ljm862: In that case, Subs is probably the best option!
In `FlashSystem` you can add 
`Subs.CVar(_configManager, CCVars.ReducedMotion, OnReducedMotionCVarChanged);` in the Initialize, it'll listen for when the change has been made. Then add a new method there called `OnReducedMotionCVarChanged`

```
    private void OnReducedMotionCVarChanged(bool isSet)
    {
        _overlay.SetReducedMotion(isSet);
    }
```

Then back in your overlay we now have a 
```
        public void ChangeReducedMotion(bool isSet)
        {
            _shader = _configManager.GetCVar(CCVars.ReducedMotion) ?
                _prototypeManager.Index<ShaderPrototype>("FlashedEffectAccessible").InstanceUnique() :
                _prototypeManager.Index<ShaderPrototype>("FlashedEffect").InstanceUnique();
        }
```
Which you can even call in the constructor for `FlashOverlay`. The only difference now being you'd need to set `_shader` to not be `readonly`!

If I'm correct, it'll function the same as you've done it :)

Quantum-cross:
> > I'd argue that the non-accessible flash remains more advantageous as the player can begin to see the peripheral of their screen sooner than a player using the reduced motion variant.
> 
> I think that it's not exactly equivalent, since you can still see a small area around the player where you cannot with the setting off. For any tider looking to steal something after being flashed this would be all they need to be able to do so with relative consistency.
> 
> Would there be a way to blank out the area surrounding the player until the effect begins to wear off? Meaning the only object of focus would be the player sprite itself without any surrounding detail.

I agree, I'll look into it. Probably a black spot to focus on that fades into the circle mask which then expands

Quantum-cross:
> I think that it's not exactly equivalent, since you can still see a small area around the player where you cannot with the setting off. For any tider looking to steal something after being flashed this would be all they need to be able to do so with relative consistency.
> 
> Would there be a way to blank out the area surrounding the player until the effect begins to wear off? Meaning the only object of focus would be the player sprite itself without any surrounding detail.

What do you think of this?

https://github.com/user-attachments/assets/8f1c729c-c734-47d5-9e2f-dd6008720fc6

Mixelz:
Looks great!

Quantum-cross:
Another version with slightly less movement, avoids the snapping of the separation of the two afterimages, now they separate over the duration of the effect.

https://github.com/user-attachments/assets/2ae6ca84-6bb4-4f47-a57d-46a13bce2891

keronshb:
Accessibility or not this options makes it a must-use. This completely negates flash use.

Quantum-cross:
> Accessibility or not this options makes it a must-use. This completely negates flash use.

Huh? It still completely blocks the screen.

walksanatora:
yeah but there is no spinning copies. so you still always can right-click and interact with stuff
even if flashed. making it ALOT easier to intereact while flashed. honestly just blacking the whole screen would be a good idea.

Quantum-cross:
I think the current iteration of the effect is in a good place. Before I make any more changes, I want to take a step back and think about the design principles of the flash.

Consider you as the user of the flash, what do you expect to happen to your victim?

1. The user is denied any new visual information about the events surrounding them for about 90% of the duration of the effect.
2. The user's movement speed is slowed for the duration of the effect.

This change only impacts number 1.

I believe that this variant of the effect still fulfills the promises of number 1.

------------

> yeah but there is no spinning copies. so you still always can right-click and interact with stuff even if flashed. making it ALOT easier to intereact while flashed. honestly just blacking the whole screen would be a good idea.

I personally believe that the gap between the floor and ceiling for this particular skill is very small. The accuracy of which you need to right click an object to get it to show up in the context menu is pretty low -- by which I mean you do not need to be very accurate.

--------------

I would like to get some feedback from higher up in the chain (would that be project manager or something?) before I make any more balance specific changes to this PR.

Quantum-cross:
> I would like to get some feedback from higher up in the chain (would that be project manager or something?) before I make any more balance specific changes to this PR.

As soon as I say that I realize that for longer flash durations like flashbangs the accessible version had a huge advantage fading in from black instead of fading from the ghost textures.

I've brought them to much closer parity by using side by side comparisons (two clients open).

I'm using the pizza as a reference for "new information."

https://github.com/user-attachments/assets/87f9b7dc-f7b2-46e7-8154-7b52791a011a

SlamBamActionman:
I need to play around with it ingame myself before I can give any official approval, but based on the videos I think this seems like a great way to handle the effect while still keeping it on par gameplay-wise.

Is it less disorienting to the player's senses? Yes, and that's the point. We shouldn't balance the game around features that cause IRL strain (I find that to be unethical design). The pizza video example demonstrates really well that it blocks out information same as the old effect. The effect is maybe not as aesthetically pleasing but that's a perfectly fine compromise. 

I may consider the doubling-effect still present to be disorienting but I'll need to test ingame first.

slarticodefast:
Wouldn't it make more sense to move this into a separate shader instead of adding an if condition to the current one?

keronshb:
> > I would like to get some feedback from higher up in the chain (would that be project manager or something?) before I make any more balance specific changes to this PR.
> 
> As soon as I say that I realize that for longer flash durations like flashbangs the accessible version had a huge advantage fading in from black instead of fading from the ghost textures.
> 
> I've brought them to much closer parity by using side by side comparisons (two clients open).
> 
> I'm using the pizza as a reference for "new information."
> flash_compare.mp4

It shouldn't only be hard to pick up/see an item, but to figure out _where_ you and others are as well.

As it is right now, there's absolutely no reason for someone to use the default option since the accessible option gives way too much of an advantage.

We might want to pull the trigger on this one and require adding confusion movement (random input while moving) and a little slowdown, otherwise the accessible option will be the powergaming option.

SlamBamActionman:
> It shouldn't only be hard to pick up/see an item, but to figure out _where_ you and others are as well.

As seen in the pizza example, the screen freezes at the moment of impact. So yes, you will have a comparatively clearer view of how everyone was positioned *as* the flash went off... but I think that positioning is still clear in the existing flash effect, and in any combat situation it'd still result in you swinging blindly in the direction the person flashing you was in.

Quantum-cross:
Powergamers will always powergame. Some people will always potato their fps games to get the highest frame rate and least visual effects, turn off all views Bob and motion blues, turn off all realistic lighting effects.

In the end, I still think only a miniscule amount of players would take advantage of it, and I've made to minimize the advantage that would be gained. I believe it is extremely minimal.

It's not my call, but I'll adjust it further based on maintainer feedback at this point.

keronshb:
> It's not my call, but I'll adjust it further based on maintainer feedback at this point.

I still say to make up for it, some sort of confused input component should be added while flashed.

BlueMoon059:
Looks like a great addition. Now I wont have to go lie down after getting a flash used on me.

Excellent work.

Powergamers will always do whatever they can to get an advantage, but if people are being shitters in their powergaming that can be handled via admin means. If we wanted to restrict any QoL features just because someone _might_ take advantage of it, there's a lot of things we'd need to remove (uncapped fps, custom keybinds etc).

Accessibility shouldn't come second to game balance.

> > 
> 
> I still say to make up for it, some sort of confused input component should be added while flashed.

Main issue with this is that a lot of the people who've complained about the flash effect comes from a place of vertigo or motion sickness. Changing someone's movement suddenly would..Also result in vertigo or motion sickness being triggered.

I'm not sure why we need to balance an accessibility feature. Other games have colourblind modes and the like, and might provide an advantage due to colour contrast difference, but we should still seek to accomidate. I'm confused as to why we need to "make up" for anything.

keronshb:
> Main issue with this is that a lot of the people who've complained about the flash effect comes from a place of vertigo or motion sickness. Changing someone's movement suddenly would..Also result in vertigo or motion sickness being triggered.

As someone with motion sickness issues, inverted motion never triggers that on games. It's just input, it's not like the camera is going to start doing that to you.

Quantum-cross:
> Wouldn't it make more sense to move this into a separate shader instead of adding an if condition to the current one?

I forgot to comment on this one -- but there is some shared code for creating the after images, so I thought I'd use a uniform bool to switch the behaviors.

If it is a separate shader, then I'd need to either have two ShaderInstance members in flashOverlay or dynamically switch between them when the client Cvar changes. Which I COULD do if you really want?

I figured with the shared code in the shader itself this is the best way to do it.

From what I've learned about fragment shaders, they recompile automatically if any uniform parameter changes, I noticed they recompile when the resolution changes, etc. The compiler will be smart enough to compile out the branch statement controlled by a single uniform bool.

NoahCHeaverin:
My gut says the effect will look slightly nicer if the black spot is white instead. I can't say for certain until i see it in action though.

Errant-4 (approved):
I like it, thank you for your contribution

slarticodefast (dismissed):
Instead of having a giant if/else in the shader this should be a seperate shader file and selected in C#

Quantum-cross:
> Instead of having a giant if/else in the shader this should be a seperate shader file and selected in C#

Sure I can do it that way, I figured there was enough shared code to consider combining it but I'll do it that way.

DandiIion:
Hero. Admittedly little odd looking but anything is better than the vomit-inducer-2000 version it was before.

Content.Client/Flash/FlashOverlay.cs
```diff
@@ -15,23 +17,42 @@ public sealed class FlashOverlay : Overlay
...
         private readonly SharedFlashSystem _flash;
         private readonly StatusEffectsSystem _statusSys;
 
         public override OverlaySpace Space => OverlaySpace.WorldSpace;
-        private readonly ShaderInstance _shader;
+        private ShaderInstance _shader;
         public float PercentComplete = 0.0f;
         public Texture? ScreenshotTexture;
 
+        [ValidatePrototypeId<ShaderPrototype>]
```
slarticodefast, line 30: This attribute is deprecated. Use a ProtoId<ShaderPrototype> instead of a string.

Quantum-cross: here I thought I was being clever trying something new

keronshb:
I think for now it looks fine, I would want to add a lasting confused motion to this and maybe a knockdown once crawling is figured out. But that's to be atomized and in a different PR.

keronshb (approved):

keronshb (changes requested):
Actually wait sorry.  I spoke too soon.
Remove the older flash effect. Zero reason to have it since everyone's going to be using the new one.

Errant-4:
Idunno about that, the older looks nicer. I am not going to switch to reduced motion for this

Quantum-cross:
> Actually wait sorry. I spoke too soon. Remove the older flash effect. Zero reason to have it since everyone's going to be using the new one.

Have you actually tested it ingame?

Fildrance:
Lowering priorotity as another option for accessibility was merged. That does not mean we dont need more nicely/neatly looking effect, but thats not P1 anymore.

AJCM-git:
In maints discussions it was decided that two completely different shaders for this is bad for maintainability. The black circle in the alternative shader is potentially problematic as well. Closing the PR, though an argument could be made to replace the main shader effect with this one, this can be reopened if thats done for further deliberation.

Issue #5188, opened 2021-11-06
Tiles megaissue


AJCM-git:
## Description
 These are the big four and most important features, once they are done crayons, carpets and catwalks should be moved from entities to tiles
  - [ ] Support for decals (https://github.com/space-wizards/space-station-14/pull/5183)
  - [ ] Support for sprite transparency
  - [ ] Support for sprite rotation
  - [ ] Support for tile smoothing
---
Some important cleanup and QoL that would make everything tidier
  - [ ] Support for parenting in the tiles prototype
  - [ ] Tiles prototype use `name` as their ID and `display_name` as their name. This is very confusing and not in line with how the rest of the codebase handles this, so changing `name` to `id` and `display_name` to name is necessary
  - [ ] Tiles prototype `name` (IDs) don't follow current rules. They use snake_case instead of PascalCase
  - [ ] The tile spawner window should use space as an eraser button. Like how the entity spawner has "erase" at the button.
  - [ ] Support for RSI with the correct copyright and licenses
  - [ ] Cleanup obsolete tiles like underplating

Peptide90:
Commenting to bring attention to this as I had a small rant about it today.

Now that we have the decal placer which has support for palettes, I feel adding a category drop down to the tile and entity spawner would be very helpful. Have it default to an all category in alphabetical order but being able to change to tile sets for things like shuttles, outdoors, dark, light, maint etc etc.

IProduceWidgets:
Issue kinda all over the place. I'm electing to close it as the main things have been resolved / are not planned wrt crayons carpets and catwalks.

PR #36290, opened 2025-04-04
Tilestack system (aka tile smoothing)

Killerqu00:
## About the PR
A base system for supporting tilestacks - non-standard tile configurations (like wood floor on steel floor in this example, not actually part of the PR)
it's not yet actually in the game, this PR only introduces the tech itself - there are different problems with tiles that need to be fixed before implementing this fully, such as tile parenting and stuff
i might have missed something, so please think of any problems that might arise

## Why / Balance
it cool

## Technical details
every grid gets a TileStackMapComponent, which stores tilestacks. Tilestacks are lists of strings which are ContentTileDefinion (we cannot replace those directly since ContentTileDefinion cannot be serialized)
TilestackSystem handles some basic functions like creating and checking for tilestacks, as well as actually giving all grids the TileStackMapComponent
ContentTileDefinition also got a field called PossibleBaseTurfs - you can place a tile BOTH on BaseTurf and PossibleBaseTurfs. Placing it on PossibleBaseTurfs creates the tilestack
then we add some additional logic to TileSystem, FloorTileSystem and ExplosionSystem... and waow floor system

## Media
https://github.com/user-attachments/assets/082da01c-4bfe-448b-9b23-1e8b8589d4ff

## Requirements

## Breaking changes
Tilestack support (aka tile smoothing) has been added. This comes with some changes to TileSystem.DeconstructTile, FloorTileSystem.OnAfterInteract and ExplosionSystem.Processing.DamageFloorTile.

**Changelog**
not player facing

### DIFF
--- a/Content.Shared/Maps/ContentTileDefinition.cs
+++ b/Content.Shared/Maps/ContentTileDefinition.cs
@@ -44,6 +44,13 @@ namespace Content.Shared.Maps
         [DataField("baseTurf")]
         public string BaseTurf { get; private set; } = string.Empty;
 
+        /// <summary>
+        ///     A list of all the possible base turfs. doesn't have to contain <see cref="BaseTurf"/>.
+        ///     You can place a tile on BaseTurf and all the tiles listed here; if it's not placed on BaseTurf, a tilestack is created in <see cref="TileStackMapComponent"/>.
+        /// </summary>
+        [DataField]
+        public string[] PossibleBaseTurfs = [];
+
         [DataField]
         public PrototypeFlags<ToolQualityPrototype> DeconstructTools { get; set; } = new();
 
--- a/Content.Shared/Maps/ContentTileDefinition.cs
+++ b/Content.Shared/Maps/ContentTileDefinition.cs
@@ -49,7 +49,7 @@ namespace Content.Shared.Maps
         ///     You can place a tile on BaseTurf and all the tiles listed here; if it's not placed on BaseTurf, a tilestack is created in <see cref="TileStackMapComponent"/>.
         /// </summary>
         [DataField]
-        public string[] PossibleBaseTurfs = [];
+        public HashSet<string> PossibleBaseTurfs = [];
 
         [DataField]
         public PrototypeFlags<ToolQualityPrototype> DeconstructTools { get; set; } = new();
--- a/Content.Shared/Tiles/FloorTileSystem.cs
+++ b/Content.Shared/Tiles/FloorTileSystem.cs
@@ -44,6 +44,18 @@ public sealed class FloorTileSystem : EntitySystem
     {
         base.Initialize();
         SubscribeLocalEvent<FloorTileComponent, AfterInteractEvent>(OnAfterInteract);
+        SubscribeLocalEvent<GridInitializeEvent>(OnGridInit);
+        SubscribeLocalEvent<GridRemovalEvent>(OnGridRemoved);
+    }
+
+    private void OnGridInit(GridInitializeEvent ev)
+    {
+        EnsureComp<TileStackMapComponent>(ev.EntityUid);
+    }
+
+    private void OnGridRemoved(GridRemovalEvent ev)
+    {
+        RemComp<TileStackMapComponent>(ev.EntityUid);
     }
 
     private void OnAfterInteract(EntityUid uid, FloorTileComponent component, AfterInteractEvent args)
@@ -170,6 +182,11 @@ public sealed class FloorTileSystem : EntitySystem
         return tileDef.BaseTurf == baseTurf;
     }
 
+    public bool HasPossibleBaseTurf(ContentTileDefinition tileDef, string baseTurf)
+    {
+        return HasBaseTurf(tileDef, baseTurf) || tileDef.PossibleBaseTurfs.Contains(baseTurf);
+    }
+
     private void PlaceAt(EntityUid user, EntityUid gridUid, MapGridComponent mapGrid, EntityCoordinates location,
         ushort tileId, SoundSpecifier placeSound, float offset = 0)
     {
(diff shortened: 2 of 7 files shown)


ScarKy0:
Few questions
Does it support placing tiles on top of planet tiles?
Is there a limit to how many can be stacked?

Killerqu00:
> Few questions
> Does it support placing tiles on top of planet tiles?
> Is there a limit to how many can be stacked?
> Is mapping several tiles on top of each other possible?

1. if you add it to yml yes
2. no, but robustly written yml will prevent tile doomstacking
3. not yet, will need different PR

ScarKy0:
Rimworld admeme players will love you for this, as well as planet forks

TheShuEd:
> 2. no, but robustly written yml will prevent tile doomstacking

i think we 100% need integration test for it or some other way to preventing doomstacking before merg

TheShuEd:
Ok, I'm interested in this PR, and will look into checking it out shortly

metalgearsloth:
Shouldn't this be on engine?

Killerqu00:
> Shouldn't this be on engine?

ContentTileDefinition is in, well, content

NikitosAseev:
What about pipes, wires, and other things placed under the tiles? Especially, what if a smuggler's bag were placed under each tile?

Killerqu00:
> What about pipes, wires, and other things placed under the tiles? Especially, what if a smuggler's bag were placed under each tile?

Seems to work as intended with the example showed in the media. Pipes and satchels show right under all the floors when plating is exposed.

Winkarst-cpu:
Related to the #5188.

FairlySadPanda:
image

*slaps maintainer*

metalgearsloth:
> > Shouldn't this be on engine?
> 
> ContentTileDefinition is in, well, content next change I'll need is tile parenting in order not to copypaste all gazillion possibilities, this will go to the engine since tile prototypes are engine-side, I think _maybe_ also mapping tools would go to engine, unsure

The issue is you have no guarantee where SetTile is called on content, or handling for stuff like dungeons for example where you would definitely want tile layering.

Content.Shared/Tiles/FloorTileSystem.cs
```diff
@@ -145,6 +147,21 @@ private void OnAfterInteract(EntityUid uid, FloorTileComponent component, AfterI
...
+                    }
+                    if (!_stackSystem.Use(uid, 1, stack))
+                        continue;
+                    // current tile gets stored in the tilestack under the new tile
+                    ref var tilestacks = ref Comp<TileStackMapComponent>(gridUid).Data;
+                    tilestacks[tile.GridIndices].Add(tile.GetContentTileDefinition().ID);
+                    PlaceAt(args.User, gridUid, mapGrid, location, currentTileDefinition.TileId, component.PlaceTileSound);
+                    args.Handled = true;
+                    return;
+                }
```
FairlySadPanda, line 164: Don't add to this over-bloated function with yet another if. This function is now well over 100 lines long and is a nested nightmare.

Content.Shared/Tiles/FloorTileSystem.cs
```diff
@@ -37,6 +37,7 @@ public sealed class FloorTileSystem : EntitySystem
     [Dependency] private readonly TileSystem _tile = default!;
     [Dependency] private readonly SharedPhysicsSystem _physics = default!;
     [Dependency] private readonly SharedMapSystem _map = default!;
+    [Dependency] private readonly TileStackSystem _tileStack = default!;
```
FairlySadPanda, line 40: Tight coupling. TILE stacks, Floor TILE System.

See Door bolts (currently) and my partialization of Airlock and Firelock (PR) as examples of partialization. If a system is an orphan without its parent system and is just there to extend and customize an underlying system, it should be a partial.

FairlySadPanda (commented):
Have covered code quality and domain structure, I'm assuming this is in the right place. Listen to Sloth et al - if this feels like it should be engine code, and it's an RT maintainer saying it, they are probably right.

Content.Shared/Tiles/FloorTileSystem.cs
FairlySadPanda: (This isn't your fault but I would show-stopper this PR if I was responsible for merging for this reason alone)

Killerqu00:
> > > Shouldn't this be on engine?
> > 
> > ContentTileDefinition is in, well, content next change I'll need is tile parenting in order not to copypaste all gazillion possibilities, this will go to the engine since tile prototypes are engine-side, I think _maybe_ also mapping tools would go to engine, unsure
> 
> The issue is you have no guarantee where SetTile is called on content, or handling for stuff like dungeons for example where you would definitely want tile layering.

i'll see what i can do with the same structure engine-side

Content.Shared/Maps/ContentTileDefinition.cs
```diff
@@ -44,6 +44,13 @@ public sealed partial class ContentTileDefinition : IPrototype, IInheritingProto
         [DataField("baseTurf")]
         public string BaseTurf { get; private set; } = string.Empty;
 
+        /// <summary>
+        ///     A list of all the possible base turfs. doesn't have to contain <see cref="BaseTurf"/>.
+        ///     You can place a tile on BaseTurf and all the tiles listed here; if it's not placed on BaseTurf, a tilestack is created in <see cref="TileStackMapComponent"/>.
+        /// </summary>
+        [DataField]
+        public string[] PossibleBaseTurfs = [];
```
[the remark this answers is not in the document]
Killerqu00: The reason why I use string here instead of ProtoId is because BaseTurf also uses string. It's definitely bad and I could fix this, but it would proooooobably require a different pr? i am not sure

Killerqu00:
I'll just work on the current implementation, because I am not capable of moving this to engine

Content.Shared/Maps/ContentTileDefinition.cs
```diff
@@ -49,7 +49,7 @@ public sealed partial class ContentTileDefinition : IPrototype, IInheritingProto
         ///     You can place a tile on BaseTurf and all the tiles listed here; if it's not placed on BaseTurf, a tilestack is created in <see cref="TileStackMapComponent"/>.
         /// </summary>
         [DataField]
-        public string[] PossibleBaseTurfs = [];
+        public HashSet<string> PossibleBaseTurfs = [];
```
TheShuEd, line 52: Dont use raw strings, use ProtoId<ContentTileDefinition>

metalgearsloth:
> I'll just work on the current implementation, because I am not capable of moving this to engine. If this is a dealbreaker, I'll just close the PR i guess

Having every caller have to maintain this isn't really feasible long-term on a codebase this large, it's just going to cause bugs instead of doing it at the source and tile code is on engine.

ScarKy0:
I vote for keeping this PR on content cause I really badly need it :3

Simyon264:
> I vote for keeping this PR on content cause I really badly need it :3

It is a technical issue. You can't just vote to keep it on content smh

ScarKy0:
> It is a technical issue. You can't just vote to keep it on content smh

Find someone who knows how to do this on engine and is willing to do it and ill agree
I just see a highly wanted feature finally being made, even if its not 100% perfect

Killerqu00:
soap

Killerqu00:
Quoting Milon on discord: "Having it on content isn't feasible, it should either be on engine or just closed."

i guess I will try to move this to engine, but it needs a full rewrite, so I guess it'll be a different PR

Content.Shared/Maps/ContentTileDefinition.cs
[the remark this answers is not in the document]
deltanedas: you can use ProtoId<ContentTileDef>

PR #36297, opened 2025-04-04
Remove embed mispredict

metalgearsloth:
I don't know why this is here but it doesn't seem to cause issues and transforms should be fully predicted so if there are bugs I will deal with them as they come up.

And if there were some esoteric issue they should've commented what it actually is instead of just not predicting it.


slarticodefast (approved):
I guess we will see if anything happens.

Issue #36286, opened 2025-04-04
Ninjas cannot hack Criminal Records Computers


Toboe5tails:
## Description

Ninjas are currently unable to hack the criminal records computer as required for their objectives. You get a doafter bar on attempt, but once the bar finishes nothing happens; no announcement, and nobody is set to wanted.

**Reproduction**

- Spawn in as Ninja
- Attempt to hack criminal records computer
- Observe nothing happening

**Media**

https://github.com/user-attachments/assets/cc67bf84-cbc8-4e6c-8afe-ee7ae0f12db4

PR #35810, opened 2025-03-13
Localize criminal records reason placeholder dataset

MilenVolf:
## About the PR
Changes type from datasetPrototype to localizedDatasetPrototype
This dataset is the penultimate one. The last one in line is Ion Storm dataset.

Couldn't test in game 'cause as comment says it's currently unused:
image

## Why / Balance
Localization is good

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

PR #36299, opened 2025-04-04
Fix ninjas not being able to hack criminal records

Tayrtahn:
## About the PR

Space ninjas can once again use their gloves to hack criminal records consoles.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/36286

## Technical details

The placeholder reasons `Dataset` prototype was converted to a `LocalizedDataset` by https://github.com/space-wizards/space-station-14/pull/35810, but `CriminalRecordsHackerSystem` was coded to read the prototype as a `Dataset` when hacking. Since no `Dataset` with that `ProtoId` exists anymore, `PrototypeManager.Index` was throwing an exception.

`CriminalRecordsHackerSystem` has been modified to use generic type inference for the `Index` call instead of explicitly calling `Index<DatasetPrototype>` - it will figure out the right type from the `ProtoId` type in the component. Additionally, the `IRobustRandom.Pick` extension method taking a `LocalizedDataset` as a parameter is now used so the resulting strings are localized.

## Media

Screenshot 2025-04-04 at 12 03 50 PM

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed ninjas being unable to hack criminal records consoles.

### DIFF
--- a/Content.Server/CriminalRecords/Systems/CriminalRecordsHackerSystem.cs
+++ b/Content.Server/CriminalRecords/Systems/CriminalRecordsHackerSystem.cs
@@ -4,7 +4,7 @@ using Content.Server.StationRecords.Systems;
 using Content.Shared.CriminalRecords;
 using Content.Shared.CriminalRecords.Components;
 using Content.Shared.CriminalRecords.Systems;
-using Content.Shared.Dataset;
+using Content.Shared.Random.Helpers;
 using Content.Shared.Security;
 using Content.Shared.StationRecords;
 using Robust.Shared.Prototypes;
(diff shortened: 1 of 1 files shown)


Content.Server/CriminalRecords/Systems/CriminalRecordsHackerSystem.cs
```diff
@@ -36,10 +36,10 @@ private void OnDoAfter(Entity<CriminalRecordsHackerComponent> ent, ref CriminalR
         if (_station.GetOwningStation(ent) is not {} station)
             return;
 
-        var reasons = _proto.Index<DatasetPrototype>(ent.Comp.Reasons);
```
slarticodefast, line 39: We should totally make casting one ProtoId into another throw an error or warning.

slarticodefast (approved):

TeenSarlacc:
probably not in the scope for this PR (not that it matters since it's already merged) but how hard would it be to make the criminal record hack repeatable?

PR #23640, opened 2024-01-06
Add: Turnstiles

DrMelon:
## About the PR
Added Turnstiles to the game. Fence-like structures that permit movement by mobs in only one direction, one at a time. There are two variants: a full-size turnstile, and a shorter variant. The short variant can be vaulted and unanchored, whereas the tall variant cannot. Both have construction graphs available. 
Ideally the turnstile would have animation, but I couldn't get anything that looked right. Maybe a better artist could make it work in the future. 

## Why / Balance
This could be used to provide one-way exits from Medbay to allow treated patients to leave, which allows medics to focus on treatment of patients and not have to keep opening the door. Similarly, this can allow for automated release of prisoners from brigs. 
See also #16419 and this design document: https://docs.spacestation14.com/en/proposals/ike709-genpop-security.html 

## Technical details
Added TurnstileComponent and SharedTurnstileSystem, prototypes, sprites, construction graphs, etc. 

## Media

YouTube video:
<a href="http://www.youtube.com/watch?feature=player_embedded&v=FPHSCA0uxuo
" target="_blank">turnstiles video</a>
https://www.youtube.com/watch?v=FPHSCA0uxuo

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: One-way turnstiles are now available to construct for all your traffic flow needs! Get an engineer to build one in Medbay to let people out automatically!

PR #36313, opened 2025-04-05
Turnstiles

EmoGarbage404:
## About the PR

Requires https://github.com/space-wizards/RobustToolbox/pull/5810

Adds turnstiles, one way doors for controlling access into areas. These can be crafted in the construction menu.

Choice code taken [DrMelon's PR](https://github.com/space-wizards/space-station-14/pull/23640/files#diff-697a1c82254e786dd242d6e55978bf6914ab32e5798c1ceebcef9d14cf08cd16). Sprites, description, sfx taken from OracleStation's turnstile implementation found [here](https://github.com/OracleStation/OracleStation/pull/419/files#diff-3c3fbeecd31a52991168835cdf353f0d460d267ca6e32ebf7d2b2e919e01d58f).

## Why / Balance

Turnstiles are needed for [genpop](https://github.com/OracleStation/OracleStation/pull/419/files#diff-3c3fbeecd31a52991168835cdf353f0d460d267ca6e32ebf7d2b2e919e01d58f) and can be used to greatly improve things like medbays. Oftentimes, 

## Technical details

Adds TurnstileSystem and Component for managing logic. Everything is self contained.

## Media

https://github.com/user-attachments/assets/3d58f35b-1d2e-47f8-ad1c-bc616c80c469

## Requirements

## Breaking changes

**Changelog**
:cl: DrMelon, EmoGarbage404
- add: Added turnstiles, gates able to limit the flow of foot traffic. The station's flow is at your fingertips.


EmoGarbage404:
excuse the mildly fucked commits i seem to have transgressed my unholy git setup in my absence

EmoGarbage404:
> I feel like it's better to add a red light when entering the wrong way rather than using a context

the red light is for the access. I think being able to readily distinguish the two is important.

Pronana:
How long from this getting merged before someone is killed by a trap that uses this door?

Pronana:
Who dies to the turnstables maints trap first:
The guy who just installed

The HoS one bad day from a voucher ban

The entite Nukie team who arrived 50 mins into the round

YoungThugSS14:
Excited for this but worried on how people are gonna use this.

Content.Shared/Doors/Components/TurnstileComponent.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// A whitelist of the things this turnstile blocks. Anything else (bullets, etc) can fly right through.
+    /// </summary>
+    [DataField]
+    public EntityWhitelist BlockWhitelist = new();
+
+    /// <summary>
+    /// The next time this turnstile can attempt to be passed through.
+    /// </summary>
+    [DataField, AutoNetworkedField, AutoPausedField]
+    public TimeSpan NextPassTime;
```
slarticodefast, line 26: needs the time offset serializer

Content.Shared/Doors/Components/TurnstileComponent.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// The minimum time between passes.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public TimeSpan PassDelay = TimeSpan.FromSeconds(2);
+
+    /// <summary>
+    /// The next time at which the resist message can show.
+    /// </summary>
+    [DataField, AutoNetworkedField, AutoPausedField]
+    public TimeSpan NextResistTime;
```
slarticodefast, line 38: time offset serializer

Content.Shared/Doors/Components/TurnstileComponent.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// Maintained hashset of entities currently passing through the turnstile.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public HashSet<EntityUid> CollideExceptions = new();
+
+    /// <summary>
+    /// animation state of the turnstile spinning.
+    /// </summary>
+    [DataField]
+    public string SpinState = "operate";
```
slarticodefast, line 50: these being strings is extremely sus
they should be validated for existence somehow
the flick implementation will probably have to be adjusted for that

Content.Shared/Doors/Components/TurnstileComponent.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// <summary>
+    /// Sound to play when the turnstile denies entry
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? DenySound = new SoundPathSpecifier("/Audio/Machines/airlock_deny.ogg");
+}
+
+[Serializable, NetSerializable]
+public enum TurnstileVisualLayers : byte
```
slarticodefast, line 72: should this be in the component, system or a separate file?

Content.Shared/Doors/Components/TurnstileComponent.cs
```diff
@@ -0,0 +1,75 @@
...
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, AutoGenerateComponentPause]
+[Access(typeof(TurnstileSystem))]
+public sealed partial class TurnstileComponent : Component
+{
+    /// <summary>
+    /// A whitelist of the things this turnstile blocks. Anything else (bullets, etc) can fly right through.
+    /// </summary>
+    [DataField]
+    public EntityWhitelist BlockWhitelist = new();
```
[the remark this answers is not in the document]
EmoGarbage404: Pedantic: this is not a blacklist. this specifically is a whitelist for what items the turnstile will use its process behavior for. Otherwise, if it fails the whitelist, it will get default behavior (pass through)

i've updated the name and comment to better reflect this purpose.

Content.Shared/Doors/Components/TurnstileComponent.cs
EmoGarbage404: Currently all types of state-based animation are just done by passing around strings, including prior methods like manually queuing animations on the client and generic visualizers. Due to the nature of how they work, it's not really possible to ensure there's going to be an RSI on a given layer and as such validation isn't possible.

I'd be happy to update the Flick implementation to match and kind of validation standard once one is created for the codebase.

deltanedas:
wb king

EmoGarbage404:
> wb king

stunting will continue til game improves

FairlySadPanda:
I'll do a sweep of this later as I touched doorcode last.

@EmoGarbage404 if you can try and move Content.Shared/Doors/Systems/TurnstileSystem.cs into DoorSystem as a partial that'd be handy, that would match the ongoing work 🫡

EmoGarbage404:
> EmoGarbage404 if you can try and move Content.Shared/Doors/Systems/TurnstileSystem.cs into DoorSystem as a partial that'd be handy, that would match the ongoing work 🫡

is there a particular reason it needs to be a partial? The similarities are more superficial than anything relating to code. They basically don't contain any overlapping logic, unlike firelocks and airlocks which are basically just extensions.

FairlySadPanda:
I'll have a watch later and see, I'm sure you're right.

FairlySadPanda:
OK, actually sat down and read it.

Yes, it's completely independent of doors. I'm actually thinking it's sufficiently independent as to be its own namespace; the only thing this shares with "door" is that they're both concept of something blocks access. 

In that way I'm a bit 50/50 on it. A turnstile can't be bolted, set to emergency access, and so on.

Whilst this encapsulates what it needs to do well, ideally "one way door" would be a feature a door can have.

Resources/Prototypes/Entities/Structures/Doors/turnstile.yml
```diff
@@ -0,0 +1,62 @@
...
+  - type: Sprite
+    sprite: Structures/Doors/turnstile.rsi
+    snapCardinals: true
+    drawdepth: Doors
+    layers:
+    - state: turnstile
+      map: [ "enum.TurnstileVisualLayers.Base" ]
+  - type: Rotatable
+  - type: Transform
+    noRot: false
```
metalgearsloth, line 16: what

Content.Shared/Doors/Systems/TurnstileSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        SubscribeLocalEvent<TurnstileComponent, PreventCollideEvent>(OnPreventCollide);
+        SubscribeLocalEvent<TurnstileComponent, EndCollideEvent>(OnEndCollide);
+    }
+
+    private void OnPreventCollide(Entity<TurnstileComponent> ent, ref PreventCollideEvent args)
+    {
+        if (args.Cancelled || !args.OurFixture.Hard)
+            return;
+
```
metalgearsloth, line 39: What's the non-hard layer doing?

Content.Shared/Doors/Systems/TurnstileSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+        // unblockables go through for free.
+        if (_entityWhitelist.IsWhitelistFail(ent.Comp.ProcessWhitelist, args.OtherEntity))
+        {
+            args.Cancelled = true;
+            return;
+        }
+
+        if (_timing.CurTime < ent.Comp.NextPassTime)
+            return;
+
```
metalgearsloth, line 62: Isn't this going to mispredict if more than 1 ent even touches it. That seems pretty bad? Movement stuff should always be fully predicted. Worst case the client just tells the server if it got rejected or not just so it doesn't feel bad ingame.

Content.Shared/Doors/Systems/TurnstileSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+                _animationPlayer.Flick(ent, ent.Comp.DenyState, TurnstileVisualLayers.Base);
+            return;
+        }
+
+        var xform = Transform(ent);
+        var otherXform = Transform(args.OtherEntity);
+
+        var moveDir = (xform.LocalPosition - otherXform.LocalPosition).GetDir();
+        var faceDir = xform.LocalRotation.GetDir();
+
```
metalgearsloth, line 76: Does this even work if someone is off-grid and touches the turnstile.

Content.Shared/Doors/Systems/TurnstileSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            if (_timing.CurTime >= ent.Comp.NextResistTime)
+            {
+                _popup.PopupPredicted(Loc.GetString("turnstile-component-popup-resist", ("turnstile", ent.Owner)), ent, args.OtherEntity);
+                ent.Comp.NextResistTime = _timing.CurTime + TimeSpan.FromSeconds(0.5);
+                Dirty(ent);
+            }
+        }
+    }
+
```
metalgearsloth, line 100: Doing stuff in attempt events is really suss, also this is going to be spammed every single frame in collision which is the most expensive part of box2d which is not good.

I think you're better off just cancelling either the hard / non-hard collision as appropriate and just running the rest of the logic in startcollideevent so it also predicts properly when contacts get fully predicted and doesn't suddenly break in the next few months.

metalgearsloth (changes requested):

EmoGarbage404:
removed engine dependency from this PR and addressed all the comments sloth had (besides the collision stuff cuz its complique)

Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
```diff
@@ -341,6 +341,22 @@ public bool IsPulling(EntityUid puller, PullerComponent? component = null)
         return Resolve(puller, ref component, false) && component.Pulling != null;
     }
 
+    public EntityUid? GetPuller(EntityUid puller, PullableComponent? component = null)
+    {
+        if (!Resolve(puller, ref component, false))
+            return null;
+
+        return component.Puller;
```
ArtisticRoomba, line 349: kid named
```suggestion
        return !Resolve(puller, ref component, false) ? null : component.Puller;
```

Content.Shared/Movement/Pulling/Systems/PullingSystem.cs
```diff
@@ -341,6 +341,22 @@ public bool IsPulling(EntityUid puller, PullerComponent? component = null)
...
+    }
+
+    public EntityUid? GetPulling(EntityUid puller, PullerComponent? component = null)
+    {
+        if (!Resolve(puller, ref component, false))
+            return null;
+
+        return component.Pulling;
```
ArtisticRoomba, line 357: kid named
```suggestion
        return !Resolve(puller, ref component, false) ? null : component.Pulling;
```

Content.Shared/Doors/Systems/SharedTurnstileSystem.cs
```diff
@@ -0,0 +1,134 @@
...
+        var xform = Transform(ent);
+        var otherXform = Transform(other);
+
+        var (pos, rot) = _transform.GetWorldPositionRotation(xform);
+        var otherPos = _transform.GetWorldPosition(otherXform);
+
+        var approachVec = pos - otherPos;
+        var rotVec = rot.ToWorldVec();
+        return Math.Abs(approachVec.ToAngle().Reduced().Theta - rotVec.ToAngle().Reduced().Theta) < Math.PI / 4;
+    }
```
ArtisticRoomba, line 127: methinks should become a helper method somewhere in the future?

ArtisticRoomba (approved):
yaml and attribution are good, testing was cool and good

Content.Shared/Doors/Systems/SharedTurnstileSystem.cs
EmoGarbage404: i mean it's literally just taking a vector and comparing angles. i guess it could be though i have no idea what would use it

ArtisticRoomba: if new code in the future starts copy/pasting this and we need to change this code for some reason (obs or buggy) then we'll have to change it in 10 different places rather than one (and know that it exists in the first place)

and it's the right implementation so might as well set it in stone(trademark)

feel free to ignore though

ScarKy0 (changes requested):
So just from testing in game, few things

1. The sound needs to be more quiet, walking through a turnstile makes my ears bleed.
2. PLEASE some directional indicator. They ALWAYS look the same, no matter which direction they allow passage in.
3. Whatever bug this is

https://github.com/user-attachments/assets/86ff4e51-ac11-4c1a-85c7-28aefe6c9ca9

EmoGarbage404:
> PLEASE some directional indicator. They ALWAYS look the same, no matter which direction they allow passage in 

to bonk into the wrong side of a turnstile is to err. And to err is to be human. It tickles the soul and nourishes the spirit. Why would a moral being deprive a person of the most basest of pleasures?

ScarKy0:
> to bonk into the wrong side of a turnstile is to err. And to err is to be human. It tickles the soul and nourishes the spirit. Why would a moral being deprive a person of the most basest of pleasures?

It is absolutely hilarious. But lack of sprites for facing direction makes them an absolute pain to map and construct as you literally dont know which way they face. Even some arrow on the sprite would massively help.

ArtisticRoomba:
There's an arrow on the sprite on mapping mode but not otherwise

ScarKy0:
> There's an arrow on the sprite on mapping mode but not otherwise

Is there? I saw nothing when spawning them in. Unless its like only full on mapping mode and not entity menu? In such case it would still need some indicator.

ArtisticRoomba:
Yeah it's only on mapping mode. It should get an indicator.

BigfootBravo:
Maybe allow them to be climbed through with a very long do-after? Feels weird that you couldn't climb over or make your way through the turnstiles, and would also help their balance a bit. Could still make them useful by just making the do-after very slow, allowing them to still fulfill their role of discouraging or restricting access in a particular direction? 

Or maybe make it be two different ones, ones that can be climbed over very slowly and ones that can't, with the latter requiring extra construction materials.

EmoGarbage404:
> Maybe allow them to be climbed through with a very long do-after? Feels weird that you couldn't climb over or make your way through the turnstiles, and would also help their balance a bit. Could still make them useful by just making the do-after very slow, allowing them to still fulfill their role of discouraging or restricting access in a particular direction?

the primary purpose of these is in restricted sec locations (genpop). if you can just climb over it, then it would be completely ineffective.

i don't mind if someone adds a climbable one later but at the very least this one needs to be secure.

BigfootBravo:
> > Maybe allow them to be climbed through with a very long do-after? Feels weird that you couldn't climb over or make your way through the turnstiles, and would also help their balance a bit. Could still make them useful by just making the do-after very slow, allowing them to still fulfill their role of discouraging or restricting access in a particular direction?
> 
> the primary purpose of these is in restricted sec locations (genpop). if you can just climb over it, then it would be completely ineffective.
> 
> i don't mind if someone adds a climbable one later but at the very least this one needs to be secure.

Very fair! I was initially not realizing that this PR was about full door size ones, with my mind initially going towards the like half-height, or chest like lower-chest high ones, so was expecting that they could be climbed over.
I do think there should be a form that can be climbed over though, eventually. Could allow more choice with which to build during midround and such, and just giving more tools to mappers, providing strong exit points that could be used against a dept if not monitored every once in a while.

insoPL:
Maybe it would be better to use pipes and disposals system for piping ~~people~~ prisoners into and from prisons. Would allow more creative mapping with prisons further away from sec and funny piping shenanigans.

metalgearsloth (approved):

metalgearsloth:
Contact code is poopy but the events are probably getting redone in box2d v3.

ScarKy0 (approved):
So the sprites just full on rotate as construction ghosts. It's ass and will need to be looked into in the future (but it's fully unrelated to this PR, afaik) so should be fine.

PR #36319, opened 2025-04-05
Fix blocked UI interaction on unpowered devices

### LINKED ISSUE #36242: Vulture. Door maintenance wires are inaccessible.
## Description

Round 80941, multiple attempted to hack doors with no results after several plasma fires. Issue may extend to other item.s

**Reproduction**

Open door panel, try to interact via wire cutter or multitool, no wiring UI.

**Additional context**

Several others reported the same issue at the same time.

ArtisticRoomba:
## About the PR
This PR fixes UI interaction (ex. opening) being blocked if the device is unpowered.

## Why / Balance
It was a bug.

Fixes #36242.

## Technical details
Right now this checks if the device is unpowered, if it is, it blocks the interaction:
```C#
    private void OnMessageAttempt(BoundUserInterfaceMessageAttempt ev)
    {
        if (ev.Actor == ev.Target)
            return;

        // no need to show menu if device is not powered.
        if (!PowerReceiver.IsPowered(ev.Target))
        {
            ShowDeviceNotRespondingPopup(ev.Actor);
            ev.Cancel();
            return;
        }
```

This is not a good idea as you're blocking every single interaction that comes through if the device is unpowered.

To fix this, we just move the check into the realm where we check if the person interacting is an AI. If the person interacting is an AI, and if the device is unpowered, then block the interaction.

The rest of that logic also blocks the interaction if it's disallowed by the component (AI wire being cut for example).

## Media
Door is depowered and I can still interact with it and cut wires.
Content Client_mI3IUpIhNQ

## Requirements

**Changelog**
:cl:
- fix: Fixed players being unable to interact with or open any machine UI if the machine is unpowered (ex. wirepanels).

### DIFF
--- a/Content.Shared/Silicons/StationAi/SharedStationAiSystem.Held.cs
+++ b/Content.Shared/Silicons/StationAi/SharedStationAiSystem.Held.cs
@@ -121,18 +121,19 @@ public abstract partial class SharedStationAiSystem
         if (ev.Actor == ev.Target)
             return;
 
-        // no need to show menu if device is not powered.
-        if (!PowerReceiver.IsPowered(ev.Target))
-        {
-            ShowDeviceNotRespondingPopup(ev.Actor);
-            ev.Cancel();
-            return;
-        }
-
         if (TryComp(ev.Actor, out StationAiHeldComponent? aiComp) &&
            (!TryComp(ev.Target, out StationAiWhitelistComponent? whitelistComponent) ||
             !ValidateAi((ev.Actor, aiComp))))
         {
+            // Don't allow the AI to interact with anything that isn't powered.
+            if (!PowerReceiver.IsPowered(ev.Target))
+            {
+                ShowDeviceNotRespondingPopup(ev.Actor);
+                ev.Cancel();
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
Summoning @Fildrance because these changes fix the changes made in #34639 which caused this bug, he might know why this was done and if I'm missing anything.

Fildrance:
> Summoning @Fildrance because these changes fix the changes made in #34639 which caused this bug, he might know why this was done and if I'm missing anything.

That was small refactor that tried to reduce amount of nested code and was not respecting the fact that sub for BoundUserInterfaceMessageAttempt was global, and the exact restriction for handling only AI related branch was that nested If. As a side activity i can propose extracting the insides into seprate method. With correct xml-doc that separation would prevent such problems in future.

FairlySadPanda:
Making a PR.

FairlySadPanda:
it took more than an hour

This PR is a suitable hotfix but unless I'm wrong the actual bug is the mangling of the logic and the failure to scope the event correctly, causing random Urists to trigger AI code unexpectedly.

Issue #36304, opened 2025-04-04
Chem synthesis kit inaccurate description


Blobadoodle:
## Description
The chemical synthesis kit's description in the uplink states that it includes toxin except it doesn't.

**Reproduction**
1. Read uplink description
2. Buy it
3. Realise it has no toxin
4. Cry after wasting your TC

**Screenshots**
![Image](https://github.com/user-attachments/assets/15c29da9-cc8b-4896-bde5-75cb8d7e7ca2)
![Image](https://github.com/user-attachments/assets/ec237712-6970-4846-a586-874c37ca7eb4)
(two vials of vestine, no toxin)

**Additional context**
idk if it needs toxin added, or the description removed

PR #36323, opened 2025-04-05
Remove toxins from the chemical synthesis kit description

PhilIngham:
## About the PR

Updated the description of the chemical synthesis kit to no longer say it has toxins inside. Now specifies it has two whole vials of vestine instead!
## Why / Balance

Stops someone buying it expecting toxin inside when there is no toxin.

Fixes [#36304](https://github.com/space-wizards/space-station-14/issues/36304)
## Technical details

Changing the words in the descriptions of the items both in game and in the syndicate uplink.
## Media

image

image

image

## Requirements

**Changelog**

:cl:
- fix: Fixed the Chemical Synthesis Kit claiming to have toxin inside when it infact did not.


muburu:
should probably add something saying what vestine is actually used for as well while you're at it, if a newer player doesn't know what vestine is they probably won't bother

PhilIngham:
Well I'd think the only reason someone would buy it would be for the vestine itself, and they'd only do that if they had an idea of what to create with it.

slarticodefast (approved):

PR #36325, opened 2025-04-05
Allow Aghosts to load biomes.

ScarKy0:
## About the PR

Allows aghosts to load in biome terrain.

## Why / Balance

Its stupid to be able to fly around on planets and have it not load.

## Technical details

Added a tag check.

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- tweak: Aghosts now load biome terrain.

### DIFF
--- a/Content.Server/Parallax/BiomeSystem.cs
+++ b/Content.Server/Parallax/BiomeSystem.cs
@@ -16,6 +16,7 @@ using Content.Shared.Light.Components;
 using Content.Shared.Parallax.Biomes;
 using Content.Shared.Parallax.Biomes.Layers;
 using Content.Shared.Parallax.Biomes.Markers;
+using Content.Shared.Tag;
 using Microsoft.Extensions.ObjectPool;
 using Robust.Server.Player;
 using Robust.Shared;
@@ -50,6 +51,7 @@ public sealed partial class BiomeSystem : SharedBiomeSystem
     [Dependency] private readonly SharedPhysicsSystem _physics = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly ShuttleSystem _shuttles = default!;
+    [Dependency] private readonly TagSystem _tags = default!;
 
     private EntityQuery<BiomeComponent> _biomeQuery;
     private EntityQuery<FixturesComponent> _fixturesQuery;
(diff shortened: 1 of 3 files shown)


IProduceWidgets:
Makes sense to me. iirc the reason we 1984ed it in the first place was bad actors using observers to load large amounts of tiles and entities. This avoids that though, so seems good.

IProduceWidgets (approved):

Content.Server/Parallax/BiomeSystem.cs
```diff
@@ -59,6 +61,7 @@ public sealed partial class BiomeSystem : SharedBiomeSystem
     private readonly HashSet<EntityUid> _handledEntities = new();
     private const float DefaultLoadRange = 16f;
     private float _loadRange = DefaultLoadRange;
+    private static readonly ProtoId<TagPrototype> MapLoaderTag = "AllowBiomeLoading";
```
slarticodefast, line 64:
```suggestion
    private static readonly ProtoId<TagPrototype> AllowBiomeLoadingTag = "AllowBiomeLoading";
```
let's keep it consistent with all the other examples of this being done

Content.Server/Parallax/BiomeSystem.cs
```diff
@@ -321,7 +324,7 @@ public void Preload(EntityUid uid, BiomeComponent component, Box2 area)
 
     private bool CanLoad(EntityUid uid)
     {
-        return !_ghostQuery.HasComp(uid);
+        return !_ghostQuery.HasComp(uid) || _tags.HasTag(uid, MapLoaderTag);
```
slarticodefast, line 327:
```suggestion
        return !_ghostQuery.HasComp(uid) || _tags.HasTag(uid, AllowBiomeLoadingTag);
```

slarticodefast (changes requested):
At some point we will finally have to give the aghost its own component instead of handling everything via tags. But I guess this is fine for now.

ScarKy0:
> At some point we will finally have to give the aghost its own component instead of handling everything via tags. But I guess this is fine for now.

Replacing it all with AghostComponent sounds like a pain, and tags can be easily applied everywhere

slarticodefast:
> Replacing it all with AghostComponent sounds like a pain, and tags can be easily applied everywhere

Yeah, I don't mean to replace all of them by hardcoding the component either.
But some might be better if they are not tags.

slarticodefast (approved):

PR #36326, opened 2025-04-05
fix special scrubber/vent tags

deltanedas:
## About the PR
vox/freezer scrubbers and vents were being broken by the `fixrotations` command since they had air sensor tags from AirSensorBase
moved them to air sensor since its nothing to do with the monitoring functionality

it also meant vox/freezer scrubbers/vents didnt have the scrubber/vent and unstackable tags

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Specific/Atmospherics/sensor.yml
+++ b/Resources/Prototypes/Entities/Structures/Specific/Atmospherics/sensor.yml
@@ -25,10 +25,6 @@
       Ammonia: stationAmmonia
       NitrousOxide: stationNO
       Frezon: danger
-  - type: Tag
-    tags:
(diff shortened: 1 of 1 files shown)


slarticodefast (commented):
Wouldn't it be better to have `AirSensorVoxBase` not inherit from `AirSensorBase` at all instead because it causes the same problem with all other datafields?

deltanedas:
true but not related to the fixrotations bug
nfi why i inherited it to begin with its abstract

CawsForConcern:
Just ran into this bug myself when mapping. It makes creating nitrogen / vox friendly areas and stations incredibly error prone -- I'd really appreciate seeing this merged since it's been a few months without a review.

deltanedas:
3 months left

PR #36327, opened 2025-04-05
Clean up TitleWindowManager.cs

VasilisThePikachu:
## Why / Balance

- I did not like how `OnHostnameChange()` always needed a string even though that string would always just be the hostname, so now it's just part of its function
- The extra function made to just trigger `OnHostnameChange()` are removed. It just runs the right function off the bat.
- Checking for `ClientRunLevel.InGame` for setting the title without the hostname, which means the previous joined server won't appear for a split second before being corrected by the new cvars being set. Or if the server prefers no host name in the titlebar by the time we connect.

## Requirements

### DIFF
--- a/Content.Client/GameTicking/Managers/TitleWindowManager.cs
+++ b/Content.Client/GameTicking/Managers/TitleWindowManager.cs
@@ -15,48 +15,35 @@ public sealed class TitleWindowManager
 
     public void Initialize()
     {
-        _cfg.OnValueChanged(CVars.GameHostName, OnHostnameChange, true);
-        _cfg.OnValueChanged(CCVars.GameHostnameInTitlebar, OnHostnameTitleChange, true);
+        _cfg.OnValueChanged(CVars.GameHostName, _ => OnHostnameChange(), true);
+        _cfg.OnValueChanged(CCVars.GameHostnameInTitlebar, _ => OnHostnameChange(), true);
 
-        _client.RunLevelChanged += OnRunLevelChangedChange;
+        _client.RunLevelChanged += (_, _) => OnHostnameChange();
     }
 
     public void Shutdown()
     {
-        _cfg.UnsubValueChanged(CVars.GameHostName, OnHostnameChange);
-        _cfg.UnsubValueChanged(CCVars.GameHostnameInTitlebar, OnHostnameTitleChange);
+        _cfg.UnsubValueChanged(CVars.GameHostName, _ => OnHostnameChange());
+        _cfg.UnsubValueChanged(CCVars.GameHostnameInTitlebar, _ => OnHostnameChange());
     }
 
-    private void OnHostnameChange(string hostname)
+    private void OnHostnameChange()
     {
         var defaultWindowTitle = _gameController.GameTitle();
 
-        // Since the game assumes the server name is MyServer and that GameHostnameInTitlebar CCVar is true by default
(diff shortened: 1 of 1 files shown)


Content.Client/GameTicking/Managers/TitleWindowManager.cs
```diff
@@ -15,48 +15,35 @@ public sealed class TitleWindowManager
...
+        _client.RunLevelChanged += (_, _) => OnHostnameChange();
     }
 
     public void Shutdown()
     {
-        _cfg.UnsubValueChanged(CVars.GameHostName, OnHostnameChange);
-        _cfg.UnsubValueChanged(CCVars.GameHostnameInTitlebar, OnHostnameTitleChange);
+        _cfg.UnsubValueChanged(CVars.GameHostName, _ => OnHostnameChange());
+        _cfg.UnsubValueChanged(CCVars.GameHostnameInTitlebar, _ => OnHostnameChange());
     }
```
metalgearsloth, line 28: Does unsubbing anonymous methods even work?

VasilisThePikachu: /shrug 

Did it to be safe, I was told it does not really matter here in the pr this code got introduced.

Xeri7 (approved):

Content.Client/GameTicking/Managers/TitleWindowManager.cs
```diff
@@ -15,48 +15,35 @@ public sealed class TitleWindowManager
 
     public void Initialize()
     {
-        _cfg.OnValueChanged(CVars.GameHostName, OnHostnameChange, true);
-        _cfg.OnValueChanged(CCVars.GameHostnameInTitlebar, OnHostnameTitleChange, true);
+        _cfg.OnValueChanged(CVars.GameHostName, _ => OnHostnameChange(), true);
+        _cfg.OnValueChanged(CCVars.GameHostnameInTitlebar, _ => OnHostnameChange(), true);
```
Simyon264, line 19: This does not work for unsubbing.

The reason for this is that this makes a new lamba.

Content.Client/GameTicking/Managers/TitleWindowManager.cs
```diff
@@ -15,48 +15,35 @@ public sealed class TitleWindowManager
...
-        _client.RunLevelChanged += OnRunLevelChangedChange;
+        _client.RunLevelChanged += (_, _) => OnHostnameChange();
     }
 
     public void Shutdown()
     {
-        _cfg.UnsubValueChanged(CVars.GameHostName, OnHostnameChange);
-        _cfg.UnsubValueChanged(CCVars.GameHostnameInTitlebar, OnHostnameTitleChange);
+        _cfg.UnsubValueChanged(CVars.GameHostName, _ => OnHostnameChange());
+        _cfg.UnsubValueChanged(CCVars.GameHostnameInTitlebar, _ => OnHostnameChange());
```
Simyon264: See above

Simyon264 (dismissed):
I am not sure this is actually needed? It's already clean.

Content.Client/GameTicking/Managers/TitleWindowManager.cs
Simyon264: dont unsub, nerd

Simyon264 (changes requested):

PR #36328, opened 2025-04-05
Refactor magic speak system to be a component added to actions

ljm862:
## About the PR

Part of #34597
This extracts the speaking spell event and makes it generic for all actions. Any action can now attach the `SpeakOnAction` component and be able to say some words in IC chat on use of the action. 

## Why / Balance

Again part of #34597. This is another step in refactoring the wizard and magic system enabling us to do more with it in the future. Also enables other actions to use speech if they wish.

## Technical details

Adds a new Component `SpeakOnActionComponent` which is attached to actions. The `ActionPerformedEvent` is now read as well by the new `SpeakOnActionSystem` which writes a message into the chat based on the value supplied to the component in the yaml file.

Unfortunately the `SpeakOnActionSystem` is in `Server`, not `Shared`. When the chat refactor has been completed, this should be a simple switch over though to `Shared`.

It also removes the `ISpeakSpell` interface and all speech related methods for the `MagicSystem`.

## Media

This first video is the wizard system working correctly with the new speech setup.

https://github.com/user-attachments/assets/81026e74-357e-42e5-8b60-69f77a1c28e7

This video below is **NOT** in the PR but by simply a demonstration that we can now set other Actions, such as the `ActionCombatModeToggle` to also use the `SpeakOnActionComponent` if we wish.

https://github.com/user-attachments/assets/31a127e4-eea2-4e45-b82b-11fd0a6f1dd6

## Requirements

## Breaking changes

`ISpeakSpell` is no longer available. You now need to attach a `SpeakOnAction` component to the ActionEvent in the yaml file and set `sentence` instead of `speech`


Content.Server/Speech/EntitySystems/SpeakOnUseSystem.cs
```diff
@@ -0,0 +1,41 @@
...
+        SubscribeLocalEvent<SpeakOnUseComponent, SpeakOnUseEvent>(OnSpeakOnUse);
+    }
+
+    private void OnSpeakOnUse(Entity<SpeakOnUseComponent> ent, ref SpeakOnUseEvent args)
+    {
+        var user = args.Performer;
+
+        // If we can't speak, we can't speak
+        if (!HasComp<SpeechComponent>(user) || HasComp<MutedComponent>(user))
+            return;
```
deltanedas, line 34: this should cancel an action attempt event as in ss13 you cant say (most) spells while muted (exceptions like mansus grasp which are "spells" but you can use them while muted)

keronshb: I'm not adding that functionality right now.

Content.Shared/Speech/Components/SpeakOnUseComponent.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Content.Shared.Speech.Components;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(SharedSpeakOnUseSystem))]
+public sealed partial class SpeakOnUseComponent : Component
+{
+    /// <summary>
+    /// The ftl id of the sentence that the user will speak.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public string? Sentence;
```
Tayrtahn, line 13:
```suggestion
    public LocId? Sentence;
```

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -701,6 +703,12 @@ public void PerformAction(EntityUid performer, ActionsComponent? component, Enti
             action.Toggled = !action.Toggled;
         }
 
+        if (HasComp<SpeakOnUseComponent>(actionId))
+        {
+            var speakEvent = new SpeakOnUseEvent(performer);
+            RaiseLocalEvent(actionId, ref speakEvent);
+        }
```
Tayrtahn, line 710: This feels like a slightly odd use of components and events. 

Would it make more sense not to do a HasComp here and just always raise an `ActionActivatedEvent` (or something) on the action entity? Then `SpeakOnUseSystem` would subscribe to that event instead. That would avoid having to hardcode `SpeakOnUseComponent` here and leave open the ability for other systems that work in a similar way.

Content.Shared/Speech/Components/SpeakOnUseComponent.cs
ljm862: Thank you! I knew string wasn't correct. I couldn't remember what the type should be. Big help!

Content.Shared/Actions/SharedActionsSystem.cs
ljm862: Yes, I think your solution is more extensible! I'm still new to the whole ECS paradigm so things like this I'm pretty sure as I'm doing it it feels wrong, just don't know the right solution for this instance. Thanks for the tip!

I did have it not checking the `HasComp` before and just raising the event as is. I didn't know which one had more overhead and expected the `HasComp` to cut out a lot of unnecessary event raises for all other actions such as `CombatStance` or something else. My idea was that we would make a more generic event and raise that all the time here when we have more listeners to catch that but as well all know, refactors are elusive creatures :D

ljm862: Ah! I'll avoid my whole concern by just hooking into the `ActionPerformedEvent` which happens about 20 lines after this at the end anyway 😅

Tayrtahn: Perfect!

Content.Shared/Speech/Components/SpeakOnUseComponent.cs
```diff
@@ -0,0 +1,14 @@
+using Content.Shared.Speech.EntitySystems;
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Speech.Components;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(SharedSpeakOnUseSystem))]
+public sealed partial class SpeakOnUseComponent : Component
```
slarticodefast, line 7: Don't call it `SpeakOnUseComponent` the `DoXonY` naming pattern is a common one, and the `OnUse` is already used for for the `UseInHandEvent` subscription. `SpeakOnAction` or similar would be better.
Also add documentation to the component.

keronshb:
Ill grab the review for this one.

metalgearsloth:
Really needs event list support for actions but eh for now.

metalgearsloth (approved):

keronshb:
Relatively straightforward, pretty easy to read. Just fix the conflict and this is good to go.

keronshb (approved):

PR #36329, opened 2025-04-05
Add interaction tests for ninja hacking abilities

Tayrtahn:
## About the PR

Adds ~three~ four interaction tests that turn the test player into a ninja antag and try to hack a criminal records console, an R&D server, a communications console, and an APC, making sure that they work.

Also adds a feature to `InteractionTest` that allows the test to spawn a station entity and register the test grid to it.

## Why / Balance

There was a bug recently (#36286) that prevented ninjas from hacking criminal records consoles and the cause was something not obviously connected to ninja code. It's better to catch these issues with automated testing than to rely on someone finding it after it's already been broken.
The test of APC battery draining was added in response to https://github.com/space-wizards/space-station-14/issues/36454.

## Technical details

Each test turns the player into a ninja antag, spawns the related hackable entity, has the player interact with it with their gloves disabled (to make sure they don't hack it), activates their gloves, and then has them interact with it again (making sure that the hack succeeds).
The tests are pretty similar, and a fair bit of common functionality was put into helpers that are used in all three tests, and may be helpful for future ninja-related tests:
- `MakeNinja` turns the target into a ninja antag and makes sure that they have the expected gear, also storing the net ids of the gear in private fields for tests to use.
- `HackTestSystem<TComp, TEvent>` is a very simple abstract class that adds a subscription for `TEvent` on `TComp` and sets a flag when the event is raised. This was needed in order to listen for the successful hack events on the ninja while avoiding duplicate subscriptions on the `SpaceNinjaComponent`. Each test adds a test component to the ninja entity and checks for changes in a child class of `HackTestSystem` customized for that component and the related event.

The ninja game rule needs a station to exist in order to be considered valid, and it's much easier to set up the station before spawning the player, so this feature was added to the setup of `InteractionTest` using an optional prototype that can be overridden in interaction test classes. It should come in handy for other tests that need a station in order to function.

## Requirements


Tayrtahn:
Added a test of APC battery draining, which will fail pending https://github.com/space-wizards/space-station-14/pull/36461.

Tayrtahn:
Added a test of the suit stealth ability, which will fail pending #37198.

slarticodefast:
I'll have to temporarily freeze this PR until https://github.com/space-wizards/space-station-14/pull/41379 is merged to minimize merge conflicts for the refactor.

PR #36331, opened 2025-04-05
[Hotfix] Fix the walking alert

youtissoum:
## About the PR

I fixed the walking alert hiding itself shortly after starting to walk with toggle walk.
I also added a test to make sure that client-sided alerts do not break again.

## Why / Balance

99% of Space Station 14 players agree that alerts should not hide themselves when they are still relevant.

## Technical details

The alert prototype now contains a `ClientOnly` field which is copied to the alert state.
Client-sided alerts are temporarily held in another dictionary when handling the state and readded after the server state is applied.

## Media

https://github.com/user-attachments/assets/a39d755c-3ea6-4475-84f7-65b1a7883e34

## Requirements

## Breaking changes

Client-sided alerts now require that the `ClientOnly` datafield on the alert prototype is set to true.

**Changelog**

:cl:
- fix: The walking alert will no longer hide itself shortly after appearing.

### DIFF
--- a/Content.Client/Alerts/ClientAlertsSystem.cs
+++ b/Content.Client/Alerts/ClientAlertsSystem.cs
@@ -52,7 +52,22 @@ public sealed class ClientAlertsSystem : AlertsSystem
         if (args.Current is not AlertComponentState cast)
             return;
 
+        // Temporarily store all clientside alerts in another dictionary.
+        var clientAlertStorage = new Dictionary<AlertKey, AlertState>();
+        foreach (var clientAlert in alerts.Comp.Alerts)
+        {
+            if (clientAlert.Value.ClientOnly)
+                clientAlertStorage.Add(clientAlert.Key, clientAlert.Value);
+        }
+
         alerts.Comp.Alerts = new(cast.Alerts);
+
+        // Reapply the stored alerts after they have been removed.
+        foreach (var clientAlert in clientAlertStorage)
+        {
+            if (!alerts.Comp.Alerts.ContainsKey(clientAlert.Key))
+                alerts.Comp.Alerts[clientAlert.Key] = clientAlert.Value;
+        }
 
         UpdateHud(alerts);
     }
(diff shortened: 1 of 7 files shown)


youtissoum:
> Can you test if the alert works correctly when you rebind walk and shuttle braking to the same key (e.g. Space) then brake while operating a shuttle?

The walking *alert* appears to function correctly but the walking itself still gets toggled through the shuttle controls. That issue wasn't caused by this PR though and seems like it would be out of scope to me.

metalgearsloth (changes requested):
This looks more like a prediction issue than a dictionary issue. You should look at why it's not being re-applied properly during prediction instead. Having to manually curate predicted code as a workaround doesn't make for good yaml.

youtissoum:
> This looks more like a prediction issue than a dictionary issue. You should look at why it's not being re-applied properly during prediction instead. Having to manually curate predicted code as a workaround doesn't make for good yaml.

The issue happens because the client alert dictionary is completely replaced whenever the server sends a new state over, which is needed or it might cause mispredicts.

In order to make client-side alerts work again I need to either mark the alert itself as client-side so I can stop the state handler from overwriting it or keep it in another place that will not get overwritten when a new state comes in.
Keeping it in another place doesn't seem possible to me without rewriting a lot of the alert system to be much more complicated.
Marking it as client-side is the current method I chose and could also work by making a new `ShowClientAlert` function which marks the alert as client-side instead of defining that on the prototype.

It's also possible to discontinue client-side alerts and make the client send an event when it needs an alert shown instead.

I'm not sure which of these options is the better one.

youtissoum:
I have decided to make a function that marks the alert as client-side instead of using a field on the prototype.
I also got rid of the test to add it later in another PR.

VerinSenpai:
Seems the issue is caused by #36004. Ironic. I'm uncertain as to what's going on in this code. I don't know why or if new should be clearing the states or what? Just an observation.

EmoGarbage404:
Superseded by #37033

PR #36336, opened 2025-04-06
Overhauled stamina slowdown behavior

### LINKED ISSUE #34895: Stamina Slowdown bugs out around 99%
When testing out different stamina values for https://github.com/space-wizards/space-station-14/pull/34890 I tested the default value of 33.

2 hits gives you slowdown like expected.
3 hits (99 stamina damage) resets your speed back to default
4th hit will stun.

The 3rd hit should be extreme slow, and this makes sense in part of why disablers are downright awful to use.

https://github.com/user-attachments/assets/f21210e4-9149-41ef-9381-c982dda1b8ad

nikitosych:
## About the PR

**Fixes #34895**

This PR resolves the problem with slowdown abruptly resetting when stamina damage is applied to an entity and reaching a damage value of 99/100. Also, the abrupt zeroing of damage after leaving a crit is replaced with gradual damage regeneration. 

This is my first PR, but I have spent a lot of time on it, so I apologize for any code errors, unfinished work, or accidental violations of PR guidelines. I will definitely take any feedback into account and will be grateful for any comments, suggestions, or corrections. Thank you for your attention! :)

## Why / Balance

Previously, stamina slowdown was inconsistently handled: entities that reached exactly 100% stamina damage (entering crit) would have their movement speed **instantly restored** upon exiting crit, while their stamina damage still slowly regenerated. This created unintuitive behavior and gave players inconsistent movement feedback.

## Technical details

This PR significantly modifies 3 classes:

1. **`StaminaSystem`** from `Content.Shared.Damage.Systems`:

- **`public void TakeStaminaDamage (...)`:**
-- In the `TakeStaminaDamage` method, before the fix, when taking damage and the `StaminaDamage` value reached certain thresholds, the `SlowedDownComponent` component-effect was temporarily added to the entity for 3 seconds. After three seconds, the effect was removed and the speed was reset to the base value. However, for several sequential hits, the effect did not add time or anything else to the effect, but rather 3 seconds from the first hit. This would cause us to reset the speed to the base speed at 99 damage. So instead of adding an effect, entity slowdown is now directly adjustable using the new `AdjustSlowdown` method explained below.

- **`public override void Update(...)`:**
-- The `Update` method is called every game tick and if there is no new stamina damage for a long time, it restores StaminaDamage (and consequently speed) by calling `TakeStaminaDamage` with a negative `Decay` value of the `StaminaComponent` component. If the recovery should occur after the exit from StamCrit (`StaminaComponent.AfterCritical = true`), then `Decay` is multiplied by the `StaminaComponent.AfterCritDecayMultiplier`. 
-- When working with the Update method, I encountered a `System.InvalidOperationException` error when modifying the source block `while (query.MoveNext(out var uid, out _))`. It was due to the fact that we are trying to iterate over a collection while modifying it. This violates the enumeration rules in .NET, which require the collection to remain unchanged while iterating. To solve this problem, I created a temporary list of enumerated entities, ensuring that changes to the original collection would not affect the iteration. 

- **`private void AdjustSlowdown(...)`:**  
-- This helper method dynamically adjusts the slowdown effect on an entity based on the current `StaminaDamage` value of the `StaminaComponent`. It replaces the previous time-limited slowdown behavior with a more granular system that scales based on damage thresholds.  
-- The method retrieves the `StaminaComponent` from the entity to access a dictionary of speed modifier thresholds (i.e., mappings from damage levels to slowdown factors).  
-- It then iterates through the thresholds to find the closest applicable damage threshold that is still below the next value in the threshold dictionary. If a suitable threshold is found, it applies the corresponding slowdown modifier using the `_stunSystem.UpdateStunModifiers` method.  

2. **`SharedStunSystem`** from `Content.Shared.Stunnable`:

- **`public void UpdateStunModifiers(…)`:**
-- This method was invented to replace `TrySlowdown`, which applied a slowdown effect for 3 seconds. I explained why the former method was a problem above. When called, the method updates or adds a slowdown modifier entry for the given `EntityUid` in the internal `_slowdowns` dictionary, using the provided `modifier` value. If a modifier of `1f` is passed, it signifies a full recovery, and any existing slowdown entry for the entity is removed instead.
-- The `modifier` parameter represents the percentage of the entity's original speed (e.g., `0.5f` = 50% speed).
-- This change allows the `StaminaSystem` to scale the movement penalty smoothly as stamina damage accumulates, replacing the older model where slowdown was temporarily applied as a timed effect regardless of stacking hits.

3. **`StaminaComponent`** from `Content.Shared.Damage.Components`:

- Added a component field `AfterCritical` to indicate whether the `StaminaDamage` field of the `StaminaComponent` component decreases after the character exits stamcrit (completion of the `ExitStamCrit` method). Defaults to `false` initially, set to `true` in the ExitStamCrit method and to `false` in the `TakeStaminaDamage` method (provided that `StaminaDamage` is restored to 0)
- Added a component field `AfterCritDecayMultiplier` that represents the multiplier applied to the decay after leaving stamina crit in the `Update(...)` method
- Added dictionary `StunModifierThresholds` that determine an entity's slowdown as a function of stamina damage.

## Media

[**_Before fix_**](https://i.imgur.com/DHbyRmC.gif) (imgur)

[**_After fix_**](https://i.imgur.com/Yx7sFGz.gif) (imgur)

## Requirements

## Changelog

:cl: Nikitosych
- fix: Fixed abrupt stamina damage reset and abrupt speed recovery after 3 seconds (especially noticeable at high damage levels)
- tweak: Stamina damage no longer instantly resets when exiting critical state; it now regenerates gradually.


nikitosych:
Oh I also made some videos with tests, I can upload them if anyone needs

keronshb:
I'll grab the review for this one too.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -262,23 +263,25 @@ public void TakeStaminaDamage(EntityUid uid, float value, StaminaComponent? comp
         // Reset the decay cooldown upon taking damage.
         if (oldDamage < component.StaminaDamage)
         {
+            
             var nextUpdate = _timing.CurTime + TimeSpan.FromSeconds(component.Cooldown);
-
+
```
Pronana, line 268: Whitespace issues

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -330,7 +333,14 @@ public override void Update(float frameTime)
         var query = EntityQueryEnumerator<ActiveStaminaComponent>();
         var curTime = _timing.CurTime;
 
+        var activeEntities = new List<EntityUid>();
         while (query.MoveNext(out var uid, out _))
+        {
+            activeEntities.Add(uid);
+        }
+
```
Pronana, line 341: Query is fine we don't need to make another list to iterate over.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -346,15 +356,21 @@ public override void Update(float frameTime)
...
                 ExitStamCrit(uid, comp);
-                continue;
-            }
+            
 
             comp.NextUpdate += TimeSpan.FromSeconds(1f);
-            TakeStaminaDamage(uid, -comp.Decay, comp);
+
+            // This constant determines how fast stamina will regenerate after exiting the stamcrit
+            const float afterCritDecayMultiplier = 5f;
```
Pronana, line 367: Values like this should be located in components and not systems that way they're easily editable by downstream.

Pronana (changes requested):
Partial Review, need to properly open this in my IDE to analyze the code better.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: > Whitespace issues

> -- When working with the Update method, I encountered a System.InvalidOperationException error when modifying the source block while (query.MoveNext(out var uid, out _)). It was due to the fact that we are trying to iterate over a collection while modifying it. This violates the enumeration rules in .NET, which require the collection to remain unchanged while iterating. To solve this problem, I created a temporary list of enumerated entities, ensuring that changes to the original collection would not affect the iteration.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: > Query is fine we don't need to make another list to iterate over.

Please familiarise yourself with the reasons why I had to make a copy of the collection. Maybe you know how to avoid the errors in another way? This approach has almost no impact on performance.

>> -- When working with the Update method, I encountered a System.InvalidOperationException error when modifying the source block while (query.MoveNext(out var uid, out _)). It was due to the fact that we are trying to iterate over a collection while modifying it. This violates the enumeration rules in .NET, which require the collection to remain unchanged while iterating. To solve this problem, I created a temporary list of enumerated entities, ensuring that changes to the original collection would not affect the iteration.

nikitosych:
> Partial Review, need to properly open this in my IDE to analyze the code better.

Thanks so much for the quick look at the PR! I'm really looking forward to hearing more of your thoughts!

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: > Values like this should be located in components and not systems that way they're easily editable by downstream.

So should I move `afterCritDecayMultiplier` and `_afterCrit` to the component?

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -45,6 +43,9 @@ public sealed partial class StaminaSystem : EntitySystem
 
     public float UniversalStaminaDamageModifier { get; private set; } = 1f;
 
+    // A flag indicating whether the entity recovers after exiting a crit
+    private bool _afterCrit = false;
+
```
Pronana, line 48: This should be stored in a component, I tested to be sure and if you stamcrit someone else your stamina recovery also gets multiplied even if you aren't stamcrit. This is a pretty major issue.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +407,57 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
         }
```
Pronana, line 405: Can't comment on EnterStamCrit, but the entire buffer system is going to need to be reworked. Currently when you exit stamcrit you're slowed and because you don't regenerate to 100% stamina you can immediately be insta stunned again by any amount of stamina damage.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +407,57 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
...
 
         component.Critical = false;
-        component.StaminaDamage = 0f;
+        _afterCrit = true;  // Set to true to indicate that stamina will be restored after exiting stamcrit
```
Pronana, line 408: There should be logic to block incoming stamina damage for a time if this is true.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +407,57 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
...
+    /// </param>
+    private void AdjustSlowdown(EntityUid uid, StaminaComponent? comp = null)
+    {
+        if (!Resolve(uid, ref comp))
+            return;
+
+        // Use a dictionary with default 'damage - speed' pairs if the entity doesn't have a component with an already existing one
+        var thresholds = HasComp<SlowOnDamageComponent>(uid)
+            ? Comp<SlowOnDamageComponent>(uid).SpeedModifierThresholds
+            : new Dictionary<FixedPoint2, float> { { 60, 0.7f }, { 80, 0.5f } };
```
Pronana, line 436: You should have your own thresholds for stamina rather than using the same ones from the SlowOnDamageComponent. You can probably store them in that component since they're extremely similar.

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,69 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+    /// </summary>
+    /// <param name="uid">Entity whose movement speed should be updated.</param>
+    /// <param name="walkSpeedModifier">New walk speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="runSpeedModifier">New run (sprint) speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="component">
+    /// Optional <see cref="StaminaComponent"/> of the entity.
+    /// </param>
+
+    public void UpdateStunModifiers(EntityUid uid,
+        float walkSpeedModifier = 1f,
```
Pronana, line 273: This does not apply correctly if you stamcrit an entity and then rejuvenate them.

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,69 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+        float runSpeedModifier = 1f, StaminaComponent? component = null)
+    {
+        if (!Resolve(uid, ref component))
+            return;
+
+        bool Approximately(float a, float b, float epsilon = 0.0001f) =>
+            Math.Abs(a - b) < epsilon; // Comparing two floats, taking into account the error (epsilon)
```
Pronana, line 280: You can probably use MathHelper.CloseTo(float a, float b) here.

Pronana (changes requested):
Booted it up in my IDE and found a few more issues to be addressed.

Content.Shared/Damage/Systems/StaminaSystem.cs
Pronana: Yes

Content.Shared/Stunnable/SharedStunSystem.cs
Pronana: Also this appears to not be predicted properly? Unfortunately I gotta start my day but I ran into prediction issues that I didn't run into on master. I'd need to properly debug it to be sure.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: > You should have your own thresholds for stamina 

All right. Any idea what exactly the thresholds should be for stamina damage?

Content.Shared/Stunnable/SharedStunSystem.cs
nikitosych: Can you elaborate on what exactly is wrong with prediction? How does the issue manifest itself?

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: You will help me a lot if you can tell me how to reproduce this bug step by step.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: So when an entity leaves the crit state, it shouldn't take damage while it recovers? Why is this a problem? While testing this locally, another player was dealing damage to me while I was recovering from stamcrit and no problems with that were detected.

Pronana: I believe Keron wants to avoid being able to easily chain stamina crits, so after leaving stamcrit you should have some buffer time to regenerate your stamina at least until full movement speed.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: Fixed that.

Content.Shared/Stunnable/SharedStunSystem.cs
Pronana: When I would get stamcrit and start moving, I would get a small client/server mispredict where my speed would jump up immediately.

You can use "showplayervelocity" in the console to see this effect. Normally when moving from like glue to a solid surface there's a small but quick ramp up of speed once you step off.

With stamina, the changes are abrupt. I'd need to debug it more but it feels like the client isn't predicting the modifiers properly and is instead relying on the server to send them.

Content.Shared/Damage/Systems/StaminaSystem.cs
Pronana: Stamina crit yourself with a stun baton, then pick it up and click on yourself the moment you start regenerating stamina and you'll get stunned again.

Content.Shared/Damage/Systems/StaminaSystem.cs
Pronana: I'd need to debug it myself but for now it's fine to leave as is to avoid breaking things.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: @keronshb What do you think about it? I'd be grateful for your opinion

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: @keronshb

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: During recovery, the player can only fall again if his `StaminaDamage` has reached 100 again. At least during testing I was not able to confirm the opposite.

https://youtu.be/JCIZhM86Jyw

Content.Shared/Stunnable/SharedStunSystem.cs
nikitosych: > You can use "showplayervelocity" in the console to see this effect. Normally when moving from like glue to a solid surface there's a small but quick ramp up of speed once you step off.

I'm really sorry, but I haven't been able to reproduce this issue. Slowdown is working just as it should with the current code. When it hits a certain threshold, slowdown is applied. And when you step in a puddle of glue, the entity slows down even more. Would you mind recording a video of this bug?

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -102,7 +100,7 @@ public float GetStaminaDamage(EntityUid uid, StaminaComponent? component = null)
 
         var curTime = _timing.CurTime;
         var pauseTime = _metadata.GetPauseTime(uid);
-        return MathF.Max(0f, component.StaminaDamage - MathF.Max(0f, (float) (curTime - (component.NextUpdate + pauseTime)).TotalSeconds * component.Decay));
+        return MathF.Max(0f, component.StaminaDamage - MathF.Max(0f, (float)(curTime - (component.NextUpdate + pauseTime)).TotalSeconds * component.Decay));
```
keronshb, line 103: As a heads up you don't need to make space changes like this when you add something. Makes the review a little harder.

Not the end of the world but just letting you know.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -218,7 +217,7 @@ private void SetStaminaAlert(EntityUid uid, StaminaComponent? component = null)
             return;
 
         var severity = ContentHelpers.RoundToLevels(MathF.Max(0f, component.CritThreshold - component.StaminaDamage), component.CritThreshold, 7);
-        _alerts.ShowAlert(uid, component.StaminaAlert, (short) severity);
+        _alerts.ShowAlert(uid, component.StaminaAlert, (short)severity);
```
keronshb, line 220: Same here, you don't need to make adjustments like this.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -265,20 +264,21 @@ public void TakeStaminaDamage(EntityUid uid, float value, StaminaComponent? comp
             var nextUpdate = _timing.CurTime + TimeSpan.FromSeconds(component.Cooldown);
 
             if (component.NextUpdate < nextUpdate)
+            {
                 component.NextUpdate = nextUpdate;
+            }
```
keronshb, line 269: Here as well.

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -56,4 +56,16 @@ public sealed partial class StaminaComponent : Component
 
     [DataField]
     public ProtoId<AlertPrototype> StaminaAlert = "Stamina";
+
+    /// <summary>
+    /// This flag indicates whether the value of <see cref="StaminaDamage"/> decreases after the entity exits stamina crit.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite), AutoNetworkedField]
+    public bool AfterCritical = false;
```
keronshb, line 64: Needs [DataField]

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -56,4 +56,16 @@ public sealed partial class StaminaComponent : Component
...
+
+    /// <summary>
+    /// This float determines how fast stamina will regenerate after exiting the stamina crit.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite)]
+    public float AfterCritDecayMultiplier = 5f;
```
keronshb, line 70: Also needs [DataField]

Content.Shared/Damage/Systems/StaminaSystem.cs
keronshb: What was it throwing on? It doesn't look like logic was changed outside of removing a continue and adjusting take stamina damage.

Content.Shared/Damage/Systems/StaminaSystem.cs
keronshb: I'd need to think about it. I don't think it should be tied to SlowOnDamage here and thresholds should be put into the Stamina Comp.

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,65 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+    /// </summary>
+    /// <param name="uid">Entity whose movement speed should be updated.</param>
+    /// <param name="speedModifier">New walk and run speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="component">
+    /// Optional <see cref="StaminaComponent"/> of the entity.
+    /// </param>
+    public void UpdateStunModifiers(EntityUid uid, float speedModifier = 1f, StaminaComponent? component = null)
+    {
+        UpdateStunModifiers(uid, speedModifier, speedModifier, component);
+    }
```
keronshb, line 309: Neither of these are needed. The Stun System already takes care of this.
There's a line for it in TakeStaminaDamage method already:
_stunSystem.TrySlowdown(uid, TimeSpan.FromSeconds(3), true, 0.8f, 0.8f);

keronshb (changes requested):
Partial review.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: I'm sorry, gonna fix that

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: I think that's done by my IDE 😅

Content.Shared/Damage/Components/StaminaComponent.cs
nikitosych: > Needs [DataField]

Why? This field should not be modifiable explicitly through the prototype

keronshb: It's not just for the prototype, it's for replays and other things.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: @keronshb 

> What was it throwing on? It doesn't look like logic was changed outside of removing a continue and adjusting take stamina damage.

Here's the answer to your question.

Content.Shared/Stunnable/SharedStunSystem.cs
nikitosych: > Neither of these are needed. The Stun System already takes care of this. There's a line for it in TakeStaminaDamage method already: _stunSystem.TrySlowdown(uid, TimeSpan.FromSeconds(3), true, 0.8f, 0.8f);

True, but because `TrySlowdown` imposes the effect for 3 seconds, it causes the `SlowedDownComponent` to be removed after three seconds and the player's speed is abruptly reset to `1f` with no smooth increase.

To quote the text of my PR:
> public void TakeStaminaDamage (...):
> -- In the TakeStaminaDamage method, before the fix, when taking damage and the StaminaDamage value reached certain thresholds, the SlowedDownComponent component-effect was temporarily added to the entity for 3 seconds. After three seconds, the effect was removed and the speed was reset to the base value. However, for several sequential hits, the effect did not add time or anything else to the effect, but rather 3 seconds from the first hit. This would cause us to reset the speed to the base speed at 99 damage. So instead of adding an effect, entity slowdown is now directly adjustable using the new AdjustSlowdown method explained below.
>
>public void UpdateStunModifiers(…):
> -- This method was invented to replace TrySlowdown, which applied a slowdown effect for 3 seconds. I explained why the former method was a problem above. When called, the method updates or adds a slowdown modifier entry for the given EntityUid in the internal _slowdowns dictionary, using the provided modifier value. If a modifier of 1f is passed, it signifies a full recovery, and any existing slowdown entry for the entity is removed instead.
> -- The modifier parameter represents the percentage of the entity's original speed (e.g., 0.5f = 50% speed).
> -- This change allows the StaminaSystem to scale the movement penalty smoothly as stamina damage accumulates, replacing the older model where slowdown was temporarily applied as a timed effect regardless of stacking hits.

keronshb: That sounds reasonable then, I thought we might've had a method that applies a status effect without a timer to handle it.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -346,15 +351,17 @@ public override void Update(float frameTime)
             if (nextUpdate > curTime)
                 continue;
 
-            // We were in crit so come out of it and continue.
+            // Handle exiting critical condition and restoring stamina damage
             if (comp.Critical)
-            {
                 ExitStamCrit(uid, comp);
```
keronshb, line 356: Consider here that ExitStamCrit sets comp.NextUpdate to current time. Without the continue, this immediately gets set again a few lines down in line 358.

Also I see the issue where ExitStamCrit removes the ActiveStaminaDamage component. 
This shouldn't be removing the ActiveStaminaDamage component anymore since ExitStamCrit is no longer setting StaminaComponent damage to 0.

You should only be removing ActiveStaminaComponent when stamina damage hits 0. Change that, and then you don't need to make another list to reiterate over.

keronshb (changes requested):
Some more changes that I noticed.

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: Fixed successfully

keronshb:
https://github.com/user-attachments/assets/8ea7a4ba-bb67-4b10-b61d-b65ae6905399

There's something going on here that I can't quite pinpoint down yet. Not entirely sure if it's a mispredict.

If you stun someone, there looks to be a grace period after stamcrit is up before they can be batonned again. This is a little inconsistent. The video does have the baton going off, but there was a non-recorded instance where I had a fully charged baton and I was swinging at the other player after stamcrit and they wouldn't take stamina damage until they were fully recovered.

Ideally what this should be is after stun, they should be allowed to be stam-critted again when in the recovery phase.

Content.Shared/Damage/Systems/StaminaSystem.cs
keronshb: Check out my latest comment.

Content.Shared/Stunnable/SharedStunSystem.cs
keronshb: @Pronana this seems fixed with the latest changes, can you double check?

ArtisticRoomba:
Isn't that an intentional feature in SS14? You're immune to stamina damage for a few seconds after you get up from a stamina crit.

Pronana:
> If you stun someone, there looks to be a grace period after stamcrit is up before they can be batonned again. This is a little inconsistent. The video does have the baton going off, but there was a non-recorded instance where I had a fully charged baton and I was swinging at the other player after stamcrit and they wouldn't take stamina damage until they were fully recovered.

> Ideally what this should be is after stun, they should be allowed to be stam-critted again when in the recovery phase.

image

It's intentional to prevent chain stunning.

Content.Shared/Stunnable/SharedStunSystem.cs
Pronana: Still happening on my end. I'm going to try and debug it myself.

Pronana: You get the same mispredict when recovering from damage slowdown on normal damage so it's probably not worth worrying about unless you can solve it easily.

nikitosych:
> Content.Client.Wx94cprrvy.mp4 
> If you stun someone, there looks to be a grace period after stamcrit is up before they can be batonned again. This is a little inconsistent. The video does have the baton going off, but there was a non-recorded instance where I had a fully charged baton and I was swinging at the other player after stamcrit and they wouldn't take stamina damage until they were fully recovered.

I noticed similar behaviour at 0:09 - 0:17.  Is that what you mean?

It's never happened to me, though I've tested hitting myself with the baton numerous times, and such an annoying bug I'd definitely notice.

Content.Shared/Stunnable/SharedStunSystem.cs
nikitosych: > You get the same mispredict when recovering from damage slowdown on normal damage so it's probably not worth worrying about unless you can solve it easily.

Are you sure the same issue doesn't happen on `master`?

Pronana: image

Source of the mispredicts is this return statement here. I'd talk with sloth about it before making adjustments though.

keronshb:
> > Content.Client.Wx94cprrvy.mp4
> > If you stun someone, there looks to be a grace period after stamcrit is up before they can be batonned again. This is a little inconsistent. The video does have the baton going off, but there was a non-recorded instance where I had a fully charged baton and I was swinging at the other player after stamcrit and they wouldn't take stamina damage until they were fully recovered.
> 
> I noticed similar behaviour at 0:09 - 0:17. Is that what you mean?
> 
> It's never happened to me, though I've tested hitting myself with the baton numerous times, and such an annoying bug I'd definitely notice.

You may need to open two clients to test it.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +391,56 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
...
+    /// </param>
+    private void AdjustSlowdown(EntityUid uid, StaminaComponent? comp = null)
+    {
+        if (!Resolve(uid, ref comp))
+            return;
+
+        // Use a dictionary with default 'damage - speed' pairs if the entity doesn't have a component with an already existing one
+        var thresholds = HasComp<SlowOnDamageComponent>(uid)
+            ? Comp<SlowOnDamageComponent>(uid).SpeedModifierThresholds
+            : comp.StunModifierThresholds;
```
keronshb, line 421:
```suggestion
```

Remove this and just use the comp thresholds.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +391,56 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
...
+
+        var closest = FixedPoint2.Zero;
+
+        // Iterate through the dictionary in the similar way as in Damage.SlowOnDamageSystem.OnRefreshMovespeed
+        foreach (var thres in thresholds)
```
keronshb, line 426:
```suggestion
        // Iterate through the dictionary in the similar way as in Damage.SlowOnDamageSystem.OnRefreshMovespeed
        foreach (var thres in comp.StunModifierThresholds)
```

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +391,56 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
...
+        if (closest != FixedPoint2.Zero)
+        {
+            _stunSystem.UpdateStunModifiers(uid, thresholds[closest], comp);
+        }
+        else if (closest == FixedPoint2.Zero)
+        {
+            // Recover the full speed if the closest threshold is zero
+            _stunSystem.UpdateStunModifiers(uid, 1f, comp);
+        }
+    }
```
keronshb, line 443: Instead of doing this, just add another threshold for 0 so you can restore back to normal.

keronshb (changes requested):

Content.Shared/Damage/Systems/StaminaSystem.cs
nikitosych: Makes sense, gonna change that

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -56,4 +57,22 @@ public sealed partial class StaminaComponent : Component
...
+    /// This float determines how fast stamina will regenerate after exiting the stamina crit.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite), DataField]
+    public float AfterCritDecayMultiplier = 5f;
+
+    /// <summary>
+    /// Thresholds that determine an entity's slowdown as a function of stamina damage. Only used if thresholds are not defined for the same entity in <see cref="SlowOnDamageComponent"/>
+    /// </summary>
+    [ViewVariables(VVAccess.ReadOnly), DataField("stunModifierThresholds")]
+    public Dictionary<FixedPoint2, float> StunModifierThresholds = new() { { 60, 0.7f }, { 80, 0.5f } };
```
keronshb, line 77:
```suggestion
    /// <summary>
    /// Thresholds that determine an entity's slowdown as a function of stamina damage.
    /// </summary>
    [ViewVariables(VVAccess.ReadOnly), DataField("stunModifierThresholds")]
    public Dictionary<FixedPoint2, float> StunModifierThresholds = new() { { 60, 0.7f }, { 80, 0.5f } };
```

Forgot this in the earlier review but you can remove the Slow on Damage Comp part here

nikitosych:
> You may need to open two clients to test it.

> Ideally what this should be is after stun, they should be allowed to be stam-critted again when in the recovery phase.

Followed this up on two clients. Indeed, when a character leaves the stamkrit, there is a short period of time where they don't take stamina damage. As Pronana pointed out, this is done on purpose. However, we can allow the character to allow taking damage immediately by this change:

```csharp
// Give them buffer before being able to be re-stunned
component.NextUpdate = _timing.CurTime + component.StunTime; // + StamCritBufferTime;
```

keronshb:
> Isn't that an intentional feature in SS14? You're immune to stamina damage for a few seconds after you get up from a stamina crit.

I need to compare it against master but I probably want it changed at this point since you're already immune to getting stun-stacked.

keronshb:
> > You may need to open two clients to test it.
> 
> > Ideally what this should be is after stun, they should be allowed to be stam-critted again when in the recovery phase.
> 
> Followed this up on two clients. Indeed, when a character leaves the stamkrit, there is a short period of time where they don't take stamina damage. As Pronana pointed out, this is done on purpose. However, we can allow the character to allow taking damage immediately by this change:
> 
> ```cs
> // Give them buffer before being able to be re-stunned
> component.NextUpdate = _timing.CurTime + component.StunTime; // + StamCritBufferTime;
> ```

I think we might want to make this change. Since right now they already have a buffer when they were stunned (stun doesn't stack).

For now just leave it in, I need to run more tests to see if the buffer ever messes up again like it did the one time for me.

Pronana:
> I think we might want to make this change. Since right now they already have a buffer when they were stunned (stun doesn't stack).
> 
> For now just leave it in, I need to run more tests to see if the buffer ever messes up again like it did the one time for me.

Stuns are already very strong and easy to apply, plus making it so stamina regenerates after being crit rather than jumping to 100 means once the buffer runs out you can already be instantly stunned again. 

Getting rid of the buffer means you can stunlock a person as long as you have charge, which is a very very long time.

keronshb:
> > I think we might want to make this change. Since right now they already have a buffer when they were stunned (stun doesn't stack).
> > For now just leave it in, I need to run more tests to see if the buffer ever messes up again like it did the one time for me.
> 
> Stuns are already very strong and easy to apply, plus making it so stamina regenerates after being crit rather than jumping to 100 means once the buffer runs out you can already be instantly stunned again.
> 
> Getting rid of the buffer means you can stunlock a person as long as you have charge, which is a very very long time.

I'm coming from servers where this was normal.

We'll see how it goes with the buffer, still need to do some testing since i felt like the buffer expanded if i kept smacking someone while they were already stunned.

Content.Shared/Stunnable/SharedStunSystem.cs
keronshb: @nikitosych  or @Pronana have either of you talked to sloth about this? Was the IsFirstTimepredicted in this method or elsewhere?

nikitosych: > @nikitosych or @Pronana have either of you talked to sloth about this? Was the IsFirstTimepredicted in this method or elsewhere?

No, I haven't yet. Maybe we should ping him and hook him up to PR.

keronshb:
So far the only thing that feels a little jank is the grace period after someone recovers, since there's nothing that visually shows a person is immune to stam damage for an extra few seconds after getting up. I feel we can deal, but also another maint needs to give it a pass anyway.

I don't feel like I'm getting any prediction issues.

keronshb (approved):

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -56,4 +57,22 @@ public sealed partial class StaminaComponent : Component
 
     [DataField]
     public ProtoId<AlertPrototype> StaminaAlert = "Stamina";
+
+    /// <summary>
+    /// This flag indicates whether the value of <see cref="StaminaDamage"/> decreases after the entity exits stamina crit.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite), AutoNetworkedField, DataField]
```
slarticodefast, line 64:
```suggestion
    [DataField, AutoNetworkedField]
```
datafield includes VV readwrite nowadays, so this is redundant

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -56,4 +57,22 @@ public sealed partial class StaminaComponent : Component
...
+    public bool AfterCritical;
+
+    /// <summary>
+    /// This float determines how fast stamina will regenerate after exiting the stamina crit.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadWrite), DataField]
```
slarticodefast, line 70:
```suggestion
    [DataField, AutoNetworkedField]
```
same
and can't hurt to have this networked

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -56,4 +57,22 @@ public sealed partial class StaminaComponent : Component
...
+    public float AfterCritDecayMultiplier = 5f;
+
+    /// <summary>
+    /// Thresholds that determine an entity's slowdown as a function of stamina damage.
+    /// </summary>
+    [ViewVariables(VVAccess.ReadOnly), DataField("stunModifierThresholds")]
```
slarticodefast, line 76:
```suggestion
    [DataField]
```
the name is automatically converted from PascalCase to camelCase in yaml
old files still do this manually
your IDE should give you a warning for these, don't ignore it!

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +391,43 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
...
+    /// If the entity has a <see cref="SlowOnDamageComponent"/>, its custom damage-to-speed thresholds are used,
+    /// otherwise, a default set of thresholds is applied.
+    /// The method determines the closest applicable damage threshold below the crit limit and applies the corresponding
+    /// speed modifier using the stun system. If no threshold is met then the entity's speed is restored to normal.
+    /// </summary>
+    /// <param name="uid">Entity to update</param>
+    /// <param name="comp">
+    /// Stamina component of the entity. Can be null, in this case the method attempts to resolve it automatically.
+    /// </param>
+    private void AdjustSlowdown(EntityUid uid, StaminaComponent? comp = null)
```
slarticodefast, line 413: use `Entity<Component?>` pattern

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,65 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+    /// and the movement speed is refreshed accordingly.
+    /// </summary>
+    /// <param name="uid">Entity whose movement speed should be updated.</param>
+    /// <param name="walkSpeedModifier">New walk speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="runSpeedModifier">New run (sprint) speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="component">
+    /// Optional <see cref="StaminaComponent"/> of the entity.
+    /// </param>
+
+    public void UpdateStunModifiers(EntityUid uid,
```
slarticodefast, line 272: use `Entity<Component?>` pattern

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,65 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+        {
+            RemComp<SlowedDownComponent>(uid);
+            return;
+        }
+
+        EnsureComp<SlowedDownComponent>(uid, out var comp);
+
+        comp.WalkSpeedModifier = walkSpeedModifier;
+
+        comp.SprintSpeedModifier = runSpeedModifier;
```
slarticodefast, line 292: doesn't this need dirtying?

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,65 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+    /// <summary>
+    /// A convenience overload of <see cref="UpdateStunModifiers(EntityUid, float, float, StaminaComponent?)"/> that sets both
+    /// walk and run speed modifiers to the same value.
+    /// </summary>
+    /// <param name="uid">Entity whose movement speed should be updated.</param>
+    /// <param name="speedModifier">New walk and run speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="component">
+    /// Optional <see cref="StaminaComponent"/> of the entity.
+    /// </param>
+    public void UpdateStunModifiers(EntityUid uid, float speedModifier = 1f, StaminaComponent? component = null)
```
slarticodefast, line 306: use `Entity<Component?>` pattern

slarticodefast (changes requested):
Got a few comments.
I did not look at this in detail, nor did I test this, so this is not a full review.

Content.Shared/Stunnable/SharedStunSystem.cs
Pronana: > Was the IsFirstTimepredicted in this method or elsewhere?

@keronshb 
It's used in the update loop before the active stamina component enumerator to prevent it from running multiple times for the client. 

It should be fine to remove unless the update loop is sending network events to the server.

Content.Shared/Stunnable/SharedStunSystem.cs
nikitosych: > doesn't this need dirtying?

It is dirtyied later on `_movementSpeedModifier.RefreshMovementSpeedModifiers(uid)` call

```csharp
public void RefreshMovementSpeedModifiers(EntityUid uid, MovementSpeedModifierComponent? move = null)
{
    if (!Resolve(uid, ref move, false))
        return;

    if (_timing.ApplyingState)
        return;

    var ev = new RefreshMovementSpeedModifiersEvent();
    RaiseLocalEvent(uid, ev);

    if (MathHelper.CloseTo(ev.WalkSpeedModifier, move.WalkSpeedModifier) &&
        MathHelper.CloseTo(ev.SprintSpeedModifier, move.SprintSpeedModifier))
        return;

    move.WalkSpeedModifier = ev.WalkSpeedModifier;
    move.SprintSpeedModifier = ev.SprintSpeedModifier;
    Dirty(uid, move); // <-- here
}
```

slarticodefast: That dirties `MovementSpeedModifierComponent`, but not `SlowedDownComponent`, which has the changed datafields

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,65 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+    /// and the movement speed is refreshed accordingly.
+    /// </summary>
+    /// <param name="uid">Entity whose movement speed should be updated.</param>
+    /// <param name="walkSpeedModifier">New walk speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="runSpeedModifier">New run (sprint) speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="component">
+    /// Optional <see cref="StaminaComponent"/> of the entity.
+    /// </param>
+
+    public void UpdateStunModifiers(Entity<StaminaComponent?> uid,
```
slarticodefast, line 272:
```suggestion
    public void UpdateStunModifiers(Entity<StaminaComponent?> ent,
```
it's not an uid, so please rename these to fit the naming scheme we use everywhere else
same for the other cases

slarticodefast (dismissed):

Content.Shared/Stunnable/SharedStunSystem.cs
nikitosych: Done

Content.Shared/Stunnable/SharedStunSystem.cs
keronshb: @nikitosych can you try removing this if you haven't yet?

nikitosych: > @nikitosych can you try removing this if you haven't yet?

@keronshb You mean removing the `IsFirstTimePredicted`?

nikitosych:
Yes, it looks like the predict works more clearly without returning from the update cycle on the first predict. I'll comment out those lines for now.

```csharp
if (!_timing.IsFirstTimePredicted)
    return;
```

nikitosych:
Bruh now it causes errors

keronshb:
> Bruh now it causes errors

Looks like a fake test, Ill rerun.

nikitosych:
Maybe merge? 👉👈

Content.Shared/Stunnable/SharedStunSystem.cs
Pronana: All my issues seem to be resolved from what I can tell.

Content.Shared/Damage/Systems/StaminaSystem.cs
```diff
@@ -391,13 +391,43 @@ private void ExitStamCrit(EntityUid uid, StaminaComponent? component = null)
...
+
+    /// <summary>
+    /// Adjusts the movement speed of an entity based on its current <see cref="StaminaComponent.StaminaDamage"/> value.
+    /// If the entity has a <see cref="SlowOnDamageComponent"/>, its custom damage-to-speed thresholds are used,
+    /// otherwise, a default set of thresholds is applied.
+    /// The method determines the closest applicable damage threshold below the crit limit and applies the corresponding
+    /// speed modifier using the stun system. If no threshold is met then the entity's speed is restored to normal.
+    /// </summary>
+    /// <param name="ent">Entity to update</param>
+    /// <param name="comp">
```
ScarKy0, line 410: Not needed param since you just use entity.

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -248,13 +249,67 @@ public bool TrySlowdown(EntityUid uid, TimeSpan time, bool refresh,
...
+    /// or if the both modifiers are 0, the slowdown component is removed to restore normal movement.
+    /// Otherwise, the slowdown component is created or updated with the provided modifiers,
+    /// and the movement speed is refreshed accordingly.
+    /// </summary>
+    /// <param name="ent">Entity whose movement speed should be updated.</param>
+    /// <param name="walkSpeedModifier">New walk speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="runSpeedModifier">New run (sprint) speed modifier. Default is 1f (normal speed).</param>
+    /// <param name="component">
+    /// Optional <see cref="StaminaComponent"/> of the entity.
+    /// </param>
```
ScarKy0, line 270: No component field in the function below.

ScarKy0 (changes requested):
Minor stuff. I'll push a commit to address those myself.

ScarKy0 (approved):
Looks good to me.

ScarKy0:
I also fixed the CL for you, it lacked - before `add` and `tweak`

nikitosych:
OH YES!!! I'm very pleased with the work done, huge thanks to everyone who took part in the review and helped bring it all to completion!

PR #36013, opened 2025-03-22
Fixed some unintended salvage changes.

RedBookcase:
## About the PR

Noticed that 2 recent PR's made some unintended item changes, specifically for salvage.

- A PR adding inhand sprites to various items unintentionally made the Handheld Mass Scanner a 2x2 instead of a 1x2. #33689 
- A PR editing Security contraband tags swapped around the parents on the Survival Knife and Kukri, meaning the Combat Knife parent made them appear as Security contraband instead of Cargo/Security Contraband. #35281 

## Why / Balance
No balance issues, just bugfixing.

## Technical details
A few lines of YAML to make the Handheld Mass Scanner a small item and swapping around the parents on the Survival Knife and Kukri

## Media
N/A

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
N/A

**Changelog**
:cl:
- fix: Fixed some unintended changes to Salvage gear.

PR #36340, opened 2025-04-06
Fixed the empty mass scanner still being a normal item, whoops.

RedBookcase:
## About the PR
Finishes what #36013 started. I forgor that there's a separate yaml entry for an empty version of the handheld mass scanner, which is the one printed by science. So currently there's a 2x2 mass scanner and a 1x2 mass scanner in game. oops.

Additionally, it looks like an issue with the sprite layering was causing the empty version of the Handheld to show it's screen as active despite having no battery, which has also been fixed as part of the cleanup.

## Why / Balance
Finishing what I started.

## Technical details
Finishes removal of the Normal size category from both the regular handheld mass scanner and the empty version, thus fixing their unintentional size increase.

## Requirements

**Changelog**
:cl:
- fix: Fixed empty handheld mass scanners being 2x2.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Tools/handheld_mass_scanner.yml
+++ b/Resources/Prototypes/Entities/Objects/Tools/handheld_mass_scanner.yml
@@ -13,7 +13,7 @@
       - state: icon
       - state: scanner
         shader: unshaded
-        visible: true
+        visible: false
         map: [ "enum.PowerDeviceVisualLayers.Powered" ]
   - type: RadarConsole
     maxRange: 256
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
back at it again

Resources/Prototypes/Entities/Objects/Tools/handheld_mass_scanner.yml
```diff
@@ -50,7 +49,6 @@
       cell_slot:
         name: power-cell-slot-component-slot-name-default
   - type: Item
```
EmoGarbage404, line 51: everything from Item down is completely duplicated. You can remove all of this without changing anything.

EmoGarbage404 (changes requested):

Resources/Prototypes/Entities/Objects/Tools/handheld_mass_scanner.yml
RedBookcase: huh true. Why was it made that way to begin with?

RedBookcase: fixed now, turned out to be related to the item states.

PR #36343, opened 2025-04-06
Make `RandomMetadata` properly support localization

Tayrtahn:
## About the PR

`RandomMetadataComponent` can now do proper localization and is no longer only capable of producing a concatenated sequence of strings with a non-localizable separator.

## Why / Balance

Forcing the use of a fixed-order sequence of strings is _really_ bad for localization. A revenant named "The spirit of peaceful misery" in English isn't going to translate properly into other languages on a word-by-word basis ("peaceful" and "misery" would need to be swapped in many languages, such as Spanish). To properly localize that, we would need to pass the randomly-selected terms into Fluent as variables with format string to arrange them. So that's what we do.

## Technical details

Replaced the string concatenation in `RandomMetadataSystem` with logic that assembles a list of the segments, gives them names like "part0", "part1", etc. and passes them into a `Loc.GetString` call.

The `NameSeparator` and `DescriptionSeparator` DataFields of `RandomMetadataComponent` have been removed and replaced with `NameFormat` and `DescriptionFormat`, which are `LocId`s of the format strings to use to assemble the name or description. The default format string is just `{$part0}` to simplify cases where only a single segment is used (which shouldn't need to do anything special for localization).

With this change, the relevant part of the revenant's prototype looks like this:
```
  - type: RandomMetadata
    nameSegments:
    - NamesRevenantType
    - NamesRevenantAdjective
    - NamesRevenantTheme
    nameFormat: name-format-revenant
```
and the locale file looks like this:
```
name-format-revenant = The {$part0} of {$part1} {$part2}
```

This allows other localizations to rearrange the terms as needed, and enables more fancy formatting tricks in general. A Spanish localization can use `El {$part0} de {$part2} {$part1}` to get much better results than were possible before. Fluent functions like `CAPITALIZE` can now be used on these formats too.

All content using `RandomMetadata` was converted for this change.
Many of the format strings are just `{$part0} {$part1}`, and it seem tempting to just use the same format string for different use cases, but remember that these terms might be formatted differently in other languages. `FirstName LastName` is the same arrangement as `AgentTitle CodeName` in English, but it might not be in every language, so it's better to define these as separate format strings.

## Media

Here are some pictures of things continuing to look the way they did before:
Screenshot 2025-04-05 at 11 14 28 PM
Screenshot 2025-04-05 at 11 16 01 PM
Screenshot 2025-04-05 at 11 17 34 PM
Screenshot 2025-04-05 at 11 18 27 PM
Screenshot 2025-04-05 at 11 19 33 PM
Screenshot 2025-04-05 at 11 21 14 PM
Screenshot 2025-04-05 at 11 21 33 PM
Screenshot 2025-04-05 at 11 24 01 PM
Screenshot 2025-04-05 at 11 35 49 PM

## Requirements

## Breaking changes

`RandomMetadata` has been reworked to properly support localization. Instead of `NameSeparator` and `DescriptionSeparator`, the component now has `NameFormat` and `DescriptionFormat` DataFields. These are `LocId`s of formatting strings to use to assemble to the final name. The randomly-selected segments can be inserted into the formatting string by the variable names `$part0`, `$part1`, `$part2`, etc. See `Resources/Locale/en-US/random-metadata/random-metadata-formats.ftl` for examples of formatting strings.

`RandomMetadataSystem.GetRandomFromSegments` now requires a format string as a parameter instead of an optional separator.


Content.Server/RandomMetadata/RandomMetadataSystem.cs
```diff
@@ -12,7 +12,7 @@ public sealed class RandomMetadataSystem : EntitySystem
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly MetaDataSystem _metaData = default!;
 
-    private List<string> _outputSegments = new();
+    private readonly List<(string, object)> _outputSegments = new();
```
beck-thompson, line 15: this susses me out but it looks like its handled with a big switch statement later on!

Content.Server/RandomMetadata/RandomMetadataSystem.cs
```diff
@@ -28,31 +28,32 @@ private void OnMapInit(EntityUid uid, RandomMetadataComponent component, MapInit
...
     /// <summary>
     /// Generates a random string from segments and a separator.
     /// </summary>
     /// <param name="segments">The segments that it will be generated from</param>
-    /// <param name="separator">The separator that will be inbetween each segment</param>
+    /// <param name="format">The format string used to combine the segments.</param>
     /// <returns>The newly generated string</returns>
     [PublicAPI]
-    public string GetRandomFromSegments(List<ProtoId<LocalizedDatasetPrototype>> segments, string? separator)
+    public string GetRandomFromSegments(List<ProtoId<LocalizedDatasetPrototype>> segments, LocId format)
```
beck-thompson, line 48: put in breaking changes that it now requires a format (you could also just give it a default and make it optional but that's up to you!)

beck-thompson (changes requested):
look good! Only one very minor thing. I haven't tested everything but I'm gonna do that tommrow

Content.Server/RandomMetadata/RandomMetadataSystem.cs
Tayrtahn: Good catch. Added a note to breaking changes.

beck-thompson (approved):
Looks good! Everything seems go work :+1:

PR #36355, opened 2025-04-06
Fix for ghosts being unable to follow the AI when it uses a holopad

### LINKED ISSUE #36342: Following the AI breaks when they take a holopad call
## Description

Their entity changes from the eye to the holoprojection, and loses followers.

**Reproduction**

Follow an AI eye as a ghost > have the AI answer a holopad call > lose track of the AI as the ghost.

**Screenshots**

**Additional context**

I think polymorph manages to handle this, so maybe there's a solution in there?

chromiumboy:
## About the PR

Implemented a fix for ghosts being unable to follow the AI when it uses a holopad

## Why / Balance

Closes #36342

## Technical details

When SharedStationAiSystem.SwitchRemoteEntityMode is called, an event is now raised on the old AI eye before it is deleted which contains a reference to the new AI eye that is replacing it. The FollowerSystem listens for this event and switches any followers of the old AI eye to the new one  

## Media

[Evidence that the bug is fixed](https://youtu.be/P-w3BWZuUsE)

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- fix: Ghosts will now follow the AI when it activates a holopad or answers a holo-call

### DIFF
--- a/Content.Shared/Follower/FollowerSystem.cs
+++ b/Content.Shared/Follower/FollowerSystem.cs
@@ -8,6 +8,7 @@ using Content.Shared.Hands;
 using Content.Shared.Movement.Events;
 using Content.Shared.Movement.Pulling.Events;
 using Content.Shared.Polymorph;
+using Content.Shared.Silicons.StationAi;
 using Content.Shared.Tag;
 using Content.Shared.Verbs;
 using Robust.Shared.Containers;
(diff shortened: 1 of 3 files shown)


Content.Shared/Follower/FollowerSystem.cs
```diff
@@ -169,6 +171,15 @@ private void OnFollowedPolymorphed(Entity<FollowedComponent> entity, ref Polymor
...
     }
 
+    private void OnFollowedStationAiRemoteEntityReplaced(Entity<FollowedComponent> entity, ref StationAiRemoteEntityReplacementEvent args)
+    {
+        if (args.NewRemoteEntity == null)
+            return;
+
+        foreach (var follower in entity.Comp.Following)
+            StartFollowingEntity(follower, args.NewRemoteEntity.Value);
+    }
```
slarticodefast, line 181: I don't really like that we are hardcoding multiple systems into follower code like this. This could maybe be generalized to be part of `SetRelay` in `SharedMoverController.Relay.cs`, but then it would also affect someone entering a mech or cardboard box. I am not sure if that would be good or bad.

PR #36369, opened 2025-04-07
Replace uplink thieving gloves with chameleon thieving gloves

Minemoder5000:
## About the PR

Replaces the traitor uplink thieving gloves with chameleon thieving gloves, and increases their price by one.
## Why / Balance

Thieving gloves have no unique fibers, so by replacing the uplink gloves with chameleon gloves, they now leave behind holographic chameleon fibers. This makes it so the detective can find some evidence that thieving gloves were in use by someone, instead of thieving gloves being completely invisible. Their price has been increased by one TC to compensate for the chameleon ability.
Alternative to #36370.
## Technical details

uplink_catalog.yml changed localization tag, increased price.
uplink-catalog.ftl changed localization to call them chameleon and reflect the change in the description.
## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl: Minemoder
- tweak: The traitor uplink thieving gloves have been replaced with chameleon thieving gloves, the same used by the Thief.

### DIFF
--- a/Resources/Locale/en-US/store/uplink-catalog.ftl
+++ b/Resources/Locale/en-US/store/uplink-catalog.ftl
@@ -315,8 +315,8 @@ uplink-chameleon-desc = A backpack full of items that contain chameleon technolo
 uplink-clothing-no-slips-shoes-name = No-slip Shoes
 uplink-clothing-no-slips-shoes-desc = Chameleon shoes that protect you from slips.
 
-uplink-clothing-thieving-gloves-name = Thieving Gloves
-uplink-clothing-thieving-gloves-desc = Discreetly steal from pockets and improve your thieving technique with these fancy new gloves. They even look like normal gloves!
+uplink-clothing-chameleon-thieving-gloves-name = Chameleon Thieving Gloves
+uplink-clothing-chameleon-thieving-gloves-desc = Discreetly steal from pockets and improve your thieving technique with these fancy new gloves. They can change appearance to match any pair of gloves!
 
 uplink-clothing-outer-vest-web-name = Web Vest
 uplink-clothing-outer-vest-web-desc = A synthetic armor vest. This one has added webbing and ballistic plates.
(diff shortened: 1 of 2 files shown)


Everturning:
It's like this on purpose because gloves are a stealth item and need to blend in with other things because you bought them while the thief gloves come free with the antag.

SpeltIncorrectyl:
I fully support this, would end thief glove metagaming overnight.
I hate obviously knowing someone has thief gloves (because why else would a doctor be wearing black gloves) but not be able to do anything about it because it's metagaming.
It is much better if they can disguise their gloves.

EmoGarbage404:
while i do think this is a decent solution in the intermediate, i would really prefer to see something less powerful than outright chameleon. Dying via crayons would be a good middle point imo.

SpeltIncorrectyl:
There is no reason why this shouldn't be merged. As far as the metashield is concerned thief gloves are supposed to be unnoticeable by sec. If it is merged so they are chameleon then nothing is changed, sec still won't notice thief gloves. Except by merging this pr it removes a stupid metashield crutch.

SpeltIncorrectyl:
> It's like this on purpose because gloves are a stealth item and need to blend in with other things because you bought them while the thief gloves come free with the antag.

You don't have to put any effort into blending in, they are protected by the metashield. Sec is not allowed to assume that black gloves == thief gloves.

SpeltIncorrectyl:
Something I would add to the pr would be making the default style for the chameleon gloves budget insulated gloves. They are the least suspicious pair of gloves on the station and it wouldn't be out of place for most people (as opposed to black gloves).

Just thinking for the situation in which a guy orders them but doesn't change them so it's not metagamed.

ArtisticRoomba:
You're a triager, just triage the PR.

Pumkin69:
This is the better alternative of the two PRs.

slarticodefast:
> while i do think this is a decent solution in the intermediate, i would really prefer to see something less powerful than outright chameleon. Dying via crayons would be a good middle point imo.

Yeah, I think this PR is fine in the meantime. @Minemoder5000 could you add a TODO code comment? Something like `Revert these to be non-chameleon once clothes dyeing exists`

slarticodefast (approved):
Looks good to me!

PR #36371, opened 2025-04-07
Arch

metalgearsloth:
Requires https://github.com/space-wizards/RobustToolbox/pull/5814


slarticodefast:
I'm closing this since the engine PR was closed.
One day arch will be real 😢

K-Dynamic:
We should probably inform Arch that we're not using their system currently https://github.com/genaray/Arch#space-station-14

PR #36379, opened 2025-04-07
Introduce a Content sided config file

VasilisThePikachu:
## About the PR

This PR adds a server_config.toml file on the content side of the game.

The new default has what I believe to be the most touched config options for Space Station 14, sandbox player side and server host side. Along with a good description on what everything does.

## Why / Balance

Before the example config file was provided by Robust Toolbox. Which meant it could only really have RT related settings in it.

## Technical details

This PR tells dotnet on build time (Via Content.Server.csproj) to replace the config file with the one in content, which means we can have specific SS14 cvars.

Fork owners may add their own cvars they believe should be easily visible at a glance by just editing the server_config.toml inside the Content.Server folder.

## Requirements

## Breaking changes

Fork owners should change any of these defaults if they like to have a consistent experience for players downloading a sandbox server.

### DIFF
--- a/Content.Server/server_config.toml
+++ b/Content.Server/server_config.toml
@@ -53,7 +53,7 @@ id = "unknown_server_id"
 
 [events]
 # Controls if the game should run station events. (Meteors, Sleeper Syndicates ETC)
-eneabled = true
+enabled = true
 # Should admins count towards the max players count?
 # Explanation: 9 players, 2 admins. Server list shows 11/10 players. But those 2 admins are subtracted from the max players.
 # So someone who is not an admin can join, since technically the server is at 9/10 players.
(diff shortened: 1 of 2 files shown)


Content.Server/server_config.toml
```diff
@@ -0,0 +1,158 @@
...
+# Should players be able to see their own notes?
+see_own_notes = false
+
+[server]
+# Change this to have the changelog and rules "last seen" date stored separately from other servers.
+id = "unknown_server_id"
+
+[events]
+# Controls if the game should run station events. (Meteors, Sleeper Syndicates ETC)
+eneabled = true
```
metalgearsloth, line 56: eneabled

You should really make a test that makes sure this config can be parsed.

metalgearsloth:
I think these still exist and if not I'll write a test.

PR #36395, opened 2025-04-08
Reimplement water vapor condensation

B-Kirill:
## About the PR
- Reintroduced and reworked the water vapor condensation system (converting vapor to puddles)
- Puddle evaporation logic was updated to match the new condensation system (in particular, puddles do not evaporate with sufficient vapor concentration.)

More information in the technical part

## Why / Balance
This is simply an interesting mechanic that can be brought back in a new form (it previously existed in-game but was disabled due to bag optimization and other issues).

## Technical details
Added a new gas reaction `WaterVaporCondensationReaction` that occurs under these conditions:
- Temperature below 373.15 Kelvin (water boiling point)
- Pressure below 200 kPa (primarily prevents condensation in atmos)
- Vapor concentration of 5+ moles

Vapor condenses into water at a 1:1 ratio, with 0.1 moles processed per tick. Each tick processes 10% of eligible tiles (capped at 100 tiles/tick). These limits were implemented to:
- Slow down the condensation process
- Optimize performance

The puddle evaporation system was also modified:
- Water puddles won't evaporate if tile vapor exceeds 1 mole
- Added `PreventEvaporationComponent` to block evaporation
  - Can be reused by other systems
  - Solves client-server sync delays (previously caused 1-2 second description persistence issues)

Performance Notes:
The system has been well optimized for minimal gameplay impact. Benchmarks show that 100 tiles are processed from 0.1ms to 0.7ms

## Media
Condensation of vapor from a whole canister
vapor

It's just a video that everything works as it should

https://github.com/user-attachments/assets/c4b5f65d-e7cd-47ce-b1d3-03f8e67089a4

## Notes
Feedback is welcome!

## Requirements

**Changelog**
:cl:
- add: Condensation of water vapor into puddles
- tweak: Puddles do not evaporate in the presence of vapor


B-Kirill:
Yeah, the build failed again because of Poly, what's wrong with it :/

slarticodefast:
Random, unrelated test fail. I'm rerunning it for you.

Krzeszny:
Does the condensation speed depend on the temperature delta? Cause you know, room temperature water doesn't condense much.

Alpha-Two:
Based on some napkin arithmetic, water should instead convert moles:liquid at a 1 mol:4u ratio if you wanna be consistent. Obviously it doesn't actually matter and 1:1 is fine but 1:4 is slightly more accurate, imo.

B-Kirill:
> From a quick look, this would have to be implemented as separate reactions for all gases that we might want to condense. Couldn't this be made generic by using DataFields rather than consts? Take a look at WaterVaporReaction.cs

This would really be nice, but I'm having trouble transferring data right now because `ProcessEligibleTiles` is a separate static class. I'll still try to come up with a solution, but in the worst case, I'll have to leave everything as it is.

B-Kirill:
> Does the condensation speed depend on the temperature delta? Cause you know, room temperature water doesn't condense much.

I would leave everything as it is, for some technical reasons of the implementation of this system. Moreover, I don't think that such strong realism is needed in this case.
I'll still think about it though.

B-Kirill:
> Based on some napkin arithmetic, water should instead convert moles:liquid at a 1 mol:4u ratio if you wanna be consistent. Obviously it doesn't actually matter and 1:1 is fine but 1:4 is slightly more accurate, imo.

I think a ratio of 1:4 would be too much, but 1:2 sounds perfect.

B-Kirill:
Great, I broke everything..

B-Kirill:
**I'll fix it later**

PR #36398, opened 2025-04-08
Fix `lsobjectives` target player logic

Tayrtahn:
## About the PR

The `lsobjectives` command no longer requires that the executing client has a player session.

## Why / Balance

This is preventing the command from working from the server console and in integration tests.

## Technical details

Small, sneaky logic error in the old code.
```
var player = shell.Player;
if (player == null || !_players.TryGetSessionByUsername(args[0], out player))
```
That null check will cause the command to always fail if `shell.Player` is null. Presumably the intention was to fall back to the local shell's player if the `TryGetSessionByUsername` call fails, and only if neither works, report a failure.

The logic is now: if an argument was passed, try to find a player with that username. If no arguments were passed, used the local shell's player instead. If that still didn't work, report an error.

This is a slight change in behavior from before: Previously, if the specified player was not found, the command would instead report the objectives for the local player. However, that behavior is kinda weird, and probably not intended, so it has been changed to report that the target player was not found instead.

## Requirements

### DIFF
--- a/Content.Server/Objectives/Commands/ListObjectivesCommand.cs
+++ b/Content.Server/Objectives/Commands/ListObjectivesCommand.cs
@@ -18,8 +18,8 @@ namespace Content.Server.Objectives.Commands
 
         public override void Execute(IConsoleShell shell, string argStr, string[] args)
         {
-            var player = shell.Player;
-            if (player == null || !_players.TryGetSessionByUsername(args[0], out player))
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
There was, indeed, something wrong with it. Hero.

slarticodefast (approved):
Looks good! Thanks for fixing it!

PR #36399, opened 2025-04-08
Replaced Sterile Swabs in NutriMax with a Swab Dispenser

TheBlueYowie:
## About the PR
- Replaced the sterile swabs in NutriMax inventory with a swab dispenser

## Why / Balance
- Swabs are one of my favourite mechanics in Botany, however there are sometimes more Botanists mapped than botany lockers, so sometimes there are not enough dispensers for everyone. 
- Also if you use all of your swabs in your starting dispenser, you need to individually vendor the 20 swabs in the nutrimax (As you typically swab a plant multiple times). Instead we can now just get the dispenser from the vendor and dispense swabs as needed, so a nice quality of life improvement.
- I would also like to see the swab dispenser added to the hydro equipment crate and/or perhaps the BioGen, but I think these deserve more consideration and their own separate PR.
- Balance wise - The dispenser contains 30 swabs instead of the 20 in the NutriMax before, so it now has 10 more swabs total
- Performance wise - I'm aware that entity count is always a concern with botany, but swabs are typically dispensed, used, then throw in the trash right away so this should not be an issue here.

## Technical details
- Replaced 20 sterile swabs (DiseaseSwab) in NutriMax inventory with 1 swab dispenser (BoxMouthSwab)

## Media

Before:
Swabs
After:
Dispenser

## Requirements

## Breaking changes

- N/A

**Changelog**

:cl:
- tweak: NutriMax now contains a sterile swab dispenser instead of individual swabs

### DIFF
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/nutri.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/nutri.yml
@@ -15,7 +15,7 @@
     ChemistryBottleRobustHarvest: 3
     ChemistryBottleEZNutrient: 3
     Bucket: 3
-    DiseaseSwab: 20
+    BoxMouthSwab: 1
     BoxAgrichem: 1
     #TO DO:
     #plant analyzer


slarticodefast (approved):
Good idea!

Tayrtahn (approved):
Makes sense to me

PR #16687, opened 2023-05-22
No atmos stacking (attempt 2)

tom-leys:
## About the PR
Atmos Technicians were creating stacks (5-20!) of filters and similar devices on the same tile. That wasn't the intent of the system and it makes their designs visually unreadable which is not what we want. 

###This is a redux of #16331 as it needed to check only against ANCHORED unstackables as mentioned in the revert #16608 

However the main reason for stacking was because the gas filter is woefully underpowered at only 200 Litres / Sec which is especially bad for hot gasses like Tritium from a plasma fire pipe.

- Gas Mixers, Filters, pumps and similar devices are all tagged as unstackable and this is enforced with a new construction rule NoUnstackableInTile
- There is a matching check when anchoring to prevent stacking through re-anchoring structures.
- Filters now have a 5x max volume to compensate for no more stacking, so a max filterrate of 200 L/S
- Filter UI no longer enforces the max rate, as this was already enforced in the GasFilter system, the UI wasn't reading the actual max from the GasFilter component.
- Filter UI has a large enough region to type 1000 L/S into.

**Media**

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

- remove: You cannot stack Gas Filters, Mixers or several other Atmos devices like pumps and valves. Pipes can still be stacked.
- tweak: Gas filters can now filter 5x faster (up to 1000 L/S) to compensate for no more filter stacks. This is their new default rate

PR #36411, opened 2025-04-09
Fix absurd volume in gas filters

kosticia:
## About the PR
Decreased transfer rate in gas filter

## Why / Balance
Zero reasons for filter to be x5 more efficient than volumetric pump. Bug.

## Media
n/a

## Requirements

**Changelog**
:cl:
- fix: Gas filters now have accurate transfer rate.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Piping/Atmospherics/trinary.yml
+++ b/Resources/Prototypes/Entities/Structures/Piping/Atmospherics/trinary.yml
@@ -60,8 +60,8 @@
           type: GasFilterBoundUserInterface
     - type: GasFilter
       enabled: false
-      transferRate: 1000
-      maxTransferRate: 1000
+      transferRate: 200
+      maxTransferRate: 200
     - type: Flippable
       mirrorEntity: GasFilterFlipped
     - type: Construction


ArtisticRoomba:
Gas filters will see their full throughput usage in Flowmos, and the currently high transfer rates help to fulfill potential throughput losses when splitting a pipe into two (you both have a filtered output and an excess output). So this change seems a bit off to me. Have you tested these changes in various circumstances? (Trit production, chamber filtering, large waste filtering, etc)

kosticia:
> Gas filters will see their full throughput usage in Flowmos

This change can be easy reverted with flowmos. Right now 1000 l/sec is fully useless because all piping has 200 L volume. This is just confusing and can be potentially abused. 
>and the currently high transfer rates help to fulfill potential throughput losses when splitting a pipe into two (you both have a filtered output and an excess output). So this change seems a bit off to me. Have you tested these changes in various circumstances? (Trit production, chamber filtering, large waste filtering, etc)

As said, 1000L / per sec can't be used because pipes have 200L volume and just confuse players.
> (Or is it already capped at the atmospherics global max transfer rate? Godo)

Technically yes :godo: :ulybka:

ArtisticRoomba:
We're talking about gas filters right now (the chiral atmospheric devices), just to be clear. I would look at the code right now to see if this change affects anything important both now and the future but I can't right now.

Partmedia:
The reason for this change was given in #16687 . Is there a reason that this is no longer applicable?

M1tht1c:
> We're talking about gas filters right now
Hm somehow I managed to read the PR without actually reading the PR. (I think seeing flowmos mentioned made me think of scrubbers)

kosticia:
> Not necessarily true, since afaik the L/s rate is based on litres drawn from the source gas, with no regard for the output, (i.e. an air tile has 2500L, so we take 2/5ths of the gas in the tile and compress it as needed to fit into the output pipe) so the only reason they wouldn't be able to draw 1000L would be if it would cause the output to reach the pressure cap.
> 
> Considering that a single tile contains 2500L, and even a small room should have upwards of 10 tiles, currently it should take about 25 seconds to space a small room with a scrubber on panic, and after this change it would take upwards of 2 minutes. Considering that atmos techs already prefer to just space rooms with an RCD (which I think the maintainers would like to discourage?) it feels unhelpful to make scrubbers worse.
> 
> I should also mention that since scrubbers are used in a lot of burn chambers designs this would probably hurt their ability to pull out tritium before it burns(as well as make space for new fuel to be injected).

This pr doesn't change scrubbers at all. Just filters.

kosticia:
> The reason for this change was given in #16687 . Is there a reason that this is no longer applicable?

But... No? Because system with filter that filters 1000L -> pipe with volume 200L shouldn't work?

kosticia:
Goddamn. I'm stupid.

ArtisticRoomba:
Remember that in the current atmospherics system, pipes are not individual containers (as reported by a gas analyzer), they are actually a single unified volume separated by certain devices.

So the system can truly be greater than 200L, and an operation involving 1000L of gas can take place.

deltanedas:
> > The reason for this change was given in #16687 . Is there a reason that this is no longer applicable?
> 
> But... No? Because system with filter that filters 1000L -> pipe with volume 200L shouldn't work? 

thats not how atmos or speed in general works

PR #36414, opened 2025-04-09
Clarify "purple text" characters in rules & readd Space Law non-restricted item seizure

SlamBamActionman:
## About the PR

This PR aims to do two things:
- Clarify the "permanently disconnected" state for characters, also known as being "catatonic" or "soulless", and enshrine them clearer in the rules.
- Re-add the Space Law section regarding confiscating non-restricted items used in a crime.

The first is done by changing the examine text to read "Any recovery is impossible." from "Any recovery is unlikely.", and to use this as the qualifying text for when a character is counted as a non-player in the rules. It also provides examples of allowed/prohibited actions against such a character.

The second is a mere addition to Space Law text in the Search & Seizures subheading.

Additionally, a prohibited example regarding SSD/disconnected players was added. 

## Why / Balance

When a character is purple-texted, the original player *can't* under any normal circumstance return to the body. The only exception to this that I am aware of is if the body is injected with cognizine, and even then there is no guarantee the person claiming the ghost role is the same player. "Unlikely" makes it needlessly vague for new players to understand that the body is for all intents and purposes completely dead.

Seizing non-restricted items used in a crime used to be a thing but was removed in a rules rewrite at some point. Admins still consider it fair to confiscate items this way, but Space Law as written technically does not permit it.
image

## Technical details

Text editing.

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Permanently disconnected characters (i.e. catatonic/soulless/purple examine text) now have a clearer description that recovery is impossible.


Samuka-C:
I understand where you coming from, but the text "Any recovery is impossible" is kinda wrong
you can maybe recover them with cognizine, this will allow another (or maybe even the original) player to control the body.

SlamBamActionman:
> I understand where you coming from, but the text "Any recovery is impossible" is kinda wrong you can maybe recover them with cognizine, this will allow another (or maybe even the original) player to control the body.

I noted this in the Why/Balance section; it's a very specific scenario and for all intents and purposes the body itself is (before it gets cogni'd) just an object. For a Medical Intern getting a dead purpletext body dragged into medbay, it should not be implied that the body can somehow be revived on its own. 

I wouldn't consider the cogni to be a "recovery", and I think it's more suitable for cogni to be an exception to the stated text than be a niche situation the text alludes to. Plus it makes the language unambiguous that the body itself can be gibbed/meatspiked/morgued without rules consequences.

Samuka-C:
very fair

slarticodefast (approved):
Looks good from the technical side.
Self-merge whenever Nik gives the ok.

PicklOH:
This seems like it should be 2 different PRs.

PicklOH:
> How so? They're both very related changes and these clarifications are being made by both a maintainer and the admin team.

Maybe I'm dumb but I don't see how the SSD change and the Space Law change are related at all? Otherwise I don't think there are any issues with it, just pointing out it was weird they were added together.

SlamBamActionman:
Technically yeah probably should've been two, but since they're both just editing text to update clarity and require admin oversight, I slammed them together.

May my actions be condemned, and my sins never forgiven.

ArtisticRoomba:
I can't believe you could do such a thing... Your tests even cried out in protest as well.

SlamBamActionman:
Head admin approved by Nik.

slarticodefast (approved):

ScarKy0 (approved):
Looks good

PR #36419, opened 2025-04-09
Basic Resources Crate

ScarKy0:
## About the PR

Adds a "Basic Sheet Crate" for cargo to order.

## Why / Balance

I feel like this is a small QoL thing. Usually when I play cargo and we lack a certain resource (especially round start) I do not wonna have to buy 90 of each resource only to give 30 of each to medical or science.
The price of 1660 is basically ideal for roundstart to allow a fast delivery to other departments without having to wait for a bunch of money or for salvage to deliver all the materials. Midround and endgame it will still probably be more beneficial to simply buy the regular sheet crates or depend on salvage.

## Technical details

Yaml only changes.

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: A basic sheet crate can now be ordered for 1660 spesos. It contains 30 steel, glass and plastic.


Errant-4:
Wouldn't it make more sense to call it either "basic resources crate" or "basic materials crate" ?

ScarKy0:
> Wouldn't it make more sense to call it either "basic resources crate" or "basic materials crate" ?

All others were called by "sheet" so i kept it. It also makes it more intuitive to search for

TiniestShark:
The math on these feels like it could use a bit of adjustment since buying three of these gives you 90 of each batch of materials for 20 less spesos than just buying one of each crate type. It's not a _giant_ difference, but it is still something worth considering imo. Convenience should cost a bit more than specificity.

Errant-4:
> The math on these feels like it could use a bit of adjustment since buying three of these gives you 90 of each batch of materials for 20 less spesos than just buying one of each crate type. 

Cost of the 2 extra crates, probably?

RedBookcase:
I was kinda in opposition to this PR as bloat, but in the context of an incoming departmental economy PR it makes a lot more sense and allows department like med and security to order a basic subsistence-level supply of materials for themselves without having to beg cargo. With that in mind, i'm in favor of it.

ScarKy0:
> The math on these feels like it could use a bit of adjustment since buying three of these gives you 90 of each batch of materials for 20 less spesos than just buying one of each crate type

Oh please its 20 spesos. I definitely don't want this to be more expensive than default materials.
And as said, with dept economy this makes materials way more available for all departments without needing a crap ton of spesos.

Resources/Prototypes/Catalog/Fills/Crates/materials.yml
```diff
@@ -153,6 +153,18 @@
           # for some reason, the selector here adds 1 to whatever value it generates,
           # so this is actually 2-4
 
+- type: entity
+  id: CrateMaterialBasicResource
+  parent: CrateGenericSteel
+  name: basic sheet crate
+  description: 30 sheets of steel, glass and plastic.
```
Tayrtahn, line 160:
```suggestion
  description: 30 sheets of steel, glass, and plastic.
```

ScarKy0: Is one needed here? I hate english.

Tayrtahn: Needed? No. It's a style thing that will be debated until the end of time.

ScarKy0: A quick google search says the lack of comma is correct in this case, and a comma should only be placed when you connect two independant clauses. And I don't think its the case here?

ScarKy0: Ah whatever

Tayrtahn (approved):

ScarKy0:
Didn't wait for checks? The universe will now explode

TiniestShark:
> Oh please its 20 spesos. I definitely don't want this to be more expensive than default materials. And as said, with dept economy this makes materials way more available for all departments without needing a crap ton of spesos.

i mean the whole point was it costing just a smidge more to departments for the convenience and easy spread since everyone now has their own economy but whatever.

PR #36348, opened 2025-04-06
Fix docks clipping on shuttle nav

metalgearsloth:
UI scale casualty. Fixes https://github.com/space-wizards/space-station-14/issues/36295

**Changelog**
:cl:
- fix: Fixed docking ports being clipped in shuttle nav if using non-100% UI scales.

PR #36420, opened 2025-04-09
Fix docks in shuttle UI not being shown at >100% UI scale

ColonelThirtyTwo:
## About the PR

The existing culling code for removing the purple squares indicating docks in the shuttle piloting and mass scanner UIs does not account for UI scale, so they will not appear unless they are in the top left corner if the player has set their UI scale to more than 100% (common for high-res displays).

This patch changes the culling to use the scaled sizes instead, fixing the issue.

Ported from new-frontiers-14/frontier-station-14#2898, of which I am the author.

## Why / Balance

Not being able to see where you can dock most of the time is annoying. So is playing at 100% UI scale on a 4K display.

## Technical details

Change `ShuttleNavControl.xaml.cs` to use on `PixelSize` instead of `Size` for culling, which is scaled by the UI scale.

## Media

Currently, without patch:

Screenshot_20250409_165932

With patch:

Screenshot_20250409_170141

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Fix docks not always being visible in the shuttle UI when using UI scaling.

### DIFF
--- a/Content.Client/Shuttles/UI/ShuttleNavControl.xaml.cs
+++ b/Content.Client/Shuttles/UI/ShuttleNavControl.xaml.cs
@@ -303,7 +303,7 @@ public sealed partial class ShuttleNavControl : BaseShuttleControl
         const float sqrt2 = 1.41421356f;
         const float dockRadius = DockScale * sqrt2;
         // Worst-case bounds used to cull a dock:
-        Box2 viewBounds = new Box2(-dockRadius, -dockRadius, Size.X + dockRadius, Size.Y + dockRadius);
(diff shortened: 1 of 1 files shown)


slarticodefast:
Serms to be a duplicate of
https://github.com/space-wizards/space-station-14/pull/36348

ColonelThirtyTwo:
> Seems to be a duplicate of #36348

Ah, I didn't see that. I had made this fix for frontier awhile ago and figure I'd port it here. That patch probably does it better though since it also scales the dock radius.

metalgearsloth:
Apologies whoever deals with your conflicts

PR #36424, opened 2025-04-10
Wizard Teleport Scroll (Teleport Location ECS)

keronshb:
## About the PR

Two big features here!

First the Wizard Teleport Scroll, which is just a byproduct of the Teleport Location ECS. It's a wizard staple and is very much needed for the antagonist.

And the Teleport Location ECS! This is the main meat of the PR. It gets player-safe teleport points that it can relay to BUIs and the like.

## Why / Balance

The Teleport scroll is painfully needed so the Wizard just doesn't hide out in space all round, and it's way better flavor than a wizard flying in with a hardsuit.

## Technical details

Teleport Location ECS grabs a list of player safe warp entities (uses warp point component and checks if the ent is ghost only).
BUI comes with it. There's support to add speech, effects, and to close the BUI after if desired.

Wizard Teleport Scroll is added and given to the wizard as a default item. 120 second cooldown, can be made longer if needed.
- The cooldown is separate than the regular use delay!

The Wizard Hardsuit is also removed because of this. It's still in the game files but will be reworked at a later date.

## Media

https://github.com/user-attachments/assets/929c6e62-c688-4509-9cd6-e22733317b83

## Requirements

## Breaking changes

A new field was added to WarpPoints - GhostOnly
This is what allows the Teleport Location to work, since it excludes points like the Rod, Singulo, etc.

**Changelog**

:cl:
- add: Added the Wizard Teleport Scroll!


Content.Client/Teleportation/Ui/TeleportLocationsBoundUserInterface.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        base.ReceiveMessage(message);
+
+        if (_menu is null)
+            return;
+
+        if (message is TeleportLocationRequestCloseMessage)
+            CloseBUIResponse();
+    }
+
```
metalgearsloth, line 24: Kid named CloseUI

Content.Client/Teleportation/Ui/TeleportLocationsBoundUserInterface.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+    }
+
+    protected override void Open()
+    {
+        base.Open();
+
+        _menu = this.CreateWindow<TeleportMenu>();
+        SendMessage(new TeleportLocationRequestPointsMessage());
+
```
metalgearsloth, line 35: Kid named SetUIState

Content.Client/Teleportation/Ui/TeleportLocationsBoundUserInterface.cs
```diff
@@ -0,0 +1,55 @@
...
+            return;
+
+        _menu._warps = teleComp.AvailableWarps;
+        _menu.AddTeleportButtons();
+
+        _menu.TeleportClicked += (netEnt, pointName) =>
+        {
+            SendMessage(new TeleportLocationRequestTeleportMessage(netEnt, pointName));
+        };
+    }
```
metalgearsloth, line 46: SendPredictedMessage

Content.Server/Teleportation/TeleportLocationsSystem.cs
```diff
@@ -0,0 +1,58 @@
...
+    protected override void OnTeleportLocationRequest(Entity<TeleportLocationsComponent> ent, ref TeleportLocationRequestTeleportMessage args)
+    {
+        if (ent.Comp.User is null || string.IsNullOrWhiteSpace(ent.Comp.Speech))
+            return;
+
+        _chat.TrySendInGameICMessage(ent.Comp.User.Value, $"{ent.Comp.Speech.Trim()} ({args.PointName})", InGameICChatType.Speak, ChatTransmitRange.Normal);
+
+        base.OnTeleportLocationRequest(ent, ref args);
+    }
+}
```
metalgearsloth, line 58: Isn't this just a copy of ghost warp, why don't we just use that UI again?

Content.Server/Warps/WarpPointComponent.cs
```diff
@@ -14,5 +14,12 @@ public sealed partial class WarpPointComponent : Component
...
         [DataField]
         public bool Follow;
+
+        /// <summary>
+        ///     Should this warp point be accessable to ghosts only?
+        ///     Useful where you want things like a ghost to reach only like CentComm
+        /// </summary>
+        [DataField]
+        public bool GhostOnly;
     }
```
metalgearsloth, line 24: Just use a whitelist.

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// <summary>
+    ///     Should the user have some speech if they teleport?
+    ///     If enabled it will be prepended to the location name.
+    ///     So something like "I am going to" would become "I am going to (Bridge)"
+    /// </summary>
+    [DataField]
+    public string Speech = "";
+}
+
```
metalgearsloth, line 45: As above just roll it into warp point whitelist.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        _delay.TryResetDelay(ent.Owner, true, id: TeleportDelay);
+
+        if (!ent.Comp.CloseAfterTeleport)
+            return;
+
+        // Teleport's done, now tell the BUI to close if needed.
+        _uI.ServerSendUiMessage(ent.Owner, TeleportLocationUiKey.Key, new TeleportLocationRequestCloseMessage());
+    }
+}
```
metalgearsloth, line 57: Why is the open delayed but not spamming teleports? A spoofed can just spam these out.

Also no reason to wrap an entire BUI message around a closeUI call.

Content.Server/Teleportation/TeleportLocationsSystem.cs
```diff
@@ -0,0 +1,58 @@
+﻿using Content.Server.Chat.Systems;
+using Content.Server.Warps;
+using Content.Shared.Teleportation;
+using Content.Shared.Teleportation.Components;
+using Content.Shared.Teleportation.Systems;
+
+namespace Content.Server.Teleportation;
+
+public sealed partial class TeleportLocationsSystem : SharedTeleportLocationsSystem
```
metalgearsloth, line 9: Please document your code.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Content.Shared.Teleportation.Components;
+using Content.Shared.Timing;
+using Content.Shared.UserInterface;
+
+namespace Content.Shared.Teleportation.Systems;
+
+public abstract partial class SharedTeleportLocationsSystem : EntitySystem
+{
```
metalgearsloth, line 8: Same here though most of this should be wrapped into warps.

metalgearsloth (changes requested):

Content.Server/Teleportation/TeleportLocationsSystem.cs
keronshb: It has player teleports baked into it pretty hard and if the ghost should follow the warp point.

Content.Client/Teleportation/Ui/TeleportLocationsBoundUserInterface.cs
keronshb: Please explain further on why this is suggested.

Content.Client/Teleportation/Ui/TeleportLocationsBoundUserInterface.cs
keronshb: Please give an explanation why.

Content.Server/Teleportation/TeleportLocationsSystem.cs
keronshb: Why does this need to be documented? It's on the component.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
keronshb: Why does this need to be documented? It's on the component.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
keronshb: Didn't know that was a thing!

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
keronshb: What kind of whitelist are you looking for?
If you make a whitelist for Warp Point Component, it includes every single Warp Point.
If you make one for tags, the tag still needs to be added to the CC maps.
I need further info on what you need first.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
keronshb: Why does this need to be documented? It's on the component.

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
MilonPL, line 1: Do not indent xmldoc.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+    [Dependency] private readonly SharedTransformSystem _xForm = default!;
+    [Dependency] protected readonly UseDelaySystem _delay = default!;
+    [Dependency] private readonly SharedUserInterfaceSystem _uI = default!;
```
MilonPL, line 11: Sort your dependencies alphabetically, use lowercase for xform.

Content.Client/Teleportation/Ui/TeleportMenu.xaml.cs
```diff
@@ -0,0 +1,71 @@
...
+using Robust.Client.UserInterface.Controls;
+
+namespace Content.Client.Teleportation.Ui;
+
+[GenerateTypedNameReferences]
+public sealed partial class TeleportMenu : DefaultWindow
+{
+    public HashSet<TeleportPoint> _warps = new();
+
+    public string _searchText = "";
```
MilonPL, line 15: Public fields use upper camel case.

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+    [DataField]
+    public bool CloseAfterTeleport;
+
+    /// <summary>
+    ///     Should the user have some speech if they teleport?
+    ///     If enabled it will be prepended to the location name.
+    ///     So something like "I am going to" would become "I am going to (Bridge)"
+    /// </summary>
+    [DataField]
+    public string Speech = "";
```
MilonPL, line 43: Make it nullable rather than an empty string, you're checking for `IsNullOrWhiteSpace` anyways.

MilonPL (changes requested):

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    private void OnUiOpenAttempt(Entity<TeleportLocationsComponent> ent, ref ActivatableUIOpenAttemptEvent args)
+    {
+        if (_delay.IsDelayed(ent.Owner, TeleportDelay))
+        {
+            args.Cancel();
+            return;
+        }
+
+        ent.Comp.User ??= args.User;
```
Tayrtahn, line 31: Seems a little wrong to do this in an attempt event handler.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
Tayrtahn: You can use `inheritdoc` with the component as `cref`.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
keronshb: Please explain why it would be and provide an alternative if possible.

Tayrtahn: Typically attempt events should only check if it's valid for something to happen, and shouldn't change state. If there's another event that's raised when the UI is opened, that would be a better place to do it.

Content.Shared/Teleportation/Systems/SharedTeleportLocationsSystem.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        SubscribeLocalEvent<TeleportLocationsComponent, ActivatableUIOpenAttemptEvent>(OnUiOpenAttempt);
+        SubscribeLocalEvent<TeleportLocationsComponent, AfterActivatableUIOpenEvent>(OnUiOpen);
+        SubscribeLocalEvent<TeleportLocationsComponent, TeleportLocationDestinationMessage>(OnTeleportToLocationRequest);
+    }
+
+    private void OnUiOpen(Entity<TeleportLocationsComponent> ent, ref AfterActivatableUIOpenEvent args)
+    {
+        if (ent.Comp.User is null || ent.Comp.User != args.User)
+            ent.Comp.User = args.User;
```
Tayrtahn, line 31: Any reason not to just always set `ent.Comp.User = args.User`?

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+    [DataField]
+    public bool CloseAfterTeleport;
+
+    /// <summary>
+    /// Should the user have some speech if they teleport?
+    /// If enabled it will be prepended to the location name.
+    /// So something like "I am going to" would become "I am going to (Bridge)"
+    /// </summary>
+    [DataField]
+    public string? Speech;
```
Tayrtahn, line 43: This should be localized, with the location passed in so that the localization system can handle placing it in the sentence instead of always putting it on the end.

Content.Server/Teleportation/TeleportLocationsSystem.cs
metalgearsloth: The main issue with this is it also couples it to wizard and doesn't support a subset of teleports either. The component name says it's teleportlocations but functionally it only supports that.

Content.Client/Teleportation/Ui/TeleportLocationsBoundUserInterface.cs
metalgearsloth: You should be using UI states for stateful data not messages.

Content.Server/Teleportation/TeleportLocationsSystem.cs
keronshb: Kind of? You could throw it on a console and basically give it to Abductors/Clock Cult with some modification.
Definitely not as freeform as just using an AI eye on a camera console and hitting teleport/recall but basically can still be used as part of a teleport console. The effect itself can be replaced with anything like sparks, a silhouette, etc and the speech itself is optional. In theory if you put it on a console you could proooobably make the console speak.

At the very least since this uses the BUI system as a backbone it has more functionality than the ghost menu.

What thoughts did you have?

Golubgik:
why did you decide to make the teleport scroll infinite?

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
keronshb: Can you give an example? I'm having a hard time visualizing what you're asking me to do.

keronshb:
> why did you decide to make the teleport scroll infinite?

Trying something new. Using it with a longer cooldown might be a bit better, if not then we just add charges.

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
Tayrtahn: Sure thing!

First, I would change this to a `LocId?` instead of a `string?`:
```cs
public LocId? Speech;
```

Then add a new loc string for it:
```fluent
teleportation-scroll-speech-wizard = EY TCHEL TORT TU ({$location})
```

And then when you go to use it, use `Loc.GetString` and pass in the location:
```cs
var msg = Loc.GetString(ent.Comp.Speech, ("location", args.PointName));
_chat.TrySendInGameICMessage(ent.Comp.User.Value, msg, InGameICChatType.Speak, ChatTransmitRange.Normal);
```

And of course, change the yaml to use the LocId:
```yaml
  - type: TeleportLocations
    teleportEffect: WizardSmoke
    closeAfterTeleport: true
    speech: teleportation-scroll-speech-wizard
```

keronshb: OH that's what you meant. Didn't know that's how you added it as a yaml field.

Content.Server/Teleportation/TeleportLocationsSystem.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    private void OnInit(Entity<TeleportLocationsComponent> ent, ref MapInitEvent args)
+    {
+        UpdateTeleportPoints(ent);
+    }
+
+    private void OnTeleportLocationRequest(Entity<TeleportLocationsComponent> ent, ref TeleportLocationRequestPointsMessage args)
+    {
+        UpdateTeleportPoints(ent);
+    }
```
EmoGarbage404, line 32: Sloth made this comment already but instead of querying the server for this from the client, just call this function on a sub to `UiOpenedEvent` (you already do it in the shared code)

Content.Server/Warps/WarpPointComponent.cs
```diff
@@ -10,9 +10,16 @@ public sealed partial class WarpPointComponent : Component
...
         /// </summary>
         [DataField]
         public bool Follow;
+
+        /// <summary>
+        /// Should this warp point be accessable to ghosts only?
+        /// Useful where you want things like a ghost to reach only like CentComm
+        /// </summary>
+        [DataField]
+        public bool GhostOnly;
```
EmoGarbage404, line 23: i dont know if i agree with sloth that this should be a whitelist but maybe a more descriptive name? The actual difference isn't that these things are "ghost only" but rather that they are non-fixed locations or tied to objects.

EmoGarbage404 (changes requested):
Major issues.

Content.Server/Warps/WarpPointComponent.cs
keronshb: I figure a blacklist would be better.
In more cases than not you want to opt into as many warp points as possible and exclude others.
What isn't done here yet is changing maps like centcomm that have warp points that are fixed locations.

Content.Shared/Teleportation/Components/TeleportLocationsComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+
+/// <summary>
+/// A teleport point, which has a location (the destination) and the entity that it represents.
+/// </summary>
+[Serializable, NetSerializable]
+public record struct TeleportPoint(string Location, NetEntity TelePoint)
+{
+    public string Location = Location;
+    public NetEntity TelePoint = TelePoint;
+}
```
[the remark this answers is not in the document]
keronshb: Didn't know that needed it. I honestly don't know how half of this actually works with all the attributes.

Resources/Locale/en-US/teleportation/teleportation-menu-gui.ftl
```diff
@@ -0,0 +1,3 @@
+﻿## Wizard
+teleportation-scroll-window-title = Teleportation Scroll
+teleportation-scroll-speech-wizard = EY TCHEL TORT TU ({$location})
```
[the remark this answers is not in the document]
keronshb: Kinda but it's to highlight where the person is going, but I don't see the harm removing it.

Content.Client/Teleportation/Ui/TeleportMenu.xaml
```diff
@@ -0,0 +1,10 @@
+﻿<DefaultWindow xmlns="https://spacestation14.io" Title = "{Loc 'teleportation-scroll-window-title'}" MinSize="450 450" SetSize="450 450">
```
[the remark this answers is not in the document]
keronshb: Im getting a lot of errors for this so imma skip

Content.Server/Teleportation/TeleportLocationsSystem.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        var allEnts = AllEntityQuery<WarpPointComponent>();
+
+        while (allEnts.MoveNext(out var warpEnt, out var warpPointComp))
+        {
+            if (_whitelist.IsBlacklistPass(warpPointComp.Blacklist, warpEnt) || string.IsNullOrWhiteSpace(warpPointComp.Location))
+                continue;
+
+            ent.Comp.AvailableWarps.Add(new TeleportPoint(warpPointComp.Location, GetNetEntity(warpEnt)));
+        }
```
metalgearsloth, line 50: I think you want to clear warps before doing this?

Content.Server/Teleportation/TeleportLocationsSystem.cs
```diff
@@ -0,0 +1,68 @@
...
+        if (Delay.IsDelayed(ent.Owner, TeleportDelay))
+            return;
+
+        if (!string.IsNullOrWhiteSpace(ent.Comp.Speech))
+        {
+            var msg = Loc.GetString(ent.Comp.Speech, ("location", args.PointName));
+            _chat.TrySendInGameICMessage(args.Actor, msg, InGameICChatType.Speak, ChatTransmitRange.Normal);
+        }
+
+        base.OnTeleportToLocationRequest(ent, ref args);
```
metalgearsloth, line 66: This needs wrapping up in action refactor ideally.

Emisse:
map changes approval

keronshb:
Please merge this as a hotfix when approved since it significantly changes how Wizard is played, which is the intended implementation.

Content.Shared/Warps/WarpPointComponent.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    [DataField]
+    public bool Follow;
+
+    /// <summary>
+    /// What points should be excluded?
+    /// Useful where you want things like a ghost to reach only like CentComm
+    /// </summary>
+    [DataField]
+    public EntityWhitelist? Blacklist;
```
slarticodefast, line 26: This should be on the `TeleportLocationsComponent` not on every single warp point.
The teleporter item decides where you can teleport, not the target.

Content.Shared/Warps/WarpPointComponent.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Content.Shared.Warps;
+
+/// <summary>
+/// Allows ghosts etc to warp to this entity by name.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class WarpPointComponent : Component
+{
+    [ViewVariables(VVAccess.ReadWrite), DataField]
+    public string? Location;
```
slarticodefast, line 13: Datafield includes VV, so the first attribute is redundant
also documentation

PR #36439, opened 2025-04-10
air alarm panic wire snipping forces panic mode

qwerltaz:
## About the PR

Snipping the air alarm panic wire forces panic mode until the wire is mended. When the wire is snipped, the mode change drop-down menus for the air alarm and for individual scrubbbers are disabled.

Before, the panic wire just changed the mode to siphon but allowed anyone to change it back normally in the alarm UI.

## Why / Balance

Before this change, the panic wire was useless. If you had access, you could just switch the mode to panic normally. If you didn't have access, you were better off snipping the access wire, giving you control to change the mode to panic and also any other.

This wire is now more dangerous when snipped, giving agents the ability to cause localized atmospheric hazards that are slightly harder to fix by Atmos and impossible to fix by the station AI, while remaining easy to spot (red siphoning animation on all scrubbers).

## Media

**Changelog**

:cl:
- tweak: Snipping the panic wire in an air alarm now forces panic mode until the wire is mended.

### DIFF
--- a/Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
+++ b/Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
@@ -74,7 +74,13 @@
         <!-- Mode buttons -->
         <BoxContainer Orientation="Horizontal">
             <Label Text="{Loc 'air-alarm-ui-window-mode-label'}" Margin="0 0 2 0" />
-            <OptionButton Name="CModeButton" HorizontalExpand="True" />
+            <BoxContainer Orientation="Vertical" HorizontalExpand="True">
+                <OptionButton Name="CModeButton" HorizontalExpand="True" />
+                <ui:StripeBack Name="CModeSelectLocked">
+                    <RichTextLabel Text="{Loc 'air-alarm-ui-window-mode-select-locked-label'}"
+                                   HorizontalAlignment="Center" />
+                </ui:StripeBack>
+            </BoxContainer>
             <CheckBox Name="AutoModeCheckBox" Text="{Loc 'air-alarm-ui-window-auto-mode-label'}" />
         </BoxContainer>
     </BoxContainer>
--- a/Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
+++ b/Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
@@ -74,13 +74,13 @@
         <!-- Mode buttons -->
         <BoxContainer Orientation="Horizontal">
             <Label Text="{Loc 'air-alarm-ui-window-mode-label'}" Margin="0 0 2 0" />
-            <BoxContainer Orientation="Vertical" HorizontalExpand="True">
+            <Control HorizontalExpand="True">
                 <OptionButton Name="CModeButton" HorizontalExpand="True" />
                 <ui:StripeBack Name="CModeSelectLocked">
                     <RichTextLabel Text="{Loc 'air-alarm-ui-window-mode-select-locked-label'}"
                                    HorizontalAlignment="Center" />
                 </ui:StripeBack>
-            </BoxContainer>
+            </Control>
             <CheckBox Name="AutoModeCheckBox" Text="{Loc 'air-alarm-ui-window-auto-mode-label'}" />
         </BoxContainer>
     </BoxContainer>
--- a/Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
+++ b/Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
@@ -75,7 +75,7 @@
         <BoxContainer Orientation="Horizontal">
             <Label Text="{Loc 'air-alarm-ui-window-mode-label'}" Margin="0 0 2 0" />
             <Control HorizontalExpand="True">
-                <OptionButton Name="CModeButton" HorizontalExpand="True" />
+                <OptionButton Name="CModeButton" />
                 <ui:StripeBack Name="CModeSelectLocked">
                     <RichTextLabel Text="{Loc 'air-alarm-ui-window-mode-select-locked-label'}"
                                    HorizontalAlignment="Center" />
--- a/Content.Client/Atmos/Monitor/UI/Widgets/ScrubberControl.xaml.cs
+++ b/Content.Client/Atmos/Monitor/UI/Widgets/ScrubberControl.xaml.cs
@@ -71,6 +71,7 @@ public sealed partial class ScrubberControl : BoxContainer
             _data.PumpDirection = (ScrubberPumpDirection) args.Id;
             ScrubberDataChanged?.Invoke(_address, _data);
         };
+        _pumpDirection.Disabled = data.AirAlarmPanicWireCut;
 
         _copySettings.OnPressed += _ =>
         {
@@ -109,6 +110,7 @@ public sealed partial class ScrubberControl : BoxContainer
 
         _data.PumpDirection = data.PumpDirection;
         _pumpDirection.Select((int) _data.PumpDirection);
+        _pumpDirection.Disabled = data.AirAlarmPanicWireCut;
 
         _data.VolumeRate = data.VolumeRate;
         _volumeRate.Value = _data.VolumeRate;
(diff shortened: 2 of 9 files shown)


Content.Server/Atmos/Monitor/Components/AirAlarmComponent.cs
```diff
@@ -47,4 +47,7 @@ public sealed partial class AirAlarmComponent : Component
     /// </summary>
     [DataField("normalPort", customTypeSerializer: typeof(PrototypeIdSerializer<SourcePortPrototype>))]
     public string NormalPort = "AirNormal";
+
+    [ViewVariables]
+    public AirAlarmMode? ForcedMode;
```
slarticodefast, line 52: add documentation
also why is this not a datafield?

qwerltaz: I didn't think it makes sense for it to be a datafield because it's meant to be the mode currently forced by something. For panic, it's set to panic mode when the wire is cut and back to null when the wire is mended. If we set it to some arbitrary mode at the beginning, the air alarm will just stay forced in that mode with no way for players to nullify it back, until something changes the forced mode, like snipping and mending the panic wire.

slarticodefast: Datafields are not only used for prototypes, but for serialization in general. Any information that should not be lost when saving and loading the game (for example when mapping) has to be a datafield

qwerltaz: Ah, thanks for the info. I don't know what is done to datafields or when else the game is saved and loaded, so I can't tell if this should be a datafield. Should it?

VerinSenpai: Is that in the slart guide? It should be

slarticodefast: Yeah, I think it should be a datafield. Think of this scenario: You cut the wire, save the map and then load it again. In that case the variable wasn't saved and is null again, meaning someone could change the settings even though the wire is still cut.
I got that in my guide, but I should maybe emphasize that point more!

VerinSenpai: Time to get slarted.

VerinSenpai: I'll stop now

Partmedia (commented):
This feature sounds fine. For implementing this feature, this code also seems reasonable.

I would probably ask you to consider if storing `AirAlarmMode? ForcedMode` is really the most clear for long-term maintainability. If someone wanted to add more wires, how should the air alarm respond? It doesn't seem like the air alarm mode should be set by just the last wire that was cut.

So perhaps what you're looking for is just storing a `bool PanicWireCut`.

Content.Server/Atmos/Monitor/Components/AirAlarmComponent.cs
qwerltaz: Wow, the guide is magnificent! If only I had known earlier. All the dots connect at an accelerating pace.

qwerltaz:
True, your suggestion sounds better. I implemented it.

slarticodefast:
I gave this a test, but it looks like it causes a desync for the UI:
example
- open the air alrm UI
- open the panel, then cut the panic wire
- try to change the current mode
- the UI will show it has changed until it gets a new state from the server and corrects itself

I think it would make sense to disable the option button while the wire is cut.

Content.Server/Atmos/Monitor/WireActions/AirAlarmPanicWire.cs
Partmedia, line 1: I don't think this is a particularly important concern, but now `PanicWireCut` duplicates information in the air alarm's `WiresComponent`. Maybe this is necessary because `PanicWireCut` needs to be sent to the client to disable the UI. But if the `WiresComponent` data changes through a means other than `Cut` or `Mend`, that could also lead to another desync.

qwerltaz: I don't think this is an issue. I don't think the data can or will ever be able to be changed through other means. I think it's fine now.

Content.Client/Atmos/Monitor/UI/Widgets/ScrubberControl.xaml.cs
Partmedia, line 1: FYI, and I don't think this is a real problem here, but: disabling a control via the UI does not prevent users with patched or modified clients from doing this.

superjj18:
I mean, can't you just cut the acc wire and set it to panic and then fix the acc wire?  All this really does is screw over engineers and atmoswho happen to not have a screwdriver and wire cutters on them

It's not that this is OP, it's just... not really necessary?

Even the most minimally equipped engineer or tider could fix this

ArtisticRoomba:
> I mean, can't you just cut the acc wire and set it to panic and then fix the acc wire?

Yeah, that's another way of going about it.

> All this really does is screw over engineers and atmos who happen to not have a screwdriver and wire cutters on them

I will say, you gotta be having a *really* bad day if you don't have tools on you as an engineer.

I will agree that this could make fixing these things a bit unintuitive, when atmos is hard to pick up already: there is no feedback if the panic wire is cut, so you'll need to do some troubleshooting by looking at the panel. We should look to the case where these types of things happening are explicitly communicated to the player. A good example is the PA which has flashing UI elements and warning text (quite flavorful I might add, I really want to see UI similar to it in the future) if the LIMIT wire is cut.

I'd like this change a lot more if there was feedback in the UI displaying an error message similar to the PA. But that inflates the scope of this PR.

VerinSenpai:
> I will say, you gotta be having a _really_ bad day if you don't have tools on you as an engineer.

😅

qwerltaz:
Added visual feedback. It now replaces the mode selector when the wire is cut. It's similar to the particle accelerator limiter failure alert, so people should be likely to check the wires when this shows up.

image

ArtisticRoomba:
Excellent!

VerinSenpai:
That looks really good. I love that

Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
```diff
@@ -74,7 +74,13 @@
         <!-- Mode buttons -->
         <BoxContainer Orientation="Horizontal">
             <Label Text="{Loc 'air-alarm-ui-window-mode-label'}" Margin="0 0 2 0" />
-            <OptionButton Name="CModeButton" HorizontalExpand="True" />
+            <BoxContainer Orientation="Vertical" HorizontalExpand="True">
```
PJB3005, line 77: This should be a regular `Control` instead.

Content.Client/Atmos/Monitor/UI/AirAlarmWindow.xaml
```diff
@@ -74,7 +74,13 @@
...
+                <OptionButton Name="CModeButton" HorizontalExpand="True" />
```
PJB3005, line 78: `HorizontalExpand="True"` does nothing here.

PJB3005 (changes requested):
Minor XAML nits.

PJB3005 (approved):

Content.Server/Atmos/Monitor/Systems/AirAlarmSystem.cs
```diff
@@ -468,11 +468,16 @@ private string GetPort(AirAlarmComponent comp)
...
         {
             return;
         }
 
+        if (controller.PanicWireCut)
+        {
+            mode = AirAlarmMode.Panic;
+        }
+
         controller.CurrentMode = mode;
```
slarticodefast, line 481:
```suggestion
        if (controller.CurrentMode == mode)
            return;

        controller.CurrentMode = mode;
```
We keep the check here to skip the rest of the code if nothing changed.

Content.Server/Atmos/Monitor/Systems/AirAlarmSystem.cs
```diff
@@ -468,11 +468,19 @@ private string GetPort(AirAlarmComponent comp)
...
             return;
         }
 
+        if (controller.PanicWireCut)
+        {
+            mode = AirAlarmMode.Panic;
+        }
+
+        if (controller.CurrentMode == mode)
+            return;
```
slarticodefast, line 482:
```suggestion
```
nevermind, that causes the UI to not update when already in panic mode

slarticodefast (approved):

PR #36440, opened 2025-04-10
Transparent tear gas

nkokic:
## About the PR

You can now see through tear gas smoke. Gas is now more quick acting and in toe to toe with flashbang.
Characters now cough and slow down upon inhaling tear gas for around 20 seconds and become blind for half of that time.

## Why / Balance

As far as I have seen, tear gas is mostly released when a security officer dies to an explosion.
Throwing tear gas works against security in some cases, by occluding the area they are supposed to catch the suspect giving advantage to them instead with smoke screen.
This allows security to clear rooms more quickly, SWAT style.

## Technical details

All YAML code:
New TearGasSmoke prototype without occluder component.
BaseFoam is without clickable component so all smokes and foams can be clicked through, so one could arrest someone through those.
Tweaks on reagent (color change most important) and grenade prototypes.

## Media

Here is demonstration video of officer Owen performing an arrest on our _volunteer_ testing out new tear gas.

https://github.com/user-attachments/assets/a2a7d474-5b21-4d70-9684-75a491ba1c31

## Requirements

## Breaking changes

Removed Clickable component from BaseFoam. Workaround would be adding a new parent for foams to have clickable component or adding clickable to every foam prototype that needs it.

**Changelog**

:cl:
- tweak: Tear gas smoke is now white and semi-transparent.
- add: Tear gas now causes coughing, slows you down and hits a bit faster than before.


SlimmSlamm:
I don't think Tear Gas is supposed to be black either so this is a nice change although maybe make it lower view distance while inside it instead of blinding completely?

RedBookcase:
> I don't think Tear Gas is supposed to be black either so this is a nice change although maybe make it lower view distance while inside it instead of blinding completely?

It's tear gas, it should absolutely blind you. Should make you scream too honestly, if it doesn't already.

Resources/Prototypes/Reagents/narcotics.yml
```diff
@@ -402,18 +402,25 @@
...
         - !type:ReagentThreshold
-          min: 20
+          min: 5
+      - !type:MovespeedModifier
+        walkSpeedModifier: 0.5
+        sprintSpeedModifier: 0.5
+        statusLifetime: 1.5
+        conditions: # because of the remainding after effect, threshold is given so the effects ends simultaniously
+        - !type:ReagentThreshold
+          min: 5
```
IProduceWidgets, line 423: This feels overkill. It basically makes these stun grenades, and we already have flashes doing a very similar thing.

IProduceWidgets (approved):
yml seems functional.

Resources/Prototypes/Reagents/narcotics.yml
nkokic: Well the idea IS to make them similar to stun grenades. Been playing SWAT4 and Ready Or Not, the games from where I drew these features.
I have seen discussions on SS14 how tear gas is often and solely used for torturing instead for crowd control.
With this change, tear gas would be used more often as a tactical item like flashbangs, but with it's own pros and cons opposite to flashbangs.

RedBookcase:
> > Tear gas seems too translucent and hard to notice
> 
> I should note I might PR smoke grenades since tear gas loses cover ability when translucent

Smoke grenades are a thing already, they're available to traitors and thieves with the smuggler kit. Just need to make it available to Security.

Alpha-Two:
I like the change, I just think the tear gas is a little too hard to see in the provided video.

superjj18:
I feel like it doesn't scream "you will go completely blind if you walk into me" as much anymore

SlamBamActionman:
I've tested out the changes and I've done some edits to the stats and visuals for a few reason:

1) The time it took for the blindness to take effect felt a bit too slow, but the old values also lasted far too long. So while I have kept it low it's not as low as the values you proposed.
2) Screaming now has the same threshold as blindness/slowdown, and coughing has a lowered probability. The coughing was a bit too frequent, and tying the screaming to the blindness is a good way to indicate that the effect is active.
3) I've created a white smoke sprite for this effect and changed the transparency to match. The original black-transparent smoke had unclear borders and on some tiles it was hard to even see it was there. Tear gas is white anyhow, so this should be more realistic.
4) The slowdown was raised to 0.65, but with the blindness I am unsure if it's needed. It works kinda well though to simulate stumbling, especially considering the counter is fairly easy. 

image
image

I think this PR hits the nail on the head that tear gas is at the moment discouraged and even when it's used for the right context (dispersing crowds) it's kind of a hinderance for Sec still. So making it transparent and non-clickable is great.

PR #36441, opened 2025-04-10
New "PR made on master" message

VasilisThePikachu:
New message:

It appears you created a pr from your master branch, this is an [anti-patern](https://jmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/) that should be avoided.

We suggest you follow [our git usage documentation](https://docs.spacestation14.com/en/general-development/setup/git-for-the-ss14-developer.html). 

You can move your current work from the master branch to another branch by following [these commands](https://ohshitgit.com/#accidental-commit-master). And then you may recreate your PR using the new branch.

### DIFF
--- a/.github/workflows/close-master-pr.yml
+++ b/.github/workflows/close-master-pr.yml
@@ -12,7 +12,7 @@ jobs:
     steps:    
     - uses: superbrothers/close-pull-request@v3
       with:
-        comment: "Thank you for contributing to the Space Station 14 repository. Unfortunately, it looks like you submitted your pull request from the master branch. We suggest you follow [our git usage documentation](https://docs.spacestation14.com/en/general-development/setup/git-for-the-ss14-developer.html) \n\n You can move your current work from the master branch to another branch by doing `git branch <branch_name` and resetting the master branch."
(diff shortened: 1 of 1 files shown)


slarticodefast:
I would maybe avoid the term "anti-pattern" which might be confusing

beck-thompson:
> I would maybe avoid the term "anti-pattern" which might be confusing

I was about to say the same thing! I was a little confused :laughing:

VasilisThePikachu:
What is confusing about it? What would you suggest?

slarticodefast:
It is a very technical term that might not be known to someone not coming from a coding background.
Also I preferred the way the old message made clear that the PR had to be closed because of this, and the thank you was a nice touch as well.

slarticodefast (approved):
Sounds better to me!
I'll let beck double check to be sure

FairlySadPanda:
youscrewedup

consider alternatives

FairlySadPanda:
see they can't get mad at you if it's got scug art

beck-thompson (approved):
looks good!

PR #36447, opened 2025-04-10
fix cargo products

lzk228:
## About the PR

remove name and desc from cargoProduct prototypes as it uses name and desc of the product

## Why / Balance

described above

## Technical details

no

## Media

no 

## Requirements

## Breaking changes

no

**Changelog**

no


beck-thompson (approved):
good catch!

slarticodefast (approved):
If these are not used we should probably delete the datafields for them.
They aren't even localized.

PR #36450, opened 2025-04-10
feat: disable arachnids

princesspengy:
just changes a single line to prevent selection of arachnids at character creation!


Callmore:
wrong repo

princesspengy:
I did. god why is the default option in the pull request menu to submit to upstream

PR #36456, opened 2025-04-11
Smuggler follow up

IamVelcroboy:
## About the PR
Fixes the smuggler satchels overfilling with minimal changes. 
- Adds notes on fill sizes
- Reduces the alcohol bottle fill range 4->2
- Reduces cigar cases range 2->1
- Splits the generic category 
- <details><summary>🐟</summary>Sneaks in a dehydrated carp</details>  

Alternate fix to [36146](https://github.com/space-wizards/space-station-14/pull/36146)
No shade just trying to do it with less changes. I went through each of the groups to determine which were the offenders and changed those, opposed to blanket culling a bunch of things. 

## Why / Balance
Was breaking tests making people go crazy.

## Technical details
~~Depends on [36155](https://github.com/space-wizards/space-station-14/pull/36155)~~

## Media
n/a

## Requirements

## Breaking changes
n/a

**Changelog**
n/a

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
+++ b/Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
@@ -28,8 +28,8 @@
     - !type:NestedSelector
       tableId: RandomSatchelTable6
 
-  #number after each group is potential max weight. If combined PMW is too high, it could result in an over-filled bag/error.
-  #number after each group is potential max inventory capacity. If combined PMIC is too high, it could result in an over-filled bag/error.
+
+  #number after each group inidcates potential max inventory fill. If all combined fills are too high, it could result in an over-filled bag/error.
 - type: entityTable
   id: RandomSatchelTable1
   table: !type:AllSelector
--- a/Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
+++ b/Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
@@ -28,6 +28,8 @@
     - !type:NestedSelector
       tableId: RandomSatchelTable6
 
+  #number after each group is potential max weight. If combined PMW is too high, it could result in an over-filled bag/error.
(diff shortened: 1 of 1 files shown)


IProduceWidgets:
While this looks like it will work,
I feel this will be very annoying to maintain going forward as items get added and removed.
For that reason I suggest adding a new method for filling the storage that fails more gracefully and stops making new draws from the entitytable once it hits the storage limit instead of overflowing.

An alternative could be including a test that searches entitytables for the worst-case fill but that seems very hard and might take a long time to sim.

IamVelcroboy:
> While this looks like it will work, I feel this will be very annoying to maintain going forward as items get added and removed. For that reason I suggest adding a new method for filling the storage that fails more gracefully and stops making new draws from the entitytable once it hits the storage limit instead of overflowing. Or perhaps a selector that selects to grid size and then attempts to halt.
> 
> An alternative could be including a test that searches entitytables for the worst-case fill but that seems very hard and might take a long time to sim.
>

I'm just waking up but hear me out. What I'm thinking, and I can add an additional note, is that new items shouldn't be added to existing groups, but new bags/groups should be added.

However, as long as things added to any given group don't go over the size already allocated to them, it should be OK.

Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
```diff
@@ -28,22 +28,23 @@
     - !type:NestedSelector
       tableId: RandomSatchelTable6
 
+  #number after each group inidcates potential max inventory fill. If all combined fills are too high, it could result in an over-filled bag/error.
```
Killerqu00, line 31: "inidcates"

Killerqu00 (changes requested):

Resources/Prototypes/Entities/Clothing/Back/smuggler_tables.yml
IamVelcroboy: oof

slarticodefast (approved):
Looks good to me!

PR #36457, opened 2025-04-11
Exploding cigar

IProduceWidgets:
## About the PR

Adds an exploding cigar to the uplink for most service roles and the psychologist.

## Why / Balance

giphy
tumblr_o3095xQpCL1qmob6ro1_500
ddIl_w

## Technical details

All the events already existed I just connected the tubes man.

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: exploding cigars in traitor uplink for service jobs and the psych

### DIFF
--- a/Content.Server/Explosion/Components/TriggerOnSmokableExpended.cs
+++ b/Content.Server/Explosion/Components/TriggerOnSmokableExpended.cs
@@ -1,0 +1,8 @@
+namespace Content.Server.Explosion.Components;
+
+[RegisterComponent]
+public sealed partial class TriggerOnSmokableExpendedComponent : Component
+{
+    [DataField]
+    public EntityUid? LitBy = default!;
+}
--- a/Content.Server/Explosion/EntitySystems/TriggerSystem.cs
+++ b/Content.Server/Explosion/EntitySystems/TriggerSystem.cs
@@ -4,6 +4,7 @@ using Content.Server.Explosion.Components;
 using Content.Server.Flash;
 using Content.Server.Electrocution;
 using Content.Server.Pinpointer;
+using Content.Server.Nutrition.EntitySystems;
 using Content.Shared.Chemistry.EntitySystems;
 using Content.Shared.Flash.Components;
 using Content.Server.Radio.EntitySystems;
@@ -84,6 +85,9 @@ namespace Content.Server.Explosion.EntitySystems
         [Dependency] private readonly InventorySystem _inventory = default!;
         [Dependency] private readonly ElectrocutionSystem _electrocution = default!;
         [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
+        [Dependency] private readonly SmokingSystem _smokingSystem = default!;
+
+        private EntityQuery<MobStateComponent> _mobQuery;
 
         public override void Initialize()
         {
@@ -103,6 +107,7 @@ namespace Content.Server.Explosion.EntitySystems
             SubscribeLocalEvent<TriggerImplantActionComponent, ActivateImplantEvent>(OnImplantTrigger);
             SubscribeLocalEvent<TriggerOnStepTriggerComponent, StepTriggeredOffEvent>(OnStepTriggered);
             SubscribeLocalEvent<TriggerOnSlipComponent, SlipEvent>(OnSlipTriggered);
+            SubscribeLocalEvent<TriggerOnSmokableExpendedComponent, SmokableSolutionEmptyEvent>(OnSmokableExpended);
             SubscribeLocalEvent<TriggerWhenEmptyComponent, OnEmptyGunShotEvent>(OnEmptyTriggered);
             SubscribeLocalEvent<RepeatingTriggerComponent, MapInitEvent>(OnRepeatInit);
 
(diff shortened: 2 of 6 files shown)


PunishedJoe:
castro

Content.Server/Explosion/EntitySystems/TriggerSystem.cs
```diff
@@ -293,6 +298,23 @@ private void OnSlipTriggered(EntityUid uid, TriggerOnSlipComponent component, re
...
         }
 
+        /// <summary>
+        /// trigger's when a smokeable entity expends its reagents to 0.
+        /// </summary>
+        private void OnSmokableExpended(EntityUid uid, TriggerOnSmokableExpendedComponent component, ref SmokableSolutionEmptyEvent args)
+        {
+            _mobQuery = GetEntityQuery<MobStateComponent>();
+
+            var outcomp = new MobStateComponent();
```
Killerqu00, line 308: doesn't seem to be used

Content.Server/Explosion/Components/TriggerOnSmokableExpended.cs
```diff
@@ -0,0 +1,11 @@
+namespace Content.Server.Explosion.Components;
+
+[RegisterComponent]
+public sealed partial class TriggerOnSmokableExpendedComponent : Component
+{
+    /// <summary>
+    /// The entity which lit the triggering expendable entity
+    /// </summary>
+    [DataField]
+    public EntityUid? LitBy = default!;
```
slarticodefast, line 10: This is a little problematic, because you can light the cigarette, give it to someone else, but the user will still be you.
Some `XOnTriggerComponents` apply the effect to the user, which will have strange long-range effects in that case. And in my refactor all of them will have a `ApplyToUser` bool instead.
This will also throw errors in case the user is deleted by the time it does trigger.

Content.Server/Nutrition/EntitySystems/SmokingSystem.Cigar.cs
```diff
@@ -49,6 +50,10 @@ private void OnCigarInteractUsingEvent(Entity<CigarComponent> entity, ref Intera
                 return;
 
             SetSmokableState(entity, SmokableState.Lit, smokable);
+            if (TryComp<TriggerOnSmokableExpendedComponent>(entity, out var triggerComp))
```
slarticodefast, line 53: I'm not a fan of hardcoding the component here like this.
Also this will not work if the cigarette is ignited in any other way.

Resources/Prototypes/Entities/Objects/Weapons/Bombs/funny.yml
```diff
@@ -120,3 +120,40 @@
...
+  - type: Explosive
+    explosionType: Default
+    maxIntensity: 3.4
+    intensitySlope: 3
+    totalIntensity: 20
+    canCreateVacuum: false
+  - type: Tag
+    tags:
+      - Cigar
+      - Trash
```
slarticodefast, line 140:
```suggestion
    tags:
    - Cigar
    - Trash
```

Resources/Prototypes/Entities/Objects/Weapons/Bombs/funny.yml
```diff
@@ -120,3 +120,40 @@
...
+    maxIntensity: 0.01
+    intensitySlope: 1
+    totalIntensity: 0.01
+  - type: Construction
+    graph: smokeableCigarBombHarmless
+    node: cigarBombHarmless
+  - type: Tag
+    tags:
+      - Cigar
+      - Trash
```
slarticodefast, line 159:
```suggestion
    tags:
    - Cigar
    - Trash
```

Resources/Prototypes/Recipes/Crafting/Graphs/smokeables.yml
```diff
@@ -110,3 +110,25 @@
             doAfter: 5
     - node: ground
       entity: GroundTobacco
+
+- type: constructionGraph
+  id: smokeableCigarBombHarmless
+  start: start
+  graph:
+    - node: start
```
slarticodefast, line 118: fix indentation,
same below

slarticodefast (changes requested):
I'm currently refactoring trigger code and moving it to shared, so I would prefer if this PR would wait until after that is done, because this is quite a big merge conflict.
After that I can help you with rewriting this if you want.

Content.Server/Nutrition/EntitySystems/SmokingSystem.Cigar.cs
IProduceWidgets: That should be fine, it only sets the user for log purposes and otherwise doesn't do anything.

Content.Server/Explosion/Components/TriggerOnSmokableExpended.cs
IProduceWidgets: I'm happy to wait for that if you've got a way to solve the user issues. I've run into that before with triggers for interactinworld and the like as well, so I see why you want it.

IProduceWidgets: Do you mind linking the PRs? I did a quick search but I only found Slam's explosive trigger changes.

slarticodefast: still work in progress, I can ping you once I make the PR

Content.Server/Explosion/EntitySystems/TriggerSystem.cs
IProduceWidgets: I think you're right. Thanks.
At one point I was gonna have to use that to set the user, but I found a better way.

Resources/Prototypes/Recipes/Crafting/Graphs/smokeables.yml
IProduceWidgets: being honest, construction graphs explode every time I touch them, so I'm gonna pass on that for now.

slarticodefast (changes requested):
Ok the trigger refactor https://github.com/space-wizards/space-station-14/pull/39034 has been merged, so this will need some adjustments now:

- `TriggerOnSmokableExpended` should be moved to `Content.Shared.Trigger`.
- It has to be adjusted to the new trigger system so it inherits from `BaseTriggerOnXComponent` and passes in the trigger key into the `Trigger` method.
- The subscription should be moved into its own system in the same namespace.
- The hardcoding of the trigger component into the smokingsystem should be removed.
- The user should be the person smoking the cigar. You can do something similar to this to obtain the the entity

grafik

slarticodefast:
Are you coming back to this?

Issue #36454, opened 2025-04-11
ninja using gloves on power sources causes server exception


lzk228:
```
[FATL] unhandled: Robust.Shared.Utility.DebugAssertException: Entity 3068 is not the owner of the component. Component: BatteryComponent
   at Robust.Shared.Utility.DebugTools.AssertOwner(Nullable`1 uid, IComponent component) in /home/lzk/Documents/GitHub/space-station-15/RobustToolbox/Robust.Shared/Utility/DebugTools.cs:line 178
   at Content.Server.Power.EntitySystems.BatterySystem.IsFull(EntityUid uid, BatteryComponent battery) in /home/lzk/Documents/GitHub/space-station-15/Content.Server/Power/EntitySystems/BatterySystem.cs:line 219
   at Content.Server.Ninja.Systems.BatteryDrainerSystem.TryDrainPower(Entity`1 ent, EntityUid target) in /home/lzk/Documents/GitHub/space-station-15/Content.Server/Ninja/Systems/BatteryDrainerSystem.cs:line 109
   at Content.Shared.Ninja.Systems.SharedBatteryDrainerSystem.OnDoAfter(Entity`1 ent, DrainDoAfterEvent& args) in /home/lzk/Documents/GitHub/space-station-15/Content.Shared/Ninja/Systems/SharedBatteryDrainerSystem.cs:line 39
   at Robust.Shared.GameObjects.EntityEventBus.<>c__DisplayClass57_0`2.<SubscribeLocalEvent>g__EventHandler|0(EntityUid uid, IComponent comp, TEvent& args) in /home/lzk/Documents/GitHub/space-station-15/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 324
   at Robust.Shared.GameObjects.EntityEventBus.<>c__DisplayClass67_0`1.<EntSubscribe>b__0(EntityUid uid, IComponent comp, Unit& ev) in /home/lzk/Documents/GitHub/space-station-15/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 460
   at Robust.Shared.GameObjects.EntityEventBus.EntDispatch(EntityUid euid, Type eventType, Unit& args) in /home/lzk/Documents/GitHub/space-station-15/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 634
[...]

PR #36461, opened 2025-04-11
Fix ninja battery drain debug assertion

Tayrtahn:
## About the PR

Fixed a debug assert that would occur when a ninja drained power from a device.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/36454.

## Technical details

The end of `BatteryDrainerSystem.TryDrainPower` calls `BatterySystem.IsFull` to check if the ninja's battery is full, but it accidentally passes in the ninja itself as the UID instead of the ninja's battery entity. This trips the debug assertion in `IsFull`'s `Resolve` call when it checks that the `BatteryComponent` belongs to the passed UID.

Fixed by simply passing in the right UID - the `BatteryUid` field from the `BatteryDrainerComponent`.

## Requirements


slarticodefast (approved):
Looks good!

PR #36466, opened 2025-04-11
PlayerPanel Follow button

Errant-4:
## About the PR
Admin playerpanel now has a Follow button. This is basically just quick access to what is already possible via console with the `follow` command

## Why / Balance
Admin request

## Technical details
Button raises a new event, which FollowerSystem is subscribed to

## Media

Screenshot_2025-04-11_191938

## Requirements

## Breaking changes
None

**Changelog**
:cl: Errant
ADMIN:
- add: Player Panel now has a Follow button.


Content.Server/Administration/PlayerPanelEui.cs
```diff
@@ -141,6 +142,16 @@ public override void HandleMessage(EuiMessageBase msg)
                     _entity.DeleteEntity(session.AttachedEntity);
                 }
                 break;
+            case PlayerPanelFollowMessage:
+                if (!_admins.HasAdminFlag(Player, AdminFlags.Admin) ||
```
slarticodefast, line 146: I guess validating the message makes sense due to manipulated clients, but is the Admin flag the right one here? That feels a little restrictive considering you can already use the ghost warp menu.

Content.Server/Administration/PlayerPanelEui.cs
```diff
@@ -141,6 +142,16 @@ public override void HandleMessage(EuiMessageBase msg)
...
+                    !_player.TryGetSessionById(_targetPlayer.UserId, out session) ||
+                    session.AttachedEntity == null ||
+                    Player.AttachedEntity is null)
+                    return;
+
+                var ev = new FollowTargetEvent(Player.AttachedEntity.Value, session.AttachedEntity.Value) ;
+                _entity.EventBus.RaiseEvent(EventSource.Local, ev);
```
slarticodefast, line 153: does this have to be done via an event instead of simply using the API methods for it?

Content.Shared/Follower/FollowerSystem.cs
```diff
@@ -381,3 +389,19 @@ public EntityStoppedFollowingEvent(EntityUid following, EntityUid follower) : ba
     {
     }
 }
+
+/// <summary>
+///     Makes Follower follow Target
+/// </summary>
+[Serializable]
+public sealed class FollowTargetEvent : EntityEventArgs //TODO:ERRANT struct or something? Idunno
```
slarticodefast, line 397: always use recods structs where possible
also use the primary constructor

slarticodefast (changes requested):

Content.Server/Administration/PlayerPanelEui.cs
Errant-4: I would have loved to do it the easy way, too. But making RT allow you to add a new  system Dependency from an Eui is afaik an engine change (for each new thing, each time), or was I just having a skill issue?

slarticodefast: `_follower = _entMan.System<FollowerSystem>();`
in the constructor should work

Content.Server/Administration/PlayerPanelEui.cs
Errant-4: The `follow` command requires Admin

slarticodefast: The messages should always be server validated in case a client fakes them, so a check is good.
I guess if the other command has the admin flag, too, then it is fine here as well.

slarticodefast (approved):

PR #32561, opened 2024-09-30
Disable bioluminescence plant mutations because it breaks the engine

PJB3005:
No fun allowed

:cl:
- remove: Removed bioluminescence plant mutations due to it breaking the rendering engine.

PR #36479, opened 2025-04-12
Disable point light on rainbow weed

Minemoder5000:
## About the PR

This PR disables the point light and rgb controller components on rainbow weed.
## Why / Balance

50 morbillion point lights still destroy the renderer, and was disabled in #32561 as a random mutation.
## Technical details

Changed description of rainbow weed leaves
Commented out the light components and left a reason
## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl: Minemoder
- remove: Rainbow Cannabis Leaves no longer emit light due to a renderer limitation.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Hydroponics/leaves.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Hydroponics/leaves.yml
@@ -70,11 +70,13 @@
   name: rainbow cannabis leaves
   parent: LeavesCannabis
   id: LeavesCannabisRainbow
-  #description: "Is it supposed to be glowing like that...?"
-  description: "Wasn't it supposed to be glowing...?"
+  description: "Is it supposed to be glowing like that...?"
   components:
   - type: Sprite
-    sprite: Objects/Specific/Hydroponics/rainbow_cannabis.rsi
+    layers:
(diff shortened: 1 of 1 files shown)


slarticodefast:
Sad, but probably necessary. You could make it unshaded so it keeps its bright colors!

Everturning:
dislike this because its boooriiing plus I think we should have a better system for handling lighting by now so this kind of stuff doesn't come up

metalgearsloth:
> dislike this because its boooriiing plus I think we should have a better system for handling lighting by now so this kind of stuff doesn't come up

If someone codes it (renderer is very rarely touched by anyone) otherwise this is what we have bud.

PJB3005 (commented):
You know that if you want it not breaking the lighting engine, you can just disable shadow casting, right?

slarticodefast:
> You know that if you want it not breaking the lighting engine, you can just disable shadow casting, right?

That could work as well, but for now I just pushed some changes to make it unshaded, which still looks good.

example

metalgearsloth:
> You know that if you want it not breaking the lighting engine, you can just disable shadow casting, right?

I prefer unshaded over non-shadow casting lights :trollface:

PR #5868, opened 2021-12-23
Conditional map support

moonheart08:
## About the PR 
Maps can now have conditions required for them to be votable
That is all
```yaml
  conditions:
  - !type:HolidayMapCondition
    holidays: [ "Christmas" ]
```

PR #36481, opened 2025-04-12
New sharpness filter

B-Kirill:
## About the PR
Added a sharpness filter as an option in the graphics settings. The standard value is 0, it can be from 0 to 20.

## Why / Balance
Just a good feature. Those who like it can use it, otherwise just leave it at 0. It will also help some people to distinguish small details more easily.

## Technical details
Implemented using an overlay. 
`DisplaySharpening` CCvar has been added to a new `CCVars.Display.cs` file for client graphics CCvars, because all current graphics ones are stored in the Robust Toolbox, but this shader is only for SS14.

## Media

https://github.com/user-attachments/assets/3f51bef7-4c28-4e92-ae04-0702c35befa7

## Notes
Feedback is welcome!
If you want to see more like this, let me know.

## Requirements

**Changelog**
:cl:
- add: Added customizable sharpness filter in graphics settings


Content.Client/Overlays/SharpeningSystem.cs
```diff
@@ -0,0 +1,63 @@
...
+using Robust.Shared.GameObjects;
+using Robust.Shared.IoC;
+using Robust.Client.Graphics;
+using Robust.Client.Player;
+using Content.Shared.CCVar;
+using Content.Client.Overlays;
+
+namespace Content.Client.Overlays
+{
+
```
metalgearsloth, line 11: File-scoped namespaces please.

Content.Client/Overlays/SharpeningSystem.cs
```diff
@@ -0,0 +1,63 @@
...
+        private void OnPlayerAttached(EntityUid entity)
+        {
+            _overlayManager.AddOverlay(_overlay);
+        }
+
+        private void OnPlayerDetached(EntityUid entity)
+        {
+            _overlayManager.RemoveOverlay(_overlay);
+        }
+
```
metalgearsloth, line 49: Can't you just always run the overlay?

metalgearsloth:
@PJB3005 given she wrote the sharpener for rainworld (though that game did need it more). I don't see any harm leaving it upstream even if we leave it disabled.

Content.Client/Overlays/SharpeningSystem.cs
B-Kirill: > Can't you just always run the overlay?

It makes sense. Fixed in f7c534d

PJB3005 (commented):
What algorithm is this using? Also does it even make sense for this to be running as an overlay, where it can't blit to the final screen resolution directly?

B-Kirill:
> What algorithm is this using?

Mostly Unsharp Masking, although with a couple minor changes

> Also does it even make sense for this to be running as an overlay, where it can't blit to the final screen resolution directly?

In other words, do you think it's better to embed this shader in the render to process the entire screen instead of using it only in the game area using an overlay?

PJB3005:
> In other words, do you think it's better to embed this shader in the render to process the entire screen instead of using it only in the game area using an overlay?

I mean embedding it in the `Viewport` rendering logic so that the sharpening shader can take the native (nearest neighbor scaled) image, and then directly map it to the appropriate screen resolution when putting the viewport on-screen

B-Kirill:
Redone following @PJB3005 recommendations in [PR #5868](https://github.com/space-wizards/RobustToolbox/pull/5868)

Now this PR only adds a slider in the graphics settings. I will remove the draft status after accepting the PR in Robust Toolbox.

B-Kirill:
@PJB3005 
I redid everything according to your recommendation. _(Thanks for your patience)_

B-Kirill:
Well, why not open it?
Let it just exist, maybe will be added someday

PR #36487, opened 2025-04-12
Add a toggle to silicon laws to prevent accidental stating

M1tht1c:
## About the PR
Added a toggle to each silicon law that disables the state law buttons

## Why / Balance
This would make it much easier to avoid accidentally revealing that you are emaged as a borg. (As it stands it's pretty easy to reveal the emaged laws out of muscle memory)

## Technical details
Adds a single toggle button to the law display UI element, and disables any channel button depending on its state

## Media

https://github.com/user-attachments/assets/f79f2eda-198f-49e6-b246-f95fc4b0f0f7

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added a simple toggle to prevent accidentally stating certain laws

### DIFF
--- a/Content.Client/Silicons/Laws/Ui/LawDisplay.xaml
+++ b/Content.Client/Silicons/Laws/Ui/LawDisplay.xaml
@@ -6,8 +6,10 @@
                       HorizontalExpand="True"
                       VerticalExpand="True"
                       Margin="5 5 5 5">
-            <RichTextLabel Name="LawNumberLabel" StyleClasses="LabelKeyText"/>
-            <Button Name="AllowStatingLaws" Text="{Loc 'laws-ui-stating-toggle'}" ToggleMode="True" Pressed="True"/>
+            <GridContainer Columns="2">
+                <RichTextLabel Name="LawNumberLabel" StyleClasses="LabelKeyText" HorizontalExpand="False"/>
+                <Button Name="AllowStatingLaws" Text="{Loc 'laws-ui-stating-toggle'}" ToggleMode="True" Pressed="True" HorizontalExpand="False" HorizontalAlignment="Right"/>
+            </GridContainer>
             <customControls:HSeparator Margin="0 5 0 5"/>
             <RichTextLabel Name="LawLabel"/>
             <BoxContainer Name="LawAnnouncementButtons" Orientation="Horizontal" HorizontalExpand="True" Margin="0 5 0 0">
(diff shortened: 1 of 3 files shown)


M1tht1c:
Media is now attached. Also I don't think that test fail is real?

lzk228:
not sure if that's that necessary

EmoGarbage404:
i dont really understand why this is necessary. the ui only has a single button: to state laws. If you don't want to press it, just close the ui?

slarticodefast:
Yeah, I pretty much agree with Emo on this. Adding extra buttons for this just overcomplicates the UI in my opinion.
A more streamlined alternative would be a "state all" button, and a small, less intrusive checkmark for each law to decide which ones should be included.

M1tht1c:
> If you don't want to press it, just close the ui?

This is more specifically targeted at borgs with emaged laws - where you don't want to state laws 0 or 4. My personal experience is that when I'm told to state laws, my muscle memory is to run down the list and state every law. A couple of times despite being extremely self-conscious about not wanting to state law 0 I've clicked the button anyway out of habit. 

This PR is really just adding the confirmation prompt that acidifiers and the like have to the law interface(arguably a bit more important since if you acidify yourself on accident you aren't at risk of breaking a server rule). It's entirely possible this is just a me problem, but it was really simple to add so I figured I'd make a PR and see what others think.

M1tht1c:
> Adding extra buttons for this just overcomplicates the UI in my opinion.

I understand completely, the main reason I've done it this way is because this is my first time making a UI PR, and because I wanted a working implementation to get feedback on for a better UI(which I guess means this could have been a draft PR, but it is functionally complete, and feels like way to small of a PR to have a draft)

> A more streamlined alternative would be a "state all" button, and a small, less intrusive checkmark for each law to decide which ones should be included. 

Makes a lot of sense, although this feels like it could have to wait for the other laws UI PR.

> But even that I think is not really needed.

I personally feel like this feature would be very useful, as someone who has mistakenly stated law 0 far too many times(twice out of the three times I've been emaged...), but like I said in my last comment that could be entirely a me problem. (I also want to say this isn't a mald PR, I just wanted to make a PR and this came to mind as something easy that would let me learn the UI system)

PJB3005:
> A more streamlined alternative would be a "state all" button, and a small, less intrusive checkmark for each law to decide which ones should be included. But even that I think is not really needed.

For reference, this is how the UI works in SS13 (from my memory). You can't individually state laws, it's always via group like this.

For the actual UI issue at hand (accidentally stating laws): IMO there are better ways to solve this either way. In one of my borg rounds I ended up getting hacked by a ninja (without noticing) and accidentally stated law 1 (which was a very mild modification from crewsimov, but still incriminating). This UI change wouldn't even have prevented that.

My suggestion would be to just style the law *differently* if it's modified? Like have a red background or border or something like that. Enough for a player to realize "this isn't a stock law".

PR #36492, opened 2025-04-12
Local Material Silo

EmoGarbage404:
## About the PR

Adds the material silo as a way to reduce some annoying material management within a department.

The silo has a range of 25 (approx. one large department) and all lathes/material storage items linked to it will share a single pool of resources.

A new silo can be purchased via cargo for 5k (to stop spam and theft cheese)

## Why / Balance

The goal is to reduce situations where material management is unwieldy due to being split between tons of machines in a small area (science stuff being split between autolathe, protolathe, circuit imprinter, exosuit fab, flatpacker, etc)

The goal is _not_ to streamline delivery, as indicated by the limit on the range. People will still need to collect materials and bring them back to the department.

Machines are linked through the silo UI and must be within range. The silo also has to be powered to transmit materials.

## Technical details

Adds various events to SharedMaterialStorage system.

## Media

https://github.com/user-attachments/assets/0df7499b-a7e3-40cc-a5af-55ae58631626

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added the material silo. This machine allows nearby machines to share a pool of resources, cutting down on micromanagement.


Resources/Textures/Structures/Machines/silo.rsi/meta.json
DieselMohawk, line 1: The sprite is actually taken from "https://github.com/tgstation/tgstation/pull/39768/commits/d74b67828394a9842578279a6b8ab2955bb08216" and was made by MrDoomBringer

DieselMohawk (changes requested):

Resources/Textures/Structures/Machines/silo.rsi/meta.json
EmoGarbage404: Thanks for informing me of the provenance. I've updated the PR accordingly.

DieselMohawk (approved):
Peak

lzk228:
where do materials go if you link an autolathe with plastic to a silo with steel and then unlink them?
i think it should be somehow shown which materials are from silo and which are from a lathe

EmoGarbage404:
> where do materials go if you link an autolathe with plastic to a silo with steel and then unlink them? i think it should be somehow shown which materials are from silo and which are from a lathe

all the materials are in the silo. when you link a lathe, all the materials are moved to the silo. When you unlink it, the lathe has nothing.

Content.Server/Materials/OreSiloSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            clients.Add((netEnt, txt, beacon));
+        }
+
+        _userInterface.SetUiState(ent.Owner, OreSiloUiKey.Key, new OreSiloBuiState(clients));
+    }
+
+    private const float OreSiloPreloadRangeSquared = 225f; // ~1 screen
+    private readonly HashSet<EntityUid> _silosToAdd = new();
+    private readonly HashSet<EntityUid> _silosToRemove = new();
```
ScarKy0, line 72: Should be below dependancies

ScarKy0 (changes requested):
Have yet to do in-game testing. Will do that later.

ScarKy0 (changes requested):
Ok so after testing theres only one nitpick I have.
Machines should have some way to tell if theyre connected to the silo, like the "Materials" text showing up as "Materials (Silo)"

Resources/Textures/Structures/Machines/silo.rsi/meta.json
ArtisticRoomba, line 1: File needs to be 4-spaced

ArtisticRoomba (changes requested):
YAML is good, see comment about attribution.

In-game testing was very nice (prediction my beloved), no material duplication glitches or anything. Some really minor issues I observed:
- You can troll someone by sticking the silo in range of a lathe and yoinking the materials from the lathe. Armory lathes for example.
- If the machine is ever destroyed, nuked, or otherwise inaccessible, the info won't update in the linking list and it'll continue to say its linked.

Resources/Textures/Structures/Machines/silo.rsi/meta.json
ScarKy0: Wait we do 4-spaces and not 2-spaces?

ArtisticRoomba: Dev wiki conventions give examples in 4 spaces, though they don't precisely say the spacing (the example was made by PJB though).
RSIEdit outputs in 2 spaced reportedly.
Editorconfig requests that they're 4 spaced.
A maintainer has said in the past that they're supposed to be 4 spaced.

It's really a shrug, it's just painful when rider bulldozes formatting and whatever you're working with is 2 spaced but rider wants to enforce 4 spaced formatting as per the editorconfig. Maybe I'll bring it up later in the future.

ScarKy0:
I may be stupid but the "Silo Linked" label doesnt show up for me

ScarKy0 (approved):
Tested in game and I personally see no issues in the code.

IProduceWidgets:
Magic transfer feels a tiny bit silly when we've got pipes and wires everywhere.
That said I like the factory-ization. I definitely think this is a step in that direction. 

This feels a little like it will step on the toes of other potential systems like waypointed cargomules or another infrastructure system like cargo train lines. But of course we don't have those yet, so mild grains of salt.

Mechanically this seems like it will make it slightly easier for bad actors to waste a department's materials.
I think that's maybe not the biggest concern though given that usually all the machines are placed nearby one another regardless.
I also think the station economy rework could server to improve that issue a bit since the department will have a direct way to replenish.

I think its fine?
I might prefer a different system later though.

ScarKy0:
> Mechanically this seems like it will make it slightly easier for bad actors to waste a department's materials.
I think that's maybe not the biggest concern though given that usually all the machines are placed nearby one another regardless

Not neccessarily. You need to break to the silo to link a machine to it. So only people from that department can link them together. If you already have access to that department to begin with you can steal materials just as easily as without a silo.
Also its limited to 25 tiles range, not global like on downstreams

superjj18:
Feel like you may as well do what holopads do, in the sense of having different versions such as global, map wide, grid wide, and local.  Even if you don't want to implement a station wide silo, I feel like the mere option could have value for up and downstream

ScarKy0:
> Feel like you may as well do what holopads do, in the sense of having different versions such as global, map wide, grid wide, and local. Even if you don't want to implement a station wide silo, I feel like the mere option could have value for up and downstream

People can always modify the range in yaml

Content.Shared/Materials/OreSilo/SharedOreSiloSystem.cs
```diff
@@ -0,0 +1,164 @@
...
+            clientComp.Silo = ent;
+            Dirty(client, clientComp);
+
+            UpdateOreSiloUi(ent);
+        }
+    }
+
+    private void OnBoundUIOpened(Entity<OreSiloComponent> ent, ref BoundUIOpenedEvent args)
+    {
+        if (args.UiKey is OreSiloUiKey.Key)
```
MilonPL, line 70: Use Subs.BuiEvents rather than this.

Content.Shared/Materials/MaterialStorageComponent.cs
```diff
@@ -75,6 +75,12 @@ public enum MaterialStorageVisuals : byte
     Inserting
 }
 
+[ByRefEvent]
+public readonly record struct GetStoredMaterialsEvent(Entity<MaterialStorageComponent> Entity, Dictionary<ProtoId<MaterialPrototype>, int> Materials, bool LocalOnly);
```
MilonPL, line 79: Needs docs.

Content.Shared/Materials/SharedMaterialStorageSystem.cs
```diff
@@ -76,39 +95,54 @@ public int GetMaterialAmount(EntityUid uid, MaterialPrototype material, Material
...
-    public bool CanTakeVolume(EntityUid uid, int volume, MaterialStorageComponent? component = null)
+    public bool CanTakeVolume(EntityUid uid, int volume, MaterialStorageComponent? component = null, bool localOnly = false)
     {
         if (!Resolve(uid, ref component))
             return false;
-        return component.StorageLimit == null || GetTotalMaterialAmount(uid, component) + volume <= component.StorageLimit;
+        return component.StorageLimit == null || GetTotalMaterialAmount(uid, component, true) + volume <= component.StorageLimit;
+    }
+
+    public bool IsMaterialWhitelisted(Entity<MaterialStorageComponent?> ent, ProtoId<MaterialPrototype> material)
```
MilonPL, line 137: Docs.

MilonPL (changes requested):
Some minor nitpicks.

superjj18:
> > Feel like you may as well do what holopads do, in the sense of having different versions such as global, map wide, grid wide, and local. Even if you don't want to implement a station wide silo, I feel like the mere option could have value for up and downstream
> 
> 
> 
> People can always modify the range in yaml

Half my point was for admeme purposes.  Let's say the event runner wanted station at its absolute peak performance so they have centcomm say they are sending extra materials via "Blue Space Matterbeam"(patent pending) and the admin could simply use a global or map wide silo to instantly ensure the station is immediately at peak performance when it comes to materials.

I don't think any other admin tool, even the item spawner itself, could so effectively prepare the station for a threat of admeme proportions.  If the admin needs the station firing on all cylinders this is how they can do it.

ScarKy0:
> Does it include settings for being on the same grid or map

No, but doing it that way would hurt the hell out of performance

ArtisticRoomba (approved):
YAML/attribution side looks good to me. Still needs an in-game test from Milon.

MilonPL (dismissed):
The only nitpick I have is the client ItemList jumping around whenever you select an item.

annoying
This is because you're calling `ClientList.Clear()` whenever updating the UI, which resets the scroll position. Don't rebuild the list on every update, instead only remove/add new items and use `ClientList.Sort()` instead to get the correct order.

not so annoying

The UI not being predicted is also pretty annoying, but I'm not sure how to approach that because of the UI update logic being on the server.

Content.Shared/Materials/SharedMaterialStorageSystem.cs
```diff
@@ -57,17 +56,37 @@ private void OnMapInit(EntityUid uid, MaterialStorageComponent component, MapIni
         _appearance.SetData(uid, MaterialStorageVisuals.Inserting, false);
     }
 
+    /// <summary>
+    /// Gets all the materials stored on this entity
+    /// </summary>
+    /// <param name="ent"></param>
+    /// <param name="localOnly">Include only materials held "locally", as determined by event subscribers</param>
+    /// <returns></returns>
+    public Dictionary<ProtoId<MaterialPrototype>, int>GetStoredMaterials(Entity<MaterialStorageComponent?> ent, bool localOnly = false)
```
metalgearsloth, line 65:
```suggestion
    /// <returns></returns>
    public Dictionary<ProtoId<MaterialPrototype>, int> GetStoredMaterials(Entity<MaterialStorageComponent?> ent, bool localOnly = false)
```

Content.Server/Materials/OreSiloSystem.cs
```diff
@@ -0,0 +1,118 @@
...
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        // Solving an annoying problem: we need to send the silo to people who are near the silo so that
+        // Things don't start wildly mispredicting. We do this as cheaply as possible via grid-based local-pos checks.
+        // Sloth okay-ed this in the interim until a better solution comes around.
+
+        var actorQuery = EntityQueryEnumerator<ActorComponent, TransformComponent>();
+        while (actorQuery.MoveNext(out _, out var actorComp, out var actorXform))
```
metalgearsloth, line 83: On engine as its own component, and should iterate all players instead.

metalgearsloth: Fuck it just leave it as is and set a reminder for me on the weekend

IProduceWidgets:
After thinking about this further, I'm pretty sure this is going to be a mapping sin for large maps where its not at all clear these will cover the entire department.

I think we might be much better off with a another pipe system for material providing. I know, node groups are pain though. 
<3

ilovehans10:
Just my two cents here, but I feel like this would work much better if all linking was done with the multitool. This would prevent tiders from easily siphoning from a silo, and I assume would help performance as you would no longer have iterate over all `OreSiloClientComponent` items and find which are in range of a silo.

> After thinking about this further, I'm pretty sure this is going to be a mapping sin for large maps where its not at all clear these will cover the entire department. Also I think being on the edge and linking failing will be a frustrating engineering issue.

Would linkable range extenders be a possible option? That way for departments that are large/split up you could bridge gaps with the power of bluespace. To keep things balanced, you could have it where the silo extender can link (via multitool) with 1 silo that is ~50 tiles away from it, and it can service an area that is 1/2 the size of a regular silo.

> I think we might be much better off with another "pipe" system for material providing. I know, node groups are pain though.
<3

If you ask me making this rely on pipes is one way to make sure that this never gets used beyond what is set up roundstart. The process of ripping up a floor, placing pipes both manually and slowly over an area and then replacing the floor (now without the pretty departmental markings) is just a pain. On top of that it's already a pain to try to find out where atmos pipes were run after a bombing, and now we will have another type of pipes obfuscating things.

metalgearsloth (changes requested):

Content.Server/Materials/OreSiloSystem.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        if (!_userInterface.IsUiOpen(ent.Owner, OreSiloUiKey.Key))
+            return;
+        _clientLookup.Clear();
+
+        var clients = new HashSet<(NetEntity, string, string)>();
+        var xform = Transform(ent);
+
+        // Sneakily uses override with TComponent parameter
+        _entityLookup.GetEntitiesInRange(xform.Coordinates, ent.Comp.Range, _clientLookup);
```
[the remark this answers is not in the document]
EmoGarbage404: due to the nature of it i doubt you'd ever see >5 in a round.

metalgearsloth (approved):

ScarKy0:
Random test fail strikes again.

ScarKy0:
Peak

Vinkoln:
> Подумав об этом подробнее, я почти уверен, что это будет грехом картирования для больших карт, где не совсем ясно, что они будут охватывать весь отдел. Также я думаю, что нахождение на грани и неудачная компоновка будут раздражающей инженерной проблемой.
> 
> например: (exo suit fabs внизу слева, sci fabs вверху справа) изображение
> 
> Я думаю, что нам было бы гораздо лучше с другой системой "труб" для поставки материалов. Я знаю, что группы узлов - это больно. <3

it is possible to implement the connection of the silo to the machines via a multi-tool, and the connected machines are serviced in the silo interface (they can be disconnected in the silo interface), such a supply will eliminate the possibility of stealing resources from machines using the silo

PR #36495, opened 2025-04-12
Fix/shader radial compat mode

Fildrance:
## About the PR

fixing compatibility mode 

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Client/UserInterface/Controls/RadialContainer.cs
+++ b/Content.Client/UserInterface/Controls/RadialContainer.cs
@@ -1,12 +1,23 @@
 using Robust.Client.UserInterface.Controls;
 using System.Linq;
 using System.Numerics;
+using Robust.Client.Graphics;
+using Robust.Shared.Prototypes;
 
 namespace Content.Client.UserInterface.Controls;
 
 [Virtual]
 public class RadialContainer : LayoutContainer
 {
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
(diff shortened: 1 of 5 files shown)


Fildrance:
I fixed most of direct errors shader compiler were outputting, it compiles and launches but it bugs out in a very wierd way now...
So in non-compat mode everything is good and all, but in compatibility mode i get spam of those errors

ogl.debug.error: DebugSourceApi: Invalid uniform location
ogl.debug.error: DebugSourceApi: Uniform type does not match uniform method.

by playing around with shader code i figured out that it happens when i am using variable that is input parameter for shader from c#, but only vec4 array ones. I added way to push arguments of such type into shader some time ago and i'm afraid that its just somehow not working in compatibility mode (even tho code itself is executed and throws no errors).

https://github.com/space-wizards/space-station-14/pull/36495/commits/bdace6421236eee12804cebf947e7911c53af1ec this is commit with fixes, when i replace using data from highp vec4[64] sectorColors then error spam stops... 
My first guess was that my old commit in RT that introduced overload for pushing uniform Color[] into shader as vec4[] is not working for compat mode, but after 8 hours of attempting to read documentation and fix the original in RT, and then brief consulting with delusional useless ChatGPT, it seemed like the code SHOULD work fine there... Any ideas what can i look at to see where the real problem is?

Fildrance:
Needs some attention from someone who can peek into problem described in previous comment :( could be reopened then...

Issue #36507, opened 2025-04-13
SpriteFadeSystem needs work


slarticodefast:
See Sloth's review for this PR #35863 after it has been merged

PR #36509, opened 2025-04-13
Sprite fade review

metalgearsloth:
Resolves https://github.com/space-wizards/space-station-14/issues/36507

Clickable stuff needs doing later this should just keep parity with master except the part where previously if anything collides the mouse it aborts the entire loop which may not be intended.


slarticodefast:
Can you target stable with this since it's already on there?
Might need to be rebased

TheShuEd:
staging already merged in stable

slarticodefast:
But now it's not in stable for 2 weeks.
Please hotfix this.

PR #36511, opened 2025-04-13
update editorconfig docstring indentation

MilonPL:
## About the PR

removes stupid indentation now that we've decided we dont want it (see https://github.com/space-wizards/docs/pull/418#discussion_r1991741550)

image


slarticodefast:
Resharper only works for rider, right?
Is there a VS code equivalent?

MilonPL:
`xmldoc_indent_text` apparently :godo:

PR #36513, opened 2025-04-13
Bluespace lights

drteaspoon420:
## About the PR

Added new syndicate item for janitors and engineers. Bluespace light tubes and bulbs. These let you hide tiny and small items inside the tube. If the item has timer trigger (most explosives do) then the payload can be activated.
How it works:
You can insert any tiny/small item into the tube or bulb.
After installed on a fixture, if the light is **On** and it receives additional **On** signal it will do one of following:
- if the item has **OnUseTimerTrigger** Component and it is not already active, trigger it.
- else eject the item from the slot dropping it in-front of the fixture.
This prevents accidental triggering of the mechanism.

Light tube's content can be ejected also by breaking the tube.

## Why / Balance

Janitors especially need more love in their syndicate roles. Also I am sure someone figures out other inventive uses for the mechanism. I am already thinking complicated traps involving logic relays bolting doors and trapping someone with bunch of C4 lights.

## Technical details

Added `ActivatePayload` to `PoweredLightSystem` that handles triggering the payload in bluepsace bulbs and ejecting the contents. Logic to call this function is at `OnSignalReceived` where we check excess 'On' signals to the light fixture.
Also added constant of `PayloadSlotName` to `PoweredLightComponent`

## Media

image
[Screencast from 04-13-2025 04:13:21 PM.webm](https://github.com/user-attachments/assets/daca464f-7545-4232-89cc-f4008379d2ff)

## Requirements

- [ x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

There should be no breaking changes.

**Changelog**

:cl: DrTeaSpoon
- add: Bluespace Light Tubes and Bulbs mixed light box as Janitor and Engineer syndicate item.

### DIFF
--- a/Content.Server/Light/Components/PoweredLightComponent.cs
+++ b/Content.Server/Light/Components/PoweredLightComponent.cs
@@ -76,5 +76,7 @@ namespace Content.Server.Light.Components
         /// </summary>
         [DataField("unarmedHitStun")]
         public TimeSpan UnarmedHitStun = TimeSpan.FromSeconds(5);
+
+        public const string PayloadSlotName = "payloadSlot";
     }
 }
--- a/Content.Server/Light/EntitySystems/PoweredLightSystem.cs
+++ b/Content.Server/Light/EntitySystems/PoweredLightSystem.cs
@@ -20,6 +20,9 @@ using Robust.Shared.Audio.Systems;
 using Content.Shared.Damage.Systems;
 using Content.Shared.Damage.Components;
 using Content.Shared.Power;
+using Content.Shared.Explosion.Components;
+using Content.Shared.Containers.ItemSlots;
+using Content.Server.Explosion.EntitySystems;
 
 namespace Content.Server.Light.EntitySystems
 {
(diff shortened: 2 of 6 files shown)


saga3152:
They shouldn't be called bluespace light tube/bulb and be called like the regular ones to prevent meta

drteaspoon420:
I pushed the PR to get some discussion going. There are still something I am not clear what would be good. Recap on my thoughts I already posted on discord:
```'ve been thinking of giving janitors some syndicate love.

Ok, I have this idea I have been prototyping:
Allow putting item (mainly explosives) inside a light bulb. If the item has active timer or no timer, it is ejected when lamp gets 'on' signal when it's already on. If it does have valid timer trigger to activate, it does so.
The tube would fit max Small item (1x2) , bulb would fit tiny item (1x1)

Now major question is should this be unique syndicate janitor item. Or make it default feature of light tubes and bulbs.
If it were janitor Item I am thinking of mixed box of lights. But I also think this would be really cool feature for gray syndicates to take advantage of aswell so not sure if making it specifically syndicate item feature is good idea.
Another would be to just make them explosive but that is not as interesting as choosing your own payload.

Only problem I am seeing is that people start meta gaming by putting all the contraband into light tubes. so maybe making them janitor exclusive syndicate item would be for the best.
```

EmoGarbage404:
Maintainer consensus is to close this PR.

Concerns were expressed about the innocuous and ever-present nature of light tubes and how it makes finding these hidden items very difficult. Additionally, the fact that this was a unique "bluespace" thing attached to a normal item felt very arbitrary.

In the discussion, it was brought up that hidden stashes and/or traps like these should be more common and generic rather than being tied to a syndicate purchases. Something that more broadly integrates with the existing trigger system and allows for traps would be ideal.

Thank you for your work. Feel free to continue contributing in the future.

PR #36521, opened 2025-04-14
Minor fix to give Lone Operatives the correct roletype

Zokkie:
## About the PR
Lone Operatives are officially a Solo Antagonist. But because they use the same mindrole as nukeops their roletype is marked Team Antagonist when you join as them. This minor fix will now show players the correct roletype ingame.

## Why / Balance
To give them the correct roletype

## Technical details
Gave the Loneops their own mindrole with the roletype SoloAntagonist that still has MindRoleNukeops as their parent.

## Media
soloantag

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- fix: Lone Operative is now assigned the Solo Antagonist role ingame instead of the Team Antagonist role.

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -555,7 +555,7 @@
         factions:
         - Syndicate
       mindRoles:
-      - MindRoleNukeops
+      - MindRoleLoneops
 
 - type: entity
   parent: BaseTraitorRule
--- a/Resources/Prototypes/Roles/MindRoles/mind_roles.yml
+++ b/Resources/Prototypes/Roles/MindRoles/mind_roles.yml
@@ -170,6 +170,14 @@
 
 # Nukies
 - type: entity
+  parent: MindRoleNukeops
+  id: MindRoleLoneops
+  name: Loneops Operative Role
+  components:
+  - type: MindRole
+    roleType: SoloAntagonist
+
+- type: entity
   parent: BaseMindRoleAntag
   id: MindRoleNukeops
   name: Nukeops Operative Role


Zokkie:
@Minemoder5000 thanks for your feedback. The reason I went with Solo Antagonist is because @slarticodefast asked @JIPDawg in Discord once whether they were a Team or Solo Antagonist and JIP_Dawg replied: 
> Sorry, uh I think the general consensus is lone op should be solo
they have similar objectives, but don't necessairly have to help each other out

I don't mind fixing it to Team Antagonist instead if this is not the general consensus anymore. 

PS: If anyone was curious Loneop/Nukeop reinforcements have not been changed. They get the Team Antagonist roletype and they are told when taking the ghost role that they are part of the team that summoned them and needs to help them.

Errant-4 (changes requested):
Please also fix the mindrole datafield value on the SpawnPointLoneNukeOperative entity prototype. Part of why these keep getting the wrong mindrole is because it's set in two places, and one overwrites the other. But we kinda must  have it on the spawned entity in case it's used independently without a gamerule. 
(There is potential room for improvement here if we rework something so we only need to assign the mindrole in one place and make that work in both scenarios)

Zokkie:
@Errant-4 do you mean the SpawnPointLoneNukeOperative entity in ghost_roles.yml? It seems to already have the "MindRoleGhostRoleSoloAntagonist" value which seems to be the correct way to do it judging by what all the other entities have as their mindRoles datefield value in that file.

Unless you want me to replace the datafield value with "MindRoleLoneops" that I added in the mind_roles.yml? But that would mean all the other entities in the ghost_roles.yml file also have the wrong mindRoles datafield value if I understood you correctly?

Errant-4:
I apologize for the long wait, particularly about such a tiny change to the PR. Yes, that is what I meant, explanation below. I'll move your PR forward once you merge master to unconflict and make that change. Or let me know if you would rather have me deal with it, at this point.

The context/explanation: You are correct that MindRoleGhostRoleSoloAntagonist holds the correct role type, and is often used on ghostroles. But it's a generic mindrole which is used when/because a more specific one is not available. It has some current and future issues/limitations related to playtime and role ban tracking, so if a more specific mind role is available, as in this case, it should be used instead. It's not actually going to be read by anything from that particular datafield, but for consistency it should match what mindrole this entity has defined elsewhere. Hopefully, one day I/someone will get around to revamping how ghostrole code works, so it will not be mandatory to specify a mind role there that is in some cases, not going to be used by anything and thus serves to potentially create confusion 

(Internal receipt for LoneOps being solo antag, in case anyone ever asks: https://discord.com/channels/310555209753690112/1281461283157184532/1313145617735880714)

Errant-4:
(single-approve bugfix clause)

PR #36523, opened 2025-04-14
Fix to prevent the appearance of a previous character transferring over in ghost roles where it shouldn't

### LINKED ISSUE #33713: Ninja role is your character, but any reference to your character is illegal?
## Description
I have seen multiple times where the ninja played their character, but instead of being Nanotrasen aligned, they're Spider Clan, which makes sense, but it opens up a *very confusing* scenario, granted rare, where someone who died IC, in round, manages to get the role, and now... they're playing as their character, who died, as a ninja.

Now, this wouldn't be much of an issue, if it wasn't for the ghost role rules specifically stating that "You are absolutely not allowed to remember, say, the name, appearance, etc. of your previous character.
Screenshot 2024-12-03 220215

Which is an issue, because, in every way but name (Probably not DNA or Prints), you ARE the previous character if you died on that shift.

Related to that, it also means that players who you have a lot of interaction with in rounds would recognize you from your details, like your hair, the color or design of your tail, or maybe the color and design of your wings. Which, again, wouldn't be much of an issue, but this indirectly encourages people to regard you as *that very same character*, which in turn, makes it harder to disconnect the ninja from 'The guy I see on station every other shift'

This also means that from the crew's perspective, if you see someone who is obviously that guy you know, but a ninja, you're GOING to treat that differently than any other ninja. Instead of calling out "NINJA ATMOS" They might instead go "Urist?" because that just makes sense in character. Which also means, yes, now the ninja was outed as the character they normally play, despite being technically an entirely different character.

Again, this isn't really an issue, but it becomes an issue when you consider that *any other ninja would likely be outted just because they don't recognize them.*

This might end up for debate entirely, but I think I made it pretty clear what the issue is.

**Reproduction**
1. Be spectator/die in round
2. Roll ninja
3. Be you, the special, fantastic person you are
4. Watch as several people recognize you from your unique character traits

**Screenshots**
Screenshot 2024-12-03 223444

Zokkie:
## About the PR
Resolves #33713 and two more ghost roles that have the same problem.

1. Join the raffle for the Ninja ghost role
2. Get told by the game you are absolutely NOT allowed to remember the appearance of you previous character.
3. Click the button to accept and the game spawns you in having the exact same appearance of your previous character.

I assume this is not intended so I made some changes that the midround ghost roles Ninja, Wizard and Lone Operative will now have a randomized appearance when you take them (You still get to keep your species).

I also updated the Lone Operative rules description to more match the other antagonists rules description when you try to take the ghost role as it did not state at all you could not remember your previous life (which I assume should be the case if you take the  Lone Operative role).

I want to make it extra clear I did not make any changes to roundstart antagonists.  That code is still the same and will still use the character you customized.

## Why / Balance
Seems like a mistake the rules tell you to forget about your previous appearance/character but you still get the exact same appearance from your previous character.

Loneop ghost roles also not having a "forget your previous life" disclaimer also seems a mistake considering almost all other antagonists ghost roles have it

## Technical details
In events.yml the Ninja, Wizard and Lone Operative ghost role created by midround events now have the RandomHumanoidAppearance type attached to it 

In ghost-role-compenent.ftl changed Lone Operative ghost role rules description to also include you do not remember your previous character just like in Ninja/Wizard ghost role rules description.

## Media
I join the round as Captain Micheal Mitchell.
OriginalCharacter - Copy

Woeps, seems my character had an accident! Time to take this Lone Operative ghost role I guess!
JoiningAsLoneOp - Copy

Time to nuke the station with my new character that is in no way related to my previous character!
NewPerson - Copy

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- fix: Lone Operative ghost role rules description now explains that you do not remember the life of your previous character.
- fix: If you take a Ninja, Wizard or Lone Operative ghost role it will now randomize the appearance of your new character instead of using the exact appearance of your previous character.


slarticodefast:
Why does it use the character appearance from your profile in the first place?
`RandomHumanoidAppearance` only overwrites it after spawn.

Zokkie:
@slarticodefast your question made me dig around in the code even deeper and I figured out that it seems most of the magic happens in here:
`Content.Server/GameTicking/Rules/AntagLoadProfileRuleSystem.cs `

It always tries to load the player their profile for antags that use this system like Ninja/Wizard/Nukeops/Loneops instead of randomizing it.

I still lack a bit of experience but I guess after having this new perspective on the code it seems better to use the code in the AntagLoadProfileRuleSystem and tell it we want a random character instead of doing it after creating the character?

MozarteanChaos:
> I don't think this should be merged in its current form, myself and lot of other people really enjoy having set character appearances for antags like ninja and loneop. It lets you RP as a different version of that character. If only did a random appearance if you already spawned that round I would understand, but this removes spawning with your character altogether, which I think loses a lot of potential RP.

Agreed. Personally, if I observe a round and spot - for example - a wizard ghost role, I'd prefer to take that role as the character I had selected in the lobby, rather than as an entirely new character I have no attachment to and have to make up the personality of on the spot. Losing that ability wouldn't be the end of the world, but it would be kinda sad.
In other words, I think this PR is good for the situation it set out to prevent, but not so good for observers looking for ghost roles.

northerndakotas:
While I don't necessarily agree with their decision, an admin clarified on 3/25/25 that a nukie disguising themselves as their own character (With Voice mask, Agent ID) is metagaming. So I think randomizing not just lone ops but roundstart nukies too would be the best way to mechanically enforce this rule. I know I've accidentally broken it because I already looked like my character so I had assumed that I was supposed to be the same person.

slarticodefast:
Hey, I brought it up for a maintainer discussion and the consensus was that it is mostly harmless and players seem to enjoy playing as their own character as a ninja or nukeops. While metafriending can be a problem it is considered minor enough in this case.
As such I will be closing this PR.

PR #36530, opened 2025-04-14
Add InRangeUnobstructed check to storage transfer

metalgearsloth:
TryPickup doesn't have this in its codepath which I thought it did.


metalgearsloth:
Better idea

PR #36533, opened 2025-04-14
Storage fixes

metalgearsloth:
- Add size event.
- Fix reclaim not running always.
- If storage close it stops dragging the piece.

Fixes ItemSize breaking bags by both disabling it and adding an event so anyone that wants them toggleable in backpacks still has the behavior fixed. Also fixes a bad storage bug and stops drags if the storage has closed. Apparently not being able to toggle in bags is standard 13 behavior though I imagine we may want exceptions for some stuff at some point which is why it still validates it anyway.

:cl:
- tweak: Toggleable items such as stunbatons and eswords can no longer be toggled in storage.


cfif126:
Can items be untoggled in storage?

metalgearsloth:
> Can items be untoggled in storage?

Oop forgot I reverted that but yeah that's a separate pr

slarticodefast:
@metalgearsloth please don't just self-merge
ping me if you need an urgent review

PR #36537, opened 2025-04-14
Fix visitor mindshields

M1tht1c:
## About the PR
Made the mindshields visitors get actually removeable, and gives mindshields to the disaster victim heads.

## Why / Balance
Makes no sense that visitors would have unremoveable mindshields, or that disaster victims would lose theirs.

## Technical details
YAML changes only 

Changed ```EventHumanoidMindShielded``` to use ```AutoImplant```, removed redundant ```AutoImplant``` components in children, and made disaster victims actually children of ```EventHumanoidMindShielded```

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: You can now remove visitors mindshields.
- fix: Disaster victims now have mindshields as appropriate.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/ShuttleRoles/roles.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/ShuttleRoles/roles.yml
@@ -13,9 +13,6 @@
       state: captain
     - type: RandomHumanoidSpawner
       settings: VisitorCaptain
-    - type: AutoImplant
-      implants:
-      - MindShieldImplant
 
 - type: entity
   id: RandomHumanoidVisitorCE
@@ -26,9 +23,7 @@
       state: ce
     - type: RandomHumanoidSpawner
       settings: VisitorCE
-    - type: AutoImplant
-      implants:
-      - MindShieldImplant
+
 
 - type: entity
   id: RandomHumanoidVisitorCMO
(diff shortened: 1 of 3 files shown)


M1tht1c:
As it stands, this also makes CentComm ghost roles convertible too. While https://github.com/space-wizards/space-station-14/pull/35828 would fix this, I can also make a temporary fix in this PR without needing to add a new implant if people think its necessary (by just adding the mindshield component, rather than an implant, to the CentComm settings)

IProduceWidgets:
Ahh thanks. Think these auto-implants didn't exist when I first made the visitors.

I'm not sure we want any heads to have them removable though. That feels inconsistent, and until we get a rev rework where they aren't required to die, it doesn't make sense for regular heads to be removable.

M1tht1c:
> I'm not sure we want any heads to have them removable though. That feels inconsistent, and until we get a rev rework where they aren't required to die, it doesn't make sense for regular heads to be removable.

When they made mindshields removable, it was intended that the heads become convertable (revs victory tracks whether the heads are converted now).

IProduceWidgets:
I think we should keep them unremovable for heads -- even the event ones -- for consistency sake, and because with the way revs' end condition works having heads get converted will break the gamemode. At least until that gets a rework.

Errant-4 (approved):

SlamBamActionman (approved):
Works on my machine

Issue #36536, opened 2025-04-14
TileEmissions need their own comptree


metalgearsloth:
When rendering the entire station it was about 10% of performance. Most of these don't move so shouldn't be a huge cost.

PR #36538, opened 2025-04-14
Use CompTree for TileEmission

metalgearsloth:
Was like 8.5% of full-station render, mostly just helps with bigger viewports though with the engine light query PR this will also help for that as we can quickly check particular points.

Might need to move the tree comp to shared at some point if the engine pr gets merged.

Resolves https://github.com/space-wizards/space-station-14/issues/36536


TheShuEd:
> Does this let us have longer ranged scopes

absolutely not, this PR has nothing to do with that.

metalgearsloth:
> Does this let us have longer ranged scopes

da heck this got to do with scopes

VerinSenpai:
well you see, if we CompTree TileEmission, it stands to reason that scope ranges get at least doubled. Its just simple math.

Issue #36539, opened 2025-04-14
PoweredLight interactions aren't predicted


metalgearsloth:
In the year of our lord 2025.

PR #36541, opened 2025-04-14
Predict PoweredLights

metalgearsloth:
BC is moving a lot of it to shared, poweredlight + lightbulb, and args changes. No datafield breakages.

Resolves https://github.com/space-wizards/space-station-14/issues/36539

# Breaking Changes
`PoweredLightComponent` has been moved to `Content.Shared.Light.Components`.

:cl:
- add: Inserting lights now plays an insert animation.


slarticodefast:
I pushed some review changes. But I found a bug that needs to be fixed before this can be merged.
grafik
To reproduce simply walk through the portal to the engi test area.
Something is causing the lights to not toggle correctly when going through the portal.

PR #36542, opened 2025-04-14
Allowing Cats to walk

archee1:
## About the PR
Allows cats to walk and avoid slipping on puddles/hazards.

## Why / Balance
While fairly arbitrary in practice, most animals are unable to walk, however specifically aimed for syndicats, this allows them to walk to avoid tripping on welder spills or syndie soap ectetera.

## Technical details
Adds component MovementSpeedModifier to the MobCat prototype with values of 2/4 for walking/sprinting.
(Sprinting speed unchanged from parent inheritance)

## Requirements

**Changelog**

:cl:
- tweak: Cats can walk slowly to avoid slipping on puddles.
- fix: Cak's (cake cats) previously incorrect sprinting and walking speeds have been corrected

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
@@ -3035,6 +3035,9 @@
         Slash: 6
         Piercing: 6
         Structural: 15
+  - type: MovementSpeedModifier
+    baseWalkSpeed: 2
+    baseSprintSpeed: 4
 
 - type: entity
   name: space cat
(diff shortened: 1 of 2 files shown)


Killerqu00:
changelogs should be in present passive and shouldn't be IC. Also, uh, the sentence seems unfinished?
for example, I would suggest:
tweak: Syndicats can walk on puddles without slipping

archee1:
I did consider at first putting this on the base MobCat, however chose against it. Primarily because most other animals are also incapable of walking, though hamlet is able, it does seem fairly arbitrary. I'll move it down to base MobCat for now, though if its fine should we consider moving it to the BaseSimpleMob that all animals/mobs parent off of? I cant think of any mobs that don't have the ability to walk outside of simply not being able to.

beck-thompson (approved):
makes sense to me!

Issue #27748, opened 2024-05-07
Shoving sound and animation works out of range


lzk228:
https://github.com/space-wizards/space-station-14/assets/124214523/63a3066f-11f9-4dd3-a201-1a7f443bcd1f

Hmeister-real:
intended? your just shoving the air your meant to be allowed to shove towards someone (e.g. for missing a shove)

lzk228:
> intended? your just shoving the air your meant to be allowed to shove towards someone (e.g. for missing a shove)

- shoving appears only if pointed on the entiity
- shoving available even if you cant see the entity
- animation appears only if entity is shoveable, sound - always

themias:
On testing, it only plays the animation/sound when within range. You're right that it was likely fixed by #36546.

PR #36546, opened 2025-04-14
Refactor disarms

metalgearsloth:
- Move client stuff to shared
- Cleanup a bunch of stuff
- Ref events
- Fix the swing sound mispredict (I noticed it on target dummies).

## Breaking changes
- Disarm events are now raised by ref.
- DIsarm was moved fully to shared.
- DisarmMalusComponent was moved to shared.
- CalculateDisarmChance was moved to shared.

:cl:
- fix: Fix disarm swing sound even if a disarm isn't possible on the target.


lzk228:
i suppose it fixes https://github.com/space-wizards/space-station-14/issues/27748

metalgearsloth:
> i suppose it fixes #27748

Yeah I would guess so now that inrange is there?

Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
```diff
@@ -793,8 +826,112 @@ protected virtual bool DoDisarm(EntityUid user, DisarmAttackEvent ev, EntityUid
...
+
+        // At this point we diverge
+        if (_netMan.IsClient)
+        {
+            // Play a sound to give instant feedback; same with playing the animations
+            _meleeSound.PlaySwingSound(user, meleeUid, component);
+            return true;
+        }
+
+        if (_random.Prob(chance))
```
slarticodefast, line 891: this will desync between client and server
randomPredicted when?

I did not look at the rest

MilonPL: This is fine, client returns above this in `_netMan.IsClient` so wont ever reach this code. As to RandomPredicted when: right now.

Issue #36548, opened 2025-04-14
Hunger wildly mispredicts


metalgearsloth:
Client was at 126 and server was at 96, wildly off.

Steps to reproduce:
1. Use sericulture action on arachnid.
2. Client now predicts it can do it again and server doesn't think so.

PR #36549, opened 2025-04-14
Fix hunger + thirst mispredicts

metalgearsloth:
Resolves https://github.com/space-wizards/space-station-14/issues/36548


FairlySadPanda (commented):
Just throwing a lot of dirties at it definitely helps, and I noticed this a lot during testing for learning med stuff so this should help.

Once we get onto organ-izing this stuff we'll need to nail down the prediction/replication/messaging stuff for all this without needing to add dirties in numerous places to cope.

metalgearsloth:
I've started work on a helper method to make it easier but nothing has come to fruition yet. This has deltas just because some of the states were rather fat and it was easy to test.

Content.Shared/Nutrition/EntitySystems/ThirstSystem.cs
```diff
@@ -49,13 +49,19 @@ private void OnMapInit(EntityUid uid, ThirstComponent component, MapInitEvent ar
...
         component.NextUpdateTime = _timing.CurTime;
         component.CurrentThirstThreshold = GetThirstThreshold(component, component.CurrentThirst);
         component.LastThirstThreshold = ThirstThreshold.Okay; // TODO: Potentially change this -> Used Okay because no effects.
         // TODO: Check all thresholds make sense and throw if they don't.
         UpdateEffects(uid, component);
 
+        DirtyField(uid, component, nameof(ThirstComponent.NextUpdateTime));
+        DirtyField(uid, component, nameof(ThirstComponent.CurrentThirstThreshold));
+        DirtyField(uid, component, nameof(ThirstComponent.LastThirstThreshold));
+
```
MilonPL, line 64: Ugly, why not DirtyFields >:(

metalgearsloth: Well you see I was cooking on DirtyField that also took in a field ref but it's currently in limbo while I figure it out :3

MilonPL (approved):
good enough

PR #36553, opened 2025-04-14
Fix interaction popups

metalgearsloth:
One IRU codepath was using popupentity not popupclient so it duplicated. RCD also manually checked this so you got this weird codepath where the interaction is considered a valid range but RCD does not consider it a valid range.

@chromiumboy 

Is it intended that RCD range is slightly smaller than interaction range? Seems like it would be easy to just skip the RCD range check to make it consistent even if the middle-of-tile-length is further.

### DIFF
--- a/Content.Shared/Interaction/SharedInteractionSystem.cs
+++ b/Content.Shared/Interaction/SharedInteractionSystem.cs
@@ -953,7 +953,7 @@ namespace Content.Shared.Interaction
             if (!inRange && popup && _gameTiming.IsFirstTimePredicted)
             {
                 var message = Loc.GetString("interaction-system-user-interaction-cannot-reach");
-                _popupSystem.PopupEntity(message, origin, origin);
+                _popupSystem.PopupClient(message, origin, origin);
             }
 
             return inRange;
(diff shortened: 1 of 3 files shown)


chromiumboy:
> One IRU codepath was using popupentity not popupclient so it duplicated. RCD also manually checked this so you got this weird codepath where the interaction is considered a valid range but RCD does not consider it a valid range.
> 
> @chromiumboy
> 
> Is it intended that RCD range is slightly smaller than interaction range? Seems like it would be easy to just skip the RCD range check to make it consistent even if the middle-of-tile-length is further.

IIRC, I did it so the RCD could only targeted tiles a single tile away, cardinal or ordinal

Content.Shared/Interaction/SharedInteractionSystem.cs
```diff
@@ -953,7 +953,7 @@ public bool InRangeUnobstructed(
             if (!inRange && popup && _gameTiming.IsFirstTimePredicted)
             {
                 var message = Loc.GetString("interaction-system-user-interaction-cannot-reach");
-                _popupSystem.PopupEntity(message, origin, origin);
+                _popupSystem.PopupClient(message, origin, origin);
```
slarticodefast, line 956: a lot of server side systems call `InRangeUnobstructed` and this will just kill the popup for them because `PopupClient` does nothing on the server
```suggestion
                _popupSystem.PopupEntity(message, origin, origin);
```

slarticodefast (changes requested):

slarticodefast:
Closing as derelict.

PR #31045, opened 2024-08-15
No meta implanters + other

lzk228:
## About the PR

- remove names from implanters, so thay all legal will be just `implanter`, syndi will be `syndicate implanter`
- they will have suffixes
- give admeme implanter small inject and draw time
- fix locale

- also that means contraband components was removed from tracking and mindshield implanters
- all syndi implanters still contraband, tho they even have different sprite

- more info in Media part 
 
## Why / Balance

reduce meta, because even if you use an implanter, someone still can see which implant was in the implanter

## Technical details

no

## Media

**MEDIA A LITTLE OUTDATED**
syndicate implanters are named `syndicate implanter`

**Before**
Screenshot_20240815_155924
**After**
Screenshot_20240815_161037

even if you can't see which implant is inside immediately
Screenshot_20240815_161102

you can examine the implanter
Screenshot_20240815_161128

or take in hand
Screenshot_20240815_161111
 
to see which implant it contains

## Requirements

- [X] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [X I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- tweak: Removed names from implanters. No more meta. (You still can see which implant is inside via examining or holding implanter in hand).
ADMIN:
- tweak: Admeme implanter injects and draws faster than default implanter.

PR #36555, opened 2025-04-14
Add implant names to implanters

SlamBamActionman:
## About the PR

This PR (re)introduces implant names to implanters, allowing them to be identified by the entity name even when no longer containing the implant.

It also gives tips in the implanter description and as a gameplay tip that you should wipe the implanter after use.

## Why / Balance

Back when I made the PR that changed implanters and their extraction, a goal was to keep a balance between antags and Security where Sec are strongly discouraged from doing outright guesswork when it comes to implant checking, yet have no barriers if they are certain they are correct.
The PR was also intended to allow Security to do educated guesses - using deductive reasoning to extract a suspected implant, but still preventing randomly guessing.

I think these goals were met. However, since we were uncertain of if removing the metashield would impact balance in Sec's favor, we decided to be conservative via keeping labels off to see the effects, even with concerns raised by contributors. 

This has resulted in a situation that isn't an issue from a rules-perspective, but simply put is unfun. And that is Security finding an implanter, figuring out who used it (via DNA or other evidence), and then being unable to act on it. While Sec use this to mark them as suspicious or bring them in for the crime of using syndie contra, it leaves a bad taste where Sec managed to secure easily hideable evidence yet don't get anything from it. An antag can drop off an implanter at Sec's doorstep and suffer no consequence.

---

Labels on implanters would throw Security a bone in this regard. The main concern here is that it would bring back some semblance of random implant checking, in that after Sec finds a scrubbed implant, they will just start implant checking anyone they arrest for that implant. There is also an impact on people who fail to scrub their implant (likely due to being new, but can also be stress/arrogance) who can now have their implant removed.

While these are undoubtedly antag nerfs, I think they allow for some more round dynamicity and also makes hiding the implant actually important. I also don't think the first concern is going to be prevalent; under the old implanter system for sure, but the cost of guessing wrong is high now, even if you know the implant you're looking for. It's also a system players are familiar with, helps new Sec players know what implants to look out for, and further mellows out popular implants by making them indirectly more at risk.

---

The description tip is needed to help new players who have not played Detective know the importance of cleaning your implanter, something that isn't very evident otherwise. 

There have been discussions to allow recyclers to be able to destroy implanters again, though this PR will be added first, with a later evaluation seeing if the recycler change is desirable. 

## Technical details

Reverts some of the changes in #31045 in regards to implant names.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Implanters are now named after the implant they contained, even after the implanter has been used.
- add: Syndicate implanters now remind you to scrub them in the item description.

### DIFF
--- a/Resources/Locale/en-US/tips.ftl
+++ b/Resources/Locale/en-US/tips.ftl
@@ -136,3 +136,4 @@ tips-dataset-135 = Instead of picking it up, you can alt-click food to eat it. T
 tips-dataset-136 = If you're trapped behind an electrified door, disable the APC or throw your ID at the door to avoid getting shocked!
 tips-dataset-137 = If the AI electrifies a door and you have insulated gloves, snip and mend the power wire to reset their electrification!
 tips-dataset-138 = If you want to stop your prisoner from escaping from the cell right after being uncuffed, turn on combat mode while uncuffing - this will shove the prisoner down.
+tips-dataset-139 = Make sure to clean your illegal implanters with a soap or a damp rag after you use them! Detectives can scan used implanters for incriminating DNA evidence, but not if they've been wiped clean.
--- a/Resources/Prototypes/Datasets/tips.yml
+++ b/Resources/Prototypes/Datasets/tips.yml
@@ -2,4 +2,4 @@
   id: Tips
   values:
     prefix: tips-dataset-
-    count: 137
+    count: 139
--- a/Resources/Prototypes/Entities/Objects/Misc/implanters.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/implanters.yml
@@ -157,7 +157,7 @@
 
 - type: entity
   id: SadTromboneImplanter
-  suffix: sad trombone
+  name: sad trombone implanter
   parent: BaseImplantOnlyImplanter
   components:
     - type: Implanter
@@ -165,7 +165,7 @@
 
 - type: entity
   id: LightImplanter
-  suffix: light
+  name: light implanter
   parent: BaseImplantOnlyImplanter
   components:
     - type: Implanter
@@ -173,7 +173,7 @@
 
 - type: entity
   id: BikeHornImplanter
-  suffix: bike horn
+  name: bike horn implanter
   parent: BaseImplantOnlyImplanter
   components:
     - type: Implanter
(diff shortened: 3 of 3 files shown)


SlimmSlamm:
Year of the Implanter Nerfs.

lzk228:
i find it inconvenient that an implanter still keeps its name even after it doesn't have this implant inside no more

Sarahon:
Didn't we change this before to avoid the very easy detection?

SlamBamActionman:
> Didn't we change this before to avoid the very easy detection?

The original addition/removal of this feature was during the old implanter system where random checking was much easier and had to be strongly enforced via rules instead of mechanics.

keronshb:
> i find it inconvenient that an implanter still keeps its name even after it doesn't have this implant inside no more

This has a caveat of being tested and then possibly reintroducing recycling implanters if it's too easy.

walksanatora:
I feel like there is no need to make implants recyclable. there are many ways to get rid of a implant(spacing,department hiding, lizard plushie) and if you just soap it random guessing will still hurt sec. (you know how rare it is for chemist to make genetic fixing chems? and then have to burn caustic and whatever the other dammage type is phlananx uses)

keronshb:
> I feel like there is no need to make implants recyclable. there are many ways to get rid of a implant(spacing,department hiding, lizard plushie) and if you just soap it random guessing will still hurt sec. (you know how rare it is for chemist to make genetic fixing chems? and then have to burn caustic and whatever the other dammage type is phlananx uses)

Spacing is near useless, it can be found relatively easily by salvage.
Hiding it in departments is esoteric knowledge. You'd need to make sure you hide it away from a place a tider won't check since they're aggressive at finding things.
Soaping does not matter with this PR since it gives them a guaranteed way to check what implant it is.

FairlySadPanda (approved):
I've argued this point (probably past the point of toxicity) with Slam about this - I think this is a good experiment to try now that the initial implant changes have had time to settle.

I'd still like to see implanters made conceptually cooler as a fluff thing - I've got a thread about it on the forum - but this _probably_ resolves the remaining niche edge-case unfun stuff.

slarticodefast (approved):

slarticodefast:
> Didn't we change this before to avoid the very easy detection?

When the labels were removed there were different circumstances than now with the implanter draw rework. Reintroducing the labels was always something we had in mind, but like we mentioned in the previous PR that tried to reintroduce them with different mechanics, we wanted to wait a little to evaluate how players adjust before making further changes.
Since some time has passed since then I think we can give this a try and see how it goes. We will of course keep an eye on player feedback.

PR #36558, opened 2025-04-14
fix inspect mind verb

slarticodefast:
## About the PR
It was only working for the local player.

## Why / Balance
bugfix

## Technical details
Move it to the server and execute the command remotely for the client since the vv command is client side only.
Mind entities are stored in nullspace, so they are not networked by default and the client was failing to get the net entity for the command.
It was only working for the local player's mind because that one has a PVS override.

## Media
Screenshot (390)
Screenshot (391)

## Requirements

## Breaking changes
none

**Changelog**
:cl:
ADMIN:
- fix: Fixed the inspect mind command not working for other players.


ScarKy0 (approved):
Tested and works, thanks a lot!

PR #36594, opened 2025-04-15
make ShowInChat in EmoteEntityEffect true by default and change EmoteEntityEffect code

lzk228:
## About the PR

ShowInChat in EmoteEntityEffect is true by default
if it's true, there will be a message above the character's head and a message in the chat
if it's false, there will be only a message above the head, but it won't go in chat 

## Why / Balance

i always thought: why can't you see who is emoting while metabolizing a reagent, so there it is, fixing it because it look at least weird

## Technical details

no

## Media

don't wanna

## Requirements

## Breaking changes

no i suppose

**Changelog**

:cl:
- tweak: If a reagent make you emote, it will be shown in chat.

### DIFF
--- a/Content.Server/EntityEffects/Effects/Emote.cs
+++ b/Content.Server/EntityEffects/Effects/Emote.cs
@@ -17,7 +17,7 @@ public sealed partial class Emote : EntityEffect
     public string? EmoteId;
 
     [DataField]
-    public bool ShowInChat;
+    public bool ShowInChat = true;
 
     [DataField]
     public bool Force = false;


slarticodefast:
>i always thought: why can't you see who is emoting while metabolizing a reagent, so there it is, fixing it because it look at least weird

A weh foam bomb will make the chat unreadable.

lzk228:
well it is adjustable with setting false to certain reagents

Dutch-VanDerLinde:
we should really add a thing in chat so duplicate messages appear as "John Urist screams!" x2 like popups and shit

slarticodefast:
There are a lot more emote effects than these two. I would prefer to keep it disabled by default, and only activate it in selected cases. This is also a breaking change.

lzk228:
applied the review but really don't know what to write in breaking changes

slarticodefast:
Not a breaking change anymore since the default value is no longer changed.

slarticodefast:
Closing this as derelict since it has not been fixed for several months.
Conceptually I still think it's better to not show reagent emotes in chat to prevent them from being spammed with foam bombs.

PR #36597, opened 2025-04-15
Material Door rebalancing

archee1:
## About the PR
This PR targets material doors (not airlocks or firelocks)
Aims to tweak constructible material doors to encourage their use.

- Changes the de/construction time for all of them from 15s -> 8s
- halves their material costs from 20 -> 10
- destroying them returns a small portion of the materials put in
- lowers their overall health and resistances
- adds sounds when destroying them

## Why / Balance
For the most part these material doors are rarely built or mapped into stations, hopefully this should encourage them to be used at the cheaper cost and be a better alternative when needing a low security door for a construction project.

Paper doors are difficult enough already to get the materials to craft, mostly only being built when a stack of 30 paper is mapped on a station by default (bagel), then disappointingly only creates 1 door leaving 10 paper left over with no use.

Lowering their resistances and health across the board make sense given more doors being buildable out of the same amount of surplus materials, their previous resistances also didn't make much sense given they can be crafted by hand with no tools.

## Technical details
destructible threshold on metal doors lowered from 200 to 150 (also lowers the wood/paper/web doors slightly)
added thresholds for nuke-level damage to destroy doors without dropping materials
changes the damageModifierSet from StructuralMetallicStrong -> StrongMetallic
Reduces construction recipes by half (most cases 20 -> 10) and all doafter times from 15s -> 8s
Adds break sounds to all doors (except web, It uses foliage rustle for opening its kind of an outlier to the other doors)
*also shifts around some of the prototype/component orders

## Requirements

**Changelog**
:cl:
- add: Material doors now have destruction sounds and will drop a portion of their construction materials when destroyed.
- tweak: Material doors now have reduced health, resistances, construction time, and construction costs

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Doors/MaterialDoors/material_doors.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/MaterialDoors/material_doors.yml
@@ -51,7 +51,7 @@
     thresholds:
     - trigger:
         !type:DamageTrigger
-        damage: 200
+        damage: 300
       behaviors:
       - !type:DoActsBehavior
         acts: ["Destruction"]
@@ -68,6 +68,8 @@
   components:
   - type: NavMapDoor
 
+### Metal doors ###
+
 - type: entity
   id: MetalDoor
   name: metal door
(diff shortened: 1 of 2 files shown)


iaada:
Thank you for the change! Sorry about the delay <3

PR #36618, opened 2025-04-16
Predict more entities

metalgearsloth:
The spicier ones that still don't require v2. No CL because I'm lazy and no one's gonna notice this besides me.

Requires https://github.com/space-wizards/RobustToolbox/pull/5859


VerinSenpai:
I'll notice it Sloth sensei.

metalgearsloth:
Oh here's where I left you.

PR #36619, opened 2025-04-16
Predict blocking examine + fix blocking breaking persistence

metalgearsloth:
No idea why these were like this.

technical:
- stateful data needs to be datafields or it breaks persistence.
- Never use .IsServer unless you have a very good reason.
- PopupPredicted

### DIFF
--- a/Content.Shared/Blocking/BlockingSystem.cs
+++ b/Content.Shared/Blocking/BlockingSystem.cs
@@ -313,7 +313,7 @@ public sealed partial class BlockingSystem : EntitySystem
 
     private void OnVerbExamine(EntityUid uid, BlockingComponent component, GetVerbsEvent<ExamineVerb> args)
     {
-        if (!args.CanInteract || !args.CanAccess || !_net.IsServer)
+        if (!args.CanInteract || !args.CanAccess)
             return;
 
         var fraction = component.IsBlocking ? component.ActiveBlockFraction : component.PassiveBlockFraction;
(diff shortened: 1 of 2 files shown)


slarticodefast (changes requested):
`if (component.BlockingToggleAction != null)`
remove this check, having that there makes no sense an is causing the compile error

slarticodefast:
I fixed it and made the popups predicted

keronshb (changes requested):
This needs to follow the template like other PRs please.

Content.Shared/Blocking/BlockingSystem.cs
```diff
@@ -156,61 +150,53 @@ public bool StartBlocking(EntityUid item, BlockingComponent component, EntityUid
...
-            //Don't allow someone to block if someone else is on the same tile
-            var playerTileRef = xform.Coordinates.GetTileRef();
-            if (playerTileRef != null)
+        //Don't allow someone to block if someone else is on the same tile
+        var playerTileRef = xform.Coordinates.GetTileRef();
+        if (playerTileRef != null)
+        {
+            var intersecting = _lookup.GetLocalEntitiesIntersecting(playerTileRef.Value, 0f);
+            var mobQuery = GetEntityQuery<MobStateComponent>();
+            foreach (var uid in intersecting)
```
metalgearsloth, line 173: Oh yeah this is super silly but future concern.

metalgearsloth:
> `if (component.BlockingToggleAction != null)` remove this check, having that there makes no sense an is causing the compile error the rest looks good

I think I fatfingered that

slarticodefast:
At least we got a technical details section now.
But please keep the template for future PRs.

PR #36621, opened 2025-04-16
Entity storage prediction thing

metalgearsloth:
oldcode I don't think anything even uses these codepaths upstream.

### DIFF
--- a/Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
@@ -202,7 +202,7 @@ public abstract class SharedEntityStorageSystem : EntitySystem
         }
     }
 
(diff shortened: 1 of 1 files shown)


Content.Shared/Storage/EntitySystems/SharedEntityStorageSystem.cs
```diff
@@ -202,7 +202,7 @@ public void EmptyContents(EntityUid uid, SharedEntityStorageComponent? component
         }
     }
 
-    public void OpenStorage(EntityUid uid, SharedEntityStorageComponent? component = null)
+    public void OpenStorage(EntityUid uid, SharedEntityStorageComponent? component = null, EntityUid? user = null)
```
slarticodefast, line 205: Hmm, there are systems callign this but they are all on the server.
Adding the user there would then cause the user to no longer hear the sound since the server thinks it was predicted.
Same if we passed on the user from `TryOpenStorage`. Not sure how to best solve this.

PR #36624, opened 2025-04-16
Obsolete snapgrid helpers

metalgearsloth:
Really should be on engine. Requires https://github.com/space-wizards/RobustToolbox/pull/5843

### DIFF
--- a/Content.Shared/Coordinates/Helpers/SnapgridHelper.cs
+++ b/Content.Shared/Coordinates/Helpers/SnapgridHelper.cs
@@ -1,46 +1,28 @@
-using System.Numerics;
+using JetBrains.Annotations;
 using Robust.Shared.Map;
 using Robust.Shared.Map.Components;
 
-namespace Content.Shared.Coordinates.Helpers
+namespace Content.Shared.Coordinates.Helpers;
+
+public static class SnapgridHelper
 {
-    public static class SnapgridHelper
+    [Obsolete("Use SharedTransformSystem.SnapToGrid")]
+    public static EntityCoordinates SnapToGrid(this EntityCoordinates coordinates, IEntityManager? entMan = null, IMapManager? mapManager = null)
     {
-        public static EntityCoordinates SnapToGrid(this EntityCoordinates coordinates, IEntityManager? entMan = null, IMapManager? mapManager = null)
-        {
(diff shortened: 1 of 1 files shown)


Content.Shared/Coordinates/Helpers/SnapgridHelper.cs
```diff
@@ -1,46 +1,28 @@
...
-        }
+        IoCManager.Resolve(ref entMan, ref mapManager);
+        return entMan.System<SharedTransformSystem>().SnapToGrid(coordinates);
+    }
 
-        public static EntityCoordinates SnapToGrid(this EntityCoordinates coordinates, MapGridComponent grid)
-        {
-            var tileSize = grid.TileSize;
+    [PublicAPI]
+    public static EntityCoordinates SnapToGrid(this EntityCoordinates coordinates, MapGridComponent grid)
```
slarticodefast, line 17: is this one not getting obsoleted?

metalgearsloth: This one's not really harmful I was eh on it I could move it to engine.

slarticodefast: would probably be best to move the whole file so we don't end up with weird mixed imports again

slarticodefast: wait, you already did move it in the engine PR
grafik

metalgearsloth: uhh lag

Content.Server/Explosion/EntitySystems/SmokeOnTriggerSystem.cs
```diff
@@ -42,7 +42,7 @@ private void OnTrigger(EntityUid uid, SmokeOnTriggerComponent comp, TriggerEvent
             return;
 
         var coords = grid.MapToGrid(mapCoords);
```
slarticodefast, line 44: do we even need all of this?
Can't we directly use the other overload of `SnapToGrid` on coords of the uid?

slarticodefast (commented):
One part looks weird, the rest is good.

Tayrtahn:
Cleaned up merge conflicts. Engine PR needs to merge master.

Issue #35953, opened 2025-03-20
Bandanas (hat state) no longer fit in crates and closets


K-Dynamic:
## Description

Bandanas in their hat state will appear outside of containers like closets and crates when closed.

Bandanas in their mask state will fit without issue.

**Reproduction**

1. Open container with open and closed states (crate or closet).
2. Put bandana (hat state) into container.
3. Close container.
4. Bandana appears outside.

PR #36625, opened 2025-04-16
Fix certain foldable items not being possible to store in crates/lockers

SlamBamActionman:
## About the PR

Some objects such as bandanas and zippable coats wouldn't get inserted into crates, and instead remain outside when the container closed. This PR fixes that.

Fixes https://github.com/space-wizards/space-station-14/issues/35953.

## Why / Balance

Bugfix.

## Technical details

*As far as I can tell* the cancel behavior of `InsertIntoEntityStorageAttemptEvent` subscribed to in `FoldableSystem` is already fully covered by `ContainerGettingInsertedAttemptEvent`, so removing it seems fine. From testing all items appear to keep their desired behavior.

Can be tested by spawning in foldable items and trying to close a locker around them.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Unfolded bandanas/lab coats now fit in crates again.

### DIFF
--- a/Content.Shared/Foldable/FoldableSystem.cs
+++ b/Content.Shared/Foldable/FoldableSystem.cs
@@ -9,6 +9,7 @@ using Robust.Shared.Utility;
 
 namespace Content.Shared.Foldable;
 
+// TODO: This system could arguably be refactored into a general state system, as it is being utilized for a lot of different objects with various needs.
 public sealed class FoldableSystem : EntitySystem
 {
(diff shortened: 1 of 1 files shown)


slarticodefast:
Rollerbeds now also fit into crates in their folded state, but I think that makes sense.
Like you said, the removed event is already covered by the other insertion check.

slarticodefast (approved):

PR #36626, opened 2025-04-16
Predict GhostBoo

### LINKED ISSUE #36540: GhostBoo needs predicting
Title.

ljm862:
## About the PR

Predict GhostBoo. 
**Requires Chat refactor and #36541 - on hold until then.**

## Why / Balance

Resolve #36540

## Technical details

Move GhostBooEvent to Shared, make it a `record struct` and change the paths. Also pulled the Boo `OnActionPerform` listener method to Shared.
Could probably have moved other things too that use it but:
- SpookySpeaker has to be server still, uses Chat.
- Don't want to touch PoweredLight until #36541 is merged. Can keep this open and go back to it if necessary!
- XAE and revenant stuff needs the light flickering stuff pulled from ghost really

## Requirements

## Breaking changes

GhostBoo is now Shared.

### DIFF
--- a/Content.Shared/Ghost/SharedGhostSystem.cs
+++ b/Content.Shared/Ghost/SharedGhostSystem.cs
@@ -19,6 +19,8 @@ namespace Content.Shared.Ghost
         [Dependency] protected readonly EntityLookupSystem Lookup = default!;
         [Dependency] protected readonly IRobustRandom Random = default!;
 
+        private HashSet<EntityUid> _entitiesInRange = new();
+
         public override void Initialize()
         {
             base.Initialize();
@@ -47,12 +49,10 @@ namespace Content.Shared.Ghost
             if (args.Handled)
                 return;
 
-            var entities = Lookup.GetEntitiesInRange(args.Performer, component.BooRadius).ToList();
-            // Shuffle the possible targets so we don't favor any particular entities
-            Random.Shuffle(entities);
+            Lookup.GetEntitiesInRange(args.Performer, component.BooRadius, _entitiesInRange);
 
             var booCounter = 0;
-            foreach (var ent in entities)
+            foreach (var ent in _entitiesInRange)
             {
                 var handled = DoGhostBooEvent(ent);
 
@@ -64,7 +64,7 @@ namespace Content.Shared.Ghost
             }
 
             if (booCounter == 0)
-                Popup.PopupEntity(Loc.GetString("ghost-component-boo-action-failed"), uid, uid);
+                Popup.PopupPredicted(Loc.GetString("ghost-component-boo-action-failed"), uid, uid);
 
             args.Handled = true;
         }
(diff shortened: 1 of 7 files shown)


Content.Shared/Ghost/SharedGhostSystem.cs
```diff
@@ -37,6 +42,41 @@ private void OnAttempt(EntityUid uid, GhostComponent component, CancellableEntit
...
+                if (handled)
+                    booCounter++;
+
+                if (booCounter >= component.BooMaxTargets)
+                    break;
+            }
+
+            if (booCounter == 0)
+                Popup.PopupEntity(Loc.GetString("ghost-component-boo-action-failed"), uid, uid);
+
```
metalgearsloth, line 68: Random can't be predicted like this, also use popuppredicted not popupentity. Realistically you can just check for any ents in range on client.

Also please use the hashset version and pass it in rather than allocating it every time. Same with storing a list on the system.

slarticodefast: I think this system might be pretty hard to predict correctly, because the event in DoGhostBooEvent is only subscribed to on the server, so it is not marked as handled on the client, meaning it will mispredict. To fix this you would have to predict all subscriptions to `GhostBooEvent` first, which includes powered lights and chat code, making it pretty much impossible until those are refactored.

ljm862: Sounds good, I've pushed the change for the HashSet and removing randomness, not sure what you meant by checking on client here though, or the storing list.
I think if we want to make it predict properly from a client perspective we need to move all the other stuff over, but that requires Chat refactor and your powered lighting stuff so this can wait for that!

Content.Shared/Ghost/SharedGhostSystem.cs
```diff
@@ -37,6 +44,39 @@ private void OnAttempt(EntityUid uid, GhostComponent component, CancellableEntit
...
+        {
+            if (args.Handled)
+                return;
+
+            Lookup.GetEntitiesInRange(args.Performer, component.BooRadius, _entitiesInRange);
+
+            var booCounter = 0;
+            foreach (var ent in _entitiesInRange)
+            {
+                var handled = DoGhostBooEvent(ent);
```
slarticodefast, line 57: All of this will still have to stay on the server for now, or it will desync. `DoGhostBooEvent` will only return true on the server, since the subscriptions to the events are server side.
also you should not simply be removing the random shuffle, or it will prioritize entities by their ID instead of picking them randomly

slarticodefast (changes requested):

slarticodefast:
I'm closing this PR for now because the required refactors won't be resolved in the near future.
Once those are done either open a new PR or leave a comment so we can re-open this one.

PR #36635, opened 2025-04-16
Feature/auto sync node scanner

Fildrance:
## About the PR

node scanner now shows data on-line (like health analyzer)

## Why / Balance

QOL for new xeno arch

## Technical details

as all of stuff is networked, checking data is done on client side in update loop, and only attachment to artifact is networked separately in this PR

## Media

https://github.com/user-attachments/assets/78769d21-8a6f-43cc-938e-97e09248e744

## Requirements

## Breaking changes

none

**Changelog**

:cl: Fildrance
- tweak: node scanner now dislplays live artifact info (while in range) after linking to artifact


Content.Client/Xenoarchaeology/Ui/NodeScannerDisplay.xaml.cs
```diff
@@ -30,30 +43,83 @@ public void SetOwner(EntityUid scannerEntityUid)
...
+            return;
+
+        _nextUpdate = _timing.CurTime + _updateFromAttachedFrequency;
+
+        if (!_ent.TryGetComponent(_owner, out NodeScannerComponent? scannerComponent))
+            return;
+        
+        EntityUid? attachedArtifactEnt;
+        if (scannerComponent.AttachedTo == null || (attachedArtifactEnt = _ent.GetEntity(scannerComponent.AttachedTo)) == null)
+        {
```
metalgearsloth, line 65: Doing the variable inline is some cursed shit just split it out on god.

Content.Shared/Xenoarchaeology/Equipment/Components/NodeScannerComponent.cs
```diff
@@ -1,38 +1,47 @@
...
+[Access(typeof(NodeScannerSystem))]
 public sealed partial class NodeScannerComponent : Component
 {
     /// <summary>
-    /// Identity-names (3-digit codes) of nodes that are triggered on scanned artifact.
+    /// Xeno artifact entity, to which scanner is attached currently.
     /// </summary>
     [DataField, AutoNetworkedField]
-    public HashSet<string> TriggeredNodesSnapshot = new();
+    public NetEntity? AttachedTo;
```
metalgearsloth, line 19: Why isn't it an entityuid?

Content.Shared/Xenoarchaeology/Equipment/NodeScannerSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+                return;
+
+            var artifactCoordinates = Transform(attachedArtifact.Value).Coordinates;
+
+            if (!_transform.InRange(artifactCoordinates, transform.Coordinates, component.MaxLinkedRange))
+            {
+                //scanner is too far, disconnect
+                component.AttachedTo = null;
+                Dirty(uid, component);
+            }
```
metalgearsloth, line 50: 1. You slipped a return in there.
2. Just make an active comps so you're not iterating the dozens / hundreds of nodescanners when not necessary.

Content.Shared/Xenoarchaeology/Equipment/NodeScannerSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+    )
+    {
+        if (!_timing.IsFirstTimePredicted)
+            return;
+
+        if (TryComp(device, out UseDelayComponent? useDelay)
+            && !_useDelay.TryResetDelay((device, useDelay), true))
+            return;
+
+        var artifactNetEntity = EntityManager.GetNetEntity(unlockingEnt.Owner);
```
metalgearsloth, line 98: As above don't store netentity also it doesn't even serialize so this will break persistence.

metalgearsloth (changes requested):

Content.Shared/Xenoarchaeology/Equipment/Components/NodeScannerComponent.cs
Fildrance: coz its networked for client?

Fildrance: okay i'll fix it, god some misconceptions from xeno code

Content.Client/Xenoarchaeology/Ui/NodeScannerDisplay.xaml.cs
Fildrance: its not cursed, its neat!  But harder to read then usual shit. kinda happy about change for EntityUid so now its not needed at all.
:mistake:

Content.Shared/Xenoarchaeology/Equipment/NodeScannerSystem.cs
Fildrance: fair. fixed.

Content.Shared/Xenoarchaeology/Equipment/NodeScannerSystem.cs
Fildrance: NetEntity is totally not needed so hurray!

Fildrance:
@metalgearsloth could u look at how i separated fields between device comp and active-marker-comp? i'm not too comfortable with that stuff, even tho i understand the concept and positive performance implications, it seems like it is not really ECS-ish (as now we have 2 components, but they are very tightly coupled).

PR #36641, opened 2025-04-16
Fixed stinger grenade lag spikes

TGRCdev:
## About the PR

This PR modifies projectile grenades to no longer pass themselves as the "gun" that shot their contents. This prevents stinger grenades from causing a lag spike due to excessive error logging due to the bullets referencing the deleted grenade as their "gun".

## Why / Balance

Lag is bad. Also this issue would crash the debug server.

## Technical details

`SharedGunSystem.ShootProjectile` was modified so that its `gunUid` parameter is now nullable, and is `null` by default. Projectile grenades now pass `null` as the `gunUid`, instead of the grenade entity.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ProjectileGrenadeSystem.cs
+++ b/Content.Server/Explosion/EntitySystems/ProjectileGrenadeSystem.cs
@@ -77,7 +77,7 @@ public sealed class ProjectileGrenadeSystem : EntitySystem
             // slightly uneven, doesn't really change much, but it looks better
             var direction = angle.ToVec().Normalized();
             var velocity = _random.NextVector2(component.MinVelocity, component.MaxVelocity);
-            _gun.ShootProjectile(contentUid, direction, velocity, uid, null);
+            _gun.ShootProjectile(contentUid, direction, velocity, null, null);
         }
     }
 
--- a/Content.Server/Explosion/EntitySystems/ProjectileGrenadeSystem.cs
+++ b/Content.Server/Explosion/EntitySystems/ProjectileGrenadeSystem.cs
@@ -77,7 +77,7 @@ public sealed class ProjectileGrenadeSystem : EntitySystem
             // slightly uneven, doesn't really change much, but it looks better
             var direction = angle.ToVec().Normalized();
             var velocity = _random.NextVector2(component.MinVelocity, component.MaxVelocity);
-            _gun.ShootProjectile(contentUid, direction, velocity, null, null);
+            _gun.ShootProjectile(contentUid, direction, velocity);
         }
     }
 
--- a/Content.Server/Explosion/EntitySystems/ProjectileGrenadeSystem.cs
+++ b/Content.Server/Explosion/EntitySystems/ProjectileGrenadeSystem.cs
@@ -77,7 +77,7 @@ public sealed class ProjectileGrenadeSystem : EntitySystem
             // slightly uneven, doesn't really change much, but it looks better
             var direction = angle.ToVec().Normalized();
             var velocity = _random.NextVector2(component.MinVelocity, component.MaxVelocity);
-            _gun.ShootProjectile(contentUid, direction, velocity);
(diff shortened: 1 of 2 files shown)


Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
```diff
@@ -425,8 +425,11 @@ public void ShootProjectile(EntityUid uid, Vector2 direction, Vector2 gunVelocit
         Physics.SetLinearVelocity(uid, finalLinear, body: physics);
 
         var projectile = EnsureComp<ProjectileComponent>(uid);
-        Projectiles.SetShooter(uid, projectile, user ?? gunUid);
-        projectile.Weapon = gunUid;
+        if (!TerminatingOrDeleted(gunUid))
```
beck-thompson, line 428: So I did some testing, I don't think this `TerminatingOrDeleted` is necessary. From what I'm seeing, the server still errors out if you pass in the grenades UID (something that should definitely be deleting or something). I'd always assign this one (its nullable anyway so it doesn't matter) and for the bottom one just do a `shooter != null` check.

beck-thompson (changes requested):

Content.Server/Explosion/EntitySystems/ProjectileGrenadeSystem.cs
```diff
@@ -77,7 +77,7 @@ private void FragmentIntoProjectiles(EntityUid uid, ProjectileGrenadeComponent c
             // slightly uneven, doesn't really change much, but it looks better
             var direction = angle.ToVec().Normalized();
             var velocity = _random.NextVector2(component.MinVelocity, component.MaxVelocity);
-            _gun.ShootProjectile(contentUid, direction, velocity, uid, null);
+            _gun.ShootProjectile(contentUid, direction, velocity, null, null);
```
beck-thompson, line 80: Just remove the nulls

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
```diff
@@ -425,8 +425,10 @@ public void ShootProjectile(EntityUid uid, Vector2 direction, Vector2 gunVelocit
         Physics.SetLinearVelocity(uid, finalLinear, body: physics);
 
         var projectile = EnsureComp<ProjectileComponent>(uid);
-        Projectiles.SetShooter(uid, projectile, user ?? gunUid);
         projectile.Weapon = gunUid;
+        var shooter = user ?? gunUid;
+        if (shooter != null)
+            Projectiles.SetShooter(uid, projectile, shooter!.Value);
```
beck-thompson, line 431: You don't need the `!` anymore because your checking for null!

beck-thompson (changes requested):
looks good just a few things

beck-thompson (approved):
looks good! Thanks for fixing this it was annoying me

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
```diff
@@ -414,7 +414,7 @@ public abstract void Shoot(
         EntityUid? user = null,
         bool throwItems = false);
 
-    public void ShootProjectile(EntityUid uid, Vector2 direction, Vector2 gunVelocity, EntityUid gunUid, EntityUid? user = null, float speed = 20f)
+    public void ShootProjectile(EntityUid uid, Vector2 direction, Vector2 gunVelocity, EntityUid? gunUid = null, EntityUid? user = null, float speed = 20f)
     {
         var physics = EnsureComp<PhysicsComponent>(uid);
```
metalgearsloth, line 419: Is null gunuid even used anywhere.

TGRCdev: Should I remove the default argument and reintroduce the `null` in ProjectileGrenadeSystem?

MilonPL (approved):

PR #36645, opened 2025-04-17
Revert polymorphed entities before gibbing if revert on death.

SolStar2:
## About the PR
Revert polymorphed entities before gibbing if revert on death.

## Why / Balance
Gibbing should count as death and revert. Prevents brain and items like nuke disk from being lost in null space on the parent entity. Also allows polymorphs without damage transfer to be able to revert once gibbed.

## Technical details
made some small changes to the gib trigger moving acidify logic to gib system

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl: SolStar
- fix: Fixed an edge case where polymorphs would not revert if gibbed instantly.

### DIFF
--- a/Content.Server/Body/Systems/BodySystem.cs
+++ b/Content.Server/Body/Systems/BodySystem.cs
@@ -12,6 +12,8 @@ using Content.Shared.Movement.Systems;
 using Robust.Shared.Audio;
 using Robust.Shared.Timing;
 using System.Numerics;
+using Content.Server.Polymorph.Components;
+using Content.Server.Polymorph.Systems;
 
 namespace Content.Server.Body.Systems;
 
@@ -22,6 +24,7 @@ public sealed class BodySystem : SharedBodySystem
     [Dependency] private readonly HumanoidAppearanceSystem _humanoidSystem = default!;
     [Dependency] private readonly MobStateSystem _mobState = default!;
     [Dependency] private readonly SharedMindSystem _mindSystem = default!;
+    [Dependency] private readonly PolymorphSystem _polymorph = default!;
 
     public override void Initialize()
     {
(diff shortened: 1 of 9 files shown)


Content.Server/Body/Systems/BodySystem.cs
```diff
@@ -12,6 +12,8 @@
 using Robust.Shared.Audio;
 using Robust.Shared.Timing;
 using System.Numerics;
+using Content.Server.Polymorph.Components;
+using Content.Server.Polymorph.Systems;
```
slarticodefast, line 16: sort alphabetically

Content.Server/Body/Systems/BodySystem.cs
```diff
@@ -110,11 +113,23 @@ public override HashSet<EntityUid> GibBody(
             return new HashSet<EntityUid>();
         }
 
+        // If a polymorph configured to revert on death is gibbed without dying,
+        // revert it then gib so the parent is gibbed instead of the polymorph.
+        if (TryComp<PolymorphedEntityComponent>(bodyId, out var polymorph)
```
slarticodefast, line 118: Polymorphing should not be hardcoded into the BodySystem like this.
This should raise an `BeforeBeingGibbedEvent` and polypmorphsystem should subscribe to it, revert the polymorphing and change the gibbing target so that the code below can run normally.

slarticodefast (changes requested):
gibOrgans and acidifying are doing two different things, so you can't just make them  equivalent
one is deleting organs, the other is deleting equipment

AJCM-git:
@SolStar2  Still interested in getting this in? I can handle this PR if you fix conflicts

Issue #36634, opened 2025-04-16
Mail Delivery Refactor


ljm862:
The Delivery System is currently relatively barebones. It has huge potential to unlock really intricate storytelling in terms of the deliverers, recipients and even potential antagonistic intervention or even spawning of random events. To aid in the unlocking of this potential, a refactor is required to enable all the little mechanics that content developers would like to add.
Here are the key tasks we need to complete to consider the refactor finished:

 - **`DeliveryComponent` needs the following fields to enable the deliveries to track their potential earnings and losses.** - #36636
 	- `BaseSpesoReward`: Standard ammount for a successful delivery
 	- `BaseSpesoPenalty`: Standard penalty ammount (half of the reward) 
 	- `BaseSpesoMultiplier`: Aggregates other multipliers from various components and adds up at the end
 	- `PenaltyBankAccount`: Which account should be penalised?
 	- `WasPenalized`: Has the penalty already been applied?
 These fields should of course be configurable and flexible enough that external components can interact with them and not cause issue

 - **Dynamic reward and penalty messaging** included in `GrantSpesoReward()` so that when there is an issue with the delivery, the player is informed that this issue has been taken into account. - #36638

 	Will need to create some messages such as `"delivery-penalty-damaged"` - `"{reason}! INVOKING A PENALTY OF {spesos} ON {account}!"` which sends a message (speaks) upon opening and informs the recipient, cargo and anyone else about the action

 - **Calculate the reward dynamically upon opening the mail.** - #36639
Currently we use the `SpesoReward` in the yml file, we should use the new fields, `BaseSpesoReward`, etc. to dynamically asses how many spesos the department should receive. 
    - The idea is to raise an event called `GetTotalDeliveryMultiplier` which invokes the modifying components to add their multiplier onto the total multiplier and grants the reward:
 	$TotalReward = BaseSpesoReward * TotalMultiplier$
 	where $TotalMultiplier = BaseSpesoMultiplier + \sum_{i=1}^{n} \text{externalMultiplier}_i $
 	Or if it fails due to timeout, destroyed, etc:
 	$TotalPenalty = BaseSpesoPenalty * TotalMultiplier$.

     - If the `WasPenalized` flag has already been set though, we don't need to apply the penalty again and can skip any calculation relating to it.

-----
The above 3 tasks will enable penalties and rewards to be added onto the delivery items, things such as a penalty for being cut which forces it open, or a reward for delivering it very quickly. 

[...]

PR #36660, opened 2025-04-17
Delivery system now has the possibility of penalisation

ljm862:
## About the PR

For #36634. Closes all linked issues inside of it.

## Why / Balance

We now open up the possibility of penalization on deliveries for damaging them, being within a certain time, or something else entirely!

## Technical details

Adds a few fields and simplifies some code in the `DeliverySystem` too also updates the YAML file for deliveries which needs the new names and adds a Speech component which allows for announcing to the recipient upon opening.

Creates a new event which gets raised whenever a successful opening of a delivery takes place which does the multiplier calculation for penalties and rewards.

Small fix in `SharedCargoSystem` where trying to open a delivery in the dev environment would crash the server due to using an `Add` on the dictionary and `Cargo` would already be present. This was happening on a clean master so I don't believe the issue was due to anything here and this fix looks to not affect the functionality on my tests.

## Requirements

## Breaking changes

`SpesoReward` field in `DeliveryComponent` is now `BaseSpesoReward`.


ScarKy0 (changes requested):
Set the value for baseSpesoPenalty in the yaml files, since 250 isnt half for letters or packages. So you gotta make it 200 and 400 respectively.
Also add description to the PR

ljm862:
> Set the value for baseSpesoPenalty in the yaml files, since 250 isnt half for letters or packages. So you gotta make it 200 and 400 respectively. Also add description to the PR

Yes, sorry I pressed Enter when trying to rename the PR which submitted it 😅

ScarKy0 (approved):
Looks good.

ScarKy0:
Marking as do not merge for now. Talked with other maints and we came to a conclusion the logic should be in this PR as well so we don't accidentally end up with unused datafields.
Ping me on discord when you add the other stuff and I'll remove it

ScarKy0:
Removed DNM since you added the logic. Resolve the merge conflict and I'll drop a review.

Content.Shared/Delivery/DeliveryComponent.cs
```diff
@@ -48,6 +62,26 @@ public sealed partial class DeliveryComponent : Component
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool WasPenalized;
+
+    /// <summary>
+    /// If this delivery has failed its conditions and should incur a penalty on the
+    /// respective account.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool ShouldBePenalized;
```
ScarKy0, line 83: No need for this. Penalties will be handled by their respective components when that is done.
For example, FragileDeliveryComponent will pick up that the delivery got broken, it will simply run the function for invoking a penalty.

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -70,13 +76,59 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+    /// <param name="stationAccountEnt"><see cref="StationBankAccountComponent"/> entity.</param>
+    private void HandlePenalization(Entity<DeliveryComponent> ent, Entity<StationBankAccountComponent?> stationAccountEnt, float multiplier)
+    {
+        var delivery = ent.Comp;
+        var accountName = UnknownAccount;
+        if (_protoMan.TryIndex(delivery.PenaltyBankAccount, out var accountInfo))
+            accountName = Loc.GetString(accountInfo.Name);
+
+        // Extract reason from components...
+        _chat.TrySendInGameICMessage(ent, GetMessage("wah", delivery.BaseSpesoPenalty, accountName), InGameICChatType.Speak, hideChat: true);
```
ScarKy0, line 113: rename to HandlePenalty. It should also pass a string for the reason (null by default) which gets inserted into the GetMessage. I'm not really sure where you should try to get the locale string for the reason, either here or in GetMessage

ScarKy0 (changes requested):

ScarKy0:
resolved the conflicts for you

Content.Server/Delivery/DeliverySystem.cs
ljm862: Ah got it! I misunderstood and thought it was only applying on opening. I see now! 
The method is now public so should be callable whenever these things happen and I've made the GetMultiplier stuff happen on penalty calling too

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -25,6 +29,18 @@ public sealed partial class DeliverySystem : SharedDeliverySystem
...
+
+    /// <summary>
+    /// Name to use if the <see cref="DeliveryComponent.PenaltyBankAccount"/> is not set
+    /// </summary>
+    private const string UnknownAccount = "Unknown Account";
+
+    /// <summary>
+    /// Default reason to use if the penalization is triggered
+    /// </summary>
+    private const string DefaultMessage = "Warning";
```
ScarKy0, line 42: Those should be localized.
Save them here as a LocId and just call Loc.GetString() where neccessary

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -70,13 +81,66 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+        var multiplier = GetDeliveryMultiplier(ent!); // Resolve so we know it's got the component
+
+        _cargo.UpdateBankAccount(
+            stationAccountEnt,
+            (int)(delivery.BaseSpesoReward * multiplier),
+           _cargo.CreateAccountDistribution((delivery.RecipientStation.Value, account)));
+    }
+
+    /// <summary>
+    /// Runs the penalty logic: Announcing the penalty and calculating how much to charge the desginatied account
```
ScarKy0, line 100: designated*

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -70,13 +81,66 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
         if (!Resolve(ent, ref ent.Comp))
             return;
 
-        if (!TryComp<StationBankAccountComponent>(ent.Comp.RecipientStation, out var account))
+        var delivery = ent.Comp;
```
ScarKy0, line 84: no need for this, just use ent.Comp where neccessary

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -70,13 +81,66 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+
+    /// <summary>
+    /// Runs the penalty logic: Announcing the penalty and calculating how much to charge the desginatied account
+    /// </summary>
+    /// <param name="ent"><see cref="DeliveryComponent"/> entity.</param>
+    /// <param name="stationAccountEnt"><see cref="StationBankAccountComponent"/> entity.</param>
+    public void HandlePenalty(Entity<DeliveryComponent> ent, Entity<StationBankAccountComponent?> stationAccountEnt, string reasonLoc)
+    {
+        var multiplier = GetDeliveryMultiplier(ent);
+        var delivery = ent.Comp;
```
ScarKy0, line 107: same here, just use ent.Comp
its more understandable

ScarKy0 (changes requested):

ScarKy0 (approved):
Looks fine.
Marking as DNM until I can test this with some in-game examples. Thank you a lot for your help!

Content.Shared/Delivery/DeliveryComponent.cs
```diff
@@ -23,10 +25,22 @@ public sealed partial class DeliveryComponent : Component
...
+    /// The base amount of spesos that will be removed from the station bank account on a penalized delivery
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public int BaseSpesoPenalty = 250;
+
+    /// <summary>
+    /// The base multiplier to add onto the reward or penalty upon delivery.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float BaseSpesoMultiplier = 1.0f;
```
slarticodefast, line 43: I already mentioned this in the other PR, but I don't see the reason for this datafield
the default should always be one
if you want something else you can edit the other two base values

Content.Shared/Delivery/SharedDeliverySystem.cs
```diff
@@ -210,9 +210,17 @@ protected void UpdateDeliverySpawnerVisuals(EntityUid uid, int contents)
...
+    protected virtual void HandlePenalty(Entity<DeliveryComponent> ent, string? reasonLoc = null) { }
+
     protected virtual void SpawnDeliveries(Entity<DeliverySpawnerComponent?> ent) { }
 }
 
+/// <summary>
+/// Used to gather the multiplier from all different delivery components.
+/// </summary>
+[ByRefEvent]
+public record struct GetDeliveryMultiplierEvent(float Multiplier = 0.0f);
```
slarticodefast, line 222:
```suggestion
public record struct GetDeliveryMultiplierEvent(float Multiplier = 1.0f);
```
should default to one, that way you can either modify the value multiplicatively or additively in subscriptions

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -73,10 +84,72 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+        var ev = new GetDeliveryMultiplierEvent();
+        RaiseLocalEvent(ent, ref ev);
+
+        return ev.Multiplier += ent.Comp.BaseSpesoMultiplier;
+    }
+
+    private string GetMessage(string reason, int penalty, string accountName)
+    {
+        return Loc.GetString("delivery-penalty-message", ("reason", reason), ("spesos", penalty), ("account", accountName.ToUpper()));
     }
```
slarticodefast, line 153: why make this a function?

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -73,10 +84,72 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+
+        ent.Comp.WasPenalized = true;
+    }
+
+    private float GetDeliveryMultiplier(Entity<DeliveryComponent> ent)
+    {
+        var ev = new GetDeliveryMultiplierEvent();
+        RaiseLocalEvent(ent, ref ev);
+
+        return ev.Multiplier += ent.Comp.BaseSpesoMultiplier;
```
slarticodefast, line 147: remove this, this is not how multipliers should be done
see the other comments
```suggestion
```

slarticodefast (changes requested):

EmoGarbage404 (dismissed):
Conceptually, I don't know if penalties really work with department economy and if directly subtracting funds is what we should be doing.

I feel like a system that allows mail to penalize a specific cargo account could be trivially weaponized to suck funds out of cargo with basically 0 counterplay. Even just having a tech make a mistake and actively cost the account their limited budget seems like something that could open itself to toxicity.

ScarKy0:
> Conceptually, I don't know if penalties really work with department economy and if directly subtracting funds is what we should be doing.
I feel like a system that allows mail to penalize a specific cargo account could be trivially weaponized to suck funds out of cargo with basically 0 counterplay. Even just having a tech make a mistake and actively cost the account their limited budget seems like something that could open itself to toxicity.

This has already been discussed and approved via a design document. We need some sort of penalty so cargo doesn't just get free money from things like priority mail, as that would only result in a positive with no risk whatsoever, as well as just have them cut up their mail for free items. Which is something cargo will almost definitely do, seeing how their entire thing is gambling most of the time.
The only other way penalties could work is if they apply a flat negative modifier to the delivery when it is opened, which is okay. However cutting open mail should ALWAYS invoke an instant penalty like this.

EmoGarbage404:
> This has already been discussed and approved via a design document. 

From my understanding the design doc preceds the recent department economy changes.

> We do NOT want them to simply get free things in those cases.

I feel like there are still better ways of handling it. A kinda sneaky one would be to have a separate pool of "junk" mail and just always make improperly opened mail give junk. You could use the same for priority wtih "expired" items and fragile with broken ones.

ScarKy0:
The only situation cargo is really meant to be at a nat negative is when mail is stolen. If you lose too much money simply turn off the teleporter or protect it better.
In all other situations (when it's still possible to deliver something) taking a cut from the earnings is more reasonable than flat money loss, so I'll agree with that.

EmoGarbage404:
> Yes but this has no explaination of happening to the players and just feels forced.

If its happening internally, how does it feel forced? Its not like cargo knows what theyre delivering beforehand. If anything, ripping a package only to not get money *and* get trash seems like it would be better for keeping cargo from rifling through things. As opposed to basically doing small amount gambling for mail items.

ScarKy0:
> If its happening internally, how does it feel forced?

Cause it is just a very unsatisfying solution in my eyes. And this makes stealing mail absolutely pointless. The thing tearing mail open is meant to prevent is cargo leaving it on sushi belts or just belting it outside of the department for people to take for themselves (yes, it happens often). In this case, a tider could steal a package and get possibly contraband, or pills out of a letter. (I still plan to improve the entity pools, they are just not a major concern right now)
Now, if you simply get trash, the main problem is not solved. Why would anyone try to take mail knowing its only trash inside? This is meant to punish cargo for mishandling their stuff, not only prevent them from opening it by themselves.

slarticodefast:
I agree with scarky here. Currently mail is just left lying around on the cargo counter making it self-serve, with no downsides to cargo if it is stolen. Also if we add a way to tear open mail it will mean cargo will do it themselves just to get free stuff instead of delivering it (they love gambling!). A penalty should be big enough to prevent this via negative feedback, but not be abusable to bankrupt cargo. We are talking maybe 200 spesos or so, which I think is totally fine.

EmoGarbage404:
I suppose this is probably fine if it only applies to opening another person's mail. I do want to clarify that i think this would be a poor choice for something like priority or fragile mail.

walksanatora:
> If you lose too much money simply turn off the teleporter or protect it better.

it is now stolen. cant turn it of. no way to locate it. just constant dings to cargo's money with no way to stop it.

northerndakotas:
> > If you lose too much money simply turn off the teleporter or protect it better.
> 
> it is now stolen. cant turn it of. no way to locate it. just constant dings to cargo's money with no way to stop it.

If someone wants to somehow covertly steal a whole ass mailbox in order to sit in maints and open people's mail every few minutes taking a couple hundred spesos each time, I think they deserve to cause a little trouble for all that effort. It's far less destructive than damaging power or bombing the chem locker.

Regarding penalties at all, I think it's a good idea as long as a delivered parcel never subtracts money from the account, lest it become advantageous for Cargo to intentionally never deliver certain mail to avoid the penalty.

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -73,10 +84,68 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+        {
+            { ent.Comp.PenaltyBankAccount, 1.0 }
+        };
+
+        var penaltyAccountBalance = stationAccount.Accounts[ent.Comp.PenaltyBankAccount];
+        var calculatedPenalty = (int)(ent.Comp.BaseSpesoPenalty * multiplier);
+
+        // Prevents cargo from going into negatives
+        if (penaltyAccountBalance - calculatedPenalty < 0)
+            calculatedPenalty = Math.Max(0, penaltyAccountBalance);
```
slarticodefast, line 130:
```suggestion
        calculatedPenalty = Math.Max(0, penaltyAccountBalance);
```
that is already doing the comparison

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -73,10 +84,68 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+
+        _cargo.UpdateBankAccount(
+            (ent.Comp.RecipientStation.Value, stationAccount),
+            -calculatedPenalty,
+            dist);
+
+        var message = Loc.GetString("delivery-penalty-message", ("reason", reason), ("spesos", calculatedPenalty), ("account", accountName.ToUpper()));
+        _chat.TrySendInGameICMessage(ent, message, InGameICChatType.Speak, hideChat: true);
+
+        ent.Comp.WasPenalized = true;
```
slarticodefast, line 140: dirty

slarticodefast (changes requested):

Content.Server/Delivery/DeliverySystem.cs
slarticodefast: nevermind, this is different

Content.Server/Delivery/DeliverySystem.cs
```diff
@@ -73,10 +79,74 @@ protected override void GrantSpesoReward(Entity<DeliveryComponent?> ent)
...
+        var dist = new Dictionary<ProtoId<CargoAccountPrototype>, double>()
+        {
+            { ent.Comp.PenaltyBankAccount, 1.0 }
+        };
+
+        var penaltyAccountBalance = stationAccount.Accounts[ent.Comp.PenaltyBankAccount];
+        var calculatedPenalty = (int)(ent.Comp.BaseSpesoPenalty * multiplier);
+
+        // Prevents cargo from going into negatives
+        if (penaltyAccountBalance - calculatedPenalty < 0)
```
slarticodefast, line 123:
```suggestion
        if (calculatedPenalty > penaltyAccountBalance )
```
same logic, but a little more readable and one calculation less

slarticodefast (approved):
Looks good to me!

PR #36661, opened 2025-04-17
Make escape pods fail the "Maroon Target" objective

SlamBamActionman:
## About the PR

This PR makes being on the escape pod *as a target* fail the maroon target objective for antags, in which you must kill or maroon your target.

Being on CentComm also counts as surviving - this doesn't change the roundend text, but keeps the objectives consistent post roundend.

Note! The escape pod does *not* count for the "Survive to CentComm" objective in this PR.

## Why / Balance

This should (hopefully) push players a bit more towards doing objectives on-station, as waiting for the evac shuttle and blindly throwing bombs into it is no longer going to be as reliable since the target may be on a pod instead. 

This PR doesn't change the survive objective for antags because that has a larger impact on how antags and Sec have to play during the end of the round and that requires extra design work, i.e. out of scope for this PR.

## Technical details

Added a query for all entities with the `EscapePodComponent`, simple as that.

Previously the EscapePodComponent was removed upon launch (to prevent pods launching more than once somehow?), so now instead the component tracks if it has been launched or not.

## Steps to test

1. Spawn 3 clients on a station map with at least one escape pod
2. Make one traitor and give it `KillPersonObjective` until it has an objective for both other clients
3. Make one client go to an escape pod
4. Call evac with `dockemergencyshuttle` and launch it with `launchemergencyshuttle`
5. The objective for the client on the pod should count as failed, while the client remaining on the station should count as successful.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The "Kill/Maroon" objective now fails if the target is on an escape pod. 
- fix: Being at CentComm now counts as surviving post-roundend.

### DIFF
--- a/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
@@ -51,7 +51,7 @@ public sealed class KillPersonConditionSystem : EntitySystem
             return 0f;
 
         // target is escaping so you fail
-        if (_emergencyShuttle.IsTargetEscaping(mind.OwnedEntity.Value))
+        if (_emergencyShuttle.IsTargetEscaping(mind.OwnedEntity.Value, false))
             return 0f;
 
         // evac has left without the target, greentext since the target is afk in space with a full oxygen tank and coordinates off.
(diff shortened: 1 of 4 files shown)


ArtisticRoomba:
Been wanting to do this for a while.

MilonPL (approved):
Looks good.

EmoGarbage404:
This has been superseded by the recent kill objective rework.

SlamBamActionman:
I need to doublecheck the parameters but has it? 
I was under the impression that the maroon/kill objective rework only changed it so you can be either dead or alive on the evac shuttle.
I.e. taking the pod still counts as marooning.

PR #36685, opened 2025-04-18
Nuke DoesCustomBoundsIntersectWithFixture

### LINKED ISSUE #36665: DoesCustomBoundsIntersectWithFixture needs killing
This method is pointless and shouldn't exist.

VerinSenpai:
## About the PR
Takes the request quite literally.

Resolves #36665

## Technical details
I took DoesCustomBoundsIntersectWithFixture around back and beat it to death with a shovel.

## Requirements

## Breaking changes

I don't believe so. The only use for this was just above in a collision check that seemed to do nothing most tile items won't collide anyway and items like wall lights will happily collide regardless.

### DIFF
--- a/Content.Shared/RCD/Systems/RCDSystem.cs
+++ b/Content.Shared/RCD/Systems/RCDSystem.cs
@@ -446,11 +446,6 @@ public sealed class RCDSystem : EntitySystem
                     if (!fixture.Hard || fixture.CollisionLayer <= 0 || (fixture.CollisionLayer & (int) prototype.CollisionMask) == 0)
                         continue;
 
-                    // Continue if our custom collision bounds are not intersected
-                    if (prototype.CollisionPolygon != null &&
-                        !DoesCustomBoundsIntersectWithFixture(prototype.CollisionPolygon, component.ConstructionTransform, ent, fixture))
-                        continue;
-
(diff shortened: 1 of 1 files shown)


metalgearsloth:
If this breaks anything the ideal is to actually do the bounds check in the lookup itself rather than re-checking it.

Content.Shared/RCD/Systems/RCDSystem.cs
```diff
@@ -446,11 +446,6 @@ private bool IsConstructionLocationValid(EntityUid uid, RCDComponent component,
                     if (!fixture.Hard || fixture.CollisionLayer <= 0 || (fixture.CollisionLayer & (int) prototype.CollisionMask) == 0)
                         continue;
 
-                    // Continue if our custom collision bounds are not intersected
-                    if (prototype.CollisionPolygon != null &&
-                        !DoesCustomBoundsIntersectWithFixture(prototype.CollisionPolygon, component.ConstructionTransform, ent, fixture))
-                        continue;
-
```
metalgearsloth, line 453: Might be better off just iterating fixtures directly and querying, or using the engine method (I forgor which one it is). Worst case add an entitylookup that checks each fixture individually?

PR #36689, opened 2025-04-18
Replaced black gloves in security lockers with combat gloves.

ToastEnjoyer:
## About the PR
Changed the black gloves in the security officer lockers to the insulate combat gloves.

## Why / Balance
First off, combat gloves look better then just black gloves, second it makes sense that security would have combat gloves, and combat gloves are insulated which means they wont get shocked on grilles.

## Technical details
Changed ClothingHandsGlovesColorBlack to ClothingHandsGlovesCombat in security.yml

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/security.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/security.yml
@@ -69,7 +69,7 @@
         prob: 0.5
       - id: ClothingEyesGlassesSecurity
       - id: ClothingHeadsetSecurity
-      - id: ClothingHandsGlovesColorBlack
+      - id: ClothingHandsGlovesCombat
       - id: ClothingShoesBootsJack
       - id: WeaponMeleeNeedle
         prob: 0.1


LaCumbiaDelCoronavirus:
I actually dont see anything wrong with this, some people might disagree

slarticodefast:
I don't think security needs insulated gloves, they should not be hacking doors.
This also has huge balance implications on for antags, as you can no longer shock security.

ToastEnjoyer:
Partly I made this PR because of grille traps, another part is that i've seen all of security get fucked by a single HV grille.

keronshb:
They're not meant to have insuls.

Everturning:
> Partly I made this PR because of grille traps, another part is that i've seen all of security get fucked by a single HV grille.

then they just need to avoid it idk what to tell you. there are ways to test if a grille is shocked that are not running into it

YoungThugSS14:
You need a Changelog btw

RedBookcase:
Don't think this should go through for the same reason suggestions of giving Security magboots haven't been accepted. If Security needs those extra tools/utilities, they can go ask for them from other departments. If Security is all getting killed by a single Grille trap... that is a problem with Security players, not their gear.

slarticodefast:
I brought it up for maintainer discussion and giving all of security full restance to shocks removes a lot of potential for interesting trap mechanics and makes gameplay one-sided, in addition to allowing them to hack doors.
If shocks are a problem that balance should be looked at. But if I remember correctly we did nerf shock times from electrocutions across the board in the last release, so I would prefer if we see how that goes first and adjust it more at a later point if necessary.

PR #36579, opened 2025-04-15
Remove egg-plant from MegaSeed Servitor

K-Dynamic:
## About the PR

- Removed egg-plant from MegaSeed Servitor
  - Egg-plant can still be obtained by mutating eggplant or exotic seed crate

## Why / Balance

Egg-plant seems out of place in the MegaSeed Servitor, especially as an exotic and mutant plant.

Follows tg station and goonstation convention where egg-plant seeds are only obtainable by mutation or exotic seed.

## Technical details

N/A

## Media

N/A

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

N/A

**Changelog**

:cl:
- remove: Removed egg-plant from MegaSeed Servitor. (Egg-plant seeds can still be obtained by mutation or exotic seed crate.)

PR #36693, opened 2025-04-18
Increase base amount of egg in egg-plant

iaada:
## About the PR

This PR increases the amount of egg in a default 20 potency egg-plant seed from 3 to 6.

## Why / Balance

Egg-plant eggs not having the same reagents as regular eggs has always been an annoyance. Any cooking recipe asking for eggs will want 6 units, twice as much as the egg-plant produce without a clear indication the vegan egg is smaller than a normal egg. With #36579 removing the seed from round start this has made it proportionally more annoying to wait for an exotic crate just to have small eggs, or to mutate into a small egg filled with other mutation reagents.

## Requirements

**Changelog**

:cl: aada
- tweak: Egg-plants now have 6 reagent per egg at 20 potency, up from 3.

### DIFF
--- a/Resources/Prototypes/Hydroponics/seeds.yml
+++ b/Resources/Prototypes/Hydroponics/seeds.yml
@@ -932,8 +932,8 @@
   idealHeat: 298
   chemicals:
     Egg:
-      Min: 1
-      Max: 10
+      Min: 4
+      Max: 12
       PotencyDivisor: 10
 
 - type: seed


slarticodefast (approved):
Yep, that always annoyed me as well.
And I think it is a good way to counter-balance not having them round-start anymore.

ArtisticRoomba (approved):

gitjubx:
normaly you just add 10u robust and make it 50 potency quickly to get 6u egg it wasnt really a problem its like a very small bonus compared what has been taken away

PR #36696, opened 2025-04-18
Prisoner Eva Suit is now an Emergency Eva Suit

YoungThugSS14:
## About the PR

I made the prisoner eva suit an emergency eva suit.
## Why / Balance

Because why do most of the crew get emergency eva suits, but a prisoner gets a higher quality one. (ALSO THEY'RE BOTH ORANGE SO ITS LESS CONFUSING)
## Technical details

Simple YML changes
## Media

N/A But I tested it and it works.
## Requirements

## Breaking changes
N/A

**Changelog**
🆑 
- tweak: The Prisoner Eva Suit now has the same stats as an emergency eva suit.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/softsuits.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/softsuits.yml
@@ -59,7 +59,7 @@
 
 #Prisoner EVA
 - type: entity
-  parent: ClothingOuterEVASuitBase
+  parent: ClothingOuterSuitEmergency
   id:  ClothingOuterHardsuitEVAPrisoner
   name: prisoner EVA suit
   description: A lightweight space suit for prisoners to protect them from the vacuum of space during emergencies.
--- a/Resources/Prototypes/Catalog/Fills/Lockers/suit_storage.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/suit_storage.yml
@@ -47,7 +47,6 @@
     contents:
         - id: OxygenTankFilled
         - id: ClothingOuterHardsuitEVAPrisoner
-        - id: ClothingHeadHelmetEVALarge
         - id: ClothingMaskBreath
 
 #Syndicate EVA


slarticodefast:
Pretty sure the emergency EVA suit has a toggleable included helmet, whereas the one in the EVA suit has a one you have to equip manually. So reparenting will change that as well.

YoungThugSS14:
> Pretty sure the emergency EVA suit has a toggleable included helmet, whereas the one in the EVA suit has a one you have to equip manually. So reparenting will change that as well.

Yeah I removed the helmet you equip manually from the locker fill

slarticodefast:
Ok, then that should be mentioned in the PR description. Also the included helmet has to be adjusted to the right prototype.
And the helmet prototype has to be added to the `HideSpawnMenu` category so it cannot be spawned on its own anymore. You will also have to check if it is used anywhere else, for example in maps.

But personally I like the prisoner EVA suit as a two part suit, it makes it feel less high-tech.

YoungThugSS14:
> Ok, then that should be mentioned in the PR description. Also the included helmet has to be adjusted to the right prototype. And the helmet prototype has to be added to the `HideSpawnMenu` category so it cannot be spawned on its own anymore. You will also have to check if it is used anywhere else, for example in maps.
> 
> But personally I like the prisoner EVA suit as a two part suit, it makes it feel less high-tech.

alright, I’m probably going to revert the helmet changes but how would you about the stat changes on their own.

YoungThugSS14:
@slarticodefast is the test fail related?

slarticodefast:
Nope, that test is currently broken for everyone.

Resources/Prototypes/Entities/Clothing/OuterClothing/softsuits.yml
```diff
@@ -62,12 +62,19 @@
...
     sprite: Clothing/OuterClothing/Suits/eva_prisoner.rsi
   - type: Clothing
     sprite: Clothing/OuterClothing/Suits/eva_prisoner.rsi
+  - type: ClothingSpeedModifier
+    walkModifier: 0.7
+    sprintModifier: 0.7
+  - type: HeldSpeedModifier
+  - type: TemperatureProtection
+    heatingCoefficient: 0.75
+    coolingCoefficient: 0.5
```
ArtisticRoomba, line 77: might be better to reparent to the emergency EVA suit instead of duping all of this if you're going the emergency EVA suit route

ArtisticRoomba (changes requested):
I partially disagree with this change: I feel that this could increase Security's workload in terms of keeping a prisoner alive during transport in dire situations, which is already hard to do and annoying.

You're supposed to try and keep them alive under space law, it would be counterproductive to make it harder for security.

If the problem is the prisoner EVA being indistinguishable from the emergency EVA suit, then it should see a resprite, not a downgrade.

Resources/Prototypes/Entities/Clothing/OuterClothing/softsuits.yml
YoungThugSS14: I did that the first time but it brought over the built in helmet which Slarti disagreed with

ArtisticRoomba: I see, this is fine then.

YoungThugSS14:
> Note that I said _could_: of course it would be rare for a prisoner to take a spacewalk or something during transit, but security shouldn't be out of luck. But this change also limits antags ability to escape from perma if they got their hands on a prisoner EVA suit and wiggled away free.

I feel the nerf isnt TOO major, it discourages prisoners hiding out in space for like… the majority of the shift after their escape, and it makes no sense the majority of the crew gets weaker suits than prisoners

ArtisticRoomba:
You make a good point, though the reason why the Emergency EVA suits are like that is because we don't want to give everyone free access to space unharmed. I guess this change isn't as bad, though I'll wait for some more maints (maybe Slam) to chime in.

YoungThugSS14:
> You make a good point, though the reason why the Emergency EVA suits are like that is because we don't want to give everyone free access to space unharmed. I guess this change isn't as bad, though I'll wait for some more maints (maybe Slam) to chime in.

Thanks for being understanding

PR #36701, opened 2025-04-19
Buyable + destructible desk bell, desk bell variants

iaada:
## About the PR

PR adds a desk bell to the bureaucracy crate, and allows desk bells to be destroyed. It also adds two variants of the desk bell: the desk buzzer and the desk horn.

## Why / Balance

The desk bell is a very common item, however there's currently no way to get a new one if it disappears or wasn't mapped, nor is there a way to destroy one. To quote its description it's, "the cornerstone of customer service." Not having one brings shame to a receptionist.

The variants were created to allow some personal character to your front desk.

## Technical details

- `EmptyDeskBell` was created and `DeskBell` reparented to it. 2 new entities also parented to `EmptyDeskBell`.
- `BellGraph` construction graph for creating the bell variants.
- `DeskBuzzer` sound collection. Uses an existing sound that previously only triggered when throwing a phone.

## Media

https://github.com/user-attachments/assets/644d42a0-7e28-4218-8e34-11f6b2526377

## Requirements

**Changelog**

:cl: aada
- add: A desk bell is now included in the bureaucracy crate.
- add: Variants can be crafted by adding 2 lv cable or a bike horn after screwdrivering a desk bell.


slarticodefast:
Making it constructible might result in too much sound spam by some players.
I would prefer adding it to the bureaucray cargo crate.

iaada:
> Making it constructible might result in too much sound spam by some players. I would prefer adding it to the bureaucray cargo crate. But the variants could still be made with a normal bell as an ingredient.

I was thinking about it being easy to spam noise. Bureaucracy crate makes sense with the recent economy stuff.

> I'm more concerned about if they can be anchored or not, as some desk bells can't be moved

Achorability isn't something the normal desk bell has. Desk bells seem to get VV'd like this pretty often, though.
image

iaada:
Added the bell to the bureaucracy crate and removed the construction entries. The variants are still crafted with a construction graph.

slarticodefast (approved):
I pushed some indentation fixes, looks good otherwise!

Resources/Prototypes/Entities/Objects/Misc/desk_bell.yml
```diff
@@ -1,50 +1,207 @@
...
+  name: unfinished desk bell
+  description: An empty bell missing its noise maker.
+  components:
+  - type: Sprite
+    sprite: Objects/Misc/desk_bell.rsi
+    state: icon-empty
+  - type: Construction
+    graph: &BellGraph BellGraph
+    node: start
+  - type: Damageable
```
SnappingOpossum, line 13: Just changing this should work.
```suggestion
  - type: Injurable
```

PR #36704, opened 2025-04-19
Add RGA/RSI to Credits

Aeshus:
## About the PR

Somehow no one thought to display the attributions files, so here is that, finally added.

## Why / Balance

Why _hasn't_ this been added.

## Technical details

It reads all the `attributions.yml` and `meta.json` as YAML files (Yaml ~ Json), and then parse out the correct fields. 

Added some pagination stuff as the UI crashes my machine if it was all displayed at once (ScrollContainer + Lots of RichTextLabels = Crash).

There's also a few exceptions used to ensure that any issues with the files are reported. They aren't made a special type/caught specifically as they are not propagated out of the functions.

Also, I reformatted the file. Press the "ignore whitespace changes" option in the GitHub diff for easier review..

## Requirements

## Breaking changes

**Changelog**

No CL needed, should've always been here?


ArtisticRoomba:
I'm thinking of the quadrillion character long toys.rsi meta.json now.

Aeshus:
Mmmm, time to rebase!

EmoGarbage404:
while i understand the sentiment, the actual credits are borderline unreadable spam

image

image
image

It includes filepaths for literally every single sound and sprite in the game. Additionally, the same repositories are credited repeatedly via the commit links.

additionally, generating all this text causes a pronounced lag when opening the credits, which isn't ideal either.

Aeshus:
> It includes filepaths for literally every single sound and sprite in the game. Additionally, the same repositories are credited repeatedly via the commit links.

There is no way to coalesce them. This is the only comprehensive way that I see to actually credit _every single sound and sprite in the game_. Anything less than comprehensive crediting is a failure.

> additionally, generating all this text causes a pronounced lag when opening the credits, which isn't ideal either.

I originally included a tab-based implementation to only load the attributions while on the attributions tab (as this was before I did pagination, so it would actually slow the game to a crawl), but that was removed as it simply moved the time to load to another location and made the code harder to read.

I don't really see a way to handle this without a mass amount of lag, as if you simply ignore the parsing until later, the very large `attributions.yml` files will cause drastic differences in page length, plus some may get large enough to cause further UI lag.

Aeshus:
Also, the fact that is is "borderline unreadable spam" says as much about the current systems as it does about this implementation. All this version does is parse the YAML/JSON and display it on the client in a more readable format.

metalgearsloth:
> additionally, generating all this text causes a pronounced lag when opening the credits, which isn't ideal either.

No one often reads it anyway we get a ping every 6-12 months saying credits are out of date.

Aeshus:
> > additionally, generating all this text causes a pronounced lag when opening the credits, which isn't ideal either.
> 
> No one often reads it anyway we get a ping every 6-12 months saying credits are out of date.

What do you mean by a "credit [is] out of date?"

Also, I do expect people to open the credits, but the lag is only at load-time and I believe is bearable (<1s). The lag is from reading and parsing hundreds of YAML/JSON files, and once it is parsed in memory, the lag goes away.

metalgearsloth:
> What do you mean by a "credit [is] out of date?"

Someone needs to run the powershell files in Tools and actually update the credits.

Aeshus:
> Someone needs to run the powershell files in Tools and actually update the credits.

I believe the credits (at least, the GitHub ones) are automatically updated via a GitHub action.

metalgearsloth:
Patron ones are not and have to be manually done still.

Content.Client/Credits/CreditsWindow.xaml.cs
```diff
@@ -1,184 +1,353 @@
...
+
+        foreach (var stream in rgaStreams)
+        {
+            try
             {
-                if (!first)
+                var yamlStream = _resourceManager.ContentFileReadYaml(stream);
+
+                if (yamlStream.Documents[0].RootNode.ToDataNode() is not SequenceDataNode sequence)
+                    throw new Exception("Attributions file is not a list of attributions.");
```
Tayrtahn, line 175: Same deal with these exceptions.

Content.Client/Credits/CreditsWindow.xaml.cs
```diff
@@ -1,184 +1,353 @@
...
-                var text = _resourceManager.ContentFileReadAllText($"/Credits/{path}");
-                if (markup)
-                {
-                    label.SetMessage(FormattedMessage.FromMarkupOrThrow(text.Trim()));
-                }
-                else
-                {
-                    label.SetMessage(text);
-                }
+            var msg = string.Join(", ", tier.OrderBy(p => p.Name).Select(p => p.Name));
```
Tayrtahn, line 272: These names probably should be escaped before adding them to a FormattedMessage, in case some patron gets the clever idea of adding markup tags to their name.

Tayrtahn (changes requested):

Content.Client/Credits/CreditsWindow.xaml.cs
Aeshus: This change is out of scope of this PR, as that is the current functionality of the credits window.

Content.Client/Credits/CreditsWindow.xaml.cs
```diff
@@ -1,184 +1,353 @@
...
-                Button patronButton;
-                patronsContainer.AddChild(patronButton = new Button
-                {
-                    Text = Loc.GetString("credits-window-become-patron-button"),
-                    HorizontalAlignment = HAlignment.Center
-                });
+                var yamlStream = _resourceManager.ContentFileReadYaml(stream);
+
+                if (yamlStream.Documents[0].RootNode.ToDataNode() is not MappingDataNode map)
+                    throw new Exception("meta.json is not a mapping.");
```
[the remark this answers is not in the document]
Aeshus: None of these propagated out into any public APIs, and are instead only handled locally. 

Generic exceptions already proliferate through the codebase, so I don't see much point in making a special type that will never be used (it's all caught locally) nor have any meaningful difference.

Content.Client/Credits/CreditsWindow.xaml.cs
Tayrtahn: Fair enough

Content.Client/Credits/CreditsWindow.xaml.cs
```diff
@@ -1,184 +1,353 @@
...
+
+                if (!map.TryGet("states", out var statesNode))
+                    throw new Exception("Missing the states field.");
+
+                if (statesNode is not SequenceDataNode states)
+                    throw new Exception("Missing a list of states.");
+
+                var copyright = copyrightNode.ToString();
+                var files = states.Select(n => (MappingDataNode)n)
+                    .Select(n => stream.Directory.ToString().Split("/").Last() + "/" + n.Get("name") + ".png");
```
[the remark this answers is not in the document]
Aeshus: I have some architectural changes I'll be pushing up later that will resolve this.

Content.Client/Credits/CreditsWindow.xaml.cs
```diff
@@ -1,184 +1,353 @@
...
+
+                m.AddMarkupPermissive(Loc.GetString("credits-window-attributions-files",
+                    ("files", string.Join(", ", files))));
+                m.AddText("\n");
+                m.AddMarkupPermissive(Loc.GetString("credits-window-attributions-copyright",
+                    ("copyright", copyright)));
+                m.AddText("\n");
+            }
+            catch (Exception e)
```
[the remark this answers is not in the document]
Aeshus: There's two different types of exceptions that can be passed up, one from YamlDotNet that originates from the yaml parser itself failing (such as malformed/invalid input), and then there's the exceptions originating from these functions, which is from malformed/missing data fields.

There's no singular exception type that handles both cases, and this is simply a catch all, so when there is an error, it will display a message to the user instead of failing silently or crashing the client.

Aeshus:
The parsing has been moved off the main thread, so there's no more UI blocking/lag spike issues, similar to the Changelog window (so there's ~1 second of a blank tab before everything loads).

Aeshus:
Plus, there's now a "Directory" listing to make RSIs easier to handle.

SlamBamActionman (approved):
From a UI perspective it's not the easiest to navigate, but the important part is that it's in there. Gj!

PR #36708, opened 2025-04-19
Pressure Relief Valve

ArtisticRoomba:
## About the PR
Adds the Pressure Relief Valve, a new passive atmospherics device that automatically releases gas depending on the set threshold pressure.

Requires https://github.com/space-wizards/RobustToolbox/pull/5932

## Why / Balance
Atmos doesn't have good ways of automatically releasing pressure when an input pressure side is at a certain level, other than connecting an air alarm to a signal valve. Oh and it was in SS13.

In the future, this will see usage in Flow-based Atmospherics (Flowmos), as well as pipe-bursting.

In Flowmos, the gas network can be automatically vented to make room for more gas in the event of an overpressure (during normal operation, gas is added to the recirculation loop, and if left unchecked, it may clog pipes and force a full stop).

In pipe-bursting, engineers can set up pressure relief valves to relieve excess pressure on a network, in order to prevent a catastrophic failure of a large pipenet.

## Technical details
In short, the predicted code and systems revolving around it are largely inspired :trollface: from the predicted pressure pump. I'm sorry Sloth and Milon.

I will try my best to explain the math I have done for the actual atmospherics side of the system. It _seems_ to be correct and behave correctly in-game, but I don't trust myself.

image
image
image

## Media
Content Client_QScQBqRvCe
Content Client_EmbCNTjBJN
Content Client_a3KRW49Boq

## Requirements

**Changelog**
:cl:
- add: A new atmospherics valve has been added, the Inlet Pressure Regulator! You can learn more about its behavior and potential uses in the Engineering Guidebook.


Coolsurf6:
You are so unfathomably based for this. I will continue to harass people to read the engineering guide book because of your insane work on it, and anything atmos. 

_Your atmos code and knowledge scares me._

ArtisticRoomba:
> You are so unfathomably based for this. I will continue to harass people to read the engineering guide book because of your insane work on it, and anything atmos.
> 
> _Your atmos code and knowledge scares me._

I barely know any C# lol. If you locked me in a room and told me to make an actually unique system that isn't just kid named Update(Frametime) you would never see me again :godo:.

ScarKy0:
Holy shit atmos content.
We can't allow that, closing immediately.

ScarKy0 (changes requested):
Math is scary.

Content.Client/Atmos/UI/GasPressureReliefValveBoundUserInterface.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Content.Client.Atmos.UI;
+
+public sealed class GasPressureReliefValveBoundUserInterface : BoundUserInterface
+{
+    private GasPressureReliefValveWindow? _window;
+
+    public GasPressureReliefValveBoundUserInterface(EntityUid owner, Enum uiKey) : base(owner, uiKey)
+    {
+    }
```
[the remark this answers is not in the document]
VerinSenpai: looks like you're just combining init of the class with the declaration. Looks cleaner. I'm in no position to judge readability though.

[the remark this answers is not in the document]
ArtisticRoomba: I went against using that rider autocomplete because people like doing it the other way, I guess I'll use it.

ScarKy0 (changes requested):
Ok so I gave it an in-game test now.
The pressure field should autofill with a 0, cause currently it doesn't and you have to type in "0" before being able to press any buttons for increasing or decreasing pressure.
Also, the valve is still active when the pressure limit is reached, despite nothing actually happening. (Valve at 2000kPa with 2000kPa limit is basically still active and making the sound)

ArtisticRoomba:
> Also, the valve is still active when the pressure limit is reached, despite nothing actually happening. (Valve at 2000kPa with 2000kPa limit is basically still active and making the sound)

Are you sure? The valve will close and stop working when both sides equalize, I just tested it. And the valve will still continue to work and allow gas to flow because there's flow in the pipenet, if applicable.

> The pressure field should autofill with a 0, cause currently it doesn't and you have to type in "0" before being able to press any buttons for increasing or decreasing pressure.

Sure, I'll autofill it with current pressure. It kinda makes the "insert current pressure" and "insert zero" button redundant but idrc.

ScarKy0:
> Are you sure? The valve will close and stop working when both sides equalize, I just tested it. And the valve will still continue to work and allow gas to flow because there's flow in the pipenet, if applicable.

To my knowledge there was no flow anymore, but it stayed enabled. Maybe I checked incorrectly?

ArtisticRoomba:
Due to funny floats it'll keep transferring small values until it truly equalizes and the gas stops flowing, and the valve will shut off and stop making noise in that case. The UI and gas analyzer won't show those small changes but there is still technically some flow, but it only takes a few seconds to truly equalize.

If there is any amount of gas flow in the input side which will cause a pressure rise if unreleased, then the valve will open. The pressure you see in the gas analyzer is static at the set threshold pressure because the valve is working.

ScarKy0 (approved):

ArtisticRoomba:
The math for pressure equalization is partially wrong: the math has to consider the temperature change on T_2 when the gas is mixed with the T_1 gas.

slarticodefast:
Started working on fixing the math, will finish tomorrow
https://hackmd.io/@Slart/BJUt5oWylg

Content.Client/Atmos/UI/GasPressureReliefValveBoundUserInterface.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    protected override void UpdateState(BoundUserInterfaceState state)
+    {
+        base.UpdateState(state);
+
+        if (state is not GasPressureReliefValveBoundUserInterfaceState valveState || _window == null)
+            return;
+
+        _window.SetThresholdPressureLabel(valveState.ThresholdPressure);
+    }
```
slarticodefast, line 67: this is not needed since the networking is already done using component states and the UI updates on `AfterAutoHandleState`

Content.Client/Atmos/UI/GasPressureReliefValveWindow.xaml
```diff
@@ -0,0 +1,62 @@
+﻿<DefaultWindow xmlns="https://spacestation14.io"
```
slarticodefast, line 1: I'm not an UI expert but I think using FancyWindow is preferred for new UIs

Content.Shared/Atmos/Piping/Binary/Components/SharedGasPressureReliefValveComponent.cs
```diff
@@ -0,0 +1,63 @@
...
+public enum GasPressureReliefValveUiKey : byte
+{
+    Key,
+}
+
+/// <summary>
+/// Represents the state of the Gas Pressure Relief Valve user interface.
+/// </summary>
+[Serializable, NetSerializable]
+public sealed class GasPressureReliefValveBoundUserInterfaceState : BoundUserInterfaceState
```
slarticodefast, line 18: this should be fully handled by the component state

Resources/Prototypes/Entities/Structures/Piping/Atmospherics/binary.yml
```diff
@@ -150,6 +150,63 @@
...
+    range: 5
+    sound:
+      path: /Audio/Ambience/Objects/gas_hiss.ogg
+  - type: AtmosMonitoringConsoleDevice
+    navMapBlip: GasFlowRegulator
+  - type: GuideHelp
+    guides:
+    - PressureReliefValve
+
+
```
slarticodefast, line 209:
```suggestion
```

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+                    nameof(valveEntity.Comp.Enabled));
+            }
+
+            return;
+        }
+
+        // guess we're doing work now
+        valveEntity.Comp.Enabled = true;
+        _ambientSoundSystem.SetAmbience(valveEntity, true);
+        UpdateAppearance(valveEntity);
```
slarticodefast, line 108: see the comment above
this will set them every single atmos tick

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+/// Handles serverside logic for pressure relief valves. Gas will only flow through the valve
+/// if the pressure on the inlet side is over a certain pressure threshold.
+/// See https://en.wikipedia.org/wiki/Relief_valve
+/// </summary>
+[UsedImplicitly]
+public sealed class GasPressureReliefValveSystem : SharedGasPressureReliefValveSystem
+{
+    [Dependency] private readonly AtmosphereSystem _atmosphereSystem = default!;
+    [Dependency] private readonly NodeContainerSystem _nodeContainer = default!;
+    [Dependency] private readonly SharedAmbientSoundSystem _ambientSoundSystem = default!;
```
slarticodefast, line 24:
```suggestion
    [Dependency] private readonly AtmosphereSystem _atmosphere = default!;
    [Dependency] private readonly NodeContainerSystem _nodeContainer = default!;
    [Dependency] private readonly SharedAmbientSoundSystem _ambientSound = default!;
```
naming convention

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+        UpdateAppearance(valveEntity);
+
+        var t1 = inletPipeNode.Air.Temperature;
+
+        // First, calculate the amount of gas we need to transfer to bring us below the threshold.
+        var deltaMolesToPressureThreshold =
+            AtmosphereSystem.MolesToPressureThreshold(inletPipeNode.Air, valveEntity.Comp.Threshold);
+
+        // Second, calculate the moles required to equalize the pressure.
+        // Cursed rounding to avoid the floating point error-inator.
```
slarticodefast, line 117: what kind of errors happened here?
Seems weird that it is not numerically stable

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        // And finally, limit the transfer volume to the max flow rate of the valve.
+        var actualVolumeToTransfer = Math.Min(desiredVolumeToTransfer,
+            valveEntity.Comp.MaxTransferRate * _atmosphereSystem.PumpSpeedup() * args.dt);
+
+        // We remove the gas from the inlet and merge it into the outlet.
+        var removed = inletPipeNode.Air.RemoveVolume(actualVolumeToTransfer);
+        _atmosphereSystem.Merge(outletPipeNode.Air, removed);
+
+        // Calculate the flow rate in L/S for the UI.
```
slarticodefast, line 140:
```suggestion
        // Calculate the flow rate in L/s for the UI.
```
fun physics units fact:
capital letters are used for units named after people, lower case for everything else
liter is an exception (at least in the US) as it looks similar to a 1

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// The <see cref="GasPressureReliefValveComponent" /> of the entity. Will be resolved if not provided.
+    /// </param>
+    /// <param name="appearance">
+    /// The <see cref="AppearanceComponent" /> associated with the entity. Will be resolved if not provided.
+    /// <remarks>
+    /// This isn't in shared because it's pointless to predict something whose state change is
+    /// entirely handled by the server.
+    /// </remarks>
+    /// </param>
+    private void UpdateAppearance(EntityUid valveEntityUid,
```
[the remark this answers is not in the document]
ArtisticRoomba: I used `Entity<GasPressureReliefValveComponent, AppearanceComponent?>` instead as I think `GasPressureReliefValveComponent` can't be null in this instance. Let me know if I'm incorrect.

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -0,0 +1,162 @@
...
+                out PipeNode? outletPipeNode))
+        {
+            // Prevent spamming updates if the valve is already disabled.
+            if (valveEntity.Comp.Enabled)
+            {
+                valveEntity.Comp.Enabled = false;
+                _ambientSound.SetAmbience(valveEntity, false);
+                UpdateAppearance(valveEntity);
+                DirtyField(valveEntity!, nameof(valveEntity.Comp.Enabled));
+            }
```
slarticodefast, line 61: Since we have this pattern copy pasted multiple times in the code I would just move it into a `SetEnabled` helper function. And in the other cases you were setting the flow rate to 0, but not here.

Content.Client/Atmos/UI/GasPressureReliefValveWindow.xaml.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        SetToCurrentPressureButton.OnPressed += _ =>
+        {
+            if (ThresholdPressureLabel.Text != null)
+            {
+                ThresholdInput.Text = ThresholdPressureLabel.Text;
+            }
+
+            SetThresholdButton.Disabled = false;
+        };
```
slarticodefast, line 35: The two buttons are acting weird.
If I understand it correctly the `Set` button will set it to whatever is currently in the text box.
But what does the `Set to current pressure` button do? I would assume it would change it to the current pressure in the valve input node to quickly keep the current pressure stable, but it doesn't seem to do anything when testing.

slarticodefast (changes requested):

Content.Client/Atmos/UI/GasPressureReliefValveWindow.xaml.cs
ArtisticRoomba: Yeah, that button is redundant and unintuitive. I'm really really considering scope creeping the UI (sorry) so I'll give it a face-lift.

ArtisticRoomba:
Tests catch shitter trying to shitcode UI changes in 4K UHD

ArtisticRoomba:
PR now requires https://github.com/space-wizards/RobustToolbox/pull/5932

ArtisticRoomba:
PR now requires content engine update

ArtisticRoomba:
Went ahead and scopecrept the UI, implementation feels sussy though, was an MVP. I'm so sorry Slart.

Content Client_vASV1emp89

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -110,9 +113,9 @@ Can be used to calculate the amount of gas we need to transfer.
 
         // Calculate the flow rate in L/s for the UI.
         valveEntity.Comp.FlowRate = MathF.Round(actualVolumeToTransfer * args.dt, 1);
-        DirtyField(valveEntity, valveEntity.Comp, nameof(valveEntity.Comp.FlowRate));
 
         ChangeStatus(true, valveEntity);
+        SendUIInfo(p1, p2, valveEntity.Comp.FlowRate, valveEntity.Owner);
```
slarticodefast, line 141: Manually networking this via an UI message seems fine, but if we do it every single atmos tick then we have the same problem as with DirtyField. I would add a check comparing a `NextUpdate` timespan to `CurTime` and only update it once a second.

slarticodefast (changes requested):

slarticodefast:
I pushed some minor cleanup, looks good otherwise.
I'm still not completely certain on regularly dirtying it for the flowrate, but only happens on a timer and I prefer that over the UI mispredicting. If turns out to be too expensive we will have to change it again.

The last thing that needs to be done is adjusting this PR to the new pipe layers.

slarticodefast (changes requested):
Oh, another thing I noticed:
Move the component from `Components` to `Piping/Binary/Components`

Content.Shared/Atmos/Components/GasPressureReliefValveComponent.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// </summary>
+    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
+    [AutoPausedField]
+    public TimeSpan NextUiUpdate = TimeSpan.Zero;
+
+    /// <summary>
+    /// Specifies the pipe node name to be treated as the outlet.
+    /// </summary>
+    [DataField]
+    public string OutletName = "outlet";
```
slarticodefast: move this below the inled datafield

Content.Shared/Atmos/Components/GasPressureReliefValveComponent.cs
```diff
@@ -0,0 +1,88 @@
...
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
+namespace Content.Shared.Atmos.Components;
+
+/// <summary>
+/// Defines a gas pressure relief valve,
+/// which releases gas depending on a set pressure threshold between two pipe nodes.
+/// </summary>
+[RegisterComponent] [NetworkedComponent]
+[AutoGenerateComponentState(true, true)] [AutoGenerateComponentPause]
```
slarticodefast:
```suggestion
[RegisterComponent, NetworkedComponent]
[AutoGenerateComponentState(true, true), AutoGenerateComponentPause]
```

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
```diff
@@ -0,0 +1,202 @@
...
+        var deltaMolesToPressureThreshold =
+            AtmosphereSystem.MolesToPressureThreshold(inletPipeNode.Air, valveEntity.Comp.Threshold);
+
+        // Second, calculate the moles required to equalize the pressure.
+        // We round here to avoid the valve staying enabled for 0.00001 pressure differences.
+        var deltaMolesToEqualizePressure =
+            float.Round(_atmosphere.FractionToEqualizePressure(inletPipeNode.Air, outletPipeNode.Air) *
+                        inletPipeNode.Air.TotalMoles,
+                1,
+                MidpointRounding.ToPositiveInfinity);
```
Partmedia: Would it make sense to lower this rounding into `FractionToEqualizePressure`, since other (future) callers of this method would presumably run into the same issue?

Partmedia (commented):
Thank you for the excellent write-up. Is there any way to include that in-line with the source code, so that it can be referred to in the future?

Minor (non-blocking nitpick: The intent of this device is to be passive, but the input widget makes this device feel distinctly electronic to me.

We also have a core design principle of having separate devices do different things that can be connected together, which seems to call for the flow rate meter to be in a separate device. Namely, that device already exists as the "passive valve".

Consider revisiting what we want to call this device. "Pressure relief" implies that it is only for emergencies, whereas a "back-pressure regulator" implies that it could be useful for controlling the pressure of the input.

ArtisticRoomba:
> Thank you for the excellent write-up. Is there any way to include that in-line with the source code, so that it can be referred to in the future?

I'd like to store my entire write-up in the codebase somehow as documentation, as I think that abstracting it in an attempt to store it in the code wouldn't be helpful (the proof is left as a simple exercise to the reader type stuff).

Would it be ideal to start documenting this in the dev wiki, or just to start an area of documentation in the game's repo that can be referenced for people to read fully?

Stuff like heat exchanger code would be easier to read if there was just an xml doc I can look at in Rider as a general reference of the derivations/processes done.

> We also have a core design principle of having separate devices do different things that can be connected together, which seems to call for the flow rate meter to be in a separate device. Namely, that device already exists as the "passive valve".

After talking with it more with you on Discord, I understand the vision, and I plan on moving a lot of the flow rate/pressure sensing information to the passive gate, as we now (or at least now I know) have a decently defined methodology for communicating information from server to client for UI prediction.

> Consider revisiting what we want to call this device. "Pressure relief" implies that it is only for emergencies, whereas a "back-pressure regulator" implies that it could be useful for controlling the pressure of the input.

Probably going to rename it to the pressure regulator as discussed.

Partmedia:
> I'd like to store my entire write-up in the codebase somehow as documentation, as I think that abstracting it in an attempt to store it in the code wouldn't be helpful (the proof is left as a simple exercise to the reader type stuff).
> 
> Would it be ideal to start documenting this in the dev wiki, or just to start an area of documentation in the game's repo that can be referenced for people to read fully?

I think it would need to be close to the actual source code, e.g. inline, otherwise it's going to be hard to find and many people aren't going to realize that it is there.

I don't see a good way to do this now, but perhaps to start you can just leave the whole text, and literal $\LaTeX$ inline for now.

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureReliefValveSystem.cs
ArtisticRoomba: Given that this will likely be seen in a lot of places in the future I'll go ahead and move this logic into `FractionToEqualizePressure`

ArtisticRoomba:
going to do large redocumentation tmro

ArtisticRoomba:
It isn't an atmos PR if the lines of documentation surpass the LOC for the helper method.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.Gases.cs
```diff
@@ -252,6 +252,134 @@ public void ScrubInto(GasMixture mixture, GasMixture destination, IReadOnlyColle
...
+            // Ensure that P_1 > P_2 so the quadratic works out.
+            if (gasMixture1.Pressure < gasMixture2.Pressure)
+            {
+                (gasMixture1, gasMixture2) = (gasMixture2, gasMixture1);
+            }
+
+            // Establish the dimensionless ratios.
+            var volumeRatio = gasMixture2.Volume / gasMixture1.Volume;
+            var molesRatio = gasMixture2.TotalMoles / gasMixture1.TotalMoles;
+            var temperatureRatio = gasMixture2.Temperature / gasMixture1.Temperature;
```
Partmedia, line 352: Could you check what (if anything) prevents divide by near zero in these cases?

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureRegulatorSystem.cs
```diff
@@ -0,0 +1,200 @@
...
+    [Dependency] private readonly AtmosphereSystem _atmosphere = default!;
+    [Dependency] private readonly NodeContainerSystem _nodeContainer = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<GasPressureRegulatorComponent, ComponentInit>(OnInit);
+        SubscribeLocalEvent<GasPressureRegulatorComponent, AtmosDeviceUpdateEvent>(OnReliefValveUpdated);
```
Partmedia: It would be good to update this method name now.

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureRegulatorSystem.cs
```diff
@@ -0,0 +1,200 @@
...
+        SubscribeLocalEvent<GasPressureRegulatorComponent, MapInitEvent>(OnMapInit);
+    }
+
+    private void OnMapInit(Entity<GasPressureRegulatorComponent> ent, ref MapInitEvent args)
+    {
+        ent.Comp.NextUiUpdate = _timing.CurTime + ent.Comp.UpdateInterval;
```
Partmedia, line 40: I know this is a pattern used in the existing codebase, but if you are not going to randomize the initial update time, what is the point of keeping per-component next-update times? The map inits effectively "all at once" so all of these updates will also be "all at once."

ArtisticRoomba: We keep per-component update times so they're automatically paused and unpaused if we pause/unpause the game. I didn't really want to do a global update in the system (because I didn't know how to do it, and a per component update time was easier) and Slarti said it was fine.

These update times determine how often the regulator gets dirtied so the client has predicted information available to it.

Content.Server/Atmos/EntitySystems/AtmosphereSystem.Gases.cs
ArtisticRoomba: The initial conditions require that `gasMixture1` and `gasMixture2` are non-null with the pressure of `gasMixture1` being greater than `gasMixture2`. If the case is the opposite the code swaps the two via deconstruction.

Ex. if we had a `gasMixture1` at 4500 kPa and a `gasMixture2` at 0.001 kPa, the math would be $\frac{0.001 kPa}{4500 kPa} = 2.22... \cdot 10^{-7}$. If this was the opposite case for whatever reason, the gasMixtures would be swapped to ensure that `gasMixture2.Pressure < gasMixture1.Pressure` and the result would be the same.

I have also done testing and logged the ratios for the first few runs of the method in an instance where a pressure regulator is releasing pressure to a completely spaced pipe (one with zero pressure):
```
vol 3
mol 0
temp 1
heatcap 5.1997542E-08

vol 3
mol 0.0022221396
temp 1
heatcap 0.0022221394

vol 3
mol 0.004444183
temp 1
heatcap 0.0044441824

vol 3
mol 0.0066661304
temp 1
heatcap 0.0066661304

vol 3
mol 0.008887981
temp 1
heatcap 0.008887981

File: Content.Server/Atmos/EntitySystems/AtmosphereSystem.Gases.cs [hunk 20]
```diff
@@ -294,8 +357,13 @@ and the transfer fraction is solved for.
...
 
-            // Ahhh, Perry the Platypus, I see you've found my secret FLOATING POINT ERROR-INATOR!
-            return (-quadraticB + MathF.Sqrt(quadraticB * quadraticB - 4 * quadraticA * quadraticC)) / (2 * quadraticA);
+            var solvedQuadratic = (-quadraticB + MathF.Sqrt(quadraticB * quadraticB - 4 * quadraticA * quadraticC)) /
+                                  (2 * quadraticA);
+
+            // Round to avoid the FLOATING POINT ERROR-INATOR!
+            return float.Round(solvedQuadratic,
+                1,
+                MidpointRounding.ToPositiveInfinity);
```
slarticodefast: Before you were rounding the number of moles to the first decimal place, but now you round the fraction, which is between 0 and 1, so it becomes highly inaccurate.
And I would rather keep the calculation itself exact, as it may be used for other purposes and the rounding should only be done when absolutely needed.

ArtisticRoomba: Don't code at 3am :trollface:
I undid these changes and just put them back in the pressure regulator method. People will likely want different methods of rounding for whatever they use the method for.

Partmedia:
I would name this an "inlet pressure regulator" because it regulates the inlet pressure. Please update the descriptions to also reflect this. Otherwise, it is confusing which port is actually being regulated.

Partmedia:
I think the moving flow meter was a cool idea, but it is confusing what is going on when there is no flow. The flow meter bar looks "stuck":

image

Note also that real (passive) gas flow meters suspend a weight in the path of the flow, which would give an output bar proportional to the flow, instead of scroll like the current behavior.

image

Since this is a inlet port regulator, it is not clear to me what the progress bar next to "Target" represents.

We discussed some minor adjustments to the pressure setpoint widget to make it feel more "analog". Were you saving those for a future PR after this is merged?

Partmedia:
Why does the inlet pressure label frequently disagree with the gas analyzer? The output seems to match more.

image

ArtisticRoomba:
> I think the moving flow meter was a cool idea, but it is confusing what is going on when there is no flow. The flow meter bar looks "stuck":

So, and this is silly, but the idea was to replicate a water flow meter, with its spinny diamond. I then scaled back my feature creep to a level I was comfortable with, so that is the current implementation.

For each bar filled, 50 liters have gone by, so it gives a sense of volume that's at least partially physical, and you can get a good timescale, in my opinion.

Of course, the proper diegetic implementation would be a suspension flow meter.

> Since this is a inlet port regulator, it is not clear to me what the progress bar next to "Target" represents.

I'm probably going to rename it to Setpoint.

> We discussed some minor adjustments to the pressure setpoint widget to make it feel more "analog". Were you saving those for a future PR after this is merged?

In a future PR, I was going to make the inlet pressure regulator a completely analog device: it would purely tell you if it was ON/OFF using an Apollo-style [barber pole indicator](https://apolloexperience.com/2019/07/23/talkbacks/) or maybe a standard controls thing.

In the PR following that, I would probably redesign the flow meter to be a proper suspension flow meter and have it be on the passive gate.

> Why does the inlet pressure label frequently disagree with the gas analyzer? The output seems to match more.

This is an intentional technical limitation. We only dirty the UI to provide the client with pressure info on the valve every second to prevent network spamming. This will desync with the usual ~0.5 atmos $dt$.

Content.Client/Atmos/EntitySystems/GasPressureRegulatorSystem.cs
```diff
@@ -0,0 +1,31 @@
+using Content.Shared.Atmos.EntitySystems;
+using Content.Shared.Atmos.Piping.Binary.Components;
+
+namespace Content.Client.Atmos.EntitySystems;
+
+/// <summary>
+/// Represents the client system responsible for managing and updating the gas pressure relief valve interface.
```
slarticodefast:
```suggestion
/// Represents the client system responsible for managing and updating the gas pressure regulator interface.
```

Content.Shared/Atmos/EntitySystems/SharedGasPressureRegulatorSystem.cs
```diff
@@ -0,0 +1,70 @@
+﻿using Content.Shared.Administration.Logs;
+using Content.Shared.Atmos.Components;
+using Content.Shared.Atmos.Piping.Binary.Components;
+using Content.Shared.Database;
+using Content.Shared.Examine;
+
+namespace Content.Shared.Atmos.EntitySystems;
+
+/// <summary>
+/// Handles all shared interactions with the gas pressure relief valve.
```
slarticodefast:
```suggestion
/// Handles all shared interactions with the gas pressure regulator.
```

Content.Shared/Atmos/Piping/Binary/Components/GasPressureRegulatorComponent.cs
```diff
@@ -0,0 +1,88 @@
+﻿using Content.Shared.Guidebook;
+using Robust.Shared.GameStates;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
+namespace Content.Shared.Atmos.Piping.Binary.Components;
+
+/// <summary>
+/// Defines a gas pressure relief valve,
```
slarticodefast:
```suggestion
/// Defines a gas pressure regulator,
```

Content.Shared/Atmos/Piping/Binary/Components/SharedGasPressureRegulatorComponent.cs
```diff
@@ -0,0 +1,25 @@
...
+/// Represents the unique key for the UI.
+/// </summary>
+[Serializable, NetSerializable]
+public enum GasPressureRegulatorUiKey : byte
+{
+    Key,
+}
+
+/// <summary>
+/// Message sent to change the pressure threshold of the Gas Pressure Relief Valve.
```
slarticodefast:
```suggestion
/// Message sent to change the pressure threshold of the gas pressure regulator.
```

Content.Server/Atmos/Piping/Binary/EntitySystems/GasPressureRegulatorSystem.cs
slarticodefast: Most of the devices will be built during the round rather than being there on rounstart, so I would prefer to keep the updates spread apart from each other to distribute the networking load in case someone builds a few hundred of them.

Issue #36692, opened 2025-04-18
experimental RCD inconsistent charge use - Vulture Test Server


reivehnFliyt:
**Description**

When using Cyborg experimental RCD, Inconsistent use of charges was noticed. 

One grill cost all 20 charges on one use, and only 4 on another use, thereafter some cases of unlimited RCD use without depleting charges was witnessed when placing reinforced windows.

Unsure of exact cause.

ArtisticRoomba:
Possible cause could be the recent limited item charges refactor?

reivehnFliyt:
Seems to be working again now, Good the refactor seems to be implemented now.

reivehnFliyt:
Nevermind, Still using all 20 charges per use, but seems at the start of a game the first use uses just the number of charges it should

ScarKy0:
Does the EMAG and AD also break?

metalgearsloth:
> Seems to be working again now, Good the refactor seems to be implemented now.

Literally didn't change anything in that time lol.

Anyway steps to repro I beg of you.

ArtisticRoomba:
Unable to reproduce in random testing, please provide reproduction steps or maybe a round number.

reivehnFliyt:
So, Loading into a Cyborg at the beginning of a round on Vulture, Selected RCD, was using all charges in a single use after the first use of the round, right now it seems stable and not repeating after recent server update. think it might have been fixed? not sure yet, could be triggered some other way i havent figured yet.

metalgearsloth:
> So, Loading into a Cyborg at the beginning of a round on Vulture, Selected RCD, was using all charges in a single use after the first use of the round, then when recharging, as long as there were enough charges to cover the cost it would build without removing charges until fully recharged, and only then take all charges upon use.
> 
> Right now it seems stable and not repeating after recent server update. think it might have been fixed? not sure yet, could be triggered some other way i havent figured yet.

Server hasn't updated bruh but I imagine timer not updating on mapinit so will look into it.

reivehnFliyt:
yes, i see that, said the server had an update and restarted so i took a break and came back but was probably a regular restart?

anyway, just used the RCD again and i managed to trigger it. trying to get a clean screenshot now, used once and it used all 20 charges making a fluorescent light

<img width="502" alt="Image" src="https://github.com/user-attachments/assets/8a8b5055-51f8-475a-80a9-1bbc0e68c1af" />
_one reinforced window at full charge, 20 charges expended_

[...]

PR #36714, opened 2025-04-19
Fix charges

metalgearsloth:
We have the absolute value so use that.

Resolves https://github.com/space-wizards/space-station-14/issues/36692

**Changelog**
:cl:
- fix: Fix charges sometimes breaking for auto-recharge devices such as RCDs.


ScarKy0 (approved):
Simple enough.

Issue #36679, opened 2025-04-18
borgs with custom names don't have name identifier and their name in borg ui is cut


lzk228:
## Description

if you set a custom name in loadout for borg,
1. you will not have default name identifier (Si-XXXX)
2. their name will be cut in borg's maintenance panel

https://github.com/space-wizards/space-station-14/pull/35170

**Reproduction**

1. set a custom name for borg in loadout
2. join game
3. your name is only what you set, without (Si-XXXX)
4. open the borg's maintenance panel
5. try to set another name

**Screenshots**

**Additional context**

I had one solution but i find it quite bad
Just call the method, that sets a name identifier, again after an entity with NameIdentifierComponent gets custom name from loadout

walksanatora:
also try checking if the name is less then 9 chars. causse I (downstream) had a issue where it would crash the client if (since it tries to substring the nameIdentifierComp from the full name and since it's length is longer then the whole nname it crashes.) I fixed this by wrapping the substring call to be
```cs
            var name = fullName.EndsWith(nameIdentifierComponent.FullIdentifier)
                ? fullName.Substring(0, fullName.Length - nameIdentifierComponent.FullIdentifier.Length - 1) //old behaviour
                : fullName; // our full name does not contain the Name Identifer so we use the full name.
```

Tayrtahn:
I'm thinking the right approach here is to make `NameIdentifier` use `NameModifierSystem`.

PR #36736, opened 2025-04-19
Convert `NameIdentifier` to use `NameModifierSystem`

Tayrtahn:
## About the PR

`NameIdentifier` now uses `NameModifierSystem` to add its identifier to the entity's name.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/36679.
This also makes it easier to add and remove identifiers from entities cleanly, and removes the need for some slightly-funky string manipulation in the borg UI.
This may make it possible to add an identifier to the AI as well, but I haven't actually tried that out.

## Technical details

- `NameIdentifierSystem` no longer directly alters the name of the entity it's on. Instead, it subscribes to the `RefreshNameModifiers` event and adds the identifier as a modifier.
- The borg UI code was altered to get the borg's base name using `NameModifierSystem.GetBaseName` instead of trying to get it by taking a substring of the entity name (which could fail and throw an exception if the name had been made short enough).
- The logic for applying the identifier to the entity name now uses the localization system, because `NameModifierSystem` wants a LocId (and it's better that way anyway).

## Note to Admins
Due to this change, renaming an entity with an identifier will preserve the identifier in the name. For example, renaming "mouse (624)" to "Bubba the Great" will produce "Bubba the Great (624)". If you want to remove the identifier, simply remove the `NameIdentifier` component from the entity.

## Media

Borg UI functioning with a changed name:
Screenshot 2025-04-19 at 12 38 16 PM
Screenshot 2025-04-19 at 12 38 31 PM

## Requirements

## Breaking changes

`NameIdentifier` is now applied as a name modifier. If you need to get an entity's base name (without the identifier), use `NameModifierSystem.GetBaseName`.

**Changelog**

Not player-facing.

### DIFF
--- a/Content.Client/Silicons/Borgs/BorgMenu.xaml.cs
+++ b/Content.Client/Silicons/Borgs/BorgMenu.xaml.cs
@@ -1,6 +1,7 @@
 using Content.Client.Stylesheets;
 using Content.Client.UserInterface.Controls;
 using Content.Shared.NameIdentifier;
+using Content.Shared.NameModifier.EntitySystems;
 using Content.Shared.Preferences;
 using Content.Shared.Silicons.Borgs;
 using Content.Shared.Silicons.Borgs.Components;
@@ -15,6 +16,7 @@ namespace Content.Client.Silicons.Borgs;
 public sealed partial class BorgMenu : FancyWindow
 {
     [Dependency] private readonly IEntityManager _entity = default!;
+    private readonly NameModifierSystem _nameModifier;
 
     public Action? BrainButtonPressed;
     public Action? EjectBatteryButtonPressed;
@@ -31,6 +33,8 @@ public sealed partial class BorgMenu : FancyWindow
     {
         RobustXamlLoader.Load(this);
         IoCManager.InjectDependencies(this);
+
+        _nameModifier = _entity.System<NameModifierSystem>();
 
         _lastValidName = NameLineEdit.Text;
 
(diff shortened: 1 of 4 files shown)


Content.Client/Silicons/Borgs/BorgMenu.xaml.cs
```diff
@@ -15,6 +16,7 @@ namespace Content.Client.Silicons.Borgs;
 public sealed partial class BorgMenu : FancyWindow
 {
     [Dependency] private readonly IEntityManager _entity = default!;
+    private readonly NameModifierSystem _nameModifier;
```
ScarKy0, line 19: Any reason [Dependency] was skipped? I assume cause UI code?

Tayrtahn: Yep.

EmoGarbage404:
i think there may be vestigial name identifier stuff left inside of the borg code that you might want to prune.

Tayrtahn:
> i think there may be vestigial name identifier stuff left inside of the borg code that you might want to prune.

I checked all uses of `Identifier` and `FullIdentifier` and cleaned them up, but I'll take another look for anything else lurking.

Content.Client/Silicons/Borgs/BorgMenu.xaml.cs
metalgearsloth: For future archaeologists the reason is system lifetime only starts when you actually connect to the game but UI starts well before that. Anywhere you can use it for systems there's some background stuff going on to ensure it.

PR #36745, opened 2025-04-19
Fix artifact spawns parenting to players

EmoGarbage404:
## About the PR

artifacts use entity table spawner which gets unhappy when not just chilling on the ground.
When a player would use a handheld artifact inhand, it would spawn things parented to the player instead of the ground.

these changes fix that.

## Requirements

## Breaking changes

**Changelog**

no cl no fun


Tayrtahn (approved):
Shrimple enough.

PR #36759, opened 2025-04-20
Adjust entityTableContainerFill to work better with Storage

EmoGarbage404:
## About the PR

Adds in some custom handling for storage since it has some unique handling logic that doesnt get called on typical insert events.
also adds in the size-based sorting that StorageFill does.

the goal is better parity with storageFill as we start to decommission it. problem is that not using the StorageSystem insert function would cause the stack merging logic to skip which sucks.

apologies if this isn't comprehensible its nearly 2 am

## Technical details

container fill now prioritizes larger items first so things have the lowest chance of conflicts.

## Requirements

## Breaking changes

**Changelog**

no cl no fun

### DIFF
--- a/Content.Shared/Containers/ContainerFillSystem.cs
+++ b/Content.Shared/Containers/ContainerFillSystem.cs
@@ -1,15 +1,22 @@
 using System.Linq;
 using System.Numerics;
 using Content.Shared.EntityTable;
+using Content.Shared.Item;
+using Content.Shared.Storage;
+using Content.Shared.Storage.EntitySystems;
 using Robust.Shared.Containers;
 using Robust.Shared.Map;
+using Robust.Shared.Prototypes;
 
 namespace Content.Shared.Containers;
 
 public sealed class ContainerFillSystem : EntitySystem
 {
+    [Dependency] private readonly IPrototypeManager _prototype = default!;
     [Dependency] private readonly SharedContainerSystem _containerSystem = default!;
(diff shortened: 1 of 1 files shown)


Content.Shared/Containers/ContainerFillSystem.cs
```diff
@@ -68,11 +78,20 @@ private void OnTableMapInit(Entity<EntityTableContainerFillComponent> ent, ref M
...
+                    ? _item.GetItemShape(item).GetArea()
+                    : int.MaxValue);
+
             foreach (var proto in spawns)
             {
                 var spawn = Spawn(proto, coords);
-                if (!_containerSystem.Insert(spawn, container, containerXform: xform))
+
+                var success = storageComp != null && containerId == StorageComponent.ContainerId
+                    ? _storage.Insert(ent, spawn, out _, storageComp: storageComp, playSound: false)
```
ArtisticRoomba, line 91: This will return true on partial insertions, which means that if a bag is full of steel and it tries to insert more, but only inserts half, it'll be stuck with the other half but still declare it true. Could be a problem in the future?

ArtisticRoomba (commented):
One potential problem I saw

SnappingOpossum:
If you're still working on this could you also move the NestedSeletor in FoodMealHappyHonkClown back to the bottom? Had to have it at the top because of the lack of sorting.

PR #36773, opened 2025-04-20
Cleanup warnings in ChatSystem

Tayrtahn:
## About the PR

Fixes 3 warnings in `ChatSystem.cs`.

## Why / Balance

## Technical details

The current code makes the frightening move of calling `ToString()` on the passed `SoundSpecifier` and passing the resulting string to `SharedAudioSystem.PlayGlobal` as the sound path, but there doesn't seem to be any guarantee that the string will actually _be_ a path - in fact I'm pretty sure it's guaranteed _not_ to be a path, since `SoundPathSpecifier` overrides `ToString`:
```cs
override public string ToString() => $"SoundPathSpecifier({Path})";
```
So I think these methods will actually break if we ever call them and pass in a value for the `announcementSound` parameter - because we never do, the `??` operator winds up using the const string in the class instead and it works.

In any case, we now no longer call `ToString()` and instead just pass the `SoundSpecifier` into `PlayGlobal` like we should. The fallback `const string DefaultAnnouncementSound` was changed to a `readonly SoundSpecifier DefaultAnnouncementSound` to fall in line with this change and meet modern standards.

## Requirements

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -64,7 +64,7 @@ public sealed partial class ChatSystem : SharedChatSystem
     public const int VoiceRange = 10; // how far voice goes in world units
     public const int WhisperClearRange = 2; // how far whisper goes while still being understandable, in world units
     public const int WhisperMuffledRange = 5; // how far whisper goes at all, in world units
-    public const string DefaultAnnouncementSound = "/Audio/Announcements/announce.ogg";
+    public readonly SoundSpecifier DefaultAnnouncementSound = new SoundPathSpecifier("/Audio/Announcements/announce.ogg");
 
     private bool _loocEnabled = true;
     private bool _deadLoocEnabled;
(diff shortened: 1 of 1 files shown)


VerinSenpai:
I've been avoiding touching things I thought may be frozen. Is that a major concern for cleanup?

Tayrtahn:
> I've been avoiding touching things I thought may be frozen. Is that a major concern for cleanup?

That's a fair point - I'll leave that decision up to other maintainers.

In this case, the bug fix in addition to the cleanup makes me think that this is a bit more justified, but we'll see!

slarticodefast (approved):
Code looks good but I will wait for approval from @SlamBamActionman to merge this so we don't cause unneeded conflicts for the chat refactor.

Fildrance (approved):

Fildrance:
Chat system refactor was discarded and undergoing re-doing so this is fine (and removing warnings here and now is better then having them possibly removed in some possible future)

PR #36032, opened 2025-03-23
Contraband examine changes (again)

Killerqu00:
## About the PR
Contraband examine uses hover examine as requested in #32250.
The icon also changes both color and shape depending on whether or not you are allowed to possess the item.

## Why / Balance
fancy

## Technical details
replaced AddDetailedExamine with AddHoverExamine, added textures, fixed a warning while at it

## Media
image
image
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
no

**Changelog**
:cl:
- add: Contraband examine icon now shows whether or not you are allowed to possess the item.
- tweak: Contraband examine activates by hovering instead of clicking.

PR #36788, opened 2025-04-21
Allow departments and jobs to be implicitly cleared to use job/department restricted items

DinnerCalzone:
## About the PR
This PR adds support to contraband to clear specific departments and jobs to use job/department restricted items without explicitly stating them in the examine text. This implicitly clears Central Command jobs and Captain to use all job and department restricted items and clears the Head of Personnel to use items that are restricted to service jobs by some degree.

This PR also removes the redundant BaseCentcommCommandContraband prototype, since all members of the CentComm department are already part of the Command department and it is only used for the station master encryption key entity.

## Why / Balance
Currently, if the Captain were to examine a contraband item that is not restricted to either the Captain job or the Command department (for example, a pair of galoshes), the examine text states they should probably avoid carrying this item without good reason, despite holding jurisdiction over all departments. This change gives the Captain positive feedback that they may use the restricted item without bloating the contraband examine text. This also lets heads of departments be cleared to use items that are restricted to specific jobs within their department, such as the Head of Personnel being cleared to use bartender-restricted items.

## Technical details
Adds the `ImplicitlyAllowedDepartments` and `ImplicitlyAllowedJobs` to `ContrabandComponent`. `ContrabandSystem` uses these to clear listed departments and jobs in the examine text using the same logic as it does for `AllowedDepartments` and `AllowedJobs`.

## Media
Before:
Fq8xfvc

After:
JQ9bSO5

## Requirements

## Breaking changes
The BaseCentcommCommandContraband prototype has been removed and uses of it have been replaced with BaseCommandContraband.

**Changelog**
:cl:
- add: Captains will now be cleared to use job and department restricted items when examining them. HoPs will also be cleared to use items restricted to specific service jobs.

### DIFF
--- a/Content.Shared/Contraband/ContrabandSystem.cs
+++ b/Content.Shared/Contraband/ContrabandSystem.cs
@@ -39,6 +39,8 @@ public sealed class ContrabandSystem : EntitySystem
         contraband.Severity = other.Severity;
         contraband.AllowedDepartments = other.AllowedDepartments;
         contraband.AllowedJobs = other.AllowedJobs;
+        contraband.ImplicitlyAllowedDepartments = other.ImplicitlyAllowedDepartments;
+        contraband.ImplicitlyAllowedJobs = other.ImplicitlyAllowedJobs;
         Dirty(uid, contraband);
     }
 
@@ -87,8 +89,11 @@ public sealed class ContrabandSystem : EntitySystem
         // if it is fully restricted, you're department-less, or your department isn't in the allowed list, you cannot carry it. Otherwise, you can.
         var carryingMessage = Loc.GetString("contraband-examine-text-avoid-carrying-around");
         var iconTexture = "/Textures/Interface/VerbIcons/lock-red.svg.192dpi.png";
+        var implicitJobs = component.ImplicitlyAllowedJobs.Select(p => _proto.Index(p).LocalizedName).ToArray();
         if (departments.Intersect(component.AllowedDepartments).Any()
-            || jobs.Contains(jobId))
+            || departments.Intersect(component.ImplicitlyAllowedDepartments).Any()
+            || jobs.Contains(jobId)
+            || implicitJobs.Contains(jobId))
         {
             carryingMessage = Loc.GetString("contraband-examine-text-in-the-clear");
             iconTexture = "/Textures/Interface/VerbIcons/unlock-green.svg.192dpi.png";
--- a/Resources/Prototypes/Entities/Objects/base_contraband.yml
+++ b/Resources/Prototypes/Entities/Objects/base_contraband.yml
@@ -31,6 +31,8 @@
     severity: Major
 
 # base department restricted contraband, this should only be used as a parent for other contraband prototypes, not the restricted items themselves.
+# centcomm, captain, and hop are implicitly cleared for all departmental contraband
+# child prototypes can override this behavior by providing an empty set for these data fields
 - type: entity
   id: BaseRestrictedContraband
   abstract: true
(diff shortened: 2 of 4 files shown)


ArtisticRoomba:
Implicit approval might allow or encourage some things that are frowned upon, ex. the Head of ~Powergaming~ Personnel being cleared to use things like secglasses, a pistol, galoshes, outright or at the start of the round.

DinnerCalzone:
My rationale for including the HoP in the base restricted contraband is that they already have (nearly) all access to departments, but I would be willing to limit this so that they're only implicitly cleared for cross departmental contraband over service jobs, like the Bartender/Security contraband used for the bartender's shotgun.

Djungelskog2:
as an admin, the only reason this even exists in the rules is for the sake of job promotions as far as im aware (I hate it)

UbaserB (dismissed):
The symbol needs padding, it's too close to the edges of the button. Also, the icon is kind of weird I would go for something more like this:

image

DinnerCalzone:
> The symbol needs padding, it's too close to the edges of the button. Also, the icon is kind of weird I would go for something more like this:
> 
> image

I'm not quite sure how this is related to the changes in my PR? These assets were added with #36032.

UbaserB:
LOL this is what happens when i don't double check the pr

Content.Shared/Contraband/ContrabandSystem.cs
```diff
@@ -39,6 +39,8 @@ public void CopyDetails(EntityUid uid, ContrabandComponent other, ContrabandComp
         contraband.Severity = other.Severity;
         contraband.AllowedDepartments = other.AllowedDepartments;
         contraband.AllowedJobs = other.AllowedJobs;
+        contraband.ImplicitlyAllowedDepartments = other.ImplicitlyAllowedDepartments;
+        contraband.ImplicitlyAllowedJobs = other.ImplicitlyAllowedJobs;
         Dirty(uid, contraband);
```
metalgearsloth, line 44: Ignoring whether this pr is needed, this isn't how copying works. This is a shallow copy and mutating one piece of data affects the other because it's only copying the ref.

DinnerCalzone:
> as an admin, the only reason this even exists in the rules is for the sake of job promotions as far as im aware (I hate it) which could be done in pretty much any other way (Having the contraband status update according to your ID ect).

Contraband status does currently update according to your ID; changing the job preset of an ID through the ID card computer changes the job and department associated with the ID.

Djungelskog2:
> > as an admin, the only reason this even exists in the rules is for the sake of job promotions as far as im aware (I hate it) which could be done in pretty much any other way (Having the contraband status update according to your ID ect).
> 
> Contraband status does currently update according to your ID; changing the job preset of an ID through the ID card computer changes the job and department associated with the ID.

huh, dunno why we still have that then

Content.Shared/Contraband/ContrabandSystem.cs
DinnerCalzone: I was mostly just copying the logic that already existed in the system since I'm not _too_ experienced with C#, but it did seem odd to me. The thing of most significance that this method is doing is dirtying the contraband component. Would it make more sense to make deep copies of the data fields or just straight up remove the `contraband` parameter and dirty `other`?

Resources/Prototypes/Entities/Objects/base_contraband.yml
```diff
@@ -39,6 +41,8 @@
     severity: Restricted
     allowedDepartments: [  ]
     allowedJobs: [  ]
+    implicitlyAllowedDepartments: [ CentralCommand ]
+    implicitlyAllowedJobs: [ Captain ]
```
slarticodefast, line 45: the captain should not get a free pass to use all equipment
he doesn't need galoshes etc

PR #36729, opened 2025-04-19
replace obsolete IsGrid calls with HasComp<MapGridComponent>

VerinSenpai:
## About the PR
Replace obsolete IsGrid calls with HasComp<MapGridComponent>

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

PR #36793, opened 2025-04-21
Fix NRE in GasTileOverlaySystem

ElectroJr:
## About the PR
See title. Added in #36729, which was only recently merged and isn't on the test server yet. but this should get merged before next test publish.


VerinSenpai:
wait did I... was I using a variable before I filled it?

ElectroJr:
Its a pretty common mistake, the nullability checks just don't catch it when there's a struct with a default value with reference types.

VerinSenpai:
I'll try to be more careful of that. When I needed it most the linter failed me lmao

PR #36800, opened 2025-04-21
Further Allocation Tweaks

ScarKy0:
## About the PR

Cargo now takes a 50% cut instead of 75% cut of all sales. Lockboxes remain unchanged.

## Why / Balance

Currently departments have no way to make money for themselves, and with the current split they earn close to no money at all. This should lead ot a more interesting split.
50% still makes sure cargo is the main manager for funds, however departments now actually earn enough to be able to order the more expensive things.

Also made the default allocations 20% across the board. I hate that it was different for some reason. This makes it more fair and HoP can still adjust if neccessary.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Default cargo split is now 50% instead of 75%. Lockboxes remain unchanged.

### DIFF
--- a/Content.Server/Cargo/Systems/CargoSystem.Shuttle.cs
+++ b/Content.Server/Cargo/Systems/CargoSystem.Shuttle.cs
@@ -342,8 +342,8 @@ public sealed partial class CargoSystem
             {
                 distribution = new Dictionary<ProtoId<CargoAccountPrototype>, double>
                 {
-                    { sellComponent.OverrideAccount, bankAccount.PrimaryCut },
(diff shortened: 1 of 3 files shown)


Fildrance (approved):
Minor balance change for experiments, while having feature in testing. Having normal distribution would reduce probability of seeing some departments as 'inferior'. Override naming seems ok, as its what comp is called. Even tho it is currently more of _'DeliveryFee'_ it could later be part of some emag interaction or other stuff, so this naming fits better.

ScarKy0:
Fildrance is a maintainer, just has no github perms yet. That being said I count that as 2nd approval and will merge this.

Issue #36501, opened 2025-04-13
StealthOnMove needs removal or refactoring


metalgearsloth:
Content should absolutely not be touching moveevents, especially on server, unless the system is critical which stealth is not.

slarticodefast:
Now that we made mob movement have actual velocity in the `PhysicsComponent` we should be able to simply use that.

PR #36803, opened 2025-04-21
add StationTrackerComponent

MilonPL:
## About the PR

Requires https://github.com/space-wizards/RobustToolbox/pull/5893

adds a StationTrackerComponent
stations are networked now so no reason not to do this, this way client can track what station an entity is on, mostly for purposes of UI prediction so that we can just 
a) update the UI when station changes
b) not store the station on 99 different comps every time like cartridges already do

The method itself is a simplified version of GetOwningStation, although someone needs to check what API can be moved from Server to Shared now.

## Why / Balance

"hmm today i will work on PDA prediction" (clueless)

image

## Requirements

## Breaking changes

`StationGridAddedEvent` and `StationGridRemovedEvent` now have an additional field for the station to which the grid was added. All callers need to be updated accordingly.

**Changelog**


slarticodefast:
shouldn't we completely replace `GetOwningStation` with this then? Or make it use this component?

MilonPL:
They serve different purposes right now, I didn't really want to do more breaking changes right now. Realistically most PublicAPI can be moved to shared now but that's out of scope of this PR.

slarticodefast:
What are the different purposes then? Since you are saying it's "basically a cheaper GetOwningStation"

EmoGarbage404:
this is just `GetOwningStation` with networking. You should just move the original API into shared.

slarticodefast:
Isn't using `MoveEvent` for every single entity extremely expensive?
There is also `EntParentChangedMessage` which should show grid changes, so still don't get why we need this value in a component.
But we can ask @metalgearsloth

Tayrtahn:
Quoting sloth [here](https://github.com/space-wizards/space-station-14/issues/36501):
>Content should absolutely not be touching moveevents, especially on server, unless the system is critical which stealth is not.

Using `EntParentChangedMessage` would probably be better.

dffdff2423 (approved):
Millon yelled at me to review this. I did a read through it and could not find anything obviously wrong to me.

Content.Client/Station/StationSystem.cs
```diff
@@ -5,7 +5,7 @@ namespace Content.Client.Station;
 /// <summary>
 /// This handles letting the client know stations are a thing. Only really used by an admin menu.
 /// </summary>
-public sealed class StationSystem : EntitySystem
+public sealed class StationSystem : SharedStationSystem
```
slarticodefast, line 8:
```suggestion
public sealed partial class StationSystem : SharedStationSystem
```
let's do this while we are changing it anyways
makes things easier for forks

Content.Server/Station/Systems/StationSystem.cs
```diff
@@ -27,7 +27,7 @@ namespace Content.Server.Station.Systems;
 /// For jobs, look at StationJobSystem. For spawning, look at StationSpawningSystem.
 /// </summary>
 [PublicAPI]
-public sealed class StationSystem : EntitySystem
+public sealed class StationSystem : SharedStationSystem
```
slarticodefast, line 30:
```suggestion
public sealed partial class StationSystem : SharedStationSystem
```

Content.Server/Station/Systems/StationSystem.cs
```diff
@@ -353,7 +385,7 @@ public void AddGridToStation(EntityUid station, EntityUid mapGrid, MapGridCompon
         stationMember.Station = station;
         stationData.Grids.Add(mapGrid);
 
-        RaiseLocalEvent(station, new StationGridAddedEvent(mapGrid, false), true);
+        RaiseLocalEvent(station, new StationGridAddedEvent(mapGrid, station, false), true);
```
slarticodefast, line 388: I would the `station` uid to `StationGridRemovedEvent` as well to keep the two events identical

Content.Shared/Station/SharedStationSystem.cs
```diff
@@ -0,0 +1,86 @@
+using Content.Shared.Station.Components;
+using Robust.Shared.Map;
+
+namespace Content.Shared.Station;
+
+public abstract class SharedStationSystem : EntitySystem
```
slarticodefast, line 6:
```suggestion
public abstract partial class SharedStationSystem : EntitySystem
```

slarticodefast (changes requested):
Found a bug:
- Spawn an item
- Add `StationTrackerComponent` to that item
- Put the item into your pocket
- Walk off the grid into space
- The tracker won't update because the parent is the mob, and the tracked item itself does not change its parent

slarticodefast (approved):
Looks good. Self-merge after the new engine release.

PR #36806, opened 2025-04-21
fix some datasets

lzk228:
while these string were in DatasetPrototype, they could be with quotation marks or without it didn't matter, they would appear the same
but they were ported in ftl with those strings, so they shouldn't be here as it was supposed to be


slarticodefast (approved):
Thanks!

PR #36811, opened 2025-04-21
dragons can now pry doors

muburu:
## About the PR
gives dragons the ability to pry doors, with a speed modifier of 2.5 (tarantula prying speed is 1.5 for reference)

## Why / Balance
dragons have giant hands bigger than your face, makes sense that they could pry an airlock open pretty easily.
proper reason is airlocks are super annoying if you're a dragon, being able to pry them open quickly in the heat of battle lets them be more aggressive.
they can pry open bolted doors, but at that point they should just be eating it instead because it takes a bajillion years.

they're still encouraged to devour doors so their carp minions can follow them.

## Technical details
adds prying component to dragon

## Media
**powered door**
dragondoor
**airlock** (unpowered door is same speed)
dragonfirelock
the "extra click" is just me focusing the window because it was focused on the gif recording software beforehand
[this PR doesn't mess with your inputs don't worry]

## Requirements

**Changelog**
:cl:
- tweak: Dragons can now pry open doors, including airlocks.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/dragon.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/dragon.yml
@@ -159,6 +159,12 @@
     - NamesDragon
     - NamesDragonTitle
     nameFormat: name-format-dragon
+  - type: Prying
+    pryPowered: true
+    force: true
+    speedModifier: 2.5 # fast because dragon strong
+    useSound:
+      path: /Audio/Items/crowbar.ogg
 
 - type: entity
   parent: BaseMobDragon


beck-thompson (approved):
MAKES SENSE TO ME

ArtisticRoomba (approved):
I agree. Waiting for a bit to merge if anyone has complaints, thought I think this is just something the dragon should be able to do. We can see if it causes problems on Vulture, worst case.

beck-thompson:
> I agree. Waiting for a bit to merge if anyone has complaints, thought I think this is just something the dragon should be able to do. We can see if it causes problems on Vulture, worst case.

They can already the doors already as well. This is just a little faster!

superjj18:
> > I agree. Waiting for a bit to merge if anyone has complaints, thought I think this is just something the dragon should be able to do. We can see if it causes problems on Vulture, worst case.
> 
> 
> 
> They can already the doors already as well. This is just a little faster!

With the added benefit of not making an obvious hole in the side of the station that anyone, ESPECIALLY those in bridge, can see from a map.

PR #36812, opened 2025-04-21
Fix deliveries disappearing when opened in a container

ljm862:
## About the PR

I saw a discord message about items not spawning properly when you open a delivery inside your bag. So I fixed it!
Not sure if there's a github issue for this one

## Why / Balance

Items were disappearing, it's a bug and confusing for players. Also, lost mail :(

## Technical details

`EmptyContainer` -> `Remove` doesn't actually reach the "Set Parent" bit of code if there's a coordinate destination to empty the items to. So we don't set the coordinates if the delivery is in a container so it just drops it in the container instead.

## Media
Not shown in the video, but if the container is full and you open it, the delivery drops onto the floor under the container (you if your bag)

https://github.com/user-attachments/assets/75ae32aa-e476-4bad-9990-790f718b6954

## Requirements

**Changelog**

🆑 fix: Deliveries opened in a container now drop the item in the container.

### DIFF
--- a/Content.Shared/Delivery/SharedDeliverySystem.cs
+++ b/Content.Shared/Delivery/SharedDeliverySystem.cs
@@ -180,7 +180,7 @@ public abstract class SharedDeliverySystem : EntitySystem
         if (!_container.TryGetContainer(ent, ent.Comp.Container, out var container))
             return;
 
-        if (attemptPickup)
+        if (attemptPickup || _container.IsEntityInContainer(ent))
         {
             foreach (var entity in container.ContainedEntities.ToArray())
             {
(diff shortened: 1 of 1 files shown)


PJB3005 (commented):
This feels like there should be a proper helper function for this instead?

ljm862:
> This feels like there should be a proper helper function for this instead?

Some sort of `EmptyContainerIntoParentContainer` ? Yeah I think it probably makes sense to do that.
Though I also think the `SharedContainerSystem.Remove` function could do with a small refactor too so bugs like this would be harder to make. 

If I get time tomorrow I'll quickly set up a helper in the SharedSystem and I'll have a crack at a deeper refactor in about 2 weeks when I'm back home!

PR #36813, opened 2025-04-21
Remove non-functional ninja exemption from `SpawnAndDeleteEntityCountTest`

Tayrtahn:
## About the PR

Removes a chunk of code that tried to make the space ninja entity exempt from `SpawnAndDeleteEntityCountTest` (but actually did nothing).

## Why / Balance

It does nothing, and it has a TODO comment saying to fix it.

## Technical details

The deleted code checks each ProtoId in the loop to see if it matches "MobHumanSpaceNinja" so it can be skipped by the test.

But there is no entity prototype with the ID "MobHumanSpaceNinja", so it does nothing.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/EntityTest.cs
+++ b/Content.IntegrationTests/Tests/EntityTest.cs
@@ -277,11 +277,6 @@ namespace Content.IntegrationTests.Tests
             {
                 foreach (var protoId in protoIds)
                 {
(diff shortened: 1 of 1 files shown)


slarticodefast (approved):
I was wondering what that was for before 😄 
Good to have it cleaned up!

Issue #36782, opened 2025-04-21
Station Event selection logic needs complete replacement


EmoGarbage404:
## Description

The current system is an unfortunate mix of legacy code and modern re implementation that clashes heavily:
- The event entity tables (despite theoretically being perfect for defining probabilities) don't actually contain the weights and probabilities.
- The actual selection logic is basically entirely custom and works on reading component info off of prototypes (very gross)
- Despite having all the events in one place, specific info for the events is duplicated and copy pasted in the event prototypes, which is very confusing.

How would you fix this:
- The event table for station events should be the entire selection logic. It should literally call GetSpawns and then just run shit.
- Weights should be moved off of individual gamerule prototypes and into the table
- EntityTables need support for conditional selection, allowing for gamerule conditions to be stored there.
- - This can either be specific to station events or something generic (use it for holidays, etc.)

IProduceWidgets:
yea, I think youre right that most of stationeventcomp can probably be yeeted into selectors
might not even be that bad?
will depend how much stuff isnt accessible from selectors :godo~1:
I think the event stuff mostly has its own manager anyway though, so probably not too unpleasant

IProduceWidgets:
To bikeshed us a bit,
Would also be very nice to have a selector that understands when a storage is full and stops trying to add shit.
Causes heisen tests and whatnot when someone maps a drawer with too many folders in it or something. 
(definitely wasnt me)

IProduceWidgets:
linko patronus! #36819

TriviaSolari:
Unless this is already being actively worked on, I'd like to start working towards getting this issue done, and had a question: most of the fields in `StationEventComponent` (besides I think just the evac check) exist in some form within `EntityTableSelector` or as various `EntityTableConditions`; I'm assuming a refactor here would involve deleting those fields in StationEventComp and updating YAML where needed? Or should that be done as two separate PRs, one to change over the YAML and another to deprecate/delete the fields afterwards?

IProduceWidgets:
> I'd like to start working towards getting this issue done

I think the currently missing entitytable conditions are 

- ReoccurenceDelay
- OccursDuringRoundEnd

But the the bigger lift is there needs to be a way to simulate the gamerule selection like we have in the BasicStationEventScheduler.cs 
via the simulate command.

[...]

PR #36819, opened 2025-04-22
Add condition support to entity tables

EmoGarbage404:
## About the PR

Adds the ability to define conditions for entity tables. Will be useful in the server for moving more selection logic stuff into tables rather than having it as some kind of bespoke thing.

Tables aren't tied to entities so this is really just a global condition system. Still a lot of utility: just defining its bounds.

Added a player count condition with an example to show the potential.

## Why / Balance

n/a

## Technical details

n/a

## Requirements

## Breaking changes

**Changelog**

no cl no fun

### DIFF
--- a/Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
+++ b/Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
@@ -1,0 +1,31 @@
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.EntityTable.Conditions;
+
+public sealed partial class PlayerCountCondition : TableCondition
+{
+    /// <summary>
+    /// Minimum players of needed for this condition to succeed. Inclusive.
+    /// </summary>
+    [DataField]
+    public int Min = int.MinValue;
+
+    /// <summary>
+    /// Maximum numbers of players there can be for this condition to succeed. Inclusive.
+    /// </summary>
+    [DataField]
+    public int Max = int.MaxValue;
+
+    private static ISharedPlayerManager? _playerManager;
+
+    public override bool EvaluateImplementation(IEntityManager entMan, IPrototypeManager proto)
+    {
+        // Don't resolve this repeatedly
+        _playerManager ??= IoCManager.Resolve<ISharedPlayerManager>();
+
+        var playerCount = _playerManager.PlayerCount;
+
+        return playerCount >= Min && playerCount <= Max;
+    }
+}
--- a/Content.Shared/EntityTable/EntitySelectors/EntityTableSelector.cs
+++ b/Content.Shared/EntityTable/EntitySelectors/EntityTableSelector.cs
@@ -1,3 +1,4 @@
+using Content.Shared.EntityTable.Conditions;
 using Content.Shared.EntityTable.ValueSelector;
 using JetBrains.Annotations;
 using Robust.Shared.Prototypes;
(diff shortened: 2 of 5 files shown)


IProduceWidgets:
linko patronus! #36782

IProduceWidgets (approved):
looks good to me.

Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
```diff
@@ -0,0 +1,31 @@
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.EntityTable.Conditions;
+
+public sealed partial class PlayerCountCondition : TableCondition
```
Fildrance, line 6: part of yaml tree, could use xml-doc in case we will have some tool that will reveal xml-doc on picking class while yaml-coding... why dont we have it already? who knows! :(

Fildrance (changes requested):
Mostly stylistic and minor stuff :3 also a small pile of thoughts

Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
```diff
@@ -0,0 +1,31 @@
...
+    [DataField]
+    public int Min = int.MinValue;
+
+    /// <summary>
+    /// Maximum numbers of players there can be for this condition to succeed. Inclusive.
+    /// </summary>
+    [DataField]
+    public int Max = int.MaxValue;
+
+    private static ISharedPlayerManager? _playerManager;
```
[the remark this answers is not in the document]
EmoGarbage404: Being static is good since it prevents having to resolve this per condition evaluation. Otherwise, we may be doing 30 or 40 resolves.

Content.Shared/EntityTable/EntitySelectors/EntityTableSelector.cs
```diff
@@ -26,10 +27,26 @@ public abstract partial class EntityTableSelector
...
+    /// </summary>
+    [DataField]
+    public List<TableCondition> Conditions = new();
+
+    /// <summary>
+    /// If true, all the conditions must be successful in order for the selector to process.
+    /// Otherwise, only one of them must be.
+    /// </summary>
+    [DataField]
+    public bool RequireAll = true;
```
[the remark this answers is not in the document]
EmoGarbage404: i'm not opposed to meta conditions but i think for most usecases this is pretty much sufficient.

Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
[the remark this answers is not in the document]
Fildrance: Ah, this condition might be included in many other prototypes, forgot about that! Fair!

Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
```diff
@@ -0,0 +1,34 @@
...
+/// <summary>
+/// Condition that passes only if the server player count is within a certain range.
+/// </summary>
+public sealed partial class PlayerCountCondition : EntityTableCondition
+{
+    /// <summary>
+    /// Minimum players of needed for this condition to succeed. Inclusive.
+    /// </summary>
+    [DataField]
+    public int Min = int.MinValue;
```
Fildrance, line 15: are we expecting negative player count? :D

Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
```diff
@@ -0,0 +1,34 @@
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.EntityTable.Conditions;
+
+/// <summary>
+/// Condition that passes only if the server player count is within a certain range.
+/// </summary>
+public sealed partial class PlayerCountCondition : EntityTableCondition
```
Fildrance, line 9: not sure if thats critical, but maybe better rename derived type too...

image

Fildrance: when considering naming, using Rider-style search  (fuzzy search that applies extra attention for characters in upper case) should be considered - so if we enter ETCond it kinda should show us all things with EntityTableCondition first.

Fildrance (changes requested):
just a tiny bit more :< just a little tiny-tiny bit.

Fildrance:
Also, i found out that our spawn item panel calls all of spawn logic, including such conditions. Not in scope of this PR but it is kinda sad...

Content.Shared/EntityTable/Conditions/PlayerCountCondition.cs
EmoGarbage404: i don't really think this is an issue. This is primarily for YAML usage so brevity is good to prioritize over specificity. These conditions are inevitably gonna overlap in name unless there is a true generic system for it.

Fildrance: 👀  well that will become problem when (and IF) we will get some good tool with smart intellisense on yaml, which is not coming any time soon...

Fildrance (approved):
fine

PR #36821, opened 2025-04-22
Make research clients grid-local

EmoGarbage404:
## About the PR

Prevents research clients from being able to access servers on different grids

## Why / Balance

Currently, RD servers have global range. This means that basically anyone, including entities that don't interact with or aren't positively affiliated with the station can access any tech that is researched.

Forcing people to at least be on the same grid ensures that at least some level of interaction is required for them to make good use of the science system. This prevents things like antags accessing science's unlocks and off-station salvagers being able to set up factories in shuttles and such.

## Technical details

changes research code to use a shared method for accessing servers.

## Requirements

## Breaking changes

`GetServerIds`, `GetServerNames`, and `TryGetServerById` in Server.ResearchSystem all now have an additional parameter for the research client.
 
**Changelog**
:cl:
- tweak: Lathes can no longer connect to research servers on separate grids

### DIFF
--- a/Content.Server/Research/Systems/ResearchSystem.cs
+++ b/Content.Server/Research/Systems/ResearchSystem.cs
@@ -34,19 +34,20 @@ namespace Content.Server.Research.Systems
         }
 
         /// <summary>
-        /// Gets a server based on it's unique numeric id.
+        /// Gets a server based on its unique numeric id.
         /// </summary>
+        /// <param name="client"></param>
         /// <param name="id"></param>
         /// <param name="serverUid"></param>
         /// <param name="serverComponent"></param>
         /// <returns></returns>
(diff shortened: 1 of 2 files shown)


Content.Server/Research/Systems/ResearchSystem.cs
```diff
@@ -78,19 +79,32 @@ public string[] GetServerNames()
...
+            var query = AllEntityQuery<ResearchServerComponent, TransformComponent>();
+            while (query.MoveNext(out var uid, out var comp, out var xform))
+            {
+                if (xform.GridUid != clientXform.GridUid)
+                    continue;
+
+                yield return (uid, comp);
+            }
+        }
+
```
metalgearsloth, line 107: Don't we have GetEntitiesInGrid for dis.

EmoGarbage404: :mistake:

PR #36823, opened 2025-04-22
Advanced NT anti-syndicate technology

EmoGarbage404:
## About the PR

Common syndicate ID cards will now be melted inside of microwaves. this includes:
- agent ID
- syndicate ids (normal and nukie)
- clown ids from clown bundle

## Why / Balance

Syndies and nukies constantly exploit microwave gambling to try and get access for free. Making it actually require stealing IDs at least presupposes some minor amount of brainpower, instead of being completely free.

## Technical details

checks microwave tags _before_ the event handler so that they can cancel it.

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Superior Nanotrasen-made microwaves will now instantly melt puny Syndicate ID cards.

### DIFF
--- a/Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
+++ b/Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
@@ -543,16 +543,6 @@ namespace Content.Server.Kitchen.EntitySystems
             // TODO use lists of Reagent quantities instead of reagent prototype ids.
             foreach (var item in component.Storage.ContainedEntities.ToArray())
             {
-                // special behavior when being microwaved ;)
-                var ev = new BeingMicrowavedEvent(uid, user);
-                RaiseLocalEvent(item, ev);
-
-                if (ev.Handled)
-                {
-                    UpdateUserInterfaceState(uid, component);
-                    return;
-                }
-
                 if (_tag.HasTag(item, MetalTag))
                 {
                     malfunctioning = true;
@@ -564,6 +554,19 @@ namespace Content.Server.Kitchen.EntitySystems
                     _container.Insert(junk, component.Storage);
                     Del(item);
                     continue;
+                }
+
+                if (!malfunctioning)
+                {
+                    // special behavior when being microwaved ;)
+                    var ev = new BeingMicrowavedEvent(uid, user);
+                    RaiseLocalEvent(item, ev);
+
+                    if (ev.Handled)
+                    {
+                        UpdateUserInterfaceState(uid, component);
+                        return;
+                    }
                 }
 
                 var microwavedComp = AddComp<ActivelyMicrowavedComponent>(item);
--- a/Resources/Prototypes/Catalog/Fills/Backpacks/duffelbag.yml
+++ b/Resources/Prototypes/Catalog/Fills/Backpacks/duffelbag.yml
@@ -140,7 +140,7 @@
       - id: ClothingShoesClown
       - id: ClothingMaskClown
       - id: BikeHorn
-      - id: ClownPDA
+      - id: ClownPDAPlastic # no free ID for you
       - id: ClothingHeadsetService
 
 - type: entity
--- a/Resources/Prototypes/Entities/Objects/Devices/pda.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/pda.yml
@@ -368,6 +368,14 @@
 
 - type: entity
   parent: ClownPDA
+  id: ClownPDAPlastic
+  suffix: Plastic Card
+  components:
+  - type: Pda
+    id: ClownIDCardPlastic
+
+- type: entity
+  parent: ClownPDA
   id: VisitorClownPDA
   suffix: Visitor
   components:
--- a/Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/identification_cards.yml
@@ -257,6 +257,16 @@
     job: Clown
 
 - type: entity
+  parent: ClownIDCard
+  id: ClownIDCardPlastic
+  suffix: Plastic
+  components:
+  - type: Tag
+    tags: #override chameleon whitelist
+    - DoorBumpOpener
+    - Plastic
+
+- type: entity
   parent: IDCardStandard
   id: MimeIDCard
   name: mime ID card
@@ -601,6 +611,7 @@
   - type: Tag
     tags:
     - DoorBumpOpener
+    - Plastic
   - type: ChameleonClothing
     slot: [idcard]
     default: PassengerIDCard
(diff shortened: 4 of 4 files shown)


EmoGarbage404:
> Wait doesn't the Syndicate ship have Donk Co microwaves

canon lore dump: syndicate microwaves are actually bought wholesale from NT and rebranded.

SlimmSlamm:
The Agent ID and Clown Syndicate ID are stealth items though? This doesn't make much sense that NT would know about them and have precautions to deal with them?

TytosB:
this seems entirely unnecessary, the issue of nukies getting free was fixed by replacing their microwaves with donk co. now the only way they can do the microwave aa trick is if they steal a normal microwave board/machine from the station, which is balanced due to it being time consuming and risking them getting caught.

RedBookcase:
> Making it actually require stealing IDs at least presupposes some minor amount of brainpower, instead of being completely free.

To be clear this does entirely remove the value of ID gambling for nukies. If they have to steal IDs, they're already in a position to be copying those ID's accesses with agent IDs, which nukies almost always buy anyways. And since they're already stealing from and/or killing crew, they're already taking a huge amount of risk to get those IDs.

Hrosts:
Do nukies actually do that now? Seems like it's faster and easier to just kill a crewmember and copy the ID?

Minemoder5000:
> This implementation will, at best, needlessly punish new antag players who aren't aware of this mechanic, and at worst be used to check if ids are valid.

Id cards can already melt in the microwave, though.

> Do nukies actually do that now? Seems like it's faster and easier to just kill a crewmember and copy the ID?

Yes, they do. It's happened at least twice on Salamander alone.

Djungelskog2:
The clown id's seem minorly meta-able and it needlessly minorly nerfs normal syndi gameplay in an area where it wasnt an issue.

My only real issue here is with the clown ids

Djungelskog2:
> The clown id's seem minorly meta-able and it needlessly minorly nerfs normal syndi gameplay in an area where it wasnt an issue.
> 
> My only real issue here is with the clown ids

actually im still not really sure how to feel abt it since it feels kinda whacky if not mentioned anywhere and its just for this super specific scenario but whatever

walksanatora:
my main thing is... I dont put my agent ID in the microwave? I put in my true ID and copy from that. which this PR doesn't fix? all it does is add a way of checking agent IDs... which you could allready check by just... holding them and checking the verbs?

Pronana:
> my main thing is... I dont put my agent ID in the microwave? I put in my true ID and copy from that. which this PR doesn't fix? all it does is add a way of checking agent IDs... which you could allready check by just... holding them and checking the verbs?

I'm pretty sure this PR isn't meant to prevent that. It's meant to prevent nukies from jumping onto a salv debris with a microwave and buying IDs until they have AA. This doesn't really affect most gameplay just prevents one cheese strat.

walksanatora:
and if they are spending all their TC on AA. wanna know what they aint spending it on? weapons. meaning they are gonna have to be VERY precise with what they do. and also be mildly underpowered compared to normal nukie squads.

walksanatora:
(my stupid spur of the moment idea) if you wanna prevent basically "AA rolling" make it so the clown duffel/agent ID can only be bought once per nukie. no more burning it all on id cards.

Pumkin69:
I never put my agent id into the microwaves to gamble. If you are gonna make this PR i would only make it the clown id. If nukies are spending 3 TC on an agent id card to gamble for random accesses that also has a chance to blow up they should just be allowed to do that imo.

TytosB:
> access breaker costs 4 tc, gives you AA, and requires substantially LESS setup
> 
> agent ID requires 3 tc and a solid amount of time for GAMBLING on what access you get

i think this really should be the final nail in the coffin for this pr. please just close it

Minemoder5000:
The significant amount of time is the issue this is solving. Nukies figured out that the vgroid can have microwaves, and spend an hour there trying to gamble instead of nuking the station in a timely fashion.

TytosB:
> The significant amount of time is the issue this is solving. Nukies figured out that the vgroid can have microwaves, and sometimes spend an hour there trying to gamble instead of nuking the station in a timely fashion.

i doubt this has happened more than a handful of times, but even if it doesn't, i fail to see the issue. if the nukies want to waste their time getting a microwave from vgroid then thats fine? the only consequence i can see is that the crew has more time to have a normal shift before the nukies show up and start killing everyone.

Minemoder5000:
AA gives the nukies a significant advantage, which is why the donk co microwave was added to the nukie outpost. And we don't want the nukies spending forever somewhere else before engaging the station because they're meant to be the main driving force behind the round. Just because it's only happened a few times before doesn't mean it won't happen more in the future.

memengineer25:
> The significant amount of time is the issue this is solving. Nukies figured out that the vgroid can have microwaves, and sometimes spend an hour there trying to gamble instead of nuking the station in a timely fashion.

That particular strat only has the microwaves as an _incidental_ target, and _additional_ advantage from going to vgroid, not the primary objective. The real target with vgroid is the lathes and circuit imprinters, for making unlimited/specialist ammo and potentially even things like shuttle weapons.

Minemoder5000:
There is another PR that alleviates the issues with the lathes on the vgroid. Though as for microwave gambling it is exceptionally useful to be able to follow the captain through any airlock, before you kill them for AA.

walksanatora:
if you are doing your job you should allready have maints and other basic accesses from killing some people or having htem offer their IDs

Minemoder5000:
> if you are doing your job you should allready have maints and other basic accesses from killing some people or having htem offer their IDs

And this PR helps to encourage that, instead of trying to gamble every access.

memengineer25:
> There is another PR that alleviates the issues with the lathes on the vgroid. Though as for microwave gambling it is exceptionally useful to be able to follow the captain through any airlock, before you kill them for AA.

If vgroid lathes are removed or "alleviated" as you say, it will cease to be worth the cost in time and resources to go onto vgroid and nukies using the microwaves will stop. Gambling for (in reality) maybe half or three-quarters access is NOT reliable, unlike the access breaker, and is an entirely insufficient incentive for taking the time and resources to hit vgroid.

IProduceWidgets:
I don't think people microwaving the IDs to antag check is a problem.

I *do* think hidden punish mechanic is a problem.

I think this would be better solved by a punish mechanic for nukies who are on station map for too long without doing anything. Something like the IFF cloak having some resource or time based limitation.

EmoGarbage404:
After some reflection:

I think ID microwaving as a general mechanic is ass for a lot of reasons, and it doesn't necessarily need to be nukie-related to be a problem. I'd rather approach this more generally than special-casing a bunch of situations, as i think the issue is really with the system itself rather than any individual user of it.

PR #36833, opened 2025-04-22
Adds a new traitor item to the emagged coffee vending machine.

Starlight991:
## About the PR

Emagging a coffee vending machine will give out the option to dispense molten chocolate. Molten chocolate is a powerful item that heavily burns and increases the temperature of anyone it gets splashed on.
## Why / Balance

Would be a difficult but rewarding single target weapon. It only has a chance to be in stock in the emagged machine and the machines are usually in public areas therefor making it risky. also you only get 20u which is just under enough to kill someone. 
## Technical details

its all yaml from me, a yaml warrior
## Media

https://github.com/user-attachments/assets/5c88603b-7328-4558-9207-96f99d31f5a9

## Requirements

## Breaking changes

n/a
**Changelog**

:cl:
- add: When emagged, the Solar's Hot Drinks machine gains a secret new drink

### DIFF
--- a/Resources/Locale/en-US/reagents/generic.ftl
+++ b/Resources/Locale/en-US/reagents/generic.ftl
@@ -9,3 +9,4 @@ generic-reagent-effect-parched = You feel parched.
 generic-reagent-effect-thirsty = You feel thirsty.
 generic-reagent-effect-sick = You feel sick after consuming that...
 generic-reagent-effect-slicing-insides = You feel an incredibly sharp pain in your gut!
+generic-reagent-effect-melting-insides = Your insides feel like they are melting!
--- a/Resources/Locale/en-US/reagents/meta/consumable/drink/drinks.ftl
+++ b/Resources/Locale/en-US/reagents/meta/consumable/drink/drinks.ftl
@@ -93,3 +93,6 @@ reagent-desc-posca = Poor warriors' drink from a forgotten era.
 
 reagent-name-mopwata = mopwata
 reagent-desc-mopwata = Dirty, stagnant mop water.
+
+reagent-name-molten-chocolate = molten chocolate
+reagent-desc-molten-chocolate = A cup of hot chocolate mixed with a unknown chemical that has been superheated through unknown means.
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/coffee.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/coffee.yml
@@ -10,3 +10,4 @@
     DrinkTeapot: 2
   emaggedInventory:
     DrinkNothing: 2
+    DrinkMoltenChocolate: 1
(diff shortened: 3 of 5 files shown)


Resources/Locale/en-US/reagents/generic.ftl
```diff
@@ -9,3 +9,4 @@ generic-reagent-effect-parched = You feel parched.
 generic-reagent-effect-thirsty = You feel thirsty.
 generic-reagent-effect-sick = You feel sick after consuming that...
 generic-reagent-effect-slicing-insides = You feel an incredibly sharp pain in your gut!
+generic-reagent-effect-melting-insides = Your insides feel like they are melting!
```
beck-thompson, line 12: you need a newline at the end of the file

Resources/Locale/en-US/reagents/meta/consumable/drink/drinks.ftl
```diff
@@ -93,3 +93,6 @@ reagent-desc-posca = Poor warriors' drink from a forgotten era.
 
 reagent-name-mopwata = mopwata
 reagent-desc-mopwata = Dirty, stagnant mop water.
+
+reagent-name-molten-chocolate = molten chocolate
+reagent-desc-molten-chocolate = A cup of hot chocolate mixed with a unknown chemical that has been superheated through unknown means.
```
beck-thompson, line 98: same here

beck-thompson (changes requested):
Currently emag interactions are frozen! Could you make this obtainable in another way?

Entvari:
Damage is **way** too high.
Allows you to near instantly crit any target with ease at a measly TC cost

Entvari:
> pen bomb
Not starting a balance argument, but pen bomb costs nearly the same as an Emag, is completely onetime use, is far less effective against armored targets, requires you to somewhat position, doesn't come with a ton of other stuff, etc, etc.
40 heat per swing + extra DoT is actually unreal, especially considering the fact that even if you miss you can just swing again with no penalty. 
This would be a lot more acceptable if it were a onetime purchase but it'd be pretty easy to get multiple of these as most stations have several coffee vendors, this paired with the fact you ALSO get all the benefits of the Emag at the same cost is too much.

Starlight991:
Updated to no longer be obtainable from an emagged vending machine. Now craftable with chemistry using hot chocolate, vestine, plasma and heating. Fixed the newline issues as well.

Geoengi:
Now only if we could make putty copies of sec's keycards...

deltanedas:
and use duct tape to cuff people

beck-thompson (approved):
I'm unsure about balance and stuff but it looks fine!

slarticodefast:
Hey, we had a maintainer poll and it was decided not to merge this PR.
The current reagent system should be improved so that temperature of reagents actually matters and lets hot reagents deal damage instead of having a "hot" variant of an existing reagent.

Issue #36223, opened 2025-03-31
Doafters are not canceled even when the targeted items is removed from a container. (Stethoscope)


beck-thompson:
## Description

**Reproduction**

1.) Equip a stethoscope
2.) Start the doafter event
3.) Unequipped the stethoscope

The event will continue!

**Screenshots**

https://github.com/user-attachments/assets/e35385d0-d87a-4b69-9178-bfc7275b38e7

**Additional context**

The doafter system doesn't check for if the containers changes. This doesn't effect most items as a lot of items require it being held as an additional requirement, but if an item doesn't you could technically put it anywhere and the doafter event would still work! A fix would probably look somewhat similar to how `NeedHand` (or any of the other checks) works.

See: https://github.com/space-wizards/space-station-14/pull/36210#discussion_r2020275191 for some more context

minism:
I took a stab at this in #36838 but I'm not sure if this approach is correct

PR #36838, opened 2025-04-22
Implement BreakOnContainerChange for DoAfter

minism:
## About the PR
Add `BreakOnContainerChange` option to DoAfter, application is for stethoscope to be interrupted when unequipping it

## Why / Balance
See #36223

## Technical details
DoAfter checks the container ID on start and on update to compare. 

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Fix a bug where stethoscope wasn't interrupted on unequip.
-->

### DIFF
--- a/Content.Shared/DoAfter/SharedDoAfterSystem.Update.cs
+++ b/Content.Shared/DoAfter/SharedDoAfterSystem.Update.cs
@@ -3,6 +3,7 @@ using Content.Shared.Hands.Components;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Interaction;
 using Content.Shared.Physics;
+using Robust.Shared.Containers;
 using Robust.Shared.Utility;
 
 namespace Content.Shared.DoAfter;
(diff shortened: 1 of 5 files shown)


Content.Shared/DoAfter/SharedDoAfterSystem.Update.cs
```diff
@@ -237,6 +238,15 @@ private bool ShouldCancel(DoAfter doAfter,
                 return true;
         }
 
+        if (args.BreakOnContainerChange && args.EventTarget != null)
+        {
+            _container.TryGetContainingContainer((args.EventTarget.Value, null, null), out var container);
+            if (container == null || container.ID != doAfter.InitialContainerID)
```
beck-thompson, line 244: I think you shouldn't cancel if the container == null! What if it starts out not inside a container? It could work if you first check if the InitialContainerId isn't null

beck-thompson (changes requested):
So here is what worries me mostly: What if you switch containers to a new container with the same ID but is actually unique? E.g I take off the stethoscope and put it on someone elses neck.

I don't know the answer, so if you could give me your resonating for this approach that would be great!

minism:
> So here is what worries me mostly: What if you switch containers to a new container with the same ID but is actually unique? E.g I take off the stethoscope and put it on someone elses neck.
> 
> I don't know the answer, so if you could give me your resonating for this approach that would be great!

Oh, that's a good point.  Perhaps it would be better to use the entity ID of the container?  I'm new to the codebase

beck-thompson:
> Oh, that's a good point. Perhaps it would be better to use the entity ID of the container? I'm new to the codebase

Yes that's probably the best idea! The issue is that it's kind of complex because of networking stuff. Look how the the other fields do it it shouldn't be *too* hard. If you get stuck let me know :saluting_face:

iaada:
This PR has been awaiting changes for 3 months. Are you still working on this, or do you need help?

TheShuEd:
Considering that the PR has been in “awaiting changes” status for more than 2 months without any changes, and week withous answering on iaada message, I am closing this PR as abandoned. If you want to continue working on it, please let me know.

PR #36841, opened 2025-04-22
Lower interdyne herbals TC cost

ScarKy0:
## About the PR

Lowered the cost of interdynes from 2TC to 1TC

## Why / Balance

Currently interdyne cigarettes are almost pure flavor. The passive regeneration they provide from omnizine is close to nothing unless the damage is perfectly split (in most cases, it isn't).
This makes them almost purely useless outside of few niche scenarios (or if you're willing to wait around 12min to heal to full health, while writing this PR description i healed about 8 pierce as the only damage type).

With that in mind, I think they are perfectly fine to be a 1TC purchase.

## Technical details

All yaml.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Interdyne cigs now cost 1TC instead of 2TC.

### DIFF
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -684,11 +684,8 @@
   name: uplink-cigarettes-name
   description: uplink-cigarettes-desc
   productEntity: CigPackSyndicate
-  discountCategory: rareDiscounts
-  discountDownTo:
-    Telecrystal: 1
   cost:
-    Telecrystal: 2
+    Telecrystal: 1
   categories:
   - UplinkChemicals
 


Cadunkus:
Just move them to the "useless" section so newbies don't get bamboozled into buying them.

ScarKy0:
If someone makes healing not split to non-existing damage maybe they'll be useful

ArtisticRoomba:
Omnizine would be much more worth it across the board if its healing potential was spread equally amongst the damages currently present.

Presently it does what ScugKy says it does which makes stuff like combat medipens only actually effective if you've just met a laser user, a ballistic user, a sword user, and an X-Ray cannon user. Otherwise it's kind-of shit.

ArtisticRoomba (approved):
I agree with the changes given the argument, and no other maintainer has expressly disagreed with this change.

In the future, Omnizine should have its damages split to cover health types in terms of "filling" them instead of splitting them, but that's out of scope.

PR #36845, opened 2025-04-23
Fix warning CS9107

Tayrtahn:
## About the PR

Fixes an instance of warning CS9107 ("Parameter is captured into the state of the enclosing type and its value is also passed to the base constructor. The value might be captured by the base class as well") in `AnalysisConsoleBoundUserInterface.cs`.

## Why / Balance

## Technical details

Capitalized an o.

Actual explanation: using `owner` from the primary constructor in `AnalysisConsoleBoundUserInterface.Open` causes this warning, since it is both being used locally and being passed to the base constructor. The base constructor (in `BoundUserInterface`) captures the value in the `Owner` field, so we can just use that instead.

## Requirements


VerinSenpai (approved):
My names Verin, and I approve this cleanup :smiling_imp:

slarticodefast (approved):
Works!

PR #36846, opened 2025-04-23
Add 1 dragon name

DogZeroX:
## About the PR

Adds one dragon name

## Why / Balance

Funny calamity reference

**Changelog**
No :cl: no fun

### DIFF
--- a/Resources/Locale/en-US/datasets/names/dragon.ftl
+++ b/Resources/Locale/en-US/datasets/names/dragon.ftl
@@ -28,3 +28,4 @@ names-dragon-dataset-27 = Toastarinus
 names-dragon-dataset-28 = Embergill
 names-dragon-dataset-29 = Doomwing
 names-dragon-dataset-30 = Celesteus
+names-dragon-dataset-31 = Jharon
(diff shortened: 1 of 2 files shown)


slarticodefast:
You will need to increase the counter on the localized dataset by one or this won't be used.

DogZeroX:
this is why i get for being lazy and not reading the pr

slarticodefast:
@Tayrtahn I got a request for another integration test: Somehow check if there is a loc string for the dataset count + 1 because that would catch this case.

Tayrtahn:
> @Tayrtahn I got a request for another integration test: Somehow check if there is a loc string for the dataset count + 1 because that would catch this case.

I like it. Should be easy enough to add that to the existing test.

YoungThugSS14:
THE ENEMY ASCENDED BEYOND YOUR CONTROL

NoElkaTheGod:
OR WAS THAT ALL YOUR INTENTION?

slarticodefast (approved):
Seems fine!

Issue #36827, opened 2025-04-22
Jetpacks can't be thrown far


PotentiallyTom:
## Description
Trying to throw a jetpack causes it to drop to your feet, instead of going where you threw it.

**Reproduction**
1. Grab a jetpack
2. Throw it
3. Drop it on your feet instead

**Additional context**

* tested with all types of jetpack, full and empty, and blood-red magboots, all had the same issue.

* it looks like they do get a very small amount of momentum when thrown, just not enough to go anywhere.

* removing the `CanMoveInAirComponent` lets it be thrown

* This only happens while trying to throw it while you are affected by gravity

beck-thompson:
what the hell?? Just replicated it 😆

ArtisticRoomba:
Yeah this has existed since forever, it's weird

PotentiallyTom:
> Yeah this has existed since forever, it's weird

Yeah I thought for sure someone else would have submitted this

Pronana:
Normal items have friction applied to them two times. Once by tilefriction, and another by physics engine damping.

Mobs have friction applied to them once, either by tilefriction or by the sharedmover controller.

Jetpacks have friction applied to them twice. Once by sharedmover and once by the physics engine. 

Sharedmover's friction values are significantly higher than those of tilefriction so the throwing system miscalculates causing it to do the weird jump animation, and it goes nowhere because sharedmover doesn't have a way to handle a mover being in the air (it's supposed to hand over control to the tilefrictioncontroller in that case, but because a jetpack can technically always move it never does).

MilenVolf:
Fixed by [#36847](https://github.com/space-wizards/space-station-14/pull/36847)

beck-thompson:
thanks!

PR #36847, opened 2025-04-23
Mob Movement Major Refactor

### LINKED ISSUE #35655: Corpses are now following Newton's First Law (This is a bug)
## Description

IIRC, things have drag in space because it's better for the game performance, however as of now, corpses and be launched into space with far less space drag then regular objects
**Reproduction**

Kill someone
Throw them into space
Profit as they fly off so far that station IFF wouldnt be visible.
**Screenshots**

**Additional context**

Pronana:
## About the PR
This PR is a pretty massive refactor to how friction is calculated and used by mobs to lessen the disparity between the friction for a mob that can and cannot move. It also makes a lot of the values more configurable and make more sense.

This PR also uses some fixes from #35984 but makes enough changes that it's no longer a requirement.

Jetpacks for example, no longer have the bug in space with camera rotation and are no longer relay movers.

Friction is now handled by two distinct systems instead of three, and uses them hopefully consistently. Kinematic Controllers use TileFriction and SharedMoverController for friction, with the final value calculated using the Friction method from SharedMoverController. Everything else uses TileFriction and has the physics engine handle the friction. 

There shouldn't be any player facing changes in regards to gamefeel ASIDE FROM THE BUGFIXES TO ANCHORED ENTITIES so if there are, some values may need to be adjusted.

Speaking of Anchored Entities, they now all play (somewhat) nicely together. Conveyors are a different story but they have their own proprietary system for impulse and friction which is layered on top of everything else so they don't bulldoze things, they just are a bit rude.

This means ice crust, puddles, and anything else will no longer have a bulldozing contest to see who's friction value gets used! You can also now use roller skates on ice (Not recommended).

Basically everything friction related has been moved to use a modifier system and rely on CVARs instead of magic numbers whenever possible. 

## Why / Balance
Needed for crawling, it didn't feel good to suddenly gain 12x more friction because you stopped being stunned.
Also moving item friction over to the physics engine means we're only applying friction once which is probably a good thing.

In addition, pretty much all sources of friction were a bulldozing contest to see who could bulldoze friction either first or last, now they use modifiers like civilized variables + fixes a lot of bugs. 

It's also now more consistent which system is used for applying friction and the values should have parity between each-other.

Also fixes #35655 and #36827 and some others probably.

## Technical details
Complete overhaul of how friction is applied and calculated. There was no way to do this without touching so many files.
Jetpacks now bulldoze movement mods on an entity through events rather than being a relayed mover. They function exactly the same player facing wise, except you don't lose your inputs when turning it on now.

## Requirements

## Breaking changes

- JETPACKS ARE NO LONGER RELAY MOVERS!!!
- Friction has been turned into a modifier based system and *any* values that relied on the old bulldozing system will need to be adjusted.
- CVAR tileFriction mod has been divided by 5 and all tile friction has been multiplied by 5 to get rid of magic numbers (By default tiles have a friction mod of 1 instead of 0.2)
- Movers now respect the _tileFriction and a new CVAR _airFriction with baseFriction now being a modifier to these values rather than bulldozing them
- The application of friction to items has been moved entirely to the physics engine damping system
- MobFriction is now capped by acceleration when a mover is trying to move.

**Changelog**

:cl:
- add: Items can now slide on ice crust


Resources/Prototypes/Entities/Mobs/Player/dragon.yml
```diff
@@ -41,7 +41,7 @@
   - type: MovementSpeedModifier
     baseWalkSpeed: 3
     baseSprintSpeed: 5
-    weightlessModifier: 1.5
+    baseWeightlessModifier: 1.5
```
Pronana, line 44: Dragon movement is janky with this value, needs to have friction or something else adjusted to compensate.

Resources/Prototypes/Entities/Mobs/NPCs/behonker.yml
```diff
@@ -49,7 +49,7 @@
     - type: MovementSpeedModifier
       baseWalkSpeed: 3
       baseSprintSpeed: 5
-      weightlessModifier: 1.5
+      baseWeightlessModifier: 1.5
```
Pronana, line 52: Same as for dragon

Content.Shared/Physics/Controllers/SharedConveyorController.cs
```diff
@@ -94,11 +94,17 @@ protected void WakeConveyed(EntityUid conveyorUid)
 
         while (contacts.MoveNext(out var contact))
         {
+            var otherBody = contact.OtherBody(conveyorUid);
+
+            if (otherBody.BodyType == BodyType.Static || !otherBody.Hard)
+                continue;
+
```
Pronana, line 101: Remove this code as it's all sloths and is causing issues.

Content.Shared/Movement/Components/MovementModifiedByContactComponent.cs
```diff
@@ -10,3 +10,8 @@ namespace Content.Shared.Movement.Components;
 public sealed partial class SpeedModifiedByContactComponent : Component
```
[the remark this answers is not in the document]
Pronana: Oh wait I did I am stupid

Pronana:
It's time.

Pronana:
Forgor to update the PR template oops.

Pronana:
Test fail real

Pronana:
I am 95% sure that's all the breaking changes. If there's more please inform me.

ScarKy0:
Who do I bribe to review this...

PR #36848, opened 2025-04-23
Improve LocalizedDataset test to check for low Count

Tayrtahn:
## About the PR

Added logic to `LocalizedDatasetPrototypeTest` to check if a LocId exists using the dataset's Count + 1.

Also, proves its worth by fixing the Count being set too low for the `FigurinesFootsoldier` dataset.

## Why / Balance

This will catch instances where the Count has been set too low. Requested [here](https://github.com/space-wizards/space-station-14/pull/36846#issuecomment-2822852702).

## Technical details

Added an additional check for each prototype that generates the next LocId that would be used and checks if a localized string is registered with that ID. If it is, the test fails with a message suggesting that Count should be raised.

## Requirements


slarticodefast (approved):
Awesome, thanks! This should prevent a lot of accidental user error.

PR #32640, opened 2024-10-04
Adds a variety of new options for passenger loadouts.

packmore:
## About the PR
Adds a lot of new options for passenger loadouts. Mainly jumpsuits - colored jumpsuits, Hawaiian shirts, casual jumpsuits - but also adds a few new misc items, like cheap sunglasses and colored shoes.

## Why / Balance
Passenger has the least responsibility and the most adaptability when it comes to character creation, gameplay variety, and roleplay. Almost every single item added to their loadouts in this PR can be obtained from basic clothing vendors scattered around the station, cargo, and uniform printers. Nothing added to their loadouts is generally out of their control to obtain, and almost all of it can be obtained very easily. The only real exception to this is the Gar Glasses, which I would accept being removed in favor of remaining as maintenance loot.  

Very few of these outfits can be used to disguise yourself as another department, and acquiring the outfits to do so is already quite easy. However, to prevent raiders from abusing this, and to ensure that it is still somewhat readable when someone is a new join, there is a 5 hour playtime requirement on most of these additional clothing items.

## Technical details
Changes the passenger.yml loadout file, and the loadout_groups.yml file.

## Media

**https://youtu.be/hKRkzUpxIxY**

Screenshot (1523)

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- add: Added new customization options for passenger loadouts.

PR #36849, opened 2025-04-23
make loadouts menu grid

qrwas:
## About the PR
This PR changes the display of items in the Loadouts menu to a grid. 

pull request adding groups: https://github.com/space-wizards/space-station-14/pull/36951

## Why / Balance
In this PR https://github.com/space-wizards/space-station-14/pull/32640, slarticodefast noted that it would be worth changing the look of this menu, and I suggested a new look (from another account), which judging by the message, he liked. https://github.com/space-wizards/space-station-14/pull/32640#issuecomment-2815903094

It might be a good idea to reduce the size of the icons, add a maximum width, and increase the number of columns to 4-5.

This will also allow you to read the item description, if anyone needs it, of course.

## Technical details
Change LoadoutGroupContainer.xaml to display grid
Added overview text output, instead of Label of the item name in the button itself
Since there were problems with the text, and I didn't understand how to correctly add a break to the new line of the label and align the text in the center, and because it looked bad, I decided to remove the Label from the element.
The description appears only after shift+click, like an in-game overview

### Old
image5

## Requirements

**Changelog**
:cl: qrwas
- tweak: Changed the appearance of the loadouts menu. More space for the item icon instead of text!


iaada:
I like this a lot! I do think the size needs to be a little smaller though.

RangerXVII:
I think maybe 5-6 on each row would look nicer. Right now It's a little too big.

qrwas:
on 4 columns and 4 by 4 images (instead of 5 by 5) looks better

qrwas:
image2

qrwas:
I tried making groups, it looks interesting, but I want to add a background for an open group
image
image
image

qrwas:
someday...

sowelipililimute:
not a formal blocker but one concern I have with this is that some downstreams let you take implanters as a loadout option and this makes the text that distinguishes between them hover-only which isn't ideal

qrwas:
image
Perhaps a better option for implants would be to add a color change that shows that the implant is not empty.

qrwas:
> not a formal blocker but one concern I have with this is that some downstreams let you take e.g. implanters or alternate job title PDAs as a loadout option and this makes the text that distinguishes between them hover-only which isn't ideal

image
as an example

qrwas:
> adjusting the grid to show text below icon shouldn't be terribly difficult and would prevent regressing that usecase

Adding text under the icon is terrible. Both in implementation and appearance. Considering that there are names that consist of 3-4 long words.

qrwas:
> not a formal blocker but one concern I have with this is that some downstreams let you take e.g. implanters or alternate job title PDAs as a loadout option and this makes the text that distinguishes between them hover-only which isn't ideal

Well, if they accept https://github.com/space-wizards/space-station-14/pull/37288 there won't be such a problem with implants

metalgearsloth:
I still think text description under the item is useful as the sprites aren't always clear on it and tooltips are slow.

qrwas:
> > I still think text description under the item is useful as the sprites aren't always clear on it and tooltips are slow.
> 
> Could add a text box at the top of the UI but beloe the header that states the selected item, and maybe another box that states all confirmed items.

image
Something like that? Or a pop-up text when hovering with the name?

sowelipililimute:
functionally has the same main problem as the design before, the core issue is that the items need to have their text available in the grid in some manner if you wish to have players scanning a grid for different items with the same/similar sprite.

i don't see why adding text is an impossibility in a grid layout, given the abundance of space before and after the change. if you'd like to elaborate your reasoning/give an example of that design not working well, that would help.

qrwas:
This is what it looks like
image
image
image

I can try to do this with RichTextLabel, but it won't be centered.
In general, in my opinion, the text on such icons looks terrible and is unnecessary.
Example with the captain's uniform. I can already see that it is the captain's uniform, because I opened the captain's loadout. Then I just look at the appearance, and not read the text that will confirm to me: "yes, this is really the captain's uniform, did you think otherwise?"

image
image
With cigarettes. I see the packaging, and if I'm curious about what kind of cigarette it is, I just look at them, just like in the game.
Of course, a problem may arise, for example, with a cigar case for beginners: "what is this strange gray rectangle?" - examine it with shift+click, and you will find out. But this applies to any new things.

If the player has already seen it in the game, he doesn't need descriptions that take up all the space on the screen, he needs the appearance.

beck-thompson:
Thanks for the PR!

The main thing that came up was concerns over colorblind people having issues telling different items apart (such as towels) because of the lack of text. Even with tooltip text, you cannot immediately tell it apart and it's less convenient overall. I suggested adding text to the grid but the issue with that is it won't work very well with long names.

I'm closing the PR but the group pr (#36951) is still good (obviously it would have to not be a grid anymore but the idea is solid)

iaada:
There's also the option of a toggle between a grid and a list, like how the construction menu does it.

PR #36859, opened 2025-04-23
Remove redundant fields from `MovementSpeedModifierComponent`

Tayrtahn:
## About the PR

Deleted `_baseWalkSpeedVV` and `_baseSprintSpeedVV` from `MovementSpeedModifierComponent`.

## Why / Balance

They are outdated and pointless, and they produce compiler warnings.
Two more for [PZW](https://github.com/space-wizards/space-station-14/issues/33279).

## Technical details

`BaseWalkSpeed` and `BaseSprintSpeed` can both be set using View Variables, and doing so will `Dirty` the field automatically (I have tested and verified this). These extra fields presumably come from a time before that was the case, but now they are redundant.

## Requirements


slarticodefast (approved):
Yep, I tested this and it works. The normal datafields are already dirtied when changed.
I think we got a bunch of other components that have similar fields.

PR #36865, opened 2025-04-23
Holidayselector & Holiday systems in shared

IProduceWidgets:
## About the PR

I just wanted to add a selector man.

The HolidaySelector works on entity tables and is an AllSelector which only works on holidays.

Moved all the holiday stuff to shared so I could access it, with the exception of a couple things that are now in the ServerHolidaySystem instead

## Why / Balance

I sure hope you don't use holidays for balance >.>

## Technical details

Most everything was pretty much cut and paste into shared.
I changed a bunch of the namespace to reflect that,

I kept the bulk of the christmas stuff in server because it seemed annoying to move since it messes with other server systems, and also some of the components list secrets.

I left the stuff that interacts with gameticking and chat messaging in server because I think I have to.

## Requirements

## Breaking changes

You may need to change your references for the Holiday System to Shared instead of Server.

**Changelog**

:cl:
- add: a Holiday conditioned entity table selector.

### DIFF
--- a/Content.Shared/EntityTable/EntitySelectors/HolidaySelector.cs
+++ b/Content.Shared/EntityTable/EntitySelectors/HolidaySelector.cs
@@ -33,5 +33,6 @@ public sealed partial class HolidaySelector : EntityTableSelector
                 }
             }
         }
+        yield break;
     }
 }
--- a/Content.Shared/EntityTable/EntitySelectors/HolidaySelector.cs
+++ b/Content.Shared/EntityTable/EntitySelectors/HolidaySelector.cs
@@ -1,0 +1,37 @@
+using Content.Shared.Holiday;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.EntityTable.EntitySelectors;
+
+/// <summary>
+/// If the given Holiday is active, Gets spawns from all of the child selectors
+/// </summary>
+public sealed partial class HolidaySelector : EntityTableSelector
+{
+    [Dependency] private readonly HolidaySystem _holiday = default!;
+
+    /// <summary>
+    /// Holiday prototype ID
+    /// </summary>
+    [DataField(required: true)]
+    public string Holiday = default!;
+
+    [DataField(required: true)]
+    public List<EntityTableSelector> Children = default!;
+
+    protected override IEnumerable<EntProtoId> GetSpawnsImplementation(System.Random rand,
+        IEntityManager entMan,
+        IPrototypeManager proto)
+    {
+        if (_holiday.IsCurrentlyHoliday(Holiday))
+        {
+            foreach (var child in Children)
+            {
+                foreach (var spawn in child.GetSpawns(rand, entMan, proto))
+                {
+                    yield return spawn;
+                }
+            }
+        }
+    }
+}
--- a/Content.Shared/Holiday/Christmas/LimitedItemGiverComponent.cs
+++ b/Content.Shared/Holiday/Christmas/LimitedItemGiverComponent.cs
@@ -1,13 +1,13 @@
-﻿using Content.Shared.Storage;
+using Content.Shared.Storage;
 using Robust.Shared.Network;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype;
 
-namespace Content.Server.Holiday.Christmas;
+namespace Content.Shared.Holiday.Christmas;
 
 /// <summary>
 /// This is used for granting items to lucky souls, exactly once.
 /// </summary>
-[RegisterComponent, Access(typeof(LimitedItemGiverSystem))]
+[RegisterComponent]
 public sealed partial class LimitedItemGiverComponent : Component
 {
     /// <summary>
--- a/Content.Shared/Holiday/HolidaySystem.cs
+++ b/Content.Shared/Holiday/HolidaySystem.cs
@@ -1,30 +1,29 @@
 using System.Linq;
-using Content.Server.Chat.Managers;
-using Content.Server.GameTicking;
 using Content.Shared.CCVar;
-using Content.Shared.Holiday;
 using Robust.Shared.Configuration;
+using Robust.Shared.Network;
 using Robust.Shared.Prototypes;
 
-namespace Content.Server.Holiday
+namespace Content.Shared.Holiday
 {
     public sealed class HolidaySystem : EntitySystem
     {
         [Dependency] private readonly IConfigurationManager _configManager = default!;
         [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
-        [Dependency] private readonly IChatManager _chatManager = default!;
+        [Dependency] private readonly INetManager _net = default!;
         [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
 
         [ViewVariables]
         private readonly List<HolidayPrototype> _currentHolidays = new();
+        public List<HolidayPrototype> CurrentHolidays => _currentHolidays;
 
         [ViewVariables]
         private bool _enabled = true;
+        public bool Enabled => _enabled;
 
         public override void Initialize()
         {
             Subs.CVar(_configManager, CCVars.HolidaysEnabled, OnHolidaysEnableChange);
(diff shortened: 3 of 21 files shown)


lzk228:
i was thinking of adding holiday things to mail a few days ago and here it is possible now🥹

Content.Shared/Holiday/Christmas/LimitedItemGiverComponent.cs
```diff
@@ -1,13 +1,13 @@
...
 using Robust.Shared.Network;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype;
 
-namespace Content.Server.Holiday.Christmas;
+namespace Content.Shared.Holiday.Christmas;
 
 /// <summary>
 /// This is used for granting items to lucky souls, exactly once.
 /// </summary>
-[RegisterComponent, Access(typeof(LimitedItemGiverSystem))]
```
slarticodefast, line 10: Why remove that?
Also moving all the components to shared won't do much alone without also networking them and moving the corresponding system to shared. I would leave them as they are for now in this PR and move them one by one.

Content.Shared/Holiday/HolidaySystem.cs
```diff
@@ -1,30 +1,29 @@
...
 using Content.Shared.CCVar;
-using Content.Shared.Holiday;
 using Robust.Shared.Configuration;
+using Robust.Shared.Network;
 using Robust.Shared.Prototypes;
 
-namespace Content.Server.Holiday
+namespace Content.Shared.Holiday
 {
     public sealed class HolidaySystem : EntitySystem
```
slarticodefast, line 9:
```suggestion
    public abstract partial class SharedHolidaySystem : EntitySystem
```
naming convention
rename the file accordingly
the server and client systems inheriting should be simply named `HolidaySystem`

Content.Shared/Holiday/HolidaySystem.cs
```diff
@@ -1,30 +1,29 @@
...
     {
         [Dependency] private readonly IConfigurationManager _configManager = default!;
         [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
-        [Dependency] private readonly IChatManager _chatManager = default!;
+        [Dependency] private readonly INetManager _net = default!;
         [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
 
         [ViewVariables]
```
slarticodefast, line 16: not your fault, but what the hell is this doing here?

Content.Shared/Holiday/HolidaySystem.cs
```diff
@@ -51,19 +50,14 @@ public void RefreshCurrentHolidays()
...
-                _chatManager.DispatchServerAnnouncement(holiday.Greet());
-            }
-        }
-
         public void DoCelebrate()
         {
             foreach (var holiday in _currentHolidays)
             {
-                holiday.Celebrate();
+                if (_net.IsServer)
```
slarticodefast, line 57: Move outside the loop.
Or even better, make this a virtual function and override it on the server. Same for `DoGreet`. That way they can be called from shared.

Content.Shared/Holiday/HolidaySystem.cs
```diff
@@ -1,30 +1,29 @@
...
+        [Dependency] private readonly INetManager _net = default!;
         [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
 
         [ViewVariables]
         private readonly List<HolidayPrototype> _currentHolidays = new();
+        public List<HolidayPrototype> CurrentHolidays => _currentHolidays;
 
         [ViewVariables]
         private bool _enabled = true;
+        public bool Enabled => _enabled;
```
slarticodefast, line 22: wouldn't it be simpler to make it protected?

slarticodefast (changes requested):
You will need to network `DateTime.Now` from the server to the client or they may disagree depending on your computer's time settings.

EmoGarbage404 (changes requested):
instead of moving all of the holiday code into shared, just send a network message to clients containing the list of CurrentHolidays. Then all you need is to put the API in a shared system.

Content.Shared/Holiday/Christmas/LimitedItemGiverComponent.cs
IProduceWidgets: Eh, sure. I was hopeful I could move the system out of shared too, but I think that's gonna be a bigger job, and there's maybe some reasons why parts of it shouldn't be to keep secrets from clients.

IProduceWidgets:
> You will need to network `DateTime.Now` from the server to the client or they may disagree depending on your computer's time settings.

I did so with a couple pairs of events now for both DateTime.Now and the ccvar for HolidayEnabled.

Not sure I used the right approach though.

IProduceWidgets:
> instead of moving all of the holiday code into shared, just send a network message to clients containing the list of CurrentHolidays. Then all you need is to put the API in a shared system.

That probably would have been easier yeah. 
Alas, my mistress Caffeine took me away. We here now.
Bright side, you can check against the ChineseLunisolarCalendar now. uh,

Content.Shared/EntityTable/EntitySelectors/HolidaySelector.cs
```diff
@@ -0,0 +1,38 @@
+using Content.Shared.Holiday;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.EntityTable.EntitySelectors;
+
+/// <summary>
+/// If the given Holiday is active, Gets spawns from all of the child selectors
+/// </summary>
+public sealed partial class HolidaySelector : EntityTableSelector
```
EmoGarbage404, line 9: this really would ideally be handled by the condition system i have outlined in https://github.com/space-wizards/space-station-14/pull/36819.

not that this is particularly bad, i just think its a bit of a clunky way to do it with the lack of proper condition support.

IProduceWidgets: Sounds like a plan. Yoinked the selector, but Ill have to wait on this and that PR to merge to make it due to shared.

EmoGarbage404 (approved):

IProduceWidgets:
Hold merge for a sec, I am dummy with network events and race conditions.

IProduceWidgets:
Kay, think I fixed that.

I do get these on connect, but Idk maybe its fine?
image

Let me know if Im just using network events wrong, I fixed the issue by just ending the method that raised the event I need the server to respond to to continue, and then having it call the second half of what it once was. This seems to have fixed the race issue, but It does not make me feel warm and cuddly.
More details in the form of a question:
https://discord.com/channels/310555209753690112/560845886263918612/1365797321580871692

Content.Shared/Holiday/SharedHolidaySystem.cs
```diff
@@ -0,0 +1,120 @@
...
+            RefreshCurrentHolidaysPart2(now);
+        }
+
+        public void RefreshCurrentHolidaysPart2(DateTime then)
+        {
+            var now = _currentDate;
+
+            // The following is intended to interrogate race conditions since the first part of this method needs to let the server get the date and set it for shared.
+            if (_net.IsServer)
+                Log.Warning($"the value for now on Server is {now} and was {then}");
```
EmoGarbage404, line 63: this doesn't work. the second function is just going to compute immediately before the network events hit anything.

in general, why are you bothering raising these events on the server? The server should just precalculate all of this and then the client, on connect, should send a message to the server to receive the data. i don't understand what all of this is for.

EmoGarbage404 (changes requested):

Content.Shared/Holiday/SharedHolidaySystem.cs
IProduceWidgets: How do you await the response of a network event? It didn't work until I did this, but being honest I'm not sure why. 

But to answer your question; because for some reason we can hotload holidays.

IProduceWidgets: I dunno, Ill sleep on it and try again.

IProduceWidgets:
Gonna close this for now because I slept on it and realized Emo's right, but salvaging this branch is a feelsbad.

PR #36873, opened 2025-04-24
make throw insert container code more clear

lzk228:
so here is the thing, the datafield comment says
> Throw chance of hitting into the container

but it's actually the opposite!

so 75% of hitting into container, as it says, actually is 25% of hitting
i made it right

it causes a breaking change and i could just change the comment or something, but i find it not convenient

## Breaking changes

Inverted logic of `Probability` datafield in `ThrowInsertContainerComponent`. (75% of hitting actually meant 25% so it should be changed like this.) You need to update your prototypes to this change.

### DIFF
--- a/Content.Server/Containers/ThrowInsertContainerComponent.cs
+++ b/Content.Server/Containers/ThrowInsertContainerComponent.cs
@@ -16,7 +16,7 @@ public sealed partial class ThrowInsertContainerComponent : Component
     /// Throw chance of hitting into the container
     /// </summary>
     [DataField]
-    public float Probability = 0.75f;
+    public float Probability = 0.25f;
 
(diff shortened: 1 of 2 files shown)


EmoGarbage404:
there's no point in an obscure breaking change when you could just change the comment.

lzk228:
then the name "Probability" is not that logical 
in this case it's better to change it to "MissProbability"
either one way or another

slarticodefast (approved):
Eh, it's probably fine. Technically it is a breaking change, but
- it is more of a bugfix, since it did not do what the name and documentation were claiming it would do.
- this is probably hardly used anywhere else, and if so then it was likely used wrong due to the point above.

PR #36876, opened 2025-04-24
Remove beaker from the MaterialArbitrageTest whitelist

ElectroJr:
## About the PR
Removes the beaker entity from the MaterialArbitrageTest whitelist,  as the beaker's arbitrage should've been fixed in #34441


slarticodefast (approved):
Why the hell do we have this whitelist in the first place 😞

PR #36880, opened 2025-04-24
Removes the generic apron

Flareguy:
## About the PR
Removes the generic apron from the game. All existing generic aprons will be replaced with botany aprons.

## Why / Balance
So I know what you're thinking. "What the hell did the generic apron do to you?" and the answer to that is nothing, really. It's a pretty inoffensive item on the face of it. The item doesn't quite fit the artstyle, but it could be pretty easily resprited.

Despite that, this item has basically no place in the game, and only really serves to exist as a weird bit of inconsistency for maps that decide to map it in. The normal botany apron (or any other apron, for that matter) is usually going to feel significantly more cohesive & make more sense in basically every situation you put it in. They are already incredibly generic by nature of being aprons - they do not need a generic counterpart.
This is also ignoring that the item is (presumably?) just a CEV-Eris leftover. It's seemingly completely divorced from its original context.

## Technical details
Removes the generic apron and its sprites.

## Media
N/A

## Requirements

## Breaking changes
ClothingOuterApron has been removed. It has been migrated to ClothingOuterApronBotanist.

**Changelog**
no cl no apron no fun


SlimmSlamm:
This is a pretty useless PR.

slarticodefast:
@SlimmSlamm Please keep comments here constructive. Thanks.

IProduceWidgets:
I kinda don't like migrating away entities for little reason. It adds mapping overhead and as we get more maps and grids, doing so will increasingly lead to these things getting changed or removed and never fixed.

This is gonna be controversial, but Id rather you just change the prototype so it does nothing but parent the entity your migrating to instead.

Flareguy:
> This isn't really worth the breaking changes when you could just copy over one of the existing apron textures and color it yellow.

Sure. I'll close the PR since it'd be significantly easier to just remake it at that point.

PR #36882, opened 2025-04-24
Fix RoundToEqualLevels for max = levels case

beck-thompson:
## About the PR

There was odd behavior when the levels were equal to the max amount of items. This resolves the issue by just doing a specific check for the case of the levels equaling the max! There isn't really another easy fix, this is the cleanest I could think of.

See #35697 (This PR fixes the issue. Checkout the cone pr and make the cone levels normal to test this pr!)
See #35698 (Another approach)

## Requirements

## Breaking changes

`RoundToEqualLevels` in `ContentHelper.cs` now properly sets the states when the levels are equal to the maximum value.

### DIFF
--- a/Content.Shared/Rounding/ContentHelpers.cs
+++ b/Content.Shared/Rounding/ContentHelpers.cs
@@ -110,7 +110,11 @@
                 return 0;
             }
 
-            return (int) Math.Round(actual / max * levels, MidpointRounding.ToZero);
+            // This is necessary if the stack size is the same as the number of levels.
+            // If this doesn't happen, it starts at 1 instead of 0 and will skip the first state.
+            var mod = levels == (int) max ? 1 : 0;
+
+            return (int) Math.Round(actual / max * levels, MidpointRounding.ToZero) - mod;
         }
     }
 }


Content.Shared/Rounding/ContentHelpers.cs
```diff
@@ -110,7 +110,11 @@ public static int RoundToEqualLevels(double actual, double max, int levels)
                 return 0;
             }
 
-            return (int) Math.Round(actual / max * levels, MidpointRounding.ToZero);
+            // This is necessary if the stack size is the same as the number of levels.
+            // If this doesn't happen, it starts at 1 instead of 0 and will skip the first state.
+            var mod = levels == (int) max ? 1 : 0;
```
EmoGarbage404, line 115: int to double comparison is suspect at best, even with the int casting. Also, how is this meant to handle noninteger values for `max`? If you actually want it to do a Math.Floor transformation then you should do that specifically instead of what appears to be kinda suspect logic.

EmoGarbage404 (changes requested):

PR #33286, opened 2024-11-13
Four new food crates for the ATS

AgentSmithRadio:
## About the PR
I've added ~~six~~ new food crates available for purchase from the ATS. A brief summary:

1. **Getmore Bakemore Crate** - A slimmed down kitchen supplies crate with 4-6 of each Getmore baking ingredient (raisins/chocolate/pistachios).
2. ~~**Black Market Meats Freezer** - Freezer containing 4 human meat, 2 lizard meat and 2 plant meat.~~
3. ~~**Moth Pizza Delivery** - 4 cotton pizzas for stations that want to accommodate moths, otherwise identical to a normal pizza delivery crate. They spawn properly in pizza boxes with plastic knives.~~
4. ~~**Personnel Surprise Crate** - High-end hamburger and fries meal with cake, bottle of space cola, water cup party hat and mystery figure box, a way to reward/celebrate/cheer up a crew member. Not made for sharing.~~
5. **Ice Cream Delivery** - 16 pieces of assorted ice cream food items, probably going to be very popular if implemented.
6. **Snowcone Delivery** - 16 assorted snowcones for _slightly less_ than the ice cream delivery, because they deserve representation too.
7. NEW: **Happy Honk Meal Delivery** - Big bite burger, bottle of space cola, cheesy fries, a slice of apple pie, and a random toy, all in an official Happy Honk box! Now you can literally ask your QM to buy you McDonalds. Two meals per crate.

Also, while you're here, please see [my other pull request on hacked vending machine contrabandInventory menus.](https://github.com/space-wizards/space-station-14/pull/32934)

EDIT: Minor bonus updates. Capitalized the start of a sentence on Sweetie's Pistachio's description, and removed it as a possible item in the cargo fruit bounty for $5k. They're seemingly a fruit so Reptilians can eat them, but I've never even heard of them being used in the bounty before, so I've removed it as a possible item.

EDIT: Removed Black Market Meats Freezer from PR.
EDIT 2: Removed Moth Pizza Delivery as cotton pizzas are now guaranteed in normal pizza delivery orders.
[...]

PR #36890, opened 2025-04-24
Reptilians Can Eat Orange Creamsicles

AgentSmithRadio:
## About the PR
I've added a fruit tag to the orange creamsicle entity in frozen.yml to bring it in line with Reptilian dietary restrictions, and therefore make it edible to the species.

## Why / Balance
Reptilians are able to consume foods with meat and fruit tags. Otherwise inedible foods, such as cakes, are made edible when fruit flavoured (for instance, Reptilians can eat orange cake, but not normal cake, or carrot cake). On top of that, the berry creamsicle entity already had a fruit tag to make it edible to Reptilians.

This was a relatively rare entity before, but now with the [Ice Cream Delivery](https://github.com/space-wizards/space-station-14/pull/33286) crate now available from cargo, they will become a much more common item. This change gives Reptilians two options to pick from the freezer, all while staying in line with their established dietary habits.

## Technical details
Added the "fruit" tag to orange creamsicle, matching what is already there for berry creamsicle in frozen.yml.

## Media
image

## Requirements

## Breaking changes
None.

**Changelog**
:cl: AgentSmithRadio
- tweak: Reptilians can now eat orange creamsicles.
- fix: Berry creamsicles can no longer be used in cargo's generic fruit bounty.

### DIFF
--- a/Resources/Prototypes/Catalog/Bounties/bounties.yml
+++ b/Resources/Prototypes/Catalog/Bounties/bounties.yml
@@ -564,7 +564,8 @@
       - Cake
       - Pie
       - Bread
-      - Pistachios  
+      - Pistachios
+      - Creamsicle
 
 - type: cargoBounty
   id: BountyVegetable
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/frozen.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/frozen.yml
@@ -90,6 +90,9 @@
   - type: Food
     trash: 
     - FoodFrozenPopsicleTrash
+  - type: Tag
+    tags:
+    - Fruit
 
 - type: entity
   name: berry creamsicle
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/frozen.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/frozen.yml
@@ -93,6 +93,7 @@
   - type: Tag
     tags:
     - Fruit
+    - Creamsicle
 
 - type: entity
   name: berry creamsicle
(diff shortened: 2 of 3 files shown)


slarticodefast:
The fruit tag will make it count towards the fruit cargo bounty.

AgentSmithRadio:
> The fruit tag will make it count towards the fruit cargo bounty.

You're right. I realized this before, but hadn't done the math, and now it might be an issue. Generic fruit is $5000 for 12 fruit tag items (with restrictions). You could in theory spend $3600 on 3 Ice Cream Delivery crates, remove the 4 creamsicles from each crate, and cycle the bounty with a $1400 profit.

The solution to this is for me to add a new tag to the library and attach it to the fruit bounty if we want to guard against this.

IProduceWidgets:
> The solution to this is for me to add a new tag to the library and attach it to the fruit bounty if we want to guard against this.

What we really need is tracking for ordered items so they don't sell for very much or count in bounties. Similar to DwarfFortress' foreign goods debuff.

AgentSmithRadio:
> What we really need is tracking for ordered items so they don't sell for very much or count in bounties. Similar to DwarfFortress' foreign goods debuff.

That could work, but then we're talking a bounty system refactor. Some bounties, such as instruments, percussion instruments, clown mask + shoes, 2x rollerskates and the like are inherently order dependent. It's often the only way to complete those bounties. If such a system were to go in place, we'd need to remove those types of bounties first, or provide alternate sources for these items.

AgentSmithRadio:
Alrighty, I've added a Creamsicle tag to tags.yml, tagged both creamsicle entities with that tag, and then added the tag to BountyFruit. The cargo bounty loophole has been closed.

beck-thompson (approved):
looks good to me! I'm a little sused out by not tagging the rest of the icecream stuff but I think this makes sense

ArtisticRoomba (approved):
Thanks for le contribution!

PR #36895, opened 2025-04-24
Unite LogComponent, FoodSlicableComponent and partially ButcherableComponent into one. Sharp component removal.

### LINKED ISSUE #28141: Need merge LogSystem and SliceableFood system
similar logic
I also think we need to add doAfter to the slicing process, clicking through 10 logs in a second doesn't look great.

### LINKED ISSUE #28973: Butch verb should be shown only with sharp objects in hand
Like every other item ability (soap for cleaning, etc) sharp objects should has verb ONLY if they are inhand, but butch verb is always visible

### LINKED ISSUE #35042: Butcherable dupe
## Description

There is no validation for Butcherable enum Knife action, two people who's butchering entity with knifr at the same time will duplicate an item
**Reproduction**

Two players should butcher entity with knife at the same time, at the end of doAfter result entity is gonna duplicate
**Screenshots**

**Additional context**

kosticia:
## About the PR
Removes LogComponent, FoodSlicableComponent and related to them systems. Removes slicing functional from ButcherableComponent.
Replaces them via SlicableComponent that contains all functions from them.

## Why / Balance
Right now we have three components that do the same things. Sharp component removed, because tool with slicing exist.
Knife bounty now actually require knives. 
Closes #28973
Closes #28141
Closes #35042

## Technical details
Makes everything predicted, but leaves a lot of breaking changes.

## Media
https://github.com/user-attachments/assets/a32ba91c-a2db-4771-9517-3034e84f53c4

## Requirements

## Breaking changes
Removed systems and components (their logic and fields, required for it was moved to `ToolRefinableSystem` and `ToolRefinableComponent`):
* `LogComponent` (and `LogSystem`) was removed - replace with `ToolRefinableComponent` and add `BaseSlicingRefinable` prototype as parent to fill all related fields with proper defaults
* `FoodSlicableComponent` and `SliceableFoodSystem` - replace with `ToolRefinableComponent` and add `BaseSlicingRefinableFood` prototype as parent to fill all related fields with proper defaults, `ToolRefinableSolutionComponent` contains name for solution conainer id that should be used to get reagents from original entity and to which reagents should be put on refined entities.
* `ButcherableComponent` and now is used only for meatspike, so on prototypes that should not require meatspike - replace with `ToolRefinableComponent` and add `BaseSlicingRefinable` prototype as parent to fill all related fields with proper defaults
* `SharpComponent` Use `ToolComponent` with `Slicing` quality instead of it.
* `SharpDoAfterEvent` was removed, `ToolRefineDoAfterEvent` can be used instead
*  `SliceFoodEvent` was removed, `AttemptToolRefineEvent` and `BeforeToolRefinedEvent` events were introduced instead. 'Attempt' enables blocking action without any actual consequences, and 'Before' enables cancelling action after it already was in progress (for example to un-polymorph person who was bread upon cutting it down, without creating bread slices).
* renamed localization strings related to sliceable food and butcherable, new string names can be found in Resources/Locale/en-US/tool-refinable/tool-refinable.ftl

**Changelog**
:cl:
- tweak: Biggest part of items with slicing tool quality now can be used for butchering.
- tweak: Butchering is now predicted and called slicing!


kosticia:
Nice triaging

Winkarst-cpu:
Are you still working on that PR?

kosticia:
> Are you still working on that PR?

I'm waiting for the minimal maintainer review on the codebase level. Because fixing conflicts when in every moment can appear maintainer and say "SharpComponent removal is hell for forks, close your pr" doesn't makes any sense.
Also, I'm a little bit scared of what I have coded 3 months ago.

Fildrance:
BIG-BIG and important request - can you please look into making slicing into a tool? :(

Content.Server/Botany/Components/PlantHolderComponent.cs
```diff
@@ -138,4 +138,7 @@ public sealed partial class PlantHolderComponent : Component
 
     [ViewVariables]
     public Entity<SolutionComponent>? SoilSolution = null;
+
+    [DataField]
+    public string HarvestToolQuality = "Slicing";
```
Fildrance: a short xml-doc would be helpful

Content.Server/Botany/Systems/BotanySystem.Seed.cs
```diff
@@ -194,7 +195,7 @@ public IEnumerable<EntityUid> GenerateProduct(SeedData proto, EntityCoordinates
 
     public bool CanHarvest(SeedData proto, EntityUid? held = null)
     {
-        return !proto.Ligneous || proto.Ligneous && held != null && HasComp<SharpComponent>(held);
+        return !proto.Ligneous || proto.Ligneous && held != null && _tools.HasQuality(held.Value, "Slicing");
```
Fildrance: please use ProtoId<ToolQualityPrototype> as much as you can

Content.Shared/Tools/Components/ToolComponent.cs
```diff
@@ -2,6 +2,7 @@
 using Robust.Shared.Audio;
 using Robust.Shared.GameStates;
 using Robust.Shared.Utility;
+using Content.Shared.Sliceable;
```
Fildrance: seems unneeded

Resources/Prototypes/Entities/Objects/Consumable/Food/produce.yml
```diff
@@ -2223,13 +2227,10 @@
         solution: food
       - !type:DoActsBehavior
         acts: [ "Destruction" ]
-  - type: SliceableFood
-    count: 5
-    slice: FoodHolymelonSlice
-  - type: Butcherable
-    butcheringType: Knife
-    spawned:
-    - id: ClothingHeadHatHolyWatermelon
```
Fildrance, line 2209: you just removed the only legal way to do this? :( probably make sense to spawn it randomly? hmmm

Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
```diff
@@ -2457,8 +2451,16 @@
         - MobMask
         layer:
         - MobLayer
-  - type: Butcherable
-    spawned:
+  - type: DamageStateVisuals
```
Fildrance: this looks pretty harmless, but i have to ask - why is it here :<

Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
```diff
@@ -1617,9 +1612,8 @@
           horns_floppy_kobold_ears: Inherit
           horns_double: Inherit
           horns_kobold_ears: Inherit
-  - type: Butcherable
-    butcheringType: Spike
```
Fildrance, line 1857: it required spike previously, but now it doesnt? why so =w=

Resources/Prototypes/Entities/Mobs/NPCs/pets.yml
```diff
@@ -855,9 +855,8 @@
     description: ghost-role-information-punpun-description
     rules: ghost-role-information-nonantagonist-rules
   - type: GhostTakeoverAvailable
-  - type: Butcherable
-    butcheringType: Spike
```
Fildrance, line 876: same, u changed it to not require spike? why

Resources/Prototypes/Entities/Mobs/NPCs/xeno.yml
```diff
@@ -447,10 +444,9 @@
         Base: dead_small_purple_snake
       Dead:
         Base: dead_small_purple_snake
-  - type: Butcherable
-    spawned:
+  - type: Sliceable
+    slices:
     - id: FoodMeatSnake
-      amount: 2
```
Fildrance, line 465: not sure why amount is lowered :<
i dont think its healthy to do microbalance in such PRs, please make sure that numbers of spawned amount are the same wherever its possible - its refactor, after all

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+            _solutionContainer.TryAddSolution(itsSoln.Value, lostSolutionPart);
+        }
+    }
+}
+
+/// <summary>
+/// Called for doafter.
+/// </summary>
+[Serializable, NetSerializable]
+public sealed partial class TrySliceEvent : SimpleDoAfterEvent;
```
Fildrance: better call it TrySliceDoAfterEvent - so you can understand that this is DoAfter right away

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+            _solutionContainer.RemoveAllSolution(itsSoln.Value);
+
+            var lostSolutionPart = solution.SplitSolution(itsSolution.AvailableVolume);
+            _solutionContainer.TryAddSolution(itsSoln.Value, lostSolutionPart);
+        }
+    }
+}
+
+/// <summary>
+/// Called for doafter.
```
Fildrance: this is not good description - remember that reading xml-doc is usually done with intellisense which provide no context (file where class lives)

Content.Server/Kitchen/EntitySystems/KitchenSpikeSystem.cs
```diff
@@ -173,24 +175,21 @@ private void Spike(EntityUid uid, EntityUid userUid, EntityUid victimUid,
...
-            KitchenSpikeComponent? component = null, SharpComponent? sharp = null)
+            KitchenSpikeComponent? component = null)
         {
             if (!Resolve(uid, ref component) || component.PrototypesToSpawn == null || component.PrototypesToSpawn.Count == 0)
                 return false;
 
             // Is using knife
-            if (!Resolve(used, ref sharp, false) )
-            {
+            if (!_tools.HasQuality(used, "Slicing"))
```
Fildrance: same, use ProtoId and not magic string

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        if (args.Using is { } used)
+        {
+            if (!TryComp<ToolComponent>(used, out var toolComp))
+                return;
+
+            if (!_tool.HasQuality(used, comp.ToolQuality))
+            {
+                verbDisabled = true;
+                verbMessage = Loc.GetString("slice-verb-message-tool", ("uid", uid));
```
Fildrance: you put it as 'uid' but loc string is waiting for 'target'

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+        if (_tool.HasQuality(args.Used, comp.ToolQuality))
+        {
+            OnVerbUsing(uid, args.User, args.Used, comp.SliceTime.Seconds, comp.ToolQuality);
+            return;
+        }
+    }
+
+    private void AddSliceVerb(EntityUid uid, SliceableComponent comp, GetVerbsEvent<InteractionVerb> args)
+    {
+        if (!args.CanInteract)
```
Fildrance: butcherable previously also was checking CanAccess - it would not be funny if you can butcher through doors or windows

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+            return;
+
+        var text = Loc.GetString("slice-verb-name");
```
Fildrance: not too descriptive variable name, tbh :(

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+        var verbDisabled = false;
+        var verbMessage = Loc.GetString("slice-verb-message-default");
+        EntityUid tool;
+
+        if (args.Using is { } used)
```
Fildrance: not a fan of that thing, ```is not null``` looks clearer tbh

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+
+    private void OnInteraction(EntityUid uid, SliceableComponent comp, InteractUsingEvent args)
+    {
+        if (args.Handled)
+            return;
+
+        args.Handled = true;
+        if (_tool.HasQuality(args.Used, comp.ToolQuality))
+        {
+            OnVerbUsing(uid, args.User, args.Used, comp.SliceTime.Seconds, comp.ToolQuality);
```
Fildrance: why is it called OnVerbUsing, we are not using verb here, for example, we are just interacting... misleading!

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+    private void AfterSlicing(Entity<SliceableComponent> ent, ref TrySliceEvent args)
+    {
+        var hasBody = TryComp<BodyComponent>(ent, out var body);
+
+        // only show a big popup when butchering living things.
+        var popupType = PopupType.Small;
+        if (hasBody)
+            popupType = PopupType.LargeCaution;
+
+        _popup.PopupPredicted(Loc.GetString("slice-butchered-success", ("target", ent.Owner), ("tool", args.Used!.Value)),
```
Fildrance: please bring order onto line breaks. Might be worth to extract method for that
also dont use null suppression.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+            Loc.GetString("slice-butchered-success-others", ("user", args.User), ("target", ent.Owner), ("tool", args.Used!.Value)),
```
Fildrance: extract variable for that

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        if (TrySlice(ent))
+        {
+            if (hasBody)
+                _body.GibBody(ent, body: body);
+            else
+                PredictedQueueDel(ent.Owner);
+
+            var ev = new SliceEvent();
+            RaiseLocalEvent(ent, ref ev);
```
Fildrance: you raise event on entity that was queued for delete - wont it be possibly problematic? maybe we should delete it after we raise event?

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+    {
+        if (!Resolve(uid, ref comp, ref transform))
+            return false;
+
+        var slices = EntitySpawnCollection.GetSpawns(comp.Slices);
+
+        foreach (var sliceProto in slices)
+        {
+            var sliceUid = PredictedSpawnNextToOrDrop(sliceProto, uid);
+            _transform.SetLocalRotation(sliceUid, 0);
```
Fildrance: why?

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+
+            // Called only on server because random unpredicted
+            if (!_container.IsEntityOrParentInContainer(sliceUid) && _net.IsServer)
+            {
+                var randVect = _random.NextVector2(2.0f, 2.5f);
```
Fildrance: slarti developed hash-based thing that can help with that - look at usages of SharedRandomExtensions.HashCodeCombine
But needs testing ofc

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,217 @@
...
+                _physics.SetLinearVelocity(sliceUid, randVect);
+            }
+
+            // Fills new slice if comp allows.
+            if (Resolve(uid, ref food) && comp.TransferSolution)
```
Fildrance: switch places - check for bool is cheaper then resolve, better check it first

Fildrance (changes requested):
Seems mostly fine

Content.Shared/Sliceable/SliceableSystem.cs
kosticia: Pretty funny that I started used this construction after I saw it in the ArtifactNodeScannerSystem. Its actually pretty usable.

Content.Shared/Sliceable/SliceableSystem.cs
kosticia: > please bring order onto line breaks.

What?

Content.Shared/Sliceable/SliceableSystem.cs
kosticia: iirc there was an issue that spawned there enities were rotated

Content.Shared/Sliceable/SliceableSystem.cs
kosticia: Doesn't works here properly because results of random vectors almost same for every slice.

Winkarst-cpu:
Sorry for merge conflicts.

beck-thompson:
I review if you fix conflicts

kosticia:
@beck-thompson 


Now fully ready for review

Fildrance:
I see no answers about why you changed some certain numbers in yaml related to item drops. Have you seen first half of my comments?

Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
kosticia: Butcherable now only for spike

kosticia:
> I see no answers about why you changed some certain numbers in yaml related to item drops. Have you seen first half of my comments?

most of those changes already reverted. I'm probably just forgot to resolve comments

kosticia:
Or not... Welp...

Fildrance:
https://github.com/user-attachments/assets/16d42f7f-9e18-4067-86f6-50ca96ad847b

Found this sad bug while testing - pizza was fine but 'tower-cap log' and 'steel-cap log' specifically are broken for some reason.

Fildrance:
Breaking DoAfter still leads to entity being butchered on client and having bunch of errors outputted
image

Content.Shared/Kitchen/Components/KitchenSpikeComponent.cs
```diff
@@ -28,6 +30,12 @@ public sealed partial class KitchenSpikeComponent : Component
     /// </summary>
     private static readonly ProtoId<SoundCollectionPrototype> DefaultSpikeButcher = new("SpikeButcher");
 
+    /// <summary>
+    /// Tool quality that required for the entity that used for butchering.
+    /// </summary>
+    [DataField(customTypeSerializer: typeof(PrototypeIdSerializer<ToolQualityPrototype>))]
+    public string RequiredToolQuality = "Slicing";
```
Fildrance: why not ProtoId

Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
Fildrance: Tested and it does not seem so - any idea why?

Fildrance (changes requested):
bugs are sad, code nitpicks are minor, but i will come back again when bugs are fixed :<

Content.Shared/Kitchen/SharedKitchenSpikeSystem.cs
```diff
@@ -90,7 +95,7 @@ private void OnInit(Entity<KitchenSpikeComponent> ent, ref ComponentInit args)
 
     private void OnInsertAttempt(Entity<KitchenSpikeComponent> ent, ref ContainerIsInsertingAttemptEvent args)
     {
-        if (args.Cancelled || TryComp<ButcherableComponent>(args.EntityUid, out var butcherable) && butcherable.Type == ButcheringType.Spike)
+        if (args.Cancelled)
```
Fildrance: that actually leads to any interaction with butcher spike to be cancelled.

Fildrance:
This refactor would lead to most non-spike related butchering now being 1 sec long. As all humanoids should get on spike to be butchered that looks not too bad, but should be considered carefully...

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,224 @@
...
+            args.User, args.User, popupType);
+
+        if (TrySlice(ent, args.User))
+        {
+            var ev = new SliceEvent();
+            RaiseLocalEvent(ent, ref ev);
+
+            var giblings = _body.GibBody(ent);
+            if(giblings.Count == 0)
+                _destructible.DestroyEntity(ent);
```
Fildrance: doing gib and then destroy sounds like not a good idea too, because gib queue-dels on server-side, which could probably lead to having event subs on destroy that will get entities under deletion. But it workes somehow. Yeah, this is kinda mess...

Content.Shared/Sliceable/SliceableSystem.cs
Fildrance: Consistency in line breaks is key to readability - having it in one file at least is important

Content.Server/Nutrition/EntitySystems/SliceableFoodSystem.cs
```diff
@@ -1,170 +0,0 @@
...
-        SubscribeLocalEvent<SliceableFoodComponent, SliceFoodDoAfterEvent>(OnSlicedoAfter);
-        SubscribeLocalEvent<SliceableFoodComponent, ComponentStartup>(OnComponentStartup);
-    }
-
-    private void OnInteractUsing(Entity<SliceableFoodComponent> entity, ref InteractUsingEvent args)
-    {
-        if (args.Handled)
-            return;
-
-        if (!TryComp<UtensilComponent>(args.Used, out var utensil) || (utensil.Types & UtensilType.Knife) == 0)
```
Fildrance, line 45: UtensilType.Knife is now going to be useless  :|
Its that or think about how we incorporate it into new system

Fildrance:
Chopping trees now requires sawing tool, same for slicing logs. Thank gowd we already had it - i was not a fan of creating new tool quality just for that.

Fildrance (approved):
This seems functional replicates balance we had previosuly to some extent.
I need to update breaking changes tab properly now, but all in all - needs another pair of eyes, and specially for yaml.
There were 1 case where i saw that ProtoId of slice is different in before and after, so be extra careful when looking at it.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,219 @@
...
+            _transform.SetLocalRotation(sliceUid, 0);
+
+            if (slices.Count != 0 && !_container.IsEntityOrParentInContainer(sliceUid))
+            {
+                var randVect = rng.NextPolarVector2(2.0f, 2.5f);
+                _physics.SetLinearVelocity(sliceUid, randVect);
+            }
+
+            // Fills new slice if comp allows.
+            if (ent.Comp.TransferSolution && TryComp<EdibleComponent>(ent, out var edible))
```
Fildrance: Tightly couping to Edible here sounds a bit questionable, to be quite honest. I am not sure about the best way to resolve this. Putting soltuion name into SliceableComponent could be one, but that would also be quite a mess!

Content.Shared/Sliceable/SliceableComponent.cs
```diff
@@ -0,0 +1,46 @@
...
+/// and can provide newly created entities as a result.
+/// </summary>
+[RegisterComponent, Access(typeof(SliceableSystem))]
+public sealed partial class SliceableComponent : Component
+{
+    /// <summary>
+    /// Prototype ID of the entity that will be spawned after slicing.
+    /// </summary>
+    [DataField]
+    public List<EntitySpawnEntry> Slices = [];
```
iaada: `List<EntitySpawnEntry>` has been replaced by EntityTable except for specific use cases like `ButcherableComponent`.

Resources/Prototypes/Entities/Objects/Consumable/Food/produce.yml
```diff
@@ -224,9 +229,9 @@
         Blunt: 12
   - type: Produce
     seedId: steelcap
-  - type: Log
-    spawnedPrototype: SheetSteel1
-    spawnCount: 1
+  - type: Sliceable
```
iaada: Steel-cap logs might as well be sawing too to be consistent with its predecessor.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Snipers/snipers.yml
```diff
@@ -83,8 +83,9 @@
     solution: melee
   - type: SolutionTransfer
     maxTransferAmount: 2
-  - type: Sharp
-    butcherDelayModifier: 2 #It's a bayonet
+  - type: Tool
+    qualities: Slicing
+    speedModifier: 2 #It's a bayonet
```
iaada: Should be 0.5. `Sharp` multiplied, but `Tool` divides.

Resources/Prototypes/Entities/Clothing/Shoes/specific.yml
```diff
@@ -25,7 +25,7 @@
           - ToySidearm
         blacklist:
           components:
-          - Sharp
+          - Tool
```
iaada: There is no knife that fits in the shoes now because the plastic knife became a tool. I agree with that change, but to keep parity, can you change the whitelist to `ToyKnife` and add the tag to the plastic knife.

Content.Shared/Nutrition/Components/ButcherableComponent.cs
iaada, line 1: Can you append to the component description that this component is now specifically for the meathook.

Resources/Prototypes/Entities/Mobs/NPCs/regalrat.yml
```diff
@@ -126,7 +126,8 @@
   - type: Grammar
     attributes:
       gender: male
-  - type: Sharp
+  - type: Tool
+    qualities: Slicing # Should work 🙏🙏🙏🙏
```
iaada: This does not work, probably because of problems with displaying the verb.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+    [Dependency] private readonly IGameTiming _gameTiming = default!;
+
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<SliceableComponent, TrySliceDoAfterEvent>(AfterSlicing);
+        SubscribeLocalEvent<SliceableComponent, GetVerbsEvent<InteractionVerb>>(AddSliceVerb);
+        SubscribeLocalEvent<SliceableComponent, InteractUsingEvent>(OnInteraction);
+    }
+
+    private void OnInteraction(EntityUid uid, SliceableComponent comp, InteractUsingEvent args)
```
iaada: Every subscription should be using the modern declaration.
```suggestion
    private void OnInteraction(Entity<SliceableComponent> ent, ref InteractUsingEvent args)
```

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+    {
+        if (args.Handled)
+            return;
+
+        if (_toolSystem.HasQuality(args.Used, comp.ToolQuality))
+        {
+            args.Handled = true;
+            CreateDoAfter(uid, args.User, args.Used, comp.SliceTime.Seconds, comp.ToolQuality);
+        }
+    }
```
iaada:
```suggestion
    {
        if (args.Handled || !_toolSystem.HasQuality(args.Used, comp.ToolQuality))
            return;

          args.Handled = true;
          CreateDoAfter(uid, args.User, args.Used, comp.SliceTime.Seconds, comp.ToolQuality);
        }
    }
```

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+            Message = verbMessage ?? Loc.GetString("slice-verb-message-default"),
+            Act = () =>
+            {
+                CreateDoAfter(uid, user, tool.Value, comp.SliceTime.Seconds, comp.ToolQuality);
+            },
+        };
+        args.Verbs.Add(verb);
+    }
+
+    private void CreateDoAfter(EntityUid uid, EntityUid user, EntityUid used, float time, string qualities)
```
iaada: Every non-subscription method should have documentation.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    private void AddSliceVerb(EntityUid uid, SliceableComponent comp, GetVerbsEvent<InteractionVerb> args)
+    {
+        if (!args.CanInteract || !args.CanAccess)
+            return;
+
+        var user = args.User;
+        if (!TryGetTool(uid, comp, args, user, out var verbDisabled, out var verbMessage, out var tool))
+            return;
```
iaada: This method is causing problems in displaying the verb and lost functionality from `SharpSystem`. Previously if an item could be butchered the verb was always visible but disabled. If you held the right tool (or *were* the right tool) the verb would be enabled. Currently this method *always* returns false and so the verb is never displayed.

Fundamentally this method is doing too much. Whether to display the verb and getting the tool should be two separate things.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+            ("tool", args.Used.Value)
+        );
+        _popup.PopupPredicted(slicingDoneMessageForUser, slicingDoneMessageForOthers, args.User, args.User, popupType);
+        if (TrySlice(ent, args.User))
+        {
+            var ev = new SliceEvent();
+            RaiseLocalEvent(ent, ref ev);
+
+            _body.GibBody(ent);
+            _destructible.DestroyEntity(ent);
```
iaada: Why both? Gibbing also queues the entity for deletion.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+    private bool TrySlice(Entity<SliceableComponent> ent, EntityUid user)
+    {
+        var slices = EntitySpawnCollection.GetSpawns(ent.Comp.Slices);
+
+        var rndSeed = SharedRandomExtensions.HashCodeCombine((int) _gameTiming.CurTick.Value, user.Id, ent.Owner.Id);
```
iaada: Leave a TODO here to switch to the proper implementation when the engine PR is merge.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+        var rng = new System.Random(rndSeed);
+
+        (Entity<SolutionComponent> sourceSoln, Solution sourceSolution)? solutionInfo = default;
+        if (ent.Comp.TransferSolution && !TryGetSourceSolutionForTransfer(ent, out solutionInfo))
+            return false;
+
+        foreach (var sliceProtoId in slices)
+        {
+            var sliceUid = PredictedSpawnNextToOrDrop(sliceProtoId, ent);
+            _transform.SetLocalRotation(sliceUid, 0);
```
iaada: This issue was resolved in engine. Setting the rotation is unnecessary.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+        var ev = new BeforeFullySlicedEvent
+        {
+            User = user
+        };
+        RaiseLocalEvent(ent, ev);
+
+        return true;
+    }
+
+    private bool TryGetTool(
```
iaada: This method is causing problems in displaying the verb and lost functionality from `SharpSystem`. Previously if an item could be butchered the verb was always visible but disabled. If you held the right tool (or *were* the right tool) the verb would be enabled. Currently this method *always* returns false and so the verb is never displayed.

Fundamentally this method should be about whether the player should see the verb, and getting the tool should be second.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+
+    private bool TrySlice(Entity<SliceableComponent> ent, EntityUid user)
+    {
+        var slices = EntitySpawnCollection.GetSpawns(ent.Comp.Slices);
+
+        var rndSeed = SharedRandomExtensions.HashCodeCombine((int) _gameTiming.CurTick.Value, user.Id, ent.Owner.Id);
+        var rng = new System.Random(rndSeed);
+
+        (Entity<SolutionComponent> sourceSoln, Solution sourceSolution)? solutionInfo = default;
+        if (ent.Comp.TransferSolution && !TryGetSourceSolutionForTransfer(ent, out solutionInfo))
```
iaada: This is really sus'ing me out. The solution transfer stuff does not work. I do want to see it, but the whole implementation needs to be rethought to have the container name in the component and for the remaining solution to have more authority on how many slices are spawned. It's second PR territory imo.

Image

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,256 @@
...
+                var (sourceSoln, sourceSolution) = solutionInfo.Value;
+                var sliceVolume = sourceSolution.Volume / FixedPoint2.New(slices.Count);
+
+                var lostSolution = _solutionContainer.SplitSolution(sourceSoln, sliceVolume);
+                FillSlice(sliceUid, lostSolution);
+            }
+        }
+
+        var transform = Transform(ent);
+        _audio.PlayPredicted(ent.Comp.Sound, transform.Coordinates, user, AudioParams.Default.WithVolume(-2));
```
iaada:
```suggestion
        _audio.PlayPredicted(ent.Comp.Sound, Transform(ent).Coordinates, user, AudioParams.Default.WithVolume(-2));
```

iaada (changes requested):

iaada (changes requested):
After thinking on it more these systems should be merged into `ToolRefineableSystem`. `SliceableFoodComponent` is significantly different because of the solution transfer and should remain as a separate `ToolRefineableSolutionComponent`. `ToolRefineable` should be updated to EntityTable.

Content.Shared/Sliceable/SliceableComponent.cs
Fildrance: Are you sure it would be good idea to make this change in this PR? Because this will lead to much harder to read diff in this hell of a mess PR... i can propose to make move to EntityTable in separate PR :< i can, for sure, change comps to ToolRefinable - there must be no problem with that...

Content.Shared/Sliceable/SliceableSystem.cs
Fildrance: Well, because you cannot gib... logs? :( gib works only with BodyComponent marked entities, sadly.

Content.Shared/Sliceable/SliceableSystem.cs
Fildrance: Changed other method in system in same way too.

Content.Shared/Sliceable/SliceableSystem.cs
Fildrance: I think i fixed it (showing verb), without changing code too much. It should work like old code.

Content.Shared/Sliceable/SliceableSystem.cs
Fildrance: I fixed method outputs and naming - its more of a getting verb data, yes, so it is 'TryGetVerbDataAndTool'. And getting tool is a side-activity of a kind. Splitting the thing into different parts would inevitably lead to code duplication, sadly. At least i see it this way.

Fildrance:
I replaced direct dependency to FoodComponent with name of solution ('food') hard-coded into SliceableComponent. This is not ideal, but that will lead to minimal amount of additional yaml changes. I am not 100% sure that is acceptable to be perfectly honest... Ideally we would split chemicals from mobs, for example, by drawing as much 'extra' as we can from their bloodstream, but doing it in this PR looks a bit too much.

Resources/Prototypes/Entities/Clothing/Shoes/specific.yml
Fildrance: should it be ToyKnife or are we going to be okay with re-using ToySidearm?

Fildrance:
> After thinking on it more these systems should be merged into `ToolRefineableSystem`. `SliceableFoodComponent` is significantly different because of the solution transfer and should remain as a separate `ToolRefineableSolutionComponent`. `ToolRefineable` should be updated to EntityTable.

you mean, that all basic split logic will go into ToolRefineableComponent, and extra-data (names of solutions from and to which we need to move) should go into ToolRefineableSolutionComponent?

That would need Sound added to ToolRefineableComponent... also it have fuel variables, which is a bit odd for our example, and default values are not zero. What about events? ToolRefinablSystem have only WelderRefineDoAfterEvent, which is pretty specific, sadly... while our system have TrySliceDoAfterEvent (why is it Try? a bit odd, but it was this way so not sure) and SliceEvent.

Also there would be a bit of a pain, related to verb. Are we sure we want all of this in this PR? It sounds alluring, i am not gonna lie.

Content.Shared/Sliceable/SliceableComponent.cs
iaada: It's something that could be done in a second PR I suppose.

Resources/Prototypes/Entities/Clothing/Shoes/specific.yml
iaada: ToySidearm is fine.

Content.Shared/Sliceable/SliceableSystem.cs
iaada: Right I guess that makes sense. Body code stuff...

iaada:
I really think the solution stuff is too specific to try and fit it all into one component. I also feel like while we're in the business of consolidating all these components might as well finish the job.

But maybe instead of `ToolRefinableComponent` we use the pseudo-refactor `SimpleToolUsageComponent`? That handles all of the verb stuff and the using of a tool on the entity. Then `NewToolRefinableComponent` (whatever we end up on) can work by subscribing to `SimpleToolDoAfterEvent` to spawn the items. It can then create `RefinedEvent` to substitute what `SliceEvent` is doing.

Although sound would still need to be added (probably as its own component), and I'm not sure about how to fix the fuel usage from `ToolRefinable`. I suppose the simple answer would be `ConsumeFuelOnToolUsageComponent` subscribed to `SimpleToolDoAfterEvent`, but it would lose the `DoAfterArgs.AttemptFrequency` canceling the doafter when you are low on fuel. I don't think `SimpleToolUsageSystem` will escape changes, particularly `NeedHand = tool != user` from `UseTool` seems necessary.

Fildrance:
> SimpleToolUsageComponent

Sadly, not exactly - SimpleToolUsage have only one tooltip to show when verb is disabled, while we change it based on our attempt to get tool logic.

Fildrance:
@iaada could you, please, propose some scheme by which i should split SliceableComponent fields into existing components? I'm a bit hesitant about next step :/ partially due to amount of yaml changes and need for another sister-component...

iaada:
Assuming the starting point is `ToolRefinableComponent`, I picture a `ToolRefinableSolutionComponent` with it being clearly marked as requiring the first component and containing both solution datafields. Then instead of `ToolRefinableComponent.SolutionToSplit != null` you use a TryComp for `ToolRefinableSolutionComponent` to go down the path for solution splitting. It should log a warning if it fails to find a solution on the spawn result.

The problem is that you need to know all of the items getting created to properly split the solution, so I struggle to think of a way to do it that isn't basically how it's already done. Though another way to do it could be to have the list of `EntityUid`s in the `RefinedEvent` called on the entity right before destruction, and sub to it with `ToolRefinableSolutionComponent` to do the solution management there. That might be better in case other components want to do something to the results after an item gets refined. DNA is the only thing that comes to mind, but there might be other uses for such an event.

Fildrance:
@iaada can you please look at changes in ToolRefinableSystem? I am not finished but futher changes with yaml will require A LOT of yaml digging, so i'd like to get an OK before i start ;( basically look at https://github.com/space-wizards/space-station-14/pull/36895/commits/8c82b3de391c9eaaec109a7fe18eef82bfc5e0ea

I saw no opportunity to add logic to SimpleToolUsageSystem, sadly.
Also i had to leave SliceableSystem with 1 simple event handler (due to one of conditions bound to MobStateComponent).

Content.Shared/Sliceable/SliceableSystem.cs
iaada, line 1: A whole system to manage one mobstate subscription should probably just go in MobStateSystem.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+    [Dependency] private readonly SharedDestructibleSystem _destructible = default!;
+    [Dependency] private readonly IGameTiming _gameTiming = default!;
 
     public override void Initialize()
     {
         base.Initialize();
+
+        SubscribeLocalEvent<ToolRefinableComponent, GetVerbsEvent<InteractionVerb>>(AddVerb);
         SubscribeLocalEvent<ToolRefinableComponent, InteractUsingEvent>(OnInteractUsing);
         SubscribeLocalEvent<ToolRefinableComponent, WelderRefineDoAfterEvent>(OnDoAfter);
```
iaada: This event should be renamed to `ToolRefineDoAfterEvent` to avoid confusion.

Content.Shared/Sliceable/SliceableSystem.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    /// <inheritdoc />
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<MobStateComponent, GetIsToolRefineBlockedEvent>(GetIsRefinable);
+    }
+
+    private void GetIsRefinable(Entity<MobStateComponent> ent, ref GetIsToolRefineBlockedEvent args)
+    {
+        if (args.RequiredToolQuality == SlicingToolQuality && !_mob.IsDead(ent, ent))
```
iaada: I don't think it needs to check the tool quality here. This is ignoring `ToolRefinableComponent.QualityNeeded` and hardcoding slicing for all mobs. The tool quality should be checked by `ToolRefinableSystem` before sending this event.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+            fuel: component.RefineFuel
+        );
+    }
+}
+
+/// <summary>
+/// Event for checking if tool refining of entity is blocked in some complex way.
+/// </summary>
+[ByRefEvent]
+public record struct GetIsToolRefineBlockedEvent(
```
iaada: The convention for events like this tends to be calling them "attempt" like `ToolRefineAttemptEvent` and using `args.Cancelled` for the bool.

I also don't think it should have the tool quality as a param. In theory nobody else should care about the tool quality except us and ToolSystem.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+    ProtoId<ToolQualityPrototype> RequiredToolQuality,
+    bool IsRefinable = true,
+    string? BlockCause = null
+);
+
+/// <summary>
+/// Called after slicing of the entity.
+/// </summary>
+[ByRefEvent]
+public record struct BeforeToolRefineFinishedEvent(EntityUid User)
```
iaada: I think this should use the typical naming scheme of `AfterToolRefinedEvent`, with a better summary about when this is raised, maybe what to do with it. Specifically it's not clear what `Cancelled` is cancelling without going all the way to where the event is raised.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+                    FillResult(spawnedUid, comp.SolutionToSet, lostSolution);
+                }
+            }
+        }
+
+        if (component.Sound != null)
+            _audio.PlayPredicted(component.Sound, Transform(uid).Coordinates, args.User, AudioParams.Default.WithVolume(-2));
+
+        var ev = new BeforeToolRefineFinishedEvent(args.User);
+        RaiseLocalEvent(uid, ref ev);
```
iaada: What do you think about playing the sound being done in another component. Something like `ToolRefinableSoundComponent` subscribed to `AfterToolRefinedEvent`?

Another way to use this event could be to handle `ToolRefinableSolutionComponent`. If the event had the list of entities spawned, you could do the solution splitting in an event subscription. (And another way to use the list of entities in the event could be transferring DNA.)

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+
+            var lostSolutionPart = solution.SplitSolution(itsSolution.AvailableVolume);
+            _solutionContainer.TryAddSolution(itsSoln.Value, lostSolutionPart);
+        }
+    }
+
+    private void AddVerb(Entity<ToolRefinableComponent> ent, ref GetVerbsEvent<InteractionVerb> args)
+    {
+        var used = args.Using;
+        if (ent.Comp.VerbText == null || !args.CanInteract || !args.CanAccess || !used.HasValue)
```
iaada: The `!used.HasValue` check is invalidating your `if (args.Using is not null)` later on. IDE wasn't picking it up because of renaming `args.Using` to `used`, but not being consistent.

Content.Shared/Tools/Components/ToolRefinableComponent.cs
```diff
@@ -35,4 +37,56 @@ public sealed partial class ToolRefinableComponent : Component
...
+    /// </summary>
+    [DataField]
+    public LocId? PopupForUser;
+
+    /// <summary>
+    /// Popup text to display for players near action performer
+    /// on action performed (end of DoAfter).
+    /// </summary>
+    [DataField]
+    public LocId? PopupForOther;
```
iaada, line 85: I feel like all these should be given generic default values. A dev should have to opt-in to making the action quiet on their component. It also might cut down on yml changes.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+    private void OnDoAfter(Entity<ToolRefinableComponent> ent, ref WelderRefineDoAfterEvent args)
     {
-        if (args.Cancelled)
+        if (args.Cancelled || args.Used == null)
+            return;
+
+        var (uid, component) = ent;
+
+        var getIsBlocked = new GetIsToolRefineBlockedEvent(args.Used.Value, component.QualityNeeded);
+        RaiseLocalEvent(ref getIsBlocked);
```
iaada: If it wasn't for the cancelled popup I'd say to use the built in attempt event from DoAfters `DoAfterAttemptEvent<WelderRefineDoAfterEvent>`. Though it can't handle the popup, so it's probably fine.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
 
-        args.Handled = _toolSystem.UseTool(
-            args.Used,
-            args.User,
-            uid,
-            component.RefineTime,
-            component.QualityNeeded,
-            new WelderRefineDoAfterEvent(),
-            fuel: component.RefineFuel);
+        var (uid, component) = ent;
```
iaada: According to Chzblz, Electro says that doing this is not the best. Wish I knew more. https://discord.com/channels/310555209753690112/770682801607278632/1429086789112561745

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+            tool = used;
+        }
+        else if (_toolSystem.HasQuality(user, comp.QualityNeeded))
+        {
+            tool = user;
+        }
+        else
+        {
+            return;
+        }
```
iaada:
```suggestion
        var tool = used ?? user;
        var verbDisabled = false;
        string? verbMessage = null;
        
        if (!_toolSystem.HasQuality(tool, ent.Comp.QualityNeeded))
        {
            verbDisabled = true;
            verbMessage = ent.Comp.ToolMissingQualityTooltip == null
                ? null
                : Loc.GetString(ent.Comp.ToolMissingQualityTooltip, ("target", ent.Owner));
        }
```

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
iaada, line 1: The structure could use a little work. The event subscriptions should all be together, in the order they're in the initialize, and preferably with a region around them too.

I also feel there's a general lack of comments. Particularly there's no summaries for any of the helper methods, the system itself needs a summary, at at least something for in-lines.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+
+        PopupOnRefine(uid, component, args.Used.Value, args.User);
+
+        // TODO: Use RandomPredicted https://github.com/space-wizards/RobustToolbox/pull/5849
+        var rndSeed = SharedRandomExtensions.HashCodeCombine((int)_gameTiming.CurTick.Value, args.User.Id, uid.Id);
+        var rng = new System.Random(rndSeed);
+
+        var spawns = EntitySpawnCollection.GetSpawns(component.RefineResult, rng);
+        List<EntityUid> spawned = new(spawns.Count);
```
iaada:
```suggestion
        var spawned = new List<EntityUid>(spawns.Count);
```

Content.Shared/Tools/Components/ToolRefinableComponent.cs
```diff
@@ -35,4 +37,56 @@ public sealed partial class ToolRefinableComponent : Component
...
+
+    /// <summary>
+    /// Is usage of 'caution' type popup required for <see cref="PopupForUser"/> and <see cref="PopupForOther"/>,
+    /// or small popup will suffice.
+    /// </summary>
+    [DataField]
+    public bool IsUsingCautionPopup;
```
iaada: Instead of a bool I think you could just let them decide. 
```suggestion
    public PopupType PopupType = PopupType.Small;
```

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
             return;
 
-        var xform = Transform(uid);
-        var spawns = EntitySpawnCollection.GetSpawns(component.RefineResult, _random);
-        foreach (var spawn in spawns)
+        _body.GibBody(uid);
+        _destructible.DestroyEntity(uid);
+    }
+
+    private void PopupOnRefine(EntityUid uid, ToolRefinableComponent component, EntityUid used, EntityUid user)
```
iaada: Typically shouldn't separate them when they can tuple.

```suggestion
    private void PopupOnRefine(Entity<ToolRefinableComponent> ent, EntityUid used, EntityUid user)
```

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+                ("tool", used)
+            );
+        var slicingDoneMessageForOthers = component.PopupForOther == null
+            ? null
+            : Loc.GetString(
+                component.PopupForOther,
+                ("user", user),
+                ("target", uid),
+                ("tool", used)
+            );
```
iaada: These aren't so large that they need extra lines.

```suggestion
        var slicingDoneMessageForUser = component.PopupForUser == null
            ? null
            : Loc.GetString(component.PopupForUser, ("target", uid), ("tool", used));
        var slicingDoneMessageForOthers = component.PopupForOther == null
            ? null
            : Loc.GetString(component.PopupForOther, ("user", user), ("target", uid), ("tool", used));
```

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
+        string solutionName,
+        [NotNullWhen(true)] out (Entity<SolutionComponent> sourceSoln, Solution sourceSolution)? solutionInfo
+    )
+    {
+        solutionInfo = default;
+
+        if (!_solutionContainer.TryGetSolution(source, solutionName, out var sourceSoln, out var sourceSolution))
         {
-            SpawnNextToOrDrop(spawn, uid, xform);
+            return false;
```
iaada, line 239: Don't need brackets.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,274 @@
...
         }
 
-        Del(uid);
+        solutionInfo = (sourceSoln.Value, sourceSolution);
+        return true;
     }
+
+    private void FillResult(EntityUid sliceUid, string solutionName, Solution solution)
+    {
+        if (_solutionContainer.TryGetSolution(sliceUid, solutionName, out var itsSoln, out var itsSolution))
```
iaada:
```suggestion
        if (!_solutionContainer.TryGetSolution(sliceUid, solutionName, out var itsSoln, out var itsSolution))
        return false;
```

Resources/Locale/en-US/sliceable/sliceable.ftl
iaada, line 1: This should be tool-refinable.ftl, maybe with a comment to soft group slicing.

Resources/Prototypes/Entities/Objects/base_item.yml
```diff
@@ -170,3 +170,26 @@
         cell_slot:
           name: power-cell-slot-component-slot-name-default
           startingItem: PowerCellHigh
+
+- type: entity
+  id: SlicingRefinableBase
+  abstract: true
+  components:
```
iaada, line 177: I'm not sure about this, but I'm also not against it. Either way [abstract goes before id](https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#yaml-conventions), and `Base` should go before `SlicingRefinable`.

iaada (changes requested):
Gave a full review. The components look ready to go ahead with yml. I still think `ToolRefinableComponent.RefineResult` should be an `EntityTableSelector` instead of a `List<EntitySpawnEntry>`. Though it would be even more yml changes, so I can settle with it being an issue posted after this PR is finished.

Content.Shared/Sliceable/SliceableSystem.cs
Fildrance: It ensures blocking slicing for NON DEAD mobs explicitly, and it is ugly, but having MobStateSystem references from ToolRefinable, or preventing any tool refine action based on mob being not dead - sounds odd too!

Content.Shared/Tools/Components/ToolRefinableComponent.cs
Fildrance: generic defaults are null - people never got problems with this, i honestly dont think it would be good change to have spam of popups or verbs which people were not missing :(
All this stuff is used by only one subtype of refining - slicing, and none other (AS OF NOW).

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
Fildrance: It was said that it would be problem when using arch - to be honest, i am not sure there is any work on this anymore, since the only one working on this was Sloth. 
But yeah, i would like to know more about why this would be a problem too ;( Deconstructor is just a method with out params, how this can prevent stuff...

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
Fildrance: but thats more symbols! :(

Content.Shared/Sliceable/SliceableSystem.cs
iaada: I think the `IsDead` check is fine, it's specifically checking the tool here that feels wrong. The tool should have already been checked in `ToolRefinableSystem` or `ToolSystem`. We don't need to hardcode slicing as the only way to refine a mob.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
iaada: Far as I'm aware the convention is to always instantiate local variables with `var` whenever possible.

Content.Shared/Sliceable/SliceableSystem.cs
Fildrance: I see that MobStateSystem.Subscribers.cs which fits into it greatly!

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
Fildrance: Code inside uses them separately, so by using Entity<> we introduce a tiny bit more garbage :| a thing to think about.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
Fildrance: Agreed that it should be BeforeToolRefinedEvent and it should be called before spawning and splitting solution

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
Fildrance: Agreed upon not touching it because that would be not really useful type of granularity - 
1. event just for 1 method call
2. we dont really want refining WITHOUT any sound (would be poor design) - even tho we cannot set any good default for action

Fildrance:
As a side idea of how to lessen amount of fields for new component - tool actually have UseSound field, which could probably be used, but there are edge cases - hatchet should both chop logs and cut buns, and that should use different sounds ideally, i suppose.

iaada:
In the future we could probably integrate this with tools system even more closely for the sound and verb stuff, but that's going to take refactoring of tools themselves which is definitely out of scope.

Resources/Prototypes/Entities/Objects/Consumable/Food/meat.yml
```diff
@@ -833,10 +846,11 @@
         Quantity: 320
       - ReagentId: Fat
         Quantity: 80
-  - type: SliceableFood
-    count: 16
-    sliceTime: 5
-    slice: FoodMeat #That's... So much meat
+  - type: ToolRefinable
+    refineTime: 16
```
iaada: Previous behavior was 5 seconds.

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,284 @@
...
+            verbDisabled = true;
+            verbMessage = ent.Comp.ToolMissingQualityTooltip == null
+                ? null
+                : Loc.GetString(ent.Comp.ToolMissingQualityTooltip, ("target", ent.Owner));
+        }
+        // make an attempt to ensure refinement is not blocked.
+        var attemptEvent = new AttemptToolRefineEvent(tool);
+        RaiseLocalEvent(ref attemptEvent);
+
+        if (!attemptEvent.IsCancelled)
```
iaada:
```suggestion
        if (attemptEvent.IsCancelled)
```

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,284 @@
...
+        if (!_toolSystem.HasQuality(tool, ent.Comp.QualityNeeded))
+        {
+            verbDisabled = true;
+            verbMessage = ent.Comp.ToolMissingQualityTooltip == null
+                ? null
+                : Loc.GetString(ent.Comp.ToolMissingQualityTooltip, ("target", ent.Owner));
+        }
+        // make an attempt to ensure refinement is not blocked.
+        var attemptEvent = new AttemptToolRefineEvent(tool);
+        RaiseLocalEvent(ref attemptEvent);
```
iaada: Event needs to be directed.
```suggestion
        RaiseLocalEvent(args.Target, ref attemptEvent);
```

Content.Shared/Mobs/Systems/MobStateSystem.Subscribers.cs
```diff
@@ -140,6 +140,14 @@ private void OnStateEnteredSubscribers(EntityUid target, MobStateComponent compo
         }
     }
 
+    private void OnAttemptToolRefine(Entity<MobStateComponent> ent, ref AttemptToolRefineEvent args)
+    {
+        if (!IsDead(ent, ent))
+        {
+            args = args with { IsCancelled = false, BlockCause = Loc.GetString("refined-slice-verb-target-isnt-dead") };
```
iaada:
```suggestion
            args = args with { IsCancelled = true, BlockCause = Loc.GetString("refined-slice-verb-target-isnt-dead") };
```

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,284 @@
...
-            args.User,
-            uid,
-            component.RefineTime,
-            component.QualityNeeded,
-            new WelderRefineDoAfterEvent(),
-            fuel: component.RefineFuel);
+        var component = ent.Comp;
+        var uid = ent.Owner;
+        var attemptEvent = new AttemptToolRefineEvent(args.Used);
+        RaiseLocalEvent(ref attemptEvent);
```
iaada: Event needs to be directed.

```suggestion
        RaiseLocalEvent(args.Target, ref attemptEvent);
```

Content.Shared/Tools/Systems/ToolRefinableSystem.cs
```diff
@@ -1,55 +1,284 @@
...
+    private void OnDoAfter(Entity<ToolRefinableComponent> ent, ref ToolRefineDoAfterEvent args)
+    {
+        if (args.Cancelled || args.Used == null)
+            return;
+
+        var component = ent.Comp;
+        var uid = ent.Owner;
+
+        var getIsBlocked = new AttemptToolRefineEvent(args.Used.Value);
+        RaiseLocalEvent(ref getIsBlocked);
```
iaada: Event needs to be directed.

```suggestion
        RaiseLocalEvent(args.Target, ref getIsBlocked);
```

Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
```diff
@@ -132,8 +132,8 @@
       reagents:
       - ReagentId: GroundBee
         Quantity: 1
-  - type: Butcherable
-    spawned:
+  - type: ToolRefinable
+    refineResult:
     - id: null # Should give nothing when you butcher it so we set the item id it needs to spawn to null
```
iaada: This causes a resolve failure from attempting to apply a linear velocity to an entity without a `PhysicsComponent`. Either `RefineResult` needs to be nullable, or a different component should be used here.

Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
```diff
@@ -835,8 +831,8 @@
       reagents:
       - ReagentId: UncookedAnimalProteins
         Quantity: 1
-  - type: Butcherable
-    spawned:
+  - type: ToolRefinable
+    refineResult:
     - id: null # Should give nothing when you butcher it so we set the item id it needs to spawn to null
```
iaada: Same problem as bees.

iaada (changes requested):
I caught a couple things, but it's looking almost done!

Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
Fildrance: As was discussed, i added warn and skip for such entites. I see that using null in yml in general is not something out of ordinary, but there is little to no usage of null for such cases (in a list that is marked 'required'). As a viable option - i can see removing 'required' from refineResult and then we wont need any such empty blocks. But at the same time it will still be misleading usage of component - it sounds like it SHOULD produce something as an output, so maybe correct way to fix such cases would be adding some component that would... delete on tool interaction?

Much more confusing thing is that bees have human organs :( but thats a bit out of scope for this PR

SnappingOpossum: A null id should be passed over by GetSpawns and not added to the list, is that not working correctly?

SnappingOpossum: Ah, it's ignored for the IRobustRandom version, not the System.Random one.

iaada (approved):
There's nothing else I can see! Thank you so much for taking over this PR. You're finally free <3

Fildrance (approved):
Yeap

PR #36896, opened 2025-04-24
Added warning when attempting to run RUN_THIS on a zip repo download

VasilisThePikachu:
It keeps happening aaaaaa

This may help


slarticodefast (approved):
Tested, works. Good idea!

slarticodefast:
I think you need to set a relative file path to get it working

PR #30603, opened 2024-08-03
Add insulated atmos gloves

SlamBamActionman:
## About the PR

Adds the Insulated Atmos Gloves, and adds them to the Atmospheric Technician's Locker.

## Why / Balance

Multiple maps (Cluster, Saltern, Bagel) have Combat Gloves mapped in Atmospherics, which they shouldn't: #11365 
While these _should_ be removed, doing so removes a cosmetic alternative for atmos players and forces them use normal insulated gloves.

The sheer prevalence of insulated gloves is a detriment to detective investigation gameplay, as insulated gloves are widely used with no variation in fibers. Adding variation to the available insulated gloves will give detectives more to work with while keeping counterplay options available to antags. The atmos gloves leave the unique "blue insulated fibers" on items touched. 

The gloves are put in the atmospherics locker to make them available to atmos techs without needing to be explicitly mapped, and to keep atmos from stealing gloves from Engineering (which they have to do right now).

Giving atmos their own gloves also lays the groundwork for giving them different functionality compared to normal insuls (e.g. giving them heat resistance to pick up heated objects like heated pipes, a mechanic normal engineers are unlikely to interact with).

## Technical details

Like normal insulated gloves, but bloo.

After this PR, mappers should aim to remove/replace combat gloves mapped in Atmospherics.

## Media

image
image
image

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- add: Added insulated atmos gloves, found in atmospherics lockers.

PR #36907, opened 2025-04-25
Gave combat gloves and mercenary gloves faster strip times.

RedBookcase:
## About the PR
Combat gloves have annoyed me for a while since they say they're shockproof and fireproof but they offer nothing in the way of fire resistance. I was considering how to make them more unique and realized that making them faster at stripping things off of people would give them an edge for the people most commonly using them (Security and NukeOps) without being too insane of a balance change.

Both Combat Gloves and Mercenary Gloves now strip items off players 1.5 seconds faster, in line with standard Syndicate thieving gloves. They are not stealthy however, so targets still get a pop-up warning them that their shit's being stolen and they see the doafter bar as well.

## Why / Balance
Gives these two gloves a unique minor niche beyond just being "Insuls that make security more likely to annoy you".

## Technical details
4 lines of yaml. 3 are for adding striptimereduction to combat gloves, and the 4th is changing the combat gloves description to remove the mention of being fireproof. Merc gloves get the strip time as a result of being parented off combat gloves.

## Requirements

**Changelog**

:cl:
- tweak: Combat Gloves are now more effective at stripping targets.


ArtisticRoomba:
One of the problems I have with this is that Atmos sometimes gets these gloves (they're supposed to be fireproof but you know) so Atmos gets free reduced strip timer. They would have to be unmapped.

RedBookcase:
> One of the problems I have with this is that Atmos sometimes gets these gloves (they're supposed to be fireproof but you know) so Atmos gets free reduced strip timer. They would have to be unmapped.

It's a bit weird atmos gets combat gloves at all imo, them being marked as engineering or security has always seemed weird even if them being fireproof is supposed to be a thing. Atmos has the gear to be completely fireproof without anything in the glove slot so I don't see why they need them in the first place.

LaCumbiaDelCoronavirus:
> It's a bit weird atmos gets combat gloves at all imo, them being marked as engineering or security has always seemed weird even if them being fireproof is supposed to be a thing. Atmos has the gear to be completely fireproof without anything in the glove slot so I don't see why they need them in the first place.

Let em keep it for fun, they get 1 pair at most and that's not even guaranteed on all maps.
I doubt an atmos tech with faster-stripping gloves will cause any severe harm

slarticodefast:
The PR itself is probably fine (a buff to nukies I guess?), but like Roomba says we should unmap the combat gloves from atmos in that case.

TaoNewt:
I like the idea in theory, the differentiation gives them better reason to be contraband which is good and the mechanial advantage gives them value to the players.

Unfortunately there might be a roleplay/meta problem associated with specifically this buff. Many security players frequently attempt take advantage of talk times to grab gear off of crew to check it, fortunately with current strip time you can pretty consistently stop them from doing this. Ive seen people refer to this as meta but ive seen people defend it too, its not a rulebreak but the only reason this works is as the player has their hands on the keyboard and needs to click out of talking to move. Distracting the player OC for game advantage is meta mechanics (getting someone talking for ic distraction makes sense too, but this is what the long doafter differentiates, you wont be so distracted by talking that the other person can walk right up to you and undress you). Im going to reference this as 'soft meta' as its unpoliceable so should be reinforced by gameplay mechanics like it currently is. Im concerned that if a noticable buff is given to security strip times, you remove the games mechanical policing of this and just make the 'soft meta' completely abusable.

The good thing is this is specific to _this_ buff and not other ones that could justify contraband status and give mechanical advantage. For example: you could give them increased dissarming chance, this makes sense for securities role and what gloves do, also it brings a weak mechanic into being usable. You could even give them increased stam damage on push to give security a weak back-up non lethal option. 
I think you had the right idea that the gloves needed something more, but it looks like the solution you picked was the only solution that introduces a bigger problem than it solves.

keronshb:
Sec shouldn't be able to strip people faster.

slarticodefast:
> Sec shouldn't be able to strip people faster.

Sec doesn't have combat gloves. Combat gloves are insulated as well, so giving them to sec would be a bad idea.

shamblestf:
> Sec shouldn't be able to strip people faster.

This would only really apply to the warden and hos, and the warden does it quite often and is always at sec so i dont really see the issue? the average secoff wont have combat gloves ever in a round most of the time

keronshb:
For some reason I thought we merged the combat glove PR.
In that case then this is probably fine.

RedBookcase:
I realized the Ninja Gloves also falsely claimed to give the user fire protection so I've updated their description as well.

EmoGarbage404:
> Sec doesn't have combat gloves. Combat gloves are insulated as well, so giving them to sec would be a bad idea.

@slarticodefast this is kinda incorrect. Both the HOS and the Warden start with either combat gloves directly or with them in their lockers.

slarticodefast:
Ok, but those two are exceptions then.
What I meant was we shouldn't make *all* of sec shock proof, because it completely disables potential traps for them and allows them to hack doors.

KaiserMaus:
Still under discussion?

Errant-4 (approved):

SlamBamActionman (approved):
We'll have to nix the combat gloves from Atmos on the maps where they are mapped - if they need drip, #30603 is still available. 

But otherwise I feel this is an okay way to distinguish them from regular insuls.

slarticodefast (changes requested):
`ThievingComponent` was refactored after this PR was made and it won't work on the current master branch.
The stealthy bool now has a player alert that toggles it, and the strip time reduction is only applied when that toggle is active.
You will need to make this more flexible so that you can only reduce the striptime without the alert.

Errant-4:
> `ThievingComponent` was refactored after this PR was made and it won't work on the current master branch. The stealthy bool now has a player alert that toggles it, and the strip time reduction is only applied when that toggle is active. You will need to make this more flexible so that you can only reducte the striptime without the alert.

~~I tested it and it seemed to change the time. Did I just imagine that~~ Never mind

SlamBamActionman:
This PR requires changes to `ThievingComponent` due to the aforementioned refactor. Giving a new datafield that reduces the time differently for stealthy vs. non-stealthy stripping is desirable. Will you be returning to this PR?

RedBookcase:
> This PR requires changes to `ThievingComponent` due to the aforementioned refactor. Giving a new datafield that reduces the time differently for stealthy vs. non-stealthy stripping is desirable. Will you be returning to this PR?

Not likely ATM no.

PR #36910, opened 2025-04-25
Releasing an under-pressure lockout is now a verb

chromiumboy:
## About the PR

Releasing an under-pressure lockout on a gas vent no longer requires a screwdriver and is instead accomplished using the right-click menu

The relevant sections of the guidebook have been updated to reflect this change

## Why / Balance

It is not communicated to the player through any tooltips that an under-pressure lockout can be released via the use of a screwdriver. The only information that references this mechanism is set quite deep in the guidebook. By moving it to the right-click menu, it will be more discoverable for newer players

This PR is also required for #36124, so that using a screwdriver on a gas vent to release a lockout will not simultaneously change its associated pipe layer (and vice versa)

## Technical details

N/A

## Media

pressure-lockout

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: Gas vent under-pressure lockout releases are now performed using the right-click menu - a screwdriver is no longer required

### DIFF
--- a/Content.Server/Atmos/Piping/Unary/EntitySystems/GasVentPumpSystem.cs
+++ b/Content.Server/Atmos/Piping/Unary/EntitySystems/GasVentPumpSystem.cs
@@ -390,7 +390,7 @@ namespace Content.Server.Atmos.Piping.Unary.EntitySystems
             var v = new Verb
             {
                 Priority = 1,
-                Icon = new SpriteSpecifier.Texture(new("/Textures/Interface/VerbIcons/unlock.svg.192dpi.png")),
+                Icon = new SpriteSpecifier.Texture(new ResPath("/Textures/Interface/VerbIcons/unlock.svg.192dpi.png")),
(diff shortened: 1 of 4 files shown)


LaCumbiaDelCoronavirus:
You should really just be able to click on it with an empty hand to do this too, more convenient!

ArtisticRoomba:
> You should really just be able to click on it with an empty hand to do this too, more convenient!

No, that will add to the problem we are trying to solve.

ArtisticRoomba (approved):
Code looks good to me and in-game test went well.

I agree with the changes in general, one of the biggest complaints I heard from new players was that pressure-lockout was very unintuitive to solve.

I thought about the potential problems this could bring up with people now having wholesale access to overriding pressure lockout, but most of the problems were either irrelevant to the situation at hand or overpowered by something else.

Merging in a few hours if no other maintainer has complaints.

PR #36918, opened 2025-04-25
Delivery random multipliers

ScarKy0:
## About the PR

Deliveries will now have a random multiplier applied to them when they spawn.

## Why / Balance

This is just meant as a simple "test" of multipliers working, as well as a way to show players that mail does actually grant money when opened. (Which apparently wasn't well known!)

This has no effect on cargo earnings as by average the amount is the same, but might end up making cargo prioritize certain deliveries over others.

## Technical details

New component that randomizes a multiplier when it initializes, and then adds it whenever GetDeliveryMultiplier is ran.
A new system for managing gathering multipliers and basic behaviour of modifiers (in the future).

## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Deliveries now have a random multiplier added to them, slightly modifying their earnings.
- tweak: Deliveries now show how many spesos they grant on examine.


Content.Shared/Delivery/DeliveryRandomMultiplier.cs
```diff
@@ -0,0 +1,30 @@
...
+/// Applies a random multiplier to the delivery on init.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class DeliveryRandomMultiplierComponent : Component
+{
+    /// <summary>
+    /// The highest the random multiplier can go.
+    /// </summary>
+    [DataField]
+    public float MaxMultiplier = 0.2f;
```
slarticodefast, line 16: you should make clear that multipliers are applied additively, not multiplicatively
I would call it MaxMultiplierOffset for that and add a comment
same for the other two

Content.Shared/Delivery/SharedDeliverySystem.cs
```diff
@@ -56,10 +56,21 @@ private void OnDeliveryExamine(Entity<DeliveryComponent> ent, ref ExaminedEvent
...
         if (ent.Comp.IsOpened)
         {
-            args.PushText(Loc.GetString("delivery-already-opened-examine"));
+            // Should always show at the top
+            args.PushText(Loc.GetString("delivery-already-opened-examine"), 99);
         }
 
-        args.PushText(Loc.GetString("delivery-recipient-examine", ("recipient", recipientName), ("job", jobTitle)));
+        // Should always show second.
+        args.PushText(Loc.GetString("delivery-recipient-examine", ("recipient", recipientName), ("job", jobTitle)), 98);
```
slarticodefast, line 64: use a push group to make sure they always show up together instead of relying on no other system having these priorities

slarticodefast (changes requested):

Content.Shared/Delivery/DeliveryRandomMultiplier.cs
```diff
@@ -0,0 +1,31 @@
+using Robust.Shared.GameStates;
```
EmoGarbage404, line 1: filename should be `DeliveryRandomMultiplierComponent`

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    private void OnRandomMultiplierMapInit(Entity<DeliveryRandomMultiplierComponent> ent, ref MapInitEvent args)
+    {
+        ent.Comp.CurrentMultiplierOffset = _random.NextFloat(ent.Comp.MinMultiplierOffset, ent.Comp.MaxMultiplierOffset);
+        Dirty(ent);
+    }
+
+    private void OnGetRandomMultiplier(Entity<DeliveryRandomMultiplierComponent> ent, ref GetDeliveryMultiplierEvent args)
+    {
+        args.Multiplier += ent.Comp.CurrentMultiplierOffset;
```
EmoGarbage404, line 28: mixing additive and multiplicative modifiers is kinda suspect. It means that the internal handling order of these events determines some very notable things. 

Your event should probably have a base price, multipliers, and flat modifiers. That way your adjustment can be based off of the base price (and not the price after some arbitrary modifiers) and can apply things in a deterministic order.

With the current implementation, you couldn't set the multiplier to 0 to void the value, as there is a chance that an event handling it afterwards would increment the multiplier.

Resources/Locale/en-US/delivery/delivery-component.ftl
```diff
@@ -1,5 +1,6 @@
 delivery-recipient-examine = This one is meant for {$recipient}, {$job}.
 delivery-already-opened-examine = It was already opened.
+delivery-earnings-examine = Delivering this will earn the station [color=yellow]{$spesos}[/color] spesos.
```
EmoGarbage404, line 3: big approvers from me

image

EmoGarbage404 (dismissed):

PR #36921, opened 2025-04-25
Make 10u of "Atomic Bomb" drink instead of 11u

Kiri-Yoshikage:
## About the PR
Atomic Bomb recipe now gives 10u of cocktail instead of 11. Finally playable.

## Why / Balance
Consistency. All the other "rare" cocktails are made in 10 units,, but not this one for some reason. Also it just feels better to have even numbers.

## Technical details
Changed the folowing line: products: AtomicBomb: 11 to products: AtomicBomb: 10

## Media

Not needed.

## Requirements

## Breaking changes
None?.. If I understand this correctly. If I'm wrong - please tell me what to add here, sorry, first time trying to make a PR.

**Changelog**

:cl:
- tweak: Atomic Bomb cocktail is now mixed in 10u instead of 11.

### DIFF
--- a/Resources/Prototypes/Recipes/Reactions/drinks.yml
+++ b/Resources/Prototypes/Recipes/Reactions/drinks.yml
@@ -90,7 +90,7 @@
     Uranium:
       amount: 1
   products:
-    AtomicBomb: 11
+    AtomicBomb: 10
 
 - type: reaction
   id: B52


Kiri-Yoshikage:
> I'd like to suggest reducing the amount of B52 required, to match. Currently, this turns 11u of ingredients into 10u of cocktail, which is probably fine? But not ideal, IMO.

Other "rare" cocktails such as gildlager and patron are also made of 10u of drink and 1u of material, and they give 10u of cocktail. For some reason atomic bomb is "special" which is really annoying when you actually mix it since there's like 3 units of leftovers

slarticodefast (approved):
Should be fine, even if it does not conserve volume.

MozarteanChaos:
> > I'd like to suggest reducing the amount of B52 required, to match. Currently, this turns 11u of ingredients into 10u of cocktail, which is probably fine? But not ideal, IMO.
> 
> Other "rare" cocktails such as gildlager and patron are also made of 10u of drink and 1u of material, and they give 10u of cocktail. For some reason atomic bomb is "special" which is really annoying when you actually mix it since there's like 3 units of leftovers

Ah, in that case, 's all good.

ArtisticRoomba (approved):
Antoine Lavoisier is rolling in his grave right now

PR #36923, opened 2025-04-25
Add noir glasses

slarticodefast:
## About the PR
Classic SS13 item.
Required equipment for every hard-boiled detective.

## Why / Balance
fun

## Technical details
Add a simple overlay using the greyscale shader.
Use EquipmentHudSystem so that the component works both on clothing items or when directly attached to a player.

## Media

https://github.com/user-attachments/assets/3036318c-4a45-4671-b40c-76547631ef35

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- add: Added noir glasses, which make everything black and white. You can find them in the detective's locker.


IProduceWidgets:
based. I've been wanting these awhile.

If you wanted to the shader for sepia is also pretty easy to make.
Stretch goal though, if you don't I probably will at some point.

slarticodefast:
I was thinking about making it a sin city style shader instead, so it keeps dark red colors, but makes everything else black and white. But I will have to play around with shaders a little more for that. It can be easily adjusted later.

TeenSarlacc:
> *This is me, John Detective. And this is my station, which i am supposed to take care of. But it's... It's not that easy...*

you are obligated to always narrate in the 2nd person whenever you wear these

Hanzdegloker:
question and two statements.
Flash protection?
Players might not like to swallow the pill of no sec hud statuses.
Lastly good work lol.

slarticodefast:
> Players might not like to swallow the pill of no sec hud statuses.

Yeah, I added flash protection, so it's a valid replacement to the sec glasses.
Doesn't have status icons though, having this much style comes at a cost.

YoungThugSS14:
> > Players might not like to swallow the pill of no sec hud statuses.
> 
> Yeah, I added flash protection, so it's a valid replacement to the sec glasses. Doesn't have status icons though, having this much style comes at a cost. It's mostly a gimmick item, so I think that is fine.

TBH detectives shouldn’t be having sec glasses anyways, they aren’t meant to be officers

IProduceWidgets:
If you wanted to add a variant the b&w fits kinda well for librarian flavour too I think.

Content.Shared/Overlays/BlackAndWhiteOverlayComponent.cs
```diff
@@ -0,0 +1,10 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Overlays;
+
+/// <summary>
+/// Makes the entity see everything in black and white by adding an overlay.
+/// When added to a clothing item it will also grant the wearer the same overlay.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class BlackAndWhiteOverlayComponent : Component;
```
EmoGarbage404, line 10: access

EmoGarbage404 (approved):
sleek

Content.Shared/Overlays/BlackAndWhiteOverlayComponent.cs
slarticodefast: It's a marker component. There is nothing in there.

PR #36930, opened 2025-04-25
Revert ringer text entry changes

Centronias:
## About the PR
Small revert to exactly how text is entered in the PDA ringer.

## Why / Balance

For a bit of context, there are broadly two ways to edit your ringer:
1. highlight and replace a note
2. backspace and enter a new note

Prior to #35907, both options were possible. After it, only option 1 works, and option 2 no longer works. This seems to be because an empty text box is "invalid" and reverted to the previous note, as though a bad note were entered.

We use Wizden master changes pretty quickly after they get merged down on [Moffstation](https://github.com/moff-station/moff-station-14), so we ran into this as a useability issue, specifically for traitors who use entry option 2, who reported this as a hard bug which prevented their access to their uplinks.

As far as I can tell, there are no downsides to this revert, and it just works like it used to while keeping all of the predictive benefits of the refactor.

## Technical details

Reverted `RingtoneMenu`'s `input.OnTextChanged` callback changes from #35907 .

## Media

https://github.com/user-attachments/assets/43f15ac9-a5e6-4156-8c49-9a2da0401047

<details>
  <summary>Ringtone behavior prior to #35907</summary>
https://github.com/user-attachments/assets/81a9fcf7-edf8-4ef8-9c7f-2fdde4d9b5ea
</details>

## Requirements

## Breaking changes

Shouldn't be any.

**Changelog**

N/A

### DIFF
--- a/Content.Client/PDA/Ringer/RingtoneMenu.xaml.cs
+++ b/Content.Client/PDA/Ringer/RingtoneMenu.xaml.cs
@@ -50,27 +50,14 @@ namespace Content.Client.PDA.Ringer
                     input.CursorPosition = input.Text.Length; // Resets caret position to the end of the typed input
                 };
 
-                input.OnTextChanged += args =>
+                input.OnTextChanged += _ =>
                 {
-                    // Convert to uppercase
-                    var upperText = args.Text.ToUpper();
+                    input.Text = input.Text.ToUpper();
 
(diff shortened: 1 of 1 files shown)


slarticodefast:
Superceded by https://github.com/space-wizards/space-station-14/pull/36936

PR #36939, opened 2025-04-26
Resolve a warning in SharedAnomalySystem and quell some linter whining.

VerinSenpai:
## About the PR
Replace obsolete grid.GetAnchoredEntities with SharedMapSystem.GetAnchoredEntities call. Also dealt with a few funny yellow squigglies. ~~Project Shut that damn Linter up (only in my dreams)~~ One thing to note here, I didn't do it in this PR, but I believe a good chunk of the methods here can be made private. The linter also implies that one of them can be made static, but I don't know what that implies ~~(C# noob)~~ so I left that alone.

Warning resolved is CS0618

## Requirements

### DIFF
--- a/Content.Shared/Anomaly/SharedAnomalySystem.cs
+++ b/Content.Shared/Anomaly/SharedAnomalySystem.cs
@@ -187,7 +187,8 @@ public abstract class SharedAnomalySystem : EntitySystem
             // Logging before resolve, in case the anomaly has deleted itself.
             if (_net.IsServer)
                 Log.Info($"Ending anomaly. Entity: {ToPrettyString(uid)}");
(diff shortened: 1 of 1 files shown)


Content.Shared/Anomaly/SharedAnomalySystem.cs
```diff
@@ -483,4 +484,6 @@ public partial record struct AnomalySpawnSettings()
     public bool SpawnOnSeverityChanged { get; set; } = false;
 }
 
-public sealed partial class ActionAnomalyPulseEvent : InstantActionEvent { }
+public sealed partial class ActionAnomalyPulseEvent : InstantActionEvent
+{
+}
```
EmoGarbage404, line 489:
```suggestion
public sealed partial class ActionAnomalyPulseEvent : InstantActionEvent;
```

PR #36944, opened 2025-04-26
Make departmental orders consoles print slips

sowelipililimute:
## About the PR
- departmental request computers now print slips that can be inserted into the cargo request computer and approved there instead of departmental staff 

## Why / Balance
- total elimination of cargo desk job wasn't well received on deltav or other mrp servers
- move back some vetting into cargo's hands w/o "well i think this is a bad thing to spend our funds on because blahblahblah" being a thing because it's not cargo's funds it's the departments' funds
- good for situations like
  - why is the clown trying to order universals
  - oh you have an order for a glass crate but we have 1,000 in the back let me get you some of that

## Technical details
- CargoAccountPrototype now has an AcquisitionSlip prototype to be spawned for that account
- CargoOrderConsoleComponent now keeps track of if it is a `SlipPrinter`, as well as sfx + time bookkeeping for that
- CargoOrderData now keeps track of the account responsible for an order, which it will draw from for funds
- slip printer consoles don't keep track of a set of added/approved orders, but instead print a slip based on the AcquisitionSlip prototype of the account when you complete the form window
- this AcquisitionSlip contains a CargoSlipComponent, which essentially encapsulates the data needed to add a cargo order + its account to the pending orders database

## Media
grafik
https://github.com/user-attachments/assets/f006d173-f3af-438d-a6a6-ee0367ee019a

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Departmental cargo request computers now print slips that can be given to Cargo for final approval, rather than directly approving the orders themselves


lzk228:
so instead of approving all things inside your department, you have to go with that slip to cargo so they can approve it?

sowelipililimute:
yes

superjj18:
> so instead of approving all things inside your department, you have to go with that slip to cargo so they can approve it?

> yes

That reality is only marginally better than what it was before department economy isn't it?

I'm almost tempted to say this probably shouldn't be a upstream change since the issues forks may have are the results of it making some of their SOP outdated(I know starlight has that issue because QM is supposed to have forms for ALL orders, now they can't so I can see how starlight and other paperwork heavy forks would love this change).  As much as I love the forks, we shouldn't design upstream to cater to their SOPs.

My whole appeal was I could press a button on a console, and have a decent chance of actually receiving the mats my department needs without having to abandon my department for several minutes, or even need to leave it at all.

Maybe a command access lock for departmental order approval would be a better change for right now?

walksanatora:
speaking of starlight the agreed upon change was that dept. order consoles were changed to be ONLY pre-approved materials for that department. with cargo still having "AO" (All Orders).
but yeah this honestly seems like a WORSE situation then the old cargo cause now you need to run to your department. get it approved there. run to cargo with the paper get it RE APPROVED by cargo and then continue waiting.

IProduceWidgets:
I think receipt printing is pretty funny other issues aside.

Content.Client/Paper/UI/PaperVisualsComponent.cs
```diff
@@ -55,6 +55,24 @@ public sealed partial class PaperVisualsComponent : Component
     [DataField("headerMargin")]
     public Box2 HeaderMargin = default;
 
+    /// <summary>
+    ///     A path to an image which will be used as a header on the paper
+    /// </summary>
+    [DataField]
+    public string? FooterImagePath;
```
MilonPL, line 62: This should probably be a ResPath. We can convert it to string for now to work with TextureButton.

Content.Client/Paper/UI/PaperVisualsComponent.cs
```diff
@@ -55,6 +55,24 @@ public sealed partial class PaperVisualsComponent : Component
...
+
+    /// <summary>
+    ///     Modulate the header image by this color
+    /// </summary>
+    [DataField]
+    public Color FooterImageModulate = Color.White;
+
+    /// <summary>
+    ///     Any additional margin to add around the header
+    /// </summary>
```
MilonPL, line 72: Don't indent xmldoc.

Content.Server/Cargo/Systems/CargoSystem.Orders.cs
```diff
@@ -24,6 +25,8 @@ public sealed partial class CargoSystem
         [Dependency] private readonly SharedTransformSystem _transformSystem = default!;
         [Dependency] private readonly EmagSystem _emag = default!;
 
+        private static readonly SoundPathSpecifier BeepSound = new("/Audio/Effects/Cargo/beep.ogg");
```
MilonPL, line 28: This should be on the component and using a collection.

Content.Server/Cargo/Systems/CargoSystem.Orders.cs
```diff
@@ -56,6 +56,55 @@ private void OnInteractUsing(EntityUid uid, CargoOrderConsoleComponent component
             args.Handled = true;
         }
 
+        private void OnInteractUsingSlip(EntityUid uid, CargoOrderConsoleComponent component, ref InteractUsingEvent args, CargoSlipComponent slip)
```
MilonPL, line 59: `Entity<T>`

Content.Shared/Cargo/Components/CargoOrderConsoleComponent.cs
```diff
@@ -96,6 +96,30 @@ public sealed partial class CargoOrderConsoleComponent : Component
...
+    /// The time between prints.
+    /// </summary>
+    [DataField("printDelay")]
+    public TimeSpan PrintDelay = TimeSpan.FromSeconds(5);
+
+    /// <summary>
+    /// The sound made when printing occurs
+    /// </summary>
+    [DataField("printSound")]
+    public SoundSpecifier PrintSound = new SoundPathSpecifier("/Audio/Machines/printer.ogg");
```
MilonPL, line 122: soundcollection + remove tags + autopausedfield

Content.Server/Cargo/Systems/CargoSystem.Orders.cs
```diff
@@ -259,12 +311,48 @@ private void OnRemoveOrderMessage(EntityUid uid, CargoOrderConsoleComponent comp
...
+            var msg = new FormattedMessage();
+
+            msg.AddMarkupOrThrow(Loc.GetString("cargo-acquisition-slip-body",
+                ("product", product.Name),
+                ("description", product.Description),
+                ("unit", product.Cost),
+                ("amount", args.Amount),
+                ("cost", product.Cost * args.Amount),
+                ("orderer", args.Requester),
+                ("reason", args.Reason)));
```
MilonPL, line 345: AddMarkupPermissive, we're handling user input here.

MilonPL (dismissed):

Krunklehorn:
If the acquisition slip is sent through a fax machine, the duplicate doesn't work. Only the original.

IProduceWidgets:
> If the acquisition slip is sent through a fax machine, the duplicate doesn't work. Only the original.

might be better to prevent faxing them alltogether, like we do with spesos.

walksanatora:
> I think receipt printing is pretty funny other issues aside.

same think also happens with nanotask slips.

insoPL:
There should be a system for making interdepartment orders independent of ats.

Med orders steel from crago, cargo then completes the order one way or another(ats, mining). They get money when they drop off order in Med. 

Then Med can order additional chemmaster from sci.

Sec orders bica pills from Med, etc

EmoGarbage404:
very interested in playtesting this before stable merge.

PR #36950, opened 2025-04-26
Cleanup warnings: CS0414

B-Kirill:
## About the PR
Remove 4x `CS0414` warnings

[CS0414](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0414): The private field 'field' is assigned but its value is never used.

## Why / Balance
[https://github.com/space-wizards/space-station-14/issues/33279](https://github.com/space-wizards/space-station-14/issues/33279)

## Requirements

### DIFF
--- a/Content.Server/Maps/MapMigrationSystem.cs
+++ b/Content.Server/Maps/MapMigrationSystem.cs
@@ -17,9 +17,9 @@ namespace Content.Server.Maps;
 /// </summary>
 public sealed class MapMigrationSystem : EntitySystem
 {
-#if DEBUG
+#pragma warning disable CS0414
     [Dependency] private readonly IPrototypeManager _protoMan = default!;
-#endif
+#pragma warning restore CS0414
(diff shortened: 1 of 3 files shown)


Content.Server/Maps/MapMigrationSystem.cs
```diff
@@ -17,7 +17,9 @@ namespace Content.Server.Maps;
 /// </summary>
 public sealed class MapMigrationSystem : EntitySystem
 {
+#pragma warning disable CS0414
```
Tayrtahn, line 20: Probably better to wrap this in `#if DEBUG` instead of suppressing it.

B-Kirill: > ~Probably better to wrap this in `#if DEBUG` instead of suppressing it.~

By the way, I also thought so, but when I did a similar PR in Robust Toolbox, it was explained to me that suppression for dependencies would be much more convenient.

PR #36954, opened 2025-04-26
Adding a throwing speed modifier 

Thinbug0:
## About the PR

Added a ThrowSpeedModifier component, so the speed at which an object is thrown can be easily modified trough yaml

its pretty limiting and boring that stuff like spears or knives fly at exactly the same speed as toolboxes

## Why / Balance

Theres no easy way to modify the speed at which something is thrown atm. Spears, darts, bolas, etc all fly at the same speed when thrown

## Technical details

Oki so this turned out to be way more of an adventure that i originally hoped for but atm 

Added a new BeforeBeingThrownEvent to the ThrowingSystem, and moved the BeforeThrowEvent there too. So now whenever TryThrow is used, both of those events are called 

The ThrowSpeedModifierSystem listens to the first event, and applies a flat modifier and a multiplier to the speed value (the modifiers are 0 and 1 by default so they dont have any effect).  
This system also checks if the value would go below 0, setting it to just 0 if thats the case

Im not sure what to do about the check in line 231 in HandsSystem, it says it should happen afterwards, but for that to be possible it would need to happen inside the event, and i think doing that would require sending a bunch more values to TryThrow for this specific scenario only.
Its a confusing situation to solve and i dont know the right approach so im gonna leave it be for now, hoping i can get some help 

## Media

(small demonstration, yeah the names are outdated)

https://github.com/user-attachments/assets/8985530a-1e9d-463f-bd95-23c5f4350cd5

actual name of the component and variables
wawawawawa

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: The speed at which objects are thrown can now be easily modified trough yaml by a ThrowingSpeedModifier component


slarticodefast:
Not sure if this should be part of ItemComponent or its own component.

Content.Shared/Item/ItemComponent.cs
```diff
@@ -60,6 +60,12 @@ public sealed partial class ItemComponent : Component
     /// </summary>
     [DataField, AutoNetworkedField]
     public Vector2i StoredOffset;
+
+    /// <summary>
+    /// Modifies the speed at which the entity is thrown by hand by adding its value to it, default speed is 11.
+    /// </summary>
+    [DataField]
```
slarticodefast, line 67: Should be networked.

Content.Server/Hands/Systems/HandsSystem.cs
```diff
@@ -225,6 +226,11 @@ hands.ActiveHandEntity is not { } throwEnt ||
             direction *= distance / length;
 
             var throwSpeed = hands.BaseThrowspeed;
+            // Checks for and applies the throwing speed modifier
+            if (EntityManager.TryGetComponent<ItemComponent>(throwEnt, out var itemComponent))
```
slarticodefast, line 230: Use shorthands.

Content.Server/Hands/Systems/HandsSystem.cs
```diff
@@ -225,6 +226,11 @@ hands.ActiveHandEntity is not { } throwEnt ||
...
+            {
+                throwSpeed += itemComponent.ThrowingSpeedModifier;
```
slarticodefast, line 232: Make sure it doesn't become negative.

Thinbug0:
> Not sure if this should be part of ItemComponent or its own component. Might also make sense to use an event for this so multiple systems can modify it.

i wasnt sure either, but yeah it being its own component is prolly better

ill pester you later about what using an event means in this context, im currently very eepy deeby and gotta nap

EmoGarbage404 (changes requested):
is there a reason this is hand-specific? Wouldn't you want this behavior to exist for any form of launching the item?

Thinbug0:
> is there a reason this is hand-specific? Wouldn't you want this behavior to exist for any form of launching the item?

well i still gotta move it to its own component instead of being in the item component, so this is gonna make more sense after that 

but since theres multiple ways an object may be thrown, i think being able to modify the strength of an specific way to throw it would work better

i would be oki with changing that, but thats my current reasoning for the decision

Thinbug0:
actually thinking back on that, it would lead to having like an extra component per way of throwing something and that sounds like slop

so yeah its prolly better if it affects all throwing, tho its gonna take a bit bcause i need to go trough every system

Content.Shared/Throwing/ThrowSpeedModifierComponent.cs
```diff
@@ -0,0 +1,12 @@
+namespace Content.Shared.Throwing;
+
+[RegisterComponent, AutoGenerateComponentState]
+public sealed partial class ThrowSpeedModifierComponent : Component
```
slarticodefast, line 4: add documentation to the component itself

Content.Server/Hands/Systems/HandsSystem.cs
```diff
@@ -225,6 +225,12 @@ hands.ActiveHandEntity is not { } throwEnt ||
             direction *= distance / length;
 
             var throwSpeed = hands.BaseThrowspeed;
+            // Checks for and applies the throwing speed modifier
+            if (TryComp<ThrowSpeedModifierComponent>(throwEnt, out var speedComp)
+                && throwSpeed + speedComp.Modifier > 0)
```
slarticodefast, line 230: it should still apply in that case
just use Math.Max
also I think it is better to make this an event so multiple systems can modify it
and make a subscription with `ThrowSpeedModifierComponent`

Thinbug0: im still working on this im just having a hard time understanding how to do that/what that means

Thinbug0:
oki so since im turning this into an event, theres already a BeforeThrow event that happens righttt before the throwing itself. So im gonna listen to that

slarticodefast:
Small tip for testing the change with the darts, you can simply not stage and commit that change so it does not become part of the PR and you don't have to revert it later. Also we got yaml hotreloading (although it might be broken at the moment until the next engine release) so you can test these things by just modifying the protoype while in-game.

Thinbug0:
> Small tip for testing the change with the darts, you can simply not stage and commit that change so it does not become part of the PR and you don't have to revert it later. Also we got yaml hotreloading (although it might be broken at the moment until the next engine release) so you can test these things by just modifying the protoype while in-game.

i havent gotten hotreloading to work, so i just been manually adding and removing the component whenever im testing. In this case i did it bcause i was gonna ask for general help, but it got solved immediately and it was just a typo 

but yeah ill keep that in mind :3

Thinbug0:
oki this should be almost done so im prolly gonna undraft it tomorrow, i wanna make sure i didnt mess up another system

Thinbug0:
oki so, i did the stuffies for the most part, moved the BeforeThrowEvent to TryThrow and also added a BeforeBeingThrownEvent that the new ThrowSpeedModifierSystem uses

but i do need helpie on what to doon that small check inside the HandsSystem (i left a message besides it so youll see it)
im a tad confused on what it actually does and i think it may cause issues, since its happening before the event now

Thinbug0:
oki so the failed check is cake and a bug so yeah its not related to this

Thinbug0:
its been long enough that i utterly and completely forgor i even did this, hoping the code wasnt too poopy

(i did just see the new real merge conflict but im too eepy and this is kinda delicate i think so yeah gonna deal with it later)

PR #36955, opened 2025-04-26
Fix autorecharge

metalgearsloth:
Probably needs a 2nd over.


slarticodefast:
If this is broken on staging it should be hotfixed.

PR #24004, opened 2024-01-13
Fix solution editor UI

PJB3005:
Fixes #23645

The problem is that the solution editor UI is an EUI, so the UI updates before the game states are applied.

A correct fix would be to move it to a BUI in some way, but that's a little involved as we don't really have pre-existing code that uses BUIs in a manner good for this. I decided against this because I realized we'd want to have more tools similar to this and tbh I kinda figured integrating it with VV would be a better fix instead, so...

This is a bad workaround to manually synchronize the UI updates against game timing. It's not pretty but it works.

PR #36969, opened 2025-04-27
Admin Tool: Observe entities in an extra viewport

slarticodefast:
## About the PR
Allows admins to keep an eye on some players or any other entity by opening a camera window that follows them around.
Supports zooming via mouse wheel, window resizing and popout mode.
It can be opened using a verb or the `camera` console command, which takes any uid or player name.
The player panel also has a new button to open it for that player.

Thanks to @beck-thompson for making the window resizing and zoom work!

## Technical details
Attaches a dummy entity with the `EyeComponent` to the target when the UI is opened and uses the same code as surveillance cameras to create an extra viewport.
To prevent cheat clients from knowing they are being watched we move it to another visibility layer. This layer is already in use by admins in stealth mode. We now fully reserve that layer for this purpose by adding a enum flag for it.
The user of the camera gets a PVS override for it and a view subscription so that the entities around the camera are networked to them.
Since EUIs are updated before the networked gamestate is applied, we have to delay network states using a frame update. Without this the client would not know the net entity send in the EUIstate. See https://github.com/space-wizards/space-station-14/pull/24004 for more details. Using a BUI instead is not a good alternative here for the reasons I mention in the code comments.

Requires https://github.com/space-wizards/RobustToolbox/pull/6049

## Media

https://github.com/user-attachments/assets/0e1e7dd9-2ec5-4a49-87f7-9d3ad9308403

example

example

## Requirements

## Breaking changes
none

**Changelog**
:cl: slarticodefast, beck-thompson
ADMIN:
- add: Added the "Open Camera" admin verb and the "camera" console command which allow you to observe entities in separate window using a remote camera.


IProduceWidgets:
Thank god. Ill finally know when the nuke is armed from my ~~ERT staging~~ admin testing area.

KaiserMaus:
1. Buy 2k/4k big screen 
2. Join as admin into the round
3. Add all player cameras on new big screen 
4. Enjoy true 1984 experience. Big brother watching you!

lzk228:
just casually admining a server
image

Sparlight:
> example

ENHANCE Corgi.

Nice work honestly, this will be a great tool!

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml.cs
```diff
@@ -0,0 +1,104 @@
...
+        CameraView.Eye.Zoom = new Vector2(Zoom, Zoom);
+        args.Handle();
+    }
+
+    public void SetState(AdminCameraEuiState state)
+    {
+        _nextState = state;
+    }
+
+    // I know that this is awful, but I copied this from the solution editor anyways.
```
PJB3005, line 73: Ahem.

Content.Client/Administration/UI/AdminCamera/AdminCameraEui.cs
```diff
@@ -0,0 +1,119 @@
...
+        };
+    }
+
+    private void PopOut()
+    {
+        _lastLocation = _window.Position;
+
+        var monitor = _clyde.EnumerateMonitors().First();
+
+        var clydeWindow = _clyde.CreateWindow(new WindowCreateParameters
```
PJB3005, line 54: Why not use `OSWindow` for this?

PJB3005 (commented):
Not a full review I just caught this at 1 am

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml.cs
slarticodefast: See the explanation in the lines below.
A BUI is really not the right tool for this. They have costly interaction and range checks hardcoded in that cannot be bypassed except with a few components that disable *all* range checks for BUIs. And then it will still close the window if the user of the window ghosts or moves to another entity in another way.
What I would prefer is to have a simulation EUI that updates after the gamestate is applied. But that is out of scope for this PR.

Content.Server/Administration/UI/AdminCameraEui.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    private EntityUid CreateCamera(EntityUid target, ICommonSession observer)
+    {
+        // spawn a camera entity attached to the target
+        var coords = target.ToCoordinates();
+        var camera = _entityManager.SpawnAttachedTo(CameraProtoId, coords);
+
+        // The camera entity has its visibility flags set to 0 so that cheat clients can't see it.
+        // We add an override so that only the user can.
+        _pvs.AddForceSend(camera, observer);
```
ElectroJr, line 92: This shouldn't be necessary, view subscribers should already implicitly get force-sent

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml.cs
```diff
@@ -0,0 +1,104 @@
...
+    // This is needed because EUIs update before the gamestate is applied, which means it will fail to get the uid from the net entity.
+    // The suggestion from the comment in the solution editor saying to use a BUI is even worse:
+    // - BUIs have range checks hardcoded in, which can only be bypassed by having GhostComponent or IgnoreUIRangeComponent.
```
ElectroJr, line 76: FWIW `InterfaceData.InteractionRange` allows ranges to be configured per BUI

slarticodefast: grafik
still has the action blocker checks that cannot be circumvented otherwise

Content.Client/Administration/UI/AdminCamera/AdminCameraEui.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        if (_OSWindow.ClydeWindow == null || _OSWindow.Root == null)
+            return;
+
+        _control.Orphan();
+
+        _OSWindow.Root.AddChild(_control);
+
+        _OSWindow.ClydeWindow.RequestClosed += _ => _window.Close();
+        _OSWindow.ClydeWindow.DisposeOnClose = true;
```
PJB3005, line 68: Can't you do this directly with the APIs available on `OSWindow`?

beck-thompson: I just tried this and it doesn't quite work! The issue is listening for `Closing` will also close the window when you pop it back in, not just when the button is pressed. `RequestClosed` *only* goes off when the client clicks the x button.

Is there another equilient method that would work instead? I don't know much about this stuff so if you have more ideas let me know :laughing:

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml.cs
ElectroJr: There's also an option to disable the attempt event. Its on by default so people don't forget to check for ghosts/dead/incapacitated players interacting with buis

slarticodefast: Ok, `RequireInputValidation` seems to be able to do that. I will check again and adjust the comments accordingly, maybe I get it working with a BUI.
But when I tried it out before there were other problems with the UI closing again when it should not be.

slarticodefast: I pushed some changes to the comments. In addition I made another branch where I tried changing it to a BUI instead, which caused two problems:
- The UI cannot be opened for the client in the same tick that the bound entity was spawned in. I assume this happens because opening the UI on the client relies on component states.
- Using a bound UI will also bind it to a user entity rather than a user session (see `UserInterfaceUserComponent`). This is undesirable for admin tooling as it will mean the UI is closed if the user changes his currently controlled entity, for example when ghosting or controlling a mob.

TaoNewt:
I know the 1984 thing is funny and all, but is 'power admining' like that actually possible?

Say you connected to all wizden servers as admin, and on the second monitor had 32 (4x8) pop-out windows of the 32 most sus players. Would that be a untenable performance hit?
Could an independent server have all 80 of their online players be displayed on a one admins monitors? 

Cause all jokes aside, if that is as useful as it sounds, having popped-out windows automatically make themselves into an array would be needed, otherwise having loads open would be a faff. If an array was made, it could be taken a step further, add 'blank' pop out cams that auto open on priority server logs, 'player' exploded welding fuel tank or 'player' sent a-help, the 'blank' pop-out cam immediately un-blanks and attaches to them.

Pixeltheaertist:
I actually really do like the idea of the option to open a panel that has a setting of "auto focus", which pretty much will focus onto players causing "High" or "Severe" loggings, or a player ahelping, etc. Could be hella useful to see what is happening at that EXACT moment, because in the time it takes to find their name and teleport to them, they could be done or away from the situation

beck-thompson:
> I know the 1984 thing is funny and all, but is 'power admining' like that actually possible?

> Say you connected to all wizden servers as admin, and on the second monitor had 32 (4x8) pop-out windows of the 32 most sus players. Would that be a untenable performance hit?
Could an independent server have all 80 of their online players be displayed on a one admins monitors?

> Cause all jokes aside, if that is as useful as it sounds, having popped-out windows automatically make themselves into an array would be needed, otherwise having loads open would be a faff. If an array was made, it could be taken a step further, add 'blank' pop out cams that auto open on priority server logs, 'player' exploded welding fuel tank or 'player' sent a-help, the 'blank' pop-out cam immediately un-blanks and attaches to them.

Yeah that would be nice and those are all good ideas! It really wouldn't be too hard to add that functionality. Also, for having 30+ windows open at a time it would basically be equivalent to having 30 more players connected to the server on the same connection. Obviously, if the connections were individual the server could handle it but:
1.) If multiple admins were doing this, it could add up quickly
2.) I'm unsure if one connection could handle the throughput of 30+ players pvs

As for the automatic popup stuff, yeah that's possible and probably wouldn't be too hard! The issue though is this PR already has some slightly suspect methods and it just needs to get reviewed in this simple state before more functionality is added.

slarticodefast:
For the performance side, this is probably not that bad because it it only sends the entities for two observed entities with overlapping PVS range once. So the worst case scenario would be that you send the whole station to the observing admin. Multiple admins observing every single player on all servers would be worse of course, both for performance and probably for our admins' mental health.
Regarding automatic popups, I would prefer to not cause any notification fatigue with too many windows popping for false alerts. But a very simple alternative is to add a button to the existing chat alert messages admins receive once the chat refactor is done.

TaoNewt:
The clickable message to pop open the window is definately better than the version i suggested.

BigfootBravo:
A system similar to this could be handy for AI to be able to make like static windows based on a camera, to monitor like 1 area constantly! A similar existed at the very least on Goonstation in SS13, possibly more, and I found it pretty handy to be able to split my attention while still being able to monitor a specific area.

DandiIion:
This'd be a really good feature for AI 👀

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml
```diff
@@ -0,0 +1,22 @@
+<Control
+    xmlns="https://spacestation14.io"
+    xmlns:viewport="clr-namespace:Content.Client.Viewport"
+    MouseFilter="Stop">
+    <PanelContainer StyleClasses="BackgroundDark" Name="AdminCameraWindowRoot" Access="Public">
+        <BoxContainer Orientation="Vertical" VerticalExpand="True" Access="Public">
```
PJB3005: `VerticalExpand` does nothing here.

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml
```diff
@@ -0,0 +1,22 @@
...
+            <!-- Camera -->
+            <Control VerticalExpand="True" HorizontalExpand="True" Name="CameraViewBox">
```
PJB3005: `HorizontalExpand` does nothing here.

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml
```diff
@@ -0,0 +1,22 @@
...
+                <viewport:ScalingViewport Name="CameraView"
+                                          VerticalExpand="True"
+                                          HorizontalExpand="True"
```
PJB3005: These properties do nothing here.

PJB3005:
I cannot get the code to compile locally and can therefore not review this PR as it currently stands.

slarticodefast:
We merged master into both the engine and content PR and I tested to make sure it compiles properly.
Please have another look when you got time.

Content.Client/Administration/UI/AdminCamera/AdminCameraEui.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        _control.Orphan();
+
+        _OSWindow.Root.AddChild(_control);
+
+        // Only close when the user actually clicks the close button (this doesn't trigger when its popped back in)
+        _OSWindow.RequestClosed += _ => _window.Close();
+
+        // You can't close the window, otherwise the EUI will complain that your sending requests without it open.
+        _window.Visible = false;
```
PJB3005: This entire EUI code is a mess. You should be closing this window properly here, not hiding it.

PJB3005 (changes requested):

Content.Client/Administration/UI/AdminCamera/AdminCameraControl.xaml
beck-thompson: This one is necessary

PJB3005: That literally is not possible. The box container is vertically oriented, so `HorizontalExpand` is ignored.

PJB3005: image

The layout does indeed work fine with this property gone.

beck-thompson: ah I see! Yeah I missread it as `VerticalExpand`! :partying_face:

Content.Client/Administration/UI/AdminCamera/AdminCameraEui.cs
beck-thompson: The issue is I can't close the window without the EUI manager getting mad that there is no window open. How would you recommend solving this? Should there be some new type of EUI that doesn't need a UI?

PJB3005: > The issue is I can't close the window without the EUI manager getting mad that there is no window open. How would you recommend solving this? Should there be some new type of EUI that doesn't need a UI?

Read your own EUI code. The EUI manager has no idea what a window is. You're telling it to close when the window is closed, so don't do that.

beck-thompson: yeah I was being a goober :laughing:

This was tough though - now it closes the window properly but overall it seems a little odd

PJB3005 (approved):
Scar took too long to press merge so now I get merge credit

little-meow-meow:
This is a great change. I'd like to use it and see about integrating something like this for the station AI, but I'm noticing that it causes my game to cycle between allocating and freeing about 6 gigabytes of memory every few seconds. Is this something you can replicate?

image

beck-thompson:
> This is a great change. I'd like to use it and see about integrating something like this for the station AI, but I'm noticing that it causes my game to cycle between allocating and freeing about 6 gigabytes of memory every few seconds. Is this something you can replicate?

Unfortunately I can not replicate that, can you give me more specifics? Was this on a fork, was it during a live round, how were you using the camera etc.

little-meow-meow:
I'm investigating further. I'm testing this on the space-wizards repo from master as of an hour ago. I'm running the game in debug mode on the dev map and performing the test right after compilation (i.e. neither client nor server has any considerable uptime.) I'm spawning a urist next to me at the spawn of the dev map and opening a camera window on him.

Here's an annotated screenshot of my total system memory usage from this morning's test. I hadn't thought any better in the moment to monitor more specifically than my total memory usage, but I can tell you from a quick follow-up observation that the increased memory consumption is limited to the client (no change on the server) and my total system memory usage fluctuates no more than 0.2 gigabytes while doing nothing in my testing environment.

image

Point 1: I have opened the camera window on urist. Total memory usage is 31.3 GB. Memory usage spikes within seconds and oscillates between roughly 37 and 34 GB. Peak at 37.4 GB.

Point 2: I move to the weapons room to escape the camera window's PVS. Memory usage drops to 34 GB, soon resumes oscillating between 36.5 and 34 GB.

Point 3: I close the camera window. Memory usage immediately drops to a baseline of 33 GB, up 1.7 GB from before opening the camera.

I'll dig deeper with dotMemory.

beck-thompson:
Wow that is extremely odd 🤔

I did basically exactly the same test and my memory usage didn't change even at all...

little-meow-meow:
Ah, okay. "Stretch viewport to fit game window" is the deciding factor. I'm using a 3840x2160 monitor. When I resize the game window above a point (around 1090 pixels height) while using the camera, the viewport resizes and memory usage immediately spikes. dotMemory says all the extra usage is unmanaged.

image

slarticodefast:
For visibility reasons, could you create an issue where you summarize your findings?
If we manage to improve the performance for this that would be great, so the snapshots are really useful.

PR #36979, opened 2025-04-27
Dans soaked smokes redux

Hanzdegloker:
## About the PR

Like half of the flavors for Dan's Soaked smokes just don't spawn natrually ingame and that's a tragedy, also a good few are just kinda eh and don't do anything. 

## Why / Balance

I think Dan's soaked smokes are fun, they're kinda like gambling but with minimal upsides and minimal downsides. But a good few flavors just don't do anything OR the inhalation speed of the reagent is so slow it cannot trigger the desired effect. Of course also a bunch of flavors existing but being basically non-existent during normal gameplay is very strange. 

## Technical details

Fixed two minor mispelling in two cigs, reworked a few cigs and standardized the reagent amount and max size of cigs (for God knows why dan's cigs, which are the questionable quality cigs, had double the reagent storage though it was never used). The total amount of soaked smokes choices for the fill was only increased by one. All possible soaked smoke flavors now spawn in during normal gameplay in the mixed pack. The nasty and medical packs were I imagine workarounds from long ago and with all cigs being able to spawn I see no reason to keep them seeing as they so comically rare and yet still barely impactful what's the point? For that reason I just turned instances of either into just normal mixed packs for simplicity's sake.

## Media

I hope this is simple enough to not require it.

## Requirements

## Breaking changes

God I hope not.

**Changelog**

:cl:
- tweak: Previously unused flavors alongside some reworked ones now appear in Dan's soaked smokes packs.


EmoGarbage404 (changes requested):
You need to convert the fill to an entity table because I'm 90% sure how its done now is going to introduce the possibility of heisentests.

Hanzdegloker:
> You need to convert the fill to an entity table because I'm 90% sure how its done now is going to introduce the possibility of heisentests.

I'm gonna be very real with you, I asked about and was concerned about this and got no response what so ever on the discord but I knew this would would get someone's attention lol. Thankyou, gettin to work now.

Hanzdegloker:
> needs entries in migration.yml for all the removed prototypes.

Copythat

Hanzdegloker:
I feel kinda bad always requesting reviews from you Emo in specific over these mistakes in a simple PR lol.
Here's to hopefully not making the same ones twice anyhow.

Chartman95:
I would like to voice here that half the fun of Dan's soaked smokes _was_ the extra capacity so you could custom soak and smoke them yourself. Now there's zero room for extra reagents and regular cigarettes have to be the new and reduced capacity soaked cigs.
It's removing a fun element from the game, for little explained reason is what I'm saying.

Chartman95:
> We should have, a custom cig making crate which can be bought from Cargo which clearly shows the cigs can store more and are reagentless so you don't have to fear whatever Dan put in the cig is going to mix with what you want to put in. Dan's soaked smokes are right under the syndie cigs which have 40u size instead of 20u, I think the extra storage was just a copy-paste thing.

It would be just like Dan to get his discount smokes from a back alley Syndie dealer, wouldn't it?
Regardless, can we avoid removing a feature until it's alternative is in place?

> Fair point, you can make a PR to revert that part if you want. Looks like the intention behind that was just overlooked.

Im still new to GitHub but is a revert different to a regular PR just changing code?

slarticodefast:
Don't revert the whole PR. Just change the solution container volumes in a new one.

PR #36980, opened 2025-04-27
Fragile Deliveries

ScarKy0:
## About the PR

Deliveries now have a 25% chance to spawn as a fragile delivery.
Fragile deliveries get 10% more profit, but lose 10% on being broken. (Either by tripping, throwing, etc)

## Why / Balance

Missing feature from the design doc, and wanted by players. Also adds a little more flavor to cargo itself.

## Technical details

DeliveryFragileComponent, stores the "Broken" state. Subscribes to BreakageEventArgs to decide when it should break.
It also gets all neccessary components from the ComponentTable (damage on impact, etc)

## Media

image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Deliveries can now spawn as fragile-type! Deliver them intact to earn a minor speso bonus.

### DIFF
--- a/Content.Shared/Delivery/DeliveryFragileComponent.cs
+++ b/Content.Shared/Delivery/DeliveryFragileComponent.cs
@@ -15,13 +15,13 @@ public sealed partial class DeliveryFragileComponent : Component
     /// Multiplier to use when the delivery is intact.
     /// </summary>
     [DataField]
-    public float IntactMultiplierOffset = 0.1f;
+    public float IntactMultiplierOffset = 0.15f;
 
     /// <summary>
     /// Multiplier to use when the delivery is broken.
     /// </summary>
     [DataField]
-    public float BrokenMultiplierOffset = -0.1f;
+    public float BrokenMultiplierOffset = -0.33f;
 
     /// <summary>
     /// Whether this priority has already been broken or not.
--- a/Content.Shared/Delivery/DeliveryFragileComponent.cs
+++ b/Content.Shared/Delivery/DeliveryFragileComponent.cs
@@ -1,0 +1,31 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Delivery;
+
+/// <summary>
+/// Component given to deliveries.
+/// Allows the delivery to be broken.
+/// If intact, applies a small multiplier, otherwise substracts it.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(DeliveryModifierSystem))]
+public sealed partial class DeliveryFragileComponent : Component
+{
+    /// <summary>
+    /// Multiplier to use when the delivery is intact.
+    /// </summary>
+    [DataField]
+    public float IntactMultiplierOffset = 0.1f;
+
+    /// <summary>
+    /// Multiplier to use when the delivery is broken.
+    /// </summary>
+    [DataField]
+    public float BrokenMultiplierOffset = -0.1f;
+
+    /// <summary>
+    /// Whether this priority has already been broken or not.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Broken;
+}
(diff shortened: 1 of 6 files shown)


TeenSarlacc:
is it possible to only have fragile items in the fragile packages? (bottles, plates etcetera)

TeenSarlacc:
> is it possible to only have fragile items in the fragile packages? (bottles, plates etcetera)

and have said items break when the package is broken (open up package, 500 gallons of alcohol spill out

ScarKy0:
> and have said items break when the package is broken (open up package, 500 gallons of alcohol spill out)

You can have deliveries get deemed "Fragile" based on the contents, that is what downstreams do in their mail. However I dislike the solution, and I dislike stuff inside of deliveries breaking when this is done. It just feels like unneccessary annoyance.

Hitlinemoss:
>  I dislike stuff inside of deliveries breaking when this is done. It just feels like unneccessary annoyance.

I think it's important for the sake of verisimilitude. It would feel weird for a "fragile" package to be opened only to contain, like, a piece of paper or articles of clothing. Likewise, it would feel weird for a fragile package to be "broken" but when you actually open it the bottles/glasses/etc inside are perfectly intact.

ScarKy0:
> I think it's important for the sake of verisimilitude.

Yeah but at the same time it leads to some annoying gameplay. The main thing here is that modifiers are meant to add money to what cargo earns, breaking contents n stuff just feels like making it more annoying for no real reason besides realism.

Morp9x:
No one will care that this mail is fragile, if it breaks, it will cost the same as regular mail (1% difference If I understood correctly). You should objectively introduce a much higher penalty for damaging mail. A fragile parcel that contains clothes is also a very strange absurdity.

ScarKy0:
> it will cost the same as regular mail (1% difference If I understood correctly)

If you deliver it intact you get +10% (so a default letter will be 550 instead of 500) and if you deliver it damaged its -10% (default letter will be 450)

ScarKy0:
Like you can argue nobody will care but that is more or less the point. Cargo can choose to be careful or deliver stuff faster to get bonus money, a higher penalty will make it annoying and breaking stuff inside the boxes will remove the reward for whoever is the recipient.
Either way, all this can be changed, this is just a basic system.

IProduceWidgets:
> > I think it's important for the sake of verisimilitude.
> 
> Yeah but at the same time it leads to some annoying gameplay. The main thing here is that modifiers are meant to add money to what cargo earns, breaking contents n stuff just feels like making it more annoying for no real reason besides realism.

good for janitors

ArcaneOcto:
I'm going to have to agree with the "breakable contents" party, as I think it would have some cool RP. The "wtf" moment when cargo delivers to me a binch of broken glass is just as fun as whatever small trinket I would have gotten imo. I also think it would be neat to be delivered fancy drinks if cargo _is_ successful. 

To be fair however, this is a fairly small detail and I think fragile deliveries will be cool regardless of how they're implemented!

slarticodefast:
I think breaking it should be way more impactful. A 10% deduction is nothing. Should be 50% at least. if breaking it only costs 50 spesos people will just throw it around regardless.

ScarKy0:
> I think breaking it should be way more impactful. A 10% deduction is nothing. Should be 50% at least. if breaking it only costs 50 spesos people will just throw it around regardless.

I know the rewards are a bit small but I also don't want it to be like "Oh boy free money!". Especially with how hard it is to have it accidentally break. It would be essentially a free 50% money bonus.

slarticodefast:
I mean 50% decuction when failing and 10% increase when successful

ScarKy0:
> I mean 50% decuction when failing and 10% increase when successful

Oh yeah I can work with that.

ScarKy0:
There, -33% if its broken. I'll need to do a pass on balance after looking at vulture anyways.

slarticodefast (approved):
I think the modifiers should have a bigger impact in general, but we can do that in a general balance pass like you say.

PR #34223, opened 2025-01-05
Updated air scrubber and air vent sprites.

ScarKy0:
## About the PR

Updated air scrubber and air vent sprites to be the tg ones.
Requested by an art director.

## Media

image
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Changed the air vent and scrubber sprites.

PR #36982, opened 2025-04-27
Revert "Updated air scrubber and air vent sprites."

EmoGarbage404:
# About this PR
Reverts space-wizards/space-station-14#34223

Opting to revert due to some feedback from maints
[meeting notes](https://hedgedoc.spacestation14.com/BdvjeoukSjuiISWyAwB1QA?view)

General issues brought up at maint meeting:
- Identifying the mode type on scrubbers is much harder since these are all recolors instead of proper different animations like before.
- They shouldn't have shading on the face of them much less directional shading that changes when they rotate. General statement about flat shading on floor objects goes here.
- Kinda stick out from other atmos devices too much
- Too smooth and overly detailed--artstyle clash

Some things to improve:
- PJB brought up having the lip of the vent/fan be a separate layer that only shows without a tile
- - basically have a secondary bezel that protrudes over playing but is otherwise smooth on tiles
- The fan spinning is good and probably should stay.


ArtisticRoomba (approved):

slarticodefast:
This should should target staging or we will have the other sprites for the next two weeks

Issue #36906, opened 2025-04-25
ChemMaster pilling resets dosage size on interact


archee1:
## Description
when using the chemmaster to create pills or swap contents (input or output slots) the dosage resets to 20u, and the count tries to max out for the most pills it can at 20u, this is incredibly annoying and only serves to be inconvenient.

**Reproduction**
have 200u of any chemical in the chemmaster
place a pill canister in the chemmaster
go to output and set dosage to 5u and count to 1, then hit create
*dosage has now reset to 20u, and count will attempt to make 9

Kittygyat:
Issue replicated.

PR #36994, opened 2025-04-28
ChemMaster UI Reset Fix

r3ndd:
## About the PR
This makes it so the ChemMaster pill number and dosage do not reset back to their default values after creating a pill.

## Why / Balance
Addressing #36906 . The ChemMaster would wipe the pill number and dosage settings upon creating a pill, making it an arduous process to re-use the same settings multiple times in a row.

## Technical details
This makes pill number and pill dosage a shared value by the client and server, which is updated by the server when a pill is created. It adds SelectedPillNumber and SelectedPillDosage fields to ChemMasterBoundUserInterfaceState and updates them in ChemMasterSystem>OnCreatePillsMessage. Finally, it changes ChemMasterWindow>UpdateDosageFields to only use a default value when the shared pill number is zero.

## Requirements

## Breaking changes
Added two fields to ChemMasterBoundUserInterfaceState: SelectedPillNumber and SelectedPillDosage

**Changelog**
:cl:
- tweak: ChemMaster pill number and dosage no longer resets after creating a pill.

### DIFF
--- a/Content.Server/Chemistry/Components/ChemMasterComponent.cs
+++ b/Content.Server/Chemistry/Components/ChemMasterComponent.cs
@@ -15,6 +15,12 @@ namespace Content.Server.Chemistry.Components
         [DataField("pillType"), ViewVariables(VVAccess.ReadWrite)]
         public uint PillType = 0;
 
+        [DataField("pillNumber"), ViewVariables(VVAccess.ReadWrite)]
+        public uint PillNumber = 0;
+
+        [DataField("pillDosage"), ViewVariables(VVAccess.ReadWrite)]
+        public uint PillDosage = 20;
+
         [DataField("mode"), ViewVariables(VVAccess.ReadWrite)]
         public ChemMasterMode Mode = ChemMasterMode.Transfer;
 
--- a/Content.Client/Chemistry/UI/ChemMasterBoundUserInterface.cs
+++ b/Content.Client/Chemistry/UI/ChemMasterBoundUserInterface.cs
@@ -42,10 +42,10 @@ namespace Content.Client.Chemistry.UI
                 new ChemMasterSetModeMessage(ChemMasterMode.Discard));
             _window.CreatePillButton.OnPressed += _ => SendMessage(
                 new ChemMasterCreatePillsMessage(
-                    (uint) _window.PillDosage.Value, (uint) _window.PillNumber.Value, _window.LabelLine));
+                    (uint)_window.PillDosage.Value, (uint)_window.PillNumber.Value, _window.LabelLine));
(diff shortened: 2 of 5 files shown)


Content.Server/Chemistry/Components/ChemMasterComponent.cs
```diff
@@ -15,6 +15,12 @@ public sealed partial class ChemMasterComponent : Component
         [DataField("pillType"), ViewVariables(VVAccess.ReadWrite)]
         public uint PillType = 0;
 
+        [DataField("pillNumber"), ViewVariables(VVAccess.ReadWrite)]
```
EmoGarbage404, line 18:
```suggestion
        [DataField]
```
This is all applied automatically

Content.Server/Chemistry/Components/ChemMasterComponent.cs
```diff
@@ -15,6 +15,12 @@ public sealed partial class ChemMasterComponent : Component
...
+        public uint PillNumber = 0;
+
+        [DataField("pillDosage"), ViewVariables(VVAccess.ReadWrite)]
```
EmoGarbage404, line 21:
```suggestion
        [DataField]
```
here too

EmoGarbage404 (approved):
pretty cool. only have small comments

Content.Server/Chemistry/Components/ChemMasterComponent.cs
r3ndd: Ah okay, good to know!

Quantum-cross (approved):
Looks good

metalgearsloth (changes requested):
Review still needs doing unless firefox is trolling me.

r3ndd:
My bad, I didn't commit the suggestions. It should be good to go now.

r3ndd:
@metalgearsloth Is there anything else need for this PR?

r3ndd:
@metalgearsloth Just a friendly nudge about this PR.

Errant-4 (changes requested):
The values are not saved until you create pills, and are reloaded when you change pill type. This means that if you set new values first and then change the pill type, your change will reset. This feels like a UI annoyance, it would be better if the values were stored as soon as they change since that's what the user would expect

r3ndd:
Ok, I went ahead and make these fields sync with the server when their values change. I also did this for the bottle dosage.

r3ndd:
I adjusted this to send separate messages when the pill number and dosage are adjusted. I also made the UI use SpinBox.OverrideValue when updating the spin boxes using server values for number/dosage, as this prevents it from sending redundant update messages back.

PR #36995, opened 2025-04-28
Chem master more unit transfer buttons

abadaba695:
Added buttons to transfer 15u, 20u and 30u in the chem master.

Some nice QOL to keep it in line with the chem dispenser, was a bit frustrating to not have these options as a chemist/botanist.

https://imgur.com/j8syjs0

- [X ] I have added media to this PR or it does not require an ingame showcase.

🆑 
- add: Added 15u, 20u, and 30u transfer buttons in the chem master.


slarticodefast:
Please add a changelog entry and keep the PR description template in the future.

EmoGarbage404 (approved):

slarticodefast (approved):
Seems fine.

Huaqas:
As someone who exclusively plays chemist, I really don't like this. It makes the Chemmaster UI super cluttered and bloated. And it is exasperated the issue of a lack of visual clarity between individual rows of buttons.

I have rarely ever needed to withdraw 15u of something from the buffer of a Chemmaster and I don't think we need to have a permanent button that you'll only use maybe once or twice per shift. As for 20u and 30u, you can simply press the 10u button multiple times to acquire that quantity.

Adding these amounts begs the question, why don't we have 2u, 3u, 4u, 35u, 40u, 45u, 55u, 60u, etc.

It makes sense for the Chemical Dispenser to have more selections for dispensing amounts, because you only have to display those options once in the UI. If you had to copy the button interface for every single jug in the chemical dispenser, it would look ridiculous and bloated.

This change only really make sense for people who chose to not use the Chemical Dispenser by withdrawing all the jugs and emptying them into the Chemmaster. But I think it's silly to bloat the UI of the Chemmaster to be more accommodating for people who choose not to use the Chemical Dispenser.

PR #36999, opened 2025-04-28
Fix examine prediction

metalgearsloth:
If our current ID is different to the received one we should just discard it even if it's for the same entity.

Note that this doesn't quite fix my prediction issue as client and server are using slightly different timings for the event.

### DIFF
--- a/Content.Client/Examine/ExamineSystem.cs
+++ b/Content.Client/Examine/ExamineSystem.cs
@@ -37,7 +37,6 @@ namespace Content.Client.Examine
         public const string StyleClassEntityTooltip = "entity-tooltip";
 
         private EntityUid _examinedEntity;
-        private EntityUid _lastExaminedEntity;
         private Popup? _examineTooltipOpen;
         private ScreenCoordinates _popupPos;
         private CancellationTokenSource? _requestCancelTokenSource;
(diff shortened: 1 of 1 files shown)


EmoGarbage404 (approved):
this coder was found ALIVE in their home.

:(

Content.Client/Examine/ExamineSystem.cs
```diff
@@ -416,15 +415,14 @@ public void DoExamine(EntityUid entity, bool centeredOnCursor = true, EntityUid?
             if (!IsClientSide(entity))
             {
                 // Ask server for extra examine info.
-                if (entity != _lastExaminedEntity)
+                unchecked
+                {
                     _idCounter += 1;
-                if (_idCounter == int.MaxValue)
-                    _idCounter = 0;
+                }
```
Tayrtahn, line 421: I'm not sure if this is an issue, but this is going to wrap the value to `int.MinValue` instead of `0`, right?

metalgearsloth: Correct, I don't think negative values are an issue (if they were ON GOD WHY DO WE USE INT NOT UINT) but that's intended.

Issue #18166, opened 2023-07-20
The inmovable rod deletes, not gibs


DogZeroX:
## Description

It's not as funny
**Reproduction**

**Screenshots**

**Additional context**

SlamBamActionman:
Looks like it gibs to me (ignore the microwaved guy)

<img width="538" height="362" alt="Image" src="https://github.com/user-attachments/assets/15f80195-d7be-4acb-b0af-b4750af23465" />

Issue #29526, opened 2024-06-28
Buckled entities will be deleted if their chair/bed is destroyed by an explosive.


PolarTundra:
## Description
Anything to which an entity can be buckled being destroyed by an explosive will cause the entity in question to be deleted. This includes anything they were holding, wearing or carrying.  It seems to have to be an explosive, or erase mode.

**Reproduction**
Buckle an entity to a chair or bed, anything that accepts buckled entities. Then detonate a bomb with a sufficient payload to destroy the chair or bed. If the bed of chair is destroyed, the entity should be deleted wholly.

**Additional context**

 As stated, this will not happen if the thing that the entity is buckled to is destroyed by melee damage. It will occur with Erase Mode, an immovable rod deleting the chair/bed, or an explosive doing so. Unsure if there are more possibilities.
 
 Pretty significant bug. Particularly for med of shuttlebombings, which is how I unfortunately came to learn about this bug's existence the hard way.

bruhmogus:
additional possibility: the object can also be transmogrified to cause this

Tyzemol:
apparently simply screwdriving to deconstruct the chair/bed works too
so deleting the entity in any way

lzk228:
should be fixed by #29718

Hreno:
I've tested this and the buckled entity doesn't get deleted if disassembling or destroying the chair.

But the entity still does get deleted if admin-deleting the chair though.

FairlySadPanda:
dun dun dun

IProduceWidgets:
Im gonna guess this is failing for cases where its not QueueDelete but just Delete

Pronana:
So I tested this and the problem is very stupid.

If you gib an entity on a buckled chair, everything is fine.

If you destroy a buckled chair while an entity is on it, everything is fine.

But if you gib an entity at the same time the chair is destroyed it gets deleted.

What's happening is Urist McHands gibs *before* the chair is deleted. This causes Urist McHands to be deleted and to drop his gibs into the chair for some reason. 

Items are dropped fine from testing which means it's the gibbing system that's fucking up *somewhere* and I can't figure out where.

PR #37007, opened 2025-04-28
Literally rewrite the Immovable Rod code just to stop Wizards being absolute gits

### LINKED ISSUE #17320: immovable rod soul count increased by two per kill
## Description

**Reproduction**

step into immovable rod as human

**Screenshots**

**Additional context**

i think first it gibs you and then it destroys brain - thus 2 souls

### LINKED ISSUE #23955: The immovable rod can delete the nuke
## Description
I just watched an armed nuke in a nukie round get deleted by the rod. Round 43630 on Salamander.
Related to issue #18166

FairlySadPanda:
## About the PR
CLANG, but less! The wizard's immovable rod polymorph has been rebalanced to be a brutal, Kool-Aid Man-style entrance or exit strategy with still some utility in a fight.

## Why / Balance
The rod currently does 190 damage, which is a LOT for a spell with this amount of utility, especially because it combines that with a focus on just deleting everything it touches. Fireball is your classic BOOM spell, Jaunt is your classic run-away spell. Immovable Rod is when you want to make an entrance.

Whilst investigating this I soaped myself into doing a rewrite of much of the system code and porting it to Shared because AUGH

## Technical details
* Move ImmovableRod to Shared.
* Detecting polymorphed entities is still on Server because old polymorph code AUGH
* Made ImmovableRods obey an EntityWhitelist to give designers options for what to disallow from being destroyed. Resolves #23955
* Slowed down the Wizard's rod to 20/s and increased its duration to 1.25s to compensate, making it feel far less jarring to use (especially now it's predicted)
* The wizard's rod now does 50 piercing damage and two hundred stamina to mobs, and 200 structural to structures. It no-longer auto-gibs or auto-destroys. It's still absolutely brutal to walls.
* Polymorph durations can now be sub-second intervals. Yeah this just was an int. Why? I dunno.
* Fix hitting people twice with the Immovable Rod because both Fix1 and Flammable were being hit. Resolves #17320 (probably).
* Probably does _not_ resolve 29526 which I suspect was the reason @PJB3005 got deleted to begin with, but resolves it for the case of the wizard being a shitter
* Shuffle all the ImmovableRod code around just to make it nicer to be around and more legible
* Removed Immovable Rod tile deletion code (it needs moving onto its own component anyway).

## Known Bugs

Sometimes people are just... sideways, after being hit by the Wizard's rod?? I suspect this is a wider PVS issue with animations.

## Media

https://github.com/user-attachments/assets/36bc0708-9892-48ff-9e35-b928911d06df

## Requirements

## Breaking changes
ImmovableRod has been moved to Content.Shared, with the usual namespace changes associated with that. Migrate any changes to Content.Server's ImmovableRod appropriately.

**Changelog**
:cl:
- tweak: Immovable Rods now deal huge amounts of structural damage, and quite a lot of piercing damage, instead of just deleting whatever they hit. CLANG!
- tweak: The wizard's immovable rod polymorph has been rebalanced, now intended to function as a middle ground of destruction and movement.
- tweak: Immovable Rods now don't destroy tiles by default.
- add: Three rare new immovable rods have been added.
- ADMIN: Several new admeme immovable rods have been added.

resolves https://github.com/space-wizards/space-station-14/issues/18166


keronshb:
This is the PR I prefer.
Will note a separate PR will still need to be opened for the Rod ECS to sanity check the deleting.

Can you show a video hitting some people, regular walls, and R Walls?

FairlySadPanda:
Will do once the PC I jumped on to do this PR remembers, yes, it has the .NET SDK installed 🔫

FairlySadPanda:
FWIW on design:

Other funny damage spells -> deal direct damage
Jaunt -> GTFO
Rod -> in the middle but a short duration, good at destroying structures and making a dramatic entrance

TytosB:
how is it not good? wizards have plenty of spells that let them kill large amounts of random people. if the wizard wants to destroy the station, they are both allowed to and given plenty of tools to do so. im not sure why we would need to nerf this one specifically.

FairlySadPanda:
I hate soap

keronshb:
Seems fine. Interestingly I can't seem to actually make a singulo despite how many rods I can spawn.
Would be out of scope probably to fix this, but can you give it a shot and see if you can spawn one? (Especially if two wiz rods collide)

Pronana:
Not part of the review but I feel like 50 damage, even if it's armor piercing and stuns, is way too low. 
I think it's fine if the rod crits anyone it touches, fireball usually kills when it hits and has a much shorter cooldown.

FairlySadPanda:
> Not part of the review but I feel like 50 damage, even if it's armor piercing and stuns, is way too low. I think it's fine if the rod crits anyone it touches, fireball usually kills when it hits and has a much shorter cooldown.

True, but fireball doesn't also get you out of dodge as well (e.g. it can't be used at near-point-blank, it doesn't bust you through walls).

If this is too rarely picked afterward then the maints can buff the damage up, maybe to 100 or so.

IProduceWidgets:
> Sometimes people are just... sideways, after being hit by the Wizard's rod?? I suspect this is a wider PVS issue with animations.

You would be correct. Seen it happen on live without your PR.

FairlySadPanda:
FFS man I'm not soaping myself into writing a mob animation state machine

keronshb:
> FFS man I'm not soaping myself into writing a mob animation state machine

Don't worry about that, it's out of scope. It's a weird bug that happens.

Pronana:
> If this is too rarely picked afterward then the maints can buff the damage up, maybe to 100 or so. This is a rework of the spell rather than a perfect rebalance.


Probably wont affect its usage cause wiz doesn't have many damaging spells. Iirc it's just rod fireball and smite.

> Oh and this makes you bleed a lot so it's more than 50 damage!

True! Bleed is pretty killer.

FairlySadPanda:
Triagers: just to flag that this is higher than a P2 for priority. Currently the wizard rod can trigger https://github.com/space-wizards/space-station-14/issues/29526, a P0 issue.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+        // we are deliberately including paused entities. rod hungers for all
+        foreach (var (rod, trans) in EntityManager.EntityQuery<ImmovableRodComponent, TransformComponent>(true))
+        {
+            if (!rod.DestroyTiles)
+                continue;
+
+            if (!TryComp<MapGridComponent>(trans.GridUid, out var grid))
+                continue;
+
+            _map.SetTile(trans.GridUid.Value, grid, trans.Coordinates, Tile.Empty);
```
EmoGarbage404, line 49: should update this to `AllEntityQuery<T>` (this is the enumerator version) while you're here.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        SubscribeLocalEvent<ImmovableRodComponent, StartCollideEvent>(OnCollide);
+        SubscribeLocalEvent<ImmovableRodComponent, MapInitEvent>(OnMapInit);
+        SubscribeLocalEvent<ImmovableRodComponent, ExaminedEvent>(OnExamined);
+    }
+
+    private void OnMapInit(EntityUid uid, ImmovableRodComponent component, MapInitEvent args)
+    {
+        if (!EntityManager.TryGetComponent(uid, out PhysicsComponent? phys))
+            return;
```
EmoGarbage404, line 65: TryComp

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        // Oh nyo.
+        if (HasComp<ImmovableRodComponent>(ent))
+            HandleRodRodCollision(uid, ent);
+        // Mobs never automatically get deleted by Rodney, because being instantly deleted is not any fun.
+        else if (TryComp<MobStateComponent>(ent, out var mob))
+            HandleRodMobCollision(uid, component, args);
+        // Admiral Roddington XIV may have been limited in his desire to delete everythng.
+        // What's the point of an immovable rod if it deletes all those organs it's just splattered?
+        else if (component.ShouldDelete && !_whitelist.IsWhitelistPass(component.DeletionBlacklist, ent))
```
EmoGarbage404, line 104: we have IsWhitelistFail which is a little bit more explicit. feel free to just ignore this comment though.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+        // :panik:
+        var coords = Transform(uid).Coordinates;
+        _popup.PopupCoordinates(Loc.GetString("immovable-rod-collided-rod-not-good"), coords, PopupType.LargeCaution);
+
+        // :kalm:
+        Del(uid);
+        Del(ent);
+
+        // :panik:
+        Spawn("Singularity", coords);
```
EmoGarbage404, line 122: This at least needs to be `SpawnAtPosition` and may need to be called before the deletion.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private void HandleRodMobCollision(EntityUid uid, ImmovableRodComponent component, StartCollideEvent args)
+    {
+        var ent = args.OtherEntity;
+
+        // TODO: Immovable rods are weird because they're projectiles that don't get consumed after they collide.
+        // We could store who the projectile has already hit, but prediction makes that non-trivial, and I've
+        // not found a good example in the codebase that mimics the rod's desired behaviour.
+        // Temporary hack: use the same collision layer that MobCollision uses.
```
EmoGarbage404, line 132: it would be preferable to leverage projectile penetration for this code instead but it might be a bit of an overhaul. It may be too much effort at the present moment.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+        if (args.OtherFixtureId != "flammable")
```
EmoGarbage404, line 133: perhaps just check for a hard fixture instead of specifically checking this one. Lots of entities use non-hard fixtures like this for checking special types of collision.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+            return;
+
+        // Ma'am, this is a Christian Minecraft server.
+        _popup.PopupEntity(Loc.GetString("immovable-rod-penetrated-mob", ("rod", uid), ("mob", ent)), uid, PopupType.LargeCaution);
```
EmoGarbage404, line 137: mmph.

EmoGarbage404 (changes requested):
mostly fine

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
FairlySadPanda: We need something consistent and sane for "thing hit mob". I'm deferring to "mirror how mob collision works"; we need a more universal system. Like a lot with mob code it's just shitcode under the hood.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
FairlySadPanda: Rod code being tied to wizard rod code is tbh just a bad idea in general

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
FairlySadPanda: The check specifically needs to be

whitelist not exist: true
whitelist fails: true
whitelist passes: false

AFAIK IsWhitelistFail doesn't do this - it's

whitelist not exist - false
whitelist fails - true
whitelist passes - true

FairlySadPanda:
There's a fix for the deletion warpath of immovable rods et al for buckled chairs, so this is now a bit lower prio. Will address Emo's feedback later.

FairlySadPanda:
Thoughts on making immovable rods not just call QueueDel?

Pronana:
> Thoughts on making immovable rods not just call QueueDel? That basically redefines how they work conceptually and removes the feature flag inside the YAML. Structure damage is there to allow for structures to get merc'd anyway and it'd mean that there's a reduction in wacky edge cases

I'd have the regular rod call destruction behavior that way it can be cancelled and raises an event when it's deleted.

keronshb:
This is a critical change, please leave the P0

slarticodefast:
Oh, didn't read the last comment before changing the label again.
I agree it is bad, but not servers on fire bad, so I think P1 is more fitting.

FairlySadPanda:
@EmoGarbage404 @slarticodefast Changes done.

Makefile
```diff
@@ -0,0 +1,27 @@
+# Common repo commands, so you don't have to memorize things after being told them once.
```
Pronana: Unintended file commit detected

Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
```diff
@@ -55,29 +59,19 @@
...
-    destroyTiles: false
+  - type: ImmovableRod # Deliberately significantly weaker than the non-spell counterpart.
+    maxSpeed: 20 # Because the Wizard's camera follows the rod, it needs slowing down a lot.
     randomizeVelocity: false
-    shouldGib: false
     damage:
       types:
-        Blunt: 190
+        Piercing: 50 # OH GOD, MY INSIDES!
+        Structural: 200 # OH GOD, MY CAR!
```
Pronana, line 74: 200 structural is not enough, there's strong potential they fail to destroy a reinforced wall or firelock and end up stuck inside of it.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,133 @@
...
+        /*
+         * TODO: Immovable rods are weird. Ergo, always only collide with one fixture on a mob.
+         */
+        if (args.OtherFixtureId != ent.Comp.MobCollisionFixtureId)
+            return;
+
+        var entityHitByRod = args.OtherEntity;
+
+        var popup = Loc.GetString(ent.Comp.OnMobCollisionPopup, ("rod", ent), ("mob", entityHitByRod));
+        _popup.PopupEntity(popup, ent, PopupType.LargeCaution);
```
Pronana, line 114: Immovable rod still says "Eviscerates entity" but it doesn't eviscerate them anymore.

Pronana (commented):
Removing QueDel is a good idea, but imo it should be changed to use destruction behavior which is cancellable and raises events. 

Right now it's not too different from a projectile that just doesn't delete itself on collide.

Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
```diff
@@ -92,16 +86,17 @@
   - type: NoSlip
 
 - type: entity
-  parent: ImmovableRodKeepTiles
-  id: ImmovableRodKeepTilesStill
-  suffix: Keep Tiles, Still
+  parent: ImmovableRod
+  id: ImmovableRodStill
+  name: Immobile Rod
```
Pronana, line 91: Shouldn't be capitalized. Also a suffix would be nice because I initially read it as immovable rod even though it's the immobile rod.

Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
FairlySadPanda: Bumped it to 1k, shared with the regular rod

FairlySadPanda:
> Removing QueDel is a good idea, but imo it should be changed to use destruction behavior which is cancellable and raises events.
> 
> Right now it's not too different from a projectile that just doesn't delete itself on collide.

You are entirely correct, and at this point I think you could break out the "persistant projectile", "tile destruction" and "rod theming" into three different components.

Really I've just omega-nerfed rod; it doesn't kill you on hit, it doesn't nuke tiles. But this is in response to how rods are actually used.

There's some other rotten rod code I'd love to touch. I think if the rod is dealing damage on hit it could easily also deal that damage to itself; this would give us a better way of controlling rod destructiveness (a rod that flies for 30 seconds straight through the station but doesn't hit much, a rod that hits a LOT in 10 seconds) than the timer we use currently.

At some point this goes from engineering to design, though. Rods are a survival round thing generally and a good survival round doc would help figure out how the random survival events work.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,133 @@
...
+            _popup.PopupCoordinates(popup, coords, PopupType.LargeCaution);
+
+            SpawnAtPosition(rod.Comp.SpawnOnRodCollision, coords);
+        }
+
+        rod.Comp.HasCollidedWithRod = true;
+        target.Comp.HasCollidedWithRod = true;
+
+        Del(rod);
+        Del(target);
```
EmoGarbage404, line 100: pretty sure you need to queuedel in collision or bad things will happen

EmoGarbage404 (changes requested):

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private void OnMapInit(EntityUid uid, ImmovableRodComponent component, MapInitEvent args)
+    {
+        if (!TryComp<PhysicsComponent>(uid, out var physics))
+            return;
+
+        _physics.SetLinearDamping(uid, physics, 0f);
+        _physics.SetFriction(uid, physics, 0f);
+        _physics.SetBodyStatus(uid, physics, BodyStatus.InAir);
```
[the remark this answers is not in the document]
Pronana: Reminding myself to look at this when I get home this feels sus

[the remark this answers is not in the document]
Pronana: You can define these values in yaml as emo said. Also remember to keep the tileFrictionModifier to 0 so that the damping is always 0.

Setting friction to 0 should be unnecessary given that friction is based on two colliding entities (aka an item rubbing up against a reinforced wall for example) and I believe the immovable object's physics body is the same as a bullet in that it doesn't experience this type of friction.

Setting Body Status again can be done in yaml.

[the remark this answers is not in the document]
Fildrance: fixed

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
Fildrance: fixed

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
```diff
@@ -0,0 +1,99 @@
...
+    ///     Should this rod spawn anything when it hits another rod?
+    /// </summary>
+    [DataField]
+    public string? SpawnOnRodCollision = "Singularity";
+
+    /// <summary>
+    ///     If this rod spawns anything when it hits another rod, what popup text should appear?
+    /// </summary>
+    [DataField]
+    public string OnRodCollisionPopup = "immovable-rod-collided-rod-not-good";
```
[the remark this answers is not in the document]
Fildrance: fixed

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
```diff
@@ -0,0 +1,99 @@
...
+    ///     How much stamina damage should this rod do to appropriate mobs?
+    /// </summary>
+    [DataField]
+    public float StaminaDamage = 200.0f;
+
+    /// <summary>
+    ///     Should this rod spawn anything when it hits another rod?
+    /// </summary>
+    [DataField]
+    public string? SpawnOnRodCollision = "Singularity";
```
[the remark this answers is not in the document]
Fildrance: fixed

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
```diff
@@ -0,0 +1,99 @@
+using Content.Shared.Damage;
+using Content.Shared.Whitelist;
+using Robust.Shared.Audio;
+
+namespace Content.Shared.ImmovableRod;
+
+[RegisterComponent]
```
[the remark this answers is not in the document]
Fildrance: fixed

Fildrance:
Removed usage of friction on Physics component - i failed to find any meaningful usage of this thing.

perryprog:
I suppose this resolves #18166 too.

FairlySadPanda:
Someone should really action this P1, it's only because wizards are rare and have more obviously fucked powers that this has escaped big malds since PJB encountered it.

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
```diff
@@ -0,0 +1,104 @@
...
+/// Marker for entities that represent unstoppable object,
+/// that crushes through obstacles and mobs.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class ImmovableRodComponent : Component
+{
+    /// <summary>
+    /// Tracks the number of mobs that this rod has slaughtered.
+    /// </summary>
+    public int MobCount = 0;
```
slarticodefast, line 18: shouldn't this be a datafield?

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// What fixture is used to detect that the rod has hit a mob?
+    /// </summary>
+    [DataField]
+    public string MobCollisionFixtureId = "flammable"; // Temporary default; aligns with mob collision code.
+
+    /// <summary>
+    /// Has this rod already hit another immovable rod? Used to avoid "oops, two singularities!"
+    /// bugs because collisions are evaluated on the same simulation tick.
+    /// </summary>
+    public bool HasCollidedWithRod;
```
slarticodefast, line 103: datafield

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+            return;
+
+        var (worldPos, worldRot) = _transform.GetWorldPositionRotation(uid);
+        var vel = worldRot.ToWorldVec() * component.MaxSpeed;
+
+        if (component.RandomizeVelocity)
+        {
+            vel = component.DirectionOverride.Degrees switch
+            {
+                0f => _random.NextVector2(component.MinSpeed, component.MaxSpeed),
```
slarticodefast, line 51: just make the angle nullable instead of interpreting a 0 degree angle as a random direction

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+                _ => worldRot.RotateVec(component.DirectionOverride.ToVec()) *
+                     _random.NextFloat(component.MinSpeed, component.MaxSpeed),
+            };
+        }
+
+        _physics.ApplyLinearImpulse(uid, vel, body: physics);
+        Transform(uid).LocalRotation = (vel - worldPos).ToWorldAngle() + MathHelper.PiOver2;
```
slarticodefast, line 58: this should have a setter in the transformsystem

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+        /*
+         * Note for future maintainers: it used to be the case that Immovable Rods (and wizards that polymorphed into
+         * one) literally _deleted_ things they hit. As in, they called QueueDel() on them.
+         *
+         * Do not make the Project Lead write threatening mald PRs:
+         * https://github.com/space-wizards/space-station-14/pull/37004
+         */
+
+        // Clang!
+        if (_random.Prob(ent.Comp.HitSoundProbability))
```
slarticodefast, line 72: Wouldn't all of this completely mispredict?
The component and subscriptions are in shared, but nothing is networked. And the server and client will roll different random results here. This needs either [RandomPredicted](https://github.com/space-wizards/RobustToolbox/pull/5849) or the workaround used in ClumsySystem.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+            _audio.PlayPvs(ent.Comp.HitSound, ent);
```
slarticodefast, line 73: wouldn't this play the sound multiple times due to prediction?

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        if (TryComp<ImmovableRodComponent>(args.OtherEntity, out var othComp))
```
slarticodefast, line 75: it will have a lot of collisions so use the query versions of TryComp and HasComp here for better performance.

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private void HandleRodCollision(Entity<ImmovableRodComponent> rod, Entity<ImmovableRodComponent> target)
+    {
+        if (rod.Comp is { SpawnOnRodCollision: not null, HasCollidedWithRod: false } && !target.Comp.HasCollidedWithRod)
+        {
+            var coords = Transform(target).Coordinates;
+            var popup = Loc.GetString(rod.Comp.OnRodCollisionPopup);
+
+            _popup.PopupCoordinates(popup, coords, PopupType.LargeCaution);
```
slarticodefast, line 90: won't work with prediction

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            PredictedSpawnAtPosition(rod.Comp.SpawnOnRodCollision, coords);
+        }
+
+        rod.Comp.HasCollidedWithRod = true;
+        target.Comp.HasCollidedWithRod = true;
+
+        QueueDel(rod);
+        QueueDel(target);
```
slarticodefast, line 99: PredictedQueueDel

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+        /*
+         * TODO: Immovable rods are weird. Ergo, always only collide with one fixture on a mob.
+         */
+        if (args.OtherFixtureId != ent.Comp.MobCollisionFixtureId)
+            return;
+
+        var entityHitByRod = args.OtherEntity;
+
+        var popup = Loc.GetString(ent.Comp.OnMobCollisionPopup, ("rod", ent), ("mob", entityHitByRod));
+        _popup.PopupEntity(popup, ent, PopupType.LargeCaution);
```
slarticodefast, line 113: prediction

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        if (ent.Comp.StaminaDamage > 0)
+            _stamina.TakeStaminaDamage(entityHitByRod, ent.Comp.StaminaDamage, ignoreResist: true);
+    }
+
+    private void OnExamined(EntityUid uid, ImmovableRodComponent component, ExaminedEvent args)
+    {
+        args.PushText(
+            component.MobCount > 0
+            ? Loc.GetString("immovable-rod-consumed-souls", ("rod", uid), ("amount", component.MobCount))
```
slarticodefast, line 128: it's not a datafield and not networked, so this will show wrong numbers to other clients
also it will mispredict

Content.Shared/Polymorph/PolymorphPrototype.cs
```diff
@@ -60,7 +60,7 @@ public sealed partial record PolymorphConfiguration
     /// can be null if there is not one
     /// </summary>
     [DataField(serverOnly: true)]
-    public int? Duration;
+    public float? Duration;
```
slarticodefast, line 63: doesn't really have anything to do with this PR?

Resources/Prototypes/Polymorphs/polymorph.yml
```diff
@@ -211,7 +211,7 @@
     transferName: true
     transferDamage: false
     inventory: None
-    duration: 1
+    duration: 1.25
```
slarticodefast, line 214: same

slarticodefast (changes requested):

slarticodefast:
Just a quick look, not a full review
I didn't test it in-game yet

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
Fildrance: Its not REALLY that big of a deal, its just sound effect of really fast-moving entity, people cannot react to it and its not really controlled... but it might be useful to leave comment to fix it with random predicted, because its nice :godo:

slarticodefast: You could also just not make this random and let it play a sound every N collisions instead

Content.Server/ImmovableRod/ImmovableRodSystem.cs
```diff
@@ -1,147 +1,17 @@
...
-            // oh god.
-            var coords = Transform(uid).Coordinates;
-            _popup.PopupCoordinates(Loc.GetString("immovable-rod-collided-rod-not-good"), coords, PopupType.LargeCaution);
-
-            Del(uid);
-            Del(ent);
-            Spawn("Singularity", coords);
-
+        // Save thyself from the wrath of Rod.
+        if (TryComp<PolymorphedEntityComponent>(ent, out var polymorphed) && polymorphed.Parent == args.OtherEntity)
```
slarticodefast, line 12: Oh, and this will also make it mispredict.
Why are polymorphed entities even immune in the first place?

Resources/Prototypes/Entities/Objects/Fun/immovable_rod.yml
```diff
@@ -55,29 +63,19 @@
     minSpeed: 1
     maxSpeed: 5
 
-- type: entity
-  parent: ImmovableRodDespawn
-  id: ImmovableRodKeepTiles
```
slarticodefast, line 60: the removed prototypes need a map migration
no clue if anyone has mapped one of these, but it won't hurt either

Content.Shared/ImmovableRod/ImmovableRodSystem.cs
slarticodefast: > Its not REALLY that big of a deal, its just sound effect of really fast-moving entity, people cannot react to it and its not really controlled... but it might be useful to leave comment to fix it with random predicted, because its nice :godo:

Then just leave all the code on the server and unpredicted instead of having it mispredict.

Content.Server/ImmovableRod/ImmovableRodSystem.cs
FairlySadPanda: "If the rod is a polymorphed wizard, and the rod somehow hits the non-polymorphed version of the wizard, don't actually collide with the non-polymorphed version of the wizard".

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
FairlySadPanda: No? It's just a var that tracks how many people have died due to the rod, it does not interact with YAML.

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
FairlySadPanda: No - this does not interact with YAML, and tracks if this rod has hit another rod.

To be honest a worryingly large amount of this code is hacky patches like this to avoid underlying physics issues.

Content.Shared/ImmovableRod/ImmovableRodComponent.cs
perryprog: Isn't it needed for map persistence though?

Content.Server/ImmovableRod/ImmovableRodSystem.cs
FairlySadPanda: So: 

Realistically, the wizard should not be able to hit themselves with their own polymorph. The proper fix to this is to stop this being able to happen.

Do I have any idea why this happens? No.

Do I want to have to rewrite the Polymorph system to stop this happening? No.

The best I can do is mark up the explanation for why this TryComp is doing what it's doing and TODO removing it once the base cause of the issue is fixed.

FairlySadPanda:
OK so the current plan on this:

Rods have some innate randomness. But Wizard rods don't worry about this. This feels like something that should be driven by different components.

Then we can make the Wizard polymorph into something that is a "rod" but we've not got this millstone of a tall gameplay system that needs to cover how a polymorphed antagonist, an admeme projetile AND a random station event  work inside one bit of code.

This essentially leads to a bit of a soap moment; this PR rewrites the Immovable Rod code, but the correct way to do this is to delete the code and start again with composition in mind. I can use the work done here as inspiration, but realistically the solution starts by deleting ImmovableRodComponent and ImmovableRodSystem, and thus this work is incompatible.

Maintainers seem to be fine with the intended new design for the Rod spell, so that will feature as part of the new work once we return to this point.

Before arguing this point: Go look the Immovable Rod YAML code. Does this look sensibly-developed to you?

FairlySadPanda:
Thank you to @Fildrance, @slarticodefast, @Princess-Cheeseballs and others for feedback. This PR has been reworked into https://github.com/space-wizards/space-station-14/pull/39877

PR #31247, opened 2024-08-21
PKA Modkits + Rebalance

EmoGarbage404:
## About the PR

Adds several modkits that can be used to buff the PKA. 2 out of the 3 can be installed at once but there can't be duplicates. They are researchable via the "Kinetic Modifications" tech and can also be rarely found in salvage.

The 3 modkits are:
- Damage: Raises damage to 25 blunt and 30 structural
- Range: Raises range from 4 tiles to about 6.5
- Fire Rate: Raises fire rate from 0.5 to 0.8

To compensate, rebalances the Proto-kinetic Accelertor:
- The range has been decreased to roughly 4 tiles
- The damage has been decreased to 15 blunt and 15 structural.
- Raises the size class to large and increases the inventory size to 2x3

## Why / Balance

The baseline PKA has really powerful stats that outclass most of its contemporaries. Decent damage, infinite ammo, and high range make it a forgiving and easy choice for salvagers but also make it better than actual guns and higher-level melee weapons in a lot of scenarios. Bringing the stats down a little bit carves out a bit of a niche for the PKA where it's less of an outright gun and more of a mid-range kiting option, which is a lot more interesting of an archetype.

Of course, dangerous monsters make strong ranged options necessary to an extent, so to facilitate this and to encourage collaboration with science, the modkits can serve as ways to make the PKA more viable. This gives the weapon a bit of a progression which makes it a more unique item for salvage while also giving them more of a reason to work with science in order to unlock the tech.

## Media

image
image
image

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl:
- add: Added 3 modkits for the proto-kinetic accelerator, granting increased damage, range, and fire rate. They can be researched via the new "Kinetic Modifications" arsenal tech.
- tweak: Reduced damage and range on the proto-kinetic accelerator.

PR #37012, opened 2025-04-28
re-buffs proto kinetic accelerator

TytosB:
## About the PR

re-buffed pka to its damage and range before https://github.com/space-wizards/space-station-14/pull/31247
## Why / Balance

the nerf to pkas to compensate for the addition of its modkits seems entirely wrongheaded. as someone who plays an unhealthy amount of salvage i can tell you that the weapon really wasn't that strong, mainly due to its very poor fire rate already compensating for its damage, newer salvagers especially really struggled with the speed of carps and enemies with special abilites like hivelords, goliaths, and acid spitting xenos. not to mention that all mob spawn rates and health pools were balanced around the old pka damage. the real focus here should be time to kill, not damage per hit, and the pka's time to kill was in the perfect sweet spot for the enemies salvs were meant to fight while not really being useful at all in pvp [again, due to its horrible fire rate, which even the fire rate modkit doesn't really change]

some will say that this will now make a pka WITH the modkits enitrely too powerful, and ill have to disagree. the modkits are a tier 2 arsenal tech, that is late shift technology, assuming sci even studies it, as in many shifts they simply wont. the original pr argued that this encouraged more communication between salv and sci, but sci was already the most common dept salv talked to-the pka cannon, industrial ore processors, sci being the main consumer of salvs mats, shuttle parts from shuttlecraft tech, etc etc. 

one thing i really think i need to stress though is that this really was a flat nerf to the pka, which didnt need one. even if you could give the new pka all 3 modkits [you can't, max of 2] it would only be slightly better than the pre-nerf pka. as it stands to get a pka thats even just 'as good' as the pre-nerf version you need 2 expensive modkits matwise that are locked behind a tier 2 tech. 

i also reverted the size nerf as it felt unnecessary on its own. its kind of a separate balance issue based on salvager storage space and ease of storing guns that's separate from a debate over the damage and fire rate of said gun, and nerfing it when youre already neutering the gun makes no sense to me. 

## Requirements

## Breaking changes

:cl:
- tweak: adjusted pka damage and range


slarticodefast:
Let's at least wait a little and see how players adjust to the other balance changes and the mod kits before making any more major changes to the balance. This is way too early.

EmoGarbage404:
i dont mind rebuffing damage but i think keeping the size and range is good for the more general balance of salvage combat on the whole

TytosB:
> i dont mind rebuffing damage but i think keeping the size and range is good for the more general balance of salvage combat on the whole

i can see the argument for a range nerf, but why size? salv is already kind of packed to the gills inventory wise, since their whole thing is looting random bits of tech and scrap to bring back to the stations

EmoGarbage404:
Having limited inventory makes for more engaging hauling, as it necessitates going in bigger groups, taking multiple trips, and bringing secondary storage containers. Especially for timed missions and things involving travel time, it gives it a bit of nuance. I also think keeping the PKA a little bit clunkier is ideal, since it really is intended to only be a starter weapon for salvagers.

TytosB:
> You really, realistically, shouldn't be taking on goliaths/xeno turrets without modkits, and flat out shouldn't be using your pka against hivelords regardless. Expeditions and the VGroid are meant to be where you go mid-round, after actually doing your job for a while and acquiring upgrades.

yes vgroid is mid round, but because youre meant to build a shuttle to be able to reach it, the enemies there are strong but doable with the current balance, adding another barrier to being able to have a real chance to fight them when you already need a shuttle and a large time investment to get there is excessive.


> New-to-the job salvagers are allowed to struggle until they learn how to use the rest of their tools effectively. Carp on debris for instance can be fought using your grappling hook off-grid. Getting someone to teach them that will make them struggle less than being able to kill a carp in 2 hits rather than 3.

yes but they already struggle with the tools as they are, salv is one of the most difficult roles to learn, and uniquely barring sec- screwing up means you die, often round removed if your corpse slips off a ruin. making the main starter weapon significantly worse when it was already balanced well is unnecessary. and its funny you mention carp hits to kill, as they used to be 3 hits but had their hp specifically nerfed to make them 2 hit for pka, so 2 hit carps were a pre-agreed upon balance.


> You literally said yourself, sci and salv talk a lot because of their mutual needs. Why would adding "can you get pka modkits asap?" to the list of mutual exchanges not result in them actually doing it? You're overestimating how long it takes to hit tier 2 techs when sci focuses on them.

well if youre just adding another thing to the to do list then you arent really facilitating increased dept interaction, are you? the list of things salv comes to beg sci for is just 5 things instead of 4 now. i fail to see how this meaningfully improves gameplay in any way for either role. and for what? to make the pka, salv main weapon, needlessly weaker and throw off the hp balance of every salv enemy?

TytosB:
> also as a side-comment be sure to make branches off of your master instead of your previous branches

thanks, sorry. not used to making non-mapping prs.

deltanedas:
> all of which can be fought with the aformentioned grappling-hook-in-space method.

relying solely on the fact that simplemob AI cant cope with space

in ss13 pka oneshots carp in low pressure anyway so why aim for 3 anyway

TytosB:
aight heres what im thinking, re buff the damage, but set the range to something like lifetime .22 i think this would still be short enough to make the range modkit worth it, while being long enough to make the pka not a glorified melee weapon, as the current range of only 4ish tiles is very unforgiving in my opinion

as for the size change, i still feel like adding an extra two squares will just increase storage clutter and inventory shuffling and not really improve gameplay in the intended way, as the change of a 2x1 is basically just one small item, enough to be slightly annoying but not really engaging. i get the intention but it having to carefully manage inventory is, personally, an issue i already struggle with as a salv, and im one of those nutters that uses a duffle bag.

superjj18:
It really doesn't feel like the nerf was very well thought out, and as a result made what should have been a hyped new feature(mod kits) and turned it into a lame and disappointing nerf.

rebalancing to more of a "slight optional buff" rather than a "major non-optional nerf with slight optional relief" is an easy way to restore the hype that was lost from the mod kits with the initial change.

As sad as I am to see it go, I get the range nerf, but the other two mod kits better be worth it

EmoGarbage404 (changes requested):
After a bit of pondering, i'd say bring the damage back up but keep the size the same. The range can maybe come up a tile or two but i wouldn't want it back to how it was before. It should be a decidedly mid-range weapon.

Pronana:
Finally got around to using it and yeah I think the nerfs were way too heavy handed.

The biggest drawback in my opinion is the range it's way too short. 

The shorter range makes it a borderline melee weapon on laggier servers as mobs will often hit you from what feels like further away, even in dev environment if you circle a mob and shoot at it, it can still occasionally strike at you due to ping differences, and that's just one mob. If you're fighting multiple you're effectively forced to cheese them or have a ton of meds in your bag, it really sucks. It needs at least a full tile more range if not two.

Damage just feels bad at the moment. It's sluggishly slow to break windows and asteroids and it makes me want to just bring a baseball bat instead.

TytosB:
tweaked the range to .22, roughly 5.5 tiles in testing. size set to 2x3, damage unchanged

EmoGarbage404 (approved):
Fine with me

slarticodefast (approved):

PR #37013, opened 2025-04-28
make scrubber widenet in panic mode

### LINKED ISSUE #37008: scrubbers in panic mode should have widenet turned on
so it could panic more quickly

lzk228:
## About the PR

title

## Why / Balance

so it could panic faster fixes #37008

## Technical details

one-liner

## Media

no

## Requirements

## Breaking changes

none

**Changelog**

:cl:
- tweak: Scrubber now will turn on widenet in panic mode.

### DIFF
--- a/Content.Shared/Atmos/Piping/Unary/Components/SharedVentScrubberComponent.cs
+++ b/Content.Shared/Atmos/Piping/Unary/Components/SharedVentScrubberComponent.cs
@@ -62,7 +62,7 @@ namespace Content.Shared.Atmos.Piping.Unary.Components
             FilterGases = new(GasVentScrubberData.DefaultFilterGases),
             PumpDirection = ScrubberPumpDirection.Siphoning,
             VolumeRate = 200f,
-            WideNet = false
+            WideNet = true
         };
 
         public static GasVentScrubberData ReplaceModePreset = new GasVentScrubberData


ArtisticRoomba:
An actual about/why section:

Currently there is a massive noobtrap with scrubbers. Most people would expect that an air alarm set to panic would drain the room quite quickly of air. This is not the case and people instead resort to spacing rooms using the RCD which is a bad habit and should be killed in the future (it'll be highly discouraged when Flowmos is real).

Panic mode scrubbers on widenet work much faster, which allows them to be a viable option without having to tell everyone "make sure you turn panic mode on". It's just something that should be the default instead of some weird hear-say knowledge.

Generally this is a good change for Atmos, thinking in the future.

Unfortunately this is not a bugfix so it needs two approvals.

ArtisticRoomba (approved):

slarticodefast (approved):
I always thought it already did that.
Good change.

ScarKy0:
I was about to press merge :(

VerinSenpai:
Y'all make me so happy 😭

ArtisticRoomba:
> I was about to press merge :(

shouldn't have done in-game testing then :godo:

deltanedas:
air alarms never automatically use panic and AI can set wide+all gases anyway

ArtisticRoomba:
> The argument was that you should get the faster fix if you have an atmos tech run around and actually deal with issues. It sounds like this is no longer a concern?

What delta said, and you still have to do this.

PR #37020, opened 2025-04-28
Better Godmode

Pronana:
## About the PR
Added an event to cancel destruction and a check in the gib system to stop if what's being gibbed has godmode.
Rod can also no longer delete entities with godmode.

## Why / Balance
Oversight in godmode functionality I assume based on comment from Emo. Also prevents you from grinding the smite snail.

## Technical details
DestructionEventArgs now has a cancelled bool. 

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: You can no longer grind the smite snail


IProduceWidgets (approved):

pheenty:
Godmode should also make you not slow down from temperature (or maybe rather make your temperature contant normal temperature), and I'm pretty sure it still doesn't do that.

IProduceWidgets:
> Godmode should also make you not slow down from temperature (or maybe rather make your temperature contant normal temperature), and I'm pretty sure it still doesn't do that.

hmmm, thats gonna be weedy because temperature is tied to homeostasis is tied to atmos.

EmoGarbage404 (approved):

Content.Shared/Destructible/SharedDestructibleSystem.cs
```diff
@@ -26,7 +28,7 @@ public void BreakEntity(EntityUid owner)
 /// <summary>
 ///     Raised when entity is destroyed and about to be deleted.
 /// </summary>
-public sealed class DestructionEventArgs : EntityEventArgs
+public sealed class DestructionEventArgs : CancellableEntityEventArgs
```
EmoGarbage404, line 31: instead of destruction event args being cancellable, you should add a new cancellable event that is raised before this one. With the present logic, you can only cancel the deletion, and not the actual effects of DestructionEventArgs, which is probably unwanted.

EmoGarbage404 (changes requested):

Content.Server/Kitchen/EntitySystems/ReagentGrinderSystem.cs
```diff
@@ -123,10 +125,8 @@ public override void Update(float frameTime)
                         if (solution.Volume > containerSolution.AvailableVolume)
                             continue;
 
-                        var dev = new DestructionEventArgs();
-                        RaiseLocalEvent(item, dev);
-
-                        QueueDel(item);
+                        if(!_destructible.DestroyEntity(item))
+                            continue;
```
EmoGarbage404, line 129:
```suggestion
                        _destructible.DestroyEntity(item);
```
got lost in the sauce

EmoGarbage404 (approved):

PR #37029, opened 2025-04-29
Increase benchmark BuildTimeout

ElectroJr:
## About the PR
Makes benchmarks useable again. Currently the build always times out on my & sloth's machine.
Also changes the default map, seeing as atlas has been removed.

I'm also using this PR to add a new benchmark, because its not really worth creating a separate PR for that.


metalgearsloth (approved):

Content.Benchmarks/ComponentQueryBenchmark.cs
```diff
@@ -29,7 +29,7 @@ namespace Content.Benchmarks;
 [CategoriesColumn]
 public class ComponentQueryBenchmark
 {
-    public const string Map = "Maps/atlas.yml";
+    public const string Map = "Maps/saltern.yml";
```
slarticodefast, line 32: why change it to saltern again?
it is deprecated and not updated anymore

ElectroJr: The actual map doesn't really matter, just about any map that has a decent range of entity prototypes on it would work, though being somewhat smaller would help run the benchmarks faster. So I literally just sorted by size and used the first map that I recognized as an actual game map
explorer_vvU6aBjU0R

I don't play the game enough to be familiar with relic & reach, and centcomm is not a typical game map.

EmoGarbage404 (approved):
lgtm

PR #37030, opened 2025-04-29
Show other speso colours, add larger denominations (Frontier#1496)

whatston3:
## About the PR

Speso stacks of different sizes now display differently, with the layer of the largest denomination under its count being shown.

The sprites from Frontier were also ported, including larger bills, and an S with yellow and orange instead of a yellow and pastel yellow.

## Why / Balance

Upon request.

Different values of cash having immediate recognition is helpful to distinguish between small and large amounts (is this $1 bill a stack of 3 or a stack of 2 million spesos?).  Having different sprites for amassing larger amounts of cash feels better for the player.

## Technical details

A new LayerFunction field is added to the StackComponent telling which, if any, function is applied on the stack amounts before selecting a layer. Adds a new component, StackThresholdComponent, for storing threshold data.  This field and component are set on the SpaceCash entity.

Compared with the original Frontier PR, the data structure was dropped in favour of ref parameters (only used in local functions, no real point).

## Media

Each of the bills is shown in-game below.

https://github.com/user-attachments/assets/60be6e6a-b53f-4ed4-bdb8-74fb2ec636c3

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Speso stacks now show different denominations depending on their size.


whatston3:
@MilonPL, as requested.

Content.Client/Stack/StackSystem.cs
```diff
@@ -66,10 +66,60 @@ private void OnAppearanceChange(EntityUid uid, StackComponent comp, ref Appearan
...
+
+        /// <summary>
+        /// Adjusts the actual and maxCount to change how stack amounts are displayed.
+        /// </summary>
+        /// <param name="uid">The entity being applied.</param>
+        /// <param name="comp">The </param>
+        /// <param name="actual">The actual number of items in the stack. Altered depending on the function to run.</param>
+        /// <param name="maxCount">The maximum number of items in the stack. Altered depending on the function to run.</param>
+        /// <returns>Whether or not a function was applied.</returns>
+        private bool ApplyLayerFunction(Entity<StackComponent> ent, ref int actual, ref int maxCount)
```
ArtisticRoomba, line 86: XMLDoc doesn't actually match the params

Content.Client/Stack/StackSystem.cs
```diff
@@ -66,10 +66,60 @@ private void OnAppearanceChange(EntityUid uid, StackComponent comp, ref Appearan
...
+        /// Selects which layer a stack applies based on a list of thresholds. Each threshold passed results in the next layer being selected.
+        /// </summary>
+        /// <param name="comp">The threshold parameters to apply.</param>
+        /// <param name="actual">The number of items in the stack. Will be set to the index of the layer to use.</param>
+        /// <param name="maxCount">The maximum possible number of items in the stack. Will be set to the number of selectable layers.</param>
+        private static void ApplyThreshold(StackLayerThresholdComponent comp, ref int actual, ref int maxCount)
+        {
+            // We must stop before we run out of thresholds or layers, whichever's smaller. 
+            maxCount = Math.Min(comp.Thresholds.Count + 1, maxCount);
+            int newActual = 0;
```
ArtisticRoomba, line 112: Use var

Content.Shared/Stacks/StackComponent.cs
```diff
@@ -95,4 +102,13 @@ public StackComponentState(int count, int? maxCount, bool lingering)
...
         }
     }
+
+    public enum StackLayerFunction
+    {
+        // No operation performed.
+        None,
+        // Arbitrarily thresholds the stack amount for each layer.
+        // Expects entity to have the StackLayerThresholdComponent.
+        Threshold
```
ArtisticRoomba, line 112: You can use <summary> XMLDocs on enum states, use them here

Content.Shared/Stacks/StackThresholdComponent.cs
```diff
@@ -0,0 +1,13 @@
...
+[RegisterComponent]
+public sealed partial class StackLayerThresholdComponent : Component
+{
+    /// <summary>
+    /// A list of thresholds to check against the number of things in the stack.
+    /// Each exceeded threshold will cause the next layer to be displayed.
+    /// Should be sorted in ascending order.
+    /// </summary>
+    [DataField(required: true)]
+    public List<int> Thresholds = new List<int>();
```
ArtisticRoomba, line 12: You can just use a collection expression here.

```suggestion
    public List<int> Thresholds = [];
```

ArtisticRoomba (changes requested):
Minor nitpicks, everything else looks good.

Content.Shared/Stacks/StackThresholdComponent.cs
MilonPL: I personally think using `new()` is more explicit and better than collection expressions.

ArtisticRoomba: Sure, but the `List<int>` part is redundant. You can just do `new ();`

Either way works.

Content.Shared/Stacks/StackComponent.cs
```diff
@@ -95,4 +102,13 @@ public StackComponentState(int count, int? maxCount, bool lingering)
             Lingering = lingering;
         }
     }
+
+    public enum StackLayerFunction
```
slarticodefast, line 106:
```suggestion
    [Serializable, NetSerializable]
    public enum StackLayerFunction
```

whatston3: Dumb question for you - haven't looked into Lidgren at all - if a type is serialized into a packet (say it was this enum being sent on its own), is there some unique identifier that it receives as being NetSerializable?  Can you ever "run out of space" for NetSerializable things?  Hypothetically?  In practice?

whatston3:
Changes applied.

EmoGarbage404 (approved):

MilonPL (changes requested):

Content.Shared/Stacks/StackThresholdComponent.cs
```diff
@@ -0,0 +1,13 @@
+namespace Content.Shared.Stacks;
+
+[RegisterComponent]
+public sealed partial class StackLayerThresholdComponent : Component
```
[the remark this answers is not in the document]
slarticodefast: also add documentation to the component please

whatston3:
Requested changes should be applied.

MilonPL (approved):

ArtisticRoomba (approved):

Content.Shared/Stacks/StackComponent.cs
whatston3: it is a mystery

PR #37038, opened 2025-04-29
Add Cyborg crew indicator

SlamBamActionman:
## About the PR

This PR adds a green border to all jobs that are considered "crew" in default crewsimov, visible to cyborg/AI.

## Why / Balance

Helps clarify who are crew in default crewsimov, as that's otherwise relegated to rules only. Helps with some confusion that may arrive with e.g. agent ID and nukies. This overlay shares the space used for mindshields, which should be fine because there shouldn't be a situation where they overlap.

The coloring is a bit understated compared to the mindshield since silicon will see these a lot and don't need the visual distraction, as well as to not push for it too hard when there's a law change.

Silicon rules definition will need to be adjusted to reference this eventually. 

This PR does not remove the border if there's a law change.

## Technical details

Instead of making this a separate component, I've opted to include this in `ShowJobIconsSystem`. This is because the border relies on the job icon information, which is already being handled in that system.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Cyborgs now see an indicator for jobs that are considered "crew" under the normal crewsimov lawset.


crazybrain23:
> I'd just give them a crew manifest too at this rate unless we want Operator Uniform to be able to order borgs around simply because they have a Clown icon (same going with visitors since the V icon is ambigious)

This PR just reinforces the existing enforcement from admins about what HUD icons counted as crew. The only other thing we might do is clarify in the rules that the green border means crew.

PicklOH:
> I'd just give them a crew manifest too at this rate unless we want Operator Uniform to be able to order borgs around simply because they have a Clown icon (same going with visitors since the V icon is ambigious)

I thought this was intended. Nukies are crew with Clown IDs? Agent ID (which everyone buys as a Nukie) let you pick whatever so Nukies are almost always crew.

walksanatora:
i feel like it should be to the side a-la the syndicate vizor cause I know of atleast one downstream that adds silicons with eyeglass slots. and secglasses would conflict. (or if there is a eventual security borg with built-in sec glasses. same downstream is considering detective borgs.)

ScarKy0:
> I'd just give them a crew manifest too at this rate unless we want Operator Uniform to be able to order borgs around simply because they have a Clown icon (same going with visitors since the V icon is ambigious)

Thats very much intentional.
I think ALL icons should be crew. Silicons are meant to be strictly neutral.

SlimmSlamm:
I don't like this. Cyborgs are cool because they consider anyone with an ID on station as crew. This includes nukies. It makes their gameplay much more interesting and more of a Free Agent rather than just a fully aligned crew member non antag.

MozarteanChaos:
> I don't like this. Cyborgs are cool because they consider anyone with an ID on station as crew. This includes nukies. It makes their gameplay much more interesting and more of a Free Agent rather than just a fully aligned crew member non antag.

From what I can tell, this PR would still make nukies with agent IDs count as crew - it just makes this clear mechanically, rather than expecting borg players to remember which job icons are crew and which ones aren't, because some job icons are _already_ not considered crew.

SlamBamActionman:
This PR still expects borgs with changed laws to follow them based on their interpretation of the laws, and not the border. Even if it says "Only Wardens are crew" the green border will still persist for all other station jobs, and it is up to the player to act according to their laws.

SlamBamActionman:
The PR has been updated to make the border change yellow for *all* HUD icons if there is a law change. For AI, this happens on any non-Crewsimov law board, and goes back to green if Crewsimov is reinserted. 

It has also been expanded to the outside of the HUD icon, to be compatible with mindshields.

walksanatora:
I vote visitors should NOT be crew. a bit rude to wash up on a station and start ordering their borgs. if you wanna be crew go to the HoP and get actualized with propper accesses.

superjj18:
What about for non-crewsimov lawsets where these indicators would be incorrect? 

Robocop for instance doesn't care about crew or non-crew.

superjj18:
> I vote visitors should NOT be crew. a bit rude to wash up on a station and start ordering their borgs. if you wanna be crew go to the HoP and get actualized with propper accesses.

By this logic, passengers shouldn't be considered crew either :trollface: get a job tider

walksanatora:
passengers are officially sent to the station to be there. visitors just show up in space nearby and just join.

Content.Client/Overlays/ShowJobIconsSystem.cs
```diff
@@ -16,11 +17,44 @@ public sealed class ShowJobIconsSystem : EquipmentHudSystem<ShowJobIconsComponen
     [ValidatePrototypeId<JobIconPrototype>]
     private const string JobIconForNoId = "JobIconNoId";
 
+    [ValidatePrototypeId<SecurityIconPrototype>]
+    private const string CrewBorder = "CrewBorderIcon";
+    [ValidatePrototypeId<SecurityIconPrototype>]
+    private const string CrewUncertainBorder = "CrewUncertainBorderIcon";
```
slarticodefast: Use `ProtoId<SecurityIconPrototype>` instead of a string.
Also I'm not certain, but I think `ValidatePrototypeId` is completely deprecated.

Content.Client/Overlays/ShowJobIconsSystem.cs
```diff
@@ -16,11 +17,44 @@ public sealed class ShowJobIconsSystem : EquipmentHudSystem<ShowJobIconsComponen
...
+
+    [ViewVariables]
+    public bool IncludeCrewBorder = false;
```
slarticodefast: Does VV even do anything here?

Content.Shared/Overlays/ShowJobIconsComponent.cs
```diff
@@ -5,5 +5,12 @@ namespace Content.Shared.Overlays;
 /// <summary>
 ///     This component allows you to see job icons above mobs.
 /// </summary>
-[RegisterComponent, NetworkedComponent]
-public sealed partial class ShowJobIconsComponent : Component { }
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState(raiseAfterAutoHandleState: true)]
+public sealed partial class ShowJobIconsComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public bool IncludeCrewBorder = false;
```
slarticodefast: 9wfqto

Content.Server/Silicons/Laws/SiliconLawSystem.cs
```diff
@@ -34,6 +35,9 @@ public sealed class SiliconLawSystem : SharedSiliconLawSystem
     [Dependency] private readonly UserInterfaceSystem _userInterface = default!;
     [Dependency] private readonly EmagSystem _emag = default!;
 
+    [ValidatePrototypeId<SiliconLawsetPrototype>]
+    private const string DefaultCrewLawset = "Crewsimov";
```
slarticodefast: use ProtoId
also this should not be hardcoded here
some borgs don't have this as a default

Content.Server/Silicons/Laws/SiliconLawSystem.cs
```diff
@@ -303,6 +307,11 @@ protected override void OnUpdaterInsert(Entity<SiliconLawUpdaterComponent> ent,
 
         while (query.MoveNext(out var update))
         {
+            if (TryComp<ShowJobIconsComponent>(update, out var jobIconComp))
+            {
+                jobIconComp.UncertainCrewBorder = DefaultCrewLawset != provider.Laws;
+                Dirty(update, jobIconComp);
+            }
```
slarticodefast: doesn't this only work for the AI?

Content.Client/StatusIcon/StatusIconOverlay.cs
```diff
@@ -106,8 +106,8 @@ protected override void Draw(in OverlayDrawArgs args)
                         accOffsetR += texture.Height;
                         countR++;
                     }
-                    yOffset = (bounds.Height + sprite.Offset.Y) / 2f - (float)(accOffsetR - proto.Offset) / EyeManager.PixelsPerMeter;
-                    xOffset = (bounds.Width + sprite.Offset.X) / 2f - (float)texture.Width / EyeManager.PixelsPerMeter;
+                    yOffset = (bounds.Height + sprite.Offset.Y) / 2f - (float) (accOffsetR - proto.Offset) / EyeManager.PixelsPerMeter;
```
slarticodefast: wrong convention
we don't use a space behind a cast
check if your .editorconfig is being read correctly, your IDE should complain about this
```suggestion
                    yOffset = (bounds.Height + sprite.Offset.Y) / 2f - (float)(accOffsetR - proto.Offset) / EyeManager.PixelsPerMeter;
```
same for the other cases

slarticodefast (dismissed):
Instead of making this part of `ShowJobIconsComponent` it might be better to separate this into its own component like `ShowMindShieldIconsComponent` does.

SlamBamActionman:
> Instead of making this part of `ShowJobIconsComponent` it might be better to separate this into its own component like `ShowMindShieldIconsComponent` does.

I thought of this, but it would need to go into the same system regardless since getting the job eligibility requires subscribing to the same 
`SubscribeLocalEvent<StatusIconComponent, GetStatusIconsEvent>(OnGetStatusIconsEvent);`
and we aren't be able to have duplicate subscriptions with StatusIconComponent. 

It should be possible to make it its own component and TryComp it, but it would still be in the `ShowJobIconsSystem`.

Content.Server/Silicons/Laws/SiliconLawSystem.cs
SlamBamActionman: Yes, but that is because that is the AI's lawboard insertion event.

For regular borgs it is handled in the EnsureSubvertedSiliconRole function. AIs with changed lawboards do not count as subverted and therefore needs the extra check.

Content.Server/Silicons/Laws/SiliconLawSystem.cs
SlamBamActionman: I am not sure I agree with this on account of the code of the selected highlighting is specifically designed for Crewsimov. Other lawsets may have a different definition of crew and should therefore not be covered by it.

hollowarc:
It would work best if icons only changed to yellow if the laws getting changed included the word "crew" so everyone isnt turned yellow when the clown must drink gargleblasters to live or if the lawset doesnt even involve crew in which there should have no indicator

slarticodefast:
Hey, I got a branch where I fixed the merge conflict and tried doing some cleanup, especially getting rid of the update loop and making the job icon event based:
https://github.com/slarticodefast/space-station-14/tree/pr/37038
Turns out that is more complicated than expected though, since a new id card spawning inside a pda is not covered by this (to test it give the first client a sec hud and connect with a second client, which will then have no job icon). If you manage to fix that it might be a little cleaner.
But I'll leave it up to you if you want to use this or not.

Otherwise I found one more bug while testing, the med HUD is shifted to the left from your position changes:
grafik

SlamBamActionman:
> Hey

Thanks for giving this PR a bit of a poke while it was in a merge conflict state; I'll give it a re-look!

SlamBamActionman:
Oh no

SlamBamActionman:
I've fixed the bug that Slart noticed; additionally, xenoborgs have the icon border (though they're ofc yellow by default), Nukie borgs don't, but the derelict assault borg does since it shows crew icons.

crazybrain23 (approved):
I can say with confidence that most, if not all, admins would approve of this change. This is how we currently enforce the "crew" rule internally and would make things much clearer to players.

Princess-Cheeseballs:
Was gonna say "I'm an admin/maint so I can just review and merge" but:
1. merge conflicts
2. another maint already approved?

It's good to merge once conflicts are solved it seems since it has maint + admin approval.

ToastEnjoyer (changes requested):
I think this should be added to the guidebook

K-Dynamic:
Personally I would prefer a green check and red cross in a checkbox for colourblind players, but this works for now.

PicklOH:
> I think this should be added to the guidebook

A guidebook and rule PR should be atomized here imo. Let the work make it in and we can worry about wording for those later?

IProduceWidgets:
My only design critique remains that it conflicts with the mindshield ui. When we discussed it many months ago it was pointed out that theres no current standard way to have both huds on upstream.
Of course I believe some pointed out that's not the case for downstreams. As well as being a possible easy place to misspecify a prototype in the future.

Mostly adding this for posterity since I don't think theres any intent to change that, but also the PR is so old now that requesting a ui redesign feels unsavory.

Kowlin:
> My only design critique remains that it conflicts with the mindshield ui. When we discussed it many months ago it was pointed out that theres no current standard way to have both huds on upstream. Of course I believe some pointed out that's not the case for downstreams. As well as being a possible easy place to misspecify a prototype in the future.
> 
> Mostly adding this for posterity since I don't think theres any intent to change that, but also the PR is so old now that requesting a ui redesign feels unsavory.

Worst case this could always be changed later, but right now this would be a amazing QoL improvement for game admins.

PR #31902, opened 2024-09-06
Tweak - Reflective vest and Energy Katana should reflect only in correct slots

BIGZi0348:
## About the PR

Added new case for reflect component to take into account - positioning, because:

Equippable items, like swords, that reflect bullets and lasers should do it only in your hands.

Equippable items, like armor, that reflect bullets and lasers should do it only while you wearing them.

## Why / Balance

Reflective vest should not work as a OP riot laser shield.

Energy katana should give you a feeling like you are cool ninja deflecting bullets and stuff, slicing them in mid air. Not a lucky frying pan from PUBG.

## Technical details

Added a bool value InRightPlace and value Placement to ReflectComponent.
Placement denotes where should reflective item reflect.

Added method IsInRightPlace() that's returns true if item is in right place and false otherwise.
It's placed in OnReflectEquipped and OnReflectHandEquipped.

Added check to TryReflectProjectile and TryReflectHitscan methods for new logic.

## Media

Current work of reflective properties on Energy Katana and Reflective Vest:

https://github.com/user-attachments/assets/7fc0848b-a579-4fea-8de4-4d4fd7409b68

New work of reflective properties on Energy Katana and Reflective Vest:

https://github.com/user-attachments/assets/51ac35ea-21a7-4ca0-9d78-7ad8992d3c87

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- tweak: Reflective vest now reflects lasers only while equipped.
- tweak: Energy katana now reflects lasers and projectiles only while in hands.

PR #37039, opened 2025-04-29
minor ReflectionSystem refactor

slarticodefast:
## About the PR
Supercedes https://github.com/space-wizards/space-station-14/pull/31902
Does the same, but simplifies the logic a lot at the same time instead of making it more complicated.
This allows reflective vests only to reflect when equipped and energy katanas only when holding them in your hands.

## Why / Balance
no more holding the energy vest as an OP laser shield
no more having the katana in your belt while it reflects

## Technical details
I removed the custom logic the system has for relaying the event to the inventory via an enumerator. This was messy because we had to add or remove a component to the wearer depending on where it got equipped and check all the other slots every time we did. Now it's just a bool and uses regular relay events.
This also saves us a few resolves in the helper functions.
We use the `SubscribeWithRelay` helper function to relay the event to both the hands and inventory slots. The `baseEvent` is subscribed with another method that does not check if the item is equipped correctly, so that directly shooting at a reflective entity still reflects on its own.

## Requirements

## Breaking changes
Removed `ReflectUserComponent`. This was only internally used by the system.
No API changes.

**Changelog**
:cl: BIGZi0348, slarticodefast
- tweak: Reflective vest now reflects lasers only while equipped.
- tweak: Energy katana now reflects lasers and projectiles only while in hands.


pheenty:
opened 3 minutes ago
merge conflict
speedrun
image

beck-thompson:
I feel like this is just a bandaid to fixing the actual issue which is to make hitscans entities :thinking:
It would be harder but allow for a lot cooler stuff and make this not ugly

slarticodefast:
> I feel like this is just a bandaid to fixing the actual issue which is to make hitscans entities 

I mean, that was there before. I just made it use relay event so we can check if the item is in the correct inventory slot.

Content.Shared/Weapons/Reflect/ReflectComponent.cs
```diff
@@ -13,23 +15,48 @@ public sealed partial class ReflectComponent : Component
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool ReflectingInHands = true;
+
+    /// <summary>
+    /// Can only reflect when placed correctly.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool InRightPlace;
+
```
EmoGarbage404, line 39: this seems to be unused? isn't this a leftover from the other PR?

EmoGarbage404 (approved):

Content.Shared/Weapons/Reflect/ReflectComponent.cs
slarticodefast: the equip/unequip subscriptions set it and it is checked by the relay in `OnReflectUserCollide` and `OnReflectUserHitscan`

BIGZi0348:
Pulled your branch for some local tests and yeah, it's behaving incorrectly.

Energy katana not reflecting in hands.

https://github.com/user-attachments/assets/016750cc-25db-4586-9c66-48a75ba353c0

slarticodefast:
Good catch! Fixed.
The relay for the hands slot was not set up. Still a negative diff PR though😄 
example

BIGZi0348:
Then, It is done.

PR #37057, opened 2025-04-30
Species are now picked at random in the developer environment!

beck-thompson:
## About the PR

Title! Species are now chosen at random round start when in dev mode.

## Why / Balance

There are many many species specific bugs that come up (See #37054 for a recent example). A lot of the time this is due to the dev environments character just not being one that fully tests all edge cases. Having them be random allows for easier testing of features with a wide variety of species! This wont catch all bugs, but it hopefully will reduce this issue by at little bit.

## Technical details

Just enabled the CCvar in dev mode and also added some additional logic so it will just randomly choose a round start species instead of something from a specific table if `ICRandomSpeciesWeights` is left blank.

## Requirements

## Breaking changes

Species are now picked at random when in development mode! If you don't want this feature, turn `random_characters` to `false` in the `development.toml` config file.

### DIFF
--- a/Content.Server/Station/Systems/StationSpawningSystem.cs
+++ b/Content.Server/Station/Systems/StationSpawningSystem.cs
@@ -140,8 +140,29 @@ public sealed class StationSpawningSystem : SharedStationSpawningSystem
         if (_randomizeCharacters)
         {
             var weightId = _configurationManager.GetCVar(CCVars.ICRandomSpeciesWeights);
-            var weights = _prototypeManager.Index<WeightedRandomSpeciesPrototype>(weightId);
-            speciesId = weights.Pick(_random);
+
+            // If blank, choose a round start species.
+            if (string.IsNullOrEmpty(weightId))
+            {
+                var roundStart = new List<ProtoId<SpeciesPrototype>>();
+
+                var speciesPrototypes = _prototypeManager.EnumeratePrototypes<SpeciesPrototype>();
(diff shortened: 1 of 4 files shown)


slarticodefast:
Yeah, I have seen many instances of species specific bugs being overlooked because they weren't tested for since you always spawn with the same character. The one mentioned above made slimes crash the game when toggling an item and it has been around for a week before it was found.
While this does not replace good integration test coverage, it will help identify these issues faster across the development team.

walksanatora:
yes but also at the same time. what if I want to test ONE specific species. wont this PR prevent me from being able to quickly launch in as the species (since it is set as my char)

beck-thompson:
> yes but also at the same time. what if I want to test ONE specific species. wont this PR prevent me from being able to quickly launch in as the species (since it is set as my char)

Don't worry about that! You can just edit `Resources/ConfigPresets/Build/development.toml`
and set `random_characters = false` when your doing that and it'll let you select characters like normal!

slarticodefast:
> yes but also at the same time. what if I want to test ONE specific species. wont this PR prevent me from being able to quickly launch in as the species (since it is set as my char)

Like Orks said, either disable the randomisation or spawn an Urist and take it over which only takes a few seconds.

EmoGarbage404 (approved):

kosticia:
I have spawned as vox without mask 👍

EmoGarbage404:
image

working as intended

PR #36697, opened 2025-04-18
Resprited the Chief Engineer's mantle/manica

august-sun:
## About the PR
Resprited the Chief Engineer's mantle to become the Chief Engineer's manica.

## Why / Balance
Mantles are a recognition of a player's time spent working in command, and as such, should have the option/opportunity to more distinctly stand out. The manica is an example of this, an asymmetric cosmetic that allows the CE to stand out.

In short: drip or drown, CE

## Technical details
Swapped sprites with new versions, updated meta.json.

## Media
image
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:

- tweak: Resprited the Chief Engineer's mantle.

PR #37060, opened 2025-04-30
Revert "Resprited the Chief Engineer's mantle/manica"

EmoGarbage404:
#About this PR
Reverts space-wizards/space-station-14#36697

The CE mantle resprite had some issues that we'd like to see addressed. In the meantime, we're reverting to the old sprite.

image

Core complaints:
- shoulders mess up silhouette and make them look almost hunchback, and also doesn’t connect properly with most outfits
- makes the arm have weird internal shape (aka. the segments and the fact it covers up the whole arm) alongside layering over the hand in a way that looks strange
- asymmetrical in a way that it messes up the character’s balance (almost like half of their body is 2x heavier)
- looks unsupported like it’s going to fall off the character’s shoulders at any moment (too far out/not enough straps)
- some clothing sprites clip outside the mantle


august-sun:
> If we did want to bring this sprite back as a new mastery item, it would be good to create a new category so it doesn't replace the CE mantle
> 
> The new sprite is okay imo but I liked the old one more.

I'd be okay with this if we wanted to just keep the mantles uniform!

PR #37066, opened 2025-04-30
Decal Spawners

southbridge-fur:
## About the PR

Added decal spawners! Spawners that spawn decals for mapping, round start variation, and whatever else you might want to use them for!

## Why / Balance

Map makers are often frustrated with having to place dirt decals, since they can be tedious and clog up workflows. 
But not anymore! 
With decal spawners, dirt, plants, and whatever else we may think of can now be automated away!
These spawners operate like any other spawners, in that they can be placed in maps to spawn decals round-start with plenty of variation. 

I've also taken the liberty to implement some round-start variations involving some of these spawners, just so all maps can have some level of starting dirt. 

Also as a janitor main I like having something to do on maps where I actually have to clean things, and having random variations helps include more variety between rounds where some areas will be more dirty than others. 

## Technical details

Added a new Component
- `RandomDecalSpawnerComponent`

With the following parameters

- `Decals` - A list of decals to select from randomly.
- `Radius` - The radius to use for spawning decals.
- `Prob` - The probability that a decal spawns when an attempt is made to spawn one.
- `MaxDecals` - The Maximum amount of decals to spawn.
- `MaxDecalsPerTile` - The maximum amount of decals to spawn per tile in the Radius.
- `RandomRotation` - Whether to give the decal a random rotation.
- `SnapRotation` - Whether to snap the rotation to 90 degrees.
- `SnapPosition` - Whether to snap the position to the center of a tile.
- `zIndex` - The zIndex for the decals.
- `Color` - The color for the decals.
- `RandomColorList` - A list of colors to select from, if this is set it will override the `Color` parameter.
- `Cleanable` - Whether the decals are cleanable or not.
- `TileWhitelist - A whitelist for tiles to only spawn decals on. Overrides the blacklist.
- `TileBlacklist` - A blacklist for tiles to avoid spawning on.
- `DeleteSpawnerAfterSpawn` - Whether to delete the entity after the decals are spawned.

## Media

This is the current list of spawners I've added. Plenty more could be added but this is just the list I've made. They're fairly easy to make with plenty of options for folks who want them.
image

Spawners can be given tile blacklists, which prevents them from spawning on particular tiles. 
image

Some examples of round start variation
image
image
image
image

For folks who aren't familiar with yaml, these are the options that get exposed to map makers in the spawners, allowing for a great deal of customization.
image

## Requirements

## Breaking changes

**Changelog**

:cl: Southbridge
MAPS:
- add: Added Decal Spawners


Resources/Prototypes/Entities/Markers/Spawners/Random/Decals/dirt.yml
```diff
@@ -0,0 +1,153 @@
+- type: entity
+  parent: DecalSpawnerBase
+  id: DecalSpawnerDirtBase
+  abstract: true
+  components:
+  - type: RandomDecalSpawnerDistributed
+    decals:
+    - DirtLight
+    - DirtyMedium
```
southbridge-fur, line 9: I gotta fix this

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    /// <summary>
+    /// A list of tile names to avoid placing decals on.
+    /// </summary>
+    /// <remarks>
+    /// Note that due to the nature of tile-based placement, it's possible for decals to "spill over" onto nearby tiles.
+    /// This is mostly so dirt decals don't go on diagonal tiles that won't work for them.
+    /// </remarks>
+    [DataField]
+    public List<String> TileBlacklist = new();
```
Killerqu00, line 69: Raw strings aren't supposed to be used here, I would probably suggest using List<ProtoId<ContentTileDefinition>> or List<ProtoId<ITileDefinition instead>>

Killerqu00: image
github ate my comment. i want a refund immediately

southbridge-fur: done

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
```diff
@@ -0,0 +1,77 @@
+using Robust.Shared.Prototypes;
+using Content.Shared.Maps;
+
+namespace Content.Server.Spawners.Components;
+
+public abstract partial class RandomDecalSpawnerComponent : Component
```
slarticodefast, line 6: add documentation to the componentes themselves

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    /// <summary>
+    /// A list of decals to randomly select from when spawning.
+    /// </summary>
+    [DataField]
+    public List<String> Decals = new();
```
slarticodefast, line 12: use proto ids instead of strings

Content.Server/Spawners/Components/RandomDecalSpawnerDistributedComponent.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Robust.Shared.Prototypes;
+
+namespace Content.Server.Spawners.Components;
+
+[RegisterComponent, EntityCategory("Spawner")]
+public sealed partial class RandomDecalSpawnerDistributedComponent : RandomDecalSpawnerComponent
```
slarticodefast, line 6: the inheritance here is sus and probably not needed
just combine the two spawners

Content.Server/Spawners/EntitySystems/RandomDecalSpawnerSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<RandomDecalSpawnerDistributedComponent, MapInitEvent>(OnMapInit);
+        SubscribeLocalEvent<RandomDecalSpawnerScatteredComponent, MapInitEvent>(OnMapInit);
+    }
+
+    public void OnMapInit(EntityUid uid, RandomDecalSpawnerDistributedComponent component, MapInitEvent args)
```
slarticodefast, line 28: use `Entity<Component>` pattern
same for all the others
and subscription methods should be private

Content.Server/Spawners/EntitySystems/RandomDecalSpawnerSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+            QueueDel(uid);
+    }
+
+    public bool TrySpawn(EntityUid uid, RandomDecalSpawnerDistributedComponent component)
+    {
+        if (!TryComp<RandomDecalSpawnerDistributedComponent>(uid, out var comp))
+            return false;
+
+        if (component == null)
+            return false;
```
slarticodefast, line 48: just use `Resolve` and make the parameter nullable
`comp` and `component` are identical
and the null check is not needed because TryComp already returns false if it was

Content.Server/Spawners/EntitySystems/RandomDecalSpawnerSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+        if (component.Decals.Count == 0)
+            return false;
+
+        var tileBlacklist = new List<ITileDefinition>();
+        if (component.TileBlacklist.Count > 0)
+        {
+            tileBlacklist = GetTileDefs(component.TileBlacklist);
+        }
+
+        // I feel like there's a better way to do this, but I can't find it
```
slarticodefast, line 59: yeah, this looks sus

Content.Server/Spawners/EntitySystems/RandomDecalSpawnerSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+        return true;
+    }
+
+    public bool TrySpawn(EntityUid uid, RandomDecalSpawnerScatteredComponent component)
+    {
+        if (!TryComp<RandomDecalSpawnerScatteredComponent>(uid, out var comp))
+            return false;
+
+        if (component == null)
+            return false;
```
slarticodefast, line 101: same as above

slarticodefast (commented):
I did not look at everything, so not a full review, just a few quick comments

Content.Server/Spawners/Components/RandomDecalSpawnerDistributedComponent.cs
southbridge-fur: alright

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
southbridge-fur: no can do, the decal system expects strings https://github.com/space-wizards/space-station-14/blob/213a9ef36d5dc48e4304eadd11402000bf767210/Content.Server/Decals/DecalSystem.cs#L285

slarticodefast: they will automatically cast back into a string
the benefit of a protoId is that it gets validated for existance by the linter

southbridge-fur: alright done

southbridge-fur:
alright now all the changes have been addressed

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// Whether decals should snap to 90 degree orientations, does nothing if RandomRotation is false.
+    /// </summary>
+    [DataField]
+    public bool SnapRotation = false;
+
+    /// <summary>
+    /// Whether decals should snap to the center omf a grid space or be placed randoly.
+    /// </summary>
+    [DataField]
+    public bool SnapPosition = false;
```
EmoGarbage404, line 63: you have a handful of these fields that set values on the decals, but i think they should include a nullable case for using the default decal values. That way, you don't need to manually specify that dirt snaps to grid, as the dirt decals have that behavior by default.

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// <summary>
+    /// A list of tile prototype IDs to avoid placing decals on.
+    /// </summary>
+    /// <remarks>
+    /// Note that due to the nature of tile-based placement, it's possible for decals to "spill over" onto nearby tiles.
+    /// This is mostly so dirt decals don't go on diagonal tiles that won't work for them.
+    /// </remarks>
+    [DataField]
+    public List<ProtoId<ContentTileDefinition>> TileBlacklist = new();
```
EmoGarbage404, line 97: this definitely should also have a whitelist option

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
```diff
@@ -0,0 +1,104 @@
...
+    public int MaxDecals = 1;
+
+    /// <summary>
+    /// The maximum amount of decals to spawn within a tile.
+    /// </summary>
+    /// <remarks>
+    /// A value <= 0 is considered unlimited.
+    /// </remarks>
+    [DataField]
+    public int MaxDecalsPerTile = 0;
```
EmoGarbage404, line 45: it'd be more standard to use null to indicate the base case rather than <= 0

Content.Server/Spawners/EntitySystems/RandomDecalSpawnerSystem.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            if (comp.Prob < 1f && _random.NextFloat() > comp.Prob)
+                continue;
+
+            // The vector added here is just to center the generated decals to the tile the spawner is on.
+            var localPos = xform.Coordinates.Position + _random.NextVector2(comp.Radius) + new Vector2(-0.5f, -0.5f);
+            var position = new EntityCoordinates(xform.GridUid.Value, localPos);
+
+            var tileRef = _map.GetTileRef(xform.GridUid.Value, grid, position);
+            var tileRefStr = tileRef.ToString();
```
EmoGarbage404, line 65: i dont think this checks if the tile you've gotten is just space, which would be an issue.

Resources/Prototypes/Entities/Markers/Spawners/Random/Decals/dirt.yml
```diff
@@ -0,0 +1,157 @@
...
+    - DirtLight
+    - DirtMedium
+    - DirtHeavy
+    maxDecalsPerTile: 1
+    snapPosition: true
+    zIndex: 1
+    prob: 0.8
+    color: '#FFFFFF7F'
+    cleanable: true
+    tileBlacklist: # Everything here just looks bad if it gets regular dirt on it
```
EmoGarbage404, line 17: see: my comment on a whitelist. we have a lot more non-square tiles than normal square ones.

EmoGarbage404 (changes requested):

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
southbridge-fur: I can never trust the default values myself so I don't use them. Plus using default values would require someone using this to know what the default value of the tile does rather than simply set it themselves in this component.

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
southbridge-fur: Fair, I just thought I'd get dinged for having both

Content.Server/Spawners/Components/RandomDecalSpawnerComponent.cs
southbridge-fur: I mean I could allow null as well but in what case would anyone set this value to null if it always defaults to 0?

southbridge-fur: Like this wouldn't be overridden by a default value from the decal, this codebase already uses 0 to indicate infinite or unlimited in plenty of other components, and anyone using this would have no reason to set a null value? I'll still change it I'm just really confused as to why.

Content.Server/Spawners/EntitySystems/RandomDecalSpawnerSystem.cs
southbridge-fur: If I had to look up every single side effect and edge case of every single method I called I'd never get anything done. And this little bit of code took me two days to write because I had to troll through so much to be able to put any of this together. I'm begging for documentation.

Resources/Prototypes/Entities/Markers/Spawners/Random/Decals/dirt.yml
southbridge-fur: Surprisingly, we don't. I started writing out the whitelist and it already surpassed this list in length even ignoring all the concrete tiles.

southbridge-fur: The best way to handle this might just be to add a field to the tile prototypes themselves which indicates the types of decals which can be placed on them. That however isn't easy since it would require changes to RobustToolbox for something in SS14

ArtisticRoomba:
mario middle finger jpeg pointing towards trailing whitespace check

PR #37069, opened 2025-04-30
Bomb Deliveries

ScarKy0:
## About the PR

Added the bomt type delivery.
It start ticking down when unlocked/broken/expired, the longer it takes to blow up the higher the chance it does so.

## Why / Balance

Feature for mail, also a very funny thing that can happen like once per game.

## Technical details

DeliveryBombComponent, the neccessary sprites and layers. You get the deal.

## Media

https://github.com/user-attachments/assets/3e4f2048-e0ff-40f4-a969-825894471724

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Deliveries can now very rarely spawn as bomb-type! They grant A LOT of spesos... at a price.


lzk228:
hey cap you got a letter from... Urist McKaczynski?...

SlimmSlamm:
I think this would work well as an Uplink Item as well with letting you change the name and job like an Agent ID.

ps3moira:
way too long, should blow up the instant it's opened

ScarKy0:
> way too long, should blow up the instant it's opened

People keep saying its unfun(though I agree with you, it would be better)
If anything can be balanced out later

superjj18:
> way too long, should blow up the instant it's opened

Regular mail should have a small chance of being a hidden bomb, maybe not instant but <5 seconds

TeenSarlacc:
> Regular mail should have a small chance of being a hidden bomb, maybe not instant but \u003C5 seconds

>open up envelope 
>primed singularity grenade
>:godo:

Content.Shared/Delivery/SharedDeliverySystem.cs
```diff
@@ -259,6 +259,14 @@ public void UpdateBrokenVisuals(Entity<DeliveryFragileComponent> ent, bool isFra
         _appearance.SetData(ent, DeliveryVisuals.IsFragile, isFragile);
     }
 
+    public void UpdateBombVisuals(Entity<DeliveryBombComponent> ent)
+    {
+        if (!HasComp<DeliveryComponent>(ent))
+            return;
```
slarticodefast, line 265: is this needed here?

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -148,6 +220,29 @@ private void UpdatePriorty(float frameTime)
...
+
+        while (priorityQuery.MoveNext(out var uid, out var bombData))
+        {
+            if (bombData.NextExplosionRetry > curTime)
+                continue;
+
+            bombData.NextExplosionRetry += bombData.ExplosionRetryDelay;
+
+            if (!bombData.Primed)
+                continue;
```
slarticodefast, line 237: you might want to add a `PrimedDeliveryBombComponent` to use in the query so you don't have to update the timespans for all of them

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -148,6 +220,29 @@ private void UpdatePriorty(float frameTime)
...
+
+            if (_random.NextFloat() < bombData.ExplosionChance)
+                _explosion.TriggerExplosive(uid);
```
slarticodefast, line 240: this will desynch between server and client
use randompredicted once it's merged into the engine

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -119,12 +130,73 @@ private void OnGetFragileMultiplier(Entity<DeliveryFragileComponent> ent, ref Ge
...
+
+    private void OnExplosiveBreak(Entity<DeliveryBombComponent> ent, ref BreakageEventArgs args)
+    {
+        if (!ent.Comp.PrimeOnBreakage)
+            return;
+
+        PrimeBombDelivery(ent);
+    }
+
+    private void PrimeBombDelivery(Entity<DeliveryBombComponent> ent)
```
slarticodefast, line 181: I think you can make this one a public API function in case forks want to use it
and make it check if it's already primed

Content.Shared/Delivery/DeliveryBombComponent.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// How often will this bomb retry to explode.
+    /// </summary>
+    [DataField]
+    public TimeSpan ExplosionRetryDelay = TimeSpan.FromSeconds(5);
+
+    /// <summary>
+    /// The time at which the next retry will happen
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public TimeSpan NextExplosionRetry;
```
slarticodefast, line 30: time offset serializer, autopausedfield

slarticodefast (dismissed):

Content.Shared/Delivery/DeliveryModifierSystem.cs
ScarKy0: Should all the explosion logic be on PrimedDelivery and all the priming logic stay on DeliveryBomb then?

Content.Shared/Delivery/SharedDeliverySystem.cs
ScarKy0: Just a check to know whether the DeliveryBomb is for sure added to a Delivery, in case its added somewhere else by accident or stuff

slarticodefast: It's probably fine without?
All it needs are the sprite layers, deliverycomp doesn't really matter for the visual update

ScarKy0: Yeah but it also ensures it has the sprite layers since all the deliveries have them, but i can remove it then

Content.Shared/Delivery/DeliveryModifierSystem.cs
slarticodefast: you can make it an empty marker component
it's mostly to optimize the entity query
you can see a lot of examples of this in the code, for example `ActiveChargerComponent`

Content.Shared/Delivery/SharedDeliverySystem.cs
slarticodefast: I would say it's better because in that case you want to get the error message telling you the layers are missing

Content.Shared/Delivery/DeliveryModifierSystem.cs
slarticodefast: I didn't say that we should ignore it.
This will either need to do the manual workaround by setting the seed, wait for random predicted or not be predicted at all. I prefer the second option. I guess we will have to annoy sloth with this

ScarKy0: Welp guess we can wait then

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -119,12 +132,86 @@ private void OnGetFragileMultiplier(Entity<DeliveryFragileComponent> ent, ref Ge
...
+        _delivery.UpdateBombVisuals(ent);
+    }
+
+    private void OnPrimedExplosiveMapInit(Entity<PrimedDeliveryBombComponent> ent, ref MapInitEvent args)
+    {
+        if (!TryComp<DeliveryBombComponent>(ent, out var bomb))
+            return;
+
+        bomb.NextExplosionRetry = _timing.CurTime;
+        Dirty(ent);
```
PJB3005, line 147: You're not modifying any networked state, and even if you were, it's on a different component than contained in `ent`. This doesn't make sense.

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -119,12 +132,86 @@ private void OnGetFragileMultiplier(Entity<DeliveryFragileComponent> ent, ref Ge
     }
     #endregion
 
+    #region Explosive
+    private void OnExplosiveMapInit(Entity<DeliveryBombComponent> ent, ref MapInitEvent args)
+    {
+        _delivery.UpdateBombVisuals(ent);
+    }
+
+    private void OnPrimedExplosiveMapInit(Entity<PrimedDeliveryBombComponent> ent, ref MapInitEvent args)
```
PJB3005, line 141: Map init isn't the right place to put this. You need component started instead. This code doesn't even run if a component is added on an existing entity, does it?

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -119,12 +132,86 @@ private void OnGetFragileMultiplier(Entity<DeliveryFragileComponent> ent, ref Ge
     }
     #endregion
 
+    #region Explosive
+    private void OnExplosiveMapInit(Entity<DeliveryBombComponent> ent, ref MapInitEvent args)
+    {
+        _delivery.UpdateBombVisuals(ent);
```
PJB3005, line 138: This should be started or initialized instead, not map init.

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -119,12 +132,86 @@ private void OnGetFragileMultiplier(Entity<DeliveryFragileComponent> ent, ref Ge
...
+        PrimeBombDelivery(ent);
+    }
+
+    [PublicAPI]
+    public void PrimeBombDelivery(Entity<DeliveryBombComponent> ent)
+    {
+        if (HasComp<PrimedDeliveryBombComponent>(ent))
+            return;
+
+        EnsureComp<PrimedDeliveryBombComponent>(ent);
```
PJB3005, line 198:
```suggestion
        if (EnsureComp<PrimedDeliveryBombComponent>(ent))
            return;
```

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -119,12 +132,86 @@ private void OnGetFragileMultiplier(Entity<DeliveryFragileComponent> ent, ref Ge
...
+
+        _delivery.UpdateBombVisuals(ent);
+
+        _ambientSound.SetAmbience(ent, true);
+
+        Dirty(ent);
```
PJB3005, line 204: Again, you didn't actually change any networked state on `ent`. Does this even do anything?

Content.Shared/Delivery/DeliveryModifierSystem.cs
```diff
@@ -148,6 +235,27 @@ private void UpdatePriorty(float frameTime)
             }
         }
     }
+
+    private void UpdateBomb(float frameTime)
+    {
+        var bombQuery = EntityQueryEnumerator<DeliveryBombComponent, PrimedDeliveryBombComponent>();
```
PJB3005, line 241: Swap the order of these around, so the game enumerates over `PrimedDeliveryBombComponent` as primary component (as those will be less numerous).

Content.Shared/Delivery/DeliveryModifierSystem.cs
PJB3005: Can we even predict explosions? This feels like a totally reasonable thing to not have prediction for.

PJB3005 (dismissed):

Content.Shared/Delivery/DeliveryModifierSystem.cs
PJB3005: Correct, I just checked the code. Explosions can't even be predicted. Just leave this code server-only.

Content.Shared/Delivery/DeliveryModifierSystem.cs
ScarKy0: It does. MapInit usually seems to be the way to go since ComponentStartup has PVS issues (also seems to run every time entity enters PVS on client, unless I misremember)

walksanatora:
I kinda wish it had a package variant. so it could spawn with a creature (mothroach) inside. so now you have a dilema. do you open the box causing it to explode but rescuing the mothroach. or do you let it sit there alone, trapped.

ScarKy0:
> I kinda wish it had a package variant. so it could spawn with a creature (mothroach) inside. so now you have a dilema. do you open the box causing it to explode but rescuing the mothroach. or do you let it sit there alone, trapped

It can be both letter AND package

Content.Shared/Delivery/DeliveryModifierSystem.cs
ScarKy0: NextExplosionRetry is networked. Though since this is just startup I think there would be no need to dirty anyways?
Also yeah, i forgot to change the target entity, oops.

walksanatora:
media didn't show it so I assumed it didn't exists. o7 cant wait to terroize my friends with a mothroach dilema.

Content.Shared/Delivery/DeliveryModifierSystem.cs
metalgearsloth: I think this one is actually fine as mapinit.

ScarKy0:
Approved by sloth over discord, so gonna self merge

PR #37077, opened 2025-04-30
Skeletons leave glove fiber evidence

### LINKED ISSUE #35161: Skeletons do not leave fibers
## Description

Apparently having no fingerprint means you do not leave behind even your fibers.
**Reproduction**

Be a skeleton. Wear insuls. Pick something up and toss it back onto the ground. Scan it with the forensic scanner. There would be no fibers.
**Screenshots**

Image
**Additional context**

Faced this on a downstream with half-organic (like skeletons) xenomorphs as a playable race that didn't have fingerprints. Then saw the same behavior with IPCs on another downstream. And from the testing that's also the case with upstream skeletons.

themias:
## About the PR

This PR allows anyone with hands to leave glove fibers, even if they are missing fingerprints.

Fixes #35161 

## Why / Balance

We want anyone touching with their hands to leave forensics, either fingerprint or glove fibers.
In the case of those without fingerprints, we'd still want to try leaving fibers.

Currently, only skeletons lack fingerprints, but in the future we might have people with prosthetic hands, burned off fingerprints, etc.

## Technical details

We register the event with HandsComponent instead of FingerprintComponent.

I've tested with cyborg, mech, and aghost. If there's an edge case involving something with hands, let me know.

## Media

https://github.com/user-attachments/assets/bd24e33d-e448-4c29-9dd7-2c58e935b971

## Requirements

**Changelog**

:cl:
- fix: Skeletons wearing gloves now leave fiber evidence

### DIFF
--- a/Content.Server/Forensics/Systems/ForensicsSystem.cs
+++ b/Content.Server/Forensics/Systems/ForensicsSystem.cs
@@ -19,6 +19,7 @@ using Content.Shared.Weapons.Melee.Events;
 using Robust.Shared.Random;
 using Content.Shared.Verbs;
 using Robust.Shared.Utility;
+using Content.Shared.Hands.Components;
 
 namespace Content.Server.Forensics
 {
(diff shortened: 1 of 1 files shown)


beck-thompson (approved):
Looks good!

PR #37079, opened 2025-04-30
Make universal access config better

Samuka-C:
## About the PR
Universal access config now starts with a universal ID and now has infinite range

## Why / Balance
As a admin is very annoying either having to spawn a universal ID to use in the universal access config or having to remove the universal ID from your PDA, also, is annoying having to be physically close to the stuff to use the universal access config

## Technical details
Some yml edit to add the universal ID to the universal access config and fixed a bug with the AccessOverrider system

## Media

https://github.com/user-attachments/assets/7fef6105-dbd0-41b7-af35-9389efa73ea9

## Requirements

## Breaking changes

**Changelog**
🆑 Samuka 
ADMIN:
- tweak: universal access config now starts with a universal ID
- fix: admins can now use any access config from any range

### DIFF
--- a/Content.Server/Access/Systems/AccessOverriderSystem.cs
+++ b/Content.Server/Access/Systems/AccessOverriderSystem.cs
@@ -201,7 +201,7 @@ public sealed class AccessOverriderSystem : SharedAccessOverriderSystem
         if (!PrivilegedIdIsAuthorized(uid, component))
             return;
 
-        if (!_tagSystem.HasTag(player, BypassInteractionRangeChecksTag) && !_interactionSystem.InRangeUnobstructed(uid, component.TargetAccessReaderId))
+        if (!_interactionSystem.InRangeUnobstructed(player, component.TargetAccessReaderId))
         {
             _popupSystem.PopupEntity(Loc.GetString("access-overrider-out-of-range"), player, player);
 
--- a/Content.Server/Access/Systems/AccessOverriderSystem.cs
+++ b/Content.Server/Access/Systems/AccessOverriderSystem.cs
@@ -4,7 +4,6 @@ using Content.Shared.Access;
 using Content.Shared.Access.Components;
 using Content.Shared.Access.Systems;
 using Content.Shared.Administration.Logs;
-using Content.Shared.Tag;
 using Content.Shared.Database;
 using Content.Shared.DoAfter;
 using Content.Shared.Interaction;
@@ -26,12 +25,9 @@ public sealed class AccessOverriderSystem : SharedAccessOverriderSystem
     [Dependency] private readonly AccessReaderSystem _accessReader = default!;
     [Dependency] private readonly ISharedAdminLogManager _adminLogger = default!;
     [Dependency] private readonly SharedInteractionSystem _interactionSystem = default!;
-    [Dependency] private readonly TagSystem _tagSystem = default!;
     [Dependency] private readonly PopupSystem _popupSystem = default!;
     [Dependency] private readonly SharedAudioSystem _audioSystem = default!;
     [Dependency] private readonly SharedDoAfterSystem _doAfterSystem = default!;
-
-    private static readonly ProtoId<TagPrototype> BypassInteractionRangeChecksTag = "BypassInteractionRangeChecks";
 
     public override void Initialize()
     {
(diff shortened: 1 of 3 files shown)


Content.Server/Access/Systems/AccessOverriderSystem.cs
```diff
@@ -197,7 +197,7 @@ private void TryWriteToTargetAccessReaderId(EntityUid uid,
         if (!PrivilegedIdIsAuthorized(uid, component))
             return;
 
-        if (!_interactionSystem.InRangeUnobstructed(uid, component.TargetAccessReaderId))
+        if (!component.infiniteRange && !_interactionSystem.InRangeUnobstructed(uid, component.TargetAccessReaderId))
```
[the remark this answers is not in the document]
Samuka-C: image

okay, here it is, the bypass is only checked when clicking stuff, but the UI of the access overrider checks for unobstructed range when you click on the UI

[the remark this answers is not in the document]
Samuka-C: there, changed

Content.Server/Access/Systems/AccessOverriderSystem.cs
```diff
@@ -197,7 +201,7 @@ private void TryWriteToTargetAccessReaderId(EntityUid uid,
         if (!PrivilegedIdIsAuthorized(uid, component))
             return;
 
-        if (!component.infiniteRange && !_interactionSystem.InRangeUnobstructed(uid, component.TargetAccessReaderId))
+        if (!_tagSystem.HasTag(player, BypassInteractionRangeChecksTag) && !_interactionSystem.InRangeUnobstructed(uid, component.TargetAccessReaderId))
```
slarticodefast, line 204: no, i mean `InRangeUnobstructed` already does check for that tag
this should not be needed at all
and if it is not working it is a bug we should fix instead

slarticodefast (changes requested):

Content.Server/Access/Systems/AccessOverriderSystem.cs
Samuka-C: well, then I guess is bugged, cause as for right now, as a admin, you can't change the access of anything using the access config unless you are really close to it. my guess is cause the bypassInteractionRange tag is only checked when you click on a entity (which means you can interact with any door at any range) but not checked when you click the UI button, since that is not a entity and the accessOverrider system checks for the range when you try to change the access. but I am not sure as I am not very familiar on the inner workings of... the game.

- solution 1: use my temporary™ fix until that is properly fixed
- solution 2: remove my fix and only merge the universal access reader starting with the universal ID (already helps a lot) and wait for the issue to be fixed by someone that knows more than me about this stuff

what you think is best?

slarticodefast: Give me a moment, I will do some testing to see if I can fix the bug.

Content.Server/Access/Systems/AccessOverriderSystem.cs
```diff
@@ -197,7 +201,7 @@ private void TryWriteToTargetAccessReaderId(EntityUid uid,
         if (!PrivilegedIdIsAuthorized(uid, component))
             return;
 
-        if (!_interactionSystem.InRangeUnobstructed(uid, component.TargetAccessReaderId))
+        if (!_tagSystem.HasTag(player, BypassInteractionRangeChecksTag) && !_interactionSystem.InRangeUnobstructed(uid, component.TargetAccessReaderId))
```
slarticodefast, line 204:
```suggestion
        if (!_interactionSystem.InRangeUnobstructed(player, component.TargetAccessReaderId))
```
there we go, this should fix the bug
revert the other changes in this file, please

can you also add 
`inHandsOnly: true` to the `ActivatableUI` component in the prototype? That prevents some weird behaviour where you can use it from a distance

slarticodefast (changes requested):

Content.Server/Access/Systems/AccessOverriderSystem.cs
Samuka-C: add to the access config proto?

Samuka-C: done

slarticodefast (approved):
Looks good.
Needs a second approval since the id card is technically not a bugfix.

Issue #37032, opened 2025-04-29
Throwing the player mispredicts a larger throw than the server does after #35843


DrSmugleaf:
## Description

**Reproduction**
- Copy xeno headbutt from rmc14
- Use it on someone else
- Watch from the target's perspective
- ![Image](https://github.com/user-attachments/assets/594c1828-b2b1-4149-b0a5-d264af098cf0)

This broke 25 usages of throwing in rmc14 and only a 1 hour bisect adventure told me why

**Screenshots**

https://github.com/user-attachments/assets/7c5bb9c4-9ec0-4bd5-b3a6-3b94ebc64a0c

**Additional context**
See https://github.com/space-wizards/space-station-14/pull/35843

Git patch to add an ability for testing (add XenoHeadbuttComponent to yourself)
<details>

```
Index: Resources/Prototypes/Accents/a.yml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/Resources/Prototypes/Accents/a.yml b/Resources/Prototypes/Accents/a.yml
new file mode 100644
--- /dev/null
+++ b/Resources/Prototypes/Accents/a.yml
@@ -0,0 +1,14 @@
+- type: entity
+  id: ActionXenoHeadbutt
+  name: Headbutt (10) # TODO RMC14 proper plasma costs
+  description: Charge towards the targeted enemy and push back the first enemy hit, piercing armor slightly. Does less damage but has more knockback when crest is lowered.
+  components:
+  - type: EntityTargetAction
+    itemIconStyle: NoItem
+    icon:
+      sprite: _RMC14/Actions/xeno_actions.rsi
+      state: headbutt
+    event: !type:XenoHeadbuttActionEvent
+    range: 3
+    useDelay: 6
+    deselectOnMiss: false
Index: Content.Shared/Abilities/XenoHeadbuttActionEvent.cs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/Content.Shared/Abilities/XenoHeadbuttActionEvent.cs b/Content.Shared/Abilities/XenoHeadbuttActionEvent.cs
new file mode 100644
--- /dev/null
+++ b/Content.Shared/Abilities/XenoHeadbuttActionEvent.cs
@@ -0,0 +1,5 @@
+using Content.Shared.Actions;
+
[...]

PR #37086, opened 2025-05-01
Fix throwing prediction

metalgearsloth:
- Disposals is still janky but I think that's disposals in general not being predicted and the disposals throw not being predicted and short-lived.
- Would need to check RMC.
- Couldn't repro the underlying issues however thrown items don't slip anymore so (and we also don't predict their land / stopping anymore so).

Ref event breaking change on StopThrowEvent though nothing here uses it.

Probably fixes https://github.com/space-wizards/space-station-14/issues/37032

No one notices upstream so CL pointless.

### DIFF
--- a/Content.Shared/Throwing/LandEvent.cs
+++ b/Content.Shared/Throwing/LandEvent.cs
@@ -10,8 +10,5 @@ namespace Content.Shared.Throwing
     /// Raised when a thrown entity is no longer moving.
     /// </summary>
     [ByRefEvent]
-    public record struct StopThrowEvent
-    {
-        public EntityUid? User;
-    }
+    public record struct StopThrowEvent(EntityUid? User);
 }
--- a/Content.Shared/Throwing/LandEvent.cs
+++ b/Content.Shared/Throwing/LandEvent.cs
@@ -9,7 +9,8 @@ namespace Content.Shared.Throwing
     /// <summary>
     /// Raised when a thrown entity is no longer moving.
     /// </summary>
-    public sealed class StopThrowEvent : EntityEventArgs
+    [ByRefEvent]
+    public record struct StopThrowEvent
     {
         public EntityUid? User;
     }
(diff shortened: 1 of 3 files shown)


Content.Shared/Throwing/LandEvent.cs
```diff
@@ -9,7 +9,8 @@ namespace Content.Shared.Throwing
     /// <summary>
     /// Raised when a thrown entity is no longer moving.
     /// </summary>
-    public sealed class StopThrowEvent : EntityEventArgs
+    [ByRefEvent]
+    public record struct StopThrowEvent
     {
         public EntityUid? User;
     }
```
slarticodefast, line 16:
```suggestion
    public record struct StopThrowEvent(EntityUid? User = null);
```
just give it a primary constructor

Content.Shared/Throwing/ThrownItemSystem.cs
```diff
@@ -108,8 +110,12 @@ public void StopThrow(EntityUid uid, ThrownItemComponent thrownItemComponent)
                 }
             }
 
-            EntityManager.EventBus.RaiseLocalEvent(uid, new StopThrowEvent { User = thrownItemComponent.Thrower }, true);
-            EntityManager.RemoveComponent<ThrownItemComponent>(uid);
+            var ev = new StopThrowEvent
+            {
+                User = thrownItemComponent.Thrower
+            };
+            RaiseLocalEvent(uid, ref ev);
```
slarticodefast, line 117:
```suggestion
            var ev = new StopThrowEvent(thrownItemComponent.Thrower);
            RaiseLocalEvent(uid, ref ev);
```

slarticodefast (changes requested):

slarticodefast (commented):
Tested and I couldn't find any of the old slipping predictions issues.
Self-merge after the review has been addressed.

slarticodefast (approved):
I just pushed the changes myself.

PR #36386, opened 2025-04-07
Battery (SMES/substation) interface

PJB3005:
## About the PR
Added a proper interface to SMES/substations. You can use it to turn input/output on/off, and change max charge/discharge rates. It also has a ton of numbers and generally looks like the best UI you've ever seen.

## Why / Balance
We were chatting in Discord and I lost my  shit when I realized that it's still not possible for engineering players to actually change the charge rate on SMES properly.

## Technical details
See commits.

## Media

https://github.com/user-attachments/assets/53c0bb4b-2dfa-47dc-8355-5672fd052a8b

## Breaking changes
None.

**Changelog**
:cl:
- add: SMES and substations now have an (extremely fancy) interface to control them with.

PR #37088, opened 2025-05-01
Rate limit progress bar spamming for MagicMirror color change

AryaRaiin:
## About the PR

Added LastColorChangeTick to the MagicMirrorSystem so that OnTryMagicMirrorChangeColor no longer results in progress bar spam when using the barber scissors to change hair color. Now, 5-6 progress bars at maximum are shown.

## Why / Balance

The progress bar while changing hair color could be spammed to block vision during gameplay, as demonstrated in Media.

## Technical details

Added Robust.Shared.Timing to Content.Server/MagicMirror/MagicMirrorSystem.cs to allow monitoring game ticks.
Added variable LastColorChangeTick.
LastColorChangeTick determines the last time the color change function is called and prevents it from being canceled and recreated on the same tick.

## Media

image

## Requirements

## Breaking changes

No breaking changes noted.

### DIFF
--- a/Content.Server/MagicMirror/MagicMirrorSystem.cs
+++ b/Content.Server/MagicMirror/MagicMirrorSystem.cs
@@ -12,6 +12,7 @@ using Content.Shared.Popups;
 using Content.Shared.Tag;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Prototypes;
+using Robust.Shared.Timing;
 
 namespace Content.Server.MagicMirror;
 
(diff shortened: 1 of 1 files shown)


metalgearsloth:
PJB just added more BUI tools plz use those specifically for this https://github.com/space-wizards/space-station-14/pull/36386

AryaRaiin:
The BUI tools could be used for further improvement! I'll see what I can do to incorporate them.

Even with the BUI tools PjB made, I would still strongly encourage the rate limit per game tick. It's very needed until the options of input press-release are added. Currently, the sliders spam input like crazy.

PJB3005:
> Even with the BUI tools PjB made, I would still strongly encourage the rate limit per game tick.

That is exactly what I added in my PR. 

This solution you wrote isn't even good because it:

1. Will use the first input in a tick instead of the last, which can result in dropping the final value.
2. Still spams the hell out of the network.

PJB3005 (changes requested):
See above

AryaRaiin:
Gotcha, it sounds like you've got this under control, PjB, so I'll let you handle it. Thanks!

PR #37102, opened 2025-05-01
Pointing arrow smite no longer lasts forever

perryprog:
## About the PR

The "anger pointing arrow" smite was intended to only last for five arrow points, but instead last for infinite of them. This makes that not the case.

## Why / Balance

Critical balance for the current gameplay meta or something. Also it bothered me.

## Technical details

We were previously trying to remove the component from the arrow, not the player's entity.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Pointing/EntitySystems/RoguePointingSystem.cs
+++ b/Content.Server/Pointing/EntitySystems/RoguePointingSystem.cs
@@ -32,7 +32,7 @@ namespace Content.Server.Pointing.EntitySystems
             var angering = _random.Pick(targets);
             angering.Comp.RemainingAnger -= 1;
             if (angering.Comp.RemainingAnger <= 0)
-                RemComp<PointingArrowAngeringComponent>(uid);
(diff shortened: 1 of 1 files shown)


lzk228:
add admin cl for it

lzk228 (approved):

perryprog:
Looks like there's another silly thing where the entity query is looking across maps for targets, so arrows on other maps, when someone has PointingArrowAngering, will go shoot off to some arbitrary location and blow up there. I can fix this too but... this is also just a meme smite that's meant to be explosive and messy.

Tayrtahn (approved):
Good fix. Thanks! 

Admins: if you preferred the infinite-arrows behavior, let me or another maintainer know! This change is being made so the code works as intended, but if you'd like the intended behavior to be different, we're happy to change it!

ScarKy0:
Admin opinion, never used it but it is funny. Might actually use it more since its not as destructive now.

slarticodefast (approved):
Thanks for the bugfix!

PR #32297, opened 2024-09-19
Fix Doctors Delight metabolism Rate

BramvanZijp:
## About the PR
Doctors Delight now metabolizes 0.25 units per second for both the medicine and drink aspect, which results in it using 0.5 per second as opposed to previously using 1 per second.

## Why / Balance
It currently using 1 per second is likely an unintentional sideeffect of it having both the medicine metabolism and the drink metabolism.

## Technical details
N/A

## Media
This PR does not require an ingame showcase.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl: BramvanZijp
- fix: Fixed an issue that caused Doctors Delight to metabolize twice the normal speed, which also halved its effectiveness.

PR #37106, opened 2025-05-01
Fixed holy water metabolism rate

OnyxTheBrave:
## About the PR
Fixed Holy water metabolizing at 1u a second rather than .5u a second. This is to bring it in the same way that TDD was changed in PR #32297 

## Why / Balance
Similar to the above PR, having it metabolize twice led to the unintentional effect of if being used up twice as fast as it should.

## Technical details
Moved the healing properties into one singular category to avoid "double dipping" when metabolizing.

## Media
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Holy water now correctly metabolizes at .5 unites per second

### DIFF
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -1094,19 +1094,18 @@
       effects:
       - !type:SatiateThirst
         factor: 3
-    Medicine:
-      effects:
       - !type:HealthChange
-        conditions:
-        - !type:TotalDamage
-          max: 50
-        damage:
-          types:
+         conditions:
+         - !type:TotalDamage
+           max: 50
+         damage:
+           types:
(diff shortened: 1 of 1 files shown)


EmoGarbage404 (approved):

Resources/Prototypes/Reagents/medicine.yml
```diff
@@ -1094,19 +1094,18 @@
...
-        conditions:
-        - !type:TotalDamage
-          max: 50
-        damage:
-          types:
+         conditions:
+         - !type:TotalDamage
+           max: 50
+         damage:
+           types:
```
EmoGarbage404, line 1102:
```suggestion
        conditions:
        - !type:TotalDamage
          max: 50
        damage:
          types:
```

PR #37107, opened 2025-05-02
Ichor double-metabolize fix + Very minor cleanup

Pronana:
## About the PR
Edited the yaml of ichor to combine drink and medicine metabolisms so it doesn't metabolize twice as fast as it should

## Why / Balance
I saw other yaml fixes for other reagents with the same issue and decided I should do the same

## Technical details
Minor yaml changes + removed the halving of reagents injected in the devour system and halved the reagents the dragon gets in yaml to compensate.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Devour/DevourSystem.cs
+++ b/Content.Server/Devour/DevourSystem.cs
@@ -29,8 +29,6 @@ public sealed class DevourSystem : SharedDevourSystem
         if (component.FoodPreference == FoodPreference.All ||
             (component.FoodPreference == FoodPreference.Humanoid && HasComp<HumanoidAppearanceComponent>(args.Args.Target)))
         {
-            ichorInjection.ScaleSolution(0.5f);
-
             if (component.ShouldStoreDevoured && args.Args.Target is not null)
             {
                 ContainerSystem.Insert(args.Args.Target.Value, component.Stomach);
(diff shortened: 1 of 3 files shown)


Pronana:
While I'm here, I would like to put forwards the suggestion of buffing the amount of ichor a dragon gets from eating a person significantly.

Currently, the dragon gets 7.5 ichor per person eaten, which before meant you healed 5 brute per unit split up as (12.5 blunt, 12.5 pierce and 12.5 slash) Now it's been upped to 25 of each (huge)

Except this isn't really huge, this means that two bullets from any gun hitting a dragon makes that exchange not worth it (or three pellets from a single shotgun blast). The best dragons right now kite in space and are super dodgy because you can't afford to take that damage ever and that's quite lame. 

The dragon isn't so bulky that you can't solo it with a WT and a dream (I've done it a few times before) so I don't think giving it better healing wouldn't be a horrible decision. 

In return it's space movement multiplier could probably be brought down a bit. It's currently 1.5 which means it has a movement speed of 12 in space. For comparison a person in a jet pack has a movement speed of 7.2 in space. This doesn't include snaking which is a multiplicative modifier which allows dragons to reach speeds of around 17 or higher when making big turning maneuvers.

PR #37112, opened 2025-05-02
Celliminol and H-32

Nrd-K:
## About the PR

PR adds two new medicines: Celliminol and H-32
## Why / Balance

These drugs can be another way to revive the player through cryo pod rather than cloning.

The recipes of Celliminol and H-32 are balanced due to:
1 - A rare opportunity for a player to receive a huge amount of radioactive radiation and cellular damage.
2 - Adding Omnizine to the recipe .
3 - Drugs are required exclusively for use in cryo pod.
4 - The absence of negative effects is due to the rarity of receiving radioactive or cellular damage.

The drugs themselves have already been added to one of the SS14: Sunrise projects, where they have perfectly demonstrated their usefulness.
## Media

Before:
Снимок экрана 2025-05-02 103800
The creation process Celliminol:
Снимок экрана 2025-05-02 102316
Снимок экрана 2025-05-02 102356
Снимок экрана 2025-05-02 103344
The creation process H-32:
Снимок экрана 2025-05-02 103650
Снимок экрана 2025-05-02 103727
After H-32:
Снимок экрана 2025-05-02 103832

After Celliminol:
Снимок экрана 2025-05-02 103918

The images are intended to show the performance of Celliminol and H-32
## Requirements

## Breaking changes

**Changelog**

:cl: Nrd
- add: Added Celliminol and H-32 
Celliminol (5u): Doxarubixadone [5], Omnizine: [1], Blood [10]
H-32 (4u):  Arithrazine [1], Bicaridine[1], Cryoxadone[1], Plasma[1]

### DIFF
--- a/Resources/Locale/en-US/reagents/meta/medicine.ftl
+++ b/Resources/Locale/en-US/reagents/meta/medicine.ftl
@@ -147,3 +147,9 @@ reagent-desc-potassium-iodide = Will reduce the damaging effects of radiation by
 
 reagent-name-haloperidol = haloperidol
 reagent-desc-haloperidol = Removes most stimulating and hallucinogenic drugs. Reduces druggy effects and jitteriness. Causes drowsiness.
+
+reagent-name-h-32 = H-32
+reagent-desc-h-32 = A reagent developed in the field with increased radiation.
+
+reagent-name-celliminol = celliminol
+reagent-desc-celliminol = A cryogenics chemical. A powerful chemical developed by NanoTrasen.
--- a/Resources/Locale/en-US/reagents/meta/physical-desc.ftl
+++ b/Resources/Locale/en-US/reagents/meta/physical-desc.ftl
@@ -96,3 +96,5 @@ reagent-physical-desc-holy = holy
 reagent-physical-desc-slimy = slimy
 reagent-physical-desc-neural = neural
 reagent-physical-desc-unidentifiable = unidentifiable
+reagent-physical-desc-radioactive = radioactive
+reagent-physical-desc-bloody = bloody
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -1421,3 +1421,43 @@
       - !type:AdjustReagent
         reagent: MindbreakerToxin
         amount: -3.0
+
+- type: reagent
+  id: H-32
+  name: reagent-name-h-32
+  group: Medicine
+  desc: reagent-desc-h-32
+  physicalDesc: reagent-physical-desc-radioactive
+  flavor: medicine
+  color: "#cc1a79"
+  worksOnTheDead: true
+  metabolisms:
+    Medicine:
+      effects:
+      - !type:HealthChange
+        conditions:
+        - !type:Temperature
+          max: 213.0
+        damage:
+          types:
+            Radiation: -2
+
+- type: reagent
+  id: Celliminol
+  name: reagent-name-celliminol
+  group: Special
+  desc: reagent-desc-celliminol
+  physicalDesc: reagent-physical-desc-bloody
+  flavor: medicine
+  color: "#81eb81"
+  worksOnTheDead: true
+  metabolisms:
+    Medicine:
+      effects:
+        - !type:HealthChange
+          conditions:
+          - !type:Temperature
+            max: 213.0
+          damage:
+            types:
+             Cellular: -5
--- a/Resources/Prototypes/Recipes/Reactions/medicine.yml
+++ b/Resources/Prototypes/Recipes/Reactions/medicine.yml
@@ -662,3 +662,32 @@
       amount: 1
   products:
     Haloperidol: 5
+
+- type: reaction
+  id: H-32
+  reactants:
+    Arithrazine:
+      amount: 1
+    Bicaridine:
+      amount: 1
+    Cryoxadone:
+      amount: 1
+    Plasma:
+      amount: 1
+      catalyst: true
+  products:
+    H-32: 4
+
+- type: reaction
+  id: Celliminol
+  impact: High
+  minTemp: 500
+  reactants:
+    Doxarubixadone:
+      amount: 5
+    Omnizine:
+      amount: 1
+    Blood:
+      amount: 10
+  products:
+    Celliminol: 5


SlamBamActionman:
The point of Cellular is that it's a difficult and slow to heal damage type. This is because it's dealt primarily through actions that you yourself take, adding a balancing weight to the action; e.g implant checking or eating a brain to round remove. Adding a more powerful cellular healing option would go against that, so I am against its inclusion. 

Note: If anything doxarubixadone is being watched closely if it needs to be nerfed for being too good facilitating implant checking, though we haven't gotten to that point yet.

Nrd-K:
This "more powerful cell healing method" is even harder to create, is only used in cryopods, and is meant to be used in extreme cases. Also, knowing the SS14 community, the number of players on the servers, and the rules of the different servers in the form of "skill tables", Celliminol will be quite difficult to create, since few people will want to waste Omnizine, effort, and time for a medicine that may not be needed during the entire shift (like many other cryo drugs, especially Necrosol, which is actually a replacement for many regular drugs, but is also rarely used despite the huge number of ways to get the types of damage it heals), and it will only be used in cases where the only way to revive a player is with it.

Nrd-K:
> also, please take a picture of the guidebook entry for the chemicals, their effects, and their recipe all opened. It should make understanding what the chemical is supposed to do and reading the recipe easier.

I'm sorry, here are the screenshots
Снимок экрана 2025-05-03 211224
Снимок экрана 2025-05-03 211202

Nrd-K:
@Emisse 
Hello Emisse, I'm very sorry for the ping (if I did everything right and that's what the ping is). The fact is that I noticed that it was you who made the PR about Necrosol and I would like to hear your opinion about Celliminol. (I really hope you'll forgive me for the ping)

Hrosts:
> This "more powerful cellular healing method" is even harder to create, is only used in cryopods, and is meant to be used in extreme cases. 

Extreme cases such as? The first use case of this thing I see is just being a buffed 2.5 times doxa which requires only 4 units of omnizine to heal 200 units worth of cellular damage. That's a pithy amount of Omni and is barely a wall. For this case it only speeds up healing of cellular damage, which is slow to heal by design. This definitely skirts the "cellular is a gamemech limiter for bad activity" rule, as in it takes 10 seconds plus the freeze time to heal one failed implant check worth of damage.

The second case is to heal people who somehow needs anti-cellular deadchems to be revived, which is so rare it barely matters over the first case. If you got cellular and died, you will have 2 to 4 other damage types on you that already have deadchems for them.

It's overpowered for its one use, it's mostly pointless for its other use. If it was for something like Caustic (I don't remember if there's already a deadchem for it), it would be fine. The conflict with Cellular's special role breaks it, I think.

Nrd-K:
> > This "more powerful cellular healing method" is even harder to create, is only used in cryopods, and is meant to be used in extreme cases.
> 
> Extreme cases such as? The first use case of this thing I see is just being a buffed 2.5 times doxa which requires only 4 units of omnizine to heal 200 units worth of cellular damage. That's a pithy amount of Omni and is barely a wall. For this case it only speeds up healing of cellular damage, which is slow to heal by design. This definitely skirts the "cellular is a gamemech limiter for bad activity" rule, as in it takes 10 seconds plus the freeze time to heal one failed implant check worth of damage.
> 
> The second case is to heal people who somehow need anti-cellular deadchems to be revived, which is so rare it barely matters over the first case. If you got cellular and died, you will have 2 to 4 other damage types on you that already have deadchems for them.
> 
> It's overpowered for its one use, it's mostly pointless for its other use. If it was for something like Caustic (I don't remember if there's already a deadchem for it), it would be fine. The conflict with Cellular's special role breaks it, I think.

First of all, I must clarify that there are 4 omnisin units to repair 150 damage, not 200. Secondly, I am always ready for suggestions on how to improve it by making it more complex, adding something to it, etc., so I will gladly take into account your opinion on this matter. Thirdly, I was thinking about making his prescription more complicated, but I wasn't completely sure if it was necessary, because basically this medicine on Sunrise was created to counteract new antagonists, changelings, which could cause up to 300 cell damage when killing a person, but as I see it, given the lack of antagonist data in the game itself, I think I will still increase his need for omnisine or reduce the amount of medicine produced. If there is a suggestion in what proportions this medicine can be made, I will listen to it and take it into account.

MozarteanChaos:
> First of all, I must clarify that there are 4 omnisin units to repair 150 damage, not 200.

1u of omnizine makes 5u of celliminol, which heals 50 damage. This means 4u of omnizine results in 200 genetic healing.

Nrd-K:
> > First of all, I must clarify that there are 4 omnisin units to repair 150 damage, not 200.
> 
> 1u of omnizine makes 5u of celliminol, which heals 50 damage. This means 4u of omnizine results in 200 genetic healing.

Aww... I wanted a slightly wrong outcome, you're absolutely right, I calculated it wrong, now I'll try to balance everything, thank you!

EthanQix:
Cellular damage is, by design, a very slow damage type to heal. Adding a chemical that is a straight 2.5x healing speed upgrade of Doxarubixadone runs contrary to that design, especially for a measly cost of 0.2u of Omnizine per unit.

Maybe balance that speed upgrade with some side effects ?

Nrd-K:
> Cellular damage is, by design, a very slow damage type to heal. Adding a chemical that is a straight 2.5x healing speed upgrade of Doxarubixadone runs contrary to that design, especially for a measly cost of 0.2u of Omnizine per unit.
> 
> Maybe balance that speed upgrade with some side effects ?

I am not a fan of negative effects when working with omnisin. It is still necessary to experiment with the recipe of the drug, in my opinion, or reduce the effect.

Hrosts:
> because basically this medicine on Sunrise was created to counteract new antagonists, changelings, which could cause up to 300 cell damage when killing a person
This is a fork-specific issue that comes from the fork using the damage type in the way different from the upstream. Currently, there are only a few ways to get cellular damage here:
- Failing the implant check
- Eating brains
- Drinking benzene
- Breathing frezon
None of these require a 5 damage per second heal to be treated, none of these require a deadchem. The one of these that _might_ require it deals this much damage intentionally, with knowledge and expectation that it will be a pain in the ass to heal.
You're being super polite and eager, and it's not like I'm any authority here, but I really don't see the point for this one on the upstream.

Nrd-K:
> 
> This is a fork-specific issue that comes from the fork using the damage type in the way different from the upstream. Currently, there are only a few ways to get cellular damage here:
> 
> * Failing the implant check
> * Eating brains
> * Drinking benzene
> * Breathing frezon
> 
> None of these require a 5 damage per second heal to be treated, none of these require a deadchem. The one of these that _might_ require it deals this much damage intentionally, with knowledge and expectation that it will be a pain in the ass to heal. You're being super polite and eager, and it's not like I'm any authority here, but I really don't see the point for this one on the upstream.

Based on your own words, almost any medicine for the dead is unnecessary, because all damage above 200 is inflicted with the knowledge that it will be difficult to cure. As for the ways to cause such damage, the medicine is aimed at treating such unexpected outcomes in which it is necessary, and judging by the reactions to the PR itself, there are such cases. As for the amount of treatment, it is offset by the cost of manufacturing, which I have already adjusted and it has become even more difficult than before.

Nrd-K:
Let's start with the fact that I don't know when you could accidentally raise the temperature to 200, but at the same time you couldn't also "accidentally" get huge cellular damage by drinking the same gasoline "accidentally" or just rotting somewhere in the technical tunnels "accidentally". You are talking about the same probabilities of random events as I do, only mine are less frequent. Next, if we speak in your language, we should again remove the rest of the substances that heal the dead, only to give this effect to other weaker substances (prescribe cryoxadone to heal the dead instead of necrozol, etc.), will this change the balance? I don't think it will change much, because drugs to heal the dead are hard to make anyway. By giving doxa this ability, you immediately hit your entire concept, which you have been telling me about all this time. If, from your point of view, cellular damage should take a long time to heal, being the result of your erroneous actions, etc., then why should it be completely treated with a medicine (including a dead one, of course), prepared from two substances that are brewed in the first minutes of the shift? In fact, this is a contradiction in itself. It is difficult to inflict this damage - the damage is a mistake of one's own actions - it takes a long time to heal on its own - but at the same time, everything is treated on corpses with the first available medicine made at the beginning of the shift? In this case, there is simply no logic ... And if you immediately think about complicating dox, then consider that there is the same celliminol, only now people will not have ordinary dox ...

SlamBamActionman:
Benzene and rot damage are very low chances of being a concern in treatment, which is why we have not patched them. For benzene, the damage dealt is minor per unit, meaning no antag would aim to use it in any meaningful capacity (though I think a benzene foam bomb could be fun!). For rot, the cellular damage on a corpse tends to not be very high, and even then you have to go through the whole rot treatment first anyways. Both these instances can be served by current cellular medicine.

We removed cellular damage from vent slimes because those *are* common.

As-is, I see no need for Celliminol upstream.

Hrosts:
> Let's start with the fact that I don't know when you could accidentally raise the temperature to 200

Are we playing the same game? Almost every time there's a trit/plasma fire at least one person gets ashed or close to ashing via tons of heat damage.

> Next, if we speak in your language, we should again remove the rest of the substances that heal the dead

You're just making up points that I don't agree with, even after I explained to you the difference both in how frequent the damage is, and why cellular is different from the others.

Nrd-K:
> > Let's start with the fact that I don't know when you could accidentally raise the temperature to 200
> 
> Are we playing the same game? Almost every time there's a trit/plasma fire at least one person gets ashed or close to ashing via tons of heat damage.
> 
> > Next, if we speak in your language, we should again remove the rest of the substances that heal the dead
> 
> You're just making up points that I don't agree with, even after I explained to you the difference both in how frequent the damage is, and why cellular is different from the others.

First of all, you were talking about random events, the fact that someone set the plasma on fire is hardly a coincidence in most cases. Secondly, I have already expressed my position to you, if you want to hear my answer, then look at the messages above, it has not changed. I've heard and seen your position for a long time, and even after that, I don't want to change my position, I don't understand how you haven't understood it yet. I suggest we end our conversation with you at this point, because no matter how much we prove something to each other, we will not come to a common opinion.

slarticodefast:
I agree with Slam, Cellular damage is currently used as a game mechanic to act to discourage certain behaviour, and making it easily healable bypasses that effect.
Additionally I think we should wait for a more in-depth medical system before adding more healing reagents as most current ones are similar with slightly different stats. This will need a larger rework along the future med development and I would rather not add content that makes this harder to do at the moment.

PR #37128, opened 2025-05-02
Tranexamic Metabolism Buff

archee1:
## About the PR
Makes Tranexamic-Acid last 2.5x longer and lowers its bleed-reduction and overdose to compensate

## Why / Balance
Currently tranexamic acid is rarely used, aside from the small amount in an emergency-medipen the use of the chem is almost non existent (aside making sutures).

This change should allow you to use a syringe of tranex before a gunfight and leave it without being pale, or dying due to making poor decisions by rushing to gain any value while its metabolising (due to it only lasting 30s)
With the nature of most gunfights or red alert situations you typically don't have the time to inject yourself with a syringe and it then typically only sees use post-fight in places where a single gauze would be doing its job better.

## Technical details
yaml changes
adds metabolism rate of 0.2 (from 0.5) making it last 2.5x longer
lowers bleed-rate to 0.6 (from 1.5) making it 2.5x slower (in testing this is still enough to prevent going pale after being shot from 0-100 pierce)
lowers overdose bloodloss damage to 1.5 (from 3.0) to compensate for lasting longer in metabolism

## Requirements

**Changelog**
:cl:
- tweak: Tranexamic Acid now metabolizes and clots blood slower, with a less harsh overdose.

### DIFF
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -758,11 +758,12 @@
   color: "#ba7d7d"
   metabolisms:
     Medicine:
+      metabolismRate: 0.2 # makes it viable for pre-combat rather than a burst post-combat
       effects:
       # Medium-large quantities can hurt you instead,
-      # but still technically stop your bleeding.
+      # but still stop your bleeding.
       - !type:ModifyBleedAmount
-        amount: -1.5
+        amount: -0.6
       - !type:HealthChange
         conditions:
         - !type:ReagentThreshold
@@ -1206,7 +1207,7 @@
         conditions:
         - !type:ReagentThreshold
           min: 12
-          
+
 - type: reagent
   id: Opporozidone #Name based of an altered version of the startreck chem "Opporozine"
   name: reagent-name-opporozidone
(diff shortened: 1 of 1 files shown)


minthydra:
> If we do this then we should merge the PR that removes blood clotting from puncturase

Another nerf to nukies :godo:

archee1:
> If we do this then we should merge the PR that removes blood clotting from puncturase

i knew punct does this, but is this still the case? because the guidebook and yaml for puncturase doesnt say anything about blood clotting.

also for nukies this would just mean lacing tranex into your meds at like 1:9 ratios (also the recent change to inaprov's effects make it reduce bleedrates by 0.25 for twice as long as this change to tranex would actually probably need to keep the bleed rate up to keep it relevant against just using inaprov...

slarticodefast:
Hello, and thank you for your contribution!
At the moment maintainers are facing a high backlog and unfortunately can’t review every single pull request. As this PR pertains to the addition of new content and/or balance changes deemed low-impact, we have opted to close it instead of leaving it open indefinitely.
PRs may be reopened in the future once the backlog has been reduced, and we encourage you to keep contributing in other areas of the codebase.

Issue #36314, opened 2025-04-05
DamageSpecifiers handle groups weirdly


IProduceWidgets:
## Description

DamageSpecifiers have logic to split damage amongst a group.
https://github.com/space-wizards/space-station-14/blob/af9526197a86b562bfee763eb7e68e169d64fc25/Content.Shared/Damage/DamageSpecifier.cs#L105
But there is no internal tracking of these groups from when the specifier was set.
This prevents us later from using the group as a frame of reference and forces us to act on specific types unless we want to specify the group specifically at in the method.

**Reproduction**

Use bicar > realize it heals damage equally in all brute types instead of healing any brute type up to its set group amount > guh.

**Screenshots**

**Additional context**

PR #37129, opened 2025-05-02
Add EvenHealthChange effect

EmoGarbage404:
## About the PR

Adds a new HealthChange effect that attempts to distribute damage more fairly when healing.

## Why / Balance

For the current system, if you have an effect that heals 3 brute, what you actually have is an effect that heals 1 blunt, 1 slash, and 1 piercing. This is not really what we want in a lot of cases, as it means that the effect is 1/3 as strong on single damage-type sources.

Ideally, we split it relative to the damages present. So if we have only 3 blunt damage, we heal all 3 blunt instead of only 1.

That is what this PR does.

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Omnizine no longer "wastes" healing when administered to patients without even amounts of all damage types.

### DIFF
--- a/Content.Server/EntityEffects/Effects/EvenHealthChange.cs
+++ b/Content.Server/EntityEffects/Effects/EvenHealthChange.cs
@@ -1,0 +1,136 @@
+using Content.Shared.Damage;
+using Content.Shared.Damage.Prototypes;
+using Content.Shared.EntityEffects;
+using Content.Shared.FixedPoint;
+using Content.Shared.Localizations;
+using JetBrains.Annotations;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
+
+namespace Content.Server.EntityEffects.Effects;
+
+/// <summary>
+/// Version of <see cref="HealthChange"/> that distributes the healing to groups
+/// </summary>
+[UsedImplicitly]
+public sealed partial class EvenHealthChange : EntityEffect
+{
+    /// <summary>
+    /// damage to heal, collected into entire damage groups
+    /// </summary>
+    [DataField]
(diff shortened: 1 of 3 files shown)


ArtisticRoomba:
Holy peak

IProduceWidgets:
related to this issue: https://github.com/space-wizards/space-station-14/issues/36314

IProduceWidgets:
I don't know if this is the right way to go about this. It really feels like this should be tightly coupled with DamageDicts so they can be used in places that aren't just health changes. Currently you cant really use DamageDicts with comparators because of the issue where the constructor simply spreads the damage across the group and then disposes of the information.

EmoGarbage404:
> I don't know if this is the right way to go about this. It really feels like this should be tightly coupled with DamageDicts so they can be used in places that aren't just health changes. Currently you cant really use DamageDicts with comparators because of the issue where the constructor simply spreads the damage across the group and then disposes of the information about its creation.
> 
> i.e. If i want a check for "5 of any brute" there's not a simple way to do this generically currently because when I specify a DamageDict I can never get "5 of any" I can only get "one of each", but having a way to check how the DamageDict was created, through groups or specific types, would also solve most of this problem.

the problem with this is that damage dicts aren't intended to preserve specific group information. The ability to specify groups in yaml is basically just a nice QOL feature. Internally, everything is just treated as a damage type without any more context. To be honest, i'd make more sense to move away from defining damage as groups at all, since it tends to be so unhelpful.

slarticodefast:
A damage model rework would be great, but that is a bigger project.
I think this PR is a good solution for now.

slarticodefast (approved):

PR #37131, opened 2025-05-02
Stelloxadone

ScarKy0:
## About the PR

Added stelloxadone, a cryochem used to cure posion and radiation at the cost of genetic damage.

## Why / Balance

Adding a way to cure poison/rads without the need for necrosol.
Also in an ideal world cryogenics would be used mainly for curing dead people. Shoving someone into a cryotube while theyre alive is just kinda annoying due to having to wait to heat up, so nobody wants to.
After stellox is added cryox will be basically fine to turn into a asphyx/bloodloss chem that works on the dead, instead of the cold omnizine it is.

## Technical details

Yaml.

## Media

image
(Yes this is just frontier)

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- add: Added Stelloxadone! A cryochemical used to treat poison and radiation damage in living patients as well as corpses, though it comes at a cost of genetic damage.

PR #37132, opened 2025-05-03
Cryoxadone Tweaks

ScarKy0:
## About the PR

Cryoxadone was tweaked to work on the dead. However it only heals asphyx, bloodloss and restores blood.

## Why / Balance

Cryoxadone is currently overtuned. It's basically omnizine at no downside other than not working on the dead. It makes cryochems boring and tasteless as it does everything on it's own.
This PR should be paired with #37236 and #37131 for best effect, otherwise cryox will just lose and nothing else will be gained.

## Technical details

Yaml.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Cryoxadone now works on the dead.
- tweak: Cryoxadone has been reworked. It no longer heals basic damage types. Instead it heals asphyxiation, bloodloss and increases blood level.

### DIFF
--- a/Resources/Locale/en-US/reagents/meta/medicine.ftl
+++ b/Resources/Locale/en-US/reagents/meta/medicine.ftl
@@ -14,7 +14,7 @@ reagent-name-bicaridine = bicaridine
 reagent-desc-bicaridine = An analgesic which is highly effective at treating brute damage. It's useful for stabilizing people who have been severely beaten, as well as treating less life-threatening injuries.
 
 reagent-name-cryoxadone = cryoxadone
-reagent-desc-cryoxadone = Required for the proper function of cryogenics. Heals all standard types of damage, but only works in temperatures under 213K. It can treat and rejuvenate plants when applied in small doses.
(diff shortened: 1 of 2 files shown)


ScarKy0:
Same deal as with the two other linked PRs you get the idea :godo:
This PR wouldn't work without the other two anyways.

superjj18:
Rest in peace the first meaningful change to medbay since the death of cloning over a year ago.  I wish I was being hyperbolic, but I genuinely cant think of anything other than paramedic lights and sirens, some health analyzer upgrades like the medtek cartridge, and some minor tweaks here and there.

Genuinely very sad for Med related changes, with very little light to be seen at the end of this tunnel.   If the maints don't have something planned for the foreseeable future of medbay then its hard not to view this revert as a massive fumble given the bleak state of medbay content.

Cryo will continue to fail to be viable, and nothing will change despite the community waiting literally meme-levels of time(two more weeks) for a med update.  At what point does it become legitimately better to just start asking for forks to port their med changes to upstream?(which is what was supposed to happen for newmed but it seems that too is unlikely at best)

PR #37136, opened 2025-05-03
Nerf science magboots

EmoGarbage404:
## About the PR

Science magboots are a research item that makes what is supposed to be a limited piece of job equipment into a common station item. The fact that basically everyone can get a set of magboots that are on-par with engineer's and salvager's equipment is a really nasty piece of power creep.

This PR adds a battery charge system to the science magboots to make them less overpowered and not just an infinitely printable version of normal magboots.

## Why / Balance

If the whole appeal of science magboots is being mass-printable, they shouldn't simultaneously be equal to rare roundstart magboots without any further investment.

## Technical details

n/a
## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Magboots printed in science now require power-cells to function.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Shoes/magboots.yml
+++ b/Resources/Prototypes/Entities/Clothing/Shoes/magboots.yml
@@ -67,13 +67,21 @@
     stealGroup: ClothingShoesBootsMagAdv
 
 - type: entity
-  parent: [ClothingShoesBootsMagBase]
+  parent: [ClothingShoesBootsMagBase, PowerCellSlotMediumItem]
   id: ClothingShoesBootsMagSci
+  name: experimental magboots
+  description: Not every experiment is a success...
   components:
   - type: Sprite
     sprite: Clothing/Shoes/Boots/magboots-science.rsi
   - type: Clothing
     sprite: Clothing/Shoes/Boots/magboots-science.rsi
+  - type: PowerCellDraw
+    drawRate: 6
+  - type: ClothingSpeedModifier
+    walkModifier: 0.75
+    sprintModifier: 0.7
+  - type: ToggleCellDraw
 
 - type: entity
   parent: ClothingShoesBootsMagBase


Spangs04:
You could give it a "charge" system in that they can only be turned on for a minimal time as the internal charge drains, and then require a "recharge" timer for the boot to recharge (and it can't activate during this) to full before it can activate again. Almost like the mass-produced ones can only hold a magnetized charge for a limited amount of time before they need to go on cooldown. That way, they are not as strong as the normal magboots, but they would not have to require batteries, as I feel them having a slow self-recharge would make them worse, but not completely gutted. I do agree that it needs to change with how magboots are currently being produced.

Djungelskog2:
> Science magboots are a research item that makes what is supposed to be a limited piece of job equipment into a common station item. The fact that basically everyone can get a set of magboots that are on-par with engineer's and salvager's equipment is a really nasty piece of power creep.

Isnt the point of science being able to print them for the sake of the whole "Being able to make the station from scratch" thing? Ive also never really seen issues with it and I certainly wouldnt call it powercreep if its literally just the same item

Djungelskog2:
If the issue is them being mass producible then wouldn't the cost or even research tier be the area to aim for?

superjj18:
Who is this change for lmao

Like who actually wants this

metalgearsloth:
> Who is this change for lmao
> 
> Like who actually wants this

Make a useful comment or, better yet, don't comment at all.

YoungThugSS14:
> Who is this change for lmao
> 
> Like who actually wants this

It’s a cool change to me tbh

pheenty:
Magboots are nearly useless, 20% slowdown is huge. I'd better buff engi magboots to 15%

Pronana:
> I recommend looking at minimum speed levels (the lowest your move speed can get) rather than tweaking the numbers.

Minimum speed is applied only for certain components iirc. Having a good minimum speed system would require a movement refactor although probably not a large one.

dvir001:
If anything, you should probably make all magboots require power and just give them other drain rates.
I mostly just find it weird only one magboots type require the cell when no others do.

Or probably add a note on Advanced / Normal / Bloodred about not needing cells from some internal power recharging.

Pumkin69:
magboots use cases are not that important other than nukies and in that case i dont know why we would buff lube ops by nerfing crew no slips even further.

Everturning:
> > Who is this change for lmao
> > Like who actually wants this
> 
> Make a useful comment or, better yet, don't comment at all.

idk i think it's a useful question to ask who the audience for this change is. who is complaining about magboots being printable from sci? if the intention is to make the game more fun, how does this change make the game more fun?

EmoGarbage404:
i still think it's wack but also i don't care enough for it.

kosticia:
He didn't delete the branch!!!! He will come back with bigger nerf!!1!

PR #16021, opened 2023-05-02
Northstar Gloves

HerCoyote23:
First PR!

## About the PR

Adds the gloves of the north star! They can be bought at the uplink for 8 tc, and allow you to punch very fast, while saying a battlecry.
Stats:
Atkspeed = 4
Damage = 8 Blunt
DPS = 32

By interacting with them inhand, you can set your battlecry. Every time you attack, you ~~'say' it as a popup (having you actually say it was causing too much adminlog spam)~~
you say it, but it's not sent to admin or chat log.

In adding north star gloves, this PR also adds the MeleeSpeech component, which can be varedited onto any melee weapon.
In addition to that, now there's a way to 'say' while having it not appear on adminlogs(to reduce spam), and a way to make activateable UIs only activate with the right click menu, not useinhand.
**Media**

https://user-images.githubusercontent.com/131214189/235754146-dc46241a-1b0a-4d58-bb93-1c53038a2b69.mp4
(Yes the video is watermarked, I had to compress it and didn't want to make an account for that website)

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl: HerCoyote23
- add: Added Gloves of the North Star

PR #37141, opened 2025-05-03
Northstar rebalance

beck-thompson:
## About the PR

Northstars now have a chance to hit peoples clothing right off their bodies! It also does less damage than before as well.

## Why / Balance

Right now, northstars are really only used to gib people (boring!). This PR will give them a powerful and unique ability that is much more fun. Hopefully this will let them be purchased and used in more interesting ways!

## Technical details

Really just adds the `ThrowEquippedOnHitComponent` and a basic system that listens for hit events!

## Media

https://github.com/user-attachments/assets/29470426-814c-4a13-b44f-0b9a12c41ebe

## Requirements

**Changelog**

:cl:
- tweak: Northstars damage has gone from 8 -> 2.5 blunt. Northstars now have a chance to throw targets clothing off their body!
- add: Boxing gloves now have a 50% chance to hit northstar gloves out of targets hands.


Pronana:
I like the component but I don't think reducing the number of options a Syndie melee user has is a good idea. Could be fun on a new and unique item though.

minthydra:
> I like the component but I don't think reducing the number of options a Syndie melee user has is a good idea. Could be fun on a new and unique item though.

Fun and good pr but again dito maybe if you made these into a brand new item maybe colour them red and white instead of blue and white :godo:

Djungelskog2:
pretty funny but I think super-quick disarming and more tactical stuff should be saved for CQC techniques

kosticia (commented):
Pretty sure that its can be predicted. Also a smaaaal nitpick.

Content.Server/Weapons/Melee/ThrowEquippedOnHitSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<ThrowEquippedOnHitComponent, MeleeHitEvent>(OnSpeechHit);
+    }
+
+    private void OnSpeechHit(Entity<ThrowEquippedOnHitComponent> ent, ref MeleeHitEvent args)
```
kosticia, line 24: Why its called On SpeechHit?

SlimmSlamm:
This just removes what the item was originally supposed to be? There are objectives to RR people. You can't just remove those items that do it and call them boring items when being RR'd is an intentional mechanic. You should instead just make a new item with the disarm ability instead of removing options.

beck-thompson:
> This just removes what the item was originally supposed to be? There are objectives to RR people. You can't just remove those items that do it and call them boring items when being RR'd is an intentional mechanic. You should instead just make a new item with the disarm ability instead of removing options.

Here is the original PR: #16021. Nowhere in the PR did it mention that it's main purpose was to gib targets.

If you guys really think that having a 8 second gib tool is important, a new item should be made that specifically does that so you don't need to be forced into buying northstars. I am sure that if instead of this, I just changed the damage type to slash the usage rate would go to zero as the esword is just better in the vast majority of situations.

Content.Server/Weapons/Melee/ThrowEquippedOnHitSystem.cs
beck-thompson: copy paste gone wrong.

For the prediction, I didn't see much use in it for two reasons:
1.) Predicting randomness is super annoying
2.) Throwing is unpredicted anyway so it wouldn't look better or anything.

Also, there were a lot of odd glitches when its done in shared related to virtual items so I didn't want to bother :laughing:

If you think Its important I can investigate more, but my personal opinion is that it isn't worth

Archisteks:
> > This just removes what the item was originally supposed to be? There are objectives to RR people. You can't just remove those items that do it and call them boring items when being RR'd is an intentional mechanic. You should instead just make a new item with the disarm ability instead of removing options.
> 
> Here is the original PR: #16021. Nowhere in the PR did it mention that it's main purpose was to gib targets.
> 
> If you guys really think that having a 8 second gib tool is important, a new item should be made that specifically does that so you don't need to be forced into buying northstars. I am sure that if instead of this, I just changed the damage type to slash the usage rate would go to zero as the esword is just better in the vast majority of situations.

400BluntDamage/32DPS = 12.5 seconds?
Also, there ARE plenty of other gibbing tools, some of which are even faster than northstars (Minibombs explode in 5 seconds, and holoparasites have 36 Blunt DPS). Others include the suspicious toolbox (2TC, also includes free tools, insuls, welder protection, and flash immunity), baseball bat (a bit slow, but free) 5L oxygen tank (very slow, but hits silently), chemist flamethrower mix in an extinguisher (doesn't even leave blood or organs), truncheon via emagged protolathe, and I'm probably forgetting a bunch of others.

In that entire list of options, the main purpose of northstars to me is as a high DPS melee weapon that can't get dropped if you get stunned, cuffed, or shoved. It synergizes well with the freedom implant, leaving you with little combat downtime.

beck-thompson:
> 400BluntDamage/32DPS = 12.5 seconds?

This is what I get for:
1.) Doing math incorrectly because I'm silly
2.) Testing with a timer :laughing:
(I was also timing from dead, not from full health)

> In that entire list of options, the main purpose of northstars to me is as a high DPS melee weapon that can't get dropped if you get stunned, cuffed, or shoved. It synergizes well with the freedom implant, leaving you with little combat downtime.

Hmmm that's fair :thinking:! I just personally never saw them used in any other way as a fast gibbing tool. I wish we had more data on this stuff so we could actually tell instead of just kind of going off what we see :laughing:

Pronana:
> Hmmm that's fair :thinking:! I just personally never saw them used in any other way as a fast gibbing tool. I wish we had more data on this stuff so we could actually tell instead of just kind of going off what we see :laughing:

Northstars are a very fun brawling weapon due to their high attack speed. I've used before during DAGD to avoid getting disarmed and they're really fun. 

Also them gibbing is kind of a flavor win considering the show they're a direct reference to.

SlimmSlamm:
> Here is the original PR: #16021. Nowhere in the PR did it mention that it's main purpose was to gib targets.

It is very clear the entire intention behind this is to provide an option for quick but loud gibbing and this is not an SS14 original item. You bringing this PR up is irrelevant.



> If you guys really think that having a 8 second gib tool is important, a new item should be made that specifically does that so you don't need to be forced into buying northstars

What? This doesn't make much sense considering you are removing the option and suggesting it to be remade. Gloves of the Northstars in SS13 don't disarm and there are mechanics in the game that we can bring over to disarm opponents instead. As others have mentioned CQC. If you want disarming to be a bigger mechanic then you should work on martial arts instead of removing options for syndicate to use.



> a new item should be made that specifically does that so you don't need to be forced into buying northstars

Nobody is being forced into buying Gloves of the Northstar. It's an option to gibb people and the only way to gibb someone fast. It's a very loud option too which balances it out. Currently Gloves of the Northstar are like the 20th most bought item in the past 15 days it's not OP and it is a 12 second gibb.

As another stated there are better options than the Gloves of the Northstar but they're still a fun option to use and this is not an item I would see anyone use at all. I'd much rather see martial arts which has the intention of what you want to do with this PR.

Content.Server/Weapons/Melee/ThrowEquippedOnHitSystem.cs
kosticia: Yea, shared random sucks, I take my words back

PicklOH:
>Hmmm that's fair 🤔! I just personally never saw them used in any other way as a fast gibbing tool. I wish we had more data on this stuff so we could actually tell instead of just kind of going off what we see 😆
>

I see Northstars used for brawling far more than I see them used for gibbing. The minibomb is the prime culprit for that.

> Nobody is being forced into buying Gloves of the Northstar. It's an option to gibb people and the only way to gibb someone fast. It's a very loud option too which balances it out. Currently Gloves of the Northstar are like the 20th most bought item in the past 15 days it's not OP and it is a 12 second gibb.
> 
Where does everyone get the usage stats? I'm curious to look at them myself.

beck-thompson:
> Where does everyone get the usage stats? I'm curious to look at them myself.

Sadly due to various reasons there isn't any easy way to get it, currently you just have to ask on the discord but any maint can give you the stats for the last x days or whatever :)

keronshb:
> It is very clear the entire intention behind this is to provide an option for quick but loud gibbing and this is not an SS14 original item. You bringing this PR up is irrelevant.

Naw. Northstars don't exist to gib in most 13 servers. The implementation in 14 wasn't even to make a loud gibbing tool and the PR doesn't even mention it.

Entvari:
2.5 damage seems too low in my opinion, stripping equipment is very cool mechanically but this makes the time to kill way too low, especially against unarmed targets where the new benefit would be functionally irrelevant.
If it still retains its 4 attack speed that's only 10dps, an 8 TC syndicate weapon should not be out DPS'd by baseball bats.
Syndie Melee is already pretty stale, taking away the actual damaging factor of the 1 out of 2 viable syndie melees is a bad call.

If the sole intention of this was to diversify the Northstar's usage from "viable melee that gibs" to "melee that has a unique role" then it needs to retain that whole viable thing, could have it so that it retains a similar damage to standing targets but deals reduced blunt to prone targets, etc.
Targets who are armored (the intended target of this Northstar) revision are either Captain or Sec generally, both of which possess melees (saber / baton) that massively overshadow the performance of a 10dps glove, with or without armor stripping.

beck-thompson:
> 2.5 damage seems too low in my opinion, stripping equipment is very cool mechanically but this makes the time to kill way too low, especially against unarmored targets where the new benefit would be functionally irrelevant.
If it still retains its 4 attack speed that's only 10dps, an 8 TC syndicate weapon should not be out DPS'd by baseball bats.
Syndie Melee is already pretty stale, taking away the actual damaging factor of the 1 out of 2 viable syndie melees is a bad call.

Yeah I think your correct, here my idea for an updated stat line:

image
(Top one is something like capts armor, bottom is unarmored)

Thoughts?

Entvari:
> Thoughts?

Additional Asphyx damage is certainly a solution I'd like to see in practice due to regeneration and various armors, etc, but is absolutely a step in the right direction. Good idea.

Pronana:
> Thoughts?

Armor piercing north-stars seems cool but most people on the station aren't going to have more than 40% blunt resist so in most scenarios e-sword will still be better. Especially considering how rare heat resistance is. 

Could also be interesting to give them a bit of structural considering that the blunt damage has been reduced.

HerCoyote23:
While conceptually the idea of knocking people's equipment off is cool, I think that in practice it will be __extremely annoying__ to play against, for a number of reasons. 
Whoever you're hitting has to choose between continuing the combat or just donating all their stuff to the local tiders. Plus it'll be really easy to lose track of what you've lost once the fight's over; people will realize 5-10 minutes later that they dropped their oxy tank/gloves/hat/etc.
In addition, this seems really powerful when combined with any other source of damage. Even if the weapon itself doesn't do much DPS, you can just punch until you strip their armor(which took ~1s in the video), then blast them with a pistol or whatever. This would also work in space -- strip their spacesuit, now they're forced to disengage(and probably get shot/hurt while they're trying to retrieve it)
Plus losing your stuff just isn't fun.

beck-thompson:
> While conceptually the idea of knocking people's equipment off is cool, I think that in practice it will be extremely annoying to play against, for a number of reasons.
Whoever you're hitting has to choose between continuing the combat or just donating all their stuff to the local tiders. Plus it'll be really easy to lose track of what you've lost once the fight's over; people will realize 5-10 minutes later that they dropped their oxy tank/gloves/hat/etc.
In addition, this seems really powerful when combined with any other source of damage. Even if the weapon itself doesn't do much DPS, you can just punch until you strip their armor(which took ~1s in the video), then blast them with a pistol or whatever. This would also work in space -- strip their spacesuit, now they're forced to disengage(and probably get shot/hurt while they're trying to retrieve it)
Plus losing your stuff just isn't fun.

Just for reference (It isn't really clear anywhere) its ~3% chance to drop any particular piece of equipment so it takes about 33 hits or 6.6 seconds on average to hit the armor off. Do you think this should be less or just not a thing at all?

RedBookcase:
> While conceptually the idea of knocking people's equipment off is cool, I think that in practice it will be _extremely annoying_ to play against, for a number of reasons. Whoever you're hitting has to choose between continuing the combat or just donating all their stuff to the local tiders. Plus it'll be really easy to lose track of what you've lost once the fight's over; people will realize 5-10 minutes later that they dropped their oxy tank/gloves/hat/etc. In addition, this seems really powerful when combined with any other source of damage. Even if the weapon itself doesn't do much DPS, you can just punch until you strip their armor(which took ~1s in the video), then blast them with a pistol or whatever. This would also work in space -- strip their spacesuit, now they're forced to disengage(and probably get shot/hurt while they're trying to retrieve it) Plus losing your stuff just isn't fun.

I don't see this as too much of an issue personally, usually if you're being attacked by someone with Northstars something like losing your Seclite or pocket o2 tank is way down on the list of priorities compared to "Killing or cuffing the person beating the shit out of me". Dropping items and having the potential for them to be stolen or taken is an interesting idea and the same thing would already happen if you end up being crit or killed as a secoff anyhow.

The only thing I would like to see exempted from the knockoff ability is probably jumpsuits and hardsuits. Losing those would be very obnoxious. Plus, punching someone for so long they individually lose the contents of both their pockets and their PDA is more funny than having them all drop when the jumpsuit's knocked off.

PicklOH:
> **Plus losing your stuff just isn't fun.**

This is the most important part.

Pumkin69:
make it a new item

beck-thompson:
Updated sheet:

image

In summery:
Time to kills are *basically* the same around the board - I don't think there is really any meaningful difference there (I think at most it would be ~1 more second to kill unarmored targets). There is an asterisk here, if you do manage to hit off someones armor the TTK could go down by a decent amount with the gloves.

The time to gib takes about ~50% longer across the board - however, if the opponent is armored it is probably a lot closer as your extremely likely to punch off their armor before 50 seconds have elapsed.

So basically, the new gloves are better against armored opponents, slightly weaker on unarmored (But again, the TTK is still extremely fast we are talking like 3 seconds to kill vs 4), and take ~50% longer to gib.

Thoughts?

PicklOH:
I personally think you should just make a new item to incorporate the knockoff effect.

It takes an additional second to strip someone of any armor that's making a difference in that sheet once you actually crit them, and you probably gain more than a second in gibbing time. Making them better against armored opponents just makes killing them all the easier.

Northstars are fine the way they are.

beck-thompson:
> I personally think you should just make a new item to incorporate the knockoff effect.

> It takes an additional second to strip someone of any armor that's making a difference in that sheet once you actually crit them, and you probably gain more than a second in gibbing time. Making them better against armored opponents just makes killing them all the easier.

The issue is that northstars are already not purchased *that* much, adding a new weapon that will be purchased the same (at best) is probably not a good idea. It would also be kind of confusing if there are two boxing gloves with extremely similar abilities in the uplink

archee1:
as much as I hate northstars they have a place in the games ethos, HOWEVER knocking peoples clothing off is both hilarious and powerful, though I think it would fit better as a unique item, since northstars are pretty well solidified as they are right now, perhaps the knuckledusters from paradise are a good thing to implement this as (they did organ damage which we don't have, though knocking clothes off is a good simulacrum for now).
ref;
https://www.paradisestation.org/wiki/index.php?title=Syndicate_Items#:~:text=for%202%20seconds.-,Syndicate%20Knuckleduster,-10%20TC

Pronana:
> Thoughts?

The numbers look alright and seem fair. Maybe could have an extremely minor damage increase since you're giving up better DPS against unarmored targets for a very minor increase in DPS against the most armored target on the station. 

Having 0.25 more blunt damage means your DPS against a secoff will still be lower than the e-sword but at 21 dps instead of 22.5 also without a reflect chance and no insuls. Gib time would also only drop by a little under 2 seconds against unarmored targets (22.8 seconds). 

Also to note, this would make skeletons immune to northstars.



As a fist of the north star fan however, I would ask for two minor things for flavor purposes:

A crumb of structural, at least enough to break reinforced windows again.

A yaml note that asphyx damage should be something cooler once we get a better med system (something akin to pressure point damage so maybe organ damage, nerve damage, or internal bleeding)

Also does is still strip people? Because I feel like that should be moved to its own weapon personally still.

archee1:
> A crumb of structural, at least enough to break reinforced windows again.

you would need more than a crumb, RGlass has flat 10 structural, most things of importance have atleast 5 or 10 which ends up really annoying to try give stuff a meaningful impact at breaking walls without it obliterating chairs and stuff.

ps3moira:
Instead of decreasing or increasing the punches,  it should just be the default punch damage of your species but let it keep it's speed of punches. I also don't think you should have your gloves be able to fly? out of your hands, that defeats their unique weapon capability.

beck-thompson:
> A crumb of structural, at least enough to break reinforced windows again.

> A yaml note that asphyx damage should be something cooler once we get a better med system (something akin to pressure point damage so maybe organ damage, nerve damage, or internal bleeding)

I tried to get this to work, but sadly it just isn't really possible. As @archee1 said, the flat 10 reduction makes it basically impossible. That really needs to change it's kind of impossible to balance stuff properly.

(And yeah, the asphyx damage note is a good idea I added that!)

beck-thompson:
> How does this PR interact with the suit clothing slot blocker PR #35172

> It'll be annoying to be unable to reequip clothing through your hardsuit while someone else can punch them off through your hardsuit

I'm pretty sure it would work like you'd think (helmet would need to be hit off first) but I'll test it when that gets merged!

Winkarst-cpu:
> I'm pretty sure it would work like you'd think (helmet would need to be hit off first) but I'll test it when that gets merged!

With my PR, you will not be able to:

1. Equip/Unequip jumpsuits and shoes while a hardsuit is equiped.
2. Equip/Unequip stuff like headsets, goggles, and masks while a helmet is equiped.

beck-thompson:
> When hardsuit helmets are knocked off, are they despawned or become interactable items?

I'm pretty sure they just can't be knocked off!

ToastEnjoyer:
This looks like a really cool feature to them! Amazing

PR #37148, opened 2025-05-03
Fix brains, borgs etc not counting as marooned

ScarKy0:
## About the PR

Borgs and brains now correctly count as marooned even while on the evac shuttle.

## Why / Balance

Unintentional change that was made along with the new maroon objective. This results in people having to dispose of the brains instead of letting them get borged and stuff, which is just a dick thing to do.

## Technical details

Added a new GetCharacterUnrevivableIcEvent which serves a similiar purpose to GetCharacterDeadIcEvent. However the main differance is it is supposed to check whether the character must receive a new body to be considered alive. (Brains are unrevivable as the body is gone, borgs are unrevivable as the orginal body cannot be revived anymore, being a zombie is considered unrevivable as you must be cloned, aka given a new body)

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Borgs, brains, zombies and similiar now correctly count as "marooned" even while on the evac shuttle.

### DIFF
--- a/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
+++ b/Content.Server/Objectives/Systems/KillPersonConditionSystem.cs
@@ -46,9 +46,6 @@ public sealed class KillPersonConditionSystem : EntitySystem
             requireMaroon = false;
         }
 
-        Log.Debug(_emergencyShuttle.IsTargetEscaping(mind.OwnedEntity.Value).ToString());
-        Log.Debug(_mind.IsCharacterUnrevivableIc(mind).ToString());
-
         if (requireDead && !targetDead)
             return 0f;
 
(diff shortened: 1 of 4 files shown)


PicklOH:
I have to comment that its just safer all around to get rid of the brain anyways because of the lack of the NLR in most cases.

ScarKy0:
> I have to comment that its just safer all around to get rid of the brain anyways because of the lack of the NLR in most cases.

Yes but we also don't wonna encourage just eating the brain.

walksanatora:
I thought that the "cannon" reason why dead was not counted for maroon was cause they could just be cloned. CC has the facilities. so zombified would be clonable at CC.

ScarKy0:
> I thought that the "cannon" reason why dead was not counted for maroon was cause they could just be cloned. CC has the facilities. so zombified would be clonable at CC.

That is still the case, however it seems it does not provide a great gameplay loop

M1tht1c:
> I thought that the "cannon" reason why dead was not counted for maroon was cause they could just be cloned. CC has the facilities. so zombified would be clonable at CC.

Clearly romerol does irreparable damage to the DNA - that would explain why you cant cure someone once they've been zombified too

PR #31970, opened 2024-09-08
add material composition to some salv treasure

deltanedas:
## About the PR
iron made of steel
silver and gold coins have a steel core and silver/gold coating because no zinc
diamond and adamantine coins have less than 1 diamond so recycler doesn't give you anything :trollface:

## Why / Balance
if you are low on specific mats you can choose to recycle valuable treasure, essentially buying them at a steep price.

**Changelog**
:cl:
- tweak: Coins and Supercharged CPUs can now be recycled for rarer materials.

PR #37149, opened 2025-05-03
Revert "add material composition to some salv treasure"

EmoGarbage404:
# About this PR
Reverts space-wizards/space-station-14#31970

The original PR misses the point of treasure items, which is to have sellables. Making people able to turn them into mats is an absolute noob trap that just fucks over salvagers. Most of these items give you an unbelievably steep value lost with little purpose, so I don't even see a reason in having it. Most of these materials are trivially obtainable on wrecks or through scrap anyways.

Additionally removes materials from rings for the same reason

:cl:
- remove: You can no longer scrap coins and other salvaging treasures. They're simply too valuable to junk.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Hands/base_clothinghands.yml
+++ b/Resources/Prototypes/Entities/Clothing/Hands/base_clothinghands.yml
@@ -69,11 +69,8 @@
   abstract: true
   id: GoldRingBase
   components:
-    - type: PhysicalComposition
-      materialComposition:
-        Gold: 200 # 2 bars
-    - type: StaticPrice
-      price: 300
+  - type: StaticPrice
+    price: 300
 
 - type: entity
   abstract: true
@@ -81,8 +78,5 @@
   name: silver ring
   description: Looks slightly less valuable than a gold one.
   components:
-    - type: PhysicalComposition
-      materialComposition:
-        Silver: 200 # 2 bars
-    - type: StaticPrice
-      price: 275
+  - type: StaticPrice
+    price: 275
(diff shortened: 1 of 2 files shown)


kosticia:
Idk but fact that that one coin costs like 1/5 of fueltank and takes really alot of space makes them not valid as treasure in my opinion

EmoGarbage404:
> Idk but fact that that one coin costs like 1/5 of fueltank and takes really alot of space makes them not valid as treasure in my opinion

they're literal 1x1 items in the inventory and have some of the highest values by volume. you wouldn't toss out a 2x2 1k value item so i have no idea why the same logic doesn't apply to something smaller.

kosticia:
> they're literal 1x1 items in the inventory and have some of the highest values by volume. you wouldn't toss out a 2x2 1k value item so i have no idea why the same logic doesn't apply to something smaller.

Sounds logical.

ruddygreat:
2 cents from someone that mostly plays on an mrp fork - why?
The valuable mats from the treasures (particularly the gold & silver from coins, though the rings as well) are imo much more useful to the station than the spesos they would've sold for unless cargo has had some comedically awful bounties.

RedBookcase:
I actually do like recycling some of these as Salvage as often times Cargo has plenty of cash and even getting small amounts of gold or silver feels more impactful for helping science when compared to an extra 200 or 300 spesos for Cargo. I wouldn't mind making rings drop some more gold or silver, or removing the material compositions from the steel coin (That one is absolutely a noob trap and sucks) and the diamond coin (that one is worth selling and the mats are awful), but I would miss the rest of these.

PJB3005:
> 2 cents from someone that mostly plays on an mrp fork - why?

The reason "why" is clearly stated in the PR description.

PR #37151, opened 2025-05-03
Fix skeletons spawning in folded body bags

youtissoum:
## About the PR

This PR fixes skeletons spawning in folded body bags by making CanInsert check if the container is folded.

## Why / Balance

If a skeleton spawns inside a folded body bag, it's impossible for them to escape without admin intervention. That is not good.

## Technical details

The CanInsert function now raises a new `EntityStorageInsertedIntoAttemptEvent` to the container.
This event is cancelled by the foldable system if the container is folded.

## Media

https://github.com/user-attachments/assets/21642b91-2e3f-4e5e-8bc6-55a6c73d16a5

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Skeletons will no longer spawn inside of folded body bags.

### DIFF
--- a/Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
+++ b/Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
@@ -150,6 +150,9 @@ public sealed class EntityStorageComponentState : ComponentState
 public record struct InsertIntoEntityStorageAttemptEvent(EntityUid ItemToInsert, bool Cancelled = false);
 
 [ByRefEvent]
+public record struct EntityStorageInsertedIntoAttemptEvent(EntityUid ItemToInsert, bool Cancelled = false);
+
+[ByRefEvent]
 public record struct StorageOpenAttemptEvent(EntityUid User, bool Silent, bool Cancelled = false);
 
 [ByRefEvent]
--- a/Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
+++ b/Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
@@ -146,9 +146,16 @@ public sealed class EntityStorageComponentState : ComponentState
     }
 }
 
+// Why does this even pass the entity being inserted if it's raised to the entity being inserted...
+/// <summary>
+/// Raised to the entity being inserted whenever checking if an entity can be inserted into a container.
+/// </summary>
 [ByRefEvent]
 public record struct InsertIntoEntityStorageAttemptEvent(EntityUid ItemToInsert, bool Cancelled = false);
 
+/// <summary>
+/// Raised to the container whenever checking if an entity can be inserted into it.
+/// </summary>
 [ByRefEvent]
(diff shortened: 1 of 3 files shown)


Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
```diff
@@ -149,6 +149,9 @@ public EntityStorageComponentState(bool open, int capacity, bool isCollidableWhe
 [ByRefEvent]
 public record struct InsertIntoEntityStorageAttemptEvent(EntityUid ItemToInsert, bool Cancelled = false);
 
+[ByRefEvent]
+public record struct EntityStorageInsertedIntoAttemptEvent(EntityUid ItemToInsert, bool Cancelled = false);
+
```
slarticodefast, line 154: add documentation to these two events so that people know what each of them does

Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
```diff
@@ -146,9 +146,19 @@ public EntityStorageComponentState(bool open, int capacity, bool isCollidableWhe
     }
 }
 
+// Why does this even pass the entity being inserted if it's raised to the entity being inserted...
```
slarticodefast, line 149:
```suggestion
```
probably in case the whitelist should be checked for another entity for some reason
or it sued to be a broadcast event

Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
```diff
@@ -146,9 +146,19 @@ public EntityStorageComponentState(bool open, int capacity, bool isCollidableWhe
...
+/// <summary>
+/// Raised to the entity being inserted whenever checking if an entity can be inserted into a container.
```
slarticodefast, line 151:
```suggestion
/// Raised on the entity being inserted whenever checking if an entity can be inserted into an entity storage.
```
an entity storage is a container, but not all containers are entity storages

Content.Shared/Storage/Components/SharedEntityStorageComponent.cs
```diff
@@ -146,9 +146,19 @@ public EntityStorageComponentState(bool open, int capacity, bool isCollidableWhe
...
+/// </summary>
 [ByRefEvent]
 public record struct InsertIntoEntityStorageAttemptEvent(EntityUid ItemToInsert, bool Cancelled = false);
 
+/// <summary>
+/// Raised to the container whenever checking if an entity can be inserted into it.
```
slarticodefast, line 157:
```suggestion
/// Raised on the entity storage whenever checking if an entity can be inserted into it.
```

slarticodefast (changes requested):

Tayrtahn (approved):

slarticodefast (approved):
Thanks for the bugfix!

Zarq0:
hilarious, bug of 2025 candidate in my heart fr

PR #34002, opened 2024-12-22
Changeling devour and transform 

poklj:
## About the PR

Adds Changelings Devour and Transform Componentry and Systems to Achieve what is set out in The [Changeling Doc ](https://docs.spacestation14.com/en/space-station-14/round-flow/proposals/changeling.html)

With the new components and Systems a Player is able to Consume Other humanoids for their Identities to later Assume their visage, speaking and all things that make up them.
## Why / Balance

Followed https://docs.spacestation14.com/en/space-station-14/round-flow/proposals/changeling.html directly.

Assumptions made not defined within the doc:
Devour can be used on a living being
Devour does 25 damage per second after windup (10/10/5 slash/pierce/blunt)

Double pressing the transform button will transform into the last Identity Consumed

Fingerprints are the Tell for changelings, While everything from their Gender, DNA, TypingBubble, voice descriptor are changed, their Fingerprints will always be the original Identities prints

## Technical details

Added ChangelingTransformComponent and System for controlling Transformation
Added ChangelingDevourComponent and System for managing the action of Consuming people
Added ChangelingIdentityComponent and System for Storing references to Consumed Identities and managing aspects that require additional functionality for devouring and Consuming (like spawning clone's of victims into nullspace as well as assisting with Pvs)

Identities when consumed are stored in nullspace via the HumanoidAppearanceComponent's Clone appearance function, and are paused. These entities will be Forced Into the PVS of the changeling so that later when the Transform Ui is utilized, these entities can be used to show the sprites of the identity.

During transform all information is transferred from these nullspace entities onto the acting entity.

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

Moved DNA into Shared
Moved HumanoidAppearanceComponent::CloneAppearance into Shared

## TODO:
- [x] Replace all Wehs with appropriate sound cues
  - [x] Devour Windup (2 seconds)
  - [x] Devour Attempt (10 seconds)
    - [x] Discuss the Slurp noise: keep it, Make it a wilhelm chance or an april fools revert
  - [x] Transform Attempt (5 seconds)
- [x] Discuss/get guidance on how Husking should behave
  - [x] Preliminary implementation on Husking
- [x] Discuss/get guidance on how Transform limitations should behave (if at all, can possibly be left for a future PR)
[...]

PR #37153, opened 2025-05-03
Make CloningSystem Shared and make Cloning components its own method

poklj:
## About the PR

Moved Cloning to shared and split out cloning of components for use in #34002 

## Why / Balance

Cloning is capable of being done in shared and the function of copying a list of components is required for #34002 

## Technical details

Moved Cloning to shared.

split out the copy comps into its own function for later use in other systems

## Media

Purely technical.

## Requirements

## Breaking changes

Moved CloningSystem to SharedCloningSystem

**Changelog**

No CL no Fun


poklj:
> While we do have predicted spawning, that doesn't work for any entity yet, so spawning still has to be done server side, or you will end up with a client side entity. Having the code in shared incentivised user error.

I think this approach would be best to make it funcional from sh

PR #37154, opened 2025-05-03
[HOTFIX] Movement Rewrite Hotfix Shuttles now respect their friction values

Pronana:
## About the PR
Fixes shuttles slowing down during FTL.

Shuttles now have the friction values they had before the update and use a modifier system to respect the new system.

Previously shuttles relied on tileFriction being 0 in the air and overwrote the 0.2 damping modifiers the physics engine gave every object. 

However since friction now uses damping, those yaml overrides were getting overwritten. So shuttles have had their friction overrides changed to a modifier based system which works with the new friction system.

Instead of shuttle's friction values being mistakenly set to 0.2 they are now correctly set as:

0.05 when in space
0 when in FTL 

## Why / Balance
Bugfixes shuttles slowing down and eventually stopping in FTL which is unintended behavior.

## Technical details
Removed Shuttles bulldozing damping since now everything uses damping and it's getting counter bulldozed
Make it so shuttles instead use damping modifers like everything else.
Adjusted the damping modifiers to match their old friction values
Made it so shuttles have 0 friction when in FTL.

## Requirements

## Breaking changes

Shuttle AngularDamping and LinearDamping values have been removed  
Shuttle BodyModifier and DampingModifier have been added

**Changelog**

### DIFF
--- a/Content.Server/Shuttles/Components/ShuttleComponent.cs
+++ b/Content.Server/Shuttles/Components/ShuttleComponent.cs
@@ -57,12 +57,12 @@ namespace Content.Server.Shuttles.Components
         public DirectionFlag ThrustDirections = DirectionFlag.None;
 
         /// <summary>
-        /// Damping applied to the shuttle's physics component when not in FTL.
+        /// Damping modifier applied to the shuttle's physics component when not in FTL.
         /// </summary>
-        [DataField("linearDamping"), ViewVariables(VVAccess.ReadWrite)]
-        public float LinearDamping = 0.05f;
+        [DataField]
+        public float BodyModifier = 0.25f;
 
-        [DataField("angularDamping"), ViewVariables(VVAccess.ReadWrite)]
-        public float AngularDamping = 0.05f;
+        [DataField]
+        public float DampingModifier;
     }
 }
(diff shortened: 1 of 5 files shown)


Content.Server/Shuttles/Components/ShuttleComponent.cs
```diff
@@ -57,12 +57,12 @@ public sealed partial class ShuttleComponent : Component
...
         /// </summary>
-        [DataField("linearDamping"), ViewVariables(VVAccess.ReadWrite)]
-        public float LinearDamping = 0.05f;
+        [DataField]
+        public float BodyModifier = 0.25f;
 
-        [DataField("angularDamping"), ViewVariables(VVAccess.ReadWrite)]
-        public float AngularDamping = 0.05f;
+        [DataField]
+        public float DampingModifier;
```
slarticodefast, line 66: please add documenation

Pronana: Updated it, tell me if it looks good.

Pronana:
Heisentest, I doubt those comments were load bearing.

Pronana:
Redirected onto staging cause I think it's important enough to go in before stable

Pronana:
Redirected to stable.

ScarKy0 (approved):

EmoGarbage404 (approved):

PR #29860, opened 2024-07-09
Remove Disabler Damage + Allow Pacifists to Use Disablers

Cojoke-dot:
## About the PR

Removes the heat damage from Disablers and allows Disablers to be used by pacifists.
## Why / Balance

Currently, Pacifists can use stun batons but not disablers, which does not make much sense to me. Permitting this would allow more self-defense options for pacifists, especially for heads of staff and pacifist security members. Parent the Disaber SMG projectile to the Disabler projectile because… parenting is cool 😎(changes to Disabler bolt should affect both)
## Technical details

magic of yml
## Media

https://github.com/space-wizards/space-station-14/assets/83733158/a1fdbe77-41e0-43e8-88e8-1679a18bdb26

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- tweak: Pacifists can now use disablers

PR #37164, opened 2025-05-04
Allows disabler, practice disabler, disabler SMG, and practice laser rifle to be used by pacifists

august-sun:
## About the PR
Allows pacifists to use the practice disabler, practice laser rifle, disabler, and disabler SMG.

## Why / Balance
As the disabler no longer deals anything but stamina damage, it stands to reason this is in-line with a pacifist's ability to use any other non-lethal tool. The practice laser rifle falls under this, as well.

## Technical details
Added PacifismAllowedGunComp to relevant YML lines.

## Requirements

## Breaking changes

**Changelog**

:cl:

- tweak: Pacifists can now use the practice disabler, practice laser rifle, disabler, and disabler SMG.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -228,6 +228,7 @@
     fireCost: 62.5
   - type: StaticPrice
     price: 300
+  - type: PacifismAllowedGun
 
 - type: entity
   name: laser rifle
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -472,6 +472,7 @@
     tags:
     - Taser
     - Sidearm
+  - type: PacifismAllowedGun
 
 - type: entity
   name: disabler
@@ -500,6 +501,7 @@
     guides:
     - Security
     - Antagonists
+  - type: PacifismAllowedGun
 
 - type: entity
   name: disabler SMG
@@ -538,6 +540,7 @@
   - type: Appearance
   - type: StaticPrice
     price: 260
+  - type: PacifismAllowedGun
 
 - type: entity
   name: taser


EmoGarbage404 (approved):

keronshb:
Can you also add one for throwing items as well? Would be goated if they can throw soap and bola.

ArtisticRoomba:
doesn't the practice laser rifle still do damage, just a little bit?

RedBookcase:
> doesn't the practice laser rifle still do damage, just a little bit?

Yep, looks like it deals a single point of heat damage, for whatever reason.

Cojoke-dot:
The forever loop of pacifist sec... I've fought for this for an eternity and the moment I start doing stuff again it comes back up 
#29860

slarticodefast:
Pinging @SlamBamActionman for sec balance. I know you are on a short break, so sorry for the ping, but I remember there being a long previous discussion about pacifist sec offs.

keronshb:
> Pinging @SlamBamActionman for sec balance. I know you are on a short break, so sorry for the ping, but I remember there being a long previous discussion about pacifist sec offs.

https://github.com/space-wizards/space-station-14/pull/29860

The tune changed for stun batons but that's the previous PR.

SlamBamActionman:
I am a bit pacifist-elitist and think that, from a gameplay perspective, a disabler is fairly equivalent to using a regular gun in terms of the effect; you may not kill the vent critter or whatever, but it gives you a ranged option that allows you to escape or apprehend. The only other ranged option you got is the stinger grenade, which is very limited in supply and hard to utilize. The stun baton is strong too, but melee opens you up to a lot more damage.

So the lack of range, purely from a gameplay perspective, makes pacifist SecOffs just plain better than everyone. :trollface: 

*That being said* I can recognize that I am a bit biased here. Were pacifists to get disabler access, I think at worst you'll make thieves a little harder to catch; anyone else is still playing a slightly nerfed normal role, just now slightly less so. 

If we're going on following pacifism "rules"; disablers don't deal damage, ergo they should not be blocked for pacifists.

keronshb:
> It's not really well communicated and kinda arbitrary that thieves/pacifists can use stun batons but not bolas or disablers

This is where it gets weird. If you just have pacifists not able to use anything that harms, then by proxy they should be able to use anything that not harms.

But then if you just give them everything in the usual sec toolbelt + bola, then what really is the difference between playing a pacifist and non pacifist other than being able to lethal (which, you still can with meatspike/incinerator/spacing/shuttlegibbing/plasma flooding)

Id rather it be blanket "they can use disabler/bola" than a weird in between.

SlamBamActionman:
> But then if you just give them everything in the usual sec toolbelt + bola, then what really is the difference between playing a pacifist and non pacifist other than being able to lethal (which, you still can with meatspike/incinerator/spacing/shuttlegibbing/plasma flooding)

I do think it's worth noting that being limited to not using lethals is a detriment with worthwhile gamplay impact:
- Slimes and spiders? Run or use stun batons, but you can never kill them. Need to pair up.
- Snakes? You're SOL.
- Dragon? You can at best flash carps, dragons is kind of a wipe though... So instead you can focus on splashing people with water and retrieving bodies.
- Zombies? Honestly really fun because you become full support, flashing zombies and building electrified grilles.
Not to mention what you do if you're a non-Sec and some random tider jumps you with a crowbar.

Bola always felt like it should be possible to throw as pacifist.

august-sun:
Doing the throw changes is beyond my capability, I'm pretty sure, but related to the question, do I need to remove the practice laser rifle from the PR, or do I need to remove the damage from practice lasers?

august-sun:
> would it be possible to make it so they can also fire the hos shotgun when its on disable mode?

Not unless you edited the component or system to include pacifism somehow specifically for that mode, which is beyond the scope of this PR.

keronshb:
> would it be possible to make it so they can also fire the hos shotgun when its on disable mode?

To be honest we're getting way too far into the weeds with this. 
I get you, it makes sense, but I want to avoid going to the extent of getting into minutia code of what a pacifist can and can't use, even if it means sacrificing something like the toggle.

superjj18:
Enables Pacifist sec officers which is nice and somewhat silly in a good way.

They are a bit of a wimp but just want everyone to be safe :) no nukie is immune to their infinite mercy and kindness!

PR #37168, opened 2025-05-04
Fix goliath tentacle exploit

EmoGarbage404:
## About the PR

Fixes goliath tentacles being unable to hit players if they stand still

## Why / Balance

obvious exploit lets people cheese goliaths with a minimum-effort strat.

## Technical details

this is what beams do and i have no idea why. im so sorry, god.

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Goliath tentacles no longer miss if you stand still.

### DIFF
--- a/Content.Shared/Stunnable/SharedStunSystem.cs
+++ b/Content.Shared/Stunnable/SharedStunSystem.cs
@@ -17,6 +17,7 @@ using Content.Shared.StatusEffect;
 using Content.Shared.Throwing;
 using Content.Shared.Whitelist;
 using Robust.Shared.Audio.Systems;
+using Robust.Shared.Physics;
 using Robust.Shared.Physics.Components;
 using Robust.Shared.Physics.Events;
 using Robust.Shared.Physics.Systems;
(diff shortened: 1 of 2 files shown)


metalgearsloth:
> this is what beams do

I'd rather take a look in engine as beam code is not exactly ideal.

slarticodefast:
Slightly off-topic, but why the hell do we have both `StunOnContactComponent` and `StunOnCollideComponent`?
They both seem to use `StartCollideEvent`.

CookieMasterT:
Huh, always thought this was intentional since it causes the most fun (not risk free) way to fight goliaths.
Where you have to:
- Run away from the goliath
- React with a slight window to the tentacle attack (stand still) or get stunned for 5 seconds and let them catch up
- And then move away to avoid its melee attack as the goliath is moving towards you

And it made some sense since it would be logical that it would be harder to pull you down, knocking you over, if you stood still (steady, on both legs)

Nukesthestation:
> And it made some sense since it would be logical that it would be harder to pull you down, knocking you over, if you stood still (steady, on both legs)

idea: why not make it either: one deal stamana damage if you're standing still (so that you can maybe only do it like 1-2 times), or two, give it a chance to stun?

SlimmSlamm:
I always interpreted it as the Goliath grabbing your leg but not actually tripping you unless you move. It’s a pretty short window to react to so I don’t really think it’s a big deal but I don’t care for it much either.

ArtisticRoomba (approved):
Everything looks good.

This PR has been given the seal of approval by the coalition of good salvage players.

PR #37170, opened 2025-05-04
Disarm Offhand on Shove

### LINKED ISSUE #37167: Cannot disarm left hand
## Description

When shoving a player, you are unable to shove the item they hold in their left hand

**Reproduction**

1. Put an item in a person's left hand
2. Shove repededly 
3. Item only leaves their hand after they fall

Cojoke-dot:
## About the PR

If the target of a shoves main hand is empty, disarm out of the targets other hands from right to left.
## Why / Balance

Too many times I have someone come up and grab an item out of my hand then keep it in their offhand so I can't get it back 😢 
It feels weird to not be able to shove something out of someones hand when there is no indication what hand is active on the target.
Closes #37167 
## Technical details

Lets you set the item to being disarmed in the DisarmedEvent, which gets sent about to ThrowHeldItem, where you can now override the item being thrown.
## Media

Screenshot 2025-05-04 004756

https://github.com/user-attachments/assets/aeaa7134-2079-4174-9ad6-35426b8cd30a

## Requirements

## Breaking changes

ThrowHeldItem(EntityUid, EntityCoordinates, float = 0.1) is now ThrowHeldItem(EntityUid, EntityCoordinates, EntityUid? = null, float = 0.1)

**Changelog**

:cl:
- tweak: You can now shove items out of peoples offhands.


LaCumbiaDelCoronavirus:
From first glance not totally sure on how you coded this, but it might be more ideal to specify the `Hand` that you're trying to disarm something from rather than specifying the entity you're actually trying to disarm out of someone's hands

Cojoke-dot:
> From first glance not totally sure on how you coded this, but it might be more ideal to specify the `Hand` that you're trying to disarm something from rather than specifying the entity you're actually trying to disarm out of someone's hands

Yeah I guess, but it checks if the item is in the entities hands before it attempts to throws it, which it did before this pr, so I think it would be more efficient to just define what is being thrown.

YoungThugSS14:
This is great but how did that urist have a 3rd hand..

Cojoke-dot:
> I'd prefer to swap this over to a shove system when we have the capability for it, since rng disarm is really stinky and having it effect the offhand too is gonna make it that much more overpowered.

What does a shove system mean?

keronshb:
> > I'd prefer to swap this over to a shove system when we have the capability for it, since rng disarm is really stinky and having it effect the offhand too is gonna make it that much more overpowered.
> 
> What does a shove system mean?

Somewhat similar to TGs but not entirely.

When someone is shoved, it should put them into an off-balance type state. Then, when they're shoved again and collide with something else two things should happen:
1) The off balance person should drop their item(s)
2) The off balance person should be knocked down
3) If the other entity was another player, that person should also be knocked down (but not lose their items)

Cojoke-dot:
> When someone is shoved, it should put them into an off-balance type state. Then, when they're shoved again and collide with something else two things should happen:
> 
> 1) The off balance person should drop their item(s)
> 
> 2) The off balance person should be knocked down
> 
> 3) If the other entity was another player, that person should also be knocked down (but not lose their items)

Ah so no rng, you just need to hit both shoves then a collision. I guess we would need knockdown first but that shouldn’t be toooooo long till that happens.

keronshb:
> > When someone is shoved, it should put them into an off-balance type state. Then, when they're shoved again and collide with something else two things should happen:
> > 
> > 1. The off balance person should drop their item(s)
> > 2. The off balance person should be knocked down
> > 3. If the other entity was another player, that person should also be knocked down (but not lose their items)
> 
> Ah so no rng, you just need to hit both shoves then a collision. I guess we would need knockdown first but that should be toooooo long till that happens.

Just some caveats too:
- Shover should lose stamina per shove
- The shoved should lose stamina when being knocked into something (but there should be a grace period so you can't spam it)

Cojoke-dot:
Well, this sounds cool, I don't see a reason to keep this pr open or update it then if the plan is to move to a system like ss13

keronshb:
No one's working on it right now, it definitely has some issues (not shoving people through walls, for example). So it's up for grabs.

PR #37172, opened 2025-05-04
DungeonData rework

metalgearsloth:
I moved everything back to inline fields because the data dictionary stuff doesn't realy work with my biome plans.

To facilitate this I have done a couple things:
- Every single EntitySpawnEntry uses the prototype version no exceptions.
- All data is inline and yaml lintable.
- The prototype config can specify which dungeons are passed through to that layer so we can keep stuff like VGRoidSmallPaths decoupled. ReservedTiles also now exclude any dungeons that are passed in to that layer. This will probably be subject to change but I do very much like being able to pass dungeons into prototype configs.
- IDC about the copy-paste between the 3 planet configs we can fix it later.
- I don't think I broke any configs but it may be possible.

## Breaking changes
- ReplaceTile + FillGrid dungeon layers no longer operate on every tile on the grid and only operate on the supplied dungeons. If you use prototype dungeon layers you can probably just specify the inheritDungeons datafield to specify All.
- Dungeon data is now inlined from the dictionary again (really sorry for this one). On the bright side the yaml linter should yell loudly.


EmoGarbage404 (changes requested):
should probably use EntityTables instead of EntitySpawnEntry / EntProtoId as that lets you opt to either inline or use a prototype rather than being stuck with one or the other

PR #37175, opened 2025-05-04
Energy swords welding

B-Kirill:
## About the PR
Adds a welding function to an energy sword and similar weapons

| Weapon       | Speed modifier |
|------------------|---------------|
| Welding tool         | 1  |
| Energy pen         | 0.8 | 
| Energy dagger         | 0.7 | 
| Energy sword         | 0.5  | 
| Energy cutlass         | 0.4 | 
| Double-bladed energy sword         | 0.4 | 

## Why / Balance
It is quite logical that a weapon with such a high temperature can weld metal. To prevent this from being unbalanced, the welding speed for each type of weapon has been changed. By the way, it still requires eye protection.

It even has a logical rationale:

- Energy dagger - small, low-power (compared to other types), ideal for welding, although it requires a little care.
- Energy sword - not very good for welding, the length of the weapon gets in the way a little, and you need to control the force so that you don't just burn through the metal.
- Energy cutlass - like an energy sword, however, it has an uneven blade, which also interferes with welding.

## Media

image

## Requirements

**Changelog**
:cl: B_Kirill
- add: Some melee energy weapons can now work like welding tool.


Cojoke-dot:
1. It's really powerful, infinite repair on nukie Borgs and the ability to seal off airlocks from sec while still being a powerful weapon
2. Logically have you ever tried to weld with a sword. I could agree that an epen may work but a full on sword?

B-Kirill:
> 1. It's really powerful, infinite repair on nukie Borgs and the ability to seal off airlocks from sec while still being a powerful weapon
> 2. Logically have you ever tried to weld with a sword. I could agree that an epen may work but a full on sword?

Well yeah... then I can remove the welding from the double sword altogether, and make the debuff even stronger for the regular one, so it will be better.

B-Kirill:
I have greatly reduced the welding speed of the sword and cutlass. I also removed the ability to weld with a double sword.
_Although maybe I should have given it the minimum time as a cutlass, too?_

TeenSarlacc:
is there a way to make it so you can only unweld doors with the swords, or not allow them to repair borgs?

pheenty:
As if the nukie borg wasn't ridiculously weak anyway...

keronshb:
This would be really busted with borgs that have an esword.

shamblestf:
I think this would make nukie borgs really oppressive, even if you remove the desword welding

I think keeping it on edagger might not be too bad, it would be interesting to see nukies buy edaggers to use exclusively as a welding tool

Entvari:
To all complaining about nukie borgs, you guys are aware nukies have welders right? 
Making E-Swords weld would make literally zero difference to nukie borgs as in this PR they're slower than standard welders, therefore you'd have no reason to use it over a base welder unless you're terrible at prepping.

RedBookcase:
Do Eswords/Edaggers have the ability to act as a cutting tool already? If so, are there any cases where giving energy weapons both cutting and welding functionality would cause issues?

Pronana:
They should at the very least get a big penalty to tool efficiency if this gets merged.

pheenty:
I wonder if people at least read the PR description

B-Kirill:
> Do Eswords/Edaggers have the ability to act as a cutting tool already? If so, are there any cases where giving energy weapons both cutting and welding functionality would cause issues?

Nope


> I wonder if people at least read the PR description

Nope

NoElkaTheGod:
I think the speed penalty is perhaps a bit too harsh? Like, there's literally no way anyone would ever use an esword as a welder if it's **5 times slower** than a normal welder (which you can find anywhere in like 30 seconds). 

Infinite fuel also doesn't matter that much, because:
1. As an antag, rarely do you have to weld so much that it'd require a refueling.
2. Even if you do have to refuel, finding a fuel tank is not **that** hard.

Given all that, i'd say that a 0.5 speed modifier for an esword would be pretty balanced. It's still two times slower than a normal welder, which is significant, but not too slow to ever be useful. Right now the only way i see myself using an esword as a welder is just to show off, which woon't even work that well because it would take ages to weld anything.

NoElkaTheGod:
Also yes, repairing assault borgs takes a lot of time even with a normal welder, so if nukies have enough time to repair them with an esword, they would also have enough time to find a fuel tank.

pheenty:
> I think the speed penalty is perhaps a bit too harsh? Like, there's literally no way anyone would ever use an esword as a welder if it's **5 times slower** than a normal welder (which you can find anywhere in like 30 seconds).
> 
> Infinite fuel also doesn't matter that much, because:
> 
> 1. As an antag, rarely do you have to weld so much that it'd require a refueling.
> 2. Even if you do have to refuel, finding a fuel tank is not **that** hard.
> 
> Given all that, i'd say that a 0.5 speed modifier for an esword would be pretty balanced. It's still two times slower than a normal welder, which is significant, but not too slow to ever be useful. Right now the only way i see myself using an esword as a welder is just to show off, which woon't even work that well because it would take ages to weld anything.

Extremely based

B-Kirill:
> I think the speed penalty is perhaps a bit too harsh? Like, there's literally no way anyone would ever use an esword as a welder if it's **5 times slower** than a normal welder (which you can find anywhere in like 30 seconds).

Hmm, that makes sense.

Samuka-C:
I feel like it would make more sense if energy blades could unweld stuff, but not weld

like, you can use to unweld a door but not weld it

walksanatora:
no I think a e-blade could melt the door down the middle to weld it shut. since welds even by a welder are not paticulary clean.

IProduceWidgets:
> To all complaining about nukie borgs, you guys are aware nukies have welders right? Making E-Swords weld would make literally zero difference to nukie borgs as in this PR they're slower than standard welders, therefore you'd have no reason to use it over a base welder unless you're terrible at prepping.

Its giving the borgs themselves a welder which is the buff. Now two borgs are self sufficient and all syndi borgs are mutual healers.

B-Kirill:
> Its giving the borgs themselves a welder which is the buff. Now two borgs are self sufficient and all syndi borgs are mutual healers.

Theoretically, yes, but in practice it is generally ineffective. It takes a VERY long time. Fixing borg even by welding tool takes much longer than, for example, welding doors. And with the current speed modifier, it's so slow that players will already have time to kill them 5 times while they stand around the corner and fix the borg.

superjj18:
>I think the speed penalty is perhaps a bit too harsh? Like, there's literally no way anyone would ever use an esword as a welder if it's 5 times slower than a normal welder (which you can find anywhere in like 30 seconds).

It's a matter of utility.  In perfect situations of course use a welder rather than the esword.  No welder?  Then a robust player would use the esword.

You would never break a bulb and then weld it to get a glass sheet when you could just ask cargo for one, unless you have circumstances where that isn't a possibility.

Ss13/14's best quality is that skilled players essentially are mini-mcgyvers.  Many problems have uniquely creative solutions where you get in a position of "you have a clear goal, and restrictive conditions, can you find a solution via creativity?"

NoElkaTheGod:
> It's a matter of utility. In perfect situations of course use a welder rather than the esword. No welder? Then a robust player would use the esword.
> 
> You would never break a bulb and then weld it to get a glass sheet when you could just ask cargo for one, unless you have circumstances where that isn't a possibility.
> 
> Ss13/14's best quality is that skilled players essentially are mini-mcgyvers. Many problems have uniquely creative solutions where you get in a position of "you have a clear goal, and restrictive conditions, can you find a solution via creativity?"

While that is true, using something with .2 welding speed would take so extremely long you'd probably die of old age before it happens. I think that .5 is a nice middle ground, where it's not so good you can basicaly use it in every scenario, but not too bad to ever be used.

Pixeltheaertist:
also #37182 exists, just a heads up so it might not be just for "admin stuff" if it gets merged

B-Kirill:
> also #37182 exists, just a heads up so it might not be just for "admin stuff" if it gets merged

Yeah, then I need to change that

slarticodefast:
Personally I would prefer to keep weapons and tools separate items. These really don't need any buffs, especially for borg ops.

B-Kirill:
> Personally I would prefer to keep weapons and tools separate items. These really don't need any buffs, especially for borg ops.

Considering the welding speed, this is not even a buff, but a joke. It's just for a variety of mechanics, it's really hard to call it something out of balance.

SlamBamActionman:
~~So we discussed this in a maintainer voicechat and there were mixed opinions, mostly the concern being
a) Infinite fuel for Nukies with assault borgs
b) Swords are pretty large and unwieldy; feels unintuitive that they would be used to weld things when they really are for cutting (which you can already do while hitting).~~

~~As such, we feel that it's probably best to limit it to only the energy dagger, which is otherwise a relatively underutilized item and smaller in size. It also makes it match the cybersun pen (which has screwdriver utility). We'll have to keep an eye on this for the future if it ends up being a meta choice that empowers infinite sustain assault borgs, but otherwise it has approval.~~

We discussed this a bit more (I was a bit fast typing my first response!) and came to the following conclusion:

Ideally we'd want to see a separation between "repairing" and "welding shut"; for the sake of consistency it'd probably be better if all energy melees have the "welding shut" functionality and "repairing" is exclusive to the welding torches. 

Additionally, is it necessary to have the WelderComponent?` It might be possible to achieve welding by using a ToolComponent, since that wouldn't require keeping track of fuel usage. It seems excessive to add 3 new properties to the component to achieve the desired behavior.

slarticodefast:
Yeah, `WeldingComponent` right now is only there for keeping track of fuel consumption, so the energy swords should not have that in the first place (rather than disabling it with three different datafields). And ideally the component would be refactored into a more general `ToolFuelConsumptionComponent`.

B-Kirill:
Changes:
- Switched completely to using `ToolComponent`, old problems have been fixed, apparently.
- Had to add `AutoNetworkedField` for `UseSound` in `ToolComponent` because `ComponentToggler` was breaking the sound for the client.
- Removed the welding option for the hypereutactic-blade as it is strange for a weapon of this size.

I am still thinking, maybe I should remove the welding option for borg ops?

KaiserMaus:
> Changes:
> 
> * Switched completely to using `ToolComponent`, old problems have been fixed, apparently.
> * Had to add `AutoNetworkedField` for `UseSound` in `ToolComponent` because `ComponentToggler` was breaking the sound for the client.
> * Removed the welding option for the hypereutactic-blade as it is strange for a weapon of this size.
> 
> I am still thinking, maybe I should remove the welding option for borg ops?

Leave welding on hyper just for meme

IProduceWidgets:
Just food for thought, is this better than a really slow "cut a big ass hole in the wall" interaction? Because those would be mutually exclusive.

KaiserMaus:
I saw a PR with a sword color selection in a radial menu. I would like to see something like dragons devour or RCD disassemble(because it have animation) and wielding in radial menu of ESwords.

B-Kirill:
The slicing sound for these weapons is currently broken because you cannot specify separate sounds for each quality field in `ToolComponent`.

PR #37177, opened 2025-05-04
Minor forensic evidence refactor + UI

BarryNorfolk:
## About the PR
Refactored the Forensic evidence a little to enable it to be more extensible with less effort for downstream developers.
Also, in my eyes, improved the UI a little by using collapsible headings and not showing evidence groups that have no entries.

## Why / Balance
I'd like to add some kind of forensics for rifling, and some for explosive residues, but the system for it is cumbersome and I dislike that residues means "Things that have cleaned something".

Instead we can rely on prototypes and allow people to add to it downstream instead, then we have less copy and pasting of the same `text.AppendLine(blah);` everywhere. The cleaning action of residues can also be moved out so that they are a separate category of thing not tied to the main evidence groups.

No change to balance.

## Technical details
The configuration of "cleanables" has been altered so that we can have more direct control over things. "Rifling" evidence shouldn't be cleanable by soap, but maybe SHOULD be cleanable by some other mechanism that I haven't thought of yet. Blacklisting controls that to a decent degree, though there I am also open to a suggestion for having a whitelist to cover unknowns that downstream devs might add.

DNAs special handling can be moved into a blacklist to ensure that gibbing doesn't allow for DNA to be removed.

## Media
All the evidence we can get:
image

Just one bit of evidence:
image

## Requirements

## Breaking changes
Forensics evidence has moved to a ProtoID based dictionary instead of a per-evidence List or Hashmap, this is likely to conflict with any downstream modifications that add new types of evidence or work with the forensics system.

Control over whether DNA is cleanable has moved to a ProtoID based blacklist, so you have more control over what can be removed from entities' evidence lists.

ResidueComponent has been removed and its attributes moved into the `CleansForensicsComponent`.

Finally, `CleansForensicsComponent` now also has a ProtoID based blacklist for controlling what kind of evidence they can clean.

**Changelog**
:cl:
- tweak: Minor tweak to Forensic UI, now with collapsible headers.
- tweak: Soaps and other things able to clean evidence now come under the heading "Cleaning Agents" instead of "Residues".
- tweak: Allow forensic evidence to be more easily extensible by downstreams.


themias:
Originally, I think the plan was to have more residues, but that never happened.
If you clean with a rag, that gets added as a fiber. I don't know if we want that to be put in the new column as well.

I do have a favor to ask. Can you update the residue to be consistent with the fiber naming convention of color first?
Slippery brown residue -> Brown slippery residue

BarryNorfolk:
aaahh I see, that sort of thing. Neat, I'll take a look and work it.

BarryNorfolk:
I want to rework this a little bit, don't like the `GetOrNew` everywhere.

BarryNorfolk:
Now it has less GetOrNew everywhere and gives a public interface for adding more generic evidence.

Content.Server/Destructible/Thresholds/Behaviors/SpawnEntitiesBehavior.cs
```diff
@@ -84,12 +89,12 @@ public void TransferForensics(EntityUid spawned, DestructibleSystem system, Enti
...
 
             var comp = system.EntityManager.EnsureComponent<ForensicsComponent>(spawned);
-            comp.DNAs = forensicsComponent.DNAs;
+            comp.Evidence[DNAEvidence] = forensicsComponent.Evidence[DNAEvidence];
 
             if (!system.Random.Prob(0.4f))
                 return;
-            comp.Fingerprints = forensicsComponent.Fingerprints;
-            comp.Fibers = forensicsComponent.Fibers;
+            comp.Evidence[FingerprintsEvidence] = forensicsComponent.Evidence[FingerprintsEvidence]; //TODO: These will rpboably break
```
ghost, line 96: if (forensicsComponent.Evidence.TryGetValue(FingerprintsEvidence), out var evidence)
comp.Evidence[FingerprintsEvidence] = evidence;

maybe?

Content.Server/Forensics/Systems/ForensicsSystem.cs
```diff
@@ -125,25 +160,7 @@ private void OnRehydrated(Entity<ForensicsComponent> ent, ref GotRehydratedEvent
...
-            }
-
-            foreach (var residue in src.Residues)
-            {
-                dest.Residues.Add(residue);
-            }
+            dest.Evidence = src.Evidence;
         }
 
         public List<string> GetSolutionsDNA(EntityUid uid)
```
ghost, line 167: I wander if as part of this refactor it would be good to move GetSolutionsDNA/GenerateFingerprints into its own small systems that would implement some GenerateEvidence and have everything connected through events.

So things like #40259 could be just done through adding small system and event subscribtions.

BarryNorfolk: Making it more generic and easy to add evidence is something definitely we can think about doing in the future. If we could get this in then I would be down for adding more to it if I can :)

Issue #35378, opened 2025-02-22
Hypereutactic Blade passive audio infinitely stacks.


Archisteks:
## Description
Turning the Hypereutactic Blade on causes a new passive audio to be generated, turning it off does not cause any existing passive audio to stop. It can stack infinitely and become unpleasantly loud, even if one is using it normally and not specifically attempting to make it extra loud.

**Reproduction**
Get Hypereutactic Blade. Activate it, it makes noise. Deactivate it, it still makes noise. Reactivate it, it makes even more noise. Repeat ad infinitum.

**Screenshots**

**Additional context**
Round #78293, vulture.

Pumkin69:
This is an issue with eswords as well i think the way you recreate it is

turn on esword
have someone pick up turned on esword
they turn it off
audio of humming still playing?

something like that, its really fuckin annoying.

themias:
Same issue as #37369. This one was first, but that one has more info, so I'll close this in favor of it.

PR #37182, opened 2025-05-04
Readds the Hypereutactic Blade for traitors, adds Hypereutatic blade for Nukies

keronshb:
## About the PR

This adds the Hypereutactic Blade back into the Traitor Uplinks for 18 TC with a very rare discount down to 15 TC.
Also adds a new item, the Hypereutatic Blade to the Nukie uplink at 16 tc, with a very rare discount down to 13 TC.

## Why / Balance

It's been added as an admin only item for awhile with differentiating results. I've heard on one hand it's very good, but I've also heard mostly on the other hand it's just ok. So I'm going off of the "it's just ok" crowd with some further input from Imp's Downstream.

On Imp the reflect chance was lowered to 75% (matches ESword here) and the overall sentiment is the same. It looks cool, the weapon is fun, but it's nowhere near optimal or extremely good (like how DSword was). So I added this to the Hypereutatic Blade, since it makes sense.

## Technical details

The Hypereutactic Blade was added back into the Traitor uplink at 18 TC.
It has a very rare discount for 15TC
It's banned from the Surplus Crates.

The Hypereutatic blade was added to the Nukie uplink at 16 TC, 13 discount. Also banned from Crates. Lower reflect chance than the Hypereutactic blade too (75%)

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added the Hypereutactic Blade back to the Traitor Uplink
- add: Added the Hypereutatic Blade to the Nukie uplink

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
@@ -298,7 +298,7 @@
   #  activeSound:
   #    path: /Audio/Weapons/ebladehum.ogg
   #    params:
-    #    volume: 3
+  #      volume: 3
   - type: ComponentToggler
     components:
     - type: Sharp


slarticodefast (approved):
I guess we can give it a try and see how it goes.
Another alternative would be to make it surplus bundle only (at the same cost or higher) so that is a rare, lucky item.
Code looks good, I'll leave the PR open for a while before merging so people have a chance to give feedback.

ScarKy0:
Nukies should DEFINITELY have the cost be like 25 to 30TC
From what I ran and tested nukies SHREDDED stations with just a single one of those

keronshb:
> Nukies should DEFINITELY have the cost be like 25 to 30TC From what I ran and tested nukies SHREDDED stations with just a single one of those

I've seen and heard the opposite on Imp (though to note Imp's has a nerfed reflect chance).

ScarKy0:
> I've seen and heard the opposite on Imp (though to note Imp's has a nerfed reflect chance).

It's either that or nerf the reflect chance as well then, cause this is way too busted to give to nukies. Normal traitors seem fine though

Archisteks:
I've used it as a nukeop commander on 100% reflect chance and it was, indeed, busted. Definitely lower reflect to 75% if we're making it available in uplinks.

Pumkin69:
have we fixed the bug where it just keeps stacking the esword humming audio on itself til its literally unbearable.

slarticodefast:
> have we fixed the bug where it just keeps stacking the esword humming audio on itself til its literally unbearable.

If we don't have one yet, could you create an issue for this including reproduction steps?

RedBookcase:
Do the inhands still clip through the UI

image

RedBookcase:
> > have we fixed the bug where it just keeps stacking the esword humming audio on itself til its literally unbearable.
> 
> If we don't have one yet, could you create an issue for this including reproduction steps?

https://github.com/space-wizards/space-station-14/issues/35378

keronshb:
I do think the inhand visuals and sound is out of scope. The sound stacking was not introduced by the hypereu blade itself.
Inhand visuals doesn't have support but it could be the joke that it's so massive it extends and cuts itself off.

RedBookcase:
> I do think the inhand visuals and sound is out of scope. The sound stacking was not introduced by the hypereu blade itself.

The Hypereu is the only item in the game that seems to suffer from it since the regular DEsword has been removed so i'd say it's important to fix it if it's gonna be added. And it's especially obnoxious given that its loud enough to still be audible while offscreen.

keronshb:
> > I do think the inhand visuals and sound is out of scope. The sound stacking was not introduced by the hypereu blade itself.
> 
> The Hypereu is the only item in the game that seems to suffer from it since the regular DEsword has been removed so i'd say it's important to fix it if it's gonna be added. And it's especially obnoxious given that its loud enough to still be audible while offscreen.

Going by the issue it also effects the regular esword.

RedBookcase:
> Going by the issue it also effects the regular esword. So while I do think it should be fixed, I don't think it's in scope for this PR.

It does not effect the regular esword, I tested the method mentioned there and it did not cause any problems with the esword. 

Simply turning the Hyper on and off causes the sfx to stack permanently. There's no need for any weird exploits or special methods to make the audio bug out - just turning it on and off does it. It's definitely not ready for play in this state.

Pronana:
> It does not effect the regular esword, I tested the method mentioned there and it did not cause any problems with the esword.
> 
> Simply turning the Hyper on and off causes the sfx to stack permanently. There's no need for any weird exploits or special methods to make the audio bug out - just turning it on and off does it. It's definitely not ready for play in this state.

It's wielding issue irrc. Because it's wielded it starts the sound twice for each hand but only stops it once.

keronshb:
I just do not think it is in scope for me readding the Hypereutactic blade to fix this bug.
Yes it should be fixed.
The blade itself did not cause it.
This PR is just about readding it. The existence of the blade itself did not cause this bug.
As a maintainer, I'm saying it needs to be _atomized_ and is out of scope.

slarticodefast:
I think the audio bug will have to be fixed before this can be merged, even if the sword is not directly at fault. It is too game-breaking.

keronshb:
> I think the audio bug will have to be fixed before this can be merged, even if the sword is not directly at fault. It is too game-breaking. I agree with the atomization, it can be done in a separate PR, but will still have to be done first. The visual bug is non-blocking on the other hand.

What if I just disable the on-sound for the dsword/hyper? It's not that immersion breaking if it's not in.

slarticodefast:
I guess that could work as a bandaid fix.

keronshb:
> I guess that could work as a bandaid fix.

Done :trollface:

Pixeltheaertist:
now it's a stealth weapon :trollface: yeehaw

slarticodefast:
Since Scarky had some balance concerns for the TC price I'll open up a maintainer discussion.

keronshb:
> My solution to the audio bug was making it multi handed and changing the sprite held prefix to be the wielded sprite
> 
> The only problem with that is you cannot one-hand it anymore when it is deactivated though

This probably works better since it says 2 handed anyway

keronshb:
> > My solution to the audio bug was making it multi handed and changing the sprite held prefix to be the wielded sprite
> > The only problem with that is you cannot one-hand it anymore when it is deactivated though
> 
> This probably works better since it says 2 handed anyway

This actually makes a different bug that forces you to drop it.

RedBookcase:
Fix for the sound stacking inbound #38070

Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
```diff
@@ -293,11 +293,12 @@
             Slash: 12
             Heat: 12
             Structural: 15
+  # Disabled until the wield active sfx no longer stacks
   - type: ItemToggleActiveSound
     activeSound:
       path: /Audio/Weapons/ebladehum.ogg
       params:
```
slarticodefast:
```suggestion
  #- type: ItemToggleActiveSound
  #  activeSound:
  #    path: /Audio/Weapons/ebladehum.ogg
  #    params:
```
I would rather comment it out than reducing the volume

Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
```diff
@@ -293,11 +293,12 @@
...
-        volume: 3
+        volume: -9999999
```
slarticodefast:
```suggestion
    #    volume: 3
```

Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
```diff
@@ -383,6 +384,17 @@
     reflectProb: 1.0
     spread: 75
 
+# Nukie variant, reduced reflection rate.
+- type: entity
+  parent: HyperEutacticBlade
+  id: HyperEutaticBlade
+  name: hypereutatic-blade
+  description: Often mistaken for the HyperEutactic Blade. This mass produced, off-brand, knockoff gets the same job done but with less reflection.
```
slarticodefast:
```suggestion
  description: Often mistaken for the Hypereutactic Blade. This mass produced, off-brand, knockoff gets the same job done but with less reflection.
```

Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
```diff
@@ -293,11 +293,12 @@
...
-    activeSound:
-      path: /Audio/Weapons/ebladehum.ogg
-      params:
-        volume: 3
+  # Disabled until the wield active sfx no longer stacks
+  #- type: ItemToggleActiveSound
+  #  activeSound:
+  #    path: /Audio/Weapons/ebladehum.ogg
+  #    params:
+    #    volume: 3
```
slarticodefast:
```suggestion
  #      volume: 3
```

slarticodefast (approved):
This got approval in the maintainer discussion. Giving the 100% reflection version to nukies was seen as too powerful in the playtests done by the admins, but the reduced reflection variant might address this. The exact balance and price will still have to be observed in the future.

PR #37190, opened 2025-05-05
Flatpacker resprited

vectorassembly:
## About the PR
Flatpacker resprite.

## Why / Balance
Old one didn't fit with other tech.

## Technical details
None.

## Media
image

## Requirements

- [ x ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ x ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
None

**Changelog**

:cl:
- tweak: Changed flatpacker sprite.


RedBookcase:
Looks a lot like a condenser with that big glass tube

vectorassembly:
> Looks a lot like a condenser with that big glass tube

original had it, so I decided to leave it.

RedBookcase:
> > Looks a lot like a condenser with that big glass tube
> 
> original had it, so I decided to leave it.

Yea, I think it's just a result of this one having visual depth compared to the old one. Actually, do you think the old sprite was intentionally flat bc it was for the Flatpacker? Sorta like the retro laser being the only gun that faces left instead of right? Nevermind, i'm rambling.

vectorassembly:
> > > Looks a lot like a condenser with that big glass tube
> 
> > 
> 
> > original had it, so I decided to leave it.
> 
> 
> 
> Yea, I think it's just a result of this one having visual depth compared to the old one. Actually, do you think the old sprite was intentionally flat bc it was for the Flatpacker? Sorta like the retro laser being the only gun that faces left instead of right? Nevermind, i'm rambling.

Maybe, yeah.

Warpzoned:
Awesome stuff! A tad weird that it's lopsided though, was there an intended design choice for both halves to be at different levels? The top of the flatpacker's right part is a bit too barren for how much area it covers as well, maybe a panel could be put in as filler but you decide, as otherwise it's pretty much perfect in my eyes, great work! ^^)b

Spessmann:
I think the keyboard should glow green like the other lathes

vectorassembly:
> Awesome stuff! A tad weird that it's lopsided though, was there an intended design choice for both halves to be at different levels? The top of the flatpacker's right part is a bit too barren for how much area it covers as well, maybe a panel could be put in as filler but you decide, as otherwise it's pretty much perfect in my eyes, great work! ^^)b

oh yeah, I wanted it to look modular. Thanks for feedback.

vectorassembly:
> I think the keyboard should glow green like the other lathes

thought of that, I didn't like it, so I made it a regular keyboard, instead of some glowing button rack.

EmoGarbage404 (changes requested):
i'm admittedly a bit biased because i really do like the old flatpacker sprite, but i'd be remiss to not point out some issues with this new iteration.

image

The layering between the two halves of the packer is really weird? it looks like one is almost above the other but it isn't really clearly establish what the connection between them is.

Additionally, the right-hand side of the machine seems to almost be floating. If it's meant to be set back in space, then it needs to be positioned underneath the other side of the machine. As it stands, it seems a bit spatially confused.

The warning stripe also seems strangely placed. It doesn't really wrap around the whole machine and kinda is just a single ribbon placed on the top. It might look better around the bottom like how the hyperlathes do it.

As a comment on the thematic side of things, if we're moving away from the funny space magic glass window, i'd really like the machine to be more representative of something being "packed." The tall glass tube really reads too similar to the condenser and i think it makes it a little bit too generic for such a specialized machine.

vectorassembly:
thanks for feedback, you can tweak it however you want, cc after all, i'm just busy with other sprites and some personal art projects.

vectorassembly:
> i'm admittedly a bit biased because i really do like the old flatpacker sprite, but i'd be remiss to not point out some issues with this new iteration.
> 
> 
> 
> image
> 
> 
> 
> The layering between the two halves of the packer is really weird? it looks like one is almost above the other but it isn't really clearly establish what the connection between them is.
> 
> 
> 
> Additionally, the right-hand side of the machine seems to almost be floating. If it's meant to be set back in space, then it needs to be positioned underneath the other side of the machine. As it stands, it seems a bit spatially confused.
> 
> 
> 
> The warning stripe also seems strangely placed. It doesn't really wrap around the whole machine and kinda is just a single ribbon placed on the top. It might look better around the bottom like how the hyperlathes do it.
> 
> 
> 
> As a comment on the thematic side of things, if we're moving away from the funny space magic glass window, i'd really like the machine to be more representative of something being "packed." The tall glass tube really reads too similar to the condenser and i think it makes it a little bit too generic for such a specialized machine.

also, original sprites is more "spatially confusing", especially at hazard stripes and the top part.

EmoGarbage404:
> thanks for feedback, you can tweak it however you want, cc after all, i'm just busy with other sprites and some personal art projects, so not really motivated to think of that rn.

that's fine. Look forward to seeing what you do in the future.

Issue #37035, opened 2025-04-29
Clowns don't waddle (MobAnimatorComponent Mega-Issue)


FairlySadPanda:
## Description
Clowns don't waddle. Clowns are supposed to waddle. My immersion is ruined.

## Additional Info

We can't add stuff like clown waddling in easily for two reasons
1. The movement code is janky and needs resolving. Both @metalgearsloth and @Pronana have been working tirelessly on this front.
2. Mob animation code is currently smeared all over the content codebase and is hard to work within without breaking something, especially with replication/prediction/PVS making things complicated.

A number of legacy bugs and current issues have their root cause, or are affected by, the lack of an overarching mob animation state machine.

Examples of this include:
1. Clown waddling being introduced, hotfixed repeatedly, then removed
2. Buckling visual bugs
3. Spessmen being sideways after traumatic events for no obvious reason
4. Many animations are currently coded in different places, such as footstep SFX timings, floating visuals, becoming bloodied after taking damage, and so on.

This is already out of control and getting worse as the game ages. If we're adding crawling - [see here for the issue
](https://github.com/space-wizards/space-station-14/issues/34894) - and doing mob movement refactors etc, we really need to get a handle on this before this is completely unfixable.

# TODOs
- [ ] Create a reference document of all the random animation states that mobs get into, including things not traditionally thought of as animation like swapping sprites for bloodied state, footstep timing SFX, and so on
- [ ] Research how Unity, Unreal, Godot, GMS2 and other common game engines write animators.
- [ ] Confirm there's no good off-the-shelf C# solution (or port-able solution) we can borrow/steal.
- [ ] If there is a good solution we can borrow/steal, borrow/steal that solution and make it suitable for RT ECS and replication/prediction.
- [ ] If there is not a good solution we can borrow/steal, write a solution that is suitable for RT ECS and replication/prediction.
- [ ] Agree with RT maintainers if the core state machine code should be engine-level or SS14-facing.
- [ ] Agree the cut-off point between YAML and C# for defining an animation (avoid "coding in YAML" to prevent @PJB3005 having to pull the brakes)
- [ ] Write MobAnimatorComponent and MobAnimatorSystem, and the rest of the owl (including events, managers, prototypes)
- [ ] Port each current animation over to MobAnimatorComponent.
- [ ] Re-implement clown waddling

FairlySadPanda:
## Unity Animations

* An **animator** is a thing that animates an entity. It has a animation controller.
[...]

PR #37196, opened 2025-05-05
Stun and Stamina Visuals

Princess-Cheeseballs:
## About the PR
Added animations for being low on stamina and getting stunned because I was manipulated/soaped into doing it.

Requires [#5909](https://github.com/space-wizards/RobustToolbox/pull/5909)

Requires  #37199

## Why / Balance
A visual indicator was desired and is also required before crawling can be merged.

## Technical details
StaminaSystem is now SharedStaminaSystem so that I can do animations on client
Added Animation Generators to the Client Stun System which are public for anyone to use
Added an event you can raise to make StunSystem do their funny animations
Added an event the StunSystem raises when it's animation is over (to prevent animation overlap)

## Media

https://github.com/user-attachments/assets/12bf7e90-8fc7-4a17-afeb-8a5e42f20247

https://github.com/user-attachments/assets/bdb059b8-1c6b-4e2b-9424-af8946f0f6f6

## Requirements

## Breaking changes

StaminaSystem renamed to SharedStaminaSystem and now has client and server StaminaSystems

**Changelog**

:cl:
- add: Added Stamina and Stun animations


OnsenCapy:
It doesn't really fit the pixel art animation from SS14. Looks out of place.

YoungThugSS14:
> It doesn't really fit the pixel art animation from SS14. Looks out of place.

This type of shakeyness is in 13, why shouldn’t it be in 14

Pronana:
Test fails are because Engine change needs to be merged

EthanQix:
> > It doesn't really fit the pixel art animation from SS14. Looks out of place.
> 
> This type of shakeyness is in 13, why shouldn’t it be in 14

Wrong question to ask, this isn't SS13 it's SS14. The correct question is "why should this be in SS14".

EmoGarbage404:
This type of animation looks fine. I don't know why people are pretending it isn't already in the game, because we already have electrocution, jittering, etc. and those look fine.

ScarKy0:
> This looks really janky, especially when you're in stamcrit.
Instead of an animation I'd highly prefer an overlay of sorts on the sprite to show they're winded. Like the sprite with an overlay that makes them look washed out or something.

We currently use jittering a lot, and I feel like jittering on the ground from pain (as you would after getting hit by a taser or stun batons) feels very fitting as it is. Using an overlay would make no sense as no other place uses a behaviour like this, on top of being something very hard to make fitting.

ilovehans10:
How much does the jittering affect their hitbox? It looks like the hitbox moves with the sprite, and I know that in the past when something similar was implemented for clown shoes it was undesired and removed. To my understanding, this was because the clown shoes animation would act strangely for players with high ping, but I'm unsure if that issue would exist here.

keronshb:
> > This looks really janky, especially when you're in stamcrit.
> > Instead of an animation I'd highly prefer an overlay of sorts on the sprite to show they're winded. Like the sprite with an overlay that makes them look washed out or something.
> 
> We currently use jittering a lot, and I feel like jittering on the ground from pain (as you would after getting hit by a taser or stun batons) feels very fitting as it is. Using an overlay would make no sense as no other place uses a behaviour like this, on top of being something very hard to make fitting.

The animation could be smoother if we're going that route, it just looks extremely jank in this case.

Pronana:
> Why the SSD indicator gotta jitter too

SSD indicator position is auto calculated unfortunately so it jitters with animation.

Pronana:
> How much does the jittering affect their hitbox? It looks like the hitbox moves with the sprite, and I know that in the past when something similar was implemented for clown shoes it was undesired and removed. To my understanding, this was because the clown shoes animation would act strangely for players with high ping, but I'm unsure if that issue would exist here.

The physical hitbox is unchanged, but yes the sprite does move which affects clicking

Pronana:
> Also consider before stamcrit, if a player is knocked down, they should be able to click on themselves to shove up. Having such a jittery animation makes this a huge problem.

The animation is generated using different animation variables fed into a keyframe generator. They can be adjusted easily to your liking through datafields.

Tayrtahn:
I like the concept, but I'm not sure about the rotation. I'm wondering if it would look better with just the jittering and no rotation.

HerCoyote23:
I think this looks good specifically for batons, but wouldn't make sense in a lot of other cases where you can get stamcrit. For example, boxing/rigged gloves or truncheon can stamcrit you, and in that case I don't see why you would be jittering like you've just been electrocuted. 
I'd say it would be better if you just gave a brief jitter status effect to the baton and disabler bolts, as they're energy weapons so it visually makes a bit more sense.

ScarKy0:
> For example, boxing/rigged gloves or truncheon can stamcrit you, and in that case I don't see why you would be jittering like you've just been electrocuted.

If I get beat up to the point I can't stand up I totally expect to be jittering from pain.
Also making it just on stun batons/disabler hits kinda feels like unneccessary soap.

Pronana:
> This looks really janky, especially when you're in stamcrit.
> 
> Instead of an animation I'd highly prefer an overlay of sorts on the sprite to show they're winded. Like the sprite with an overlay that makes them look washed out or something.

Define overlay, do you mean like a color tint or a proper texture/sprite?

Pronana:
> I think this looks good specifically for batons, but wouldn't make sense in a lot of other cases where you can get stamcrit. For example, boxing/rigged gloves or truncheon can stamcrit you, and in that case I don't see why you would be jittering like you've just been electrocuted. I'd say it would be better if you just gave a brief jitter status effect to the baton and disabler bolts, as they're energy weapons so it visually makes a bit more sense.

You actually could do this if you wanted, the animation methods have inputs for rotation and frequency, you can also turn off rotation and only have the jittering.

image
image

All you need to do is feed the values in that you like and get the animation you want. You'd have to remove the generic listener for stun animation from the stun system and start and manage events per stun source but you could do it.

FairlySadPanda (changes requested):
This PR should be broken out into two PRs, one to do the renaming to SharedStaminaSystem and one to implement the animations.

Pronana:
> I like the concept, but I'm not sure about the rotation. I'm wondering if it would look better with just the jittering and no rotation.

Here's the animation with reduced jitter and no rotation
https://github.com/user-attachments/assets/dcc0c384-096f-4b29-8857-3fc088dc467e

Here's the animation with no rotation but same jitter
https://github.com/user-attachments/assets/b75e5526-3bf6-425d-bf59-6ad2caaae0c3

Here's the animation with rotation but reduced jitter
https://github.com/user-attachments/assets/6e6812c9-be34-4654-aa35-3156f2fd09f2

Pronana:
> This PR should be broken out into two PRs, one to do the renaming to SharedStaminaSystem and one to implement the animations.

That's probably a good idea

> As with other mob animation things I'd strongly encourage folks to move to a unified mob animation system rather than continuing to bolt-on random animation actions onto code. #37035

I had to dig into the animation system myself and I'm gonna be honest, this is more of an engine level issue than it is a SS14 issue. The animation system doesn't have proper support for adjustment layers or delta keyframes so if two animations modify the same field (aka two animations or even just two keyframe layers affect the position of a sprite, very common) they will always bulldoze each other and it will look like shit (trust me I tried).

But yeah I agree we do need a proper animation controller for stuff like this, but it's gonna require big changes under the hood first. Otherwise, for waddling and stun we'd need a waddling animation, a stun animation, and a waddling+stun animation growing parabolicly for each new animation we add, and that's not feasible.

ps3moira:
The animation could be a little closer to SS13 and be more subtle

Pronana:
> Having no rotation looks a lot better in both instances. I know Emo said it looks fine, but I feel like rotation isn't really very fitting for how space station characters normally move - I feel like their paper-doll-ness only makes them really lend themselves to simple linear-feeling animations. With the subtle rotation here, it makes it feel like the character is being unnaturally contorted around in a photoshop canvas for a meme, or something.

That's fair. However I added the rotation because we needed a way to distinguish specfically stamcrit from knockdown and the rotation is visually distinct.

Any alternatives you suggest?

Flareguy:
> > Having no rotation looks a lot better in both instances. I know Emo said it looks fine, but I feel like rotation isn't really very fitting for how space station characters normally move - I feel like their paper-doll-ness only makes them really lend themselves to simple linear-feeling animations. With the subtle rotation here, it makes it feel like the character is being unnaturally contorted around in a photoshop canvas for a meme, or something.
> 
> That's fair. However I added the rotation because we needed a way to distinguish specfically stamcrit from knockdown and the rotation is visually distinct.
> 
> Any alternatives you suggest?

Jittering is already probably good enough, but if you wanted a more obvious indicator, I feel like you could probably have a status-effect-looking thing above the character. Like dizzy stars or whatever.

slarticodefast (commented):
Not sure if we need to have all of the animation parameters as datafields in both components. Would they need to be adjusted per entity in some cases?
If you keep them autonetworked then use field deltas though, or dirtying will become too expensive.

Errant-4 (commented):
I do not find this visual approach to be intuitive. I think we should consider some effect that is more immediately recognizable. Think about how this is always indicated in still images while being immediately readable
image

Princess-Cheeseballs:
> Not sure if we need to have all of the animation parameters as datafields in both components. Would they need to be adjusted per entity in some cases?
> 
> If you keep them autonetworked then use field deltas though, or dirtying will become too expensive.

They don't need to be autonetworked, it just helped a lot when tuning the animation since it let me edit the datafields without having to restart my game.

I should also probably draft this PR while I work on some of the alternative stun animations suggested.

Princess-Cheeseballs:
https://github.com/user-attachments/assets/1534c7c7-8e72-4cd8-810c-60193f5e1194

Sorry for the wait, got stuck wrangling the appearance system and such, and also wasn't happy with the first few iterations of the stars (Don't exactly have a lot of pixels to work with). 

I'd love to get some feedback, personally I'm not sure if I even like the swirling stars now that its done.

keronshb:
> Base.Profile.2025.06.16.-.21.05.44.16.1.mp4
> 
> Sorry for the wait, got stuck wrangling the appearance system and such, and also wasn't happy with the first few iterations of the stars (Don't exactly have a lot of pixels to work with).
> 
> I'd love to get some feedback, personally I'm not sure if I even like the swirling stars now that its done.

i think it might be too much jitter but i wanna see what others think.
the stars is hilarious, looks good to have

K-Dynamic:
Please no rotational jitter, just regular linear jitter is fine

ArcaneOcto:
Stars look fantastic! I agree that the rotational jitter is meh

keronshb:
Try it with just horizontal jitter and I think we're good.

Princess-Cheeseballs:
> Try it with just horizontal jitter and I think we're good.

https://github.com/user-attachments/assets/00d3ca2e-3eef-4e50-9c47-7f82991eb4e4

Princess-Cheeseballs:
If the animation looks good I can do some final cleanup (cause the code is a bit of a mess from going through multiple iterations of the animation) and undraft the PR.

keronshb:
> > Try it with just horizontal jitter and I think we're good.
> 
> Base.Profile.2025.06.18.-.21.40.08.18.mp4

I def think that looks way better. Has my stamp of approval.

ScarKy0:
The jitter speed and stuff might need some fine tuning but thats for a future PR. It looks great!

Princess-Cheeseballs:
Cleaned up the code, removed unneeded variables, added some comments should be ready for review.

keronshb (approved):
I'm approving for visuals. The code itself is fine.

Content.Client/Damage/Systems/StaminaSystem.cs
```diff
@@ -1,7 +1,125 @@
...
+        // Don't animate if we're dead
+        if (TryComp<MobStateComponent>(entity, out var state) && state.CurrentState == MobState.Dead)
+            return;
+
+        entity.Comp.StartOffset = sprite.Offset;
+
+        PlayAnimation(entity, sprite);
+    }
+
+    private void StopAnimation(Entity<StaminaComponent> entity, SpriteComponent? sprite = null)
```
slarticodefast:
```suggestion
    private void StopAnimation(Entity<StaminaComponent?, SpriteComponent?> ent)
```
use `Entity<T>` pattern

Content.Client/Damage/Systems/StaminaSystem.cs
```diff
@@ -1,7 +1,125 @@
...
+            return;
+        }
+
+        if (!HasComp<AnimationPlayerComponent>(entity))
+            return;
+
+        PlayAnimation(entity, sprite);
+    }
+
+    private void PlayAnimation(Entity<StaminaComponent> entity, SpriteComponent sprite)
```
slarticodefast: same

Content.Client/Stunnable/StunSystem.cs
```diff
@@ -1,9 +1,227 @@
...
+            UpdateAppearance((entity, sprite, entity.Comp));
+        }
+
+        private void OnAppearanceChanged(Entity<StunnedComponent> entity, ref AppearanceChangeEvent args)
+        {
+            if (args.Sprite != null)
+                UpdateAppearance((entity, args.Sprite, entity.Comp));
+        }
+
+        private void UpdateAppearance(Entity<SpriteComponent?, StunnedComponent> entity)
```
slarticodefast:
```suggestion
        private void UpdateAppearance(Entity<StunnedComponent, SpriteComponent?> entity)
```
keep it consistent with the other methods

Content.Client/Stunnable/StunSystem.cs
```diff
@@ -1,9 +1,227 @@
...
+using Content.Shared.Mobs;
 using Content.Shared.Stunnable;
+using Robust.Client.Animations;
+using Robust.Client.GameObjects;
+using Robust.Shared.Animations;
+using Robust.Shared.Random;
+using Robust.Shared.Timing;
+using Robust.Shared.Utility;
 
 namespace Content.Client.Stunnable
```
slarticodefast: use file scoped namespace

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -75,4 +76,79 @@ public sealed partial class StaminaComponent : Component
...
+    public float FrequencyMod = 1.75f;
+
+    /// <summary>
+    /// Jitter keyframes per animation
+    /// </summary>
+    [DataField]
+    public int Jitters = 4;
+
+    [AutoNetworkedField, DataField]
+    public Vector2 LastJitter;
```
slarticodefast, line 147: add documentation
also this seems to be purely client side, so why network it?

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -75,4 +76,79 @@ public sealed partial class StaminaComponent : Component
...
+    /// Maximum vector amount we add to the JitterAmplitudeMin
+    /// </summary>
+    [DataField]
+    public float JitterAmplitudeMod = 0.05f;
+
+    /// <summary>
+    /// Min and max multipliers for JitterAmplitude in the X direction for the random value we will choose
+    /// </summary>
+    [DataField]
+    public (float, float) JitterMinMaxX = (0.5f, 1f);
```
slarticodefast: why not use a Vector2 for this?

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -75,4 +76,79 @@ public sealed partial class StaminaComponent : Component
...
+
+    /// <summary>
+    /// Min and max multipliers for JitterAmplitude in the X direction for the random value we will choose
+    /// </summary>
+    [DataField]
+    public (float, float) JitterMinMaxY = (0.125f, 0.25f);
```
slarticodefast: same

Content.Shared/Damage/Components/StaminaComponent.cs
```diff
@@ -75,4 +76,79 @@ public sealed partial class StaminaComponent : Component
...
+    public float AnimationThreshold = 50;
+
+    /// <summary>
+    /// Minimum y vector displacement for breathing at AnimationThreshold
+    /// </summary>
+    [DataField]
+    public float BreathingAmplitudeMin = 0.04f;
+
+    /// <summary>
+    /// Maximum y vector amount we add to the BreathingAmplitudeMin
```
slarticodefast, line 96: Why breathing? The name seems a little confusing

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -352,6 +363,14 @@ private void OnUnequipAttempt(EntityUid uid, StunnedComponent stunned, IsUnequip
     }
 
     #endregion
+
+    public virtual void TryStunAnimation(Entity<StunnedComponent?> entity)
+    {
+        if (!Resolve(entity, ref entity.Comp))
+            return;
+        // Here so server can tell the client to do things
+        entity.Comp.Visualized = true;
```
slarticodefast: doesn't this need dirtying?

Content.Shared/Stunnable/SharedStunSystem.cs
```diff
@@ -107,9 +109,18 @@ private void OnMobStateChanged(EntityUid uid, MobStateComponent component, MobSt
 
     }
 
+    private void OnStunShutdown(Entity<StunnedComponent> ent, ref ComponentShutdown args)
+    {
+        // This exists so the client can end their funny animation if they're playing one.
+        UpdateCanMove(ent, ent.Comp, args);
+
+        if (_timing.IsFirstTimePredicted)
+            ent.Comp.Visualized = false;
```
slarticodefast: why change it while the component is shutting down anyways?

slarticodefast (commented):
Just had a quick look, not a full review

Princess-Cheeseballs:
> Since we have the stars, the jitter while in crit could be slightly weaker. Also, the jitter and stars don't work when stunning another entity + don't update when stunning somebody outside of your PVS.
> 
> They only update after like, a solid while of being stunned.

Stars update properly now, jittering still isn't updating properly.

Princess-Cheeseballs:
Should work now, animation plays for all clients properly and stars update properly.

There is a minor issue where SSD players don't show stars which appears to be related to a prediction issue where the game thinks they are asleep until the server tells them they're not.

Content.Shared/Damage/Components/StaminaComponent.cs
Princess-Cheeseballs: It's used purely for the "breathing" part of the layered animation. It'd be a lot more clear if the Animation System was better but I added code comments to the Animation Constructor to explain what it does.

Fildrance:
Current code is not going to play stun animation if you enter PVS after network message about animation start was sent, this is not quite nice :< one of the ideas is to sub for AfterAutoHandleState for stun and put TryPlayAnimation there, but that will need to work nicely with https://github.com/space-wizards/docs/pull/307 if possible

Content.Server/Damage/Systems/StaminaSystem.cs
```diff
@@ -1,4 +1,6 @@
-﻿using Content.Shared.Damage.Systems;
+﻿using Content.Shared.Damage.Components;
+using Content.Shared.Damage.Systems;
+using Robust.Shared.Player;
```
ScarKy0:
```suggestion
using Content.Shared.Damage.Systems;
```
Unused

Content.Client/Damage/Systems/StaminaSystem.cs
```diff
@@ -1,7 +1,125 @@
...
+
+        // If the animation is running, the system should update it accordingly
+        // If we're below the threshold to animate, don't try to animate
+        // If we're in stamcrit don't override it
+        if (entity.Comp.AnimationThreshold > entity.Comp.StaminaDamage || _animation.HasRunningAnimation(entity, StaminaAnimationKey))
+            return;
+
+        // Don't animate if we're dead
+        if (TryComp<MobStateComponent>(entity, out var state) && state.CurrentState == MobState.Dead)
+            return;
```
ScarKy0: Could probably be replaced with MobStateSystem's IsDead so if anything changes there this will not need to be updated.

Content.Client/Stunnable/StunSystem.cs
```diff
@@ -1,9 +1,243 @@
...
+    }
+
+    /// <summary>
+    ///     Add stun visual layers
+    /// </summary>
+    private void OnComponentInit(Entity<StunnedComponent> entity, ref ComponentInit args)
+    {
+        if (_timing.IsFirstTimePredicted)
+            return;
+
```
ScarKy0:
```suggestion
    {
        if (!_timing.IsFirstTimePredicted)
            return;

```
I'm not a prediction expert but shouldn't this be negated so it only runs on first prediction, instead of on every but first?

Content.Client/Stunnable/StunSystem.cs
```diff
@@ -1,9 +1,243 @@
...
+        if (!_spriteSystem.LayerMapTryGet((entity, entity.Comp2), StunVisualLayers.StamCrit, out var index, false))
+            return;
+
+        // Don't animate if we're not conscious
+        if (!Blocker.CanConsciouslyPerformAction(entity))
+            entity.Comp1.Visualized = false;
+
+        _spriteSystem.LayerSetVisible((entity, entity.Comp2), index, entity.Comp1.Visualized);
+        _spriteSystem.LayerSetRsiState((entity, entity.Comp2), index, "stunned");
```
ScarKy0: Instead of having "stunned" directly you should either keep the layer name in StunnedComponent or as a variable below dependancies

ScarKy0 (changes requested):
Some small stuff

Princess-Cheeseballs:
Reviews addressed, works with AppearanceComponent, I am hydrated, I slept well last night, and I ate a balanced breakfast it should be gtg.

Content.Client/Stunnable/StunSystem.cs
```diff
@@ -1,9 +1,252 @@
...
+    }
+
+    /// <summary>
+    ///     Add stun visual layers
+    /// </summary>
+    private void OnComponentInit(Entity<StunnedComponent> entity, ref ComponentInit args)
     {
+        if (!_timing.IsFirstTimePredicted)
+            return;
+
```
ScarKy0: Is this check even neccessary? How does one predict adding sprite layers.

Content.Client/Stunnable/StunSystem.cs
```diff
@@ -1,9 +1,252 @@
...
+    /// <param name="sprite">The spriteComp we're rotating</param>
+    /// <param name="frequency">How many times per second we're rotating</param>
+    /// <param name="rotateClamp">Maximum angle of rotation (in radians)</param>
+    /// <param name="startAngle">Default starting angle of rotation (used because we don't have adjustment layers)</param>
+    /// <returns></returns>
+    public Animation GetTwitchAnimation(SpriteComponent sprite,
+        float frequency,
+        (float, float) rotateClamp,
+        Angle startAngle)
+    {
```
ScarKy0: Since we don't use twitching anymore, this can probably be removed.

ScarKy0:
After testing in-game everything works as it should, so after the reviews are addressed i'm gonna drop my approval

Princess-Cheeseballs:
> After testing in-game everything works as it should, so after the reviews are addressed i'm gonna drop my approval

Crawling just a week away

ScarKy0 (approved):
Peak

Content.Client/Damage/Systems/StaminaSystem.cs
```diff
@@ -1,7 +1,126 @@
...
+    private void OnAnimationCompleted(Entity<StaminaComponent> entity, ref AnimationCompletedEvent args)
+    {
+        if (args.Key != StaminaAnimationKey || !args.Finished || !TryComp<SpriteComponent>(entity, out var sprite))
+            return;
+
+        // stop looping if we're below the threshold
+        if (entity.Comp.AnimationThreshold > entity.Comp.StaminaDamage)
+        {
+            _animation.Stop(entity.Owner, StaminaAnimationKey);
+            _sprite.SetOffset((entity, sprite), entity.Comp.StartOffset);
```
SlamBamActionman, line 95: This function is slightly problematic in that it does not account for other systems setting the offset; as luck would have it this isn't really done, but you can observe it with the CryoPod system which does use it. If you stun someone and put them in a cryopod, they will be offset incorrectly in the pod.

Content.Client/Stunnable/StunSystem.cs
```diff
@@ -1,9 +1,206 @@
...
+
+        if (!_spriteSystem.LayerMapTryGet((entity, entity.Comp2), StunVisualLayers.StamCrit, out var index, false))
+            return;
+
+        if (!Appearance.TryGetData<bool>(entity, StunVisuals.SeeingStars, out var stars))
+            return;
+
+        // Don't animate if we're not conscious
+        if (!Blocker.CanConsciouslyPerformAction(entity))
+            Appearance.SetData(entity, StunVisuals.SeeingStars, false);
```
SlamBamActionman: (I may be incorrect but) I don't think there's much use in setting the data here on the clientside, since the point of AppearanceData is that it can be used by other clients to replicate the sprite state. The check should probably either be done in shared, or the check sets the stars bool to false without updating the AppearanceData.

Content.Client/Damage/Systems/StaminaSystem.cs
Princess-Cheeseballs: Not sure if there's a solution to this other than specifically stopping and starting the animation when entering an exiting a cryo-pod. Animation system doesn't have any way to pause and it just sets values directly without any practical way to get a total offset from an animation.

Content.Client/Stunnable/StunSystem.cs
Princess-Cheeseballs: Yeah let me move this to shared and make it event based.

Princess-Cheeseballs: Should also prevent calling the AppearanceChangedEvent twice

Content.Client/Damage/Systems/StaminaSystem.cs
Princess-Cheeseballs: You can do the same bug with meth currently. There's not really any solution I can find that doesn't involve hardcoding a fix to the cryopod system directly, and should probably just be another PR that also fixes the same issue with meth.

SlamBamActionman: I would make an issue about it (unless there is one) and then leave it at that; it'd be too much of a refactor to include that in this PR and honestly it's not much of a blocker.

SlamBamActionman:
Seems to work as intended from testing and code looks good. There might be some future visualizer refactor necessary to separate out the animation visuals from the stun code itself (and I consider the way you've implemented this to make that fairly easy), but there's no common standard settled on there, so this works well for now. 

FairlySadPanda:
Sick. I can take this ball and run now with mob animation rewriting

ScarKy0:
Crawling just a week away

PR #37197, opened 2025-05-05
Add support for lattice removal prevention

SlamBamActionman:
## About the PR

This PR adds support for the `BlockLatticeRemovalComponent`, which makes it impossible to remove the lattice of a grid. 

This does not update any shuttles; that has to be done in separate PRs.

## Why / Balance

This component is intended for Evac shuttles. Evac shuttles has a bit of a strange issue:

1. We can't allow extending the Evac shuttle grid, as that could make it possible to block docking by building new lattice along its sides.
2. We want to make it repairable in the event of a bombing.
3. Strong bombs can remove lattice.

These combined aren't possible since when the lattice is removed, that space is considered "not-grid" and can therefore not be repaired. So the solution this PR proposes is to keep the lattice on Evac; this still allows for it to space the environment, but crew are also able to quickly patch the hole.

## Technical details

Added the `ReplaceTileAttemptEvent` event, which is raised whenever a tile should be replaced by another. 

## Requirements

## Breaking changes

**Changelog**

No CL

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
+++ b/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
@@ -507,7 +507,8 @@ public sealed partial class ExplosionSystem
         int maxTileBreak,
         bool canCreateVacuum,
         List<(Vector2i GridIndices, Tile Tile)> damagedTiles,
-        ExplosionPrototype type)
+        ExplosionPrototype type,
+        EntityUid gridUid)
     {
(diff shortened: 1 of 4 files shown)


ArtisticRoomba:
My hero.

SlamBamActionman:
#36989 is a better solution than this PR tho ngl

Pixeltheaertist:
We need this so bad on Trieste i was just about to code this exact thing

EmoGarbage404:
Superseded by #36989

PR #37199, opened 2025-05-05
StaminaSystem to SharedStaminaSystem

Pronana:
## About the PR
Changed StaminaSystem to SharedStaminaSystem and changed from sealed to abstract

## Why / Balance
Needed for #37196
Needed for Stamina animations even if above PR is rejected

## Technical details
Just file renaming

## Requirements

## Breaking changes

StaminaSystem is now SharedStaminaSystem

**Changelog**

### DIFF
--- a/Content.Client/Damage/Systems/StaminaSystem.cs
+++ b/Content.Client/Damage/Systems/StaminaSystem.cs
@@ -1,0 +1,7 @@
+﻿using Content.Shared.Damage.Systems;
+
+namespace Content.Client.Damage.Systems;
+
+public sealed class StaminaSystem : SharedStaminaSystem
+{
+}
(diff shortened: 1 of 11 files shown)


Content.Client/Damage/Systems/StaminaSystem.cs
```diff
@@ -0,0 +1,7 @@
+﻿using Content.Shared.Damage.Systems;
+
+namespace Content.Client.Damage.Systems;
+
+public sealed class StaminaSystem : SharedStaminaSystem
```
slarticodefast, line 5:
```suggestion
public sealed partial class StaminaSystem : SharedStaminaSystem
```
should be partial so that forks can easier extend it from another file
same for all the others

PR #37203, opened 2025-05-05
Playtime commands improved time input format

Errant-4:
## About the PR
The 'playtime_' commands can now take time specified as the new brief alphanumeric timespans, such as "1.5h" for 90 minutes. (seconds and minutes can also be specified, technically, but are not likely to be used). A number input without time unit will be read as minutes, so there is no incompatibility with the old method.
The time displayed in the console feedback messages is now also formatted the same as the times in the playtime stat window.

PR is done and works, but only with engine update.
Requires [#5910](https://github.com/space-wizards/RobustToolbox/pull/5910)

## Why / Balance
Entering playtimes as minutes is inconvenient, when we probably typically allocate them by hours.
The old format for the command's output was very ambiguous about the time units being shown, now its consistent with other showcases of playtime.

## Media
image

## Requirements

## Breaking changes
None

**Changelog**
:cl: Errant
ADMIN:
- tweak: The 'playtime_' commands can now take time specified as brief alphanumeric timespan, such as "1.5h" for 90 minutes.

### DIFF
--- a/Content.Server/Administration/Commands/PlayTimeCommands.cs
+++ b/Content.Server/Administration/Commands/PlayTimeCommands.cs
@@ -1,5 +1,6 @@
 ﻿using Content.Server.Players.PlayTimeTracking;
 using Content.Shared.Administration;
+using Content.Shared.Localizations;
 using Content.Shared.Players.PlayTimeTracking;
 using Robust.Server.Player;
 using Robust.Shared.Console;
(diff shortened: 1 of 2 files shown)


AJCM-git:
@Errant-4 coming back to this?

Errant-4:
I'm on it, but gonna have to do some sleuthing because it broke after I merged master

Issue #18688, opened 2023-08-05
Paused ents shouldn't be raising interaction events


metalgearsloth:
1. It causes so many mapping issues from random stuff like doors or forensics being clicked on.
2. A bunch of systems have to add handlers for this shit when left-click should just be blocked.
3. Leaving verbs as is is okay I think so mappers can just verb if they really want an open door or whatever.
4. Items shouldn't be able to be picked up into admin hands when paused. The container handles just means dozens of systems have to now handle this situation instead of just preventing it entirely and letting admins move entities without container / parent changes.

See https://github.com/space-wizards/space-station-14/pull/18435 which looks more like a bandaid.

ProPeperos:
https://github.com/space-wizards/space-station-14/pull/37210 got merged, should this be closed?

PR #37210, opened 2025-05-06
Prevent opening doors in map editor by accident

### LINKED ISSUE #42835: doors should ignore interaction while paused
### What version did the issue occur in?

stable

### Description

since doors were added if you clicked a door by accident while mapping it would get a bunch of components added and stuck in the opening state until the map is loaded.
it should either be completely ignored (1 line easy) or immediately open without a delay (harder to do)

it should also be refactored to use timespans not accumulating frametime

### Reproduction

a

### Additional Context

_No response_

SpaceManiac:
## About the PR
Less-controversial alternative to #33796.

Resolves #42835

## Why / Balance
At present if a map editor accidentally clicks a door, it invisibly enters the "Opening" state (or "Closing" for firelocks) and the door has to be erased and replaced or the YAML has to be manually edited or the door will open on roundstart.

## Technical details
Paused doors can't be opened or closed.

## Requirements

### DIFF
--- a/Content.Shared/Doors/Systems/SharedDoorSystem.cs
+++ b/Content.Shared/Doors/Systems/SharedDoorSystem.cs
@@ -358,6 +358,9 @@ public abstract partial class SharedDoorSystem : EntitySystem
         if (!Resolve(uid, ref door))
             return;
 
+        if (Paused(uid))
+            return;
+
         var lastState = door.State;
 
         if (!SetState(uid, DoorState.Opening, door))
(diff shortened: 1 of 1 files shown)


EmoGarbage404 (dismissed):
shouldn't these checks be in CanOpen and CanClose?

IProduceWidgets:
I really feel like mappers need to be able to map opened doors.

- Windoors have a hold open state
- paper, wood, metal, etc doors have open states

SpaceManiac:
> shouldn't these checks be in CanOpen and CanClose?

Seems so.

> An alternative that's been suggested is to just make the change instant when the entity is paused instead of erroneously starting the door animation.

I like this too, but maintainers rejected it in #33796, and again when I argued this case on Discord.

SlamBamActionman (approved):
As per #18688 there should really just be a general solution, but this does solve a long-standing bug and closes #42835 regardless.

PR #37224, opened 2025-05-06
AI context menu fix

B-Kirill:
## About the PR
All objects with `StationAiWhitelist` component had the "Open actions" option in the context menu, even if no actions were provided for them. Now, to show this option, `ShowContextMenu` must be `true` in the component.

## Why / Balance
This could confuse the players.

## Media
_No need_

## Requirements

**Changelog**
:cl:
- fix: AI context menu buttons now only appear on objects with valid interactions.

### DIFF
--- a/Content.Shared/Silicons/StationAi/SharedStationAiSystem.Held.cs
+++ b/Content.Shared/Silicons/StationAi/SharedStationAiSystem.Held.cs
@@ -157,6 +157,9 @@ public abstract partial class SharedStationAiSystem
 
     private void OnTargetVerbs(Entity<StationAiWhitelistComponent> ent, ref GetVerbsEvent<AlternativeVerb> args)
     {
+        if (!_uiSystem.HasUi(args.Target, AiUi.Key))
+            return;
+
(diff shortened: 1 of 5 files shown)


EmoGarbage404 (changes requested):
instead of manually having to specify the bool on all users of this, can we not just add a check to see if the UI the action would open exists? We already check if it's open, so checking that it even exists earlier seems reasonable.

B-Kirill:
> instead of manually having to specify the bool on all users of this, can we not just add a check to see if the UI the action would open exists? We already check if it's open, so checking that it even exists earlier seems reasonable.

Well, I hadn't even thought about it, but it would really be much better this way.

B-Kirill:
Random fail of the build, please restart.
_By the way, would anyone mind if I create empty commits or something similar to restart the build in such cases?_

metalgearsloth:
> Random fail of the build, please restart. _By the way, would anyone mind if I create empty commits or something similar to restart the build in such cases?_

go nuts

EmoGarbage404:
holy clean fix

PR #37228, opened 2025-05-06
Make animals drop giblets into container or floor when they inserted into container

### LINKED ISSUE #25401: ORGAN MANIFESTATION
## Description

Round 46899
Somehow, when right clicking myself I began seeing multiple sets of organs of the chest cavity. I was able to take the hearts out at first and obviously I decided to eat them then. Another individual was able to interact with the organs stuck to me later, so it wasn't just me experiencing this on my end.

**Reproduction**

Prior to this I was doing a lot of hardsuit helmet changes, removing and putting on my hardsuit, and messing with a gas mask and oxygen walking around tesla.
I had two cockroaches in my inventory, notably, and I had 2 sets of the chest cavity organs. it would make sense if they came from the cockroaches.
This was on Round 46899, US West Salamander at Wizard's Den.

I was informed "yes. if they decay in your inventory the gibs get stuck to your character for.... some reason
this is probably useful information to include in the bug report" I am not sure if this (cockroaches decaying in my duffel bag) is the case but it would be worth trying.

**Screenshots**

image
image
image

**Additional context**

N/A

kosticia:
## About the PR
Now animals drop their organs on floor instead of attaching to container. 

## Why / Balance
Closes #25401.
I don't think that its intented.

## Technical details
One line change. From SetCoordinates method to SetWorldPosition.

## Media
https://github.com/user-attachments/assets/261cd504-86c8-4697-94ad-d46bd78e0783

## Requirements

## Breaking changes
N/A.

**Changelog**
:cl:


kosticia:
Fuck wrong title

EmoGarbage404 (dismissed):
the gibs really ought to just be using SpawnNextTo to begin with so that they spawn inside of the container. Attaching to the floor is better than the locker but both are weird when really it should all be staying inside.

PR #37235, opened 2025-05-06
Tweak guidebook to not reset when re-opened

sudobeans:
## About the PR
This changes the guidebook so that if the same guidebook is opened, the guidebook window isn't reset. As a result, closing and reopening the main guidebook will retain:
- Expanded tree items
- current search (on pages with a "Filter items" box)
- current scroll height

This works when using the main guidebook, but doesn't work as well when using handheld guidebooks since those always open to the first page.

## Why / Balance
Currently, the guidebook resetting each time it is open is frustrating, especially when you are trying to use the reference tables. For example, if I had pulled up the "Moonshine" in "Drinks", and then I closed and re-opened the guidebook, I would have to look it up all over again.

This PR makes it so I wouldn't have to look it up all over again.

## Technical details
Previously, the `GuidebookWindow.UpdateGuidebook` function would always rebuild the entire tree UI and select the given guide. Now it uses the new `GuidebookWindow.SameAsLastUpdate` method to determine if the guides changed, and only rebuilds if the guides changed.

It would be convenient if the guidebook items saved your position as well, but they're currently coded to always open to the first page (see [`GuidebookSystem.cs:99`](https://github.com/space-wizards/space-station-14/blob/a1dacfc62b2d8578ba268e682c10d98f0dbdba51/Content.Client/Guidebook/GuidebookSystem.cs#L99)) and I couldn't figure out how to change this without spaghetti code or refactoring.

## Media

Example of re-opening the guidebook **before** this PR. Notice how the filter and tree are reset.

https://github.com/user-attachments/assets/3ce65942-87e2-4b3b-bf12-3fdaf520e054

Example of re-opening the guidebook **after** this PR. Notice how the filter and tree are not reset.

https://github.com/user-attachments/assets/1eae5401-f98e-446c-92c3-114f5966ecb5

## Requirements

## Breaking changes
- GuidebookWindow.LastEntry renamed to GuidebookWindow.Selected.
- UpdateGuides will not rebuild the GuidebookWindow's tree and selected guide if the guides haven't changed.

**Changelog**
:cl:
- tweak: Main guidebook window doen't reset when re-opened.


sudobeans:
Rewrote the history since I screwed up during a merge.

southbridge-fur (approved):
Lgtm

PR #37237, opened 2025-05-06
Genpop closet cargo orders

### LINKED ISSUE #37218: Genpop Prisoner Closets Cannot be replaced if destroyed.
Genpop Prisoner Closet Sub-Issue 1:

They can be broken but there is no way to get new ones, so if you bomb the locker area sec's ability to use this system is crippled, adding a way to order these from cargo is advised.

themias:
## About the PR

This PR allows the new gen pop lockers to be ordered from cargo.

The sprites for the replacements will all show #&NoBreak;1, but that is a limitation until we can spray paint lockers, like in #31328.
(The prisoner's name will still appear on the locker upon examine.)

Closes #37218 

## Why / Balance

These need to be replaceable if they are destroyed or unanchored/stolen.

## Media

https://github.com/user-attachments/assets/3baaf0d4-46ab-482a-9059-e80b6f6597ef

## Requirements

**Changelog**

:cl:
- tweak: Genpop prisoner closets can be ordered from cargo

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_security.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_security.yml
@@ -81,8 +81,8 @@
 - type: cargoProduct
   id: SecurityGenPopCloset
   icon:
-    sprite: Objects/Misc/id_cards.rsi
-    state: orange
+    sprite: Structures/Storage/closet.rsi
+    state: genpop
   product: LockerPrisoner
   cost: 250
   category: cargoproduct-category-name-security
--- a/Resources/Prototypes/Catalog/Cargo/cargo_security.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_security.yml
@@ -77,3 +77,13 @@
   cost: 2500
   category: cargoproduct-category-name-security
   group: market
+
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Catalog/Cargo/cargo_security.yml
```diff
@@ -77,3 +77,13 @@
   cost: 2500
   category: cargoproduct-category-name-security
   group: market
+
+- type: cargoProduct
+  id: SecurityGenPopCloset
+  icon:
+    sprite: Objects/Misc/id_cards.rsi
```
slarticodefast, line 84: the ID card is a little confusing as the preview icon, I would directly use the locker sprite

themias: Since the icons don't allow more than one state, I didn't like the look of the locker without the door. It looks like the EOD locker uses a helmet as a stand in. In this case, it really is the locker we're after rather than its contents, so I'll update it to the locker base.

EmoGarbage404 (approved):

slarticodefast (approved):

PR #37239, opened 2025-05-06
Aloxadone Tweaks

ScarKy0:
## About the PR

Tweaked the recipe and effects of Aloxadone to get them more in terms with my other cryochem changes.

## Why / Balance

Aloxadone is currently severely underused, just like other cryochems. However the reason for that is majorly different, which is how Aloe generally can be used.

At the current moment, creation of meshes and aloe cream is not affected by the amount of aloe reagent within the plant. This makes the healing amounts between Alox and topicals vary greatly.
1 unrobusted leaf of aloe, when microwaved becomes 50 heat, 50 cold, 50 shock and 15 caustic healing.
1 unrobusted leaf of aloe, when made into a mesh, becomes 100 heat, 100 cold, 100 shock and 100 caustic.
1 unrobusted leaf of aloe, when made into alox, becomes 24 heat, 0 cold, 24 shock and 8 caustic.
This split is simply not fair, especially considering the expensive nature of the reagents needed to create alox.

This PR tweaks the numbers a bit. Healing of 24/0/24/6 becomes 64/64/64/24, which puts it in between cream and meshes. If the aloe was to be robusted to 49, it would yield about (rounded) 192/192/192/72 healing.

Now those numbers may sound extremely high, especially for robusted aloe. However, unlike topicals cryo is unable to move and leaves the user almost fully unable to act for the duration, as well as adds difficulty for self use.

Now, about adding cold healing to alox. While the main drawback of cryo is meant to be cold, it is usually easily preventable one. With the recent thermal exchange changes, it takes few seconds to warm up after cryo, and max 2 ointment uses or a shot of lepo to immediately heal that damage up, however that is not the main reason.
As it stands right now, your body ashes upon reaching 1500 heat. This is not the case for cold damage, allowing it to stack infinitely. This change is meant to allow for healing extreme amounts of cold damage if neccessary, as there is no way to handle them beyond the usual topical spam, which these changes are meant to prevent/give an alternative to.

Obviously those values can be tweaked as neccessary, however alox is currently not worth using whatsoever over turning it all into cream or meshes.

## Technical details

Yaml.

### Don't let this fool you, its actually 4 heat, 4 cold, 4 shock and 1.5 caustic
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Aloxadone has been tweaked. The recipe has been altered to be simplier to make and the healing values have been increased.


ScarKy0:
This is a cry for help

archee1:
I love medbay when you all stand around the dude who stepped in the plasma pool for the 2nd time who has 800 burns, thats my favourite part.

Aloxadone desperately needs this regardless of the other cryo-chem changes, that stuff doesnt get enough done on its own. O7

PR #37241, opened 2025-05-06
Fire damage system fixes

Alpha-Two:
## About the PR
1. Having greater than 100% fire protection no longer heals burn damage (possible with Elite Web Suit + Atmos Fire Helmet according to @RedBookcase )
2. Examine text is fixed, no longer will the CE helmet say it has 80% Fire resistance (it has 20%)
## Why / Balance
Bugs

## Technical details
MathF.Min and removing a "1f -"

## Requirements

## Breaking changes
none

**Changelog**
:cl: UpAndLeaves
- fix: wearing an elite web vest and an atmos fire helmet will no longer heal your burn damage when on fire.

### DIFF
--- a/Content.Shared/Clothing/EntitySystems/FireProtectionSystem.cs
+++ b/Content.Shared/Clothing/EntitySystems/FireProtectionSystem.cs
@@ -25,7 +25,7 @@ public sealed class FireProtectionSystem : EntitySystem
 
     private void OnArmorExamine(Entity<FireProtectionComponent> ent, ref ArmorExamineEvent args)
     {
-        var value = MathF.Round((1f - ent.Comp.Reduction) * 100, 1);
+        var value = MathF.Round(ent.Comp.Reduction * 100, 1);
 
         if (value == 0)
             return;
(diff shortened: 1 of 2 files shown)


Content.Shared/Clothing/EntitySystems/FireProtectionSystem.cs
```diff
@@ -25,7 +25,7 @@ private void OnGetProtection(Entity<FireProtectionComponent> ent, ref InventoryR
 
     private void OnArmorExamine(Entity<FireProtectionComponent> ent, ref ArmorExamineEvent args)
     {
-        var value = MathF.Round((1f - ent.Comp.Reduction) * 100, 1);
+        var value = MathF.Round(ent.Comp.Reduction * 100, 1);
```
EmoGarbage404, line 28: why is this being changed? Doesn't it display the reduction, not the coefficient?

RedBookcase: > why is this being changed? Doesn't it display the reduction, not the coefficient?

Fire resistance isn't a coefficient like other damage resistances - it's a flat reduction. But it gets displayed inversely as if it _were_ a coefficient. So something like the atmos fire suit, which has an 80% fire reduction, gets listed as having only a 20% fire reduction. Very misleading.

image

image

RedBookcase:
> Maybe push this to stable? Seems exploitable currently

Even when testing it with maximum firestacks, running through lava and dousing myself with phlog, I was never even able to heal 1 heat damage a second. Given that the atmos fire helmet + elite web vest combo already makes you basically immune to heat damage it's not gonna cause anything game breaking in my opinion.

EmoGarbage404:
man fuck this code

Issue #35128, opened 2025-02-13
Strange interaction with mice in locked lockers


Booblesnoot42:
## Description
If a) a player-controlled mouse is placed in a lockable locker, b) the locker is locked, and c) the mouse player attempts to move, then the mouse will begin "kicking at the door." After the doAfter, the locker will unlock, but because the mouse cannot open the locker door they remain stuck.

**Reproduction**
1. Spawn a lockable locker.
2. Spawn a mouse.
3. Place the mouse in the locker.
4. Close and lock the locker.
5. Ghost and control the mouse.
6. Attempt to move.
7. After the locker unlocks, attempt to move again.

Issue #28097, opened 2024-05-17
Creatures without hands are able to unbuckle other creatures


MjrLandWhale:
## Description

Had a round on mander where I played as a mouse. I found a catatonic player hiding in the bathroom, and found that I was able to unbuckle them from the toilet even though I have no hands. As expected, I was unable to buckle them back into their seat, as I had no hands.

**Reproduction**

Attempt to unbuckle a buckled creature as a mouse.
**Screenshots**

**Additional context**

lzk228:
seems easy to fix
just add check for userUid handscomp
https://github.com/space-wizards/space-station-14/blob/6082df7eac5d6ca443f4a0740227c4d281552f9a/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs#L385-L386
https://github.com/space-wizards/space-station-14/blob/6082df7eac5d6ca443f4a0740227c4d281552f9a/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs#L396-L422

Issue #10666, opened 2022-08-17
Various mouse issues as in the animal not peripheral.


Androidonator:
## Description
1. You can rotate shuttle consoles as mouse
2. Setup cameras even they have been setup already. I didn't bother to check if anything got changed.
3. When jumping of someone's head you can phase trough walls if they are moving.

**Reproduction**
1. Right click shuttle console as mouse --> rotate
2. Right click camera --> setup
3. Bother someone so they grab you or put you on their heads --> wait for some amount of time --> start moving when they start moving 
**Screenshots**
N/A

**Additional context**
Playing mouse too much

Androidonator:
When I am looking trough the animals yaml it looks like all animals might have the same issue. Not sure where to find it.

IProduceWidgets:
replicated 
> You can rotate shuttle consoles as mouse
> Setup cameras even they have been setup already. I didn't bother to check if anything got changed.

PR #37246, opened 2025-05-07
Add checks for various complex interactions

Krunklehorn:
## About the PR
Prevents abuse/weirdness by considering the following actions complex:
- Switching gun modes, open/closing gun bolts, racking guns
- Toggling radiation collectors, emitters, welders & stun batons
- Unbuckling others (can still unbuckle yourself)
- Setting up cameras and camera routers
- Rotating and flipping objects
- ~Resisting locks~

Also ensures the following actions require hands:
- Switching gun modes, open/closing gun bolts, racking guns

Additionally, unbuckling cannot occur even if the user (ex: a dragon) tries to pull the target off. I can tackle this if necessary.

https://discord.com/channels/310555209753690112/1369481928893534409
~#35128~
#28097
#10666 (1 & 2, not 3)

## Why / Balance
Radiation collectors, cameras, camera routers and guns should obviously require humanoid-level intellect.
For everything else, a more fleshed out solution might consider the relative size of the user and target, if necessary.

## Media

https://github.com/user-attachments/assets/216f3e24-afdd-44df-a497-c7702694bd40

Content Client_zh2jGnx5XK
Content Client_UrILTNGzAX
Content Client_aBdbjpqSio
image

## Requirements

## Breaking changes
`SharedBuckleSystem.Interaction:OnStrapDragDropTarget()` now checks `CanBuckle()` to exit early, skipping the doafter.
`SharedBuckleSystem.Interaction:AddUnbuckleVerb()` also checks `CanUnbuckle()` to prevent the verb.
`StunBatonSystem:TryTurnOn()` is now a protected override method.

**Changelog**
:cl:
- fix: Animals can no longer fiddle with guns, radiation collectors, emitters, welders, stun batons, cameras and camera routers.
- fix: Animals can no longer rotate and flip buildings or unbuckle others.

### DIFF
--- a/Content.Server/Resist/ResistLockerSystem.cs
+++ b/Content.Server/Resist/ResistLockerSystem.cs
@@ -50,6 +50,9 @@ public sealed class ResistLockerSystem : EntitySystem
         if (!Resolve(target, ref storageComponent, ref resistLockerComponent))
             return;
 
+        if (!_actionBlocker.CanComplexInteract(user))
+            return;
+
         var doAfterEventArgs = new DoAfterArgs(EntityManager, user, resistLockerComponent.ResistTime, new ResistLockerDoAfterEvent(), target, target: target)
         {
             BreakOnMove = true,
--- a/Content.Server/Resist/ResistLockerSystem.cs
+++ b/Content.Server/Resist/ResistLockerSystem.cs
@@ -50,9 +50,6 @@ public sealed class ResistLockerSystem : EntitySystem
         if (!Resolve(target, ref storageComponent, ref resistLockerComponent))
             return;
 
-        if (!_actionBlocker.CanComplexInteract(user))
-            return;
-
         var doAfterEventArgs = new DoAfterArgs(EntityManager, user, resistLockerComponent.ResistTime, new ResistLockerDoAfterEvent(), target, target: target)
         {
             BreakOnMove = true,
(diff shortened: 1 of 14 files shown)


Content.Server/Resist/ResistLockerSystem.cs
```diff
@@ -50,6 +50,9 @@ private void AttemptResist(EntityUid user, EntityUid target, EntityStorageCompon
         if (!Resolve(target, ref storageComponent, ref resistLockerComponent))
             return;
 
+        if (!_actionBlocker.CanComplexInteract(user))
```
slarticodefast, line 53: I'm pretty sure mice are supposed to be able to escape an EntityStorage. Otherwise you could lock someone playing an animal into a locker and they could not espace, which would be quite annoying for that player and basically round remove them. The issue you linked mentions it starts the doafter, but the escaping itself is broken, so that should be fixed instead of completely disabling it.

beck-thompson (approved):
Seems good!

pheenty:
> I would also like to have them be unable to eject every item from a disposals bin but i'll approve anyway

This means that you'd be able to hide from zombies in bins and they wouldn't be able to so anything about it

Errant-4:
> > I would also like to have them be unable to eject every item from a disposals bin but i'll approve anyway
> 
> This means that you'd be able to hide from zombies in bins and they wouldn't be able to so anything about it

Solution, it should be possible to beat the shit out of disposal bins

PR #16300, opened 2023-05-10
Regular boxes are now destructible

PeterFuto:
#16187

Syndie boxes are indestructible.
Added a base type.

- tweak: Regular boxes are now destructible

PR #37247, opened 2025-05-07
Stealth Cardboard Box Destructible

Chartman95:
## About the PR
This PR makes removes the `BaseBigBox` abstract entity, combining and replacing it with the `BigBox` entity (the large cardboard box), and makes it the parent of the `StealthBox` entity (the stealth cardboard box). The `StealthBox` entity is given a large base health pool of 300 but is no longer indestructible as it was previously

## Why / Balance
The `StealthBox` had an infinite capacity to absorb damage and projectiles, which was both inconsistent with all other storage objects and abusable in that it could be used as an infinite shield against projectiles. This was seen especially when tail-dragged by lizards with two handed weapons.
The Stealth Cardboard Box's functionality has not changed but it now has a moderate health pool (currently 300 before reductions.)

## Technical details
Removed the `BaseBigBox` abstract as it was deemed unnecessary.
Combined the the `BigBox` entity with the old `BaseBigBox` abstract to serve as the new parent entity for the `StealthBox`.
Gave the `StealthBox` entity a damage pool of 300 before being destroyed.
Changed references to `BaseBigBox` to `BigBox` in `big_boxes.yml`, `bigbox.yml`, and `crates.yml`.

## Media
**Before:**
https://github.com/user-attachments/assets/c478b94a-a3f8-4510-ae77-e7725eee76e7

**After:**
https://github.com/user-attachments/assets/39afea21-5550-4b83-9828-882c684cea32

## Requirements

## Breaking changes
`BaseBigBox` entity no longer exists as a parent abstract or crafting graph. 

**Changelog**
:cl:
- tweak: Stealth Cardboard Boxes no longer indestructible.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
+++ b/Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
@@ -59,7 +59,7 @@
 - type: entity
   id: StealthBox
   suffix: stealth
-  parent: BaseBigBox
+  parent: BigBox
   description: Kept ya waiting, huh?
   components:
     - type: Damageable
--- a/Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
+++ b/Resources/Prototypes/Entities/Structures/Storage/Closets/big_boxes.yml
@@ -1,8 +1,7 @@
 ﻿- type: entity
-  id: BaseBigBox
+  id: BigBox
   name: cardboard box #it's still just a box
   description: Huh? Just a box...
-  abstract: true
   components:
     - type: Transform
       noRot: true
@@ -51,10 +50,19 @@
       tags:
         - DoorBumpOpener
     - type: Construction
-      graph: BaseBigBox
-      node: basebigbox
+      graph: BigBox
+      node: bigbox
       containers:
         - entity_storage
+    - type: InteractionOutline
+    - type: Destructible
+      thresholds:
+      - trigger:
+          !type:DamageTrigger
+          damage: 15
+        behaviors:
+        - !type:DoActsBehavior
+          acts: [ "Destruction" ]
 
 - type: entity
   id: StealthBox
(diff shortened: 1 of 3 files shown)


slarticodefast:
> This seems to be how it was originally coded to work but someone forgot to set the parent to be the default carboard box instead of the base box entity which has no damage threshold.

That doesn't seem to be the case to me because now `BaseBigBox` isn't needed anymore. The interaction outline is already disabled by the stealth component. So the only difference and reason to make an abstract parent is to explicitly make them indescructible.

Chartman95:
> Edit: Ok, I looked up the git history and this PR #16300 made only normal cardboard boxes destructible by adding the abstract parent. We can do the same for the syndie boxes, but it would be a balance change and not a bugfix.

Ah, my mistake then. I had assumed the regular box being flimsy and the Stealth box having a damage modifier and the comment "Syndicate boxes should have a bit of protection." meant that it was supposed to be just slightly harder to kill.

As it stands the stealth box has an infinite capacity to block projectiles and soak damage, albeit while passing the damage on to any occupants. It can still be used as a static means to block infinitely which can be dragged around - which didn't seem to be an intended feature to me.

Should I change the wording of this PR to indicate its a balance change as you suggested? Or is more thought required into giving the Stealth Box a different damage modifier type and or its own health pool? Either are easily implemented.

slarticodefast:
If we want to make it destructible for balance reasons, then mention those in the PR description.
As for the yaml code, please clean up the unused base prototype.

keronshb:
The box is already kind of useless, I don't want to make it destructable.
The person inside can get damaged, there's no point in it being able to be destroyed.

superjj18:
> The box is already kind of useless

I would say underutilized rather than useless

Chartman95:
> The person inside can get damaged, there's no point in it being able to be destroyed.

Is it not inconsistent that an object that can take damage and absorb/block projectiles cannot eventually be destroyed? 
I could  clean up the abstract as suggested and give the box the same properties and health of say, a reinforced wall or something? 
Surely a clake, C4 or or at least the nuke should destroy it?

keronshb:
> > The person inside can get damaged, there's no point in it being able to be destroyed.
> 
> Is it not inconsistent that an object that can take damage and absorb/block projectiles cannot eventually be destroyed? I could clean up the abstract as suggested and give the box the same properties and health of say, a reinforced wall or something? Surely a clake, C4 or or at least the nuke should destroy it?

Nah I think it's funnier that way. 

> It's funny because it's absurdly powerful with nukies (5 nukies hiding in a box in a corner + 100% move speed)

This is technically not intended but it's funny.

Chartman95:
> If we want to make it destructible for balance reasons, then mention those in the PR description.
> As for the yaml code, please clean up the unused base prototype.

OK, I have done as suggested.
Base abstract entity has been removed and replaced with the default cardboard box entity.
The StealthBox has been given an initial damage pool of 300 (which is effectively increased by its Flimsymetallic modifier).
Happy to tweak that value to be however tanky is deemed fit for the box so long as its no longer invincible.

Chartman95:
> 300 is way too much

Well I had to pick somewhere between 15 and infinity.
Three hundred seemed conservative all things considered.

ArtisticRoomba:
Yeah, as Keron said, the stealthbox should ideally be an implant which would alleviate this entire concern and make it a cool item and viable.

PR #37248, opened 2025-05-07
Doxa Tweaks

ScarKy0:
## About the PR

Tweaked the Doxa recipe to make it more difficult. Doxa now works on the dead and shares effects similiar to Phalanx.
Requires #37131

## Why / Balance

Mainly to be in-tune with the other cryochem changes. The recipe has been made to require Phalanx as to make it a bit more difficult to obtain, and as such was also given the same downsides that phalanx has (cannot be used alongside arith, and in this case also stellox).
There is not many ways you can get RRd in-game by genetic damage, so the balance implications arent huge except the ability to cure secoffs after failed implant checks, however the requirement for them to sit in medical and await cryo treatment (instead of just being given a jug of phalanx by med and leaving) feel like they balance it out to me.
The values and metabolism can always be adjusted.

## Technical details

All yaml.

### Ignore it says cryox, this PR just depends on the Stellox PR and I didn't feel like retaking screenshots.
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Doxarubixadone has been tweaked. The recipe has been altered to be slightly more expensive. It now has a side effect similiar to Phalanximine (deals caustic when used alongside arith and stellox), but in turn also works on the dead.

### DIFF
--- a/Resources/Locale/en-US/reagents/meta/medicine.ftl
+++ b/Resources/Locale/en-US/reagents/meta/medicine.ftl
@@ -17,7 +17,7 @@ reagent-name-cryoxadone = cryoxadone
 reagent-desc-cryoxadone = Required for the proper function of cryogenics. Heals all standard types of damage, but only works in temperatures under 213K. It can treat and rejuvenate plants when applied in small doses.
 
 reagent-name-doxarubixadone = doxarubixadone
-reagent-desc-doxarubixadone = A cryogenics chemical. Heals cellular damage caused by dangerous gasses and chemicals.
+reagent-desc-doxarubixadone = A cryogenics chemical. Heals cellular damage caused by dangerous gasses and chemicals. Works regardless of the patient being alive or dead.
 
 reagent-name-dermaline = dermaline
 reagent-desc-dermaline = An advanced chemical that is more effective at treating burn damage than kelotane.
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -232,7 +232,7 @@
           - !type:Temperature
             max: 213.0
           - !type:ReagentThreshold
-            reagent: Cryoxadone
+            reagent: Stelloxadone
             min: 1
           damage:
             types:
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -204,6 +204,7 @@
   physicalDesc: reagent-physical-desc-bubbling
   flavor: medicine
   color: "#32cd32"
+  worksOnTheDead: true
   metabolisms:
     Medicine:
       effects:
@@ -213,7 +214,29 @@
             max: 213.0
           damage:
             types:
-             Cellular: -2
+             Cellular: -3
+             Radiation: 1.5
+             Caustic: 1.5
+        - !type:HealthChange
+          conditions:
+          - !type:Temperature
+            max: 213.0
+          - !type:ReagentThreshold
+            reagent: Arithrazine
+            min: 1
+          damage:
+            types:
+              Caustic: 3
+        - !type:HealthChange
+          conditions:
+          - !type:Temperature
+            max: 213.0
+          - !type:ReagentThreshold
+            reagent: Cryoxadone
+            min: 1
+          damage:
+            types:
+              Caustic: 3
 
 - type: reagent
   id: Dermaline
--- a/Resources/Prototypes/Recipes/Reactions/medicine.yml
+++ b/Resources/Prototypes/Recipes/Reactions/medicine.yml
@@ -71,7 +71,7 @@
   reactants:
     Cryoxadone:
       amount: 1
-    UnstableMutagen:
+    Phalanximine:
       amount: 1
   products:
     Doxarubixadone: 2


Gentleman-Bird:
A balance implication I noticed is that this version of Doxa can be used to RR dead bodies if antags can get ahold of it. Shove a dead body in space or somewhere with a freezer, and unhealable damage can be stacked up until the body is unrevivable. Also, an inattentive medical doctor can accidentally RR a body this way if they inject too much in.

Nrd-K:
In my opinion, the dox should not have been so complicated, it was already, let's say, a transitional stage from phalanximine to the use of cryopods. According to your PR, he can be said to pass the stage of the usual cryo-substance, going straight to the treatment of the dead, which is not good for me. Using your PR about traumaxodone as an example. First, people will use bicaridin, then bruizin, lacerinol, and SO on, then they will switch to your traumaxodone and only then they will go to necrozole, which heals the dead. In this example, there is a sense of progression. And in the case of your dox change, it goes straight to the stage of treating the dead after Phalanximine. In my PR, I just suggest adding this next stage of treatment of the dead in the form of Celliminol, which will be possible only after the creation of the dox
#37112

MozarteanChaos:
> A balance implication I noticed is that this version of Doxa can be used to RR dead bodies if antags can get ahold of it. Shove a dead body in space or somewhere with a freezer, and unhealable damage can be stacked up until the body is unrevivable. Also, an inattentive medical doctor can accidentally RR a body this way if they inject too much in.

It'd require using more than an entire small beaker's worth of doxa to RR with it - 50u would deal 150 damage. Most cryo setups come stocked with 50u beakers, so I doubt accidental RRs will be very likely with this.
It's also... not actually an RR, as this PR is stated to depend on stelloxadone, which is a corpse-compatible toxin medicine.

Nrd-K:
> > A balance implication I noticed is that this version of Doxa can be used to RR dead bodies if antags can get ahold of it. Shove a dead body in space or somewhere with a freezer, and unhealable damage can be stacked up until the body is unrevivable. Also, an inattentive medical doctor can accidentally RR a body this way if they inject too much in.
> 
> It'd require using more than an entire small beaker's worth of doxa to RR with it - 50u would deal 150 radiation damage. Most cryo setups come stocked with 50u beakers, so I doubt accidental RRs will be very likely with this. It's also... not actually an RR, as this PR is stated to depend on stelloxadone, which is a corpse-compatible toxin medicine. Juggling the two will be _annoying,_ but each heals more than the other causes, so... it's theoretically possible.

In the case of the amount of damage, and it's huge, I agree, it won't be better than phalanximine, because doctors will have to inject several cryo medications just to heal the other types of damage that the new dox will cause, and this is a huge waste of other medications. Also, in the case of phalanximine, it is possible to regulate what a person infuses through a syringe, when, as a cryo pod, he will consume dox on his own and he will kill the patient faster at the front than he will cure
In general, as for me, with such a dock it is already easier to use phalanximine than cryo pods.

MozarteanChaos:
> > It'd require using more than an entire small beaker's worth of doxa to RR with it - 50u would deal 150 radiation damage. Most cryo setups come stocked with 50u beakers, so I doubt accidental RRs will be very likely with this. It's also... not actually an RR, as this PR is stated to depend on stelloxadone, which is a corpse-compatible toxin medicine. Juggling the two will be _annoying,_ but each heals more than the other causes, so... it's theoretically possible.
> 
> In the case of the amount of damage, and it's huge, it won't be better than phalanximine, because doctors will have to inject several cryo medications just to heal the other types of damage that the new dox will cause, and this is a huge waste of other medications. Also, in the case of phalanximine, it is possible to regulate what a person infuses through a syringe, when, as a cryo pod, he will consume dox on his own and he will kill the patient faster at the front than he will cure In general, as for me, with such a dock it is already easier to use phalanximine than cryo pods.

You can regulate how much doxa is administered using cryo - simply put a different amount in the beaker, or take the beaker out when the healing is done. All the cryopods do is gradually and automatically inject a chemical, but you can still decide how much of that chemical is available to them.
Also, I feel it's worth noting that with the values currently proposed by this PR, doxa heals _exactly 10x faster_ than phalanx. To some doctors, that speed increase will be worth the risk and inconvenience of cryo.

Nrd-K:
> 
> You can regulate how much doxa is administered using cryo - simply put a different amount in the beaker, or take the beaker out when the healing is done. All the cryopods do is gradually and automatically inject a chemical, but you can still decide how much of that chemical is available to them. Also, I feel it's worth noting that with the values currently proposed by this PR, doxa heals (and damages) _exactly 10x faster_ than phalanx. To some doctors, that speed increase will be worth the risk and inconvenience.

I know how cryo works, but I'm taking into account possible scenarios. Anyway, a person will not measure the exact amount of healing before the unit, but will only pour an approximate amount, secondly, let's take a method for obtaining the maximum dose of cellular damage, which is currently almost impossible to obtain. In this case, you will need about 66 units (in the case of 200 damage), and it will cause about 200 damage in turn (taking into account the 200 cell damage), but of other types... And thirdly, your suggested method, which was to get the beaker when the treatment was over. In this case, there is a high probability of giving much more damage, again due to a human reaction. Fourth, it made sense to use dox when a person has an average number of cells, but why use it now if it literally changes one type of damage to two others, for which you will again have to select an approximate amount of medicine?
You can also take many other scenarios, from the banal inattention of doctors to a beginner in this business, but the fact remains that despite faster healing, the waste of medication, as well as the total time (taking into account the use of all medications), becomes much more than with the previous dox

iaada:
Do you intend to continue this PR with the other cryo changes getting reverted?

Issue #22490, opened 2023-12-14
Pinpointers in a backpack can point the wrong way


Errant-4:
## Description
Pinpointer sprite needs to remain upright when it's in-inventory-box is rotated, because the pointer arrow is aligned to the sprite, not the screen

**Reproduction**
Put pinpointer in a backpack and rotate it

**Screenshots**
![image](https://github.com/space-wizards/space-station-14/assets/35878406/d4168ffe-abb9-431c-9e71-ccdb0d48879c)

**Additional context**

Errant-4:
Any item with `noRot: true` set on their SpriteComponent will display their inventory sprite rotated 180 degrees, when the tetris block is rotated 90 or 270 degrees. 

![image](https://github.com/space-wizards/space-station-14/assets/35878406/d6a572ed-8411-4328-b7ac-8fa4b03abcec)

I'm not sure if there is any actual reason for this behavior in the first place, but I haven't found the relevant code that does it yet

PR #37249, opened 2025-05-07
Fix pinpointer showing the wrong way in a backpack

psykana:
## About the PR
Band-aid fix to https://github.com/space-wizards/space-station-14/issues/22490

## Why / Balance
Running AWAY from the disk as a nukie is not fun

## Technical details
Pinpointer now checks if it's in a storage and accounts for sprite rotation

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Flipped pinpointer in a backpack now points the right way


Content.Client/Pinpointer/PinpointerSystem.cs
```diff
@@ -18,13 +23,26 @@ public override void Update(float frameTime)
...
             if (!pinpointer.HasTarget)
                 continue;
             var eye = _eyeManager.CurrentEye;
             var angle = pinpointer.ArrowAngle + eye.Rotation;
 
+            // Pinpointer may be flipped in a backpack etc, adjust for it
+            if (_container.TryGetContainingContainer((uid, null, null), out var container))
+            {
+                if (!TryComp<HandsComponent>(container.Owner, out var _))
+                {
```
EmoGarbage404, line 37: would this not work the same if you just removed these two checks? TryGetStorageLocation should only return true if the item is currently in a bag, anyways.

Content.Client/Pinpointer/PinpointerSystem.cs
```diff
@@ -18,13 +23,26 @@ public override void Update(float frameTime)
...
+                    _storage.TryGetStorageLocation(uid, out var _, out var _, out var itemStorageLocation);
+                    if (itemStorageLocation.Direction == Direction.East || itemStorageLocation.Direction == Direction.West)
+                    {
+                        angle += Angle.FromDegrees(180);
+                    }
```
EmoGarbage404, line 42: is this due to that bug where some items in the bag don't rotate properly? i feel like it should just be getting the radian angle from the direction and adding it.

EmoGarbage404 (changes requested):

Content.Client/Pinpointer/PinpointerSystem.cs
psykana: Hands have a storage component but no `storage.StoredItems` array `TryGetContainingContainer` crashes.

Content.Client/Pinpointer/PinpointerSystem.cs
psykana: From the bag's standpoint, item can be rotated in 4 directions but pinpointer sprite can only have 2 positions (see pic). So I need to compensate for it only in two cases.

Samuka-C:
I think that is a pretty funny feature that allows for player skill (or lack of there of) to create a funny situation.

ArtisticRoomba:
The description says something along the lines of "ensure upright for best accuracy" so I think it would be pretty funny if it remained :godo:

Content.Client/Pinpointer/PinpointerSystem.cs
EmoGarbage404: hands shouldn't have a storage component? The call does a specific check for the storage container as well. If it's crashing that should probably be looked into via a git issue or something.

superjj18:
> I think that is a pretty funny feature that allows for player skill (or lack of there of) to create a funny situation.

That only works if the behavior is realistic.  It doesn't matter how you rotate a compass, its still gonna point north.

Samuka-C:
> > I think that is a pretty funny feature that allows for player skill (or lack of there of) to create a funny situation.
> 
> That only works if the behavior is realistic.  It doesn't matter how you rotate a compass, its still gonna point north. 

It's not a compass, is a screen

is like a GPS without a auto rotating screen

is going to point to the wrong direction if you rotate it

superjj18:
> It's not a compass, is a screen
> 
> is like a GPS without a auto rotating screen
> 
> is going to point to the wrong direction if you rotate it

My phone has a compass on it's screen, disable auto-rotation and it still points north regardless of how its turned.

Functionally it's far closer in application to a compass than a gps as a result of it always appearing flat on the ground like how you would use an actual compass.  This feels closer to a bug than a feature.

Content.Client/Pinpointer/PinpointerSystem.cs
psykana: I must have used a wrong branch while working on this. You're right, this check is not necessary, it works fine.
Please review again, test fail seems to be unrelated.

walksanatora:
the internal compass is gyroscopic so it always points the right way. the outweshell and screen is not hgence why you gotta keep it upright.

Content.Client/Pinpointer/PinpointerSystem.cs
```diff
@@ -18,13 +22,23 @@ public override void Update(float frameTime)
...
         {
             if (!pinpointer.HasTarget)
                 continue;
             var eye = _eyeManager.CurrentEye;
             var angle = pinpointer.ArrowAngle + eye.Rotation;
 
+            // Pinpointer may be flipped in a backpack etc, adjust for it
+            if (_container.TryGetContainingContainer((uid, null, null), out var container))
+            {
+                _storage.TryGetStorageLocation(uid, out var _, out var _, out var itemStorageLocation);
```
slarticodefast, line 35: `TryGetStorageLocation` should be part of the if condition or it may fail to get the `Direction` below if it is in another type of container.
Also the `TryGetContainingContainer` is not needed as that is already done inside `TryGetStorageLocation`

slarticodefast (changes requested):

superjj18:
> the internal compass is gyroscopic so it always points the right way. the outweshell and screen is not hgence why you gotta keep it upright.

Again I don't think a single person genuinely thinks it's a GPS aside from the people who desperately want to keep this bug in the game because they think it's funny.

It doesn't create routes, it points to a thing.  It's is a pinpointer.

ArtisticRoomba:
Personally I find it exceptionally funny. Keep it upright next time lest you get trolled :godo:

Anno-Midi:
I agree that this is funny enough to be called a feature, and the description is pretty clear.

TheShuEd:

PR #37251, opened 2025-05-07
Make the Syringe gun a science research

JrInventor05:
## About the PR

Added the opportunity to study launcher syringe.

## Why / Balance

Why not?

## Technical details

It is studied in a service branch in  3 tear for 15k points, the cost of creating a launcher syringe of  2 silver, 10 steel, 2 plastic and 1 glass, mini-syringe - 2 steel, 1 plastic and 1 glass. After studying, it becomes available in medical fechfabs and protolate. Fluids are not immediately absorbed by the body from the syringe☺.

## Media

image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: The syringe gun is now a T3 civilian science research.


JrInventor05:
test fail not related

slarticodefast:
Its Player facing name is "Syringe Gun". The proto id is only called that way because it reuses pneumatic cannon and pie launcher code.
Also this will need some more balance discussion than "why not". It can become quite a powerful weapon being able to inject any reagent at a distance, even through hardsuits. If mass-produced it will be extremely good against nukies.

kosticia:
I agree with slarti. Also you changelog missing :cl: symbol.

JrInventor05:
> Its Player facing name is "Syringe Gun". The proto id is only called that way because it reuses pneumatic cannon and pie launcher code. Also this will need some more balance discussion than "why not". It can become quite a powerful weapon being able to inject any reagent at a distance, even through hardsuits. If mass-produced it will be extremely good against nukies.

So, mini syringes introduces reagents slowly and storage of surgery gun is small. If there was a large ammunition load and a high rate of reagent injection, I agree it would not be very good. Name of technology and other I eill change soon.

EmoGarbage404 (changes requested):
Have a few pieces of feedback after some discussion:
- Move the technology from tier 2 up to tier 3.
- Reduce the volume of the mini syringes to 5u down from 15u
- Add some silver into the gun recipe so that it is harder to make.

JrInventor05:
I'll do it soon!

ScarKy0:
t3 sounds kinda sad, with the lowered capacity and stuff it could totally be a t2

TaoNewt:
>Fluids are not immediately absorbed by the body from the syringe☺.

So injected all at once after a delay, or injected x units per second?
the latter sounds better

ScarKy0:
> So injected all at once after a delay, or injected x units per second?
the latter sounds better

2u on impact and 1u every 3s or so

TaoNewt:
> 2u on impact and 1u every 3s or so

I like that a lot, although it would probably need to be re-tuned for the 5u capacity change.

JrInventor05:
> > So injected all at once after a delay, or injected x units per second?
> > the latter sounds better
> 
> 2u on impact and 1u every 3s or so

Every 2s =)

EmoGarbage404 (approved):
i lovee.....

slarticodefast (approved):
We will probably need to make it only hold one syringe at a time and make the printing recipe more expensive for balance.
But we can give it a try and see how it goes.

JrInventor05:
YES!!!

PR #37256, opened 2025-05-07
Mob pull spin fix

Pronana:
## About the PR
Mobs no longer spin as if they had 0 friction when being pulled (because they had 0 friction before oops)

## Why / Balance
Unintended change. Bugfix.

## Technical details
Angular Velocity now has friction applied to it through the same method as Linear Velocity and only for Kinematic Controllers.
This is how it worked before.

Added a version of the Friction method that takes a float instead of a Vector2 for Angular Velocity

## Requirements

## Breaking changes

**Changelog**


EmoGarbage404 (approved):
looks fine but this is a sloth PR in spirit

Issue #37244, opened 2025-05-07
Dev random species spawns you as vox


metalgearsloth:
You immediately get toxicity poisoning and it's incredibly annoying, might be better off reverting unless fixed promptly.

Tayrtahn:
I would argue that spawning as a vox in dev is actually quite useful for making sure that things work for different species. But they absolutely need to get their proper loadout.

This seems to be an issue with how the species override works with the player profile. I just tested, and if you select a vox character on a dev server and then spawn in as a random species, you'll be given the vox loadout regardless of what species you wind up as! The species of the selected profile must be bleeding through somehow.

Issue #37214, opened 2025-05-06
Random species do not have their correct profile on the ID computer.


beck-thompson:
**Reproduction**

Have the random species ccvar set check ID console.

**Screenshots**

![Image](https://github.com/user-attachments/assets/3bf7cf12-e82e-4062-a462-2c5b61b02eb3)

**Additional context**

This probably is a deeper issue than just the station records...

PR #37258, opened 2025-05-07
Move random species selection earlier in player spawning logic

Tayrtahn:
## About the PR

Moved code around in `StationSpawningSystem` and `GameTicker` so that the logic for selecting a random species for the player (used in the dev environment) happens earlier in the process. This allows it to override the player's profile so they spawn with species-appropriate equipment and have the correct information in the station records.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/37214.
Fixes https://github.com/space-wizards/space-station-14/issues/37244.

## Technical details

The current random species selection logic waits until just before spawning the player entity before picking a random species. Notably, this means that the profile used for determining loadout and entered into the station records isn't updated with the random character, but instead uses the player's selected profile. There are probably more bugs caused by this too; for example, I believe that the random selection will ignore any species restrictions on jobs (not used on WizDen, but relevant to downstreams).

To fix these issues, the random species selection logic was moved from `StationSpawningSystem` to `GameTicker.Spawning`. The random profile now replaces the selected profile before being sent out with spawning events and method calls, so it will be used consistently.

## Media

A randomly-selected moth character showing up correctly in the station records:
mothman finally caught on record

A randomly-selected vox character with the appropriate species loadout.
vox with nitrogen

## Requirements

## Breaking changes

`StationSpawningSystem.SpawnPlayerMob` no longer handles the random species selection from the `ic.random_characters` CVar. This logic is now handled in `GameTicker`. If you're directly calling `SpawnPlayerMob` and using `ic.random_characters`, you may need to add additional logic to handle it.

### DIFF
--- a/Content.Server/GameTicking/GameTicker.Spawning.cs
+++ b/Content.Server/GameTicking/GameTicker.Spawning.cs
@@ -8,11 +8,16 @@ using Content.Server.Ghost;
 using Content.Server.Spawners.Components;
 using Content.Server.Speech.Components;
 using Content.Server.Station.Components;
+using Content.Shared.CCVar;
 using Content.Shared.Database;
 using Content.Shared.GameTicking;
+using Content.Shared.Humanoid;
+using Content.Shared.Humanoid.Prototypes;
 using Content.Shared.Mind;
 using Content.Shared.Players;
 using Content.Shared.Preferences;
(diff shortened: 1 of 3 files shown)


EmoGarbage404:
bugs caught with random species cvar: 1

PR #37264, opened 2025-05-07
Craftable Brass Tiles Changed from PlatingBrass turf to Plating

imatsoup:
## About the PR
Changed the baseTurf of the craftable brass floor tiles from PlatingBrass to Plating so that those floor tiles can be placed over top of hull plating tiles.

## Why / Balance
These plates are craftable from your inventory the same as the other floor tiles, but without any indication can only be placed on lattice. My understanding is that these tiles are for the as of now unimplemented clockwork cultist, but it is a tad confusing that these floor tiles work differently from the other tiles. I would like to either have these tiles behave the same as the others in the meantime, or update the item description to indicate that they must be placed on lattice.

## Technical details
Changes to floor.yml for tiles FloorBrassFilled and FloorBrassReebe.

## Requirements

**Changelog**

- tweak: Craftable brass floor tiles can now be placed on top of hull plating in addition to lattice.

### DIFF
--- a/Resources/Prototypes/Tiles/floors.yml
+++ b/Resources/Prototypes/Tiles/floors.yml
@@ -194,7 +194,7 @@
   id: FloorBrassFilled
   name: tiles-brass-floor-filled
   sprite: /Textures/Tiles/Misc/clockwork/clockwork_floor_filled.png
-  baseTurf: PlatingBrass
+  baseTurf: Plating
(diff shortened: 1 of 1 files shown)


imatsoup:
Another possible change could be to just remove the tiles from the crafting menu if they are just placeholder for the clockwork cult's stuff

PR #37265, opened 2025-05-08
Small InventorySystem.Equip Unequip Reason bugfix

Cojoke-dot:
## About the PR

Small bugfix, it was setting the reason for the cancel to attemptEvent.Reason instead of the actual reason, which would be itemAttemptEvent.Reason, it was only like this for unequipping and was correct for equipping
## Why / Balance

bugfix, idk if anything actualy uses this upstream but I noticed it when looking around
## Technical details

changed a variable
## Requirements

**Changelog**

Nothing player facing


slarticodefast (approved):
Yep, looks like a copy paste error.
Thanks for the cleanup!

PR #37271, opened 2025-05-08
Validate Cargo Markets

EmoGarbage404:
## About the PR

Adds a new `CargoMarketPrototype` to validate the `Group` field on cargo orders.

## Why / Balance

needed for a future PR.

## Technical details

Adds a new CargoMarketPrototype. Currently doesn't store any information. 

This main purpose of this is just to add proper validation for things that will use it in the future.

## Requirements

## Breaking changes

Any custom groups specified in cargo order prototypes now need a corresponding `CargoMarketPrototype`.

**Changelog**

no cl no fun

### DIFF
--- a/Content.Shared/Cargo/Prototypes/CargoMarketPrototype.cs
+++ b/Content.Shared/Cargo/Prototypes/CargoMarketPrototype.cs
@@ -10,5 +10,5 @@ public sealed partial class CargoMarketPrototype : IPrototype
 {
     /// <inheritdoc/>
     [IdDataField]
-    public string ID { get; } = default!;
+    public string ID { get; private set; } = default!;
 }
--- a/Content.Shared/Cargo/Prototypes/CargoMarketPrototype.cs
+++ b/Content.Shared/Cargo/Prototypes/CargoMarketPrototype.cs
@@ -1,0 +1,14 @@
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Cargo.Prototypes;
+
+/// <summary>
+/// Defines a "market" that a cargo computer can access and make orders from.
+/// </summary>
+[Prototype]
+public sealed partial class CargoMarketPrototype : IPrototype
+{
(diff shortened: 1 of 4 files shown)


Content.Shared/Cargo/Prototypes/CargoProductPrototype.cs
```diff
@@ -93,6 +93,6 @@ public string Description
         ///     The prototype group of the product. (e.g. Contraband)
         /// </summary>
         [DataField]
-        public string Group { get; private set; } = "market";
+        public ProtoId<CargoMarketPrototype> Group { get; private set; } = "market";
```
slarticodefast, line 96: Might need to be changed to a list of markets in the future, but that can be done whenever we need it.

Content.Shared/Cargo/Prototypes/CargoMarketPrototype.cs
```diff
@@ -0,0 +1,14 @@
...
+
+/// <summary>
+/// Defines a "market" that a cargo computer can access and make orders from.
+/// </summary>
+[Prototype]
+public sealed partial class CargoMarketPrototype : IPrototype
+{
+    /// <inheritdoc/>
+    [IdDataField]
+    public string ID { get; } = default!;
```
slarticodefast, line 13: Any reason you are not using `{ get; private set; }` like all the other prototypes are doing?
No clue why a setter would be needed though.

EmoGarbage404: This is just the default prototype template in rider, but i can add it since making it readonly is good.

Tayrtahn: Notably, not having a setter (private or otherwise) on a DataField property is actually an error - but there's currently a bug where prototypes aren't checked by the data definition analyzer (see https://github.com/space-wizards/RobustToolbox/pull/5699)

PR #37272, opened 2025-05-08
Prevent AI from interacting with objects on other grids

B-Kirill:
## About the PR
AI should not be able to interact with objects on other grids, but the interaction menu still opened via Alt+ click.

## Why / Balance
Players could abuse this and bolt the airlocks on the shuttles.

## Media
_No need_

## Requirements

**Changelog**
:cl:
- fix: AI can no longer interact with objects on other grids using hotkeys.

### DIFF
--- a/Content.Shared/Silicons/StationAi/SharedStationAiSystem.Held.cs
+++ b/Content.Shared/Silicons/StationAi/SharedStationAiSystem.Held.cs
@@ -160,6 +160,10 @@ public abstract partial class SharedStationAiSystem
         if (!_uiSystem.HasUi(args.Target, AiUi.Key))
             return;
 
+        // No cross-grid
+        if (Transform(args.Target).GridUid != Transform(args.User).GridUid)
+            return;
+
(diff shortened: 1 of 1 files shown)


EmoGarbage404 (changes requested):
should move this check (and honestly the hasUi check) to _after_ the first if block. that way, you run the least expensive operations first and quit out early.

B-Kirill:
I want to add another similar fix to this PR, it may take some time

B-Kirill:
Okay, that's it for sure now

B-Kirill:
@EmoGarbage404 
Can you please check this PR after my changes, after all it is a bug fix (writing just in case you have lost this PR)?

B-Kirill:
> I personally agree with the first change but not the second. you are a AI wired to the station. you should not need to be able to physically "see" things to manipulate them aslong as they have power and are connected to the station.

Yes, your logic makes sense. But in the current implementation, AI shouldn't do that, so that's what I'm fixing.

PR #37273, opened 2025-05-08
Add groups to bounties

EmoGarbage404:
## About the PR

Adds a new group field to cargo bounty, adds a check to make sure only bounties of a certain group are picked for the regular station bounties.

## Why / Balance

Needed for future PR.

## Technical details

Adds a new `CargoBountyGroupPrototype` and a field on `CargoBountyPrototype` and `StationCargoBountyDatabaseComponent`.

## Requirements

## Breaking changes

**Changelog**

no cl no fun

### DIFF
--- a/Content.Server/Cargo/Components/StationCargoBountyDatabaseComponent.cs
+++ b/Content.Server/Cargo/Components/StationCargoBountyDatabaseComponent.cs
@@ -1,4 +1,6 @@
 using Content.Shared.Cargo;
+using Content.Shared.Cargo.Prototypes;
+using Robust.Shared.Prototypes;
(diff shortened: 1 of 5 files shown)


slarticodefast:
I think this should just be part of the coming PR. Because if we don't merge that one for some reason we end up with dead, unused code.
Also it would be worth considering turning bounties into entities. That way we can select them from an entity table instead of using linq. And I previously tried to make the bounty request amount random, but that is hard to do with a prototype that is indexed in multiple places, because we would have to store the amount that was chosed somewhere.

EmoGarbage404:
Sure

Issue #36403, opened 2025-04-09
`HTNSystem` breaks out if more than 128 NPCs need pathing


ArtisticRoomba:
## Description
Right now the HTNSystem refuses to process more than 128 NPCs. In the sense that it gives up and doesn't process any of them.

https://github.com/space-wizards/space-station-14/blob/cc167b0c2f3be40c3a98dc75b775b4d521f3fb3e/Content.Server/NPC/HTN/HTNSystem.cs#L178-L187

Some points made on the discord condensed, not in any real order.
- If we're past 128 mobs, then add the excess mobs to a list and move the processing over to the next tick. Then the next batch of 128 mobs can be processed on that tick, and any excess will be left over for the next one.
- The actual logic in `Update` states that even if the NPC is waiting for a plan, it updates the count and counts towards the 128.

**Reproduction**
Just spawn like a bunch of mothroach crates or something.

**Screenshots**
![Image](https://github.com/user-attachments/assets/98246888-2fe2-48eb-a424-036e99c9d505)

EmoGarbage404:
Fixed by #37056

PR #37275, opened 2025-05-08
Add hivebots

Flareguy:
## About the PR
Adds hivebots (ported from SS13.) Though weak individually, their strength in numbers is not to be underestimated.
These are intended to go on to salvage wrecks later on. The only thing this PR does is add the prototypes for them.

Worth noting is that the ``MobAtmosExposed`` parenting has been moved from BaseSimpleMob to SimpleSpaceMobBase, since it was getting in the way of hivebots being robotic. I do not know if this affects the functionality of any existing mobs, but it probably shouldn't, since things that want to be atmos exposed are probably going to be parenting one of the other 2 simplemob prototypes.

## Why / Balance
Most of the current salvage mobs are incredibly similar-feeling and boring, both thematically and visually. Hivebots should help rectify this by a lot, since swarming nature & projectiles should help offer interesting combat-puzzle-like gameplay.

## Technical details
N/A

## Media
[3Ybo8PIv67.webm](https://github.com/user-attachments/assets/ef25dbba-33ce-4792-a3f1-c2d17ca9229d)

## Requirements

## Breaking changes
N/A

**Changelog**
no cl no fun


Resources/Prototypes/Entities/Mobs/base.yml
```diff
@@ -238,3 +238,32 @@
     bloodlossHealDamage:
       types:
         Bloodloss: -1
+
+# Used for robotic mobs, like hivebots and cyborgs
+- type: entity
+  save: false
+  id: MobRobotic
```
EmoGarbage404, line 245: this really needs to be combined or built into `MobSiliconBase`. The other one looks a bit too specialized but this is really weird parenting

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
```diff
@@ -0,0 +1,117 @@
...
+      types:
+        Slash: 6 # Pretty whatever on their own, but their strength is in numbers
+  - type: Tag
+    tags:
+      - DoorBumpOpener
+  - type: MovementSpeedModifier
+    baseWalkSpeed: 1.5
+    baseSprintSpeed: 2.5
+  - type: PointLight
+    radius: 1.0
```
EmoGarbage404, line 57: you should set `castShadows` to false here just because these are meant to be swarms and can potentially spawn a fair number of them

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
```diff
@@ -0,0 +1,117 @@
...
+  name: hivebot
+  suffix: Ranged
+  description: Agitating, mechanized pests. This one has a gun.
+  components:
+  - type: Sprite
+    drawdepth: Mobs
+    sprite: Mobs/Aliens/hivebot.rsi
+    layers:
+    - map: [ "enum.DamageStateVisualLayers.Base" ]
+      state: ranged_attack
```
EmoGarbage404, line 75: imagine if you split up the visor into an unshaded layer... would be cool

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
```diff
@@ -0,0 +1,117 @@
...
+    rootTask:
+      task: SimpleRangedHostileCompound
+  - type: RechargeBasicEntityAmmo
+    rechargeCooldown: 0.50
+  - type: BasicEntityAmmoProvider
+    proto: BulletHivebot
+    capacity: 1
+    count: 1
+  - type: Gun
+    fireRate: 0.50
```
EmoGarbage404, line 86: not sure if this is intended but this means the fire rate is 0.5 _bullets per second_. You have the recharge set to half a second so im just verifying you actually meant to have this speed instead of 0.5 seconds inbetween shots (which would be a fireRate of 2)

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
```diff
@@ -0,0 +1,117 @@
...
+    rootTask:
+      task: SimpleHostileCompound
+    blackboard:
+      NavSmash: !type:Bool
+        true
+  - type: NpcFactionMember
+    factions:
+    - SimpleHostile
+  - type: Sprite
+    drawdepth: Mobs
```
EmoGarbage404, line 18:
```suggestion
```
probably gets inherited from something up the tree

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
```diff
@@ -0,0 +1,117 @@
...
+- type: entity
+  parent: MobHivebot
+  id: MobHivebotRanged
+  name: hivebot
+  suffix: Ranged
+  description: Agitating, mechanized pests. This one has a gun.
+  components:
+  - type: Sprite
+    drawdepth: Mobs
+    sprite: Mobs/Aliens/hivebot.rsi
```
EmoGarbage404, line 72:
```suggestion
```

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
```diff
@@ -0,0 +1,117 @@
...
+
+- type: entity
+  parent: MobHivebotRanged
+  id: MobHivebotStrong
+  name: strong hivebot
+  description: Agitating, mechanized pests. This one looks much stronger then normal.
+  components:
+  - type: Sprite
+    drawdepth: Mobs
+    sprite: Mobs/Aliens/hivebot.rsi
```
EmoGarbage404, line 104:
```suggestion
```

EmoGarbage404 (changes requested):

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
Flareguy: This was in fact intentional. I didn't want them to shoot very fast since I knew that it would probably be really hard to manage in swarms with the PKA - though, obviously, I'm more then open to buffing them later if they're too weak

Resources/Prototypes/Entities/Mobs/NPCs/hivebot.yml
Flareguy: I wanted to opt for a pointlight instead, since IMO unshaded layers tend to look kind of bad - especially considering the hivebot's lighting is intended to be pretty soft.
Though, the pointlight doesn't actually seem to work for some reason, and I can't really throw my head against the wall at that without hot reloading, so I guess that could be a reason to not add it. But I'd rather just opt to not do it for the afermentioned reasons.

EmoGarbage404 (approved):

EmoGarbage404:
> It's probably worth keeping #36403 in mind with this addition. I don't think it should be a blocker, but it might increase the priority of that issue.

i actually fixed that bug with #37056 and forgot to close the issue

Tayrtahn:
Oh, well that works!

DogZeroX:
I say give them pistol shooting sound, seems weird they shoot one bullet and has the smg one :3

Flareguy:
> I say give them pistol shooting sound, seems weird they shoot one bullet and has the smg one :3

I actually had the pistol sound before & changed it because it felt like it didn't fit as well as the SMG one. Comparatively, the SMG firing sound is a lot punchier and has less... uh... sustain, I guess you'd call it? So it was both less annoying and more threatening-sounding, both qualities that seemed desirable for a combat mob like this one.

EmoGarbage404:
new test fails are indeed real

ArtisticRoomba (approved):
merging when tests pass

ArtisticRoomba:
yeah these are SUPER FUN to encounter

PR #37283, opened 2025-05-08
Gyroscope throttling

Krunklehorn:
## About the PR
Allows players to reduce a thruster's influence to one of four predefined modes. Currently only enabled on gyroscopes.
Maximum: 100%
High: 50%
Medium: 20%
Low: 5%

## Why / Balance
Makes it easier to fly shittles since gyroscopes are so powerful by default.

## Technical details
Same code, sound and examine text from the electric grill, along with a fix for `IsOn` behavior if the `TryGet<ShuttleComponent>` fails.

## Media
https://github.com/user-attachments/assets/61cb7db1-9f7e-45d3-85b9-f4612a332d68

## Requirements

## Breaking changes
`ThrusterSystem:Enable/DisableThruster()` no longer set IsOn if the `TryGet<ShuttleComponent>` fails. Make sure you are not relying on this behavior.
`ThrusterComponent` now controls the power output of the `ApcPowerReceiver`. Copy the `powerLoad` field to `basePowerLoad` for anything inheriting `BaseThruster`. See `thrusters.yml`.
.
**Changelog**
:cl:
- add: Gyroscopes can now be alt-clicked to reduce their influence. Great for personal shuttles!

### DIFF
--- a/Content.Server/Shuttles/Components/ThrusterComponent.cs
+++ b/Content.Server/Shuttles/Components/ThrusterComponent.cs
@@ -1,6 +1,7 @@
 using System.Numerics;
 using Content.Server.Shuttles.Systems;
 using Content.Shared.Damage;
+using Robust.Shared.Audio;
 using Robust.Shared.GameStates;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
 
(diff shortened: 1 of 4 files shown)


slarticodefast:
The verb needs to be predicted, but you will have to move the component to shared for that as well.

EmoGarbage404 (changes requested):
flight controls like this should really be in the shuttle console UI. Having the gyroscope right next to the console is unfeasible for a lot of shuttles and having to exit out of the ui to change the rotational speed is really clunky.

Alpha-Two:
> flight controls like this should really be in the shuttle console UI. Having the gyroscope right next to the console is unfeasible for a lot of shuttles and having to exit out of the ui to change the rotational speed is really clunky.

I wouldn't expect this to be changed via the flight UI necessarily, as it should be something that's set-and-forget as opposed to something that you'd change on the fly. Especially since if you're adding mass to a ship, you'd probably be messing around in the engine area anyway.

ilovehans10:
If you ask me it would make more sense to limit the maximum rpm of the gyro instead of an arbitrary percentage. That way muscle memory could be developed for piloting even when using shittles of varying sizes.

EmoGarbage404:
> I wouldn't expect this to be changed via the flight UI necessarily, as it should be something that's set-and-forget as opposed to something that you'd change on the fly. Especially since if you're adding mass to a ship, you'd probably be messing around in the engine area anyway.

this is literally just adjusting the turning force on the shuttle. it makes no sense to not just be able to finely control the output from the UI.

EmoGarbage404:
After some more proper maint discussion:

- small shuttles having absurd turning is an issue and something we'd like to see fixed
- the proposed solution has notable shortcomings with regards to usability
- ideally there would be a better shuttle UI that allows more precise control of rotation
- in lieu of that, it would be better to scale gyroscopes with the grid size (up to a certain point) so that the rotational force is similar between both big and small shuttles.

PR #37288, opened 2025-05-08
Add colorful implants

qrwas:
## About the PR
Implants in implanters will now have their own color.
The color was chosen based on some logic, taking into account "what does the implant do?" and "what color matches that?"
To make this work, I created a new system `ItemChangeLayerColorSystem` which operates on two components: `ChangeLayersColorComponent` and `ItemLayersColorComponent`

## Why / Balance
A small life-improving change that will allow you to visually determine which implant is in front of you
## Technical details
Added a new system `ItemChangeLayerColorSystem` which operates on two components: `ChangeLayersColorComponent` and `ItemLayersColorComponent`. 
- `ChangeLayersColorComponent` - contains the layers and a whitelist for which the color should be changed.
- `ItemLayersColorComponent` - contains the color that the system will take
## Requirements

- [x] Add a new component to store color
- [x] Add a component to define layers that will change color
- [x] Determine colors for existing implants (I would love to hear suggestions regarding the color of certain implants.)
- [x] Add appropriate components to implants and implanters

**Changelog**
:cl: qrwas
- tweak: Each implant has its own color while in the implanter


qrwas:
I am currently having a problem with reconnecting (new connection). Implants removed by the implanter have a default white color, and only change it after re-extraction. I tried adding new checks, but that didn't work.
Also, to put it mildly, the code may not be the best (at least because it doesn't work properly yet), although it performs almost all the functions assigned to it

slarticodefast:
Can't you just use `ItemMapperComponent` and a generic visualizer? Should be possible completely in yaml rather than setting the layers manually.

qrwas:
The color changes in the spawn menu, but when spawning - this color is not there. Cool.

But now there is synchronization with removed implants.

image

🫠

slarticodefast:
Before you continue doing this in C#, I would really recomming trying to reuse the existing code for this and just do it in yaml.

qrwas:
> Before you continue doing this in C#, I would really recomming trying to reuse the existing code for this and just do it in yaml.

I saw that GenericVisual supports changing color, but there is no example of how it works.
image

Or Are you suggesting making layers for each implant and turning them on depending on which one is currently in the implant? That's the worst possible option.

slarticodefast:
True, a new layer for each implant would be a little messy.
But the generic visualizer should be able to handle this without you having to write a custom visualizer. Like the code comment above says, in yaml you simply need to write down the string representation of the color value you are using to set the color of the greyscale layer. This still means you would have to list every color value for the implants.
So a new visualizer for this would be useful, but I would make it generic rather than tied explicitly to implants. Something like itemmapper, but instead of toggling a layer it sets its color.

qrwas:
> True, a new layer for each implant would be a little messy. But the generic visualizer should be able to handle this without you having to write a custom visualizer. Like the code comment above says, in yaml you simply need to write down the string representation of the color value you are using to set the color of the greyscale layer. This still means you would have to list every color value for the implants. So a new visualizer for this would be useful, but I would make it generic rather than tied explicitly to implants. Something like itemmapper, but instead of toggling a layer it sets its color.

making it universal sounds interesting, but I don't yet imagine what it should look like. maybe as it works now, except for adding a field of the required layer that will change the color, and another component that will contain the color to which it needs to be changed. And globally the changes will not be significant in the implementation that I made, but it will be universal. But until it works fully, it's too early to think about unification

slarticodefast:
> making it universal sounds interesting, but I don't yet imagine what it should look like. maybe as it works now, except for adding a field of the required layer that will change the color, and another component that will contain the color to which it needs to be changed. And globally the changes will not be significant in the implementation that I made, but it will be universal. But until it works fully, it's too early to think about unification

Sounds good, but I would probably make it mostly identical to ItemMapper, so that it uses a whitelist to set the selected layer to a specified color, instead of splitting it into two components.

qrwas:
> > making it universal sounds interesting, but I don't yet imagine what it should look like. maybe as it works now, except for adding a field of the required layer that will change the color, and another component that will contain the color to which it needs to be changed. And globally the changes will not be significant in the implementation that I made, but it will be universal. But until it works fully, it's too early to think about unification
> 
> Sounds good, but I would probably make it mostly identical to ItemMapper, so that it uses a whitelist to set the selected layer to a specified color, instead of splitting it into two components.

whitelisting means constantly updating old files, which doesn't scale very well. and where to get the unique color is also a question, except to define it in the same whitelist, instead of adding the main component once, and then adding new items with the secondary component, never returning to the main component 🤔

Tayrtahn:
Rather than adding a component to store the color, it looks like `AppearanceComponent` can have values set in YAML:

https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Shared%2FGameObjects%2FComponents%2FAppearance%2FAppearanceComponent.cs#L40-L48

Maybe that could be used?

qrwas:
> Rather than adding a component to store the color, it looks like `AppearanceComponent` can have values set in YAML:
> 
> https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Shared%2FGameObjects%2FComponents%2FAppearance%2FAppearanceComponent.cs#L40-L48
> 
> Maybe that could be used to attach colors to the implant definitions?

you can see the changes. I already did it :) Comments above are a discussion on how this can be unified in the future

qrwas:
@slarticodefast It finally worked. Should I make universal components and systems in this PR, or in the future?

qrwas:
Although it will be a bit problematic to do this now, given how the starting check of a specific slot (`ImplanterComponent.Implant`) currently works.

qrwas:
I thought that doing it right away was the best option.

qrwas:
i hate Atmos.EntitySystems 🫡

Content.Client/Storage/Systems/ItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (component.SpriteLayers.Count == 0)
+            {
+                InitLayers((uid, component, spriteComponent, args.Component));
+            }
+
+            UpdateLayers((uid, component, spriteComponent, args.Component));
+        }
+    }
+
+    private void InitLayers((EntityUid uid, ChangeLayersColorComponent component, SpriteComponent spriteComponent, AppearanceComponent Component) ent)
```
Tayrtahn, line 31: Consider using `Entity<T1, T2, T3>` instead of a tuple here. If a tuple is needed for some reason, use more distinct and descriptive parameter names.

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        foreach (var (layerName, val) in component.MapLayers)
+        {
+            val.Layer = layerName;
+        }
+
+        if (EntityManager.TryGetComponent(uid, out AppearanceComponent? appearanceComponent))
+        {
+            var dictionary = new Dictionary<string, Color>();
+            foreach(var key in component.MapLayers.Keys)
```
Tayrtahn, line 72:
```suggestion
            foreach (var key in component.MapLayers.Keys)
```

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
+using Content.Shared.Implants;
+using Content.Shared.Implants.Components;
+using Content.Shared.Storage.Components;
+using Content.Shared.Whitelist;
+using Robust.Shared.Containers;
+using Robust.Shared.Prototypes;
+using System.Linq;
```
Tayrtahn, line 7:
```suggestion
using Content.Shared.Storage.Components;
using Content.Shared.Whitelist;
using Robust.Shared.Containers;
using Robust.Shared.Prototypes;
```

Content.Server/Storage/EntitySystems/ItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,8 @@
+using Content.Shared.Storage.EntitySystems;
+using JetBrains.Annotations;
+
+namespace Content.Server.Storage.EntitySystems;
+
+/// <inheritdoc/>
```
Tayrtahn, line 6: There's no xmldoc in `SharedItemChangeLayerColorSystem` to inherit from here.

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+
+namespace Content.Shared.Storage.EntitySystems;
+
+public abstract class SharedItemChangeLayerColorSystem : EntitySystem
+{
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
+    [Dependency] private readonly IPrototypeManager _proto = default!;
+    [Dependency] private readonly SharedContainerSystem _container = default!;
```
Tayrtahn, line 15:
```suggestion
```
Unused.

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+        if (!TryComp<ItemLayersColorComponent>(ent, out var itemLayerColor))
+            return;
+
+        UpdateAppearance(uid, ent, itemLayerColor, comp, appearance);
+    }
+
+    private void UpdateAppearance(EntityUid parent, EntityUid item, ItemLayersColorComponent itemLayerColor, ChangeLayersColorComponent comp, AppearanceComponent appearance)
+    {
+        var dict = new Dictionary<string, Color>();
+        if(!TryGetLayer(item, comp, out var layer))
```
Tayrtahn, line 55:
```suggestion
        if (!TryGetLayer(item, comp, out var layer))
```

Content.Shared/Storage/Components/ChangeLayersColorComponent.cs
```diff
@@ -0,0 +1,53 @@
...
+    ///     Can be initialized if already existing on the sprite, or inferred automatically
+    /// </summary>
+    [DataField("spriteLayers")]
+    public List<string> SpriteLayers = new();
+}
+
+[Serializable, NetSerializable]
+public sealed class ColorLayerData : ICloneable
+{
+    public readonly LayerColor[] QueuedEntities;
```
Tayrtahn, line 29: Why is this array of layername-to-color mappings called `QueuedEntities`?

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+            return;
+
+        dict.Add(layer, itemLayerColor.Color);
+        _appearance.SetData(parent, StorageMapVisuals.LayerChanged, new ColorLayerData(dict), appearance);
```
Tayrtahn, line 59: I'm pretty sure this should use its own Enum instead of the one from `StorageMapVisuals`.

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        if (EntityManager.TryGetComponent(uid, out AppearanceComponent? appearanceComponent))
+        {
+            var dictionary = new Dictionary<string, Color>();
+            foreach(var key in component.MapLayers.Keys)
+            {
+                dictionary.Add(key, Color.White);
+            }
+            _appearance.SetData(uid, StorageMapVisuals.InitLayers, new ColorLayerData(dictionary), appearanceComponent);
```
Tayrtahn, line 76: Same here.

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    private bool TryGetLayer(EntityUid ent, ChangeLayersColorComponent itemLayerColor, out string showLayer)
+    {
+        foreach (var mapLayerData in itemLayerColor.MapLayers.Values)
+        {
+            var count = _whitelistSystem.IsWhitelistPassOrNull(mapLayerData.Whitelist, ent);
+            if (count)
```
Tayrtahn, line 85:
```suggestion
            if (_whitelistSystem.IsWhitelistPassOrNull(mapLayerData.Whitelist, ent))
```

Content.Shared/Storage/Components/ItemLayersColorComponent.cs
```diff
@@ -0,0 +1,18 @@
+using Robust.Shared.Animations;
+
+namespace Content.Shared.Implants;
+
+[RegisterComponent]
+public sealed partial class ItemLayersColorComponent : Component
```
Tayrtahn, line 6: Needs xmldocs.

Content.Shared/Storage/Components/ItemLayersColorComponent.cs
```diff
@@ -0,0 +1,18 @@
...
+    [DataField("color")]
+    private Color _color = Color.Green;
+
+    [Animatable]
+    [ViewVariables(VVAccess.ReadWrite)]
+    public Color Color
+    {
+        get => _color;
+        set => _color = value;
+    }
```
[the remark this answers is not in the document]
Tayrtahn: Either way, also add xmldocs.

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Linq;
+
+namespace Content.Shared.Storage.EntitySystems;
+
+public abstract class SharedItemChangeLayerColorSystem : EntitySystem
+{
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
+    [Dependency] private readonly IPrototypeManager _proto = default!;
+    [Dependency] private readonly SharedContainerSystem _container = default!;
+    [Dependency] private readonly EntityWhitelistSystem _whitelistSystem = default!;
```
Tayrtahn, line 16:
```suggestion
    [Dependency] private readonly EntityWhitelistSystem _whitelistSystem = default!;
    
```

Tayrtahn (changes requested):

Content.Shared/Storage/Components/ChangeLayersColorComponent.cs
qrwas: Changed the name to LayersColors

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
qrwas: I don't think this is really necessary. The system resides in Storage, and works in a similar way to ItemMapper. This will simply create a new Enum that will perform exactly the same function.

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        var dict = new Dictionary<string, Color>();
+        if(!TryGetLayer(item, comp, out var layer))
+            return;
+
+        dict.Add(layer, itemLayerColor.Color);
+        _appearance.SetData(parent, StorageMapVisuals.LayerChanged, new ColorLayerData(dict), appearance);
+    }
+
+    private void OnComponentInit(EntityUid uid, ChangeLayersColorComponent component, ref ComponentInit args)
```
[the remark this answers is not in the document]
qrwas: Just change the event to ComponentStartup?

Content.Shared/Storage/EntitySystems/SharedItemChangeLayerColorSystem.cs
qrwas: I understood why this is needed... appereanceComponent stores the date in the middle as a key and date, and if both this new component and ItemMapper are used, it will simply break.

qrwas:
ready. In theory, all problems are solved

qrwas:
It was the third Sunday of the review 🫠

qrwas:
pupupu...

qrwas:
🤨

qrwas:
please check this small QoL pr 🫡

qrwas:
😴

lzk228:
i don't know if it was already said, but i think it's bad that child color is tied to the implanter, not the implant

qrwas:
> i don't know if it was already said, but i think it's bad that child color is tied to the implanter, not the implant

The color is now specified on the implant itself in the subdermal_implants.yml file, and the visual is applied to the implanter, because as such, the implant does not exist materially.

qrwas:
so you can use the implant's essence and get its ItemLayersColor somewhere else. Accordingly, when the implant is removed from the body, the implanter displays the color of the removed implant, not the one that was in it to begin with

qrwas:
happy!

qrwas:
@Tayrtahn Could you check my pr?

Resources/Textures/Objects/Specific/Medical/implanter.rsi/implanter1.png
iaada: Credit yourself for making these greyscale in the json.

Resources/Textures/Objects/Specific/Medical/syndi_implanter.rsi/implanter1.png
iaada: This too.

qrwas:
Now, it not my fall test

PR #35406, opened 2025-02-22
THE WIZARD

keronshb:
## About the PR

https://youtu.be/QL8ASZOvvS8

The Wizard is finally here. The antagonist from SS13 that everyone knows and loves (doc coming soon™).
Something special about the Wizard, you never know when one will show up!!

It has a chance of rolling as its own mode, as a sub game mode for other modes, AND as a midround ghost role!

EXTRA SPECIAL THANKS TO CHRIS REMO for allowing us to use The Wizard as lobby music!

Would also like to thank absolutely everyone who helped make Wizard a reality with good quality code.
@slarticodefast @ScarKy0 @ActiveMammmoth 
(If I forgot anyone please let me know and I will tag you)

## Why / Balance

We're in desperate needs of antags and SS13 parity.
While not every SS13 spell is in, they're planned for the Wizard! Extras are also planned so keep an eye out.

The Wizard is """Balanced""" around three key concepts.
1. A Wizard Did It
2. Extremely rare antag
3. A lot more limited spending than a Traitor or a Nukie

## Technical details

Free Objective ECS & Base Objective was added. Free Objective ECS is as simple as it sounds, it automatically greentexts that objective.

Added a SubWizard gamerule with a 0.05 probability

Added a Wizard Event for midround spawning with 1 weight.

Added a Wizard Guide.

## Media

image

image

image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

Renamed midround.yml to subgamemodes.yml

**Changelog**

:cl:
- add: Added the Wizard Antagonist!

PR #32782, opened 2024-10-13
Add Diona rooting

SlamBamActionman:
## About the PR

This PR adds a new species-unique feature to the Diona that allows them to get a built-in toggleable "magboots", at the cost of absorbing liquids on the ground and a slowdown. This feature is called "Rooting", simulating the diona steadying themselves using their root-like leg appendages.

## Why / Balance

The Diona have long been seen as a sub-optimal species due to their main downside: not being able to wear shoes. This downside is interesting but due to how many mechanics of the game are centered around shoes it has made the Diona adifficult species to justify for certain jobs and scenarios. Among these include Janitorial, Salvage and Nukies, as these are roles that rely on footwear-specific items. 

Giving Diona built-in magboots have been suggested before in #24596, however this was an always-on feature and it was suggested the Diona would have a slower base speed as result. 

Instead this PR implements this as a toggle in the form of the "Rooting" action and status effect. While rooted, Diona move at 80% speed (same as magboots) and have a similar effect of preventing slipping and floating in no-grav. The one exception is slipping caused by injuring items on the ground, such as glass shards and D4s. This is due to the rooting not being equivalent to wearing shoes, and to prevent a diona walking over a stack of multiple glass shards and immediately dying from burst damage.

While rooted, the affected entity will also attempt to transfer reagents from puddles to the entity's chemstream. This happens at a rate of 0.75u per second. In effect, this means that a diona can passively "drink" from puddles. This is largely meant as a downside; while some chemicals on the floor may be neutral or even helpful, being rooted over e.g. welding fuel would make the diona take damage over time.

## Technical details

Most of the absorption code in `RootableSystem` is set up in a similar way to how `SmokeSystem` handles it, though centered from the rootable entity. The no-slip and no-grav code is fairly boilerplate event handling.

`SlipAttemptEvent` has been given an optional entity property that includes the entity causing the slip (used to determine if it has `DamageUserOnTriggerComponent`). 

## Media

https://github.com/user-attachments/assets/65a1ea7d-1e75-4a79-bc15-92c31cbda4c6

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- add: Diona can now root, keeping them from slipping but absorbing fluids from the ground.

PR #37289, opened 2025-05-08
Put those dawgs away! [Let Diona Wear Shoes]

EmoGarbage404:
## About the PR

Something something sins of the father.

Gives diona the ability to wear shoes
Edits the feet and leg textures slightly so that the feet don't clip over the pants and look bad with shoes.

## Why / Balance

Being able to wear shoes has always been and continues to be a crippling downside that makes diona unable to do various jobs. The number of jobs that have some item of footwear as a required element is far too high for one species to arbitrarily lose out on the ability.

Additionally, there's just no reason for them not to have it? It's not like Diona are meant to be some ultra-hardcore species that is supposed to suck at every job. And if they were, there are more interesting ways to do it. Just being unable to wear essential equipment is bad.

## Technical details

n/a

## Media

image

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Diona have figured out how to wear shoes.


SlimmSlamm:
I disagree with this change. Choosing a race should have plenty of pros and cons and this will make them have even less of a negative. Races should never be pure cosmetic and allowing to have such diverse gameplay between all of them should always what we aim for in my opinion and this is just two steps in the wrong direction.

SlimmSlamm:
I forgot to mention but there should be a PR currently open that allows themselves to root into the ground making it impossible to be pushed or fall over. I think there are plenty of ways we could compensate for Diona not being able to wear shoes. An idea I have always had but never talked about is allowing Diona to be genetically mutated through Botany which would allow themselves to get unique abilities every round. Unfortunately there is no species design document so I can’t really do anything with the concept yet.

muburu:
what about https://github.com/space-wizards/space-station-14/pull/32782????

EmoGarbage404:
> what about #32782 ????

my suggestion would be just to only allow rooting when no shoes are being worn.

slarticodefast:
Yeah, I would prefer to see https://github.com/space-wizards/space-station-14/pull/32782 merged. They are kinda exclusive because shoes would make the puddle absorption feature when rooted useless.

SlimmSlamm:
> > I disagree with this change. Choosing a race should have plenty of pros and cons and this will make them have even less of a negative. Races should never be pure cosmetic and allowing to have such diverse gameplay between all of them should always what we aim for in my opinion and this is just two steps in the wrong direction.
> 
> absolutely true, but a player should never feel forced to disable entire antagonists/jobs purely because of their chosen species, i simply don't see what positive comes from a diona nukie slipping on a lube puddle and instantly dying to something outside of his control, or a diona player choosing not to take a space ninja role because he wouldnt have speedboots/noslips etc etc

I agree with this but I’d also like to state it’s only like this because of the early access playtest and the lack of a design document for species. I believe the design of Diona was not a well implemented on its original release but I don’t think we should be going backwards here either. I would rather see the other rooting PR merged since it’s a very awesome racial ability

DrWhoFan13:
I'll be perfectly honest, this just removes some of what makes Diona _unique._

Nukesthestation:
> > what about #32782 ????
> 
> my suggestion would be just to only allow rooting when no shoes are being worn.

I feel like it would still get shadowed by this pr, magboots, or any other method of no slips would be objectively better do to the fact that you wouldn't run the risk of getting poison or something of the like in your system, 
I will admit that I myself have thought that Diona couldn't wear shoes, but I just took it as a cultural thing

ps3moira:
I think trying to solve an intended feature defeats the purpose of having species that offer different playstyles. With the diona's arguably pretty powerful ability to survive death when being gibbed, being unable to wear shoes in my opinion is fairly balanced.

EthanQix:
I think implementing #32782 would be a much more interesting way to implement this, instead of making Dionas more generic.

SlimmSlamm:
> Could you list off a few jobs that absolutely need shoes and why?

Salvage, Nukies, Janitor. All of these are pretty self explainable. I’m however still against this PR in favor of the root ability.

superjj18:
I actually agree with this, shoes are not an optional item.  Diona is one of the weakest races with the only exception being in zombies they are essentially suicide bombers who can infect pretty much anyone by gibbing themselves into 3 mini-zombies(freakin lame from a survivor viewpoint by the way with the new infection stat system).

the gimmick of not being able to walk on glass is stupid(probably unintended too, who when designing this thought "and if they step on glass they should get instaslipped", probably no one), especially in the context of a damn tree.  Diona is not a strong race, arguably the weakest in the whole game, it wont be the end of the world to give them shoes and actually make them slightly closer to viable.

Diona's main advantage is literally that they are superior at dying, nuff said.

superjj18:
> I'll be perfectly honest, this just removes some of what makes Diona _unique._

If not wearing shoes is the most relevant unique feature of Diona, that is a damning condemnation of Diona's design.

walksanatora:
there is a open PR for diona rooting.

Nimfar11:
Not being able to wear shoes may not be the strongest gameplay design decision, but it's what makes Diona special and sets her apart from the usual standard. Makes her stand out throughout the game, not the difficulty of dying, which doesn't happen very often. Even little things like that create its own image of a race. I think the best solution is to make each species special, rather than fitting them all to a common standard.

TaoNewt:
> An idea I have always had but never talked about is allowing Diona to be genetically mutated through Botany which would allow themselves to get unique abilities every round. Unfortunately there is no species design document so I can’t really do anything with the concept yet.

I think myra is currently writing the design doc, so it's a really good time to make a draft. Also I really want to see a full write-up of your idea because it sounds really interesting and I wonder how it would work alongside genetics when that gets made.

Pixeltheaertist:
I agree with this, I mean, if a race's main gimmick is that it can't wear shoes... why? Shoes are incredibly useful and often times non-optional. Magboots, noslips, speed boots, etc are all very very important and make some roles harder or damn near impossible. Races should be balanced in ways that doesn't roadblock entire parts of the game. Diona are by far the weakest species in the game at the moment, this PR just makes them slightly more enjoyable to play with no real major detriment. They still have their gimmicks with robust harvest, and we can likely still add rooting, and the nymphs. Not having shoes is a small part of Diona design which makes them significantly more unfun to play as because it's straight up unfair at times.

ilovehans10:
Next thing people will be calling for Vox to no longer be poisoned by oxygen because it makes the species harder to play as...

Well-made species are made unique through a combination of many minor variances that differentiate them from the other species. Removing one of these features won't make them game unplayable, but it may change a unique species to just another reskin of human (looking at you dwarf). If Dionna really NEED boots, then making species-specific root covers for them could be interesting, and won't make them more generic.

End of the day, having design documents for species would help with this, as intended unique features could be outlined so that it is clear just how many or few factors are actually contributing to a species' uniqueness. This isn't my first time asking for a maintainer PR to have a design document back it, so I know how this will go ([promise of a design doc followed by no design doc + a merge + bypassing a freeze](https://github.com/space-wizards/space-station-14/pull/35406)).

superjj18:
> Next thing people will be calling for Vox to no longer be poisoned by oxygen because it makes the species harder to play as...

Apples and oranges(vox and diona).  Vox have a negative trait that only affects them if they fail to prepare.  Diona are just fucked in general with no counter play until either this PR or rooting comes in.

It's not a matter of "difficulty" it's a matter of accessability.

K-Dynamic:
#32782 got merged but I don't think it should be mutually exclusive with this PR

slarticodefast:
> https://github.com/space-wizards/space-station-14/pull/32782 got merged but I don't think it should be mutually exclusive with this PR

I would at least wait for a while to see how it affects gameplay before making more major changes like this PR though

EmoGarbage404:
Updated the PR to interface with rooting. 

Now, you can no longer root while wearing shoes.

image

Once you are rooted, you can't put on footwear.

image

superjj18:
Do roots count as shoes when dealing with broken glass on the floor?

slarticodefast:
This PR was brought up in a voice call with a bunch of maintainers in it, and the consensus was that it is not needed:
- It makes diona too similar to other species, and in general we are trying to make them more unique and special.
-This PR was made before diona rooting #32782 was merged, which serves as a well-received alternative. It addressed the main problem of diona nukies not having magboots or noslips by giving them an innate anti-slip ability. The rest of the downsides seem relatively minor in comparison.
- If damage and stun through glass shards turn out to still be a problem in the future that can be adjusted on its own.

Ideally these design decisions would be written down in a diona species design doc, but you know - backlog 😔 

PR #37292, opened 2025-05-08
fix unremovable items used in construction

Errant-4:
## About the PR
Unremovable items can no longer be used in a construction step, most prominently this means that borgs can no longer use up/lose their tools

## Why / Balance
Unremovable items should not be removable

## Technical details
Any items that have the Unremovable component can no longer be used in construction steps. Since this would also prevent borgs from using their steel stack and such, an exception is made for items that have StackComponent.Lingering, since that should be specifically indicating that this item was in fact meant to be used up, and will not disappear once the stack is empty.

## Requirements

## Breaking changes
None

**Changelog**
:cl: Errant
- fix: Borgs can no longer craft items using up their tools as an ingredient, losing them forever.


TheShuEd:
Thats good fix!

PR #37299, opened 2025-05-09
EVA suit re-resprite

Flareguy:
## About the PR
Resprites EVA suits, again.

## Why / Balance
It's been 2 years, my understanding of the style has improved by a good bit, and the current ones have some pretty notable issues. If you have a tendency to scrutinize sprites, you've probably noticed that they just feel... off. This tries to rectify that by making their shading significantly more in-line with other items & reduces the general noisiness their designs had.
Generally speaking, these are *much* closer to just being cleaned up versions of the classic SS13 space suits - which is pretty ideal. The sprites always had a pretty good base, they just had some cruft to them.

## Technical details
N/A

## Media
Content Client_FHaQg6E9iM

## Requirements

## Breaking changes
N/A

**Changelog**
not big enough for a changelog probably


archee1:
could i suggest a side-by-side in the media to better show the changes.

My only irk with it is it feels too clean, and emergency eva looks less 'inflated' which was how I imagined people slow-mo walking in it.

Djungelskog2:
I really like these the old ones felt weird to me

Flareguy:
> could i suggest a side-by-side in the media to better show the changes.
> 
> My only irk with it is it feels too clean, and emergency eva looks less 'inflated' which was how I imagined people slow-mo walking in it.

Here's a side-by-side comparison:
image

I guess I'll talk a bit about my general specific feelings on the old EVA suit here. I avoided going too in-depth in the PR but since I have an opprutunity to elaborate, I won't spare the details.

Firstly - the removal of the blubberiness is deliberate. While a nice detail that did kind of help to convey the slowness of the suit, I just feel like it made the silhouette really awkward. I wasn't sure how I could really keep it like it was while still maintaining the more chunky kind of feel the classic bulky suit staircase pattern that the normal EVA suit now uses, instead of its previously rounded one. I never actually did try spriting it this way, so there is a possibility that I could make it work... but I felt like it ultimately just wasn't worth bothering with, since it let me make the suits quicker & I felt it was not really a detail that was necessary to conveying the purpose of the suit (it already spawns in lockers and is very emergency-looking with its patterns and colors.)

The other issue, the cleanliness, is kind of deliberate to an extent. The old softsuits are just *way* too damn detailed. The shading is really overdone & doesn't resemble how clothing is shaded in SS14 normally at all - it's got this weird asymmetric diagonal staircase pattern, and employs way too much anti-aliasing. Combined with the blue stripe spam on the normal suit, the end result is distinctly off-feeling. This especially applies to the helmet, which has a ton of anti-aliasing and colors that end up resulting in it just being way messier then it realistically should be.
I *will* note that the old SS13 EVA suits had asymmetric shading that I didn't carry over to this, which is probably a big part of why they feel so simple. Ultimately, I felt like it just wasn't really necessary. I suppose it could be good for balancing out the general of the contrast of the sprite - if someone felt like it needed to have that, I probably wouldn't object to that notion.

EmoGarbage404:
can we shutter the single jarring out-of-style eris EVA suit in this PR?

EmoGarbage404:
the really bad looking gray one

PR #37306, opened 2025-05-09
Map renderer rework

CyberTropic:
## About the PR

- Fix tiles being rendered upside down
- Added ability to render from files directly without using prototypes using -f
- Added option to show hidden markers using -m

## Why / Balance

Lets people render maps from files without making a prototype first. Paving the way to mapdiffbot.

## Technical details

Use TryLoadMap and TryLoadGrid to load the maps from files. ~~Since we don't know ahead of time whether a file is a grid or a map, we rely on the user to tell us, otherwise we throw an error.~~ We use the deserializer to check the file type before loading the map.
Added customOffset argument to painter functions, this is used when rendering planet maps, which don't have a LocalAABB component and aren't necessarily aligned with the 0,0 coordinate.
Since not all maps have a LocalAABB component, we use the tiles instead to calculate dimensions.
The rest of the changes are to add new arguments.

## Media

*Regular station map from file, markers enabled*
image

*Shuttle grid*
briggle-0

*map grid*
nukieplanet-0

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: MapRenderer can now render map files from a file directly without needing a prototype.
- fix: Fixed maprenderer drawing tiles upside down.

### DIFF
--- a/Content.MapRenderer/CommandLineArguments.cs
+++ b/Content.MapRenderer/CommandLineArguments.cs
@@ -12,6 +12,8 @@ public sealed class CommandLineArguments
     public bool ExportViewerJson { get; set; } = false;
     public string OutputPath { get; set; } = DirectoryExtensions.MapImages().FullName;
     public bool ArgumentsAreFileNames { get; set; } = false;
+    public bool FilesAreGrids { get; set; } = false;
+    public bool ShowMarkers { get; set; } = false;
 
     public static bool TryParse(IReadOnlyList<string> args, [NotNullWhen(true)] out CommandLineArguments? parsed)
     {
@@ -57,6 +59,18 @@ public sealed class CommandLineArguments
                 case "-f":
                 case "--files":
                     parsed.ArgumentsAreFileNames = true;
+                    parsed.FilesAreGrids = false;
+                    break;
+
+                case "-fg":
+                case "--files-grids":
+                    parsed.ArgumentsAreFileNames = true;
+                    parsed.FilesAreGrids = true;
+                    break;
+
+                case "-m":
+                case "--markers":
+                    parsed.ShowMarkers = true;
                     break;
 
                 case "-h":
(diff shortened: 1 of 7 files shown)


Content.MapRenderer/Painters/MapPainter.cs
```diff
@@ -54,17 +60,47 @@ await client.WaitPost(() =>
...
+                    else
+                    {
+                        if (mapLoader.TryLoadMap(resPath, out var loadedMaps, out var loadedGrids))
+                        {
+                            grids = loadedGrids.ToArray();
+                        }
+                    }
+                });
+            }
+
```
metalgearsloth, line 96: Wouldn't it be better to just read the metadata instead? That way the caller doesn't need to handle it.

metalgearsloth (dismissed):

southbridge-fur:
the legend

Content.MapRenderer/Painters/MapPainter.cs
CyberTropic: Do you mean like, read the file and then use the deserializer to find the type? I'll look into that some more, but it seemed like a bit of haunted code because we'd be reading the file twice, first here then in tryLoad.

CyberTropic: Okay, that was actually more straightforward than I thought it would be

EmoGarbage404 (approved):
kiss

Content.MapRenderer/CommandLineArguments.cs
```diff
@@ -95,7 +101,9 @@ Changes the path the rendered maps will get saved to.
         Defaults to Resources/MapImages
     -f / --files
         This option tells the map renderer that you supplied a list of map file names instead of their ids.
-        Example: Content.MapRenderer -f box.yml bagel.yml
+        Example: Content.MapRenderer -f /Maps/box.yml /Maps/bagel.yml
```
juliangiebel, line 104: Why would you change how you call the -f parameter? If I would've seen this before the PR got merged I'd have asked you to change this. This breaks the mapserver so every mapserver instance running can't render maps anymore.
The worst part is that this is a breaking change that was neither documented in the PR description nor announced on discord or relayed to me in any shape or form.

Content.MapRenderer/Program.cs
```diff
@@ -156,12 +129,14 @@ private static async Task Run(CommandLineArguments arguments)
                 };
```
juliangiebel, line 129:
```cs
var mapViewerData = new MapViewerData
{
    Id = map,
    Name = Thread.CurrentThread.CurrentCulture.TextInfo.ToTitleCase(map)
};
````
Id is supposed to be the entity id of the map. This breaks the mapviewer as maps are retrieved by their id. Just using the relative path is going to break things. Please at least use the file name without extension

Content.MapRenderer/CommandLineArguments.cs
CyberTropic: Wasn't aware that's how mapserver worked, but it should be a straightforward fix to just search through the /Maps folder for the filename instead of requiring the full path, I'll post a follow up PR.

PR #37307, opened 2025-05-09
Add test that mob damage thresholds have alerts

Tayrtahn:
## About the PR

Adds an integration test that checks every entity prototype with a `MobThresholdsComponent` and makes sure that each possible state for the mob is mapped to an `AlertPrototype` in the component's `StateAlertDict`.

## Why / Balance

I'm working on an integration test for setting off the station's nuke, and ran into [an error](https://github.com/space-wizards/space-station-14/blob/394b5e71f6c75032714cfa9583ec764d9ab372c9/Content.Shared/Mobs/Systems/MobThresholdSystem.cs#L382) for the cleanbot, honkbot, and other non-cyborg silicons that get killed by the explosion.

This test lets us catch this error and others like it without having to encounter it at runtime.

## Technical details

Iterates through all entity prototypes with a `MobThresholdsComponent`, and looks at the `Thresholds` DataField to find all of the states the mob can be in. The `StateAlertDict` DataField is then inspected to make sure that an entry exists for each state.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/Damageable/MobThresholdsTest.cs
+++ b/Content.IntegrationTests/Tests/Damageable/MobThresholdsTest.cs
@@ -1,0 +1,35 @@
+using Content.Shared.Mobs.Components;
+
+namespace Content.IntegrationTests.Tests.Damageable;
+
+public sealed class MobThresholdsTest
+{
+    [Test]
+    public async Task ValidateMobThresholds()
+    {
+        await using var pair = await PoolManager.GetServerClient();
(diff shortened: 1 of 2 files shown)


Tayrtahn:
I'm submitting this without fixing the bug initially to show that the test catches it. I'll push the bug fix after the tests finish.

Tayrtahn:
error output

EmoGarbage404 (approved):
strongest test-writer

PR #37309, opened 2025-05-09
Fix the component toggler

Samuka-C:
## About the PR
It solves a issue from the component toggler

## Why / Balance
If you use a item with component toggler with parent true (that is, it will add components to the parent and not the actual thing with the component toggler) and the item changes parents (you drop it, put in your bag, or if is a item in a module, you change the module) then once it toggles back, it doesn't remove the components from the entity it gave to.

## Technical details
I made so the component toggle component now remembers which entity it gave the components to and then it will try to remove the components from that entity and not their current parent.

## Media
Doesn't need any

## Requirements

## Breaking changes

**Changelog**
I don't think I need to make a fix changelog since the only thing affected by this bug were the cloaking device, that isn't available to players.

### DIFF
--- a/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
+++ b/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
@@ -26,7 +26,10 @@ public sealed class ComponentTogglerSystem : EntitySystem
 
             EntityManager.AddComponents(target, ent.Comp.Components);
         } else {
-            var target = ent.Comp.Target;
+            if (ent.Comp.Target == null)
+            return;
+
+            var target = ent.Comp.Target.Value;
 
             if (TerminatingOrDeleted(target))
             return;
--- a/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
+++ b/Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
@@ -16,13 +16,22 @@ public sealed class ComponentTogglerSystem : EntitySystem
 
     private void OnToggled(Entity<ComponentTogglerComponent> ent, ref ItemToggledEvent args)
     {
-        var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
-        if (TerminatingOrDeleted(target))
+        if (args.Activated) {
+            var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
+
+            ent.Comp.Target = target;
+
+            if (TerminatingOrDeleted(target))
             return;
 
-        if (args.Activated)
             EntityManager.AddComponents(target, ent.Comp.Components);
-        else
+        } else {
+            var target = ent.Comp.Target;
+
+            if (TerminatingOrDeleted(target))
+            return;
+
             EntityManager.RemoveComponents(target, ent.Comp.RemoveComponents ?? ent.Comp.Components);
+        }
     }
 }
--- a/Content.Shared/Item/ItemToggle/Components/ComponentTogglerComponent.cs
+++ b/Content.Shared/Item/ItemToggle/Components/ComponentTogglerComponent.cs
@@ -29,4 +29,7 @@ public sealed partial class ComponentTogglerComponent : Component
     /// </summary>
     [DataField]
     public bool Parent;
+
+    [DataField]
+    public EntityUid? Target = null;
 }


Content.Shared/Item/ItemToggle/Components/ComponentTogglerComponent.cs
```diff
@@ -29,4 +29,7 @@ public sealed partial class ComponentTogglerComponent : Component
     /// </summary>
     [DataField]
     public bool Parent;
+
+    [DataField]
+    public EntityUid? Target = null;
```
slarticodefast, line 34:
```suggestion
    public EntityUid? Target;
```
nullable types default to null, so no need to write it out
also add documentation

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
```diff
@@ -16,13 +16,25 @@ public override void Initialize()
 
     private void OnToggled(Entity<ComponentTogglerComponent> ent, ref ItemToggledEvent args)
     {
-        var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
-        if (TerminatingOrDeleted(target))
+        if (args.Activated) {
```
slarticodefast, line 19: fix formatting
```
if (condition)
{
    // weh
}
else
{
    // hew
}
```

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
```diff
@@ -16,13 +16,25 @@ public override void Initialize()
...
+            var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
+
+            ent.Comp.Target = target;
+
+            if (TerminatingOrDeleted(target))
```
slarticodefast, line 24: shouldn't this check be there before assigning the datafield?

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
```diff
@@ -16,13 +16,25 @@ public override void Initialize()
...
             return;
 
-        if (args.Activated)
             EntityManager.AddComponents(target, ent.Comp.Components);
-        else
+        } else {
+            if (ent.Comp.Target == null)
+            return;
+
+            var target = ent.Comp.Target.Value;
```
slarticodefast, line 32: just use this directly instead of assingning it to a variable

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
```diff
@@ -16,13 +16,25 @@ public override void Initialize()
...
+
+            if (TerminatingOrDeleted(target))
             return;
 
-        if (args.Activated)
             EntityManager.AddComponents(target, ent.Comp.Components);
-        else
+        } else {
+            if (ent.Comp.Target == null)
+            return;
```
slarticodefast, line 30: fix indentation

slarticodefast (changes requested):

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
Samuka-C: okay, but I just copied how it was in the same file 

```
if (TerminatingOrDeleted(target))
return;
```

Content.Shared/Item/ItemToggle/Components/ComponentTogglerComponent.cs
Samuka-C: fixed

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
Samuka-C: fixed

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
Samuka-C: fixed

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
Samuka-C: fixed (and fixed on the thing that was there before me too)

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
Samuka-C: fixed

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
```diff
@@ -16,13 +16,26 @@ public override void Initialize()
 
     private void OnToggled(Entity<ComponentTogglerComponent> ent, ref ItemToggledEvent args)
     {
-        var target = ent.Comp.Parent ? Transform(ent).ParentUid : ent.Owner;
-        if (TerminatingOrDeleted(target))
-            return;
+        if (args.Activated)
```
slarticodefast, line 19:
```suggestion
        if (args.Activated)
```
fix whitespace

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
```diff
@@ -16,13 +16,26 @@ public override void Initialize()
...
+
+            ent.Comp.Target = target;
 
-        if (args.Activated)
             EntityManager.AddComponents(target, ent.Comp.Components);
-        else
-            EntityManager.RemoveComponents(target, ent.Comp.RemoveComponents ?? ent.Comp.Components);
+        } 
+        else 
+        {
```
slarticodefast, line 31:
```suggestion
        }
        else
        {
```

slarticodefast (changes requested):

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
Samuka-C: fixed

Content.Shared/Item/ItemToggle/ComponentTogglerSystem.cs
Samuka-C: fixed

slarticodefast (approved):
Looks good to me, thanks for the bugfix!

PR #37320, opened 2025-05-10
Fix meteors deceleration

B-Kirill:
## About the PR
Meteors no longer slow down during flight

## Why / Balance
It's strange to look at a rotating meteor frozen in the middle of the station

## Media
_Definitely not necessary, it just works_

## Requirements

**Changelog**
:cl:
- fix: Meteors no longer slow down during flight.


Pronana:
It's a bit too late to comment but yeah this is the correct way to fix this. 

If there's any other space objects I missed that should have no offgrid friction, set the tileModifier to 0 since that directly modifies the body modifier in the tileFrictionController.

This is generally the correct solution and I'm leaving this comment for future reference.

Pronana:
Also we don't need the linear and angular damping in the physicscomponent yaml since that gets overwritten anytime the physics updates.

PR #37332, opened 2025-05-10
Force Gun De AAing

### LINKED ISSUE #35990: Force Gun AA
## Description

The Force Gun can fling objects/containers in between and through occupied tiles like walls or doors. A player can enter a container and use the force gun to phase through to the other side. 

**Reproduction**

1. Urist gets Force Gun and a container (I only used the crate and locker for my testing.)
2. Urist gets in container and uses the Force Gun on said container to drive it.
3. Urist drives container in between a wall/door/etc.  
4. Urist is now on the other side.
5. Profit

**Examples**

Locker Ex.1

Image

Locker Ex.2 (Farthest I Could Do)

Image

Crate Ex.1 

Image

Crate Ex.2 (Farthest I Could Do)

Image

Prole0:
## About the PR
The Force Gun got AA capabilities, it was fun while it lasted.

## Why / Balance
Fixes #35990, no locker or crate AA. Velocity would be too low to accomplish AA. 

## Technical details
Halved ThrowForce from 30f to 15f, tried to go higher but then some of its capabilities returned. At 20f it got back those AA capabilities. 

However, this will reduce the Force Gun's Velocity in general. Which is also the reason why it has AA, Velocity gives it AA.

## Media

30f (Before)

https://github.com/user-attachments/assets/cf4fbc2d-9000-4236-b555-9c06b08b0633

15f (After)

https://github.com/user-attachments/assets/b7fd56f5-4b04-46a8-8f6d-dd5526fc6c60

## Requirements

## Breaking changes

**Changelog**

No CL, maybe?


Prole0:
Also affects the Admin ForceGun, both versions use the same component.

B-Kirill:
Perhaps a better solution would be to ban the use of this gun on objects if the user is a child of this object (meaning contained in it) or to ban its use altogether if the player uses it while in a container.

Prole0:
> Perhaps a better solution would be to ban the use of this gun on objects if the user is a child of this object (meaning contained in it) or to ban its use altogether if the player uses it while in a container.

It seems like there is something similar, there's a pop-up saying "It's too cramped"
Pop-Up Example

Prole0:
I also have to clarify that this can still be done outside of the closet/crate. And with most items affected by the force gun. Which still leaves this as an option.


https://github.com/user-attachments/assets/4018f2e5-9194-4303-9226-67d339cc79a4


https://github.com/user-attachments/assets/d0909fc1-ca0d-4c38-b966-970bd691b912


Harder to do but I got it within a minute for the closet and 3 for the rcd. And theirs easier tech to do the same.

B-Kirill:
Wow, this gun needs to be repaired thoroughly

Callmore:
the issue here isn't with the force gun, it's much more likely with the game's physics itself, since your clipping though the wall, it's more like your sm64 BLJ'ing though it.
the solution is probably the same solution to most of these physics problems, substepping

ArtisticRoomba:
like what callmore said this is generally just a core physics problem, this happens with people getting launched by containment fields and they phase through the wall and get launched out into space or get stuck in a layer in a thick wall

pheenty:
Whitehole and supermatter grenades do that too

Prole0:
> Whitehole and supermatter grenades do that too

Yep, physics problem. So should be physics solution to fix it. Keeping peak in, closing.

pheenty:
Forcegun is probably the least issue out of these by the way, it's locked behind T3 research and has no actual use

Prole0:
> Forcegun is probably the least issue out of these by the way, it's locked behind T3 research and has no actual use

Yea, it's a gimmick too. An easy counter to this would be opening the closet/crate. Putting them in a partially filled tile which would get them stuck in-between. Turning off grav, etc.

A more common problem with these physics bugs is players getting stuck in walls through the supermater grenades or an entity pushing them through. Mostly the latter now, cause I haven't seen a supermatter used lately.

PR #37336, opened 2025-05-10
Add ability for Interaction tests to spawn a station entity

Tayrtahn:
## About the PR

Adds the option for integration tests derived from `InteractionTest` to specify a station prototype that will be spawned and cleaned up automatically as part of the test.

This was separated from https://github.com/space-wizards/space-station-14/pull/36329 for easier review.

## Why / Balance

A bunch of systems rely on having a station and its associated data set up: station records, mail deliveries, salvage magnet stuff, etc.

This doesn't mean spawning in an actual station map; it just needs the nullspace entity that holds station-related data. These prototypes can be found in [Resources/Prototypes/Entities/Stations](https://github.com/space-wizards/space-station-14/tree/34f8be84b6989a49d614879be8db005864f310a4/Resources/Prototypes/Entities/Stations).

## Technical details

Adds a nullable, virtual `EntProtoId` `StationPrototype` to `InteractionTest`, defaulting to null.

During test setup, if `StationPrototype` is non-null, `StationSystem.InitializeNewStation` is called, passing in `StationPrototype` and the test grid. A reference to the station's `NetEntity` is stored in a protected field of `InteractionTest` for easy access by tests.

During test teardown, the station entity is deleted if it exists.

Adding this functionality into `InteractionTest` allows the station to be set up at the right time so that station systems can respond to the player being spawned (such as by adding them to the crew records). It also reduces code duplication by not making multiple tests reimplement this functionality.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/Interaction/InteractionTest.cs
+++ b/Content.IntegrationTests/Tests/Interaction/InteractionTest.cs
@@ -26,6 +26,8 @@ using Robust.Shared.Timing;
 using Robust.UnitTesting;
 using Content.Shared.Item.ItemToggle;
 using Robust.Client.State;
+using Content.Server.Station.Systems;
+using Content.Server.Station;
 
 namespace Content.IntegrationTests.Tests.Interaction;
 
@@ -110,6 +112,7 @@ public abstract partial class InteractionTest
     protected SharedMapSystem MapSystem = default!;
     protected ISawmill SLogger = default!;
     protected SharedUserInterfaceSystem SUiSys = default!;
+    protected StationSystem StationSys = default!;
 
     // CLIENT dependencies
     protected IEntityManager CEntMan = default!;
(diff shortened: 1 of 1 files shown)


Content.IntegrationTests/Tests/Interaction/InteractionTest.cs
```diff
@@ -194,6 +209,15 @@ public virtual async Task Setup()
         TargetCoords = SEntMan.GetNetCoordinates(Transform.WithEntityId(MapData.GridCoords.Offset(new Vector2(1.5f, 0.5f)), MapData.MapUid));
         await SetTile(Plating, grid: MapData.Grid);
 
+        if (StationPrototype is not null)
+        {
+            await Server.WaitPost(() =>
+            {
+                var station = StationSys.InitializeNewStation(new StationConfig() { StationPrototype = StationPrototype }, [MapData.Grid]);
```
slarticodefast, line 216: Would it maybe make sense to instead have a `StationConfig` parameter instead of `StationPrototype`?
That way it would be possible to test them in the full context they are normally used in.

ElectroJr, line 216: `StationComponentOverrides` in StationConfig is not marked as nullable, and the constructor does a ` = default!` because its a required yaml field, so this could lead to NREs. Otherwise LGTM, beyond maybe doing as slarti suggested.
```suggestion
                var station = StationSys.InitializeNewStation(new StationConfig() { StationPrototype = StationPrototype, StationComponentOverrides = new() }, [MapData.Grid]);
```

ElectroJr: Yeah that's probably a better idea. Though actually using the StationComponentOverrides is probably a hassle, unless you manually parse it from yaml with SerializationManager.

TheShuEd:

PR #37342, opened 2025-05-11
Add support for rotatable tiles in map renderer

SlamBamActionman:
## About the PR

This PR adds support for rotatable/mirrored tiles in the tile render.

It also fixes a bug with the map renderer where apparently all tiles have been rendering at a 180 degree rotation. This was partially noticed when decals didn't line up for things like regular grey tiles, but the cause of the misalignment was misidentified and instead the decals were moved one pixel. So this is now fixed.

## Why / Balance

Without it, the map renderer looks all wrong!

## Technical details

It checks the tile's `RotationMirroring` property and modifies the tile sprite based on it. The property has normal rotations in 0-3, and mirrored rotations in 4-7.

For some reason the way the renderer works is that it flips the entire sprite horizontally at the end? So for that reason we flip the tile if it *isn't* mirrored, so that it can be flipped back again afterwards.

### Rotated tiles:

Before:
image

After:
image

### 180 rotation bugfix:
Base sprite:
eighties

Before:

image

After:
image

## Requirements

## Breaking changes

**Changelog**

No CL


EmoGarbage404:
goncflicts

EmoGarbage404 (commented):
Does this work properly now that the other PR also fixed the tiles being upside down? It does it in a different place so my immediate thought is that the two would cancel out or cause some sort of issue.

Issue #37269, opened 2025-05-08
sitting on a chair that youre dragging causes dev client crash


TaoNewt:
## Description
Sitting in the chair that youre dragging crashes the client and server.

**Reproduction**
Be actively dragging seat, drag your own sprite onto the seat at same time.

**Screenshots**

**Additional context**
in vs code, not sure if that matters tho.

PR #37344, opened 2025-05-11
Debug Assert when buckling and pulling

insoPL:
## About the PR

This PR unexpectedly grew, so I started to split it, i need to split tryStopPull  into  canStopPull  and StopPull:
https://github.com/space-wizards/space-station-14/pull/38029

image

Debug only assert fix:
1. pull chair 
2. hop onto the chair
3. assert

1. pull chair
2. someone else put you on that chair
3. assert

## Why / Balance

N/A

## Technical details

checks for pulling, stops the pulling if buckleUid is the same as pulled object

asserts added to toolbox here, i am not aware of any bug caused by this on release build
https://github.com/space-wizards/RobustToolbox/pull/4490

fixes:
https://github.com/space-wizards/space-station-14/issues/37269
## Media

N/A

## Requirements

## Breaking changes

public bool CanUnbuckle(Entity<BuckleComponent?> buckle, EntityUid user, bool popup) was changed into
public bool CanUnbuckle(Entity<BuckleComponent?> buckle, EntityUid? user, bool popup)

**Changelog**


Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -257,6 +258,9 @@ private bool CanBuckle(EntityUid buckleUid,
         if (!_container.IsInSameOrNoContainer((buckleUid, null, null), (strapUid, null, null)))
             return false;
 
+        if (TryComp<PullableComponent>(strapUid, out var pullable) && pullable.Puller == buckleUid)
+            _pullingSystem.TryStopPull(strapUid, pullable);
+
```
EmoGarbage404, line 263: This really isn't the proper place for this code. "Can" methods that check if behavior can occur should not contain actual visible behavior within them. It would make sense to either cancel the buckle if you are pulling it, _or_ move the `TryStopPull` call into the `Buckle()` or `SetBuckledTo()` logic

EmoGarbage404 (changes requested):

Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
insoPL: I don't want to change game logic by canceling the buckle. I moved it to Buckle

Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -360,6 +361,9 @@ private void Buckle(Entity<BuckleComponent> buckle, Entity<StrapComponent> strap
         else if (user != null)
             _adminLogger.Add(LogType.Action, LogImpact.Low, $"{ToPrettyString(user):player} buckled {ToPrettyString(buckle)} to {ToPrettyString(strap)}");
 
+        if (TryComp<PullableComponent>(strap.Owner, out var pullable) && pullable.Puller == buckle.Owner)
+            _pullingSystem.TryStopPull(strap.Owner, pullable);
```
EmoGarbage404, line 365: I think you'll need to change `TryStopPull` to have some sort of "forced" behavior where it doesn't allow it to fail. Presently, the current code doesn't handle a situation where the `TryStopPull` call fails to actually stop the pull (since it's not guaranteed) causing the assert to still trigger.

EmoGarbage404 (changes requested):

insoPL:
I refactored whole buckling thing now its little bit more readable and should stop the buckle if it can't stop the pulling action.,

insoPL:
Should be good now

metalgearsloth:
Needs breaking change due to API breakage.

Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -215,144 +216,150 @@ protected void SetBuckledTo(Entity<BuckleComponent> buckle, Entity<StrapComponen
...
+    {
+        if (!TryComp<PullableComponent>(strap.Owner, out var pullable))
+            return true; // strap owner is not pullable, self pulling not possible
+
+        if (pullable.Puller != buckle.Owner)
+            return true;
+
+        return _pullingSystem.TryStopPull(strap.Owner, pullable);
+    }
+
```
metalgearsloth, line 362: This should be part of canbuckle because it's possible for CanBuckle to pass and trybuckle to fail which should never happen.

insoPL: So you are tellin me that:
1. TryStopPull can't be part of CanBuckle because its changing state.
2. TryStopPull can't be part of Buckle because it can fail.
3. TryStopPull can't be it's own thing because  if CanBuckle  passes TryBuckle also has to pass.

Ok, I will need to refactor TryStopPulll into CanStopPull and StopPull
then I can do CanStopPull in CanBuckle  and StopPull StopBuckle.

metalgearsloth: IDK about buckle but trystoppull shouldn't be in canbuckle and you should be able to call canbuckle then buckle and not have behavior different to trybuckle

insoPL: As Emo said putting any "Try" method into buckle is not optimal:
"Presently, the current code doesn't handle a situation where the TryStopPull call fails to actually stop the pull (since it's not guaranteed) causing the assert to still trigger"
I can't cancel buckling either as it suppose to buckle after positive verification from CanBuckle

Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -215,141 +216,139 @@ protected void SetBuckledTo(Entity<BuckleComponent> buckle, Entity<StrapComponen
...
         {
             if (popup)
-                _popup.PopupClient(Loc.GetString("buckle-component-no-hands-message"), user);
+                _popup.PopupClient(Loc.GetString("buckle-component-no-hands-message"), userUid);
 
             return false;
         }
 
-        if (buckleComp.Buckled && !TryUnbuckle(buckleUid, user, buckleComp))
+        if (buckle.Comp.Buckled && !TryUnbuckle(buckle.Owner, userUid, buckle.Comp))
```
insoPL, line 256: TryUnbuckle is state changing  method inisde can method

insoPL: i will change it to CanBuckle recursion
and move Buckle part into Buckle

insoPL:
This PR unexpectedly grew, so I started to split it:
https://github.com/space-wizards/space-station-14/pull/38029

Issue #37225, opened 2025-05-06
Re-Built Genpop Turnstiles cannot be access configured for genpop entry/exit.


BramvanZijp:
Genpop Prisoner Closet Sub-Issue 6:

You can build new ones, sure, but you cannot configure them because the access configurator does not include the 2 genpop accesses, nor do security ID cards have them (mapped ones have both security AND genpop enter/exit lock)

BramvanZijp:
Thatll be issue 7 I guess

perryprog:
Fixed via #39043. For your other noted things, feel free to make new issues about those if there aren't already existing ones (such as #37227).

Issue #37222, opened 2025-05-06
Genpop Prisoner IDs form very easy trash with no easy way to get rid of them.


BramvanZijp:
Genpop Prisoner Closet Sub-Issue 5:

You are going to be left with a lot of trash ID cards, somehow making them dissolve or disintigrate after being cleared from a closet, or X minutes after the sentence ending, may be usefull.

This issue also includes the fact that the exit access never expires. Meaning if you smuggle your old genpop ID back in in IE an implant, it is an easy escape tool. So that too needs to be solved, and it would be solved by them disintigrating or de-activating when cleared.

themias:
Maybe they could be recycle-able?

BramvanZijp:
> Maybe they could be recycle-able?

Would be an unlimited material farm.

I think that if a security officer selects "Clear ID" on the locker, or after X minutes of the sentence being over automatically, the ID should AT THE VERY LEAST be "de-activated", losing its ability to open the turnstiles, and also get marked as a junk item that can be recycled by a recycler WITHOUT giving back any materials.

ALTERNATIVELY, you could make it so that the lockers have to be supplied with plastic to print IDs, and then have the IDs be recyclable for that plastic that it cost to make.

ScarKy0:
They are recyclable and give no materials, just break.

themias:
> They are recyclable and give no materials, just break. I'm pretty sure at least?

Checking now, they have the recyclable tag and no material comp, so they shouldn't produce anything.

BramvanZijp:
> They are recyclable and give no materials, just break. I'm pretty sure at least?

I still think they should de-activate if you clear them, and have them clear automatically after sentence is up for X period.
You dont want people smuggling old prisoner IDs back in and having those actually function. (commit vandalism, get released after a few minutes, put your old ID in a storage implant, commit murder and get permad, and then use that old ID to get out)

PR #37350, opened 2025-05-11
Normalize turnstile construction to be more like airlocks -- use door electronics for access configuration

Quantum-cross:
## About the PR

Make the turnstile construction more like doors.

- Metal sheet + Metal rod -> Turnstile assembly
- Anchor the assembly
- Add LV wire -> wired node
- Add door electronics -> electronics node
- Use a screwdriver -> Turnstile (finished) node

## Why / Balance

In the same vein of normalizing machine and computer construction, I believe that turnstiles should be as similar as possible to airlock construction. This also solves the problem (#37225) of re-configuring genpop turnstile access after they have been destroyed or deconstructed, as the door electronics can provide access configuration.

This also increases the effort to deconstruct a turnstyle as they now require a crowbar to pull out the board.

## Technical details

- Turnstile assembly entity has been added, mirroring what has been done with airlock assemblies
- DoorElectronicsGenpop(Enter|Leave) prototypes have been added with accesses pre-configured into door electronics
- Turnstile components have been added to allow a board slot with the access reader
- TurnstileGenpop(Enter|Leave) entities have been changed to contain the proper board in its fill

## Requirements

## Breaking changes

None anticipated, previously mapped `TurnstileGenpopEnter` and `TurnstileGenpopLeave` should work as expected.

**Changelog**
:cl:
- add: Turnstiles now require board electronics, which means access can be configured

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/Electronics/door_access.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/Electronics/door_access.yml
@@ -244,6 +244,22 @@
   - type: AccessReader
     access: [["Brig"]]
 
+- type: entity
+  parent: DoorElectronics
+  id: DoorElectronicsGenpopEnter
+  suffix: Genpop Enter, Locked
+  components:
+  - type: AccessReader
+    access: [["GenpopEnter"], ["Security"]]
+
+- type: entity
+  parent: DoorElectronics
+  id: DoorElectronicsGenpopLeave
+  suffix: Genpop Leave, Locked
+  components:
+  - type: AccessReader
+    access: [["GenpopLeave"], ["Security"]]
+
 # Medical
 - type: entity
   parent: DoorElectronics
--- a/Resources/Prototypes/Entities/Structures/Doors/turnstile.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/turnstile.yml
@@ -1,4 +1,47 @@
 - type: entity
+  id: TurnstileAssembly
+  name: turnstile assembly
+  description: A mechanical door that permits one-way access and prevents tailgating.
+  components:
+  - type: Clickable
+  - type: InteractionOutline
+  - type: Sprite
+    sprite: Structures/Doors/Airlocks/Standard/basic.rsi
+    state: "assembly"
+  - type: Physics
+  - type: Fixtures
+    fixtures:
+      fix1:
+        shape:
+          !type:PhysShapeAabb
+          bounds: "-0.45,-0.45,0.45,0.45"
+        density: 110
+        mask:
+        - FullTileMask
+        layer:
+        - HumanoidBlockLayer
+  - type: Anchorable
+    delay: 2
+  - type: Pullable
+  - type: Transform
+    anchored: true
+    noRot: true
+  - type: Damageable
+    damageContainer: StructuralInorganic
+    damageModifierSet: StructuralMetallicStrong
+  - type: Destructible
+    thresholds:
+    - trigger:
+        !type:DamageTrigger
+        damage: 500
+      behaviors:
+      - !type:DoActsBehavior
+        acts: [ "Destruction" ]
+  - type: Construction
+    graph: Turnstile
+    node: assembly
+
+- type: entity
   id: Turnstile
   parent: BaseStructure
   name: turnstile
@@ -55,7 +98,14 @@
       behaviors:
       - !type:DoActsBehavior
         acts: ["Destruction"]
+  - type: ContainerContainer
+    containers:
+      board: !type:Container
+  - type: ContainerFill
+    containers:
+      board: [ DoorElectronics ]
   - type: AccessReader
+    containerAccessProvider: board
   - type: Construction
     graph: Turnstile
     node: turnstile
@@ -84,13 +134,15 @@
   parent: Turnstile
   suffix: Genpop Enter
   components:
-  - type: AccessReader
-    access: [["GenpopEnter"], ["Security"]]
+  - type: ContainerFill
+    containers:
+      board: [ DoorElectronicsGenpopEnter ]
 
 - type: entity
   id: TurnstileGenpopLeave
   parent: Turnstile
   suffix: Genpop Leave
   components:
-  - type: AccessReader
-    access: [["GenpopLeave"], ["Security"]]
+  - type: ContainerFill
+    containers:
+      board: [ DoorElectronicsGenpopLeave ]
--- a/Resources/Prototypes/Recipes/Construction/Graphs/structures/turnstile.yml
+++ b/Resources/Prototypes/Recipes/Construction/Graphs/structures/turnstile.yml
@@ -3,12 +3,11 @@
   start: start
   graph:
   - node: start
-    actions:
-    - !type:DeleteEntity { }
     edges:
-    - to: turnstile
+    - to: assembly
       completed:
-      - !type:SnapToGrid
+      - !type:SetAnchor
+        value: false
       steps:
       - material: MetalRod
         amount: 4
@@ -17,17 +16,87 @@
         amount: 1
         doAfter: 2
 
-  - node: turnstile
-    entity: Turnstile
+  - node: assembly
+    entity: TurnstileAssembly
+    actions:
+    - !type:SnapToGrid {}
+    - !type:SetAnchor {}
     edges:
+    - to: wired
+      conditions:
+      - !type:EntityAnchored {}
+      steps:
+      - material: Cable
+        amount: 5
+        doAfter: 2.5
     - to: start
+      conditions:
+      - !type:EntityAnchored
+        anchored: false
       completed:
       - !type:SpawnPrototype
         prototype: PartRodMetal1
         amount: 4
-      - !type:DeleteEntity
+      - !type:SpawnPrototype
+        prototype: SheetSteel1
+        amount: 1
+      - !type:DeleteEntity {}
       steps:
       - tool: Welding
-        doAfter: 4.0
+        doAfter: 3
+
+  - node: wired
+    entity: TurnstileAssembly
+    edges:
+    - to: electronics
+      conditions:
+      - !type:EntityAnchored {}
+      steps:
+      - component: DoorElectronics
+        store: board
+        name: construction-graph-component-door-electronics-circuit-board
+        icon:
+          sprite: "Objects/Misc/module.rsi"
+          state: "door_electronics"
+        doAfter: 3
+    - to: assembly
+      completed:
+      - !type:SpawnPrototype
+        prototype: CableApcStack1
+        amount: 5
+      steps:
       - tool: Cutting
-        doAfter: 2.0
+        doAfter: 2.5
+
+  - node: electronics
+    edges:
+    - to: turnstile
+      conditions:
+      - !type:EntityAnchored {}
+      steps:
+      - tool: Screwing
+        doAfter: 2.5
+    - to: wired
+      conditions:
+      - !type:EntityAnchored {}
+      completed:
+      - !type:EmptyAllContainers
+        pickup: true
+        emptyAtUser: true
+      steps:
+      - tool: Prying
+        doAfter: 5
+  - node: turnstile
+    entity: Turnstile
+    edges:
+    - to: wired
+      conditions:
+      - !type:EntityAnchored {}
+      completed:
+      - !type:EmptyAllContainers
+        pickup: true
+        emptyAtUser: true
+      steps:
+      - tool: Prying
+        doAfter: 5
+


EmoGarbage404 (changes requested):
what's the reasoning for all the door parity added with turnstiles? You straight up wrote nothing about it in the PR description.

Just from a basic overview, being able to bolt, electrify, and cut the power to disable it all seem like pretty noteworthy changes. Electrifying in particular seems particularly painful due to the fact that people can make particularly nasty inescapable traps with these. I understand it's harder with the beefed up construction but it's definitely not impossible.

Being able to bolt them also is weird IMO, since part of the purpose behind them is to enable prisoners to freely move through genpop. What does it even look like if you bolt the turnstile while people are going through, or while there's a pull chain?

I have similar feelings about cutting power. Should prisoners be trapped in genpop just because power is out, even if their sentence is done? This is a tangible balance change that isn't addressed at all.

As for construction, the new electronics for turnstiles need to be made accessible in game. AFAIK you can't manually add the access from the normal configurator and even if you could, sec doesn't actually have it on their IDs. You'd best make a crate so they can order some from cargo.

Quantum-cross:
> what's the reasoning for all the door parity added with turnstiles? You straight up wrote nothing about it in the PR description.

I wrote: "In the same vein of normalizing machine and computer construction, I believe that turnstiles should be as similar as possible to airlock construction."

We need configurable access on turnstiles so I thought that making it parity to airlocks would be intuitive from an engineering sense.

> Just from a basic overview, being able to bolt, electrify, and cut the power to disable it all seem like pretty noteworthy changes. 

I understand that these are significant balance changes, and I'm not personally comfortable enough to judge balance changes myself, I'm hoping to open a dialog on which of these features if any should come through.

> Electrifying in particular seems particularly painful due to the fact that people can make particularly nasty inescapable traps with these. I understand it's harder with the beefed up construction but it's definitely not impossible.

The check for shock only happens on the initial collision in the correct or incorrect direction. If you are shocked walking through the correct direction, you will be able to walk through the rest of the way after standing back up.

> Being able to bolt them also is weird IMO, since part of the purpose behind them is to enable prisoners to freely move through genpop.

Sure, but I feel like it makes sense to be able to bolt them in an emergency, say... zeds?

> What does it even look like if you bolt the turnstile while people are going through

The check for bolt is only on the initial collision in the correct direction. This makes sense realistically in my head too as I'm imagining there being a spring loaded pawl system when the turnstile makes a full rotation.

> or while there's a pull chain?

As is right now it lets the chain through, and kind of messes up the light indicators a bit. If this PR continues, once it's determined what the behavior should be, I'll ensure that it's fixed.

> I have similar feelings about cutting power. Should prisoners be trapped in genpop just because power is out, even if their sentence is done? This is a tangible balance change that isn't addressed at all.

Good question. I added power receiver because it doesn't make sense for a device with door electronics to use no power.

> As for construction, the new electronics for turnstiles need to be made accessible in game. AFAIK you can't manually add the access from the normal configurator and even if you could, sec doesn't actually have it on their IDs. You'd best make a crate so they can order some from cargo.

The electronics are standard door electronics that can be printed at the Autolathe, you don't need a configurator from a head, just a multitool.

image

BramvanZijp:
DISCLAIMER: I am a Triager, not a Maintainer. Despite me being listed as a Member, this is just me providing an OPINION how any other contributor could/would. I have no say in if this PR should or should not be merged, but I do have personal feedback.

I do agree these should get more door parity, here is my in-depth thoughts:
1. These are one way doors with access restrictions, that should require electronics, which also makes replacing them feasible. This also solves them being too easy to build for trolling. Make sure the deconstruction steps are also updated if they havent been. so that you can for example remove and re-insert the door elecs (after fixing the door elec access) if they got emagged without having to fully de and re construct.
2. They should not be able to be electrified. Not by the AI, and not by cutting 1/2 power wires. I agree that will cause problems.
3. Being able to bolt it in emergencies is a good thing, keep that for both AI and door remotes, as well as the wire.
4. I believe there should also be a DIR (Directional) wire that, if cut, or for 15/30 seconds if pulsed, removes the direction requirement, allowing you to go through both directions without the turnstile resisting your efforts.
5. I am not sure if this is currently the case, but the prototypes for the genpop enter/exit turnstiles should be updated to have higher security protection for the door panels, simular to vault doors. This means a plasteel plate and electrified grille inside the panel to make hacking it a tough job. If thats too much, the basic security, such as that in place for head of staff offices (just a steel plate welded inside) would do too.
6. IMO the need for power should be removed entirely, that will cause issues with the primarily application being use in perma. Remove the appropriate wires too. You can replace it with a lock wire that suspends access requirements so that if you can hack it, you can still force it open just like regular doors, just for this it is by removing access instead of removing power.

BramvanZijp:
@EmoGarbage404 Would you be satisfied with the PR if the above changes are made?

Quantum-cross:
>     1. These are one way doors with access restrictions, that should require electronics, which also makes replacing them feasible. This also solves them being too easy to build for trolling. Make sure the deconstruction steps are also updated if they havent been. so that you can for example remove and re-insert the door elecs (after fixing the door elec access) if they got emagged without having to fully de and re construct.

The construction and deconstruction steps are nearly identical to airlocks, and door electronics can be pulled out after snipping the wires.

I have not implemented an emag interaction. There are a few things to consider with this.

1. Does emagging remove the direction requirement, the access requirement, or both? 
2. With normal airlocks, it's reasonably obvious that an airlock is emagged because it gets bolted open. This cannot happen visually with a turnstile, but maybe the fact that your prisoners are suddenly gone might be a good indicator...
3. Another option for a visual distinction would be the indicator lights constantly blinking "green" or even "yellow" which kind of mimics an emergency access kind of indicator. This emagged indicator state would be reset on re-construction of the turnstile (the first deconstruction step will turn it into a turnstile assembly, and it must be partially deconstructed to get to the door electronics board to repair).

>     2. They should not be able to be electrified. Not by the AI, and not by cutting 1/2 power wires. I agree that will cause problems.

I would probably agree with this as well, I only added the power because it "realistically made sense", but mechanics comes before that.

>     3. Being able to bolt it in emergencies is a good thing, keep that for both AI and door remotes, as well as the wire.

This also increases the difficulty in hacking with only a wirecutter and not a multitool, and accidentally bolting the turnstile with a wirecutter without the ability to unbolt it will absolutely raise eyebrows for a failed jailbreak.

Also now that I am thinking of this, the player constructed turnstiles should have a different wires layout than the genpop mapped ones so that a bad actor can't construct their own turnstile and memorize the layout before they get brigged.

>     4. I believe there should also be a DIR (Directional) wire that, if cut, or for 15/30 seconds if pulsed, removes the direction requirement, allowing you to go through both directions without the turnstile resisting your efforts.

This would be funny, but without the danger of a power wire, wouldn't this make jailbreaks incredibly easy with just a screwdriver and multitool (and almost easy with a wirecutter if you don't hit the bolt wire)?

>     5. I am not sure if this is currently the case, but the prototypes for the genpop enter/exit turnstiles should be updated to have higher security protection for the door panels, simular to vault doors. This means a plasteel plate and electrified grille inside the panel to make hacking it a tough job. If thats too much, the basic security, such as that in place for head of staff offices (just a steel plate welded inside) would do too.

I can add high security versions, but I believe the standard versions should be the ones mapped (unless you want this to be a mitigation for my objection to point 4.)

>     6. IMO the need for power should be removed entirely, that will cause issues with the primarily application being use in perma. Remove the appropriate wires too.

I can remove power requirements and the power wires.

>    You can replace it with a lock wire that suspends access requirements so that if you can hack it, you can still force it open just like regular doors, just for this it is by removing access instead of removing power.

How about a `SOL` wire which would represent the solenoid holding the spinner in place, a cut or pulse on this wire would turn on the `SOL` wire indicator, representing an activation or partial activation of the solenoid, during which you could begin a do-after. The result of the do-after would be popup that says "You manage to force the turnstile mechanism!". In this state, next entry through the turnstile will be allowed regardless of access restrictions.

Some testing needs to be done to ensure in the case of say 3 people pushing against the turnstile while someone performs the do-after, if multiple people try to enter in the same tick it should prioritize the do-after actor.

-----

Overall, I'm totally cool and open to changes here.

I was actually going to suggest rolling back everything except the simple addition of door electronics,

I just added all the other shit because it's so damn easy with ECS and it's cool.

-----

@BramvanZijp thank you for the thoughtful changes

BramvanZijp:
> > ```
> > 1. These are one way doors with access restrictions, that should require electronics, which also makes replacing them feasible. This also solves them being too easy to build for trolling. Make sure the deconstruction steps are also updated if they havent been. so that you can for example remove and re-insert the door elecs (after fixing the door elec access) if they got emagged without having to fully de and re construct.
> > ```
> 
> The construction and deconstruction steps are nearly identical to airlocks, and door electronics can be pulled out after snipping the wires.
> 
> I have not implemented an emag interaction. There are a few things to consider with this.
> 
> 1. Does emagging remove the direction requirement, the access requirement, or both?
> 2. With normal airlocks, it's reasonably obvious that an airlock is emagged because it gets bolted open. This cannot happen visually with a turnstile, but maybe the fact that your prisoners are suddenly gone might be a good indicator...
> 3. Another option for a visual distinction would be the indicator lights constantly blinking "green" or even "yellow" which kind of mimics an emergency access kind of indicator. This emagged indicator state would be reset on re-construction of the turnstile (the first deconstruction step will turn it into a turnstile assembly, and it must be partially deconstructed to get to the door electronics board to repair).
> 
> > ```
> > 2. They should not be able to be electrified. Not by the AI, and not by cutting 1/2 power wires. I agree that will cause problems.
> > ```
> 
> I would probably agree with this as well, I only added the power because it "realistically made sense", but mechanics comes before that.
> 
> > ```
> > 3. Being able to bolt it in emergencies is a good thing, keep that for both AI and door remotes, as well as the wire.
> > ```
> 
> This also increases the difficulty in hacking with only a wirecutter and not a multitool, and accidentally bolting the turnstile with a wirecutter without the ability to unbolt it will absolutely raise eyebrows for a failed jailbreak.
> 
> Also now that I am thinking of this, the player constructed turnstiles should have a different wires layout than the genpop mapped ones so that a bad actor can't construct their own turnstile and memorize the layout before they get brigged.
> 
> > ```
> > 4. I believe there should also be a DIR (Directional) wire that, if cut, or for 15/30 seconds if pulsed, removes the direction requirement, allowing you to go through both directions without the turnstile resisting your efforts.
> > ```
> 
> This would be funny, but without the danger of a power wire, wouldn't this make jailbreaks incredibly easy with just a screwdriver and multitool (and almost easy with a wirecutter if you don't hit the bolt wire)?
> 
> > ```
> > 5. I am not sure if this is currently the case, but the prototypes for the genpop enter/exit turnstiles should be updated to have higher security protection for the door panels, simular to vault doors. This means a plasteel plate and electrified grille inside the panel to make hacking it a tough job. If thats too much, the basic security, such as that in place for head of staff offices (just a steel plate welded inside) would do too.
> > ```
> 
> I can add high security versions, but I believe the standard versions should be the ones mapped (unless you want this to be a mitigation for my objection to point 4.)
> 
> > ```
> > 6. IMO the need for power should be removed entirely, that will cause issues with the primarily application being use in perma. Remove the appropriate wires too.
> > ```
> 
> I can remove power requirements and the power wires.
> 
> > You can replace it with a lock wire that suspends access requirements so that if you can hack it, you can still force it open just like regular doors, just for this it is by removing access instead of removing power.
> 
> How about a `SOL` wire which would represent the solenoid holding the spinner in place, a cut or pulse on this wire would turn on the `SOL` wire indicator, representing an activation or partial activation of the solenoid, during which you could begin a do-after. The result of the do-after would be popup that says "You manage to force the turnstile mechanism!". In this state, next entry through the turnstile will be allowed regardless of access restrictions.
> 
> Some testing needs to be done to ensure in the case of say 3 people pushing against the turnstile while someone performs the do-after, if multiple people try to enter in the same tick it should prioritize the do-after actor.
> 
> Overall, I'm totally cool and open to changes here.
> 
> I was actually going to suggest rolling back everything except the simple addition of door electronics,
> 
> I just added all the other shit because it's so damn easy with ECS and it's cool.
> 
> @BramvanZijp thank you for the thoughtful changes

The EMAG interaction should 100% be removing the access requirement, maybe also removing directional.

The DIR wire is not just there for fun, its also so that if an asshat clown places them infront of sec facing inwards, you can hack them to get out.

There are prototypes for the turnstiles mapped for permas, being genpop enter and genpop exit. I think these should be slightly harder to hack to make escapes more challanging. Changing these prototypes to either have plasteel and rods, or just steel, would accomplish this. Since otherwise the new wirepanel could be abused to allow trivial escapes.

As for the SOL/Lock wire, do whatever you want there, maintainers may give more clear instructions.

Quantum-cross:
> The DIR wire is not just there for fun, its also so that if an asshat clown places them infront of sec facing inwards, you can hack them to get out, etc. Thats something you need to keep in mind: These can be built anywhere, used for anything.


Just to clarify what I was getting at here, is that prisoner IDs retain the Genpop Enter access until their sentence is up, so an accessible DIR wire with just a screwdriver + cutters/multitool would allow prison escape probably too easily. I understand your concern and now I need to think about it.

SpeltIncorrectyl:
People being able to build turnstiles with a bunch of sticks was crazy. Glad it's finally going to need an airlock circuit.

BramvanZijp:
> > The DIR wire is not just there for fun, its also so that if an asshat clown places them infront of sec facing inwards, you can hack them to get out, etc. Thats something you need to keep in mind: These can be built anywhere, used for anything.
> 
> Just to clarify what I was getting at here, is that prisoner IDs retain the Genpop Enter access until their sentence is up, so an accessible DIR wire with just a screwdriver + cutters/multitool would allow prison escape probably too easily. I understand your concern and now I need to think about it.

For starters, if they have a screwdriver + cutters/multitool, then security has already fucked up. But thats also why a steel plate is a good idea.

BramvanZijp:
> It would be interesting if a turnstile could "consume" an access on an ID card, so that the Genpop Enter and Genpop Exit accesses are one time use on turnstiles.

Problem is, you can trigger it but not walk through it. Furthermore, all maps have 2 exit turnstiles: One to reach the locker room, and one to exit it after you get your stuff. This part of the issue is actually described in the comments of [#37222]

Quantum-cross:
> Problem is, you can trigger it but not walk through it.

Then trigger the access deletion on the exit side

> Furthermore, all maps have 2 exit turnstiles: One to reach the locker room, and one to exit it after you get your stuff. This part of the issue is actually described in the comments of [#37222]

Then this feature can be enabled on the second exit turnstile. I'd have to think of a way to configure this but it should be possible.

Maybe a better idea is an access wiper that looks kind of like those anti-theft pylons or metal detectors on the correct sides of the turnstiles that will wipe a specified access when walking through it. Probably make it so that sec access is required to toggle it between Genpop Entry to Genpop Exit.

Anyway this is probably getting off topic, one thing at a time I suppose

Quantum-cross:
Here's my proposed design after hearing the discussions:

- Construction requires LV cable and Door electronics just like an airlock
- Deconstruction requires welder, crowbar, screwdriver, wirecutters, wrench (NO insuls for base turnstile)
- Turnstile can be hit with Authentication Disruptor
  - This works automatically because the turnstile has a AccessReader component
  - This will clear all accesses on the door electronics and also clear all the logs just like any airlock
  - Since airlocks bolt open as a tell that they may have been disrupted, "emergency access" style lights can be turned on until the door electronics has been removed and reinserted (fixed or not)
- Turnstiles can be Emagged
  - This will rotate the entry direction by 180 degrees, but not touch access because that's the auth disruptor's job
  - This interaction can be repeated
  - Yes this can be done manually by breaking the turnstile down to the frame and rotating it
  - This allows an actor to flip the direction of a turnstile quickly and quietly -- maybe they'll flip the genpop entrance so that they or someone else could exit through the entrance? Maybe flip it again to cover their tracks?
  - Look if someone has an emag or auth disruptor in genpop sec has a problem.
- Turnstiles do not require power to operate, despite the electronics and lights (maybe they're powered by generating electricity from the spinning 😄)
- Proposed wire panel
  - LOG
  - BOLT
  - BLIT
  - AIA
  - SOLN
  - 3 Dummy wires (to match 8 wires on airlock)
- SOLN (solenoid) will allow you force your way through a turnstile using minimal tools
  - When this wire is cut or pulsed will allow you to begin a long do-after
  - After this do-after is complete, a popup will appear saying something on the order of "You force the mechanism of the turnstile to momentarily let you through"
  - The actor of the do-after will be whitelisted to enter the turnstile for a brief amount of time (something like 3 seconds?).
  - After successfully entering the turnstile, this whitelist expires.
  - If the turnstile is not entered before the time expires a popup should appear "You lose your grip on the turnstile mechanism as it snaps back to the locked position"
  - The do-after can be shortened using a crowbar
  - The do-after can be lengthened if attempting to force the turnstile in the opposite direction
  - All of these interactions require tampering with the SOLN wire
  - It only requires you to ***begin*** the do-after while the SOLN wire is tampered with, the state of SOLN should not matter at the end of the do-after
- BOLT
  - Turnstiles should be able to be bolted.
  - Bolt indicator will turn on (unless BLIT is cut)
  - Can be bolted with wire, door remote, or AI interaction
  - An attempt to enter a bolted turnstile will result in the "resists your efforts" message
  - A turnstile force attempt via SOLN should fail at the end of the do-after if the turnstile is bolted at the end.
  - A turnstile force attempt via SOLN should fail to start if bolted
  - Players that enter a turnstile will always be able to exit even if it is bolted mid-traversal
  - If a player dragging another player enters a turnstile, and the turnstile bolts before the dragged player enters the turnstile, the dragged entity should be allowed through.
- Higher security variants
  - I propose only three options for beefing the security of the turnstile, I don't think we need super max turnstiles with two layers of plasteel etc etc.
  - Medium Security
    - slap some steel on it
    - weld it
    - call it a day
  - Stop Resisting variant
    - slap some steel rods on it
    - will electrocute if the turnstile has LV cable to it.
    - turnstile will not have the "extension cable" component so the LV wire must be directly under the turnstile similar to your standard electrified grill
    - This allows bypass without insuls if for instance the power goes out for whatever reason
  - Perma Variant
    - Medium security variant + the electrified grill


@EmoGarbage404 can you review this design?

ilovehans10:
Honestly including an emergency access option might be handy for emergencies. If you ask me, emergency access would still lock the direction, but would allow any ID through. This would mean that if the brig gets spaced, the warden could make it easier for engineers to enter the brig.

Quantum-cross:
> Honestly including an emergency access option might be handy for emergencies. If you ask me, emergency access would still lock the direction, but would allow any ID through. This would mean that if the brig gets spaced, the warden could make it easier for engineers to enter the brig.

I would have liked to implement emergency access, allowing both directions and no access check.

Unfortunately, emergency access mode is coupled into the airlock system right now and would probably be another PR.

BramvanZijp:
@Quantum-cross Want me to mark this as waiting on that PR?

Quantum-cross:
I'm going to atomize this PR and the pry/bolt refactors more to make this more palatable. It's been on my list of things to do but @ScarKy0 requested it on discord which is another motivational push.

Quantum-cross:
I just wanna ask to keep this PR open (and drafted) until all the features have been atomized into other PRs. Thank you!

PR #31392, opened 2024-08-24
decouple ItemToggle from PowerCellDraw

deltanedas:
## About the PR
idk why i did this but now it actually works without it

## Why / Balance
fixes #31377 

## Technical details
(uid, toggle) -> uid
added ToggleCellDraw which everything but borg has (borg uses ItemToggle for toggling access, so it would lose access if power ran out)

## Requirements
- [X] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes
no

**Changelog**
:cl:
- fix: Fixed borgs losing access when they run out of power.

PR #37364, opened 2025-05-12
fix: don't retroactively drain disabled batteries

perryprog:
## About the PR

If something that used PowerCellDraw temporarily disabled said draw, once it became re-enabled the system would play catch-up trying to drain the battery for all the time since the component was disabled. This fixes that by setting our next update time to be in the future rather than possibly the past.

## Why / Balance

I like playing borg on long rounds.

## Technical details

When `ToggleCellDrawSystem` was extracted in #31392, this issue was worked around for entities with both `ToggleCellDraw` and `PowerCellDraw` by adding a `QueueUpdate` method (which existed in a different form before that PR). Since `NextUpdateTime` should never need changed now, I've removed this, as well as the MapInit that was calling it too.

The issue I primarily was trying to address is that if a borg is being played, but has its mind remove causing its power draw to be disabled, upon return the battery—even if it's a microreactor—will be drained as its power draw is run on every tick, until `Update` catches back up.

I've tested this and everything seems to still work as expected. There's also an alternate solution of moving the `QueueUpdate` call to be within `SetDrawEnabled` but I feel like this makes more sense.

## Requirements

## Breaking changes

- `SharedPowerCellSystem.QueueUpdate` is removed and no longer needs to be called.

**Changelog**

🆑 
- fix: Borgs in certain circumstances will no-longer draw much more power than they should.

### DIFF
--- a/Content.Server/PowerCell/PowerCellSystem.Draw.cs
+++ b/Content.Server/PowerCell/PowerCellSystem.Draw.cs
@@ -20,10 +20,11 @@ public sealed partial class PowerCellSystem
             if (!comp.Enabled)
                 continue;
 
-            if (Timing.CurTime < comp.NextUpdateTime)
+            var now = Timing.CurTime;
+            if (now < comp.NextUpdateTime)
                 continue;
 
-            comp.NextUpdateTime += comp.Delay;
+            comp.NextUpdateTime = now + comp.Delay;
 
             if (!TryGetBatteryFromSlot(uid, out var batteryEnt, out var battery, slot))
                 continue;
--- a/Content.Server/PowerCell/PowerCellSystem.Draw.cs
+++ b/Content.Server/PowerCell/PowerCellSystem.Draw.cs
@@ -20,11 +20,10 @@ public sealed partial class PowerCellSystem
             if (!comp.Enabled)
                 continue;
 
-            var now = Timing.CurTime;
-            if (now < comp.NextUpdateTime)
+            if (Timing.CurTime < comp.NextUpdateTime)
                 continue;
 
-            comp.NextUpdateTime = now + comp.Delay;
+            comp.NextUpdateTime += comp.Delay;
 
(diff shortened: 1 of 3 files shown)


Content.Server/PowerCell/PowerCellSystem.Draw.cs
```diff
@@ -20,10 +20,11 @@ public override void Update(float frameTime)
             if (!comp.Enabled)
                 continue;
 
-            if (Timing.CurTime < comp.NextUpdateTime)
+            var now = Timing.CurTime;
+            if (now < comp.NextUpdateTime)
                 continue;
 
-            comp.NextUpdateTime += comp.Delay;
+            comp.NextUpdateTime = now + comp.Delay;
```
slarticodefast, line 27: this will eat the remainder every update and becomes inprecise
see my writeup on that here https://hackmd.io/@Slart/S1hsoGFm1l
instead this should be updated once when re-enabled

slarticodefast (changes requested):

Content.Server/PowerCell/PowerCellSystem.Draw.cs
perryprog: Ah, makes sense; I saw this style used elsewhere so I assumed it was fine, guh. For my own understanding—isn't this only adding longs to longs? I'm not sure I see where the imprecision would be coming from.

Super fantastic resource by the way—link from dev docs when? :>

slarticodefast: Yeah, people copy paste this style everywhere and I'm fighting hard to contain it.

The imprecision happens because CurTime will be larger than NextUpdateTime in each step. So the actual update rate will be lower than intended. This can cause a bunch of bugs over time. For example we used to have singulooses because of exactly this.

My guide will be uploaded to the docs once I finished writing it.

slarticodefast (approved):
Looks good to me, thanks for the bugfix!

PR #36022, opened 2025-03-23
Remove all trailing whitespaces in .yml files

CrafterKolyan:
## About the PR
Removes all trailing whitespaces in .yml files

## Why / Balance
Makes writing code in Space Station 14 repo easier with use of autoformatters as it allows to not get a lot of random diffs related to autoformatting.

## Technical details
Done using:
```bash
find -name "*.yml" | xargs sed -i -e "s/[[:space:]]\+$//"
```

## Requirements
- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

PR #37367, opened 2025-05-12
Add trailing whitespace check workflow

slarticodefast:
## About the PR
Adds a new workflow checking for trailing whitespace in text files, similar to the CRLF check.
This workflow only checks files that have been edited in the corresponding PR, so we don't get one giant merge conflict but can clean it up over time without introducing new cases.

Do not merge until I did some more testing to make sure it works as intended.

## Why / Balance
A lot of contributors add trailing whitespace by accident because they don't use an IDE or use web edits. This usually automatically gets fixed once someone edits the file using an IDE, but each time it messes up the git history and potentially causes merge conflicts.
See https://github.com/space-wizards/space-station-14/pull/36022

## Technical details
Uses https://github.com/marketplace/actions/changed-files
and a simple bash script.

## Requirements

## Breaking changes
none

**Changelog**
no formating errors no fun

### DIFF
--- a/.github/workflows/check-trailing-whitespace.yml
+++ b/.github/workflows/check-trailing-whitespace.yml
@@ -1,0 +1,40 @@
+﻿name: Trailing Whitespace Check
+
+on:
+  pull_request:
+    types: [ opened, reopened, synchronize, ready_for_review ]
+
+jobs:
+  build:
+    name: Trailing Whitespace Check
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4.2.2
+    - name: Get changed text files
+      id: changed-files
+      uses: tj-actions/changed-files@v46.0.5
+       with:
+         files: |
+           **.cs
+           **.yml
+           **.swsl
+           **.json
+           **.py
+    - name: Check for trailing whitespace
+      env:
+        ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+      run: |
+        has_trailing_whitespace=0
+        for file in ${ALL_CHANGED_FILES}; do
+          echo "Checking $file"
+          if grep -qP '[ \t]+$' "$file"; then
+            echo "::error file=$file::Trailing whitespace found"
+            has_trailing_whitespace=1
+          fi
+        done
+
+        if [ "$has_trailing_whitespace" -eq 1 ]; then
+          echo "Trailing whitespace detected. We recommend using an IDE to prevent this from happening."
+          exit 1
+        fi


kosticia:
So ironically that whitespace test doesn't have newline symbol at the end :trollface:

slarticodefast:
> So ironically that whitespace test doesn't have newline symbol at the end :trollface:

The dollar sign in the grep search includes that. And we have a separate test for checking CRLF new lines.

kosticia:
> The dollar sign in the grep search includes that. And we have a separate test for checking CRLF new lines.

I mean...

Winkarst-cpu:
> I added an EOF new line check as well. We are finally free from all the messed up yaml git diffs.

Machines will steal your job!!!

kosticia:
> I added an EOF new line check as well. We are finally free from all the messed up yaml git diffs.

heroic pr.

ciaran:
Possible to sneak in `dotnet format --verify-no-changes` here perhaps?

ciaran:
Also, I think there's an argument for any formatting checks being added to pre-commit hooks as well to avoid people potentially making a week’s worth of commits and then realising when they go to make a PR that all their formatting was wrong.

slarticodefast:
> Possible to sneak in dotnet format --verify-no-changes here perhaps?
Is it possible to apply that to only the changed files in the PR? And would it work with non-C# files?
Also might be a little too much to completely enforce. In some rare cases breaking formatting rules is more readable.

>Also, I think there's an argument for any formatting checks being added to pre-commit hooks as well to avoid people potentially making a week’s worth of commits and then realising when they go to make a PR that all their formatting was wrong.

Fixing any whitespace errors takes like a minute to do. And if they use an IDE like they should it would be literally impossible to make any such errors. This only comes up with web edits or contributors using notepad to edit yaml.

ciaran:
> Is it possible to apply that to only the changed files in the PR? And would it work with non-C# files?

I’ve done some testing; you can specify files with `--include` and it will ignore any files that aren’t part of the project, so that shouldn‘t be a problem.

E.g. I tried the following:

```
dotnet format style --no-restore --verify-no-changes --include=Resources/Prototypes/Reagents/medicine.yml Content.Server/Medical/HealthAnalyzerSystem.cs
```

This will validate the formatting of `HealthAnalyzerSystem.cs` and ignore the preceding Yaml file. Non-zero exit means a formatting error was found.

There is also `dotnet format whitespace` which is much quicker since it doesn’t load the actual project, but checks whitespace formatting only.

> Also might be a little too much to completely enforce. In some rare cases breaking formatting rules is more readable.

You can disable formatting for a block like this:

```
#pragma warning disable format
…
#pragma warning enable format
```

Unfortunately there doesn’t seem to be a single-line/statement option.

ciaran:
> Do you think it's realistic to check yaml indentation conventions? Because that would be incredible

I did look into options for this, IIRC `yamlfmt` can be configured to match the existing style with the exception of wrapping inline lists with a space (`foo: [ 1 ]`)

slarticodefast:
> Do you think it's realistic to check yaml indentation conventions? Because that would be incredible

Ideally we would use an existing formatting checker for that.
But the simple alternative would be:
- count the number of spaces at the start of each line
- if it increases, check that it increases by 2

Or are there any edge cases I'm missing?

And it probably should be part of the yaml linter, but then we would have to fix it everywhere at once rather than on a file by file basis.

Errant-4:
I guess either way, that should be a separate PR. Forcing that level of changes when someone edits some old yaml trying to adjust just 1 line is certainly a more significant change than this one

Errant-4 (approved):

little-meow-meow:
Is there a specific reason this uses `grep -qP` rather than `grep -nP`? It would be nice to see the few affected lines from the failed workflow log.

PR #37375, opened 2025-05-12
Booze storage and chef's closet now contain paper

Hitlinemoss:
## About the PR
The booze storage and the chef's closet now each contain 3 sheets of blank paper.

## Why / Balance
Paper is useful for the bar/kitchen to make menus, ingredient requests for botany, etc. Giving them a few sheets of paper in their lockers means the bartender/chef doesn't (usually) need to leave their workplaces to go hunt for paper in the library or something.

## Technical details
Adjusted the values in Resources\Prototypes\Catalog\Fills\Lockers\service.yml.

## Media

MyServer - Space Station 14 5_12_2025 12_54_31 PM
MyServer - Space Station 14 5_12_2025 12_54_46 PM

## Requirements

## Breaking changes

N/A

**Changelog**

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/service.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/service.yml
@@ -21,6 +21,8 @@
         amount: 2
       - id: RagItem
         amount: 2
+      - id: Paper
+        amount: 3
 
 #- type: entity
 #  id: LockerFormalFilled
@@ -39,6 +41,8 @@
       - id: BoxMousetrap
         amount: 2
       - id: SprayBottleWater
+      - id: Paper
+        amount: 3
       - id: ReagentContainerFlour
         amount: 2
       - id: FoodBoxCloth


Hitlinemoss:
> I personally just abuse the nanotask when I need paper. but this is a step in the right direction.

NanoTask *kind* of works for printing paper on demand, but then the paper is labelled "NanoTask item" so that makes it harder to identify as a menu.

Were I a more skilled programmer, I'd want to add a new "menu paper" item so that it's obvious at a glance that the random paper on the kitchen counter is actually a menu, but then you'd not only need to program the item itself but also program places where you can actually find/make it.

slarticodefast (approved):
Alternatively it could be mapped or come in a folder, since loose sheets of paper are a little unusual, but should be fine.

slarticodefast:
Ok, this was brought up for a quick maintainer discussion and the job lockers should only contain immediately applicable job equipment, loose sheets of paper feel a little out of place here. And it is easy to grab from the library or to print some at a fax machine.

EmoGarbage404:
Thanks for the PR.

After some discussion, it was felt that lockers should be relegated strictly to the most relevant job-related equipment. Almost any job can arguably utilize some level of common goods. However, it's best for those things to be kept in common areas or just mapped directly in things like paper trays. 

As such, we're opting to close this PR. Feel free to contribute again in the future.

Hitlinemoss:
> Thanks for the PR.
> 
> After some discussion, it was felt that lockers should be relegated strictly to the most relevant job-related equipment. Almost any job can arguably utilize some level of common goods. However, it's best for those things to be kept in common areas or just mapped directly in things like paper trays.
> 
> As such, we're opting to close this PR. Feel free to contribute again in the future.

I agree with this sentiment and the closing of this PR, but now I'm curious about something. The chef's closet contains an emergency crowbar: why is that? As far as I'm aware the chef doesn't need to use that for anything related to their job specifically. Is that something I should make a PR to remove?

PR #37381, opened 2025-05-12
Cleanup warnings in `HumanoidAppearanceSystem`

Tayrtahn:
## About the PR

Fixed 18 warnings in `HumanoidAppearanceSystem`.

## Why / Balance

## Technical details

Replaced `SpriteComponent` methods with `SpriteSystem` methods. Many private methods signatures were changed to allow the uid to be passed in.

Due to a change in behavior from `SpriteComponent.LayerMapReserveBlank` to `SpriteSystem.LayerMapReserve`, a small amount of logic needed to be changed as well. If `LayerMapReserveBlank` tries to reserve a layer that already exists, it returns the ID of the existing layer, whereas `LayerMapReserve` throws an exception. To resolve this, `LayerMapTryGet` was used to mimic the previous behavior.

## Media

Humanoids still appearing:
Screenshot 2025-05-12 at 2 37 41 PM
Screenshot 2025-05-12 at 2 37 52 PM
Screenshot 2025-05-12 at 2 38 01 PM

## Requirements


Content.Client/Humanoid/HumanoidAppearanceSystem.cs
```diff
@@ -65,34 +74,38 @@ private void UpdateLayers(HumanoidAppearanceComponent component, SpriteComponent
...
         bool sexMorph = false,
         Color? color = null)
     {
-        var layerIndex = sprite.LayerMapReserveBlank(key);
+        var component = entity.Comp1;
+        var sprite = entity.Comp2;
+
+        var layerIndex = _sprite.LayerMapTryGet((entity, sprite), key, out var index, false)
+            ? index
+            : _sprite.LayerMapReserve((entity, sprite), key);
```
slarticodefast, line 108: Just to make sure why this is being changed, `LayerMapReserveBlank` and `LayerMapReserve` are not completely identical, right? The latter seems to cause an exception if the layer already exists. Maybe we should ask @ElectroJr if that API change was on purpose.

ElectroJr: I don't remember if there was a specific reason for the change, but the old component-method behaviour makes more sense, so I'll open an engine PR to make it not throw an exception.

Tayrtahn: Sounds good. I've changed the code back to the old logic (with the new `SpriteSystem` method calls). This PR will need to wait for the next engine update.

slarticodefast (approved):

PR #37386, opened 2025-05-12
Dragons no longer round remove cyborgs when devouring them

Centronias:
## About the PR

Dragons no longer round remove cyborgs when devouring them, because now the cyborg body gets gibbed and the brain goes in the dragon

## Why / Balance

The number of times I have been round removed because a dragon ate me is greater than four, and my poor cyborgsonas deserve better.

## Technical details

More or less just added a special case to devouring for borgs.

## Media

Small, so I didn't record anything, but lemme know if you want proof and I can provide it.

## Requirements

## Breaking changes

None.

**Changelog**

:cl:
- tweak: Cyborgs are no longer literally deleted when devoured by a dragon, round-removing the cyborg player. Instead, their chassis is destroyed and their brain goes in the dragon's stomach.

### DIFF
--- a/Content.Server/Devour/DevourSystem.cs
+++ b/Content.Server/Devour/DevourSystem.cs
@@ -4,12 +4,14 @@ using Content.Shared.Chemistry.Components;
 using Content.Shared.Devour;
 using Content.Shared.Devour.Components;
 using Content.Shared.Humanoid;
+using Content.Shared.Silicons.Borgs.Components;
 
 namespace Content.Server.Devour;
 
 public sealed class DevourSystem : SharedDevourSystem
 {
     [Dependency] private readonly BloodstreamSystem _bloodstreamSystem = default!;
+    [Dependency] private readonly BodySystem _body = default!;
 
     public override void Initialize()
     {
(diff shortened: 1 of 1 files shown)


Content.Server/Devour/DevourSystem.cs
```diff
@@ -19,31 +21,55 @@ public override void Initialize()
...
-            QueueDel(args.Args.Target.Value);
+            // Humanoids go into the stomach entirely
+            if (HasComp<HumanoidAppearanceComponent>(target))
+            {
+                ContainerSystem.Insert(target, devourer.Stomach);
+                return;
+            }
+
+            // Borgs get gibbed and their brain goes in the stomach
+            if (TryComp<BorgChassisComponent>(target, out var borg))
```
slarticodefast, line 59: I really don't like hardcoding the component into the system like this.

Content.Server/Devour/DevourSystem.cs
```diff
@@ -19,31 +21,55 @@ public override void Initialize()
...
+        _audioSystem.PlayPvs(entity.Comp.SoundDevour, entity);
+    }
+
+    private void InsertEntityToDevourerStomachOrDelete(DevourerComponent devourer, EntityUid target)
+    {
+        if (devourer.ShouldStoreDevoured)
         {
-            QueueDel(args.Args.Target.Value);
+            // Humanoids go into the stomach entirely
+            if (HasComp<HumanoidAppearanceComponent>(target))
```
slarticodefast, line 52: wouldn't this delete all non-humanoid mobs being devoured?

slarticodefast (changes requested):

Content.Server/Devour/DevourSystem.cs
Centronias: It gets to live with its twin hard coding the Humanoid Appearance Component :)

But seriously, is there any better way to do this?

I was thinking maybe just mobs in general could go in the stomach, but that'd mean a downed borg could be in the stomach yelling on the radio while being carried around.

Content.Server/Devour/DevourSystem.cs
Centronias: If the devourer doesn't store things, the entity is always deleted.

Otherwise, entities with Humanoid Appearance always go in the stomach.
Otherwise, borgs get gibbed and the brain goes in the stomach. (This is the only addition)
Otherwise the entity is destroyed.

The logic before was very happy to delete non humanoid mobs, which is exactly what happened to cyborgs.

slarticodefast:
Superceded by https://github.com/space-wizards/space-station-14/pull/38087

PR #37391, opened 2025-05-12
Attributions checking in CI for PRs

SaphireLattice:
## About the PR
Added a script to check for attribution issues that can run in CI and fail the build if any are found

## Why / Balance
A lot of the times people forget to add appropriate `attributions.yml` entry when contributing assets. This should prevent that.

## Technical details
Added the `Tools/rga_check.py`, which checks that all the assets have appropriate `attributions.yml` entry, and that said entry passes source checks as defined in `Resources/licensing.yml`. The script needs to run in the root dir of the repository. There's a lot of flags that control the verbosity of it, too.

The info about the checks done and what they represent can be seen in the comments for `ValidationFailure` enum. Feedback about unclear wording or needing a better description will be VERY appreciated, especially for `validation_failure_names` list!

The script when running in CI will only output info relevant to the files changed by the PR. So in attributions files it will only validate entries of files that have been added/changed. However if the changeset is ONLY modifying attributions files, the script will validate all the asset files in same directory.

Still on todo/wishlist: making the source check properly aware of github users existing, and handling the relevant URLs appropriately. Another wishlist thing is making it aware of SS13 repos, and having it parse paths out, so that we can check if paths are known to be valid or invalid according to another configurable list. The asset ecosystem of SS13 is complex, and SS14 is inheriting a lot of this complexity due to using assets from these repositories, so I think we would benefit a lot from this.

Additionally source links pointing at SS13 repositories are inconsistent, and can point to either a repo overall, a commit, or a specific file blob. Having the script also check for this being in some specific format (I am leaning to commit specific file blob URLs) would be nice.

The source list is *INCOMPLETE* and needs a lot of help in making sure that it will not annoy ever single person who contributes some asset. We don't want them to dive into the `licensing.yml` file unless it's particularly necessary.

Needs someone to test it with the lists of changed files from PRs that modified assets.

## Requirements

## Breaking changes
Added CI script for checking attributions, please make sure `Resources/licensing.yml` is set up appropriately

### DIFF
--- a/.github/workflows/validate-rgas.yml
+++ b/.github/workflows/validate-rgas.yml
@@ -33,12 +33,7 @@ jobs:
     - name: Get changed files
       uses: tj-actions/changed-files@v45
       id: changed-files
-    - name: List all changed files
-      env:
-        ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
-      run: |
-        for file in ${ALL_CHANGED_FILES}; do
-          echo "$file" >> ${{ runner.temp }}/validate-rga-filter.txt
-        done
+      with:
+        write_output_files: true
     - name: Validate RGA attribution
-      run: python3 Tools/rga_check.py -i "" ${{ runner.temp }}/validate-rga-filter.txt
+      run: python3 Tools/rga_check.py -i "" cat .github/outputs/all_changed_and_modified_files.txt
--- a/.github/workflows/validate-rgas.yml
+++ b/.github/workflows/validate-rgas.yml
@@ -1,14 +1,17 @@
-name: RGA schema validator
+name: RGA schema and attribution validator
 on:
   push:
     branches: [ master, staging, stable ]
   merge_group:
   pull_request:
-    types: [ opened, reopened, synchronize, ready_for_review ]
+    paths:
+      - 'Resources/**'
+      - 'Tools/rga_check.py'
+      - 'RobustToolbox/**'
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/validate-rgas.yml
```diff
@@ -17,9 +20,21 @@ jobs:
...
       with:
         schema: RobustToolbox/Schemas/rga.yml
         path_pattern: .*attributions.ya?ml$
         validators_path: RobustToolbox/Schemas/rga_validators.py
         validators_requirements: RobustToolbox/Schemas/rga_requirements.txt
+    - name: Install Python dependencies
+      run: |
+        pip3 install --ignore-installed --user pyyaml
+    - name: Get changed files
+      uses: tj-actions/changed-files@v45
```
slarticodefast: I tried using this tool for the Trailing Whitespace check. As you can see this sometimes fails for no reason and gives you a ton of files that have not been changed. I did not confirm this, but my suspicion is that it either happened when a branch is very old and the search depth for the branching commit in the script was not high enough, or when a contributor did not fork directly from the wizden repo. I had to revert that check again because of this.
https://github.com/space-wizards/space-station-14/pull/38179
So you will either need to make sure this works, or we make this a manual workflow to prevent heisentests.

SaphireLattice:
Alright fuck it, have the script without the CI stuff, woo. That can be written later, no?

VerinSenpai:
So question. Is it typical to only want to check changed files, or is that a time saving measure?

PR #28941, opened 2024-06-13
Geras fixes/tweaks 2: Now with polymorph changes

Vermidia:
## About the PR

- [X] Update guidebook to be accurate to slime people as they currently are
- [X] Geras and normal slimes are faster/slower than humanoids (respectively)
- [X] Slime people do higher damage but slower, Geras do it faster but weaker (both average 5 DPS)
- [X] Polymorph correctly handles zombification (Zombie -> Zombie and back for both parent and child)
  - [X] Geras can turn into Zombified slime people and vice versa while staying zombified
- [X] Polymorph can now modify bloodlevel and transfer chemicals from parent to child and child to parent
- [X] Polymorph persists identity though child if transfer name is on
  - [X] Polymorph makes sure criminal record icons also are correct through polymorphs
- [X] Firestacks & bleeding persist through polymorphs
- [X] Temperature persists through polymorphs
- [X] Polymorphed Event for components to subscribe to
  - [X] FollowerComponent no longer sends the followers to nullspace
- [x] Add description when a slime person contains something (same for Geras)
- [ ] ~~(Will Try) Add doafter when trying to take something out of a slime person/geras (and you aren't said slime person)~~ I tried
- [x] Slime inventory and Geras inventory are synced
- [X] Increase slime people's blood reduction rate slightly
  - [X] Make Geras passive damage & bloodloss healing mirror slimes more

~~Waiting for #27417 before attempting synced inventories and other modifications.~~ The guidebook's mention of grille movement has been removed, as it isn't implemented and it'd be too much trouble to do so for how little it'd actually do.

## Why / Balance

Fixes #27418, Fixes #27261, Fixes #27272 (probably), Fixes #27250, Fixes #27239, Fixes #29011, Fixes #28507, Fixes #28781, Fixes #27485, Fixes #28778

Lots of these were bugs and in general the polymorph system will need to transfer lots of information between entities, so this is a start to modifying parts of it to handle some common cases (chemicals, blood, zombification).

## Technical details

Changes are done in both the PolymorphSystem for Bloodstream and Zombification transfers. Parts were modified so normal slimes and geras have legs that properly modify their speeds.

## Media

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

https://github.com/space-wizards/space-station-14/assets/32827189/c5b89645-4c5b-4e36-b6ab-bd24ed64b52a

[...]

PR #37401, opened 2025-05-13
Polymorph Improvements

SolStar2:
## About the PR
Uses polymorphed event to transfer most needed components.

always
- Injected anoms die and drop core to prevent continued point generation from paused anomaly
- Bible user transfers
- Kill tracking transfers
- Temperature transfers

Transfer Status Effects
- Eye damage is transferred
- Status Effects transfer

Damage Transfer
- flame stacks transfer
Transfer Bloodstream
- bloodstream and reagents transfer

Identity
- Name transfers
- Criminal records transfer
- DNA transfers

Vow
- Mime vow transfers

Code ported from: https://github.com/space-wizards/space-station-14/pull/28941

## Why / Balance
Polymorphs are becoming more frequent with wizard and downstream systems. This allows polymorphs to transfer components such as zombification that could persist between forms.

Fixes https://github.com/DeltaV-Station/Delta-v/issues/3522 , https://github.com/DeltaV-Station/Delta-v/issues/3603

## Technical details
More changes to come as https://github.com/space-wizards/RobustToolbox/pull/5854 gets merged.
An extra anomaly core is spawned when an entity with an injected anomaly is polymorphed.

## Requirements

## Breaking changes

**Changelog**

:cl: SolStar, Vermidia
- tweak: Injected anomalies die when an injected person is polymorphed.
- tweak: Identity and DNA transferred along with name if polymorph transfers name.
- tweak: Flame stacks now transfer if polymorph transfers damage.
- tweak: Polymorphs may now transfer bloodstream.
- tweak: Polymorphs may now transfer status effects.
- tweak: Polymorphs may now transfer mime vows.
- fix: Tempature persists between polymorphs
- fix: Bible users retain their ability to use bibles through polymorphs.

### DIFF
--- a/Content.Server/Abilities/Mime/MimePowersSystem.cs
+++ b/Content.Server/Abilities/Mime/MimePowersSystem.cs
@@ -7,6 +7,7 @@ using Content.Shared.Coordinates.Helpers;
 using Content.Shared.Maps;
 using Content.Shared.Paper;
 using Content.Shared.Physics;
+using Content.Shared.Polymorph;
 using Robust.Shared.Containers;
 using Robust.Shared.Map;
 using Robust.Shared.Timing;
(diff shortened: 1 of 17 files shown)


NoElkaTheGod:
Killing an anomaly by polymorphing (especially as a kitsune or something similar, who can do it at any time) seems kinda weird. Maybe it should spawn an identical non-infection anomaly at the same location?

slarticodefast:
Hmm, this does have some significant overlap with the CloningSystem, which allows you to copy over components based on a yaml whitelist, instead of having to add a subscription for every single one of them like you do here. That is how we are handling changeling transformation in this PR https://github.com/space-wizards/space-station-14/pull/34002
It seems to me that the intention behind polymorphing is to temporarily change you into a completely new mob rather than preserving as many components as possible. And using it for transformations like this will introduce quite a range of bugs because you are a new entity after being polymorphed, which will influence things like maroon objectives, the crew manifest and so on. If polymorphing is used like that it may need an ever-growing amount of bugfix PRs like this one.

SolStar2:
The issue in this case is that because you transform into a fox you don't have humanoidappearance. we just want to apply the sex without needing humanoid apperance.

SolStar2:
That was how I originally was doing things but ran into issues where many things are not as simple as copying over components. I figured it makes more sense to use events and have each system only be responsible for transferring of their own state from one entity to another.

SolStar2:
Things that people may not want to transfer are tied to options in the polymorph configuration which is sent in the event.

slarticodefast:
I asked some other maintainers for an opinion and the sentiment was that this is not the right approach for fixing this problem as polymorphing is not made with such a usecase in mind.
The general rule is avoid specific systems (polymorph) infecting general systems (temperaturesystem), which means the eventbus is the wrong call here.

PR #37405, opened 2025-05-13
Changing sex now properly updates sounds.

SolStar2:
## About the PR
Load sounds based on new sex in event.

## Why / Balance
Simple small fix to help downstream.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Speech/EntitySystems/VocalSystem.cs
+++ b/Content.Server/Speech/EntitySystems/VocalSystem.cs
@@ -49,7 +49,7 @@ public sealed class VocalSystem : EntitySystem
 
     private void OnSexChanged(EntityUid uid, VocalComponent component, SexChangedEvent args)
     {
-        LoadSounds(uid, component);
+        LoadSounds(uid, component, args.NewSex);
     }
 
     private void OnEmote(EntityUid uid, VocalComponent component, ref EmoteEvent args)


beck-thompson:
Could you link the PR that this will help fix?

SolStar2:
https://github.com/space-wizards/space-station-14/pull/37405/files

SolStar2:
https://github.com/DeltaV-Station/Delta-v/pull/3725

slarticodefast:
Hey, I'm trying to figure out why the new trailing whitespace check fails for your PR. Apparently it complains about your repository being a shallow clone and fails to get the changed files correctly.
grafik
How did you clone your repository? Did you make any changes to the master branch?

Content.Server/Speech/EntitySystems/VocalSystem.cs
```diff
@@ -49,7 +49,7 @@ private void OnShutdown(EntityUid uid, VocalComponent component, ComponentShutdo
 
     private void OnSexChanged(EntityUid uid, VocalComponent component, SexChangedEvent args)
     {
-        LoadSounds(uid, component);
+        LoadSounds(uid, component, args.NewSex);
```
slarticodefast, line 52: If set to null this should already use the value in the `HumanoidAppearanceComponent` which is set before the event is even raised.
I think you just have to use the correct `HumanoidAppearanceSystem.SetSex` API method in your downstream PR to male this work correctly instead of raising this event manually.
That will also ensure markings are applied correctly and the changed values are dirtied.

slarticodefast (changes requested):

Content.Server/Speech/EntitySystems/VocalSystem.cs
SolStar2: I wanted to be able to apply a sex without requiring humanoid appearance.

slarticodefast: Ok, that does make sense.

slarticodefast (approved):

SolStar2:
let me try rebasing and force pushing

PR #37410, opened 2025-05-13
Add Nuclear explosion type [Redo]

B-Kirill:
## About the PR
_This is the second attempt due to problems with the old branch and bad code in general_

Adds new `Nuclear` type of explosion, which is now used by a nuclear warhead. It deletes some objects (see the technical part).
Unlike the default one, it additionally deals 2 radiation damage:
```
  damagePerIntensity:
    types:
      Heat: 5
      Blunt: 5
      Piercing: 5
      Radiation: 2
```

## Why / Balance

- It is quite logical that a nuclear explosion should cause radiation damage.
- Removes a large number of objects that the default type of explosion cannot handle. A nuclear explosion ends the round in 99.99% of cases, so the balance will not be affected by removing items. But the server will not lag so much, especially if someone decides to fly through the remnants of the station on a shuttle.

## Technical details
The deletion is based on the matching of the components. 

A white and black list of components has been added to the yml file (they can also be added to other types of explosions and everything will work fine.):
```
- type: explosion
  id: Nuclear
  deleteComponents:
    - "Item"
    - "Cable"
    - "DisposalTube"
    - "WallMount"
    - "GasMiner"
  blacklistedComponents: []
```

String component IDs (e.g. "Item") are converted to `Type` via `IComponentFactory`. This also eliminates invalid components derived from yml, which is a fork-friendly method.

**Caching:**
On first use, component lists (`DeleteComponents`, `BlacklistedComponents`) are cached into `HashSet<Type>`, eliminating repeated string conversions. Cache is also marked as `[NonSerialized]`.

**Explosion Processing:**
- Entities with `BlacklistedComponents` are skipped (but they are still can take damage from the explosion).
- Entities with `DeleteComponents` are deleted if not in a container.
- The processed entities are added to a separate `HashSet`, which does not allow them to be processed again, which prevents errors and improves performance.

In terms of performance, of course, the new checks have an impact, but they do not take up more than 50% of the entity processing time. Also, the optimization benefit gained from removing such a huge number of objects on the map will significantly exceed the disadvantages of some additional load on the server during the explosion.

## Media
Default:
image

Nuclear:
image

## Requirements

**Changelog**
:cl:
- add: Nuclear explosion now destroys items, cables, pipes, and the like.

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
+++ b/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
@@ -12,6 +12,7 @@ using Content.Shared.Maps;
 using Content.Shared.Physics;
 using Content.Shared.Projectiles;
 using Content.Shared.Tag;
+using Robust.Server.Containers;
 using Robust.Shared.Map;
 using Robust.Shared.Map.Components;
 using Robust.Shared.Physics;
@@ -28,6 +29,7 @@ namespace Content.Server.Explosion.EntitySystems;
 public sealed partial class ExplosionSystem
 {
     [Dependency] private readonly FlammableSystem _flammableSystem = default!;
+    [Dependency] private readonly ContainerSystem _container = default!;
 
     /// <summary>
     ///     Used to limit explosion processing time. See <see cref="MaxProcessingTime"/>.
@@ -67,6 +69,8 @@ public sealed partial class ExplosionSystem
 
     private List<EntityUid> _anchored = new();
 
+    private HashSet<EntityUid> _processedForDeletion = new();
+
     private void OnMapRemoved(MapRemovedEvent ev)
     {
         // If a map was deleted, check the explosion currently being processed belongs to that map.
@@ -84,6 +88,8 @@ public sealed partial class ExplosionSystem
     /// </summary>
     public override void Update(float frameTime)
     {
+        _processedForDeletion.Clear();
+
         if (_activeExplosion == null && _explosionQueue.Count == 0)
             // nothing to do
             return;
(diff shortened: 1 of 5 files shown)


B-Kirill:
Um.. it's a pretty weird error..

EmoGarbage404 (changes requested):
I don't see a reason why this needs to be a separate explosion type. The performance overhead from new types is very noteworthy (we already have a few we ought to remove) and this seems barely different from the default one. The radiation damage is not significant with the strength of the explosion and the deletion stuff really ought to be handled properly by normal destructible behavior.

B-Kirill:
> I don't see a reason why this needs to be a separate explosion type. The performance overhead from new types is very noteworthy (we already have a few we ought to remove) and this seems barely different from the default one. The radiation damage is not significant with the strength of the explosion

Okay, there are no questions, everything is clear.


>the deletion stuff really ought to be handled properly by normal destructible behavior

Maybe I didn't understand? That is, to manually add destructibility from a nuclear explosion to each object? Some are specially made indestructible. Pipes, for example, should break anyway, but no, they don't because they're under the floor. I mean, what other solutions can there be, at least I don't see any?

B-Kirill:
> Subfloor objects not breaking with the tile definitely is a bug. While subfloor objects shouldn't take damage when they are covered by a tile that isn't broken, there seems to be an issue regarding transferring the damage onto the actual subfloor entities once the tile is destroyed.

Okay... then:
- Subfloor objects not breaking - a bug that someone will fix someday
- Wallmounts - add destruction thresholds to objects that don't have any yet
- Items - damage thresholds are definitely not enough here. Not all items can be destroyed for the sake of balance, but the rest will be.. it takes a long time to add it. 

What do you think about the idea of adding a bool flag for a regular type of explosion that would remove items? Well, or why this PR is needed at all then...

EmoGarbage404:
After some internal discussion, we came to the conclusion that this is not the solution we want. More individual objects should just support destructible behavior and bugs with subfloor objects should be fixed.

Issue #37133, opened 2025-05-03
new TryStopNukeOpsFromConstantlyFailing heisentest


slarticodefast:
```
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
  Failed TryStopNukeOpsFromConstantlyFailing [10 s]
  Error Message:
   System.AggregateException : One or more errors occurred. (Multiple failures or warnings in test:
  1) SERVER: 10.703s [ERRO] job: Job failed on exception:
Robust.Shared.Utility.DebugAssertException: Encountered empty chunk while enumerating tiles
   at Robust.Shared.Map.Enumerators.GridTileEnumerator.MoveNext(Nullable`1& tileRef) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/Map/Enumerators/GridTileEnumerator.cs:line 33
   at Content.Server.Atmos.EntitySystems.AtmosphereSystem.InvalidateAllTiles(Entity`2 entity) in /home/runner/work/space-station-14/space-station-14/Content.Server/Atmos/EntitySystems/AtmosphereSystem.GridAtmosphere.cs:line 328
   at Content.Server.Atmos.EntitySystems.AtmosphereSystem.OnGridAtmosphereStartup(EntityUid uid, GridAtmosphereComponent component, ComponentStartup args) in /home/runner/work/space-station-14/space-station-14/Content.Server/Atmos/EntitySystems/AtmosphereSystem.GridAtmosphere.cs:line 59
   at Robust.Shared.GameObjects.EntityManager.LifeStartup[T](EntityUid uid, T component, CompIdx idx) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.LifeCycle.cs:line 56
   at Robust.Shared.GameObjects.EntityManager.AddComponentInternal[T](EntityUid uid, T component, ComponentRegistration reg, Boolean overwrite, Boolean skipInit, MetaDataComponent metadata) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Components.cs:line 352
   at Robust.Shared.GameObjects.EntityManager.AddComponent[T](EntityUid uid, T component, Boolean overwrite, MetaDataComponent metadata) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Components.cs:line 304
[...]

PR #37419, opened 2025-05-13
Gyroscope throttling take 2, automatic, inertia based

Krunklehorn:
## About the PR
Previously: #37283
Reduces the influence and power load of gyroscopes on small shuttles.

## Why / Balance
Makes it easier to fly shittles, this time without requiring input from the user.
Does not affect larger shuttles, they still require additional gyroscopes.

## Technical details
No more grill code. This is all based on Shuttle->Physics->Inertia and a subscription to `MassDataChangedEvent`.
A threshold is used to ramp down the gyroscope's angular thrust and power load when the shuttle's inertia drops below it.
The default threshold is roughly half the inertia of the standard 4-door cargo shuttle.

A future PR for UI controls can simply tack on another scalar inside `GetInertiaThresholdScale()`.

Includes a fix for `IsOn` behavior if `TryGet<ShuttleComponent>` fails.

## Media
https://github.com/user-attachments/assets/a8acdfad-6288-4e32-ad1c-0e874de90afa

## Requirements

## Breaking changes
`ThrusterSystem:Enable/DisableThruster()` no longer set IsOn if the `TryGet<ShuttleComponent>` fails. Make sure you are not relying on this behavior.
`ThrusterComponent` now controls the power output of the `ApcPowerReceiver`. Copy the `powerLoad` field to `basePowerLoad` for anything inheriting `BaseThruster`. See `thrusters.yml`.
`ThrusterSystem:SetAngularThrust()` was renamed to `SetAngularThrustVisualState()`.

**Changelog**
:cl:
- add: Gyroscopes now throttle down on smaller shuttles, making them much easier to use.


Krunklehorn:
I encountered a bug possibly be related to #33279.

Since `TileChangedEvent` is raised before `PhysicsComponent` updates it's data, anything that responds to it will be reading old values.

In this case, subscribing to `MassChangedEvent` solved the problem, but only just barely. Technically the call to `DisableThrusters()` is one update behind.

Other systems may be relying on old or stale data causing bugs in the same way.

Content.Server/Shuttles/Systems/ThrusterSystem.cs
```diff
@@ -131,6 +134,25 @@ private void OnShuttleTileChange(EntityUid uid, ShuttleComponent component, ref
...
+    {
+        // The shuttle's inertia has changed, so subtraction would be incorrect. We have to re-accumulate.
+        ent.Comp.AngularThrust = 0f;
+
+        var allThrusters = EntityQueryEnumerator<ThrusterComponent, ApcPowerReceiverComponent, TransformComponent>();
+
+        while (allThrusters.MoveNext(out var thrusterUid, out var thrusterComp, out var thrusterPowerReceiver, out var thrusterXform))
+        {
+            if (thrusterXform.GridUid != ent.Owner)
+                return;
```
metalgearsloth, line 147: This is cooked use continue.

metalgearsloth:
> I encountered a bug possibly be related to #37133.
> 
> Since `TileChangedEvent` is raised before `PhysicsComponent` updates it's data, anything that responds to it will be reading old values.
> 

A) we know
B) gridfixturechangedevent is probably what is required.

Content.Server/Shuttles/Systems/ThrusterSystem.cs
```diff
@@ -131,6 +134,25 @@ private void OnShuttleTileChange(EntityUid uid, ShuttleComponent component, ref
...
+            if (thrusterXform.GridUid != ent.Owner)
+                continue;
+
+            if (ent.Comp.AngularThrusters.Contains(thrusterUid))
+                ent.Comp.AngularThrust += thrusterComp.Thrust * GetInertiaThresholdScale((thrusterUid, thrusterComp), thrusterXform);
+
+            UpdatePowerLoad((thrusterUid, thrusterComp), thrusterPowerReceiver);
+        }
+    }
+
```
metalgearsloth, line 155: This can be pretty frequent might be better to cache the thrusters per grid.

ilovehans10:
Is there a way to manually configure this still, or does this simply make it so that shuttles with low mass all have the same maximum angular velocity?

Krunklehorn:
The old PR used fixed settings, this one is based on a threshold configurable through yaml.

Further control from an updated shuttle console UI, physical switch, wires, etc. can be added later.

AJCM-git:
@Krunklehorn Still interested in getting this in? I can handle this PR if you fix conflicts

Krunklehorn:
Ready!

Content.Server/Shuttles/Components/ThrusterComponent.cs
```diff
@@ -31,6 +37,12 @@ public sealed partial class ThrusterComponent : Component
         [ViewVariables(VVAccess.ReadWrite), DataField("thrust")]
         public float Thrust = 100f;
 
+        /// <summary>
+        /// Throttles the influence of gyroscopes on small shuttles. The default value is half the inertia of the standard 4-door cargo shuttle.
+        /// </summary>
+        [DataField]
+        public float inertiaThreshold = 850f;
```
AJCM-git: fix casing

AJCM-git (changes requested):
should really be in a UI but it works as a step towards that as is

Krunklehorn:
Should be good again, don't know why the test was cancelled.

AJCM-git (approved):

PR #37425, opened 2025-05-13
Slime-Person Blood and Organs Color

Bixkitts:
## About the PR
Made blood spilled by slime people the same color as the slime person.
Organs also can inherit colors now.

## Why / Balance
A distinctly sought-after visual improvement, according to my sources.

## Technical details
I used the existing DNA system and piggybacked on it to attach color data.

The code affects the color calculation for substances and as such
appears to work for syringes, beakers etc.
Different colored slime blood mixing appears to work also,
as can be seen in the screenshots.

## Media
Blue slime leaking blue blood, mixing with the dark purple slime blood.
Light purple slime blood too.

image
Organ colors!
Thanks to @DissidentBullet for the whitened organ sprites and idea.
image

## Requirements

## Breaking changes
No APIs changed.

**Changelog**

:cl: 
- tweak: Slime person blood is now person colored


muburu:
this is super cool but my only concern is that it'll make it hard to see what reagent it actually is at a glance if you don't know it's blood

slarticodefast:
The tests are really acting up today for some reason, let me rerun them for you.
For the new trailing whitespace check, for some reason it is complaining about your fork being a shallow copy. Could you tell me how you cloned your fork so I can investigate?
grafik

Bixkitts:
> The tests are really acting up today for some reason, let me rerun them for you. For the new trailing whitespace check, for some reason it is complaining about your fork being a shallow copy. Could you tell me how you cloned your fork so I can investigate? grafik

- Forked it on Github website (a loooong time ago)
- Cloned my fork on bash CLI with SSH auth
- Updated main branch by pulling from upstream with CLI
- Made a branch with git checkout, worked on that branch
- Left in whitespace and stuff coz I'm not a C# dev and am fumbling/bumbling with VScode
- Committed and pushed it from CLI
- Committed and pushed from CLI again
- Opened PR on github website

slarticodefast:
Thanks, I'll see if I can fix that test. We can ignore it for now.

Bixkitts:
> this is super cool but my only concern is that it'll make it hard to see what reagent it actually is at a glance if you don't know it's blood

Indeed, but such is the cost of violence.
Perhaps a _solution_ would be having various visual effects tied to reagents,
like bubbling/sparkling etc.

LaCumbiaDelCoronavirus:
Also, was the blood intentionally made transparent? If so it would be pretty hard to distinguish from other reagents as slime, should be opaque

NoElkaTheGod:
> Also, was the blood intentionally made transparent? If so it would be pretty hard to distinguish from other reagents as slime, should be opaque

I mean, slime people are transparent, so it'd be pretty weird if their blood was opaque. Why are they transparent when other slimes are not is another question.

Bixkitts:
> Also, was the blood intentionally made transparent? If so it would be pretty hard to distinguish from other reagents as slime, should be opaque

Slime people are transparent, I made the daring assumption that their nutritional constituents would follow suit~
(And it was easier to write)

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -492,13 +503,32 @@ public List<ReagentData> GetEntityBloodData(EntityUid uid)
...
+        if (TryComp<BloodstreamComponent>(uid, out var bloodstreamComp))
+        {
+            var bloodProto = _prototypeManager.Index<ReagentPrototype>(bloodstreamComp.BloodReagent);
+            bloodColorData.SubstanceColor = bloodProto.SubstanceColor;
+        }
+        // Color.White is the default color that gets set if
+        // the blood color is unconfigured.
+        // So we use this as an indicator to set the blood color
+        // to the skin color.
+        if (Color.White == bloodColorData.SubstanceColor)
```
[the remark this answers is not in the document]
Bixkitts: Considered that!
I didn't want to have every of the 100s of packed ReagentPrototype data
 have an extra byte of data for "Is this slime blood?"
That being said, you're right this is not cognitively optimal.
I propose making a "Color.CopySkinColor" which does not test true for
`
Color.CopySkinColor == Color.White
`
or you know, just check `"Slime" == bloodPrototype.ID`.
Working on it rn.

Content.Shared/Chemistry/Components/Solution.cs
```diff
@@ -100,7 +100,7 @@ public void UpdateHeatCapacity(IPrototypeManager? protoMan)
             _heatCapacity = 0;
             foreach (var (reagent, quantity) in Contents)
             {
-                _heatCapacity += (float) quantity *
+                _heatCapacity += (float)quantity *
```
beck-thompson, line 103: don't do style changes in other parts of the code! This could also be your IDE just automatically doing it

Content.Shared/Chemistry/Components/Solution.cs
```diff
@@ -836,9 +836,17 @@ public Color GetColorWithout(IPrototypeManager? protoMan, params string[] withou
...
+                    BloodColorData? bloodData = reagentData.OfType<BloodColorData>().FirstOrDefault();
+                    if (null != bloodData)
+                    {
+                        mixColor = bloodData.SubstanceColor;
+                    }
+
                     continue;
                 }
 
+
```
beck-thompson, line 849: Remove this

Content.Shared/Chemistry/Components/Solution.cs
```diff
@@ -836,9 +836,17 @@ public Color GetColorWithout(IPrototypeManager? protoMan, params string[] withou
                 {
                     first = false;
                     mixColor = proto.SubstanceColor;
+                    List<ReagentData> reagentData = reagent.EnsureReagentData();
```
beck-thompson, line 839: Why are you only doing this if its the first one? I could be misunderstanding, but I think you probably want the overall color to be impacted no matter how much of the reagent is inside.

Content.Shared/Chemistry/Reagent/BloodColorData.cs
```diff
@@ -0,0 +1,27 @@
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.Chemistry.Reagent;
+
+[ImplicitDataDefinitionForInheritors, Serializable, NetSerializable]
+public sealed partial class BloodColorData : ReagentData
```
beck-thompson, line 6: Add a comment - and make it detailed because this is all pretty confusing :laughing:

Also call out slime blood color as a specific example

Content.Server/Body/Components/BloodstreamComponent.cs
```diff
@@ -9,6 +9,7 @@
 using Robust.Shared.Audio;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype.Set;
```
beck-thompson, line 12: This is unused

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -471,18 +473,27 @@ public void ChangeBloodReagent(EntityUid uid, string reagent, BloodstreamCompone
...
+    }
+
+    private void RefreshBloodData(Entity<BloodstreamComponent> entity)
     {
         if (_solutionContainerSystem.ResolveSolution(entity.Owner, entity.Comp.BloodSolutionName, ref entity.Comp.BloodSolution, out var bloodSolution))
         {
             foreach (var reagent in bloodSolution.Contents)
             {
                 List<ReagentData> reagentData = reagent.Reagent.EnsureReagentData();
-                reagentData.RemoveAll(x => x is DnaData);
```
beck-thompson, line 480: Why did you remove this line? They seemed to be specifically targeting DNA. I worry that just clearing the list might have unintended consequences. Probably safer to just remove DNA and the color (if you even need to remove the color)

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -471,18 +473,27 @@ public void ChangeBloodReagent(EntityUid uid, string reagent, BloodstreamCompone
...
+                reagentData.Clear();
                 reagentData.AddRange(GetEntityBloodData(entity.Owner));
             }
         }
-        else
```
beck-thompson, line 484: Probably best to not remove this unless I'm missing something

Content.Server/Fluids/EntitySystems/PuddleSystem.cs
```diff
@@ -69,7 +69,7 @@ public sealed partial class PuddleSystem : SharedPuddleSystem
     [ValidatePrototypeId<ReagentPrototype>]
     private const string CopperBlood = "CopperBlood";
 
-    private static string[] _standoutReagents = [Blood, Slime, CopperBlood];
+    private static string[] _standoutReagents = [Blood, CopperBlood];
```
beck-thompson, line 72: Why did you remove this? Does it make it look weird?

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -492,13 +503,28 @@ public List<ReagentData> GetEntityBloodData(EntityUid uid)
...
         var bloodData = new List<ReagentData>();
         var dnaData = new DnaData();
+        var bloodColorData = new BloodColorData();
 
         if (TryComp<DnaComponent>(uid, out var donorComp) && donorComp.DNA != null)
             dnaData.DNA = donorComp.DNA;
         else
             dnaData.DNA = Loc.GetString("forensics-dna-unknown");
 
+        if (TryComp<BloodstreamComponent>(uid, out var bloodstreamComp))
```
beck-thompson, line 513: Its probably better to raise an event on the entity that's like `GetBloodColorOverride` or something and have the slime system (if that isn't really a thing, just make a new system) that listens and gives the color.

This would fix that issue of slimes having white blood and also allow them to have different colors as well.

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -471,18 +473,27 @@ public void ChangeBloodReagent(EntityUid uid, string reagent, BloodstreamCompone
     }
 
     private void OnDnaGenerated(Entity<BloodstreamComponent> entity, ref GenerateDnaEvent args)
+    {
+        RefreshBloodData(entity);
+    }
+    private void OnHumanoidAppearanceLoaded(Entity<BloodstreamComponent> entity, ref LoadedHumanoidAppearanceEvent args)
```
beck-thompson, line 479: I think it probably would be better to just have a generic "RefreshBloodDataEvent" or something and have the dna and the slime system (see other comment for more info) raise that. You might also want to split the DNA and blood color into two separate functions but I'm unsure

beck-thompson (changes requested):
A few comments!

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: Didn't work without changing this, I ran it in the debugger.

It would leave residual behind duplicates of whatever isn't removed here.
The only things ever added to this list are DnaData and BloodColorData so
it effectively works the same as before.
And given the way GetEntityBloodData() inherently works,
you're gonna get duplicates of anything that isn't deleted anyhow.

I'll see if I can clean this up a little, but it's no worse for specifically handling blood here.

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: It was right there, and I believe Log is deprecated we're supposed to inject the dependency for logging?
Just removed it because I was lazy it's not critical info anyways but I can handle it.

Content.Server/Fluids/EntitySystems/PuddleSystem.cs
Bixkitts: It straight up doesn't work when this is left in, because it does a separate pass for `_standoutReagents`
where it gets the color from the `reagentPrototype` (which would be slime green).
Removing it makes it resolve the color through the slime-checking special case in `Solution.GetColorWithout()`.

Content.Shared/Chemistry/Components/Solution.cs
Bixkitts: I see yellow lines, I touch up as I read.
Ironic considering the empty lines and stuff I then proceed to add.
Noted.

Content.Shared/Chemistry/Reagent/BloodColorData.cs
Bixkitts: Yep.
Such are the workarounds for immutable reagent data...

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: The color stuff is fixed already, all I had to do was `"Slime" == bloodProto.ID`
instead of checking for `Color.White` so the Slime can remain green and only
_do the thing_ when it's an entity with Slime blood+`humanoidAppearanceComponent`.
But ye I could shovel into a system where it's more explicit and not a special case sneakily nested inline here.

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: I was also thinking about splitting the function somehow, I'm a bit unsure there.

But I'm sure on the HumanoidAppearanceLoaded event because it
specifically fires when the humanoid appearance is loaded,
and other people are free to subscribe to it and use it
for that purpose.
I was surprised there wasn't already this event existing.

Content.Server/Body/Components/BloodstreamComponent.cs
Bixkitts: Automagically added :shrug:

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: Oh I misunderstood.
Have a separate slimeBlood system or whatever dispatch these
into one RefreshBloodData handler here instead of subscribing directly here.
Ye sure.

Bixkitts:
While I was at it, I made slime organs inherit slime color as well.
It doesn't work on the held sprite though, I can't find the event for that rn.

Bixkitts:
Didn't get around to making the new system, I'll do it tomorrow

Content.Shared/Chemistry/Components/Solution.cs
Bixkitts: Update: Because I was dumb and was setting the wrong variable, and putting it into "first" somehow made it work even though it was wrong, but now it's fixed.

Content.Shared/Chemistry/Components/Solution.cs
Bixkitts: I activated the .editorconfig formatting and it got worse sorry :x

Resources/Prototypes/Entities/Mobs/Species/slime.yml
Bixkitts, line 1: A neutral white looks so much better than a neon green, at least while I look into changing the
color for this too.

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -492,13 +536,23 @@ public List<ReagentData> GetEntityBloodData(EntityUid uid)
...
         var bloodData = new List<ReagentData>();
         var dnaData = new DnaData();
+        var bloodColorData = new BloodColorData();
 
         if (TryComp<DnaComponent>(uid, out var donorComp) && donorComp.DNA != null)
             dnaData.DNA = donorComp.DNA;
         else
             dnaData.DNA = Loc.GetString("forensics-dna-unknown");
 
+        if (TryComp<BloodstreamComponent>(uid, out var bloodstreamComp))
```
beck-thompson, line 546: Get rid of this try comp

Content.Shared/Chemistry/Components/Solution.cs
beck-thompson: Remember to fix all the formatting changes

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,13 @@
...
+/// <summary>
+/// Subscribe to this event and set the
+/// BloodOverrideColor to override blood
+/// reagent color
+/// </summary>
+[ByRefEvent]
+public record struct BloodColorOverrideEvent()
+{
+    public EntityUid Owner;
+    public BloodstreamComponent? BloodstreamComp;
```
beck-thompson, line 12: Remove this

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,13 @@
+using Content.Server.Body.Components;
+
+/// <summary>
+/// Subscribe to this event and set the
```
beck-thompson, line 4: Only wrap around at 80 characters - do this for all your comments they are very squished :laughing:

Content.Server/Body/Events.cs
beck-thompson, line 12: I think this pattern would be better:
1.) Pass in the current color of the blood to the event
2.) Modify the color in the slime system to the color you want
3.) Update the blood color to whatever the color the event has in the bloodstream system

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -268,9 +277,33 @@ private void OnHealthBeingExamined(Entity<BloodstreamComponent> ent, ref HealthB
...
+        if (TryComp<AppearanceComponent>(part, out AppearanceComponent? appearanceComp))
+        {
+            _appearanceSystem.SetData(part, GoreVisuals.ColorTint, color, appearanceComp);
+        }
+    }
+
     private void OnBeingGibbed(Entity<BloodstreamComponent> ent, ref BeingGibbedEvent args)
     {
         SpillAllSolutions(ent, ent);
+        if (!TryComp<BloodstreamComponent>(ent, out var bloodstreamComp)
```
beck-thompson, line 291: I think you can do everything in this file in the slime system instead

Content.Server/Body/Components/SlimeBloodComponent.cs
```diff
@@ -0,0 +1,13 @@
+using Content.Server.Body.Systems;
+
+namespace Content.Server.Body.Components
```
beck-thompson, line 3: Use file scoped namespaces:

`namespace Content.Server.Body.Components;`

Content.Server/Body/Components/SlimeBloodComponent.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// Used by the SlimeSystem to respond
+    /// to BloodColorOverrideEvents.
+    /// </summary>
+    [RegisterComponent, Access(typeof(SlimeBloodSystem))]
+    public sealed partial class SlimeBloodComponent : Component
```
beck-thompson, line 10: You can just do:

`public sealed partial class SlimeBloodComponent : Component`

Content.Server/Body/Components/SlimeBloodComponent.cs
```diff
@@ -0,0 +1,13 @@
+using Content.Server.Body.Systems;
+
+namespace Content.Server.Body.Components
+{
+    /// <summary>
+    /// Used by the SlimeSystem to respond
```
beck-thompson, line 6: Just say: "Changes the slimes blood to the color of its skin"

Content.Server/Body/Components/BloodstreamComponent.cs
```diff
@@ -182,5 +182,12 @@ public sealed partial class BloodstreamComponent : Component
 
         [DataField]
         public ProtoId<AlertPrototype> BleedingAlert = "Bleed";
+
+        /// <summary>
+        /// A color to override the color of the blood
+        /// set by external components responding to the BloodColorOverrideEvent.
+        /// Otherwise blood just uses it's reagent color.
+        /// </summary>
+        public Color? BloodOverrideColor = null;
```
beck-thompson, line 191: I think you can just get rid of this

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -268,9 +277,33 @@ private void OnHealthBeingExamined(Entity<BloodstreamComponent> ent, ref HealthB
         }
     }
 
+    private void ColorGibbedPart(EntityUid part, Color color)
+    {
+        if (TryComp<AppearanceComponent>(part, out AppearanceComponent? appearanceComp))
```
beck-thompson, line 282: You don't need this trycomp

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -471,13 +504,24 @@ public void ChangeBloodReagent(EntityUid uid, string reagent, BloodstreamCompone
     }
 
     private void OnDnaGenerated(Entity<BloodstreamComponent> entity, ref GenerateDnaEvent args)
+    {
+        RefreshBloodData(entity);
+    }
+    private void OnHumanoidAppearanceLoaded(Entity<BloodstreamComponent> entity, ref LoadedHumanoidAppearanceEvent args)
```
beck-thompson, line 510: To clarify what I think you should do:
1.) Change `GenerateDnaEvent` to a generic `RefreshBloodDataEvent`
2.) Remove `LoadedHumanoidAppearanceEvent`
2.) Raise `RefreshBloodDataEvent` on map init of the slimebloodcomponent (I'm pretty sure that will be fine)

If that doesn't work let me know

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -268,9 +277,33 @@ private void OnHealthBeingExamined(Entity<BloodstreamComponent> ent, ref HealthB
...
+        {
+            return;
+        }
+        var bloodData = GetEntityBloodData(ent);
+        var bloodColorData = bloodData.OfType<BloodColorData>().FirstOrDefault();
+        if (null == bloodColorData)
+            return;
+        foreach (EntityUid part in args.GibbedParts)
+        {
+            if (TryComp<OrganComponent>(part, out OrganComponent? organComp))
```
beck-thompson, line 302: This is for all your trycomps but you can actually just do `TryComp<OrganComponent>(part, out var organComp)`

Content.Client/Body/Systems/GibVisualizerSystem.cs
```diff
@@ -0,0 +1,39 @@
...
+        base.Initialize();
+
+        SubscribeLocalEvent<OrganComponent, VisualsChangedEvent>(OnVisualsChanged);
+    }
+
+    // TODO: This is supposed to callback when an item is held but it doesn't?
+    // Want to tint held slime organs too.
+    private void OnVisualsChanged(EntityUid uid, OrganComponent component, VisualsChangedEvent args)
+    {
+        var item = GetEntity(args.Item);
```
beck-thompson, line 25: What were you trying to do here?

beck-thompson (changes requested):

beck-thompson:
Oh also if you can't figure out how to get the inhand colors working I can try to fix it if you want as w

Content.Shared/Chemistry/Components/Solution.cs
Bixkitts: So do I need to run the formatter only on stuff I've directly modified and not on modified files?
Does that mean disabling/reconfiguring the formatter and undoing all those random formatting changes?

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: But even if I change the event to work with the color rather than the bloodstream component
(like suggested in the other comment),
I still need to resolve the bloodstream component to get the blood reagent being used here.

Do you mean I should try pull all this stuff out into the slimeblood system?

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: Yeah if the appearance and DNA data are all loaded when OnMapInit runs, should work nicely.

Content.Shared/Chemistry/Components/Solution.cs
beck-thompson: Don't bother for the moment, I'll ask some other people if you can just leave them in. In general though, having random changes makes it harder to review and also messes up the git history! The only ones you 100% should remove are files you never touched getting formatting updates

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: Eh, well it could apply to any living thing where you want to modify
the blood and gib color at runtime.
Was thinking to patch this for the normal slime mobs too.
But I'll make sure to pull out anything slime-specific to slimebloodsystem

Content.Server/Body/Events.cs
Bixkitts: er, you mean make it take a color instead?

Content.Server/Body/Components/SlimeBloodComponent.cs
Bixkitts: Sure.
But for future reference, only the slimeblood system is to access these fields, do I not use this for that?

Content.Server/Body/Components/BloodstreamComponent.cs
Bixkitts: Yeah, makes more sense to pass the color through the event and modify it like that.

Content.Server/Body/Systems/BloodstreamSystem.cs
beck-thompson: Yeah good point. Is this possible?

1.) Raise the event with a null color
2.) If no one responds to the event, just don't add bloodcolor data
3.) If someone does respond add blood color data with the override color

Content.Server/Body/Events.cs
beck-thompson: yep!

Content.Server/Body/Systems/BloodstreamSystem.cs
beck-thompson: Sorry your right this was from in the middle of the review: Make a new system that just handles the organ color changes

Content.Client/Body/Systems/GibVisualizerSystem.cs
Bixkitts: Sifting through the code looking for an event that looks like it would give me the held version of the item
so I can tint the sprite there, this one looked like I was on the right track.
But idk, maybe need to lock in and read into the sprite system more it's probably just
a layer or something and I don't need an extra subscription.

Content.Server/Body/Events.cs
Bixkitts: If the function doesn't all fit at once on my 800x600 screen with letters the size of
my toes, I probably wrote it wrong anyhow :nerd_face:

Bixkitts:
> Oh also if you can't figure out how to get the inhand colors working I can try to fix it if you want as well!

Imma have a crack at it, just need to _study the source_ more. I'll tap out if I can't, thanks!

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: But I set appearance data with that! :o
I'll look into it!

Content.Shared/Chemistry/Components/Solution.cs
beck-thompson: Its fine to leave them in! Just make sure you don't do any formatting changes on files you didn't touch (I think there is at least one of those). Just undo the formatting on those files

Content.Server/Body/Systems/BloodstreamSystem.cs
beck-thompson: > But I set appearance data with that! :o
> I'll look into it!

It automatically resolves in the `SetData` command

Bixkitts:
Didn't color the held version yet, and some formatting needs to be cleaned up. Otherwise everything discussed worked out.

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: Done

Bixkitts:
So blood is colored, organs colored, held organs colored and non-humanoid slime slime is colored.
And the formatting changes to files I otherwise didn't touch are undone.
That's just about everything, except ripping apart `BloodstreamSystem.GetEntityBloodData()`
which seemed unnecessary, it's currently chill.
Ready for review again.

Content.Shared/Chemistry/Components/Solution.cs
Bixkitts: Done

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -268,9 +276,27 @@ private void OnHealthBeingExamined(Entity<BloodstreamComponent> ent, ref HealthB
         }
     }
 
+    private void ColorGibbedParts(EntityUid ent, HashSet<EntityUid> gibbedParts)
+    {
+        var bloodData = GetEntityBloodData(ent);
+        var bloodColorData = bloodData.OfType<BloodColorData>().FirstOrDefault();
+        if (null == bloodColorData)
```
beck-thompson, line 283: This is fine but usually its done `bloodColorData == null`

Content.Shared/Chemistry/Components/Solution.cs
```diff
@@ -831,16 +831,21 @@ public Color GetColorWithout(IPrototypeManager? protoMan, params string[] withou
                 {
                     continue;
                 }
+                Color reagentColor = proto.SubstanceColor;
 
+                List<ReagentData> reagentData = reagent.EnsureReagentData();
+                BloodColorData? bloodData = reagentData.OfType<BloodColorData>().FirstOrDefault();
```
beck-thompson, line 837: Do:
`var bloodData = reagentData.OfType<BloodColorData>().FirstOrDefault();`

Content.Shared/Chemistry/Components/Solution.cs
```diff
@@ -831,16 +831,21 @@ public Color GetColorWithout(IPrototypeManager? protoMan, params string[] withou
                 {
                     continue;
                 }
+                Color reagentColor = proto.SubstanceColor;
```
beck-thompson, line 834: Do:

`var reagentColor = proto.SubstanceColor;`

Content.Shared/Chemistry/Components/Solution.cs
```diff
@@ -831,16 +831,21 @@ public Color GetColorWithout(IPrototypeManager? protoMan, params string[] withou
...
 
+                List<ReagentData> reagentData = reagent.EnsureReagentData();
```
beck-thompson, line 836: Do:

`List<ReagentData> reagentData = reagent.EnsureReagentData();`

Content.Shared/Humanoid/Events.cs
```diff
@@ -0,0 +1,17 @@
+namespace Content.Shared.Humanoid;
+/// <summary>
+/// Raised on an entity when their HumanoidAppearanceComponent changes
+/// </summary>
+[ByRefEvent]
+public record struct LoadedHumanoidAppearanceEvent()
```
beck-thompson, line 6: You can remove this now

Content.Client/Body/Systems/GibVisualizerSystem.cs
```diff
@@ -0,0 +1,46 @@
+using Content.Shared.Body.Organ;
+using Content.Server.Body.Systems;
+using Robust.Client.GameObjects;
+
+using Content.Shared.Hands;
```
beck-thompson, line 5: Do:

`namespace Content.Client.Body.Systems;`

Content.Client/Body/Systems/GibVisualizerSystem.cs
```diff
@@ -0,0 +1,46 @@
...
+        foreach (string layer in args.RevealedLayers)
+        {
+            var index = _spriteSystem.LayerMapReserve((args.User, sprite), layer);
+            _spriteSystem.LayerSetColor((args.User, sprite), index, color);
+        }
+    }
+
+    protected override void OnAppearanceChange(EntityUid uid, OrganComponent comp, ref AppearanceChangeEvent args)
+    {
+        if (AppearanceSystem.TryGetData<Color>(uid, GoreVisuals.ColorTint, out var color))
```
beck-thompson, line 38: To avoid nested if statments like this you should use [guard statements](https://en.wikipedia.org/wiki/Guard_(computer_science))! They are very handy. With guards it would look like this:

```
        if (!AppearanceSystem.TryGetData<Color>(uid, GoreVisuals.ColorTint, out var color) || args.Sprite == null)
            return;

        _spriteSystem.SetColor((uid, args.Sprite), color);
```

That's a lot nicer, but you can do even better and simplify to:

```
        if (!AppearanceSystem.TryGetData<Color>(uid, GoreVisuals.ColorTint, out var color))
            return;

        _spriteSystem.SetColor(uid, color);
```

The SetColor command:
image

It will resolve the sprite component and ensure it exists by itself without needing for you to pass it in. `uid` by itself will also automatically be casted into an `Entity<SpriteComponent?>` so you can also just pass it in raw without doing `(ud, null)`.

Content.Client/Body/Systems/GibVisualizerSystem.cs
```diff
@@ -0,0 +1,46 @@
...
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<OrganComponent, HeldVisualsUpdatedEvent>(OnHeldVisualsUpdated);
+    }
+
+    private void OnHeldVisualsUpdated(EntityUid uid, OrganComponent component, HeldVisualsUpdatedEvent args)
+    {
+        if (!(TryComp<SpriteComponent>(args.User, out var sprite)
```
beck-thompson, line 24: This is another guard statement but its would usually look like this:

```
        if (!TryComp<SpriteComponent>(args.User, out var sprite) || !AppearanceSystem.TryGetData<Color>(uid, GoreVisuals.ColorTint, out var color))
            return;
```

I personally think putting it on one line when its only two statements is clearer (it doesn't matter much though) and turning `!(x & y) -> !x || !y` is usually whats done

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -470,36 +496,48 @@ public void ChangeBloodReagent(EntityUid uid, string reagent, BloodstreamCompone
...
         if (TryComp<DnaComponent>(uid, out var donorComp) && donorComp.DNA != null)
             dnaData.DNA = donorComp.DNA;
         else
             dnaData.DNA = Loc.GetString("forensics-dna-unknown");
-
         bloodData.Add(dnaData);
 
+        var ev = new BloodColorOverrideEvent { OverrideColor = null };
+        RaiseLocalEvent(uid, ref ev);
+        if (null != ev.OverrideColor)
```
beck-thompson, line 532: `        if (ev.OverrideColor != null)`

beck-thompson (changes requested):
I dont have time to debug this now, but when I spawn as a slime person it seems to not bleed the skin color I selected. Use `golobby` and `sudo cvar ic.random_characters false` so you can test it out with other colors

Content.Client/Body/Systems/GibVisualizerSystem.cs
Bixkitts: Yes that makes sense, De Morgan law, more legible, sure.

Which one is more performant depends on how often you get the branching to short circuit and what the compiler generates - I thought the compiler would output the same for both, BUT the optimized IL is actually subtly different for each!
I'll do !x || !y either way.

Content.Server/Body/Systems/BloodstreamSystem.cs
Bixkitts: I habitually write rvalues first in comparisons to avoid accidentally writing `if (var = rvalue) {...}` and losing hair to that... but I see C# will typically scream about assignments being comparisons anyhow due to type safety or other reasons.

Content.Client/Body/Systems/GibVisualizerSystem.cs
Bixkitts: Done!

Content.Shared/Chemistry/Components/Solution.cs
Bixkitts: Did you mean `var reagentData`?

Content.Shared/Chemistry/Components/Solution.cs
Bixkitts: So we use var for type inference wherever possible?

beck-thompson: Yeah! If you use rider it'll also give you a little warning telling you to do that

Content.Shared/Chemistry/Components/Solution.cs
beck-thompson: :( yes

Content.Server/Body/Systems/BloodstreamSystem.cs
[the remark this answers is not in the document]
Bixkitts: Thanks, resolved it

Bixkitts:
Pipeline says there's a build+test failure but I don't see anything egregious in commit [d395e58](https://github.com/space-wizards/space-station-14/pull/37425/commits/d395e58b9c80b125720b68ccbb6ef7199ebc8c58) where it starts,
I just re-added the event (which was working before).

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public Color? OverrideColor;
+}
+
+[ByRefEvent]
+public record struct RefreshBloodEvent()
+{
+}
+[ByRefEvent]
+public record struct ColorGibsEvent()
```
beck-thompson, line 19: Remove this

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public HashSet<EntityUid>? Gibs;
+}
+
+[ByRefEvent]
+public record struct ColorGibPartEvent()
```
beck-thompson, line 25: Remove this

Content.Server/Body/Systems/GibsColorSystem.cs
```diff
@@ -0,0 +1,38 @@
+using Content.Server.Body.Components;
+using Content.Shared.Body.Components;
+using Content.Shared.Chemistry.EntitySystems;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.Body.Systems;
+
+public sealed class GibsColorSystem : EntitySystem
```
beck-thompson, line 8:
```
using Content.Server.Body.Components;
using Content.Shared.Body.Components;
using Content.Shared.Chemistry.EntitySystems;

namespace Content.Server.Body.Systems;

public sealed class GibsColorSystem : EntitySystem
{
    [Dependency] private readonly SharedAppearanceSystem _appearanceSystem = default!;
    [Dependency] private readonly SharedSolutionContainerSystem _solutionContainerSystem = default!;
    [Dependency] private readonly IPrototypeManager _prototype = default!;

    {
        base.Initialize();
        SubscribeLocalEvent<SlimeBloodComponent, BeingGibbedEvent>(OnColorGibs);
    }

    private void OnColorGibs(Entity<SlimeBloodComponent> ent, ref BeingGibbedEvent ev)
    {
        if (!TryComp<BloodstreamComponent>(ent, out var bloodstream))
            return;

        if (!_solutionContainerSystem.ResolveSolution(ent.Owner, bloodstream.BloodSolutionName, ref bloodstream.BloodSolution, out var bloodSolution))
            return;

        var bloodColor = bloodSolution.GetColor(_prototype);
        foreach (var part in ev.GibbedParts)
        {
            _appearanceSystem.SetData(part, GoreVisuals.ColorTint, bloodColor);
        }
    }
}
```

If you do it that way you can just remove two of the events. You'll also need to make a new component for slimes (I just reused `SlimeBloodComponent` but you shouldn't) that's called like `ColorOrgansComponent` or something

beck-thompson (changes requested):
OK so the last error you have is for an odd reason. When players get spawned it:
1.) Spawns an entity with the default profile:
image
2.) Spawns the entity with a real profile:
image

The reason why map init doesn't work is because its triggered on the first one when the profile is the default (so it just uses the default skin color) when the second one sets the actual skin color. That means your event is probably correct and is what we should use.

The issue is this: 
image

When that default profile is loaded, the solutions haven't been initialized yet (The `OnInit` of the `HumanoidAppearanceComponent` is run before `BloodStream`) so it fails. The simple solution would be to reorder the init events so bloodstream goes first but I'm not sure that is a good idea.

Don't make any changes for now I'm going to ask around and see what the best solution would be

Content.Server/Body/Systems/GibsColorSystem.cs
Bixkitts: The GibbedParts HashSet I loop through here contains EntityUids for everything expunged from from a gibbed entity;
Organs, limbs, cellphones, just whatever. Usually 20-ish items.
Hence the second event (ColorGibPartEvent) is raised to (somewhat) branchlessly invoke the coloring logic for specific gibs we want to color.

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -1,3 +1,4 @@
+using System.Linq;
```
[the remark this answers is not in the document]
Bixkitts: It can be difficult to tell what's a no-longer-needed dependency _I_ added,
or one that was there all along!

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,28 @@
+
```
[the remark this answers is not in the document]
Bixkitts: Done, I'll amend commit

Content.Server/Body/Systems/GibsColorSystem.cs
Bixkitts: And replacing the `ColorGibsEvent` with a `BeingGibbedEvent` is a cool idea, I should have thunk of that,
HOWEVER I explicitly raise the `ColorGibsEvent` before running `SpillAllSolutions()` because
if `SpillAllSolutions()` runs first, the blood substance can no longer have it's color determined
(because ALL of it is on the floor).

If I subscribe to `BeingGibbedEvent`, there's a non-deterministic chance I could spill all the blood first
and hence have nothing to color the gibs with.

Bixkitts: Running `OnColorGibs` with the `SlimeBloodComponent` instead of the `BloodstreamComponent`?
Makes sense, prevents the event from running on all things with blood.
I'd do that.

Content.Server/Body/Systems/BloodstreamSystem.cs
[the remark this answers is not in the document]
beck-thompson: > It can be difficult to tell what's a no-longer-needed dependency I added,
or one that was there all along!

I really recommend using rider, it tells you which ones are unused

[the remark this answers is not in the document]
Bixkitts: Oh it's free for non commercial use?
Hell yeah.
VSCode/Nvim also tell me which ones are unused, but
not which ones are unused AND added by me (I'm not supposed to delete unused dependencies that were already around).

Content.Server/Body/Systems/GibsColorSystem.cs
Bixkitts: Fair, I'll do that- I didn't see I could force it to execute later in the
event invocation than the `BloodstreamSystem`. But what about the `ColorGibPartEvent`?

Content.Server/Body/Events.cs
Bixkitts: This exists solely because not all GibbedParts are to be colored.
See https://github.com/space-wizards/space-station-14/pull/37425#discussion_r2104775668

Content.Shared/Body/Components/GibSplatterComponent.cs
```diff
@@ -0,0 +1,10 @@
+namespace Content.Shared.Body.Components;
+
+/// <summary>
+/// If attached to an entity if it's a part in a BeingGibbedEvent,
+/// this component will make sure the giblet will get tinted/splattered with blood.
+/// </summary>
+[RegisterComponent]
+public sealed partial class GibSplatterComponent : Component
```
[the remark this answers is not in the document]
Bixkitts: I had dreams of a world where it wasn't, but I shall!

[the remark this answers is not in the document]
beck-thompson: If you ever want to come back and make it more generic for other species, completely feel free to change it up and rename!

Content.Server/Body/Systems/BloodstreamSystem.cs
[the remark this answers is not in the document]
beck-thompson: Rider is really nice, even if you don't think you'd like it I'd recommend at least giving it a shot

Content.Server/Body/Events.cs
beck-thompson: OK I think that's a good way of doing it then!

beck-thompson:
Spent a while trying to think of a nice solution but couldn't find any. I say:
1.) Add a comment to `LoadedHumanoidAppearanceEvent` mentioning that it might get raised more than once when an entity is spawned (basically explain the issue we found here)
2.) Just remove the error check and add a comment that mentions it can fail because of component initiation order

Its sus but the rest of the solutions I could think of are worse. Having it fail the check isn't that big of a deal imo

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,27 @@
...
+/// reagent color
+/// </summary>
+[ByRefEvent]
+public record struct BloodColorOverrideEvent()
+{
+    public Color? OverrideColor;
+}
+
+[ByRefEvent]
+public record struct RefreshBloodEvent()
```
beck-thompson, line 14: Add comments to all of these new events you made

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,27 @@
+namespace Content.Server.Body;
+/// <summary>
+/// Subscribe to this event and set the
+/// BloodOverrideColor to override blood
```
beck-thompson, line 4: When writing comments, go out to the 80 character mark! When your comments have line breaks that much its hard to read.

beck-thompson (approved):
Minor comment stuff but the rest looks good!

Content.Server/Body/Events.cs
```diff
@@ -0,0 +1,27 @@
...
+/// reagent color
+/// </summary>
+[ByRefEvent]
+public record struct BloodColorOverrideEvent()
+{
+    public Color? OverrideColor;
+}
```
slarticodefast, line 11:
```suggestion
public record struct BloodColorOverrideEvent(Color? OverrideColor);
```

slarticodefast (dismissed):

Bixkitts:
`BloodColorData` -> `ReagentColorData`.
You could use `ReagentColorData` to color any reagent for any reason
so I think it makes sense where it is in the Solution color calculation.
Does it need to be more performant?

Content.Shared/Chemistry/Components/Solution.cs
```diff
@@ -831,16 +831,20 @@ public Color GetColorWithout(IPrototypeManager? protoMan, params string[] withou
                 {
                     continue;
                 }
-
+                var reagentColor = proto.SubstanceColor;
+                var reagentData = reagent.EnsureReagentData();
+                var bloodData = reagentData.OfType<BloodColorData>().FirstOrDefault();
+                reagentColor = bloodData?.SubstanceColor ?? reagentColor;
```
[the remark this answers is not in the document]
slarticodefast: That works for me 😄

TheShuEd:
I think this is a bad implementation of this functionality from the code point of view. We really need a refactor of reagents and solution entities that Jezy did to implement full editing of reagents (including color) in runtime

TheShuEd:
I am very much against the whole part with repainting puddles in this PR

Bixkitts:
> I think this is a bad implementation of this functionality from the code point of view. We really need a refactor of reagents and solution entities that Jezy did to implement full editing of reagents (including color) in runtime

I don't entirely disagree, and if you have a refactored reagent system I can integrate with this I would.
I found the immutable reagent prototype data and overriding the reagent color a bit cumbersome here!

However, I think the current system makes sense for 99% of cases.
Most other solutions would be substantially cache-unfriendlier for little gain
(If I'd want to touch anything, I'd make reagent prototype color data contiguous in memory if it isn't already).
You probably wouldn't want to be instantiating a runtime-modifiable color for every reagent, for example.

Additionally, the solution code here would look 90% the same if the color data were stored and modified at runtime in whatever other way.
Overall, despite a little weirdness I think the `ReagentId` container is chill.
I'd maybe swap like the ReagentData list for a map.

EmoGarbage404 (commented):
apologies for the late review but i had some insight that may be useful.

first of all, i don't think the reagent color datum is that bad of a solution. For the framework we have, it's lightweight and generally a decent way to handle it. However, I will concede that the implementation is a bit lackluster due to the way prototypes are often read.

People will often just grab the color off of the reagent, which is obviously not ideal for this kind of thing. My suggestion would be to hide the color datafield on the reagent prototype using either the protected/private access modifiers or the `AccessAttribute` (look into the permissions for this).

From there, you can add an API that retrieves the color off of the reagent, using the color from the datum when needed. That way, instead of being inconsistently colored depending on where the color was sourced in code, it would always appear to be a single color.

If that's done, i think the code quality would be acceptable to merge. DM me if you have questions.

Bixkitts:
I dismissed it, as it would entail asking "Hey, do we have special color data?"
on every access (99.9% of the time they don't).
Or giving _every_ reagent color data, forcing color access through this,
and writing it's default color in there (defeats the point of the reagent
prototype/flyweight pattern)...

But you're right and having one way to access the color instead of two is way better either way, I'll do it soon!

Bixkitts:
I tried some more invasive methods also, but this seemed like a
fairly not-evil solution.
I also propose renaming `ReagentId` to `ReagentDecorator`.

The call sites using that specific obsolete `ReagentId(string)` constructor are
clearly marked as the places that should maybe be refactored
to use `ReagentId` if possible (and get dynamic reagent colors).

However, the Guide UI exemplifies a situation where just a string
is used to resolve a reagent+color, and it makes sense,
so I might give these sorts of classes access to `ReagentPrototype` also.

Content.Server/Body/Components/SlimeBloodComponent.cs
```diff
@@ -0,0 +1,15 @@
+namespace Content.Server.Body.Components;
+/// <summary>
```
Simyon264: Formatting, put a newline after the namespace

Content.Client/Body/Systems/GibVisualizerSystem.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Content.Client.Body.Systems;
+
+/// <summary>
+/// Ensures entities with <see cref="SlimeGibSplatterComponent"/> have the correct color.
+/// </summary>
+public sealed class GibVisualizerSystem : VisualizerSystem<SlimeGibSplatterComponent>
+{
+    [Dependency] private readonly SpriteSystem _spriteSystem = default!;
+    public override void Initialize()
```
Simyon264: Put a newline after the dependency

Content.Shared/Chemistry/Reagent/ReagentId.cs
```diff
@@ -35,6 +36,28 @@ public ReagentId()
         Data = new();
     }
 
+    [Obsolete("If you are using this, you don't care about getting extended reagent runtime data like DNA and dynamic color")]
+    public ReagentId(string prototype)
+    {
+        Prototype = prototype;
+        Data = null;
+    }
```
Simyon264, line 44: You are adding a new API and are adding it as obsolete?

Simyon264 (changes requested):

Content.Shared/Chemistry/Reagent/ReagentId.cs
Bixkitts: The `ReagentPrototype.SubstanceColor` field is now only accessible through the `ReagentId.GetColor()`
function (which is the new API).
Most practical usages of reagent color in code, e.g. in the substances system, properly
instantiate and use the `ReagentId` decorator class everywhere (which attaches potential DNA+Runtime Color data).
However, there are a few cases where a reagent color is derived directly from a raw prototype string
e.g. in the guidebook.
These specific cases now need to shove their raw string up in this specific `ReagentId` constructor if they
want to get access to the `ReagentPrototype.SubstanceColor` field, which
yells at you for using it, because if at all possible you shouldn't be.
Some cases, like the Guidebook, may need to be made into friend classes,
but I think some of them could be rewritten! (e.g. printing the correct
runtime color of reagent text when examining a puddle).

Bixkitts: But I'm here to color slime blood and chew bubble gum, so I'd leave it obsolete for now and
try patch it later, before I'm overwhelmed by merge conflicts.

Lev-tobecontinued:
is this still being worked on?

Bixkitts:
> is this still being worked on?

Sure, I was simply waiting for further comment.
Should I solve the merge conflicts?
Should I work on the blood color/attribute fetching API more, or do that in another PR (if at all)?
Because a significantly better API for that as @EmoGarbage404 suggested is likely it's own entire undertaking.

WinkRussian:
just merge this its been almost a full year

Princess-Cheeseballs:
While this is a desired feature, the changes to the reagent prototype, as well as ReagentId are suspect, as well this PR has had requested changes for months now. In addition it has been heavily conflicted by the recent body refactors and would need a new approach. I might PR this myself in the future if someone else doesn't get to it first, and if you want to revisit this PR I recommend starting from scratch due to the numerous changes to API since this PR was created, and the fact this PR has code quality issues.

PR #37436, opened 2025-05-14
Wall inheritance cleanup

IamVelcroboy:
## About the PR
Things here were getting pretty messy
- Fixes diagonal versions of walls inheriting rebound
- Fixes diagonal walls having more HP than their full wall variants
- Fixes tags being all over the place
- Fixes diagonal walls not inheriting features from full wall variants
- Fixes many walls having the wrong damage modifiers
- Fixes other random things walls were inheriting for no reason
- Fixes Radiation Blocker not being applied consistently
- Reduces lots of redundant copy/paste
- Fixes `tilewalls` cmd from changing tiles under diagnal walls

## Todo: 
- Should walls be inheriting the `Structure` tag from `BaseStructure`? only the meat wall previously was.
- I didn't touch RCDable walls but it seems inconsistent a bit. 

## Technical details
~~I made a second parent for walls so it wouldn't shock the system for downstreams.~~

## Media
n/a

## Requirements

## Breaking changes
~~Downstreams will want to add the new parent to their walls probably~~

**Changelog**
n/a

### DIFF
--- a/Content.Server/Construction/Commands/TileWallsCommand.cs
+++ b/Content.Server/Construction/Commands/TileWallsCommand.cs
@@ -25,6 +25,7 @@ public sealed class TileWallsCommand : IConsoleCommand
 
     [ValidatePrototypeId<TagPrototype>]
     public const string WallTag = "Wall";
+    public const string DiagonalTag = "Diagonal";
 
     public void Execute(IConsoleShell shell, string argStr, string[] args)
     {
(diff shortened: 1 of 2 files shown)


IamVelcroboy:
Which sprite?

IamVelcroboy:
No that was silly, I didn't need to do that at all. Should be better this way

PR #37441, opened 2025-05-14
Allow scrap generators to be fultoned & Decrease Fulton delay

ViceEmargo:
## About the PR
Let the scrap generators found in space be able to be whisked away.
Reduce the amount of time it takes for a Fultoned object to make its way to the beacon.
## Why / Balance

Fultons are an excellent utility item primarily utilized by Salvagers. Many of the objects they will encounter are these scrap generators, which are a great source of material. However, they were unable to be fultoned, which I feel reduced the appeal of obtaining the fulton and beacon.
I personally found it unappealing to spend the time and hassle making carrying a single broken generator from a magnet'd Space Debris back to station. This change would make be an additional boost of enjoyment to finding the fulton and beacon. (or their blueprints)

I believe the Fulton delay should be reduced from 45 seconds to 30 seconds to account for the broadcast on Cargo radio that tells the listeners that the magnetized debris will disappear in 30 seconds.
In the event that a salvager collects their loot and utilizes a fulton within that previous 15 second window, the actively fultoned loot will be all lost to the depths of space.

This change will have a side effect of additionally "buffing" the Thief's kit which includes a fulton kit.

## Technical details
Adds the Anchorable component to BaseScrapLarge which currently only the scrap, broken down generators inherit from.
Changes the duration of both FultonComponent and FultonedComponent from 45 seconds to 30 seconds.
(From my testing, only changing the duration on FultonedComponent actually resulted in the correct change.)

## Media

https://github.com/user-attachments/assets/15ad057b-704a-4708-8f14-ca47cd26a6fe

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Scrap generators can now be fultoned
- tweak: Fulton delay reduced to 30 seconds from 45

### DIFF
--- a/Content.Shared/Salvage/Fulton/FultonComponent.cs
+++ b/Content.Shared/Salvage/Fulton/FultonComponent.cs
@@ -32,7 +32,7 @@ public sealed partial class FultonComponent : Component
     /// How long the fulton will remain before teleporting to the beacon.
     /// </summary>
     [ViewVariables(VVAccess.ReadWrite), DataField("duration")]
-    public TimeSpan FultonDuration = TimeSpan.FromSeconds(45);
+    public TimeSpan FultonDuration = TimeSpan.FromSeconds(30);
 
     [ViewVariables(VVAccess.ReadWrite), DataField("whitelist"), AutoNetworkedField]
     public EntityWhitelist? Whitelist = new()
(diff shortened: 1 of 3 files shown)


YoungThugSS14:
There’s already a PR for it but Orks is busy with other things https://github.com/space-wizards/space-station-14/pull/34594 😔😔😔

ViceEmargo:
> There’s already a PR for it but Orks is busy with other things #34594 😔😔😔

My hubris hath been tarnished. I really didn't know there was a PR to do this, huh. Thank you for letting me know. It looks like a more worthwhile solution will be in place in time.

PR #37442, opened 2025-05-14
Prediction of Acquistion Slip error message

BigfootBravo:
## About the PR
Using a pen on an acquisition slip no longer spams the error message to the player. It now instead only sends the message once.

## Why / Balance
This is just a quality of life adjustment. Reduction of spammed messages whenever this situation would arise.

## Technical details
Popup for this error message is now predicted. Changed "PopupEntity" in PaperSystem.cs to ~"PopupPredicted"~ "PopupClient."

## Media
Not needed.

## Requirements

## Breaking changes
None.

**Changelog**

:cl:
- fix: No more spammed error message when trying to write on a acquisition slip.

### DIFF
--- a/Content.Shared/Paper/PaperSystem.cs
+++ b/Content.Shared/Paper/PaperSystem.cs
@@ -122,7 +122,7 @@ public sealed class PaperSystem : EntitySystem
                 if (entity.Comp.EditingDisabled)
                 {
                     var paperEditingDisabledMessage = Loc.GetString("paper-tamper-proof-modified-message");
-                    _popupSystem.PopupEntity(paperEditingDisabledMessage, entity, args.User);
+                    _popupSystem.PopupPredicted(paperEditingDisabledMessage, entity, args.User);
 
                     args.Handled = true;
                     return;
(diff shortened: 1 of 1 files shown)


Content.Shared/Paper/PaperSystem.cs
```diff
@@ -122,7 +122,7 @@ private void OnInteractUsing(Entity<PaperComponent> entity, ref InteractUsingEve
                 if (entity.Comp.EditingDisabled)
                 {
                     var paperEditingDisabledMessage = Loc.GetString("paper-tamper-proof-modified-message");
-                    _popupSystem.PopupEntity(paperEditingDisabledMessage, entity, args.User);
+                    _popupSystem.PopupPredicted(paperEditingDisabledMessage, entity, args.User);
```
beck-thompson, line 125: I honestly have no clue why this is a PopupEntity (other players could see it) when every other one uses PopupClient (only the user can see it). I say just change it to a PopupClient as well like the others!

beck-thompson (changes requested):

Content.Shared/Paper/PaperSystem.cs
beck-thompson: (Also `PopupPredicted` *is* the correct function to replace `PopupEntity` with - I just think it shouldn't be a global popup)

BigfootBravo: There we go, done.

beck-thompson (approved):

PR #37445, opened 2025-05-14
Slightly re-nerf zombification speed

Alpha-Two:
#34472 fixed a bug with `PendingZombieComponent` having way too long a grace period, and also doubled the infection damage. This reduced the TTK from infection alone from ~19 minutes to ~2 minutes. This PR brings the TTK back up to ~5 minutes. 

## About the PR
- Changed the default grace period from Zero to 2 minutes.
- Reduced damage from poison down from 0.4 to 0.3. This will still kill voxes, but at a slower rate.

## Why / Balance
General player feedback I've been hearing is that it is too short, and also prevents most counterplay, considering there is practically no way to be cured if you're more than 90 seconds away from the nearest source of ambuzol. This PR will keep the infection deadly, but allow people to actually be saved by the cure.

## Technical details
Editing component fields

## Requirements

## Breaking changes

**Changelog**
:cl: UpAndLeaves
- tweak: You now have three extra minutes to get to a cure before succumbing to Romerol and other zombie infections.

### DIFF
--- a/Content.Shared/Zombies/PendingZombieComponent.cs
+++ b/Content.Shared/Zombies/PendingZombieComponent.cs
@@ -17,7 +17,7 @@ public sealed partial class PendingZombieComponent : Component
     {
         DamageDict = new ()
         {
-            { "Poison", 0.4 },
+            { "Poison", 0.3 },
         }
     };
 
@@ -34,7 +34,7 @@ public sealed partial class PendingZombieComponent : Component
     /// The amount of time left before the infected begins to take damage.
     /// </summary>
     [DataField("gracePeriod"), ViewVariables(VVAccess.ReadWrite)]
-    public TimeSpan GracePeriod = TimeSpan.Zero;
+    public TimeSpan GracePeriod = TimeSpan.FromMinutes(2);
 
     /// <summary>
     /// The minimum amount of time initial infected have before they start taking infection damage.


superjj18:
I secretly loved that zombie infection sickness was bugged and didn't work T-T

At very least my chances of survival weren't nuked because chem is dead and a dead zombie diona burst into 3 mini-zombies with near perfect accuracy.

The sickness while thematic and making total sense, is just insanely oppressive, ESPECIALLY if chem gets ganked by an II(not hard, considering right next door in medbay is arguably the largest supply of bodies to infect, so you can both gank chem and start a proper outbreak).  I almost think brigmed should include a proper chem set up just so chem isn't the sole source of Dylo and Ambuzol-/+

IProduceWidgets (approved):
Does what it says on the tin.
I'll point out this does make Vox 3x as resistant to zombie virus compared to other species as opposed to the 2x as resistant.

EmoGarbage404:
does the math on this PR make sense? 100 health / 0.3 dps = ~ 333.33 seconds => ~5.5 minutes. Adding on the 2 minute grace period, this seems higher than the 5 minutes posted on the PR.

Not that i'm directly opposed.

robinthedragon:
tbf, zombie bites do a bunch of damage iirc, if we assume zero medical care then you'll probably have lost enough of your health to bites themselves that a 5 minute TTK is more realistic then a 7 minute one

Alpha-Two:
> does the math on this PR make sense? 100 health / 0.3 dps = ~ 333.33 seconds => ~5.5 minutes. Adding on the 2 minute grace period, this seems higher than the 5 minutes posted on the PR.
> 
> Not that i'm directly opposed.

Yeh I did the maths wrong, it's a bit longer than said, but still what is intended.

VerinSenpai:
Honestly I hadn't realized the time to die had dropped that much. Having that grace period adds to the excitement imo since for a time you're not sure if you're infected or got lucky.

SlamBamActionman:
I think this solution is better than #38051 because honestly just a 0.4 -> 0.3 is still just a minute and a half longer resistance; it's the 2 minute grace period that actually matters. 

I'm a bit biased in favor of Vox because I play them and think that a slight zombie virus resistance is minor in comparison to the major downside the species has. Considering the self-regen stops at 20 damage a Vox has 9.7 minutes before succumbing to the virus which seems like it's not broken compared to 7.5 (it's 133 seconds more, wheras previously it was 50 seconds more).

As for the overall change: currently the zombie virus feels like it is almost binary. If you get infected, you're dead. And yes, zombie viruses are supposed to be lethal, but the window for ambuzol is so small that the intended counter feels non-existent. It also makes it difficult to do many fun "I'm... infected... Shit, so this is it... Here, take my ID and go on without me. I'll hold them off... FOR NANOTRASEN!" roleplay moments. So I'm for this change conceptually.

SlamBamActionman:
Looking at bed values:

> For the current damage values:
> Being on a regular bed is -0.1, so you take 0.3 damage
> Being on a medical bed is -0.2, so you take 0.2 damage
> Being on a regular bed sleeping is -0.3, so you take 0.1 damage
> Being on a medical bed is -0.6, so you heal 0.2
> 
> For this PR:
> Being on a regular bed is -0.1, so you take 0.2 damage
> Being on a medical bed is -0.2, so you take 0.1 damage
> Being on a regular bed sleeping is -0.3, so you take 0 but don't heal
> Being on a medical bed is -0.6, so you heal 0.3

This seems fine; you don't gain any health from sleeping in a regular bed outside Medbay and I don't think I need to explain why sleeping during an active zombie apocalypse is bad.

Maintainers Slartibartfast and ScarKy0 have approved this PR on Discord so I will go ahead and merge it.

SlamBamActionman (approved):

PR #37457, opened 2025-05-14
remove actions from maps

deltanedas:
## About the PR
yet again removing serialized actions from every map

## Why / Balance
need for action refactor to pass tests

## Breaking changes
its naturally conflict bait but its easy to fix:
1. check out --ours for the map
2. search Action and remove the action entities
3. remove any action related components like ActionContainer, ContainerContainer (if only the action container has any entities) from gas tanks jetpacks flashlights etc


ScarKy0 (approved):
These all should absolutely be different PRs but i'll see what others decide on.
I checked the diff and it does in fact only remove mapped actions (how did this even happen??)

ArtisticRoomba:
Does this middle finger a lot of large mapping PRs (genpop) that are currently being worked on right now?

deltanedas:
its a minute at most to fix them see breaking changes

VerinSenpai:
wait, ActionToggleInternals was mapped to a ruin? am I reading that right?

deltanedas:
yes someone picked up a gas tank or something

Issue #37118, opened 2025-05-02
Items keep rotating when thrown on a conveyor


slarticodefast:
## Description
They loose angular friction somehow.
This was introduced in https://github.com/space-wizards/space-station-14/pull/35931

**Reproduction**
- Make a conveyor loop
- throw an item on it
- observe

**Screenshots**

![Image](https://github.com/user-attachments/assets/82b4a40c-01a6-47b4-9725-2858697e1ea0)

PR #37468, opened 2025-05-15
Conveyors spin fix

Pronana:
## About the PR

Fixes https://github.com/space-wizards/space-station-14/issues/37118

Also fixes conveyors applying friction when turned off and when an item is in the air.

## Why / Balance

Bugfixes.

## Technical details

- Checks if an item is actually conveyed and on the ground before applying friction mods to it.

- Conveyors now apply friction to angularVelocity.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Items no longer spin on Conveyors


Pronana:
Splitting the simple fix into its own PR and gonna make two mutually exclusive fixes for the mob issue

Pronana:
I cannot stop failing TryNukeOpsFromConstantlyFailing today

slarticodefast:
Hmm, multiple TryComps every single tick for every conveyed item seem really bad for performance.
I'll need to look at the code in more detail, but does this have any advantage over the simpler approach that is used in https://github.com/space-wizards/space-station-14/pull/38468 ?

Princess-Cheeseballs:
> Hmm, multiple TryComps every single tick for every conveyed item seem really bad for performance.
> 
> I'll need to look at the code in more detail, but does this have any advantage over the simpler approach that is used in https://github.com/space-wizards/space-station-14/pull/38468 ?

Give me a bit to re-open and look at/test this code it's been a while. 

From a curosry glance I think the Fixture TryComp could be removed, the Physics TryComp is meant to prevent conveyors from applying Friction to items in the air.

Especially after Slam's conveyor PR got merged I need to double check everything still is needed.

slarticodefast:
That TryComp did not have any effect anyways since the `Resolve` inside `IsConveyed` would have handled it when not passed in as a parameter. Performance-wise this still has the same problem though.

slarticodefast:
Can't the body status be checked only when it changes instead of using a trycomp every frame?
That is probably something that ConveyedComponent should already be checking for.

Princess-Cheeseballs:
> Can't the body status be checked only when it changes instead of using a trycomp every frame? That is probably something that ConveyedComponent should already be checking for.

I'll see if I can implement this, from testing both PRs the main differences for the other are:

- Items still spin if Conveyor is on. 
- Friction is applied to entities in the air, even if the conveyor is off.

Princess-Cheeseballs:
> Can't the body status be checked only when it changes instead of using a trycomp every frame? That is probably something that ConveyedComponent should already be checking for.

BodyStatus changing doesn't raise an event it seems.
image

Princess-Cheeseballs:
> > Can't the body status be checked only when it changes instead of using a trycomp every frame? That is probably something that ConveyedComponent should already be checking for.
> 
> BodyStatus changing doesn't raise an event it seems. image

Actually I thought of a way better solution hold on.

Princess-Cheeseballs:
Both TryComps are gone and it works as intended, should be gtg.

TheRealMasterChief117:
uuuh.. still not merged?

slarticodefast (approved):

PR #37470, opened 2025-05-15
Revenants become capable of spawning earlier in the round

Kittygyat:
## About the PR
I let Revenants potentially spawn 15m into the round, down from the previous 45m requirement.

## Why / Balance
Revenants, a snowball solo antagonist, tend to spawn too near to the end of the round to actually build-up or even _use_  their abilities or cause any issues, and this helps corrects that by letting them spawn around similar times to other antagonists, giving them the time to snowball & have a bit more impact on the round that they need

## Technical details
Changed the Revenants "earliestStart: 45" to "earliestStart: 15" in the "events.yml" folder

## Media
N/A

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Revenants can now potentially spawn 30m earlier than before

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -277,7 +277,7 @@
   - type: StationEvent
     weight: 7.5
     duration: 1
-    earliestStart: 45
+    earliestStart: 15
     minimumPlayers: 20
   - type: RandomSpawnRule
     prototype: MobRevenant


CookieMasterT:
Ok, but no one's dead this early.

Kittygyat:
> Ok, but no one's dead this early.

The vast majority of dead people _will_ be morgued past the +45m mark, being unabsorbable, but the Revenant can absorb narcoleptics or catch people _as_ they crit earlier in the round with the extra time, instead of only ever spawning atleast +45m in, giving them precious little time to snowball before evac

EmoGarbage404:
I think simply allowing revenant to spawn earlier in a round without any further reinforcement is a bad idea. The role is very dependent on having some amount of chaos and death going on, so spawning too early may just be boring. We might want to consider an alternative spawning condition, however, whenever there are a certain amount of bodies on the station.

VerinSenpai:
On top of what Emo said, an issue I've got with this is the revenant tends to be difficult and even annoying to fight against, often conducting hit and run tactics like spacing an area and then dipping. I'm a touch biased as an engineering main but I definitely don't wanna see this happening earlier in the round unless a more effective way to deal with the rev is added.

Cojoke-dot:
Maybe if rev had any actual reason to live I would agree, but at the current moment they have 3 abilities that barely affect the round and no objectives.

Kittygyat:
Hmm, unsure what I could do to address these issues within the scope of this PR, however I could adjust their spawning threshold to the regular wait times of other antagonists (~20m) until revenants are reworked or changed enough to justify a higher wait time, but I'd like to see them actually in-play & have the potential to snowball in more rounds, until the time comes that they _actually get_ reworked/adjusted.

I could try look into adding some of the suggestions mentioned in the comments, but they might be outside of my current capabilities.

slarticodefast:
Since Emo also added a concern, the revenant will be reworked anyways in the future and we are going to shift towards using the dynamic gamemode I will be closing this PR.

PR #37461, opened 2025-05-14
Conveyor Fixes

Pronana:
## About the PR
Added some extra logic to conveyors and movers to fix a few minor issues. Still not perfect.

Also doubled the friction conveyors apply because it felt too slippery as a mover, should also help prevent items from getting thrown off.

Fixes #37117
Fixes #37118

## Why / Balance
The doubling of Conveyor Friction was because it felt too slippery when walking on it with 5 friction, plus items would often get bumped and slide off.
Also fixes some bugs.

## Technical details

- Added an event that is raised after tileDef is acquired to override tileDef Friction and Acceleration values

- Conveyors now override tileDef friction values for movers. 

- The tileDef override compensates for differences in baseFriction to try and ensure you have the same friction as items moving on a conveyor.

- Added angularFriction to conveyors so items no longer spin. 

- Conveyors now check if the object they're trying to convey is on the ground before trying to convey it. 

- Doesn't fix big objects getting stuck on conveyors or ice crust/ glue issues.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: 
- tweak: Conveyors have more friction
- fix: Items no longer spin on conveyors

PR #37473, opened 2025-05-15
Conveyor Fixes -Alt-

### LINKED ISSUE #37117: Mobs get conveyed at a different speed than items
## Description
Likely caused by https://github.com/space-wizards/space-station-14/pull/36847
I assume because mobs calculate friction differently somehow
@Pronana 

**Reproduction**
- Make a conveyor loop
- stand on it
- put an item on it
- compare

**Screenshots**

Image

### LINKED ISSUE #37118: Items keep rotating when thrown on a conveyor

### LINKED ISSUE #37462: Ice Crust and Glue affect conveyed mobs
## Description
Ice Crusts and Glue below conveyors will slow and slide mobs causing them to have wildly different friction compared to items.

This also results in mobs being unable to resist conveyors due to their speed and acceleration being changed to be well below the conveyor's acceleration

**Reproduction**
- Make a conveyor
- Put glue or ice crust underneath (the entity not the tile)
- Turn the conveyor on
- Try to walk in the opposite direction of the conveyor

**Screenshots**

**Additional context**
Ideally conveyors would have better ways to bulldoze values and account for modifiers or we could just add bool AffectsConveyed = false to friction and movementcontacts

Pronana:
## About the PR
The slightly more performative solution to conveyors and ice crust issues.

Requires/Replaces: #37468
Mutually Exclusive with: #37461 (Merge only one or neither)

Fixes #37117
Fixes #37118
Partially Fixes #37462 glue still remains an issue

## Why / Balance
The doubling of Conveyor Friction was because it felt too slippery when walking on it with 5 friction, plus items would often get bumped and slide off.

Also fixes issues.

This is the less performative but more accurate to how conveyors are supposed to work version. Raises an extra event every tick for every mover for the cost of working better with conveyors.

## Technical details

- Added an event that is raised after tileDef is acquired to override tileDef Friction and Acceleration values

- Added an event that is raised at the end of HandleMoveMovement to override all friction modifiers in case of emergency (AKA for Conveyors because they're special)

- Added the TileFrictionBulldozerComponent for Ice crust and any other anchored (hypothetical) tileEntity which allows it to bulldoze the properties of the tile beneath

- Added a linked TileFrictionOverwrittenComponent that overrides tileDef values if an entity has it, which is given and modified by the TileFrictionBulldozerSystem

- Increased the fixture size of ice crust to be a full tile

- Conveyors now override tileDef friction values for movers. 

- The tileDef override compensates for differences in baseFriction to try and ensure you have the same friction as items moving on a conveyor.

- Added angularFriction to conveyors so items no longer spin. 

- Conveyors now check if the object they're trying to convey is on the ground before trying to convey it. 

- Conveyed entities now check if they're actually being conveyed before modifying their friction or acceleration

- Doesn't fix big objects getting stuck on conveyors or glue issue.

## Requirements

## Breaking changes

**Changelog**

:cl: 
- tweak: Conveyors have more friction
- fix: Items no longer spin on conveyors


Content.Shared/Friction/TileFrictionBulldozerComponent.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Content.Shared.Friction;
+
+/// <summary>
+/// This is used to let an entity bulldoze tileFriction for colliding entities with a specified value
+/// </summary>
+[RegisterComponent]
+public sealed partial class TileFrictionBulldozerComponent : Component
```
beck-thompson, line 7: This probably needs another name, I don't quite get what it means at a glance :laughing:

Content.Shared/Friction/TileFrictionBulldozerComponent.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    [DataField]
+    public float? Friction;
```
beck-thompson, line 10: These need comments

Content.Shared/Friction/TileFrictionBulldozerSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+        var uid = args.OtherEntity;
+
+        if (!TryComp<PhysicsComponent>(uid, out var physics) || !TryComp<TileFrictionOverwrittenComponent>(uid, out var tile))
+            return;
+
+        if (!CalculateTileFriction((uid, tile), physics))
+            RemComp<TileFrictionOverwrittenComponent>(uid);
+    }
+
+    private bool CalculateTileFriction(Entity<TileFrictionOverwrittenComponent> entity, PhysicsComponent? physics = null, EntityUid? ignore = null)
```
beck-thompson, line 59: This needs a comment - what is it trying to do?

Content.Shared/Friction/TileFrictionOverwrittenComponent.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Content.Shared.Friction;
+
+/// <summary>
+/// An entity with this component will have their tileFriction overwritten with these value
+/// </summary>
+[RegisterComponent]
+public sealed partial class TileFrictionOverwrittenComponent : Component
+{
+    [DataField]
+    public float? Friction;
```
beck-thompson, line 10: Comments

Content.Shared/Physics/Controllers/SharedConveyorController.cs
```diff
@@ -58,10 +60,29 @@ public override void Initialize()
...
+        if(!TryComp<FixturesComponent>(ent, out var fixture) || !IsOnConveyor((ent, fixture)))
+            return;
+
+        args.MobAcceleration = 1f;
+        args.Handled = true;
+    }
+
+    private void OnFrictionBulldozeEvent(Entity<ConveyedComponent> ent, ref FrictionBulldozeEvent args)
+    {
+        args.Friction = 10f;
```
beck-thompson, line 83: magic value?

Content.Shared/Physics/Controllers/SharedConveyorController.cs
```diff
@@ -155,7 +181,8 @@ public override void UpdateBeforeSolve(bool prediction, float frameTime)
                 // they'll go too slow.
                 if (!_mover.UsedMobMovement.TryGetValue(ent.Entity.Owner, out var usedMob) || !usedMob)
                 {
-                    _mover.Friction(0f, frameTime: frameTime, friction: 5f, ref velocity);
+                    _mover.Friction(0f, frameTime: frameTime, friction: 10f, ref velocity);
+                    _mover.Friction(0f, frameTime: frameTime, friction: 10f, ref angularVelocity);
```
beck-thompson, line 185: Is this the same magic value of 10 as the other one?

beck-thompson (changes requested):

Content.Shared/Friction/TileFrictionBulldozerComponent.cs
Pronana: Any good suggestions? It basically just replaces the friction of the tile (presumably) underneath it.

Content.Shared/Physics/Controllers/SharedConveyorController.cs
Pronana: Basically, we should probably have the Conveyed job return a friction value but that would be out of scope for this PR.

beck-thompson: Is it the same magic number as the other one?

Pronana: Yep

beck-thompson: If you truly can't somehow factor it out, then at least make it a constant in the system. That is sus but better than leaving the separate :laughing:

Content.Shared/Friction/TileFrictionBulldozerComponent.cs
beck-thompson: Make title friction override? When first looked at the PR, I was confused what the difference between this and the tileFrictionOverrideComp does because the names make them sound like they do the same thing

Content.Shared/Physics/Controllers/SharedConveyorController.cs
Pronana: I'll do that for all the current hard-coded values and add a note that they should be moved to the ConveyorComponent when the system inevitably gets another needed rewrite. 

Slam might(?) be working on one right now for their Xeno map, but I'm not sure.

Princess-Cheeseballs:
Closing this and the other fix, raising events and trycomps every tick is bad in the tileFrictionController. Instead tileFrictionController just needs to be rewritten to cache a lot of its modifiers and not raise events every tick.

PR #37478, opened 2025-05-15
Icecream for Vox

MissKay1994:
## About the PR
Minor bug fix for Vox

## Why / Balance
I tagged everything but frozen foods as Vox edible. 

## Technical details
Adds VoxFood tag to frozen food base in frozen.yml

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- fix: Vox now scream for Iceskreem


slarticodefast (changes requested):
I feel like we need to come up with a better solution to digestion than marking every single item with a tag for each species

beck-thompson:
Thanks for the PR! I was actually talking to someone last night about this, just tagging every food that doesn't work isn't a good solution. The issue is twofold:
1.) Its very easy to forget to add the tag on new food.
2.) Tags aren't inherited very well and, because food has a lot of tags, just adding the tag to the parents doesn't work very well.

The correct solution is to make the vox stomach whitelist non exclusive and only additive (e.g they eat all normal food AND also any food with their tags.) Someone is working on this at the moment so there should be a fix up soon!

PR #37483, opened 2025-05-15
rolebans respect IP exemptions

### LINKED ISSUE #37453: IP Exemptions do not work or get corrupted somehow
## Description
IP Exemptions do not make an account exempt from bans.

A ban hit a second player on the same network. Second player asked for and was granted an IP exemption, but reported that it does not work.
The issue was tested and reproduced on a staff member's main and test account. The exemption had no effect.

Some details of the issue are confusing/unclear
When I first tried to repro in dev, the issue did not manifest. This was confirmed multiple times. (This is why it was tested on live). After some unknown step (switching from an old version of stable to up-to-date master, or updating old stable) , the issue manifested. However, the issue does not go away even if I go back to older versions, even as far back as 2024 September. This suggests that the stored data of the ban or the exemption may have gotten corrupted, somehow.  But if the ban exemption is queried, it still shows
Image
I cannot account for how or why the issue could not be reproduced at first. Occam's razor says that the testing was somehow performed incorrectly, but I re-checked everything multiple times so I do not think this was the case

**Reproduction**

Ban Account-1, check IP, uncheck HW
Grant an IP exemption to Account-2
If Account-2 is exempt from ban, please report to confirm that I did not hallucinate it working at first
? ? ? switch branches or something I don't know
Account-2 is no longer exempt from the ban
Any subsequent exemptions are ineffective out of the gate

Errant-4:
## About the PR
Accounts can be given an IP Exemption, protecting them from bans targeting that IP address (so long as the Account or HWID in question is not also specifically targeted). However, currently this only works on server bans. PR extends this functionality to role bans as well

## Why / Balance
Missing functionality. There is no reason why IP Exemptions should bypass server bans but not role bans, the bypass was simply never added for the role ban codepath. It probably wasn't noticed until now because IP exemptions are quite rare to begin with.
Fixes #37453

## Technical details
Just passes the exemption data to the part where ban conditions are checked, and requires there to be no IP exemption to consider an IP ban valid.

## Media
https://github.com/user-attachments/assets/e9864c99-b330-450a-bbf9-647181a3dc69

## Requirements

## Breaking changes
None

**Changelog**
:cl: Errant
ADMIN:
- fix: Role bans now respect IP exemptions.

### DIFF
--- a/Content.Server/Database/ServerDbSqlite.cs
+++ b/Content.Server/Database/ServerDbSqlite.cs
@@ -219,8 +219,10 @@ namespace Content.Server.Database
             // So just pull down the whole list into memory.
             var queryBans = await GetAllRoleBans(db.SqliteDbContext, includeUnbanned);
 
+            var exempt = await GetBanExemptionCore(db, userId) ?? default;
+
             return queryBans
-                .Where(b => RoleBanMatches(b, address, userId, hwId, modernHWIds))
+                .Where(b => RoleBanMatches(b, address, userId, hwId, modernHWIds, exempt))
                 .Select(ConvertRoleBan)
                 .ToList()!;
         }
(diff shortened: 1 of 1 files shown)


IProduceWidgets:
Am I missing something? Would this need us to have placed IP-range rolebans? I can't think of any time we would do that.

also, we would want someone with an IP exemption for a gameban to still be roleban-able, right?
-- although this addresses that via the guid/hwid as long as we did those.

Errant-4:
> Am I missing something? Would this need us to have placed IP-range rolebans? I can't think of any time we would do that.
> 
> also, we would want someone with an IP exemption for a gameban to still be roleban-able, right? -- although this addresses that via the guid/hwid as long as we did those.

From what I understand. Situation:

Player A and Player B are on the same network. (Think the "my brother got banned" story, except it's true)
Player A gets banned. Their account, IP and HWID are banned.
Player B cannot play, because they are on the same IP.
Player B appeals and gets an IP exemption, because they did not do anything wrong. 
Player B can now bypass a server ban. But not a role ban, because role bans do not respectt IP exemptions
If Player B's ACCOUNT also gets rolebanned, their IP exemption will not protect them from it - an IP exemption only allows you to bypass an IP ban on an account that is not itself banned

This PR just makes it so the IP exemption will also work for the role ban

PR #37486, opened 2025-05-15
Fix EOF newline check

Quantum-cross:
## About the PR

Command substitution strips newline characters, so it's not actually comparing the last character against `\n`

## Why / Balance

Fix

## Technical details

I know it looks really dumb adding an x to the output here but it's required because command substitution will strip ALL trailing newlines, but we're exactly interested in trailing newlines!
We can trick it by ensuring that the output of the command substitution does not end in a newline, thus preserving any newlines before the 'x'.
See https://bash-hackers.gabe565.com/syntax/expansion/cmdsubst/

## Why does command substitution strip newlines?

I have a calculator program, `calc`. 

if I type `calc "4+4"` I expect to get 8.

If I type that in a terminal I expect it to look like:

```
clownOS$ calc "4+4"
8
clownOS$ 
```

In order to make it look like that, the `calc` program actually outputs `8\n`, if it only outputs `8` it will look like:

```
clownOS$ calc "4+4"
8clownOS$ 
```

Now say I want to use the result in a command substitution.

If command substitution DIDN'T strip newlines, it would look like this:

```
clownOS$ echo "The answer is $(calc "4+4")!"
The answer is 8
!
clownOS$ 
```

when you would actually expect

```
clownOS$ echo "The answer is $(calc "4+4")!"
The answer is 8!
clownOS$ 
```

So command substitution strips the newline so that you don't need to tell the application what context its running in to decide whether to output a newline.

## Requirements

## Breaking changes

none

### DIFF
--- a/.github/workflows/check-trailing-whitespace.yml
+++ b/.github/workflows/check-trailing-whitespace.yml
@@ -39,8 +39,13 @@ jobs:
 
           # Check for missing EOF newline
           if [ -f "$file" ] && [ -s "$file" ]; then
-            last_char=$(tail -c 1 "$file")
-            if [ "$last_char" != "" ] && [ "$last_char" != $'\n' ]; then
+            # I know it looks really dumb adding an x to the output here but it's required because
+            # command substitution will strip ALL trailing newlines, but we're exactly interested in trailing newlines!
+            # We can trick it by ensuring that the output of the command substitution does not end in a newline, thus
+            # preserving any newlines before the 'x'.
+            # See https://bash-hackers.gabe565.com/syntax/expansion/cmdsubst/
+            last_char=$(tail -c 1 "$file"; echo -n "x")
+            if [ "$last_char" != "x" ] && [ "$last_char" != $'\nx' ]; then
               echo "::error file=$file::Missing newline at end of file"
               has_missing_eof_newline=1
             fi


.github/workflows/check-trailing-whitespace.yml
```diff
@@ -39,8 +39,13 @@ jobs:
...
           if [ -f "$file" ] && [ -s "$file" ]; then
-            last_char=$(tail -c 1 "$file")
-            if [ "$last_char" != "" ] && [ "$last_char" != $'\n' ]; then
+            # I know it looks really dumb adding an x to the output here but it's required because
+            # command substitution will strip ALL trailing newlines, but we're exactly interested in trailing newlines!
+            # We can trick it by ensuring that the output of the command substitution does not end in a newline, thus
+            # preserving any newlines before the 'x'.
+            # See https://bash-hackers.gabe565.com/syntax/expansion/cmdsubst/
+            last_char=$(tail -c 1 "$file"; echo -n "x")
+            if [ "$last_char" != "x" ] && [ "$last_char" != $'\nx' ]; then
```
sowelipililimute, line 48: you can write this test as `[[ -n "$(tail -c 1 <"$file")" ]]` instead of doing this x at the end thing

sowelipililimute (changes requested):

.github/workflows/check-trailing-whitespace.yml
Quantum-cross: I wasn't sure if the shell in the action was running bash or not. I'll give it a try.

deltanedas:
why is this being done in a bash script instead of something like https://github.com/pre-commit/pre-commit-hooks/

slarticodefast:
> why is this being done in a bash script instead of something like https://github.com/pre-commit/pre-commit-hooks/

Would those also apply to web-edits? Because that is where most unintended whitespace changes come from.

.github/workflows/check-trailing-whitespace.yml
Quantum-cross: This specifically relies on the fact that the command substitution will strip the newline... which I agree works but it feels bad and it not very clear what is happening to anyone who looks at it.

I personally would like a solution that explicitly looks at the last character and says "yes this is a newline character"

deltanedas:
you can have an action or the dedicated ci run for all prs and it will fix webedit ops

slarticodefast:
I reverted the whitespace check for now, the script that was checking for which files were changed was causing false positives, confusing contributors.
We will need to come up with another approach for this in the future.

PR #37494, opened 2025-05-15
CC Genpop and AI Satellite

Alpha-Two:
My first mapping PR!

## About the PR
Major changes:
- Security now gets a small genpop system. The firing range is moved to the east, and other stuff is shuffled around to make room for it.
- There is now an AI satellite south of the Afterhours bar/science lounge, accessible by Command.
- The central hub is now NT-locked. The AI core remains there, but there is a new AI core in the satellite that the RD will be able to access.

Minor changes:
- Added two medkits to the Afterhours bar backroom.
- Added sleeper pods south of the docking bay
- Adjusted lights in some hallways
- Added a crew monitoring console to Telecoms, and a crew monitor to the central hub. (Doesn't actually work and shows server-not-found, but it's there anyway in case there's a fix for it later)
- Added station beacons all around the station. These are NOT warpable, to avoid cluttering the ghost warp list.

## Why / Balance
Genpop because genpop.

AI Satellite, welllllll. Emisse said that there's a very difficult balance between how all of CC interacts, which I understand. I believe my change to the AI satellite works because:
1. The idea of the AI having its own satellite fits better with most of the rest of the maps in-game.
2. Moving the "main" core that is accessible to players out of the hub allows the hub to be NT-locked.
    1. This is good as it prevents EORG getting too stale and just being "Run to central -> NT keycard -> DS gear -> Kill everyone"
    2. This shouldn't affect admin testing, as there will still be a secondary AI core in the hub.
    3. For admemes, I do not see why NT-locking the hub would cause issues, as it was already Command-locked.

## Technical details
Mapping changes 👍 

## Media
CentComm-0

## Requirements

## Breaking changes
Nuh uh

**Changelog**
:cl: UpAndLeaves
MAPS:
- add: Added an AI satellite to CC, south of the Afterhours bar.
- add: CC security now uses a genpop system.
- tweak: The CC hub is now locked to CentComm personnel only.


LaCumbiaDelCoronavirus:
why lock cc hub behind centcom access doe
also ai sat is 1 reinforced window and windoor away from ai it could be better than that

ArtisticRoomba (changes requested):
I'm letting Emisse hold the final decision on this one, but this is what I have to say.

- Making the hub CC only defeats the point of originally making it for command to inhabit at the end of the shift
- If there's problems with people instantly stealing the CC ID to run to the armory then remove the ID or change armory access to be inaccessible. Another alternative question is to ask: is this really a problem?
- CC AI core being a satellite defeats the purpose of it being in the hub: to still have the AI socializing with the crew at the end of the shift. Holopads around CC would help though.

Personally the changes that are perfectly fine are mini genpop and shooting range. Everything else I disagree with.

UpAndLeaves:
@ArtisticRoomba Do these changes seem good? (reverted AI changes, made central accessible to command again, put the shit in the locker)

PR #37498, opened 2025-05-16
Knives now do damage when thrown

liegend:
# About the PR

The kitchen knife and cleaver now deal damage when thrown.

## Why / Balance

a knife flies into you and you don't take damage, it's not realistic
the knife **won't** stick into the body and objects, just for the sake of realism.

## Technical details

added - type: DamageOtherOnHit in knife.yml
## Media

https://github.com/user-attachments/assets/f4acd0ad-557d-4d81-8a93-e2a7d735e6ef

## Requirements

- [X]I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/knife.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/knife.yml
@@ -46,7 +46,7 @@
   - type: DamageOtherOnHit
     damage:
       types:
-        Slash: 8
+        Slash: 9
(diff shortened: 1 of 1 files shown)


beck-thompson:
Thanks for the PR!

Its a fun idea but there are some balance issues. People used to use autoclickers to spam throw knives / `DamageOtherOnHit` objects before and it wasn't fun to deal with. Unfortunately adding it back to knives will lead to the same issue again.

PR #37502, opened 2025-05-16
Fix tile-events returns

metalgearsloth:
Should really be continues.


slarticodefast (changes requested):
AutomaticAtmosSystem
grafik
ExplosionSystem.Gridmap
grafik
PathfindingSystem.Grid
grafik
PinpointerSystem.Navmap
grafik
NavMapSystem
grafik
RequiresTileSystem
grafik

These all seem to have the same problem
see https://github.com/space-wizards/space-station-14/pull/37229/files

slarticodefast:
Pinging @Tayrtahn to confirm in case some are on purpose

slarticodefast (changes requested):

slarticodefast (approved):

Tayrtahn:
Good fixes! Thanks for cleaning these up!

PR #37507, opened 2025-05-16
Don't group damage in the HealthChange.cs guidebook text

EmoGarbage404:
## About the PR

Changes the chem guidebook text to not group damage in the normal `HealthChange` effect.

## Why / Balance

It's more obvious now with the `EvenHealthChange` effect, but healing damage is not based around groups and treating it as such is super misleading.

If a player sees someone has 20 brute damage and administers 10u of bicaridine, the actual effectiveness can vary from 100% to 33%, due to how the split damage is distributed. As such, it makes more sense to be specific about how the damage is split so that players can better understand how the healing works.

## Technical details

Just removes a bunch of complicated shitcode

## Media

old:
image

new:
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: The guidebook is now more explicit about how damage/healing from chemicals is split.


ScarKy0:
PEAK

PR #37513, opened 2025-05-16
Give VGRoid its name back

EmoGarbage404:
## About the PR

i made an oops when doing debris stuff.

## Why / Balance

makes it really hard to find when its far out which is lame.

## Technical details

/
## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Fixed the mining asteroid not having a name.

### DIFF
--- a/Resources/Prototypes/Entities/Stations/base.yml
+++ b/Resources/Prototypes/Entities/Stations/base.yml
@@ -110,9 +110,6 @@
           - type: Gravity
             enabled: true
             inherent: true
-          - type: IFF
-            flags: HideLabel
-            color: "#d67e27"
           protos:
           - VGRoid
 


IProduceWidgets:
Two thoughts:
- will make it easier to skip salvage progression until grappling hooks aren't infinite range.
- being slightly harder to find does make it more of a requirement to have jetpack etc, which might be desirable.

EmoGarbage404:
> Two thoughts:
> 
>     * will make it easier to skip salvage progression until grappling hooks aren't infinite range.
> 
>     * being slightly harder to find does make it more of a requirement to have jetpack etc, which might be desirable.

this is just the old behavior that was unintentionally modified.

PR #37515, opened 2025-05-16
Add debug power cells

ToastEnjoyer:
## About the PR
Added debug power cells and power cages.

## Why / Balance
These can be used for various things.

## Technical details
Added debug power cells/cages.

## Requirements

## Breaking changes

**Changelog**
N/A
-->


ScarKy0:
Won't just using a microreactor or antique cell have the same effect since they regenerate?

ToastEnjoyer:
> Won't just using a microreactor or antique cell have the same effect since they regenerate?

This is basically for mappers or other reasons, this is not supposed to be a normal obtainable item.

PR #37525, opened 2025-05-16
Force Wall Tweaks

keronshb:
## About the PR

Force Wall was tweaked to be more fair and to prevent cheesing with wall layers to be invulnerable.

## Why / Balance

Force Wall had some glaring issues:
- The despawn time was longer than the use action time
- If a wizard hid in the wall hitbox they couldn't be hit (they also couldn't interact out of it either)

Now the despawn timer is shorter than the use action time
12s => 10s
And the Action Use Delay is longer:
10s => 15s

A Special Wall Layer was added to compensate for this. It's almost the same as a Wall Layer except Impassible and InteractImpassible are removed from it. 
- Non-casters still can't go past it (including mice and flying creatures)
- It can't be shot through with bullets, or lasers, or any projectile from the wizard
- The wizard now CAN interact out of it, but they can also be hit while inside of it too

## Technical details

SpecialWallLayer was added to collision groups, it's similar to a Wall Layer except that anyone inside of it can be interacted with and can interact outside of it.

## Media

https://github.com/user-attachments/assets/d5aa226e-5d03-4fcc-849d-c8c599ec9878

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Force Wall timers changed so it despawns faster than it can be cast.
- tweak: Force Wall users can now interact while inside of the wall, but also can be attacked while inside of the wall.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Walls/walls.yml
+++ b/Resources/Prototypes/Entities/Structures/Walls/walls.yml
@@ -1301,30 +1301,30 @@
   id: WallForce
   name: force wall
   components:
-    - type: TimedDespawn
-      lifetime: 10
-    - type: Tag
-      tags:
-        - Wall
-    - type: Physics
-      bodyType: Static
-    - type: Fixtures
-      fixtures:
-        fix1:
-          shape:
-            !type:PhysShapeAabb
-            bounds: "-0.5,-0.5,0.5,0.5"
-          mask:
-            - FullTileMask
-          layer:
-            - WallLayer
-    - type: Airtight
-    - type: Sprite
-      sprite: Structures/Magic/forcewall.rsi
-      state: forcewall
-    - type: Icon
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/Entities/Structures/Walls/walls.yml
```diff
@@ -1301,30 +1301,30 @@
...
+  - type: Fixtures
+    fixtures:
+      fix1:
+        shape:
+          !type:PhysShapeAabb
+          bounds: "-0.5,-0.5,0.5,0.5"
+        mask:
+          - SpecialWallLayer
+        layer:
+          - SpecialWallLayer
```
slarticodefast, line 1320:
```suggestion
        mask:
        - SpecialWallLayer
        layer:
        - SpecialWallLayer
```

slarticodefast (approved):
Works as intended!

PR #37531, opened 2025-05-17
Atmos Air Grenades

ArtisticRoomba:
## About the PR
Adds air grenades, a portable grenade that rapidly fills a room with air. They have enough air in them to fill 30 tiles of pure vacuum up to 101.325 kPa (standard atmospheric) at room temp.

The reasoning behind having around 3000 mol of gas in the size of a grenade? Yuh huh.

## Why / Balance
Atmospherics doesn't have a lot of quick repair solutions available to combat spacing, and people are often dead or are in the process of dying as atmos techs frantically struggle to restore air to a room.

Air grenades provide a middleground solution. While it doesn't fill up most rooms to where everyone is breathing, it'll significantly reduce the barotrauma damage that they take, which is huge if spacing damage was actually real on upstream instead of it being a suggestion.

Air grenades aren't super powerful, as Atmos only starts with one of them per atmos tech. And there's only 2-3 atmos techs on station. Air grenades aren't purchasable right now because I want to see how powerful these turn out.

Your standard air canister currently contains less air than these grenades. This is bad! I'm going to buff those moderately in the near future.

In the future, a custom payload may be implemented, which will allow you to make your own custom grenades using a gas payload canner. This will allow Atmos techs to create their own custom mixes :trollface: for restoring air :trollface: in an emergency :trollface:.

## Technical details
Another update loop. I'm sorry Sloth.
Pretty naive implementation as well.

A new ReleaseGasOnTriggerSystem/Component has been added.

In short, the grenade is inactive until it's made active via a received trigger.

When triggered, it'll start releasing gas into the atmosphere. It updates once per second.

In order to prevent trolling forks who have space wind turned on, the gas can be optionally released as a ratio of the current gas mixture. An [exponential rise](https://www.desmos.com/calculator/fx9gfrhoim) can also be specified as taking ever so smaller ratios of the gas mix will lead to the grenade sitting there for a long time.

If there's no more gas to output, or there's no atmosphere to be found, it just nukes itself.

## Media
Sprite is temporary. Just putting this up for shitcode review.

Content Client_45A8PpiVV3

## Requirements

**Changelog**
:cl:
- add: Air grenades have been added! These can fill up a spaced room of ~30 tiles with fresh air. They can be found in Atmospheric Technician's lockers. Use them wisely!


PicklOH:
These are super cool! Awesome PR. I can't wait for the custom payload for totally safe and legitimate reasons.

NoElkaTheGod:
Why does a small grenade has 3000 mol of gas in it? Bluespace tech, of course.

slarticodefast:
More bereaking changes for my trigger refactor :death:
I guess I can fix these though.

EmoGarbage404:
black smoke is kinda strange. im kinda torn if it needs a proper visual or just like a hissing sound effect to communicate air is moving.

Resources/Prototypes/Entities/Objects/Weapons/Throwable/grenades.yml
```diff
@@ -461,6 +461,35 @@
...
 
+- type: entity
+  parent: [ BaseEngineeringContraband, GrenadeBase ] # Prevent inheriting DeleteOnTrigger from SmokeGrenade
+  id: AirGrenade
+  name: air grenade
+  description: A special solid state chemical grenade used for quickly releasing standard air into a spaced area. Fills up to 30 tiles!
+  components:
+  - type: Sprite
+    sprite: Objects/Weapons/Grenades/metalfoam.rsi
+  - type: SmokeOnTrigger
```
slarticodefast, line 472: I would remove the smoke and replace it by a visual effect with a timed despawn
you can use SpawnOnTrigger for that
or add a sprite layer for it

Content.Shared/Explosion/Components/OnTrigger/ReleaseGasOnTriggerComponent.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// for exponential rise calculations.
+    /// </summary>
+    [DataField]
+    public int TimesReleased;
+
+    /// <summary>
+    /// The gas mixture that will be released to the current tile atmosphere when triggered.
+    /// </summary>
+    [DataField]
+    public GasMixture Air { get; set; }
```
slarticodefast, line 72: getters and setters in components are no longer needed

Content.Server/Explosion/EntitySystems/ReleaseGasOnTriggerSystem.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            var fraction = comp.RemoveFraction ?? 1f;
+            var giverGasMix = comp.Air.RemoveRatio(fraction);
+            var environment = _atmosphereSystem.GetContainingMixture(uid, false, true);
+
+            if (environment == null)
+            {
+                QueueDel(uid);
+                continue;
+            }
```
slarticodefast, line 54: wouldn't this immediately delete the grenade when in space?

ArtisticRoomba: Yeah, it'll middle finger them if they throw it into general space. Undesirable?

Content.Server/Explosion/EntitySystems/ReleaseGasOnTriggerSystem.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        var curTime = _timing.CurTime;
+        var query = EntityQueryEnumerator<ReleaseGasOnTriggerComponent>();
+
+        while (query.MoveNext(out var uid, out var comp))
+        {
+            if (!comp.Active || comp.NextReleaseTime > curTime)
+                continue;
+
+            var fraction = comp.RemoveFraction ?? 1f;
```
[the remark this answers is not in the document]
ArtisticRoomba: I was... eeby and sleepy and not thinking right

Content.Server/Explosion/EntitySystems/ReleaseGasOnTriggerSystem.cs
slarticodefast: I think it's just a little weird if it directly disappears

ArtisticRoomba: I was planning on having the grenade exist afterwards and when the component is done just do remcomp

superjj18:
>which is huge if spacing damage was actually real on upstream instead of it being a suggestion

I feel like this is true only for slimes who prepare to handle spacing.  As a slime atmos it's actually fun to forgo the typical hardsuit and just use a winter jacket(at the risk of round removal via ashing/rads).  If ever I find a spaced area rather than running back to engi, getting a suit and slowly walking back, I instead just say "fuck it" and either fix the problem or die trying.  It's like a time trial where the cost of failure is drag of shame back to medbay.

Honestly that play style is definitely going to benefit the most from this item.  You would still need to fix the breach first but it would allow you to immediately stop dying once you do solve the issue.

Content.Shared/Explosion/Components/OnTrigger/ReleaseGasOnTriggerComponent.cs
```diff
@@ -0,0 +1,91 @@
...
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
+namespace Content.Shared.Explosion.Components.OnTrigger;
+
+/// <summary>
+/// Contains a GasMixture that will release its contents to the atmosphere when triggered.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+[AutoGenerateComponentPause]
+public sealed partial class ReleaseGasOnTriggerComponent : Component
```
EmoGarbage404, line 13: should add Access if you have a shared system for this

EmoGarbage404 (approved):
good to merge after the smallest comment ever

Content.Shared/Explosion/Components/OnTrigger/ReleaseGasOnTriggerComponent.cs
ArtisticRoomba: Local maintainer for the upstream project gets caught being a C# noob in 4K UHD

PR #34976, opened 2025-02-08
Stomach processing before bloodstream

nkokic:
## About the PR

Stomach now acts as a centrifuge and processes the chemicals before putting them into chemstream.
Fixed issue https://github.com/space-wizards/space-station-14/issues/26386 by replacing list of ReagentDeltas with list of Solutions which behave as an array of beakers, pipe or simply it now simulates intestines.

## Why / Balance

This finally allows a way for chemicals to be processed differently when injected into bloodstream or digested through stomach.
Stomach now doesn't need to know what chemical and for how long it has been inside. Does job regardless of what chemical it is and through what origin it has been through.

## Technical details

C#: 
Added reagent mixer specification when updating chemicals.
Removed list of ReagentDeltas and replaced with list of Solutions.
Size of solution list depends on digestionResolution which is calculated from DigestionDelay divided by UpdateInterval.
Calculation and list solution instancing is done once at the start. Used OnComponentInit event.
Each stomach update frame, stomach solution and transfers solution content to its next neighboring solution. Same with other solutions in list. Last solution in list transfers to body to be metabolized. This was done in reverse to avoid instant passing.

YAML: 
Attached reagent mixer, centrifuge, to every stomach.

## Media

Diagram of chemical flow. 5 units are transferred each 5 seconds to next solution in this example.
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Stomach can now processes chemicals differently.
- fix: Fixed stomach not processing reagents from reactions made in stomach.

PR #37553, opened 2025-05-17
fix: digest reagents created from in-stomach reactions

### LINKED ISSUE #25751: Drinking fresium with water creates ice that can't be digested
## Description
If you drink fresium and wash it down with a glass of water reaction creates ice inside of your stomach that doesn't digest for some reason. Furthermore if you drink some coffee/tea/green tea ice will mix into iced coffee/iced tea/iced green tea that also won't digest.

**Reproduction**
1. Drink 5u of freisum
2. Drink 5u of water
3. Drink ~60u of coffee
Now you are unable to drink or eat for the rest of the shift. Even rejuvenate command won't fix this. Only way to satiate yourself is syringes with nutriment/saline etc.

**Additional context**
Reasons mentioned above are my speculation because i couldn't find stomach component, but from testing I'm pretty sure that they are true.

### LINKED ISSUE #26386: Drinking reagents that can create reaction inside of your stomach makes resulting reagent impossible to digest
## Description
Originally described in https://github.com/space-wizards/space-station-14/issues/25751. But as it turns out ANY combination of reagents can lead to the same effect.

**Reproduction**
1. Put any combination of reagents that can mix into new reagent inside cryobeaker(i.e. 30 ambuzol 30 omnizine)
2. Drink this mixture
3. Wait until it's supposed to digest
4. Find out that it never digests
5. Now you can't eat or drink anything for the rest of the shift

perryprog:
## About the PR

If a reaction occurs in a stomach we should make sure we add the produced reagents to our tracked reagent deltas to ensure they get processed after an appropriate delay.

This does not handle removal of tracking of the consumed reactants, as this is meant to be a stopgap solution until something like #35071 is merged, as this can be a fairly significant bug. (Eat some chips and water, end up with a stomach that is permanently full.) The behavior without the removal is... "fine". Not totally accurate, but this is a rare scenario anyway.

## Why / Balance

Fixes #26386, fixes #25751.

## Media
<details>
<summary>Super exciting screenshots</summary>
image
image
</details>

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Body/Systems/StomachSystem.cs
+++ b/Content.Server/Body/Systems/StomachSystem.cs
@@ -3,6 +3,8 @@ using Content.Shared.Chemistry.EntitySystems;
 using Content.Shared.Body.Organ;
 using Content.Shared.Chemistry.Components;
 using Content.Shared.Chemistry.Components.SolutionManager;
+using Content.Shared.Chemistry.Reaction;
+using Content.Shared.Chemistry.Reagent;
 using Robust.Shared.Timing;
 using Robust.Shared.Utility;
 
@@ -20,6 +22,8 @@ namespace Content.Server.Body.Systems
             SubscribeLocalEvent<StomachComponent, MapInitEvent>(OnMapInit);
             SubscribeLocalEvent<StomachComponent, EntityUnpausedEvent>(OnUnpaused);
             SubscribeLocalEvent<StomachComponent, ApplyMetabolicMultiplierEvent>(OnApplyMetabolicMultiplier);
+            SubscribeLocalEvent<StomachComponent, ReactionAttemptEvent>(OnReactionAttempt);
+            SubscribeLocalEvent<StomachComponent, SolutionRelayEvent<ReactionAttemptEvent>>(OnReactionAttempt);
         }
 
         private void OnMapInit(Entity<StomachComponent> ent, ref MapInitEvent args)
@@ -30,6 +34,27 @@ namespace Content.Server.Body.Systems
         private void OnUnpaused(Entity<StomachComponent> ent, ref EntityUnpausedEvent args)
         {
             ent.Comp.NextUpdate += args.PausedTime;
+        }
+
+        private void OnReactionAttempt(Entity<StomachComponent> ent, ref ReactionAttemptEvent args)
+        {
+            if (args.Cancelled)
+                return;
+
+            // If our stomach has a reaction (e.g., table salt + water) which
+            // produces reagents, make sure we add that to our reagent deltas.
+            foreach (var (product, amt) in args.Reaction.Products)
+            {
+                ent.Comp.ReagentDeltas.Add(new StomachComponent.ReagentDelta(new ReagentQuantity(product, amt)));
+            }
+        }
+
+        private void OnReactionAttempt(Entity<StomachComponent> ent, ref SolutionRelayEvent<ReactionAttemptEvent> args)
+        {
+            if (args.Name != DefaultSolutionName)
+                return;
+
+            OnReactionAttempt(ent, ref args.Event);
         }
 
         public override void Update(float frameTime)


Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -45,6 +49,27 @@ private void OnEntRemoved(Entity<StomachComponent> ent, ref EntRemovedFromContai
...
+        }
+
+        private void OnReactionAttempt(Entity<StomachComponent> ent, ref SolutionRelayEvent<ReactionAttemptEvent> args)
+        {
+            if (args.Name != DefaultSolutionName)
+                return;
+
+            OnReactionAttempt(ent, ref args.Event);
+        }
+
```
Princess-Cheeseballs: image

Unfortunately this solution isn't going to work. This screenshot is from just drinking pure water.
This adds every possible reaction, even ones that won't happen to the stomach delta. 

It also happens during an attempt event, meaning even if you could verify the reaction would happen, it still could just get cancelled anyways (although currently there aren't enough things cancelling for that to be an issue. 

This is likely going to need a more complex solution as there isn't really an event you can hook into for after a reaction occurs, at least not one caused by chemical reactions that I could find. 

This will very likely require at least some rewriting of reaction code, which would be appreciated and I would also review since I'm working on entity effects right now which are tied heavily to reaction code currently.

perryprog: Wow, three months ago me was NOT cooking. In light of this maybe we just wait for metabolizing bloodstream, since that should fix this issue. I’ll probably do some poking around though just in case.

perryprog:
I redid the logic for tracking reagents to just be in response to any solution container updates, and that seems to work okay. It's not quite as simple as before but that's how it goes, I suppose.

Princess-Cheeseballs:
> I redid the logic for tracking reagents to just be in response to any solution container updates, and that seems to work okay. It's not quite as simple as before but that's how it goes, I suppose.

I'll try to look at it later today and give a full review. Considering this is reagent and metabolism code no matter the solution it's probably not gonna be great. But I we can hopefully the best not great solution.

Content.Shared/Chemistry/Reagent/ReagentId.cs
```diff
@@ -75,7 +75,21 @@ public override bool Equals(object? obj)
 
     public override int GetHashCode()
     {
-        return HashCode.Combine(Prototype, Data);
+        // We need to make we take the hash code of Data by value in order for
```
Princess-Cheeseballs: Minor spelling mistake.

Content.Shared/Chemistry/Reagent/ReagentId.cs
```diff
@@ -75,7 +75,21 @@ public override bool Equals(object? obj)
...
+            if (Data?.Count != 0)
+            {
+                foreach (var data in Data ?? [])
+                {
+                    hash = hash * 23 + data.GetHashCode();
+                }
+            }
+        }
+
+        return HashCode.Combine(Prototype, hash);
```
Princess-Cheeseballs: Curious as to why you changed this, were there issues with collision before that weren't noticed?

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -1,138 +1,167 @@
...
+    {
+        if (args.SolutionId != DefaultSolutionName)
+            return;
 
-        public bool CanTransferSolution(
-            EntityUid uid,
-            Solution solution,
-            StomachComponent? stomach = null,
-            SolutionContainerManagerComponent? solutions = null)
+        foreach (var reagent in args.Solution.Contents)
```
Princess-Cheeseballs, line 132: This works much better, still doesn't delete reagents but as you said this is a stop gap measure.

I'm gonna think on this for a bit, but I'm leaning more towards having a method that removes unused deltas even if it's gonna suck.

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -1,138 +1,167 @@
...
-                var queue = new RemQueue<StomachComponent.ReagentDelta>();
-                foreach (var delta in stomach.ReagentDeltas)
+            foreach (var (id, deltas) in stomach.ReagentDeltas.ToList())
+            {
+                // Get the deltas with their timestamps incremented. Note we
+                // can't just set a timespan that the reagents are digested
+                // at, since our update interval may change after the delta
+                // was added.
+                var nextDeltas = deltas
+                    .Select(delta => delta with { Lifetime = delta.Lifetime + stomach.AdjustedUpdateInterval })
```
Princess-Cheeseballs: This doesn't really work as advertised. Every upadate of x time interval, we add x timer interval to the lifetime. You would expect a 10x slower update interval would result in 10x slower digestion, but it just means that we add 10x the amount of time when we update. The Adjusted interval should just be the base interval *if* we're using TimeSpan.

Content.Shared/Body/Components/StomachComponent.cs
```diff
@@ -6,87 +6,81 @@
...
+/// A simple pair of a reagent's quantity and the time it's been present in the
+/// stomach.
+/// </summary>
+/// <remarks>
+/// We can't have it be the timespan the reagent should be digested since the
+/// amount the lifetime it gets incremented by is actually dependent on
+/// <see cref="StomachComponent.AdjustedUpdateInterval"/>, which can change at
+/// any time.
+/// </remarks>
+public record struct ReagentDelta(ReagentQuantity Reagent, TimeSpan Lifetime);
```
Princess-Cheeseballs: Alright if we're gonna be cleaning this up, here's a not so cursed idea. 

Instead of accumulating time, what if the LifeTime was an integer that counted the number of times the stomach updated. That's *basically* what it does already. 

DigestionDelay would just be a DigestionCycle counter that is equal to 20, and every time digestion updates the counter gets reduced by 1. 

We're not calculating this based on the time the food enters the stomach, just when the stomach updates so there's not really a reason to use a time accumulator from my perspective.

When you hit 0, the food is digested and we remove it. 

It's either that or you make it a timespan based on CurTime + DigestionDelay, and account for the metabolic multiplier when stomach updates.

Content.Shared/Chemistry/Reagent/ReagentId.cs
perryprog: I've run into it a few times before (I literally grabbed this off another open PR I have), but hashcodes of lists are by reference (it's just the default object.GetHashCode) because ????????????? so you need to make sure you hash it by value if you want to use it with value equality. You can test by reverting that and seeing how it breaks; it's very fun to debug :)

Thanks for the review; will address tomor—well, today. In a number of hours!

Princess-Cheeseballs: Ah so it was broken? Doesn't surprise me, if you could give me reproduction steps that would help so I can properly catalogue this bug for the maints. It might be better to spin this off if it's bad enough.

perryprog: It's only an issue if you use a ReagentId in something that uses hashed lookups, which I think only I've been doing funnily enough. (The other PR I needed it for was for reagent dispensers I think.)

```c#
Dictionary<ReagentId, string> stuff = [];
var refId = new ReagentId("foo", null);
stuff.Add(refId, "bar");
stuff.ContainsKey(new ReagentId("foo", null)); // false
stuff.ContainsKey(refId); // true
stuff.Keys.Contains(new ReagentId("foo", null)); // false
stuff.Keys.Contains(refId); // true
stuff.First().Key == new ReagentId("foo", null); // TRUE??????
```

Princess-Cheeseballs: I think it's worth spinning off into its own PR just so it's not bogged down in stomach code. Good catch and good fix! The FixedPoint2 addition can stay here since it's just API stuff.

perryprog: Okie lemme get that real quick.

Princess-Cheeseballs: Hero.

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: Ha, you're totally right—this is a bug with what's live as well!

Princess-Cheeseballs: Good thing metabolism is changed so rarely it basically doesn't matter.

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: I might be brain farting, but I'm not seeing the issue—currently this does get rid of unused deltas, since they're removed once their lifetime expires regardless of if they referred to a sensical amount of solution.

Content.Shared/Body/Components/StomachComponent.cs
perryprog: I think incrementing lifetime by base digestion update and having the stomach update once every AdjustedUpdateInterval should be enough, right?

-----

> We're not calculating this based on the time the food enters the stomach

I mean, we sort of are—the units of everything you're suggesting are still units of time, if I'm understanding you right.

If we wanted to do this correctly I think we'd have to "attempt" to record the expected digestion time $d_t$ of each reagent delta and when the metabolic rate $r$ is updated you recalculate the time that all deltas should have by $r(d_t-t)$. (We'd also have to of course handle OnUnpaused ourselves.)

Princess-Cheeseballs: > I mean, we sort of are—the units of everything you're suggesting are still units of time, if I'm understanding you right.

Not really, because amount of time we're in the stomach is dedicated by the update interval. This is much more obvious if you slow metabolism down.

If we have a base update interval of 10 seconds for the stomach, and we eat some food which has a lifetime of 20 seconds to digest, the amount of time it actually takes to digest varies between about 10-20 seconds.

Let's say we eat a piece of food at the same tick a stomach update happens. 10 seconds are now added to the delta lifetime despite 0 seconds passing, and in 10 seconds we'll add another 10 seconds resulting in digestion completing in 10 seconds. If we eat food lets say 5 seconds into a cycle, we get an actual digestion time of 15 and so on.

As a result, we're not actually tracking time but a fixed amount of stomach updates before we pass the reagent into the bloodstream. This is fine honestly. With an update interval of 1 second it doesn't really matter, especially since it takes 20 cycles to pass to bloodstream.

The alternative, while more accurate would be harder to do: Kill update interval and store the next update based on the time the next reagent that needs to enter bloodstream. You'd have to manipulate the delta timespans manually based on metabolism changes and pauses for this to work, and frankly, I don't think the extra effort is worth it.

Content.Shared/Body/Systems/StomachSystem.cs
Princess-Cheeseballs: We're removing them in the stomach update loop, but in this system if there's less of a reagent than we expect, we just continue and we also aren't removing reagents if they don't exist.

Not sure it's worth the extra effort to do the removal considering the stomach update loop can handle it itself, I'd just not worry about it for now.

Content.Shared/Body/Components/StomachComponent.cs
perryprog: Alright, I think I see what you're saying—I would still argue that it's best represented by TimeSpans, it's just that there are definitely inaccuracies inherent to an update loop that make it a little silly.

While your alternative would be more accurate, that's a lot more overhead since you then have to iterate through every single stomach's dictionary of ReagentDeltas _every single tick_. For what I'm saying in my last paragraph of the previous reply, you'd still have an update interval of every second (and always every second) for each stomach, before you then compare the timespans to see which have expired.

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: Are you sure? I'm not seeing that in testing. Adding a delta with `stomach.ReagentDeltas.Add(new("asdf", null), [new(new("asdf", 5), TimeSpan.Zero)]);` sees the delta removed after twenty seconds like expected.

Princess-Cheeseballs: I might be miscommunicating, the stomach is removing the deltas correctly within the update loop. 

The thing I'm mulling over is whether or not we should remove deltas when the solution updates and they're no longer relevant.

Content.Shared/Body/Components/StomachComponent.cs
Princess-Cheeseballs: Considering that we aren't tracking time accurately for the deltas because we're confined to accuracy of the stomach update loop, it would probably just be better to have the lifeTime be an integer for "number of updates" rather than TimeSpan.

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: OH. Okay, sorry, I'm with you now.

Uhhhhhhhh yeah I think it's "fine".

Content.Shared/Body/Components/StomachComponent.cs
perryprog: Hm. I think at that point I'd rather go the route of just storing the digestion time and updating that as needed—honestly it shouldn't be that bad, and it might make things clearer.

Content.Shared/Body/Systems/StomachSystem.cs
Princess-Cheeseballs: Neither solution is good so I think I'm gonna say it's fine as well. The best solution is to rework how reactions and metabolism works (They really need it), and doing that is gonna probably require rewriting pieces of this anyways. 

Leaving the deltas in sucks, but so does removing them so if another maint reviews and asks for them to be removed I won't disagree with em.

Content.Shared/Body/Components/StomachComponent.cs
Princess-Cheeseballs: Brave. Good luck.

perryprog: Wait frick it might need custom serialization hooks too; gah. Doing things the right way is the WORST.

perryprog:
One funny detail I noticed is that if you chow down on something on a stasis bed, then get up, the digestion times of all the new deltas gets adjusted to be a tenth of a second apart, since, uh, math.

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -1,138 +1,188 @@
...
-            var query = EntityQueryEnumerator<StomachComponent, OrganComponent, SolutionContainerManagerComponent>();
-            while (query.MoveNext(out var uid, out var stomach, out var organ, out var sol))
-            {
-                if (_gameTiming.CurTime < stomach.NextUpdate)
-                    continue;
+            if (curTime < stomach.NextUpdate)
+                continue;
 
-                stomach.NextUpdate += stomach.AdjustedUpdateInterval;
+            stomach.NextUpdate += stomach.UpdateInterval;
```
Princess-Cheeseballs, line 69: Because we're still doing this, the extra logic for tracking ingested reagents accurately from time they're added isn't really being seen. 

If we're gonna go the full mile and track reagents accurately, have custom handlers for pausing and stuff, then we need to kill UpdateInterval completely, and structure around that. 

Otherwise it's just better to use an int for "UpdateCycles" and accept the mild inaccuracies.

Princess-Cheeseballs (changes requested):
Leaving the update interval in, but tracking time accurately means we have the opposite issue. Assuming our update interval is 10 seconds (from a stasisbed), and a digestion time of 20 (although it would be more like 200 in our case but less numbers good). 

If we update at 0, 10, 20, 30, ect. and eat an item at the 1 second mark with our item being digested at 21 seconds, then it'll take until the 30 second mark for the item to digest, adding 9 seconds to digestion time.

So before where we would end up removing time, we now add time.

perryprog:
I switched the update loop to be always one second (regardless of metabolic multiplier), so it's only an error of at most that, and I feel like that's reasonable?

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: I'm not following what you're saying, sorry.

Princess-Cheeseballs:
> I switched the update loop to be always one second (regardless of metabolic multiplier), so it's only an error of at most that, and I feel like that's reasonable?

That's reasonable yeah, but in that case I don't think we need to accurately track when the food entered. TimeSpan is really only good if we need accurate time, but our accuracy is down to the second (or whatever the update interval is).

Content.Shared/Body/Systems/StomachSystem.cs
Princess-Cheeseballs: Basically, having reagents store an accurate timespan for when they should update doesn't really make sense if the accuracy of our update loop is limited by UpdateInterval (which is 1 second by default).

perryprog:
Isn't the sub-second accuracy still necessary for food eaten off a stasis bed that you then get onto (or emagged but opposite)? With quantization to a second you would have ticks where it digests a bunch of bites at once instead of having each being (doafter duration times ten) apart.

Princess-Cheeseballs:
> Isn't the sub-second accuracy still necessary for food eaten off a stasis bed that you then get onto (or emagged but opposite)? With quantization to a second you would have ticks where it digests a bunch of bites at once instead of having each being (doafter duration times ten) apart.

Wouldn't it still do the same if we just tracked digestion cycles? You eat 10 burgers within 1 digestion cycle, the next cycle all 10 get updated. All that changes is the time in between cycles. 

Whether or not we use TimeSpan for the reagent deltas it will still behave like this.

perryprog:
Okay, let me make sure I'm understanding your idea correctly—the circumstance I'm imagining is:

1. Person eats two bites of burger, with like 0.9 seconds between each bite, before then lying down on a stasis bed, all before the update interval is reached.
2. Once the next update interval is reached, the digestion cycles for each burger gets incremented.
3. This continues and the burgers begin digestion _at the same time_.

With what I have now, you would have seen the bites get digested 9–10 seconds apart.

Princess-Cheeseballs:
> Okay, let me make sure I'm understanding your idea correctly—the circumstance I'm imagining is:
> 
>     1. Person eats two bites of burger, with like 0.9 seconds between each bite, before then lying down on a stasis bed, all before the update interval is reached.
> 
>     2. Once the next update interval is reached, the digestion cycles for each burger gets incremented.
> 
>     3. This continues and the burgers begin digestion _at the same time_.
> 
> 
> With what I have now, you would have seen the bites get digested 9–10 seconds apart.

Ah I see what you're saying, yeah this would result in more accurate behavior because we're always updating in 1 second intervals versus increasing the intervals to 10 seconds and updating based on integers.

perryprog:
By the way, I unironically love how in-depth we're going for this being mechanically consistent for something that really "does not matter" in a certain sense. It's so silly but I love it, so thanks. :P (The horror of a burger bite being digested five seconds later than it should!!)

Princess-Cheeseballs:
> By the way, I unironically love how in-depth we're going for this being mechanically consistent for something that really "does not matter" in a certain sense. It's so silly but I love it, so thanks. :P (The horror of a burger bite being digested five seconds later than it should!!)

If we wanted full mechanical depth, instead of using update interval, have the next update be determined by the next reagent delta that should be processed. Would require changing how you iterate through them a bit though.

perryprog:
Yeah, that'd definitely be best but I feel like that really adds a lot of processing overhead for a benefit at that point even I can't justify. (You're iterating through every single stomach's reagent delta every single tick!)

Princess-Cheeseballs:
> Yeah, that'd definitely be best but I feel like that really adds a lot of processing overhead for a benefit at that point even I can't justify. (You're iterating through every single stomach's reagent delta every single tick!)

It would only be every tick if you aren't setting next update properly. In reality it would be more like every [DoAfterTime] modified by metabolic rate modifiers, but yeah I don't think it's worth it, arguably what we're doing here isn't worth it either but c'est la vive.

perryprog:
Oh, I see what you mean—but yeah, it'd still be a lot more overhead (though I wonder if it gets balanced out by how many update loops that run where every stomach has zero reagent deltas). Anyway, in light of all that, anything you still want me to adjust?

Princess-Cheeseballs:
> Oh, I see what you mean—but yeah, it'd still be a lot more overhead (though I wonder if it gets balanced out by how many update loops that run where every stomach has zero reagent deltas). Anyway, in light of all that, anything you still want me to adjust?

Doing a proper review now so await changes on that. If you want the extra challenge go ahead and try that and test it but after I do my review.

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -1,138 +1,188 @@
...
 
-            return true;
+    public bool TryTransferSolution(
+        EntityUid uid,
+        Solution solution,
+        StomachComponent? stomach = null,
+        SolutionContainerManagerComponent? solutions = null)
+    {
+        if (!Resolve(uid, ref stomach, ref solutions, logMissing: false)
+            || !_solutionContainerSystem.ResolveSolution((uid, solutions), DefaultSolutionName, ref stomach.Solution)
```
Princess-Cheeseballs: CanTransferSolution already checks for this stuff, so we shouldn't have to resolve twice if we can help it...

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -1,138 +1,188 @@
...
+    {
+        return Resolve(uid, ref stomach, ref solutions, logMissing: false)
+               && _solutionContainerSystem.ResolveSolution((uid, solutions), DefaultSolutionName, ref stomach.Solution, out var stomachSolution)
+               // TODO: For now no partial transfers. Potentially change by design
+               && stomachSolution.CanAddSolution(solution);
+    }
 
-            return true;
+    public bool TryTransferSolution(
+        EntityUid uid,
```
Princess-Cheeseballs: Entity\<T> typing

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -1,138 +1,188 @@
...
-            }
+    public bool CanTransferSolution(
+        EntityUid uid,
+        Solution solution,
+        StomachComponent? stomach = null,
+        SolutionContainerManagerComponent? solutions = null)
+    {
+        return Resolve(uid, ref stomach, ref solutions, logMissing: false)
+               && _solutionContainerSystem.ResolveSolution((uid, solutions), DefaultSolutionName, ref stomach.Solution, out var stomachSolution)
+               // TODO: For now no partial transfers. Potentially change by design
```
Princess-Cheeseballs: This comment is old and inaccurate, we can do partial transfers after IngestionSystem refactor.

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -1,138 +1,188 @@
...
+    }
 
-            _solutionContainerSystem.TryAddSolution(stomach.Solution.Value, solution);
-            // Add each reagent to ReagentDeltas. Used to track how long each reagent has been in the stomach
-            foreach (var reagent in solution.Contents)
-            {
-                stomach.ReagentDeltas.Add(new StomachComponent.ReagentDelta(reagent));
-            }
+    public bool CanTransferSolution(
+        EntityUid uid,
```
Princess-Cheeseballs: Entity\<T> typing.

Princess-Cheeseballs (changes requested):
It's amazing, I look at this system and I don't feel total despair. Good job.

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: Oh sick hell yeah

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: Or I just delete the other method :)

perryprog:
> If you want the extra challenge go ahead and try that

I'll be real all I want is my friends to stop asking me to break out SCSI magic whenever they eat chips and water :P (at least adjreagent works now).

Princess-Cheeseballs:
> > If you want the extra challenge go ahead and try that
> 
> 
> 
> I'll be real all I want is my friends to stop asking me to break out SCSI magic whenever they eat chips and water :P (at least adjreagent works now).

One day reagent code will be good.

Content.Shared/Body/Systems/StomachSystem.cs
```diff
@@ -158,31 +158,29 @@ private void UpdateDigestionTimes(Entity<StomachComponent> ent, Func<TimeSpan, T
...
+            || !_solutionContainerSystem.ResolveSolution((ent, ent.Comp2),
+                DefaultSolutionName,
+                ref ent.Comp1.Solution,
+                out var stomachSol)
+            || stomachSol.AvailableVolume == FixedPoint2.Zero)
             return false;
-        }
 
-        return _solutionContainerSystem.TryAddSolution(stomach.Solution.Value, solution);
+        _solutionContainerSystem.AddSolution(ent.Comp1.Solution.Value, solution);
```
Princess-Cheeseballs, line 178: I think we can just use TryAddSolution and not have to check the stomachsol for available volume, leaving just the resolve.

Princess-Cheeseballs (changes requested):
Small thing.

Content.Shared/Body/Systems/StomachSystem.cs
perryprog: I don't think so, since TryAddSolution only adds it if the solution fully fits—AddSolution is needed for partial transfers. I suppose the available volume check isn't really needed since that's included in AddSolution.

Princess-Cheeseballs: Hm, well its current and only use is in ingestion system which should check that there's enough volume for the split it creates, so I guess it's fine. Means we at least get some of the solution if we fail. 

Honestly now I want to rewrite that part of ingestion system and just kill this API but that's out of scope.

perryprog:
Me four days ago:

> It’s a pretty simple fix

past me is such a liar

Princess-Cheeseballs:
> Me four days ago:
> 
> > It’s a pretty simple fix
> 
> past me is such a liar

NEVER trust anything reagent related to be simple.

nkokic:
Hello @perryprog
I used to work on the issue, except I completely avoided working with reagents and just used [solution "tunnel"](https://github.com/space-wizards/space-station-14/pull/34976).
With that I was also working on PR preventing replenishing blood by drinking since the future [metabolising bloodstream PR](https://github.com/space-wizards/space-station-14/pull/35071) would somewhat enable it.

Now I want to ask you, are there any concerns that this [PR](https://github.com/space-wizards/space-station-14/pull/40066), which lets stomach breakdown chems, might be incompatable with your PR. It relies on stomach updating chemicals, but with a mixer parameter. To me it looks like it should work out very nicely, but I want to ask anyway.

perryprog:
Hi! I don't have time to look into it in detail, but really the only thing this PR is meant to properly fix is stomachs not digesting reagents created from in-stomach reactions (e.g., if you drink water and eat some salt, the saline that creates should be digested instead of permanently reducing your stomach's capacity).

ThatGuyUSA:
Is this PR obsoleted by #42419 ?

Princess-Cheeseballs:
Yeah. Obsolted.

PR #37558, opened 2025-05-18
Fix prisoner closet UI not working

EmoGarbage404:
## About the PR

Fixes a crash that occurs when opening the prisoner closet UI.

## Technical details

@iaada for future reference

The `GenpopLockerMenu` class attempts to use a dependency for `IConfigurationManager`. However, since this is not an EntitySystem or a class that automatically injects dependencies, we must first call `IoCManager.InjectDependencies(this);` in order to actually populate the class dependencies. This class, since it didn't use any before, did not do this, which caused a Null Reference Exception when trying to call GetCvar.

## Requirements

## Breaking changes

**Changelog**

no cl no fun

### DIFF
--- a/Content.Client/Security/Ui/GenpopLockerMenu.xaml.cs
+++ b/Content.Client/Security/Ui/GenpopLockerMenu.xaml.cs
@@ -16,14 +16,11 @@ public sealed partial class GenpopLockerMenu : FancyWindow
 
     public event Action<string, float, string>? OnConfigurationComplete;
 
-    // CCVar.
-    private int _maxIdJobLength;
-
     public GenpopLockerMenu(EntityUid owner, IEntityManager entMan)
     {
         RobustXamlLoader.Load(this);
+        IoCManager.InjectDependencies(this);
 
-        _maxIdJobLength = _cfgManager.GetCVar(CCVars.MaxIdJobLength);
 
         Title = entMan.GetComponent<MetaDataComponent>(owner).EntityName;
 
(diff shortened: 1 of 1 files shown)


Content.Client/Security/Ui/GenpopLockerMenu.xaml.cs
```diff
@@ -16,14 +16,11 @@ public sealed partial class GenpopLockerMenu : FancyWindow
 
     public event Action<string, float, string>? OnConfigurationComplete;
 
-    // CCVar.
-    private int _maxIdJobLength;
-
     public GenpopLockerMenu(EntityUid owner, IEntityManager entMan)
     {
         RobustXamlLoader.Load(this);
+        IoCManager.InjectDependencies(this);
```
metalgearsloth, line 22: This is part of why I think all controls should have blank ctors so we can just have a test sweep all controls, and construct them just to check their dependencies actually loaded.

iaada:
Oh shit, that's my mistake sorry. I did know about injecting dependencies, but I did not give enough thought to it when I update my PR.

EmoGarbage404:
> Oh shit, that's my mistake sorry. I did know about injecting dependencies, but I did not give enough thought to it when I update my PR.

it's alright. it's a pretty easy thing to miss and i don't really blame you for it. Just something to keep in mind while working on UI.

Tayrtahn:
-mutters something about integration tests-

PR #37560, opened 2025-05-18
Rebuild ATS Anchor room and add indestructible spine.

Minemoder5000:
## About the PR

Rebuilt the ATS Anchor room to add a turret and shuttle miners.
## Why / Balance

Prevents anyone from getting into the ATS anchor and stealing the generators.
Prevents the buy/sell pallets from being moved closer to the station by cutting off the anchor.
Gas miners helps repressurize the ATS.
## Technical details

Modified the trading_outpost.yml
## Media

2jyQnpt
hVhfh0L

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: Minemoder
MAPS:

- tweak: The ATS station anchor room was rebuilt to prevent stealing from the room, and to add shuttle miners to repressurize the ATS.

PR #31667, opened 2024-08-31
Increase gas canister volumes (take 2)

IProduceWidgets:
## About the PR

Previously gas canisters didn't hold enough gas to suitably pressurize any meaningful amount of tiles. This has led to them being mostly ignored for use in fixing spacings, as well as not fulfilling their intended uses aboard shuttles and evac shuttles.

## Why / Balance

To make gas canisters useful.
This will likely have some strange balance implications due to how this will interact with the more dangerous gases, but I think making them useful for the main distro gases is more important, and we can come up with ways to hinder their misuse to accompany this change.

What follows is [a discord blurb](https://discord.com/channels/310555209753690112/1008709214006427689/1256275687652720661)
> I did some back of the napkin canister exploration today, here are my findings, and a recommendation based on some things you mentioned the other day.
>
>A tile at common temp takes a little over 100 mols of 78/22 air mix to reach 101kpa. 
>
>current canisters hold 1871.71 (weird number) mols at that temp by default. This is the relevant metric I think despite the ability to change temp to augment that because a hotter or colder mix will also quickly make it unsuitable for repressurizing areas due to temperature damage.
>
>canisters at common temp can *almost* repressurize 18 tiles to 101kpa but quickly falls off at anything higher than that.
>
>We could bump this up and make canisters more useful. Ideally I think canisters should be able to pressurize *most* rooms on their own. For that I think we likely need *at least* a 4x increase in mols & volume (here they are tightly coupled due to restricting temp).
>
>I've added a quick visual of how large these multiples would *feel* in game.
image
image
> Something I did not explore is how this would effect canister bombs, but Im not sure those are even something we want long term, and this is far more important imo.
> Something I found that doesn't work in the current context is using canisters as station pressure banks. Unfortunately pipenets are orders of magnitude larger than canisters. For context Oasis' pipenet (admittedly a very large station, but thats where you would want a volume buffer the most due to the longer time to repressurize the pipenet and larger number of rooms/tiles) is 3460x larger than a canister.
> So if we wanted pressure banks, they would need to be new, probably immobile at least when pressurized, entities.
> Canisters should work as decent pressure banks for shuttles though with my proposed changes. Higher multiples preferred however for that application, especially as the pipenets currently start empty of gas and so any connected canister on a shuttle grid will initially fill the pipenet and require changing nearly immediately.

## Requirements

[...]

PR #37564, opened 2025-05-18
Buff gas canister volumes moderately

ArtisticRoomba:
## About the PR
Gas canister volumes have been buffed to 1500 L. For reference, they used to have a volume of 1000 L. The volume of a standard tile is 2500 L.

The prices of these canisters have increased respectively, how they are still an attractive option.

## Why / Balance
Gas canisters are in a poor state when talking about player rewards. A player expects a gas tank to fill a very large air volume when dumped to the atmosphere! After all, they contain ~1871 mols of gas at 4500 kPa, 20C.

This is not the case. They currently only have enough volume to fill about 18 tiles. This is poor, as air grenades fill 30 tiles!

This PR buffs canisters so that regular air canisters at 4500 kPa can fill rooms (to 101.325 kPa) up to ~26 tiles. This makes these canisters an attractive option for refilling rooms quickly, as slogging that canister across the station, or buying them from Atmospherics, pays off. Also note that oftentimes the goal of hastily dumping a canister is to not fill the room up to a perfect, livable atmosphere, but to fill it to the point where people stop taking extreme barotrauma damage.

Do note that Atmospheric Technicians can fill these canisters up to 9000 kPa using a volumetric pump, enabling even higher molar amounts, leading to more tiles filled!

This PR is an alternative to another PR that extremely buffs canisters beyond the capacity of a standard tile and introduces mini canisters, which I think is unrealistic and too extreme of a buff, when Atmospherics can simply fill their cans to a higher pressure.

## Technical details
YAML

## Media
Not necessary, but I did testing

## Requirements

**Changelog**
:cl:
- tweak: The volume of large gas canisters have been increased to 1500L to encourage their usage in resolving pressure problems in spaced rooms. The gas that starts inside of the tanks and the price of them has increased to compensate.

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_atmospherics.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_atmospherics.yml
@@ -4,7 +4,7 @@
     sprite: Structures/Storage/canister.rsi
     state: grey
   product: AirCanister
-  cost: 300
+  cost: 600
   category: cargoproduct-category-name-atmospherics
   group: market
 
@@ -14,7 +14,7 @@
     sprite: Structures/Storage/canister.rsi
     state: blue
   product: OxygenCanister
-  cost: 300
+  cost: 600
   category: cargoproduct-category-name-atmospherics
   group: market
 
@@ -34,7 +34,7 @@
     sprite: Structures/Storage/canister.rsi
     state: red
   product: NitrogenCanister
-  cost: 300
+  cost: 500
   category: cargoproduct-category-name-atmospherics
   group: market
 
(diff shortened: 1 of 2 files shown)


murolem:
ATS unspacing finally here

ArtisticRoomba:
ats unspacing is being untrolled in #37560

LaCumbiaDelCoronavirus:
This is what liquid o2/n2 canisters are for: see https://github.com/space-wizards/space-station-14/pull/36247#issuecomment-2770754783

slarticodefast (approved):
Seems reasonable to me.
I didn't like the previous PR with the big canisters because that one allowed them to store 4 times the gas amount of the tile they are standing in (I know, something something bluespace, but it's unintuitive).
I'll leave it open for a while to allow others to give feedback before merging.

See https://github.com/space-wizards/space-station-14/pull/31667

ArtisticRoomba:
Liquid canisters will be useful in the future when Atmospherics wants to spend hard earned money on liquid oxygen canisters instead of emergency equipment, when they don't wait to wait for their gas miner to mine gas. But that's far in the future.

IProduceWidgets:
https://github.com/space-wizards/space-station-14/pull/31667

Issue #37552, opened 2025-05-17
Items can be picked up through windows.


VerinSenpai:
## Description
As the title suggests. An item can be picked up through a window if it is close enough to it.

**Reproduction**
Place a line of windows. Something you'd be certain you couldn't normally reach around.
Place down an item, any item. Putting the item down by the window should cause it to repel from the window a touch.
Pick the item up from the other side.
Profit.

**Screenshots**

https://github.com/user-attachments/assets/2b748a93-c01f-421c-8b32-b14934ee2bb0

VerinSenpai:
Update. Can be picked up through walls too, if you can see the tip of it.

![Image](https://github.com/user-attachments/assets/c888ab9f-cc27-4fa5-a0f1-f64f6705e7d9)

PR #37570, opened 2025-05-18
Fix item stuck check

metalgearsloth:
Items can no longer be picked up through windows

Fixes https://github.com/space-wizards/space-station-14/issues/37552

:cl:
- fix: Fixes being able to grab items through walls.

### DIFF
--- a/Content.Shared/Interaction/SharedInteractionSystem.cs
+++ b/Content.Shared/Interaction/SharedInteractionSystem.cs
@@ -34,6 +34,7 @@ using Robust.Shared.Input.Binding;
 using Robust.Shared.Map;
 using Robust.Shared.Network;
 using Robust.Shared.Physics;
+using Robust.Shared.Physics.Collision.Shapes;
 using Robust.Shared.Physics.Components;
 using Robust.Shared.Physics.Systems;
 using Robust.Shared.Player;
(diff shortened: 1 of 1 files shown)


Content.Shared/Interaction/SharedInteractionSystem.cs
```diff
@@ -854,7 +856,20 @@ private Ignored GetPredicate(
...
+                        !_physicsQuery.TryComp(otherEnt, out var otherBody) ||
+                        !otherBody.CanCollide ||
+                        ((int) collisionMask & otherBody.CollisionLayer) == 0x0)
+                    {
+                        continue;
+                    }
+
+                    ignored.Add(otherEnt);
+                }
             }
```
metalgearsloth, line 873: Old code would ignore any intersecting ents which is bad because those intersecting ents (like a wall) may be in the way instead of just checking the item's position which is generally more flexible on knowing if it's in a seam.

slarticodefast (commented):
I'm testing right now and I can no longer grab items from inside a window like it allowed you to do before
isn't that exactly what this code is there for?
a range of 0.01 only works if they are almost perfectly on top of each other
grafik

slarticodefast:
Ok, apparantly that is a different code path.
Also the predicate is collected twice, once for the target and once for the origin.
Works as intended.

VerinSenpai:
Don't suppose this impacts soap being able to slide inside of wall tiles and be picked up from the other side?

VerinSenpai:
Don't suppose this impacts soap being able to slide inside of wall tiles and be picked up from the other side?

VerinSenpai:
Not sure why that posted twice, also just seen slartis comment. Ignore me

VerinSenpai:
> I'm testing right now and I can no longer grab items from inside a window like it allowed you to do before
> isn't that exactly what this code is there for?
> a range of 0.01 only works if they are almost perfectly on top of each other
> grafik
> 
Also, that was finicky before anyhow. You could grab items in a window if it had been flung into it by something but not if a window had been rcd'd on top of it. I think id prefer it if items lodged in walls got pushed out if I'm being honest 😅

PR #37572, opened 2025-05-18
Fixed talking when sleeping while on crit

zHonys:
## About the PR

Fixed bug where a crited player could talk if used the sleep action

## Why / Balance

Bugfix

## Technical details

I changed SleepingSystem, added two lines to check if the MobState is "Alive" on OnSpeakAttempt method

## Media
https://github.com/user-attachments/assets/f1d227ce-32fb-422a-86ad-535de3367ab3

## Requirements

**Changelog**

:cl:
- fix: Falling asleep while crit will no longer cause you to say whatever is typed in your chatbox.

### DIFF
--- a/Content.Server/Damage/ForceSay/DamageForceSaySystem.cs
+++ b/Content.Server/Damage/ForceSay/DamageForceSaySystem.cs
@@ -4,6 +4,7 @@ using Content.Shared.Damage.Events;
 using Content.Shared.Damage.ForceSay;
 using Content.Shared.FixedPoint;
 using Content.Shared.Mobs;
+using Content.Shared.Mobs.Components;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.Stunnable;
 using Robust.Shared.Player;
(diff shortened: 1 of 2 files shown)


EmoGarbage404 (dismissed):
ok this is actually a really weird bug to wrangle so i had to dig.

Your issue is actually in `DamageForceSaySystem`, specifically `TryForceSay`. When your sleep state changes, it forces you to send whatever is in your message bar (and makes an exception so that you can have final words and whatnot).

The issue is that it doesn't check that you are actually able to send the message _prior_ to the state change. So what the code actually needs to do is make sure you can speak before it tries to force you to say something in `TryForceSay`.

zHonys:
Yeah there are a couple places you could fix this from, i choose `SleepingSystem` because if seemed fitting. I'll look at `DamageForceSaySystem` later today. It is pretty weird and I think it has existed since ForceSay was introduced in 2023, but by being so specific no one had reported it yet

SlamBamActionman (approved):
I've tested this a bit and I think the code holds up. Checked with stuns and that doesn't seem to trigger when crit.

Arguably you should not be able to choose to fall asleep in the first place? Maybe? Regardless, this solves other extremely niche cases like being noct'd while crit. Thank you for your contribution!

zHonys:
Oh nice

PR #37573, opened 2025-05-18
Fix weather audio persisting even when associated comp is deleted

### LINKED ISSUE #35081: Audio from planet/map weather persists even after the associated WeatherComponent is deleted
## Description
When there is active weather on a map, deleting the associated _(server-side)_ `WeatherComponent` doesn't stop weather audio (if you are on the map and are able to hear it in the first place), meaning you will hear it forever.
 Did not test if this effect remains after reconnecting.

**Reproduction** 
0. Have working audio..
1. Find id of the map (not grid, but map) that you're on
2. Set weather of that map to anything of your choice other than `'null'`
3. Optionally wait the full hardcoded 15 seconds for weather to fully transition from clear weather
4a. Press F7 and go to objects
4b. Select 'Maps'
4c. Right-click on the map with corresponding map id and open View Variables
5. Go to Server Components tab
6. Search for `WeatherComponent`
7. Delete `WeatherComponent` (by pressing the X on the right-hand side)
8. Notice that the audio of the weather stays and changes to full volume regardless of whether you are in an indoor space (where it is usually muffled),  *the actual weather effect is removed*

**Screenshots**
Would add media but I can't do so right now.

**Additional context**
You can also notice this issue with #34973 (steps to test are provided in the comments), by FTLing early from the expedition planet and noticing persisting audio despite having left the planet

LaCumbiaDelCoronavirus:
## About the PR
See title
fix also applies for weather audio persisting when leaving maps with weather that get deleted right after you leave (aka expeds)

## Why / Balance
bugfix
Fixes #35081

## Technical details
added OnWeatherRemoved on weathersystem: calls EndWeather when the comp is shutting down
and changed a call for SharedAudioSys from .Stop(EntUid stream) to .SetState(EntUid stream, AudioState.Stopped) [i had to do it]

## Media
n/a

## Requirements

### DIFF
--- a/Content.Client/Weather/WeatherSystem.cs
+++ b/Content.Client/Weather/WeatherSystem.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Numerics;
 using Content.Shared.Light.Components;
 using Content.Shared.Weather;
@@ -24,6 +25,7 @@ public sealed class WeatherSystem : SharedWeatherSystem
     {
         base.Initialize();
         SubscribeLocalEvent<WeatherComponent, ComponentHandleState>(OnWeatherHandleState);
+        SubscribeLocalEvent<WeatherComponent, ComponentShutdown>(OnWeatherRemoved);
     }
 
     protected override void Run(EntityUid uid, WeatherData weather, WeatherPrototype weatherProto, float frameTime)
@@ -136,6 +138,7 @@ public sealed class WeatherSystem : SharedWeatherSystem
         return true;
     }
 
+    private void OnWeatherRemoved(EntityUid uid, WeatherComponent component, ref ComponentShutdown args) => component.Weather.ToList().ForEach(w => EndWeather(uid, component, w.Key));
     private void OnWeatherHandleState(EntityUid uid, WeatherComponent component, ref ComponentHandleState args)
     {
         if (args.Current is not WeatherComponentState state)
--- a/Content.Shared/Weather/SharedWeatherSystem.cs
+++ b/Content.Shared/Weather/SharedWeatherSystem.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using Content.Shared.Light.Components;
 using Content.Shared.Light.EntitySystems;
 using Content.Shared.Maps;
@@ -27,6 +28,7 @@ public abstract class SharedWeatherSystem : EntitySystem
     {
         base.Initialize();
         _blockQuery = GetEntityQuery<BlockWeatherComponent>();
+
         SubscribeLocalEvent<WeatherComponent, EntityUnpausedEvent>(OnWeatherUnpaused);
     }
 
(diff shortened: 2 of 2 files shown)


Content.Client/Weather/WeatherSystem.cs
```diff
@@ -1,3 +1,4 @@
+using System.Linq;
```
EmoGarbage404, line 1: unused import statement?

Content.Shared/Weather/SharedWeatherSystem.cs
```diff
@@ -232,6 +235,8 @@ protected virtual bool SetState(EntityUid uid, WeatherState state, WeatherCompon
         return true;
     }
 
+    private void OnWeatherRemoved(EntityUid uid, WeatherComponent component, ref ComponentShutdown args) => component.Weather.ToList().ForEach(w => EndWeather(uid, component, w.Key));
```
EmoGarbage404, line 238: should use Entity<T> version as per modern standards.

Content.Shared/Weather/SharedWeatherSystem.cs
```diff
@@ -216,9 +219,9 @@ protected virtual void EndWeather(EntityUid uid, WeatherComponent component, str
         if (!component.Weather.TryGetValue(proto, out var data))
             return;
 
-        _audio.Stop(data.Stream);
-        data.Stream = null;
         component.Weather.Remove(proto);
+        _audio.SetState(data.Stream, Robust.Shared.Audio.Components.AudioState.Stopped, true); // _audio.Stop(); is cooked here and doesn't work
+        data.Stream = null;
```
EmoGarbage404, line 224: what part of it is cooked? it just deletes the audio entity. changing this to a SetState call just orphans the audio ent and never cleans it up.

EmoGarbage404 (changes requested):

Content.Shared/Weather/SharedWeatherSystem.cs
LaCumbiaDelCoronavirus: it never stops, I looked at all the other methods that called Play on it and they never happened after the last time i tried to stop the audio so i assumed it was a problem with Stop(). would like to know what else I could do to fix it other than this

FYI breakpoints always showed data.Stream as not-null

metalgearsloth: This seems like an engine bug that should be fixed still afaict?

LaCumbiaDelCoronavirus: > This seems like an engine bug that should be fixed still afaict?

From what I've debugged I think its a problem that ive exclusively seen in weather that occurs with `.Stop()` not being in an update loop -- more specifically the second early return in `.Stop()`, where it always early returns in this scenario as `IsFirstTimePredicted` is almost always false whenever `.Stop()` gets called from weather being ended.

LaCumbiaDelCoronavirus:
fuck oops

PR #37578, opened 2025-05-18
Limit shuttle collision damage

Errant-4:
## About the PR
The velocity of a shuttle now gets reduced much more quickly, when destroying tiles. This means that high-speed ramming will make significantly "shallower" areas of destruction, but low-speed impacts don't get much less dangerous for your own ship than before

## Why / Balance
With a maximum velocity of 60, heavier ships could potentially delete an entire department's worth of the station. While this is _really cool_, it is generally not good for roundflow.

## Technical details
cvar adjustment

## Media
edit: Actual value was reduced to 8 so damage is somewhat higher than shown on the middle image

ramming

## Requirements

## Breaking changes
shuttle.impact.slowdown was increased from it's debut value of 0.8 to 10, greatly decreasing the potential maximum damage of a high-speed high-mass impact.

**Changelog**
:cl: Errant
- tweak: High-energy shuttle impacts now deal much less damage.

### DIFF
--- a/Content.Shared/CCVar/CCVars.Shuttle.cs
+++ b/Content.Shared/CCVar/CCVars.Shuttle.cs
@@ -258,7 +258,7 @@ public sealed partial class CCVars
     /// </summary>
     [CVarControl(AdminFlags.VarEdit)]
     public static readonly CVarDef<float> ImpactSlowdown =
-        CVarDef.Create("shuttle.impact.slowdown", 0.8f, CVar.SERVERONLY);
+        CVarDef.Create("shuttle.impact.slowdown", 10f, CVar.SERVERONLY);
 
     /// <summary>
     /// Minimum velocity change from impact for special throw effects (e.g. stuns, beakers breaking) to occur
--- a/Content.Shared/CCVar/CCVars.Shuttle.cs
+++ b/Content.Shared/CCVar/CCVars.Shuttle.cs
@@ -258,7 +258,7 @@ public sealed partial class CCVars
     /// </summary>
     [CVarControl(AdminFlags.VarEdit)]
     public static readonly CVarDef<float> ImpactSlowdown =
-        CVarDef.Create("shuttle.impact.slowdown", 10f, CVar.SERVERONLY);
+        CVarDef.Create("shuttle.impact.slowdown", 8f, CVar.SERVERONLY);
 
     /// <summary>
     /// Minimum velocity change from impact for special throw effects (e.g. stuns, beakers breaking) to occur


ArtisticRoomba (dismissed):
This changes CVAR defaults which is a breaking change(?)

Errant-4:
> This changes CVAR defaults which is a breaking change(?)

I mean, this has been out for like 8 hours

Errant-4:
Breaking change note added

millionse:
Fun removal

Ilya246:
i think it being as destructive as i put it is fine, since shuttles capable of doing real damage via ramming are rare, the only one that actually matters is probably nukie ship
and i think nukie ship should be able to do ram ops effectively, and if it's too strong, could add some sort of alert that goes off if a grid with sufficient inertia is approaching at high speed, even if invisible

Errant-4:
I'm tinkering with the numbers some more

Errant-4:
> i think it being as destructive as i put it is fine, since shuttles capable of doing real damage via ramming are rare, the only one that actually matters is probably nukie ship and i think nukie ship should be able to do ram ops effectively, and if it's too strong, could add some sort of alert that goes off if a grid with sufficient inertia is approaching at high speed, even if invisible besides, even then grids with large inertia are really hard to accelerate (nukie ship can only barely get to 20 speed i think), so reward for effort

With the current numbers, the cargo shuttle can delete the entire bridge on Saltern. There is no reason to introduce that amount of griefing potential for basically no gain other than "it is conceptually cool that something like that could happen in a situation where it was allowed"
The nukies have the "justification" for using such a tool, so there, the issue is rather the huge power shift this would cause: they should not be able to wipe out basically the _entire_ Security (if it's by the station edge) or Supply department and everyone within, at the bargain cost of a ship they probably would not have used later anyway, and _maybe_ sacrificing one nukie (but if you move the helm to the back of the ship, maybe not even that!)
As for the practicality of ramming with the Infiltrator, with some moderate effort you could rearrange the thrusters and gain significantly quicker acceleration. If the ship is going for ramming, there is not really a reason not to do so.

I agree that there are more interesting ways to deal with the risk that heavy ramming could cause, but none of those are coded yet and I do not think we should leave things as-is until that time. If any downstreams disagree with the number we will settle on, it is trivially easy to change it in the server config

ArcaneOcto:
What of the destruction was dialed back to just like a 3rd of the original value? The nerf proposed here seems slightly extreme to me tbh.

NoElkaTheGod:
I'd agree that the nerf is a bit extreme. Being able to delete an entire department is not cool, but so is only being able to delete like 1 room.

Emisse:
nukie shuttle plowing a single department is objectively funny but do we know how often this kinda thing is getting used in vulture, any statistics or anything yet

ScarKy0:
> nukie shuttle plowing a single department is objectively funny but do we know how often this kinda thing is getting used in vulture, any statistics or anything yet

So far its used often cause its a new feature. But realistically its not really practical cause youll definitely lose a nukie or 2 in the collision unless you prepare properly

Errant-4:
> Good PR. Even with this nerf, given the visual above it's still gonna be a hell of a time for engineers to seal breaches, let alone have Atmos or even get machines/computers back to what they were. Additionally, collisions already put the cargo shuttle at risk. Good for engineers using the mapped shuttle parts to fix the shuttle, bad for engineers now having to tell salvagers that they can't give them shuttle parts.

Yeah one issue is that the breaches they make are often going to be wide, and that's going to happen no matter how much this number would get increased. And some bridges have the important consoles right in front of the window (see Saltern), so those will be gone even if just a shuttle decides to say hi. Perhaps that is a mapping consideration that we will just have to adapt to

IProduceWidgets:
> > This changes CVAR defaults which is a breaking change(?)
> 
> I mean, this has been out for like 8 hours

Its not even on stable yet afaik.

Emisse:
> Fun removal

dont make comments like these

superjj18:
It's been like eight years and the holdo maneuver is still getting shit on lmao

Actually kinda cool though that nukies can literally just drive straight into the hos' office, certainly an entrance

Errant-4:
We have decided to increase the damage a bit.
We might revisit this number once we have better mechanics/features to counter such tactics, but for now, this is necessary.

PR #37580, opened 2025-05-18
Cryo-OverHaul re-licensed by neuPanda for Wizden MIT

neuPanda:
## About the PR
re-submitting PR as properly licensed under MIT rather than the AGPL3 it was originally
adds improvements to cryopods and findings from the last years worth of feedback 

## Why / Balance
avoid legal daruma
improve cryo pods making them useful

## Technical details
updated cryopod systems and fixed code licensing

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: cryopods and chems
- fix: Licensing issue

### DIFF
--- a/Content.Server/Body/Components/MetabolizerComponent.cs
+++ b/Content.Server/Body/Components/MetabolizerComponent.cs
@@ -55,12 +55,12 @@ namespace Content.Server.Body.Components
         public bool RemoveEmpty = false;
 
         /// <summary>
-        ///     How many reagents can this metabolizer process at once?
+        ///     How many poisons can this metabolizer process at once?
         ///     Used to nerf 'stacked poisons' where having 5+ different poisons in a syringe, even at low
         ///     quantity, would be muuuuch better than just one poison acting.
         /// </summary>
-        [DataField("maxReagents")]
-        public int MaxReagentsProcessable = 3;
+        [DataField("maxPoisons")]
+        public int MaxPoisonsProcessable = 3;
 
         /// <summary>
         ///     A list of metabolism groups that this metabolizer will act on, in order of precedence.
--- a/Content.Server/Body/Systems/MetabolizerSystem.cs
+++ b/Content.Server/Body/Systems/MetabolizerSystem.cs
@@ -146,7 +146,7 @@ namespace Content.Server.Body.Systems
             var list = solution.Contents.ToArray();
             _random.Shuffle(list);
 
-            int reagents = 0;
+            int poisons = 0;
             foreach (var (reagent, quantity) in list)
             {
                 if (!_prototypeManager.TryIndex<ReagentPrototype>(reagent.Prototype, out var proto))
@@ -163,9 +163,9 @@ namespace Content.Server.Body.Systems
                     continue;
                 }
 
-                // we're done here entirely if this is true
-                if (reagents >= ent.Comp1.MaxReagentsProcessable)
-                    return;
+                // Already processed all poisons, skip to the next reagent.
+                if (poisons >= ent.Comp1.MaxPoisonsProcessable && proto.Metabolisms.ContainsKey("Poison"))
+                    continue;
 
 
                 // loop over all our groups and see which ones apply
@@ -223,8 +223,9 @@ namespace Content.Server.Body.Systems
                 {
                     solution.RemoveReagent(reagent, mostToRemove);
 
-                    // We have processed a reagant, so count it towards the cap
-                    reagents += 1;
+                    // We have processed a poison, so count it towards the cap
+                    if (proto.Metabolisms.ContainsKey("Poison"))
+                        poisons++;
                 }
             }
 
--- a/Content.Server/Medical/CryoPodSystem.cs
+++ b/Content.Server/Medical/CryoPodSystem.cs
@@ -111,7 +111,10 @@ public sealed partial class CryoPodSystem : SharedCryoPodSystem
                     continue;
                 }
 
-                var solutionToInject = _solutionContainerSystem.SplitSolution(containerSolution.Value, cryoPod.BeakerTransferAmount);
+                // Filter out a fixed amount of each reagent from the cryo pod's beaker
+                var solutionToInject = _solutionContainerSystem.SplitSolutionReagentsEvenly(containerSolution.Value, cryoPod.BeakerTransferAmount);
+                //  for every .25 units used, .5 units per second are added to the body, making cryo-pod more efficient than injections
+                solutionToInject.ScaleSolution(cryoPod.PotencyMultiplier);
                 _bloodstreamSystem.TryAddToChemicals(patient.Value, solutionToInject, bloodstream);
                 _reactiveSystem.DoEntityReaction(patient.Value, solutionToInject, ReactionMethod.Injection);
             }
--- a/Content.Shared/Chemistry/Components/Solution.cs
+++ b/Content.Shared/Chemistry/Components/Solution.cs
@@ -580,8 +580,10 @@ namespace Content.Shared.Chemistry.Components
         }
 
         /// <summary>
-        /// Splits a solution without the specified reagent prototypes.
+        /// splits the solution taking the specified amount of reagents proportionally to their quantity.
         /// </summary>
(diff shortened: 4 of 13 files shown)


ScarKy0:
Thanks a bunch!

ScarKy0:
> Please fill out the about section and the technical details properly so we know what changes to expect

Yeah thats what im working on right now, im atomizing it out and gonna fill out info afterwards.
The PR started off like this to show the code is under MIT

Cojoke-dot:
Maybe put what you changed about cryo in the changelog lol, players aren't going to be able to see the code

ScarKy0:
> Maybe put what you changed about cryo in the changelog lol, players aren't going to be able to see the code

Read what I said above

ScarKy0:
no words can express how much i dislike that the extension does all discards seperately

EnrichUranium:
I disagree on the statement that cryo pods are not currently useful.   Also, if you give me free chemicals for just sticking someone in the pod, I'm keeping that at room temperature and using that instead of syringes.

northerndakotas:
So with the syringe it took ~15 seconds to heal 15 damage including the injection time. With cryo it took ~12 second to heal 15 damage, and ~22 seconds to heal 30 damage including the time it took to put them in the pod. Each subsequent injection will add an extra 5 seconds to non-cryo treatment, while further chems in the pod add no additional time. Both procedures used the same amount of medicine, but cryo healed twice as much. (Note: crit patients take injections faster so that would probably even out the time difference between the two.)

To summarize, this would make cryopods faster, easier, more effective, and more efficient. 

Now I would personally be all for making cryopods the premier space age healing method, since it's the one from all the movies. But, I do want to make sure this is what you meant to do.

neuPanda:
the initial intent was to ensure that proper credit for my creations were given with the understanding that once they are licensed here rather than having been copied and pasted thus making the copy here AGPL3 and improperly credited you can change what you want with it. However, it needs to be properly licensed 

to expand on the content its self

these chems have been play tested on multiple ship and station servers for the last year and i have been working on trying to understand the needs of them. there are a few tweaks that need to be made, primarily the potency of Alox as its complexity versus how much it healed made it less than desirable.  initially there was pushback against SalicylicAcid but it was quickly understood that this actually added value, threw its complexity it made the chem feel more advanced.  further the complexity on Trauma and Alox decrease the "fuck it just throw everybody in a freezer, and instead make it so that other forms of healing are still viable as the cryo chems are more resource heavy.

as for the benefits of cryopods, yes they are faster, easier, more effective, and efficient. this is by design as they are advance treatment systems designed for such. however there is normally only one or two to a station, they require an atoms configuration, and assuming your using the cryo chems, require temperature management of the patient as to not re-damage them when they come out.

what i have found across the multiple servers i maintain that i tried to license here was that while cryo is stronger, it is not so strong as to replace mass casualties or low to medium impact injuries.  it shines against those who's bodies are ravaged, and even with the ability to use standard chems which means lepro would reset thermals, people still prefer to use the needle approach or hypo and reserve cryo for the dead who are beyond hope.  i find it being about equal to surgery in down streams in that reguard.

finally there were some changes to chemical makeup that was part of this change, Oppo being one of them, there was a recent change here that reset the oppo recipe in the down streams that i maintain the cryo on that angered a lot of people.  the recipe uses 2 parts formaldehyde, which is silver plus vinegar which is ethanol plus oxygen

walksanatora:
remember we also just merged a ton of new cryo chems and rebalanced cryox.

whatston3:
> I disagree on the statement that cryo pods are not currently useful. Also, if you give me free chemicals for just sticking someone in the pod, I'm keeping that at room temperature and using that instead of syringes.

The compromise position on Frontier with this was creating a Cryo-specific metabolism and limiting the chemicals that this pod worked with to those.

There is genuinely no point to use anything apart from cryo pods when they increase the yield of every chemical.

I maintain my position from the first time you PRed this: https://github.com/new-frontiers-14/frontier-station-14/pull/1443#issuecomment-2164117853

And still think that the compromise from https://github.com/new-frontiers-14/frontier-station-14/pull/1533 was warranted.  This is busted as-is.

neuPanda:
> > I disagree on the statement that cryo pods are not currently useful. Also, if you give me free chemicals for just sticking someone in the pod, I'm keeping that at room temperature and using that instead of syringes.
> 
> The compromise position on Frontier with this was creating a Cryo-specific metabolism and limiting the chemicals that this pod worked with to those.
> 
> There is genuinely no point to use anything apart from cryo pods when they increase the yield of every chemical.
> 
> I maintain my position from the first time you PRed this: [new-frontiers-14/frontier-station-14#1443 (comment)](https://github.com/new-frontiers-14/frontier-station-14/pull/1443#issuecomment-2164117853)
> 
> And still think that the compromise from [new-frontiers-14/frontier-station-14#1533](https://github.com/new-frontiers-14/frontier-station-14/pull/1533) was warranted. This is busted as-is.

and i still disagree with you and have evidence from several servers to support my clames.  in fact the argument i have gotten from many is they dislike the way cryo works on frontier compared to my implementation.  you all nerfed cryo before it had a chance to even be tested. i had to take them to other servers to test them there and in general the request if for them to be stronger, i mentioned the difference in the cryo pod efficiency and the note is that its not worth the time to treat that way due to the limited number of pods, there is not enough benefit especially in mass casualty situations.



----------------------------------------------------------------------------------


> remember we also just merged a ton of new cryo chems and rebalanced cryox.

this was code that was merged in but not properly licensed, the code merged in was copied from an AGPL3 code base without a link to it

whatston3:
> its not worth the time to treat that way due to the limited number of pods

Triage situations are fair, granted - beyond that, assuming you have sufficient chemicals and time, your optimal play becomes cryo pods and only cryo pods - precisely measure your dosage of whatever it is, insert, and bugger off.

aeoli-fly:
This is a poor way to rebalance cryo imo, the video itself shows this. These work at the same pace as a syringe injection, just twice as efficient. You might as well just keep cryopods at room temperature and shove people in there for regular treatment, with syringes as backup in case you have more than 2 patients. Cryo is already strong, it's getting stronger and more mechanically nuanced with the various added chemicals, this just removes nuance from an already simplistic gameplay loop and makes cryo the best treatment option in any given situation.

SpaceLizard24:
This does seem far too strong. There's no reason to make cryo cold anymore, theres no reason not to to get an empty beaker, fill it with all the chems you need, throw someone into cryo and forget about them. This gives no reason to use a syringe over a cryopod unless the player simply doesn't know, needs to get dexalin or epi in the patient immediately, or just they don't have access to a pod. 

Tldr this change is too powerful.

SpaceLizard24:
> that its not worth the time to treat that way due to the limited number of pods
Sci can very easily make cryo pods, the only challenge is the piping, which can be easily done because you don't need to cool the pod. 
And imo balancing something around the fact "there is a limited amount of it" when it's super easy to get more isn't great balance

Pumkin69:
For this change to not be very strong you need to make it so every chemical has a cryo metabolism so you need to be below freezing cold for the chemical to be stronger. That way people wont overuse cryo because having to strip winter clothing off and then get warmed up after you get treated sucks way more than getting stuck with a syringe and walking out.

I could be completely wrong and the change could be fine but it seems like people will abuse it like cloning.

superjj18:
The idea that cryo would become meta because of this change is hilarious.  It could make chems 3x as potent and even then I'd doubt it would overtake syringe healing.  

In emergencies I've seen medbays heal command members with topicals despite the round start cryoxodone beakers being right next door because the cryopods weren't even worth the mental energy to turn on compared to just doing it the old fashion way.

Syringe: fast and precise

Cryo: slow and efficient, gives patient a massive speed debuff

Please god give medbay and cryo some content, the most useful cryochems are essentially paywalled behind either ridiculously expensive recipes, have super niche use-cases, or both.

At very least the cryopods would be worth turning on simply because chem will yell at medbay to stop wasting the expensive chems on non-cryo'd patients.

I do think low temp should be required though.  Not exactly "cryo" without cold.

At very least test merge this

ScarKy0:
Since this PR has been merge conflicted for >3 months im opting to close it.
Also with discomed soonish cryo might actually become good again. so lets hope for that.

PR #37582, opened 2025-05-18
Metabolism metabolizes all reagents, limits poisons only

ScarKy0:
## About the PR

Atomized out of #37580

Metabolizers now metabolize all reagents, instead limiting the amount of different poisons that can be metabolized at a time.

## Why / Balance

`Used to nerf 'stacked poisons' where having 5+ different poisons in a syringe, even at low quantity, would be muuuuch better than just one poison acting.`
This was the original intention behind MaxReagents. So why limit the amount of reagents and not by poison?
This is a very minor change and probably wont even be noticable.

## Technical details

Renamed MaxReagents to MaxPoisons. Only count reagents in MaxReagents if they are in the Poison group.

## Media

https://github.com/user-attachments/assets/405aa3c5-904a-4bfb-b05c-80b7c4adb1fd

## Requirements

## Breaking changes

`MaxReagentsProcessable` field in MetabolizerComponent has been renamed to `MaxPoisons`.

**Changelog**

:cl: neuPanda
- tweak: Metabolism now metabolizes all reagents at once, only limiting the amount of poisons.

### DIFF
--- a/Content.Server/Body/Components/MetabolizerComponent.cs
+++ b/Content.Server/Body/Components/MetabolizerComponent.cs
@@ -55,12 +55,12 @@ namespace Content.Server.Body.Components
         public bool RemoveEmpty = false;
 
         /// <summary>
-        ///     How many reagents can this metabolizer process at once?
+        ///     How many poisons can this metabolizer process at once?
         ///     Used to nerf 'stacked poisons' where having 5+ different poisons in a syringe, even at low
         ///     quantity, would be muuuuch better than just one poison acting.
         /// </summary>
-        [DataField("maxReagents")]
-        public int MaxReagentsProcessable = 3;
+        [DataField]
+        public int MaxPoisons = 3;
 
         /// <summary>
         ///     A list of metabolism groups that this metabolizer will act on, in order of precedence.
--- a/Content.Server/Body/Systems/MetabolizerSystem.cs
+++ b/Content.Server/Body/Systems/MetabolizerSystem.cs
@@ -165,7 +165,7 @@ namespace Content.Server.Body.Systems
                 }
 
                 // Already processed all poisons, skip to the next reagent.
-                if (poisons >= ent.Comp1.MaxPoisons && proto.Metabolisms.ContainsKey(PoisonGroup))
+                if (poisons > ent.Comp1.MaxPoisons && proto.Metabolisms.ContainsKey(PoisonGroup))
                     continue;
 
 
--- a/Content.Server/Body/Systems/MetabolizerSystem.cs
+++ b/Content.Server/Body/Systems/MetabolizerSystem.cs
@@ -28,6 +28,7 @@ namespace Content.Server.Body.Systems
 
         private EntityQuery<OrganComponent> _organQuery;
         private EntityQuery<SolutionContainerManagerComponent> _solutionQuery;
+        private static readonly string PoisonGroup = "Poison";
 
         public override void Initialize()
         {
@@ -146,7 +147,7 @@ namespace Content.Server.Body.Systems
             var list = solution.Contents.ToArray();
             _random.Shuffle(list);
 
-            int reagents = 0;
+            int poisons = 0;
             foreach (var (reagent, quantity) in list)
             {
                 if (!_prototypeManager.TryIndex<ReagentPrototype>(reagent.Prototype, out var proto))
@@ -163,9 +164,9 @@ namespace Content.Server.Body.Systems
                     continue;
                 }
 
-                // we're done here entirely if this is true
-                if (reagents >= ent.Comp1.MaxReagentsProcessable)
-                    return;
+                // Already processed all poisons, skip to the next reagent.
+                if (poisons >= ent.Comp1.MaxPoisons && proto.Metabolisms.ContainsKey(PoisonGroup))
+                    continue;
 
 
                 // loop over all our groups and see which ones apply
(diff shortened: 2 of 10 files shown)


iaada:
I think limiting reagents is important in general, not just for poisons. Consider a syringe with 15 different healing medicines in it, or a pill with 40. You could full heal from something like that in half a second.

EmoGarbage404 (changes requested):
> `Used to nerf 'stacked poisons' where having 5+ different poisons in a syringe, even at low quantity, would be muuuuch better than just one poison acting.`
> This was the original intention behind MaxReagents. So why limit the amount of reagents and not by poison?
This is a very minor change and probably wont even be noticable.

charitable reading of original text: the [original pr that added the functionality](https://github.com/space-wizards/space-station-14/pull/5246) has this:
"Fixes (sorta) an issue that was common in SS13, where stacking poisons/meds is wildly more effective than just using 1 med or 1 poison, because theres no cap on how many reagents can be processed at once. This adds a cap, so there's diminishing returns. Shuffles the array first so no insertion order nonsense"

The solution offered here prevents strengthening poison but instead just powercreeps all other chemicals. The metabolism system is built around metabolizing 1u of chems each tick (with a cap of 3-5u) so breaking off that cap means that you open up significantly stronger chem combination

> This is a very minor change and probably wont even be noticable.

this is most definitely not a minor change and has an incredibly prominent effect on the way that chems are administered and their strengths. This change needs proper justification because as given it doesn't make sense.

slarticodefast:
> I think limiting reagents is important in general, not just for poisons. Consider a syringe with 15 different healing medicines in it, or a pill with 40. You could full heal from something like that in half a second.

Yeah, pretty sure you can make immortality juice now by mixing every medicine there is, allowing you to metabolize them all in parallel. This is definitely not a minor change.

HelpTheSteak:
This has the unintended consequence of making slimes metabolism limit of 6 purely a downside

robinthedragon:
> Consider a syringe with 15 different healing medicines in it, or a pill with 40. You could full heal from something like that in half a second.

no, not really. mixing chems is pretty limited by rate of administration and lack of overlapping chems - most of the workable medicine combos for speedier healing (like diph + dylo, dex + epi, tric + omni + TDD, punct + dex + saline) already work under current metabolic limits 

and if you're diluting your chems with other ones for budget omni, most of the time all you accomplish is reducing the healing per unit and wasting medicine (if I was shot and have 70 piercing, 10u punct will heal me better then 2u arith 2u bic 2u derm 2u dex+ 2u dylo)

sure, maybe like... kelo/omni/py/derm heals a lot faster than py/derm, but it also heals a lot less per unit, and is more complicated to obtain. I can't think of any other insane combos, since brute chems are mutually exclusive, and there's only two heals for each type of toxin

i guess epi/inap/dex+/dex for super crit-b-gone? but crit-b-gone is already stupid fast

maybe lacing all the chems with omni would be a bit too good? but omni should be Generally Useful, and omni + specialized chem likely heals less per unit then the specialized chem on its own , so there's a tradeoff 

I think this change is good for the game - it *does* buff medbay, but the buff requires proper planning to use effectively. all that mixing chems willy-nilly will achieve is wasting chems, and the more complicated (and, thus, fast) your mixture is, the less likely it is you encounter a patient with the right damage for it to work at peak efficiency 

if specific combinations - like kelo/derm/pyra/omni for burns - prove problematic, then additional drug interactions can be added to counteract them

robinthedragon:
> > I think limiting reagents is important in general, not just for poisons. Consider a syringe with 15 different healing medicines in it, or a pill with 40. You could full heal from something like that in half a second.
> 
> Yeah, pretty sure you can make immortality juice now by mixing every medicine there is, allowing you to metabolize them all in parallel. This is definitely not a minor change.

I agree this isn't a minor change, but overdose limits and drug interactions prevent immortality juice, and specialized immortality juice (which is more feasible under the current system) would outclass generalist immortality juice

also if you have 100u of 5 different chems it'll leave your system in about 20 seconds. even if you're immortal for those 20 seconds (you're not: you'd be forced to use bic instead of specialized brute meds cuz of razorium, and it's not that hard to out-DPS bic. bic sucks). and even assuming you can easily and reliably administer 100u of chems simultaneously (consuming 5u of chem/second means you need to have 50% uptime on a hypospray to continuously be under the effect of the chems), it's absurdly expensive for the amount of time you actually spend under it's effects

ScarKy0:
Tbh I had no idea this could cause so many issues to begin with

retequizzle:
won't pretend to be an expert in the medical field of SS14 but don't overdoses kill this idea in the cradle anyway? since instead of layering toxins you're using two toxins as a base and dumping the rest of the space of the syringe/pill popping on purely exceeding overdose threshholds

5u amatox, 5u weed killer, 20u bruizine congratulations you now have a more potent poison than lexorin + amatox and it fits in a hypospray

Pronana:
This is a shadow slime nerf but as others have also pointed out a pretty big buff to chem mixing.

Granted brutes have an anti-mixing reaction but the biggest thing here is that this allows for a true permanent meth mix.

The limiting factor preventing you from chugging 500u of deso is that you can't also injest enough other chems to outdo the damage easily, especially without sacrificing your ability to use other healing chems.

If this gets merged making the ultimate combat stim will be limited purely by how many chems you can make and how quickly you can chug them.

EmoGarbage404:
As it stands i don't think this is a positive change. It opens up a lot of very overpowered chem mix options and seems like it would be very prone to abuse without further tweaking or balancing.

ScarKy0:
Yeah I didn't think it could lead to so many issues, ill just close it

PR #37583, opened 2025-05-18
reduced motion flash effect version 2

slarticodefast:
## About the PR
Based on the work of Quantum-cross in this PR https://github.com/space-wizards/space-station-14/pull/36288
In the internal maintainer discussion we don't want two completely different shaders for this for future maintainability and to make sure players can't get a combat advantage from the setting.

## Why / Balance
Accessibility feature

## Technical details
Instead of adding a completely new shader this PR just adds a timescale to the rotating motion in the shader.
To prevent it from looking the same each time we also add a random phase to the sin/cos functions.

## Media

https://github.com/user-attachments/assets/f129470c-f850-42be-86c2-26b54652e9a2

## Requirements

## Breaking changes

**Changelog**
:cl: Quantum-cross, slarticodefast
- add: Add reduced motion alternative to the flash effect

### DIFF
--- a/Content.Client/Flash/FlashOverlay.cs
+++ b/Content.Client/Flash/FlashOverlay.cs
@@ -1,8 +1,10 @@
+using Content.Shared.CCVar;
 using Content.Shared.Flash;
 using Content.Shared.Flash.Components;
 using Content.Shared.StatusEffect;
 using Robust.Client.Graphics;
 using Robust.Client.Player;
+using Robust.Shared.Configuration;
 using Robust.Shared.Enums;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Timing;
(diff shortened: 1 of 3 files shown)


Quantum-cross:
I don't suffer from this personally, but from what I'm gathering is that the fullscreen "crossfade" from "effect" to "normal gameplay" is the major issue. This version doesn't address that at all.

With the repeating lines of the floor tiles, they will intersect and kind of "moiré" into the repeating lines of the fading in tiles as you move and that's why I addressed the issue like I did with the black spot.

The black spot is something that a motion sensitive person can look at that will have zero motion associated with it.

slarticodefast:
Thanks for the feedback.
Interestingly while I usually don't have motion sickness problems, personally I consider the black circle as much more irritating than the original effect.
Maybe we can try going into a completely different direction and just replace effect with a gaussian blur that slowly fades uniformly across the screen? That could soften up the floor lines.

Quantum-cross:
> Thanks for the feedback. Interestingly while I usually don't have motion sickness problems, personally I consider the black circle as much more irritating than the original effect.

Would the black spot on my version be better if it were white as someone else suggested?

Also, hear me out for a moment here, I'm not sure the fact that you find the black spot irritating is really relevant. As a matter of fact I'd argue that it moves farther away from the "nobody will use the old one anymore" argument...

```
[----------A----------B-------C--------------D-----]
^                                                 ^
No problem                     I can't stop vomiting

A - A non-motion sensitive person with the old flash effect -- this looks neat and makes sense
B - A non-motion sensitive person with my reduced motion effect -- this looks dumb and the black spot is dumb
D - A motion sensitive person with the old flash effect -- I need to physically look away while being flashed
C - A motion sensitive person with my reduced motion effect -- I can actually play the game regularly now, I really don't care that the black spot looks dumb.
```

This is just kind of made off vibes and my discussions with other people.

> Maybe we can try going into a completely different direction and just replace effect with a gaussian blur that slowly fades uniformly across the screen? That could soften up the floor lines.

I'm just not sure that this will fix affected persons concerns.

slarticodefast:
> Would the black spot on my version be better if it were white as someone else suggested?

The flash effect was orginally white instead of black and that was really bad as well. The sudden increase in brightness actually flashed your eyes.

keronshb:
> The black spot is something that a motion sensitive person can look at that will have zero motion associated with it.

I get the opposite, where the black spot + motion background actually activates my motion sickness.

It may be better to have a separate crosshair option that someone can toggle to figure out their point of reference. This way it's agnostic of the reduced motion option and people can toggle it on/off as they want. Think of like a crosshair option built into a monitor.

What's presented in this PR is infinitely better than what's previously proposed and even the current effect.

For Flashes/Flashbangs the main idea is distortion. So both visually and mechanically.
Mechanically it's partially implemented with a rather annoying slowdown. Once crawling is in I'm going to change how it works mechanically (less slowdown, confused input, maybe a knockdown that doesn't lose stamina). But that's out of scope for either PR.

Visually this PR handles it relatively well since you can't exactly see where you are since everything is frozen.

Alternatives would be 
- A blur effect that doesn't freeze the screen but you can technically see where you are. Fades like how this effect does.
- A white overlay with the same slow fade, but that may cause other issues like seizures
- A black overlay of the above.
- A separate crosshair option in accessibility.

slarticodefast:
https://github.com/user-attachments/assets/862dd2b7-6d38-4002-a933-1a0e078f5d90

Ok, another attempt: This also removes the blending towards the end of the effect. That does make it look a little less "smooth" and visally appealing with the hard cutoff, but should prevent the grid patterns from the tiles overlapping in that transition.

Quantum-cross:
I think boring is fine, it's for accessibility. 

Quick flashes to black from my understanding isn't as much of an issue for people with actual epilepsy, mostly just flashes of high luminance and flashes of the color red.

PR #37584, opened 2025-05-18
add reduced motion drug overlay

slarticodefast:
## About the PR
Ok, this one is hopefully less controversial than the reduced motion flash shader.
The seeing rainbow overlay was completely disabled for players that have the reduced motion setting activated in #28496
This PR resolves the TODO by reimplementing it without any motion and disabling the screen warp, leaving only a constant color overlay. No motion at all.

## Why / Balance
accessibility

## Technical details
Add a timescale for the movement speed and set it to 0 when the reduced motion cvar is activated.
To make sure it doesn't look the same every time we add a random phase to the sin/cos functions on startup.
The `effectScale` was split up into `colorScale` and `warpScale`, where the latter is disabled for reduced motion as well.

## Media
Video at maximum effect strength.
Below that the alpha channel is simply lower, making it less visible.

https://github.com/user-attachments/assets/bf69075d-f863-4523-b759-032cd8adcf36

## Requirements

## Breaking changes
The `effectScale` parameter for the `rainbow.swsl` shader was split up into `colorScale` and `warpScale`.

**Changelog**
:cl:
- add: Added a reduced motion version of the seeing rainbows overlay.

### DIFF
--- a/Content.Client/Drugs/RainbowOverlay.cs
+++ b/Content.Client/Drugs/RainbowOverlay.cs
@@ -24,16 +24,27 @@ public sealed class RainbowOverlay : Overlay
 
     public float Intoxication = 0.0f;
     public float TimeTicker = 0.0f;
+    public float Phase = 0.0f;
 
     private const float VisualThreshold = 10.0f;
     private const float PowerDivisor = 250.0f;
+    private float _timeScale = 0.0f;
+    private float _warpScale = 0.0f;
 
     private float EffectScale => Math.Clamp((Intoxication - VisualThreshold) / PowerDivisor, 0.0f, 1.0f);
 
     public RainbowOverlay()
     {
         IoCManager.InjectDependencies(this);
+
         _rainbowShader = _prototypeManager.Index<ShaderPrototype>("Rainbow").InstanceUnique();
+        _config.OnValueChanged(CCVars.ReducedMotion, OnReducedMotionChanged, invokeImmediately: true);
+    }
+
+    private void OnReducedMotionChanged(bool reducedMotion)
+    {
+        _timeScale = reducedMotion ? 0.0f : 1.0f;
+        _warpScale = reducedMotion ? 0.0f : 1.0f;
     }
 
     protected override void FrameUpdate(FrameEventArgs args)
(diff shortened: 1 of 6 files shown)


keronshb (approved):

keronshb:
Mergable once test pass.

slarticodefast:
Best keep it open for a day or two to gather some feedback, just to be sure.
And while the effect doesn't look perfect visually I hope that it is an acceptable compromise and better than having no effect at all and being locked out of this part of gameplay.

walksanatora:
I feel like it either needs reduced transparency (makes it harder to see) or a static offset so the screen is still warped but statically. just something so again reduced motion is not just objectively better then the normal effect.

slarticodefast:
> static offset so the screen is still warped but statically

I tried out static warping, but that still results in a lot of movement when you walk around. This overlay was completely disabled with the setting because it caused health problems for someone in the past, so I would rather err on the side of caution. I think game-balance wise this is fine as this overlay is used purely for roleplay reasons, and not for combat like the flash overlay.

Content.Client/Drugs/RainbowOverlay.cs
```diff
@@ -24,16 +24,27 @@ public sealed class RainbowOverlay : Overlay
...
         IoCManager.InjectDependencies(this);
+
         _rainbowShader = _prototypeManager.Index<ShaderPrototype>("Rainbow").InstanceUnique();
+        _config.OnValueChanged(CCVars.ReducedMotion, OnReducedMotionChanged, invokeImmediately: true);
+    }
+
+    private void OnReducedMotionChanged(bool reducedMotion)
+    {
+        _timeScale = reducedMotion ? 0.0f : 1.0f;
+        _warpScale = reducedMotion ? 0.0f : 1.0f;
```
EmoGarbage404, line 47: This doesn't seem to actually update the shader parameters if it's already active? I think one of the key things we'd want to support is someone seeing the shader _then_ deciding to disable it due to nausea or whatnot.

slarticodefast: What do you mean?
See the video, it works instantly when toggling the setting.
The shader paramters are set on every Draw call.

PR #37586, opened 2025-05-18
remove ignition functional from welder component

kosticia:
## About the PR
Technical pr, removed one event.

## Why / Balance
this isn't void, isn't method, isn't event. This is trolling, literally ZERO reasons for welder component leave fixed amount of heat per tick when we have IgnitionSourceSystem and component togglers.

## Media

https://github.com/user-attachments/assets/30f02358-a880-4842-84e0-ae078ff412e3

## Requirements

## Breaking changes
WelderComponent no longer ignite gas. Use IgnitionSourceComponent instead of it.

**Changelog**
:cl:

### DIFF
--- a/Content.Server/IgnitionSource/IgnitionSourceSystem.cs
+++ b/Content.Server/IgnitionSource/IgnitionSourceSystem.cs
@@ -2,6 +2,7 @@ using Content.Server.Atmos.EntitySystems;
 using Content.Shared.IgnitionSource;
 
 namespace Content.Server.IgnitionSource;
+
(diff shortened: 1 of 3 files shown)


slarticodefast (changes requested):
Good catch! IgnitionSourceComponent should already be doing this,
Small change request, please make the base function for `TurnOn` no longer virtual since that isn't needed anymore.

kosticia:
> Small change request, please make the base function for `TurnOn` no longer virtual since that isn't needed anymore.

Done

slarticodefast (approved):
Thanks for the code cleanup!

PR #37588, opened 2025-05-18
Add keybind for swapping hands in the other direction

slarticodefast:
## About the PR
If you have more than two hands you can now swap them in the other direction via a new keybind.
Defaults to Shift+X (with X being the normal hand swap).
If you have exactly two hands both keybindings work the same.
Especially useful for borgs modules.

Sadly I did not figure out how to enable the mouse wheel as a keybind option yet. But that would probably conflict with UI scrolling.

## Why / Balance
Borg QOL
needed for 4 hand arachnids.

## Technical details
Just copy paste everything for the normal handswap key.
Also fixes the ftl string names for the shift+numpad hotbar slots. These are autoconverted from PascalCase to kebab-case and were missing a - sign, so they did not show up in the menu correctly.

## Media

https://github.com/user-attachments/assets/41965463-4542-45d3-a2fd-7250b2174d3f

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- add: Added a new keybind for swapping hands in the other direction (if you got more than two). Defaults to Shift+X. Useful for cycling through borg modules.


muburu:
only thing missing now is the module remembering your last selected hand when you switch to it

EmoGarbage404:
our ux is so heinous this is great

PR #37603, opened 2025-05-19
Mob Spin Fix Update

Pronana:
## About the PR
Had the float for velocity be changed to an absolute value when checking if it's below the mininmumFrictionSpeed
I was confused when sometimes mobs would still spin when being pulled until i realized that angular velocity can be negative.

## Why / Balance
When checking for minimum friction speed, it should account for negative velocities.

## Technical details
velocity -> Math.Abs(velocity)

## Requirements

## Breaking changes

**Changelog**


slarticodefast (approved):
Looks good, thanks for the bugfix!

Issue #25034, opened 2024-02-08
Banana phone should look and act like an actual banana phone


UbaserB:
## Description

It should keep the ability it had to contact the honkmother while also working as a handheld intercom for some minor trolling (afterall, it is for the clown). The sprite currently looks like a resprited red phone, just yellow which is a bit misleading considering the name. This would probably give it more purpose too.

ThatGuyUSA:
I suppose❓ 

<img width="404" height="263" alt="Image" src="https://github.com/user-attachments/assets/77fd0ac4-4241-4908-90cb-20d8498ff8d0" />

PR #37608, opened 2025-05-19
Banana phone

AsnDen:
## About the PR
1. Resprited of banana phone. I hope the top is more banana now 
2. Added radio features for banana phone instead intercom. It broadcasts to common channel. Can't receive messages. You can't turn it off. It can work without telecomms.

### About toggling it
With all banana features (praying, music) it's possible to turn radio mode off only when banana phone is on the ground. It's impossible to turn it off in hands, so I've decided that it would be better to make it untoggable.

## Why / Balance
#25034 

## Requirements

**Changelog**

:cl:
- add: Banana phone gets new sprite.
- add: Banana phone always broadcasts to common channel.


iaada:
While I do love clown additions, never being able to turn off the intercom is a bit much. I like to keep the phone on me at all times, and with this I couldn't do that without being exceedingly annoying.

IProduceWidgets:
I dont think making it a radio makes much sense, but if you are going to try to do it add the radio ui to it.

TeenSarlacc:
what if the banana phone had its own radio channel, so only people with banana phones could listen in on it?

sudobeans:
The sprite looks good!

It seems the test is supposed to make sure components that get added on spawn are all in the prototype.

The test fails because an `ActiveListenerComponent` and an `ActiveRadioComponent` get added when the radio is spawned. `RadioMicrophone`'s `enabled` property is set to `true`, and [RadioDeviceSystem.cs:66](https://github.com/AsnDen/wiz-wiz-wizard-station/blob/e38ebc887a1413a28a034be4eeabefe343bf51d9/Content.Server/Radio/EntitySystems/RadioDeviceSystem.cs#L66) spawns an ActiveListener component when the microphone is enabled. 

```csharp
            EnsureComp<ActiveListenerComponent>(uid).Range = component.ListenRange;
```

The case is very similar for the `ActiveRadioComponent`. See [RadioDeviceSystem.cs:74](https://github.com/AsnDen/wiz-wiz-wizard-station/blob/e38ebc887a1413a28a034be4eeabefe343bf51d9/Content.Server/Radio/EntitySystems/RadioDeviceSystem.cs#L74).

Possible solutions:
- add the ActiveListener component to the prototype
- make it not be enabled when it's started

AsnDen:
Now banana phone (and other handheld radios) are toggable with right-click context menu. So banana phone no more always active

AsnDen:
> what if the banana phone had its own radio channel, so only people with banana phones could listen in on it?

The issue originally says banana phone is for trolling. I think broadcasting to common will make more fun than to secret channel

SlamBamActionman:
One issue that I have seen with the "Common Radio" radio that is map-exclusive on Plasma is that it can cause the "talking into radio without intending to" issue that is sometimes observed with sec radios. Except for the common radio it affects the entire station. 

The station does not need to hear the clown's every single conversation, I would nix that part of this PR.

AsnDen:
> The station does not need to hear the clown's every single conversation, I would nix that part of this PR.

I can remove it but issue that this PR is based on says that it should have this ability for "some minor trolling"
> ... while also working as a handheld intercom for some minor trolling (afterall, it is for the clown).

SlamBamActionman:
> I can remove it but issue that this PR is based on says that it should have this ability for "some minor trolling"
> 
> > ... while also working as a handheld intercom for some minor trolling (afterall, it is for the clown).

I think the idea of the phone doing something cheeky is good, but I don't think risking an always-on Common radio chatter is the way to go. Just muddies up conversations. Not sure if Ubaser considered it when writing the issue.

It's fine if the PR is just a resprite/verb fix, but if you can figure out some other fun thing to go with it, that's worth exploring imo. Idk maybe slipping someone or calling with it makes you get a pie face or something.

AsnDen:
> ... I don't think risking an always-on Common radio chatter is the way to go. Just muddies up conversations. Not sure if Ubaser considered it when writing the issue.

Ok. May I wait for another review to be sure that I really need to remove radio channel? (I'm not sure how concern label works. Tell me, please, if I need to remove it before another review)

I haven't planned any other mechanics for banana phone here and haven't planned coding either.

So if another maintainer says that radio should be removed for now, I will remove it and make verb for handheld radios in separate PR

Probably it is possible to make it turning off automatically in some time. But I am not going to implenet this for now

ArtisticRoomba:
Closing this PR as derelict as it is merge conflicted and inactive. I agree with Slam's comments. It can be re-opened with the changes made.

PR #37625, opened 2025-05-19
Make electrified security grill on MaxSecurity airlock uncuttable without insuls

Quantum-cross:
## About the PR

This brings the max security electrified grille on an airlock in line with every other power line and electrified grille in the game.

If you are not wearing sufficient insulation, you will not be able to cut the grille.

## Why / Balance

Seems weird that the feature of the highest security door is just a little zappy zap and a small inconvenience.

## Technical details

I was looking all over for where to put this in the construction system for a cancelable action and I ended up putting it in the unholy hell of whatever `HandleInteraction` is.

This is handled by a `DataField` `TryElectrocute` on `ToolConstructionGraphStep`.

There doesn't seem to be any other place that can easily "cancel" the construction action given a condition on the user of construction.

Literally the only place that tries to electrocute the user on a construction step is this airlock step so...

## Media

BEFORE:

https://github.com/user-attachments/assets/68dada04-870f-42b4-87cb-f736378fc9d9

AFTER:

https://github.com/user-attachments/assets/1ef3e146-7395-4af7-b57a-87e5326f51bd

## Requirements

## Breaking changes

none

**Changelog**
:cl:
- fix: You can no longer cut the electrified grille of max security doors without electrical insulation

### DIFF
--- a/Content.Server/Construction/ConstructionSystem.Interactions.cs
+++ b/Content.Server/Construction/ConstructionSystem.Interactions.cs
@@ -1,17 +1,20 @@
 using System.Linq;
 using Content.Server.Administration.Logs;
 using Content.Server.Construction.Components;
+using Content.Server.Electrocution;
 using Content.Server.Temperature.Components;
 using Content.Shared.Construction;
 using Content.Shared.Construction.Components;
 using Content.Shared.Construction.EntitySystems;
 using Content.Shared.Construction.Steps;
 using Content.Shared.DoAfter;
+using Content.Shared.Electrocution;
 using Content.Shared.Interaction;
 using Content.Shared.Interaction.Components;
 using Content.Shared.Prying.Systems;
 using Content.Shared.Radio.EntitySystems;
 using Content.Shared.Stacks;
+using Content.Shared.Stunnable;
 using Content.Shared.Temperature;
 using Content.Shared.Tools.Systems;
 using Robust.Shared.Containers;
@@ -360,6 +363,19 @@ namespace Content.Server.Construction
                         return _toolSystem.HasQuality(interactUsing.Used, toolInsertStep.Tool)
                             ? HandleResult.Validated
                             : HandleResult.False;
+                    }
+
+                    if (toolInsertStep.TryElectrocute && EntityManager.TryGetComponent<ElectrifiedComponent>(uid, out var electrified))
+                    {
+                        var currentValue = electrified.Enabled;
+                        electrified.Enabled = true;
+
+                        EntityManager.EntitySysManager.GetEntitySystem<ElectrocutionSystem>().TryDoElectrifiedAct(uid, user.Value, electrified: electrified);
+
+                        electrified.Enabled = currentValue;
+
+                        if(EntityManager.HasComponent<StunnedComponent>(user.Value))
+                            return HandleResult.False;
                     }
 
                     // If we're handling an event after its DoAfter finished...
--- a/Content.Shared/Construction/Steps/ToolConstructionGraphStep.cs
+++ b/Content.Shared/Construction/Steps/ToolConstructionGraphStep.cs
@@ -15,6 +15,8 @@ namespace Content.Shared.Construction.Steps
 
         [DataField("examine")] public string ExamineOverride { get; private set; } = string.Empty;
 
+        [DataField] public bool TryElectrocute { get; private set; }
+
         public override void DoExamine(ExaminedEvent examinedEvent)
         {
             if (!string.IsNullOrEmpty(ExamineOverride))
--- a/Resources/Prototypes/Recipes/Construction/Graphs/structures/airlock.yml
+++ b/Resources/Prototypes/Recipes/Construction/Graphs/structures/airlock.yml
@@ -365,7 +365,6 @@
     edges:
     - to: highSecurity
       completed:
-      - !type:AttemptElectrocute
       - !type:GivePrototype
         prototype: PartRodMetal1
         amount: 2
@@ -373,6 +372,7 @@
       - !type:WirePanel {}
       steps:
       - tool: Cutting
+        tryElectrocute: true
         doAfter: 0.5
 
     - to: superMaxSecurityUnfinished


slarticodefast (changes requested):
This should absolutely not be hardcoded into construction graphs like this.

Quantum-cross:
Meanwhile, `HandleStep` is designed to handle cancellation by returning `HandleResult.False`

```
        /// <summary>
        ///     Takes in an entity, a <see cref="ConstructionGraphStep"/> and an object event, and handles any possible
        ///     construction interaction. Unlike <see cref="HandleInteraction"/>, if this succeeds it will perform the
        ///     step's completion actions. Also sets the out parameter to the user's EntityUid.
        /// </summary>
        /// <remarks>When <see cref="validation"/> is true, this method will simply return whether the interaction
        ///          would be handled by the entity or not. It essentially becomes a pure method that modifies nothing.</remarks>
        /// <returns>The result of this interaction with the entity.</returns>
```

Quantum-cross:
The real question I have is that why doesn't `ConstructionGraphStep` have an abstract method `HandleStep`?

Quantum-cross:
Holy shit this code is making me crash out hard

- It can't go in actions because once you're in actions it's too late you need to explicitly go back a node and you can't cancel
- it can't go in conditions because you don't get passed the tool user because it's only conditions specifically on the state of the entity. 
- the only logical place to put it is in the step action

Please tell me what I'm missing here

Quantum-cross:
Sorry if I'm repeating myself a billion times I'm also just understanding the system myself and typing it out helps solidify it.

`IGraphCondition` checks: "Is the entity itself in the right state to traverse the node and stop if not"
`ConstructionGraphStep` checks: "Is the state of the actor or the interaction of the actor correct to traverse the node and stop if not"
`IGraphAction` handles: "We've ALREADY traversed the node, now perform all of these actions and stop if an action deletes the entity"

`AttemptElectrocute` is currently an `IGraphAction` and I'm fairly confident it should be in `ConstructionGraphStep`.

It just feels really dirty because `ConstructionSystem.Interactions.cs` is kind of stinky and might need a refactor...

slarticodefast:
Should probably use some kind of interaction attempt event and cancel that before this part of interaction is even reached.

Quantum-cross:
> Should probably use some kind of interaction attempt event and cancel that before this part of construction is even reached. I can't say what the perfect solution is either without a more detailed look though.

So to do this, I would need to:

- Subscribe to `InteractUsingEvent` on airlocks
- Query the construction graph of the airlock to see what node it's on
- Query to see if we're on a construction edge
- if we're on an edge, check the step index that we're on
- see if the next step requires wire cutters
- check that the conditions of the edge are satisfied
- ... (basically beginning to duplicate all construction code at this point)
- then after all that we finally can tell we're going to complete an edge with an `AttemptElectrocute` completed action, then we can finally electrocute the user and then somehow cancel the construction action (I haven't even thought about that part yet)

*Or we can just have the tool step optionally try to electrocute the user and cancel if successful*

This is the pattern that `WiresPanel` `PowerWireAction` does. The cut action itself will attempt to electrocute, and if successful, well return false for unsuccessful.

-----

I want to make it clear that while I am being a little sassy here -- it comes from a place of me spending hours last night scrambling, trying not to commit what I committed.

> I can't say what the perfect solution is either without a more detailed look though.

The construction system has a very high cognitive complexity, and I absolutely don't blame you for saying "there has to be a better way".

You'll either have to dive in yourself or trust me.

-----

I don't like the solution, but any other solution I could think of involved hacking interfaces to the `ConstructionSystem` (requiring changes to every single `IGraphAction`) and I even still couldn't get it to work quite right. Any of these hacks would be in spite of a real refactor.

This little piece of code is relatively unobtrusive and works flawlessly, and given the current architecture of the `ConstructionSystem` it's where it "should" be.

slarticodefast:
You could still make it more generic by raising and event similar to how `OnConstructionTemperatureEvent` does it. That way multiple systems could hook into it and return a `HandleResult` without explicitly having to hardcode the component here.
Electrifying and de-electrifying the grill would have to be handled through an `IGraphAction`.

Quantum-cross:
> You could still make it more generic by raising and event similar to how `OnConstructionTemperatureEvent` does it. That way multiple systems could hook into it and return a `HandleResult` without explicitly having to hardcode the component here. Electrifying and de-electrifying the grill would have to be handled through an `IGraphAction`.

I feel like this supports my method even more.

`ToolConstructionGraphStep`
- Given a configuration value (`TryElectrocute`)
- We want to check some external condition
- Test the condition using `TryDoElectrifiedAct`
- Return `HandleResult.False` if it fails

`TemperatureConstructionGraphStep`
- Given a configuration value (`MinTemperature` and `MaxTemperature`)
- We want to check some external condition
- Test the condition by comparing the configured temperatures to the `InternalTemperatureComponent` or `TemperatureComponent`
- Return `HandleResult.False` if it fails

Quantum-cross:
> You could still make it more generic by raising and event similar to how `OnConstructionTemperatureEvent` does it. That way multiple systems could hook into it and return a `HandleResult` without explicitly having to hardcode the component here. Electrifying and de-electrifying the grill would have to be handled through an `IGraphAction`.

`OnConstructionTemperatureEvent` just checks for an `ActivelyMicrowavedComponent`. Are you suggest I add an `ActivelyElectrifiedGrilleComponent` just to raise an event on it?

Electrocution checks should only be done by `ElectrocutionSystem` on `ElectrifiedComponent`, much like how `InternalTemperatureComponent` and `TemperatureComponent` are "hard-coded" into the handling of `TemperatureConstructionGraphStep`.

I can't think of any reason other systems need to hook into a `TryElectrocute == true` flag on `ToolConstructionGraphStep`.

slarticodefast:
Ok, let me elaborate a little more:
Hardcoding very specialized components via `TryComps` into generic systems like construction graphs is bad practice because it makes the generic system depend on the specialized code. Dependencies like this should always flow only the other way around. If someone now has a fork that has no electricity and they would have to rewrite the construction code and undo your PR in order to make it still work. In addition it is harder to extend than via events and sets a bad precendence - we regularly get lots of PRs like this and it would result in a giant list of `TryComps` in this system in a few years.

slarticodefast:
> OnConstructionTemperatureEvent just checks for an ActivelyMicrowavedComponent. Are you suggest I add an ActivelyElectrifiedGrilleComponent just to raise an event on it?

You could add an `OnToolUse` bool to `ElectrifiedComponent` maybe. It could be a little tricky because of other methods that enable or disable the component.

Quantum-cross:
Oh yeah, with the caveat that we need to handle the `Enabled` status of `ElectrifiedComponent` externally (I honestly don't know when this gets set to false so it should just... work...)

Quantum-cross:
this is much less terrible

AJCM-git:
@Quantum-cross coming back to this?

Issue #37624, opened 2025-05-19
smes/substation power flickering can trigger epilepsy


Darkmajia:
## Description
oscillating power sources and the power flickering that results from them after smes/substations have been drained can be an epilepsy hazard. this is easiest to reproduce with the tesla, but tegs with choking burn chambers can constantly dip above and below power thresholds and spam light flickers too.

one of our players on impstation has reported having a seizure over this, and it's not specific to our server: i've reproduced this on the current upstream version

**Reproduction**
- use admin tricks to empty smes and a departmental substation
- wait for apcs to drain
- start up the tesla engine
- watch the lights go crazy

**Screenshots**
this video is an actual in-round scenario from impstation with a burn chamber-choked teg, on saltern:
https://github.com/user-attachments/assets/ce1f7f9c-ccef-4c8c-85f6-b17279411c71

this video is on the current upstream version, on box:
https://github.com/user-attachments/assets/1bbbbc51-1a1f-4f37-84b5-bbd14a27b847

**Additional context**
while this _can_ be alleviated with redundant power sources (ame, solars, etc), the potential for this to happen in the first place is a legitimate health concern

ArtisticRoomba:
Working on this, numbers are hard to get right but it's looking promising

PR #37626, opened 2025-05-19
Significantly improve Tesla generator power stability

ArtisticRoomba:
## About the PR
Modifies Tesla coils to make their power generation output over a longer period of time, in order to keep them constantly outputting while in a coil array in an active Tesla setup.

## Why / Balance
Following #37624 I went ahead and dove into it and found out that the Tesla coil power problem had gotten worse and now was supreme ass. No idea why nobody was complaining about this before.

Basically, tesla coils, whenever they are struck, instantly get 2 MJ dumped into their battery bank. This then gets dumped to the grid at a limit of 1 MW. Even in a 4 or 6-coil setup, this rapid release of power causes breakages in the supply of power, as all of the power is already dumped to the grid and wasted (SMESes can't charge that fast) before the next lightning strike from the Tesla comes in.

This is bad, as SMESes do *not* provide power to compensate for deficits instantaneously, intentionally.

This PR smoothes out and eliminates the gaps in power, which means that the station is always powered and thus no flickering is caused.

This also works towards the goal of making engines more attractive.

## Technical details
YAML changes.

## Media
I did a lot of testing, but here's proof.
Content Client_mDr3i3UF3h

## Requirements

**Changelog**
:cl:
- tweak: Tesla coils can now hold a max capacity of 5 MJ, and receive 5 MJ of energy when struck by lightning.
- tweak: Tesla coils can now only output a maximum of 350 kW to the grid. This was done to make power flow smoother (it was triggering epilepsy in extreme circumstances).

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Power/Generation/Tesla/coil.yml
+++ b/Resources/Prototypes/Entities/Structures/Power/Generation/Tesla/coil.yml
@@ -56,11 +56,11 @@
     sprite: Structures/Power/Generation/Tesla/coil.rsi
     state: coil
   - type: Battery
-    maxCharge: 2000000
+    maxCharge: 5000000
     startingCharge: 0
   - type: BatteryDischarger
   - type: TeslaCoil
-    chargeFromLightning: 2000000
+    chargeFromLightning: 5000000
   - type: LightningTarget
     priority: 4
     hitProbability: 0.5
--- a/Resources/Prototypes/Entities/Structures/Power/Generation/Tesla/coil.yml
+++ b/Resources/Prototypes/Entities/Structures/Power/Generation/Tesla/coil.yml
@@ -56,7 +56,7 @@
     sprite: Structures/Power/Generation/Tesla/coil.rsi
     state: coil
   - type: Battery
-    maxCharge: 1000000
+    maxCharge: 2000000
     startingCharge: 0
   - type: BatteryDischarger
   - type: TeslaCoil
@@ -67,8 +67,8 @@
     lightningResistance: 10
     lightningExplode: false
   - type: PowerNetworkBattery
-    maxSupply: 1000000
-    supplyRampTolerance: 1000000
+    maxSupply: 350000
+    supplyRampTolerance: 350000
   - type: Anchorable
   - type: Rotatable
   - type: Pullable


ArtisticRoomba:
This is an engineering problem, it would be hard to fix this in terms of an accessibility setting. We need to design our systems to have cleaner power output in general, reduce systems that have large jumps and periods of deficit, while still having distinct periods of "oh no it's not generating any power" which is the point of SMESes having power ramping. If engineering blows up, the station should brown out, and you should know.

The better solution is to implement Pow3r low-power states in lightbulbs, ex. have incandescent lightbulbs dim, fluorescent lights flicker, etc. That's for the future though.

Though I could be looking at it wrong and PJB will come in and kill me with hammers.

VerinSenpai:
I think you're in the right mindset. It honestly doesn't make sense that the Tesla would not be capable of preventing random brownouts. As for the teg that is generally down to setup as mentioned in the issue. I definitely think there should be something for accessibility to deal with ghost boos and inefficient power setups. Engines being capable of meeting power demand when setup properly is an important start though.

mqole:
how possible would it be to give TEG a battery bank like mentioned for the tesla? just something to ensure power is dumped to the grid more gradually, to compensate for any potential flickers. on the fork this issue occurred on initially, we've got a pretty consistent TEG setup that is capable of meeting power demands, the problem is just the stuttering in start/stop due to pump kPa lockout (whole seperate can of worms). on TEG-only stations like Plasma, this can still be a big issue. although maybe its just not an easy fix and needs to wait for a big overhaul of power states.

ArtisticRoomba:
It's a combination of the both. It's true that power can spike heavily in extremely basic setups. This is due to the pumps clogging and halting flow, which halts power generation, which causes flickering. This can be alleviated by using volumetric pumps as well as having proper pipenet volume buffers (to prevent clogs due to no room for gas to go). At the end of the day this is a player issue/mapping issue with bad design causing inconsistent power due to clogged pumps.

I will possibly look at a powernet averager similar to radiation collectors, as those temporary store/average power out due to Pow3r and the radiation system updating at different time intervals.

I'm also probably going to look at improving the base TEG setups around the current maps as I've noticed that they're noobtraps right now.

ArtisticRoomba:
DNM until I test more edge cases I thought of

VerinSenpai:
Sure thing hoss. 👍

ArtisticRoomba:
This is a pull request not a chatroom

ArtisticRoomba:
Good to merge

Issue #34928, opened 2025-02-06
Hristov inaccuracy when using separated hud layout


shamblestf:
Tried 20 times to upload the video directly to here but i couldnt so youtube video: https://www.youtube.com/watch?v=RSJgchS-sFI

Hitting things directly north, east, west, south is usually fine, but any other direction and it becomes unusable

Errant-4:
Thanks for reporting this, I have relayed it to Slam.
In the meantime, I have found that this only happens if you use the Separated hud mode. Presumably, Slam is a newchat-using heretic and so never encountered the bug himself

shamblestf:
>  I have found that this only happens if you use the Separated hud mode (where the view is offset from the center). Presumably, Slam is a newchat-using heretic and so never encountered the bug himself

we seperated hud gamers are oppressed once again...

Errant-4:
It's not fixed yet

iaada:
This is still the case and also happens when shooting directly upward, not just diagonal.

ThatGuyUSA:
Resolved by #35087!!

PR #37636, opened 2025-05-20
fix zoomed aiming on separated UI, cleanup unused dependencies

Alkheemist:
## About the PR

Upstreaming https://github.com/new-frontiers-14/frontier-station-14/pull/3416
Fixes https://github.com/space-wizards/space-station-14/issues/34928
Instead of finding the centre of the window, tries to get the centre of the viewport for calculating the eye offset.
This code will probably still break if there ever is a separated UI layout with the chat on the left side.

## Why / Balance

Bugfix

## Technical details

Instead of using the whole window pixel size, we get the main viewport from the eyemanager and find its size instead using __eyeManager.MainViewport as Control;_

## Requirements

## Breaking changes

None identified.

**Changelog**

:cl:
- fix: When using the separated UI layout, snipers shoot where you point when firing north/south.

### DIFF
--- a/Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
+++ b/Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
@@ -5,6 +5,7 @@ using Robust.Client.Graphics;
 using Robust.Client.Input;
 using Robust.Shared.Map;
 using Robust.Client.Player;
+using Robust.Client.UserInterface;
 
 namespace Content.Client.Movement.Systems;
 
@@ -13,12 +14,11 @@ public sealed partial class EyeCursorOffsetSystem : EntitySystem
     [Dependency] private readonly IEyeManager _eyeManager = default!;
     [Dependency] private readonly IInputManager _inputManager = default!;
     [Dependency] private readonly IPlayerManager _player = default!;
-    [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly IClyde _clyde = default!;
 
     // This value is here to make sure the user doesn't have to move their mouse
     // all the way out to the edge of the screen to get the full offset.
-    static private float _edgeOffset = 0.9f;
+    private readonly float _edgeOffset = 0.9f;
 
     public override void Initialize()
     {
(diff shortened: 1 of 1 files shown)


Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
```diff
@@ -13,12 +14,11 @@ public sealed partial class EyeCursorOffsetSystem : EntitySystem
     [Dependency] private readonly IEyeManager _eyeManager = default!;
     [Dependency] private readonly IInputManager _inputManager = default!;
     [Dependency] private readonly IPlayerManager _player = default!;
-    [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly IClyde _clyde = default!;
 
     // This value is here to make sure the user doesn't have to move their mouse
     // all the way out to the edge of the screen to get the full offset.
-    static private float _edgeOffset = 0.9f;
+    private readonly float _edgeOffset = 0.9f;
```
whatston3, line 21: If readonly, why not const?

SlamBamActionman:
> @SlamBamActionman I think this supercedes your PR, right? Because this solution is much nicer than the extra update loop. #35087

After testing, it does not solve the issue the extra update loop in #35087 aims to solve, which was that the mouse position during a camera offset is invisibly incorrect. 

It does solve the issue caused by the separated UI mode, which lead to an incorrect camera offset when moving the mouse diagonally in that mode. This is also being done in #35087 in [EyeCursorOffsetSystem.cs](https://github.com/space-wizards/space-station-14/pull/35087/files#diff-db4e29314993524ea46f5958c1172ed147fcc2487635fcd21c5833617342b81aL44), though in a different way.

Content.Client/Movement/Systems/EyeCursorOffsetSystem.cs
Alkheemist: I only changed it to readonly to appease the formatter. It would also work as a const.

Alkheemist:
huh, I completely missed https://github.com/space-wizards/space-station-14/pull/35087 when I set out to fix this as it wasn't linked to https://github.com/space-wizards/space-station-14/issues/34928. If I had known about this it would've saved me a lot of time on the weekend trying to get the main viewport through an increasingly strange route before discovering (read: whatstone pointing out) that casting to a control worked.

It doesn't address the extra issue with the mouse position flickering because we didn't set out to fix that, our main concern was making it work at all with separated UI. If there's an existing PR to address both I can close this one and wait for that to go in.

VigersRay:
Why is still not merged? Wizdens want for every players disabled separated mode?

metalgearsloth:
> Why is still not merged? Wizdens want for every players disabled separated mode?

Kid named read the discussion.

VigersRay:
> > Why is still not merged? Wizdens want for every players disabled separated mode?
> 
> Kid named read the discussion.

im so lazy for use translator and read disscusion

Princess-Cheeseballs:
Superceded by #35087

Issue #37635, opened 2025-05-20
areaInsert causes frame drops as available storage decreases


K-Dynamic:
## Description

areaInsert within Storage component causes frame drops as available storage decreases, even if there's no space available. This makes items like 'spell of all-consuming cleanliness' unusable for admins.

This effect seems to scale with areaInsertRadius.

**Reproduction**

Trash bag example.
1. Enable frame rate counter.
2. Spawn more trash than could fit inside trash bag storage.
3. Equip trash bag and pick up trash.
5. Observe lag and drop in frame rate as space decreases.

metalgearsloth:
Okay I've looked at it before and it's just storage being comically slow because:
1. For every caninsert it iterates every possible slot / config the item it could fit into.
2. It re-gets every single stored item's itemcomp + storage slot and rebuilds the entire occupied grid for every single slot above.

So yeah doing this however many thousands of times is going to suck more than a few frames.

PR #37638, opened 2025-05-20
Optimise storage a quadrillion times

metalgearsloth:
Fixes https://github.com/space-wizards/space-station-14/issues/37635

Draft so I can cleanup the API a bit and get some more savings but caching what tiles are occupied already gives us stupid savings.

CBF taking a video but framegraph
image

This is as about as fast as I think I can get it short of having a linkedlist of free tiles or something but I don't think it's worth it. At the very least storage should never lag again (short of the clientside frameupdate callers may need to use the new codepaths).

:cl:
- fix: Picking up items with area pickups (e.g. trash bags) no longer lags the game.

### DIFF
--- a/Content.Client/Storage/Systems/StorageSystem.cs
+++ b/Content.Client/Storage/Systems/StorageSystem.cs
@@ -63,6 +63,8 @@ public sealed class StorageSystem : SharedStorageSystem
             component.SavedLocations[loc.Key] = new(loc.Value);
         }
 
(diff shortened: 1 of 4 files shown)


metalgearsloth:
Okay it was still noticeable on the framegraph but I added a fastpath for 99% of usecases and now it doesn't even register.

EmoGarbage404:
thats crazy that the profiler reported a performance increase characteristic of a quadrillion thats so epic

PR #37640, opened 2025-05-20
Predict Flashes

slarticodefast:
## About the PR
Usual prediction PR, needed for trigger refactor.
~~Currently a draft because I'm waiting on https://github.com/space-wizards/RobustToolbox/pull/5849 to have the random chance in FlashArea not mispredict.~~
Also fixes a few interaction bugs with the flash lantern and interrogator lamp.

## Why / Balance
prediction good

## Technical details
Moved everything to shared.
Replaced the deprecated timer spawn with an update loop. This is only used for resetting the visuals, ideally we replace this with something like `Flick()` (see https://github.com/space-wizards/RobustToolbox/pull/5810 ) in the future.
Use a `UseDelay` for the cooldown instead of hardcoding it to 4 seconds.
Make the parameters for `Flash` and `FlashArea` sane by using a `TimeSpan` for the duration rather than milliseconds.
`AfterFlashedEvent` is properly raised on the target, user and flash item as the code comment says it would be. Added an check in the subscription to account for that.
Split the `FlashVisuals` enum into `FlashVisuals` and `FlashVisualLayers` because those are completely different things.

Previously for the interrogator lamp and the flash lantern there were multiple event subscriptions blocking each other depending on order they were handled in. If the flash got triggered first, then the light did not get toggled on and vice versa. The previous manual subscription ordering was useless because one used `UseInHandEvent` and the other `ActivateInWorldEvent`. Also it was flashing a second time when the light was turned off. To fix this we instead activate the flash in a `LightToggleEvent` subscription and disable the use in hand activation using a DataField. Light code is quite messy at the moment and we got multiple implementations of toggleable lights that should all be unified and should use ItemToggle instead, but that is out of scope for this PR.

Currently is affected by this bug https://github.com/space-wizards/RobustToolbox/issues/5802 which means the start of the flash sound (in particular the click noise) is cut off sometimes. But that is not this PRs fault.

For the random probability I used the same trick of setting the rng seed that `SharedDiceSystem` uses so that we have predicted randomness. This is the same implementation that the `RandomPredicted` PR uses, but it should be replaced with the API method once that is merged.

## Media

https://github.com/user-attachments/assets/8740d224-e703-4d4a-87a8-c4167ec52694

## Requirements

## Breaking changes
`DamagedByFlashingComponent`, `FlashImmunityComponent ` were moved to `Content.Shared.Flash.Components`.
`FlashComponent` now uses TimeSpans for the durations and some datafields were changed.
The cooldown for `FlashComponent` is no longer hardcoded to 4 seconds, but uses a use delay. Make sure to add a `UseDelayComponent` to prevent spamming it.
`Flash` and `FlashArea` had their duration parameters changed to use `TimeSpans`.
`AfterFlashedEvent` is now always raised on the user, the target and the flash, not only when used in a melee attack. A `Melee` bool was added to account for that.
`FlashVisuals` was split up into `FlashVisuals` and `FlashVisualLayers`.

**Changelog**
:cl:
- fix: The interrogator lamp and flash lantern now properly flash when toggled on.


Content.Shared/Flash/Components/DamagedByFlashingComponent.cs
```diff
@@ -0,0 +1,19 @@
+using Content.Shared.Damage;
+using Content.Shared.Flash;
```
ScarKy0:
```suggestion

```
Unused.

Content.Shared/Flash/Components/FlashComponent.cs
```diff
@@ -1,55 +1,79 @@
...
+    /// These two times should be the same.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public TimeSpan FlashingTime = TimeSpan.FromSeconds(4);
 
-        [DataField("range")]
-        [ViewVariables(VVAccess.ReadWrite)]
-        public float Range { get; set; } = 7f;
+    /// <summary>
+    /// How long the target will loose vision when melee attacked with the flash.
```
ScarKy0:
```suggestion
    /// For how long the target will lose vision when melee attacked with the flash.
```
Tiny nitpick.

Content.Shared/Flash/Components/FlashComponent.cs
```diff
@@ -1,55 +1,79 @@
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public TimeSpan MeleeDuration = TimeSpan.FromSeconds(5);
 
-        [ViewVariables(VVAccess.ReadWrite)]
-        [DataField("aoeFlashDuration")]
-        public int AoeFlashDuration { get; set; } = 2000;
+    /// <summary>
+    /// How long the target will loose vision when used in hand.
+    /// </summary>
```
ScarKy0:
```suggestion
    /// <summary>
    /// For how long the target will lose vision when used in hand.
    /// </summary>
```
Same here.

Content.Shared/Flash/FlashEvents.cs
```diff
@@ -0,0 +1,22 @@
...
+/// </summary>
+[ByRefEvent]
+public record struct FlashAttemptEvent(EntityUid Target, EntityUid? User, EntityUid? Used, bool Cancelled = false) : IInventoryRelayEvent
+{
+    SlotFlags IInventoryRelayEvent.TargetSlots => SlotFlags.HEAD | SlotFlags.EYES | SlotFlags.MASK;
+}
+
+
+/// <summary>
+/// Called when a player is successfully flashed.
```
ScarKy0:
```suggestion
{
    SlotFlags IInventoryRelayEvent.TargetSlots => SlotFlags.HEAD | SlotFlags.EYES | SlotFlags.MASK;
}

/// <summary>
/// Called when a player is successfully flashed.
```
Extra whitespace.

Content.Shared/Flash/SharedFlashSystem.cs
```diff
@@ -1,15 +1,267 @@
...
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
+    [Dependency] private readonly SharedAudioSystem _audio = default!;
+    [Dependency] private readonly SharedChargesSystem _sharedCharges = default!;
+    [Dependency] private readonly EntityLookupSystem _entityLookup = default!;
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly ExamineSystemShared _examine = default!;
+    [Dependency] private readonly SharedPopupSystem _popup = default!;
+    [Dependency] private readonly SharedStunSystem _stun = default!;
+    [Dependency] private readonly TagSystem _tag = default!;
+    [Dependency] private readonly IRobustRandom _random = default!;
```
ScarKy0:
```suggestion

```
Unused.

Content.Shared/Flash/SharedFlashSystem.cs
```diff
@@ -1,15 +1,267 @@
...
+        if (_useDelay.IsDelayed(ent.Owner))
+            return false;
+
+        if (TryComp<LimitedChargesComponent>(ent.Owner, out var charges)
+            && _sharedCharges.IsEmpty((ent.Owner, charges)))
+            return false;
+
+        _sharedCharges.TryUseCharge((ent.Owner, charges));
+        _audio.PlayPredicted(ent.Comp.Sound, ent.Owner, user);
+        Log.Debug($"{ent.Owner} {user}");
```
ScarKy0:
```suggestion

```
:godo:

ScarKy0 (changes requested):
Tiny nitpicks ill fix myself, will still need to test in game

ScarKy0 (approved):
As someone who never used flashes they still somehow feel way more responsive.
Conversion etc works as expected so I think this is all fine then.

PR #37658, opened 2025-05-21
Significantly improve TEG power generation stability

### LINKED ISSUE #37624: smes/substation power flickering can trigger epilepsy

ArtisticRoomba:
## About the PR
Smooths out the TEG's power generation so it no longer bounces between 0 and max power when the TEG is experiencing pressure fluttering.

## Why / Balance
Fixes #37624

Power bouncing around between 100-200 kW and max power can cause extreme power flickering, as the SMESes will bounce between ramping up and supplementing no power (passthrough). A fluttering TEG can be easily achieved by having a volumetric pump push gas through a circulator with a flow restrictor (ex. radiator with no extraction pump) on the other side.

This was causing seizures, and it was generally horrible to experience in round, and a big noobtrap.

## Technical details
Applies an exponential moving average when we supply power to the grid.
`EMA(t) = α * CurrentValue + (1 - α) * EMA(t-1)`

Note that this is a *gigantically* naive solution and feels really bad IMO as this magics a lot of power. However the TEG still responds semi-rapidly to shutoffs and startups, which is what we want, if there's a big power interruption the station should flicker and know engineering just exploded or something. The TEG also previously magics power into existence for the purposes of power ramping but that's not really an excuse, this feels really big.

If nothing else, this is for forks to cherry-pick.

## Media
Extreme power fluttering before implementation (`teg_power` is what you should be looking at)
Content Client_gFntKvWw7m

Power fluttering dampened after the implementation
Content Client_ldbDQCbKCm

Power responding properly to an increase in heat (due to the burn chamber igniting after extended draw)
Content Client_bpre1VcWJd

## Requirements

**Changelog**
:cl:
- fix: The TEG now outputs smoother power when pressure across the circulators flutter. (This was causing epilepsy in extreme circumstances).


mqole:
hero, tysm for working on this so quickly after the issue was created <3

ArtisticRoomba:
can't wait to get skinned alive by the review

NoElkaTheGod:
Atmos mains looking for a way to abuse this for free power 0.0001 seconds after this PR gets merged:

ArtisticRoomba:
I can easily see this being abused however I'm hoping that they don't find out how to do it until pumps get lukewarm reworked to be something other than two different flavors of pressure pumps.

NoElkaTheGod:
Be like Nolla Games. Players found a bug? Make a feature and nerf it if neccessary :godo:

slarticodefast (approved):
Ideally we would have some sort of heat reservoir on both the hot and cold end that get in contact with the gas flowing through the TEG and smooth out the power curve while still conserving energy rather than magically adjusting the power curve.
But given how problematic the power flickering is for some players I will merge this as a bandaid. Thanks for looking into this!

PR #37675, opened 2025-05-21
Ore + entitytable fixes

metalgearsloth:
Iterate every dungeon not just last.


metalgearsloth:
Supercedes https://github.com/space-wizards/space-station-14/pull/37641 because too much breakage.

Issue #9834, opened 2022-07-17
Need 2x2 thrusters


metalgearsloth:
Should provide more than 4x the thrust, like 16x to encourage its use on bigger ships like atlas.

Peptide90:
Even though we need stuff to use up power, having thrusters use piped gas would be a neat option, or burn plasma.

I have a doc somewhere. Something something make the inside face atmos blocking so you can access it from inside the ship, and also leads to the ship being spaced if destroyed.

Just-a-Unity-Dev:
willing 2 code just need sprites

metalgearsloth:
> willing 2 code just need sprites

/tg/ has one though you may need to slap on the burn effect to it.

PR #37681, opened 2025-05-21
Large thruster

AsnDen:
## About the PR
Adds 2x2 thruster. It's better than 4 normal thrusters but consumes more energy.

No craft (I guess to make it craftable we should add 2x2 frame first). Impossible to anchor.

When destroyed leaves 1x1 frame

### Pulling issue (see media)
While pulled it faces wrong direction. I don't know how to fix this. It's probably caused by sprite offset and fixture offset. Without offset object will be in the middle of the tile.

### 3x3 thruster
While doing this I've seen 3x3 thruster. If u ask I can also add that one.

## Why / Balance
It was asked here #9834 

## Media
Pulling issue

https://github.com/user-attachments/assets/05824c65-dd7f-47e0-98c6-425a40bfd543

Flight

https://github.com/user-attachments/assets/d162eb37-6ad8-4ed6-95c5-844c28994ffc

## Requirements

**Changelog**

:cl:
- add: Large thruster (2x2) was added.


Nukesthestation:
If it can't be crafted, how is it meant to be acquired in-game?

AsnDen:
> If it can't be crafted, how is it meant to be acquired in-game?

I guess pre-mapped. I don't know why did metalgearsloth need it.

If it will be asked, I can add craft later

AsnDen:
Also this is unanchorable. So even if you craft it you won't be able to move it

slarticodefast (approved):
They work great in combination with shuttle collisions 😄

PR #26749, opened 2024-04-05
Better Airlock & Meat spike forensics

SlamBamActionman:
## About the PR

Airlocks now include fingerprint/fiber forensics when bumped into, as opposed to only when manually clicked. 

Meat spikes now include DNA forensics when a creature is spiked, and while spiked it displays the creature's name.
Attaching someone to a meat spike includes fingerprints/fiber forensics.
Meat carved off a spike no longer has the person's name, but instead includes the DNA.
Carving meat gives DNA forensics on the knife used.

## Why / Balance

Airlocks not including forensics when bumped created an annoying dissonance where seemingly equal actions (bumping & manually clicking) resulted in different outcomes not apparent to the user. You can still avoid leaving forensics on a door by throwing your PDA, but that's a much more deliberate action.

Meat spikes generally needed some touch-up in the forensics department; identifying meat by a name tag was a bit weird and not diagetic. 

Note that cooking food does not preserve DNA data; best start frying up some burgers, Syndie-chefs!

## Technical details

The PR makes `BeforeDoorOpenedEvent` into a virtual method to allow the Server system `AirlockSystem` to use it. 
Also includes `unknown DNA` for meat carved from non-DNA holders (e.g. monkeys), to future-proof against #26699.

## Media

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:
- tweak: Meat spikes and carved meat can now be scanned for forensics.
- remove: Carved meat no longer includes the victim's name.
- fix: Airlocks now include fingerprint/fiber forensics when bumped!

PR #37686, opened 2025-05-21
Improve meat spike forensics & behavior

SlamBamActionman:
## About the PR

This PR improves the meat spike's forensics, ensuring better naming, DNA and forensics evidence.

It also no makes meat spikes no longer delete worn items; they instead drop to the ground when someone is meatspiked.

- Meat spikes now include DNA forensics when a creature is spiked, and while spiked it displays the creature's name.
- Attaching someone to a meat spike or carving meat off a meat spike now includes hand forensics.
- Meat carved off a spike no longer has the person's name, but instead includes the DNA.
- Carving meat gives DNA forensics on the knife used.

This PR supercedes #26749. 

## Why / Balance

Because meat spikes were using outdated code and just lacked general forensics evidence to it. 

## Technical details

Raises some kitchen spike events when spiking and carving to apply forensics, and the naming of the kitchen spike is done with the proper Name Modifier system instead of hard-setting the entity name.

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Meat carved from meat spikes no longer include the victim's name.
- fix: Meat spikes now include the victim's DNA and the spiker's hand forensics when spiking and carving.
- fix: Worn items no longer get deleted when meat spiked.

### DIFF
--- a/Content.Server/Kitchen/EntitySystems/KitchenSpikeSystem.cs
+++ b/Content.Server/Kitchen/EntitySystems/KitchenSpikeSystem.cs
@@ -1,5 +1,6 @@
 using Content.Server.Administration.Logs;
 using Content.Server.Body.Systems;
+using Content.Server.Forensics;
 using Content.Server.Kitchen.Components;
 using Content.Server.Popups;
 using Content.Shared.Chat;
@@ -7,6 +8,7 @@ using Content.Shared.Damage;
 using Content.Shared.Database;
 using Content.Shared.DoAfter;
 using Content.Shared.DragDrop;
+using Content.Shared.Forensics.Components;
 using Content.Shared.Humanoid;
 using Content.Shared.IdentityManagement;
 using Content.Shared.Interaction;
@@ -15,6 +17,7 @@ using Content.Shared.Kitchen;
 using Content.Shared.Kitchen.Components;
 using Content.Shared.Mobs.Components;
 using Content.Shared.Mobs.Systems;
+using Content.Shared.NameModifier.EntitySystems;
 using Content.Shared.Nutrition.Components;
 using Content.Shared.Popups;
 using Content.Shared.Storage;
@@ -39,6 +42,7 @@ namespace Content.Server.Kitchen.EntitySystems
         [Dependency] private readonly SharedAudioSystem _audio = default!;
         [Dependency] private readonly MetaDataSystem _metaData = default!;
         [Dependency] private readonly SharedSuicideSystem _suicide = default!;
+        [Dependency] private readonly NameModifierSystem _nameModifier = default!;
 
         public override void Initialize()
         {
@@ -47,6 +51,7 @@ namespace Content.Server.Kitchen.EntitySystems
             SubscribeLocalEvent<KitchenSpikeComponent, InteractUsingEvent>(OnInteractUsing);
             SubscribeLocalEvent<KitchenSpikeComponent, InteractHandEvent>(OnInteractHand);
             SubscribeLocalEvent<KitchenSpikeComponent, DragDropTargetEvent>(OnDragDrop);
+            SubscribeLocalEvent<KitchenSpikeComponent, RefreshNameModifiersEvent>(OnRefreshNameModifiers);
 
             //DoAfter
             SubscribeLocalEvent<KitchenSpikeComponent, SpikeDoAfterEvent>(OnDoAfter);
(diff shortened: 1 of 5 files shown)


themias:
If you cut the meat into cutlets, do it retain the DNA?

I guess at a certain point, we get to, "does the cooked hamburger have the DNA?", so I don't know how far we'd want to go with that.

TeenSarlacc:
> If you cut the meat into cutlets, do they retain the DNA?
> 
> I guess at a certain point, we get to, "does the cooked hamburger have the DNA?", so I don't know how far we'd want to go with that.

turning my kill target into meat smoothie so the det can't scan it :godo:

SlamBamActionman:
> If you cut the meat into cutlets, or grind into "animal protein" for meatballs, do they retain the DNA?
> 
> I guess at a certain point, we get to, "does the cooked hamburger have the DNA?", so I don't know how far we'd want to go with that.

It doesn't afaik, but I see that more as a cooking code issue than meat spike. We'd probably want a way to make recipes define whether they destroy forensics or not, (e.g. cutlets keep DNA, but cooking a burger doesn't), but that's a story for another PR.

Content.Server/Kitchen/EntitySystems/KitchenSpikeSystem.cs
```diff
@@ -194,16 +209,31 @@ private bool TryGetPiece(EntityUid uid, EntityUid user, EntityUid used,
 
             var item = _random.PickAndTake(component.PrototypesToSpawn);
 
-            var ent = Spawn(item, Transform(uid).Coordinates);
-            _metaData.SetEntityName(ent,
-                Loc.GetString("comp-kitchen-spike-meat-name", ("name", Name(ent)), ("victim", component.Victim)));
+            var piece = Spawn(item, Transform(uid).Coordinates);
+
+            var ev = new KitchenSpikeGetPieceEvent(piece);
+            RaiseLocalEvent(user, ref ev);
```
slarticodefast, line 215: This probably doesn't even need to be an event, or is there anything else that would use this in the future?
Also since you are subscribing with the ForensicsComponent this will only happen if the user has blood, dna or other traces of some form on them (not from their own DNAComponent etc).
And doesn't SharedInteractionSystem already handle this with `DoContactInteraction`?

Content.Shared/Kitchen/Components/KitchenSpikeComponent.cs
```diff
@@ -11,18 +11,27 @@ public sealed partial class KitchenSpikeComponent : Component
...
+
+    [DataField]
     public string MeatSource0 = "?";
-    public string Victim = "?";
+
+    [DataField]
+    public string? VictimName = null;
+
+    [DataField]
+    public string? VictimDna = null;
```
slarticodefast, line 31: documentation

Content.Server/Kitchen/EntitySystems/KitchenSpikeSystem.cs
```diff
@@ -169,13 +184,13 @@ private void Spike(EntityUid uid, EntityUid userUid, EntityUid victimUid,
                                                     ("this", uid)),
                                     uid, PopupType.LargeCaution);
 
+            component.VictimName = Name(victimUid);
```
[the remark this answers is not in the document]
SlamBamActionman: I think the name is necessary here, since it is meant to be a completely undressed character on the meatspike. Any ID or face covering gets removed upon spiking.

Content.Server/Kitchen/EntitySystems/KitchenSpikeSystem.cs
slarticodefast: But can't you dirctly call DoContactInteraction for the meat here instead of raising an event?

SlamBamActionman: The fool I am only looked at `ContactInteractionEvent` which is only used in one place and therefore frightened me, rather than `DoContactInteraction` which is used in many places and therefore safe.

`DoContactInteraction` makes sense, I'll use that! :)

SlamBamActionman: In short: I misread, like a dumdum. 🙇

slarticodefast:
I'll go ahead and close this since this has been merge conflicted to hell by https://github.com/space-wizards/space-station-14/pull/38723 and will be even more merge conflicted once I merge the forensics prediction PR, so this will have to be completely rewritten from scratch.
I'll make an issue for this.

PR #37696, opened 2025-05-21
Cogni on the nukie planet so people can disconnect without an admins help.

Goo-Jargn:
About the PR
Added Cogni to the syndiejuice machine.

Why / Balance
Getting the chance to play a nuclear operative, commander or agent is an exceptionally rare event, but unfortunately things happen without warning, whether the emergency is social, medical or otherwise, there are times where you have to get up.
This can be rough because if no admin is on, you have effectively just lost a player, granted there are some adminless remedies, i.e buying a reinforcement, but that deeply digs into your valueable telecrystal.
By providing nukies cognie innately, they are able to full on prevent these rare, yet troubling circumstances entirely without the need of third part intervention, saving both the admins and the players stress.

Technical details
Added one line of code to vending_machines.yml.

Media
image

Requirements
[ X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
[X ] I have added media to this PR or it does not require an ingame showcase.
Breaking changes
Changelog
Added cognizine to the syndiejuice so admin intervention is no longer required in the event of a nukie having an emergency and needing to leave.

### DIFF
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/chemvend.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/chemvend.yml
@@ -54,6 +54,7 @@
     JugSulfur: 1
     JugWeldingFuel: 1
     PlasmaChemistryVial: 1
+    ChemistryBottleCognizine: 1
   contrabandInventory:
     DrinkLithiumFlask: 1
     StrangePill: 3


TeenSarlacc:
wouldn't having the nukie be available as a ghostrole when the player leaves be a better option?

YoungThugSS14:
> wouldn't having the nukie be available as a ghostrole when the player leaves be a better option?

Yeah with this, Nukies can cogni their monkeys and kobalds so I don’t think this is a good idea

ScarKy0:
> Yeah with this, Nukies can cogni their monkeys and kobalds so I don’t think this is a good idea

Their monkeys and kobolds are already ghostroles, it doesnt make a differance

YoungThugSS14:
> > Yeah with this, Nukies can cogni their monkeys and kobalds so I don’t think this is a good idea
> 
> Their monkeys and kobolds are already ghostroles, it doesnt make a differance

Yeah but now they could speak

EmoGarbage404:
I do not think this is a good solution. Players leaving mid-round is unfortunate but just providing the nukies with free ghost roles is not a great way to handle it. At the very least this should be some kind of ghost role lottery that opens up rather than being at the nukie's discretion.

Goo-Jargn:
> wouldn't having the nukie be available as a ghostrole when the player leaves be a better option?

Could you tell me how to go about this in that case?

iaada:
There might be a concern of players getting their role "stolen" as a result of temporary internet issues. The original player should still get their role back if they only disconnected for a brief time.

Goo-Jargn:
> There might be a concern of players getting their role "stolen" as a result of temporary internet issues. The original player should still get their role back if they only disconnected for a brief time.

cogni only works on people who use /ghost

Nukesthestation:
> cogni only works on people who use /ghost

Doesn't that mean it won't work anyway, since most of the time, players just disconnect and not /ghost?

Goo-Jargn:
> > cogni only works on people who use /ghost
> 
> Doesn't that mean it won't work anyway, since most of the time, players just disconnect and not /ghost?

there is two types of "soul" text.
Yellow is for people who have disconnected, it says "the soul remains inert" or something along those lines, this occurs if you simply just leave. Cogni does not work on this.
Purple text is for people who have used /ghost. Cogni works on exclusively *them*.

Nukesthestation:
> Yellow is for people who have disconnected, it says "the soul remains inert" or something along those lines, this occurs if you simply just leave. Cogni does not work on this.

So there's still a problem if the leaving nukie doesn't /ghost (half the time I find that most don't /ghost and just disconnect), meaning that this change would not affect anything (this is just based off what I have seen my self)

Goo-Jargn:
> > Yellow is for people who have disconnected, it says "the soul remains inert" or something along those lines, this occurs if you simply just leave. Cogni does not work on this.
> 
> So there's still a problem if the leaving nukie doesn't /ghost (half the time I find that most don't /ghost and just disconnect), meaning that this change would not affect anything (this is just based off what I have seen my self)

Yes and no, those people who just flat out dc dont bother to /ghost since their team mates cant do squat about it anyways is my guess.
And this is more of a blanket solution for the mean time until I can figure out how to make it a ghost role. I am MAYBE hoping to try and figure out if there can be a timer on "this motherfucker aint comin back", like some odd 10-15 minutes?

Nukesthestation:
> Yes and no, those people who just flat out dc dont bother to /ghost since their team mates cant do squat about it anyways is my guess.

I mean also before /ghost did nothing iirc, so I think it was just prefrance since either way an admin would need to intervene, personally the few times where I got nukie and something came up irl and I had to go, I used the disconnect not /ghost

Samuka-C:
What if a nukie /ghosted, it became a ghostrole, and if it was SSD (yellow text) for some time (5-10 minutes) it ghosted automatically, then also turning into a ghostsrole

Goo-Jargn:
> What if a nukie /ghosted, it became a ghostrole, and if it was SSD (yellow text) for some time (5-10 minutes) it ghosted automatically, then also turning into a ghostsrole

Not a bad idea, ill try working on this in a bit
Be warned, itll take a while, I barely managed to figure out this one.
But god damn it ill find a way to make admin intervention less required somewhere!!!

PR #37697, opened 2025-05-21
Removed recoil on hitscan laser weapons.

RedBookcase:
## About the PR
Hitscan laser weapons have had their recoil reduced to zero.

## Why / Balance
According to all my google related research, and in keeping with popular sci-fi trends, lasers should not have appreciable recoil even when used as weapons. While the amount of recoil from any sort of focused-light beam is greater than zero, it is not strong enough to produce any sort of noticeable recoil (see flashlights).

Of course realism concerns are secondary to gameplay and ultimately that's the real focus of the PR. Laser weapons are a distinct class of weaponry when compared to kinetic firearms and they should both act and feel differently when compared against one another. Lasers act differently (hitscan beams that can pass through glass vs. kinetic projectiles that have travel time) but they feel similar to actually fire as a result of having identical recoil. This PR seeks to change that by removing recoil from hitscan lasers to make them feel more distinct and unique.

Projectile-based energy weapons (Disablers, the CHIMP, the Energy Shotgun, etc.) have kept their recoil at the default value of one. As they fire projectiles with travel time instead of hitscan lasers I feel that they are distinct from one another and did not attempt to tweak their recoil.

Tasers have a lowered recoil of 0.25 now however. Not that it matters much since they're not obtainable in regular gameplay but they shouldn't have the same kick as a ballistic weapon.

## Technical details
changes in battery_guns.yml to tweak laser recoil. The base versions of the different energy weapons have been given a recoil of zero and the projectile-based prototypes have been updated to keep their recoil the same. Notes also added for future reference.

## Media
Here's a test video I made when I was fiddling with laser recoil. Showcases recoilless lasers, the lower-recoil taser, and regular-recoil disablers.

https://github.com/user-attachments/assets/94b83df2-e387-43aa-9fe8-77034f10cc92

## Requirements

**Changelog**

:cl:
- tweak: Hitscan laser weapons no longer have recoil.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -14,6 +14,7 @@
     - suitStorage
   - type: AmmoCounter
   - type: Gun
+    cameraRecoilScalar: 0 #Notice: Lasers have 0 recoil, other energy projectile weapons like disabler and eshotgun vary.
     fireRate: 2
     selectedMode: SemiAuto
     availableModes:
@@ -37,6 +38,7 @@
   - type: AmmoCounter
   - type: Gun
     fireRate: 2
+    cameraRecoilScalar: 0 #Notice: Lasers have 0 recoil, other energy projectile weapons like disabler and eshotgun vary.
     selectedMode: SemiAuto
     availableModes:
     - SemiAuto
@@ -467,6 +469,7 @@
   - type: Gun
     fireRate: 2
     projectileSpeed: 35 # any higher and this causes issues in lag
+    cameraRecoilScalar: 1 #Placeholder - it fires a projectile, not a hitscan beam.
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/taser2.ogg
   - type: ProjectileBatteryAmmoProvider
@@ -527,6 +530,7 @@
   - type: Gun
     selectedMode: FullAuto
     fireRate: 4.5
+    cameraRecoilScalar: 1 #Placeholder - it fires a projectile, not a hitscan beam.
     availableModes:
       - SemiAuto
       - FullAuto
@@ -568,6 +572,7 @@
     slots:
     - Belt
   - type: Gun
+    cameraRecoilScalar: .25 #I watched a YouTube video of a taser test fire and it looked like it had a teeny tiny amount of recoil
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/taser.ogg
   - type: ProjectileBatteryAmmoProvider
@@ -683,6 +688,7 @@
   - type: Gun
     projectileSpeed: 10
     fireRate: 1.5
+    cameraRecoilScalar: 1 #Placeholder - it fires a projectile, not a hitscan beam.
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/taser2.ogg
   - type: ProjectileBatteryAmmoProvider
@@ -770,6 +776,7 @@
     sprite: Objects/Weapons/Guns/Battery/energy_shotgun.rsi
   - type: Gun
     fireRate: 2
+    cameraRecoilScalar: 1 #Placeholder - it fires a projectile, not a hitscan beam.
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/laser_cannon.ogg
   - type: ProjectileBatteryAmmoProvider
@@ -831,6 +838,7 @@
     sprite: Objects/Weapons/Guns/Battery/temp_gun.rsi
   - type: Gun
     fireRate: 1
+    cameraRecoilScalar: 1 #Placeholder - it fires a projectile, not a hitscan beam.
     soundGunshot:
       path: /Audio/Weapons/Guns/Gunshots/taser2.ogg
   - type: ProjectileBatteryAmmoProvider


TeenSarlacc:
while you're tweaking things, the foam force rifle has a lot of recoil too, at least imo.

slarticodefast:
Oh, you mean camera recoil. Because these weapons also push you backwards when shooting them in space. You might want to adjust that as well.

TaoNewt:
Laser weapons are all very distinct from kinetics, further differentiation isn't required on that ground as they feel and perform very differently. 
Mechanical, visual and audible feedback are important contributions to having something feel good ingame, stripping away a form of feedback impacts the experience.

slarticodefast:
> Does that have a specific component tied to it?

Well, currently it's a magic number hardcoded into the `SharedGunSystem`
grafik
should probably be moved either to the gun, ammo, or projectile's component

Redrover1760:
I dunno, I like the little camera recoil lasers give when I shoot em. Gives more oomph to each shot.

Content.Shared/Weapons/Ranged/Components/GunComponent.cs
```diff
@@ -56,6 +56,18 @@ public sealed partial class GunComponent : Component
...
     public float CameraRecoilScalarModified = 1f;
 
+    /// <summary>
+    /// The base value applied to the calculation governing knockback when fired in zero grav.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float ImpulseStrength = 25f;
+
+    /// <summary>
+    /// Modified version of ImpulseStrength for posterity(?).
```
beck-thompson, line 66: Just remove this comment

Content.Shared/Weapons/Ranged/Components/GunComponent.cs
```diff
@@ -56,6 +56,18 @@ public sealed partial class GunComponent : Component
...
+    /// </summary>
+    [DataField, AutoNetworkedField]
```
beck-thompson, line 68: I think for this particular case you can just do `[AutoNetworkedField]`

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
```diff
@@ -511,14 +512,13 @@ protected void MuzzleFlash(EntityUid gun, AmmoComponent component, Angle worldAn
...
-    public void CauseImpulse(EntityCoordinates fromCoordinates, EntityCoordinates toCoordinates, EntityUid user, PhysicsComponent userPhysics)
+    public void CauseImpulse(Entity<GunComponent> gun, EntityCoordinates fromCoordinates, EntityCoordinates toCoordinates, EntityUid user, PhysicsComponent userPhysics)
     {
         var fromMap = TransformSystem.ToMapCoordinates(fromCoordinates).Position;
         var toMap = TransformSystem.ToMapCoordinates(toCoordinates).Position;
         var shotDirection = (toMap - fromMap).Normalized();
 
-        const float impulseStrength = 25.0f;
-        var impulseVector =  shotDirection * impulseStrength;
+        var impulseVector =  shotDirection * gun.Comp.ImpulseStrength;
```
beck-thompson, line 521: After looking into this I think you should be using `ImpulseStrengthModified`. That is like the "real" value and `ImpulseStrength` is just the starting value

beck-thompson (changes requested):

beck-thompson (approved):
code wise it looks good!

Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
```diff
@@ -831,6 +845,8 @@
     sprite: Objects/Weapons/Guns/Battery/temp_gun.rsi
   - type: Gun
     fireRate: 1
+    impulseStrength: 25
```
pheenty, line 852: 25 is the default value, it's redundant to have it everywhere

RedBookcase: This gun (and most of the other projectile energy weapons) parents off of BaseWeaponBattery, meaning that without it's own entry for impulseStrength and cameraRecoilScalar it'll inherit the no-recoil values from BaseWeaponBattery.

RedBookcase:
> Laser weapons are all very distinct from kinetics, further differentiation isn't required on that ground as they feel and perform very differently. Mechanical, visual and audible feedback are important contributions to having something feel good ingame, stripping away a form of feedback impacts the experience.

And i'd argue it impacts the experience in a positive way, as this brings lasers more in-line with their intended status in the game. Lasers are noted as being "Favoured by Nanotrasen Security for being cheap and easy to use.", and removing recoil definitely makes them easier to use. You lose the feedback of the weapon recoiling with every shot, but that allows you to focus more on the type of feedback that is undoubtedly more important to the shooter - "Did I hit my target?"

EmoGarbage404 (changes requested):
i understand that logically lasers should not have recoil, but the loss of the camera movement significantly hurts the visuals of firing the laser. I'm fine with removing the physical impulse, but the camera should still have some movement. I think the taser looked like a good reference point.

Content.Shared/Weapons/Ranged/Components/GunComponent.cs
```diff
@@ -56,6 +56,15 @@ public sealed partial class GunComponent : Component
...
     public float CameraRecoilScalarModified = 1f;
 
+    /// <summary>
+    /// The base value applied to the calculation governing knockback when fired in zero grav.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float ImpulseStrength = 25f;
+
+    [AutoNetworkedField]
+    public float ImpulseStrengthModified;
```
EmoGarbage404: comment + datafield

EmoGarbage404: the naming of these two variables is a bit strange. I think it's more common to use "BaseX" and "X" rather than the `Modified` suffix.

EmoGarbage404 (changes requested):

Content.Shared/Weapons/Ranged/Components/GunComponent.cs
slarticodefast: I would prefer to keep the naming consistent with the existing datafields. So using Modified makes sense here.

RedBookcase: Fixed

RedBookcase: Kept it `Modified` since that was the established norm with GunComponent

RedBookcase:
Seems a lot of the .cs involved in this PR has also been changed in the past few months and I am not up to dive into .cs again to figure out how to adapt this properly. Anyone who'd like to try implementing it is welcome to copy my work!

PR #37712, opened 2025-05-22
Devices with access restrictions list those restrictions in their examination description

chromiumboy:
## About the PR

Entities with the AccessReaderComponent now list those restrictions in their examination description. The examination message can be customized on a per-entity basis

## Why / Balance

It's not immediately apparent to players what access privileges are required to interact with certain objects, particularly if an access configurator has been used to customize their access settings. This change informs players as to who can use the object

## Technical details

The AccessReaderSystem now listens for ExaminedEvent and populates the examination UI with the appropriate details

## Media

An example where the description of the cryogenic sleep unit now specifies that Cryogenics access is required to access certain functions
cryo_example

An example where an airlock has had its access settings altered - players can now readily see who can open it
airlock-example

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Devices with access restrictions now list those restrictions in their examination description

### DIFF
--- a/Content.Shared/Access/Components/AccessReaderComponent.cs
+++ b/Content.Shared/Access/Components/AccessReaderComponent.cs
@@ -77,6 +77,16 @@ public sealed partial class AccessReaderComponent : Component
     /// </summary>
     [DataField]
     public bool BreakOnAccessBreaker = true;
+
+    /// <summary>
+    /// The examination text associated with this component.
+    /// </summary>
+    /// <remarks>
+    /// The text can be supplied with the 'access' variable to populate it
+    /// with a comma separated list of the access levels contained in <see cref="AccessLists"/>
+    /// </remarks>
+    [DataField]
+    public LocId ExaminationText = "access-reader-examination";
 }
 
 [DataDefinition, Serializable, NetSerializable]
--- a/Content.Shared/Access/Systems/AccessReaderSystem.cs
+++ b/Content.Shared/Access/Systems/AccessReaderSystem.cs
@@ -1,8 +1,10 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Text;
 using Content.Shared.Access.Components;
 using Content.Shared.DeviceLinking.Events;
 using Content.Shared.Emag.Systems;
+using Content.Shared.Examine;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Inventory;
 using Content.Shared.NameIdentifier;
@@ -37,11 +39,61 @@ public sealed class AccessReaderSystem : EntitySystem
     {
         base.Initialize();
 
+        SubscribeLocalEvent<AccessReaderComponent, ExaminedEvent>(OnExamined);
         SubscribeLocalEvent<AccessReaderComponent, GotEmaggedEvent>(OnEmagged);
         SubscribeLocalEvent<AccessReaderComponent, LinkAttemptEvent>(OnLinkAttempt);
 
         SubscribeLocalEvent<AccessReaderComponent, ComponentGetState>(OnGetState);
         SubscribeLocalEvent<AccessReaderComponent, ComponentHandleState>(OnHandleState);
+    }
+
+    private void OnExamined(Entity<AccessReaderComponent> ent, ref ExaminedEvent args)
+    {
+        if (!GetMainAccessReader(ent, out var mainAccessReader))
+            return;
+
+        var sb = new StringBuilder();
+        var accessSubgroupCount = mainAccessReader.Value.Comp.AccessLists.Count;
+
+        for (int i = 0; i < accessSubgroupCount; i++)
+        {
+            var accessList = mainAccessReader.Value.Comp.AccessLists[i];
+
+            // Add the names of all access levels in the subgroup
+            for (int j = 0; j < accessList.Count; j++)
+            {
+                var accessLevel = accessList.ElementAt(j);
+                var accessName = Loc.GetString("access-reader-unknown-id");
+
+                if (_prototype.TryIndex(accessLevel, out var accessProto) && !string.IsNullOrWhiteSpace(accessProto.Name))
+                    accessName = Loc.GetString(accessProto.Name);
+
+                sb.Append(Loc.GetString("access-reader-access-label", ("access", accessName)));
+
+                if (j < (accessList.Count - 1))
+                    sb.Append(" & ");
+            }
+
+            if (i < (accessSubgroupCount - 1))
+            {
+                // Add a comma if the list will consist of more than two elements
+                if (accessSubgroupCount > 2)
+                    sb.Append(", ");
+                else
+                    sb.Append(" ");
+            }
+
+            // Add a conjunction if there is only one more element left to add
+            if (i == (accessSubgroupCount - 2))
+                sb.AppendFormat("{0} ", Loc.GetString("access-reader-list-conjunction"));
+        }
+
+        // String is invalid - either there were no access restrictions or their names are invalid
+        if (string.IsNullOrWhiteSpace(sb.ToString()))
+            return;
+
+        var accessMessage = Loc.GetString(mainAccessReader.Value.Comp.ExaminationText, ("access", sb));
+        args.PushMarkup(accessMessage);
     }
 
     private void OnGetState(EntityUid uid, AccessReaderComponent component, ref ComponentGetState args)
--- a/Resources/Locale/en-US/access/systems/access-reader-system.ftl
+++ b/Resources/Locale/en-US/access/systems/access-reader-system.ftl
@@ -1,1 +1,5 @@
 access-reader-unknown-id = Unknown
+access-reader-access-label = [color=yellow]{$access}[/color]
+access-reader-examination = Access is restricted to those with {$access} access.
+access-reader-examination-functionality-restricted = {$access} access is required to use certain functions.
+access-reader-list-conjunction = or
--- a/Resources/Prototypes/Entities/Structures/Doors/Firelocks/firelock.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/Firelocks/firelock.yml
@@ -128,6 +128,7 @@
       price: 150
     - type: AccessReader
       access: [ [ "Engineering" ] ]
+      examinationText: access-reader-examination-functionality-restricted
     - type: PryUnpowered
       pryModifier: 0.5
     - type: PointLight
--- a/Resources/Prototypes/Entities/Structures/Machines/holopad.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/holopad.yml
@@ -74,6 +74,7 @@
     speakerVolume: Speak
   - type: AccessReader
     access: [[ "Command" ]]
+    examinationText: access-reader-examination-functionality-restricted
   - type: ActivatableUI
     key: enum.HolopadUiKey.InteractionWindow
   - type: ActivatableUIRequiresPower
--- a/Resources/Prototypes/Entities/Structures/Storage/Closets/Lockers/lockers.yml
+++ b/Resources/Prototypes/Entities/Structures/Storage/Closets/Lockers/lockers.yml
@@ -400,6 +400,7 @@
         type: GenpopLockerBoundUserInterface
   - type: AccessReader # note! this access is for the UI, not the door. door access is handled on GenpopLocker
     access: [["Security"]]
+    examinationText: access-reader-examination-functionality-restricted
   - type: Lock
     locked: false
     useAccess: false
--- a/Resources/Prototypes/Entities/Structures/cryogenic_sleep_unit.yml
+++ b/Resources/Prototypes/Entities/Structures/cryogenic_sleep_unit.yml
@@ -19,6 +19,7 @@
   - type: AccessReader
     breakOnAccessBreaker: false
     access: [["Cryogenics"]]
+    examinationText: access-reader-examination-functionality-restricted
   - type: InteractionOutline
   - type: Cryostorage
   - type: Physics


Resources/Locale/en-US/access/systems/access-reader-system.ftl
```diff
@@ -1 +1,5 @@
 access-reader-unknown-id = Unknown
+access-reader-access-label = [color=yellow]{$access}[/color]
+access-reader-examination = Access is restricted to those with {$access} access.
+access-reader-examination-functionality-restricted = {$access} access is required to use certain functions.
+access-reader-list-conjunction = or
```
lzk228, line 5: wait... we have a special thing for that
https://github.com/space-wizards/space-station-14/blob/ecff1f88103c5b313b61cc54ff21aa6776261711/Content.Shared/Localizations/ContentLocalizationManager.cs#L128-L137
won't it do?

chromiumboy: Huh, nice! I'll have a look

chromiumboy: So this function formats things oddly - it writes the list as 'X or Y or Z' etc. I've updated the function so it makes a comma-separated list

BigfootBravo:
I think this is good, but maybe remove the ability for doors to have this feature? A door with hacked accesses, be it via the slow way of manually deconing it and changing the circuit, or the use of some other traitor item, would very very easily and quickly give up that they have been altered.

NoElkaTheGod:
Me on my way to enable all access on a door so that description no longer fits on the screen:

BigfootBravo:
> Maybe add a new wire to show or not show this for the airlocks
> 
> this way if you are sabotaging a door and you don't want people to know it has extra access you can cut that wire


This could work a bit better, but if that wire is cut, it would still indicate at a glance if a door has been altered at all. As if the door is intact you could see the accesses, and if not it would be hidden.

Maybe it could be like a additional component to a door, like how plates can be added and welded to increase wire security,  and so doors with this component, added and set in a similar process, would show the accesses? Like you add a special *display board* or something to it, gotten from a lathe, or maybe just adding extra wires to it at a specific stage of (de)construction, like the addition of steel to add a plate over the wires. Like the secure plates, it could roundstart on the more secure doors, but be able to be manually added as new functionality by an engineer? Or maybe make it so the only way for doors to get it is for a engineer to add this new component in the fjrst place, giving engies some early, roundstart jobs, or just more things to do over the round?

chromiumboy:
Hmm. I don't think airlock access lists should be an optional extra, it should either be there or not. It would be easy to have doors report 'Access is restricted to authorized crewmembers' but it's a bit of a cop out. How likely are veteran players going to be randomly inspecting doors to see if their access settings have been changed?

ScarKy0:
How about just saying its restricted but a diagnostics hud tells you the exact access

Samuka-C:
> How about just saying its restricted but a diagnostics hud tells you the exact access

I think is perfect

chromiumboy:
> How about just saying its restricted but a diagnostics hud tells you the exact access

Yeah, we could do that. But I was thinking more about new players, it'd be helpful for them to be able to look at a door and understand who can access it. There's always a trade off being giving players more information and antag shenanigans. Does this change noticeably degrade the antag experience more than what other players could benefit?

Samuka-C:
I think it degrades antag experience if it always showed the access

some antags might add extra access to door for easy escape routes or surprise attacks, thiefs might make custom doors in maints to hide their loot and rely on rare access tags to keep the door from being open by anyone.

if a secoff checks a random door in maints and it's locked, some wouldn't bat a eye, but if they saw it was chem access for example, they would much more likely to suspect of being a thief's den and would mark all chemists as suspect

Samuka-C:
> Getting off-topic but a good use for diagnostic HUD would be to read cables behind panels and check if something is emagged

the diagnostic HUD needs more interactions overall, only being able to see borg hp is kinda useless

EmoGarbage404:
maintainer thoughtpost ahead:

One the one hand, this is obviously a very big UX improvement. Stations can often have confusing layouts and reused door designs with different accesses can be disorienting (looking at you, service dept.). New players would obviously benefit from having something like this.

However, i do concede that it does eliminate a lot of the potential for antagonists and ne'er-do-wells to modify the station with traps and secret areas.

My off-the-cuff idea would be to have each door design associated with a particular set of accesses, display _those_ in the examine text, and then have an additional line if the accesses of the door do not match it exactly.

For example:
red security door -> "Access is restricted to those with Security access"
red security door (modified) -> "Access is restricted to those with Security access, though _the access appears to be modified_"

This obviously only works if every access has a unique sprite, though that is something we should probably have anyways.

chromiumboy:
> My off-the-cuff idea would be to have each door design associated with a particular set of accesses, display _those_ in the examine text, and then have an additional line if the accesses of the door do not match it exactly.
.
I like this approach. It can be extended to devices other than doors as well. The description would display the access settings of the device as they were set when it spawns. But if you're wearing a diagnostic HUD, it also displays any modified access settings

EmoGarbage404:
Yes!

EmoGarbage404:
> Hmm, I think it should not be possible to see if access for doors have been modified at all. It restricts way too many options for trap making or other interesting mechanics. Checking if a device has been tampered with should not be visible at a glance.

perhaps seeing the modified accesses at all should remain relegated to the diagnostic HUD? I do think that being able to see what access you need for a door is pretty invaluable, especially when we have lots of identical looking doors with varied accesses (think bridge)

chromiumboy:
> > My off-the-cuff idea would be to have each door design associated with a particular set of accesses, display those in the examine text, and then have an additional line if the accesses of the door do not match it exactly.
> 
> Hmm, I think it should not be possible to see if access for doors have been modified in the examination text. It restricts way too many options for trap making or other interesting mechanics. Checking if a device has been tampered with should not be visible at a glance. If you need to know that is what the access configurator is for.
> 
> For other devices like the cryogenics pod this examination text is a good idea though!

That's why I amended Emo's suggestion to rwqyire the diagnostic HUD to see if the access settings were changed. There wouldn't be any indication otherwise. Diagnostic HUDs aren't very common

slarticodefast:
Yeah, that making it HUD only would work!

ike709:
I think it should be a new examine button (like the Help button or the inspect ID button) just to reduce description bloat. This info shouldn't really be relevant most of the time unless someone is actively trying to figure out why they don't have access.

BigfootBravo:
> I think it should be a new examine button (like the Help button or the inspect ID button) just to reduce description bloat. This info shouldn't really be relevant most of the time unless someone is actively trying to figure out why they don't have access.

This would be good for reasons you describe, but I do also think it should require the diagnostic hud, like others have mentioned. Atleast some tool is required, and active inspection to look it over.

Otherwise I could see people just examining every door they come across for any oddities, which would be very easy if this addition was added without such consideration.

chromiumboy:
I've updated the PR in line with the discussed changes, but I need a better way of copying the AccessList datafield (one that won't break tests). And no, I'm not going to duplicate the field across 761 yaml entity prototypes... Open to suggestions

Content.Shared/Access/Components/AccessReaderComponent.cs
```diff
@@ -34,6 +34,13 @@ public sealed partial class AccessReaderComponent : Component
     [DataField("access")] [ViewVariables(VVAccess.ReadWrite)]
     public List<HashSet<ProtoId<AccessLevelPrototype>>> AccessLists = new();
 
+    /// <summary>
+    /// An unmodified copy of the original list of the access groups that grant access to this reader.
+    /// </summary>
+    [DataField]
+    [ViewVariables(VVAccess.ReadWrite)]
+    public List<HashSet<ProtoId<AccessLevelPrototype>>>? AccessListsOriginal = null;
+
```
metalgearsloth, line 43: This needs to be networked or client code can just straight up fail.

Nukesthestation:
i think that it should say something like:  "Access is normally restricted to those with (access)" if you are not using the diagnostic hubs as at least to me, I think that right now it just saying "access is restricted to those with (access)" implies that it is always that specific access and never something else which could be confusing for somepeople if they see a door that says what it is restricted to but someone without that access can just open it contradicting what it says. (This might just be me idk so take my word with a grain of salt ig)

Samuka-C:
If I construct a door and add random accesses to it, what shows in the description?

chromiumboy:
> If I construct a door and add random accesses to it, what shows in the description?

... didn't you see all the acreenshots in the media section?

Samuka-C:
> > If I construct a door and add random accesses to it, what shows in the description?
> 
> ... didn't you see all the acreenshots in the media section?

it doesn't show this specific scenario, or if it does, which one is it?

chromiumboy:
> > > If I construct a door and add random accesses to it, what shows in the description?
> > 
> > 
> > ... didn't you see all the acreenshots in the media section?
> 
> it doesn't show this specific scenario, or if it does, which one is it?

Oh, sorry, I see what you mean. If the door has no default access restrictions, like a freshly built airlock, there won't be any messages about the settings when you're not wearing a HUD (the examine message looks like how they currently are in game). If you are wearing a HUD, and the settings have been changed, it'll look like screenshot three

Samuka-C:
> > > > If I construct a door and add random accesses to it, what shows in the description?
> > > 
> > > 
> > > ... didn't you see all the acreenshots in the media section?
> > 
> > it doesn't show this specific scenario, or if it does, which one is it?
> 
> Oh, sorry, I see what you mean. If the door has no default access restrictions, like a freshly built airlock, there won't be any messages about the settings when you're not wearing a HUD (the examine message looks like how they currently are in game). If you are wearing a HUD, and the settings have been changed, it'll look like screenshot three

I see, thanks, however, there is a small issue for evil dooers

If you want to add extra access to a door without stealing a access config, you need to remove the door eletronics which deconstructs the door, so when you add the custom access and then add the door eletronics back, it will reconstruct the door

so the door will not show the access, and then is obvious even to those without diagnostic huds that the door was messed with

chromiumboy:
> > > > > If I construct a door and add random accesses to it, what shows in the description?
> > > > 
> > > > 
> > > > ... didn't you see all the acreenshots in the media section?
> > > 
> > > 
> > > it doesn't show this specific scenario, or if it does, which one is it?
> > 
> > 
> > Oh, sorry, I see what you mean. If the door has no default access restrictions, like a freshly built airlock, there won't be any messages about the settings when you're not wearing a HUD (the examine message looks like how they currently are in game). If you are wearing a HUD, and the settings have been changed, it'll look like screenshot three
> 
> I see, thanks, however, there is a small issue for evil dooers
> 
> If you want to add extra access to a door without stealing a access config, you need to remove the door eletronics which deconstructs the door, so when you add the custom access and then add the door eletronics back, it will reconstruct the door
> 
> so the door will not show the access, and then is obvious even to those without diagnostic huds that the door was messed with

I haven't tested that particular scenario, but if you build a door with electronics taken from another door (i.e. one that you deconstructed) it should show the access settings like in screenshots two and three. So this shouldn't be an issue

chromiumboy:
I checked, and yes, it's not an issue if you modify some door electronics and build a new airlock with them

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -37,17 +40,74 @@ public override void Initialize()
...
     }
 
+    private void OnExamined(Entity<AccessReaderComponent> ent, ref ExaminedEvent args)
+    {
+        if (!GetMainAccessReader(ent, out var mainAccessReader))
+            return;
+
+        // If there is no copy of the  original settings, copy the current access settings
+        if (mainAccessReader.Value.Comp.AccessListsOriginal == null)
+            mainAccessReader.Value.Comp.AccessListsOriginal = new(mainAccessReader.Value.Comp.AccessLists);
```
Fildrance: there is no need to copy it 'just in case' - you probably would be better off with just local variable (Original is empty - just put reference to Comp.AccessLists, use AccessListsOriginal otherwise)

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -467,6 +587,15 @@ public void RemoveAccesses(Entity<AccessReaderComponent> ent, List<ProtoId<Acces
     /// <param name="dirty">If true, the component will be marked as changed afterward.</param>
     public void RemoveAccess(Entity<AccessReaderComponent> ent, HashSet<ProtoId<AccessLevelPrototype>> access, bool dirty = true)
     {
+        if (dirty)
```
Fildrance: Having cancel only work in case dirty is true looks a bit sus :|

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -37,17 +40,74 @@ public override void Initialize()
...
+            {
+                if (!mainAccessReader.Value.Comp.AccessListsOriginal.Any(y => y.SetEquals(accessSubgroup)))
+                {
+                    accessHasBeenModified = true;
+                    break;
+                }
+            }
+        }
+
+        var canSeeAccessModification = accessHasBeenModified &&
```
Fildrance, line 73: maybe something like 
```

        var canSeeAccessModification = accessHasBeenModified
                                       && (
                                           HasComp<ShowAccessReaderSettingsComponent>(ent)
                                           || _inventorySystem.TryGetInventoryEntity<ShowAccessReaderSettingsComponent>(args.Examiner, out _)
                                        );
```
godo

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -37,17 +40,74 @@ public override void Initialize()
...
+            (HasComp<ShowAccessReaderSettingsComponent>(ent) || _inventorySystem.TryGetInventoryEntity<ShowAccessReaderSettingsComponent>(args.Examiner, out _));
+
+        if (canSeeAccessModification)
+        {
+            var localizedCurrentNames = GetLocalizedAccessNames(mainAccessReader.Value.Comp.AccessLists);
+
+            var currentSettingsMessage = localizedCurrentNames.Count > 0 ?
```
Fildrance: question of stylistic but i'd perosnally prefer something like :>
```
            var accessesFormatted = ContentLocalizationManager.FormatListToOr(localizedCurrentNames);
            var currentSettingsMessage = localizedCurrentNames.Count > 0
                ? Loc.GetString("access-reader-access-settings-modified-message", ("access", accessesFormatted))
                : Loc.GetString("access-reader-access-settings-removed-message");
```

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -37,17 +40,74 @@ public override void Initialize()
...
+            return;
+        }
+
+        var localizedOriginalNames = GetLocalizedAccessNames(mainAccessReader.Value.Comp.AccessListsOriginal);
+
+        // If the string list is empty either there were no access restrictions or the localized names were invalid
+        if (localizedOriginalNames.Count == 0)
+            return;
+
+        var originalSettingsMessage = Loc.GetString(mainAccessReader.Value.Comp.ExaminationText, ("access", ContentLocalizationManager.FormatListToOr(localizedOriginalNames)));
```
Fildrance: consider as an option - extracting variables and keeping 1 invocation per line as a rule makes lines shorter and usually easier to read :3
```
        var originalAccessesFormatted = ContentLocalizationManager.FormatListToOr(localizedOriginalNames);
        var originalSettingsMessage = Loc.GetString(mainAccessReader.Value.Comp.ExaminationText, ("access", originalAccessesFormatted));
```

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -100,6 +161,14 @@ private void OnEmagged(EntityUid uid, AccessReaderComponent reader, ref GotEmagg
         Dirty(uid, reader);
     }
 
+    private void OnConfigurationAttempt(Entity<AccessReaderComponent> ent, ref AccessReaderConfigurationAttemptEvent args)
+    {
+        // The first time that the access list of the reader is modified,
+        // make a copy of the original settings
+        if (ent.Comp.AccessListsOriginal == null)
```
Fildrance: having this on attempt is sad, but i understand why - it would be quite bothersome to keep original access collection copy through all transformations and output it in event after transformation is done, and its not clear which one is more memory wasteful. UI frameworks like Avalonia prefer give better UX so they absolutely do leave you with original/final version of objects upon receiving event, but we sadly dont have such strict code requirements :> so whatevs

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -727,4 +856,34 @@ public void LogAccess(Entity<AccessReaderComponent> ent, string name, TimeSpan?
...
+        {
+            var sb = new StringBuilder();
+            var accessSubset = accessHashSet.ToList();
+
+            // Combine the names of all access levels in the subset into a single string
+            foreach (var access in accessSubset)
+            {
+                var accessName = Loc.GetString("access-reader-unknown-id");
+
+                if (_prototype.TryIndex(access, out var accessProto) && !string.IsNullOrWhiteSpace(accessProto.Name))
```
Fildrance: that should be Resolve and not TryIndex, no?

Content.Shared/Access/Systems/AccessReaderSystem.cs
```diff
@@ -727,4 +856,34 @@ public void LogAccess(Entity<AccessReaderComponent> ent, string name, TimeSpan?
...
+                    accessName = Loc.GetString(accessProto.Name);
+
+                sb.Append(Loc.GetString("access-reader-access-label", ("access", accessName)));
+
+                if (accessSubset.IndexOf(access) < accessSubset.Count - 1)
```
Fildrance: How is that reachable? :( i tried some combos, but never got it
And & is not localizable thing, kinda.

Content.Shared/Localizations/ContentLocalizationManager.cs
```diff
@@ -132,7 +132,7 @@ public static string FormatListToOr(List<string> list)
                 <= 0 => string.Empty,
                 1 => list[0],
                 2 => $"{list[0]} or {list[1]}",
-                _ => $"{string.Join(" or ", list)}"
+                _ => $"{string.Join(", ", list.GetRange(0, list.Count - 1))}, or {list[^1]}"
```
Fildrance, line 135: brother, i dont think you are right there..

image

Content.Shared/Access/Systems/AccessReaderSystem.cs
chromiumboy: :death:

Content.Shared/Access/Systems/AccessReaderSystem.cs
chromiumboy: It only triggers if the hashset contains more than one access level. This never actually happens upstream, as none of our access readers require multiple access levels to trigger them. But the system supports it, so I figured I would too....

Content.Shared/Access/Systems/AccessReaderSystem.cs
chromiumboy: It's so that the AccessReaderConfigurationAttemptEvent doesn't get raised multiple times when RemoveAccesses is looping over this function multiple times. I could move the event out, but then, someone might want to call RemoveAccess directly, and it won't get raised

Content.Shared/Localizations/ContentLocalizationManager.cs
chromiumboy: What do you mean?

Content.Shared/Access/Systems/AccessReaderSystem.cs
chromiumboy: I could make the method private, and make calls external to the system use a 'TryRemoveAccess' function with the event raised there

Content.Shared/Localizations/ContentLocalizationManager.cs
Fildrance: Fresh eye found out that those are AND and OR concat (my blurred tired vision saw them as 100% equal code). 
My bad.
But we i have to go look where that thing was used :sadge: fine otherwise

Content.Shared/Access/Systems/AccessReaderSystem.cs
Fildrance: Having all API calls changed to TryRemove would be fitting, because it will clearly telegraph possibility of change not getting applied. I am feeling a not sure how big of a change that would be tho O_o

chromiumboy: Yeah, calling any function from an external system where an event might cancel it should be a 'try' function TBH. Gunna make the change later

Content.Shared/Access/Systems/AccessReaderSystem.cs
chromiumboy: I've made the conjunction localizable:

image

Fildrance (approved):
Looks good to me!

PR #37719, opened 2025-05-22
Tech disk resprite and pricing changes

murolem:
## About the PR

GAMBA

Changes tech disks from having a random color that doesn't mean anything to a color that fits a specific research discipline, plus some extra markings for tier (they even glow in dark!).

Also changes pricing from 100 per disk to tier-based:
- 100 for tier 1
- 500 for tier 2
- 1500 for tier 3

With current random distribution when printing these (25 10 1), per 100 disk you would get about 70 tier 1, 27 tier 2, and 3 tier 3, all totaling 25000.

There's one thing I would like to fix but I have no idea how and it is predictable examine. Not required, but a nice to have if someone is willing to explain how (pretty please).

## Why / Balance

Speedups the gambling process by allowing to view disk's discipline and tier by just looking at it. May also positively impact locating them in dark places for salvages as tier 2/3 have a few marks that glow in dark.

Increase in price should make these more for grabs by salv and make sci do more gamba in general. And it just made sense for higher tiers to be more pricy - shine doesn't come free.

## Technical details

- `TechnologyDiskComponent` got 2 new optional fields: `Discipline` and `Tier`.
- `StaticPriceComponent` moved from server to shared due to having to set price dynamically. 
- `TechnologyDiskSystem` expanded to handle now more configurable `TechnologyDiskComponent` and to control visuals on the disk proto.
- Disk proto expanded to have GenericVisualizer visuals. Also added separate disk variations for all three tiers that hold the pricing for their respectable tiers.

## Media

Content Client_ZW03bktr77

Content Client_COc1Kzk1Ci

## Requirements

## Breaking changes

- `StaticPriceComponent` moved from server to shared. When referencing in code, now requires `using Content.Shared.Cargo.Components;` directive.

**Changelog**

:cl:
- tweak: Retextured technology disk to have the color of their discipline and extra marks for each tier. Also increased sell price for higher-tier disks.


iaada:
I like it, bit I'm not a fan of the three pixels in the top right for tier 3. I'd rather see the silver slider thing be gold.

Content.Server/Research/TechnologyDisk/Systems/DiskConsoleSystem.cs
```diff
@@ -38,10 +49,24 @@ public override void Update(float frameTime)
...
+    private void SpawnDisk(DiskConsoleComponent console, EntityCoordinates coordinates)
+    {
+        if (!_protoMan.TryIndex(console.DiskPrototype, out var diskProto))
+        {
+            _sawmill.Error($"Failed to spawn a tech disk: disk entity prototype '{console.DiskPrototype}' not found");
+            return;
+        }
+
+        Spawn(diskProto.ID, coordinates);
+    }
```
slarticodefast, line 68: This is not needed. Just change `DiskPrototype` from a `string` to a `EntProtoId` which automatically validates it.

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -20,41 +23,182 @@ public sealed class TechnologyDiskSystem : EntitySystem
     [Dependency] private readonly SharedPopupSystem _popup = default!;
     [Dependency] private readonly SharedResearchSystem _research = default!;
     [Dependency] private readonly SharedLatheSystem _lathe = default!;
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
+    [Dependency] private readonly IComponentFactory _compFactory = default!;
+    [Dependency] private readonly ILogManager _log = default!;
+
+    private ISawmill _sawmill = default!;
```
slarticodefast, line 30: just use `Log.Error` or `Log.Warning` instead of sawmill

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -20,41 +23,182 @@ public sealed class TechnologyDiskSystem : EntitySystem
...
+
+    /// <summary>
+    /// Maps disk tiers to tech disk entity prototypes that hold StaticPrice for each tier.
+    /// </summary>
+    private readonly Dictionary<int, EntProtoId<StaticPriceComponent>> _diskTierToDiskProtoWithSellPrice = new()
+    {
+        [1] = "TechnologyDiskT1",
+        [2] = "TechnologyDiskT2",
+        [3] = "TechnologyDiskT3"
+    };
```
slarticodefast, line 40: this should not be hardcoded like this

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -20,41 +23,182 @@ public sealed class TechnologyDiskSystem : EntitySystem
...
         SubscribeLocalEvent<TechnologyDiskComponent, AfterInteractEvent>(OnAfterInteract);
         SubscribeLocalEvent<TechnologyDiskComponent, ExaminedEvent>(OnExamine);
     }
 
     private void OnMapInit(Entity<TechnologyDiskComponent> ent, ref MapInitEvent args)
+    {
+        var uid = (EntityUid)ent;
+
+        TryPickAndSetRecipe(ent);
+        TryPickAndSetSellPrice(ent, uid);
```
slarticodefast, line 58: this should not modify the static price (which like the name implies should stay static) but be a `PriceCalculationEvent` subscription

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -91,3 +251,20 @@ private void OnExamine(Entity<TechnologyDiskComponent> ent, ref ExaminedEvent ar
...
 }
+
+[Serializable, NetSerializable]
+public enum TechDiskDisciplineVisuals : byte
+{
+    Industrial,
+    Arsenal,
+    Experimental,
+    CivilianServices
+}
```
slarticodefast, line 262: Hmm, not sure if I like hardcoding these into an enum.
Ideally they would stay prototype based.

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -91,3 +251,20 @@ private void OnExamine(Entity<TechnologyDiskComponent> ent, ref ExaminedEvent ar
...
+
+[Serializable, NetSerializable]
+public enum TechDiskTierVisuals : byte
+{
+    T1,
+    T2,
+    T3
+}
```
slarticodefast, line 270: this can just be an integer

Resources/Prototypes/Entities/Objects/Specific/Research/disk.yml
```diff
@@ -45,23 +45,90 @@
...
+        t2_border:
+          True: { visible: true }
+          False: { visible: false }
+      enum.TechDiskTierVisuals.T3:
+        t3_marks:
+          True: { visible: true }
+          False: { visible: false }
+        t3_border:
+          True: { visible: true }
+          False: { visible: false }
```
slarticodefast, line 98: Having a giant list of conditions like this is very hard to expand later
Maybe a custom visualizer that just picks a prefix from the prototype is better here.

slarticodefast (changes requested):

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
metalgearsloth: Also compfactory is available via the Factory property now.

Resources/Prototypes/Entities/Objects/Specific/Research/disk.yml
```diff
@@ -45,23 +45,90 @@
...
     - state: datadisk_label
-  - type: RandomSprite
-    available:
-    - enum.DamageStateVisualLayers.Base:
-        datadisk_base: Sixteen
-  - type: TechnologyDisk
-  - type: StaticPrice
-    price: 100
+    - state: datadisk_t2_marks
+      shader: unshaded
```
EmoGarbage404, line 56: i would remove the unshaded layers as having small bit of unshaded textures without pointlights tends to not look great.

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
slarticodefast: Don't set the static price at all and completely remove this.
Just modify the price in the `PriceCalculationEvent` subscription with `TechnologyDiskComponent` depending on the tier, which is just an integer.
You won't need the component factory stuff or `StaticPriceComponent` either.

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
deltanedas: you can infact just use a string for visuals

murolem:
> I like it, but I'm not a fan of the three pixels in the top right for tier 3. I'd rather see the silver slider thing be gold.

It does look pretty good ngl.

Content Client_jcxya6O0rm

Resources/Prototypes/Entities/Objects/Specific/Research/disk.yml
murolem: Does having a pointlight make it better?

I really like these marks, they make the cards easy to spot and tiers easily distinguishable, which is the most important thing for gamba. Any chance I could keep em?

Point lights:

Looks pretty good in dark
Content Client_6JO2ErFc57

Doesn't look as good with other lighting
Content Client_uHkjjzseHl

Offset still offsets when picked up, not good
Content Client_w3x1C7B2Om

Content.Server/Research/TechnologyDisk/Systems/DiskConsoleSystem.cs
murolem: I couldn't get it to automatically validate.

Using `(EntProtoId)console.DiskPrototype` validates it on spawn which crashes the game. I could put it in try catch maybe?

Replacing `console.DiskPrototype` type from string to `EntProtoId` doesn't work because there's a `customTypeSerializer` on the attribute. I couldn't figure out what to put there, there's no `EntityPrototypeIdSerializer`.

Getting rid of `customTypeSerializer` to replace the type with `EntProtoId` doesn't work either as it doesn't do any validation on startup, only when disk is spawned thus crashing the game.

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
murolem: The event is defined in server, while the system is in shared. How to subscribe to it?

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
murolem: > Hmm, not sure if I like hardcoding these into an enum. Ideally they would stay prototype based.

You mean like separate protos for each discipline? I did that initially and I had a lot of copy pasted sprite stuff.

> you can infact just use a string for visuals

Appearance `SetData` requires enum for a key. How to use a string in this case?

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
murolem: Doesn't look like I can use integer for enum keys.

Resources/Prototypes/Entities/Objects/Specific/Research/disk.yml
murolem: Custom visualizer sounds more complicated :)

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
slarticodefast: don't use an enum at all
The tech tier is already an integer in the TechnologyPrototype

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
slarticodefast: Just move the event to Shared

Content.Server/Research/TechnologyDisk/Systems/DiskConsoleSystem.cs
slarticodefast: the customTypeSerializer can be removed
EntProtoId already handles that

murolem: Okay nevermind I've been told the validation runs as part of the linter NOT on startup.

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
murolem: How do I set proper visuals without using a enum?

slarticodefast: _appearance.SetData can take any object, not only enums

murolem: > _appearance.SetData can take any object, not only enums

I did not realize I still had to use an enum key for the second argument.

slarticodefast:
I went ahead and made a separate PR for moving the price events myself because I need this for another PR of mine as well.
Having this in a separate PR should get it merged faster and minimize conflicts for forks.
https://github.com/space-wizards/space-station-14/pull/37782

murolem:
Test fail seems unrelated

Content.Server/Research/TechnologyDisk/Components/DiskConsoleComponent.cs
```diff
@@ -16,8 +15,8 @@ public sealed partial class DiskConsoleComponent : Component
     /// <summary>
     /// The prototype of what's being printed
     /// </summary>
-    [DataField("diskPrototype", customTypeSerializer: typeof(PrototypeIdSerializer<EntityPrototype>)), ViewVariables(VVAccess.ReadWrite)]
-    public string DiskPrototype = "TechnologyDisk";
+    [DataField("diskPrototype"), ViewVariables(VVAccess.ReadWrite)]
```
slarticodefast:
```suggestion
    [DataField]
```
redundant

Content.Shared/Research/TechnologyDisk/Components/TechnologyDiskComponent.cs
```diff
@@ -9,6 +9,20 @@ namespace Content.Shared.Research.TechnologyDisk.Components;
 [AutoGenerateComponentState]
 public sealed partial class TechnologyDiskComponent : Component
 {
+    /// <summary>
+    /// A discipline to constrain the disk to.
+    /// </summary>
+    [DataField]
+    [AutoNetworkedField]
```
slarticodefast:
```suggestion
    [DataField, AutoNetworkedField]
```

Content.Shared/Research/TechnologyDisk/Components/TechnologyDiskComponent.cs
```diff
@@ -9,6 +9,20 @@ namespace Content.Shared.Research.TechnologyDisk.Components;
...
+    public ProtoId<TechDisciplinePrototype>? Discipline;
+
+    /// <summary>
+    /// A tier to constrain the disk to.
+    /// </summary>
+    [DataField]
+    [AutoNetworkedField]
```
slarticodefast:
```suggestion
    [DataField, AutoNetworkedField]
```

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -21,46 +22,121 @@ public sealed class TechnologyDiskSystem : EntitySystem
     [Dependency] private readonly SharedPopupSystem _popup = default!;
     [Dependency] private readonly SharedResearchSystem _research = default!;
     [Dependency] private readonly SharedLatheSystem _lathe = default!;
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
     [Dependency] private readonly NameModifierSystem _nameModifier = default!;
 
+    /// <summary>
+    /// Mapping of disk tiers to disk prices.
+    /// </summary>
+    private readonly Dictionary<int, int> _diskPricePerTier = new()
```
slarticodefast: Move this into a datafield on the disk component instead of hardcoding it into C#

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -21,46 +22,121 @@ public sealed class TechnologyDiskSystem : EntitySystem
...
+    private void TrySetVisuals(Entity<TechnologyDiskComponent> ent, EntityUid uid)
+    {
+        TrySetTierVisuals(ent, uid);
+        TrySetDisciplineVisuals(ent, uid);
+    }
+
+    /// <summary>
+    /// Attempts to set tier visuals based on chosen tier.
+    /// </summary>
+    private void TrySetTierVisuals(Entity<TechnologyDiskComponent> ent, EntityUid uid)
```
slarticodefast:
```suggestion
    private void TrySetTierVisuals(Entity<TechnologyDiskComponent> ent)
```
`ent` already contains the uid, just use `ent.Owner` to get it

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -95,6 +186,20 @@ private void OnExamine(Entity<TechnologyDiskComponent> ent, ref ExaminedEvent ar
         args.PushMarkup(message);
     }
 
+    private void OnPriceCalculation(Entity<TechnologyDiskComponent> ent, ref PriceCalculationEvent args)
+    {
+        if(!ent.Comp.Tier.HasValue)
+            return;
+
+        var tier = ent.Comp.Tier.Value;
```
slarticodefast:
```suggestion
        if(ent.Comp.Tier is not { } tier)
            return;
```
a little more compact syntax doing the same thing

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -21,46 +22,121 @@ public sealed class TechnologyDiskSystem : EntitySystem
...
+        if (!tier.HasValue)
+            return;
+
+        _appearance.SetData(uid, TechDiskVisuals.Tier, tier.Value);
+    }
+
+    /// <summary>
+    /// Attempts to set discipline visuals based on chosen discipline.
+    /// </summary>
+    private void TrySetDisciplineVisuals(Entity<TechnologyDiskComponent> ent, EntityUid uid)
```
slarticodefast: same as above

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -21,46 +22,121 @@ public sealed class TechnologyDiskSystem : EntitySystem
...
-        ent.Comp.Recipes.Add(_random.Pick(techs));
+        ent.Comp.Recipes.Add(bundle.recipe);
         Dirty(ent);
         _nameModifier.RefreshNameModifiers(ent.Owner);
     }
 
+    /// <summary>
+    /// Attempts to set tier and discipline visuals based on chosen tier and discipline.
+    /// </summary>
+    private void TrySetVisuals(Entity<TechnologyDiskComponent> ent, EntityUid uid)
```
slarticodefast: same

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -21,46 +22,121 @@ public sealed class TechnologyDiskSystem : EntitySystem
...
+        }
 
         //get a list of every distinct recipe in all the technologies.
-        var techs = new HashSet<ProtoId<LatheRecipePrototype>>();
+        var bundles = new HashSet<(ProtoId<LatheRecipePrototype> recipe, ProtoId<TechDisciplinePrototype> discipline)>();
         foreach (var tech in _protoMan.EnumeratePrototypes<TechnologyPrototype>())
         {
             if (tech.Tier != tier)
                 continue;
+            if(ent.Comp.Discipline != null && tech.Discipline != ent.Comp.Discipline.Value)
```
slarticodefast:
```suggestion
            if (ent.Comp.Discipline != null && tech.Discipline != ent.Comp.Discipline.Value)
```

PR #37721, opened 2025-05-22
Add "unsafe" contraband examine text

UpAndLeaves:
## About the PR
1. AlertLevelComponent and AlertLevelPrototype have been moved to Shared, to support the contraband change.
2. StationTrackerComponent now holds on to the last seen station.
3. Support has been added for "allowed alert levels" for contraband. If they are specified, the usual green "in the clear" text will be replaced with an "unsafe" text if not on the correct alert level.

## Why / Balance
The intended use is as thus (not done yet, but planned for the future):

The contents of the Armoury would be marked with a new contraband status, perhaps titled "Emergency Use Only". If not on Red, Delta, Gamma, or Epsilon alert, security personnel would be warned with the new examine text that they should not be using these guns.
It's only a warning and not red text as
a) Currently it is not an arrestable offence under space law.
b) There are numerous times when red alert SHOULD be called, but a comms console can't be accessed to change the alert level.

## Technical details
AlertLevelComponent and AlertLevelPrototype moved to shared, and references updated as required.

Contraband examine now does the following steps:
1. Does the user have the required access to hold the contraband? If no, return red text.
2. Are there any alert level values listed in the contraband? If no, return green text.
4. Add StationTrackerComponent to the user and check their last station.
5. Is the current alert level unsuitable for the contraband? If yes, return yellow text.
6. Return green text.

Slight issue: if the first time a person examines an alert-level-restricted item is not on a grid, the above steps will not return yellow text until the first time a player touches a grid.

## Media
Not in game, nothing to show

## Requirements

## Breaking changes
AlertLevelComponent + AlertLevelPrototype move.

**Changelog**
no cl cuz no player facing changes.

### DIFF
--- a/Content.Server/AlertLevel/AlertLevelDisplaySystem.cs
+++ b/Content.Server/AlertLevel/AlertLevelDisplaySystem.cs
@@ -31,7 +31,7 @@ public sealed class AlertLevelDisplaySystem : EntitySystem
         if (TryComp(uid, out AppearanceComponent? appearance))
         {
             var stationUid = _stationSystem.GetOwningStation(uid);
-            if (stationUid != null && TryComp(stationUid, out AlertLevelComponent? alert))
+            if (stationUid != null && TryComp(stationUid, out Shared.AlertLevel.AlertLevelComponent? alert))
             {
                 _appearance.SetData(uid, AlertLevelDisplay.CurrentLevel, alert.CurrentLevel, appearance);
             }
--- a/Content.Shared/Contraband/ContrabandSystem.cs
+++ b/Content.Shared/Contraband/ContrabandSystem.cs
@@ -3,11 +3,14 @@ using Content.Shared.CCVar;
 using Content.Shared.Examine;
 using Content.Shared.Localizations;
 using Content.Shared.Roles;
+using Content.Shared.Station;
 using Content.Shared.Verbs;
+using Content.Shared.AlertLevel;
 using Robust.Shared.Configuration;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Utility;
 using System.Linq;
+using Content.Shared.Station.Components;
 
 namespace Content.Shared.Contraband;
 
@@ -20,6 +23,7 @@ public sealed class ContrabandSystem : EntitySystem
     [Dependency] private readonly IPrototypeManager _proto = default!;
     [Dependency] private readonly SharedIdCardSystem _id = default!;
     [Dependency] private readonly ExamineSystemShared _examine = default!;
+    [Dependency] private readonly SharedStationSystem _station = default!;
 
     private bool _contrabandExamineEnabled;
     private bool _contrabandExamineOnlyInHudEnabled;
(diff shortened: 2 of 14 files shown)


EmoGarbage404:
the moving of the alert level stuff to shared is fine but the new unsafe contra examine level should be kept to the PR adding the feature.

Content.Shared/Contraband/ContrabandSystem.cs
```diff
@@ -101,8 +105,27 @@ private void OnDetailedExamine(EntityUid ent, ContrabandComponent component, ref
...
-            carryingMessage = Loc.GetString("contraband-examine-text-in-the-clear");
             iconTexture = "/Textures/Interface/VerbIcons/unlock-green.svg.192dpi.png";
+
+            if (!component.AllowedAlertLevels.Any())
+            {
+                carryingMessage = Loc.GetString("contraband-examine-text-in-the-clear");
+            }
+            else
+            {
+                EnsureComp<StationTrackerComponent>(args.User);
```
slarticodefast, line 116: Like mentioned on discord, not sure about the performance impact of this.
Let's best ask @ElectroJr about the metadata flags this uses.

Also this will fail if you first look at an item while in space, as the last visited station has not been tracked yet.

Content.Shared/Contraband/ContrabandSystem.cs
```diff
@@ -101,8 +105,27 @@ private void OnDetailedExamine(EntityUid ent, ContrabandComponent component, ref
...
+                var currentStation = _station.GetLastStation(args.User);
+
+                if (TryComp<AlertLevelComponent>(currentStation, out var alertLevelComponent) &&
+                    !component.AllowedAlertLevels.Contains(alertLevelComponent.CurrentLevel))
```
slarticodefast, line 120: just use the system API instead of doing this manually

Resources/Locale/en-US/contraband/contraband-severity.ftl
```diff
@@ -8,6 +8,7 @@ contraband-examine-text-Magical = [color=#b337b3]This item is highly illegal Mag
 
 contraband-examine-text-avoid-carrying-around = [color=red][italic]You probably want to avoid visibly carrying this around without a good reason.[/italic][/color]
 contraband-examine-text-in-the-clear = [color=green][italic]You should be in the clear to visibly carry this around.[/italic][/color]
+contraband-examine-text-unsafe = [color=goldenrod][italic]It is probably unsafe to continue carrying this around.[/italic][/color]
```
slarticodefast, line 11: "unsafe" is not that clear at communicating to the player that this is alert level based.

slarticodefast (changes requested):

slarticodefast:
And yeah, moving the alert level to shared is best atomized into a separate PR.

Content.Server/AlertLevel/AlertLevelDisplaySystem.cs
```diff
@@ -31,7 +31,7 @@ private void OnDisplayInit(EntityUid uid, AlertLevelDisplayComponent alertLevelD
         if (TryComp(uid, out AppearanceComponent? appearance))
         {
             var stationUid = _stationSystem.GetOwningStation(uid);
-            if (stationUid != null && TryComp(stationUid, out AlertLevelComponent? alert))
+            if (stationUid != null && TryComp(stationUid, out Shared.AlertLevel.AlertLevelComponent? alert))
```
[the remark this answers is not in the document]
Alpha-Two: will be revolved in separate PR

Content.Shared/Contraband/ContrabandSystem.cs
Alpha-Two: Failure shouldn't be too awful, as it will just revert to the old Green/Red behaviour and not take alert level into account.

Happy to take suggestions though

Content.Shared/Contraband/ContrabandSystem.cs
Alpha-Two: How exactly should this be done?

Alpha-Two:
Test fails will continue until morale is improved (#37722 is merged)

PR #37722, opened 2025-05-22
AlertLevelComponent and AlertLevelPrototype to Shared

Alpha-Two:
required for #37721 

## About the PR
Title

## Why / Balance
Shared is good 👍 

## Technical details
Move files, update references, dirty fields when required

## Media
None

## Requirements

## Breaking changes
AlertLevelComponent and AlertLevelPrototype have been moved to shared

**Changelog**
behind the scenes

### DIFF
--- a/Content.Server/AlertLevel/AlertLevelDisplaySystem.cs
+++ b/Content.Server/AlertLevel/AlertLevelDisplaySystem.cs
@@ -31,7 +31,7 @@ public sealed class AlertLevelDisplaySystem : EntitySystem
         if (TryComp(uid, out AppearanceComponent? appearance))
         {
             var stationUid = _stationSystem.GetOwningStation(uid);
-            if (stationUid != null && TryComp(stationUid, out AlertLevelComponent? alert))
+            if (stationUid != null && TryComp(stationUid, out Shared.AlertLevel.AlertLevelComponent? alert))
             {
                 _appearance.SetData(uid, AlertLevelDisplay.CurrentLevel, alert.CurrentLevel, appearance);
             }
--- a/Content.Shared/AlertLevel/AlertLevelComponent.cs
+++ b/Content.Shared/AlertLevel/AlertLevelComponent.cs
@@ -1,12 +1,13 @@
-using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype;
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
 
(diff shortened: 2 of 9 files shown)


slarticodefast (changes requested):
The same review points mentioned in https://github.com/space-wizards/space-station-14/pull/37721 still have to be addressed.

Content.Server/AlertLevel/AlertLevelDisplaySystem.cs
```diff
@@ -31,7 +31,7 @@ private void OnDisplayInit(EntityUid uid, AlertLevelDisplayComponent alertLevelD
         if (TryComp(uid, out AppearanceComponent? appearance))
         {
             var stationUid = _stationSystem.GetOwningStation(uid);
-            if (stationUid != null && TryComp(stationUid, out AlertLevelComponent? alert))
+            if (stationUid != null && TryComp(stationUid, out Shared.AlertLevel.AlertLevelComponent? alert))
```
slarticodefast, line 34:
```suggestion
            if (stationUid != null && TryComp(stationUid, out AlertLevelComponent? alert))
```

Alpha-Two:
@slarticodefast I think that's everything adjusted

slarticodefast:
Uff, the integration test never noticed this because it wasn't a datafield like it should have been 💀 
Try this
grafik

Content.Shared/AlertLevel/AlertLevelComponent.cs
```diff
@@ -0,0 +1,40 @@
...
+/// signify its alert level state.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState(fieldDeltas: true)]
+public sealed partial class AlertLevelComponent : Component
+{
+    /// <summary>
+    /// The current set of alert levels on the station.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public AlertLevelPrototype? AlertLevels;
```
slarticodefast, line 17: Like you noticed from the test, prototypes cannot be serialized, so this cannot be a datafield.
But I think keeping this as `ViewVariables` is fine in this case, as this is only used for chaching, and will be reinitialized if the map is loaded again.

Alpha-Two: Should that be VV and AutoNetworked or just VV?

slarticodefast:
I'll be closing this one in favor of the refactor that also predicts it.

PR #37723, opened 2025-05-22
Makes cargo sale computer properly indestructible

EmoGarbage404:
## About the PR

Part of the larger push to make the ATS more resistant to sabotage.

Now instead of simply being unanchorable, the sale consoles are completely indestructible and unable to be deconstructed.

Additionally removes the board from QM's locker due to not being necessary

## Why / Balance

ATS sabotage is something we want to move away from as it overly damages cargo's gameplay in a negative way. The computers are already unanchorable to prevent the most trivial sabotage, this is just going a step further.

Requiring QM to repair the computers if one is broken is better than nothing, but it still is a finite (QM only gets one board) and it enables goofy things like building a sale computer on-station (which is nonfunctional and shouldn't be a thing).

Making it properly indestructible makes it in-line with things like the pallets and the new spine on the ATS.

## Technical details

prototype rearranging.

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: The cargo sale computer is now unable to be destroyed or deconstructed.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
@@ -6,7 +6,6 @@
     - id: BoxFolderQmClipboard
     - id: CargoBountyComputerCircuitboard
     - id: CargoRequestComputerCircuitboard
-    - id: CargoSaleComputerCircuitboard
     - id: CargoShuttleConsoleCircuitboard
     - id: SalvageMagnetMachineCircuitboard
     - id: SalvageJobBoardComputerCircuitboard
--- a/Resources/Prototypes/Entities/Structures/Machines/Computers/base_structurecomputers.yml
+++ b/Resources/Prototypes/Entities/Structures/Machines/Computers/base_structurecomputers.yml
@@ -1,20 +1,13 @@
 - type: entity
-  parent: BaseStructureComputer
-  id: BaseComputer
+  id: BaseComputerInteractable
+  abstract: true
   name: computer
-  placement:
-    mode: SnapgridCenter
   components:
   - type: Animateable
   - type: MeleeSound
     soundGroups:
       Brute:
         collection: GlassSmash
-  - type: Construction
-    graph: Computer
-    node: computer
-    containers:
-      - board
   - type: Computer
   - type: ApcPowerReceiver
     powerLoad: 200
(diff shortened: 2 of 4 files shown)


Cojoke-dot:
This would affect nukies most I guess, no more hitting ats to stop cargo from getting more money for guns. Stealth ops, maybe warops, would be harder but I didn't like hitting ats anyways, it got to the point it was more procedure for nukies that's boorriiinnnggg.

Nukesthestation:
> Part of the larger push to make the ATS more resistant to sabotage.
>Now instead of simply being unanchorable, the sale consoles are completely indestructible and unable to be deconstructed.

With this pr #37560 and #36989 the ATS is basically immune to sabotage, and what feels like to me,IMO, makes cargo basically immune to sabotage (or at least, extremely difficult to)

Pronana:
A necessary change but I agree with Roomba that we should look over what is acceptable for harming the ATS.

Right now it's actually trivial to sabatoge the ATS. 

If I'm playing Nukies sabatoging the ATS is an unspoken part of any plan unless we're doing an esoteric op that requires hijacking the ATS (very rare and much riskier).

It's so easy I've never been punished for it as a Nukie. So easy I've seen rounds where it's been sabatoged by a syndie in the first 10 mins and there's no possible way to know who it was.

I will be happy to see this meta die.

Hitlinemoss:
I agree with the spirit of this change, but I think you're moving in the wrong direction. I'd rather not have the ATS be artificially invincible, since A.) antagonists can just directly target the cargo shuttle anyway (and making *that* invincible is bad for obvious reasons) and B;) it's weird from a verisimilitude standpoint for the ATS specifically to be way way more reinforced than any of the other secure areas on the station.

Instead, I think it'd be better for the ATS to just be easier to repair. In this case, I think instead of making the sale computers invincible you just make the boards something you can print from a circuit imprinter so that replacing them doesn't require the sole irreplaceable spare board from the QM's locker.

Pronana:
> Instead, I think it'd be better for the ATS to just be easier to repair. In this case, I think instead of making the sale computers invincible you just make the boards something you can print from a circuit imprinter so that replacing them doesn't require the sole irreplaceable spare board from the QM's locker.

I would normally agree, but the damage you can do to the ATS can make it virtually impossible to repair, and some things (like the buy and sell palettes) shouldn't be printable. Even if the whole ATS was repairable and you could print the parts, the time it would take and the fact its off station would effectively mean it's disabled for the whole round and evac has to be called (or cargo just doesn't get to play the game for the rest of the round anymore). 

A more mechanically interesting solution would be very nice but making the ATS easier to repair isn't going to be it unless it is trivially easy to repair and at that point, why isn't it just invincible.

Hitlinemoss:
The main problem I have with making the ATS too sabotage-proof is that it kind of just removes a strategy for antagonists outright. ATS sabotage as it is now is a bit strong, of course, but I don't think entirely removing it is great either, because then antags don't really have many other ways to prevent the crew from ordering emergency supplies.

I'll also note that you're probably overestimating how difficult it is to repair the ATS; assuming the buy/sell pallets are indestructible, all you really "need" to keep the ATS functional is those pallets, a sale computer, and a jrpacman generator as a power supply. You don't even really need atmospherics since cargo can just wear emergency EVA suits or something if the ATS is spaced. (Granted, an ATS with no atmosphere, dock, or anchor would be far more tedious to use.)

PR #35447, opened 2025-02-24
New Arsenal T3: Temperature Gun Revitalization

ArtisticRoomba:
## About the PR
Adds the Temperature Gun from SS13 as a Tier 3 Arsenal researchable technology.

This is a revitalization of PR #31762, as Flareguy has stated in discord that they are no longer continuing the PR and it was okay for someone else to re-open to continue work on it.

I've copied the rest of the PR below:

The temperature gun is an energy-based weapon that can shoot body-temperature-changing bolts (either hot or cold.) Hot has double the ammo of cold since it's not as useful due to lacking a slowdown. Cold does not do much raw damage, but will slow the enemy down.

Also moves the svalinn to T2 since there's now (hopefully) enough good weapon techs in T3. Emo has wanted this for a while, since presumably the svalinn is not very good.

On the surface, this weapon may seem a bit insane, but in testing this was pretty hard to balance. I would advise just shipping it to see how it plays out - I would rather the weapon get nerfed after the fact then a new T3 having made a bad impression on people.

## Why / Balance
We don't have enough T3s in arsenal (or really enough T3s in general.) The temperature gun should serve as a fun and powerful weapon with an interesting gimmick & extra utility (slowdown on hit.) In a way, it can be seen as a more fair version of the taser - the target can still move while slowed, and the slowdown is delayed + lasts only a few seconds if you're only capable of getting one hit in.

## Technical details
Just YAML changes.

## Media
https://github.com/user-attachments/assets/e84c55f9-fd1c-4e30-a767-71fe2c516030

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
:cl: Flareguy, ArtisticRoomba
- add: Added the Temperature Gun. It can shoot body-temperature-changing bolts. Unlock them through arsenal science today!
- remove: Experimental Battery Ammo technology is no longer available to research..

PR #37724, opened 2025-05-22
Svallin Stuff!!!!

SpaceLizard24:
## About the PR
Made the Svalinn laser pistol a T2 arsenal research for 10000 points. It is only printable in a security techfab. It costs 15 steel and 5 gold (from 20 steel and 5 gold).

## Why / Balance
I think the Svalinn is a very cool gun, but when used as a secondary. Lasers are powerful, controlling the frontline. Their main downside is opportunity cost, as you could be dealing more damage if you were carrying a long ballistic gun. However, laser pistols solve this problem by making lasers very easy to carry, so you can carry both a laser weapon and a long ballistic gun. 

The only laser pistols in the game are Captain's laser pistol (one of a kind), Retro Laser Blaster (salvage only), Advanced Laser pistol (T3 research), Makeshift Laser Pistol (to be honest idk how to even get this, is it maints loot?) and finally, Svalinn (now practical to obtain with it being a T2).

This lets the Svalinn fill the niche of laser sidearm. Now, this won't be horribly busted, as 7 damage a shot is horrible. The gun has horrible DPS. However, it has accessibility, by being easy to mass produce, having easy to reload ammo (batteries can be found and recharged nearly anywhere), it's small size (2x1), and it's not even a longarm, thus you're allowed to carry it on green alert!

Now, justifying how I said lasers are good is hard. That's like an entire game design thing. I THINK lasers are good because you can shoot people very accurately (thanks to hitscan), and you can fire on people through glass airlocks and windows. Thus, you can deal damage while your enemy with ballistic guns cannot. 

Tl;Dr: Pocket laser good. Only accessible pocket laser is Svalinn (if this gets merged.) Svalinn not too good.

This also has the benefit of security getting a cheap weapon that doesn't require either constant ammo refills or too powerful that it's dangerous to give out to civilians.

I'm also gonna be honest, I wrote this after eating about 75% of my daily sugar intake in 1 hour. So it may not be very coherent.

## Technical details
I honestly kinda blacked out, and woke up to working code. I messed with a few files. However, mostly all of the old svalinn code was still there, so I just redid that.

## Media
Svalinn

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: The Svalinn laser pistol can now be researched as a T2 Arsenal technology, and is printable in a Security Techfab.

### DIFF
--- a/Resources/Prototypes/Recipes/Lathes/Packs/security.yml
+++ b/Resources/Prototypes/Recipes/Lathes/Packs/security.yml
@@ -128,6 +128,7 @@
   - WeaponLaserCarbine
   - WeaponXrayCannon
   - WeaponTemperatureGun
+  - WeaponLaserSvalinn
 
 - type: latheRecipePack
   id: SecurityDisablers
--- a/Resources/Prototypes/Recipes/Lathes/security.yml
+++ b/Resources/Prototypes/Recipes/Lathes/security.yml
@@ -277,7 +277,7 @@
   id: WeaponLaserSvalinn
   result: WeaponLaserSvalinn
   materials:
-    Steel: 2000
+    Steel: 1500
     Gold: 500
 
 - type: latheRecipe
--- a/Resources/Prototypes/Research/arsenal.yml
+++ b/Resources/Prototypes/Research/arsenal.yml
@@ -191,6 +191,18 @@
   technologyPrerequisites:
   - SalvageWeapons
 
+- type: technology
+  id: ExpermentalBatteryAmmo
+  name: research-technology-experimental-battery-ammo
+  icon:
+    sprite: Objects/Weapons/Guns/Battery/svalinn.rsi
+    state: icon
+  discipline: Arsenal
+  tier: 2
+  cost: 10000
+  recipeUnlocks:
+  - WeaponLaserSvalinn
+
 # Tier 3
 
 - type: technology


ArtisticRoomba:
The Svallin was previously removed following #35447, citing that:
- The gun isn't very interesting and should be phased out to make room for more interesting weapons
- It was a nightmare to balance considering that everyone and their mother could get this in regular rounds + it didn't do a crazy amount of damage
- The cell based ammo didn't prove exceptionally special during the entire phase of the gun's life in the game

Could you potentially answer these and state why the Svallin would be a good fit despite its recent removal for the above reasons?

SpaceLizard24:
>The gun isn't very interesting and should be phased out to make room for more interesting weapons

I think the svalinn is very interesting, it was just nearly impossible to access and overshadowed. If science was going for T3 arsenal, 9/10 times there would be a red alert before svalinn was researched, and sec could patrol with long guns. Also, at that point, you may as well print an advanced laser pistol, which is better than the svalinn.

The svalinn was also being used as a primary weapon for head-on combat, due to it being accessable from a protolathe. The svalinn is meant to be a sidearm, accompanied by a ballistic weapon. The svalinn should only be used when firing through glass, when out of ammo for a primary weapon, or when laying cover fire without wasting real ammo. 

Now, being printed in a security techfab, it will mostly be obtained by secoffs, who already have an MK58 as their primary. 

The svalinn is also an interesting gun for the reasons I listed above. 

>It was a nightmare to balance considering that everyone and their mother could get this in regular rounds + it didn't do a crazy amount of damage

Being secfab only now means only security, people with a warrant, or criminals can obtain (besides some visitors, which already have it). The svalinn will not be given to everyone and everyone anymore, unless intended to by sec or a nefarious criminal with an emaged protolathe. 

For the low damage, like I said, people were using it wrong. Now that it will frequently be accompanied by an MK58, it will be used as a proper secondary weapon (reasons listed above).

>The cell based ammo didn't prove exceptionally special during the entire phase of the gun's life in the game

This is what I'm most unsure about. However, I think the battery ammo allows it to be a more accessable weapon. For most lasers or ballistics, you need to retreat to security (or a newly built base during nukies war) to restock on ammo and charge. For the svalinn, you won't need to. There's cell rechargers everywhere, meaning you can recharge your gun anywhere. If you used your primary weapons, you'll still need to retreat to reset those, but at least you'll always have your svalinn. 

Plus, microreactor cells give infinite ammo. And antique and high capacity cells also aren't often used, and this would be a good use. 

Thank you for coming to my ted talk.

Pronana:
The Svallin is basically a ranged welding torch in terms of damage. Restricting it to be a research for secoffs is interesting, especially as an Mk58 sidegrade but I probably wouldn't ever use it because of the truly pitiful damage. 

> Plus, microreactor cells give infinite ammo. And antique and high capacity cells also aren't often used, and this would be a good use.

Antique cells are extremely rare, and microreactor cells are expensive. High cap cells could be swapped easily allowing secoffs to have a near bottomless laser mag but that doesn't really matter because the damage is pitiful. 

I've seen antags with a bag full of highcaps attempt to try and make the Svallin work and it just, doesn't. Anyone with any other weapon will win a fight against you because your DPS output will be so comically low (14 heat DPS per second, half of a laser rifle) not even to mention that people can dodge lasers if they know how to strafe. 

Tider with a baseball bat is scarier than anyone with a svalinn to me. I do love the sprite of the svalinn and I would love to see it find a niche, but it really needs a bigger rework and a genuine purpose before it should be added back imo.

EmoGarbage404:
As it stands I don't really see a niche for the svalinn to fill. A battery powered energy gun is intentionally not something we do: it really makes it just a magazine gun with infinite ammo.

Additionally, like pronana said, the damage output has always been pitiful to a point of not being viable.

this is a gun that really needs some proper examination and balancing before coming back.

Pronana:
> As it stands I don't really see a niche for the svalinn to fill. A battery powered energy gun is intentionally not something we do: it really makes it just a magazine gun with infinite ammo.

I will say if the svalinn didn't use batteries but used something else that wasn't as replenishable and got consumed it'd be very interesting.

I saw another PR for chemically gas powered lasers that was actually neat in that while it was battery powered, it also required gas cans that you couldn't replenish.

If the Svalinn was like a compact version of that, or needed to consume reagents or solid pieces of plasma/uranium or something it could be very interesting.

ArtisticRoomba:
Unfortunately I'm going to close this PR as I agree with what Emo said and the original move we made back in #35447. It doesn't give a very good argument for the re-implementation and said re-implementation doesn't address the core problems as to why the Svallin was bad.

The Svallin needs a more interesting thing going for it in order to return. Feel free to contribute these changes (or anything else) in the future.

SpaceLizard24:
I'm gonna be honest, I feel like you guys missed my main argument. I doubt this pr is gonna get reopened, so I'll keep it short.

The svalinn is a horrible weapon _on its own_. It has pitiful dps. 
However, nearly everyone's experience with the svalinn comes from using it as a primary weapon, due to it being in a protolathe.
In a a secfab, it will never be used on its own, as all secoffs have an MK58.
Thus, the svalinn will be used with a decent pistol.
Because of this, the low DPS is okay, as you should be using the svalinn to shoot through glass, and can switch to your MK58 when close, dealing free damage with the svalinn. 

I know it's kinda hard to say "everyone's experience with the svalinn is wrong" but I genuinely believe that. Thanks to it being under arsenal T3 and accessable in a protolathe, I haven't seen a single secoff or person with another gun ever also have a svalinn. 

Tl;Dr, svalinn was being used wrong. 80% sure this PR will make svalinn be used right.

SpaceLizard24:
My idea is to use the svalinn through glass airlocks for free damage before using your MK58. That's basically 60% of why I believe the svalinn is good.
Anyways I'm planning another balance change for the svalinn, so I'll just do that instead.

superjj18:
> - The cell based ammo didn't prove exceptionally special during the entire phase of the gun's life in the game

While never special, that honestly should be standard on most energy weapons tbh

PR #37725, opened 2025-05-22
Remove capacitors and matter bins

Simyon264:
## About the PR

Removes capacitors and matter bins.

## Why / Balance

They currently have no difference from the other machine parts. So I've opted to make them all just one machine part. After discussion with Emo, we opted to use the Manipulator as its the most distinct part of them all.

Textures were left in in case they are readded with more unique features.

## Technical details

Rider search and then manual replacing. 

For the migrations I had to edit the old migrations because otherwise they would fail as the prototype they are replacing it with is no longer real.

## Requirements

## Breaking changes

The prototypes `MatterBinStockPart` and `CapacitorStockPart` have been removed.

**Changelog**

:cl: Simyon
- remove: Capacitors and Matter bins have been combined into the Manipulator.

### DIFF
--- a/Resources/migration.yml
+++ b/Resources/migration.yml
@@ -131,24 +131,19 @@ DrinkGoldschlagerGlass: DrinkGildlagerGlass
 # 2024-01-07
 ClosetBase: ClosetSteelBase
 
-# 2024-01-08 (Modified 2025-05-22)
+# 2024-01-08
 SalvagePartsT4Spawner: SalvageLootSpawner
 SalvagePartsT3Spawner: SalvageLootSpawner
 SalvagePartsT3T4Spawner: SalvageLootSpawner
 SalvagePartsT2Spawner: SalvageLootSpawner
-
-AdvancedCapacitorStockPart: MicroManipulatorStockPart
-SuperCapacitorStockPart: MicroManipulatorStockPart
-QuadraticCapacitorStockPart: MicroManipulatorStockPart
-
-NanoManipulatorStockPart: MicroManipulatorStockPart
+#SuperCapacitorStockPart: CapacitorStockPart
+#QuadraticCapacitorStockPart: CapacitorStockPart
+#NanoManipulatorStockPart: MicroManipulatorStockPart
 PicoManipulatorStockPart: MicroManipulatorStockPart
 FemtoManipulatorStockPart: MicroManipulatorStockPart
-
-AdvancedMatterBinStockPart: MicroManipulatorStockPart
-SuperMatterBinStockPart: MicroManipulatorStockPart
-BluespaceMatterBinStockPart: MicroManipulatorStockPart
-
+#AdvancedMatterBinStockPart: MatterBinStockPart
+#SuperMatterBinStockPart: MatterBinStockPart
+#BluespaceMatterBinStockPart: MatterBinStockPart
 AnsibleSubspaceStockPart: null
 FilterSubspaceStockPart: null
 AmplifierSubspaceStockPart: null


keronshb:
You forgot your media: 
https://www.youtube.com/watch?v=lkZaFLlwCHw

Simyon264:
> You forgot your media: [youtube.com/watch?v=lkZaFLlwCHw](https://www.youtube.com/watch?v=lkZaFLlwCHw)

Psshht, don't tell them.

Pronana:
Our monthly Liltenhead PR has arrived.

lzk228:
is there the point to still keep the name "manipulator"?

Pronana:
> is there the point to still keep the name "manipulator"?

Especially considering APCs and Substations use them now when it doesn't make sense they'd have moving parts a new more generic name could be nice.

iaada:
Can I suggest "modular machine part" as a name for the omni-part, similar to the "modular receiver."

superjj18:
At least now it's easier to canabalize smaller machines like battery chargers for that last machine part needed for a comms server or something

Simyon264:
> Can I suggest "modular machine part" as a name for the omni-part, similar to the "modular receiver."

Yea, that name seems good. I'll see about changing it tomorrow once I wake up.

Mixelz:
I think by removing the parts, its going to feel ostensibly a little more "bland" when making  any sort of machine since they'll all feel exactly the same to craft by using the same parts.

I do feel that despite all being made of the same material they did do a fair bit of heavy lifting in adding a small amount of variety and "flavor" to machine crafting, making the crafting components of each machine feel more unique between them. This sort of feels like a sanding down of that variety due to not wanting duplicated parts, ironically making machine recipes duplicated themselves in the process beyond the unique board.

As a follow-up, I'd want to ask why the first solution to remove them instead of making them cost different materials to make? Would that also not fix the problem of them all being the same thing while keeping the aforementioned recipe variety?

kosticia:
Ted thank you for the pr🙏

Simyon264:
> I think by removing the parts, its going to feel ostensibly a little more "bland" when making any sort of machine since they'll all feel exactly the same to craft by using the same parts.
> 
> I do feel that despite all being made of the same material they did do a fair bit of heavy lifting in adding a small amount of variety and "flavor" to machine crafting recipes, making the crafting components of each machine feel more unique between them. This sort of feels like a sanding down of that variety due to not wanting duplicated parts, ironically making machine recipes duplicated themselves in the process beyond the unique board each one takes.
> 
> As a follow-up, I'd want to ask why the first solution was to remove them instead of making them cost different materials to make? Would that also not fix the problem of them all being the same thing while keeping the aforementioned recipe variety?

While I see the point that there's flavour from different machine parts, the flavour feels artificial to me. Making them cost different amount of materials doesn't really add to the variety imo, the process is still just "examine, go to lathe, print, examine again" which is also a more generic issue of construction graphs. Reducing the amount of machine parts to just one removes the memorization required and makes it a little easier to build stuff along with reducing friction and the amount of inventory space required to store these parts.

liltenhead:
> I think by removing the parts, its going to feel ostensibly a little more "bland" when making any sort of machine since they'll all feel exactly the same to craft by using the same parts.
> 
> I do feel that despite all being made of the same material they did do a fair bit of heavy lifting in adding a small amount of variety and "flavor" to machine crafting recipes, making the crafting components of each machine feel more unique between them. This sort of feels like a sanding down of that variety due to not wanting duplicated parts, ironically making machine recipes duplicated themselves in the process beyond the unique board each one takes.
> 
> As a follow-up, I'd want to ask why the first solution was to remove them instead of making them cost different materials to make? Would that also not fix the problem of them all being the same thing while keeping the aforementioned recipe variety?

Machines already do have different parts required. Beakers, igniters, etc. You'd still need to craft the right parts that have different costs and uses, but the machine parts is really just a left over system from something that was removed awhile ago.

Pronana:
Honestly you could probably replace matter bins with beakers in a lot of these recipies. Materials already deconstruct into reagents anyways.

TPgameing:
Good riddance, although...
If we ever do get R.P.E.D (Rapid Part Exchange Devices) Along with their abilities to improve machinery, it's going to maybe be a bit interesting to balance.

ArtisticRoomba:
We nuked machine component upgrading (RPED) for good reason

HelpTheSteak:
> Good riddance, although... If we ever do get R.P.E.D (Rapid Part Exchange Devices) Along with their abilities to improve machinery, it's going to maybe be a bit interesting to balance.

We already had that but it got removed for a reason I don't remember

SpaceLizard24:
I personally don't like this change. I'm probably the minority though

It feels fun needing to get the specific parts to create something. It's more engaging with the game to print X capacitors and Y matter bins, juggling which parts go where.
It also gives a uniqueness to crafting recipes. Temperature control requires capacitors. Any lathe requires matter bins and manipulators. So on. 
I really don't think it would be a good change to just print 20 omni parts and call it a day. Seems too low effort and kinda makes machines less unique

walksanatora:
adding onto that it makes it so you have to actually remember and pay attention to the slight diffrences instead of just "waiter waiter more parts please" to fill out the requirement. so what if it is "shallow" to have more parts. it doesn't make sense to use "omni parts" that look like arms in a APC. if i were to give my 2 cents I personally would prefer to make it so that omni parts would be a slightly more expensive but the "fuck I just missed some base components" so you dont need to be exact. but if you are exact with components it would be SLIGHTLY cheaper.

DogZeroX:
This is great. I hate having to engage with machine construction and this alleviate the pain a lot, as inmersive as it would be to have different parts, it is annoying actually having to manage building the manually and all that.

PuroSlavKing:
They literally merge any PR :trollface:

metalgearsloth:
> They literally merge any PR :trollface:

If you have nothing constructive to add then please don't say anything as you have a history of doing this.

nabegator:
It is so over

PR #37734, opened 2025-05-23
Update `NodeHelpers` and `Node.GetReachableNodes` to use `Entity<MapGridComponent>`

Tayrtahn:
## About the PR

Updated the API for `NodeHelpers` so the methods take `Entity<MapGridComponent>` instead of just `MapGridComponent`, as well as a reference to `SharedMapSystem`. This enables them to use `SharedMapSystem.GetAnchoredEntities` instead of the obsolete `MapGridComponent.GetAnchoredEntities`.

The old versions of the `NodeHelpers` methods were marked as obsolete. Cleaning up the uses of the old versions required a change to the abstract method `Node.GetReachableNodes`, so that it also takes an `Entity<MapGridComponent>` instead of a `MapGridComponent`. All classes that implement `GetReachableNodes` were updated to use the new signature.

## Why / Balance

17 warnings for [PZW](https://github.com/space-wizards/space-station-14/issues/33279).
It's also just generally good to bring some old parts of the API up to more modern standards. Passing around components without references to their owner entities is funky.

## Requirements

## Breaking changes

- The abstract method `Node.GetReachableNodes` now takes an `Entity<MapGridComponent>?` instead of a `MapGridComponent?`. Classes that extend `Node` will need to be updated to override the new method signature.
- The methods of `NodeHelpers` now take an `Entity<MapGridComponent>` instead of just `MapGridComponent`, as well as a reference to `SharedMapSystem`.

### DIFF
--- a/Content.Server/Electrocution/ElectrocutionNode.cs
+++ b/Content.Server/Electrocution/ElectrocutionNode.cs
@@ -17,7 +17,7 @@ namespace Content.Server.Electrocution
         public override IEnumerable<Node> GetReachableNodes(TransformComponent xform,
             EntityQuery<NodeContainerComponent> nodeQuery,
             EntityQuery<TransformComponent> xformQuery,
-            MapGridComponent? grid,
+            Entity<MapGridComponent>? grid,
             IEntityManager entMan)
         {
             if (CableEntity == null || NodeName == null)
--- a/Content.Server/NodeContainer/EntitySystems/NodeGroupSystem.cs
+++ b/Content.Server/NodeContainer/EntitySystems/NodeGroupSystem.cs
@@ -350,12 +350,13 @@ namespace Content.Server.NodeContainer.EntitySystems
         private IEnumerable<Node> GetCompatibleNodes(Node node, EntityQuery<TransformComponent> xformQuery, EntityQuery<NodeContainerComponent> nodeQuery)
         {
             var xform = xformQuery.GetComponent(node.Owner);
-            TryComp<MapGridComponent>(xform.GridUid, out var grid);
+            if (!TryComp<MapGridComponent>(xform.GridUid, out var grid))
+                yield break;
 
             if (!node.Connectable(EntityManager, xform))
                 yield break;
 
-            foreach (var reachable in node.GetReachableNodes(xform, nodeQuery, xformQuery, grid, EntityManager))
+            foreach (var reachable in node.GetReachableNodes(xform, nodeQuery, xformQuery, (xform.GridUid.Value, grid), EntityManager))
             {
                 DebugTools.Assert(reachable != node, "GetReachableNodes() should not include self.");
 
--- a/Content.Server/NodeContainer/Nodes/NodeHelpers.cs
+++ b/Content.Server/NodeContainer/Nodes/NodeHelpers.cs
@@ -10,9 +10,10 @@ namespace Content.Server.NodeContainer.Nodes
     /// </summary>
     public static class NodeHelpers
     {
-        public static IEnumerable<Node> GetNodesInTile(EntityQuery<NodeContainerComponent> nodeQuery, MapGridComponent grid, Vector2i coords)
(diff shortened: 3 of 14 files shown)


Content.Server/NodeContainer/EntitySystems/NodeGroupSystem.cs
```diff
@@ -350,12 +350,13 @@ private List<Node> FloodFillNode(Node rootNode)
         private IEnumerable<Node> GetCompatibleNodes(Node node, EntityQuery<TransformComponent> xformQuery, EntityQuery<NodeContainerComponent> nodeQuery)
         {
             var xform = xformQuery.GetComponent(node.Owner);
-            TryComp<MapGridComponent>(xform.GridUid, out var grid);
+            if (!TryComp<MapGridComponent>(xform.GridUid, out var grid))
```
slarticodefast: I think this wasn't a guard statement on purpose in case there are nodes that are not anchored to a grid.
The whole code seems to be made to be more flexible than just with pipes and cables in mind.

slarticodefast (commented):
Look at that, I just needed the grid uid for a new `Node` type I was writing and was about to do the same changes as here.

Content.Server/Electrocution/ElectrocutionNode.cs
```diff
@@ -14,10 +14,10 @@ public sealed partial class ElectrocutionNode : Node
         [DataField("node")]
         public string? NodeName;
 
-        public override IEnumerable<Node> GetReachableNodes(TransformComponent xform,
+        public override IEnumerable<Node> GetReachableNodes(Entity<TransformComponent> xform,
```
TemporalOroboros: I don't know if this is actually forbidden by the style guide, but it just looks wrong. If you would be willing to go through and move all the params like this to their own line I would appreciate that.

Content.Server/NodeContainer/Nodes/PipeNode.cs
```diff
@@ -202,10 +203,10 @@ public override IEnumerable<Node> GetReachableNodes(TransformComponent xform,
         /// <summary>
         ///     Gets the pipes that can connect to us from entities on the tile or adjacent in a direction.
         /// </summary>
-        private IEnumerable<PipeNode> LinkableNodesInDirection(Vector2i pos, PipeDirection pipeDir, MapGridComponent grid,
-            EntityQuery<NodeContainerComponent> nodeQuery)
+        private IEnumerable<PipeNode> LinkableNodesInDirection(Vector2i pos, PipeDirection pipeDir, Entity<MapGridComponent> grid,
+            EntityQuery<NodeContainerComponent> nodeQuery, SharedMapSystem mapSystem)
```
TemporalOroboros: Please give each of the params their own line if they're going to be split up.

TemporalOroboros (changes requested):
Looks good! If you would be willing to fix the merge conflict and fix some style nits I will approve and merge.

I would have done it myself, but the 'allow maintainers to make edits' option is broken for me and it won't let me push to your branch.

Content.Server/NodeContainer/Nodes/NodeHelpers.cs
```diff
@@ -42,29 +49,50 @@ public static IEnumerable<Node> GetNodesInTile(EntityQuery<NodeContainerComponen
...
+            foreach (var uid in mapSystem.GetAnchoredEntities(grid, coords + (1, 0)))
                 yield return (Direction.East, uid);
 
-            foreach (var uid in grid.GetAnchoredEntities(coords + (-1, 0)))
+            foreach (var uid in mapSystem.GetAnchoredEntities(grid, coords + (-1, 0)))
                 yield return (Direction.West, uid);
         }
+
+        [Obsolete("Use the overload that passes in Entity<MapGridComponent> and SharedMapSystem")]
+        [SuppressMessage("ReSharper", "EnforceForeachStatementBraces")]
```
Tayrtahn:
```suggestion
```

PR #37735, opened 2025-05-23
Biome rework

metalgearsloth:
Reworks biome code to use dungeons and is pretty much entirely rewritten. Advantage is that we get to use cpujob now and makes biomes a lot more flexible and reduces code burden of maintenance.

## Breaking changes
- Biomes have been entirely reworked and need porting to dungeons via the SampleEntity / SampleTile / SampeDecal layers. Standard biomes should be easy to port but if you're using marker layers that will take a bit more work to get off the ground.
- Dungeons have had a bunch of changes around inheritance for reserved tiles across layers to make them more flexible and also namespace changes to standardise them.
- Fixed AllTiles not including CorridorExteriorTiles.

Review:
- I checked asteroids + scrap and they seemed okay.
- I flew around and it seems okayTM.
- Ore layers always persist as they modify entities live so not entirely sure how to solve this one.
- Expeds are probably broken in some way but we want to remove them anyway so eh?
- It should be a lot easier for admins to add stuff dynamically.

Random notes:
- Adding layers at runtime should be a lot easier as it's just a dungeonconfig (entirely yaml) + console command. Using the layer string as a hash for the seed also helps with adding repeat layers (e.g. to spam ores or the likes).
- Loading in large chunks takes a long-ass time so need to probably optimise it or potentially re-think it in future. At least biomes using dungeons means we can focus effort + QOL on one area and not two.
- Expeds are probably broken in some way but we intend to remove them, I just didn't do it in this PR.


EmoGarbage404:
the server is made for testing

Content.Server/Gateway/Systems/GatewayGeneratorSystem.cs
```diff
@@ -111,7 +110,7 @@ private void GenerateDestination(EntityUid uid, GatewayGeneratorComponent? gener
         };
         AddComp(mapUid, restricted);
 
-        _biome.EnsurePlanet(mapUid, _protoManager.Index<BiomeTemplatePrototype>("Continental"), seed);
+        _biome.EnsurePlanet(mapUid, _protoManager.Index("BiomeContinental"), seed);
```
slarticodefast, line 113: why the hell is this a hardcoded, non-validated string
This is currently crashing vulture
`UnknownPrototypeException: Unknown EntityPrototype prototype: BiomeContinental`

PR #37736, opened 2025-05-23
Cleanup warning in `EventHorizonSystem`

Tayrtahn:
## About the PR

Fixes 1 warning in `EventHorizonSystem.cs`.

## Why / Balance

## Technical details

- `CanConsumeTile` was changed to take `Entity<MapGridComponent>` instead of `MapGridComponent` as a parameter.
- The old version was marked obsolete.
- Now that we have an EntityUid for the grid, we can use `SharedMapSystem.GetAnchoredEntities` instead of `MapGridComponent.GetAnchoredEntities`, getting rid of a warning.

## Media

Singulo still doing its thing:
Screenshot 2025-05-22 at 9 58 58 PM

## Requirements

### DIFF
--- a/Content.Server/Singularity/EntitySystems/EventHorizonSystem.cs
+++ b/Content.Server/Singularity/EntitySystems/EventHorizonSystem.cs
@@ -270,7 +270,7 @@ public sealed class EventHorizonSystem : SharedEventHorizonSystem
         var toConsume = new List<(Vector2i, Tile)>();
         foreach (var tile in tiles)
         {
-            if (CanConsumeTile(hungry, tile, grid, eventHorizon))
+            if (CanConsumeTile(hungry, tile, (gridId, grid), eventHorizon))
                 toConsume.Add((tile.GridIndices, Tile.Empty));
         }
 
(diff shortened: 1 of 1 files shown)


Content.Server/Singularity/EntitySystems/EventHorizonSystem.cs
```diff
@@ -284,16 +284,23 @@ public int AttemptConsumeTiles(EntityUid hungry, IEnumerable<TileRef> tiles, Ent
     /// Checks whether an event horizon can consume a given tile.
     /// This is only possible if it can also consume all entities anchored to the tile.
     /// </summary>
-    public bool CanConsumeTile(EntityUid hungry, TileRef tile, MapGridComponent grid, EventHorizonComponent eventHorizon)
+    public bool CanConsumeTile(EntityUid hungry, TileRef tile, Entity<MapGridComponent> grid, EventHorizonComponent eventHorizon)
```
slarticodefast, line 287:
```suggestion
    public bool CanConsumeTile(Entity<EventHorizonComponent> hungry, TileRef tile, Entity<MapGridComponent> grid)
```
might as well make it even more future proof

slarticodefast (commented):
Well, this gets rid of one warning, but the `grid.Owner` introduces a new one.
But I guess we can then just remove the obsolete methode in the future.
Although the whole system might use some cleanup.

Tayrtahn:
Ah, but it's not a warning if it's inside an obsolete method!

slarticodefast (approved):

PR #37745, opened 2025-05-23
Add overkill thresholds to more objects

B-Kirill:
## About the PR
Adds overkill thresholds to more objects (don't know what else to call them). In the game, most objects already have them, but there are a few that don't.

If anyone doesn't know, this is a special trigger in which an object is destroyed, but does not leave items after destruction (does not affect the contents of containers). 
Here is an APC as an example:
```
  - type: Destructible
    thresholds:
    - trigger: <--- This is the overkill threshold
        !type:DamageTrigger
        damage: 200
      behaviors:
      - !type:DoActsBehavior
        acts: [ "Destruction" ]
      - !type:PlaySoundBehavior
        sound:
          collection: MetalGlassBreak
    - trigger: <--- This is the regular threshold
        !type:DamageTrigger
        damage: 50
      behaviors:
      - !type:SpawnEntitiesBehavior
        spawn:
          SheetSteel1: <--- This one creates items after destruction
            min: 1
            max: 1
      - !type:DoActsBehavior
        acts: [ "Destruction" ]
      - !type:PlaySoundBehavior
        sound:
          collection: MetalGlassBreak
```

Vending machines have also been slightly modified in this regard. Now they throw away goods only when they are damaged, not completely broken, so that the goods of this vending machine will not appear out of thin air...

## Why / Balance
It is a kind of optimization

## Requirements

**Changelog**
Nope

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Decoration/curtains.yml
+++ b/Resources/Prototypes/Entities/Structures/Decoration/curtains.yml
@@ -39,6 +39,12 @@
     damageModifierSet: Wood
   - type: Destructible
     thresholds:
+    - trigger: # Excess damage, don't spawn entities
+        !type:DamageTrigger
+        damage: 25
+      behaviors:
+      - !type:DoActsBehavior
+        acts: ["Destruction"]
     - trigger:
         !type:DamageTrigger
         damage: 5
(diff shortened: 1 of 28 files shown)


Resources/Prototypes/Entities/Structures/Decoration/curtains.yml
```diff
@@ -39,6 +39,12 @@
     damageModifierSet: Wood
   - type: Destructible
     thresholds:
+    - trigger: # Excess damage, don't spawn entities
+        !type:DamageTrigger
+        damage: 25
```
EmoGarbage404, line 44: should be higher since 25 isn't really that hard to hit with a normal weapon

EmoGarbage404 (approved):

PR #37755, opened 2025-05-23
Add system for invisible mapping tiles/entities

PJB3005:
I wanted to map something that would need invisible walls/floors. So here's the tools for it.

Tile prototypes can now have an "invisibleSprite" specified. This sprite will be used when the mapping "show invisible" overlay is active.

Added a new MapInvisible component that toggles the visibility of the entity' Sprite based on the same.

Visibility of these is toggled with a show_invisible command. Mapping client setup enables this by default.

Added an "invisible floor" and 'invisible wall" entity for your mapper pleasures. Gave the mime's invisible wall a clear suffix to distinguish it.

## Media

https://github.com/user-attachments/assets/4de22804-1648-4dd1-a377-a739e4195717

## Breaking changes
None


Resources/Locale/en-US/mapping/commands.ftl
```diff
@@ -0,0 +1,4 @@
+﻿cmd-show_invisible-desc = Toggle visility of normally-invisible map entities, such as invisible walls.
+cmd-show_invisible-help = Usage: show_invisible
```
Pok27, line 2:
```suggestion
cmd-show_invisible-help = Usage: {$command}
```

PR #37760, opened 2025-05-23
Limit survivor event spell purchases to after 30 minutes

minus1over12:
## About the PR

The 2 wizard survivor event spells (Summon Magic & Summon Guns) can now only be purchased starting 30 minutes in, in line with what the syndicate bomb requires.
## Why / Balance

Everyone getting free agent status pretty much signals what is going to end up ending the round. While this is fine, it shouldn't happen 5 minutes into the round. 30 minutes was chosen to match the timer on the syndicate bomb.
## Technical details

The time restriction on syndicate bombs was copied to the event spells.
## Media

Screenshot showing the survivor event spells having a delay

## Requirements

## Breaking changes

**Changelog**

:cl: WarPigeon

- tweak: Summon Magic and Summon Guns can now only be purchased from the spellbook after the round is 30 minutes in.

### DIFF
--- a/Resources/Prototypes/Catalog/spellbook_catalog.yml
+++ b/Resources/Prototypes/Catalog/spellbook_catalog.yml
@@ -305,6 +305,7 @@
   - !type:ListingLimitedStockCondition
     stock: 1
   disableRefund: true
+  restockTime: 1800
 
 - type: listing
   id: SpellbookEventSummonMagic
@@ -319,6 +320,7 @@
   - !type:ListingLimitedStockCondition
     stock: 1
   disableRefund: true
+  restockTime: 1800
 
 # Upgrades
 - type: listing


ArtisticRoomba:
The alternative solution I heard to this problem was to block using the spell until the Wizard was on station.

MozarteanChaos:
> > The alternative solution I heard to this problem was to block using the spell until the Wizard was on station.
> 
> I like this more, as this won't lead to a round being essentially a greenshift for 30 mins until the wizard can buy one or two spells

This PR only limits the survivor event spells, which there are two of - the wizard can still use their wizcoins on any other spells. If they choose to wait half an hour before setting foot on the station, just so they can buy an event spell before doing so... well, that sounds very boring for the wizard, so I'm not entirely sure why they'd waste their time doing that?

SlimmSlamm:
> just so they can buy an event spell before doing so... well, that sounds very boring for the wizard, so I'm not entirely sure why they'd waste their time doing that?

What would the point in even having the spell then? It's a fun spell people want to use and if the logic here is that if they're waiting to use the spell for 30 minutes then why have it at all

PR #37769, opened 2025-05-24
SVALINN GOOD

SpaceLizard24:
## About the PR
Made the Svalinn be able to fire a 4 round burst. However, the Svalinn now becomes inaccurate after prolonged rapid fire, or overheating.
The 4 round burst is twice as fast as semi auto. Semi auto will still overheat, however it will be significantly more controlled than the burst mode.

Note that I'm open to suggestions, especially the description. Although I tested some numbers (using the scientific method of guess and check) and I feel that these are good. You can still get a lucky hit at moderate inaccuracy, but it's very unlikely to get a mid range hit at high inaccuracy.

## Why / Balance
The Svalinn is a boring gun (according to some people), so this gives it a coat of paint. 
The Svalinn is better when you're moving towards the enemy (or when they're approaching you!) as the accuracy becomes less of a penalty, and the faster RoF becomes more of a boon.
The Svalinn can still be used in long range combat, but not as well. 
Imagine the Svalinn like a flintlock pistol for a pirate. You fire one shot, or one battery, and then move in closer to do proper combat against the enemy with your cutlass, rifle, shotgun. Or, you can draw another flintlock or Svalinn and keep at it... 

Also if this gets merged you should totally also merge #37724
## Technical details

1 minAngle. For reference, this is the same as a single shot with the WT550
20 maxAngle, meaning the gun will be mostly luck at far range during rapid fire.
2.2 angleIncrease, meaning the gun won't gain too much inaccuracy per shot. It's the rapid shots that do in the accuracy.
4 angleDecay, meaning this gun will dissipate it's max 20 accuracy in 4.75 seconds of not shooting. 
2 fireRate
8 burstFireRate, meaning the gun fires four times as fast when under burst mode. It fires its entire 4 round burst in half a second.
4 shotsPerBurst. Yeah, self explanatory. 
0.6 burstCooldown, meaning once you click the mouse, you'll need to wait 0.6 seconds before firing another burst. 

I mostly ripped off the Drozid and WT550.

Also, the Svalinn used to be an item. Now, it's a gun. I don't know what this changes honestly, but I don't think it breaks anything.
Note I am not a programmer. I don't think I did anything wrong, but I could have.

## Media

## Breaking changes

**Changelog**

:cl:
- tweak: The Svalinn now has a rapid-fire 4 round burst mode. 
- tweak: The Svalinn now becomes inaccurate under prolonged fire.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
@@ -105,7 +105,7 @@
   name: svalinn laser pistol
   parent: [ BaseWeaponPowerCellSmall, BaseSecurityContraband ]
   id: WeaponLaserSvalinn
-  description: A cheap and widely used laser pistol.
+  description: A cheap and widely used laser pistol. Prone to overheating when rapidly fired, losing accuracy.
   components:
   - type: Sprite
     sprite: Objects/Weapons/Guns/Battery/svalinn.rsi
@@ -115,8 +115,19 @@
     - state: mag-unshaded-4
       map: ["enum.GunVisualLayers.MagUnshaded"]
       shader: unshaded
-  - type: Item
+  - type: Gun
+    minAngle: 1
+    maxAngle: 20
+    angleIncrease: 2.2
+    angleDecay: 4
     sprite: Objects/Weapons/Guns/Battery/svalinn.rsi
+    fireRate: 2 
+    burstFireRate: 8
+    availableModes:
+    - Burst
+    - SemiAuto
+    shotsPerBurst: 4
+    burstCooldown: 0.6
   - type: MagazineVisuals
     magState: mag
     steps: 5


ArtisticRoomba:
please add media to this pull request

SpaceLizard24:
i deleted an arrow by mistake. fixed

RedBookcase:
> Also, the Svalinn used to be an item. Now, it's a gun. I don't know what this changes honestly, but I don't think it breaks anything.
Note I am not a programmer. I don't think I did anything wrong, but I could have.

These are not mutually exclusive. The Item type is important as it helps define sprites for the item. You should definitely put it back. (Due to parenting inheritance, it's currently inheriting it's item type from the item it's parented from)

RedBookcase:
> i will fix it tomorrow, after making 100% sure that it's necessary to have the gun tag thing. I don't think it is, but i just wanna double check before i break even more things.

For what you've added having the type: Gun is necessary as you've modified the weapon to be different from the thing it's parented from.

SpaceLizard24:
okay, this should work. Anyways, i have three concerns with this change.
one, if you get an antique power cell, the svalinn is actually really really good? like, 24 damage every 0.6 seconds at close range through glass, and hitscan is actually pretty good? Although you do have to be in close range, so yeah. It'll probably be fine
two, this being handed to heads with microreactor cells. like, infinite lasers are good, period. it's why the captain's pistol is good. nukies will weep. Although this will require good coordination from science and security, so it's unlikely to happen. Plus, more people with self recharging lasers means more opportunities for antags to take them. 
three, the description. can't think of a better one, though.

SpaceLizard24:
okay, cool. what's the trailing whitespace check?

Cojoke-dot:
I believe this is a change that requires an in game showcase, maybe a video or something to show how fast you can take down a stationary unarmored. At least something to show the weapon firing

RedBookcase:
> dummies didn't display damage for some reason. i tested it with other weapons, nothing

It's a known bug with the shooting targets, not related to anything on your end.

Resources/Prototypes/Entities/Objects/Weapons/Guns/Battery/battery_guns.yml
```diff
@@ -117,6 +117,18 @@
...
+  - type: Gun
+    minAngle: 1
+    maxAngle: 20
+    angleIncrease: 2.2
+    angleDecay: 4
+    fireRate: 2
+    burstFireRate: 8
+    availableModes:
+    - Burst
+    - SemiAuto
```
beck-thompson, line 129: I personally think it would be more interesting if it was burst only. Makes it a lot more unique

RedBookcase: > Idk if this is good or bad, but I think it being a 2 for 1 weapon will give it a chance to be useful.

In my experience weapons with multiple fire modes are rarely used off of their default/best mode. Nobody ever uses burst fire, despite smgs like the C20-r having it, because full auto/single shot is just preferred. I think making the Svalinn burst only would be more interesting because people would actually have to interact with the "Inaccurate and unweildy" part of the gun.

SpaceLizard24: So. While you guys did convince me over that, when I tried changing it, the game would crash when I saw a svalinn. 
Basically, no semi auto mode, the game crashes. I'm not sure how to fix it.

RedBookcase: Make sure you add the line `selectedMode: Burst` above the availableModes line. You're parenting off of the base battery weapons which probably have semi auto as the selected mode so you need to specify it in the Svalinn prototype

beck-thompson: when I tested it yeah that seems to be the issue so do what redbookcase said

SpaceLizard24:
tada

SpaceLizard24:
You know, I really like the svalinn. It's funny.

SpaceLizard24:
My pr is still awesome so if you could merge it or something that'll be great. Haha just kidding. Unless?

RedBookcase:
> My pr is still awesome so if you could merge it or something that'll be great. Haha just kidding. Unless?

Keep in mind you still need to propose and PR a method to acquire/produce the Svalinn as well as just updating the weapon itself.

RedBookcase:
Also, should probably change the description to simply note that it's a burst fire weapon and remove the bit about overheating.

SpaceLizard24:
> Keep in mind you still need to propose and PR a method to acquire/produce the Svalinn as well as just updating the weapon itself.

#37724 

>Also, should probably change the description to simply note that it's a burst fire weapon and remove the bit about overheating.

Yeah. I'll change it probably by the time you read this.

SpaceLizard24:
Hi, I was wondering if this pr will get triaged. It's been a month, and I'm curious if I should just close the pr. Thanks!

SpaceLizard24:
Well, that answers that. Thanks!

SpaceLizard24:
hi its been three months. please. i miss the svalinn...

SlamBamActionman:
Hello! Sorry for the delay in reviewing, we've had quite the extensive backlog with PRs.

I've tested this PR ingame now and I am of mixed feelings. On one hand the Svalinn is kind of a peashooter in that using it feels underwhelming (can't even fully crit someone on a full medium battery) and that to make it feel somewhat viable you either need to load up on large spare batteries or be a late-stage Scientist or lucky Salvager with a recharging power cell. Which is all going to be unlikely since the most common place you'll see Svalinns today is from escape pod ghost roles. The Svalinn *should* be a pretty bad gun but perhaps it's a bit too bad in its current state. 

As for the changes in this PR, I can't say I am vibing with them. I understand the intent but it does some stuff I don't like.

The damage it deals is comparable to a laser carbine in DPS (25.5 vs 28), which is a bump up from its previous 14. This is tempered by the accuracy penalty for rapid firing. Based on some calculations and testing it seems like you need to wait an additional 1.35s to the 0.6s forced delay between shots for the accuracy penalty to disappear. This means to fire the Svalinn with full accuracy (assuming you even know how long to wait), the DPS drops to 10.9 which means it's effectively worse for sustained accurate fire. 

So that means to get full use of the gun you have to use it at mid-to-close range, which is generally where lasers underperform since the low damage is meant to balance out the window penetration and (normally) high accuracy. 

Ultimately I don't think the Svalinn works well in that context and while I agree the four-round burst is kind of cool, I also don't see it functioning here. Maybe it could end up feeling a bit nicer with just a two-round burst and no accuracy penalty, or with some other gimmick. As a member of the Combat workgroup I'm going to be closing this PR but don't take it as changes to the Svalinn being undesirable, just that this implementation isn't what the gun needs.

SpaceLizard24:
Hi! As the #1 Svalinn laser pistol fan, I wanted to explain my logic of this PR and consider what should be done for a new change.
Skip down if you don't want to hear me rant about my favourite gun. 

So, the theory wasn't for this gun to be used as a substitute laser carbine. The intention was the Svalinn to be 3 things.
1. A way to deal free damage when approaching a glass airlock/window
2. An complementary sidearm to the MK58
3. A weapon useful for civilians, while not too dangerous to trouble security.

1. The concept with the four round burst and poor accuracy was to force you to close the gap if you want to continue shooting with decent accuracy. If you do this when fighting a conventional weapon (C-20R, Lecter, etc), you'll lose. Simply put, the DPS isn't enough. So, the solution is to use it when approaching glass airlocks. Fire one or two bursts, and then switch to your main weapon that can deal damage.
2. The MK58 is a decent weapon, but suffers from low ammo. The Svalinn could complement the MK58, either being a powerful finishing weapon (as you don't have to worry about the accuracy if you're firing only one burst, and 28dps is decent, especially with hitscan lasers that are hard to dodge), a backup gun if you need it, or a way to shoot through glass for free.
3. The Svalinn isn't really good at killing people, because by the third burst you're missing a lot of shots. However, this means it's pretty good at maiming people. If a civilian carries it and they get attacked, they can injure the attacker. They probably won't win the fight if the attacker is decently armed, but it'll at least make attackers think twice before trying to pick a fight with a Svalinn holder. On the other hand, a security officer can decently rush down a Svalinn user. They'll get injured, but they're station aligned, so they have no issues with going to medical after. Plus, civilians don't need to find a recharger after a shooting, they just need a cell recharger.

The above is of course purely theoretical, I haven't done any testing outside of a vacuum, and can't prove that any of this would actually happen. Assuming my other PR to readd the Svalinn would have been merged alongside this, the most realistic outcome would be science simply doesn't research the gun, or security doesn't print and distribute it.

**RANT OVER**

Now for the actual important part, what I want for a new Svalinn PR I will make.

I want the gun to be both strong enough to make attackers think twice before picking a fight, to be a useful utility tool, while also being safe enough to hand to civilians. I also want the gun to be easily rechargeable to civilians, which is already a mechanic. 
I don't want the gun to kill. I want the gun to damage and wound, while still being useful. The simple solution would be giving it low ammo, but because it uses batteries, that's not an option. The reduced accuracy was the first thing I thought of to make it reasonable. Maybe it could deal stamina damage as well?

Is all of that reasonable, and if it works well, would that PR in theory be merged?

SlamBamActionman:
I think I am not able to connect the intended purpose with who the gun is available to and why. 

You describe the intended purpose to be "sidearm to close the distance to an airlock". I sorta see how the mechanics are meant to tie in there, but I don't think I see the purpose of the tactic. 

If you're trying to push towards an airlock, it's likely because you have a stronger ballistic weapon that you want to get in and use. With the sight distances in the game, that means you'll fire off at most one-two shots before needing to put away the Svalinn and change over to your primary (assuming the target is right on the other side of the airlock which isn't guaranteed). That's a fair bit of juggling for a pretty small amount of damage. 

But at what point would a civilian want to do that? If they are forced into combat with a laser gun they'd be more likely to stay put behind the airlock for cover to eliminate the threat or retreat while shooting. And this Svalinn doesn't really work well in that scenario. Security already has better weapons for such a scenario and will generally fight threats where their ballistics will be strong enough to dispatch of the threat on the other side quickly, and if not the Svalinn's not gonna be a good source of damage anyways. 

As-is, having a gun that only under the best of circumstances can be used as an (even more than before) underwhelming poking tool or as a situational low-damage starter just doesn't mesh with me for civilian use and is outclassed by other Security weapons.

PR #37775, opened 2025-05-24
allow jetpacks to stay enabled when on grids and lightly rework jetpack system

### LINKED ISSUE #34094: Using two jetpacks makes you sparkle
## Description

as in when you use two jetpacks and go on a grid, the jetpack effect from one of the jets never goes away 
(and the jetpack doesn't go back to looking like it's turned off but that's less funny)

**Reproduction**

1. Use jet
2. Use different jet
e. Go on land 
ceil(pi). Sparkly little shithead syndrome 

**Screenshots**

**Additional context**

LaCumbiaDelCoronavirus:
# _**This does NOT allow you to fly on grids with a jetpack!**_

## About the PR
Jetpacks will be able to be turned on/off when you're on a grid, letting you turn one on while still on a grid and start flying when you go into space; hitting a grid while jetpacking won't turn it off anymore. **It doesn't use fuel when on a grid!!**

## Why / Balance
It's annoying to hit a tile while flying in space and get stopped immediately, and its very detrimental in some combat situations

## Technical details
Whether a jetpack is actually flying and whether it's turned on is now separate:
- Flying is flying, and you can only fly when the jetpack is on and you're allowed to fly (i.e., not on a grid regardless of if it has gravity)
- People holding jetpacks that are on now get JetpackUserComponent, rather than only when they're flying around in space with it.

Alot of event-subscribed methods have been changed to use `Entity<IComp>, Event` instead of `EntityUid, IComp, ref Event`.
Alot of methods were also renamed, but only private ones so no breaking changes. More public methods have been added to accommodate for these changes.

Fixes #34094 

## Media
https://github.com/user-attachments/assets/0bfc9d65-de07-4a5b-9bc2-e2f7653f3c87

(initial spamming of jetpack particles when you leave a grid is a problem fixed in this PR, media not updated yet)

## Breaking changes
`SharedJetpackSystem.SetEnabled()` has been changed to take `Entity<JetpackComponent>`, instead of `EntityUid, JetpackComponent`.

## Requirements

## Actual Breaking Changes
`SharedJetpackSystem.CanEnable` now takes `Entity<JetpackComponent>` instead of `EntityUid, JetpackComponent`. ActiveJetpackComponent.LastCoordinates is now nullable and defaults to null.

**Changelog**
:cl:
- tweak: Jetpacks can now be turned on/off while you're on a grid, however don't propel you or use fuel until you are in space.
- tweak: Jetpacks now don't automatically turn off when you get on a grid.
- fix: You should no longer be able to have more than 1 jetpack on at once.

### DIFF
--- a/Content.Client/Movement/Systems/JetpackSystem.cs
+++ b/Content.Client/Movement/Systems/JetpackSystem.cs
@@ -59,6 +59,7 @@ public sealed class JetpackSystem : SharedJetpackSystem
                 continue;
             }
 
+            // Only spawn particles if we're far-enough from the last place at which we spawned particles, and a long-enough timespan has passed since then.
             if (_transform.InRange(transform.Coordinates, comp.LastCoordinates.Value, comp.MaxDistance) && _timing.CurTime < comp.TargetTime)
                 continue;
 


ScarKy0:
The mispredict when stepping off grid goes hard

LaCumbiaDelCoronavirus:
> The mispredict when stepping off grid goes hard

:suspect: 
<sub><sup>(not me i think)</sup></sub>

slarticodefast:
Yeah, that mispredict needs to be fixed first.

VerinSenpai:
You beautiful sob

Pumkin69:
thank you so much this was my biggest gripe with jetpacks

Content.Client/Movement/Systems/JetpackSystem.cs
```diff
@@ -46,26 +45,33 @@ public override void Update(float frameTime)
...
         // TODO: Please don't copy-paste this I beg
         // make a generic particle emitter system / actual particles instead.
-        var query = EntityQueryEnumerator<ActiveJetpackComponent, TransformComponent>();
+        var query = EntityQueryEnumerator<ActiveJetpackComponent>();
 
-        while (query.MoveNext(out var uid, out var comp, out var xform))
+        while (query.MoveNext(out var uid, out var comp))
         {
-            if (_transform.InRange(xform.Coordinates, comp.LastCoordinates, comp.MaxDistance))
+            var transform = Transform(uid);
```
perryprog: Why not include Transform in the query as before, if you're immediately going to grab it anyway?

Content.Client/Movement/Systems/JetpackSystem.cs
```diff
@@ -46,26 +45,33 @@ public override void Update(float frameTime)
...
+            if (comp.LastCoordinates is not { } lastCoords)
+            {
+                comp.LastCoordinates = currentCoords;
+                continue;
+            }
+            else if (_transform.InRange(transform.Coordinates, lastCoords, comp.MaxDistance))
             {
                 if (_timing.CurTime < comp.TargetTime)
                     continue;
             }
```
perryprog:
```suggestion
            }

            if (_transform.InRange(transform.Coordinates, lastCoords, comp.MaxDistance)
                && _timing.CurTime < comp.TargetTime)
                continue;
```

Content.Shared/Movement/Components/JetpackComponent.cs
```diff
@@ -7,6 +7,12 @@ namespace Content.Shared.Movement.Components;
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
 public sealed partial class JetpackComponent : Component
 {
+    /// <summary>
+    /// Whether the jetpack is turned on, but not whether it's actually flying!
+    /// </summary>
+    [AutoNetworkedField]
+    public bool Enabled = false;
+
```
perryprog: Not a fan of the name. I don't love any of these either, but maybe `PrimedToFly` or `AutoEnable` or something? Not sure.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
         var jetpackQuery = GetEntityQuery<JetpackComponent>();
 
-        var query = EntityQueryEnumerator<JetpackUserComponent, TransformComponent>();
-        while (query.MoveNext(out var uid, out var user, out var transform))
+        var query = EntityQueryEnumerator<JetpackUserComponent>();
+        while (query.MoveNext(out var uid, out var user))
         {
-            if (transform.GridUid == gridUid && ev.HasGravity &&
-                jetpackQuery.TryGetComponent(user.Jetpack, out var jetpack))
+            var transform = Transform(uid);
```
perryprog, line 76: No need to remove transform from the query.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
 
-    private void OnMapInit(EntityUid uid, JetpackComponent component, MapInitEvent args)
+    private void OnMapInit(Entity<JetpackComponent> jetpack, ref MapInitEvent args)
     {
-        _actionContainer.EnsureAction(uid, ref component.ToggleActionEntity, component.ToggleAction);
-        Dirty(uid, component);
+        var (_, jetpackComponent) = jetpack;
+
+        _actionContainer.EnsureAction(jetpack.Owner, ref jetpackComponent.ToggleActionEntity, jetpackComponent.ToggleAction);
+        Dirty(jetpack);
```
perryprog: Nit, but I'd recommend

```suggestion
    private void OnMapInit(Entity<JetpackComponent> ent, ref MapInitEvent args)
    {
        _actionContainer.EnsureAction(ent, ref ent.Comp.ToggleActionEntity, ent.Comp.ToggleAction);
        Dirty(ent);
```

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
+            var transform = Transform(uid);
+
+            var jetpackUid = user.Jetpack;
+            if (transform.GridUid == gridUid && jetpackQuery.TryGetComponent(jetpackUid, out var jetpackComponent))
             {
-                _popup.PopupClient(Loc.GetString("jetpack-to-grid"), uid, uid);
+                var jetpack = new Entity<JetpackComponent>(jetpackUid, jetpackComponent);
+
+                var canFly = CanFlyOnGrid(gridUid);
+                SetEnabled(jetpack, jetpackComponent.Enabled, flyIfEnabled: canFly, user: uid);
```
perryprog: Nitpick, but I think this is fine if you inline most of this. The main thing is that there's an implicit operator for tuple of EntityUid and a component into `Entity<Comp>`, though, so you don't need to do the `new Entity<...>` stuff.

```c#
if (transform.GridUid == gridUid && jetpackQuery.TryGetComponent(user.Jetpack, out var jetpackComponent))
{
    var canFly = CanFlyOnGrid(gridUid);
    SetEnabled((user.Jetpack, jetpackComponent), jetpackComponent.Enabled, flyIfEnabled: canFly, user: uid);
```

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
     {
-        if (args.Container.Owner != ent.Comp.JetpackUser)
-            SetEnabled(ent, ent.Comp, false, ent.Comp.JetpackUser);
+        var (userUid, _) = ent;
+        var jetpackUid = ent.Comp.Jetpack;
+
+        if (!TryComp<JetpackComponent>(jetpackUid, out var jetpackComponent))
+            return;
+
+        var userGrid = Transform(userUid).GridUid;
```
perryprog: Can use `args.Transform.GridUid`.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
     }
 
-    private void OnJetpackMoved(Entity<JetpackComponent> ent, ref EntGotInsertedIntoContainerMessage args)
+    private void OnJetpackUserCanWeightless(Entity<JetpackUserComponent> ent, ref CanWeightlessMoveEvent args) => args.CanMove = true;
+
+    private void OnJetpackUserEntParentChanged(Entity<JetpackUserComponent> ent, ref EntParentChangedMessage args)
     {
-        if (args.Container.Owner != ent.Comp.JetpackUser)
-            SetEnabled(ent, ent.Comp, false, ent.Comp.JetpackUser);
+        var (userUid, _) = ent;
```
perryprog: `Entity<T>` will automatically be cast to an `EntityUid` when needed, so you don't need this destructor—just pass in `ent` where you passed in `userUid`.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
+
+        var userGrid = Transform(userUid).GridUid;
+
+        var canFly = CanFlyOnGrid(userGrid);
+        var jetpackEnabled = jetpackComponent.Enabled;
+
+        if (!canFly && jetpackEnabled)
+            _popup.PopupClient(Loc.GetString("jetpack-to-grid"), userUid, userUid);
+
+        SetEnabled(new Entity<JetpackComponent>(jetpackUid, jetpackComponent), jetpackEnabled, flyIfEnabled: canFly, user: userUid);
```
perryprog: Same deal, can just do
```c#
SetEnabled((jetpackUid, jetpackComponent), jetpackEnabled, flyIfEnabled: canFly, user: userUid);
```

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
-
-    public void SetEnabled(EntityUid uid, JetpackComponent component, bool enabled, EntityUid? user = null)
+    /// <summary>
+    /// Sets the provided jetpack to whether it is enabled, and if so, whether it can fly.
+    /// <paramref name="enabled"/> is only whether the jetpack should appear active, but
+    /// <paramref name="flyIfEnabled"/> is whether the user should actually be able to fly with it.
+    /// </summary>
+    /// <param name="enabled">Whether the jetpack should appear active, but not necessarily let the user fly.</param>
+    /// <param name="flyIfEnabled">Whether the jetpack should let the user fly, if <paramref name="enabled"/> is true. If null, defaults to whether the user is already flying.</param>
+    public void SetEnabled(Entity<JetpackComponent> jetpack, bool enabled, EntityUid? user = null, bool? flyIfEnabled = false)
```
perryprog: This is a public method, so changes to its signature are considered breaking. You can either leave it in the old style, or add an overload with the old signature that's tagged with `[Obsolete("Use Entity<T> variant")]` and just calls this version.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
+            activeJetpackComponent = null;
+            return false;
+        }
+
+        return TryComp(jetpackUserComponent.Jetpack, out activeJetpackComponent);
     }
+
+    /// <returns>Whether the provided entity is using an active jetpack to fly.</returns>
+    public bool IsUserFlying(EntityUid uid)
+        => _jetpackUserQuery.TryComp(uid, out var jetpackUserComponent) && _activeJetpackQuery.TryComp(jetpackUserComponent.Jetpack, out _);
```
perryprog:
```suggestion
    public bool IsUserFlying(EntityUid uid)
        => _jetpackUserQuery.TryComp(uid, out var jetpackUserComponent) 
           && _activeJetpackQuery.HasComp(jetpackUserComponent.Jetpack);
```

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
+
+    /// <remarks>
+    /// This should return only whether you can <i>turn the jetpack on/off</i>, such as according to whether it has enough fuel to work,
+    /// not necessarily meaning being able to fly with it.
+    /// Return value should stay regardless of <see cref="CanFlyOnGrid"/>.
+    /// </remarks>
+    protected virtual bool CanEnable(Entity<JetpackComponent> jetpack)
```
perryprog: This should be made abstract.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
perryprog: This dirty also isn't needed.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
 
-        if (TryComp(uid, out TransformComponent? xform) && !CanEnableOnGrid(xform.GridUid))
-        {
-            _popup.PopupClient(Loc.GetString("jetpack-no-station"), uid, args.Performer);
+        var user = args.Performer;
+        var (uid, component) = jetpack;
+        var toggled = !component.Enabled;
 
+        // If they're already using a jetpack that isn't this one and trying to turn this one on, don't let them.
+        if (TryComp<JetpackUserComponent>(user, out var userComponent) && userComponent.Jetpack != uid)
```
perryprog:
```suggestion
        if (_jetpackUserQuery.TryComp(user, out var userComponent) && userComponent.Jetpack != uid)
```

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
+        {
+            _popup.PopupClient(Loc.GetString("jetpack-already-using"), user);
             return;
         }
 
-        SetEnabled(uid, component, !IsEnabled(uid));
+        // You can still turn the jetpack on/off when on a grid that doesn't permit flying; you just won't be able to fly!
+        var canFly = TryComp(uid, out TransformComponent? xform) && CanFlyOnGrid(xform.GridUid);
```
perryprog: While you're here:

```suggestion
        var canFly = CanFlyOnGrid(Transform(jetpack).GridUid);
```

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +21,278 @@ public abstract class SharedJetpackSystem : EntitySystem
...
-    private bool IsEnabled(EntityUid uid)
-    {
-        return HasComp<ActiveJetpackComponent>(uid);
-    }
-
-    public void SetEnabled(EntityUid uid, JetpackComponent component, bool enabled, EntityUid? user = null)
+    /// <summary>
+    /// Sets the provided jetpack to whether it is enabled, and if so, whether it can fly.
+    /// <paramref name="enabled"/> is only whether the jetpack should appear active, but
+    /// <paramref name="flyIfEnabled"/> is whether the user should actually be able to fly with it.
```
perryprog: I think your parameter descriptions are detailed enough that you can leave these two lines out. (Thanks for adding docs, but of course no good deed goes unpunished :^)).

perryprog (commented):
Not a maintainer so you can feel free to ignore this, but I had a few notes when reading through. Nothing substantive in terms of logic (bit too tired to try to follow that right now).

In general you usually don't "need" to deconstruct `Entity<T>`'s, and it often leads to nicer code if you don't.

As much as I like them myself, it is pretty uncommon in this code base to use expression bodies for methods, so do with that as you will.

I do think the naming of things could be a lot clearer, though I recognize that it's an unintuitive thing _to_ name :P.

Content.Client/Movement/Systems/JetpackSystem.cs
LaCumbiaDelCoronavirus: Good catch, it was a mistake

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
LaCumbiaDelCoronavirus: its *generally* better practice iirc

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
LaCumbiaDelCoronavirus: I forgor

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
LaCumbiaDelCoronavirus: ok that's epic

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +22,269 @@ public abstract class SharedJetpackSystem : EntitySystem
...
-        SubscribeLocalEvent<JetpackComponent, MapInitEvent>(OnMapInit);
+        SubscribeLocalEvent<JetpackComponent, DroppedEvent>(OnJetpackDropped);
     }
 
-    private void OnJetpackUserWeightlessMovement(Entity<JetpackUserComponent> ent, ref RefreshWeightlessModifiersEvent args)
+    private void OnJetpackUserWeightlessMovement(Entity<JetpackUserComponent> jetpackUser, ref RefreshWeightlessModifiersEvent args)
     {
+        if (!IsUserFlying(jetpackUser))
+            return;
+
```
Princess-Cheeseballs, line 55: The component isn't created in time for this to reliably work. This needs a different solution.

LaCumbiaDelCoronavirus: comp was being made after modifiers were refreshed instead of before so oops (tested, works)

Content.Shared/Movement/Components/ActiveJetpackComponent.cs
```diff
@@ -4,16 +4,26 @@
...
+    /// Defaults to null so that prediction reset doesnt spam particles.
+    /// </summary>
+    public EntityCoordinates? LastCoordinates = null;
 
-    public TimeSpan TargetTime = TimeSpan.Zero;
+    /// <summary>
+    /// The time after which particles can be emitted from this jetpack.
+    /// </summary>
+    public TimeSpan TargetTime = TimeSpan.MinValue;
 }
```
Princess-Cheeseballs: All these should be datafields

Content.Shared/Movement/Components/ActiveJetpackComponent.cs
```diff
@@ -4,16 +4,26 @@
 namespace Content.Shared.Movement.Components;
 
 /// <summary>
-/// Added to an enabled jetpack. Tracks gas usage on server / effect spawning on client.
+/// Added to an enabled and flying jetpack. Tracked to know when to start handling gas usage on server, and spawn flying particles on client.
 /// </summary>
 [RegisterComponent, NetworkedComponent]
 public sealed partial class ActiveJetpackComponent : Component
 {
     public float EffectCooldown = 0.3f;
```
Princess-Cheeseballs: This should be a TimeSpan

Content.Shared/Movement/Components/ActiveJetpackComponent.cs
```diff
@@ -4,16 +4,26 @@
...
+    /// Max distance from <see cref="LastCoordinates"/> that this jetpack must be to emit particles.
+    /// </summary>
     public float MaxDistance = 0.7f;
 
-    public EntityCoordinates LastCoordinates;
+    /// <summary>
+    /// The last position at which particles were emitted from this jetpack.
+    /// Defaults to null so that prediction reset doesnt spam particles.
+    /// </summary>
+    public EntityCoordinates? LastCoordinates = null;
```
Princess-Cheeseballs, line 25: Don't need to specify null

Content.Client/Movement/Systems/JetpackSystem.cs
```diff
@@ -46,26 +47,29 @@ public override void Update(float frameTime)
...
+            var currentCoords = _transform.GetMoverCoordinates(transform.Coordinates);
+            if (comp.LastCoordinates is not { } lastCoords)
             {
-                if (_timing.CurTime < comp.TargetTime)
-                    continue;
+                comp.LastCoordinates = currentCoords;
+                continue;
             }
+            else if (_transform.InRange(transform.Coordinates, lastCoords, comp.MaxDistance) && _timing.CurTime < comp.TargetTime)
+                continue;
```
Princess-Cheeseballs: Don't need the else there because we continue anyways. Just have it be an if statement.

Content.Client/Movement/Systems/JetpackSystem.cs
```diff
@@ -46,26 +47,29 @@ public override void Update(float frameTime)
...
+            // As long as the component is clientside it bulldozes *EVERYTHING*.
+            var currentCoords = _transform.GetMoverCoordinates(transform.Coordinates);
+            if (comp.LastCoordinates is not { } lastCoords)
             {
-                if (_timing.CurTime < comp.TargetTime)
-                    continue;
+                comp.LastCoordinates = currentCoords;
+                continue;
             }
+            else if (_transform.InRange(transform.Coordinates, lastCoords, comp.MaxDistance) && _timing.CurTime < comp.TargetTime)
```
Princess-Cheeseballs: I'd also add a comment about this that we don't want to make new particles unless we're out of range or the timer is up.

Content.Shared/Movement/Components/JetpackComponent.cs
```diff
@@ -7,6 +7,10 @@ namespace Content.Shared.Movement.Components;
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
 public sealed partial class JetpackComponent : Component
 {
+    /// <summary>Whether the jetpack is turned on, but not whether it's actually flying!</summary>
+    [AutoNetworkedField]
+    public bool Enabled = false;
+
```
Princess-Cheeseballs: Datafield

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -20,182 +22,269 @@ public abstract class SharedJetpackSystem : EntitySystem
...
+        => args.AddAction(ref component.ToggleActionEntity, component.ToggleAction);
+
+    /// <summary>
+    /// Sets the provided jetpack to whether it is enabled, and if so, whether it can fly.
+    /// <paramref name="enabled"/> is only whether the jetpack should appear active, but
+    /// <paramref name="flyIfEnabled"/> is whether the user should actually be able to fly with it.
+    /// </summary>
+    /// <param name="enabled">Whether the jetpack should appear active, but not necessarily let the user fly.</param>
+    /// <param name="flyIfEnabled">Whether the jetpack should let the user fly, if <paramref name="enabled"/> is true. If null, defaults to whether the user is already flying.</param>
+    public void SetEnabled(Entity<JetpackComponent> jetpack, bool enabled, EntityUid? user = null, bool? flyIfEnabled = false)
```
Princess-Cheeseballs: I would split this into two methods, one with the flyIfEnabled bool and one without and make it not nullable. I'd have both call another private method where the current flyIfEnabled ??=IsUserFlying is that both call.

Content.Shared/Movement/Components/ActiveJetpackComponent.cs
LaCumbiaDelCoronavirus: I had to execute this workaround because otherwise, upon `ActiveJetpackComponent` being added to the jetpack, it would spam-spawn particles for the few moments that the clientside state was being bulldozed with the serverside state. See PR media for how it looks like before i 'fixed' it.

I would really like to know if there's a better way to do this.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -197,48 +196,83 @@ private bool CanFlyOnGrid(EntityUid? gridUid)
...
+        if (!Container.TryGetContainingContainer((uid, null, null), out var container))
+            return false;
+
+        user = container.Owner;
+        return true;
+    }
+
+    private void SetJetpackState(Entity<JetpackComponent> jetpack, EntityUid user, bool enabled, bool flyIfEnabled)
+    {
+        Log.Debug($"Setting jetpack state to {(enabled ? "on" : "off")}, while {(flyIfEnabled ? "flying" : "not flying")}.");
```
Princess-Cheeseballs, line 254: Remove.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -197,48 +196,83 @@ private bool CanFlyOnGrid(EntityUid? gridUid)
...
+
     /// <summary>
     /// Sets the provided jetpack to whether it is enabled, and if so, whether it can fly.
-    /// <paramref name="enabled"/> is only whether the jetpack should appear active, but
-    /// <paramref name="flyIfEnabled"/> is whether the user should actually be able to fly with it.
     /// </summary>
-    /// <param name="enabled">Whether the jetpack should appear active, but not necessarily let the user fly.</param>
-    /// <param name="flyIfEnabled">Whether the jetpack should let the user fly, if <paramref name="enabled"/> is true. If null, defaults to whether the user is already flying.</param>
-    public void SetEnabled(Entity<JetpackComponent> jetpack, bool enabled, EntityUid? user = null, bool? flyIfEnabled = false)
+    /// <param name="enabled">Whether the jetpack should be on/off, but not necessarily flying.</param>
```
Princess-Cheeseballs, line 203: remember to put the other params in the comments.

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -197,48 +196,83 @@ private bool CanFlyOnGrid(EntityUid? gridUid)
...
+                EndUserFlying(user, jetpack);
         }
         else
         {
-            RemoveUser(user.Value, jetpack);
-            if (IsFlying(uid))
-                EndUserFlying(user.Value, jetpack);
+            RemoveUser(user, jetpack);
+            if (IsFlying(jetpack))
+                EndUserFlying(user, jetpack);
```
Princess-Cheeseballs, line 268: Not sure how I feel about us checking for IsFlying potentially multiple times when disabling a jetpack.

LaCumbiaDelCoronavirus: I don't think it's really a big problem, however there's also something that isn't exactly a noticeable issue but could be bad anyway: `EntParentChangedMessage` gets spammed on client whenever an entity's parent changes, rather than once for every time that happens, which means this check gets done *many* times there.

Princess-Cheeseballs: That's because of prediction running parent changing multiple times. Spamming isn't the issue it's an important part of networking. So yeah you should probably try and optimize this check around that fact.

Princess-Cheeseballs:
Hello, are you planning on coming back to this?

LaCumbiaDelCoronavirus:
no 😂

PR #37792, opened 2025-05-24
Update UIControllers to use `Log`

Tayrtahn:
## About the PR

Updates `UIController`-derived classes that use logging to use the `Log` property added by https://github.com/space-wizards/RobustToolbox/pull/5967.

## Why / Balance

Fixes 9 warnings for https://github.com/space-wizards/space-station-14/issues/33279
Reduces a little boilerplate for some `UIControllers` that were already doing it the right way.

## Technical details

- Replaces all calls to the static `Logger` with calls to `UIController.Log`, which is a controller-specific `ISawmill` instance.
- Some classes were already handling logging correctly by adding `ILogManager` as a dependency and using `GetSawmill` to get an instance on initialize. This is a bit redundant with the engine change, so these were changed to just use `Log` too.

## Requirements


Content.Client/UserInterface/Systems/Bwoink/AHelpUIController.cs
```diff
@@ -48,6 +48,8 @@ public sealed class AHelpUIController: UIController, IOnSystemChanged<BwoinkSyst
     private bool _bwoinkSoundEnabled;
     private string? _aHelpSound;
 
+    protected override string SawmillName => "c.s.go.es.bwoink";
```
Tayrtahn, line 51: For backward-compatibility.

Content.Client/UserInterface/Systems/Character/CharacterUIController.cs
```diff
@@ -28,21 +28,16 @@ namespace Content.Client.UserInterface.Systems.Character;
...
     [UISystemDependency] private readonly CharacterInfoSystem _characterInfo = default!;
     [UISystemDependency] private readonly SpriteSystem _sprite = default!;
 
-    private ISawmill _sawmill = default!;
-
     public override void Initialize()
     {
         base.Initialize();
 
-        _sawmill = _logMan.GetSawmill("character");
```
Tayrtahn, line 44: This will be logged as "ui.character" now. I'm assuming that's not a problem, but if it is let me know and I'll add an override.

slarticodefast: Should be fine I think.

slarticodefast (approved):
Good to merge once the RT PR is merged and the engine has been updated on the content repo.

PR #37793, opened 2025-05-24
New Discord integration fixes

PJB3005:
Fix admin chat relay.

Fix leaked task instance.

Fix warning about gateway intents on startup.

Fix packaging.


Content.Packaging/ServerPackaging.cs
```diff
@@ -47,6 +47,7 @@ public static class ServerPackaging
         // Python script had Npgsql. though we want Npgsql.dll as well soooo
         "Npgsql",
         "Microsoft",
+        "Discord",
```
Simyon264, line 50: I wish we had docs on this ngl. Had no idea you had to add this here for nuget packages to show up there.

PR #37796, opened 2025-05-24
Removes references to GetAnchoredEntitiesEnumerator(Vector2i)

Cojoke-dot:
## About the PR

Title
## Why / Balance

#33279, `grid.GetAnchoredEntitiesEnumerator(Vector2i)` is Obsolete
## Technical details

Replace places using `grid.GetAnchoredEntitiesEnumerator(Vector2i)` with `mapSystem.GetAnchoredEntitiesEnumerator(uid, grid, Vector2i)`

## Requirements

### DIFF
--- a/Content.Shared/Movement/Systems/SharedMoverController.cs
+++ b/Content.Shared/Movement/Systems/SharedMoverController.cs
@@ -567,7 +567,7 @@ public abstract partial class SharedMoverController : VirtualController
 
         // If the coordinates have a FootstepModifier component
         // i.e. component that emit sound on footsteps emit that sound
-        var anchored = grid.GetAnchoredEntitiesEnumerator(position);
+        var anchored = _mapSystem.GetAnchoredEntitiesEnumerator(uid, grid, position);
 
         while (anchored.MoveNext(out var maybeFootstep))
         {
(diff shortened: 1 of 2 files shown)


Content.Shared/Movement/Systems/SharedMoverController.cs
```diff
@@ -567,7 +567,7 @@ private bool TryGetFootstepSound(
 
         // If the coordinates have a FootstepModifier component
         // i.e. component that emit sound on footsteps emit that sound
-        var anchored = grid.GetAnchoredEntitiesEnumerator(position);
+        var anchored = _mapSystem.GetAnchoredEntitiesEnumerator(uid, grid, position);
```
Tayrtahn, line 570: This should be the uid of the grid, not the mover. The value isn't actually used, so it works like this anyway, but the right value should be used.
```suggestion
        var anchored = _mapSystem.GetAnchoredEntitiesEnumerator(xform.GridUid.Value, grid, position);
```

Content.Shared/Random/Rules/NearbyTilesPercent.cs
```diff
@@ -45,7 +45,7 @@ public override bool Check(EntityManager entManager, EntityUid uid)
             // Only consider collidable anchored (for reasons some subfloor stuff has physics but non-collidable)
             if (IgnoreAnchored)
             {
-                var gridEnum = grid.GetAnchoredEntitiesEnumerator(tile.GridIndices);
+                var gridEnum = mapSys.GetAnchoredEntitiesEnumerator(uid, grid, tile.GridIndices);
```
Tayrtahn, line 48: Same goes here.

Tayrtahn (changes requested):

Tayrtahn (approved):

PR #37797, opened 2025-05-24
Normal and Cotton Dough Slices/Ropes can be cooked using heat

SyaoranFox:
## About the PR
Buns and Bagels (normal and cotton) can now be cooked using heat

## Why / Balance
In order to allow Chef to become more interactive, we should try to move away from relying entirely on microwaves. Heat means that Chef can cook using ANY source of heat.

## Technical details
Dough slices and ropes have had their construction tree expanded to incorporate most of the products in its production chain. I didn't move the tree to include loaves of bread to simplify how many things needed to be changed for refactoring considerations.

Temperature and InternalTemperature was added to FoodBakingBase, InternalTemperature is the main component that allows the cooking trigger to work.

I renamed the cooking trees from DoughRope and CottonDoughRope to CookingDough and CookingCottonDough. I would like to expand these trees to include loaves as well, and to refactor how the recipes/cooking is organised. Currently it is very messy and does not make logical sense (however this would be future stepped PRs)

Currently I set the baking temperature to 330, this is the same temperature to turn Raw Meat into Steaks. This should be adjusted to something more appropriate later.

## Media
cookinBreads

## Requirements

## Breaking changes
Some names were changed here, however I was unable to find any conflicts. I have no tested what happens if I make 100 dough pieces, and then set a trit fire. This may cause everything to trigger at once.

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/bagel.yml
+++ b/Resources/Prototypes/Entities/Objects/Consumable/Food/Baked/bagel.yml
@@ -53,7 +53,7 @@
           Quantity: 5
 
 - type: entity
-  id: FoodBagelCotton
+  id: FoodCottonBagel
   parent: FoodBagelBase
   name: cotton bagel
   description: A delicious bagel made with cotton dough.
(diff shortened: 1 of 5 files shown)


southbridge-fur (commented):
While I'm not against the idea conceptually, as someone who bakes a lot of bread usually you need the dough to at least be on a stone or in a tin to cook it over an open flame. I'd like to see these recipes adjusted to use the frankly under-utilized baking equipment we already have. You could totally put the dough ropes in a pie tin and then cook those to create bagels for example.

Issue #24151, opened 2024-01-16
Wagging your tail doesnt work for large tail marking


UbaserB:
## Description

This makes me very sad as i want to wag my tail but there's no animation.

**Screenshots**

![image](https://github.com/space-wizards/space-station-14/assets/134914314/73468fdc-4f2e-4c26-b50a-b500d8d7abf5)

TiniestShark:
Large tail wag has been merged, all lizards can properly express their happy

PR #37806, opened 2025-05-25
Large Tail Wag

TiniestShark:
## About the PR
The Large tail has been apparently missing a wag animation. This fixes that.

## Why / Balance
big tail lizard should be able to be happy too, also resolves [#24151] 

## Technical details
Single section yaml change

## Media
wag beta 2

## Requirements

**Changelog**


Pronana:
> Also I feel like the base of the tail for the back sprite doesn't quite match the motion of the left and right directions

That's a common issue with the regular tail wags as well I feel

TiniestShark:
> Darn beat me to it
> 
> Also I feel like the base of the tail for the back sprite doesn't quite match the motion of the left and right directions

It doesn't but that's because the base large tail is pretty much centered on the lizard, compared to the other tails which all curve off at the base. There's less to move with it and when I tried it just looked awkward.

PR #37817, opened 2025-05-25
fix: deglitchify drowsiness effect on macOS

perryprog:
## About the PR

Adjusted the radial blur shader (used by the drowsiness status effect) to not affect the alpha channel.

## Why / Balance

For some reason, and I presume this is either a macOS or Rosetta issue, this shader has never worked for me (see media)... so I made it work. 

## Technical details

I have no idea why this fixes it or what the issue I was having was?? Recreating the original shader in ShaderToy I'm unable to recreate the problem... regardless, this seems like a fine compromise and I don't think it should affect the visual quality of the effect?

Note: This change should be tested by someone else on different hardware to ensure I didn't fix it by breaking it for someone else. I also don't have a good reference for what this effect originally looked like. The screenshots below were done by just giving myself 10u of chloral hydrate. 

## Media

<details><summary>Media</summary>

-  What I saw before (at any strength)
  image
- Light intensity
  image
- Medium/high? intensity
   image

</details>

## Requirements

## Breaking changes

**Changelog**

No

### DIFF
--- a/Resources/Textures/Shaders/radial_blur.swsl
+++ b/Resources/Textures/Shaders/radial_blur.swsl
@@ -4,11 +4,13 @@ const highp int SampleCount = 10; // a higher number makes the shader look bette
 
 // a simple radial blur
 void fragment() {
-    highp vec2 uv = FRAGCOORD.xy * SCREEN_PIXEL_SIZE.xy;
-    highp vec2 direction = vec2(0.5, 0.5) - uv;
+    highp vec3 col = texture(SCREEN_TEXTURE, UV).xyz;
+    highp vec4 color = zTexture(UV);
+    highp vec2 direction = vec2(0.5, 0.5) - UV;
     for (int i=1; i <= SampleCount; i++)
     {
-        COLOR += zTextureSpec(SCREEN_TEXTURE, uv + float(i) * Strength / float(SampleCount) * direction);
+        col += zTextureSpec(SCREEN_TEXTURE, UV + float(i) * Strength / float(SampleCount) * direction).xyz;
     }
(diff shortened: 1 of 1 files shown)


slarticodefast (dismissed):
Thanks for looking into it. The effect does look differently than before though, it is supposed to become stronger towards the edge of the screen. Here is a screenshot at max intensity from 50u of haloperidol. Or did you just have a lower intensity?
Screenshot (400)

perryprog:
Oh yeah haloperidol exists. Looks like my screenshots are just a low intensity—here's what I have from 50u of haloperidol.
image

slarticodefast:
I asked Bhijn & Myr for shader feedback and this seems good, but I'll try to find someone else with a mac so we can figure out what is going on and confirm this is working
grafik

deathride58:
PJB might be able to help out soon, [given that she's got a Mac arriving](https://discord.com/channels/310555209753690112/1340464307430752396/1382369532496318615). Our guess is that it's probably an engine-level renderer bug (which is likely to have impacted downstreams if a shader as simple as the current one is ending up with weird corruption during a for loop)

perryprog:
Glad to hear! This is definitely the only shader I've seen with this issue, even including various funky downstream shaders, for what it's worth.

Tayrtahn:
I can confirm that the current shader is glitchy and that the modification here works as shown on my M1 MacBook!

slarticodefast:
Well, I don't see a difference on my windows computer, so I don't see a reason not to merge this as a workaround until PJB can investigate the renderer.
Thanks for the bugfix!

PR #27369, opened 2024-04-26
Flash overlay rework and bugfixes

slarticodefast:
## About the PR
This reworks the flash effect to use the status effect system.
Fixes #22217
Fixes #27992
Fixes #27072
Fixes #27743

## Why / Balance
bugfixes and code improvements

## Technical details
The flash effect works by taking a screenshot and having two copies of that screenshot circle around you, overlayed with a black screen. The problem was that this screenshot was taken after the shader was applied, resulting in stacking multiple effects with each use. I solved this by resetting the saved screenshot back to null after the effect has dissipated, which prevents the shader from being drawn until a new screenshot has been taken.

Edit:
I also changed the flash effect to use the status effects system. Previously the FlashOverlay was added in EntryPoint.cs and active at all times, even when the player never gets flashed during the round. Now it is properly attached and detached with the status effect.

I was able to remove FlashableComponent, which was basically doing the same thing as the status effect system, only worse. Even the code comments said it was awful.

I changed InRangeUnobstructed to InRangeUnoccluded. Previously it used FlashableComponent as a whitelist in the predicate, so that most mobs don't block the line of sight of the flash. Now it just checks if the flash is in view and in range. This means that smoke, which occludes vision, will now block flashes and non-occluding entities like the shadow anomaly will not.

Rejuvenate will now remove the status effect from the player.

## Media
before:

https://github.com/space-wizards/space-station-14/assets/161409025/7aabb037-87bb-4605-9127-fcf7232fc957

disabling the shader shows the underlying problem better:
testing2

https://github.com/space-wizards/space-station-14/assets/161409025/e6834a65-b127-4c9a-94f1-662eae49006c

after:

https://github.com/space-wizards/space-station-14/assets/161409025/86875909-5c66-4c95-be2c-f52b4fbfae05

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes
Removed FlashableComponent. If you want to allow mobs to be affected by flashes, allow the new Flashed status effect instead.

**Changelog**
:cl:
- fix: Fixed the flash effect getting darker with each appearence.
- fix: Fixed short-sightedness making you immune to flashes.

PR #37824, opened 2025-05-25
reduced motion flash effect version 3

slarticodefast:
## About the PR
See https://github.com/space-wizards/space-station-14/pull/36288 and https://github.com/space-wizards/space-station-14/pull/37583
Since the previous PRs all had someone who did not like how the effect looked like I made another very simple alternative effect that we can merge in the meantime until a better shader has been made. This should allow the affected players to be able to play in Revolutionary mode again for now.

The effect is a simple black overlay with the same falloff curve as used in the shader, so it should be comparable concerning combat balance.

In the past the flash shader was bugged for quite a while, causing it to look like this (see https://github.com/space-wizards/space-station-14/pull/27369). So we already know that this works from a gameplay perspective.

## Why / Balance
accessibility

## Technical details
If the reduced motion cvar is enabled, just draw a black color rectangle over the screen instead of the shader.
The alpha value is chosen from the same falloff curve used for blending the flash shader.

## Media

https://github.com/user-attachments/assets/5e96677c-5da7-4b3f-853b-c86f216ca514

## Requirements

## Breaking changes

**Changelog**
:cl: Quantum-cross, slarticodefast
- add: The flash effect is now replaced with a reduced motion variant if the accessibility setting is enabled in the game options.

### DIFF
--- a/Content.Client/Flash/FlashOverlay.cs
+++ b/Content.Client/Flash/FlashOverlay.cs
@@ -24,8 +24,8 @@ namespace Content.Client.Flash
 
         public override OverlaySpace Space => OverlaySpace.WorldSpace;
         private readonly ShaderInstance _shader;
-        private bool _reducedMotion = false;
-        public float PercentComplete = 0.0f;
+        private bool _reducedMotion;
+        public float PercentComplete;
         public Texture? ScreenshotTexture;
 
         public FlashOverlay()
@@ -82,6 +82,10 @@ namespace Content.Client.Flash
             var worldHandle = args.WorldHandle;
             if (_reducedMotion)
             {
+                // TODO: This is a very simple placeholder.
+                // Replace it with a proper shader once we come up with something good.
+                // Turns out making an effect that is supposed to be a bright, sudden, and disorienting flash 
+                // not do any of that while also being equivalent in terms of game balance is hard.
                 var alpha = 1 - MathF.Pow(PercentComplete, 8f); // similar falloff curve to the flash shader
                 worldHandle.DrawRect(args.WorldBounds, new Color(0f, 0f, 0f, alpha));
             }
--- a/Content.Client/Flash/FlashOverlay.cs
+++ b/Content.Client/Flash/FlashOverlay.cs
@@ -1,8 +1,10 @@
+using Content.Shared.CCVar;
 using Content.Shared.Flash;
 using Content.Shared.Flash.Components;
 using Content.Shared.StatusEffect;
 using Robust.Client.Graphics;
 using Robust.Client.Player;
+using Robust.Shared.Configuration;
 using Robust.Shared.Enums;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Timing;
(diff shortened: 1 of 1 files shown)


Content.Client/Flash/FlashOverlay.cs
```diff
@@ -15,12 +17,14 @@ public sealed class FlashOverlay : Overlay
...
         [Dependency] private readonly IPlayerManager _playerManager = default!;
         [Dependency] private readonly IGameTiming _timing = default!;
+        [Dependency] private readonly IConfigurationManager _configManager = default!;
 
         private readonly SharedFlashSystem _flash;
         private readonly StatusEffectsSystem _statusSys;
 
         public override OverlaySpace Space => OverlaySpace.WorldSpace;
         private readonly ShaderInstance _shader;
+        private bool _reducedMotion = false;
```
Errant-4: No need to assign the default value. 
Next line, same

Content.Client/Flash/FlashOverlay.cs
```diff
@@ -74,10 +80,18 @@ protected override void Draw(in OverlayDrawArgs args)
                 return;
 
             var worldHandle = args.WorldHandle;
-            _shader.SetParameter("percentComplete", PercentComplete);
-            worldHandle.UseShader(_shader);
-            worldHandle.DrawTextureRectRegion(ScreenshotTexture, args.WorldBounds);
-            worldHandle.UseShader(null);
+            if (_reducedMotion)
```
Errant-4, line 83: If we intend this as a placeholder it should probably be commented

Content.Client/Flash/FlashOverlay.cs
```diff
@@ -15,12 +17,14 @@ public sealed class FlashOverlay : Overlay
...
         [Dependency] private readonly IGameTiming _timing = default!;
+        [Dependency] private readonly IConfigurationManager _configManager = default!;
 
         private readonly SharedFlashSystem _flash;
         private readonly StatusEffectsSystem _statusSys;
 
         public override OverlaySpace Space => OverlaySpace.WorldSpace;
         private readonly ShaderInstance _shader;
+        private bool _reducedMotion = false;
         public float PercentComplete = 0.0f;
```
slarticodefast:
```suggestion
        private bool _reducedMotion;
        public float PercentComplete;
```

Content.Client/Flash/FlashOverlay.cs
```diff
@@ -74,10 +80,18 @@ protected override void Draw(in OverlayDrawArgs args)
                 return;
 
             var worldHandle = args.WorldHandle;
-            _shader.SetParameter("percentComplete", PercentComplete);
-            worldHandle.UseShader(_shader);
-            worldHandle.DrawTextureRectRegion(ScreenshotTexture, args.WorldBounds);
-            worldHandle.UseShader(null);
+            if (_reducedMotion)
+            {
+                var alpha = 1 - MathF.Pow(PercentComplete, 8f); // similar falloff curve to the flash shader
```
slarticodefast, line 89:
```suggestion
                // TODO: This is a very simple placeholder.
                // Replace it with a proper shader once we come up with something good.
                // Turns out making an effect that is supposed to be a bright, sudden, and disorienting flash 
                // not do any of that while also being equivalent in terms of game balance is hard.
                var alpha = 1 - MathF.Pow(PercentComplete, 8f); // similar falloff curve to the flash shader
```

Errant-4 (approved):

Issue #37823, opened 2025-05-25
Total lack of standardization between running/walking speed modifiers.


RedBookcase:
## Description
Equipped gear which effects movement speed (such as hardsuits) completely lacks any form of standardization between the debuffs imposed on running and walking speed. Some of these items take what I think is the correct approach and simply give a flat debuff to all movement speed (Example - Engineering Hardsuit). Some gear gives a higher debuff to running speed than to walking speed(Spationaut Hardsuit), and some gear does the opposite and gives a higher debuff to walking speed (CE Hardsuit). 

**Screenshots**

Engineering Hardsuit

![Image](https://github.com/user-attachments/assets/2ba5cbc5-a56f-429f-b3a4-a55eb8a864e2)

Spationaut Hardsuit

![Image](https://github.com/user-attachments/assets/dc432e71-a47b-4748-a523-44caea909e0a)

CE Hardsuit

![Image](https://github.com/user-attachments/assets/2ce401c2-8713-4d69-8df2-08818221aefc)

**Additional context**
In my opinion even having separate modifiers for the running and walk speed is needless added complexity which only serves to add more clutter to the often already cluttered equipment stats popup. The separation of running and walking speed is extremely arbitrary as well - the Warden's Hardsuit has a flat 30% modifier to it's movement speed, but the Juggernaut Hardsuit has a 35% decrease to running speed and only a 10% decrease to walking speed. Why?! Why is walking and running in a Juggsuit nearly identical, speed-wise? Why is it so important that the Juggernaut suit have this separated speed if the Warden's Hardsuit doesn't need it?

![Image](https://github.com/user-attachments/assets/6c3f079f-d13d-4ce8-8eba-74695503e462)

I think that having separate modifiers for running and walking speed CAN be interesting or useful, to be clear. If we ever add in something like leg shackles, it would make sense if they gave you a huge slowdown to running speed but let you walk relatively normally, as an abstract way to represent tripping and fumbling when your legs are cuffed together. But I see no reason to apply that level of minute detail to something like a Hardsuit. We should just standardize their movement speeds across the board based off of whatever modifier's given for the running speed.

Thank you for coming to my TED Talk.

slarticodefast:
Yeah, for clothing items it both values should probably just be identical.

Pronana:
[...]

PR #37828, opened 2025-05-25
Standardized running and walking speed debuffs for equipped gear.

RedBookcase:
## About the PR
All equipped clothing and gear that imposes a debuff on movement speed has been standardized. Equipped items no longer have separate values for running and walk speed debuffs - instead they are now debuffed equally.

## Why / Balance
There is no standardization among equipped gear in regards to walking and running speed debuffs. Some gear debuffs running more heavily than walking - some does the opposite. Some gear only debuffs running, and some only debuffs walking.

Having separate lines on inspect for running and walking speed debuffs is extra clutter for the already often cluttered item stats popup.

By standardizing the movement debuff to a single flat number, we can make speed debuffs easier to understand intuitively. If you wear clothing with a movement speed debuff, you'll know that both your running and walking speed are being debuffed by the same amount. Simple, straightforwards, easy.

Bigger rant on the issue here (with pictures!): https://github.com/space-wizards/space-station-14/issues/37823

## Technical details
All yaml changes. A full list of gear that has had it's speed changed by this PR follows. All gear has had it's walking speed debuff changed to match the sprint speed debuff.

Hardsuits: ClothingOuterHardsuitBase, ClothingOuterHardsuitSpatio, ClothingOuterHardsuitEngineeringWhite, ClothingOuterHardsuitMedical, ClothingOuterHardsuitLuxury, ClothingOuterHardsuitSyndieElite, ClothingOuterHardsuitJuggernaut, 

Other Outerwear: ClothingOuterBioGeneral, ClothingOuterArmorChangeling, ClothingOuterArmorBone, ClothingOuterSuitFire,

Shoes: ClothingShoesBootsMagBase, ClothingShoesBootsMagSyndie, ClothingShoesSpaceNinja, ClothingShoesClownLarge

Didn't touch: Duffel bags (They only slow you down while running, which seems like a unique differentiation for a backpack to have), XenoArtifactSpeedUp (Affects move speed while holding artifacts with the speed ability, outside the scope of the PR as it isn't equipped)

## Requirements

**Changelog**

:cl:
- tweak: The running and walking speed debuffs on equipped gear have been standardized.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
@@ -418,7 +418,7 @@
         Heat: 0.9
         Radiation: 0.8
   - type: ClothingSpeedModifier
-    walkModifier: 0.7
+    walkModifier: 0.65
     sprintModifier: 0.65
   - type: HeldSpeedModifier
   - type: ExplosionResistance
@@ -445,6 +445,7 @@
         Piercing: 0.4
   - type: ClothingSpeedModifier
     walkModifier: 0.8
+    sprintModifier: 0.8
   - type: HeldSpeedModifier
   - type: ExplosionResistance
     damageCoefficient: 0.4
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/base_clothingouter.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/base_clothingouter.yml
@@ -114,7 +114,7 @@
     heatingCoefficient: 0.01
     coolingCoefficient: 0.01
   - type: ClothingSpeedModifier
-    walkModifier: 0.4
+    walkModifier: 0.6
     sprintModifier: 0.6
   - type: HeldSpeedModifier
   - type: Item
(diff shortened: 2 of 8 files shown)


EmoGarbage404 (approved):
makes sense and afaik there's no balance reason for these to be separate

Pronana:
Mentioned it in the issue, I'd rather just see walking speed be the same in all hardsuits (no penalty) and have a check to cap walking speed to sprinting speed.

RedBookcase:
> Mentioned it in the issue, I'd rather just see walking speed be the same in all hardsuits (no penalty) and have a check to cap walking speed to sprinting speed.

I think it makes more sense that a hardsuit would slow you down in roughly equal amounts regardless of if you're walking or running, given that it's a big clunky suit of pressurized metal and fabric. As for your point on avoiding slipping - I think that there's just not enough of a niche there for it to matter. The suits that have big slowdowns already slow you down enough to avoid slipping, and the ones with less slowdown don't have enough of a debuff to make it that big of an issue when you walk.

Pronana:
> I think it makes more sense that a hardsuit would slow you down in roughly equal amounts regardless of if you're walking or running, given that it's a big clunky suit of pressurized metal and fabric. As for your point on avoiding slipping - I think that there's just not enough of a niche there for it to matter. The suits that have big slowdowns already slow you down enough to avoid slipping, and the ones with less slowdown don't have enough of a debuff to make it that big of an issue when you walk.

Fair enough, I think it would be interesting if "walking speed" was pretty constant and reliable across all suits though with the only exception being if your sprint speed dips below it. 

I mainly say this because I really like how the jug suit feels where walking speed reduction is only 10% while sprint speed takes a big penalty.

Either way these numbers should have some standardization because they are basically completely random at the moment.

PR #37832, opened 2025-05-25
Slat tiles and decals

Aearo-Deepwater:
## About the PR
Added a new set of tiles - horizontal and vertical slats - in steel, dark and white variants, along with light and heavy dirt decals, overlay decals, and edge decals. 

## Why / Balance
These offer a sort of directionality that the other steel tiles mostly lack. I think they'd be good for things like bridges, ramps and pathways, as well as generally offering up some new floor detailing options. See the attached media for some examples of what I mean.

## Media

**Example Scenes:**
2025-05-21-141359_1422x911_scrot
2025-05-21-134511_1116x313_scrot
2025-05-21-135614_1370x946_scrot
2025-05-25-164149_1397x1209_scrot
2025-05-25-171218_1123x1177_scrot

**Tiles:**
2025-05-25-161110_1170x406_scrot

**Dirt Decals:**
(With existing light and heavy steel tile dirt along the top for comparison.)
2025-05-25-161323_1170x496_scrot

**Overlay Decals:**
(Standard departmental colors, with a couple of regular steel tiles for comparison.)
2025-05-25-161747_399x596_scrot

**Single/Double Slat Decals:**
2025-05-25-162325_581x591_scrot

## Requirements

## Changelog
:cl: SeaWyrm
- add: Horizonal and vertical steel slat floor tiles, as well as their dark and white variants, can now be created with a Cutter lathe.


ArcaneOcto:
oh hell yeah. These are super sexy

LevitatingTree:
delectable slat

IProduceWidgets:
I can't help but see them as stairs

Aearo-Deepwater:
I think I'm going to take another look at the footstep sounds and make sure they're consistent.
Actually, I kinda want them to sound different from the regular tile footstep sounds. More rattle-y or something. If there's a footstep sound that matches that.

Aearo-Deepwater:
> I think I'm going to take another look at the footstep sounds and make sure they're consistent. 

Okay, I did that.

Aearo-Deepwater:
I realized some of this is immediately player-facing, so I added a changelog to the PR's main description.

ToastEnjoyer:
I really like these, great sprite work!

TiniestShark:
I really hate to say this but I don't think these tiles really work ingame. They're just really busy visually and make everything look far too messy. They also mesh with the base tiles in a not intuitive looking way and I don't think there's much of a use case for them on upstream. I'm sure other forks would probably enjoy having them, but they stand out harshly against other tiles and even as a decal just don't look great against the quarter tiling upstream aims for.

I talked about this with EmoGarbage to make sure I wasn't alone in my opinion and it was agreed these wouldn't work for us upstream and the decision to close was made. I do apologize and hope this isn't too discouraging, especially with the effort on display.

(I also keep thinking they're one of those moving walkways or a conveyor but that's just a me thing.)

Aearo-Deepwater:
Dang.
Gotta admit, that's kind of surprising, because they're basically the same as the mini-tiles but with fewer lines. Are they really busier than those tiles? They don't mesh with them...?

Like:
2025-07-08-214602_525x626_scrot
2025-07-08-214545_504x597_scrot
2025-07-08-214518_520x621_scrot
2025-07-08-214904_505x596_scrot
I feel like the slats stand out as much as or less than any of these?

Anyway... I don't want to be obnoxious about it or anything, but I hope you'll forgive me wanting to defend my work at least a little bit. :)

ArcaneOcto:
Ngl closing this feels a little outrageous. It's very much less (or just as) busy than some of the tiles we have in game. 100% on Aearo-Deepwater's side here.

Emisse:
opening for more discussion amongus fart directors, for transparency i like these tiles but hate some of our other tiles (checkered looking at you)

Aearo-Deepwater:
So uh
Thanks very much, Emisse, for giving it a second chance!
But I've gotta ask, is it in fact being discussed somewhere? Or is it currently stuck in limbo?
Anything I can do to help get it un-stuck?

Aearo-Deepwater:
@Emisse Any news on this? Where does it stand?
Should I be requesting reviews in the pr-review-request channel again?

Resources/Prototypes/Entities/Objects/Misc/tiles.yml
```diff
@@ -189,6 +189,62 @@
...
+  name: dark steel vertical slat tile continuous
+  parent: FloorTileItemDark
+  id: FloorTileItemDarkVerticalSlatsContinuous
+  components:
+  - type: Sprite
+    state: dark-vertical-slats
+  - type: FloorTile
+    outputs:
+      - Plating
+      - FloorDarkVerticalSlatsContinuous
```
Errant-4: Elements listed under any heading should not be indented if their ` - ` prefix already effectively indents them
```suggestion
    outputs:
    - Plating
    - FloorDarkVerticalSlatsContinuous
```

Same for the other entities below

slarticodefast (changes requested):
Instead of having separate horizontal and vertical tiles this should just use tile rotation, see https://github.com/space-wizards/RobustToolbox/pull/5652

Or does that look off with the spriting?

Aearo-Deepwater:
@slarticodefast I think it'd look off with the spriting, but I also had no idea that existed, so dunno.

Aearo-Deepwater:
@slarticodefast Not sure I understand how this works... there's a 'mirror tile' button that seems to always be disabled, and no way in the UI to rotate them?

Errant-4:
> Not sure I understand how this works... there's a 'mirror tile' button that seems to always be disabled, and no way in the UI to rotate them?

Correct, its not *on* the UI. There is ia hotkey for rotating blueprints/placement. I think the default was the middle mouse button? Should be somewhere near the end of the keybinds

Aearo-Deepwater:
@Errant-4 Middle mouse changes whether the tile placement UI says north/south/east/west, but the placed tiles still don't look any different. Does it only work with some tiles, or something like that?

The "bordered" versions of these tiles wouldn't work for this anyway, since they have those... rivets, or dimples, or whatever the heck they are that the regular steel tiles have, and those line up differently between the two orientations. The "continuous" tiles might be okay, as long as the mapper pays attention to which sides have the one-pixel borders and which don't.

Errant-4:
> @Errant-4 Middle mouse changes whether the tile placement UI says north/south/east/west, but the placed tiles still don't look any different. Does it only work with some tiles, or something like that?
> 
You need to set `allowRotationMirror: true` on the tile prototypes in `floors.yml` to make them rotatable. For those that will visually work, then, I suppose.

Aearo-Deepwater:
2025-10-24-120607_1349x867_scrot
Hmm... well, like I said, mappers would have to be careful with this. But it doesn't fundamentally not work or anything.

Dang, wish I'd known about this back when I first started - woulda saved me a bunch of effort. :stuck_out_tongue: Oh well, so it goes.

Errant-4:
I guess ideally there would be yaml support for only allowing a tile to switch between two of the four directions

Aearo-Deepwater:
Wait, before I get any farther along deleting all the duplicate-under-rotation tiles, is there a way for players to rotate them yet?
Because if not, there's the awkward situation where they might crowbar a tile and not be able to put it back the same way. I feel like that's a reasonable argument for keeping the duplicates.
@Errant-4 @slarticodefast

Aearo-Deepwater:
Still some misgivings, but it's done.
I hope tile objects become player-rotatable soon, because until they do, players will only be able to place vertical continuous slats.

Errant-4:
Please note that the main properties of old yaml prototypes are often in the wrong order. It's not a huge issues (which is why it so often goes ignored) but we should strive to keep things standardized. If you are copying an existing one as a template, you have to make sure that it matches the [conventions](https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#yaml-conventions).
If this PR did not involve so many changes in so many files already it would have been useful to fix all prototypes in that file to avoid this problem in the future, but as is, I will make that fix in a separate PR to keep this one simpler

Errant-4 (approved):

Aearo-Deepwater:
Fair enough. I shoulda double-checked that.
In my defense, my impression is that this file was changed semi-recently to add the new parent to everything; I think it was fair of me to assume that whoever did that had gotten the order right... I know not all the yaml follows the conventions correctly, but a big, recent change like this I'd figure would have.
All good, though. I'll change it.

Resources/Textures/Objects/Tiles/tile.rsi/meta.json
```diff
@@ -293,6 +305,12 @@
         {
             "name": "white-mono"
         },
+	{
```
beck-thompson: fix the spacing on this and also make sure to add yourself to credits

Aearo-Deepwater: Yup, I'm in 'em, just in the referenced other file.

beck-thompson: Sorry, I'm confused - why aren't you in this one?

Aearo-Deepwater: Yeah, sorry, that wasn't a very clear explanation.
It's in Resources/Textures/Tiles/attributions.yml, which is mentioned at the top of this meta.json as a place for additional attributions.

PR #37833, opened 2025-05-26
Fix the ability to hit and break overlapping walls

Quantum-cross:
## About the PR

Allows you to break walls that are overlapping on top of each other, like those created by rock anomaly pulses.

## Why / Balance

Bugfix / regression

## Technical details

Added a predicate in `SharedInteractionSystem.GetPredicate`.

This predicate checks the target for `Wall` tag, and will then add other `Wall` tagged entities on the same exact coordinates to the set of entities that are ignored for a user to interact with it.

## Media

https://github.com/user-attachments/assets/99397921-0b9c-4a12-83e2-1a13791f4e74

## Requirements

## Breaking changes

Negatory

**Changelog**

:cl:
- fix: You can break overlapping walls with melee weapons again!

### DIFF
--- a/Content.Shared/Interaction/SharedInteractionSystem.cs
+++ b/Content.Shared/Interaction/SharedInteractionSystem.cs
@@ -91,6 +91,7 @@ namespace Content.Shared.Interaction
         public const string RateLimitKey = "Interaction";
 
         private static readonly ProtoId<TagPrototype> BypassInteractionRangeChecksTag = "BypassInteractionRangeChecks";
+        private static readonly ProtoId<TagPrototype> WallTag = "Wall";
 
         public delegate bool Ignored(EntityUid entity);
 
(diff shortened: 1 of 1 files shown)


VerinSenpai:
does this also apply to meteors?

VerinSenpai:
Oh my God praise be

metalgearsloth:
Isn't this going to make every single interaction in the game incredibly expensive

Quantum-cross:
I understand, Ill look into a better way to fix this, I found the code path that causes it at least

BackeTako:
> What other cases did this big happen with? I was only aware with walls

I am pretty sure ive seen the meat anomally spawn flesh blockers ontop of eachoher that have the same issue

Quantum-cross:
the problem can still occur at least with the asteroid wall if the anomaly is pulsed rapidly, as the asteroid wall does not become hard until a few seconds, thus that tile can be a valid spawn location for another asteroid wall.

In most gameplay situations this should be an issue, but it's a pretty effective bandaid, and the fix makes sense in terms of `CanSpawnOnEntities`, if it's spawning two on one tile then it's kind of breaking that contract by itself...

cmorley191:
Is this still being worked on? The current state of the fix looks good to me; would correctly prevent duplicate spawns from anomalies.

TheShuEd (approved):

Quantum-cross:
@EmoGarbage404 the code that you originally commented isn't relevant anymore. The scope of this PR is drastically reduced but should fix a very common cause of anomalies spawning things on top of each other.

PR #32130, opened 2024-09-13
Remove player workarounds for PowerGridCheckRule

eoineoineoin:
## About the PR

Adds extra handling to events that could occur during a PowerGridCheckRule.

## Why / Balance

During a PowerGridCheckRule, players can still re-enable power which had been remotely turned off in two ways:

- A player with sufficient access can just enable an APC
- All players can disassemble and re-assemble the APC

These actions are kinda cheat-y and against the rules, but are _real_ common to see during a game. 

## Technical details

Adds some callbacks to handle APC construction and APC breaker toggle events. Does not allow the APC to be powered on in either of the above cases.

Needed to make a new dummy component (PowerGridCheckNotifyComponent) and added it to the APC prototype - not sure if there was a better way to do this, but couldn't subscribe to the events for the ApcComponent itself, as that was causing duplicate subscriptions.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- fix: Fixed bug where APCs could turn on during a station power-off event

PR #37841, opened 2025-05-26
Power outages are shorter and only affect the main (largest) grid

Krunklehorn:
## About the PR
Title. Leaves the ATS, cargo shuttle and others functional.
Shaves off 10-20 seconds to prevent affecting the gravity generator.

## Why / Balance
Originally did what #32130 was, then more features were requested.

## Technical details
`PowerGridCheckRule` uses `StationSystem:GetLargestGrid()` when checking if an APC should be affected.

## Requirements

## Breaking changes
`ApcSystem:OnApcStartup()` is no longer static.

**Changelog**
:cl:
- fix: APC power outage events now only affect the main station grid.


PJB3005 (changes requested):

Krunklehorn:
I tried, but it still says "Failed to download action"

LaCumbiaDelCoronavirus:
> Alternative PR doing the same thing differently: #32130

It hasn't been touched since last year and still has a year-old merge conflict..
However I'd prefer the solution in that PR (which is easier to work with) over this one which could be improved

pheenty:
Tbh I think this shouldn't be changed; but APCs in general should shock you when crowbaring them without insulated gloves.
This way regular personnel still cannot bypass it, but engineers can fix something if really needed.

Krunklehorn:
> > Alternative PR doing the same thing differently: #32130
> 
> It hasn't been touched since last year and still has a year-old merge conflict.. However I'd prefer the solution in that PR (which is easier to work with) over this one which could be improved

> Tbh I think this shouldn't be changed; but APCs in general should shock you when crowbaring them without insulated gloves. This way regular personnel still cannot bypass it, but engineers can fix something if really needed.

Either way works. Just seems like a noob trap not knowing 2/3s of the station is likely going to re-power early. Antagonists should feel more confident in the dark.

Krunklehorn:
> Please make the power event slightly shorter then, it can last long enough to disable gravity.
> 
> The even also disables ATS and shuttles annoyingly enough.

Checking for `BecomesStationComponent` on grid works. Tested on Amber in release mode and ATS, cargo shuttle etc. remained functional.

Is there a better alternative?

LaCumbiaDelCoronavirus:
> Is there a better alternative?

maybe StationMemberComponent

PicklOH:
> Tbh I think this shouldn't be changed; but APCs in general should shock you when crowbaring them without insulated gloves. This way regular personnel still cannot bypass it, but engineers can fix something if really needed.

APCs just need wire interfaces like doors that you need to snip if you wanna tamper with em.

Krunklehorn:
Ready once again.

Content.Server/Power/EntitySystems/ApcSystem.cs
kontakt, line 1: I don't know if any of the changes in this file are still relevant, I think their cases have been addressed.
Keeping the cleanup would be good though.

Content.Server/StationEvents/Events/PowerGridCheckRule.cs
kontakt, line 1: I don't think this approach is ideal.
Could you look at how the station target is chosen and see if there's a good way to add the `BecomeStation` check there?

kontakt (dismissed):

Content.Server/StationEvents/Events/PowerGridCheckRule.cs
Krunklehorn: You're right, `BecomesStation` doesn't propagate when a grid splits. Why don't we enable that and rename it to something more appropriate?

Krunklehorn: Alternatively, here is `GridEventEligible` component with true-by-default flags, propagated on station split. There are a variety of events which may or may not wish to configure themselves on a per-grid basis and the mapping effort should be one or two lines.

Krunklehorn: We decided to simply check for the largest grid: https://discord.com/channels/310555209753690112/1532587342983073833

EthanQix: Is that not going to be a problem with VGroid, or when the station is split into chunks due to a major accident ?

K-Dynamic: How about check for the one with the most players

AJCM-git:
could this potentially roll it only on flavor grids that are larger than the station but part of it?

Content.Server/StationEvents/Events/PowerGridCheckRule.cs
Krunklehorn: > Is that not going to be a problem with VGroid, or when the station is split into chunks due to a major accident ?

There is no StationMember component on the VGroid or grids spawned by salvage magnets. I've been testing this on release Elkridge.

Krunklehorn:
> could this potentially roll it only on flavor grids that are larger than the station but part of it?

I'm not sure. The vgroid and salvage magnet pulls don't have `StationMember` but I don't know what you mean by flavor.

PR #37236, opened 2025-05-06
Traumoxadone

ScarKy0:
## About the PR

Added Traumoxadone for healing brute damage in dead patients.

## Why / Balance

Currently the only way to properly heal someone dead is via spending 50 bruisepacks on them and hoping you can revive them afterwards. Adding a cryochem capable of doing this would encourage medical to actually learn how cryogenics work, which seem to be massively underused currently.
Also in an ideal world cryogenics would be used mainly for curing dead people. Shoving someone into a cryotube while theyre alive is just kinda annoying due to having to wait to heat up, so nobody wants to. Another step is adding stellox and making cryox only heal asphyx and bloodloss.

## Technical details

Yaml only.

## Media

image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- add: Added Traumoxadone! A cryochemical used to treat brute damage in living patients as well as corpses.

PR #37844, opened 2025-05-26
HOTFIX Cryochem revert

Errant-4:
## About the PR
Reverts #37131 #37236 and #37132

## Why / Balance
Maintainer vote decided that the new chems should not work on the dead, however without that ability, there was little use to have them. Cryoxadone was also changed to work on the dead but also essentially nerfed to accomodate the new chemicals (and fill a new niche), so without them we need it back as it was, so that's also a revert rather than an edit.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Changelog/Changelog.yml
+++ b/Resources/Changelog/Changelog.yml
@@ -3661,33 +3661,6 @@
   url: https://github.com/space-wizards/space-station-14/pull/37069
 - author: ScarKy0
   changes:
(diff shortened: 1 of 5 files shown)


cfif126:
deadass

Awlod:
Well, it's stupid, first because of stupidity on the part of the main developers, we lost the person who was involved in newmed, and now because of "voting support" we are losing the treatment of corpses from the same poisons. This is nonsense.

Issue #37694, opened 2025-05-21
Storages sometimes erroneously fail to place


EmoGarbage404:
## Description

Sometimes very full storage containers will have placement failures along the very edges. I only could get this to replicate with duffels and satchels (which exceed the one chunk boundary) so perhaps this has something to do with how the quick placement encodes chunks.

**Reproduction**

- get satchel
- fill up entirely with small items (1x2)
- notice that the last item will not place. This appears to happen for the whole rightmost column. Placing it elsewhere works fine.

**Screenshots**

![Image](https://github.com/user-attachments/assets/63fd3806-ae49-4319-a5a4-34d41e56584e)
example of storage situation that fails. The ui also doesn't match the logic: since it shows it as placeable.

**Additional context**

Appeared after #37638

PR #37852, opened 2025-05-26
Fix storage fastpath logical oversight

EmoGarbage404:
## About the PR

Second fix for #37694

fixes issues where some weirdly shaped objects (mostly guns) would not fit in storage.

## Technical details

The logical precondition for the fastpath is that the item being placed will always occupy the top-left spot in storage. This was being checked by seeing if any of the rectangles in the item's shape occupy 0,0. 

This doesn't actually verify correctly, however. If an item has an L-shape, then the item can be oriented such that the "cutout" of the L is in the top-left spot. This causes the fastpath logic to run, which is incorrect in this situation.

What we actually need to do is verify that the shape is concave (which i do by checking just the shape being a plain rectangle)

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Fixed a bug that would stop guns and other irregularly shaped objects from fitting into storage.

### DIFF
--- a/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
@@ -1326,7 +1326,7 @@ public abstract class SharedStorageSystem : EntitySystem
         // This uses a faster path than the typical codepaths
         // as we can cache a bunch more data and re-use it to avoid a bunch of component overhead.
 
-        // So if we have an item that occupies 0,0 we can assume that the tile itself we're checking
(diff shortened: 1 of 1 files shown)


slarticodefast:
How often does this happen? Do we need to hotfix?

Tayrtahn (approved):

EmoGarbage404:
> How often does this happen? Do we need to hotfix?

it's not really the most common issue: it only happens with a small subset of shapes (the l-shapes used only for handguns) and in a single orientation. it's also not hard to really work around. i could go either way on hotfix.

slarticodefast (approved):

PR #35341, opened 2025-02-20
Decouple Lights from Toggleable Visuals (and headphone music notes bugfix)

Centronias:
## About the PR
Fixed headphone music note visuals.

... by doing a refactor wherein I decoupled `ToggleableLightVisuals` from lights.

## Why / Balance

- Fixes the headphones visuals bug ( resolves #31755 )
- Makes it so that item toggling can change visuals of an entity that doesn't have lights

## Technical details

- Combine enum keys `ToggleableLightVisuals` and `ToggleVisuals` into `ToggleableVisuals`
- Rename `ToggleableLightVisualsComponent` to `ToggleableVisualsComponent` and `ToggleableLightVisualsSystem` to `ToggleableVisualsSystem`
  - (The `SpriteLayer` field on the component is now required because the old default of `light` doesn't make sense anymore) 
- Make it so that `ToggleableVisualsComponent` works even when there's not a light attached to the entity
  - (Amazingly this seems to have only applied to  Headphones, but I can only imagine there are many other things people would like to do with simple toggleable visuals)
- Explicitly make `ItemTogglePointLightComponent`'s purpose to make `ToggleVisualsComponent` apply to `PointLightComponent`s on the same entity.
  - Add field `ToggleableVisualsColorModulatesLights`, which makes the `Color` appearance value of `ToggleableVisuals` modulate the color of lights on the same entity
  - Lots of prototype updates to uptake the above

## Media

https://github.com/user-attachments/assets/f790ba0a-fd07-42fb-b541-1831c95e5b7c

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

### `ToggleableLightVisuals` and `ToggleVisuals` replaced by `ToggleableVisuals`

Apply the following replacements:
- `ToggleVisuals.Toggled` -> `ToggleableVisuals.Enabled`
- `ToggleVisuals.Layer` -> `ToggleableVisuals.Layer`
- `ToggleableLightVisuals.Enabled` -> `ToggleableVisuals.Enabled`
- `ToggleableLightVisuals.Color` -> `ToggleableVisuals.Color`

### `ToggleableLightVisualsComponent` replaced by `ToggleableVisualsComponent`
[...]

PR #37853, opened 2025-05-26
Fixes the telescopic shield visual bug

SpeltIncorrectyl:
## About the PR

Fixed a bug where the telescopic shield's sprite would not change when you toggle it.

## Technical details
TLDR: The `ToggleableLightVisualsSystem` is in a bad spot right now, as it does a lot more than just lights.
There is a proper pr #35341 to uncouple the system but that is 3 months old and not jet merged - and I realised I could very quickly fix the telescopic shield bug in an afternoon by changing 2 lines of code.

The telescopic shield and the energy sword work the exact same way, via the `ToggleableLightVisualsSystem`, and they both change sprite when toggled. Yet the energy sword works and the telescopic shield does not. Why is this?

In order for the item to change it needs the `ToggleLightsVisuals.Enabled` appearance data to be set. It is set a few times in the code (like when people get set on fire, another of the many many things this one system does :grimacing:). But what we need is it to be set in response to an item being toggled, and this only happens at once place. In the `ItemTogglePointLightSystem`, and only for items with point lights. The energy sword has a point light but the telescopic shield doesn't. That's it, that's the bug.

So the bug is that the `ToggleLightsVisuals.Enabled` is set in the wrong place in the code for some items and not others. So I removed the line that sets it from `ItemTogglePointLightSystem`, but I didn't know where to put it. My first thought was for it to be in the `ToggleableLightVisualsSystem` but that is Client and the data has to be set in Shared or it will mispredict. So I set it in the `UpdateVisuals` method in the `ItemToggleSystem` alongside where `ToggleVisuals.Toggle` is set. I think it makes the most sense there.

When I first tried to fix this bug I thought I should refactor it to use `ToggleVisuals.Toggled` instead of `ToggleableLightVisuals.Enabled` - but that can't work because this stuff is also used for people burning. Refactoring it so people who are burning alive are *toggled on* is an utterly insane idea and one I abandoned.

This whole thing needs a big proper refactor which is already in progress via #35341. However that pr is taking a very long time to review and I was playing yesterday and the teleshield bug was bugging me so I just fixed it with two lines of code.

## Media

Here are images of the telescopic shield working! Hallelujah!
Screenshot from 2025-05-26 18-12-30
Screenshot from 2025-05-26 18-12-37

## Requirements

**Changelog**

:cl:
- fix: Fixed the telescopic shield visual bug - it now changes sprite when toggled!

### DIFF
--- a/Content.Client/Toggleable/ToggleableLightVisualsSystem.cs
+++ b/Content.Client/Toggleable/ToggleableLightVisualsSystem.cs
@@ -8,6 +8,7 @@ using Content.Shared.Toggleable;
 using Robust.Client.GameObjects;
 using Robust.Shared.Utility;
 using System.Linq;
+using Content.Shared.Item.ItemToggle.Components;
 
 namespace Content.Client.Toggleable;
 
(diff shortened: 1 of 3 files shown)


SpeltIncorrectyl:
> I feel like #35341 is a more comprehensive solution to what this PR is doing.

That's objectively true but this pr is 3 diff and will fix the bug right now. I don't know how long it will take to merge #35341 when it's been open for 3 months.

SpeltIncorrectyl:
#35341 can just be merged after this whenever it is ready, I added a line of code that they will have to remove but that's it.
There is no dichotomy here when this is tiny 3 diff fix.

PR #37857, opened 2025-05-26
No-Slowdown Raidsuit

Entvari:
## About the PR
Removes the 10% slowdown from the Syndie Raidsuit.

## Why / Balance
The Raidsuit has been nonviable in a vast majority of Nukie scenarios for ages, it's most defining trait; the fact it's not spaceproof, results in it being pretty horrible outside of a few very fringe scenarios, such as Blitzops or Crew Disguised Stealthops, though even in these cases it's hard to argue its better than its alternatives (Regular Hardsuits and Webvest).
This PR aims to help alleviate the issue of it being borderline redundant by taking away its slowdown, this puts it at the same speed as the Webvest whilst providing more generalist defensive stats as opposed to being entirely focused on Pierce.
Thanks to https://github.com/space-wizards/space-station-14/pull/35890 Raidsuits are also inherently bundled with the syndie backpack, this should allow the Raidsuit to serve as a defacto "Move fast" alternative to hardsuits.
Unfortunately though Raidsuit will likely still remain an "alternative" instead of a viable go-to, purely because its defining feature is pretty awful, changing Raidsuit's spacing properties would be changing Raidsuit into something completely different altogether, though hopefully this PR should encourage its use a bit more.

## Technical details
All YML, Ymlwarrior strikes again.

## Media
None

## Requirements

## Breaking changes

None

**Changelog**

:cl:
- tweak: Raidsuit no longer has slowdown.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/armor.yml
@@ -240,8 +240,8 @@
   #- type: StaminaResistance
   #  damageCoefficient: 0.45
   - type: ClothingSpeedModifier
-    walkModifier: 0.9
-    sprintModifier: 0.9
+    walkModifier: 1.0
+    sprintModifier: 1.0
   #Shoulder mounted flashlight
   - type: ToggleableLightVisuals
     spriteLayer: light


SpeltIncorrectyl:
Read this as radsuit and was really confused lol

notquitehadouken (approved):
Test fail unrelated

keronshb:
tbh I'd rather remove raidsuits and just give the hardsuits the speed/armor/spacing toggle.

LittleNorthStar:
> tbh I'd rather remove raidsuits and just give the hardsuits the speed/armor/spacing toggle.

If it's that easy people will just swap suit mode the second there's spacing removing the risk to the reward, although it would be nice to see some method to pick between hardsuits and raidsuits on ops start

ArtisticRoomba:
> tbh I'd rather remove raidsuits and just give the hardsuits the speed/armor/spacing toggle.

This sounds very close to modsuits, this should see more discussion.

keronshb:
> > tbh I'd rather remove raidsuits and just give the hardsuits the speed/armor/spacing toggle.
> 
> This sounds very close to modsuits, this should see more discussion.

It's not modsuits at all. It was an action toggle.

TPgameing:
There's a very good reason i am against this.
The Raidsuit has 70% Protection against pierce, it would be the equivalent of if the full team was wearing Commander Hardsuits before the nerf, the raid suit is already **REALLY** good, i think the 10% speed penalty is a good balance to stop it from being absolutely broken.

Pronana:
> There's a very good reason i am against this. The Raidsuit has 70% Protection against pierce, it would be the equivalent of if the full team was wearing Commander Hardsuits before the nerf, the raid suit is already **REALLY** good, i think the 10% speed penalty is a good balance to stop it from being absolutely broken.

I agree this makes the raid suit extremely good, but that's why I'm also for it.

Sacrificing spacing gives the crew a big advantage as well, they are able to wear their better cheaper armor and also move faster in response, plus it means the Nukies are able to use explosives less effectively without screwing themselves over. 

They should get a really good suit of armor for having to sink a good TC cost and also deal with new downsides as a result.

Entvari:
> There's a very good reason i am against this. The Raidsuit has 70% Protection against pierce, it would be the equivalent of if the full team was wearing Commander Hardsuits before the nerf, the raid suit is already **REALLY** good, i think the 10% speed penalty is a good balance to stop it from being absolutely broken.

Raidsuit currently serves basically no purpose because of the spacing weakness, 8tc for what is essentially a Webvest with more general damage resist that additionally slows you, and there's good reason you never see Nukies running Webvest despite its great stats.
Spacing makes Raidsuit extremely weak because in almost all Nukie environments its inevitable, spacing damage quickly racks up blunt, which means if you're tended for that you can't be tended for pierce, and vice versa.

As it stands the spacing weakness alone makes Raidsuit nowhere near worth picking, and that massive downside will remain provided this PR gets merged, but it'll receive slightly more mobility to compensate for the fact it has this.
If you're sacrificing the single most powerful aspect of a hardsuit, you deserve to be compensated handsomely by the defense stats.

TPgameing:
> > There's a very good reason i am against this. The Raidsuit has 70% Protection against pierce, it would be the equivalent of if the full team was wearing Commander Hardsuits before the nerf, the raid suit is already **REALLY** good, i think the 10% speed penalty is a good balance to stop it from being absolutely broken.
> 
> I agree this makes the raid suit extremely good, but that's why I'm also for it.
> 
> Sacrificing spacing gives the crew a big advantage as well, they are able to wear their better cheaper armor and also move faster in response, plus it means the Nukies are able to use explosives less effectively without screwing themselves over.
> 
> They should get a really good suit of armor for having to sink a good TC cost and also deal with new downsides as a result.

Space protection shouldn't be required if you are doing raid ops or anything fast, as space breaches won't likely have time to form.
And for crew, any nuclear operative armor is really advantageous, so them getting a raid suit is no different.
And lasty, one of my friends pointed out, The crew also gets very little space protection, and would much more rather also fight in a pressurized environment.

EmoGarbage404:
I'm kinda dubious as to whether this PR effectively addresses any of the issues with raidsuits.

I think one of the main downsides to this item is the level of commitment you have to put in for it to work: the entire team needs to buy them and you have to hard commit to a single strategy with basically no backup plan. Making the suit better doesn't really address this and just makes it a more viable strategy _if_ the team decides to commit to it (in which case, do we really want to push nukies towards hyper-aggro blitzops?).

Internally, i consider this change similar to the various sus toolbox buffs: it doesn't really matter how good you make it because people just aren't interested in the concept that it's offering. I don't think trading space-proof nukies for a bag and better armor is interesting or particularly viable to begin with, and I'd rather see a more comprehensive redesign of this to make it a more sane pick.

PicklOH:
Agree with everything the author has stated here. I've done Raidsuit Ops very recently and it was mediocre at best. You really are losing more than you're gaining generally with the Raidsuit, so buffs like this should help offset that balance. If I were at the helm of this PR I would even push it a bit further and prehaps give it a cost decrease.

I do not think this minor buff will push it over the edge or force us into a BlitzOps meta at all.

Seeing the comments asking for a redesign, I see where you're coming from, but I think asking for a total redesign is a bit much. I prefer small buffs to this item to enable it to better fill that small niche it has as tankier, not-so-space-proof Nukies as opposed to anything more or less.

Point blank, the Raidsuit exists as a Worse than Hardsuit, Not Much Better than Webvest buy and this buff helps give it a bit more oomph. I see no reason not to let this change through.

Entvari:
> With no slowdown it's just directly better version of any crew member armor, including very limited captain's carapace

Crew has hardsuits so this isn't even true, but it's already the case with bloodreds anyways like what lol??

PubliclyExecutedPig:
I'd just like to comment as someone who plays starlight, a server which has already buffed raidsuit, even more than what this does

its not broken and still barely gets used lol, its more of a loneops buff tbh

edit: i realise the balance is different so this may not apply, but in i think its fine still

superjj18:
Another good buff would be(if it doesn't already) cold protection.  This would allow them to quickly traverse outside the station for a few seconds without being crippled by cold(would reduce their ability to vent internal heat so might wanna add a very minor heat resistance to compensate).

This would mean that yes, it doesn't have spacing protection, at very least you can sacrifice health from barotrauma and air loss for effective movement through spaced areas.

AJCM-git:
Discussion stalled for a year, closing for now.

PR #37861, opened 2025-05-26
Clastine, Explosive Syndie Chem

Entvari:
## About the PR
Introduces a new Vestine product, Clastine, made with 2:2:1 Chlorine:Plasma:Vestine.
Has low damage explosive properties and reacts with Chlorine, but also metabolizes INTO Chlorine.
This creates a functionally unique explosive chemical that autoignites upon being injected into someone, creating several small, low damage explosives.
Both possible to synthesize and purchase in uplink, far cheaper to do the former.

## Why / Balance
Offensive Syndiechems are overall pretty dry and not very fun to use. Nocturine is an instant victory and Lexorin is underwhelming.
Clastine serves to exist as a functionally unique pseudo-stealth explosive designed for killing single targets, not for gibbing or inflicting damage unto the area.
The 30U bottle costs 8TC from uplink, with a rare discount reducing it to 4TC.
The synthesis cost is somewhat expensive, requiring droves of plasma and chlorine, though using Vestine bundles would allow you to acquire considerably more reagent than purchasing it pre-made, if you're willing to extend the effort.

10u of Clastine is enough to kill an unarmored target, anything extra does nothing as it does not react on corpses.

Clastine is pretty much undeniably outclassed by its alternatives (primarily Nocturine and other single target explosives like Minibombs or Penbombs), though I feel it carves enough of its own niche to warrant inclusion.

Damage and ease of acquisition may need tinkering with depending on its functionality in practice, though from my testing it seems to be okay.

## Technical details
all YML

## Media
clastine2
clastine1

## Requirements

## Breaking changes
None

**Changelog**

:cl:
- add: Added Clastine, a new Syndicate pyrotechnic chemical purchasable in the uplink for 8TC.

### DIFF
--- a/Resources/Locale/en-US/reagents/meta/pyrotechnic.ftl
+++ b/Resources/Locale/en-US/reagents/meta/pyrotechnic.ftl
@@ -18,3 +18,6 @@ reagent-desc-welding-fuel = Used by welders to weld.
 
 reagent-name-fluorosurfactant = fluorosurfactant
 reagent-desc-fluorosurfactant = A perfluoronated sulfonic acid that forms a foam when mixed with water.
+
+reagent-name-clastine = clastine
+reagent-desc-clastine = An otherwise inert chemical that explodes violently when in contact with Chlorine, ironically that's what it metabolises into. Mix with care.
--- a/Resources/Locale/en-US/store/uplink-catalog.ftl
+++ b/Resources/Locale/en-US/store/uplink-catalog.ftl
@@ -387,6 +387,9 @@ uplink-combat-medipen-desc = A single-use medipen containing chemicals that rege
 uplink-nocturine-chemistry-bottle-name = Nocturine Bottle
 uplink-nocturine-chemistry-bottle-desc = A chemical that puts your target straight to sleep.
 
+uplink-clastine-chemistry-bottle-name = Clastine Bottle
+uplink-clastine-chemistry-bottle-desc = A medium power explosive that autoignites when metabolised, use with care.
+
 uplink-stimpack-name = Hyperzine Injector
 uplink-stimpack-desc = The legendary chemical produced by Donk Co. for the Syndicate. Injecting yourself with this will increase your run speed and let you recover from stuns faster for 30 seconds.
 
--- a/Resources/Prototypes/Catalog/uplink_catalog.yml
+++ b/Resources/Prototypes/Catalog/uplink_catalog.yml
@@ -642,6 +642,19 @@
   - UplinkChemicals
 
 - type: listing
+  id: UplinkClastineChemistryBottle
+  name: uplink-clastine-chemistry-bottle-name
+  description: uplink-clastine-chemistry-bottle-desc
+  productEntity: ChemistryBottleClastine
+  discountCategory: rareDiscounts
+  discountDownTo:
+    Telecrystal: 4
+  cost:
+    Telecrystal: 8
+  categories:
+  - UplinkChemicals
+
+- type: listing
   id: UplinkCombatMedkit
   name: uplink-combat-medkit-name
   description: uplink-combat-medkit-desc
--- a/Resources/Prototypes/Entities/Objects/Specific/chemistry-bottles.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/chemistry-bottles.yml
@@ -238,6 +238,21 @@
           Quantity: 30
 
 - type: entity
+  id: ChemistryBottleClastine
+  suffix: clastine
+  parent: BaseChemistryBottleFilled
+  components:
+  - type: Label
+    currentLabel: reagent-name-clastine
+  - type: SolutionContainerManager
+    solutions:
+      drink:
+        maxVol: 30
+        reagents:
+        - ReagentId: Clastine
+          Quantity: 30
+
+- type: entity
   id: ChemistryBottleEphedrine
   suffix: ephedrine
   parent: BaseChemistryBottleFilled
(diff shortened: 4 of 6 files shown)


LevitatingTree:
the name is great

Aearo-Deepwater:
Should be "An otherwise inert chemical that explodes violently when in contact with chlorine; ironically, that's what it metabolises into."
And while I'm being pedantic, that's not really irony. "Conveniently" or something might be a better word choice.
🤓

Entvari:
God damn merge conflicts
> Should be "An otherwise inert chemical that explodes violently when in contact with chlorine; ironically, that's what it metabolises into." And while I'm being pedantic, that's not really irony. "Conveniently" or something might be a better word choice. 🤓

The semicolon is justified, fuck my stupid illiterate life, but i'm intending to keep the 'ironic' because the guidebook is generally not written from the perspective of a terrorist, CLF3 for instance says you shouldn't get it near you, rather than "You can use this to blow people to smithereens!", i'll fix the former grammar issue when i get round to fixing the conflict.

K-Dynamic:
Tested and works ingame

I will say it's rather powerful but that's more a problem with chems, hypos and metabolism in general.

Test with 5u, purchased from syndie and nukie uplinks
image

Test with 30u
image

Chemical process
dotnet_XldP6wli0a

iaada:
I think it's a really cool chemical. I like seeing more recipes with vestine and plasma. A straight bomb mix of clastine + chlorine is twice as strong as a potassium bomb of the same amount, which is fair enough given the rarity. But the damage when injected is just too high. 5 units to kill a civilian and 7 to kill a captain is too much. That's 6 civilians dead for 6tc.

## Nocturine
Time to "crit" = 6 seconds as of #40231

## Clastine
| Volume | Damage | TTC | TTK |
| --- | --- | --- | --- |
| Urist | ~ | ~ | ~ |
| 5u | 105 | 3 sec | N/A |
| 10u | 225 | 3 sec | 5 sec |
| 15u | 225 | 3 sec | 5 sec |
| Captain | ~ | ~ |
| 5u | 72 | N/A | N/A |
| 7u | 103 | 4 sec | N/A |
| 10u | 146 | 4 sec | N/A |
| 15u | 204 | 4 sec | 7 sec |
| Bloodred | ~ | ~ |
| 5u | 56 | N/A | N/A |
| 10u | 112 | 5 sec | N/A |
| 15u | 170 | 5 sec | N/A |

## Lexorin
| Volume | Damage | TTC | TTK |
| --- | --- | --- | --- |
| Any Target | ~ | ~ |
| 5u | 92 | 11 sec | N/A |
| 10u | 200 | 11 sec | 21 sec |
| 15u | 200 | 11 sec | 21 sec |

Entvari:
I scaled the chemicals usability with the other two super lethal vestine chems, nocturine and lexorin.
Lex does a flat 10 damage per .5u, which puts it ever so slightly lower damage than a Clastine dose of the same quantity, however unlike Clastine it is in no way silent, and you are given less time to register what's actually killing you. Furthermore, Clastine is the only chemical actually subject to damage resistances, making it far less effective against high profile armored targets like Cap or HOS and especially CE and RD compared to it's contemporaries.

The same goes for Nocturine, Noct effectively instantly kills with no counterplay, unless you have a godlike reaction time to hearing a prick noise and begging to be saved on common, there's nothing you can do, but Clastine when injected still takes a full time to kill, is innately loud, and immediately obvious what it is.

To save yourself from Noct you have to immediately clock that someone has hypo'd you and call for help, even still you're unlikely to be saved.
To save yourself from Lexorin you need to register someone has hypo'd you, and notice that your health is draining to asphyx before five seconds pass, once again even still you're not at any gaurantee.
To save yourself from Clastine you have the same situation as Lexorin, however rather than silent asphyxiation damage, you're taking loud, obvious and attention grabbing explosive brute damage, you probably wouldn't even need to call out on radio for someone to get concerned about what just happened.
Someone can be Clastine'd and is probably alot more likely to be saved before its too late, this isn't really the case with Lex and especially with Noct.

My justification for Clastine's niche is that it is loud, easy to recognize due to distinct evenly spaced explosions, however is more effective in small doses, angling it to be better for loud combat with hypodarts, syringegun, or god-forbid running at people with a hypo. Of course, you could still use it in a stealth setting, but it'd be arguably worse than Lexorin and objectively worse than Nocturine.

iaada:
You're right that lexorin and nocturine are fair comparisons. I updated my chart above to better compare. The fundamental differences here are in the time to crit.

This chem (sometimes) deals less and is much louder than lexorin. But it crits/kills far faster, and lexorin is not available in an uplink. Nocturine's time to "crit" has been a known problem, and as of #40231 its "TTC" is 6 seconds - longer than it takes Clastine to kill.

In comparison to the other two syndie chems, Clastine gives you half as much time to react as Nocturine, and 1/4th as much time as Lexorin. The exploding pen (another unavoidable explosive) gives you 4 seconds to react - slightly longer than Clastine - and costs 4 TC for one kill compared to Clastine's 1 TC per kill. Looking at these numbers, I would argue that Clastine is *more* lethal than the other two chems, not less.

I still think it's a very cool chemical, but I also think it kills far too quickly and cheaply.

SuperGDPWYL:
imo the time to crit should be around 6 seconds for an unarmored target., but it should be stronger than lexorin given it gives an indicator of exactly what's happening to everyone who sees it, unlike lexorin where you need to scan them to realise they were poisoned

K-Dynamic:
I feel like any concerns regarding clastine are really about the hypopen having a 0.5s delay between injections.

iaada:
I still find this chem very cool but I only see it existing outside of a direct uplink purchase. I'd like to see it as something a chemist has to actually synthesize, and the time to kill also needs to be weakened. If my concerns are addressed please open a new PR. Thank you for your contribution ❤️

K-Dynamic:
> ed. If my concerns are addressed please open a new PR. Thank you for your contribution ❤️

I feel like they could address the changes in this PR rather than opening another one.

PR #37866, opened 2025-05-27
reduced the number of slots in bookshelves

Moomoobeef:
## About the PR
Currently, bookshelves have an absurd amount of slots, which is strange for something that can only hold books and paper. I reduced the slot-count to 40, and since one book takes up 2, and there are 20 fill-level sprites (not including empty) that corresponds to one sprite per book.

In addition, I also increased the number of books that randomly-filled bookcases contain. I did this so there will be fewer bookcases with just a couple, or even just one book in them. Now librarians will feel less compelled to dismantle all but one bookshelf.

## Why / Balance
It looked ugly honestly. Also I didn't like that adding/removing books to/from the shelf did not always correspond to a sprite change.

## Media
books

## Requirements

## Breaking changes
shouldn't be any

**Changelog**
:cl:
- tweak: Bookshelves now contain a more reasonable amount of books.

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Books/bookshelf.yml
+++ b/Resources/Prototypes/Catalog/Fills/Books/bookshelf.yml
@@ -16,14 +16,14 @@
     # Randomly generated books
     - id: BookRandomStory
       amount: !type:RangeNumberSelector
-        range: 0, 6
+        range: 4, 8
     # Guidebooks
     - !type:NestedSelector
       tableId: RandomGuidebookTable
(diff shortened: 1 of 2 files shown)


IProduceWidgets:
I almost want to say remove like 3/4 of the fill levels in yml and reduce it even more. I've never seen anyone interact with more than like 3 books in a round.

slarticodefast (approved):
Makes sense to me!

beck-thompson (approved):
makes sense!

PR #37869, opened 2025-05-27
high-vis vest is now dimmer

Moomoobeef:
## About the PR
The high-vis vest currently is absolutely *dazzling* in its brightness. You are a fucking spotlight in all but the brightest areas. So I made it less bright

## Technical details
As you may already know, we do not have emissive maps. Things are "unshaded" in order to glow, and unshaded means FULLBRIGHT, no lighting. I found a hack though. You see, you can have a transparent sprite that *is* shaded, over the top of an unshaded sprite, and it will occlude, including partially. I also discovered that the high-vis vests unshaded sprite is already transparent (probably because it came from TG where they do have working emissive maps), so I just put it over top of itself; and voila. I basically hacked in an emissive map, though admittedly the emissive level is inverted (but I didn't change it because I like how it looks already)

So now, you can make things that are unshaded have an "emissive map", just add an alpha channel to the unshaded sprite, and whatever emissiveness you want, invert that (since it works subtractively, not additively). then have two layers of this sprite, with one being unshaded and the one on top of it being shaded.

I'm sure there are problems with this approach but I haven't found them ~~so lets go get some technical debt aylmao~~

## Media
image
image
image

## Requirements

## Breaking changes
shouldn't be any

**Changelog**

:cl:
- tweak: high-vis vests are dimmer now.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/vests.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/vests.yml
@@ -13,6 +13,8 @@
     - state: icon
     - state: icon-unshaded
       shader: unshaded
+    - state: icon-unshaded
+      shader: shaded
   - type: Clothing
     sprite: Clothing/OuterClothing/Vests/hazard.rsi
     clothingVisuals:
@@ -20,6 +22,8 @@
       - state: equipped-OUTERCLOTHING
       - state: equipped-OUTERCLOTHING-unshaded
         shader: unshaded
+      - state: equipped-OUTERCLOTHING-unshaded
+        shader: shaded
   - type: Tag
     tags:
     - HiViz


Moomoobeef:
actually I might make them dimmer... what do ya'll think?

Moomoobeef:
> I think I liked the full bright better, felt like they look irl.
> 
> event-stewards-in-yellow-jacket-green-hi-vis-high-visibility-jackets-HAAMWG

this image was taken with flash. Most people do not have a flashlight on their head in real life, and so high-vis vests don't actually look like that most of the time.

The reason this is so bright in this photo is because of the afformentioned flash and because of how high-vis reflectors actually work, they are using retro-reflectors, which are so named because they reflect light back toward the direction it came from (which is also why we use them for road signage, that way a sign can appear highly illuminated just from a cars own headlights)

but to the human eye, unless you are wearing a headlamp, they don't actually look like this.

Moomoobeef:
ba-bam

slarticodefast:
Thanks!

PR #37879, opened 2025-05-27
Dwarf Accent for All

Entvari:
## About the PR
Dwarf accent has been added to the character customization screen for all species.
This does not effect Dwarf itself.

## Why / Balance
The Dwarf accent is closer to a regional dialect as opposed to a genetic way of talking, such as Lizard's dragged S.
As a Scottish person who inherently talks like a SS14 Dwarf IRL I thought it was a bit odd that the accent was restricted behind a species as opposed to being generally available like German or Spanish.
Additionally with frequent talk about Dwarf getting the axe for being too similar to humans, making this accent generally available better future-proofs it.
It'd also just be quite nice to add as another option for RP character customization due to how distinct and oddly content-packed this accent is.

## Technical details
All YML

## Media
dwarf3
dwarf1
dwarf2

## Requirements

## Breaking changes
None

**Changelog**

:cl:
- add: The Dwarf accent has been added to Character Customization for all species, Dwarves themselves are unchanged.

### DIFF
--- a/Resources/Locale/en-US/traits/traits.ftl
+++ b/Resources/Locale/en-US/traits/traits.ftl
@@ -60,5 +60,8 @@ trait-french-desc = Your accent seems to have a certain «je ne sais quoi».
 trait-spanish-name = Spanish accent
 trait-spanish-desc = Hola señor, donde esta la biblioteca.
 
+trait-dwarfaccent-name = Dwarven accent
+trait-dwarfaccent-desc = You cannae help speaking like ae proud Dorf.
+
 trait-painnumbness-name = Numb
 trait-painnumbness-desc = You lack any sense of feeling pain, being unaware of how hurt you may be.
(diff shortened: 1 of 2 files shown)


Entvari:
> mb also rename this to the Scottish accent?

This was a thought but it was mainly for parity with some of the accents replacements, some of which are explicitly Dwarven instead of Scottish, namely leaflover and greenbeard


> The problem is this is that it won't work with cloning or paradox clones as you can only have one replacement accent at a time and if you get zombified it gets overwritten. For dwarves this still works as the newly spawned Urist McDwarf defined in the species prototype starts out with the accent.

Daaamn okay I'll have a look at this but an accent rewrite would be considerably out of scope, especially considering how little I know of the codebase.

slarticodefast:
Any plans on further working on this? If not then I would close it for now and we can look at it again after the chat refactor.

slarticodefast:
Closing for now, maybe we can look at improving accents to be status effects in the future, which would solve the problem.

PR #37883, opened 2025-05-27
Add various commands for playing localized sounds

perryprog:
## About the PR

This adds two new commands—one toolshed, one... not, for playing sound at a given location or playing a sound from your attached entity respectively.
1. `/playlocalsound <path> [volume] [follow=false] [users...]` works like `/playglobalsound`, except it has a `follow` argument that controls if the played sound is parented to the command's runner. It additionally can take a filter of users who are able to hear the sound.
2. `playsound:` as a family of toolshed commands, with similar subcommands to `spawn:`. Note that the toolshed variant does not allow user filtering.

I also drive-by fixed a warning in PlayGlobalSound.

## Why / Balance

I uh. Like haunting my friends.

## Technical details

Pretty standard, but a few notes—
1. Sounds that are played "attached" to an entity are still subject to PVS visibility restrictions. This effectively means that you cannot, as a ghost, play audio that is attached to yourself and audible to non-ghost-seeing clients. I don't /love/ this, but I'm not sure how to get around it in a simple way. I'm not sure if this is where a PVS override or how to do that. Alternatively, just changing the visibilitymask on the spawned audio entity could be done, but that feels similarly janky.
2. No ResPath completion in Toolshed :(
3. I couldn't decide if the toolshed implementation should return the created audio entities or if it should pass through its input... I did the former for now since technically you can always `tee` around it if you need the latter behavior.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- add: /playlocalsound has been added for playing localized sounds. (+FUN)
- add: Toolshed commands for playing sounds at a point have been added. (+FUN)

### DIFF
--- a/Content.Server/Administration/Commands/PlayLocalSoundCommand.cs
+++ b/Content.Server/Administration/Commands/PlayLocalSoundCommand.cs
@@ -1,0 +1,129 @@
+using Content.Shared.Administration;
+using Robust.Server.Player;
+using Robust.Shared.Audio;
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Console;
+using Robust.Shared.ContentPack;
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.Administration.Commands;
+
+[AdminCommand(AdminFlags.Fun)]
+public sealed class PlayLocalSoundCommand : IConsoleCommand
+{
+    [Dependency] private readonly IEntityManager _entManager = default!;
+    [Dependency] private readonly IPlayerManager _playerManager = default!;
+    [Dependency] private readonly IPrototypeManager _protoManager = default!;
+    [Dependency] private readonly IResourceManager _res = default!;
+
+    public string Command => "playlocalsound";
+    public string Description => Loc.GetString("play-local-sound-command-description");
+    public string Help => Loc.GetString("play-local-sound-command-help");
+
+    public void Execute(IConsoleShell shell, string argStr, string[] args)
+    {
+        var filter = Filter.Empty();
+        var audio = AudioParams.Default;
+
+        var replay = true;
+        var followRunner = false;
+
+        var playerEntity = shell.Player?.AttachedEntity;
+        if (playerEntity == null)
+        {
+            shell.WriteLine(Loc.GetString("play-local-sound-command-no-entity"));
+            return;
+        }
+
+        if (args.Length == 0) // No arguments, show command help.
+        {
+            shell.WriteLine(Loc.GetString("play-local-sound-command-help"));
+            return;
+        }
+
+        // No players specified, so set filter here
+        if (args.Length <= 3)
+        {
+            // Filter.Broadcast does resolves IPlayerManager, so use this instead.
+            filter = Filter.Empty().AddAllPlayers(_playerManager);
+        }
+
+        // At least <path> [volume] specified
+        if (args.Length >= 2)
+        {
+            if (int.TryParse(args[1], out var volume))
+            {
+                audio = audio.WithVolume(volume);
+            }
+            else
+            {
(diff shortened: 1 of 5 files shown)


VerinSenpai:
My comments are merely suggestions of course, though if you don't do it here I'll be doing it down the line >.>

perryprog:
Still pretty new to the codebase and C#, hit me with 'em; it's greatly appreciated.

VerinSenpai (changes requested):

Content.Server/Administration/Commands/PlayLocalSoundCommand.cs
```diff
@@ -0,0 +1,129 @@
...
+[AdminCommand(AdminFlags.Fun)]
+public sealed class PlayLocalSoundCommand : IConsoleCommand
+{
+    [Dependency] private readonly IEntityManager _entManager = default!;
+    [Dependency] private readonly IPlayerManager _playerManager = default!;
+    [Dependency] private readonly IPrototypeManager _protoManager = default!;
+    [Dependency] private readonly IResourceManager _res = default!;
+
+    public string Command => "playlocalsound";
+    public string Description => Loc.GetString("play-local-sound-command-description");
```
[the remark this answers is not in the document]
VerinSenpai: Note I noticed the rest of the toolshed commands have their ftl done like this. Not sure why the naming convention is different here. May need input from a maintainer on this.

perryprog:
Didn’t know about LocalizedEntityCommands; happy to switch it over later today. Thanks!

VerinSenpai (approved):
Looks good to me! Thank ya for that. Less work I gotta do later :)

perryprog:
Note that volume control for `/playlocalsound` with follow set to false won't work until space-wizards/RobustToolbox#6011. Shouldn't block this PR being merged though.

Resources/Locale/en-US/administration/commands/play-local-sound-command.ftl
```diff
@@ -0,0 +1,9 @@
+cmd-playlocalsound-desc = Plays a localized sound from your current position for the specified players or for every nearby player if no players are specified.
+cmd-playlocalsound-help = playlocalsound <path> [volume] [follow] [user 1] ... [user n]
```
Pok27, line 2:
```suggestion
cmd-playlocalsound-help = Usage: {$command} <path> [volume] [follow] [user 1] ... [user n]
```

Content.Server/Administration/Commands/PlayLocalSoundCommand.cs
```diff
@@ -0,0 +1,125 @@
...
+        var playerEntity = shell.Player?.AttachedEntity;
+        if (playerEntity == null)
+        {
+            shell.WriteLine(Loc.GetString("cmd-playlocalsound-no-entity"));
+            return;
+        }
+
+        if (args.Length == 0) // No arguments, show command help.
+        {
+            shell.WriteLine(Loc.GetString("cmd-playlocalsound-help"));
```
Pok27, line 39:
```suggestion
            shell.WriteLine(Help);
```

Content.Server/Administration/Commands/PlayLocalSoundCommand.cs
```diff
@@ -0,0 +1,125 @@
...
+        var filter = Filter.Empty();
+        var audio = AudioParams.Default;
+
+        var replay = true;
+        var followRunner = false;
+
+        var playerEntity = shell.Player?.AttachedEntity;
+        if (playerEntity == null)
+        {
+            shell.WriteLine(Loc.GetString("cmd-playlocalsound-no-entity"));
```
Pok27, line 33: Maybe
```suggestion
            shell.WriteLine(Loc.GetString("shell-only-players-can-run-this-command"));
```

PR #37886, opened 2025-05-27
Add softcrit

notquitehadouken:
## About the PR

Add soft crit

## Why / Balance

This is a feature in SS13, so I thought to myself "What if it was a feature in SS14"

It is implemented differently but the idea is the same

## Media

https://github.com/user-attachments/assets/08da4238-3a30-49fe-b10d-b61a643165f8

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added soft-crit. You now have a bit of time to say goodbye to your loved ones or whatever if you take critical damage.


Content.Shared/Damage/Components/DamageableComponent.cs
```diff
@@ -47,6 +47,31 @@ public sealed partial class DamageableComponent : Component
...
+        /// <summary>
+        ///     The amount of time it takes for <see cref="DamageEffective"/> to catch up with <see cref="Damage"/>
+        ///     This value is decreased with respect to how much damage a biological container has taken,
+        ///     becoming 0 if it has died
+        ///     Ex: If this is 2 seconds, then a person that has instantly reduced to critical damage from no damage
+        ///     would take 1 second to crit, because crit is halfway between perfectly healthy and dead.
+        /// </summary>
+        [DataField]
+        public TimeSpan DamageLerpTimeZeroDamage = TimeSpan.FromSeconds(20);
+
```
metalgearsloth, line 74: As mentioned on discord damageable should not be infected with this stuff, it should be kept separate.

CatBackGround:
@notquitehadouken SoftCrit broking many tests. It's over.

gluesniffler:
Why not use the mobthresholds system to define a new threshhold for soft crit and handle your behavior from there? I'm struggling a bit to see why you need to be tracking a separate damage dict on top of the actual damage dict, along with all the prediction implications that come along with it as you end up posting on discord.

Also, these two functions must be giving you grief related to applying state while resetting predicted entities.
```csharp
private void OnDamageChanged(Entity<DamageableComponent> uid, ref DamageChangedEvent args)
    {
        if (CalculateTimeToEffective(uid) <= TimeSpan.Zero)
        {
            RemoveActiveDamage(uid, uid.Comp);
            return;
        }

        EnsureComp<ActiveDamageComponent>(uid);
    }

    public void RemoveActiveDamage(EntityUid uid, DamageableComponent? damageable)
    {
        if (!Resolve(uid, ref damageable))
            return;
        ResetDamageEffective(uid, damageable);
        RemComp<ActiveDamageComponent>(uid);
    }
```
Shitcoder in me says to just swap to RemCompDeferred or try checking for _timing.ApplyingState where relevant, but probably the underlying logic is flawed in the first place.

PR #37907, opened 2025-05-28
Sprite Fade Overlay Restricted to overlapping clickables

Xeri7:
## About the PR
Restricted Sprite Fade to only fade if there is a clickable behind the sprite. Can be overriden using a new field on SpriteFade to preserve old behavior, mainly for larger objects.

## Why / Balance
Sprite Fade was a bit too aggressive with things like plants and trees. Was especially noticeable on potted plants

## Technical details
Added a new section of code when determining whether or not to fade, which runs through the sprites again just before the fade would occur to see if there is anything behind that it needs to fade for. Put the check last as it is the most expensive of the fade checks.
This new section is skipped if you have set ignoreClickableRestrict to true.

## Media

https://github.com/user-attachments/assets/421eae6b-5e15-4e6d-a517-bf8ca11990d7

## Requirements

**Changelog**

:cl:
- tweak: Made Sprite Fade only happen when there is something there.


sowelipililimute:
some of the bigger entities (i.e. anomaly generator or 48x32 walls on downstreams) do probably need to always fade when moused over, because otherwise the user will have difficulty locating potentially small clickable items behind large areas of obscurity. i would advise adding a datafield to the `SpriteFieldComponent` for this behaviour

Xeri7:
> some of the bigger entities (i.e. anomaly generator or 48x32 walls on downstreams) do probably need to always fade when moused over, because otherwise the user will have difficulty locating potentially small clickable items behind large areas of obscurity. i would advise adding a datafield to the `SpriteFieldComponent` for this behaviour

I thought this was a good addition, so it has been added. Simple new boolean that defaults to false but can be added in case you want to override the new behavior. Put it on the anomaly gen as an example

Content.Shared/Sprite/SpriteFadeComponent.cs
```diff
@@ -8,5 +10,6 @@ namespace Content.Shared.Sprite;
 [RegisterComponent, NetworkedComponent]
 public sealed partial class SpriteFadeComponent : Component
 {
-
+    [DataField("ignoreClickableRestrict")]
```
beck-thompson, line 13: You don't actually need the name override! Just do `[DataField]`

Also add a comment explaining what it does (Something like "If false, only fades when hovering over items" or something similar)

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -118,6 +118,23 @@ private void FadeIn(float change)
                         {
                             continue;
                         }
+
+                        if (_fadeQuery.TryGetComponent(ent, out SpriteFadeComponent? fadeComp) && !fadeComp.IgnoreClickableRestriction)
```
beck-thompson, line 122: Just do `out var fadeComp`

Content.Shared/Sprite/SpriteFadeComponent.cs
```diff
@@ -1,3 +1,5 @@
+using Content.Shared.Hands.Components;
+using Content.Shared.Item;
```
beck-thompson, line 2: Unused

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -118,6 +118,23 @@ private void FadeIn(float change)
...
+                        {
+                            var hasSpriteBeneath = false;
+                            foreach (var subEnt in state.GetClickableEntities(mapPos, excludeFaded: false))
+                            {
+                                if (subEnt != ent &&
+                                    _spriteQuery.TryGetComponent(ent, out var subSprite) &&
+                                    subSprite?.DrawDepth >= sprite.DrawDepth)
```
beck-thompson, line 129: `subSprite.DrawDepth`

beck-thompson (changes requested):
Looks good, my only concern is performance but I personally can't tell if this ok or not. The double for loop makes me sus but I can't think of a reason why you would have more than like ~30 things being clickable at one moment. I also assume `GetClickableEntities` is somewhat fast so :shrug:

IamVelcroboy:
> some of the bigger entities (i.e. anomaly generator or 48x32 walls on downstreams) do probably need to always fade when moused over, because otherwise the user will have difficulty locating potentially small clickable items behind large areas of obscurity. i would advise adding a datafield to the `SpriteFieldComponent` for this behaviour

The only thing I don't like about this is you can't hide, or hide things behind objects anymore. Before the other person had to be behind the object to see it.

beck-thompson:
> The only thing I don't like about this is you can't hide, or hide things behind objects anymore. Before the other person had to be behind the object to see it.

You could always right click and it would pop up a list! This is just a little nicer way of doing it

beck-thompson (approved):
Feels good and makes sense! Again, I'm unsure about performance but I think its fine

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -118,6 +118,23 @@ private void FadeIn(float change)
...
+                                    subSprite.DrawDepth >= sprite.DrawDepth)
+                                {
+                                    hasSpriteBeneath = true;
+                                    break;
+                                }
+                            }
+                            if (!hasSpriteBeneath)
+                                continue;
+                        }
                     }
```
metalgearsloth, line 126: This is going to make this at least twice as expensive, is there a need for this to even exist?

Xeri7: Dividing parts of response for ease of discussion:
1. Yes, it is needed, thats the part keeping it from doing the current behavior. Right now, any object with sprite fade that does not have reasonably large hard fixtures is quite difficult to interact with, as they stay faded until you can manuever to the hard fixture, and then you have to wait for the unfade. This part is basically the entire point of this PR.
2. it is more expensive but only when you are hovering over an object with SpriteFade, AND that object doesn't already meet one of the other criteria to fade out. It should have minimal performance impact overall, as that is a relatively small set of scenarios, and normally only even has one other sprite in the loop, if any.

jessicamaybe (approved):
Looks good, I merged master in and it looks like everything still works!

Content.Client/Sprite/SpriteFadeSystem.cs
```diff
@@ -112,6 +112,23 @@ private void FadeIn(float change)
...
+                                if (subEnt != ent &&
+                                    _spriteQuery.TryGetComponent(ent, out var subSprite) &&
+                                    subSprite.DrawDepth >= sprite.DrawDepth)
+                                {
+                                    hasSpriteBeneath = true;
+                                    break;
+                                }
+                            }
+                            if (!hasSpriteBeneath)
+                                continue;
```
Princess-Cheeseballs: Doing another for loop inside of the same for loop is bad behavior. If we *have* to do this, we should just get an ordered list based on the draw depth of the sprite using one loop before all of this. There's ways to do that in clickable system already iirc, and if not it's not hard to make yourself.

Princess-Cheeseballs (changes requested):

Princess-Cheeseballs (approved):
Works, and avoids needing to do 2 for loops!

Xeri7:
> Works, and avoids needing to do 2 for loops!

Has it been thoroughly tested? this seems like it would only work if the item was behind or on level with the player sprite now. I had 2 loops because I needed to check for items that were behind the object, not items below player level.
I don't have my environment set up any more right now or i would check myself.
Thanks for finally reviewing this btw, the combination of this and another PR getting stuck in review hell is why i am not doing more contributing xD

Xeri7:
Oh wait, i think i see what you are doing? It looks like we are assuming the clickable entities is in order by draw depth, i guess I just didnt know that was a safe assumption haha

Princess-Cheeseballs:
> Oh wait, i think i see what you are doing? It looks like we are assuming the clickable entities is in order by draw depth, i guess I just didnt know that was a safe assumption haha

Getclickable already orders by drawdepth yeah so we don't need to do another loop. At worst, we would do two unnested loops to order by draw depth to avoid having excess CPU cycles from doing loops inside of loops.

Xeri7:
> > Oh wait, i think i see what you are doing? It looks like we are assuming the clickable entities is in order by draw depth, i guess I just didnt know that was a safe assumption haha
> 
> Getclickable already orders by drawdepth yeah so we don't need to do another loop. At worst, we would do two unnested loops to order by draw depth to avoid having excess CPU cycles from doing loops inside of loops.

Ahhh perfect, good to know! I retract my concerns then, and thanks again for yall picking this back up and getting it through!

PR #37908, opened 2025-05-28
Adds basic meds to SyndieJuice vendor

IProduceWidgets:
## About the PR

Adds 5 basic meds to the SyndieJuice vendor on nukie planet to alleviate some of the agent guidebook quiz. 1x jug each
- Bicaridine
- Dermaline
- DexalinPlus
- Dylovene
- Saline

also yoinks the HideSpawnMenu category off jugs because I often want to map a jug and its pain to use the spawn command every time. One day they should be dynamically created prototypes from every reagent but for meow this is fine.

## Why / Balance

Save players from repetitive stress syndrome and let agents do the fun chemistry for grenades and shit instead of agent rounds being a necessary 20m guidebook quiz.

Some people might say "but nukeops too fast now" to which I reply, find a less terrible stalling mechanic. We have the technology.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The nukie SyndieJuice vendor now contains some basic medical chems.


IProduceWidgets:
I asked around and people said use Dex+ instead of Dex, but weh, lemme know if anyone cares.

beck-thompson:
I like this as an idea, my only issue is it might not change much and the agent will still spend a bunch of time making chems (just more advanced ones)

IProduceWidgets:
> I like this as an idea, my only issue is it might not change much and the agent will still spend a bunch of time making chems (just more advanced ones)

We can lean in further, I just figured it makes some sense to leave room for them to make improvements if they want to, or go fast and use cheapo chems and a med bundle or something. I was also hesitant because the advanced meds have pretty steep misuse/overdose penalties that are kinda hard to actually monitor well in an adrenaline fueled combat role. These at least are pretty mild in that regard.

If we do go for more advanced meds we could either
- give them some limited quantity via bottles / vials like we do epi and omni
- give them some syndicate specific chems

I don't like option two much because it feels a little hamfisted instead of emergent and knowledge based.

Mixelz:
Making them advanced chems would certainly be a noticeable increase to the survivability of blitz ops, so I believe that solution should be avoided all together.
I do think the base concept here is rock solid and will be a massive help in ensuring that nukie chems aren't a primary factor into why nukies take too long to show up.

beck-thompson:
Yeah no no, I don't think leaning into it further is good, I'm just not sure it will have the effect we are hoping for (Although I do think its totally worth the shot and is a good idea)

NoElkaTheGod:
I like the idea, but dylovene is kinda excessive here. Nukies rarely ever get poisoned, and if they do, that's just skill issue. Giving them some basic chems is a really good idea though, would help a lot for newkie agents and save some time for the more experienced ones.

Resources/Prototypes/Catalog/VendingMachines/Inventories/chemvend.yml
```diff
@@ -33,9 +33,13 @@
   startingInventory:
     Jug: 4
     JugAluminium: 2
+    JugBicaridine: 1
     JugCarbon: 4
     JugChlorine: 1
     JugCopper: 2
+    JugDermaline: 1
+    JugDexalinPlus: 1
+    JugDylovene: 1
```
ArtisticRoomba, line 42: Generally against giving nukies Dylo. Nukies don't really expect to get poisoned and it trolls the crew if they are making weapons centered around poison (ex. amatoxin arrows or whatever).

pheenty:
Yeah, dylo is no useful except when used with meth, even then one jug is not enough, not to mention that if you can cook meth its takes you no more than a minute to cook dylo too.
Some hyro (not arith) would be nice though, for use against x-ray cannons and uranium ammo.

TPgameing:
I'm not a big fan of this to be honest, i suck at chem myself and would LOVE to see free chems as agent. But it's suppose to be a Field medic role, there's a reason it's locked behind chem hours, it's so whoever is the agent has a good background on chemicals. It isn't suppose to be a Lube bomb factory!

Resources/Prototypes/Catalog/VendingMachines/Inventories/chemvend.yml
IProduceWidgets: bump

IProduceWidgets: @ArtisticRoomba

slarticodefast:
I'm not a fan of this to be honest, it just skips a huge part of the agent's job. For the cases where there is no agent available or he is bad at his job the med bundle already exists. But needing time to prepare is part of nukie gameplay in my opinion. If that is monotone then that is a general problem with chemistry mechanics as a whole and should be addressed as such.

Resources/Prototypes/Catalog/VendingMachines/Inventories/chemvend.yml
IProduceWidgets: Resolving this review as abandoned.

Simyon264:
I agree with Slarti.

K-Dynamic:
> I'm not a fan of this to be honest, it just skips a huge part of the agent's job. For the cases where there is no agent available or he is bad at his job the med bundle already exists. But needing time to prepare is part of nukie gameplay in my opinion. If that is monotone then that is a general problem with chemistry mechanics as a whole and should be addressed as such.

I'm not sure about this because nukie agents would have to help prepare equipment and strategy, regardless if they have to make chems or not. Making chems is also going to be monotone even if we do a rework since agents make the same meds everytime.

I would like to ask about getting a Discourse forum post about this PR and discussing it further since agent meds are extremely important to nukies.

PR #37915, opened 2025-05-29
Change smoke/foam/explosion chemistry reaction order & energy transfer

SlamBamActionman:
## About the PR

This PR does two things:

1. Foam and smoke effects have been changed to trigger *after* all other reactions.
2. Foams, smokes and explosion reactions that can eliminate all reagents no longer transfer the reagents' energy to the solution (i.e. no longer heats the container).

## Why / Balance

1. Foam and smoke effects have been changed to trigger after all other reactions. 

Due to foam and smoke reactions previously having priority over normal mix reactions, there was a strange behavior that could occur where, due to foam/smoke not allowing reagent reactions, a mix of chemicals that should normally react wouldn't do so until inhaled. Not only is this unintuitive (all reagents are in the cloud, yet somehow not mixing), but it allowed for certain inhalation mixes that wouldn't trigger effects until inhaled (most notably "bee bombs"). 

2. Foams, smokes and explosion reactions that can eliminate all reagents no longer transfer reagents' energy to the solution (i.e. no longer heats the container).

The chemistry system runs on a type of largely simplified exothermic/endothermic reaction system that, while nifty, causes some issues when bumping up against the limitations of the simulation. Due to foams/smokes/explosions having potential to delete all reagents involved in the reaction, any remaining reagents in the solution take on the full heat energy of what was deleted, regardless of how little actually remains. 

E.g. a fluorosurfactant/water mix with 99.99u of each, with 0.01u of a random other reagent, after reacting results in that other reagent becoming hotter than the core of the sun by 3 magnitudes.   

Because of this, the heat system could be used to rapidly trigger heat-induced chemical reactions. Since the energy is meant to go into the act of creating the foam/smoke/explosion itself, removing it feels like an okay alternative. Foam/smoke still transfers the heat of the original container they were in (and due to change 1, exo/endothermic reactions happen before the foam/smoke).

It is worth noting that if a more exact heat simulation system is introduced in the future, the behavior could be properly tweaked back in.

## Technical details

Surprisingly... Just yaml!

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Chemistry foam and smoke reactions now trigger after regular mix reactions.
- tweak: Foam, smoke and explosion reactions no longer transfer the reagents' energies to the container.

### DIFF
--- a/Resources/Prototypes/Recipes/Reactions/chemicals.yml
+++ b/Resources/Prototypes/Recipes/Reactions/chemicals.yml
@@ -104,6 +104,7 @@
   id: PotassiumExplosion
   impact: High
   priority: 20
+  conserveEnergy: false
   reactants:
     Water:
       amount: 1
@@ -119,8 +120,9 @@
 
 - type: reaction
   id: Smoke
-  priority: 10
+  priority: -10
   impact: High
+  conserveEnergy: false
   reactants:
     Phosphorus:
       amount: 1
@@ -137,8 +139,9 @@
 
 - type: reaction
   id: Foam
-  priority: 10
+  priority: -10
   impact: High
+  conserveEnergy: false
   reactants:
     Fluorosurfactant:
       amount: 1
@@ -154,7 +157,8 @@
 - type: reaction
   id: IronMetalFoam
   impact: High
-  priority: 10
+  priority: -10
+  conserveEnergy: false
   reactants:
     Iron:
       amount: 3
@@ -172,7 +176,8 @@
 - type: reaction
   id: AluminiumMetalFoam
   impact: High
-  priority: 10
+  priority: -10
+  conserveEnergy: false
   reactants:
     Aluminium:
       amount: 3
@@ -191,6 +196,7 @@
   id: UraniumEmpExplosion
   impact: High
   priority: 20
+  conserveEnergy: false
   reactants:
     Iron:
       amount: 1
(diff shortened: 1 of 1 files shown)


7-62x54r:
> it allowed for certain inhalation mixes that wouldn't trigger effects until inhaled (most notably "bee bombs").

And why is this bad again? Pyrotech chem is very barebones, at least make the foam/smoke explode instead of inhaling humanoids. I'm fine with energy transfer fix tho.

SlamBamActionman:
> And why is this bad again? Pyrotech chem is very barebones, at least make the foam/smoke explode instead of inhaling humanoids. I'm fine with energy transfer fix tho.

It's a confluence of mechanics that results in powerful and unintuitive behavior. If you mix a bunch of chemicals, they should mix when you mix them, no? While somewhat fun the bee bomb and gib foam mix are relying on esoteric knowledge not readily communicated to players, which wouldn't be so bad if they weren't also so extremely powerful.

Explosion chems are primarily balanced around singular explosions, and due to how foam works you'd get several triggers, resulting in the damage shooting away. Foam still serves plenty of uses for acids, toxins and fire, and explosives are easily accessible through other methods (throwing jugs my beloved).

ScarKy0 (approved):
yeah

fesdgsdahagdzf:
so the fun begins.

PR #37922, opened 2025-05-29
fix teleriot visuals

pheenty:
## About the PR

fixes teleshield visuals
## Why / Balance

bug bad fix good, also the component would be useful downstream
## Technical details

one enum one component and one method, it doesnt get any easier than that
## Media

image
## Requirements

## Breaking changes

**Changelog**
no cl no fun ig


slarticodefast:
We also got https://github.com/space-wizards/space-station-14/pull/35341 which does the same.
We will have to check which PR has the better approach, but the other one also decouples it from ItemTogglePointlight, which seems good.

Also the teleriot shield is not the only item affected by this bug. Headphones for example are as well, and likely others, too.

EmoGarbage404:
no longer necessary

PR #37923, opened 2025-05-29
Fix soap recipe so it actually works

Gentleman-Bird:
## About the PR
The current soap recipe calls for 15u fat, 10u salt, and 10u water. This recipe does not work, as the salt and water react to form saline. This PR changes the recipe to 15u fat, and 25u saline. (Made of 5u salt and 20u water)

## Why / Balance
N/A

## Technical details
2 lines of yaml code

## Media
N/A

## Requirements

## Breaking changes
N/A
**Changelog**

:cl:
- fix: Fixed the "create soap" recipe so it can actually be made

### DIFF
--- a/Resources/Prototypes/Recipes/Reactions/fun.yml
+++ b/Resources/Prototypes/Recipes/Reactions/fun.yml
@@ -31,10 +31,8 @@
   reactants:
     Fat:
       amount: 15
-    TableSalt:
-      amount: 10
-    Water:
-      amount: 10
+    Saline:
+      amount: 25
   effects:
     - !type:CreateEntityReactionEffect
       entity: Soap


Gentleman-Bird:
> This could be solved by changing the reaction priority. Chaning the reactants makes the soap recipe and the bloody soap recipe different, which feels weird.

That's also a bit of an awkward solution, since the recipe would not work if you added reagents in 5u increments. Using saline instead might not be elegant, but it seems to me to be the least bad option, aside from changing the recipe entirely.

beck-thompson (approved):
Seems a little sus but I think its ok :thinking:

EmoGarbage404:
strange but isn't life hell anyways

PR #37924, opened 2025-05-29
Change potassium-water explosion scaling

SlamBamActionman:
## About the PR

This PR changes the Potassium-Water explosion properties, flattening out the explosion curve to spread out damage across an rea for large-quantity explosions, and requiring larger damage to reach the max damage capacity.

## Why / Balance

Potassium + Water = Explosion is fun! If we can't have FNAF in the game, at the very least we can give unassuming Janitors explosion jumpscares. 

Both potassium and water are fairly common chemicals, and the reaction is trivial (unlike CLF3, which both contain rare chemicals *and* has a heat requirement). It also has low unit requirements to reach the 100 damage crit threshold (30u potassium), where its scaling quickly evens out (e.g. 100u potassium = 150 damage, 200u = 180 damage). This means potassium + water is a fairly effective aggressive weapon option; simple to create, a lot of damage. This has also resulted in  combining potassium with space cleaner which, when thrown and splashed on the ground, uses the space cleaner's effect to create water on the spot. Due to the factors mentioned above, these can be very potent and are in my opinion overtuned.

The values in this PR changes the curve to require more reagents to reach the scaling softcap, and flattens the cap such that damage is spread out in a wider area. The maximum damage a potassium explosion can deal is now 105 (whereas before it was effectively uncapped).

Images below show how the changes look for some reagent quantities, with the damage dealt edited in at the bottom of each tile. 

15u Potassium:
image

30u Potassium:
image

100u Potassium:
image

200u Potassium:
image

If you wonder about mopping: Puddles tend to even out at 20u, so a janitor mopping a potassium puddle now takes 75 damage instead of 90 damage.

## Technical details

Just yaml!

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Potassium-water explosion damage scaling has been tweaked to be less potent.

### DIFF
--- a/Resources/Prototypes/Recipes/Reactions/chemicals.yml
+++ b/Resources/Prototypes/Recipes/Reactions/chemicals.yml
@@ -112,7 +112,7 @@
   effects:
     - !type:ExplosionReactionEffect
       explosionType: Default
-      intensityPerUnit: 0.2 # 50+50 reagent for maximum explosion
+      intensityPerUnit: 0.25
       maxTotalIntensity: 100
       intensitySlope: 4
       maxIntensity: 7
--- a/Resources/Prototypes/Recipes/Reactions/chemicals.yml
+++ b/Resources/Prototypes/Recipes/Reactions/chemicals.yml
@@ -112,10 +112,10 @@
   effects:
     - !type:ExplosionReactionEffect
       explosionType: Default
-      maxIntensity: 100
-      intensityPerUnit: 0.5 # 50+50 reagent for maximum explosion
-      intensitySlope: 4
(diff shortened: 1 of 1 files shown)


gitjubx:
kinda afraid to start an argument about it but never seen really someone is getting killed on purpose with potassium explosion because its so weak or someone actually tries to reach the max potentional since mostly your limited on throwing just 1 jug...you mostly just use tiny amount that makes a tiny explosion to trigger the CLF3 explosion combo

Pronana:
This is gonna change my calculations for the minimum amount of potassium you need to detonate a secoff's grenades. Definitely for the betterment of the average security officer.

PR #37928, opened 2025-05-29
Split codewords into its own system

### LINKED ISSUE #29876: traitors have different codewords
image

Simyon264:
## About the PR

This PR splits codewords into its own system for general use in more than just traitors. It also fixes the issue where sleeper agents have different codewords than the round start traitors. 

## Why / Balance

Fixes #29876

Codewords being baked into the Traitor system was also bad code design.

## Technical details

Created 2 new prototypes: `CodewordGenerator` and `CodewordFaction`.
`CodewordGenerator` is responsible for defining what dataset to use and how many codewords to generate.
`CodewordFaction` is a simple identifier for different sets of codewords. I used this instead of a magic const because magic const values are generally not good.

For generating the codewords I added a gamerule `CodewordSystem`.
This gamerule (once added, it should always be added before any codewords are requested) takes the Dictionary of `<CodewordFaction, CodewordGenerator>` given in the rule component and uses it to generate the codewords as specified in the `CodewordGenerator`. The codewords are stored in a nullspaced entity containing the `CodewordComponent` which has the array of codewords. A referenece to the EntityUid is kept in the rule component.

Retrieving codewords is done by calling `CodewordSystem::GetCodewords(ProtoId<CodewordFaction>)` . This method will search all codeword gamerules for the requested faction. If no codewords are found for this faction, the method throws an error.

Some small changes had to be made to `TraitorCodePaperSystem` to ensure it still works with the central codeword system.

For `game_presets.yml` I had to add the gamerule at the first position always, even in modes that do not call for codewords, in case an admin adds traitors midround. I am unsure if there is a better way to do this.

## Requirements

Round Start Traitor:
image
Sleeper Agent:
image

**Changelog**

:cl: Simyon
- fix: Sleeper agents no longer have different codewords than round-start traitors.

### DIFF
--- a/Content.Server/Codewords/CodewordFaction.cs
+++ b/Content.Server/Codewords/CodewordFaction.cs
@@ -1,0 +1,14 @@
+﻿using Robust.Shared.Prototypes;
+
+namespace Content.Server.Codewords;
+
+/// <summary>
+/// This is a prototype for easy access to codewords using identifiers instead of magic strings.
+/// </summary>
+[Prototype("codewordFaction")]
+public sealed partial class CodewordFaction : IPrototype
+{
+    /// <inheritdoc/>
+    [IdDataField]
+    public string ID { get; } = default!;
+}
--- a/Content.Server/Codewords/CodewordRuleComponent.cs
+++ b/Content.Server/Codewords/CodewordRuleComponent.cs
@@ -1,0 +1,22 @@
+﻿using Robust.Shared.Prototypes;
+
+namespace Content.Server.Codewords;
+
+/// <summary>
+/// Component that defines <see cref="CodewordGenerator"/> to use and keeps track of generated codewords.
+/// </summary>
+[RegisterComponent, Access(typeof(CodewordSystem))]
+public sealed partial class CodewordRuleComponent : Component
+{
+    /// <summary>
+    /// The generators available.
+    /// </summary>
+    [DataField(required: true)]
+    public Dictionary<ProtoId<CodewordFaction>, ProtoId<CodewordGenerator>> Generators = new();
+
+    /// <summary>
+    /// The generated codewords. The value contains the entity that has the <see cref="CodewordComponent"/>
+    /// </summary>
+    [ViewVariables(VVAccess.ReadOnly)]
+    public readonly Dictionary<ProtoId<CodewordFaction>, EntityUid> Codewords = new();
+}
--- a/Content.Server/Codewords/CodewordSystem.cs
+++ b/Content.Server/Codewords/CodewordSystem.cs
@@ -1,0 +1,74 @@
+﻿using System.Linq;
+using Content.Server.Administration.Logs;
+using Content.Server.GameTicking.Rules;
+using Content.Shared.Database;
+using Content.Shared.GameTicking.Components;
+using Robust.Shared.Map;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Random;
+
+namespace Content.Server.Codewords;
+
+/// <summary>
+/// Gamerule that provides codewords for other gamerules that rely on them.
+/// </summary>
+public sealed class CodewordSystem : GameRuleSystem<CodewordRuleComponent>
+{
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+    [Dependency] private readonly IAdminLogManager _adminLogger = default!;
+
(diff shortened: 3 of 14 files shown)


Content.Server/Codewords/CodewordFaction.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Robust.Shared.Prototypes;
+
+namespace Content.Server.Codewords;
+
+/// <summary>
+/// This is a prototype for easy access to codewords using identifiers instead of magic strings.
+/// </summary>
+[Prototype("codewordFaction")]
+public sealed partial class CodewordFaction : IPrototype
```
slarticodefast, line 9:
```suggestion
[Prototype]
public sealed partial class CodewordFactionPrototype : IPrototype
```

Content.Server/Codewords/CodewordRuleComponent.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// The generators available.
+    /// </summary>
+    [DataField(required: true)]
+    public Dictionary<ProtoId<CodewordFaction>, ProtoId<CodewordGenerator>> Generators = new();
+
+    /// <summary>
+    /// The generated codewords. The value contains the entity that has the <see cref="CodewordComponent"/>
+    /// </summary>
+    [ViewVariables(VVAccess.ReadOnly)]
```
slarticodefast, line 20: won't this break persistance?

PicklOH:
Also helps for future revs :godo:

Content.Server/Codewords/CodewordSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Content.Server.Codewords;
+
+/// <summary>
+/// Gamerule that provides codewords for other gamerules that rely on them.
+/// </summary>
+public sealed class CodewordSystem : GameRuleSystem<CodewordRuleComponent>
+{
+    [ValidatePrototypeId<EntityPrototype>]
+    public static string RuleComponent = "CodewordRule";
```
EmoGarbage404, line 18: we typically do a `private static EntProtoId` instead of the validator attribute. Also, if you do that, you can use the templated version to validate that the entity prototype has a certain component (`CodewordComponent`)

Content.Server/Codewords/CodewordRuleComponent.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Robust.Shared.Prototypes;
+
+namespace Content.Server.Codewords;
+
+/// <summary>
+/// Component that defines <see cref="CodewordGeneratorPrototype"/> to use and keeps track of generated codewords.
+/// </summary>
+[RegisterComponent, Access(typeof(CodewordSystem))]
+public sealed partial class CodewordRuleComponent : Component
```
EmoGarbage404, line 9: does it make sense to attach these codewords to rules? Like, if we have two syndicate rules with codewords, wouldn't we want them to share the codewords so long as they have the same faction prototype?

maybe it'd be better to store them on some arbitrary ent in nullspace like `ClockSystem`

EmoGarbage404 (changes requested):

Content.Server/Codewords/CodewordRuleComponent.cs
Simyon264: They do share them. If the game rule isn't added, it gets added when traitor calls for it. If its added twice (or more), it checks for first match, meaning you can have multiple codeword rules with different factions in them. 

I made them a gamerule because that allows easy access to a YAML config via the rule component, and it made the most sense in my head.

Content.Server/Codewords/CodewordGeneratorPrototype.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// The adjectives to be used by the generator
+    /// </summary>
+    [DataField]
+    public ProtoId<LocalizedDatasetPrototype> CodewordAdjectives = "Adjectives";
+
+    /// <summary>
+    /// The verbs to be used by the generator
+    /// </summary>
+    [DataField]
+    public ProtoId<LocalizedDatasetPrototype> CodewordVerbs = "Verbs";
```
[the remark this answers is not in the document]
Simyon264: *shrug

This part was transferred over from the traitor system, so you're probably correct. I'll change it.

PR #37930, opened 2025-05-29
Cleanup all warnings from `Dispose` on Controls

Tayrtahn:
## About the PR

Replaced all uses and overrides of `Dispose` on UI controls.

This felt way too easy. There's probably something wrong here, so please test and review this carefully!

## Why / Balance

Together with https://github.com/space-wizards/RobustToolbox/pull/5986, this fixes 165 warnings.

## Technical details

- `Dispose` was replaced with `Close` for windows or `Orphan` for controls.
- `Dispose` overrides were swapped for `ExitedTree` overrides.
- A few other tweaks were made, notable to remove some disposal-related stuff from the admin menu and to fix a double-removal that was happening with speech bubbles.

## Requirements


VerinSenpai:
I haven't gone fully through this yet but I will say in my experience while testing this for the game bar, I noted that windows still got removed from the display tree as the warning indicates they should so I imagine you're on the right track.

Content.Client/UserInterface/Systems/EscapeMenu/EscapeUIController.cs
```diff
@@ -116,7 +116,7 @@ public void OnStateExited(GameplayState state)
     {
         if (_escapeWindow != null)
         {
-            _escapeWindow.Dispose();
+            _escapeWindow.Close();
```
VerinSenpai, line 119: question regarding this and the emotes ui, is it even necessary to call this? the window is closed at the end of round whether this is here or not.

VerinSenpai: Is this perhaps better form? asking to expand my understanding of UI code.

Tayrtahn: That's a good question, but unfortunately my own lack of experience with UI code means that I don't know the answer either!

PR #37931, opened 2025-05-29
Add a sleep delay to Nocturine

SlamBamActionman:
## About the PR

This PR introduces a delay to the effects of Nocturine, causing the forced sleep to not take effect until after 6 seconds after the effect threshold has been reached. 

The duration and build-up remains the same, e.g. 10u Nocturine still results in 10 seconds of sleep, 20u results in 50 seconds. It just offset by 6 seconds from when the 8u sleep threshold is reached.

## Why / Balance

Nocturine is one of the easiest ways to immobilize a character; it is practically instant and any counterplay boils down to "always be around people" with no other option. The hypopen + nocturine combo has been a staple of antag gameplay for years, despite its lack of counterplay giving it a reputation of being "cheap" and "unfun". However, due to its unique mechanic and TC requirement this has been considered an okay status quo, even if it's not preferable.

There's been a few attempts to nerf Nocturine, but it's been shown tricky to do without completely kneecapping its usage or making it difficult to administer. 

This PR adds a 6 second delay before the Nocturine takes effect. It still has the same duration and effect, but gives a short amount of time for a person to react. This time can be spent calling out the drugging, running to a populated area, making noise or fighting back. 

In the end the victim is probably not gonna be able to get away, but it provides *some* active counterplay for a still very strong and expensive Traitor item. Should 6 second be too short, additional time can be introduced (originally the PR had a 10 second delay).

## Technical details

`StatusEffectSystem` has been updated to support a time delay. It already had partway support for this in the `startTime` argument, but that was only recorded in the component's `Cooldown` property without being utilized to actually delay assigning the desired component. 

Also minor guidebook change to account for it.

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Nocturine now has a delay of 6 seconds before sleeping the target. The duration of sleep has not been changed.

### DIFF
--- a/Content.Shared/EntityEffects/Effects/StatusEffects/GenericStatusEffect.cs
+++ b/Content.Shared/EntityEffects/Effects/StatusEffects/GenericStatusEffect.cs
@@ -62,7 +62,16 @@ public sealed partial class GenericStatusEffect : EntityEffect
         }
     }
 
-    protected override string? ReagentEffectGuidebookText(IPrototypeManager prototype, IEntitySystemManager entSys) => Loc.GetString(
+    protected override string? ReagentEffectGuidebookText(IPrototypeManager prototype, IEntitySystemManager entSys) =>
+        Delay > 0
+            ? Loc.GetString(
+        "reagent-effect-guidebook-status-effect-delay",
+        ("chance", Probability),
+        ("type", Type),
+        ("time", Time),
+        ("key", $"reagent-effect-status-effect-{Key}"),
+        ("delay", Delay))
+            : Loc.GetString(
         "reagent-effect-guidebook-status-effect",
         ("chance", Probability),
         ("type", Type),
--- a/Resources/Locale/en-US/guidebook/chemistry/effects.ftl
+++ b/Resources/Locale/en-US/guidebook/chemistry/effects.ftl
@@ -118,6 +118,22 @@ reagent-effect-guidebook-status-effect =
                 } {NATURALFIXED($time, 3)} {MANY("second", $time)} of {LOC($key)}
     }
 
+reagent-effect-guidebook-status-effect-delay =
+    { $type ->
+        [add]   { $chance ->
+                    [1] Causes
+                    *[other] cause
+                } {LOC($key)} for at least {NATURALFIXED($time, 3)} {MANY("second", $time)} with accumulation
+        *[set]  { $chance ->
+                    [1] Causes
+                    *[other] cause
+                } {LOC($key)} for at least {NATURALFIXED($time, 3)} {MANY("second", $time)} without accumulation
+        [remove]{ $chance ->
+                    [1] Removes
+                    *[other] remove
+                } {NATURALFIXED($time, 3)} {MANY("second", $time)} of {LOC($key)}
+    } after {NATURALFIXED($delay, 3)} {MANY("second", $delay)}
+
 reagent-effect-guidebook-set-solution-temperature-effect =
     { $chance ->
         [1] Sets


PicklOH:
I can understand the reasoning behind this PR. However, I personally don't believe that everything needs counterplay. Hear me out here.

The antagonist gameplay loop isn't meant to be a fair battleground. Syndies are vastly outnumbered by the crew, and typically matched in number by security.  Whereas all of Security will be working together to catch the syndies, not all syndies are guarenteed to work together to accomplish their goals. They are given access to the catalog of fun toys that's supposed to give them an edge that allows them to make up for that discrepancy. This is one such toy, albeit a powerful one. It's true there isn't much you can do against the Noct+Pen combo, but sometimes you just lose, and people need to be ok with that. There doesn't need to be counterplay for every single thing.

Against an unarmored, isolated target, the Cobra does the same exact thing as the Noct+Pen combo, but for far cheaper. Against targets that are harder to take down, like the HoS or Captain for example, it takes more planning and timing for the combo to work. Targets like the latter are hard to take down through pure stealth, and typically require going loud. Not everyone prefers to go loud, so Noct+Pen exists for those people.

We also have to remember that not everyone is hyper-robust with 1000+ hours. There needs to be items that cater to a lower skill floor, for a higher TC price. Noct+Pen is perfect for this. People with tons and tons of hours can find myriad of means to take out their target that doesn't require it. 

I think the Noct+Pen combo can be frustrating on the receiving end, but so can hundreds of other things in this game. I can count tons of times where I have randomly exploded, or have slipped resulting in my death, or any other random thing has killed me out of the plethora of things that can do so.

TLDR: Not everything needs counterplay. Noct+Pen is expensive and that alone is already a fine balancing lever.

Goo-Jargn:
> frustrating

This is a video game.
Imagine this.
You are a person who has enough time for a shift a day, otherwise your way too busy.
Then you get noct'd and minibombed in roughly 10 minutes.
But its balanced because the antag is now kneecapped.
That is my inherent issue with nocturine.
If you can just be instantly round removed in an incredibly quick way with no spectacle involved, it just makes it feel like your time is being wasted.

robinthedragon:
> I believe this was already uttered in the other PRs, but consider making it slow the target down just before putting them to sleep, especially if the delay ends up being longer than 6 seconds.
> 
> This makes it both more noticable for the target that they're being drugged with Nocturine, but also still give an immediate advantage to the one trying to hypo the target if they happen to fight back. Also, it just fits in with the flavor of a sleep drug to make the character woozy and lethargic just before knocking them unconscious

ye, i agree; chloral already slows people down before knocking them out, and chloral with good RNG* is a lot more fun then noct. your target gets disoriented and slow for a bit, then falls asleep

\*chloral with good rng acts similarly to noct in this PR: 5-10 seconds to kick in, then they're out for a few seconds.
bad RNG is pitifully bad. i've been shot with a bunch of tranq darts and only fallen asleep like two minutes afterwards

SlamBamActionman:
> PicklOH's reply

I think it's worth comparing is the "loudness" of the different kill methods, and look at where they differ. I am not talking about only volume (though that is certainly part of it), but how inconspicuous you are before, during and after.

The cobra is a very apt comparison because of how similar yet starkly different it is to the noctpen. Cobra doesn't leave any casings, is fairly quiet (the impact sounds are a bit... Hrmmm.) and easily concealable in a pocket or storage implant. But it still has some loudness going on; not only is the sprite somewhat noticeable in your hand and dodging *is* viable, but you still leave blood and change the target's health on the crew monitor, as well as damage negation from armor causing variability and can make misses cut close. It is cheaper than the noctpen, but that comes at a cost elsewhere.

Noctpen is on the other hand a lot more reliable, faster, and the main loudness of it is the snoring and getting found out transporting the sleeping body. It also isn't as obvious; even a Cobra being fired at the other end of the room is gonna draw immediate attention, but a noctpen's covertness can draw out the reaction time for bystanders. Add to the fact there are certain jobs that can outright bypass the uplink cost and it can bypass one of its major downsides.

This PR aren't changing most of those things. And, as I say in the description, the "counterplay" you're given isn't gonna be much of a counter; if you're noct'd, you're going to go down. But now there is *some* active agency in the hand of the victim, and even if it isn't gonna tip the scales much I hope that it will feel better for the victim to know they had some amount of options.

---

Why does this matter in a game as chaotic as SS14? Because the noctpen is a planned, targeted attack. Yes, you can die from a random meteor, Wiz fireball or just being in the wrong place at the wrong time when an antag is doing a crime. But those instances are rarely planned to target you specifically, and if they are they tend to provide some sort of agency for you (not necessarily actively for you, e.g. if you suddenly get gunned down in the bar, but for other crew to react and assist). Therefore, they tend to be a lot easier to brush off because it's either an unpreventable force majeure, or there is a chance to reflect on what you could've done differently.

If this PR manages to give a sense of agency to the victim getting noct'd while still largely being in favor of the antag, I feel it has served its purpose.

SlamBamActionman:
Re: causing slowdown/sleepiness.

I am not necessarily opposed to this idea, but I think Chloral Hydrate already fits that stereotype and Nocturine is powerful because it is so utilitarian. Considering how strong nocturine already is, I am hesitant to pull down on one balance lever only to pull up on another, so I may opt to leave it out of this PR.

Goo-Jargn:
> > PicklOH's reply
> 
> I think it's worth comparing is the "loudness" of the different kill methods, and look at where they differ. I am not talking about only volume (though that is certainly part of it), but how inconspicuous you are before, during and after.
> 
> The cobra is a very apt comparison because of how similar yet starkly different it is to the noctpen. Cobra doesn't leave any casings, is fairly quiet (the impact sounds are a bit... Hrmmm.) and easily concealable in a pocket or storage implant. But it still has some loudness going on; not only is the sprite somewhat noticeable in your hand and dodging _is_ viable, but you still leave blood and change the target's health on the crew monitor, as well as damage negation from armor causing variability and can make misses cut close. It is cheaper than the noctpen, but that comes at a cost elsewhere.
> 
> Noctpen is on the other hand a lot more reliable, faster, and the main loudness of it is the snoring and getting found out transporting the sleeping body. It also isn't as obvious; even a Cobra being fired at the other end of the room is gonna draw immediate attention, but a noctpen's covertness can draw out the reaction time for bystanders. Add to the fact there are certain jobs that can outright bypass the uplink cost and it can bypass one of its major downsides.
> 
> This PR aren't changing most of those things. And, as I say in the description, the "counterplay" you're given isn't gonna be much of a counter; if you're noct'd, you're going to go down. But now there is _some_ active agency in the hand of the victim, and even if it isn't gonna tip the scales much I hope that it will feel better for the victim to know they had some amount of options.
> 
> Why does this matter in a game as chaotic as SS14? Because the noctpen is a planned, targeted attack. Yes, you can die from a random meteor, Wiz fireball or just being in the wrong place at the wrong time when an antag is doing a crime. But those instances are rarely planned to target you specifically, and if they are they tend to provide some sort of agency for you (not necessarily actively for you, e.g. if you suddenly get gunned down in the bar, but for other crew to react and assist). Therefore, they tend to be a lot easier to brush off because it's either an unpreventable force majeure, or there is a chance to reflect on what you could've done differently.
> 
> If this PR manages to give a sense of agency to the victim getting noct'd while still largely being in favor of the antag, I feel it has served its purpose.

Also to note, wizards kill just about everyone.
I love funny rod, it makes me giggle.

Nukesthestation:
> This is a video game. Imagine this. You are a person who has enough time for a shift a day, otherwise your way too busy. Then you get noct'd and minibombed in roughly 10 minutes. But its balanced because the antag is now kneecapped. That is my inherent issue with nocturine. If you can just be instantly round removed in an incredibly quick way with no spectacle involved, it just makes it feel like your time is being wasted.

As someone who abuses noct+minibomb combo, I agree that it is 
1. extremely boring 
2. extremely op as it's basically a guaranteed rr
3. if you are using a hypopen (which usually is the method used), the person getting nocted basically has 0 time to react unless the syndi messes up
4. Other than using most of your tc, if you did it in a secluded area, you can basically get off scot free (hell, when I do this, I straight up go to sec and tell them I gibed someone and admit to the crime for a lesser punishment)

Admiral-Obvious-001:
People are forgetting that during this 6 seconds, you're also subject to the crippling slowing effect of the chem, which leaves you no option other than trying to fight and leave a blood puddle for the (terrible) detective to find, or screaming on radio, which would be stopped with a radio jammer, which in turn also prevents purchase of a minibomb, which *in turn* means that the noct pen bomb combo is less bullshit because it's impossible to actually do, unless you randomly mutate it in botany (why is this a thing), or synthesize it in chem.

As it is right now, noct pen is a 500ms 0 to death.

On the point of salvage turrets, you at minimum have 2 seconds to avoid the turrets, and landmines have big flashing red lights on them.

Pronana:
> People are forgetting that during this 6 seconds, you're also subject to the crippling slowing effect of the chem

Currently it doesn't have the slowdown effect and this PR intentionally doesn't add it to nocturine with the concern of balance.

TeenSarlacc:
what if during the delay, nocturine made you dizzy and slur your words, like alcohol minus the poison and vomiting?

Nukesthestation:
i had a round as a head rev where i was able to obtain noct + hypo pen (thief) using this I was able to easily kill and then convert hos, convert cap, convert qm almost roundstart, and a bit after that cmo and then was able to just cause it to cascade of getting more converts (ended up with 47 conversions from just me, other head had 12) but because of the noct, we got basically half of command converted immediately with 0 counterplay to the noct

Goo-Jargn:
> i had a round as a head rev where i was able to obtain noct + hypo pen (thief) using this I was able to easily kill and then convert hos, convert cap, convert qm almost roundstart, and a bit after that cmo and then was able to just cause it to cascade of getting more converts (ended up with 47 conversions from just me, other head had 12) but because of the noct, we got basically half of command converted immediately with 0 counterplay to the noct. so IMO it really does need a nerf

Yeah, from what ive seen noct is always in this hot plate. (Or was the saying hot seat, I cannot recall)
People go "NERF." others say "no its fine!"
"The balance is the cost." and "it feels like shit to play against." 
And just repeat and repeat and repeat.
Genuinely, I just think nothings gonna get done with the stuff because of how controversial it is.
Personally, I just think the stuff as a concept is fundamentally flawed, Slams change would make it feel a touch better to way better depending on the timer.
Theres two constants in ss14, tiders gonna tide and nocturine being busted.

TeenSarlacc:
> so you can yell on the radio.

"aw damn, cap's drunk again" :godo:

superjj18:
This is like the only silent-ish kidnapping tool that requires preplanning(need a cuffs and a private place), opportunity(your victim needs to be alone with you in a place that won't be seen by passerby), and the robustness to not fuck it up(need to cuff the victim, need to strip the victim, etc).

If a syndie has done these three things they kinda deserve the chance to do whatever shit they had planned? This pr scoffs at the idea of "don't be alone with strangers" isn't a perfectly acceptable form of counterplay, it 100% is.

The reality is if someone can nocturn you, they could probably also just shoot you in the back of the head with roughly the same level of effectiveness and chance of success.  At very least nocturn isn't a murderbone/round removal item, it actually enables cool roleplay(if the antag isn't lame).

So I'm kinda against this as much as I hate nocturn happening to me.  Wasn't deathrattle from tracking implants recently removed for a very similar reason of "crew doesn't deserve a chance to call out for help when a antag gets the jump on them"?

pheenty:
It should probably also slur the speech and slow down above the threshold then

EmoGarbage404:
test fail is indeed real.

EmoGarbage404:
Undergoing discussion
https://forum.spacestation14.com/t/add-a-sleep-delay-to-nocturine/20994

slarticodefast (commented):
From a code perspective I think this takes the wrong approach. Rather than having status effects apply with a delay I think it would be better to have a metabolism condition for applying an effect when a certain amount of reagent has been metabolized. Currently we only have a condition for the total unmetabolized reagent amount.

Or use the drowsy effect like emo said. The way I implemented it was pretty basic and mostly a placeholder because I did not want to make the PR even bigger in size.
Drowsiness should also show you an alert icon so you get more feedback what is going on.

northerndakotas:
> Currently, nocturne is used in one of 2 ways.
> 
> You noct someone, steal their shit and run. Or you noct them, and gib them. Both of which have a 500ms reaction window for the person on the other end.
> 
> The game regrettably needs to be balanced around LRP, where the antags are, in fact, lame.

I get that those are kinda lame, but noct really is the only option for a syndi that wants to be an actual silent assassin. 

My favorite antag gimmick is noct+pen+voice mask+fake mindshield. This was already plenty hard to use effectively since you needed a secluded location completely out of shouting range, had to not flub the cuffs and headset, and had to hope the AI wasn't watching. And even with all those difficulties, it still cost you most of your TC to quietly kill one person (three at best). This PR would make this kind of playstyle basically impossible. 

At the same time, I feel like it will do very little to stop the lame plays you are describing. For someone that just wants to noct and minibomb someone, or to quickly steal their gear off them, what is 6 seconds more when your target is heavily slowed anyway? The person getting minibombed still will get gibbed with no counterplay, the only difference is there's a slightly higher chance the killed will get caught. I doubt that would do a lot to stop these kinds of antags.

If you still want to nerf it in this way, could you at least skip the 6 seconds if 16u are injected? Or something along those lines. At least then I would have the option to waste all of that TC to make my silent weapon silent.

SlamBamActionman:
Had a quick chat with Ed and this PR seems to be better served by waiting until the [refactor](https://github.com/space-wizards/space-station-14/pull/37238) is done, so that I can implement the Drowsiness changes and fix any bugs without worrying about disturbing the refactor.

SlamBamActionman:
This PR has been superseded by https://github.com/space-wizards/space-station-14/pull/40231 since it uses the new effect system.

PR #37942, opened 2025-05-29
fix: Block EntityStorage from inserting into mechs

### LINKED ISSUE #37194: Ripley mech can Round Remove borgs/people
## Description

Borgs can round remove borgs/people if they enter something like a recharger or a coffin, as they can get picked up by the clamp and then there is no way to get out. If you break the mech they are gone forever.
**Reproduction**

Enter a ripley, find a borg or somebody in like a container (the shift that it happened on it was a borg in a recharger), Pick them up with the clamp, watch as they can't get out no matter what you do.
**Screenshots**

N/A
**Additional context**

Was on salamander 5/4/2025

perryprog:
## About the PR

Nominally dumps contents of EntityStorages onto the ground inside of inside a mech when they are opened (usually by someone inside the EntityStorage).

## Why / Balance

Fixes #37194.

## Technical details

This reworks the EntityStorage removal logic to throw an event on attempts to dump its storage into another container. This allows systems to choose whether that makes sense for that to happen, and if not, they can cancel it causing the contents to be dumped out onto the grid or map.

This also means the hard-coded exception for HandsComp can be moved over to the HandsSystem which is nice.

Testing various circumstances, everything seems okay, even including the forbidden items that are EntityStorage.
- Opening a locker in a mech that even has hydraulic clamps and a person in it drops everything to the ground.
- You can escape an unlocked lamp held by someone.
- You can escape an unlocked lamp that holds you and steel that's in a disposal unit. Contents spill into the disposal unit, so you can then leave it while the steel is still in the disposal unit.

As a fail safe for the mech itself, on destruction it will now dump /all/ its contents, instead of just dumping its actual equipment.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Items and borgs can no longer become stuck in mechs.

### DIFF
--- a/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Drop.cs
+++ b/Content.Shared/Hands/EntitySystems/SharedHandsSystem.Drop.cs
@@ -3,6 +3,7 @@ using Content.Shared.Database;
 using Content.Shared.Hands.Components;
 using Content.Shared.Interaction;
 using Content.Shared.Inventory.VirtualItem;
+using Content.Shared.Storage.Components;
 using Content.Shared.Tag;
 using Robust.Shared.Containers;
 using Robust.Shared.Map;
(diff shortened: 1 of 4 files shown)


perryprog:
Setting P1 since round removal bugs are bad.

slarticodefast (approved):
Thanks for the bugfix!

PR #37943, opened 2025-05-30
Allow cargo request computers to send orders to the primary account

sowelipililimute:
## About the PR
This adds a third mode of operation for cargo request computers, Send To Primary.

## Why / Balance
- Cargo having a say in orders is desirable i.e. "we have 500 steel from salvagers, don't waste your money on that and let us deliver you some materials"
- People not having to walk up to Cargo to place an order is desirable
- Conclusion: Allow people to place orders in their department's request computers, but still give Cargo the final say.

## Technical details
- replace `SlipPrinter` with `Mode` because I couldn't figure out an elegant way to handle mutually incompatible behaviour between being a slip printer and being an order console that sends to primary
- hide Approve button on non-DirectOrder consoles on the client side
- change order enumeration + counting to consider orders on behalf of the account present in the primary account

## Media
grafik

## Requirements

## Breaking changes

- `CargoOrderConsoleComponent#SlipPrinter` has been replaced with `CargoOrderConsoleComponent#Mode` that specifies console behavior. Specify `DirectOrder` for `false`, `PrintSlip` for `true`, or `SendToPrimary` to match existing non-Cargo request computers and use the new behaviour.
- `CargoSystem#GetOutstandingOrderCount` is no longer static and expects to be passed an `Entity<StationCargoOrderDatabaseComponent>` and not just the plain component.

**Changelog**

:cl:
- tweak: Departmental request computers now send their orders directly to Cargo, instead of requiring paper slips be brought to Cargo for insertion

### DIFF
--- a/Content.Client/Cargo/UI/CargoConsoleMenu.xaml.cs
+++ b/Content.Client/Cargo/UI/CargoConsoleMenu.xaml.cs
@@ -203,6 +203,9 @@ namespace Content.Client.Cargo.UI
         /// </summary>
         public void PopulateOrders(IEnumerable<CargoOrderData> orders)
         {
+            if (!_orderConsoleQuery.TryComp(_owner, out var orderConsole))
+                return;
+
             Requests.DisposeAllChildren();
 
             foreach (var order in orders)
@@ -237,6 +240,7 @@ namespace Content.Client.Cargo.UI
                 row.Cancel.OnPressed += (args) => { OnOrderCanceled?.Invoke(args); };
 
                 // TODO: Disable based on access.
+                row.SetApproveVisible(orderConsole.Mode != CargoOrderConsoleMode.SendToPrimary);
                 row.Approve.OnPressed += (args) => { OnOrderApproved?.Invoke(args); };
                 Requests.AddChild(row);
             }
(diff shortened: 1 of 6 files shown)


Hitlinemoss:
Honestly, I think I prefer physical slips for two main reasons:
1.) It forces departments like science or engineering to actually interact with the cargo desk instead of just plugging orders into a computer and waiting for them to show up. Ideally, if department A depends on department B for something, departments A and B should actually interact with each other instead of just silently hoping they actually do what needs done. With digital orders, the only time someone would interact with the cargo desk is if they're annoyed their order hasn't shown up yet.
2.) Interacting with cargo at the desk makes it easier for them to just directly tell you "we're not going to buy that item" or "we have a bunch of that item in the cargo bay already", rather than them needing to send out a radio message that might not get seen.

superjj18:
Absolutely amazing PR, this is what it should have been from the start.  We even said it during the request slip pr that they basically nuked the usefulness of department cargo consoles back to only slightly more useful than pre-department economy update

Now department economy can actually be department economy and not just the "Beg cargo for shit" update

EmoGarbage404:
i love playtesting

Issue #37902, opened 2025-05-28
TEG exponential moving average needs to be replaced with a better solution


ArtisticRoomba:
## Description
Right now the TEG's power output is smoothed using an exponential moving average. This magics an insane amount of power to the grid and isn't a very good solution.

This was brought on by https://github.com/space-wizards/space-station-14/pull/37658, we merged this as a band-aid as it was causing large problems (triggering epilepsy).

It was discussed to replace this with a large thermal buffer to prevent large jumps while still keeping power output realistic to what it should be. However any solution that ensures that power isn't being magiced out of thin air while still smoothing the TEG in pressure fluttering situations is good.

**Media**
See the pull request media.

ArtisticRoomba:
So what I'm probably going to do as a naive first try:
- Add an imaginary configurable thermal mass to each circulator (clearly there's a one ton copper block hidden under the hood of each circulator)
- Gas from the pipes is transferred to that thermal mass at a certain efficiency (slowly but surely)
- Have the efficiency rating extract our desired % from those thermal masses

It'll take some tweaking as we just decreased the efficiency to make extracting energy from low specific heat gasses (oxygen and nitrogen) very hard on purpose. So if we increase the efficiency when going from pipe to buffer, we'll just be back to square one when it comes to killing single loops.

My only problem with the above solution is that the TEG naturally wants to produce and dump as much power humanly possible to the grid. This will always result in the TEG draining the thermal mass of energy before the peaks and valleys of a fluttering TEG can warm the thermal buffer. So we're proxying the heat only to follow the same behavior.

This is probably tangentially related, but we could enforce manual TEG throttling using TEG temperature ranges. Have the TEG consume thermal energy from the heat buffers as if it was matching to the grid via power ramping. However, if the circulators got too hot (from atmosians dumping too much heat into the system) it would overheat and be damaged. This would accomplish two things:
- Prevent atmosians from running everything pedal to the metal, a goal of the atmos roadmap
- Encourage atmosians to monitor their TEG and preform grid matching, another goal of the atmos roadmap (major setup and forget solutions shouldn't be a thing)

Let me know what you think of my analysis but that's something I had concern about when thinking about it a bit further.

ArtisticRoomba:
I will re-implement the TEG throttling logic which was removed in https://github.com/space-wizards/space-station-14/pull/29050/.

Game plan is:
- Gas transfers 10% of heat to some imaginary thermal mass (necessary to keep single-loops dead).
[...]

PR #37957, opened 2025-05-30
Show TEG theoretical supply on inspect

ArtisticRoomba:
## About the PR
Presents TEG theoretical power supply information alongside its current power supply on inspect.

## Why / Balance
This helps lower the bar for performing "grid matching", a practice where engineers throttle the TEG's working gas flow rate in order to match the TEG's theoretical power production to the current demand of the station. This saves fuel, and if you're doing manual burn chamber refills and relights, this reduces how often you'll do them.

Previously it was not super intuitive to do this: you had to isolate the TEG from the grid in order to multitool the HV net to get the TEG's theoretical supply information. If you didn't do this, you would likely get other information (from an active engine or otherwise) mixed in.

This just brings it into the inspect, allowing engineers to look at it real easy and just power accordingly.

This is a temporary measure until the TEG is reworked (see #37902), as it'll probably get a UI (or the sensor monitoring console might get some love? I'd have to see how I feel about it).

## Technical details
We just use PushGroup to push info on the TEG's current max supply.

Sorry for the import cleanup, Rider just does this. I can revert if necessary.

## Media
Content Client_Kh3EH6g4Tp

## Requirements

**Changelog**
:cl:
- add: The TEG now shows its capable peak power output on inspect, making it easier for engineers to match the TEG's power output with the current grid demand (to save fuel!).

### DIFF
--- a/Content.Server/Power/Generation/Teg/TegSystem.cs
+++ b/Content.Server/Power/Generation/Teg/TegSystem.cs
@@ -1,10 +1,7 @@
-﻿using Content.Server.Atmos;
-using Content.Server.Atmos.EntitySystems;
+﻿using Content.Server.Atmos.EntitySystems;
 using Content.Server.Atmos.Piping.Components;
 using Content.Server.Audio;
-using Content.Server.DeviceNetwork;
 using Content.Server.DeviceNetwork.Systems;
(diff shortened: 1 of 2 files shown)


chromiumboy:
Looks good to me. Tested fine in game

PR #37959, opened 2025-05-30
Replace usages of customTypeSerializer PrototypeIdListSerializer with something that doesn't take 20 separate words to type out

DrSmugleaf:
## About the PR

Basic programming principle of using types that define the usage of the data, instead of just "string", without having to look at the specific serializer that something is using
Also people are still using the custom type serializer attribute instead of just using ProtoId<T> and EntProtoId which is awesome, so getting rid of existing usages so that stops happening as much

## Requirements

## Breaking changes

Usages of customTypeSerializer PrototypeIdListSerializer have been replaced with ProtoId<T>. Usage of existing fields may need to be changed if for some reason you were also using a list of strings instead of ProtoId<T>.
Of note, IBiomeWorldLayer now uses ProtoId<ContentTileDefinition> and ProtoId<DecalPrototype> and IBiomeWorldLayer uses ProtoId<ContentTileDefinition> for their property types. Existing implementations of either will need to be updated to match the new type.


slarticodefast (approved):
Thanks for the cleanup!

Issue #37956, opened 2025-05-30
Chamelon Projector can cause massive slowdown and incorrectly assigns a Dynamic Physics BodyType


Princess-Cheeseballs:
## Description
Title. The Chameleon Projector incorrectly sets the player's Physics BodyType to Dynamic instead of KinematicController when anchoring and unanchoring which is bad for more reasons than just the slowdown.

This causes the PhysicsEngine to start doing unecessary calculations on the player including friction calculations which can incur massive slowdowns. 

If the player gets, stunned, pulled, interacts with a UI or many other triggers it will update their friction value from the default of 0.2 (not noticeable but still present) to a higher value, usually 6-30 (up to a massive near 80% slowdown).

This can also sometimes cause the Physics Engine to throw asserts and errors as it should. 

**Reproduction**
- Use Chameleon Projector
- Anchor yourself
- Unanchor yourself
- Find a way to remove control from your player (Stun/Computer UI/Critting yourself ect.)
- You will now move permanently slower

**Screenshots**

**Additional context**

Made a mega issue about this because this isn't the first time and most certainly wont be the last.
Players should never have the Dynamic Physics BodyType.

Princess-Cheeseballs:
Raising to P1 because this is an extremely common physics bug, and while it could be worse it but to be fixed multiple times per week by the admin team. This should be fixed asap.

Issue #37955, opened 2025-05-30
Physics Body Changes MegaIssue(?)


Princess-Cheeseballs:
## Description
I'm naming this a mega-issue because this is the second time this exact thing has caused a bug and there needs to be notes or coding conventions put in place to tell people to stop doing this exact thing. In addition since it interfaces with the movement refactor I'm going to be treating it the same as the other MegaIssue I made in that I need to search the whole codebase for examples and fix them one by one.

Basically, there are some pretty major issues that arise from physics BodyTypes being changed around poorly, the first example was the gorilla gauntlet freezing players after throwing them because it set their BodyType to static on land assuming that all Anomalies are static. This was solved.

The next is an issue with the Chameleon Projector, when anchoring it sets the player to Static (normal) but when you unanchor it sets the player's BodyType to Dynamic instead of Kinematic controller. As far as I'm aware no dynamic entities should be able to use a Chameleon Projector so this is an easy fix. I'm making a PR as I'm typing this.

However, this opens the question of, how many more instances in the codebase are there of this exact issue, of changing BodyTypes and not changing them back properly or assuming they need to be one thing when they should be another thing. These are effectively bug landmines waiting to be stepped on.

**Reproduction**

**Screenshots**

**Additional context**
Players should always be either KinematicControllers. Physics entities should be Dynamic. Moving Projectiles should be Kinematic. Stationary entities should be Static.

Pronana:
![Image](https://github.com/user-attachments/assets/c742f862-c84c-4eb8-9294-ea51444081b2)

This is the suspect in example for the ChameleonProjector which might make this an engine level issue

Pronana:
In terms of looking at just the things which use the engine level Unanchor method:

Throwing System could cause an anchored player to become a Dynamic Body.
Shields correctly set the body type after unanchoring (Heroic).
Admin TetherGun System can (maybe) cause this. 

Really the biggest issue is the Chameleon Projector which is good to know. 
However this is looking at just this method, from that one anomaly bug there could be others gameside that still cause issues.

Pronana:
Looking at SetBodyType:

- The BeamSystem should set a BodyType of Kinematic, not Dynamic (Why does a visual line used for lightning and tech anomalies need physics(?)
- SmokeSystem is also questionable for having a Dynamic bodytype, although having smoke get blown around could be cool I don't think the current system was built with that support in mind so it should be static or kinematic.
[...]

PR #37960, opened 2025-05-30
Chameleon Projector Physics Fix

Princess-Cheeseballs:
## About the PR
Fixed Chameleon Projector changing a player's BodyType from KinematicController to Dynamic upon anchoring and unanchoring while disguised.

Fixes: #37956

## Why / Balance
Lead to some physics issues involving speed and acceleration due and could easily lead to more in the right circumstances.

I chose this approach since it covers a lot more bases since there are other systems which unanchor and anchor entities and having handlers for players inside all of them would be inefficient. 

This approach helps with: #37955 by forcing KinematicController on entities that should always be KinematicControllers

## Technical details
Added a Subscriber for the InputMoverComponent to the SharedMoverController to override the setting of BodyType in Unanchoring to be Kinematic Controller instead of Dynamic.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Chameleon Projector will no longer cause you to be permanently slowed down

### DIFF
--- a/Content.Shared/Movement/Systems/SharedMoverController.cs
+++ b/Content.Shared/Movement/Systems/SharedMoverController.cs
@@ -92,7 +92,6 @@ public abstract partial class SharedMoverController : VirtualController
         MapQuery = GetEntityQuery<MapComponent>();
 
         SubscribeLocalEvent<MovementSpeedModifierComponent, TileFrictionEvent>(OnTileFriction);
-        SubscribeLocalEvent<InputMoverComponent, AnchorStateChangedEvent>(OnAnchorState);
 
         InitializeInput();
         InitializeRelay();
@@ -626,11 +625,5 @@ public abstract partial class SharedMoverController : VirtualController
             args.Modifier *= ent.Comp.BaseWeightlessFriction;
         else
             args.Modifier *= ent.Comp.BaseFriction;
-    }
-
-    private void OnAnchorState(Entity<InputMoverComponent> entity, ref AnchorStateChangedEvent args)
-    {
-        if (!args.Anchored)
-            PhysicsSystem.SetBodyType(entity, BodyType.KinematicController);
     }
 }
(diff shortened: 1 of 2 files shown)


Pronana:
This is currently an issue on master, but the chameleon projector is used rarely enough I don't think it's worth the hassle of hotfixing. 

I can rebase to stable if Maints disagree with me on that.

slarticodefast (changes requested):
I think it makes sense this way since player controlled entities should always be kinematic controllers. But this does cause a problem with artifacts: They start out as Dynamic, and when anchoring and unanchoring them they end up as KinematicControllers. Should they even have the InputMoverComponent unless they are made sentient?

Princess-Cheeseballs:
> I think it makes sense this way since player controlled entities should always be kinematic controllers. But this does cause a problem with artifacts: They start out as Dynamic, and when anchoring and unanchoring them they end up as KinematicControllers. Should they even have the InputMoverComponent unless they are made sentient?

I think in my PhysicsAssert PR I made them KinematicControllers by default since they have InputMover: #37970. This will also cause issues if brains are anchored and unanchored and other input movers that "Can't move" but I would file that under really needing a ChangedPhysicsComponent. 

Ideally both should be merged since input movers should always be KinematicControllers if they're able to move.

slarticodefast (approved):
Ok, makes sense in combination with the other one.
I'll need to do some testing for the other PR first.
This is good to merge when the other PR is approved as well.

Princess-Cheeseballs:
Raising to P1 because this is an extremely common physics bug, and while it could be worse it but to be fixed multiple times per week by the admin team. This should be fixed asap.

Content.Shared/Movement/Systems/SharedMoverController.cs
```diff
@@ -626,4 +627,10 @@ private void OnTileFriction(Entity<MovementSpeedModifierComponent> ent, ref Tile
         else
             args.Modifier *= ent.Comp.BaseFriction;
     }
+
+    private void OnAnchorState(Entity<InputMoverComponent> entity, ref AnchorStateChangedEvent args)
+    {
+        if (!args.Anchored)
+            PhysicsSystem.SetBodyType(entity, BodyType.KinematicController);
+    }
```
slarticodefast: Oh, I just noticed:
This should be in `SharedMoverController.Input.cs`

Princess-Cheeseballs: True...

PR #33887, opened 2024-12-16
Chameleon controller implant (Clothing fast switch)

beck-thompson:
## About the PR

Added a chameleon controller implant that you can get in the standard chameleon bundle!

## Why / Balance

Right now it takes a very long time to switch all the clothing one by one. This will allow you to still have fine grained control, but also quickly switch entire jobs in a matter of seconds. With the AI ever watching hopefully this lets antags using the chameleon clothing get a little more of an edge. It also uses your loadout from your profile, so you can make a decent amount of customization changes!

I'm planning on adding more stuff to this implant as well (E.g chameleon ID icon change) but wanted to keep it simple for the first PR.

## Technical details

It is a lot... Most of the file changes are just UI stuff, the thing that you should probably focus on is `ChameleonControllerSystem.cs` and `SharedStationSpawningSystem.cs`. The loadout system is confusing so there is a lot of odd logic. Basically, I get the users profile, try to find if they have a loadout for a given job, if they do, use that as the clothing, if they don't and if they don't create a new one with the default!
```
profile.Loadouts.TryGetValue(jobProtoId, out var loadout);
loadout ??= new RoleLoadout(jobProtoId);
loadout.SetDefault(profile, session, _proto);
```

Then, I got through every slot and try to get the item stored for that slot with the new `GetGearForSlot` function.
```
var proto = _stationSpawningSystem.GetGearForSlot(loadout, slot.Name) ?? ((IEquipmentLoadout) startingGearPrototype).GetGear(slot.Name);
```

If their profile doesn't have anything for that slot, I see if there is any "starting gear" for that goes in the slot.

## Media

https://github.com/user-attachments/assets/db872245-4c61-4e7b-92f7-9464e6fecc2d

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: Beck
- add: A chameleon controller implant that allows you to quickly switch between jobs clothing! Comes packaged in the chameleon  kit.

PR #37963, opened 2025-05-30
You now can only change chameleon clothing with an implant

beck-thompson:
## About the PR

Title, you now have to be implanted with the chameleon controller implant to be able to control chameleon clothing.

## Why / Balance

The main reason for this PR is to reduce metagaming of security officers. Now they can freely interact with chameleon clothing and it wont reveal itself unless you have the implant! Also to clarify: You can still manually edit specific pieces of clothing you aren't forced to use the preset outfits.

This does reduce the fun of randomly finding a piece of chameleon clothing as crew and then being able to play dressup with it but I personally think the tradeoff is worth it.

## Requirements

**Changelog**

:cl:
- tweak: You can now only interact with chameleon clothing if you are implanted with a chameleon controller implant.

### DIFF
--- a/Content.Shared/Implants/SharedChameleonControllerSystem.cs
+++ b/Content.Shared/Implants/SharedChameleonControllerSystem.cs
@@ -1,4 +1,4 @@
-﻿using Robust.Shared.Prototypes;
+﻿using Content.Shared.Clothing.EntitySystems;
 
 namespace Content.Shared.Implants;
 
@@ -11,6 +11,7 @@ public abstract partial class SharedChameleonControllerSystem : EntitySystem
         base.Initialize();
 
         SubscribeLocalEvent<ChameleonControllerOpenMenuEvent>(OpenUI);
+        SubscribeLocalEvent<ChameleonControllerImplantComponent, ImplantRelayEvent<CanAccessChameleonClothingEvent>>(OnCanAccessRelay);
     }
 
     private void OpenUI(ChameleonControllerOpenMenuEvent ev)
@@ -24,5 +25,10 @@ public abstract partial class SharedChameleonControllerSystem : EntitySystem
             return;
 
         _uiSystem.OpenUi(implant.Value, ChameleonControllerKey.Key, ev.Performer);
+    }
+
+    private void OnCanAccessRelay(Entity<ChameleonControllerImplantComponent> ent, ref ImplantRelayEvent<CanAccessChameleonClothingEvent> args)
+    {
+        args.Event.CanAccess = true;
     }
 }
--- a/Content.Shared/Clothing/EntitySystems/SharedChameleonClothingSystem.cs
+++ b/Content.Shared/Clothing/EntitySystems/SharedChameleonClothingSystem.cs
@@ -122,7 +122,13 @@ public abstract class SharedChameleonClothingSystem : EntitySystem
 
     private void OnVerb(Entity<ChameleonClothingComponent> ent, ref GetVerbsEvent<InteractionVerb> args)
     {
-        if (!args.CanAccess || !args.CanInteract || ent.Comp.User != args.User)
+        if (!args.CanAccess || !args.CanInteract)
+            return;
+
+        var ev = new CanAccessChameleonClothingEvent();
+        RaiseLocalEvent(args.User, ref ev);
+
+        if (!ev.CanAccess)
             return;
 
         // Can't pass args from a ref event inside of lambdas
@@ -213,4 +219,10 @@ public abstract class SharedChameleonClothingSystem : EntitySystem
             }
         }
     }
+}
+
+[ByRefEvent]
+public sealed class CanAccessChameleonClothingEvent : EntityEventArgs
+{
+    public bool CanAccess = false;
 }
--- a/Content.Shared/Implants/SharedSubdermalImplantSystem.cs
+++ b/Content.Shared/Implants/SharedSubdermalImplantSystem.cs
@@ -9,6 +9,7 @@ using Robust.Shared.Containers;
 using Robust.Shared.Network;
 using Robust.Shared.Prototypes;
 using System.Linq;
+using Content.Shared.Clothing.EntitySystems;
 
 namespace Content.Shared.Implants;
 
@@ -33,6 +34,7 @@ public abstract class SharedSubdermalImplantSystem : EntitySystem
         SubscribeLocalEvent<ImplantedComponent, MobStateChangedEvent>(RelayToImplantEvent);
         SubscribeLocalEvent<ImplantedComponent, AfterInteractUsingEvent>(RelayToImplantEvent);
         SubscribeLocalEvent<ImplantedComponent, SuicideEvent>(RelayToImplantEvent);
+        SubscribeLocalEvent<ImplantedComponent, CanAccessChameleonClothingEvent>(RelayToImplantEvent);
     }
 
     private void OnInsert(EntityUid uid, SubdermalImplantComponent component, EntGotInsertedIntoContainerMessage args)


Nukesthestation:
Will this affect thief gloves?

beck-thompson:
> Will this affect thief gloves?

Bruh yes it will I'll fix that :saluting_face:

beck-thompson:
> There are a few options here. Either you're SOL without a chameleon implant (not great), the implant could be an uplink item in itself (1 TC?), or the gloves/mask come in a box where you set the chameleon visuals once when you open the box (sort of like the undetermined thieving satchel). Thieves are probably shafted if they aren't given some solution.

Yeah my idea was to just have those get an exception, I feel like having an implant would kind of suck for just one item. It is a little unintuitive, but I think its the best solution?

SlamBamActionman:
> Yeah my idea was to just have those get an exception, I feel like having an implant would kind of suck for just one item. It is a little unintuitive, but I think its the best solution?

I think it's fine for now, but I'd hesitate dropping the metashield just yet until they get a more stable longterm solution.

Content.Shared/Implants/SharedChameleonControllerSystem.cs
```diff
@@ -25,4 +26,9 @@ private void OpenUI(ChameleonControllerOpenMenuEvent ev)
 
         _uiSystem.OpenUi(implant.Value, ChameleonControllerKey.Key, ev.Performer);
     }
+
+    private void OnCanAccessRelay(Entity<ChameleonControllerImplantComponent> ent, ref ImplantRelayEvent<CanAccessChameleonClothingEvent> args)
+    {
+        args.Event.CanAccess = true;
+    }
```
IProduceWidgets, line 33: Without a component bool to require the implant it will kill uses like #36894

IProduceWidgets (changes requested):
Without a component bool to require the implant #36894

Content.Shared/Implants/SharedChameleonControllerSystem.cs
beck-thompson: yeah same with like the thieving gloves, I just forgot about them but like I said in the other comment I'll just add a bool

SlimmSlamm:
I'd be fine with this change if the chameleon kits came with the implant or it was free.

beck-thompson:
> I'd be fine with this change if the chameleon kits came with the implant or it was free.

It does!

superjj18:
Unfun but necessary, maybe we can have a second version of Cham clothing without this restriction so we can still enjoy the rare silliness when it shows up in rare loot in stuff like like salv or grand lottery

Edit: emag interaction?

SlamBamActionman:
Worth mentioning explicitly;

The Chameleon Controller Implant was merged in #33887 and is already part of the chameleon bundle. So this is extending the functionality of that existing implant, not introducing a new one.

> Unfun but necessary, maybe we can have a second version of Cham clothing without this restriction so we can still enjoy the rare silliness when it shows up in rare loot in stuff like like salv or grand lottery

Afaik outside antags, chameleon is only available to crew on the syndicate shuttle wreck? I don't think it's unreasonable to just replace that gamerloot with something else.

northerndakotas:
> I think this is a good way to move towards getting rid of chameleon items from the metashield, but it does raise some questions. Chameleon thieving gloves and voice masks are also chameleon, but of course those don't come with an implant.

Doesn't the existence of the implant already break the metashield on chameleon items? If implants aren't metashielded, thus everyone knows that chameleon controller implants exist, wouldn't they also know what can be controlled by the implant?

beck-thompson:
> imo the implant locking should only enable if someone with an implant equips it / gets an implant with it equipped
> so thieving gloves voice mask etc still work and gamer loot dress up can still happen

I'm leaning more towards this solution! Seems like the best of both worlds

walksanatora:
make it lockable so you could be a thief or something and give a cham thing to a syndicate friend without them having to buy a whole nother implant just to change the color of their gloves.

superjj18:
>Afaik outside antags, chameleon is only available to crew on the syndicate shuttle wreck? I don't think it's unreasonable to just replace that gamerloot with something else.

Unreasonable? Not at all.
Unfun? Definitely

keronshb:
> I'm not sure if I like this... if security were to find someone with a voice mask or gloves with Cham fibers they can immediately try and implant check you to remove the functionality from a lot of your equipment.

Could just make the implant permanent.

But then there's the issue of the implanter laying around. Though that could be solved by allowing the recycler to get rid of them.

Cojoke-dot:
> Could just make the implant permanent.
> 
> But then there's the issue of the implanter laying around. Though that could be solved by allowing the recycler to get rid of them.

At that point just make it where only antags can use the cham function. No point to even have it as an implant if you can't remove it and the implanter is destroyed that easily. 

If it's not permanent though then a secoff could use the implant to super easily check people if the cham gear metashield is removed. 

Idk about this still, it feels weird to have an implant decide if you can use your gear. It's too easily exploitable without a metashield in place, which I assume this pr would lift for cham gear.

superjj18:
> But then there's the issue of the implanter laying around. Though that could be solved by allowing the recycler to get rid of them.

Im convinced the people who complain about being unable to hide a implanter simply lack creativity.  Just shove the shit in a plushie and space it lol

"Oh but then salv might find it" the average salv is strapped with so much contraband that they honestly shouldn't even be talking sec considering sec could nuke salv the second they decide "salv really shouldn't have contraband"

walksanatora:
> Just shove the shit in a plushie and space it lol

or put it in any of the many potted plants of the station. or in a plush in a plant. or put it in a un-powered disposals bin. soo many options to hide it even just on station (cause I have had cargo find my spaced ones one too many times)

Errant-4 (commented):
I agree with the goal of this change but I strongly disagree with the solution. I think making chameleons unusable without an implant is too restrictive and goes against the spirit of the game

I have an alternate solution:  give all chameleons voice trigger functionality, which removes (or adds back) the Chameleon verb ((and also the voice trigger verbs/examine info)). Not only makes this possible to properly hide your chameleon gear and allows us to remove the metashield, it adds a layer of cloak-and-dagger to chameleon equipment because now you need to consider if you want to have a codeword that you can easily speak to disable the UI in plain sight, but risk it being unlocked unintentionally (or possibly by others, depending on how tiny we make the pickup range on it); vs a "distinct" word, that will possibly signal to Security that you are doing chameleon shenanigans and they will just repeat it right back at your gear.

The chameleon controller should allow setting the verbs for and disabling all worn chameleons at the same time, to make setting up several of them with a voice keyword not be absolute pain. Potentially it could allow you to also toggle the lock on your chameleons _with your mind_ without having to say the word.

I wanted to make a quick demo for this to sell the idea, but it was hamstrung by TriggerSystem not being in Shared. However, I'm told that Slarti is currently moving that to Shared, so this might not need remain a dream for that long

archee1:
while i agree it sucks to get meta checked by the verb menu, it is a direct rulebreak of 4.4(?) metashield stealth items, and should be dealt via ahelp. typically when someone sees you running around in cham items that make no sense like a lone commanders helmet with the expensive coat and non-functioning galoshes, it makes sense to them they're gonna want to check its a chameleon item.

I think a better solution would be a re-useable manual that teaches you how to alter chameleon items, which can be learned similar to how spellbooks used to function, where you do a brief doafter then gain the ability, chameleon gimmicks are already inherently weak with how easily most people become aware of them, adding an extra implanter you have to dispose of makes them even weaker in that regard.

Also with implants being currently out of the metashield due to implanters providing a full list of implants, this wouldnt technically cover them better under the metashield since now chameleon clothes are a thing that exists by extension of the knowledge of the chameleon controller implant. (unless this doesnt apply, however i have seen many people state and infer that implants are all unshielded now due to the implanter changes)

SlamBamActionman:
> while i agree it sucks to get meta checked by the verb menu, it is a direct rulebreak of 4.4(?) metashield stealth items, and should be dealt via ahelp. typically when someone sees you running around in cham items that make no sense like a lone commanders helmet with the expensive coat and non-functioning galoshes, it makes sense to them they're gonna want to check its a chameleon item.

The goal is to move away from admin-enforced game mechanics. The metashield is more often than not a crutch to make up for a mechanic not being sufficiently meta-proof, so it should not be used as an argument against a mechanic trying to remove it.

> I think a better solution would be a re-useable manual that teaches you how to alter chameleon items, which can be learned similar to how spellbooks used to function, where you do a brief doafter then gain the ability, chameleon gimmicks are already inherently weak with how easily most people become aware of them, adding an extra implanter you have to dispose of makes them even weaker in that regard.

The implanter is already added for the sake of the full chameleon outfit change. Yes, it is another item to dispose of, but if you get a chameleon set you are disposing of things regardless. 

Additionally, a skill book would still need to be discarded.

> Also with implants being currently out of the metashield due to implanters providing a full list of implants, this wouldnt technically cover them better under the metashield since now chameleon clothes are a thing that exists by extension of the knowledge of the chameleon controller implant. (unless this doesnt apply, however i have seen many people state and infer that implants are all unshielded now due to the implanter changes)

Implants are unshielded. I agree it butts heads with the stealth metashield to a degree, but that is only until we have mechanics to sufficiently delete chameleon stuff from the metashield.

For now I imagine admins rule it the same as the uplink; you can know uplinks exist, but you can't go around confiscating PDAs for it.

beck-thompson:
With the updated metashield this needs a more comprehensive solution

PR #37968, opened 2025-05-30
fix dragon devour action

### LINKED ISSUE #37944: Dragon can no longer devour airlocks/walls/etc
## Description

Devour no longer works on walls/airlocks/etc. Reported on Discord.

**Reproduction**

Try using devour on anything that's not a human/animal mob.

@deltanedas 

**Screenshots**

**Additional context**

deltanedas:
## About the PR
- fixed not being able to eat walls, fixes #37944
- fixed it not having a hand overlay because it never had one apparently
- fixed the action not having a whitelist so the green/red interaction outline is now correct except for mobs because for whatever reason they cant have one when not drag-dropping??
- can no longer target yourself it didnt work anyway due to being alive

## Media
the hand overlay is real

the outline is real

eating bodies is still real

would cl for wall eating but its not on stable so

### DIFF
--- a/Content.Shared/Actions/SharedActionsSystem.cs
+++ b/Content.Shared/Actions/SharedActionsSystem.cs
@@ -417,15 +417,13 @@ public abstract class SharedActionsSystem : EntitySystem
             return comp.CanTargetSelf;
 
         var targetAction = Comp<TargetActionComponent>(uid);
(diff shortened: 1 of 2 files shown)


ScarKy0 (approved):
I love chewing on metal

PR #37971, opened 2025-05-31
EVENT BASED WEIGHTLESSNESS

Princess-Cheeseballs:
## About the PR
Weightlessness is no longer determined every tick plus every single other time something wants to know if its weightless or not.

Instead Weightlessness is tracked through the WeightlessnessComponent, having the Component with Weightless set to true causes you to be weightless, otherwise you will never be weightless.

## Why / Balance
IsWeightless having to be solved for and raising inventory relayed events every tick is perhaps a bad thing so I changed it. There were also a bunch of notes saying that it needed to be done and now it is being done. 

I also changed "gravityState" to "weightless" in the MovementIgnoreGravityComponent because it doesn't make sense to have "gravityState: true" mean an entity is weightless.

## Technical details
New Component WeightlessnessComponent.
Weightless status is now updated by RefreshWeightless which raises events and is only updated when it needs to be (changing grids, wearing magboots/moonboots, component additions and changes)

## Requirements

## Breaking changes

Weightlessness is no longer calculated every tick, you will need to update weightlessness with RefreshWeightless or SetWeightless in SharedGravitySystem. 

IsWeightless has been changed to check for a cached weightlessness value, it no longer raises events or checks the grid. Any entity that cares about being weightless will need the GravityAffectedComponent or else they will always be considered not weightless. 

MovementIgnoreGravity has had its Datafield name changed to weightless

GravityComponent now autogenerates its network state and has had EnabledVV removed in exchange for just Enabled. 

**Changelog**

no cl no fun


Pronana:
There's gonna be test fails I can feel it. But this time I wont explode my branch fixing them.

Pronana:
All tests pass :D

EmoGarbage404:
@TemporalOroboros please explain why you close PRs in the future. This is not in accordance with the PR review guidelines. As such, I'm reopening.

TemporalOroboros:
I think I fat fingered the close button on mobile and didn't notice. Sorry, my bad.

Princess-Cheeseballs:
Raising to p1 because of liltenhead...

Content.Shared/Gravity/WeightlessnessComponent.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Robust.Shared.GameStates;
+
+namespace Content.Shared.Gravity;
+
+/// <summary>
+/// This Component allows a target to be considered "weightless" when Weightless is true. Without this component, the
+/// target will never be weightless.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class WeightlessnessComponent : Component
```
FairlySadPanda:
```suggestion
public sealed partial class CanBeWeightlessComponent: Component
```

Nitpick: This component specifically covers the possibility that an entity can be weightless, not that it actually is - that is done via the Weightless property of this component. This risks confusing YAML warriors.

Content.Shared/Gravity/WeightlessnessComponent.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// If true, this entity will be considered "weightless"
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool Weightless = true;
```
FairlySadPanda: Nitpick: If this component is intended to be attached and make something weightless, but be disableable, then expressing it that way around is clearer. That would make `public bool Weightless = true;` turn into `public bool Disabled;`, and would turn the name of this class into `WeightlessComponent`.

FairlySadPanda:
I've dropped the difficulty to D2; the bulk of the work is in SharedGravitySystem, and none of it is especially odious, although it needs context on how gravity works.

Content.Shared/Gravity/WeightlessnessComponent.cs
EmoGarbage404: im preferential to "GravityAffectedComponent"

Content.Shared/Gravity/SharedGravitySystem.cs
```diff
@@ -1,171 +1,284 @@
...
+public record struct IsWeightlessEvent(bool IsWeightless = false, bool Handled = false) : IInventoryRelayEvent
+{
+    SlotFlags IInventoryRelayEvent.TargetSlots => ~SlotFlags.POCKET;
+}
+
+/// <summary>
+/// Raised on an entity when their weightless status changes.
+/// </summary>
+[ByRefEvent]
+public record struct WeightlessnessChangedEvent(bool Weightless);
```
EmoGarbage404: make this event readonly

Content.Shared/Throwing/ThrowingSystem.cs
```diff
@@ -236,7 +236,7 @@ public void TryThrow(EntityUid uid,
         RaiseLocalEvent(user.Value, ref pushEv);
         const float massLimit = 5f;
 
-        if (pushEv.Push || _gravity.IsWeightless(user.Value))
```
EmoGarbage404, line 245: if this was meant to be moved to the gravity system, then either remove the impulse call or return the isweightless check

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
```diff
@@ -390,7 +390,7 @@ private void AttemptShoot(EntityUid user, EntityUid gunUid, GunComponent gun)
         var shooterEv = new ShooterImpulseEvent();
         RaiseLocalEvent(user, ref shooterEv);
 
-        if (shooterEv.Push || _gravity.IsWeightless(user, userPhysics))
```
EmoGarbage404, line 394: ditto

Content.Shared/Gravity/SharedGravitySystem.cs
```diff
@@ -1,171 +1,284 @@
...
-            return _gravityQuery.HasComp(entity.Comp.GridUid) ||
-                   _gravityQuery.HasComp(entity.Comp.MapUid);
-        }
+        return true;
+    }
 
+    public bool IsWeightless(Entity<WeightlessnessComponent?> entity)
+    {
+        // If we can be weightless and are weightless, return true, otherwise return false
+        return Resolve(entity, ref entity.Comp, false) && entity.Comp.Weightless;
```
EmoGarbage404: nit: would be nice to use an entityquery here since performance is a concern wrt mob stuff

Content.Shared/Gravity/SharedGravitySystem.cs
```diff
@@ -1,171 +1,284 @@
...
-                }
-                else
-                {
-                    _alerts.ClearAlert(uid, WeightlessAlert);
-                }
-            }
-        }
+        // Only update if we're changing our weightless status
+        if (entity.Comp.Weightless == weightless)
+            return;
```
EmoGarbage404: i dont get why a `bool?` is getting passed in here. This should just call `TryWeightless` (ideally named something better, or just put into here) and then it should set the weightless value to the result of that function call. The early exit seems very strange to have here, since the whole point of a refresh method is _not_ to directly set a value.

Content.Shared/Gravity/SharedGravitySystem.cs
```diff
@@ -1,171 +1,284 @@
...
+
+        if (newWeightless == entity.Comp.Weightless)
+            return;
+
+        entity.Comp.Weightless = newWeightless;
+        Dirty(entity);
+
+        var ev = new WeightlessnessChangedEvent(entity.Comp.Weightless);
+        RaiseLocalEvent(entity, ref ev);
+    }
```
EmoGarbage404: Just make this a normal setter function
- name it SetWeightless
- boolean input value
- keep the guard clause, event raise, dirty, etc.

alternatively, roll this into the refresh method since it doesn't have a purpose to exist on its own.

Content.Shared/Gravity/SharedGravitySystem.cs
```diff
@@ -1,171 +1,284 @@
...
+            return;
+        component.EnabledVV = state.Enabled;
+        var ev = new GravityChangedEvent(uid, component.EnabledVV);
+        RaiseLocalEvent(uid, ref ev, true);
+    }
+
+    private void OnGetState(EntityUid uid, GravityComponent component, ref ComponentGetState args)
+    {
+        args.State = new GravityComponentState(component.EnabledVV);
+    }
```
EmoGarbage404: can we do auto state handling in the year of our lord 2025

Content.Shared/Throwing/ThrowingSystem.cs
Princess-Cheeseballs: Not sure what you mean? The component for weightlessness now subscribes to the event removing the need for an IsWeightless check

Content.Shared/Gravity/SharedGravitySystem.cs
Princess-Cheeseballs: there's a rule that if you don't touch it it doesn't exist.

PR #37973, opened 2025-05-31
Minor revisions to planet and setmapatmos commands

VerinSenpai:
## About the PR
Made some changes to setmapatmos and planet commands to make them consistent with other LECs. Also moved their respective ftl files to the commands locale folder and renamed them accordingly. The persistence locale folder has been sent to the big recycle bin in the sky.

Some questions:
First, there is another ftl in the maps locale folder called gamemap.ftl with a locale string (gamemap-could-not-use-map-error) that does not appear to be used anywhere, or at least I couldn't find any usages of it. Can this be removed?

Secondly, do we have a preference for using EntityManager.System or dependencies? I see some inconsistencies with how these are used such as in PlanetCommand.cs where SharedMapSystem is brought in as a depen but BiomeSystem is brought in with EntManager or is there a reasoning behind the differences?

## Requirements

### DIFF
--- a/Content.Server/Atmos/Commands/SetMapAtmosCommand.cs
+++ b/Content.Server/Atmos/Commands/SetMapAtmosCommand.cs
@@ -14,10 +14,7 @@ public sealed class AddMapAtmosCommand : LocalizedEntityCommands
     [Dependency] private readonly IEntityManager _entities = default!;
     [Dependency] private readonly SharedMapSystem _map = default!;
 
-    private const string _cmd = "cmd-set-map-atmos";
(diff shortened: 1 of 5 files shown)


VerinSenpai:
I'll make that happen then. I thought that was how it was but I noticed it didn't get done or perhaps was confused and thought it was done during cleanup.

VerinSenpai:
Just realized I mistakenly clumped this in with my resolve removal prs. That was a mistake. This only brings these two commands up to what I'm calling par for LEC commands.

PR #37985, opened 2025-05-31
Everyone have access to escape pod

Laserkaier:
## About the PR
I removed the external access requirements from the evac pods.

## Why / Balance
Greytides can now be evacuated on evac pods. No downside.

## Technical details
AirlockExternalGlassShuttleEscape moved to airlocks.yml

## Requirements

## Breaking changes

**Changelog**

:cl: 
- fix: Now everyone has access to evacuation pods.


Hitlinemoss:
I do like this change, but I will note that this will affect gameplay balance by allowing antagonists to more easily escape the station unseen by sec without needing to get externals access somehow.

deltanedas:
you can just break a window to do that anyway this is just needed because of a recent docking bug

IProduceWidgets:
> you can just break a window to do that anyway this is just needed because of a recent docking bug

Do you have an issue to link? I think last time we had docking port issues with pods it was just a race condition between stationinit / gridspawncomponent and mapinit.

This might not be a great solution because (aside from being slightly bandaid-y) we're trying to move away from holofans, and not all escape pods have full airlocks yet. So its possible opening one when the pod is gone/spaced will space the area.

IProduceWidgets:
Someone put a lot of work into this similar idea which I think is a better designed solution to the issue at hand. #30143

Essentially; it would be neat if when the shuttle arrived (certain?) airlocks became AA.

PR #37987, opened 2025-05-31
Linear Damping final fixes

### LINKED ISSUE #37804: Linear and Angular Damping Mega Issue
## Description
A lot of systems in the codebase relied on the fact that friction was applied from two different sources to items (tileFrictionController and the PhysicsIslandSolver) resulting in a general assumption that airFriction would be 0, and that if you wanted something to be frictionless, you would set the LinearDamping to 0 and set its body to be in the air. 

This has instead been changed to where friction is applied through LinearDamping (the way shuttles do/did it) and now LinearDamping while in the air is 0.2 (what LinearDamping was usually before) and 0.05 when offGrid (what mobFriction was when offGrid).

Because of this, many systems which set LinearDamping or AngularDamping either on MapInit, ComponentInit, or through YAML no longer work properly as that value gets overwritten every tick, examples include:

Shuttles slowing down in space because they initially had no tileFriction and applied their friction through setting the linearDampingValue directly (immediately gets overwritten)

Admin Pinball smite now has friction (Should be getting fixed soon)

A cursory search of the codebase shows there's more examples which exist that are very niche due to the friction being applied now being very small, and it often being the case that these physics bodies don't exist long enough to matter.

Any System (aside from the TileFrictionController) that sets LinearDamping or AngularDamping needs to be fixed up either by just removing the SetLinearDamping or SetAngularDamping and accepting it now has a small bit of friction applied, or by properly applying friction modifiers such that the friction being applied is the same as it was before (It may be necessary to calculate this if the system is complex enough). 

**Reproduction**
- Search the codebase for Damping
- Cry

**Screenshots**

**Additional context**
I caused this, it's all my fault. This is my cross to bear.

Princess-Cheeseballs:
## About the PR
Went through the repository and searched for the last few straggler still setting LinearDamping or AngularDamping manually. 

Or which mistakenly had the wrong friction value (usually something that wasn't supposed to have friction but now had a little bit of friction).

Helps fix: #37804

Mostly yaml changes with some code deletions. 

Made it so movers without the MovementSpeedModifierComponent have the correct default friction value.

Added a comment

## Why / Balance

Code cleanup.

I didn't touch AdminSmites or ImmovableRodSystem because there are already PRs open which fix those specific issues and more. 

GhostPlushie I set to be weightless always because the original damping values were 1/10th of the in air damping values, and it had a friction modifier of 1/10th. 

## Requirements

## Breaking changes

**Changelog**

people wont notice these extremely minor friction changes


Princess-Cheeseballs:
The toys.yml conflict has arrived

Princess-Cheeseballs:
Never trust Rider when the merge master -> merge conflicted PR goes well.

PR #37989, opened 2025-06-01
Adding a new trait: Monochromacy!

TheFlyingSentry:
## About the PR
This adds the Trait Monochromacy a new trait (or disability depending on who you ask) that makes you see in black and white

## Why / Balance
The honest reason was i saw it on the TG wiki and I LOVE sight related disabilities (Blindness my beloved) and after a bit of research I noticed that it would be really simple to add since the noir glasses PR added a black and white filter

## Technical details
oops all yaml

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added a new disability called Monochromacy, total colorblindness.

### DIFF
--- a/Resources/Locale/en-US/traits/traits.ftl
+++ b/Resources/Locale/en-US/traits/traits.ftl
@@ -15,6 +15,9 @@ permanent-blindness-trait-examined = [color=lightblue]{CAPITALIZE(POSS-ADJ($targ
 trait-lightweight-name = Lightweight drunk
 trait-lightweight-desc = Alcohol has a stronger effect on you.
 
+trait-monochromancy-name = Monochromancy
+trait-monochromancy-desc = You are fully colorblind, everything you perceive ranges from blacks to whites.
+
 trait-muted-name = Muted
 trait-muted-desc = You can't speak.
 
--- a/Resources/Prototypes/Entities/Mobs/Player/clone.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/clone.yml
@@ -17,6 +17,7 @@
   # traits
   # - LegsParalyzed (you get healed)
   - LightweightDrunk
+  - Monochromacy
   - Muted
   - Narcolepsy
   - Pacified
--- a/Resources/Prototypes/Traits/disabilities.yml
+++ b/Resources/Prototypes/Traits/disabilities.yml
@@ -45,6 +45,14 @@
       cloneable: true
 
 - type: trait
+  id: Monochromancy
+  name: trait-monochromancy-name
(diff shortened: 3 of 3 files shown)


TheFlyingSentry:
Currently the overlay does not deal with examining, things inside your inventory... or overlays in general
but if I'm honest I'm happy with it in it's current state

beck-thompson (approved):
Only sus thing is not shading the inventory / right click menu - I personally think that's fine though it would be really hard to do

slarticodefast:
Hmm, it does kinda make the glasses a little redundant. How about using a more fitting colorblindness effect instead?

TheFlyingSentry:
i think it's best to keep the glasses for people who play all jobs with the same character and still want the noir look for det

slarticodefast:
I kinda intended the glasses to be a special item for the detective only, since it fits the noir theme. So I'm just a little sad to see them become useless.

TheFlyingSentry:
That's fair.  I dont want to take away the novelty of the noir glasses since this would give other jobs, medical for instance a harder time since their entire job revolves around knowing the color of chems at a glance (once that gets fixed since right clicking as of now dosent show the chem in black and white)

TheFlyingSentry:
also in my honest opinion the noir glasses would make a perfect thief target
poor det having everything stolen from them...

TheFlyingSentry:
I know that for some people vision affecting traits tend to mess with them alot.
Monochromacy just so happens to be black and white which tends to not mess with people as badly as say vulps deuteranopia (of which depending on the server has a setting to turn it off)

slarticodefast:
> Monochromacy just so happens to be black and white which tends to not mess with people as badly as say vulps deuteranopia 

Oh, I was about to offer making a proper colorblindness shader we could have used for the trait instead.

+1 for the thieving objective though

TheFlyingSentry:
OH i wouldn't mind a full on colorblindness shader in you're fine with making in that case, sure!

TheFlyingSentry:
The only problem i see with adding other colorblindness effects is people going into it not knowing what they're in for and having to leave half way through a shift because they can't stomach it

slarticodefast:
Emo came up with a good idea, I'll make the noir glasses use a sin city shader instead 😄 
So we can still have this as a trait
sin-city-2-screen1

BigfootBravo:
Vision based traits my beloved! So many possibilities for traits, or just other filters and such!

TeenSarlacc:
does having the trait affect you observing reagents, or can you still see the reagent color?

TheFlyingSentry:
You can still see the reagent color when you examine it in any way sadly

PR #38000, opened 2025-06-01
Added bulk cleanades crate

Hitlinemoss:
## About the PR

Added the bulk cleanades crate, containing two cleanade boxes for $400.

## Why / Balance

Cleanade boxes are already purchasable from cargo via the janitorial supplies crate ($560). However, the other contents of said crate are kind of useless unless a department really needs a full janitorial kit:
- Buckets, mops, trash bags, and wet floor signs are all printable via autolathe
- Space cleaner spray bottles are refillable, the janitor's space cleaner dispenser holds 5000u of space cleaner, chemistry can make even more space cleaner, and the spray bottles themselves can be printed via autolathe anyway
- Mop buckets, plungers, and soap can't be printed from an autolathe, but they're all infinite-use items (soap can technically be eaten or liquefied, but a janitor generally has no reason to do so)

When a janitorial supply crate is ordered, it's usually just for the single box of cleanades, which means the rest of the box's contents generally sit around in the cargo bay and take up space (unless someone remembers to sell them or give them to some other department or something). Having an explicit cleanades crate reduces clutter; if someone just wants cleanades, they can just buy the cleanades crate.

The price of the bulk cleanades crate was determined as follows:
- The janitorial supply crate costs $560, but is only actually worth $170 (including the price of the crate itself); i.e, the cost is about 3.3x the sale price.
- A plastic crate containing two cleanades would be worth $120 (including the price of the crate itself); 3.3x that is $396, which I rounded up to $400

## Technical details

- Added CrateServiceCleanerGrenades to Resources\Prototypes\Catalog\Fills\Crates\service.yml
- Added ServiceCleanerGrenades to Resources\Prototypes\Catalog\Cargo\cargo_service.yml

## Media

moss-cleanadescrate 1
moss-cleanadescrate 2

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Cargo can now order the bulk cleanades crate, containing two cleanade boxes for $400.

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_service.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_service.yml
@@ -11,9 +11,8 @@
 - type: cargoProduct
   id: ServiceCleanerGrenades
   icon:
-    sprite: Objects/Storage/boxes.rsi
-    state: box
-    state: flashbang
+    sprite: Objects/Weapons/Grenades/janitor.rsi
+    state: icon
   product: CrateServiceCleanerGrenades
   cost: 400
   category: cargoproduct-category-name-service
--- a/Resources/Prototypes/Catalog/Cargo/cargo_service.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_service.yml
@@ -9,6 +9,17 @@
   group: market
 
 - type: cargoProduct
+  id: ServiceCleanerGrenades
+  icon:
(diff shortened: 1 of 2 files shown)


EmoGarbage404 (approved):
might be nice to break open a lot of the bulk crates into smaller cheaper crates so this is a good step.

beck-thompson (approved):

PR #38006, opened 2025-06-01
Cargo food crate adjustments

Hitlinemoss:
## About the PR

Most cargo food orders have been increased in price. In addition, the disaster pizza delivery and softdrinks bulk crate have been removed entirely.

Specific price changes:
- emergency pizza delivery ($450 -> $1600)
- Happy Honk meal delivery ($1400 -> $2800)
- ice cream delivery ($1200 -> $2000)
- MRE crate ($1000 -> $1400)
- snowcone delivery ($1100 -> $2000)
- softdrinks crate ($1200 -> $1500)

## Why / Balance

Food ordered from cargo is really, really cheap. This is especially notable with the emergency pizza delivery, which can essentially feed a department for a whole shift for under $500. Cargo being able to buy the station food in bulk for cheap can invalidate the kitchen, which is ostensibly supposed to be the department providing proper food to the station. At the same time, though, food crates shouldn't be so expensive that they can't be purchased when the kitchen is unavailable (e.g kitchen gets bombed, chef gets arrested), or that it'd be impractical for department heads to reward their employees. Essentially, these price changes are intended to strike a balance where cargo ordering food is useful in emergencies / for fun, but not a practical way to feed the entire station long-term.

Prices were calculated by taking the sale value of the crates' contents, multiplying by 6, rounding to the nearest increment of $100, and adding $100 on top of that to cover the value of the crate itself. The MRE crate is multiplied by 3 instead because nutribricks/brownies are stated to be cheap, long-lasting, shit-tasting foods that should logically be less expensive than "luxury" foods like pizza or ice cream. (Snowcones and soft drinks actually sell for $0, so I had to kind of just eyeball those ones; I kept the price of the softdrinks crate relatively modest since it can technically be used as a bar restock and I don't want to make it expensive for actual service workers to order ingredients they want to use for their jobs.)

The disaster pizza delivery and softdrinks bulk crate were removed since they were essentially identical to just ordering multiple "normal-sized" crates (disaster pizza delivery is 4x the price of emergency pizza delivery and contains 4x the pizzas; softdrinks bulk crate is 2x the price of softdrinks crate and contains 2x the soda), which means they kind of just clutter up the item list without much benefit.

In the future, I might want to adjust the contents of some of these crates further (maybe making it so that the pizza / Happy Honk crates only contain one meal each, and reducing their prices to compensate), but I want to get this broader-scope PR looked at first.

## Technical details

- Adjusted prices in Resources\Prototypes\Catalog\Cargo\cargo_food.yml
- Removed CrateFoodPizzaLarge and CrateFoodSoftdrinksLarge from Resources\Prototypes\Catalog\Cargo\cargo_food.yml and Resources\Prototypes\Catalog\Fills\Crates\food.yml
- Added CrateFoodPizzaLarge and CrateFoodSoftdrinksLarge to Resources\migration.yml
- Replaced CrateFoodPizzaLarge with CrateFoodPizza in Resources\Prototypes\Procedural\salvage_rewards.yml
- Replaced FoodPizzaLarge and FoodSoftdrinksLarge with FoodPizza and FoodSoftdrinks in Resources\Prototypes\GameRules\cargo_gifts.yml (increased the number of crates so that the total quantity of food is the same)

## Media

moss-newfoodprices

## Requirements

## Breaking changes

CrateFoodPizzaLarge and CrateFoodSoftdrinksLarge have been removed. I've added them to Resources\migration.yml and replaced references to them in other files to attempt to prevent any problems arising from this.

**Changelog**

:cl:
- tweak: Most cargo food orders are now more expensive.
- remove: Disaster pizza deliveries and softdrinks bulk crates have been removed.

### DIFF
--- a/Resources/Prototypes/GameRules/cargo_gifts.yml
+++ b/Resources/Prototypes/GameRules/cargo_gifts.yml
@@ -63,9 +63,9 @@
     description: cargo-gift-pizza-large
     dest: cargo-gift-dest-bar
     gifts:
-      FoodPizzaLarge: 1              # 16 pizzas
+      FoodPizza: 4              # 16 pizzas
       FoodBarSupply: 1
-      FoodSoftdrinksLarge: 1
+      FoodSoftdrinks: 2         # 32 sodas
 
 - type: entity
   id: GiftsEngineering
--- a/Resources/Prototypes/Procedural/salvage_rewards.yml
+++ b/Resources/Prototypes/Procedural/salvage_rewards.yml
@@ -8,7 +8,7 @@
     CrateMaterialSteel: 1.0
     # things the station might want
     CrateEngineeringAMEJar: 0.25
-    CrateFoodPizzaLarge: 0.25
+    CrateFoodPizza: 0.25
     CrateFoodSoftdrinks: 0.25
     CrateFunInstrumentsVariety: 0.25
     CrateSalvageEquipment: 0.25


slarticodefast (approved):
Yeah, makes sense to me. Completely making half the service department redundant with a single cargo order should not be that cheap. And in case there is no service crew you can still order them, just at a higher price, or through vending machine restocks.

SlamBamActionman (approved):
One day we will have a fully realized statistic calculation of what the correct cargo crate price should be.

But until then, this PR seems to be adjusting things in the right direction. Thank you for contributing!

PR #38011, opened 2025-06-02
Fixes electropacks not ignoring insulation

TheFlyingSentry:
## About the PR
Electropacks used to not ignore insulation, now they do

## Why / Balance
This makes electropacking an engineer or john tider not a joke.
Yes there is a comment in SharedElectrocutionSystem that states that theres a problem with people being electrocuted gaining control over the mouse but this seems more niche of an issue compared to being able to bypass electropacks with insulated gloves
## Technical details
Yaml moment

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines

## Breaking changes

**Changelog**
:cl:
- tweak: Electropacks now shock people regardless of if they have insulated gloves or not

### DIFF
--- a/Content.Server/Explosion/EntitySystems/TriggerSystem.cs
+++ b/Content.Server/Explosion/EntitySystems/TriggerSystem.cs
@@ -152,7 +152,7 @@ namespace Content.Server.Explosion.EntitySystems
                 return;
             }
 
-            _electrocution.TryDoElectrocution(containerEnt, null, shockOnTrigger.Comp.Damage, shockOnTrigger.Comp.Duration, true);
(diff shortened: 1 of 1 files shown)


TheFlyingSentry:
image
the comment in question

TheFlyingSentry:
I will also note that if you wear a chef hat with a non sentient mouse you're able to completely ignore any type of stun

slarticodefast (approved):
Makes sense to me, wearing a clothing item that is intended to directly shock you should not be circumvented by wearing gloves.
The code comment does not even apply here: It is saying to not to add InsulatedComponent to mice, as wearing them on your head would make you immune to shocks via the relay event similarly how insuls do. But setting `ignoreInsulation` to `true` means that relay event is disabled.

slarticodefast:
Oh, and could you make the changelog a little more clear? Something like "Insulated gloves no longer protect you from electro packs".

TheFlyingSentry:
I've updated the changelog o7

PR #37990, opened 2025-06-01
Fixed some issues around 'IsTileBlocked'

Cojoke-dot:
## About the PR

Fixes #34828
Fixes #35675
Fixes #34829
Also fixes being able to pry tiles under people, I guess nobody realized you were still meant to be able to do that
## Why / Balance
buuugggssss, too lazy to make 2 different prs for the fixes when one would require another. I need this merged for yet ANOTHER pr too
## Technical details

'IsTileBlocked' now uses a 'Box2' instead of 'Box2Rotated' when calling 'GetEntitiesIntersecting'. No idea why 'Box2Rotated' was not working, it was not returning anything from the 'GetEntitiesIntersecting'. Other systems seem to use 'Box2' anyways

Also changed 'MobMask' to 'MobLayer' because 'IsTileBlocked' checks the layer, not the mask.
## Media

https://github.com/user-attachments/assets/b6bb7f27-68f8-444e-a10e-4f9e2ae8acd3

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

'IsTileBlocked' now uses a 'Box2' instead of a 'Box2Rotated' when calling 'GetEntitiesIntersecting'

**Changelog**

:cl:
- fix: Mimes can no longer spawn walls in places they shouldn't
- fix: You can't pry up tiles underneath mobs anymore

PR #37939, opened 2025-05-29
Kills TurfHelpers

TemporalOroboros:
## About the PR

This PR moves all of the remaining usages of `TurfHelpers` to use either `TurfSystem`, `EntityLookupSystem`, or both.
It also deletes the `TurfHelpers` static class.

## Why / Balance

1. `TurfHelpers` relied on a few obsolete `MapGridComponent` methods and moving usages over to a system helps ECS `MapGridComponent`.
2. Progress on [PZW](https://github.com/space-wizards/space-station-14/issues/33279).

## Technical details

1. Both variants of the `TurfHelpers.GetTileRef` extension method were moved to `TurfSystem` and all uses thereof were changed to use the system method.
2. Both variants of the `TurfHelpers.GetContentTileDefinition` extension method were moved to `TurfSystem` and all uses thereof were changed to use the system method.
3. Both variants of the `TurfHelpers.IsSpace` extension method were moved to `TurfSystem` and all uses thereof were changed to use the system method.
4. The remaining uses of `TurfHelpers.GetEntitiesInTile` were inlined using `TurfSystem.GetTileRef`, `EntityLookupSystem.GetWorldBounds`, and `EntityLookupSystem.GetEntitiesIntersecting`.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

`TurfHelpers` is kill. Use the equivalent methods on `TurfSystem` or `EntityLookupSystem` as needed.

PR #37786, opened 2025-05-24
Cleanup warning in `SpiderSystem`

Tayrtahn:
## About the PR

Fixes 1 warning in `SpiderSystem.cs`.

## Why / Balance

https://github.com/space-wizards/space-station-14/issues/33279

## Technical details

Replaces `EntityCoordinates.GetEntitiesInTile` with `EntityLookupSystem.GetEntitiesIntersecting`.

## Media

Shiva making some webs:
Screenshot 2025-05-24 at 1 30 05 PM

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

PR #38022, opened 2025-06-02
fix tile placement not being blocked by walls

Errant-4:
## About the PR
It's no longer possible to place tiles under anchored walls/windows/grilles (anything with the Impassable collisiongroup)

## Why / Balance
This was actually not meant to be possible, but the check that tested for this broke at some point.
This broken test was by now also obsolete, so the fix also removes an obsolete call, line goes down!
This was also annoying engineers, who could lose tiles by accidentally clicking on the wall, and then not be able to get it back without partially deconstructing the wall.

## Technical details
FloorTileSystem used an obsolete call to query every entity on the tile, and then check if they were walls/fully blocking. However, this query no longer returns any results (I did not bother to investigate why, because the call is also obsolete), so it never sees the walls. It has now been replaced with a short-range GetEntitiesInRange

## Media
https://github.com/user-attachments/assets/1ccd63c9-b7ad-4aa7-8e2c-9b5ee69836ed

## Requirements

## Breaking changes
None

**Changelog**
:cl: Errant
- fix: Tiles can no longer be placed under existing walls, so they will no longer eat your steel if you misclick.

### DIFF
--- a/Content.Shared/Tiles/FloorTileSystem.cs
+++ b/Content.Shared/Tiles/FloorTileSystem.cs
@@ -106,7 +106,7 @@ public sealed class FloorTileSystem : EntitySystem
         // otherwise check it isn't blocked by a wall
         if (!canAccessCenter)
         {
-            foreach (var ent in location.GetEntitiesInTile(lookupSystem: _lookup))
+            foreach (var ent in _lookup.GetEntitiesInRange(location, 0.05f))
             {
                 if (physicQuery.TryGetComponent(ent, out var phys) &&
                     phys.BodyType == BodyType.Static &&


IProduceWidgets:
GetEntitiesInTile doesn't see walls? Thats goofy

IProduceWidgets:
There's only 3* more uses in content, maybe time to take it to the woodshed?
https://github.com/search?q=repo%3Aspace-wizards%2Fspace-station-14%20GetEntitiesInTile&type=code

image

Errant-4:
> GetEntitiesInTile doesn't see walls? Thats goofy

It never went into that loop, but like I said, I did not look too long after I saw I should rip that function out anyway. I'll look into replacing the other 3 instances in a separate PR

Cojoke-dot:
> It never went into that loop, but like I said, I did not look too long after I saw I should rip that function out anyway. I'll look into replacing the other 3 instances in a separate PR

I had a similar issue in #37990, It looks like 'GetEntitiesIntersecting' doesn't return anything if it uses a 'Box2Rotated'. I didn't understand the system well enough to try and fix it in that pr.

slarticodefast:
Wouldn't this only work for entities that have their coordinate center in the middle of the tile?
Whereas the previous method checked for everything within a polygon shaped like the tile.
Also see my comment here https://github.com/space-wizards/space-station-14/pull/37786

Errant-4:
> Wouldn't this only work for entities that have their coordinate center in the middle of the tile? Whereas the previous method checked for everything within a polygon shaped like the tile. Also see my comment here #37786

Technically true, however. It is impossible for a tile-sized anchored structure to not have its center be in the center of the tile. Can you think of or show a practical example where this will be a problem?

In any case, I do plan to fix up GetEntitiesInTile, but it's going to have to be part of a bigger refactor. In the meanwhile, this will work

slarticodefast:
> It is impossible for a tile-sized anchored structure to not have its center be in the center of the tile. Can you think of or show a practical example where this will be a problem?

The local center can have an arbitrary offset. The recently removed tile aligned barsigns would be an example. Although it doesn't have to be multi-tile for that.

Errant-4:
I will also say that even if this fails to work in some arguably obscure cases, it is far better than the current code which never works. And like I mentioned, I am working on a proper solution but I think this should be merged meanwhile, because this was raised by a player as an active engineering annoyance

Errant-4:
#37939 made this fix obsolete, but also did not actually fix the issue, so back to the drawing board

iaada:
Are you planning to continue this? I tracked the bug down to here https://github.com/space-wizards/RobustToolbox/issues/6068

Errant-4:
@iaada  Not going to work on this in the foreseeable future, perhaps I should just close it. Feel free to take a stab at the bug if you have the time

PR #38023, opened 2025-06-02
Better robotics console

Samuka-C:
## About the PR
Added a option in the RoboticsConsoleComponent to disable the control buttons.
Added a HP gauge to the robotics console UI.
Empty MMI wont count as having a brain in the robotics console UI.

## Why / Balance
The option to disable the control buttons is gonna be useful for my xenoborgs later since I only want the xenoborg computer to be useful to check the status of the xenoborgs and not actually destroy them.
The HP gauge is a quality of life.
I believe if a borg has a empty MMI inside it shouldn't broadcast that it has a brain installed, is just confusing.

## Technical details
New boolean datafield in RoboticsConsoleComponent that is used to determine the visibility of the buttons in the UI.
The BorgTransponder system now calculates the HP percentage similar to the show health overlay and send that data via network to the robotics console just like the charge data and then is displayed with similar text format.
The BorgTransponder system now also checks to see if the brain entity has a brain-slot and if it's empty, it returns that it doesn't have a brain installed.

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl: Samuka
- fix: The Robotics Console now correctly displays the brain installed information if the borg has a empty MMI
- tweak: The Robotics Console now shows the borg physical integrity

### DIFF
--- a/Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
+++ b/Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
@@ -28,6 +28,8 @@ public sealed partial class RoboticsConsoleWindow : FancyWindow
 
     public EntityUid Entity;
 
+    private bool _allowBorgControl = true;
+
     public RoboticsConsoleWindow()
     {
         RobustXamlLoader.Load(this);
@@ -72,6 +74,7 @@ public sealed partial class RoboticsConsoleWindow : FancyWindow
     public void UpdateState(RoboticsConsoleState state)
     {
         _cyborgs = state.Cyborgs;
+        _allowBorgControl = state.AllowBorgControl;
 
         // clear invalid selection
         if (_selected is {} selected && !_cyborgs.ContainsKey(selected))
@@ -85,8 +88,8 @@ public sealed partial class RoboticsConsoleWindow : FancyWindow
         PopulateData();
 
         var locked = _lock.IsLocked(Entity);
-        DangerZone.Visible = !locked;
-        LockedMessage.Visible = locked;
+        DangerZone.Visible = !locked && _allowBorgControl;
+        LockedMessage.Visible = locked && _allowBorgControl; // Only show if locked AND control is allowed
     }
 
     private void PopulateCyborgs()
@@ -137,7 +140,9 @@ public sealed partial class RoboticsConsoleWindow : FancyWindow
         BorgInfo.SetMessage(text);
 
         // how the turntables
-        DisableButton.Disabled = !(data.HasBrain && data.CanDisable);
+        DisableButton.Disabled = !_allowBorgControl || !(data.HasBrain && data.CanDisable);
+        DestroyButton.Disabled = !_allowBorgControl;
+        DangerZone.Visible = _allowBorgControl; // <-- Hide the whole button area if not allowed
     }
 
     protected override void FrameUpdate(FrameEventArgs args)
--- a/Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
+++ b/Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
@@ -130,11 +130,20 @@ public sealed partial class RoboticsConsoleWindow : FancyWindow
             _ => "blue"
         };
 
+        var hpPercentColor = data.HpPercent switch {
+            < 0.2f => "red",
+            < 0.4f => "orange",
+            < 0.6f => "yellow",
+            < 0.8f => "green",
+            _ => "blue"
+        };
+
         var text = new FormattedMessage();
         text.AddMarkupOrThrow($"{Loc.GetString("robotics-console-model", ("name", model))}\n");
         text.AddMarkupOrThrow(Loc.GetString("robotics-console-designation"));
         text.AddText($" {data.Name}\n"); // prevent players trolling by naming borg [color=red]satan[/color]
         text.AddMarkupOrThrow($"{Loc.GetString("robotics-console-battery", ("charge", (int)(data.Charge * 100f)), ("color", batteryColor))}\n");
+        text.AddMarkupOrThrow($"{Loc.GetString("robotics-console-hp", ("hp", (int)(data.HpPercent * 100f)), ("color", hpPercentColor))}\n");
         text.AddMarkupOrThrow($"{Loc.GetString("robotics-console-brain", ("brain", data.HasBrain))}\n");
         text.AddMarkupOrThrow(Loc.GetString("robotics-console-modules", ("count", data.ModuleCount)));
         BorgInfo.SetMessage(text);
--- a/Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
+++ b/Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
@@ -49,11 +49,11 @@ public sealed partial class BorgSystem
             var hpPercent = 1f;
 
             // gaze upon the horrible if ladder
-            if (_entityManager.TryGetComponent(uid, out DamageableComponent? damageable))
+            if (_entityManager.TryGetComponent<DamageableComponent>(uid, out var damageable))
             {
-                if (_entityManager.TryGetComponent(uid, out MobStateComponent? mobState))
+                if (_entityManager.TryGetComponent<MobStateComponent>(uid, out var mobState))
                 {
-                    if (_entityManager.TryGetComponent(uid, out MobThresholdsComponent? mobThresholds))
+                    if (_entityManager.TryGetComponent<MobThresholdsComponent>(uid, out var mobThresholds))
                     {
                         if (CalcProgress(uid, mobState, damageable, mobThresholds) is not { } deathProgress)
                             hpPercent = 0f;
--- a/Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
+++ b/Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
@@ -1,3 +1,4 @@
+using Content.Shared.Containers.ItemSlots;
 using Content.Shared.DeviceNetwork;
 using Content.Shared.Damage;
 using Content.Shared.FixedPoint;
@@ -22,6 +23,7 @@ public sealed partial class BorgSystem
     [Dependency] private readonly IEntityManager _entityManager = default!;
     [Dependency] private readonly MobStateSystem _mobStateSystem = default!;
     [Dependency] private readonly MobThresholdSystem _mobThresholdSystem = default!;
+    [Dependency] private readonly ItemSlotsSystem _itemSlotsSystem = default!;
 
     private void InitializeTransponder()
     {
@@ -68,7 +70,8 @@ public sealed partial class BorgSystem
                 }
             }
 
-            var hasBrain = chassis.BrainEntity != null && !comp.FakeDisabled;
+            // checks if it has a brain and if the brain is not a empty MMI (gives false anyway if the fake disable is true)
+            var hasBrain = CheckBrain(chassis.BrainEntity) && !comp.FakeDisabled;
             var canDisable = comp.NextDisable == null && !comp.FakeDisabling;
             var data = new CyborgControlData(
                 comp.Sprite,
@@ -215,5 +218,27 @@ public sealed partial class BorgSystem
         }
 
         return (0, true);
+    }
+
+    /// <summary>
+    /// Returns true if the borg has a brain and fakeDisabled is false
+    /// </summary>
+    private bool CheckBrain(EntityUid? brainEntity)
+    {
+        if (brainEntity == null)
+            return false;
+
+        EntityUid realBrainEntity = brainEntity.Value;
+        if (_entityManager.HasComponent<BorgBrainComponent>(realBrainEntity) && _itemSlotsSystem.TryGetSlot(realBrainEntity, "brain_slot", out var slot) && slot != null)
+        {
+            if (slot.ContainerSlot != null)
+            {
+                return slot.ContainerSlot.ContainedEntity != null;
+            }
+            else
+                return false;
+        }
+        else
+            return true;
     }
 }
--- a/Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
+++ b/Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
@@ -1,4 +1,9 @@
 using Content.Shared.DeviceNetwork;
+using Content.Shared.Damage;
+using Content.Shared.FixedPoint;
+using Content.Shared.Mobs;
+using Content.Shared.Mobs.Components;
+using Content.Shared.Mobs.Systems;
 using Content.Shared.Movement.Components;
 using Content.Shared.Popups;
 using Content.Shared.Robotics;
@@ -14,6 +19,9 @@ namespace Content.Server.Silicons.Borgs;
 public sealed partial class BorgSystem
 {
     [Dependency] private readonly EmagSystem _emag = default!;
+    [Dependency] private readonly IEntityManager _entityManager = default!;
+    [Dependency] private readonly MobStateSystem _mobStateSystem = default!;
+    [Dependency] private readonly MobThresholdSystem _mobThresholdSystem = default!;
 
     private void InitializeTransponder()
     {
@@ -28,7 +36,7 @@ public sealed partial class BorgSystem
         var query = EntityQueryEnumerator<BorgTransponderComponent, BorgChassisComponent, DeviceNetworkComponent, MetaDataComponent>();
         while (query.MoveNext(out var uid, out var comp, out var chassis, out var device, out var meta))
         {
-            if (comp.NextDisable is {} nextDisable && now >= nextDisable)
+            if (comp.NextDisable is { } nextDisable && now >= nextDisable)
                 DoDisable((uid, comp, chassis, meta));
 
             if (now < comp.NextBroadcast)
@@ -38,6 +46,28 @@ public sealed partial class BorgSystem
             if (_powerCell.TryGetBatteryFromSlot(uid, out var battery))
                 charge = battery.CurrentCharge / battery.MaxCharge;
 
+            var hpPercent = 1f;
+
+            // gaze upon the horrible if ladder
+            if (_entityManager.TryGetComponent(uid, out DamageableComponent? damageable))
+            {
+                if (_entityManager.TryGetComponent(uid, out MobStateComponent? mobState))
+                {
+                    if (_entityManager.TryGetComponent(uid, out MobThresholdsComponent? mobThresholds))
+                    {
+                        if (CalcProgress(uid, mobState, damageable, mobThresholds) is not { } deathProgress)
+                            hpPercent = 0f;
+                        else
+                        {
+                            if (deathProgress.inCrit)
+                                hpPercent = 0f;
+                            else
+                                hpPercent = deathProgress.ratio;
+                        }
+                    }
+                }
+            }
+
             var hasBrain = chassis.BrainEntity != null && !comp.FakeDisabled;
             var canDisable = comp.NextDisable == null && !comp.FakeDisabling;
             var data = new CyborgControlData(
@@ -45,6 +75,7 @@ public sealed partial class BorgSystem
                 comp.Name,
                 meta.EntityName,
                 charge,
+                hpPercent,
                 chassis.ModuleCount,
                 hasBrain,
                 canDisable);
@@ -70,7 +101,7 @@ public sealed partial class BorgSystem
             return;
         }
 
-        if (ent.Comp2.BrainEntity is not {} brain)
+        if (ent.Comp2.BrainEntity is not { } brain)
             return;
 
         var message = Loc.GetString(ent.Comp1.DisabledPopup, ("name", Name(ent, ent.Comp3)));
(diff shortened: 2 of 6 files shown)


Warpzoned:
Could you also change the battery charge number's colour? That blue is honestly quite jarring to make out since the background is also moderately dark (Maybe a yellow would be better?)

Otherwise looks great, awesome stuff!

Samuka-C:
> Could you also change the battery charge number's colour? That blue is honestly quite jarring to make out since the background is also moderately dark (Maybe a yellow would be better?)
> 
> Otherwise looks great, awesome stuff!

I guess I could, I will see later some nice color transitions and gradients

arenaconspiracy:
That is true, the contrast on the battery readout has way too low of a contrast. Good opportunity to fix a long standing issue.

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
```diff
@@ -127,17 +130,28 @@ private void PopulateData()
...
+        text.AddMarkupOrThrow($"{Loc.GetString("robotics-console-hp", ("hp", (int)(data.HpPercent * 100f)), ("color", hpPercentColor))}\n");
         text.AddMarkupOrThrow($"{Loc.GetString("robotics-console-brain", ("brain", data.HasBrain))}\n");
         text.AddMarkupOrThrow(Loc.GetString("robotics-console-modules", ("count", data.ModuleCount)));
         BorgInfo.SetMessage(text);
 
         // how the turntables
-        DisableButton.Disabled = !(data.HasBrain && data.CanDisable);
+        DisableButton.Disabled = !_allowBorgControl || !(data.HasBrain && data.CanDisable);
+        DestroyButton.Disabled = !_allowBorgControl;
+        DangerZone.Visible = _allowBorgControl; // <-- Hide the whole button area if not allowed
```
Tayrtahn: This disables the UI for sending the message to the server, but an altered client could still spoof the message and the server would do it anyway. You'll need to check the DataField on the server when it receives the message (in `OnDestroy`, for example).

Tayrtahn (dismissed):

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
Samuka-C: ohh, okay then, so just deny any attempts to destroy or disable if the allowBorgControl is false, right?

Samuka-C: done

Samuka-C:
image

what you guys think?

the background color is the same as the UIs in ss14 for reference

Warpzoned:
> image
> 
> what you guys think?
> 
> the background color is the same as the UIs in ss14 for reference

Looks great in my eyes 👍

ToastEnjoyer:
PR looks good. Definitely good options.

tutoumi:
borg camera when ?

Samuka-C:
> borg camera when ?

wyci

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -38,13 +48,37 @@ public override void Update(float frameTime)
             if (_powerCell.TryGetBatteryFromSlot(uid, out var battery))
                 charge = battery.CurrentCharge / battery.MaxCharge;
 
-            var hasBrain = chassis.BrainEntity != null && !comp.FakeDisabled;
+            var hpPercent = 1f;
+
+            // gaze upon the horrible if ladder
+            if (_entityManager.TryGetComponent<DamageableComponent>(uid, out var damageable))
```
beck-thompson: Use `TryComp` instead! Also you don't need to trycomp a lot of these they are resolved in the the functions themselves so you can just not pass it in

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +185,58 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
     {
         ent.Comp.Name = name;
     }
+
+    /// <summary>
+    /// Returns a ratio between 0 and 1, and whether the entity is in crit.
+    /// </summary>
+    private (float ratio, bool inCrit)? CalcProgress(EntityUid uid, MobStateComponent component, DamageableComponent dmg, MobThresholdsComponent thresholds)
```
beck-thompson: I think making this non nullable and just returning "zero" values or whatever would be better, the if statement makes this confusing.

I also think it might be nicer to instead just return one float and do the in crit stuff inside this function. E.g if they are crit, just return 0 - this will allow you to get rid of that if statement

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +185,58 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
...
+
+        if (_mobStateSystem.IsCritical(uid, component))
+        {
+            if (!_mobThresholdSystem.TryGetThresholdForState(uid, MobState.Critical, out var critThreshold, thresholds) ||
+                !_mobThresholdSystem.TryGetThresholdForState(uid, MobState.Dead, out var deadThreshold, thresholds))
+            {
+                return (1, true);
+            }
+
+            var ratio = 1 - ((dmg.TotalDamage - critThreshold) / (deadThreshold - critThreshold)).Value.Float();
```
beck-thompson: I don't think this ratio is used if they are in crit

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +185,58 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
...
+        if (_mobStateSystem.IsAlive(uid, component))
+        {
+            if (dmg.HealthBarThreshold != null && dmg.TotalDamage < dmg.HealthBarThreshold)
+                return null;
+
+            if (!_mobThresholdSystem.TryGetThresholdForState(uid, MobState.Critical, out var threshold, thresholds) &&
+                !_mobThresholdSystem.TryGetThresholdForState(uid, MobState.Dead, out threshold, thresholds))
+                return (1, false);
+
+            var ratio = 1 - ((FixedPoint2)(dmg.TotalDamage / threshold)).Float();
```
beck-thompson: Doesn't this use the dead states threshold even tough it will always return 0 (where this is called) if they are crit?

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +185,58 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
...
+            var ratio = 1 - ((dmg.TotalDamage - critThreshold) / (deadThreshold - critThreshold)).Value.Float();
+
+            return (ratio, true);
+        }
+
+        return (0, true);
+    }
+
+    /// <summary>
+    /// Returns true if the borg has a brain and fakeDisabled is false
```
beck-thompson: I don't think you check for fake disabled in here

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +185,58 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
...
+    /// </summary>
+    private bool CheckBrain(EntityUid? brainEntity)
+    {
+        if (brainEntity == null)
+            return false;
+
+        EntityUid realBrainEntity = brainEntity.Value;
+        if (_entityManager.HasComponent<ItemSlotsComponent>(realBrainEntity) && _itemSlotsSystem.TryGetSlot(realBrainEntity, "brain_slot", out var slot) && slot != null)
```
beck-thompson: I think you can just use `GetItemOrNull` and it would cover all this logic

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: how to get the damageable, mobState and mobThresholds components then?

Samuka-C: I think I solved it

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: solved it

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: fixed this method

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: fixed this method

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: yep, it used to check but then I moved the check elsewhere and forgot the summary, already fixed now

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: done, but I couldn't just use it since then it would flag posibrains (which don't have brain slots) as not brains. so it checks if it has the ItemSlotsComponent and then tries to get the item of the brain slot

Samuka-C:
> Could you also update the blue color for the percentages? Its kind of hard to read!

already did, but just didn't update the image of the PR, gonna do that soon

LevitatingTree:
colours are definitely better

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -38,13 +48,21 @@ public override void Update(float frameTime)
             if (_powerCell.TryGetBatteryFromSlot(uid, out var battery))
                 charge = battery.CurrentCharge / battery.MaxCharge;
 
-            var hasBrain = chassis.BrainEntity != null && !comp.FakeDisabled;
+            var hpPercent = 1f;
+            if (TryComp<DamageableComponent>(uid, out var damageable) &&
+                TryComp<MobStateComponent>(uid, out var mobState) &&
```
beck-thompson: You can get rid of the `MobStateComponent` and `MobThresholdsComponent` - just dont pass them into the functions they get resolved so it works out!

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +169,36 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
...
+    /// Returns true if the borg has a brain
+    /// </summary>
+    private bool CheckBrain(EntityUid? brainEntity)
+    {
+        if (brainEntity == null)
+            return false;
+
+        // if the brainEntity.Value has the component ItemSlotsComponent then it is a MMI,
+        // in that case it trys to get the "brain" of the MMI, if it is null the MMI is empty and so it returns false
+        if (TryComp<ItemSlotsComponent>(brainEntity.Value, out var slot) && _itemSlotsSystem.GetItemOrNull(brainEntity.Value, "brain_slot") == null)
```
beck-thompson: I think you can replace this with a trycomp for the `MMIComponent` and then you can use the `BrainSlotId` from the component to the the slot name instead of just putting the string there

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
```diff
@@ -120,24 +123,35 @@ private void PopulateData()
...
```
beck-thompson: Looks like this is causing some pretty nasty misspredicts (Try to interact with the borgs on the console without unlocking it)

image

I don't have the time right now to quite figure out why but it looks like it should be an easy fix :thinking:
If you need help ask on discord

This is probably part (or all) of it. On the update state your doing different logic):
image

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +169,36 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
     {
         ent.Comp.Name = name;
     }
+
+    /// <summary>
+    /// Returns a ratio between 0 and 1, 1 when they have no damage and 0 whenever they are crit (or more damaged)
+    /// </summary>
+    private float CalcHP(EntityUid uid, MobStateComponent component, DamageableComponent dmg, MobThresholdsComponent thresholds)
+    {
+        if (_mobStateSystem.IsAlive(uid, component))
```
beck-thompson: Just invert these and turn them into guards like this:

```
if (!_mobStateSystem.IsAlive(uid, component))
    return 0;

if (!_mobThresholdSystem.TryGetThresholdForState(uid, MobState.Critical, out var threshold, thresholds))
    return 1;

return 1 - ((FixedPoint2)(dmg.TotalDamage / threshold)).Float();
```

Is it common for `TryGetThresholdForState` to fail? I assume all borg should have the threshold component but I don't know.

If its a common thing for them not to, make the float nullable and have a special case in the UI where it shows unknown for the HP.
If it should basically never happen, just make sure to log an error (and also return 1 thats fine)

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: solved

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: done

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: how do I log the error?

Samuka-C: I did it

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
Samuka-C: I tried to revert just this part but the missprediction still there

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -38,13 +48,19 @@ public override void Update(float frameTime)
             if (_powerCell.TryGetBatteryFromSlot(uid, out var battery))
                 charge = battery.CurrentCharge / battery.MaxCharge;
 
-            var hasBrain = chassis.BrainEntity != null && !comp.FakeDisabled;
+            var hpPercent = 1f;
+            if (TryComp<DamageableComponent>(uid, out var damageable))
```
beck-thompson: just move this in the function as well (and have it error if it doesn't exist)!

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +167,37 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
...
+    /// Returns a ratio between 0 and 1, 1 when they have no damage and 0 whenever they are crit (or more damaged)
+    /// </summary>
+    private float CalcHP(EntityUid uid, DamageableComponent dmg)
+    {
+        if (!_mobStateSystem.IsAlive(uid))
+            return 0;
+
+        if (!_mobThresholdSystem.TryGetThresholdForState(uid, MobState.Critical, out var threshold))
+        {
+            Log.Error($"Borg({ToPrettyString(uid)}), doesn't have critical threshold. Trace: {Environment.StackTrace}");
```
beck-thompson: you don't need the stack trace on this also I'm unsure if you should be using sawmill or `Log` :thinking:. I'll ask and keep you updated

beck-thompson (changes requested):

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: done

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: removed the stack trace, but gonna wait for your response to see if I use log or sawmill

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
Samuka-C: fixed

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -1,4 +1,10 @@
+using Content.Shared.Containers.ItemSlots;
 using Content.Shared.DeviceNetwork;
+using Content.Shared.Damage;
+using Content.Shared.FixedPoint;
+using Content.Shared.Mobs;
+using Content.Shared.Mobs.Components;
```
beck-thompson: unused

Samuka-C: pretty sure they are used, since it makes a compilation error when i remove them

Samuka-C: removed the using Content.Shared.Mobs.Components; that was actually redundant

Content.Shared/Robotics/RoboticsConsoleUi.cs
```diff
@@ -18,10 +18,12 @@ public sealed class RoboticsConsoleState : BoundUserInterfaceState
     /// Map of device network addresses to cyborg data.
     /// </summary>
     public Dictionary<string, CyborgControlData> Cyborgs;
+    public bool AllowBorgControl;
```
Fildrance, line 25: Missing xml-doc i guess.

Content.Shared/Robotics/Components/RoboticsConsoleComponent.cs
```diff
@@ -50,4 +50,10 @@ public sealed partial class RoboticsConsoleComponent : Component
     [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
     [AutoNetworkedField, AutoPausedField]
     public TimeSpan NextDestroy = TimeSpan.Zero;
+
+    /// <summary>
+    /// If false, disables the disable/destroy buttons in the UI.
```
Fildrance: please mention more specifically, that it is 'disable/destroy borg'

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
```diff
@@ -120,24 +123,34 @@ private void PopulateData()
         BorgSprite.Texture = _sprite.Frame0(data.ChassisSprite!);
 
         var batteryColor = data.Charge switch {
-            < 0.2f => "red",
-            < 0.4f => "orange",
-            < 0.6f => "yellow",
-            < 0.8f => "green",
-            _ => "blue"
+            < 0.2f => "#FF0000", // red
```
Fildrance: the fact that those are all custom hardcoded colors hurt my soul. But i guess we dont have this stuff in some palette? It would be cool tho =(

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
```diff
@@ -151,4 +164,40 @@ public void SetTransponderName(Entity<BorgTransponderComponent> ent, string name
     {
         ent.Comp.Name = name;
     }
+
+    /// <summary>
+    /// Returns a ratio between 0 and 1, 1 when they have no damage and 0 whenever they are crit (or more damaged)
+    /// </summary>
+    private float CalcHP(EntityUid uid)
```
Fildrance, line 170: I suppose it would be nice to have emag interaction - it probably should be working in the same way as borg transponder works. To make it managable i'd say that we will be better off with action that opens radial menu in which you can select and de-select which data borg wants to report to console.
I do see that it is out of scope for this PR - thats more of a proposal for futher improvement.

Content.Shared/Robotics/RoboticsConsoleUi.cs
Fildrance: and empty line between fields declaration :godo:

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
Fildrance: Looks like we have no such pre-set colors as of now... well what can you do!
Related to what needs to be changed - relaying @chromiumboy words from discussion - new colors for the charge value in the PR are not good choices, the contrast is poor between the lighter colors. I'd leave them as is (revert color related changes)
Kinda agreed with that myself :|

Fildrance (changes requested):
minor changes requested :3

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
Samuka-C: I can try to find a new palette, but the old one was horrible, even if the colors had contrast against each other, the blue and green had horrible contrast against the background which made them very bad to read.

Content.Shared/Robotics/Components/RoboticsConsoleComponent.cs
Samuka-C: what?

Content.Server/Silicons/Borgs/BorgSystem.Transponder.cs
Samuka-C: Yeah, it would be probably better left for another PR

Content.Client/Robotics/UI/RoboticsConsoleWindow.xaml.cs
Samuka-C: image

what you think of this?

Samuka-C: image

Fildrance: I request art lead assisstance to answer that question >_> @EmoGarbage404 could ya possibly help us?

Samuka-C: @EmoGarbage404 please help?

sowelipililimute: wcag contrast ratios between colours and background:
- 4.05
- 6.39
- 15.08
- 12.5
- 11.8

consider using this palette instead, which has the following contrast ratios that brings everything within a less overvibrant & legible range:
- 5.8
- 6.9
- 9.81
- 7.37
- 8.3

grafik

- FF6C7F
- EF973C
- E8CB2D
- 30CC19
- 00D3B8

Samuka-C: done, fixed

Fildrance (approved):

Fildrance:
New colors being hard-coded is a mess, but we  need to somehow kill StyleNano before fixing that i guess :( so that will do

PR #38007, opened 2025-06-02
Marble tiles

Aearo-Deepwater:
## About the PR
Adds four new floor tiles: Marble, Dark Marble, Plasmarble, and Uranium Marble. Includes tile object and in-hand sprites, and cutter recipes. 

## Why / Balance
Granted, space stations aren't generally supposed to be ritzy, but in a few select places, it'd be nice to have a bit of class - and gold and silver tiles are many things, but classy isn't one of 'em. Say you've got a visiting CentCom official you want to impress, or you've realized your captain's office isn't fancy enough, or you're the QM. Or maybe you just want the bathrooms to look nice. I dunno.

There's one important(ish) balance consideration: To stay consistent with glass shard shivs and spears, I've given the plasma and uranium tiles very slightly increased damage, including one rad damage on the uranium tile. Just in case someone decides, given the materials, that they'd rather craft uranium marble instead of a uranium spear for some reason.

## Media

**Example Scenes:**
2025-06-01-193613_485x580_scrot

2025-06-01-192531_1075x1239_scrot
(It's so embarrassing when all of the CentCom officials at your award ceremony fall asleep. Every captain's worst nightmare.)

2025-06-01-185910_1433x970_scrot

2025-06-01-184627_1094x1256_scrot

2025-06-01-184118_1426x1059_scrot

2025-05-31-030743_815x832_scrot

**Tiles:**
2025-06-01-194541_1004x987_scrot

2025-06-01-195156_624x615_scrot

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl: SeaWyrm
- add: Two varieties of marble tile can now be produced at Cutter lathes.

PR #38027, opened 2025-06-02
Big chess

PotatOverlord:
## About the PR

adds a big chess + checkers thing to cargo order.

## Why / Balance

every time there's a big chess game (most common is chessops) no one can actually see the moves so it's not that entertaining, so now everyone can comment on the captian's 3rd blunder

## Technical details

adds prototype entities
CrateFunBigChess/Checkers containing corresponding entity tables
large counterparts to all the chess + checkers pieces
and 8 stack variant to pawns so they fit, 12 stack to checkers

adds Tabletop suffix to BaseTabletopPiece (to avoid confusion)
adds stacks Large[ChessPiece] & Large[colour]checker(king)
adds CrateFunBigChess/Checkers to LowValueFunCrateSpawn
adds EntityTable BigChess and BigCheckers
 
## Media

big-chess
TODO - Checkers sprites are basically unintelligible, chess ones just look off, and they both need inhands

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: A chess and checkers set big enough to show off your amazing 1300 elo moves has been added to the cargo catalogue.


Aearo-Deepwater:
I dunno if it suits what you're going for or not - the thing about putting down carpets easily is a good point - but you might want to consider these marble tiles: #38007 They have good chessboard vibes.
(Assuming they ever get merged, that is.)

IProduceWidgets:
I suggest using a checkerboard 2x2 tile since with these you need an 8x8 room at least, and those are hard to come by, whereas at half scale you only need a 4x4 room. The only thing you lose out on is having to choose whether you want the board to work in North/South or East/West orientations. Id suggest East/West since most people are on landscape monitors.

You could maybe make them flippable to support both, but that will probably need some TLC for handling flippable item stacks.

ie. this is a north/south single tile:

PotatOverlord:
I didn't initially do the checkered half tile style because I don't really know how the carpet sprites work, but I think it's worth the hassle since as you said 8 by 8 rooms are hard to come by. Also, I'll need to figure out what to do for scaling the chess pieces to the board, but that should be fairly simple. 
I think I'll add the smaller checkered board in as well as it's larger counterpart, mainly because fully sized chess pieces (which would be needed for a full tile based board like the one aero linked) wouldn't be obtainable otherwise.

PotatOverlord:
Took a little while, but the checkerboard looks great and works very well with the carpet smoothing system. Took a little testing to get the right arrangement, but it ended up being fairly simple. I'll push and add some media soon

ghost:
😏📸🖥

PotatOverlord:
I'm not entirely sure why the food yaml files are failing the test, I haven't touched them. Should I just leave it as is or fix the error?
I also can't find the trailing whitespace in the cargo file for the life of me.

PotatOverlord:
me when i can't read

PotatOverlord:
Probably should've fixed that before requesting review, sorry

PotatOverlord:
If this doesn't get merged in the next month I'm sneaking in fixed vehicle code and there's nothing anyone can do about it

Resources/Textures/Structures/Furniture/Carpets/white_board.rsi/meta.json
```diff
@@ -0,0 +1,46 @@
+{
+  "version": 1,
```
slarticodefast: needs to be 4 spaces per indentation
same for the other meta files

Resources/Prototypes/Stacks/large_tabletop_pieces.yml
```diff
@@ -0,0 +1,61 @@
+# chess
+- type: stack
```
slarticodefast: Hmm, not sure if all of these should be stacks, the only one where it really makes sense is pawns.
It could avoid some amount of entities, but I think as long as it's not possible to order a hundred chess sets it will be fine.

Resources/Prototypes/Entities/Structures/Furniture/carpets.yml
```diff
@@ -422,4 +422,97 @@
...
+  name: "checkered carpet board"
+  components:
+  - type: Clickable
+  - type: Sprite
+    sprite: Structures/Furniture/Carpets/checker_board.rsi
+  - type: Icon
+    sprite: Structures/Furniture/Carpets/checker_board.rsi
+  - type: Destructible
+    thresholds:
+      - trigger:
```
slarticodefast: fix indentation
see our [yaml conventions](https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#yaml-conventions)
same everywhere else

Resources/Prototypes/Entities/Objects/Fun/large_tabletop_pieces.yml
```diff
@@ -0,0 +1,233 @@
...
+  components:
+  - type: Sprite
+    sprite: Objects/Fun/Tabletop/chess_pieces.rsi
+    state: b_pawn
+  - type: Stack
+    stackType: LargeBlackPawn
+
+# Stacks
+
+- type: entity
```
slarticodefast, line 139: Oh god, so many stack prototypes.
Do we really need this?

Resources/Prototypes/Entities/Objects/Fun/large_tabletop_pieces.yml
```diff
@@ -0,0 +1,233 @@
+﻿# Base item for large pieces (regular tabletop ones are exploitable)
+- type: entity
+  id: BaseLargeTabletopPiece
```
slarticodefast: Give your prototypes names that fit the conventions. The flow should go from generic to specific.
BaseChessPieceLarge
ChessPieceLargeWhiteKing
etc

the name `BaseLargeTabletopPiece` is a little confusing for this, because it is chess specific, and has nothing to do with `BaseTabletopPiece`

Resources/Prototypes/Entities/Objects/Fun/large_tabletop_pieces.yml
```diff
@@ -0,0 +1,233 @@
...
+  components:
+  - type: Sprite
+    sprite: Objects/Fun/Tabletop/chess_pieces.rsi
+
+# White chess pieces
+- type: entity
+  id: LargeWhiteKing
+  name: white king
+  description: The white king, it can move one square in any direction. If the no piece can move and this piece is threatened, then black wins; however, if it is not under threat and no piece can move then the game ends in a stalemate.
+  parent: BaseLargeTabletopPiece
```
slarticodefast: follow yaml convention
https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#yaml-conventions
type > abstract > parent > id > categories > name > suffix > description > components

Resources/Prototypes/Catalog/Fills/Crates/fun.yml
```diff
@@ -477,3 +477,51 @@
     - id: ClothingNeckMantleQM
     - id: ClothingHeadsetCargo
 
+- type: entity
+  name: big chess set
+  description: Perfect for any high profile chess game. Takes up a 4x4 area.
+  id: CrateFunBigChess
+  parent: CratePlastic
+  components:
+  - type: StorageFill
```
slarticodefast: storagefill is deprecated
use containerfill or entity tables

slarticodefast (changes requested):
I'm not an art lead, but I think the big chess piece sprites could use some respriting. They look kinda weird in this size.

And check out this website https://www.take-a-screenshot.org/

Resources/Prototypes/Stacks/large_tabletop_pieces.yml
PotatOverlord: It wasn't fitting in the crate and therefore failing the tests, and I forgot the exact amount a crate could hold so I just went for the nuclear option. I tested it now and it works with just stacking the pawns so going with that

Resources/Prototypes/Entities/Objects/Fun/large_tabletop_pieces.yml
PotatOverlord: See other stack comment

Resources/Prototypes/Stacks/large_tabletop_pieces.yml
PotatOverlord: Oh I just remembered that I wanted there to be enough for any promotions that happen but they can just order another chess set for that tbh it's not that much effort

Resources/Prototypes/Entities/Objects/Fun/large_tabletop_pieces.yml
PotatOverlord: I was thinking of making large versions of the other tabletops, but I can change it if needed. It's not that much effort to make the checkers since it's just 4 prototypes so I might do that ~~to avoid going through and renaming them all~~ to make the name generic

Tongekki:
Now that we have spray painter we don’t even need the tiles

PotatOverlord:
The tiles are the majority of the bloat tbh I'll cut them out

PotatOverlord:
I dont think this adds anything so I'm closing it.

PR #38030, opened 2025-06-03
Removed contraband level from combat medipen

ToastEnjoyer:
## About the PR
_this has always bugged me..._

the combat Medi pen was listed as syndicate contraband for absolutely no reason.
## Why / Balance
Imagine going into sec and getting 10 minutes for meds. But also because this makes absolutely no sense

## Requirements

## Breaking changes

**Changelog**
There is no way that this could be significant enough for a changelog

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Medical/hypospray.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Medical/hypospray.yml
@@ -603,7 +603,7 @@
 
 - type: entity
   name: combat medipen
-  parent: [ChemicalMedipen, BaseSyndicateContraband]
+  parent: ChemicalMedipen
   id: CombatMedipen
   description: A single-use medipen containing chemicals that regenerate most types of damage.
   components:


Simyon264:
It's marked as contra because the usually, the only way to get it is via Syndicate uplink.

RedSpyy:
> It's marked as contra because the usually, the only way to get it is via Syndicate uplink.

You can also buy a syndie balloon in uplink..This is not a reason to arrest..
I don't like the contraband description system because people press shift+lmb+lmb instead of common sense

RedSpyy:
There are quite a lot of items now that are considered contraband without any particular reason.

It's good that at least someone is trying to fix this

ToastEnjoyer:
> It's marked as contra because the usually, the only way to get it is via Syndicate uplink.

Another reason, for this PR, if security finds a combat medipen they should turn it into the contraband box, they can't keep it in their inventory because of the contraband level, and nobody but CC can authorize the use in of it, despite it being a combat medipen. I also think it's dumb that you know they're a syndicate because of a medipen.

SlamBamActionman:
I don't disagree that some items should probably be of lower contraband level to more appropriately reflect their arrest status (looking at you, Syndicate Pyjamas), however...

This is explicitly an antag combat item, and just because it isn't branded with a big red **S** does not mean it's not Syndicate contraband. As long as it's primarily an antag tool, I do not see any reason to change its contraband level.

EDIT: For future PRs, I would consider this changelog-significant since it would noticeably impact player gameplay for the area it touches.

ToastEnjoyer:
> I don't disagree that some items should probably be of lower contraband level to more appropriately reflect their arrest status (looking at you, Syndicate Pyjamas), however...
> 
> This is explicitly an antag combat item, and just because it isn't branded with a big red **S** does not mean it's not Syndicate contraband. As long as it's primarily an antag tool, I do not see any reason to change its contraband level.
> 
> EDIT: For future PRs, I would consider this changelog-significant since it would noticeably impact player gameplay for the area it touches.

Would you think that the ability for security (Or medical) to be able to manufacture combat medipens would break the balance? Because I've thought about making a PR like that, I just want to hear a maint opinion.

SlamBamActionman:
> Would you think that the ability for security (Or medical) to be able to manufacture combat medipens would break the balance? Because I've thought about making a PR like that, I just want to hear a maint opinion.

I know that there has been pretty widespread resistance to manufactured medipens in the past (primarily in the form of refillable pens) since they are remarkably strong - the only way for crew to get access to some useful combat ones is through the combat medkit order via Cargo, and even then those are fairly rare. So while I can't speak for certain, I would lean towards them being considered too balance-breaking.

Issue #27686, opened 2024-05-05
Hitscan / impacts should probably be entities


metalgearsloth:
Hitscan used to be but not anymore. Realistically I think they probably should be so they netsync properly though this will have more overhead.

PR #38035, opened 2025-06-03
Hitscans are now entities

beck-thompson:
## About the PR

Hitscans are now entities as the title suggests.

## Why / Balance

Resolves: #27686

This will allow hitscan weapons to be much more interesting and unique hitscan weapons! For example, it would now be possible to make hitscan weapons explode on hit!

https://github.com/user-attachments/assets/25051b40-2765-479c-937b-2b47947ae00b

## Technical details

This is a pretty basic implementation the rough diagram is this:
In `Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Battery.cs`, a hitscan entity is created.
In `Content.Server/Weapons/Ranged/Systems/GunSystem.cs`, the gun raises an event on the hitscan entity to notify that it has been fired.

Various components on the hitscan entity will listen for stuff and react! For example:
`HitscanFiredEvent` is raised, the `HitscanBasicRaycastSystem` listens for this and creates a raycast that hits a target. `HitscanHitEntityEvent` is now raised on the hitscan entity where `HitscanBasicDamageSystem` listens for it and deals damage to the target.

The main thing this system can't do is deal with multiple "shots" at once but I didn't think it was necessary and should probably be reworked with the rest of the gun system at once (I think it would require a big change to hitscan weapons beyond just changing a field name :thinking:) - I could also be wrong though and I can try to add something :laughing:

## Requirements

## Breaking changes

Hitscans are now entities! `HitscanPrototype` has been removed and `HitscanBatteryAmmoProvider`s field now takes an entity prototype rather than a hitscan prototype. If your guns are using basic lasers from upstream, you can leave the proto id the same (all upstream ones got changed into entities with the same id), but if you have custom ones you'll need to turn them into entities.

If your using a custom hitscan prototype, you'll have to update it to be an entity. For most you can just do:
```
- type: entity
  parent: BasicHitscan
  id: IDOFYOUROLDPROTOTYPE
  components:
  # put your old damage stuff from the prototype here:
  - type: HitscanBasicDamage
    damage:
      types:
        Heat: 28
  # put the old visual stuff here:
  - type: HitscanBasicVisuals
    muzzleFlash:
      sprite: Objects/Weapons/Guns/Projectiles/projectiles.rsi
      state: OLDSTATE
    travelFlash:
      sprite: Objects/Weapons/Guns/Projectiles/projectiles.rsi
      state: OLDSTATE
    impactFlash:
      sprite: Objects/Weapons/Guns/Projectiles/projectiles.rsi
      state: OLDSTATE
```

Take a look at `Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/hitscan.yml` for more examples!

### DIFF
--- a/Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
+++ b/Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
@@ -37,18 +37,11 @@ public record struct HitscanTraceEvent
     public EntityUid? Target;
 }
 
-
-
 /// <summary>
 /// All data known data for when a hitscan is actually fired.
 /// </summary>
 public record struct HitscanRaycastFiredData
 {
-    /// <summary>
-    /// Location the hitscan was fired from.
-    /// </summary>
-    public EntityCoordinates FromCoordinates;
-
     /// <summary>
     /// Direction that the ray was fired towards.
     /// </summary>
@@ -68,11 +61,6 @@ public record struct HitscanRaycastFiredData
     /// Player who shot the gun, if null the gun was fired by itself.
     /// </summary>
     public EntityUid? Shooter;
-
-    /// <summary>
-    /// How far the hitscan tried to go to intersect with a target.
-    /// </summary>
-    public float DistanceTried;
 }
 
 /// <summary>
--- a/Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
+++ b/Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
@@ -1,4 +1,3 @@
-using System.Linq;
 using System.Numerics;
 using Content.Shared.Administration.Logs;
 using Content.Shared.Damage.Components;
@@ -66,12 +65,10 @@ public sealed class HitscanBasicRaycastSystem : EntitySystem
 
         var data = new HitscanRaycastFiredData
         {
-            FromCoordinates = args.FromCoordinates,
             ShotDirection = args.ShotDirection,
             Gun = args.Gun,
             Shooter = args.Shooter,
             HitEntity = result?.HitEntity,
-            DistanceTried = distanceTried,
         };
 
         var attemptEvent = new AttemptHitscanRaycastFiredEvent { Data = data };


K-Dynamic:
This brings a smile to my face

Jacktastic09:
Okay hear me out, this could be a cool way to buff the laser cannon!

K-Dynamic:
> Okay hear me out, this could be a cool way to buff the laser cannon!

As well as the laser ship cannons

beck-thompson:
> Okay hear me out, this could be a cool way to buff the laser cannon!

Yep! It would allow a lot more unique abilities for hitscans - emp on hit, explode on hit, armor piercing etc...

Content.Shared/Weapons/Hitscan/Components/HitscanStunComponent.cs
```diff
@@ -0,0 +1,13 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Weapons.Hitscan.Components;
+
+/// <summary>
+/// Hitscan entities that have this component will stun the target.
```
pheenty: I presume it only deals stamina damage. To deal stamina damage and to stun are different things.

Content.Shared/Weapons/Hitscan/Components/HitscanBasicDamageComponent.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// <summary>
+    /// How much damage the hitscan weapon will do when hitting a target.
+    /// </summary>
+    [DataField(required: true)]
+    public DamageSpecifier Damage;
+
+    /// <summary>
+    /// How much damage was actually dealt to a hit entity. If null, didn't hit anything!
+    /// </summary>
+    [DataField]
```
pheenty: Is there a single reason to make this a datafield?

Content.Shared/Weapons/Ranged/Components/HitscanBatteryAmmoProviderComponent.cs
```diff
@@ -1,11 +1,11 @@
...
 
 namespace Content.Shared.Weapons.Ranged.Components;
 
 [RegisterComponent, NetworkedComponent]
 public sealed partial class HitscanBatteryAmmoProviderComponent : BatteryAmmoProviderComponent
 {
-    [ViewVariables(VVAccess.ReadWrite), DataField("proto", required: true, customTypeSerializer: typeof(PrototypeIdSerializer<HitscanPrototype>))]
-    public string Prototype = default!;
+    [DataField(required: true)]
+    public EntProtoId HitscanEntityProto;
```
pheenty, line 10: "proto" is pretty descriptive already, this would just lead to more pain for downstreams. If anything, keep the datafield name "proto" at least, so yaml is not changed

IProduceWidgets:
Seems good.
I checked and all the window penetrating stuff still works.

Stop-Signs:
Holy shit, beck actually did it. Maints speedmerge trust... 🙏

ToastEnjoyer:
This honestly amazes me of the potential amazing stuff I could PR with this, looks super good.

ToastEnjoyer:
Why the hell did my message send three times...

Content.Server/Weapons/Ranged/Systems/GunSystem.Battery.cs
```diff
@@ -111,8 +112,11 @@ private void OnBatteryDamageExamine<T>(Entity<T> entity, ref DamageExamineEvent
 
         if (component is HitscanBatteryAmmoProviderComponent hitscan)
         {
-            var dmg = ProtoManager.Index<HitscanPrototype>(hitscan.Prototype).Damage;
-            return dmg == null ? dmg : dmg * Damageable.UniversalHitscanDamageModifier;
+            var dmg = ProtoManager.Index(hitscan.HitscanEntityProto);
+            if (!dmg.TryGetComponent<HitscanBasicDamageComponent>(out var basicDamageComp, EntityManager.ComponentFactory))
```
perryprog:
```suggestion
            if (!dmg.TryGetComponent<HitscanBasicDamageComponent>(out var basicDamageComp, Factory))
```

Content.Server/Weapons/Ranged/Systems/GunSystem.cs
```diff
@@ -15,6 +15,8 @@
 using Content.Shared.Weapons.Ranged.Systems;
 using Content.Shared.Weapons.Reflect;
 using Content.Shared.Damage.Components;
+using Content.Shared.Weapons.Hitscan.Components;
+using Content.Shared.Weapons.Hitscan.Events;
```
perryprog, line 14: Bunch of unused imports in here.

Content.Shared/Weapons/Hitscan/Components/HitscanBasicDamageComponent.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// How much damage the hitscan weapon will do when hitting a target.
+    /// </summary>
+    [DataField(required: true)]
+    public DamageSpecifier Damage;
+
+    /// <summary>
+    /// How much damage was actually dealt to a hit entity. If null, didn't hit anything!
+    /// </summary>
+    public DamageSpecifier? DamageDealt;
```
perryprog: I feel like this shouldn't live on a component. Instead I'd consider just raising an event like `DamagedByHitscan` which has the damage dealt.

Content.Shared/Weapons/Hitscan/Components/HitscanBasicEffectsComponent.cs
```diff
@@ -0,0 +1,29 @@
...
+    public Color? HitColor = Color.Red;
+
+    /// <summary>
+    /// Sound that plays upon the thing being hit.
+    /// </summary>
+    [DataField]
+    public SoundSpecifier? Sound;
+
+    /// <summary>
+    /// Force the hitscan sound to play rather than potentially playing the entity's sound.
```
perryprog: It's not clear what "the entity's sound" is. (I don't know what it is :P)

Content.Shared/Weapons/Hitscan/Components/HitscanBasicVisualsComponent.cs
```diff
@@ -0,0 +1,20 @@
...
+public sealed partial class HitscanBasicVisualsComponent : Component
+{
+    [DataField]
+    public SpriteSpecifier? MuzzleFlash;
+
+    [DataField]
+    public SpriteSpecifier? TravelFlash;
+
+    [DataField]
+    public SpriteSpecifier? ImpactFlash;
```
perryprog, line 28: All three fields here should have docs. Also is "Flash" the right word for travel/impact?

Content.Shared/Weapons/Hitscan/Components/HitscanReflectComponent.cs
```diff
@@ -0,0 +1,23 @@
...
+/// Hitscan entities with this component will get reflected by certain things (E.g energy swords)
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class HitscanReflectComponent : Component
+{
+    [DataField]
+    public ReflectType ReflectiveType = ReflectType.Energy;
+
+    [DataField]
+    public int MaxReflections = 3;
```
perryprog, line 22: Docs and docs. For ReflectiveType make sure you say what it's for/what the types are (or add docs to the enum itself).

Content.Shared/Weapons/Hitscan/Components/HitscanStunComponent.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Content.Shared.Weapons.Hitscan.Components;
+
+/// <summary>
+/// Hitscan entities that have this component will deal stamina damage to the target.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class HitscanStunComponent : Component
+{
+    [DataField]
+    public float StaminaDamage = 10.0f;
```
perryprog, line 15: Docs, plus also this component name feels misleading—it doesn't stun you, it deals stamina damage.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+using Robust.Shared.Map;
+using Robust.Shared.Physics;
+
+namespace Content.Shared.Weapons.Hitscan.Events;
+
+/// <summary>
+/// Raised on the hitscan entity when "fired". This could be from reflections or from the gun.
+/// </summary>
+[ByRefEvent]
+public record struct HitscanFiredEvent
```
perryprog: Maybe "HitscanTraceEvent" would be a better name, if it can happen from reflections?

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    /// <summary>
+    /// Location the hitscan was fired from.
+    /// </summary>
+    public EntityCoordinates FromCoordinates;
+
+    /// <summary>
+    /// Direction that was fired.
```
perryprog: Nit, but it's worth noting that this is (I assume) in the direction that the ray is being fired towards.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    public Vector2 ShotDirection;
+
+    /// <summary>
+    /// Gun that was fired.
+    /// </summary>
+    public EntityUid GunUid;
```
perryprog: Nit: Probably naming it just `Gun` is fine.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// <summary>
+    /// Who shot the gun. Could be the gun itself!
+    /// </summary>
+    public EntityUid Shooter;
```
perryprog: Would it make more sense to just let it be null if there's no well-defined shooter, since you already have `GunUid`?

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+[ByRefEvent]
+public record struct HitscanHitEntityEvent
+{
+    /// <summary>
+    /// Location the hitscan was fired from.
+    /// </summary>
+    public EntityCoordinates FromCoordinates;
+
+    /// <summary>
+    /// Direction that was fired.
```
perryprog: Same note regarding vector direction.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    public Vector2 ShotDirection;
+
+    /// <summary>
+    /// The entity that got hit, if null the raycast didn't hit anyone.
```
perryprog, line 51: While I think you could merge this with the below event (see later comment) which would make this nullable, if you don't do that then:
```suggestion
    /// The entity that got hit.
```

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// <summary>
+    /// Gun that fired the raycast.
+    /// </summary>
+    public EntityUid GunUid;
+
+    /// <summary>
+    /// Who shot the gun. Could be the gun itself!
+    /// </summary>
+    public EntityUid Shooter;
```
perryprog: Same note regarding `GunUid` naming and letting `Shooter` be nullable.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// <summary>
+    /// Was this canceled? Used for stuff like reflections.
+    /// </summary>
+    public bool Canceled;
```
perryprog: This should have a clearer doc. What happens if it's canceled? Does nothing happen? If it's used for stuff like reflections, this implies that not nothing happens. This is also confusing because the event raiser doesn't actually care if it gets canceled or not—it's just for other event handlers to look at.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/// <summary>
+/// Results of a hitscan raycast - useful for visuals and things that rely on more than just what entity was hit.
```
perryprog: Whom is this raised against and when is it raised?

I do think though you could merge this event with HitscanHitEntityEvent (with a different name) which would just be the above with the distance traveled and an optional HitEntity.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicDamageSystem.cs
```diff
@@ -0,0 +1,27 @@
...
+
+public sealed class HitscanBasicDamageSystem : EntitySystem
+{
+    [Dependency] private readonly DamageableSystem _damage = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<HitscanBasicDamageComponent, HitscanHitEntityEvent>(OnHitscanHit, after: [ typeof(HitscanReflectSystem) ]);
```
perryprog: Nit:
```suggestion
        SubscribeLocalEvent<HitscanBasicDamageComponent, HitscanHitEntityEvent>(OnHitscanHit, after: [typeof(HitscanReflectSystem)]);
```

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicDamageSystem.cs
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    private void OnHitscanHit(Entity<HitscanBasicDamageComponent> hitscan, ref HitscanHitEntityEvent args)
+    {
+        if (args.Canceled || args.HitEntity == null)
```
perryprog: (Only applies if you don't merge the two hitscan hit events.)
```suggestion
        if (args.Canceled)
```

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
```diff
@@ -0,0 +1,37 @@
...
+public sealed class HitscanBasicEffectsSystem : EntitySystem
+{
+    [Dependency] private readonly SharedColorFlashEffectSystem _color = default!;
+    [Dependency] private readonly SharedGunSystem _gun = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<HitscanBasicEffectsComponent, HitscanHitEntityEvent>(OnHitscanHit, after: [ typeof(HitscanBasicDamageSystem) ]);
```
perryprog: Nit:
```suggestion
        SubscribeLocalEvent<HitscanBasicEffectsComponent, HitscanHitEntityEvent>(OnHitscanHit, after: [typeof(HitscanBasicDamageSystem)]);
```

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    private void OnHitscanHit(Entity<HitscanBasicEffectsComponent> hitscan, ref HitscanHitEntityEvent args)
+    {
+        if (args.Canceled|| Deleted(args.HitEntity))
```
perryprog: Nit:
```suggestion
        if (args.Canceled || Deleted(args.HitEntity))
```

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
```diff
@@ -0,0 +1,37 @@
...
+            return;
+
+        TryComp<HitscanBasicDamageComponent>(hitscan, out var hitscanDamageComp);
+
+        var dmg = hitscanDamageComp?.DamageDealt ?? new DamageSpecifier();
```
perryprog: You don't need to make a default DamageSpecifier here—PlayImpactSound can just take the null.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        if (dmg.AnyPositive() && hitscan.Comp.HitColor != null)
```
perryprog: Then this can just be
```suggestion
        if (hitscan.Comp.HitColor != null && dmg?.AnyPositive() == true)
```
It's a microoptimization, but we want to check that HitColor is null first since that's slightly cheaper than calling AnyPositive.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
```diff
@@ -0,0 +1,37 @@
...
+            _color.RaiseEffect(hitscan.Comp.HitColor.Value, new List<EntityUid>() { args.HitEntity }, Filter.Pvs(args.HitEntity, entityManager: EntityManager));
```
perryprog: Nit—I would also suggest wrapping this to 120 chars too though.

```suggestion
            _color.RaiseEffect(hitscan.Comp.HitColor.Value, new List<EntityUid> { args.HitEntity }, Filter.Pvs(args.HitEntity, entityManager: EntityManager));
```

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        // TODO get fallback position for playing hit sound.
+        _gun.PlayImpactSound(args.HitEntity, dmg, hitscan.Comp.Sound, hitscan.Comp.ForceSound);
```
perryprog: You could use `RayCastResults.HitPos` I think.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<HitscanBasicRaycastComponent, HitscanFiredEvent>(OnHitscanFired);
+    }
+
+    private void OnHitscanFired(Entity<HitscanBasicRaycastComponent> hitscan, ref HitscanFiredEvent args)
+    {
+        var mapCords = _transform.ToMapCoordinates(args.FromCoordinates);
+        var ray = new CollisionRay(mapCords.Position, args.ShotDirection, (int) CollisionGroup.Opaque);
```
perryprog: Should the collision mask be part of the comp?

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+        var rayCastResults = _physics.IntersectRay(mapCords.MapId, ray, hitscan.Comp.MaxDistance, args.Shooter, false).ToList();
+
+        RayCastResults? result = null;
+
+        if (rayCastResults.Any())
+            result = rayCastResults[0];
+
+        // Check if laser is shot from in a container
```
perryprog: This comment should say why we need that check (prevent shooting from inside containers that aren't in the collisionmask.)

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+        if (!_container.IsEntityOrParentInContainer(args.Shooter))
+        {
+            // Checks if the laser should pass over unless targeted by its user
+            foreach (var collide in rayCastResults)
```
perryprog: Honestly I feel like this would be more clear if you LINQ'd the whole thing—IntersectRay does sort its results so it doesn't have any real performance benefits, but in my opinon it's much more readable. Something like:
```c#
var target = args.Target;
var result = _container.IsEntityOrParentInContainer(args.Shooter)
    ? rayCastResults.FirstOrNull()
    : rayCastResults.FirstOrNull(hit => hit.HitEntity == target
                                        || CompOrNull<RequireProjectileTargetComponent>(hit.HitEntity)?.Active != true);
```
(You'd also want to remove the `.ToList()` from `rayCastResults`.)

There's also an issue with the logic you have now (which the above sample shouldn't have)—if an entity with RequiresProjectileTarget is the only entity in the raytrace, then it'll still be hit from the fall-through! You'll want to account for that if you don't go the LINQ way.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                result = collide;
+                break;
+            }
+        }
+
+        var raycastEvent = new HitscanRaycastResultsEvent
+        {
+            RaycastResults = result,
+            DistanceTried = hitscan.Comp.MaxDistance,
```
perryprog: Shouldn't this be:
```suggestion
            DistanceTried = result?.Distance == 0.0f ? result.Value.Distance : hitscan.Comp.MaxDistance,
```
(`0.0f` means nothing was hit which is... weird. Not sure if that would mess with shooting inside containers or something.)

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+            FromCoordinates = args.FromCoordinates,
+            ShotDirection = args.ShotDirection,
+            GunUid = args.GunUid,
+            Shooter = args.Shooter,
+            HitEntity = result.Value.HitEntity,
+        };
+
+        RaiseLocalEvent(hitscan, ref hitEvent);
+
+        _log.Add(LogType.HitScanHit, $"{ToPrettyString(args.Shooter):user} hit {hitEvent.HitEntity:target} using hitscan.");
```
perryprog: Would ToPrettyString the hit entity and also mention the weapon entity.
```suggestion
        _log.Add(LogType.HitScanHit,
            $"{ToPrettyString(args.Shooter):user} hit {ToPrettyString(hitEvent.HitEntity):target}"
            + $" using {ToPrettyString(args.GunUid):entity}.");
```

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
```diff
@@ -0,0 +1,88 @@
...
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<HitscanBasicVisualsComponent, HitscanRaycastResultsEvent>(OnHitscanHit, before: [ typeof(HitscanReflectSystem) ]);
+    }
+
+    private void OnHitscanHit(Entity<HitscanBasicVisualsComponent> hitscan, ref HitscanRaycastResultsEvent args)
+    {
+        var distance = args.RaycastResults?.Distance ?? args.DistanceTried;
```
perryprog: Does this work okay with `RaycastResults.Distance` being zero, as mentioned previously?

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        FireEffects(args.FromCoordinates, distance, args.ShotDirection.Normalized().ToAngle(), hitscan.Comp);
```
perryprog: Shouldn't need to normalize the direction if you're just getting the angle? (Plus pretty sure it gets passed through as already normalized.)
```suggestion
        FireEffects(args.FromCoordinates, distance, args.ShotDirection.ToAngle(), hitscan.Comp);
```

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private void FireEffects(EntityCoordinates fromCoordinates, float distance, Angle angle, HitscanBasicVisualsComponent hitscan, EntityUid? hitEntity = null)
```
perryprog: Nit, but this feels like really weird argument ordering to me. You also never use `hitEntity`. 

Considering this method deals with a bunch of angles and vectors, `angle` should be a clearer argument name.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        // Lord
+        // Forgive me for the shitcode I am about to do
```
perryprog: This is a personal pet peeve of mine so I'll be blunt—this isn't an appropriate comment. Either rewrite it so it meets your standards, or describe what the limitations of it are and how it could be addressed.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
```diff
@@ -0,0 +1,88 @@
...
+        // Effects tempt me not
+        var sprites = new List<(NetCoordinates coordinates, Angle angle, SpriteSpecifier sprite, float scale)>();
+        var fromXform = Transform(fromCoordinates.EntityId);
+
+        // We'll get the effects relative to the grid / map of the firer
+        // Look you could probably optimise this a bit with redundant transforms at this point.
```
perryprog: It feels like TryGetMapOrGridCoordinates could work here, but it's a little different from what you have, so I'm not 100%.

Content.Shared/Weapons/Hitscan/Systems/HitscanReflectSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+using Content.Shared.Weapons.Hitscan.Events;
+using Content.Shared.Weapons.Ranged.Events;
+using Content.Shared.Weapons.Reflect;
+using Robust.Shared.Random;
+
+namespace Content.Shared.Weapons.Hitscan.Systems;
+
+public sealed class HitscanReflectSystem : EntitySystem
+{
+    [Dependency] private readonly IRobustRandom _random = default!;
```
perryprog: Unused

Content.Shared/Weapons/Hitscan/Systems/HitscanReflectSystem.cs
```diff
@@ -0,0 +1,50 @@
...
+        if (!ev.Reflected)
+            return;
+
+        hitscan.Comp.CurrentReflections++;
+
+        args.Canceled = true;
+
+        var fromEffect = Transform(args.HitEntity).Coordinates;
+
+        var evnt = new HitscanFiredEvent
```
perryprog: Nit: event, not evnt

Content.Shared/Weapons/Hitscan/Systems/HitscanStunSystem.cs
```diff
@@ -0,0 +1,25 @@
...
+
+public sealed class HitscanStunSystem : EntitySystem
+{
+    [Dependency] private readonly SharedStaminaSystem _stamina = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<HitscanStunComponent, HitscanHitEntityEvent>(OnHitscanHit, after: [ typeof(HitscanReflectSystem) ]);
```
perryprog: Nit:
```suggestion
        SubscribeLocalEvent<HitscanStunComponent, HitscanHitEntityEvent>(OnHitscanHit, after: [typeof(HitscanReflectSystem)]);
```

Content.Shared/Weapons/Hitscan/Components/HitscanAmmoComponent.cs
```diff
@@ -0,0 +1,10 @@
+using Content.Shared.Weapons.Ranged;
+
+namespace Content.Shared.Weapons.Hitscan.Components;
+
+/// <summary>
+/// This component is used to indicate an entity is shootable from a hitscan weapon.
+/// This is placed on the laser entity being shot, not the gun itself.
+/// </summary>
+[RegisterComponent]
```
perryprog: Shouldn't this be networked?

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
beck-thompson: To be clear - I agree but this entire function was just copy pasted directly from the old system. Do you still think I should remove it?

perryprog: Ah, I didn't realize that (I suck at checking what code is just old code that was moved)—I personally think so, yeah. It's really not that bad.

Content.Shared/Weapons/Hitscan/Components/HitscanBasicVisualsComponent.cs
beck-thompson: I agree with the names not being amazing - I only kept them like that to match how they were previously so forks would have an easy time porting over their lasers. I'm fine either way, what do you think?

perryprog: Oh, if they were existing, then probably keep yeah. My bad for not checking deleted files at all :P

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
beck-thompson: I think that's reasonable but honestly it's pretty nice not to have a bunch of null checks in the rest of the logic :laughing:

It makes sense to me, and its also how the rest of the gun code does it (The two systems should eventually be combined one day...) - thoughts? I would lean towards keeping it like this

perryprog: Yeah honestly I'm pretty sure on this one—it's pretty plausible a system could want to handle cases where the shooter is not well-defined. It's also just semantically not very meaningful in my opinion.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicDamageSystem.cs
beck-thompson: I personally like the other way more! Its a little nicer if you add more type I feel

perryprog: It is up to you, though that is the codebase's convention for collection expressoins (including those for SubscribeLocalEvent(after:) which is why I pointed it out.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
beck-thompson: looks a little scrunky again, let me know what you think

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
beck-thompson: I truly have no clue what that TODO is talking about so I'm gonna remove it :laughing:

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
perryprog: Definitely up to preference—I personally prefer without since it's less line noise, and it helps with line length.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
perryprog: I thiiiink it's talking about a fallback for if the hit entity is deleted, which is guarded for at the start of the method.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
beck-thompson: We will cleanse the code of false self deprecating comments :)

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
beck-thompson: Does that seem better? I have a hard time with variable names :laughing:

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
beck-thompson: Going back to my other comment, I didn't write this and just moved it. I normally would be totally willing to fix up older code but I really don't have much experience with coordinate stuff so I'm would kind of be going in a little blind :laughing:

If you think its important, I can do it let me know

perryprog: Oh if this was existing code heck no; coordinate stuff has so many edge cases it's crazy.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
beck-thompson: Sorry a lot of these comments dont make a lot of sense - I made a few changes to this function so it looks different now :laughing:

I'll push after I address everything the best I can

perryprog: Ahh sorry, heard—I tend to reply to notifications as soon as I get them as if I don't there's a 40% chance I will totally forget about this PR.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
beck-thompson: Perry I'm widely known as LINQs #1 hater but this is a lot clearer (I also added a comment though :laughing:) so you win

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
beck-thompson: Sorry I'm a little confused! When I tested this, if it didn't hit anything the raycasult result would just be null. I just pushed the changes so let me know what you think - I'm pretty sure the new logic is good?

perryprog: I was confused about always setting DistanceTried to the comp's MaxDistance—shouldn't it be set to whatever distance the raycast had?

perryprog:
Awesome stuff—will give a look in the mornin—erm, a few hours though I doubt there'll be any issues.

Sorry about the non-substantive comments on stuff that you were just moving around—I gotta get better at remembering to check that.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicDamageSystem.cs
beck-thompson: my convention is better trust

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
beck-thompson: Good point I honestly forgot what was going on, I changed the logic at some point addressing your reviews and it does seem to work now so :laughing:

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
beck-thompson: I tested it, it DOES work with zero but it gives a slightly odd animation. Checking for zero and then not displaying anything works fine as well (E.g if your in a box) but it looks a little weird without feedback. I think its probably the better solution though? Let me know what you think

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
beck-thompson: perry vibes are all off on this one I can't lie... Let me know what you think, I added it in the last commit

beck-thompson:
Thanks for the super in depth review perry! I addressed all the stuff you mentioned but there is one issue I want to bring up: the `HitscanTraceEvent` and `HitscanRaycastFiredEvent` event names are AWFUL and I cannot think of better ones :sob:

They look almost and sound almost exactly the same, but they really are completely separate ideas and I'm not sure how to convey that properly. If you have some better names please let me know :laughing:

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
perryprog: As long as it doesn't like set off the nuke it should be fine! :+1:

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicRaycastSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        SubscribeLocalEvent<HitscanBasicRaycastComponent, HitscanTraceEvent>(OnHitscanFired);
+    }
+
+    private void OnHitscanFired(Entity<HitscanBasicRaycastComponent> ent, ref HitscanTraceEvent args)
+    {
+        var shooter = args.Shooter ?? args.Gun;
+        var mapCords = _transform.ToMapCoordinates(args.FromCoordinates);
+        var ray = new CollisionRay(mapCords.Position, args.ShotDirection, (int) ent.Comp.CollisionMask);
+        var rayCastResults = _physics.IntersectRay(mapCords.MapId, ray, ent.Comp.MaxDistance, shooter, false).ToList();
```
perryprog:
```suggestion
        var rayCastResults = _physics.IntersectRay(mapCords.MapId, ray, ent.Comp.MaxDistance, shooter, false);
```
No need to force evaluation of the IEnumerable here.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
```diff
@@ -0,0 +1,86 @@
...
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<HitscanBasicVisualsComponent, HitscanRaycastFiredEvent>(OnHitscanHit, before: [ typeof(HitscanReflectSystem) ]);
+    }
+
+    private void OnHitscanHit(Entity<HitscanBasicVisualsComponent> hitscan, ref HitscanRaycastFiredEvent args)
+    {
+        FireEffects(args.FromCoordinates, args.DistanceTried, args.ShotDirection.ToAngle(), hitscan.Comp);
```
perryprog: Actually could arguably inline this method but I think it's fine. Just wanted to point it out.

Content.Shared/Weapons/Hitscan/Systems/HitscanReflectSystem.cs
```diff
@@ -0,0 +1,48 @@
...
+        if (!ev.Reflected)
+            return;
+
+        hitscan.Comp.CurrentReflections++;
+
+        args.Canceled = true;
+
+        var fromEffect = Transform(args.HitEntity.Value).Coordinates;
+
+        var hitFiredEvent = new HitscanTraceEvent
```
perryprog, line 40: I don't think it's worth fixing (unless you want to) but I just realized that this resets the distance traveled on each reflect, meaning you can have the distance traveled by the raycast be quite a lot more than the intended max if you get a bunch of long distance reflects.

perryprog:
That's a good point regarding `HitscanTraceEvent` and `HitscanRaycastFiredEvent`. I wonder if HitscanTraceEvent would actually be better suited as just a method (e.g., TraceHitscan or FireHitscanTrace)? That's actually more what it seems to be right now. I'm not sure if you intended for it to be something multiple things listened to—though if you needed that, I suppose you could just have it fire a BeforeX event.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
perryprog: I think I'm not sure I see where your hang up is coming from (which doesn't mean it isn't valid!). I think with your last commit it's really reasonable to me.

Like imagine if you wanted a hitscan effect that like uh... blasted the gun out of the shooter's hands or something. I dunno awful example. But you wouldn't want that to try to affect the gun if the gun was triggered by a signal or something!

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicEffectsSystem.cs
perryprog: I think it looks great! It's such a shame we can't use collection expressions here. Sandboxing is so silly.

Content.Shared/Weapons/Hitscan/Systems/HitscanBasicVisualsSystem.cs
perryprog: Way better!

perryprog (commented):
SUPER minor things:

Edit: wow that added my replies to your comments in a really weird way. Github is funky.

Content.Shared/Weapons/Hitscan/Systems/HitscanReflectSystem.cs
beck-thompson: I think this is for just making sure the raycast doesn't like go on forever :laughing:! It resetting after a reflect is probably fine. I did add a comment to the `MaxDistance` to clarify

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
beck-thompson: I trust you on this :saluting_face:

beck-thompson:
> Edit: wow that added my replies to your comments in a really weird way. Github is funky.

Yep I truly have no clue why github does that sometimes... Its very confusing :laughing:

perryprog (approved):
Hell yeah

JackTalentedCoder:
Oh god. I literally thinked about entitie'd hitscans. Waiting for approve

Avalon-Proto:
Here's a demo of what I was able to do with this on Monolith. I suggest you lot get it in soon it's great


https://github.com/user-attachments/assets/516900fb-b2de-4dd6-947e-2e96f9386f43

Content.Client/Weapons/Ranged/Systems/GunSystem.cs
```diff
@@ -266,7 +269,7 @@ public override void Shoot(EntityUid gunUid, GunComponent gun, List<(EntityUid?
                     else
                         RemoveShootable(ent.Value);
                     break;
-                case HitscanPrototype:
+                case HitscanAmmoComponent:
                     Audio.PlayPredicted(gun.SoundGunshotModified, gunUid, user);
                     Recoil(user, direction, gun.CameraRecoilScalarModified);
                     break;
```
Princess-Cheeseballs, line 276: This switch statement is doing a lot of copypasting, and since it's using only components now it can definitely be cleaned up heavily to use a generic component that holds shooting data. 

That's out of scope for this PR but I'd put a big TODO: at the top so hopefully someone does it, and make a github issue.

Content.Server/Weapons/Ranged/Systems/GunSystem.cs
```diff
@@ -141,107 +133,21 @@ public override void Shoot(EntityUid gunUid, GunComponent gun, List<(EntityUid?
...
-
-                            fromEffect = Transform(hit).Coordinates;
-                            from = TransformSystem.ToMapCoordinates(fromEffect);
-                            dir = ev.Direction;
-                            lastUser = hit;
-                        }
-                    }
+                case HitscanAmmoComponent:
+                    if (ent == null)
+                        break;
```
Princess-Cheeseballs, line 139: Same here.

Princess-Cheeseballs (changes requested):
This is more up to slam but a big benefit of making hitscans into entities is that logic can be made a lot more generic which has been done here. But I feel like it might be worth refactoring gun code further to allow for more generic logic. Some of the hitscan specific components in particular feel a bit suspect since they're doing a lot of the same things that projectile ammo is doing. 

Either this needs to do more gun code refactoring, or we need to make a github issue that guncode needs some major cleanups. Since that's in the realm of the combat workgroup I'll leave that for them to decide. If they would rather do refactors later this review can be dismissed.

beck-thompson:
> But I feel like it might be worth refactoring gun code further to allow for more generic logic. Some of the hitscan specific components in particular feel a bit suspect since they're doing a lot of the same things that projectile ammo is doing.

I agree with you it would def be worth refactoring gun code fully. The main issue is that would be a pretty large task and should also probably done with gun prediction at the same time. This PR will make the future gun code refactor easier while also allowing more unique lasers right now.

Princess-Cheeseballs:
> I agree with you it would def be worth refactoring gun code fully. The main issue is that would be a pretty large task and should also probably done with gun prediction at the same time. 

Gun prediction or bullet prediction? Cause we shouldn't wait on bullet prediction. I do think it's probably fine? Since it is a big improvement but again combat workgroup needs to call the shot on "do we address this now or in the future?" 

The amount of things in gun code that needs refactoring is piling up fast.

Content.Client/Weapons/Ranged/Systems/GunSystem.cs
beck-thompson: I'll make the github PR if / when this gets merged so there isn't a weird pr that has stuff that doesn't exist on it :laughing:

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// How far the hitscan tried to go to intersect with a target.
+    /// </summary>
+    public float DistanceTried;
+
+    /// <summary>
+    /// Set to true the hitscan is cancelled (e.g. due to reflection).
+    /// Cancelled hitscans should not apply damage or trigger follow-up effects.
+    /// </summary>
+    public bool Canceled;
+}
```
Princess-Cheeseballs, line 96: I'm not a huge fan of this event being an ordered event so that reflects andd cancellations happen first. It should probably be two events in that case. Especially since, looking at the code, this should probably be combined with the projectile reflect/cancellation events and just pass the type of projectile we're dealing with. 

I can just add this to the guncode refactor list though since I understand *why* it's done. Even if I hate it.

ToastEnjoyer: I think it should be fine until the gun refactor. It's not the _greatest_ but should be fine until then.

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -37,12 +37,12 @@ public record struct HitscanTraceEvent
     public EntityUid? Target;
 }
 
+
+
```
Princess-Cheeseballs: Whitespace moment :(

Princess-Cheeseballs (approved):
I disagree. I think forks will HATE this and it's an EVIL PR!!! Therefore I'm merging it to be EVIL!!!

ToastEnjoyer:
Oh my god dream come true

PR #38041, opened 2025-06-03
Fix reinforced plasma window blocking laser shots

### LINKED ISSUE #38039: Cannot shoot lasers through reinforced plasma windows.
## Description
Not sure if this is intentional, but reinforced plasma windows are the only type of windows in the game that you cannot shoot through with lasers. 
If this is intentional I feel like it should be made consistent, as you can still shoot through diagonal reinforced plasma windows as well as directional ones. 
If this is intentional then I also feel like reinforced uranium windows should work the same too.

**Reproduction**
Shoot laser rifle at reinforced plasma window
It just takes damage instead of going through.

**Screenshots**
N/A

**Additional context**
N/A

ScarKy0:
## About the PR

Fixes reinforced plasma window blocking laster shots
Fixes #38039

## Why / Balance

Bugfix.

## Technical details

Changed the fixture layer from WallLayer to GlassLayer. Also removed density as no other window had it for some reason.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Reinforced plasma windows no longer block laser shots.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Windows/rplasma.yml
+++ b/Resources/Prototypes/Entities/Structures/Windows/rplasma.yml
@@ -64,8 +64,7 @@
         mask:
         - FullTileMask
         layer:
-        - WallLayer
-        density: 2000
+        - GlassLayer
 
 - type: entity
   id: PlasmaReinforcedWindowDirectional


ScarKy0:
Alternative solution:
Make all plasma and uranium windows block laser shots, which would make sense.

K-Dynamic:
Works in-game
dotnet_5GwQbQTOyf

However by removing the 2000 density there's likely going to be an impact on shuttle collisions, since the value and layer was added by the PR https://github.com/space-wizards/space-station-14/pull/37422

> Alternative solution: Make all plasma and uranium windows block laser shots, which would make sense.

What about shuttle and plastitanium windows?

ScarKy0:
> However by removing the 2000 density there's likely going to be an impact on shuttle collisions, since the value and layer was added by the PR 37422

Curious... Why was plasma windows the only ones to have received the density then and other left alone? Since others have no density I don't think it will matter much

K-Dynamic:
> Curious... Why was plasma windows the only ones to have received the density then and other left alone? Since others have no density I don't think it will matter much

Probably a holdover when testing nukie shuttle collisions or something

Resources/Prototypes/Entities/Structures/Windows/rplasma.yml
```diff
@@ -64,8 +64,7 @@
         mask:
         - FullTileMask
         layer:
-        - WallLayer
```
beck-thompson, line 67: I think you can just remove the entire fixture

ScarKy0: It was specifically added in shuttle collision PR and every other window has it, im not entirely sure it can be done

beck-thompson: The uranium one doesn't have it, they should be the same so either this is wrong or the uranium one is. The parent does have this fixture, the only difference is the bounds `bounds: "-0.5,-0.5,0.5,0.5"` (The parent just doesn't have them) but when I was testing there didn't seem to be a difference.

I'm unsure which one is right

ScarKy0: I'll experiment with it a little later

ScarKy0:
Confirmed works fine without the fixture

PR #38042, opened 2025-06-03
Fix displacement map rendering in UI

### LINKED ISSUE #35363: health analyzer ui doesn't show clothing
1. scan someone with a health analyzer
2. the ui doesn't show the clothing of a person

https://github.com/user-attachments/assets/8747d100-654e-47a1-8d7b-40c4e73d8eeb

### LINKED ISSUE #37629: Any entity usiong a displacement map for sprites does not  have the sprites that are displaced show up in the examine menu
## Description
When examining an entity that uses a displacement map to correct sprites, it does not show the sprite in the examine/ med scanner/ view variable menus.

**Reproduction**
1. Pick a female moth /human and wear a jumpsuit that does not have a specific version made for the gender
2. Examine yourself

**Screenshots**
Image

Image

Image

**Additional context**

PJB3005:
918709cb4742d02e9822a651cf8e74727ba997de changed the way humans work to completely remove the stencil mask thing for clothing cutouts, in favor of displacement maps. Except it didn't actually remove the stencil parameters from the shader prototype, so this caused the shader to render incorrectly in SpriteView instances.

I introduced a new DisplacedDraw shader that doesn't have the stencil in use, as removing the old one would probably be a more annoying breaking change.

Fixes #37629
Fixes #35363

## Breaking changes
`DisplacementData` now defaults to `DisplacedDraw`, not `DisplacedStencilDraw`.

**Changelog**
:cl:
- fix: Fixed various worn clothing items being invisible when seen through UI, such as the examine popups.

### DIFF
--- a/Content.Client/Holopad/HolopadSystem.cs
+++ b/Content.Client/Holopad/HolopadSystem.cs
@@ -107,7 +107,7 @@ public sealed class HolopadSystem : SharedHolopadSystem
         // Remove shading from all layers (except displacement maps)
         for (var i = 0; i < hologramSprite.AllLayers.Count(); i++)
         {
-            if (_sprite.TryGetLayer((hologram, hologramSprite), i, out var layer, false) && layer.ShaderPrototype != "DisplacedStencilDraw")
+            if (_sprite.TryGetLayer((hologram, hologramSprite), i, out var layer, false) && layer.ShaderPrototype != "DisplacedDraw")
                 hologramSprite.LayerSetShader(i, "unshaded");
         }
 
--- a/Content.Shared/DisplacementMap/DisplacementData.cs
+++ b/Content.Shared/DisplacementMap/DisplacementData.cs
@@ -10,5 +10,5 @@ public sealed partial class DisplacementData
     public Dictionary<int, PrototypeLayerData> SizeMaps = new();
 
     [DataField]
-    public string? ShaderOverride = "DisplacedStencilDraw";
+    public string? ShaderOverride = "DisplacedDraw";
 }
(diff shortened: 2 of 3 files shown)


Happyrobot33:
This is huge for downstream forks with custom species thank you so much!

Content.Client/Holopad/HolopadSystem.cs
```diff
@@ -107,7 +107,7 @@ private void UpdateHologramSprite(EntityUid hologram, EntityUid? target)
         // Remove shading from all layers (except displacement maps)
         for (var i = 0; i < hologramSprite.AllLayers.Count(); i++)
         {
-            if (_sprite.TryGetLayer((hologram, hologramSprite), i, out var layer, false) && layer.ShaderPrototype != "DisplacedStencilDraw")
+            if (_sprite.TryGetLayer((hologram, hologramSprite), i, out var layer, false) && layer.ShaderPrototype != "DisplacedDraw")
```
Tayrtahn, line 110: This should be a static readonly const so we can validate it, no?

Tayrtahn (approved):
One suggestion, looks good otherwise.

Content.Client/Holopad/HolopadSystem.cs
PJB3005: > This should be a static readonly const so we can validate it, no?

See I thought about fixing that but then I realized this code is fragile as hell anyways so I decided to not put lipstick on the pig.

Tayrtahn: Fair enough

PJB3005:
Elevating to P0 since it breaks AO.

PR #38044, opened 2025-06-03
Increase puddle spillover volume to 50u

K-Dynamic:
## About the PR

- Increase puddle spillover volume from 20u to 50u

## Why / Balance

Assume 1u = 10 mL, 1 mL = 1 cm^3,  and 0.5 mm puddle height for 1 m x 1 m tile

Spillover volume (mL) = 0.05 cm * 100 cm * 100 cm = 1000 cm^3
Spillover volume (u) = 1000 cm^3 / 10 cm^3 = 50u

In reality, you can ignore the calculations since I assumed a puddle height which doesn't account for smearing, temperature, density etc. ([There's a whole science behind puddle heights.](https://physics.stackexchange.com/questions/144269/what-is-the-maximum-height-for-a-puddle-of-water-assuming-stp)) 

Setting the spillover volume to 50u makes spills a lot smaller. This will reduce the amount of blood covering everything in gunfights and nerfs lube foam.

## Technical details

N/A

### Gibbed Urist and welding tank spill
465828591-229028d3-c09d-4a21-bec7-93964a3f07ef

### Lube foam
465829331-bfe99d6d-9dfa-4700-b369-d21017fb9fb9

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- add: Puddles now spill over at 50u instead of 20u.

### DIFF
--- a/Content.Shared/Fluids/Components/PuddleComponent.cs
+++ b/Content.Shared/Fluids/Components/PuddleComponent.cs
@@ -15,7 +15,7 @@ namespace Content.Shared.Fluids.Components
         public SoundSpecifier SpillSound = new SoundPathSpecifier("/Audio/Effects/Fluids/splat.ogg");
 
         [DataField]
-        public FixedPoint2 OverflowVolume = FixedPoint2.New(100);
+        public FixedPoint2 OverflowVolume = FixedPoint2.New(50);
 
         [DataField("solution")] public string SolutionName = "puddle";
 


K-Dynamic:
Didn't post this earlier but current gunfights turn into massive bloodbaths
image

If we want to keep the messiness without relying on puddles, consider implementing steaks and footsteps instead.

I would also suggest adding another one or two puddle states for the new puddle volumes.

slarticodefast:
That blood puddle looks way too small for fully gibbing someone. It should at least connect, otherwise it looses its dramatic effect (and a little sillyness when someone gets smited or acidified).

slarticodefast:
And the underlying problem for the huge blood puddles from gunfights is that you can loose 100% off your bloodstream and just walk it off. You should already be totally dead when loosing like half of it, and the bleeding amount should be balanced around that. That way a gibbing puddle could stay as large as it is, but gunfights would not cause literal bloodbaths.

TaoNewt:
Hard to say, but 5 times less might be an overcorrection. I'm not saying realism is bad, but space station is definitely cartoonish too and big blood puddles are fun. What do you think about something in the 40-50u range?

K-Dynamic:
I can try 50u later to see the results

K-Dynamic:
I should also note that a good solution would be to make the puddle 'flow' (enable icon smoothing with other puddles) when it's above overflow volume

Simyon264:
While the gibbing spill visuals should probably have a bigger radius, gun fights currently just end up flooding entire sections of the station. I believe we can try these values out on vulture and tweak from there.

K-Dynamic:
I forgot about this PR. This is what 50u looks like @slarticodefast @Simyon264 
image
image

Let me know whether to go with 50 or 100u.

K-Dynamic:
Updated PR description and media to reflect new 50u spillover volume.

Approval remains active after asking Simyon over Discord.
image

K-Dynamic:
@slarticodefast is the Concern label still required? https://github.com/space-wizards/space-station-14/pull/38044#issuecomment-3067729738 shows blood gib splatter at 50u.

slarticodefast:
I still think the underlying problem is that Urists are water ballons filled with blood that loose all of it with the tiniest scratch.
But at least 50u looks better.
The PR still has 100u in the code though.

K-Dynamic:
Forgot to commit the change, enjoy 50u

PR #38045, opened 2025-06-03
Add marker spawn location functionality

SlamBamActionman:
## About the PR

This PR adds marker spawn locations; these can be used to spawn entities at set locations with a matching marker component. The supporting gamerule `MarkerLocationSpawnRule` use a string to match up the spawning with the markers.

## Why / Balance

Will be necessary for the Tourist ghost role, which are supposed to have a set spawn location in Arrivals. This is a bit of a more general solution in case there's a use for it in the future.

## Technical details

Fairly simple logic; the only thing that one needs to be careful of is not reusing existing strings to match with. Should work fine though.

## Media

Example prototypes using this in the video below:
image

https://github.com/user-attachments/assets/35d49709-1646-4add-84ac-1347b0705e7a

## Requirements

## Breaking changes

**Changelog**

No CL

### DIFF
--- a/Content.Server/StationEvents/Components/MarkerLocationSpawnRuleComponent.cs
+++ b/Content.Server/StationEvents/Components/MarkerLocationSpawnRuleComponent.cs
@@ -1,0 +1,27 @@
+﻿using Content.Server.StationEvents.Events;
+using Content.Shared.Storage;
+
+namespace Content.Server.StationEvents.Components;
+
+[RegisterComponent, Access(typeof(MarkerLocationSpawnRule))]
+public sealed partial class MarkerLocationSpawnRuleComponent : Component
+{
+    /// <summary>
+    /// Which markers will be targetted by this spawn rule, based on the given string.
+    /// </summary>
+    [DataField(required: true)]
+    public string TargetString;
+
+    /// <summary>
+    /// The entities which should be spawned at the marker entities.
+    /// </summary>
+    [DataField]
+    public List<EntitySpawnEntry> SpawnEntries = new();
+
+    /// <summary>
+    /// By default, this gamerule only targets a single random marker.
+    /// If this is set to true, all eligible spawners will be used.
+    /// </summary>
+    [DataField]
+    public bool TargetAllEligible = false;
+}
--- a/Content.Server/StationEvents/Components/MarkerLocationSpawnRuleComponent.cs
+++ b/Content.Server/StationEvents/Components/MarkerLocationSpawnRuleComponent.cs
@@ -3,11 +3,14 @@ using Content.Shared.Storage;
 
 namespace Content.Server.StationEvents.Components;
 
+/// <summary>
+/// Spawns a selection of entities at a random marker with the corresponding <see cref="MarkerSpawnLocationComponent"/>.
+/// </summary>
 [RegisterComponent, Access(typeof(MarkerLocationSpawnRule))]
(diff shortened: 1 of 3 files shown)


ketufaispikinut:
Are there any reasons why this isn't being approved?

Content.Server/StationEvents/Components/MarkerLocationSpawnRuleComponent.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// Used to match which marker entities will be eligible to be selected by this spawn rule.
+    /// </summary>
+    [DataField(required: true)]
+    public string TargetString;
+
+    /// <summary>
+    /// The entities which should be spawned at the marker entities.
+    /// </summary>
+    [DataField]
+    public List<EntitySpawnEntry> SpawnEntries = new();
```
SnappingOpossum, line 22: Ideally this should be an EntityTableSelector, but those don't currently support RobustRandoms.

Content.Server/StationEvents/Components/MarkerLocationSpawnRuleComponent.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// <summary>
+    /// By default, this gamerule only targets a single random marker.
+    /// If this is set to true, all eligible spawners will be used.
+    /// </summary>
+    [DataField]
+    public bool TargetAllEligible = false;
```
SnappingOpossum, line 29: I feel like this could instead be split into different components instead of using this, and that would allow these to be combined.

Content.Server/StationEvents/Components/MarkerLocationSpawnRuleComponent.cs
```diff
@@ -0,0 +1,30 @@
...
+/// Spawns a selection of entities at a random marker with the corresponding <see cref="MarkerSpawnLocationComponent"/>.
+/// </summary>
+[RegisterComponent, Access(typeof(MarkerLocationSpawnRule))]
+public sealed partial class MarkerLocationSpawnRuleComponent : Component
+{
+    /// <summary>
+    /// Used to match which marker entities will be eligible to be selected by this spawn rule.
+    /// </summary>
+    [DataField(required: true)]
+    public string TargetString;
```
SnappingOpossum, line 16: Might be good to make this a tag.

Content.Server/StationEvents/Components/MarkerSpawnLocationComponent.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Content.Server.StationEvents.Components;
+
+/// <summary>
+/// Makes the owner entity an eligible spawner location for <see cref="MarkerLocationSpawnRule"/>.
+/// </summary>
+[RegisterComponent, Access(typeof(MarkerLocationSpawnRule))]
+public sealed partial class MarkerSpawnLocationComponent : Component
+{
+    [DataField(required: true)]
+    public List<string> MarkerStrings = new();
```
SnappingOpossum, line 12: See above comment on tags.

SnappingOpossum (commented):
If this (or a variant) could restrict spawns to the station, VentCritterSpawnRule could also be obsoleted in favour of this.

Just on something I think would be nice to have though, could also have a mode where it gets the spawned amount once and then distributes that across the markers.

PR #38046, opened 2025-06-03
predicted hyposprays

### LINKED ISSUE #38026: HypospraySystem needs prediction
Some quality of life for med gameplay.
Should be relatively straight forward since it only depends on other shared systems.
100% soap free.
The only server side thing in there is the `BloodStreamComponent` in one function parameter, but that isn't even used anywhere, so you can just remove it.

kosticia:
## About the PR
hypospray system now predicted. this means that healing via them no longer so hard

## Why / Balance
fixes #38026

## Technical details
hypospray system in the server content has been deleted
hypospray system in the client content is now independent
sharedhypospraysystem is now just hypospray system. Its predicted. Properly. I think that properly...

## Requirements

**Changelog**
:cl:
- tweak: Hyposprays are now predicted.


kosticia:
666 pr yay

ScarKy0 (changes requested):
You're gonna have to update the popups to the correct PopupPredicted version. They have a field for message and othersmessage so if you want something to be visible only to the user just set the latter to null.

kosticia:
> You're gonna have to update the popups to the correct PopupPredicted version. They have a field for message and othersmessage so if you want something to be visible only to the user just set the latter to null.

just curious, why? All popups intended to shown only for user.

ScarKy0:
> just curious, why? All popups intended to shown only for user.

For popups for other entities. If you do PopupClient but the target is not the local entity then the popup won't be visible at all. (Example with the "You feel a tiny prick" message), unless I'm mistaking this for something else.

EDIT: Actually those might just need to be replaced with PopupEntity cause you cant predict something for another player anyways.

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -1,49 +1,40 @@
...
+    [Dependency] private readonly SharedPopupSystem _popup = default!;
+    [Dependency] private readonly SharedAudioSystem _audio = default!;
+    [Dependency] private readonly ReactiveSystem _reactiveSystem = default!;
+    [Dependency] private readonly ISharedAdminLogManager _adminLogger = default!;
+    [Dependency] private readonly UseDelaySystem _useDelay = default!;
+    [Dependency] private readonly SharedSolutionContainerSystem _solutionContainers = default!;
 
     public override void Initialize()
     {
-        base.Initialize();
```
slarticodefast: keep the base call

Content.Shared/Chemistry/Components/HyposprayComponent.cs
```diff
@@ -12,7 +11,6 @@ public sealed partial class HyposprayComponent : Component
     public string SolutionName = "hypospray";
 
     [DataField]
```
slarticodefast, line 24: add `AutoNetworkedField` to allow it to be set for admemes 😄 
and add documentation to the component and datafields

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -52,27 +43,43 @@ private void OnUseInHand(Entity<HyposprayComponent> entity, ref UseInHandEvent a
         args.Handled = TryDoInject(entity, args.User, args.User);
     }
 
-    public void OnAfterInteract(Entity<HyposprayComponent> entity, ref AfterInteractEvent args)
+    private void OnAfterInteract(Entity<HyposprayComponent> entity, ref AfterInteractEvent args)
     {
-        if (args.Handled || !args.CanReach || args.Target == null)
+        if (args.Handled || !args.CanReach || args.Target is null)
```
slarticodefast:
```suggestion
        if (args.Handled || !args.CanReach || args.Target == null)
```
just keep it as it was

kosticia:
> EDIT: Actually those might just need to be replaced with PopupEntity cause you cant predict something for another player anyways.

No no no, PopupEntity sucks at shared. It causes mispredictions too hard, like a "no battery x3" at the health analyser at every click

ScarKy0:
> No no no, PopupEntity sucks at shared. It causes mispredictions too hard, like a "no battery x3" at the health analyser at every click

Only if the popup is for your local client, if you do it for another player then it SHOULD work just fine.
I'll test this in-game later to be sure

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -104,57 +111,57 @@ public bool TryDoInject(Entity<HyposprayComponent> entity, EntityUid target, Ent
...
 
         target = targetEvent.TargetGettingInjected;
 
-        if (!EligibleEntity(target, EntityManager, component))
+        if (!EligibleEntity(target, component))
             return false;
 
         // The target event gets priority for the overriden message.
-        if (targetEvent.InjectMessageOverride != null)
+        if (targetEvent.InjectMessageOverride is not null)
```
slarticodefast:
```suggestion
        if (targetEvent.InjectMessageOverride != null)
```
might just be my personal preference
but in general it is better not to do additional changes in a PR unless needed to minimize conflicts

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -104,57 +111,57 @@ public bool TryDoInject(Entity<HyposprayComponent> entity, EntityUid target, Ent
...
             msgFormat = targetEvent.InjectMessageOverride;
-        else if (selfEvent.InjectMessageOverride != null)
+        else if (selfEvent.InjectMessageOverride is not null)
```
slarticodefast:
```suggestion
        else if (selfEvent.InjectMessageOverride != null)
```
same

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -104,57 +111,57 @@ public bool TryDoInject(Entity<HyposprayComponent> entity, EntityUid target, Ent
...
             // meleeSys.SendLunge(angle, user);
         }
 
-        _audio.PlayPvs(component.InjectSound, user);
+        _audio.PlayPredicted(component.InjectSound, target, user);
 
         // Medipens and such use this system and don't have a delay, requiring extra checks
         // BeginDelay function returns if item is already on delay
-        if (delayComp != null)
+        if (delayComp is not null)
```
slarticodefast:
```suggestion
        if (delayComp != null)
```

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -104,57 +111,57 @@ public bool TryDoInject(Entity<HyposprayComponent> entity, EntityUid target, Ent
...
             return false;
         }
 
-        _popup.PopupEntity(Loc.GetString(msgFormat ?? "hypospray-component-inject-other-message", ("other", target)), target, user);
+        _popup.PopupClient(Loc.GetString(msgFormat ?? "hypospray-component-inject-other-message", ("other", target)), target, user);
 
         if (target != user)
         {
-            _popup.PopupEntity(Loc.GetString("hypospray-component-feel-prick-message"), target, target);
+            _popup.PopupClient(Loc.GetString("hypospray-component-feel-prick-message"), target, target);
```
slarticodefast: have you tested this with two clients?
This will likely not show up for the target

beck-thompson:
Make sure it works with clumsy! I forget if the randomness is predicted properly

slarticodefast:
Hmm, looks like it already is in shared, but the random chance will mispredict because it rolls separately on the server and client.
grafik
Should be an easy fix using https://github.com/space-wizards/RobustToolbox/pull/5849 though

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
kosticia: works with two clients now

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -1,26 +1,32 @@
...
+using Robust.Shared.Timing;
 
-namespace Content.Server.Chemistry.EntitySystems;
+namespace Content.Shared.Chemistry.EntitySystems;
 
-public sealed class HypospraySystem : SharedHypospraySystem
+public sealed class HypospraySystem : EntitySystem
 {
-    [Dependency] private readonly AudioSystem _audio = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
```
ScarKy0:
```suggestion

```
Unused

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -1,26 +1,32 @@
...
 using Content.Shared.Mobs.Components;
+using Content.Shared.Popups;
 using Content.Shared.Timing;
+using Content.Shared.Verbs;
 using Content.Shared.Weapons.Melee.Events;
-using Content.Server.Body.Components;
-using System.Linq;
-using Robust.Server.Audio;
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Timing;
```
ScarKy0:
```suggestion

```
Unused

ScarKy0 (approved):

kosticia:
oops, looks like you already fixed this🫠...

Content.Shared/Chemistry/Components/HyposprayComponent.cs
```diff
@@ -1,20 +1,32 @@
 using Content.Shared.FixedPoint;
 using Robust.Shared.GameStates;
-using Robust.Shared.Serialization;
 using Robust.Shared.Audio;
 
 namespace Content.Shared.Chemistry.Components;
 
-[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+/// <summary>
+///     Component that allows entity instantly transfer liquids by interacting with objects that have solutions.
```
slarticodefast:
```suggestion
///     Component that allows an entity instantly transfer liquids by interacting with objects that have solutions.
```

Content.Shared/Chemistry/Components/HyposprayComponent.cs
```diff
@@ -1,20 +1,32 @@
...
 public sealed partial class HyposprayComponent : Component
 {
+    /// <summary>
+    ///     Solution that will be used by hypospray for injections.
+    /// </summary>
     [DataField]
     public string SolutionName = "hypospray";
 
+    /// <summary>
+    ///     Amount of the units that will be transfered when user interact with target.
```
slarticodefast:
```suggestion
    ///     Amount of the units that will be transfered.
```

Content.Shared/Chemistry/Components/HyposprayComponent.cs
```diff
@@ -1,20 +1,32 @@
...
+    /// </summary>
+    [AutoNetworkedField]
     [DataField]
-    [ViewVariables(VVAccess.ReadWrite)]
     public FixedPoint2 TransferAmount = FixedPoint2.New(5);
 
+    /// <summary>
+    ///     Sound that will be played when user interact with target.
```
slarticodefast:
```suggestion
    ///     Sound that will be played when injecting.
```

Content.Shared/Chemistry/EntitySystems/HypospraySystem.cs
```diff
@@ -189,34 +198,78 @@ private bool TryDraw(Entity<HyposprayComponent> entity, Entity<BloodstreamCompon
...
+    #region Verbs
+
+    // <summary>
+    // Uses the OnlyMobs field as a check to implement the ability
+    // to draw from jugs and containers with the hypospray
+    // Toggleable to allow people to inject containers if they prefer it over drawing
+    // </summary>
+    private void AddToggleModeVerb(Entity<HyposprayComponent> entity, ref GetVerbsEvent<AlternativeVerb> args)
+    {
+        if (!args.CanInteract || entity.Comp.InjectOnly)
```
slarticodefast:
```suggestion
        if (!args.CanAccess || !args.CanInteract || args.Hands == null || entity.Comp.InjectOnly)
```
Why did you remove these two checks?

slarticodefast (approved):
I gave it an in-game test and it feels so much smoother.
How could we ever live without this?

slarticodefast:
Regarding the clumsy system, I think it's ok for now, but I'll open an issue. Ideally this will need RandomPredicted.

beck-thompson:
> Regarding the clumsy system, I think it's ok for now, but I'll open an issue. Ideally this will need RandomPredicted.

100% worth! I just tested this out and wow your right its so much better :sob:

Issue #25933, opened 2024-03-08
Thindows block electrocution on grilles and don't block dmg to grilles


IamVelcroboy:
## Description
- If a thindow is placed over a grille with a cable underneath it, it stops players that bump it from being electrocuted.
- If you attack a grille through a thindow you can break the grille

**Reproduction**
Place cable, grill, thindow -> walk into it.
Pick up weapon, attack grille from other side of thindow -> break it.

**Screenshots**
No electrocute:
https://github.com/space-wizards/space-station-14/assets/107660393/fb39ce8a-00cc-4dd6-b3bd-e64c72263dba

Break Grille:
https://github.com/space-wizards/space-station-14/assets/107660393/4aba2de5-1877-49c9-b0c8-e1a9a2fdd51d

PR #38055, opened 2025-06-04
Fix electrified grilles under directional windows

### LINKED ISSUE #27660: Directional windows stop shock grille electricute interactions
## Description
Reclassification of #27633

**Reproduction**
Have grille with directional window on same tile
Hit shocked grille
Receive no shock damage

**Screenshots**
327570713-f7cc9831-1832-4ece-a369-93bdecab782b

**Additional context**
Reproduction tested by [Lamrr](https://github.com/Lamrr)

Quantum-cross:
## About the PR

Fixes #27660. Directional windows over electrified grilles no longer entirely disable them, and you cannot hit a grille through a directional window.

## Why / Balance

Bug fix

## Technical details

I removed the `noWindowInTile` `DataField`. This was only used on grilles and fences, and in electrocution system it would add an additional check for entities with tag `Window` and disable the power to the electrified entity.

Instead, when the grille is bumped, also do a raycast to check that the ray between the player and the center of the grille is unobstructed. This allows directional windows to block this interaction.

## Media

I'm on my laptop so I can't really take a video but:

Bump against directional window, NO SHOCK

Bump against grille side, SHOCK

Bump against plain grill, SHOCK of course

Bump against grille under window, NO SHOCK

Cannot hit grille with baseball bat on directional window side.

## Requirements

## Breaking changes

`DataField` `NoWindowInTile` on `ElectrifiedComponent` has been removed.

**Changelog**

:cl:
- fix: Directional windows no longer cause electrified grilles to stop working

### DIFF
--- a/Content.Server/Electrocution/ElectrocutionSystem.cs
+++ b/Content.Server/Electrocution/ElectrocutionSystem.cs
@@ -57,6 +57,7 @@ public sealed class ElectrocutionSystem : SharedElectrocutionSystem
     [Dependency] private readonly SharedStutteringSystem _stuttering = default!;
     [Dependency] private readonly TagSystem _tag = default!;
     [Dependency] private readonly MetaDataSystem _metaData = default!;
+    [Dependency] private readonly SharedInteractionSystem _interaction = default!;
 
     [ValidatePrototypeId<StatusEffectPrototype>]
     private const string StatusEffectKey = "Electrocution";
@@ -135,19 +136,6 @@ public sealed class ElectrocutionSystem : SharedElectrocutionSystem
     {
         if (!electrified.Enabled)
             return false;
-        if (electrified.NoWindowInTile)
-        {
-            var tileRef = transform.Coordinates.GetTileRef(EntityManager, _mapManager);
-
-            if (tileRef != null)
-            {
-                foreach (var entity in _entityLookup.GetLocalEntitiesIntersecting(tileRef.Value, flags: LookupFlags.StaticSundries))
-                {
-                    if (_tag.HasTag(entity, WindowTag))
-                        return false;
-                }
-            }
-        }
         if (electrified.UsesApcPower)
         {
             if (!this.IsPowered(uid, EntityManager))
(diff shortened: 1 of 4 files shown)


Quantum-cross:
also fixes #25933

Content.Server/Electrocution/ElectrocutionSystem.cs
```diff
@@ -161,7 +149,7 @@ private bool IsPowered(EntityUid uid, ElectrifiedComponent electrified, Transfor
 
     private void OnElectrifiedStartCollide(EntityUid uid, ElectrifiedComponent electrified, ref StartCollideEvent args)
     {
-        if (electrified.OnBump)
+        if (electrified.OnBump && _interaction.InRangeUnobstructed(args.OurEntity, args.OtherEntity, overlapCheck: false))
```
slarticodefast, line 152: Isn't doing this for every single collision extremely expensive?

Quantum-cross: The if statement will short circuit if it doesn't have OnBump

Quantum-cross: I can nest the if statements if you want that to be more clear

slarticodefast: I know that part.
But having InRangeUnObstructed do a raytrace check on each collision is still expensive even if it's only the OnBump cases.
There is a lot of entities that have it set to true, and any collision they do would call InRangeUnObstructed .
It's not even checking if the target is powered anymore like it was before.

Quantum-cross: Well that happens in the tryelecrocute. 
Before it was turning off the power if any window tag was on it. I'll see if I can optimize it when I get home tonight. 

It could do some more checks to bail early before the ray cast.

IamVelcroboy:
Hopefully this is still being worked on 👦

Quantum-cross:
Out of town right now and it dipped low of my priority list. I gotta come back to this

ToastEnjoyer:
Already fixed by this PR https://github.com/space-wizards/space-station-14/pull/39032

slarticodefast:
The other PR was for diagonal windows, not for thindows.

ToastEnjoyer:
That was completely my fault, apologies.

AJCM-git:
@Quantum-cross coming back to this?

PR #38056, opened 2025-06-04
Slightly rework pathfinding obstacle avoidance and handling

### LINKED ISSUE #17764: An NPC mob wielding a melee weapon attacking some things crashes the server
**Reproduction**

A.
1. Spawn Urist
2. Spawn Makeshift shield
3. Take control of Urist
4. Pick up the shield and raise it via the action button
5. Ghost out of Urist
6. Spawn MobXenoRouny out of aggro range/vision of Urist
7. Take control of Rouny
8. Open console and "addhand [Rouny's entityUid]"
9. Spawn a Crowbar
10. Pick up the Crowbar as Rouny
11. Ghost out of Rouny
12. Teleport Rouny into aggro range/vision of Urist
13. Having Rouny attack the Urist crashes the server with an exception [FATL] unhandled: System.AggregateException: One or more errors occurred. (Specified Entity is not the component's Owner!)

B. (found a second case while testing)
1. Spawn MobXenoRouny
2. Take control of Rouny
3. Open console and "addhand [Rouny's entityUid]"
4. Spawn a Crowbar
5. Pick up the Crowbar as Rouny
6. Ghost out of Rouny
7. Have Rouny randomly attack a wall due to HTN
8. Again, the server crashes with an exception [FATL] unhandled: System.AggregateException: One or more errors occurred. (Specified Entity is not the component's Owner!)

### LINKED ISSUE #36490: HTN NavSmash struggles in open spaces
## Description

NavSmash makes it so that mobs can break destructible entities in their navigation, notable used by carp. However this doesn't work well when the mob is allowed two directions of movement. In the videos included below, 1xX carp always attack a wall/window at the end of their navigation. The 2x2 carp _occasionally_ attacks, and the 3x3 carp _never_ attacks.

**Reproduction**

Lock some carp in a room.

**Screenshots**

https://github.com/user-attachments/assets/0403a4aa-8f5e-4418-ba16-791c7923cbe4

https://github.com/user-attachments/assets/ccb59b4a-81e5-4772-88d0-399055a9f9c5

perryprog:
## About the PR

I semi-refactored how pathfinding obstacle avoidance works, specifically relating to how it handles doors (prying/bumps), vaulting, and brute force, and also adjusted some pathfinding weights to give slightly less stupid pathfinding choices when, for example, walled off. This also changes how multiple mobs will pathfind in situations that were previously "safe". As a notable example, slimes will now bum-rush doors if they see you, opening them if the doors were all access, and mobs that can interact will now check if a door is openable by them for their access level. Xenos especially are much, much more effective at chasing you down.

Also drive-by fix a thing for the route cost overlay since it wasn't working.

## Why / Balance

For fun I was making a creepy plushie that could follow you and pry open doors and then noticed it wasn't prying open doors. Then I did this. Whoops.

Also realistic space aliens or something.

Fixes #36490, fixes #17764.

## Technical details

There is now some degree of code shared between the danger weighting and the actual handling of these obstacles. I feel like... it could be better? I'm especially not happy with the duplication in logic between TryHandleFlags and FilterObstacleEntities, but I'd like a second opinion. Happy to rework things as needed, though it seems from the comments that a greater refactor for pathfinding is in the works for at least some point. Some other notes that I would like advice on:

- Not super related to the PR but I noticed `forceSteer` in `Steer` seems to be true in every code path. Not sure where it was originally meant to be false, or if it's still needed.
- Pathfinding costs for doors versus tables doesn't seem quite right still. Mobs will pathfind to prying the door over a table vault but I can't figure out why. The counter should only have a cost of 0.5 versus a door with AccessReader having +10...
- I'm not super familiar with previous regressions of issues with mobs deciding not to pry open doors, and it's super possible I reintroduced some. 

## Media

Potato video because github file size limit.
https://github.com/user-attachments/assets/daeb16f0-e3a9-4a42-92f1-13059263700a

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Mob pathfinding is greatly improved when dealing with various obstacles. Happy slime hunting!


perryprog:
Also—not sure if it's out of scope of this PR but I can add it if it's needed—as far as I can tell we don't have a way to tell if we can damage something without actually uh, damaging it. This means the pathfinder will see a wall that has 200 health as something that can be beaten past with fists, even though the damage reduction will of course nullify any damage. This means that if there's a long enough chain of plain glass windows versus some standard walls, the AI will try to fruitlessly punch at the wall instead of going through the breakable windows.

iaada:
Looks like it resolves #36490

UpAndLeaves:
Does this PR patch NPCs pathfinding over tiny fans into space and dying?

perryprog:
> Does this PR patch NPCs pathfinding over tiny fans into space and dying?

Not if the fan leads out onto e.g., a catwalk (like in the dev map); there's no differentiation in the breadcrumbs between a tile (that isn't a "space" tile) that has atmos versus one that doesn't, from what I can tell. For pathfinding ending up entering actual space, that tends to happen due to (I think) the steering overshooting the target tile.

Content.Server/NPC/Systems/NPCSteeringSystem.Context.cs
```diff
@@ -207,19 +209,9 @@ private bool TrySeek(
             {
                 // Ignore stuck while handling obstacles.
                 ResetStuck(steering, ourCoordinates);
-                SteeringObstacleStatus status;
 
                 // Breaking behaviours and the likes.
-                lock (_obstacles)
```
perryprog, line 214: Thought this lock wasn't needed since for some reason I thought this ran on main loop, but it gets parallelized in NPCSteeringSystem's update... but with a MaxDegreeOfParallelism of 1, which is probably why I didn't notice any issues. That being said, I'm also not sure what the lock is meant to be protecting, so would like to know if/why this needs to come back.

Content.Server/NPC/Systems/NPCSteeringSystem.cs
```diff
@@ -97,6 +104,9 @@ public override void Initialize()
         _factionQuery = GetEntityQuery<NpcFactionMemberComponent>();
         _physicsQuery = GetEntityQuery<PhysicsComponent>();
         _xformQuery = GetEntityQuery<TransformComponent>();
+        _doorQuery = GetEntityQuery<DoorComponent>();
+        _climbableQuery = GetEntityQuery<ClimbableComponent>();
+        _destructibleQuery = GetEntityQuery<DestructibleComponent>();
```
perryprog: I ended up putting these queries here, even though only the destructible query is used in more than one place. I'm still not entirely sure how cached entity queries work and where you /should/ initialize them, if that even matters. (Sometimes I see them initialized in an update loop, other times in a system's initialize.)

Content.Server/NPC/Systems/NPCSteeringSystem.Obstacles.cs
```diff
@@ -57,168 +53,127 @@ private SteeringObstacleStatus TryHandleFlags(EntityUid uid, NPCSteeringComponen
...
 
-                        if (door.State != DoorState.Opening)
-                            _pryingSystem.TryPry(ent, uid, out id, uid);
-
-                        component.DoAfterId = id;
-                        return SteeringObstacleStatus.Continuing;
-                    }
-                }
+                // Interacts are bit nicer than bumps, so try interacting regardless
+                _interaction.InteractionActivate(uid, ent);
```
perryprog, line 101: This still only handles bumps by kinda just... treating doors that are bumpable and accessible as a non-obstacle and hoping that the weights and speed of the NPC work out such that it bumps it. This is super obvious with slimes which will still do a bit of the back and forth wiggle when they see someone behind a door, but in testing they seem to consistently open it after a few tries.

Content.Server/NPC/Systems/NPCSteeringSystem.Obstacles.cs
```diff
@@ -57,168 +53,127 @@ private SteeringObstacleStatus TryHandleFlags(EntityUid uid, NPCSteeringComponen
...
-            var isClimbable = (poly.Data.Flags & PathfindingBreadcrumbFlag.Climb) != 0x0;
+        if (!TryComp<MapGridComponent>(poly.GraphUid, out var grid))
+            return SteeringObstacleStatus.Completed;
 
-            // Just walk into it stupid
-            if (isDoor && !isAccessRequired)
-            {
-                var doorQuery = GetEntityQuery<DoorComponent>();
+        var obstacleEnts = _mapSystem.GetLocalAnchoredEntities(poly.GraphUid, grid, poly.Box).ToHashSet();
+        FilterObstacleEntities((uid, component), mask, layer, obstacleEnts);
```
perryprog: I suspect this isn't the most performant way of going about this (though in profiling I haven't noticed any issues, though I just realized I dunno about memory usage), and would like to know what would be better if necessary.

CrigCrag:
In my testing this seems to cause turrets to continue shooting people who have rounded corners, wasting their ammo.

perryprog:
@CrigCrag, I'm having trouble reproducing this. Can you show what your setup looks like?

Content.Server/NPC/Systems/NPCSteeringSystem.Obstacles.cs
```diff
@@ -57,168 +53,127 @@ private SteeringObstacleStatus TryHandleFlags(EntityUid uid, NPCSteeringComponen
...
-                        // TODO: Validate we can damage it
-                        if (destructibleQuery.HasComponent(ent))
-                        {
-                            attackResult = _melee.AttemptLightAttack(uid, uid, meleeWeapon, ent);
-                            break;
-                        }
-                    }
+                // Should be able to pry from CanHandleDoor:
+                // TODO: Use the verb.
+                _pryingSystem.TryPry(ent, uid, out id, uid);
```
Pok27, line 117: Could we check that a `DoAfterId` was actually created here? `TryPry` can return `true` even when the pry is rejected, while leaving `id` as `null`.

Content.Server/NPC/Systems/NPCSteeringSystem.Context.cs
```diff
@@ -563,6 +546,108 @@ private void CollisionAvoidance(
...
+        {
+            // Get rid of stuff we can phase through
+            if (!_physicsQuery.TryGetComponent(nearbyEnt, out var otherBody) ||
+                !otherBody.Hard ||
+                !otherBody.CanCollide ||
+                otherBody.BodyType == BodyType.KinematicController ||
+                (mask & otherBody.CollisionLayer) == 0x0 &&
+                (layer & otherBody.CollisionMask) == 0x0)
+            {
+                nearbyEntities.Remove(nearbyEnt);
```
Pok27: Modified `HashSet` inside a loop may cause an `InvalidOperationException` to be thrown. Could collect entities to remove in a separate list and remove them after the loop, or iterate over a copy

Pok27 (changes requested):
Will you come back to this? Need to resolve some merge conflicts

Content.Server/NPC/Systems/NPCSteeringSystem.Obstacles.cs
kontakt: Is that not a bug? Why would `TryPry` return true if it doesn't pry?

Pok27: idk, but in any case, need to check either `TryPry` or `id`

perryprog:
> Will you come back to this? Need to resolve some merge conflicts

Sadly probably not, but hopefully someone brave can salvage it.

Pok27 (approved):
I've made some changes. We need to test this thoroughly, it's a huge improvement

perryprog:
Yeah, I was super excited when I saw how big the differences were when I first tested it. I remember it being a genuinely serious buff to enemies; very excited to see that you were able to pick it up!

PR #38057, opened 2025-06-04
Added functionality to DumpableComponent, made Backpacks/Satchels/Duffels and Toolboxes dumpable.

RedBookcase:
## About the PR
Added new `DumpsTables` and `DumpsBins` components to `DumpableComponent`. These can be used to control whether the attached item automatically dumps its contents if it's used on an interactable surface or bin. By default DumpsTables is set to True and DumpsBins is set to false. You can still dump the container into a bin or onto a table, but if it's set to false you have to do it through the verb in the right-click menu, not by default.

Updated various Dumpable items to account for these new components. Trashbags & Ashtrays now don't dump onto tables but do dump into bins - no more accidentally emptying out your trash bag on the bartop. Removed dumpable component from the Snap Pops box because I have no idea why it was there in the first place.

Added type: Dumpable to BaseToolbox and ClothingBackpack, which means adding dumpable to all Toolboxes & Backpacks/Satchels/Duffels. All can now have their contents dumped out on the floor.

## Why / Balance
Having the ability to dump out your bag or toolbox makes sense intuitively and makes emptying your backpack/toolbox easier. It was already possible to effectively dump out your entire backpack by just spamming quickgrab + drop already so I don't think having the option to do it as a doafter will really be a problem. Just makes for an easier and less Carpal Tunnel inducing way to clear out your bag/toolbox.

## Technical details
changes to DumpableComponent and DumpableSystem. Some other yaml tweaks to update dumpable containers to reflect the addition of DumpsTables and DumpsBins

## Media

https://github.com/user-attachments/assets/017e2a49-3355-4433-b10a-ba781cffff25

## Requirements

**Changelog**

:cl:
- tweak: Added functionality to DumpableComponent. Backpacks/Satchels/Duffels and Toolboxes can now be dumped out on the ground.


RedBookcase:
Really hoping this is one of those "Oh yea, why can't you do that?" PRs and not one of those "This will require years of debate." PRs

YoungThugSS14:
I think toolboxes are fine, not sure about the rest.

RedBookcase:
> like we talked about in VC, make it so when you click on a table it doesn't get dumped out and it only dumps out when you click the verb button

i am betrayed

beck-thompson:
> i am betrayed

It shouldn't be too hard! Just add a bool to the component and check it

Chartman95:
If you're going to make these inventory having entities be dumpable, shouldn't all of them be likewise dumpable? Things like bottle boxes come to mind, when you have to take out each item individually to them place inside another. Very tedious and requires inventory swapping. Toolbelts, coats, anything like that could use it.

RedBookcase:
> If you're going to make these inventory having entities be dumpable, shouldn't all of them be likewise dumpable? Things like bottle boxes come to mind, when you have to take out each item individually to them place inside another. Very tedious and requires inventory swapping. Toolbelts, coats, anything like that could use it.

I'm just doing toolboxes and backpacks since they're very common and also rational for dumping out (Emptying a backpack is as easy as unzipping it and turning it upside down, emptying a toolbelt or lab coat less so, etc.)

Content.Shared/Storage/Components/DumpableComponent.cs
```diff
@@ -21,6 +21,18 @@ public sealed partial class DumpableComponent : Component
...
     public SoundSpecifier? DumpSound = new SoundCollectionSpecifier("storageRustle");
 
+    /// <summary>
+    /// Does this default to being dumped out onto a table or other placeable surface (default true - set it to false if it's for something like a backpack/toolbox)?
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool DumpsTables = true;
+
+    /// <summary>
+    /// Does this default to being dumped out into a trash bin (default true - set it to false if it's for something like a backpack/toolbox)?
```
beck-thompson: comment is wrong it defaults to false

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
```diff
@@ -39,10 +39,10 @@ public override void Initialize()
 
     private void OnAfterInteract(EntityUid uid, DumpableComponent component, AfterInteractEvent args)
     {
-        if (!args.CanReach || args.Handled)
+        if (args.Target == null)
```
beck-thompson: I think it would be nicer to combine the first two if statements

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
```diff
@@ -51,6 +51,7 @@ private void OnAfterInteract(EntityUid uid, DumpableComponent component, AfterIn
         if (!storage.Container.ContainedEntities.Any())
             return;
 
+        if (component.DumpsBins == true && HasComp<DisposalUnitComponent>(args.Target) || component.DumpsTables == true && HasComp<PlaceableSurfaceComponent>(args.Target))
```
beck-thompson: remember for booleans doing `component.DumpsBins == true` is the same as just doing `component.DumpsBins`! Same with `component.DumpsTables == true`.

It should also be a guard statement (If both of those are false then return) like:

```
if (!(component.DumpsBins == true && HasComp<DisposalUnitComponent>(args.Target) || component.DumpsTables == true && HasComp<PlaceableSurfaceComponent>(args.Target)))
    return;
```

Resources/Prototypes/Entities/Objects/Fun/snap_pops.yml
```diff
@@ -54,5 +54,3 @@
     contents:
       - id: SnapPop
         amount: 5
-  - type: Dumpable
```
beck-thompson, line 57: sus

beck-thompson (changes requested):

Resources/Prototypes/Entities/Objects/Fun/snap_pops.yml
RedBookcase: I think whoever added snap pops thought dumpable was necessary to throw them away or something. Unless being able to dump them all out adds some functionality or gimmick

beck-thompson: I truly could not think of a reason either

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
RedBookcase: Alright, all suggested changes are pushed

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
```diff
@@ -51,6 +48,10 @@ private void OnAfterInteract(EntityUid uid, DumpableComponent component, AfterIn
         if (!storage.Container.ContainedEntities.Any())
             return;
 
+        if (!(component.DumpsBins && HasComp<DisposalUnitComponent>(args.Target) || component.DumpsTables  && HasComp<PlaceableSurfaceComponent>(args.Target)))
+            return;
+
+        if (component.DumpsBins && HasComp<DisposalUnitComponent>(args.Target) || component.DumpsTables && HasComp<PlaceableSurfaceComponent>(args.Target))
```
beck-thompson: You forgot to remove this

beck-thompson (changes requested):

Content.Shared/Storage/EntitySystems/DumpableSystem.cs
RedBookcase: ah i thought you just wanted the guard statement added in on top of the regular one, i'll remove the old one

beck-thompson (approved):
looks good!

RedBookcase:
Looks like the SmartFridge update changed a lot of stuff regarding DumpableSystem and I am not up for diving into the .cs again so gonna close this one. Anyone who wants to is free to retry this.

PR #38059, opened 2025-06-04
Tweak for the visual coverage granted to station AIs by holopads

chromiumboy:
## About the PR

Holopads now provide less visual coverage to station AIs (1 tile radius, down from 7.5). Holopads must also be anchored to provide vision.

## Why / Balance

The station AI needs to be able to see a holopad to interact with it, so holopads grant the AI vision just like cameras do. However, the amount of vision coverage that they grant is excessive for this need. Reducing the vision range is also helpful for antags trying to avoid detection.

Also, since holopads aren't functional when not anchored, they don't need to provide vision when they are unanchored. This also prevents unanchored holopads from acting as mobile cameras.

## Technical details

N/A

## Media

The new visual coverage granted to the station AI by holopads
aivision1

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: Holopads now provide less visual coverage to station AIs (1 tile radius, down from 7.5). Holopads must also be anchored to the floor to provide vision.

### DIFF
--- a/Content.Shared/Silicons/StationAi/StationAiVisionComponent.cs
+++ b/Content.Shared/Silicons/StationAi/StationAiVisionComponent.cs
@@ -6,14 +6,26 @@ namespace Content.Shared.StationAi;
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(SharedStationAiSystem))]
 public sealed partial class StationAiVisionComponent : Component
 {
+    /// <summary>
+    /// Determines whether the entity is actively providing vision to the station AI.
+    /// </summary>
     [DataField, AutoNetworkedField]
     public bool Enabled = true;
 
+    /// <summary>
+    /// Determines whether the entity's vision is blocked by walls.
+    /// </summary>
     [DataField, AutoNetworkedField]
     public bool Occluded = true;
 
     /// <summary>
-    /// Range in tiles
+    /// Determines whether the entity needs to be receiving power to provide vision to the station AI.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool NeedsPower = false;
+
+    /// <summary>
+    /// Vision range in tiles.
     /// </summary>
     [DataField, AutoNetworkedField]
     public float Range = 7.5f;
--- a/Content.Shared/Silicons/StationAi/StationAiVisionComponent.cs
+++ b/Content.Shared/Silicons/StationAi/StationAiVisionComponent.cs
@@ -3,6 +3,9 @@ using Robust.Shared.GameStates;
 
 namespace Content.Shared.StationAi;
 
+/// <summary>
+/// Attached to entities that grant vision to the station AI, such as cameras.
+/// </summary>
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(SharedStationAiSystem))]
 public sealed partial class StationAiVisionComponent : Component
 {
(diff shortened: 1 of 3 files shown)


Content.Shared/Silicons/StationAi/StationAiVisionComponent.cs
```diff
@@ -6,14 +6,26 @@ namespace Content.Shared.StationAi;
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState, Access(typeof(SharedStationAiSystem))]
 public sealed partial class StationAiVisionComponent : Component
```
slarticodefast, line 10: could you add documentation to the component itself as well?

ScarKy0:
Is the decrease neccessary though?
You can already cut the AIV wire and stuff, and now has a power requirement too

spanky-spanky:
I don’t think most players are aware that holopads have vision. I’ll watch prisoners destroy cameras but leave the holopad intact. As such, I support this.

slarticodefast:
Yeah, there is not a lot of player feedback for what the AI can and cannot see. Cameras are kinda intuitive and have a red light. But holopads are mapped in pretty much every single room and have a large range at the moment.

walksanatora:
I think the vision reduction is enough. no need to remove vision on power lost. you can still see through unpowered cameras even though you probally aren't able to interact with anything when the cameras are unpowered.

Mixelz:
> Yeah, there is not a lot of player feedback for what the AI can and cannot see. Cameras are kinda intuitive and have a red light. But holopads are mapped in pretty much every single room and have a large range at the moment.

Adding onto this, we should probably check maps prior to merging as we could run into some AI blindness issues for any rooms that have a holo pad but no camera (See clown room on Plasma as an example).

chromiumboy:
> I think the vision reduction is enough. no need to remove vision on power lost. you can still see through unpowered cameras even though you probally aren't able to interact with anything when the cameras are unpowered.

It's a good point that removing vision on power loss is inconsistent with the current behavior of cameras. I could add to check to see if the entity is anchored and use that instead

northerndakotas:
I understand the idea behind turning off vision on power loss as that is intuitive behavior, however I think the vision reduction is misguided. I think this primarily because the AI does have that level of vision while using the holopad. For the AI's vision to suddenly get several times wider just because they are communicating through the holopad, and then suddenly smaller again as soon as the call ends would be very annoying to play around and not very intuitive. Moreover it would likely lead to a situation where the ideal way to play AI requires you to constantly and repeatedly appear at holopads for a few seconds at a time just to be able to see what you should already be able to see. 

That said, unanchored holopads shouldn't have vision, I agree. But if an antag doesn't realize that the AI's main communication device can see them while it's powered and connected, that's a skill issue, and a part of playing and learning.

chromiumboy:
> I understand the idea behind turning off vision on power loss as that is intuitive behavior, however I think the vision reduction is misguided. I think this primarily because the AI does have that level of vision while using the holopad. For the AI's vision to suddenly get several times wider just because they are communicating through the holopad, and then suddenly smaller again as soon as the call ends would be very annoying to play around and not very intuitive. Moreover it would likely lead to a situation where the ideal way to play AI requires you to constantly and repeatedly appear at holopads for a few seconds at a time just to be able to see what you should already be able to see.
> 
> That said, unanchored holopads shouldn't have vision, I agree. But if an antag doesn't realize that the AI's main communication device can see them while it's powered and connected, that's a skill issue, and a part of playing and learning.

The holopads are communication devices, they aren't intended to be a passive spying device - that's what cameras are for. If all the cameras have been removed in an area, and the AI wants to see why, they should be active about it

chromiumboy:
I've made the requested changes. I won't merge it just yet though, since some concerns have been raised

southbridge-fur:
I think this is a good PR. When it comes to AI vision issues in maps, most maps already have cameras covering the areas where holopads provide vision, and this simply helps make things more intuitive. I was considering opening a PR like this myself so I'm glad to see it.

Issue #37369, opened 2025-05-12
ItemToggleActiveSoundComponent has some client-side issues


Krunklehorn:
TL;DR: Client side, active sounds continue playing when they shouldn't and fail to restart when they should.

Test 1:
1) Open two clients, turn off the volume of client A
2) Stand both characters next to each other
3) On client A, spawn and activate an esword
4) On client B, leave the area then come back
5) Esword no longer humming on client B
6) Esword continues to hum on client A

Test 2:
1) Open two clients, turn off the volume of client A
2) Stand both characters next to each other
3) On client A, spawn and activate a double-bladed esword
4) On client A, deactivate the double-bladed esword
5) DESword no longer humming on client B
6) DEsword continues to hum on client A, even if they drop it, leave and come back

Test 3:
1) Open two clients
2) On client A, spawn and control a Syndicate Assault Cyborg
3) Stand both characters next to each other
4) On client A, switch modules and activate the double-bladed esword
5) On client A, deactivate the double-bladed esword by using it in hand or swapping hands
6) As expected, neither client can hear the humming anymore
7) On client A, swap to and activate the double-bladed esword
8) On client A, switch modules
9) DESword plays the deactivate sound on both clients.
5) DEsword no longer humming on client B
6) DEsword continues to hum on client A

beck-thompson:
Spent a while with @RedBookcase trying to fix this but its very confusing. The first two issues basically a missprediction because of virtual items (When the virtual item gets deleted it tries to unwield it which deactivates it which makes the noise reference become null so when you try to delete it it you cant). We couldn't really figure out the problem besides just slapping a isfirsttimepredicted 🤷 

The third thing you listed is actually completely unrelated, its because when you switch modules its server side only so the client doesn't know to delete the audio basically. The `BorgModuleUnselectedEvent` would need to be predicted to fix it

Krunklehorn:
#38070 solves tests 2 & 3, but test 1 still persists.

https://github.com/user-attachments/assets/1daff28b-b0e5-48b2-a4a5-206d12b68645

PR #38070, opened 2025-06-05
Fixed the two-handed activated sound bug.

RedBookcase:
## About the PR
Fixed the bug that caused the activated hum sound effect on weapons like the double energy sword to stack infinitely. Documented in #37369 

## Why / Balance
Bugfix

## Technical details
The result of me and Beck staring at the ItemToggleActiveSoundComponent for a few hours. The bug seemed to be caused by an issue with prediction that was effectively causing the active sound effect to be duplicated and played differently for the client/server. This fix is not super comprehensive but it works for removing the most egregious version of this bug.

Assault Borgs had a very weird version of this bug that seemed to be caused by their specific version of the double-bladed energy sword - they had a 0-handed wieldable version which was extremely scuffed. The borg double esword has been updated to not require wielding to avoid weird issues related to this stuff in the future.

Additionally fixes an issue that caused the activation sound effect to play twice, seems to have been caused by a missing args.User line.

## Requirements

**Changelog**

no cl kill me

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
@@ -387,12 +387,65 @@
 # Borgs
 - type: entity
   suffix: One-Handed, For Borgs
-  parent: EnergySwordDouble
+  parent: EnergySword #To avoid the evil wielded component
   id: CyborgEnergySwordDouble # why is this invalid if ID is BorgEnergySwordDouble
+  name: double-bladed energy sword
   description: Syndicate Command Interns thought that having one blade on the energy sword was not enough. Specially designed for syndicate cyborgs.
   components: # could add energy-draining like the L6C
-  - type: Wieldable
-    freeHandsRequired: 0 # because borg has no off-hand to wield with.  Without this, it will be unable to activate the esword
+  - type: ItemToggle
+    soundActivate:
+      path: /Audio/Weapons/ebladeon.ogg
+      params:
+        volume: 3
+    soundDeactivate:
+      path: /Audio/Weapons/ebladeoff.ogg
+      params:
+        volume: 3
+  - type: ItemToggleMeleeWeapon
+    activatedSoundOnSwing:
+      path: /Audio/Weapons/eblademiss.ogg
+      params:
+        volume: 3
+        variation: 0.250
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
```diff
@@ -387,12 +387,65 @@
...
   suffix: One-Handed, For Borgs
-  parent: EnergySwordDouble
+  parent: EnergySword #To avoid the evil wielded component
   id: CyborgEnergySwordDouble # why is this invalid if ID is BorgEnergySwordDouble
+  name: double-bladed energy sword
   description: Syndicate Command Interns thought that having one blade on the energy sword was not enough. Specially designed for syndicate cyborgs.
   components: # could add energy-draining like the L6C
-  - type: Wieldable
-    freeHandsRequired: 0 # because borg has no off-hand to wield with.  Without this, it will be unable to activate the esword
+  - type: ItemToggle
```
EmoGarbage404: this is a goofy amount of yaml duplication. you can use composition to reduce like 99% of this

EmoGarbage404 (changes requested):

Resources/Prototypes/Entities/Objects/Weapons/Melee/e_sword.yml
RedBookcase: Forgot to trim it down, should be better now

Wintoli:
You'd be a god if you'd fix the merge conflicts, been awaiting a fix for this bug since FOREVERRRR

RedBookcase:
> Are you sure? I tried this with and without your changes to the borg double e-sword yaml and the result was that with your changes the borg couldn't even activate the sword at all due to a lack of hands, and I found no audio issues with reverting it to the old "0-hands" version.

Not sure what would be causing this issue, the fixed version of the sword is toggleable in the same way that the Assault Borg's pinpointer is, and in testing i didn't have any issues with turning it on and off. It's been a while since I worked on the PR but I distinctly remember the change to the DESword being necessary to avoid the sound duplication issue (Tested out fixes for a few hours in VC with Beck when I first started working on this a few months ago, explained a bit by him here https://github.com/space-wizards/space-station-14/issues/37369#issuecomment-2941905093).

PR #38074, opened 2025-06-05
Add player autofill to rename command.

### LINKED ISSUE #38060: `rename` command should have player autofill
Image

title

VerinSenpai:
## About the PR
Resolves #38060. Simple approach, just took what was done in KickCommand and copied it over. That said, I feel like it'd be better to have some way of getting player names in IPlayerManager as it seems to be done commonly enough. Could reduce duplication but thats just my input. 

Note, I was going to add admin logging to this as well but I wasn't really sure how to go about it. I stopped when I noticed the kick command also didn't have logging. Probably should be an issue opened for that.

## Requirements

### DIFF
--- a/Content.Server/Mind/Commands/RenameCommand.cs
+++ b/Content.Server/Mind/Commands/RenameCommand.cs
@@ -1,7 +1,5 @@
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using Content.Server.Administration;
-using Content.Server.Administration.Systems;
 using Content.Shared.Administration;
 using Content.Shared.CCVar;
 using Robust.Server.Player;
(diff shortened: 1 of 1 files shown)


Content.Server/Mind/Commands/RenameCommand.cs
```diff
@@ -60,4 +61,16 @@ private bool TryParseUid(string str, IConsoleShell shell,
         entityUid = EntityUid.Invalid;
         return false;
     }
+
+    public override CompletionResult GetCompletion(IConsoleShell shell, string[] args)
+    {
+        if (args.Length == 1)
+        {
+            var players = _playerManager.Sessions.OrderBy(c => c.Name).Select(c => c.Name).ToArray();
```
slarticodefast: Can't you use CompletionHelper.SessionNames for this or is that different?

VerinSenpai: Lemme take a look.

VerinSenpai: That is exactly what that is. I've also switched to using FromOptions instead of FromHintOptions. Hopefully thats alright? I may go through and switch some of the other commands that are doing the player lookup this way while doing my resolve prs.

PR #38076, opened 2025-06-05
Resprite Maint Hatch + New Syndicate Hatch

ps3moira:
## About the PR

Upstreaming Grey Station resprites for the hatch doors as well as creating a new more clearly depicted syndicate hatch door. Made airlocks not have wall icon smoothing.
## Why / Balance

These ones blend a bit better with the walls and color palette of other airlock doors. Also unsmoothed airlocks from walls to look nicer.

## Media

image


ps3moira:
Until someone will code in proper light shading for the door lights, I removed them because they don't visually work in it's current state.

K-Dynamic:
These hatches bring back the harsher black borders, not sure if that's good. At the same time it might be okay because there's no wall smoothing.

ps3moira:
was given the ok to just remove icon smoothing for all doors, most likely reason for icon smoothing to be used still was left overs from cev-eris days.

Blackern5000:
I'm not a maint but the icon smoothing change should definitely be a separate pr, it's only barely related to the hatch stuff just because it has something to do with airlocks.

ps3moira:
> I'm not a maint but the icon smoothing change should definitely be a separate pr, it's only barely related to the hatch stuff just because it has something to do with airlocks.

Was given approval by @EmoGarbage404

Resources/Prototypes/Entities/Structures/Doors/MaterialDoors/material_doors.yml
```diff
@@ -56,7 +56,7 @@
       - !type:DoActsBehavior
         acts: ["Destruction"]
   - type: IconSmooth
-    key: walls
+    key: door
```
slarticodefast: Does this do anything?
Do doors have icon smoothing sprites themselves?

Resources/Prototypes/Entities/Structures/Doors/Airlocks/base_structureairlocks.yml
```diff
@@ -159,7 +159,7 @@
   - type: PlacementReplacement
     key: walls
   - type: IconSmooth
-    key: walls
+    key: airlock
```
slarticodefast: same

slarticodefast (commented):
The icon smoothing change should be a separate PR.
Also you are missing a bunch of prototypes for that.
`HighSecDoor`, maybe `PlasticFlapsBase` and potentially the new xeno doors, but you may want to ask slam about those because they are different in design.

PR #38082, opened 2025-06-05
Prevent human based NPCs from automatically entering forced sleep after spawn

Vonsant:
## About the PR
This change prevents human-based NPCs from immediately being forced to sleep when they spawn. In the shared `SSDIndicatorSystem.Update` loop, it now checks for the presence of `NpcFactionMemberComponent` and skips applying `ForcedSleepingComponent` to any entity that has it. This effectively ensures that NPCs (which all carry `NpcFactionMemberComponent`) will no longer be put to sleep by the SSD indicator logic. 

## Why / Balance

Human-based NPCs defined in `/Resources/Prototypes/Entities/Mobs/NPCs/human.yml` were being sent to sleep right after they spawned. That behavior is undesirable for any NPC that is meant to stay active. Ideally, i would check for `HTNComponent` (which exists on all NPCs), but since `HTNComponent` lives in **Content.Server**, i cannot reference it from **Content.Shared**. 

## Technical details

Just added small check in SSDIndicatorSystem

## Requirements

## Breaking changes

Nothing.

**Changelog**

:cl:
- fix: Prevented human‐based NPCs from being forced to sleep after spawn by excluding entities with NpcFactionMemberComponent


Content.Shared/SSDIndicator/SSDIndicatorSystem.cs
```diff
@@ -82,6 +83,7 @@ public override void Update(float frameTime)
             if(ssd.IsSSD &&
                 ssd.FallAsleepTime <= _timing.CurTime &&
                 !TerminatingOrDeleted(uid) &&
+                !HasComp<NpcFactionMemberComponent>(uid) &&
```
lzk228: as i know, all nanotrasen workers (all players that join station jobs) have this component
if making this, better to add a bool to SSDComponent whether an entity should sleep after time

PR #25117, opened 2024-02-11
Add MIDI track names to the channel window

FairlySadPanda:
## About the PR
Add any MIDI track names in a loaded MIDI file to the channels window, so if your MIDI is competent it will tell you what your channels are for. Or not. The internet MIDI landscape is ancient and strange.

## Why / Balance
It sucked to not know what MIDI channels were.

## Technical details
This is not currently replicated as I have no idea how you'd do that; there's no replicated data that says someone is actually playing music, as far as I can see, as it's all done via firing MidiEvents around.

It turned out that the way that Fluidsynth is loaded is just by throwing a byte array at it, which meant there's no way of querying MIDI files being loaded for their text events. As such I did the mature senior software engineer thing of Googling to see if someone had written a MIDI loader that could do this.

The Midi file is from https://github.com/davidluzgouveia/midi-parser and is licensed under MIT. Other than deleting a few non-sandboxed things I've not changed it. Like most ideomatic C# code it throws, so I've protected us from the awfulness of the `throw` keyword with a try-catch surrounding the dirty code. 

This is a dirty hack of a PR and I'm happy to be pointed at how to do this more nicely.

## Media
image

- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**
:cl:
- add: You can now see any channel names of a MIDI file you've loaded whilst playing.

PR #38083, opened 2025-06-05
Add the instrument names to the MIDI channel selector

Simyon264:
## About the PR

This PR makes it so that the MIDI Channel Selector uses the instrument of the track instead of numbers 0 - 15.

## Why / Balance

Unless you already knew what MIDI channels were what, a simple number of 0 - 15 is not really helpful. The instrument that is supposed to be playing is however much more useful.

## Technical details

For making the MIDI parser I used http://www.somascape.org/midi/tech/mfile.html and intense research (google) for the bitshifting stuff.

For resolving the channel names the client parses the midi looking for the program change midi event along with the track name and instrument name to use as a fallback. This then gets sent over to the server which sets it in the ActiveInstrumentComponent which other clients then can look at for bands.

## Media

https://github.com/user-attachments/assets/fcb5bbeb-fbeb-4de7-92cd-1df8518585d0

https://github.com/user-attachments/assets/db375e2b-0828-4a3d-8d01-ca0e4c78fd43

## Requirements

**Changelog**

:cl:
- tweak: The MIDI Channel Selector now shows the instrument name of the track!


VasilisThePikachu:
Potentially useful to look at #25117 for any early code requests

Content.Client/Instruments/InstrumentSystem.MidiParsing.cs
```diff
@@ -0,0 +1,59 @@
...
+            switch (midiTrack)
+            {
+                case { TrackName: not null, ProgramName: not null }:
+                    resolvedTracks.Add($"{midiTrack.TrackName} ({midiTrack.ProgramName})");
+                    break;
+                case { TrackName: not null, InstrumentName: not null }:
+                    resolvedTracks.Add($"{midiTrack.TrackName} ({midiTrack.InstrumentName})");
+                    break;
+                case { TrackName: not null }:
+                    resolvedTracks.Add($"{midiTrack.TrackName}");
```
PJB3005: This motherfucker formatting a string into itself.

Content.Client/Instruments/InstrumentSystem.MidiParsing.cs
```diff
@@ -0,0 +1,59 @@
...
+                    break;
+                case { ProgramName: not null }:
+                    resolvedTracks.Add($"{midiTrack.ProgramName}");
```
PJB3005: Ditto.

Content.Client/Instruments/InstrumentSystem.cs
```diff
@@ -252,7 +253,7 @@ public bool OpenInput(EntityUid uid, InstrumentComponent? instrument = null)
 
     }
 
-    public bool OpenMidi(EntityUid uid, ReadOnlySpan<byte> data, InstrumentComponent? instrument = null)
+    public bool OpenMidi(EntityUid uid, byte[] data, InstrumentComponent? instrument = null)
```
PJB3005, line 284: This is a breaking change. IMO it's not worth documenting, just add another overload that takes a span and copies it, but marked `[Obsolete]`.

Content.Client/Instruments/MidiParser/MidiParser.cs
```diff
@@ -0,0 +1,223 @@
...
+using System.Text;
+using Robust.Shared.Utility;
+
+namespace Content.Client.Instruments.MidiParser;
+
+public static class MidiParser
+{
+    // Based on https://www.ccarh.org/courses/253/handout/gminstruments/
+    // Maybe localize? Idk.
+    private static readonly string[] GeneralMidiInstruments =
```
PJB3005: You're not getting out of localizing these.

Content.Client/Instruments/MidiParser/MidiParser.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        if (stream.ReadString(4) != "MThd")
+        {
+            error = "Invalid file header";
+            return false;
+        }
+
+        var headerLength = stream.ReadInt32();
+        // MIDI specs define that the header is 6 bytes, we only look at the 6 bytes, if its more, we skip ahead.
+        DebugTools.Assert(headerLength == 6, $"Invalid header length, expected 6, got {headerLength}");
```
PJB3005: If you're designing this function to handle invalid formats, you should not throw here.

Content.Client/Instruments/MidiParser/MidiStreamWrapper.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// </summary>
+    public ushort ReadInt16()
+    {
+        var bytes = ReadBytes(2);
+        return (ushort)((bytes[0] << 8) | bytes[1]);
+    }
+
+    public string ReadString(int count)
+    {
+        var bytes = new byte[count];
```
PJB3005: Why not use your existing `ReadBytes()` here?

Content.Client/Instruments/MidiParser/MidiStreamWrapper.cs
```diff
@@ -0,0 +1,107 @@
...
+        if (read != count)
+            throw new Exception("Unexpected end of stream");
+
+        return _buffer;
+    }
+
+    /// <summary>
+    /// Reads a 4 byte big-endian int.
+    /// </summary>
+    public uint ReadInt32()
```
PJB3005: If these are returning uints, they really should reflect that in the function name.

Content.Client/Instruments/UI/ChannelsMenu.xaml.cs
```diff
@@ -51,15 +53,23 @@ private void OnClearPressed(BaseButton.ButtonEventArgs obj)
...
-            var item = ChannelList.AddItem(_owner.Loc.GetString("instrument-component-channel-name",
-                ("number", i)), null, true, i);
+            var label = _owner.Loc.GetString("instrument-component-channel-name",
+                ("number", i));
+            if (activeInstrument != null
+                && activeInstrument.Channels.TryGetValue(i, out var resolvedMidiChannel)
+                && resolvedMidiChannel != null
+                )
+            {
+                label = resolvedMidiChannel;
```
PJB3005: I would keep the numbers of the MIDI channel in there just for clarity. There's definitely MIDIs with multiple tracks with the same instrument, so some distinction is good.

Content.Client/Instruments/UI/InstrumentBoundUserInterface.cs
```diff
@@ -103,7 +104,13 @@ public void OpenChannelsMenu()
             _channelsMenu ??= new ChannelsMenu(this);
             EntMan.TryGetComponent(Owner, out InstrumentComponent? instrument);
 
-            _channelsMenu.Populate(instrument);
+            ActiveInstrumentComponent? activeInstrument = null;
+            if (instrument is { Master: not null })
+                EntMan.TryGetComponent(instrument.Master, out activeInstrument);
```
PJB3005: Masters can actually be chained infinitely AFAIK. Not sure if you need to account for that?

perryprog:
You're a hero for this. Mini-rant: The issue I ran into when I tried to (foolishly) attempt implementing this is I _also_ wanted track names as given by SMF to be displayed, as /many/ MIDIs will have informative text their on what instrument a track is meant to be played with that can differ from the General Midi program that they used. (This presumably is done due to the soundfont the creator used.) While this is a _massive_ pain to implement, it's something I wanted to bring up. (It can also go poorly if the MIDI uses track labels as metadata... or just uses a single track filled with eight million program change events. Yes, these exist and it's frustratingly one of my best MIDIs.)

Anyway, It's frustratingly difficult to really properly represent this into something that's just a channel selection since there's no one-to-one mapping so it's probably not possible to implement anything that takes this into account, but wanted to mention it.

Regarding actual PR: my main concern is that this doesn't properly account for program change events sent to a channel after the first one. This could lead to some confusion for MIDIs that ~~abuse~~ use this as a way to have more than sixteen instruments. That being said, it's still miles ahead of what we had before, so I don't think this is a real issue.

Also: A lot of MIDI exporters tend to put garbage program change events on channel 9 despite it being fixed as percussion. (Very often it gets a program change to 0, making it display as a grand piano.) Could maybe just hardcode this to say "Percussion". (Even though it's really "Non-chromatic percussion" :P)

It's also _probably_ not worth doing but having numbers can be nice for coordination between multiple musicians. Should probably make it be numbered 1–16 though instead of the silly 0–15. (This definitely has nothing to do with the fact that I have pages of notes about good-sounding channel assignments to use with my MIDIs just for this game. I also _definitely_ don't have a program that dumps channel change events for each given track specifically for this game either...)

I'll leave the code review to the bigger nerds but I suspect localizing the GM instrument names would be nice... (sorry).

Content.Shared/Instruments/SharedInstrumentComponent.cs
```diff
@@ -38,7 +38,13 @@ public abstract partial class SharedInstrumentComponent : Component
 /// Component that indicates that musical instrument was activated (ui opened).
 /// </summary>
 [RegisterComponent, NetworkedComponent]
-public sealed partial class ActiveInstrumentComponent : Component;
+[AutoGenerateComponentState]
+public sealed partial class ActiveInstrumentComponent : Component
+{
+    [DataField]
+    [AutoNetworkedField]
+    public List<string?> Channels = new();
```
PJB3005: Why is this a list and not an array?

PJB3005 (changes requested):

perryprog:
Ah, you do have some track name handling, but in testing I'm not seeing cases where it would come up. Barring the above issues with how track names can be absurdly weird in many MIDI files, it might be worth having some option, maybe as a checkbox in the channels menu, to display track names regardless? Not totally sure.

Content.Client/Instruments/InstrumentSystem.MidiParsing.cs
Simyon264: I was tired okay? (Bad copy and paste)

Content.Client/Instruments/MidiParser/MidiStreamWrapper.cs
Simyon264: I previously had a separate method that used its own buffer and didn't just resize the normal buffer, then I made the ReasBytes method resize and forgot to update it

perryprog:
Additional nitpick—this won't update channel selection windows without closing and reopening them, which is a bit of a pain. Would be nice if it auto-refreshed.

Simyon264:
> Additional nitpick—this won't update channel selection windows without closing and reopening them, which is a bit of a pain. Would be nice if it auto-refreshed.

I can certainly fix that

Simyon264:
> Ah, you do have some track name handling, but in testing I'm not seeing cases where it would come up. Barring the above issues with how track names can be absurdly weird in many MIDI files, it might be worth having some option, maybe as a checkbox in the channels menu, to display track names regardless? Not totally sure.

I could maybe make it a toggle, but that would increase complexity somewhat. I'll see if I can manage to make the UI look good.

BigfootBravo:
Does this work with MIDI Input? Regarding the use of like MidiPlayer being set to input into SS14, and other such things.

Simyon264:
> Does this work with MIDI Input? Regarding the use of like MidiPlayer being set to input into SS14, and other such things.

I don't have a MIDI input device, however from testing it seems that the channels will just be named with the default values.

Simyon264:
@perryprog Added support for a toggle to show the track names!
Also updates on it's own now.

https://github.com/user-attachments/assets/e5cea2da-073b-4cf0-9585-34ea2da0ba0c

perryprog:
How the heck did you copy what my window layout looks like when I play musician? :)

And that looks awesome—thanks so much for this.

PJB3005 (changes requested):
print

Simyon264:
> print


https://github.com/user-attachments/assets/69123022-49e9-4296-94b7-c8c559a673d0

Content.Client/Instruments/InstrumentSystem.cs
```diff
@@ -38,6 +41,15 @@ public override void Initialize()
...
         SubscribeLocalEvent<InstrumentComponent, ComponentHandleState>(OnHandleState);
+        SubscribeLocalEvent<ActiveInstrumentComponent, AfterAutoHandleStateEvent>(OnActiveInstrumentAfterHandleState);
+    }
+
+    private void OnActiveInstrumentAfterHandleState(Entity<ActiveInstrumentComponent> ent, ref AfterAutoHandleStateEvent args)
+    {
+        // Called in a timer so that the components update client side for resolving them in TryComps.
+
+        if (OnChannelsUpdated != null)
+            Timer.Spawn(1, OnChannelsUpdated.Invoke);
```
PJB3005: Do this from `Update()` instead.

Content.Shared.Database/LogType.cs
```diff
@@ -472,5 +472,10 @@ public enum LogType
     /// <summary>
     /// Damaging grid collision has occurred.
     /// </summary>
-    ShuttleImpact = 102
+    ShuttleImpact = 102,
+
+    /// <summary>
+    /// Events relating to midi playback.
+    /// </summary>
+    Midi = 103,
```
PJB3005: I'd make this `Instrument`.

Content.Server/Instruments/InstrumentSystem.cs
```diff
@@ -132,6 +139,44 @@ private void OnMidiStop(InstrumentStopMidiEvent msg, EntitySessionEventArgs args
...
+
+        if (!TryComp(uid, out InstrumentComponent? instrument) || !TryComp(uid, out ActiveInstrumentComponent? activeInstrument))
+            return;
+
+        if (args.SenderSession.AttachedEntity != instrument.InstrumentPlayer)
+            return;
+
+        if (msg.Tracks.Length > RobustMidiEvent.MaxChannels)
+        {
+            DebugTools.Assert($"Tried to send tracks over the limit! Received: {msg.Tracks.Length}; Limit: {RobustMidiEvent.MaxChannels}");
```
PJB3005: Dealing with invalid client data should not throw exceptions. Kick the client or log a warning or something.

Content.Server/Instruments/InstrumentSystem.cs
```diff
@@ -132,6 +139,44 @@ private void OnMidiStop(InstrumentStopMidiEvent msg, EntitySessionEventArgs args
...
+            return;
+        }
+
+        var tracksString = string.Join("\n",
```
PJB3005, line 159: Do we sanely handle newlines in admin log entries like this?

Content.Client/Instruments/MidiParser/MidiInstrument.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// Turns the given enum value into it's string representation to be used in localization.
+    /// </summary>
+    public static string GetStringRep(this MidiInstrument instrument)
+    {
+        return instrument
+            .ToString()
+            .Replace(" ", "_")
+            .Replace("-", "_")
+            .Replace("'", "")
+            .Replace("\"", "")
```
PJB3005: I don't understand. None of these enum names contain any of these characters. Why is this needed?

Content.Client/Instruments/MidiParser/MidiInstrument.cs
```diff
@@ -0,0 +1,152 @@
+﻿// ReSharper disable InconsistentNaming
+namespace Content.Client.Instruments.MidiParser;
+
+// This file was autogenerated. Based on https://www.ccarh.org/courses/253/handout/gminstruments/
+public enum MidiInstrument : byte
```
PJB3005, line 6: Make these PascalCase. Then use the helper we have somewhere to turn it into the kebab-case style for the localization key lookup.

Resources/Locale/en-US/instruments/instruments-component.ftl
```diff
@@ -19,6 +19,136 @@ instruments-component-channels-menu = MIDI Channel Selection
 instrument-component-channel-name = MIDI Channel {$number}
 instruments-component-channels-all-button = All
 instruments-component-channels-clear-button = Clear
+instruments-component-channels-track-names-toggle = Toggle Track Names
```
PJB3005: This doesn't explain whether the checkbox state corresponds to "show" or "not show".

Content.Client/Instruments/MidiParser/MidiParser.cs
```diff
@@ -0,0 +1,184 @@
...
+                        {
+                            // Program Change
+                            case 0xC0:
+                            {
+                                var programNumber = stream.ReadByte();
+                                if (track.ProgramName == null)
+                                {
+                                    if (programNumber < Enum.GetValues<MidiInstrument>().Length)
+                                        track.ProgramName = Loc.GetString($"instruments-component-menu-midi-channel-{((MidiInstrument)programNumber).GetStringRep()}");
+                                }
```
PJB3005, line 144: So I don't know how this works but don't you need to make sure this program change is on bank 0?

Content.Client/Instruments/MidiParser/MidiStreamWrapper.cs
```diff
@@ -0,0 +1,103 @@
...
+public sealed class MidiStreamWrapper
+{
+    private readonly MemoryStream _stream;
+    private byte[] _buffer;
+
+    public long StreamPosition => _stream.Position;
+
+    public MidiStreamWrapper(byte[] data)
+    {
+        _stream = new MemoryStream(data);
```
PJB3005: Mark as read-only. Can't hurt.

```suggestion
        _stream = new MemoryStream(data, false);
```

Content.Client/Instruments/UI/ChannelsMenu.xaml.cs
```diff
@@ -1,26 +1,51 @@
...
+        IoCManager.InjectDependencies(this);
         _owner = owner;
 
         ChannelList.OnItemSelected += OnItemSelected;
         ChannelList.OnItemDeselected += OnItemDeselected;
         AllButton.OnPressed += OnAllPressed;
         ClearButton.OnPressed += OnClearPressed;
+        DisplayTrackNames.OnPressed += OnDisplayTrackNamesPressed;
+
+        _owner.Instruments.OnChannelsUpdated += UpdateChannelList;
```
PJB3005, line 35: This should be added to in `EnteredTree` instead.

Content.Client/Instruments/UI/ChannelsMenu.xaml.cs
```diff
@@ -51,15 +76,64 @@ private void OnClearPressed(BaseButton.ButtonEventArgs obj)
...
+                && activeInstrument.Tracks.TryGetValue(i, out var resolvedMidiChannel)
+                && resolvedMidiChannel != null
+                )
+            {
+                if (DisplayTrackNames.Pressed)
+                {
+                    label = resolvedMidiChannel switch
+                    {
+                        { TrackName: not null, InstrumentName: not null } =>
+                            $"{i} {resolvedMidiChannel.TrackName} ({resolvedMidiChannel.InstrumentName})",
```
PJB3005: This formatting should be done in Fluent.

Content.Client/Instruments/UI/ChannelsMenu.xaml.cs
```diff
@@ -51,15 +76,64 @@ private void OnClearPressed(BaseButton.ButtonEventArgs obj)
...
         for (int i = 0; i < RobustMidiEvent.MaxChannels; i++)
         {
-            var item = ChannelList.AddItem(_owner.Loc.GetString("instrument-component-channel-name",
-                ("number", i)), null, true, i);
+            var label = _owner.Loc.GetString("instrument-component-channel-name",
+                ("number", i));
+            if (activeInstrument != null
+                && activeInstrument.Tracks.TryGetValue(i, out var resolvedMidiChannel)
+                && resolvedMidiChannel != null
+                )
```
PJB3005: Don't put this closing parenthesis on its own line.

Content.Client/Instruments/UI/ChannelsMenu.xaml.cs
```diff
@@ -51,15 +76,64 @@ private void OnClearPressed(BaseButton.ButtonEventArgs obj)
...
+            {
+                if (DisplayTrackNames.Pressed)
+                {
+                    label = resolvedMidiChannel switch
+                    {
+                        { TrackName: not null, InstrumentName: not null } =>
+                            $"{i} {resolvedMidiChannel.TrackName} ({resolvedMidiChannel.InstrumentName})",
+                        { TrackName: not null } => $"{i} {resolvedMidiChannel.TrackName}",
```
PJB3005: Ditto.

Content.Client/Instruments/UI/ChannelsMenu.xaml.cs
```diff
@@ -51,15 +76,64 @@ private void OnClearPressed(BaseButton.ButtonEventArgs obj)
...
+                        _ => label,
+                    };
+                }
+                else
+                {
+                    label = resolvedMidiChannel switch
+                    {
+                        { ProgramName: not null } =>
+                            $"{i} {resolvedMidiChannel.ProgramName}",
```
PJB3005: Ditto.

PJB3005 (changes requested):

Content.Client/Instruments/MidiParser/MidiParser.cs
perryprog: If the MIDI is using GM2 (rare, sadly) then any program changes that do happen on different banks will end up being "close enough" to the instrument in bank 0 for the same program. For example, program 29 in bank 0 is "Muted Electric Guitar", while banks 1, 2, and 3 are "Funky Guitar 1", "Funky Guitar 2", and... "Jazz Man" (that is literally all it says in the spec, lol). This is broken for SFX like Applause having other banks map it to Laughing, Screaming, etc., but it works well enough that it saves a lot of localization while still being accurate to the original MIDI.

Content.Server/Instruments/InstrumentSystem.cs
Simyon264: image

Seems to work.

Content.Client/Instruments/MidiParser/MidiInstrument.cs
Simyon264: This was a bad copy + paste from the method I used to turn the locale into an enum. Only the .ToString and .ToUpperInvariant would be needed.

Content.Server/Instruments/InstrumentSystem.cs
PJB3005: Good enough for me.

Content.Client/Instruments/UI/ChannelsMenu.xaml.cs
```diff
@@ -23,6 +27,32 @@ public ChannelsMenu(InstrumentBoundUserInterface owner) : base()
         ClearButton.OnPressed += OnClearPressed;
     }
 
+    protected override void EnteredTree()
+    {
+        base.EnteredTree();
+
+        DisplayTrackNames.OnPressed += OnDisplayTrackNamesPressed;
```
PJB3005: This should be added in the constructor instead.

PJB3005 (changes requested):

PJB3005 (approved):

PR #38084, opened 2025-06-05
add ShouldSleep flag to exclude human NPCs from auto-sleep

Vonsant:
## About the PR

This PR adds a new boolean flag `ShouldSleep` to the existing `SSDIndicatorComponent` and updates the `SSDIndicatorSystem` logic so that entities with `ShouldSleep = false` are not automatically sent to sleep by the SSD timer. NPCs (which possess an HTN component) can now opt out of the auto‐sleep behavior by setting `shouldSleep: false` in their prototype.

## Why / Balance

Previously, any human-based NPC  would be put to sleep automatically when the SSD countdown expired. This caused hostile NPCs to stand around sleeping instead of remaining active. By introducing `ShouldSleep`, it prevent HTN NPCs from being bench‐ridden by the SSD timer, ensuring that only player‐controlled or otherwise intended SSD users actually fall asleep. 
It simply fixes unintended NPC downtime.

## Technical details

* Each entity can decide whether it can be put to sleep by the SSD system.

* Timers are only set if `ShouldSleep == true`.
* Only entities allowing sleep get their `FallAsleepTime` initialized.
* Entities with `ShouldSleep = false` are skipped entirely.

No new components or systems were created.

## Media

No in‐game sprite or UI changes. No media attachments are required.

## Requirements

* [x] I have read and am following the [[Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html)](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
* [x] I have added media to this PR or it does not require an in‐game showcase.

## Breaking changes

 Every NPC prototype that relies on HTN and should be exempt from SSD sleep must now include:

  ```yaml
  - type: SSDIndicator
    shouldSleep: false
  ```

If this line is omitted, the NPC will default to `ShouldSleep = true` and will be subject to the SSD sleep timer.

**Changelog**
:cl:

* tweak: Human-based NPC no longer automatically fall asleep due to SSD timers.


EmoGarbage404 (changes requested):
if we only want non-NPCs to have this enabled, why don't you just check for the HTN component in the code rather than adding in a new boolean? It would be cleaner and wouldn't require manually configuring the YAML.

Vonsant:
> if we only want non-NPCs to have this enabled, why don't you just check for the HTN component in the code rather than adding in a new boolean? It would be cleaner and wouldn't require manually configuring the YAML.

Since HTN is only available in `content.server` and `content.client`, I can’t use it in `content.shared`. I originally planned to put it there, but was advised to add it directly in the component instead. If necessary, I can rework it again - though that would mean adding HTN into shared

EmoGarbage404:
This is a bit hideous, but you can put a virtual function on a shared system called HasHtnComponent and then override it on both the client and server to check if the component is there.

Vonsant:
If you think this approach would be better than my current implementation, I’ll work on it on sunday 👍

iaada:
I'm tagging this as stale since it's had a merge conflict for over a month, and is still awaiting changes. Let me know if you plan to continue working on it.

PR #38085, opened 2025-06-05
feat: allow map editor copying of pipes on any layer

### LINKED ISSUE #38031: "Copy" won't pick up non-standard configuration pipes
## Description

Using the map editor copy key (probably P) doesn't seem to work on pipes that aren't in standard configurations.

**Reproduction**

1. Place a pipe that isn't in standard configuration.
2. Try to copy it.
3. Be sad.

Doesn't affect atmos devices, just pipes.

perryprog:
## About the PR

Allow copying pipes via the Copy (usually P) shortcut even if they aren't on the primary layer.

## Why / Balance

Fixes #38031.

## Technical details

Kinda gross hard-coding it in like this. I guess could check the comp's alternativePrototypes and see if its primary prototype is hidden from the spawn menu? Not sure it matters though.

## Media

image

## Requirements

## Breaking changes

**Changelog**


slarticodefast (changes requested):
This should really not be hardcoded like this.

perryprog:
Hm; I'm not sure what else would be reasonable. The only other thing I can think of is adding a field to the entity prototype that always allows copying, even if hidden from the spawn menu?

perryprog:
Here's a tweak that adds an entity category that acts as an exemption. I do wonder if it should be made inheritable as that would allow just applying the category once on GasPipeLayerAlt1 and GasPipeLayerAlt2, but that would make its behavior differ from the category that Robust provides, and I'm not sure if that's worth doing.

PR #38093, opened 2025-06-06
Remade the AI Camera Static Shader

southbridge-fur:
## About the PR

Remade the shader used for the AI camera static so it doesn't randomly shift off to one side.

It's technically a little more efficient since it uses less function calls but it's not that much different.

Also this uses our default zRandom function instead of importing one.

## Why / Balance

It annoyed me.

## Technical details

🌸 Shader code 🌸 

## Media

https://github.com/user-attachments/assets/7ec93361-637c-415b-9110-4c78eb0ac118

## Requirements

## Breaking changes

**Changelog**

:cl: Southbridge
- tweak: AI camera static shader has been updated to be more consistent.

### DIFF
--- a/Resources/Textures/Shaders/camera_static.swsl
+++ b/Resources/Textures/Shaders/camera_static.swsl
@@ -1,40 +1,6 @@
-// Credited to PelicanPolice on Shadertoy
-// Free for any purpose, commercial or otherwise.
-// But do post here so I can see where it got used!
-// If using on shadertoy, do link to this project on yours :)
-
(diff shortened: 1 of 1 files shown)


beck-thompson:
Does this also effect cameras?

southbridge-fur:
> Does this also effect cameras?

I guess it does but that shouldn't be an issue

beck-thompson (approved):
seems good but I did prefer the old shader for cameras

PR #38111, opened 2025-06-06
Fix wallmount interaction

slarticodefast:
## About the PR
The side you could interact with them from was rotating with the station.
Now it works as intended again.

## Why / Balance
bugfix

## Technical details
In some of the codepaths it was using a different rotation.
While the code comment says that the one code path is for "no fixtures, e.g. wallmounts" there are wallmounts that have fixtures, for example the nanomed.
Please test this PR thoroughly before merging as this interaction check is also used elsewhere, for example for melee attacks.

## How to test:
- delete the station anchor
- spawn a wallmount nanomed and a solid wall
- reduce the interaction arc in the wallmount component, and try setting different rotations in the datafield
- repeat with different rotations of the wallmount entity
- spawn a shuttle console and 10 gyroscopes, rotate the station and repeat the test
- delete the `FixturesComponent` from the wallmount, as that takes a different code path, and repeat the test

grafik
grafik

## Requirements

## Breaking changes

**Changelog**
:cl:

- fix: Fixed wallmount interaction rotation.

### DIFF
--- a/Content.Shared/Interaction/SharedInteractionSystem.cs
+++ b/Content.Shared/Interaction/SharedInteractionSystem.cs
@@ -744,7 +744,7 @@ namespace Content.Shared.Interaction
             var inRange = true;
             MapCoordinates originPos = default;
             var targetPos = _transform.ToMapCoordinates(otherCoordinates);
-            Angle targetRot = default;
+            Angle targetRot = _transform.GetWorldRotation(otherCoordinates.EntityId) + otherAngle;
 
             // So essentially:
             // 1. If fixtures available check nearest point. We take in coordinates / angles because we might want to use a lag compensated position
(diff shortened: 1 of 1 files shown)


Fildrance (approved):
The whole wallmount interactions with nanomed looks kinda clunky - currently on master i saw that interactions with bottom and left are always okay no matter the rotation of nanomed, and now it depends on wallmount rotation and your relative position, which is... ugh... better? but at the same time can potentially hurt some maps?.. i'm not sure it will make things worse anyways.

Fildrance:
[nanomed interactions.webm](https://github.com/user-attachments/assets/bcd2d6e9-592f-414c-a2df-019825d6278d)

btw this is what i was talking about being clunky

slarticodefast:
> and now it depends on wallmount rotation and your relative position, which is... ugh... better

Pretty sure that is how wallmounts are supposed to work. You can only interact with them from the side they are rotated to. If you have a 1 tile thick wall this means you can only interact with it from one room.

Issue #38103, opened 2025-06-06
Items dispensed by wallmounts get stuck in walls


K-Dynamic:
## Description

**Reproduction**

1. Spawn wallmount that dispenses item
2. Item doesn't drop by your feet compared to wallmount closets

**Screenshots**

![Image](https://github.com/user-attachments/assets/4c1f20b4-393d-4238-ac86-2b34dbc28bcf)
![Image](https://github.com/user-attachments/assets/9023bc9d-5929-49f2-9300-347539cf59ab)
![Image](https://github.com/user-attachments/assets/6e980703-39d7-4dfa-a614-ee20326b937f)

**Additional context**
Doesn't appear to affect stored items (i.e. using a wallmount closet will drop items properly)

kosticia:
I swear I saw prs with fix of this bug for five times

kosticia:
#28279 hmmmmm...

K-Dynamic:
> I swear I saw prs with fix of this bug for five times

It's like a cockroach that refuses to die

IamVelcroboy:
Oh gods, not again 😩

PR #38102, opened 2025-06-06
Genpop wall lockers

K-Dynamic:
## About the PR

- Adds genpop wall lockers
  - Splits genpop function from prisoner locker to use as parent for prisoner wall lockers to avoid code duplication
  - ID spawns away from wall locker to avoid being stuck
- Moved appearance component to BaseWallCloset

## Why / Balance

More variety for mapping and can be used to save space in small areas.

## Technical details

N/A

## Media

image
dotnet_SS77RNJpwJ

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

N/A

**Changelog**

Not player-facing

PR #38112, opened 2025-06-06
fix wallmount vending machine dispense direction

slarticodefast:
## About the PR
It was always dispensing south.

## Why / Balance
Fixes https://github.com/space-wizards/space-station-14/issues/38103

## Technical details
We use local coordinates (EntityCoordinates) to spawn the item, so the offset should use local rotation as well.

## Media
grafik

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- fix: Fixed wallmount vending machines dispensing in the wrong direction.

### DIFF
--- a/Content.Server/VendingMachines/VendingMachineSystem.cs
+++ b/Content.Server/VendingMachines/VendingMachineSystem.cs
@@ -227,14 +227,13 @@ namespace Content.Server.VendingMachines
             }
 
             // Default spawn coordinates
-            var spawnCoordinates = Transform(uid).Coordinates;
+            var xform = Transform(uid);
+            var spawnCoordinates = xform.Coordinates;
 
             //Make sure the wallvends spawn outside of the wall.
-
             if (TryComp<WallMountComponent>(uid, out var wallMountComponent))
             {
-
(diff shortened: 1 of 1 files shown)


kosticia:
Just want notice that is pr doesn't fixes issue. Only vending machins. If someone will want add another wallmount that will spawn entities it still will cause issues. I'm not even sure that this issue can be fixed in our current gamestate.

slarticodefast:
> Just want notice that is pr doesn't fixes issue. Only vending machins. If someone will want add another wallmount that will spawn entities it still will cause issues. I'm not even sure that this issue can be fixed in our current gamestate.

Other wallmounts work fine
grafik

If someone adds a new one, for example the genpop lockers, then they will of course have to offset any item spawn location according to the wallmounts rotation. This PR https://github.com/space-wizards/space-station-14/pull/38102 just forgot to do that, as the GenpopSystem never had that in mind.

kosticia:
> If someone adds a new one, for example the genpop lockers, then they will of course have to offset any item spawn location according to the wallmounts rotation. This PR https://github.com/space-wizards/space-station-14/pull/38102 just forgot to do that, as the GenpopSystem never had that in mind.

Okey, I take my words back. This is most proper solution anyway.

PR #38122, opened 2025-06-07
Adds the Estoc DMR to the syndicate uplink

august-sun:
## About the PR
Adds the Estoc DMR to the syndicate uplink, a burst/semi-auto rifle fitted with a mid-range optic. Available for 18 TC, individual magazines available for 2 TC.

## Why / Balance
The Estoc offers new engagement dynamics for syndicate antagonists, utilizing the optic system intended for the Hristov. It's a varied counterpart to the Lecter; the Lecter has a slow rate of fire and no optic, but also features a full-auto mode, as well as no slowdown when wielding the weapon. In contrast, the Estoc features only burst and semi-auto fire, as well as a slight slowdown when wielded, but features an optic and a very quick burst of rounds that are relatively accurate at range. 

The weapon can still be fired one-handed, but suffers a significant spread penalty for doing so. Similar to the other weapons, it accepts ammunition/magazines from its crew counterpart.

## Technical details
Added the prototype and sprites for the weapon. Added the weapon to a new bundle for the syndicate uplink. Mixed and added a new firing sound for the weapon.

## Media
image
image

https://github.com/user-attachments/assets/4913f10d-41ee-4fd5-a121-f8876cf9b8a0

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added the Estoc DMR, a burst-fire marksman rifle with an optic. Available for syndicate operatives at 18 TC.


august-sun:
Sidenote, getting initial feedback here that I need to tighten the weapon spread, which I can also see, I tend to start weaker balance-wise and adjust up!

Archisteks:
#35087 might be a necessary merge before this. Hristov is borderline unusable on Separated Chat HUD currently. This would most likely also apply to the Estoc.

NoahCHeaverin:
The sound fits, I think. Copyright looks good as well, minus a bit of weirdness involving a relicensing.

Pumkin69:
I think this is a cool weapon and the sounds are nice. This also provides nukies with an actual rifle to use instead of either an SMG or an LMG so thats also very nice. It does need to be slightly more accurate at range if you are gonna make it have a scope imo but obviously we need actual data to see if thats true.

august-sun:
> I think this is a cool weapon and the sounds are nice. This also provides nukies with an actual rifle to use instead of either an SMG or an LMG so thats also very nice. It does need to be slightly more accurate at range if you are gonna make it have a scope imo but obviously we need actual data to see if thats true.

Yep, tuned it up to be a tighter burst! The newly-attached video should be a better reflection.

TiniestShark:
Solid artwork! Definitely gets my approval.

Pumkin69:
> > I think this is a cool weapon and the sounds are nice. This also provides nukies with an actual rifle to use instead of either an SMG or an LMG so thats also very nice. It does need to be slightly more accurate at range if you are gonna make it have a scope imo but obviously we need actual data to see if thats true.
> 
> Yep, tuned it up to be a tighter burst! The newly-attached video should be a better reflection.

The spread is much much better!

august-sun:
> Alright, I've taken the time to review the gun more in-depth. It's an interesting gun (and required me to rewrite my burst damage TTC calculation!) and I have the following thoughts:
> 
>     * It doesn't encroach much on any gun that antags have access to at the moment, to the point where it might cover them too well and be a good option for every situation. The downsides of the gun might weigh against it, but it's something worth keeping an eye on.
> 
>     * This weapon is a bit on the strong side compared to the C-20r. The Time To Crit with bursts weapons can be a bit deceptive, since it is heavily dependent on whether the killing shot happens at the end or start of a burst.
>       
>       * E.g. for an unarmored target, the 130 DPS makes the raw TTC 0.62s. But the _actual_ TTC is 0.56s, since we don't have to count in any additional burst delays that the DPS is concerned with.
>       * For comparison, the C-20r raw TTC is 0.66s and actual TTC is 0.75s against an unarmored target.
> 
>     * The Estoc have features that distinguish it from the C-20r; no mag auto-eject, less available ammo, unusual firing mode, 2-tile scope and slowdown. The slowdown is probably going to be the most important aspect in ensuring that it doesn't encroach on the C-20r's territory, so I'm glad that it is there.
> 
> 
> Overall I think this gun sounds interesting to try and I think it might work. I _am_ concerned about the TTC however. If we are to merge this, I would want us to have an open mind to dropping the firerate down to 14 or even 12 if it ends up being too effective, as it may be on the edge of being overtuned.

Great to hear! And thankfully enough, the burst rate and the detractors are easily tunable, the big key standout is mostly the range and accuracy. My thought process was that an evasive security or crew member could get around a shooter equipped with it, while the shooter's advantage is seeing their enemy earlier. So, happy to see it given a chance!

SlamBamActionman:
> My thought process was that an evasive security or crew member could get around a shooter equipped with it, while the shooter's advantage is seeing their enemy earlier. So, happy to see it given a chance!

I am not entirely certain that this will be effective due to its DPS just being so high, but I do think this gun will excell in holding down positions (not as much pushing due to the movement speed penalty). The Hristov's stats have the effect you describe, where flanking is very potent due to extremely long firing delay. My guess is that the Estoc will be a bit of a marksman weapon where you really wanna be accurate, as missing is more punishing (both due to the forced burst and smaller magsize). I wouldn't rush someone with an Estoc though, the DPS is far too high for that.

august-sun:
> > My thought process was that an evasive security or crew member could get around a shooter equipped with it, while the shooter's advantage is seeing their enemy earlier. So, happy to see it given a chance!
> 
> I am not entirely certain that this will be effective due to its DPS just being so high, but I do think this gun will excell in holding down positions (not as much pushing due to the movement speed penalty). The Hristov's stats have the effect you describe, where flanking is very potent due to extremely long firing delay. My guess is that the Estoc will be a bit of a marksman weapon where you really wanna be accurate, as missing is more punishing (both due to the forced burst and smaller magsize). I wouldn't rush someone with an Estoc though, the DPS is far too high for that.

In that case, I can turn the burst rate down to 14 to take some of the edge off!

beck-thompson (approved):

PR #38086, opened 2025-06-05
Add hair back to hardhats

vmzd:
## About the PR
Reverts one of the changes made in #36818 to make hardhats stop hiding hair
## Why / Balance
While I understand why the change was done, hardhats now make most characters pretty ugly

## Technical details
simple yaml removal

## Media
hardhat

## Requirements

- [ X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**
The original PR didn't explicitly mention hardhats so I don't think CL is needed

PR #38127, opened 2025-06-07
Revert "Add hair back to hardhats (#38086)"

qrwas:
This reverts commit 4cbfd1931285a80b5f390e62dc8ace1e409ae8a0. [38086]

## About the PR

Revert "Add hair back to hardhats (#38086)"
## Why / Balance

That's not the way to do it. It makes the other characters look really weird, especially with their hair sticking out.

Until there's a better option, it's easier to accept that safety measures prohibit long hair
## Media

This is what different hair looks like now:

image

That's why it was done in the first place:
image

## Requirements

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Head/hardhats.yml
+++ b/Resources/Prototypes/Entities/Clothing/Head/hardhats.yml
@@ -66,6 +66,11 @@
   - type: Tag
     tags:
     - WhitelistChameleon
+  - type: HideLayerClothing
+    layers:
+      Hair: HEAD
+      HeadTop: HEAD
+      HeadSide: HEAD
 
 - type: entity
   parent: ClothingHeadHatHardhatBase


YoungThugSS14:
I would rather have hair sticking out than people looking completely bald when they have a full head of hair

qrwas:
> I would rather have hair sticking out than people looking completely bald when they have a full head of hair

image
image
This is theoretically a problem with all long hair and helmets, not just the hard hat.
Safety is the best explanation until people can do it through a layer mask

EmoGarbage404:
i understand the logic but i think fully hiding the hair should be reserved for hats that actually cover the entire head. it doesn't make sense for long-haired characters to go bald when putting on a hat that really only covers the very top of their head. we should work on some kind of hair clipping system and use that instead of this solution.

PR #38133, opened 2025-06-07
Content fix for `LocalizedCommands` `Loc` change

Tayrtahn:
Fixes a static method using `Loc` so it works with the new instanced `Loc`.


slarticodefast:
Why is it even static in the first place?

Tayrtahn:
Presumably so the other classes in this file can call it. Seems a little clunky to me, but I dunno.

PJB3005 (changes requested):
Use the `LocalizationManager` field instead so this doesn't require a synchronized engine updat.e

PJB3005 (approved):

PR #38134, opened 2025-06-07
adjreagent now accepts float

Samuka-C:
## About the PR
The command solution:adjreagent now takes a float instead of FixedPoint2

## Why / Balance
The command was useless since it needed a FixedPoint2 and we can't provide that in the command line. now it takes a float and it can actually be used for something.

## Technical details
changed the type of the input data from FixedPoint2 to float in the function call, then, inside the function, made a simple conversion

## Media
with this PR:
image

without this PR:
image

## Requirements

## Breaking changes
command solution:adjreagent now requires a float instead of FixedPoint2 for the amount

**Changelog**
:cl: Samuka
ADMIN:
- fix: the command solution:adjreagent now actually works and no longer needs a "FixedPoint2" for the amount

### DIFF
--- a/Content.Server/Administration/Toolshed/SolutionCommand.cs
+++ b/Content.Server/Administration/Toolshed/SolutionCommand.cs
@@ -44,7 +44,7 @@ public sealed class SolutionCommand : ToolshedCommand
         _solutionContainer ??= GetSys<SharedSolutionContainerSystem>();
 
         // Convert float to FixedPoint2
-        FixedPoint2 amountFixed = FixedPoint2.New(amount);
+        var amountFixed = FixedPoint2.New(amount);
 
         if (amountFixed > 0)
         {


Content.Server/Administration/Toolshed/SolutionCommand.cs
```diff
@@ -38,18 +38,21 @@ public IEnumerable<SolutionRef> Get([PipedArgument] IEnumerable<EntityUid> input
...
             ProtoId<ReagentPrototype> proto,
-            FixedPoint2 amount
+            float amount
         )
     {
         _solutionContainer ??= GetSys<SharedSolutionContainerSystem>();
 
-        if (amount > 0)
+        // Convert float to FixedPoint2
+        FixedPoint2 amountFixed = FixedPoint2.New(amount);
```
Tayrtahn:
```suggestion
        var amountFixed = FixedPoint2.New(amount);
```

Tayrtahn (approved):
Just one minor tweak.

perryprog:
Oh you're my hero; been meaning to make this change myself for ages.

Content.Server/Administration/Toolshed/SolutionCommand.cs
Samuka-C: done

ElectroJr:
A better fix for this would have been to add a FixedPoint2 parser.

PR #38140, opened 2025-06-07
Forbid string literals for `NpcFactionSystem` methods

Tayrtahn:
## About the PR

Adds the `ForbidLiteral` attribute to `NpcFactionSystem` methods, to encourage coders to pass in `ProtoId`s instead of string literals.

Despite the "Forbid" in the name, this actually just counts as a warning and not an error.

## Why / Balance

Magic strings are bad. This encourages cleaner code through the use of validated `ProtoId`s.

## Technical details

Methods in `NpcFactionSystem` that take strings to use as `NpcFactionPrototype`s were given the `ForbidLiteral` attribute.

Existing code that used string literals for these methods was updated to do it properly (as a separate commit for ease of review).

## Requirements

### DIFF
--- a/Content.Shared/NPC/Systems/NpcFactionSystem.cs
+++ b/Content.Shared/NPC/Systems/NpcFactionSystem.cs
@@ -73,7 +73,7 @@ public sealed partial class NpcFactionSystem : EntitySystem
     /// <summary>
     /// Returns whether an entity is a member of a faction.
     /// </summary>
-    public bool IsMember(Entity<NpcFactionMemberComponent?> ent, string faction)
+    public bool IsMember(Entity<NpcFactionMemberComponent?> ent, [ForbidLiteral] string faction)
     {
         if (!Resolve(ent, ref ent.Comp, false))
             return false;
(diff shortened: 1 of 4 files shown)


Content.Shared/NPC/Systems/NpcFactionSystem.cs
```diff
@@ -121,7 +121,7 @@ public void AddFaction(Entity<NpcFactionMemberComponent?> ent, string faction, b
     /// <summary>
     /// Adds this entity to the particular faction.
     /// </summary>
-    public void AddFactions(Entity<NpcFactionMemberComponent?> ent, HashSet<ProtoId<NpcFactionPrototype>> factions, bool dirty = true)
+    public void AddFactions(Entity<NpcFactionMemberComponent?> ent, [ForbidLiteral] HashSet<ProtoId<NpcFactionPrototype>> factions, bool dirty = true)
```
slarticodefast, line 124: does the attribute work if it's a hashset of strings?

slarticodefast:
Funny, I was thinking of doing the same thing after reviewing that other PR.
`SharedToolSystem` would be another candidate for using this attribute by the way.

Content.Shared/NPC/Systems/NpcFactionSystem.cs
Tayrtahn: Yep, it works for any collection type!

It will check all members of a passed collection expression to make sure that they're not literals.

Tayrtahn:
>`SharedToolSystem` would be another candidate for using this attribute by the way.

Good call! https://github.com/space-wizards/space-station-14/pull/38250

PR #31442, opened 2024-08-25
Add text highlighting

VitusVeit:
## About the PR

This PR adds text highlighting to the ChatBox, the user can now input a series of comma-separated words that will be highlighted on the chat box and bubbles.

## Why / Balance

Already present in many SS13 servers, this feature greatly improves the readability of the comms and helps departments and individuals know when they are called/needed on the radio.

## Technical details

This PR is based on #30092, it uses the newly added function "InjectTagAroundString" to change the colors from a list of keywords. The words to be highlighted can be inserted through a LineEdit in the filter panel and will be saved on the ChatBox itself (it is currently not being saved when exiting the game).
For the highlight color, I tried to choose one that could be recognized easily from any channel including OOC and Dead Chat.

## Requirements

- [x] I have read and I am following the [Pull Request Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html). I understand that not doing so may get my pr closed at maintainer’s discretion
- [x] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl: TheCactus
- add: Added text highlighting.

PR #38144, opened 2025-06-08
Fix small issues with Text Highlighting

VitusVeit:
## About the PR

This is a small PR to fix some mistakes left by the [text highlighting PR](https://github.com/space-wizards/space-station-14/pull/31442).

## Technical details

Other than grammatical errors, this PR also **changes the regex in charge of highlighting user-excluding keywords** (keywords that start with a @). 

The old regex didn't count for cases where the keyword in question would appear without a [Name] tag present (eg. station announcements), so the keyword in that case wouldn't get highlighted.

The new regex checks if there are some double quotes before the keyword or if there isn't a [Name] tag.

## Media

Before:

After:

## Requirements

### DIFF
--- a/Resources/Locale/en-US/escape-menu/ui/options-menu.ftl
+++ b/Resources/Locale/en-US/escape-menu/ui/options-menu.ftl
@@ -50,8 +50,8 @@ ui-options-interface-label = Interface
 
 
 ui-options-auto-fill-highlights = Auto-fill the highlights with the character's information
-ui-options-highlights-color = Highlighs color:
-ui-options-highlights-color-example = This is an highlighted text!
+ui-options-highlights-color = Highlights color:
+ui-options-highlights-color-example = This is a highlighted text!
 ui-options-show-held-item = Show held item next to cursor
 ui-options-show-combat-mode-indicators = Show combat mode indicators with cursor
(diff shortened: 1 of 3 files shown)


Resources/Locale/en-US/escape-menu/ui/options-menu.ftl
```diff
@@ -50,8 +50,8 @@ ui-options-interface-label = Interface
 
 
 ui-options-auto-fill-highlights = Auto-fill the highlights with the character's information
-ui-options-highlights-color = Highlighs color:
-ui-options-highlights-color-example = This is an highlighted text!
+ui-options-highlights-color = Highlights color:
+ui-options-highlights-color-example = This is a highlighted text!
 ui-options-show-held-item = Show held item next to cursor
```
VerinSenpai, line 55: Could just be me but I feel like it'd be better to have it say "This is highlighted text".

VerinSenpai: Or revise it a little bit to say this is a highlighted message

VitusVeit: Grammar is definitely not my strong point; "This is highlighted text." sounds good though, I'll add it with the next commit.

PR #38146, opened 2025-06-08
Fix the sensor monitoring console forcing a GC every 3 seconds

ArtisticRoomba:
## About the PR
Fixes the sensor monitoring console forcing a garbage collect every 3 seconds.

## Why / Balance
Holy shit I was pissed when I was doing TEG testing.
## Technical details
Previously, every single draw call we were creating a new Vector2 array with 25000 elements.
This allocated something like 130-145 KiB per frame and caused my game to do a GC every 3 seconds.

Instead, we do the following:
- We first allocate a new vertices array in the constructor.
	- Because we know how big `_samples` is, and that each line segment needs 6 vertices to form 2 triangles, we can simply make an array that's the size that we need.
- This array is populated like normal.
	- If the array is too small, it is simply doubled (however this shouldn't happen).
- For the next draw call, `_verticies` remains allocated and the data is simply overwritten as the draw call logic will start at the beginning of the array. Any extra data beyond the amount of elements in `countVtx` is simply ignored when we go to draw primitives (as it only draws up to `countVtx`).

Allocations per frame are about 20-23 KiB without looking at the console, ~50 KiB while looking at the console.
And it doesn't do a SOH garbage collect every 3 seconds.

## Media
Before the implementation
rider64_RpEgwnLor6

After
rider64_RxCqODrUwT
rider64_67FhJtZiC7

## Requirements

## Breaking changes
N/A

**Changelog**
no CL no fun

### DIFF
--- a/Content.Client/SensorMonitoring/SensorMonitoringWindow.xaml.cs
+++ b/Content.Client/SensorMonitoring/SensorMonitoringWindow.xaml.cs
@@ -198,6 +198,17 @@ public sealed partial class SensorMonitoringWindow : FancyWindow, IComputerWindo
         private readonly TimeSpan _curTime;
         private readonly float _maxY;
 
+        /// <summary>
+        /// A reusable array of vertices for drawing the graph.
+        /// </summary>
+        private Vector2[] _vertices;
+
+        /// <summary>
+        /// A precomputed capacity for the vertex array. We do this to avoid resizing it too often.
+        /// </summary>
+        /// <remarks>Previously we were allocating a Vector2[25000] every single draw call. :)</remarks>
+        private int _vertexCapacity;
+
         public GraphView(Queue<SensorSample> samples, TimeSpan startTime, TimeSpan curTime, float maxY)
         {
             _samples = samples;
@@ -205,6 +216,10 @@ public sealed partial class SensorMonitoringWindow : FancyWindow, IComputerWindo
             _curTime = curTime;
             _maxY = maxY;
             RectClipContent = true;
+
+            // Pre-estimate vertex capacity based on the sample count we've been given.
+            _vertexCapacity = Math.Max(6 * (samples.Count - 1), 0); // Ensure no negative capacity
+            _vertices = new Vector2[_vertexCapacity];
         }
 
         protected override void Draw(DrawingHandleScreen handle)
@@ -212,9 +227,6 @@ public sealed partial class SensorMonitoringWindow : FancyWindow, IComputerWindo
             base.Draw(handle);
 
             var window = (float) (_curTime - _startTime).TotalSeconds;
-
-            // TODO: omg this is terrible don't fucking hardcode this size to something uncached huge omfg.
-            var vertices = new Vector2[25000];
             var countVtx = 0;
 
             var lastPoint = new Vector2(float.NaN, float.NaN);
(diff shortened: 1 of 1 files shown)


perryprog:
One step closer to sensor monitoring console real...

Content.Client/SensorMonitoring/SensorMonitoringWindow.xaml.cs
```diff
@@ -198,23 +198,35 @@ private sealed class GraphView : Control
...
+        /// <summary>
+        /// A reusable array of vertices for drawing the graph.
+        /// </summary>
+        private Vector2[] _vertices;
+
+        /// <summary>
+        /// A precomputed capacity for the vertex array. We do this to avoid resizing it too often.
+        /// </summary>
+        /// <remarks>Previously we were allocating a Vector2[25000] every single draw call. :)</remarks>
+        private int _vertexCapacity;
```
PJB3005: You don't need to store this separately, just use `_vertices.Length`.

Content.Client/SensorMonitoring/SensorMonitoringWindow.xaml.cs
```diff
@@ -198,23 +198,35 @@ private sealed class GraphView : Control
         private readonly TimeSpan _curTime;
         private readonly float _maxY;
 
+        /// <summary>
+        /// A reusable array of vertices for drawing the graph.
+        /// </summary>
+        private Vector2[] _vertices;
```
PJB3005: Share this between all graphs of the console somehow.

ArtisticRoomba:
New changes feature a static Vector2 that every graphview just shares. Expanded automatically if necessary.

Allocs per frame reduced from ~50 KiB to ~40 KiB.

Before looking at console
rider64_hgAY3TX4K4

After looking at console
rider64_Z0GkUOjmW2

Console rendering fine
Content Client_TDmu1GIaoq

Content.Client/SensorMonitoring/SensorMonitoringWindow.xaml.cs
```diff
@@ -193,6 +193,11 @@ private sealed class SensorStream
 
     private sealed class GraphView : Control
     {
+        // Shared static buffer for all instances of GraphView to reduce allocations.
+        // Need to use this because ArrayPool violates sandbox.
+        private static Vector2[] _sharedVertices = [];
+        private static readonly object SharedVerticesLock = new();
```
PJB3005: Do not use static for this.

ArtisticRoomba: Unsure as to what I would do, PJB. I'm not experienced in C# at all.

Only thing I can think of that would allow the `Vector2[]` to be shared across instances is to move the `Vector2[]` under the `SensorMonitoringWindow` class and have that handle the sharing logic. You think that's good?

I've been trying to contain this inside of the `GraphView` control class as I felt like it would be cool to move this into engine.

PJB3005: > Only thing I can think of that would allow the Vector2[] to be shared across instances is to move the Vector2[] under the SensorMonitoringWindow class and have that handle the sharing logic. You think that's good?

Do that.

> I've been trying to contain this inside of the GraphView control class as I felt like it would be cool to move this into engine.

This isn't a useful goal.

Content.Client/SensorMonitoring/SensorMonitoringWindow.xaml.cs
```diff
@@ -197,53 +220,65 @@ private sealed class GraphView : Control
         private readonly TimeSpan _startTime;
         private readonly TimeSpan _curTime;
         private readonly float _maxY;
-
-        public GraphView(Queue<SensorSample> samples, TimeSpan startTime, TimeSpan curTime, float maxY)
+        private readonly SensorMonitoringWindow _parentWindow;
+        private readonly object _bufferLock = new();
```
PJB3005: This lock isn't necessary.

PJB3005 (changes requested):

PJB3005 (approved):

PR #38155, opened 2025-06-08
Make RunVerbAs take and return EntityUids

perryprog:
## About the PR

Made the toolshed command RunVerbAs take EntityIds instead of NetEntities.

## Why / Balance

As far as I can tell there's no way to pass around a NetEntity in toolshed, so this command wasn't usable.

## Media

Running verb with separate runner and targets works:

image

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- fix: The runverbas command should now work better with other toolshed commands.

### DIFF
--- a/Content.Server/Toolshed/Commands/Verbs/RunVerbAsCommand.cs
+++ b/Content.Server/Toolshed/Commands/Verbs/RunVerbAsCommand.cs
@@ -14,9 +14,9 @@ public sealed class RunVerbAsCommand : ToolshedCommand
     private SharedVerbSystem? _verb;
 
     [CommandImplementation]
-    public IEnumerable<NetEntity> RunVerbAs(
+    public IEnumerable<EntityUid> RunVerbAs(
             IInvocationContext ctx,
-            [PipedArgument] IEnumerable<NetEntity> input,
+            [PipedArgument] IEnumerable<EntityUid> input,
             EntityUid runner,
             string verb
         )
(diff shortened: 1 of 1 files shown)


perryprog:
If I recall correctly that’s just how toolshed commands show the autocomplete for EntityUids. You can check with some other command that takes one like `ent`.

https://github.com/space-wizards/RobustToolbox/blob/56eda3ea92913e60c27aa029ba7877c8559dfe9d/Robust.Shared/Toolshed/TypeParsers/EntityTypeParser.cs#L57

beck-thompson (approved):
Huh yeah that's odd but your right 🤔

I looked in the git history, couldn't find any reason why these were net entities so I think it was just a mistake

ScarKy0:
Wow I never even knew this command existed

VerinSenpai:
wait isn't there a non-toolshed command that does this too?

perryprog:
Yup, /invokeverb. That one works fine.

VerinSenpai:
Is it necessary for both to exist? I'm not sure what the difference is between toolshed commands and console commands besides how they're written.

perryprog:
I think it's good to have both. There's a bit of a learning curve to toolshed's syntax (compare `ent 123 runverbas { ent 456 } "Pull"` with `invokeverb 123 456 Pull`), even though it's certainly more powerful (and much, much, much easier to implement). It also has a few gotchas—it's server-side only, and you have to always (if not +HOST) prefix the commands with `> ` which if you don't it says you don't have permissions (???).

It can also be more limited in its autocomplete—for example, there's no ResPath autocomplete yet while there is (via a helper) for console commands. (The helper can't be reused for Toolshed since the path for toolshed needs to be wrapped in quotes, and I don't think there's an easy way to deal with that.)

PR #38156, opened 2025-06-08
Updated hat descriptions

Hitlinemoss:
## About the PR

Updated the descriptions of various hats.

## Why / Balance

Some of these descriptions were awkward or grammatically incorrect and it bothered me.

## Technical details

Changes to Resources\Prototypes\Entities\Clothing\Head\hats.yml and Resources\Prototypes\Entities\Clothing\Head\soft.yml

## Media

N/A, since this is purely a change to description text.

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: Various hats have had their descriptions updated.

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Head/hats.yml
+++ b/Resources/Prototypes/Entities/Clothing/Head/hats.yml
@@ -13,7 +13,7 @@
   parent: ClothingHeadBase
   id: ClothingHeadHatBeret
   name: beret
-  description: A beret, an artists favorite headwear.
+  description: An artist's favorite headwear.
   components:
   - type: Sprite
     sprite: Clothing/Head/Hats/beret.rsi
@@ -30,7 +30,7 @@
   parent: ClothingHeadBase
   id: ClothingHeadHatBeretFrench
   name: french beret
-  description: A French beret, "vive la France".
+  description: A French beret. "Vive la France!"
   components:
   - type: Sprite
     sprite: Clothing/Head/Hats/beret_french.rsi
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/Entities/Clothing/Head/hats.yml
```diff
@@ -2,7 +2,7 @@
   parent: ClothingHeadBase
   id: ClothingHeadHatBeaverHat
   name: beaver hat
-  description: 'Gentlemen?'
+  description: "Gentlemen?"
```
slarticodefast: Are quotation marks even shown in the description without escaping them? Pretty sure they are just interpreted as a string. Same for the other cases.

Hitlinemoss: Unfortunately not. I tried to get around this in a few ways (double quotation marks, apostrophes, etc) but they either don't show up or they cause the .yml file to not compile properly.

As I'm typing this reply now it occurs to me that brackets might work for this, so I'll try that quickly.

slarticodefast: Can't you just escape the quotation marks?
`\"Gentlemen?\"`

slarticodefast: Or you might need to escape them inside a second set of quotation marks
grafik

beck-thompson (approved):
Yeah I think these are better!

slarticodefast (approved):

PR #38157, opened 2025-06-08
Add technology to tech disk names

eoineoineoin:
## About the PR

Minor QOL fix - adds technology name to tech disks.

## Why / Balance

Technology disks are a bit of a pain to use right now; you tend to make a bunch of them all at once, then have to detailed-examine each of the disks to find what technology they contain, then pickup and move the disk into an "interesting" or "not interesting" pile.

Doing this a bunch is pretty tedious and easy to make a mistake requiring you to sort through the whole pile again. By adding the technology to the disk name, you can now right-click a pile and much more easily scan through it to pick up the interesting ones.

## Technical details

Nothing terribly exciting; hooking up some already-extant systems.

## Media

Before:

After:

(In this screenshot, I've also used a hand labeler to add a label to a disk, just to test that this PR doesn't interfere with that system)  

## Requirements
Real

## Breaking changes

**Changelog**
:cl:
- add: Technology disks will now show the tech they contain in their name.

### DIFF
--- a/Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
+++ b/Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
@@ -1,6 +1,7 @@
 using Content.Shared.Examine;
 using Content.Shared.Interaction;
 using Content.Shared.Lathe;
+using Content.Shared.NameModifier.EntitySystems;
 using Content.Shared.Popups;
 using Content.Shared.Random.Helpers;
 using Content.Shared.Research.Components;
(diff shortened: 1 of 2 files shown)


beck-thompson (approved):
I never even realized these were for specific techs :sob: 

Def a good change!

Content.Shared/Research/TechnologyDisk/Systems/TechnologyDiskSystem.cs
```diff
@@ -90,4 +94,16 @@ private void OnExamine(Entity<TechnologyDiskComponent> ent, ref ExaminedEvent ar
         }
         args.PushMarkup(message);
     }
+
+    private void OnRefreshNameModifiers(Entity<TechnologyDiskComponent> entity, ref RefreshNameModifiersEvent args)
+    {
+        if (entity.Comp.Recipes != null)
+        {
+            foreach (var recipe in entity.Comp.Recipes)
```
slarticodefast, line 102: Is the null check needed here?
Wouldn't the iterator just do nothing if it was null?

eoineoineoin: It's still necessary; foreach needs a real object to call GetEnumerator() on and will raise a NullReferenceException if entity.Comp.Recipes is null.

PR #38158, opened 2025-06-08
Station size armory fills

Nox38:
## About the PR

Added entitytables to automatically fill gun safes for Kammerers, Lecters, Drozds, and Lasers. The number of guns inserted is based on total playercount.
Changed security suit storage units to fill with two hardsuits, a O2 and N2 tank, and a security jetpack.

## Why / Balance

This allows the number of guns in a map's armory to be balanced to map/security team size without the current method of manually adding more guns in the map, allowing balancing changes to be made fully through yaml. 
The current values are as follows, with the number of guns listed being the number of each type of gun, except lasers which following tradition have one extra.
LowPop: 0-30 players, standard fills (2 guns)
MidPop: 31-50 players, 3 guns
HighPop: 51-75 players, 4 guns
MaxPop: 76+ players, 5 guns

Following the merging of this PR I will remove the manually fills from each station armory.
## Technical details

Simply added new prototypes which are parented to the base gun safes.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Lockers/security.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/security.yml
@@ -181,26 +181,24 @@
       tableId: FillRifleLecterLowPop
       conditions:
       - !type:PlayerCountCondition
-        min: 0
-        max: 29
+        max: 30
     - !type:NestedSelector
       tableId: FillRifleLecterMidPop
       conditions:
       - !type:PlayerCountCondition
-        min: 30
-        max: 49
+        min: 31
+        max: 50
     - !type:NestedSelector
       tableId: FillRifleLecterHighPop
       conditions:
       - !type:PlayerCountCondition
-        min: 50
-        max: 69
+        min: 51
+        max: 75
     - !type:NestedSelector
       tableId: FillRifleLecterMaxPop
       conditions:
       - !type:PlayerCountCondition
-        min: 70
-        max: 80
+        min: 76
 
 - type: entityTable
   id: FillRifleLecterLowPop
@@ -212,7 +210,7 @@
       amount: 4
 
 - type: entityTable
-  id: FillRifleLecterMedPop
+  id: FillRifleLecterMidPop
   table: !type:AllSelector
     children:
     - id: WeaponRifleLecter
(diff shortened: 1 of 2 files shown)


LaCumbiaDelCoronavirus:
Less convoluted to just put the number of guns than mark it as lowpop midpop etc., also it would be cool if the safe's descriptions reflected how many guns they're supposed to start with so it's easier to keep track of inventory

Cojoke-dot:
Why is the lecture have 1 more gun than the others? It should probably be consistent with the number of guns to the population


What are these numbers based on? Why 4 on max pop and 3 on mid? Iirc its 10 max secoff slots on maxpop stations right? It's regularly 3 of each gun on every stations with a random amount of lasers, enforcers, and fun guns thrown around.

Nox38:
> Less convoluted to just put the number of guns than mark it as lowpop midpop etc., also it would be cool if the safe's descriptions reflected how many guns they're supposed to start with so it's easier to keep track of inventory

Ill add the descriptions, the number of guns is spesifically not part of the prototype names so they can be changed by non-mappers.

Nox38:
> Why is the lecture have 1 more gun than the others? It should probably be consistent with the number of guns to the population
> 
> What are these numbers based on? Why 4 on max pop and 3 on mid? Iirc its 10 max secoff slots on maxpop stations right? It's regularly 3 of each gun on every stations with a random amount of lasers, enforcers, and fun guns thrown around.

There are more Lecters because Lecters are by far the weakest sec longarm (laser doesn’t count). These numbers are based on my extensive theorycrafting and then testing on the armory rebalance PRs I did on every single station.
These are starting values, and would be configurable by any yaml’er rather than only the maptainer.

Radezolid:
Wouldn't it be better to use `EntityTableContainerFill` and just either:
- Use rolls on each tier, reducing copy paste.
- Use #36819

Nox38:
@Radezolid Im a simple mapper. If you can show me how to get those working, absolutely it’s a better solution. As is, this is intended as a stopgap until someone programs it to automatically fill the safes based on station pop.

Radezolid:
This is a example of the use of rolls and a entity tables.
https://github.com/space-wizards/space-station-14/blob/264573b9e1ad6fd323aacbb59ebd8a5f4a131feb/Resources/Prototypes/Catalog/Fills/Crates/fun.yml#L46-L87
What you would need is make a base lecter gun safe table, and then scale it with the conditions setting up min and maxes on when to spawn one. Like on this example:
https://github.com/space-wizards/space-station-14/blob/264573b9e1ad6fd323aacbb59ebd8a5f4a131feb/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml#L56-L59
Instead of an ID you would need to place a `NestedSelector` there

Nox38:
> Reference in new

I did it! Thanks for the suggestion!

Nox38:
Massive merge conflicts and little interest in this PR, I guess we can keep doing manual fills...

PR #38167, opened 2025-06-09
SkinColor Hues Minimum Saturation

poklj:
## About the PR

- add Saturation axis minimum to the HSV vector

- Modify the frontend to actively use the valid colorscape for Hues and TintedHues

## Why / Balance

0,0,0 skincolor characters using Hue would look like Locked characters, able to hide in fully shaded areas.
Simply clamping them to a grey gives them a better and more discernable profile without fully removing black scales/moth fluff as an option
## Technical details

Added a 0.1f MathMax to the Saturation (y) axis on the HSV for hues, as well as actually made Hues/TintedHues use that Clamping function during character creation
## Media

Before:
000-noclamp

After:
Content Client_2025-06-09_02-20-51

## Requirements

## Breaking changes

**Changelog**

:cl: Kickguy223
- fix: added a Minimum saturation to the skin skin color of Arachnids, Diona, Moth, Reptilians, and Slimes.

### DIFF
--- a/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
+++ b/Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
@@ -1113,8 +1113,9 @@ namespace Content.Client.Lobby.UI
                         RgbSkinColorContainer.Visible = true;
                     }
 
-                    Markings.CurrentSkinColor = _rgbSkinColorSelector.Color;
-                    Profile = Profile.WithCharacterAppearance(Profile.Appearance.WithSkinColor(_rgbSkinColorSelector.Color));
+                    var color = SkinColor.MakeHueValid(_rgbSkinColorSelector.Color);
+                    Markings.CurrentSkinColor = color;
+                    Profile = Profile.WithCharacterAppearance(Profile.Appearance.WithSkinColor(color));
                     break;
                 }
                 case HumanoidSkinColor.TintedHues:
@@ -1333,7 +1334,6 @@ namespace Content.Client.Lobby.UI
                     }
 
                     Skin.Value = SkinColor.HumanSkinToneFromColor(Profile.Appearance.SkinColor);
-
                     break;
                 }
                 case HumanoidSkinColor.Hues:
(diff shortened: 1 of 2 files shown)


ArtisticRoomba:
Not-Unlocked-Yet in shambles

poklj:
> Not-Unlocked-Yet in shambles

I hope you can forgive my transgression against your silly goober.

K-Dynamic:
> Not-Unlocked-Yet in shambles

I'm more worried about Watcher

By the way, this wasn't how I made Not-Unlocked-Yet - setting skin colour to 0,0,0 defaults to grey skin when saving the character file
image

What I actually did was make all the markings 0,0,0
image

Pumkin69:
If its possible to make it darker i would. This isn't black its dark grey and black animals exist in nature so i don't see why it should be so bright.

TrixxedHeart:
> If its possible to make it darker i would. This isn't black its dark grey and black animals exist in nature so i don't see why it should be so bright.

I think maybe slightly darker would be good if possible. Although do imagine what it'll look like in-game with lighting, it'll probably be much darker than it looks in customization.

Entvari:
While you're at it can you unclamp the skintone for dwarfs so i can larp as a proud green maints goblin

Errant-4 (commented):
The Saturation minimum does not seem related to what you are purportedly fixing, here? 
As far as I can see, it is only 0 Value/Lightness that makes someone's color look vantablack, and that was already restricted on the Hues colorprofile in 2024 January. Not allowing 0 saturation only makes snow white albino characters impossible

poklj:
> The Saturation minimum does not seem related to what you are purportedly fixing, here? As far as I can see, it is only 0 Value/Lightness that makes someone's color look vantablack, and that was already restricted on the Hues colorprofile in 2024 January. Not allowing 0 saturation only makes snow white albino characters impossible

honestly, i think i might modify this PR to not do any clamping and instead focus on the Ingame character editor that is done with the PR so it's more obvious what you're getting when you go to save, As the Editor never attempted to validate the colors before pressing save, only when save is pressed.

poklj:
@Errant-4 I've now stripped out the Color value clamping portion of this PR and it is now just the frontend calls to the validators as both VoxFeathers and HumanSkinTones do

Errant-4 (approved):

PR #38179, opened 2025-06-09
Revert "Add trailing whitespace check workflow"

slarticodefast:
The script used to find the files that have been changed fails sometimes for some reason.
Needs another solution, it's causing too many false positives, confusing contributors.
Maybe a pre-commit hook or something, but we would also have to smite web edits somehow.


Tayrtahn (approved):

deltanedas:
> but we would also have to smite web edits somehow

just autoban webedits or have a ci running the check for prs if they butchered the hook

PR #38180, opened 2025-06-09
hi-viz vest laser reflect

nabegator:
## About the PR

hi-viz vest now has a 30% chance to reflect lasers
## Why / Balance

Cool
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: hi-viz vest now has a 30% chance to reflect lasers

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/OuterClothing/vests.yml
+++ b/Resources/Prototypes/Entities/Clothing/OuterClothing/vests.yml
@@ -28,6 +28,11 @@
       - state: equipped-OUTERCLOTHING-unshaded
         shader: shaded
         #same trick here
+  - type: Reflect
+    reflectProb: 0.3
+    reflects:
+      - Energy
+    reflectingInHands: false
   - type: Tag
     tags:
     - HiViz


K-Dynamic:
Funny but please no

slarticodefast:
Funny idea, but I don't think this will work out balance-wise. Everyone will just grab them for powergaming reasons.

LaCumbiaDelCoronavirus:
> Funny idea, but I don't think this will work out balance-wise. Everyone will just grab them for powergaming reasons.

(they have 0 armor or cold prot or heat prot)

beck-thompson:
30% is much too high but if it was very low like 2% it would be funny and on one would bother power gaming with it

kosticia:
:trollface:

slarticodefast:
> 30% is much too high but if it was very low like 2% it would be funny and on one would bother power gaming with it

Yeah, something like 1 or 2 percent as a gimmick would be fine. But such an item should not have an actual impact on combat.

nabegator:
ok

nabegator:
5% will not affect combat

Resources/Prototypes/Entities/Clothing/OuterClothing/vests.yml
```diff
@@ -28,6 +28,11 @@
       - state: equipped-OUTERCLOTHING-unshaded
         shader: shaded
         #same trick here
+  - type: Reflect
+    reflectProb: 0.05
+    reflects:
+      - Energy
```
slarticodefast:
```suggestion
    reflectProb: 0.02
    reflects:
    - Energy
```
I brought it up in maintainer chat again, and this should be more an easter egg rather than a regular thing, so I'm changing it to 2% like previously suggested.
While the 5% chance can be debated, we would rather avoid microbalancing these things.
And it's more fun when it happens if it is rare, similar to the wilhelm scream.

PR #37756, opened 2025-05-23
Bar sign AlignTile version and fixtures for bar signs

AsnDen:
## About the PR
Now there is version of bar sign that takes full 2 tiles. I've also changed its fixtures.

## Why / Balance
#6476 

## Media

image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- add: Tile-aligned bar sign version was added.
- fix: Bar signs now have proper collision.

PR #38183, opened 2025-06-09
Remove non-aligned barsigns

slarticodefast:
## About the PR
Discussed in the maintainer meeting, and this was voted on. I think. The poll options were not entirely clearly formulated, so please double check this before merging.

Follow-up to https://github.com/space-wizards/space-station-14/pull/37756
This PR removes the non-aligned bar signs and migrates the proto Ids. As a side effect all mapped bar signs will move to the right by right by half a tile, which will need to be fixed by mappers in some cases.

## Why / Balance
The original non-aligned bar sign was likely unintented.
These are the only item in the game that snaps to the grid with an offset of half a tile, and keeping them would be inconsistent.

## Technical details
Revert the PR, but keep the fixture and sprite offset.
Since the original protoIds are now aligned, we migrate the aligned ids over to them.

## Media
old bar sign from the map viewer

in-game test
grafik

## Requirements

## Breaking changes
Bar signs are now properly aligned with the grid instead of being offset by half a tile. This means all existing bar sign on maps will move to the right by half a tile and may require mapping changes.


slarticodefast:
Oh, forgot I have to target staging. Don't merge yet, I'll rebase.

slarticodefast:
Oh no.
Rebasing will cause conflicts between stable and master because migration.yml has new entries on master.

Edit: Myra said we can resolve them.

ArtisticRoomba:
This will be very annoying for mappers to fix in some instances and I don't know why we're doing this. A lot of places have an odd tile symmetry (center of 3 instead of 4) so this change will just bulldoze it. Also consider downstream as well.

slarticodefast:
Fair point, but I think this was the outcome of the poll?
Although it's not technically a revert like the poll option says, and the rest of the sentence was cut off, so some maintainers might have misunderstood it. Maybe we should clarify it, but I also don't want the release to be blocked by this.
grafik

slarticodefast:
And if we look at ss13 the original bar signs are tile aligned, as are all other existing structures in ss14. So bar signs being the only one not aligned is a weird inconsistency.

Here a part of ss13's Meta station.
grafik

Emisse:
a good way to do this would be to make it snap to half tiles

slarticodefast:
> a good way to do this would be to make it snap to half tiles

Do we have anything else in the game that does this?
That would also mean you could no longer re-anchor it in case it gets unanchored.

IamVelcroboy:
>... as are all other existing structures in ss14. So bar signs being the only one not aligned is a weird inconsistency.

Check TVs. That might be the only other one

slarticodefast:
I made an alternative PR that simply reverts the original bar sign change and gives us more time to come up with a proper solution without further blocking the release
https://github.com/space-wizards/space-station-14/pull/38185

juliangiebel (dismissed):

s27jgould:
perhaps, just an idea, set up a 2nd prototype that is the tile aligned barsign? (although that may not be liked by some maintainers because bloat or whatever, just a way that mappers can have their cake and eat it too)

Issue #38153, opened 2025-06-08
Construction menu shouldn't have (some) duplicate entries


perryprog:
## Description

Since #32339, suffixes have been killed from the crafting menu. This means there are now non-differentiated entries in the construction menu that require either careful examination of the sprite (chain link fence), or literal guessing (signal button versus directional signal button).

**Reproduction**

1. Build a signal button on a wall.
2. Guess wrong and build the non-directional variant.
3. Tider can now open the blast door to TEG from maints because non-directional buttons are weird.

**Screenshots**

![Image](https://github.com/user-attachments/assets/062369d2-0fa3-4e91-bf65-10451913c897)

**Additional context**

Should arguably have a test to check we don't have any identical labels in the crafting menu, but I don't know if this is a great idea. 

At the least, the most annoying ones are probably just these three:
- Signal button (directional versus not)
- Signal switch (directional versus not)
- Plastic flaps (opaque versus clear)

K-Dynamic:
Surely we could localise the suffixes and append the crafting menu entries with them

perryprog:
Yeah, that would work—I think it'd probably want to be a separate identifier from spawn menu suffixes, as those aren't meant to be player-facing. Alternatively, we add a localized constriction menu name field and specify a fuller name there as needed?

Tayrtahn:
Suffixes are not meant to be player-facing, so we can't use those.

However, it looks like `ConstructionPrototype` already has a `name` field that can be used to override the entity name for that specific recipe. So this should just be a matter of adding those to any recipes with the same name.

Tayrtahn:
I just submitted a PR that demonstrates how to give recipes unique names: https://github.com/space-wizards/space-station-14/pull/38182

It's a simple matter of adding strings to FTL files and one-line changes to YAML. If anyone wants to help out by using the same technique to fix other duplicate recipes, that would be greatly appreciated!

perryprog:
Amazing; thanks for your work on this. I'll get started with the remaining entries as listed by the associated PR's test failure.

Tayrtahn:
Currently failing the test:
```  
  1)   Multiple construction prototypes have the display name 'chain link fence': FenceMetalCorner, FenceMetal, FenceMetalEnd
  2)   Multiple construction prototypes have the display name 'railing': RailingCornerSmall, RailingRound, RailingCorner, Railing
  3)   Multiple construction prototypes have the display name 'gas filter': GasFilterFlipped, GasFilter
[...]

PR #38187, opened 2025-06-09
Add test that all construction recipes have unique display names

Tayrtahn:
## About the PR

Adds an integration test that makes sure that every entry in the construction window has a unique name.

**This test is going to fail on master right now! Do not merge this until https://github.com/space-wizards/space-station-14/issues/38153 is resolved!**

## Why / Balance

https://github.com/space-wizards/space-station-14/issues/38153 suggested a test for this purpose, and I thought it seemed like a good idea. Having a test will help to find the instances that need to be fixed.

Having a unique name for each entry will make it easier to select the correct recipe, clear up ambiguity in cases where entries also have identical sprites, and make it easier to search for recipes.

## Technical details

- Adds a method to Client's `ConstructionSystem` that gets the display name for a given `ConstructionPrototype` ID.
- The test itself enumerates all `ConstructionPrototype`s and uses the new method to get their display names. A dictionary uses the display names as keys and keeps a list of every `ConstructionPrototype` with the same display name. After the all prototypes are processed, the dictionary is checked and any entries that have more than one `ConstructionPrototype` fail the test. The error message displays the IDs of all `ConstructionPrototypes` with the same display name:
```
1)   Multiple construction prototypes have the display name 'chain link fence': FenceMetalCorner, FenceMetal, FenceMetalEnd
Assert.That(recipeList, Has.Count.EqualTo(1))
  Expected: property Count equal to 1
  But was:  3
```

## Requirements

## Breaking changes

Construction recipes are now required to have unique display names in the construction window. By default, recipes use the entity name of the resulting entity, but `name` can be specified for in `ConstructionPrototype` YAML files to override the display name.

### DIFF
--- a/Content.IntegrationTests/Tests/Construction/ConstructionPrototypeTest.cs
+++ b/Content.IntegrationTests/Tests/Construction/ConstructionPrototypeTest.cs
@@ -188,8 +188,26 @@ namespace Content.IntegrationTests.Tests.Construction
                 if (!constructionSys.TryGetRecipeName(recipe.ID, out var name))
                     continue;
 
-                // Track this use
                 var users = nameUsers.GetOrNew(name);
+
+                // Don't track this if it's a mirror of one we're already tracking
+                if (recipe.Mirror is { } mirrorProto && users.Contains(mirrorProto))
+                    continue;
+
+                // Don't track this if it's an alternative version of one we're already tracking
+                var isAlternative = false;
+                foreach (var alterativeProto in recipe.AlternativePrototypes)
+                {
+                    if (users.Contains(alterativeProto))
+                    {
+                        isAlternative = true;
+                        break;
+                    }
+                }
+                if (isAlternative)
+                    continue;
+
+                // Track this use
                 users.Add(recipe);
             }
 
--- a/Content.IntegrationTests/Tests/Construction/ConstructionPrototypeTest.cs
+++ b/Content.IntegrationTests/Tests/Construction/ConstructionPrototypeTest.cs
@@ -1,10 +1,13 @@
+using System.Collections.Generic;
 using System.Numerics;
+using Content.Client.Construction;
 using Content.Server.Construction.Components;
 using Content.Shared.Construction.Prototypes;
 using Robust.Shared.GameObjects;
 using Robust.Shared.Map;
 using Robust.Shared.Maths;
 using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
 
 namespace Content.IntegrationTests.Tests.Construction
 {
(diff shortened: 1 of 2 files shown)


perryprog:
Do we want to require this for entities that are hidden from the construction menu, like GasFilter versus GasFilterFlipped?

Tayrtahn:
> Do we want to require this for entities that are hidden from the construction menu, like GasFilter versus GasFilterFlipped?

Probably not. I'll change the test to ignore those.

Content.IntegrationTests/Tests/Construction/ConstructionPrototypeTest.cs
```diff
@@ -188,8 +188,26 @@ public async Task TestUniqueNames()
...
+                foreach (var alterativeProto in recipe.AlternativePrototypes)
+                {
+                    if (users.Contains(alterativeProto))
+                    {
+                        isAlternative = true;
+                        break;
+                    }
+                }
+                if (isAlternative)
+                    continue;
```
perryprog: It doesn't _really_ matter, but it might be good to check before this (or even before the mirror test) if the prototype has Hide set, and throw it out then. Otherwise the "wrong" prototype can be shown as failing—for example, the current test fail is showing GasPipeHalfAlt2 as "the one" with the gas pipe display name.

Tayrtahn: Hah, somehow I completely missed the Hide property! That's a much better way of doing it!

Content.IntegrationTests/Tests/Construction/ConstructionPrototypeTest.cs
```diff
@@ -160,5 +163,53 @@ await server.WaitAssertion(() =>
...
+                // Track this use
+                users.Add(recipe);
+            }
+
+            Assert.Multiple(() =>
+            {
+                foreach (var (displayName, recipeList) in nameUsers)
+                {
+                    // Make sure that each name only has one use
+                    Assert.That(recipeList, Has.Count.EqualTo(1),
```
perryprog: I think this should be AtLeast, or a change to the test's logic should be made—currently this will fail if all construction graph prototypes of a given name are hidden, which is currently just the case for the two variants of APC net light switches.

Tayrtahn: Ah, good catch. Thanks!

perryprog: ~~Except for where I said AtLeast instead of AtMost.~~

perryprog:
I'm wondering if we should have this test exclude construction prototypes that have identical resulting entity prototypes, in order to get around the issue of the multiple reinforced X glass recipes. It's still not great UI wise, and there really should be a better way to handle multiple recipes for the same item, but from what I can tell that's a good bit of effort.

(Though one thing that could be done is for the construction UI itself to automatically merge entries with duplicate recipes, but this doesn't help the fact that having all of SheetRPGlass, SheetRPGlass0, and SheetRPGlass1 exist as separate prototypes is pretty silly.)

PR #36236, opened 2025-03-31
Localize and colorize grill temperature settings

Tayrtahn:
## About the PR

Modified the localization strings for electric grills to demonstrate how to handle enum values. It's probably a bit excessive for the English localization, but it should be a helpful template for other languages, and it doesn't hurt to have.

Also made the temperature setting displayed in the examine text change color based on the setting.

## Why / Balance

Addresses the concerns of #27588 and demonstrates the solution for one instance listed there.
The color thing is just kinda neat and seemed like an obvious thing to do while I was there.

## Technical details

The magic of Fluent selectors and terms.

## Media

Screenshot 2025-03-31 at 4 28 45 PM
Screenshot 2025-03-31 at 4 28 51 PM
Screenshot 2025-03-31 at 4 29 00 PM
Screenshot 2025-03-31 at 4 29 07 PM

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

Not really significant enough to warrant a changelog entry, imo.

PR #38199, opened 2025-06-10
show reflection prob on examine

kosticia:
## About the PR
shows prob of reflection with examine
works with ANY entities
yes, pens-daggers protected from metagaming.

## Why / Balance
we should remove need in looking into prototypes/code of the game just for random numbers in proto.

## Media
awdszawsawa
swadswsdw

## Requirements

**Changelog**
:cl:
- add: Reflection chances are examinable now.

### DIFF
--- a/Content.Shared/Weapons/Reflect/ReflectSystem.cs
+++ b/Content.Shared/Weapons/Reflect/ReflectSystem.cs
@@ -2,6 +2,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Numerics;
 using Content.Shared.Administration.Logs;
 using Content.Shared.Database;
+using Content.Shared.Armor;
 using Content.Shared.Hands;
 using Content.Shared.Inventory;
 using Content.Shared.Inventory.Events;
@@ -15,6 +16,7 @@ using Robust.Shared.Network;
 using Robust.Shared.Physics.Components;
 using Robust.Shared.Physics.Systems;
 using Robust.Shared.Random;
+using Content.Shared.Verbs;
 
 namespace Content.Shared.Weapons.Reflect;
 
@@ -45,6 +47,7 @@ public sealed class ReflectSystem : EntitySystem
         SubscribeLocalEvent<ReflectComponent, GotUnequippedEvent>(OnReflectUnequipped);
         SubscribeLocalEvent<ReflectComponent, GotEquippedHandEvent>(OnReflectHandEquipped);
         SubscribeLocalEvent<ReflectComponent, GotUnequippedHandEvent>(OnReflectHandUnequipped);
+        SubscribeLocalEvent<ReflectComponent, ArmorExamineEvent>(OnExamine);
     }
 
     private void OnReflectUserCollide(Entity<ReflectComponent> ent, ref ProjectileReflectAttemptEvent args)
@@ -200,4 +203,35 @@ public sealed class ReflectSystem : EntitySystem
         ent.Comp.InRightPlace = false;
         Dirty(ent);
     }
+
+    #region Examine
+    private void OnExamine(Entity<ReflectComponent> ent, ref ArmorExamineEvent args)
+    {
+        var value = MathF.Round((1f - ent.Comp.ReflectProb) * 100, 1);
+
+        if (value == 0 || ent.Comp.Reflects == ReflectType.None)
+            return;
+
+        string type;
+
+        args.Msg.PushNewline();
+        switch (ent.Comp.Reflects)
+        {
+            case ReflectType.Energy | ReflectType.NonEnergy:
+                type = Loc.GetString(ent.Comp.BothTypeLoc);
+                args.Msg.AddMarkupOrThrow(Loc.GetString(ent.Comp.ExamineLoc, ("value", value), ("type", type)));
+                break;
+
+            case ReflectType.Energy:
+                type = Loc.GetString(ent.Comp.EnergyTypeLoc);
+                args.Msg.AddMarkupOrThrow(Loc.GetString(ent.Comp.ExamineLoc, ("value", value), ("type", type)));
+                break;
+
+            case ReflectType.NonEnergy:
+                type = Loc.GetString(ent.Comp.NonEnergyTypeLoc);
+                args.Msg.AddMarkupOrThrow(Loc.GetString(ent.Comp.ExamineLoc, ("value", value), ("type", type)));
+                break;
+        }
+    }
+    #endregion
 }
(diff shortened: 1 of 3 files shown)


Content.Shared/Weapons/Reflect/ReflectSystem.cs
```diff
@@ -200,4 +202,36 @@ private void OnReflectHandUnequipped(Entity<ReflectComponent> ent, ref GotUnequi
...
+
+            case ReflectType.Energy:
+                type = Loc.GetString("reflect-examine-type-energy");
+                args.PushMarkup(Loc.GetString("reflect-examine", ("value", value), ("type", type)));
+                break;
+
+            case ReflectType.NonEnergy:
+                type = Loc.GetString("reflect-examine-type-nonenergy");
+                args.PushMarkup(Loc.GetString("reflect-examine", ("value", value), ("type", type)));
+                break;
```
slarticodefast: Pretty sure you can do this in fluent and just pass in the enum, although flag enums might be more complicated.
See https://github.com/space-wizards/space-station-14/pull/36236 for an example.

RedBookcase:
Should probably be "It has a [X%] chance to reflect" rather than "It has [X%] to reflect"

Content.Shared/Weapons/Reflect/ReflectSystem.cs
kosticia: done, but with additional string because ftl selectors doesn't support logical operations and just can't pass energy | nonenergy. Actually string can be replaced with byte, but code will be less cool.

kosticia:
> Should probably be "It has a [X%] chance to reflect" rather than "It has [X%] to reflect"

done

Content.Shared/Weapons/Reflect/ReflectSystem.cs
slarticodefast: I meant moving the whole switch case thing to fluent so you only have a single loc string.
That would can be more easily extended without having the enum hardcoded here.

kosticia: Switch killed

beck-thompson (approved):
I'm approving this but this it is sus... I can't think of a neater way of doing it :thinking:

slart will know whats best :laughing:

Content.Shared/Weapons/Reflect/ReflectSystem.cs
```diff
@@ -200,4 +202,23 @@ private void OnReflectHandUnequipped(Entity<ReflectComponent> ent, ref GotUnequi
...
+
+        if (!_toggle.IsActivated(ent.Owner) || value == 0 || ent.Comp.Reflects == ReflectType.None)
+            return;
+
+        var type = ent.Comp.Reflects.ToString();
+
+        // SOOOOOOOOOOO, DEPENDING ON THE HOW FTL SELECTOR WORKS THIS IS ONLY WAY TO DO IT PROPERLY.
+        // YES. THIS SENDS TO THE SELECTOR NON-CAMEL CASE SHIT AND REMOVES ANY ", " SO
+        // "Energy, Nonenergy" turs into "EnergyNonenergy"
+        args.PushMarkup(Loc.GetString("reflect-examine", ("value", value), ("type", type.Replace(", ", null))));
```
slarticodefast: Hmm, the problem is with this it becomes really messy with more than two reflect types because then you have to add a switch in the ftl for every single combination. There must be a way to handle enum flags more elegantly.
@Tayrtahn Do you maybe have an idea?

Tayrtahn: I suggest something like:

- Split the enum flags into a list of values (there doesn't seem to be a built-in way to do this, but stack overflow has some solutions).
- Pass each value into `Loc.GetString` to get a localized version of it.
- Pass those into `ContentLocalizationManager.FormatList` to get a single localized string. 
- Pass that string as a parameter into the final `Loc.GetString` to get the full sentence.

Going back and forth with localization like that is a bit messy, but we use something similar in a few places (entity effects, contraband examine, etc).

LaCumbiaDelCoronavirus:
гол

Tayrtahn (changes requested):
See slarti's note

Resources/Locale/en-US/reflect/reflect-component.ftl
```diff
@@ -0,0 +1,3 @@
+reflect-examine = It has a [color=lightblue]{$value}%[/color] to [color=cyan]reflect[/color] {$type}.
+NonEnergy = bullets
+Energy = energy bolts
```
slarticodefast:
```suggestion
reflect-component-examine = It has a [color=lightblue]{$value}%[/color] to [color=cyan]reflect[/color] {$type}.
reflect-component-NonEnergy = bullets
reflect-component-Energy = energy bolts
```
best give these a prefix, since loc strings are globally scoped and only "energy" is not very specific. Forks may have more entries in the enumb, which may cause a conflict.

slarticodefast (approved):
Looks good to me!

Content.Shared/Weapons/Reflect/ReflectSystem.cs
```diff
@@ -200,4 +203,30 @@ private void OnReflectHandUnequipped(Entity<ReflectComponent> ent, ref GotUnequi
...
+        // This isn't examine verb or something just because it looks too much bad.
+        // Trust me, universal verb for the potential weapons, armor and walls looks awful.
+        var value = MathF.Round(ent.Comp.ReflectProb * 100, 1);
+
+        if (!_toggle.IsActivated(ent.Owner) || value == 0 || ent.Comp.Reflects == ReflectType.None)
+            return;
+
+        var compTypes = ent.Comp.Reflects.ToString().Split(", ");
+
+        List<string> typeList = [];
```
Tayrtahn: A micro-optimization - initialize the List with the needed capacity since we know it in advance:
```suggestion
        List<string> typeList = new(compTypes.Length);
```

Resources/Locale/en-US/reflect/reflect-component.ftl
```diff
@@ -0,0 +1,3 @@
+reflect-component-examine = It has a [color=lightblue]{$value}%[/color] to [color=cyan]reflect[/color] {$type}.
```
Tayrtahn:
```suggestion
reflect-component-examine = It has a [color=lightblue]{$value}%[/color] chance to [color=cyan]reflect[/color] {$type}.
```

Tayrtahn (changes requested):

Tayrtahn (approved):

slarticodefast:

PR #38208, opened 2025-06-10
TEG Rework

### LINKED ISSUE #37902: TEG exponential moving average needs to be replaced with a better solution

### LINKED ISSUE #38090: TEG breaks conservation of energy (In a BIG way - NO I am NOT talking about the magic ramping)
## Description
I investigated the TEG code during one my rabbit hole bouts of learning how things work in game through code diving. And I found this:

Image

Wmax is the "maximum" amount of energy that _could_ be extracted. N is the efficiency (recently changed to .1 which is how I noticed this issue).

So transfer is the amount of actual energy that the TEG extracts, makes sense right? So it sets electrical energy equal to transfer and then-

What's this? It makes an "outTransfer" variable? Sets it to the inverse of efficiency? _looks below_

Ah, so it changes the gas temperatures based on that. I see- but wait. That's not how energy works. If transfer IS the amount of energy the hot gas is losing, and ALSO the _exact_ amount of electrical energy the TEG produces - then where is the heat being given to the cold gas coming from?

Perhaps the code later down is changing it- NOPE. It doesn't.

Every time the TEG processes any amount of gas, it is taking 10% of the difference in energy between them, and produces exactly that much as electrical energy. It, appropriately, saps that energy directly from the hot gas. Energy is conserved.

Then for ?reasons? An amount of energy equal to 90% of that 10% is _created from nothing_ and then given to the cold gas? Where does it come from?

Nowhere. It comes from nowhere.

The TEG is breaking conservation of energy, and by no small margin. However much power it produces? It's ALSO producing 90% of that as _extra heat_ in the cold gas.

**Reproduction**
Easy. Set up any kind of TEG, run it and behold as the amount of electrical energy it produces is exactly equal to the amount of thermal energy _lost by the hot gas_ ; and that despite seemingly 100% of that heat being turned to power, a further 90% of it is arbitrarily created and added to the cold gas.

**Screenshots**
Present in the original post, demonstrating the inherent issue in the code's logic

**Additional context**
Solution is really, really simple. Just Make it so that electrical energy is equal to the inverse fraction of outTransfer. Since outTransfer is transfer * (1 - thermalEfficiency) (.9 or 90% in this case - since thermalEfficiency is currently set to .1 or 10%), electrical energy should be equal to transfer * thermalEfficiency.

Since this would fix TEG breaking conservation without changing how much heat is going to the cold gas, this would also reduce the TEG's max output to 10% of what it currently is - and output that was recently nerfed from 65% to 10%. It may be prudent to partially reverse that nerf to compensate

ArtisticRoomba:
## About the PR
This PR reworks the TEG's power generation logic and behavior. Hard to explain, but here it goes.

The TEG's circulators now have a built in heat mass. When gas moves across the circulators, heat is exchanged between the circulators and the gas moving through them.

To generate power, the TEG now exchanges heat between the two circulator heat masses. It no longer exchanges as much power as possible, it now only exchanges enough heat to satisfy grid demand. So, yes, automatic ramping is now a thing.

The waste heat from the TEG is also dumped to the coldside.

Fixes https://github.com/space-wizards/space-station-14/issues/37902
Fixes https://github.com/space-wizards/space-station-14/issues/38090

## Why / Balance
This will hopefully accomplish the following:
- Makes the TEG more thermodynamically sound.
- Makes the TEG no longer cause extreme power flickering in situations where pressure across the circulators flutter.
- Encourages dual-loop designs a bit more.
- Opens the door for temperature-related TEG mechanics (ex. failure states) in the future.

Even though the TEG now only transfers enough heat across to make enough power to fulfill demand, running a single loop design will lead to wasted heat. Additionally, people who attempt to run single loops will find that they will still have to manually throttle their designs, compared to dual-loop designs which automatically throttle themselves.

### Circulators
Circulators now have a configurable thermal mass that starts at room temperature (or any temperature you desire). They have a high heat capacity to help flatten major dips in heat flow through the circulators (which lead to power flickering).

Circulators exchange heat between the working gas that flows through them according to the equation:
```math
dQ = k(T_{gas} - T_{circulator}) \cdot dt
```
Where:
- $dQ$ is the change in heat. This number can be positive or negative to indicate heat flow in both directions.
- $k$ is a conductivity constant, used to determine how well heat is exchanged between the heat bodies.
- $(T_{gas} - T_{circulator})$ (or $\Delta T$), the difference in temperatures between the heat mediums. This is where the "direction of heat flow" is decided.
- $dt$ is atmos delta time.

This means that circulators can only get as hot as the gas you're pumping through them. Since heat transfer is largely dependent on the heat capacity of the gasses, closed-loop gasses like Water Vapor or Plasma will perform excellently in exchanging heat, compared to burn chamber waste product gasses like Oxygen, or CO2 and Water Vapor in small amounts. 

### TEG Ramping and Power Generation
The TEG now preforms power ramping. This means that it will automatically move only the amount of heat it needs across its sides in order to generate electricity, which significantly benefits dual loop designs and makes single loop designs more work to use.

First, the TEG determines how much energy it can produce *right now* if the circulator heat masses were instantly exchanged. The equation used to calculate this is below:

```math
Q = \Delta T \cdot \frac{C_A C_B}{C_A + C_B}
```
This is derived from the following facts:
1. Let $Q$ = the heat transferred from the hot body to the cold body
2. $T_A > T_B$, so $Q > 0$ for sakes. Heat flows from A -> B.
3. The energy lost from $A$ is equal to $Q = C_A \cdot (T_A - T'_A)$
4. The same is for $B$ except it gains this heat, so $Q = C_B \cdot (T'_B - T_B)$
5. Energy is, of course, conserved. So $T'_A = T_A - \frac{Q}{C_A}$ and $T'_B = T_B + \frac{Q}{C_B}$
6. They will instantly equalize in temperature, so $T_{final} = T'_A = T'_B$
7. Solve for $Q$. So $T_A - \frac{Q}{C_A} = T_B + \frac{Q}{C_B}$. Something something the derivation is left as an exercise to the reader.

After this, it takes the minimum of the two. So we only transfer enough heat to satisfy our ramp target, or however much we can produce right now if our ramp target exceeds that.

After this, we convert our transferred portion of heat to electricity. A heat engine will produce electricity via the following equation:
```math
Q_{in} = W + Q_{out}
```
Where:
- $Q_{in}$ is the amount of heat that we're working with.
- $W$ is the amount of energy we extract from the system.
- $Q_{out}$ is the amount of energy that is left.

That's it.

## Media
N/A for now, need to do more testing. But this is good to review.

## Requirements

## Breaking changes
The `PowerSmoothingFactor` field in `TegGeneratorComponent` has been removed, in favor of a more mechanical and realistic solution to the power flickering problem.

The `Circulator` struct in `TegSensorData` now contains `Temperature` information for the TEG's circulators.

**Changelog**
:cl:
- add: The TEG circulators now contain a large thermal mass. Gases flowing through the circulators now exchange their thermal energy between this thermal mass.
- tweak: The TEG now exchanges energy between the thermal masses on each circulator in order to generate electricity.
- tweak: The TEG now preforms automatic power ramping, only transferring enough heat between the circulators to satisfy grid demand.
- fix: The TEG now better respects the laws of thermodynamics.


ArtisticRoomba:
Note that the conductivity/heat capacity constants are super work in progress, I stopped my work in balancing them after I got fed up with the sensor monitoring console lagging my game, so I fixed it in #38146 as a side project.

K-Dynamic:
Just let us use steam power in space smh

ArtisticRoomba:
Updated PR desc and CL, hopefully ready for review though the constants need some more tuning I feel.

CookieMasterT:
Frezon teg buff no way :)

ArtisticRoomba:
Requires https://github.com/space-wizards/space-station-14/pull/42197.

Dicerson:
Just want to come in here and comment that I love what you're doing here! This is a great rework, at least at the concept level (I am not code literate enough to read through the actual file implementation) and sincerely hope you continue working on this! It's about as close to a perfect TEG implementation that I can envision!

ArtisticRoomba:
Thanks. I just haven't merged this yet because it was blocked by `HeatContainer`s (which is resolved) and to be honest it shakes up a few things, and I'm not quite sure if it would be a good time to do so.

PR #38212, opened 2025-06-10
HOTFIX: Stop AI turrets from causing flickering lights

slarticodefast:
Temporary bandaid because this may cause seizures for some players.
We set the power draw to 0 so the APC has no load.
The turrets still require a powered APC to recharge, so they are still functional.


ArtisticRoomba (approved):
world's cheesiest fix though definitely a must considering how bad this could be

Issue #38078, opened 2025-06-05
construction menu grid view does not deselect when selecting items with the same localized name


qwerltaz:
## Description

In the construction menu grid view, selecting items with the same localized name in succession does not deselect the previous items. If you try to select one of the already selected items, it just gets deselected. The last item of the same-name group never lights up as selected if any previous ones are selected.

This makes for a heavy annoyance.

Does not occur in standard list view.

**Reproduction**

1. open construction menu (G)
2. toggle grid view
3. select items with the same localized name consecutively

**Screenshots**

https://github.com/user-attachments/assets/b58fce80-face-45ce-ba21-7f108aa41cbb

**Additional context**

I believe this was introduced in https://github.com/space-wizards/space-station-14/pull/32339

PR #32339, opened 2024-09-20
Localizable craftmenu

Ertanic:
## About the PR

The name, description and icon of the target crafting items are now taken directly from the prototypes. The ability to specify a name is left through localization files. See the secret door for an example.

Depends on ~~https://github.com/space-wizards/RobustToolbox/pull/5466~~.

## Why / Balance

Fixes #22814

## Technical details

Because prototypes are read only on the server, I moved some of the code from the Server segment to Shared and Client. When ConstructionPrototype comes in, it creates a ConstructionPrototype.ID-to-EntityPrototype.ID map, which the client then works with.

To take sprites directly from the prototypes, `ItemList` has been replaced by `ListContainer`, which uses `EntityPrototypeView`.

Fluent keys have been used to localize GraphStep. Therefore, the names of the `tag` and `material` steps, which are used directly in the build menu, have been moved to ftl files.

The `MaterialConstructionGraphStep` (set in StackPrototype) passes the necessary amount of material to be translated using selectors. Especially relevant for the Slavic language group.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

- tweak: The name, description and icon of the target crafting items are now taken directly from the prototypes.

PR #38214, opened 2025-06-10
Fix construction menu grid view quirky button selection

qwerltaz:
## About the PR

1. fix https://github.com/space-wizards/space-station-14/issues/38078
2. Change grid view button hover-over name to use the construction prototype name instead of the entity prototype name.
3. Return item turning green when selected. This originally happened and was lost in https://github.com/space-wizards/space-station-14/pull/32339.

## Why / Balance

1. bug fix
2. Some entities have identical names, but more specific names can be defined for the construction prototype, like in https://github.com/space-wizards/space-station-14/pull/38182. After this fix, the grid view uses those more specific names.
3. Bug fix. selected item turning green was originally there, so that one can clearly see an item is selected even if it's large and the lit background is not visible.

## Technical details

1. Change _recipeButtons to use prototype IDs as keys instead of prototype names :trollface: 

## Media

We now use the construction prototype name 'Curtains (white)' instead of entity name 'Curtains'.

image

### DIFF
--- a/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
+++ b/Content.Client/Construction/UI/ConstructionMenuPresenter.cs
@@ -217,8 +217,8 @@ namespace Content.Client.Construction.UI
                 var itemButton = new ContainerButton()
                 {
                     VerticalAlignment = Control.VAlignment.Center,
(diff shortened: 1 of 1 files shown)


beck-thompson (approved):
Good fix

PR #38220, opened 2025-06-10
Remove redunant loc string for EntityPrototype

lzk228:
~~please no more~~

## About

Remove redundant loc string from the entity prototype. EntityPrototypes have automatically generated loc string names, so we don't need to add one manually.

### DIFF
--- a/Resources/Locale/en-US/actions/actions/rootable.ftl
+++ b/Resources/Locale/en-US/actions/actions/rootable.ftl
@@ -1,2 +1,0 @@
-action-name-toggle-rootable = Rootable
-action-description-toggle-rootable = Begin or stop being rooted to the floor.


keronshb:
This needs an actual description. Otherwise I have no idea why removing loc is good.

slarticodefast:
> This needs an actual description. Otherwise I have no idea why removing loc is good.

EntityPrototypes have automatically generated loc string names, so we don't need to add one manually.

https://docs.spacestation14.com/en/ss14-by-example/fluent-and-localization.html#localizing-prototypes

lzk228:
> This needs an actual description

the description is fine and you still get an alert that says "you can't slip but you absorb reagents" or something

keronshb:
> > This needs an actual description. Otherwise I have no idea why removing loc is good.
> 
> EntityPrototypes have automatically generated loc string names, so we don't need to add one manually.
> 
> https://docs.spacestation14.com/en/ss14-by-example/fluent-and-localization.html#localizing-prototypes

This is good but for other people reading the PR, not so much.

IIRC we're trying to cut down on meme titled PRs like this with no description.

slarticodefast:
I went ahead and edited the title.

Errant-4:
~~Did this just remove the loc strings from the entity and left them cluttering the .ftl forever?~~

Oh, github trolled me and did not load the diff for the ftl so I missed it

SlamBamActionman:
> @SlamBamActionman [shame meme]

I'm never adding Loc agaaaain 😭

PR #38224, opened 2025-06-10
Added Hemophilia Trait

TheFlyingSentry:
## About the PR
Added the Hemophilia Trait, people with this trait will now bleed even longer and may have to think about making gauze or tourniquets with them... and maybe some garlic too.

## Why / Balance
It was one of the few traits when I first stared playing that I was surprised to not see.
Hopefully in the future someone will code specific blood intervals in which people will faint so it's not just "chug some more dex and saline and you'll be fine"

## Technical details
Added a new cs file in Shared/Traits (Where traits aught to go instead of literally everywhere but)
Added a couple lines in Bloodstream.cs and added a thing into the component itself

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added a new trait: Hemophilia!

### DIFF
--- a/Content.Server/Body/Components/BloodstreamComponent.cs
+++ b/Content.Server/Body/Components/BloodstreamComponent.cs
@@ -51,6 +51,12 @@ namespace Content.Server.Body.Components
         public float BleedReductionAmount = 0.33f;
 
         /// <summary>
+        ///     How much should bleeding be reduced every update interval for the hemophilia trait?
+        /// </summary>
+        [DataField]
+        public float HemophiliacBleedReductionAmount = 0.10f;
+
+        /// <summary>
         ///     How high can <see cref="BleedAmount"/> go?
         /// </summary>
         [DataField]
--- a/Content.Server/Body/Components/BloodstreamComponent.cs
+++ b/Content.Server/Body/Components/BloodstreamComponent.cs
@@ -51,12 +51,6 @@ namespace Content.Server.Body.Components
         public float BleedReductionAmount = 0.33f;
 
         /// <summary>
-        ///     How much should bleeding be reduced every update interval for the hemophilia trait?
-        /// </summary>
-        [DataField]
-        public float HemophiliacBleedReductionAmount = 0.10f;
-
-        /// <summary>
         ///     How high can <see cref="BleedAmount"/> go?
         /// </summary>
         [DataField]
--- a/Content.Server/Body/Systems/BloodstreamSystem.cs
+++ b/Content.Server/Body/Systems/BloodstreamSystem.cs
@@ -134,16 +134,8 @@ public sealed class BloodstreamSystem : EntitySystem
                 // Blood is removed from the bloodstream at a 1-1 rate with the bleed amount
                 TryModifyBloodLevel(uid, (-bloodstream.BleedAmount), bloodstream);
 
-                if (!HasComp<HemophiliaComponent>(uid))
-                {
-                    // Bleed rate is reduced by the bleed reduction amount in the bloodstream component.
-                    TryModifyBleedAmount(uid, -bloodstream.BleedReductionAmount, bloodstream);
-                }
-                else
-                {
-                    TryModifyBleedAmount(uid, -bloodstream.HemophiliacBleedReductionAmount, bloodstream);
-                }
-            }
+                // The rate in which stacks are reduced. Think of it like the rate at which your blood clots
+                TryModifyBleedAmount(uid, -bloodstream.BleedReductionAmount, bloodstream);
 
             // deal bloodloss damage if their blood level is below a threshold.
             var bloodPercentage = GetBloodLevelPercentage(uid, bloodstream);
--- a/Content.Server/Body/Systems/BloodstreamSystem.cs
+++ b/Content.Server/Body/Systems/BloodstreamSystem.cs
@@ -19,6 +19,7 @@ using Content.Shared.Mobs.Systems;
 using Content.Shared.Popups;
 using Content.Shared.Rejuvenate;
 using Content.Shared.Speech.EntitySystems;
+using Content.Shared.Traits.Assorted;
 using Robust.Server.Audio;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
@@ -132,8 +133,16 @@ public sealed class BloodstreamSystem : EntitySystem
             {
                 // Blood is removed from the bloodstream at a 1-1 rate with the bleed amount
                 TryModifyBloodLevel(uid, (-bloodstream.BleedAmount), bloodstream);
-                // Bleed rate is reduced by the bleed reduction amount in the bloodstream component.
-                TryModifyBleedAmount(uid, -bloodstream.BleedReductionAmount, bloodstream);
+
+                if (!HasComp<HemophiliaComponent>(uid))
+                {
+                    // Bleed rate is reduced by the bleed reduction amount in the bloodstream component.
+                    TryModifyBleedAmount(uid, -bloodstream.BleedReductionAmount, bloodstream);
+                }
+                else
+                {
+                    TryModifyBleedAmount(uid, -bloodstream.HemophiliacBleedReductionAmount, bloodstream);
+                }
             }
 
             // deal bloodloss damage if their blood level is below a threshold.
--- a/Resources/Prototypes/Entities/Mobs/Player/clone.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/clone.yml
@@ -16,6 +16,7 @@
   - NpcFactionMember
   # traits
   # - LegsParalyzed (you get healed)
+  - Hemophilia
   - LightweightDrunk
   - Monochromacy
   - Muted
--- a/Content.Shared/Traits/Assorted/HemophiliaComponent.cs
+++ b/Content.Shared/Traits/Assorted/HemophiliaComponent.cs
@@ -4,7 +4,13 @@ namespace Content.Shared.Traits.Assorted;
 
 /// <summary>
 /// This component is used for the Hemophilia Trait, it reduces the passive bleed stack reduction amount so entities with it bleed for longer.
-/// The reduction SHOULD be located in BloodstreamComponent.cs as HemophiliacBleedReductionAmount
 /// </summary>
 [RegisterComponent, NetworkedComponent]
-public sealed partial class HemophiliaComponent : Component;
+public sealed partial class HemophiliaComponent : Component
+{
+    /// <summary>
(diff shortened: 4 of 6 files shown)


TheFlyingSentry:
Theres probably a way to change it so I don't have to add another component but theres always a chance it get's used for vamps, med, or some other PR in the future... or something downstream. Who knows

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -132,8 +133,16 @@ public override void Update(float frameTime)
             {
                 // Blood is removed from the bloodstream at a 1-1 rate with the bleed amount
                 TryModifyBloodLevel(uid, (-bloodstream.BleedAmount), bloodstream);
-                // Bleed rate is reduced by the bleed reduction amount in the bloodstream component.
-                TryModifyBleedAmount(uid, -bloodstream.BleedReductionAmount, bloodstream);
+
+                if (!HasComp<HemophiliaComponent>(uid))
```
beck-thompson: Use an event here! Raise it on the player and listen for it in the hemophilia system (you'll need to make it)

Content.Server/Body/Components/BloodstreamComponent.cs
```diff
@@ -50,6 +50,12 @@ public sealed partial class BloodstreamComponent : Component
         [DataField]
         public float BleedReductionAmount = 0.33f;
 
+        /// <summary>
+        ///     How much should bleeding be reduced every update interval for the hemophilia trait?
+        /// </summary>
+        [DataField]
+        public float HemophiliacBleedReductionAmount = 0.10f;
```
beck-thompson: Put this on the component

beck-thompson (changes requested):
isn't this just big bleed buff? Why wouldn't you take this?

TheFlyingSentry:
> isn't this just big bleed buff? Why wouldn't you take this?

It's worded weirdly but the comment is kinda misleading. you do bleed more over time
The only thing i've made it do is remove less stacks over time from .33 stacks -> .10 stacks

TheFlyingSentry:
the .1 speed is the speed at which your blood clots
.33 is base and as it's bigger it'll clot faster
since your blood clots slower you tend to lose more blood. _the speed of bleeding is not effected_

beck-thompson:
Ahhh I see what your saying makes sense! Sounds good then

walksanatora:
so you dont bleed faster. you just bleed longer. makes sense.

TheCrimsonJupiter:
Absolutely awesome trait!

Draedon4:
next we need the anemia trait that makes bloodloss very very deadly

Content.Server/Body/Components/BloodstreamComponent.cs
TheFlyingSentry: Done

Content.Server/Body/Systems/BloodstreamSystem.cs
TheFlyingSentry: Thanks for teaching me <3

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -40,6 +41,7 @@ public sealed class BloodstreamSystem : EntitySystem
     [Dependency] private readonly SharedSolutionContainerSystem _solutionContainerSystem = default!;
     [Dependency] private readonly SharedStutteringSystem _stutteringSystem = default!;
     [Dependency] private readonly AlertsSystem _alertsSystem = default!;
+    [Dependency] private readonly HemophiliaSystem _hemophiliaSystem = default!;
```
beck-thompson: Looks like this is unused

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -1,6 +1,7 @@
 using Content.Server.Body.Components;
 using Content.Server.Fluids.EntitySystems;
 using Content.Server.Popups;
+using Content.Server.Traits.Assorted;
```
beck-thompson: also unused

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -128,13 +130,16 @@ public override void Update(float frameTime)
...
             // as well as stop their bleeding to a certain extent.
-            if (bloodstream.BleedAmount > 0)
-            {
-                // Blood is removed from the bloodstream at a 1-1 rate with the bleed amount
-                TryModifyBloodLevel(uid, (-bloodstream.BleedAmount), bloodstream);
-                // Bleed rate is reduced by the bleed reduction amount in the bloodstream component.
-                TryModifyBleedAmount(uid, -bloodstream.BleedReductionAmount, bloodstream);
-            }
+            if (bloodstream.BleedAmount <= 0)
+                continue;
```
beck-thompson: Add a newline after this

beck-thompson (changes requested):
only some minor stuff

Content.Server/Body/Systems/BloodstreamSystem.cs
TheFlyingSentry: Done

Content.Server/Body/Systems/BloodstreamSystem.cs
TheFlyingSentry: Done

Content.Server/Body/Systems/BloodstreamSystem.cs
TheFlyingSentry: Done

Content.Server/Traits/Assorted/HemophiliaComponent.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Robust.Shared.GameStates;
+
+namespace Content.Server.Traits.Assorted;
+
+/// <summary>
+/// This component is used for the Hemophilia Trait, it reduces the passive bleed stack reduction amount so entities with it bleed for longer.
+/// </summary>
+[RegisterComponent]
+public sealed partial class HemophiliaComponent : Component
```
slarticodefast: move the component to shared
I'm currently working on making the bloodstream predicted

Content.Server/Traits/Assorted/HemophiliaComponent.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    ///     How much should bleeding be reduced every update interval for the hemophilia trait?
```
slarticodefast:
```suggestion
    /// How much should bleeding be reduced every update interval for the hemophilia trait?
```

Content.Server/Traits/Assorted/HemophiliaSystem.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Content.Server.Traits.Assorted;
+
+public sealed partial class HemophiliaSystem : EntitySystem
```
slarticodefast: move to shared

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -128,13 +128,17 @@ public override void Update(float frameTime)
...
```
slarticodefast: this is not equivalent
this means you no longer take bloodloss damage from low blood level when you stopped bleeding as the guard statement skips that code now

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -128,13 +128,17 @@ public override void Update(float frameTime)
...
+
+            // Blood is removed from the bloodstream at a 1-1 rate with the bleed amount
+            TryModifyBloodLevel(uid, (-bloodstream.BleedAmount), bloodstream);
+
+            var modifyBleedLevel = new BleedStackReduceEvent(bloodstream.BleedReductionAmount);
```
slarticodefast:
```suggestion
            var ev = new BleedStackReduceEvent(bloodstream.BleedReductionAmount);
```
usual naming convention

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -128,13 +128,17 @@ public override void Update(float frameTime)
...
+            RaiseLocalEvent(uid, ref modifyBleedLevel);
+
+            // The rate in which stacks are reduced. Think of it like the rate at which your blood clots
+            TryModifyBleedAmount(uid, -modifyBleedLevel.BleedStackReductionAmount, bloodstream);
```
slarticodefast:
```suggestion
            RaiseLocalEvent(uid, ref ev);

            // The rate in which stacks are reduced. Think of it like the rate at which your blood clots
            TryModifyBleedAmount(uid, -ev.BleedStackReductionAmount, bloodstream);
```

Content.Server/Traits/Assorted/HemophiliaSystem.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<HemophiliaComponent, BleedStackReduceEvent>(OnBleedStackReduceEvent);
+    }
+
+    private void OnBleedStackReduceEvent(Entity<HemophiliaComponent> ent, ref BleedStackReduceEvent args)
+    {
+        args.BleedStackReductionAmount = ent.Comp.HemophiliacBleedReductionAmount;
```
slarticodefast: Maybe offset it by a certain value rather than overwriting it or use a multiplier?
In case the base value is ever changed

slarticodefast:
Are you coming back to this?

ScarKy0:
Addressed the review myself

Content.Shared/Body/Components/BloodstreamComponent.cs
```diff
@@ -8,6 +8,7 @@
 using Robust.Shared.Audio;
 using Robust.Shared.GameStates;
 using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization;
```
ScarKy0: Unused.

Resources/Prototypes/Entities/Mobs/Player/clone.yml
```diff
@@ -18,6 +18,7 @@
   - Clumsy
   - ImpairedMobility
   # - LegsParalyzed (you get healed)
+  - Hemophilia
```
slarticodefast:
```suggestion
  - Hemophilia
  - ImpairedMobility
  # - LegsParalyzed (you get healed)
```

TheFlyingSentry:
Totally didn't forget that I was working on this
Ty though! <3

PR #38231, opened 2025-06-10
better colors

Orsoniks:
## About the PR
Added improved coloring to the Health Examination UI, damages are now colored by the color of their usual healing chem/common sense color, getting more intense as the damage increases (EG. Brutes are colored by the colors of their advanced brute chems, shock is yellow due to common sense, cold is blue due to common sense, etc)

## Why / Balance
This makes distinguishing between damage types MUCH easier, and will aid players in learning the game, especially for medical roles

## Technical details
changes locale file

## Media
image
Full palette for all damage types (Slash, Blunt, Piercing, Asphyx, Heat, Shock, Cold, Caustic, Rad) (8-15-30-50-75-100-200)
image

## Requirements

## Breaking changes

**Changelog**
Included in the original health exam changelog i guess


luckyshotpictures:
The second and the fifth color look fairly similar to me.

Orsoniks:
> The second and the fifth color look fairly similar to me.

one is red and the other's orange, both are colored after the respective meds (Bruizine/Pyrazine)
this PR does not take color blindness into account so that might be a factor

dora-meowmeow:
Wouldn't it be better to make two PRs, one for all of the new shades of these colours with actual names, and other as this?

I think having more shades of colour would be great.

Orsoniks:
> Wouldn't it be better to make two PRs, one for all of the new shades of these colours with actual names, and other as this?
> 
> I think having more shades of colour would be great.

Huh? This PR only changes the colors though

Issue #38221, opened 2025-06-10
We need an integration test to ensure EntityPrototypes only use automatically generated Loc strings


slarticodefast:
Contributors often add loc strings for `EntityPrototype` names or descriptions, even though that is already done automatically. See https://github.com/space-wizards/space-station-14/pull/38220 for the latest example.

See https://docs.spacestation14.com/en/ss14-by-example/fluent-and-localization.html#localizing-prototypes for an explanation.

@Tayrtahn

PR #38234, opened 2025-06-10
Add test that entity prototypes don't use LocIds for name/description

Tayrtahn:
## About the PR

Adds an integration test that checks that no entity prototype uses a `LocId` for its name or description.

## Why / Balance

We have an automated system for localizing entity prototypes which should be used instead. See https://docs.spacestation14.com/en/ss14-by-example/fluent-and-localization.html#localizing-prototypes for more information.

Resolves https://github.com/space-wizards/space-station-14/issues/38221. 

## Technical details

Examines each entity prototype. For each prototype, if a name or description is set, checks if the value is registered as an ID with the localization manager. If it is, the test fails with an explanation of why this is a problem.

The test detected one entity that currently fails: the mapping entity eraser (`ActionMappingEraser`). This was fixed so the test passes.

## Requirements

### DIFF
--- a/Resources/Locale/en-US/actions/actions/mapping.ftl
+++ b/Resources/Locale/en-US/actions/actions/mapping.ftl
@@ -1,1 +1,0 @@
-action-name-mapping-erase = Erase Entity
--- a/Resources/Prototypes/Actions/mapping.yml
+++ b/Resources/Prototypes/Actions/mapping.yml
@@ -28,7 +28,7 @@
 - type: entity
   parent: BaseMappingAction
   id: ActionMappingEraser
-  name: action-name-mapping-erase
+  name: Erase Entity
   components:
   - type: Action
     icon: Interface/VerbIcons/delete.svg.192dpi.png


ArtisticRoomba:
My hero. This would have been amazing to have when I broke master for about 5 minutes when I merged an old PR before localized construction menu entries.

Content.IntegrationTests/Tests/Localization/EntityPrototypeLocalizationTest.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        Assert.Multiple(() =>
+        {
+            foreach (var proto in protos)
+            {
+                // Check name
+                if (!string.IsNullOrEmpty(proto.SetName))
+                {
+                    Assert.That(locMan.HasString(proto.SetName), Is.False,
+                        $"Entity prototype {proto.ID} has a LocId ({proto.SetName}) as a name. {NoLocIdExplanation}");
```
slarticodefast, line 36: If possible it's probably better for readability if you only show the `NoLocIdExplanation` text once at the end in case of multiple asserts.
But this is not a blocker.

Tayrtahn: I like this idea, and it would be nicer for readability, but I don't see a good way of implementing it. It would be perfect if `Assert.Multiple` had an optional parameter for a message to display in the case of any contained assertions failing, but sadly that's not a thing.

slarticodefast (approved):

slarticodefast:
@lzk228 You will finally no longer have to fix these 😄

PR #38238, opened 2025-06-10
Add very gameplay metrics

beck-thompson:
## About the PR

Right now there is zero way to save data besides looking through admin logs (painful). This is a VERY VERY basic system that will allow some way to actually see the impact of specific things. This is NOT supposed to be the be all end all of telemetry systems (This is literally just a database table with a hat :sob:) but I think its good enough to use and see what we actually want in a real fully finished system.

## Technical details

The second commit is the migration so if you want to not get confused, just look at the first one!

## Media

image

## Requirements


walksanatora:
I may be stupid but what does this add that admin logs dont have?

PJB3005 (changes requested):
This needs an actual use case.

Also, "telemetry" isn't even a good name. I would expect "telemetry" to be used for data collection on end-user systems or something.

slarticodefast:
> I may be stupid but what does this add that admin logs dont have?

For context: If we currently want to look up in-game statistics like for example the uplink store purchases we have to use regex to crawl through the admin logs. Which is inefficient, complicated and doesn't log anything not in the admin logs.

ArtisticRoomba:
Yeah crawling through the admin logs for game stats is a pain. This is especially painful when we have to delete the logs, it would be very cool to look at long term trends.

Might be good to look into Jez's unfinished telemetry/game stats branch to see if there's anything you'd like to seek inspiration from?

beck-thompson:
I just wanted to mention again: This system isn't the best way to implement this! I just think its important to have some kind of way to collect data (Even if its a little sus :laughing:)

If people think otherwise that's totally fine and don't worry I have no attachment to the PR so please feel free to close (we have to reduce the PR backlog!).

Again though, I think being able to actually see what players are doing (instead of just having to rely on inround polls and discord discussions) is important and this PR will help a LOT in doing it.

beck-thompson:
I added a discussion post! https://forum.spacestation14.com/t/add-basic-gameplay-metrics/21803

beck-thompson:
I still think this is a good idea even if it isn't perfect

Issue #38240, opened 2025-06-10
Speech bubbles reappear when reentering PVS


Tayrtahn:
Making an official issue from [this comment](https://github.com/space-wizards/RobustToolbox/issues/5960#issuecomment-2960067746)

>[This PR](https://github.com/space-wizards/RobustToolbox/issues/5960) introduces a bug with speech bubbles. Bubbles are supposed to be hidden when the user is obscured (e.g. by a wall), and as such the chat UI code sets their visibility to false. The FrameUpdate loop then counts down the timer before the bubble should disappear. However with this PR, since the visibility is set to false, the FrameUpdate loop does not happen and the timer does not go down, resulting in the bubble appearing once the speaker is back in view. 
>
>This might be possible to solve in Content with some sort of refactor to the bubble system, but that seems fairly difficult. As-is, bubbles seem to be a type of Control that does need to run FrameUpdate when invisible.

_Originally posted by @SlamBamActionman in https://github.com/space-wizards/RobustToolbox/issues/5960#issuecomment-2960067746_

PR #38241, opened 2025-06-10
Replace Speech bubble time accumulator with TimeSpan

Tayrtahn:
## About the PR

Speech bubbles now track the time that they will fade and disappear with a `TimeSpan`, rather than accumulating frametime.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/38240.

## Technical details

A recent engine change (https://github.com/space-wizards/RobustToolbox/pull/5960) made it so that controls that are not visible no longer have their `FrameUpdate` method called. This is good for efficiency, but had an unexpected side-effect for speech bubbles.

Speech bubbles track their fade and death times by a float field named `_timeLeft`. Their `FrameUpdate` method subtracts the time delta from this field every frame. But with the engine change mentioned above, when a speech bubble leaves PVS and is made invisible, `FrameUpdate` no longer gets called and `_timeLeft` stops being updated, effectively pausing its fade. When the speech bubble reenters PVS, `_timeLeft` resumes updating, which can cause the bubble to reappear.

To fix this, `_timeLeft` has been removed and replaced with a `TimeSpan` that tracks the time at which the speech bubble will disappear. This means that when a bubble leaves and reenters PVS, its death time will still be accurate and it won't reappear.

## Requirements

## Breaking changes

Technically this shouldn't be a breaking change since it's all private to `SpeechBubble`, but just to be safe:

`SpeechBubble._timeLeft` has been removed and a new `_deathTime` field now tracks the time at which the bubble will disappear. `SpeechBubble.TotalTime` and `SpeechBubble.FadeTime` are now `TimeSpan`s instead of `int`s.

### DIFF
--- a/Content.Client/Chat/UI/SpeechBubble.cs
+++ b/Content.Client/Chat/UI/SpeechBubble.cs
@@ -14,6 +14,7 @@ namespace Content.Client.Chat.UI
 {
     public abstract class SpeechBubble : Control
     {
+        [Dependency] private readonly IGameTiming _timing = default!;
         [Dependency] private readonly IEyeManager _eyeManager = default!;
         [Dependency] private readonly IEntityManager _entityManager = default!;
         [Dependency] protected readonly IConfigurationManager ConfigManager = default!;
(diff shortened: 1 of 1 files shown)


Tayrtahn:
> This should use `RealTime`, not `CurTime`.

Right, make sense. Done.

> Also, this isn't worth logging a breaking change over.

Removed.

PJB3005 (approved):

PR #38250, opened 2025-06-11
Forbid string literals for `SharedToolSystem` methods

Tayrtahn:
## About the PR

Adds the `ForbidLiteral` attribute to `SharedToolSystem` methods, to encourage coders to pass in `ProtoId`s instead of string literals.

## Why / Balance

Magic strings are bad. This encourages cleaner code through the use of validated ProtoIds.

## Technical details

Methods in `SharedToolSystem` that take strings to use as `ToolQualityPrototype`s were given the `ForbidLiteral` attribute.

Existing code that used string literals for these methods was updated to do it properly (as a separate commit for ease of review).

## Requirements


Content.Server/Fax/FaxSystem.cs
```diff
@@ -28,6 +28,8 @@
 using Content.Shared.NameModifier.Components;
 using Content.Shared.Power;
 using Content.Shared.DeviceNetwork.Components;
+using Robust.Shared.Prototypes;
+using Content.Shared.Tools;
```
slarticodefast: sort alphabetically
same for the other files

slarticodefast (changes requested):

Content.Server/Fax/FaxSystem.cs
Tayrtahn: Vs code has betrayed me

Tayrtahn: Fixed using an extension. Turns out they were all kinds of messed up even before I got there.

slarticodefast (approved):

PR #38253, opened 2025-06-11
DoAfter support for Actions

keronshb:
## About the PR

Actions now have DoAfter support! Just slap a DoAfter Component on an action and fill in the details!

## Why / Balance

DoAfter support for actions was a long time coming. You can do a lot of interesting things with it such as:
- Cast Times
- Repeating actions
  - Repeating actions that have a lower doafter time

## Technical details

- DoAfter Arg fields were also added to the DoAfterComponent so they can be relayed.
- AttemptFrequency was given serialization/netserialization
- DelayReduction was also added to the doafter component - to set a repeat delay to a lower time
- A force option was added to the doafter cancel method so we can cancel even if the doafter was completed.
- ActionDoAfterEvent was added to pass action info back to the action class after the doafter completes, so we can try to do another action
- LimitedChargesComponent also handles doafter cancellation if there's no more charges left on the action
- Doaftersystem handles the relay for actions
- A base DoAfter Action entity was added
- 

## Media

https://github.com/user-attachments/assets/c65c2974-b5a0-4963-8712-c35a8126954b

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/Actions/Events/ActionDoAfterEvent.cs
+++ b/Content.Shared/Actions/Events/ActionDoAfterEvent.cs
@@ -1,4 +1,5 @@
-﻿using Content.Shared.DoAfter;
+﻿using Content.Shared.Actions.Components;
+using Content.Shared.DoAfter;
 using Robust.Shared.Serialization;
 
 namespace Content.Shared.Actions.Events;
@@ -9,11 +10,15 @@ namespace Content.Shared.Actions.Events;
 [Serializable, NetSerializable]
 public sealed partial class ActionDoAfterEvent : DoAfterEvent
 {
-    public readonly RequestPerformActionEvent RequestEvent;
+    [NonSerialized]
+    public readonly Entity<ActionsComponent?> Performer;
 
-    public ActionDoAfterEvent(RequestPerformActionEvent requestEvent)
+    public readonly TimeSpan? OriginalUseDelay;
+
+    public ActionDoAfterEvent(Entity<ActionsComponent?> performer, TimeSpan? originalUseDelay)
     {
-        RequestEvent = requestEvent;
+        Performer = performer;
+        OriginalUseDelay = originalUseDelay;
     }
 
     public override DoAfterEvent Clone() => this;
--- a/Content.Shared/Actions/Events/ActionDoAfterEvent.cs
+++ b/Content.Shared/Actions/Events/ActionDoAfterEvent.cs
@@ -1,0 +1,20 @@
+﻿using Content.Shared.DoAfter;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.Actions.Events;
+
+/// <summary>
+/// The event that triggers when an action doafter is completed or cancelled
+/// </summary>
+[Serializable, NetSerializable]
+public sealed partial class ActionDoAfterEvent : DoAfterEvent
+{
+    public readonly RequestPerformActionEvent RequestEvent;
+
+    public ActionDoAfterEvent(RequestPerformActionEvent requestEvent)
+    {
+        RequestEvent = requestEvent;
+    }
+
+    public override DoAfterEvent Clone() => this;
+}
--- a/Content.Shared/Actions/Events/ActionDoAfterEvent.cs
+++ b/Content.Shared/Actions/Events/ActionDoAfterEvent.cs
@@ -10,15 +10,27 @@ namespace Content.Shared.Actions.Events;
 [Serializable, NetSerializable]
 public sealed partial class ActionDoAfterEvent : DoAfterEvent
 {
+    /// <summary>
+    /// <inheritdoc cref="ActionAttemptDoAfterEvent.Performer"/>
+    /// </summary>
     [NonSerialized]
     public readonly Entity<ActionsComponent?> Performer;
 
+    /// <summary>
+    /// <inheritdoc cref="ActionAttemptDoAfterEvent.OriginalUseDelay"/>
+    /// </summary>
     public readonly TimeSpan? OriginalUseDelay;
 
-    public ActionDoAfterEvent(Entity<ActionsComponent?> performer, TimeSpan? originalUseDelay)
+    /// <summary>
+    /// <inheritdoc cref="ActionAttemptDoAfterEvent.Input"/>
+    /// </summary>
+    public readonly RequestPerformActionEvent Input;
+
+    public ActionDoAfterEvent(Entity<ActionsComponent?> performer, TimeSpan? originalUseDelay, RequestPerformActionEvent input)
     {
         Performer = performer;
         OriginalUseDelay = originalUseDelay;
+        Input = input;
     }
 
     public override DoAfterEvent Clone() => this;
--- a/Content.Shared/Actions/SharedActionsSystem.cs
+++ b/Content.Shared/Actions/SharedActionsSystem.cs
@@ -36,6 +36,8 @@ public abstract class SharedActionsSystem : EntitySystem
     private EntityQuery<ActionsComponent> _actionsQuery;
     private EntityQuery<MindComponent> _mindQuery;
 
+    private const int DoAfterRepeatUseDelay = 0;
+
     public override void Initialize()
     {
         base.Initialize();
@@ -82,11 +84,33 @@ public abstract class SharedActionsSystem : EntitySystem
 
     private void OnActionDoAfterAttempt(Entity<ActionComponent> ent, ref ActionDoAfterEvent args)
     {
+        Entity<ActionComponent?>? action = (ent, ent);
+
+        if (TryComp<DoAfterComponent>(ent, out var doAfterComponent))
+        {
+            // If this doafter is on repeat and was cancelled, start use delay as expected
+            if (args.Cancelled && doAfterComponent.Repeat)
+            {
+                SetUseDelay(action, args.OriginalUseDelay);
+                RemoveCooldown(action);
+                StartUseDelay(action);
+                UpdateAction(ent);
+                return;
+            }
+
+            args.Repeat = doAfterComponent.Repeat;
+
+            // Set the use delay to 0 so this can repeat properly
+            if (doAfterComponent.Repeat)
+            {
+                SetUseDelay(action, TimeSpan.FromSeconds(DoAfterRepeatUseDelay));
+            }
+        }
+
         if (args.Cancelled)
             return;
 
-        // Run validations again
-        TryPerformAction(args.RequestEvent, args.User, skipDoAfter: true);
+        PerformAction(args.Performer, ent);
     }
 
     private void OnActionMapInit(Entity<ActionComponent> ent, ref MapInitEvent args)
@@ -272,21 +296,16 @@ public abstract class SharedActionsSystem : EntitySystem
     /// </summary>
     private void OnActionRequest(RequestPerformActionEvent ev, EntitySessionEventArgs args)
     {
-        TryPerformAction(ev, args.SenderSession.AttachedEntity);
-    }
-
-    public bool TryPerformAction(RequestPerformActionEvent ev, EntityUid? sessionUser, bool skipDoAfter = false)
-    {
-        if (sessionUser is not { } user)
-            return false;
+        if (args.SenderSession.AttachedEntity is not { } user)
+            return;
 
         if (!_actionsQuery.TryComp(user, out var component))
-            return false;
+            return;
 
         var actionEnt = GetEntity(ev.Action);
 
         if (!TryComp(actionEnt, out MetaDataComponent? metaData))
-            return false;
+            return;
 
         var name = Name(actionEnt, metaData);
 
@@ -295,25 +314,25 @@ public abstract class SharedActionsSystem : EntitySystem
         {
             _adminLogger.Add(LogType.Action,
                 $"{ToPrettyString(user):user} attempted to perform an action that they do not have: {name}.");
-            return false;
+            return;
         }
 
         if (GetAction(actionEnt) is not {} action)
-            return false;
+            return;
 
         DebugTools.Assert(action.Comp.AttachedEntity == user);
         if (!action.Comp.Enabled)
-            return false;
+            return;
 
         var curTime = GameTiming.CurTime;
         if (IsCooldownActive(action, curTime))
-            return false;
+            return;
 
         // check for action use prevention
         var attemptEv = new ActionAttemptEvent(user);
         RaiseLocalEvent(action, ref attemptEv);
         if (attemptEv.Cancelled)
-            return false;
+            return;
 
         // Validate request by checking action blockers and the like
         var provider = action.Comp.Container ?? user;
@@ -325,18 +344,19 @@ public abstract class SharedActionsSystem : EntitySystem
         };
         RaiseLocalEvent(action, ref validateEv);
         if (validateEv.Invalid)
-            return false;
+            return;
 
-        if (HasComp<DoAfterComponent>(action) && !skipDoAfter)
+        Entity<ActionsComponent?> performer = (user, component);
+
+        if (HasComp<DoAfterComponent>(action))
         {
-            var attemptDoAfterEv = new ActionAttemptDoAfterEvent(user, ev);
+            var attemptDoAfterEv = new ActionAttemptDoAfterEvent(performer, action.Comp.UseDelay);
             RaiseLocalEvent(action, ref attemptDoAfterEv);
-            return false;
+            return;
         }
 
         // All checks passed. Perform the action!
-        PerformAction((user, component), action);
-        return true;
+        PerformAction(performer, action);
     }
 
     private void OnValidate(Entity<ActionComponent> ent, ref ActionValidateEvent args)
@@ -571,6 +591,7 @@ public abstract class SharedActionsSystem : EntitySystem
         ev.Performer = performer;
         ev.Action = action;
 
+        // TODO: This is where we'd add support for event lists
         if (!action.Comp.RaiseOnUser && action.Comp.Container is {} container && !_mindQuery.HasComp(container))
             target = container;
 
@@ -584,12 +605,11 @@ public abstract class SharedActionsSystem : EntitySystem
             return; // no interaction occurred.
 
         // play sound, start cooldown
-        if (ev?.Toggle == true)
+        if (ev.Toggle)
             SetToggled((action, action), !action.Comp.Toggled);
 
         _audio.PlayPredicted(action.Comp.Sound, performer, predicted ? performer : null);
 
-        // TODO: move to ActionCooldown ActionPerformedEvent?
         RemoveCooldown((action, action));
         StartUseDelay((action, action));
 
--- a/Content.Shared/Actions/SharedActionsSystem.cs
+++ b/Content.Shared/Actions/SharedActionsSystem.cs
@@ -110,7 +110,10 @@ public abstract class SharedActionsSystem : EntitySystem
         if (args.Cancelled)
             return;
 
-        PerformAction(args.Performer, ent);
+        // TODO: Needs a way to cancel DoAfter if there are no charges left
+
+        // Validate again for charges, blockers, etc
+        TryPerformAction(args.Input, args.Performer, skipDoActionRequest: true);
     }
 
     private void OnActionMapInit(Entity<ActionComponent> ent, ref MapInitEvent args)
@@ -292,20 +295,29 @@ public abstract class SharedActionsSystem : EntitySystem
     #region Execution
     /// <summary>
     ///     When receiving a request to perform an action, this validates whether the action is allowed. If it is, it
-    ///     will raise the relevant <see cref="InstantActionEvent"/>
+    ///     will raise the relevant action event
     /// </summary>
     private void OnActionRequest(RequestPerformActionEvent ev, EntitySessionEventArgs args)
     {
         if (args.SenderSession.AttachedEntity is not { } user)
             return;
 
+        TryPerformAction(ev, user);
+    }
+
+    /// <summary>
+    /// <see cref="OnActionRequest"/>
+    /// </summary>
+    /// <param name="skipDoActionRequest">Should this skip the initial doaction request?</param>
+    private bool TryPerformAction(RequestPerformActionEvent ev, EntityUid user, bool skipDoActionRequest = false)
+    {
         if (!_actionsQuery.TryComp(user, out var component))
-            return;
+            return false;
 
         var actionEnt = GetEntity(ev.Action);
 
         if (!TryComp(actionEnt, out MetaDataComponent? metaData))
-            return;
+            return false;
 
         var name = Name(actionEnt, metaData);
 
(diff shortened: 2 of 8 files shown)


Golubgik:
I love u

Content.Shared/Actions/Events/ActionDoAfterEvent.cs
```diff
@@ -0,0 +1,37 @@
...
+/// <summary>
+/// The event that triggers when an action doafter is completed or cancelled
+/// </summary>
+[Serializable, NetSerializable]
+public sealed partial class ActionDoAfterEvent : DoAfterEvent
+{
+    /// <summary>
+    /// <inheritdoc cref="ActionAttemptDoAfterEvent.Performer"/>
+    /// </summary>
+    [NonSerialized]
```
slarticodefast: why is this needed?
Are events even serialized anywhere?

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -30,11 +32,15 @@ public abstract class SharedActionsSystem : EntitySystem
...
     [Dependency] private   readonly SharedInteractionSystem _interaction = default!;
     [Dependency] private   readonly SharedTransformSystem _transform = default!;
+    [Dependency] private   readonly SharedDoAfterSystem _doAfter = default!;
+    [Dependency] private   readonly INetManager _netMan = default!;
 
     private EntityQuery<ActionComponent> _actionQuery;
     private EntityQuery<ActionsComponent> _actionsQuery;
     private EntityQuery<MindComponent> _mindQuery;
 
+    private const int DoAfterRepeatUseDelay = 0;
```
slarticodefast: why hardcode a 0 here?

Content.Shared/DoAfter/DoAfterComponent.cs
```diff
@@ -1,19 +1,133 @@
...
+    /// <summary>
+    /// What should the delay be reduced to after completion?
+    /// </summary>
+    [DataField]
+    public TimeSpan? DelayReduction;
+
+    // This region of fields are for setting parameters for a do after
+    // Eventually DoAfterArgs should be completely merged with this class
+    // Still requires some setup currently, but this way it hardcodes doafters a lot less
+    #region DoAfterArgsSettings
```
slarticodefast: Having all these datafields in this component seems weird.
The component is for allowing an entity to have DoAfters at all, and the `DoAfters` dictionary is storing all the doafters that are currently active.
But you are giving this component a second, completely different effect: starting a doafter in case the entity is an action.
You should instead make this a separate component.
And is should be a single datafield containing a `DoAfterArgs`, instead of duplicaing each entry inside.

slarticodefast (commented):
Not a full review, just a quick look.

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: See comment on line 107. This is also so it's not magic numbered in.

Content.Shared/Actions/Events/ActionDoAfterEvent.cs
keronshb: Multiple events are yes.

Content.Shared/DoAfter/DoAfterComponent.cs
keronshb: The issue with adding DoAfterArgs to something is that the ctor is expecting an event and entities, so I don't think that would work as expected. What ideas do you have to solve that?

Content.Shared/DoAfter/SharedDoAfterSystem.cs
```diff
@@ -29,12 +32,48 @@ public abstract partial class SharedDoAfterSystem : EntitySystem
...
+            NeedHand = actionDoAfter.NeedHand,
+            BreakOnHandChange = actionDoAfter.BreakOnHandChange,
+            BreakOnDropItem = actionDoAfter.BreakOnDropItem,
+            BreakOnMove = actionDoAfter.BreakOnMove,
+            BreakOnWeightlessMove = actionDoAfter.BreakOnWeightlessMove,
+            MovementThreshold = actionDoAfter.MovementThreshold,
+            DistanceThreshold = actionDoAfter.DistanceThreshold,
+            BreakOnDamage = actionDoAfter.BreakOnDamage,
+            DamageThreshold = actionDoAfter.DamageThreshold,
+            RequireCanInteract = actionDoAfter.RequireCanInteract
```
TheShuEd: I agree with Slarti, this is bad from a logical standpoint. Why can't we add DoAfterArgs as a serializable structure directly in DoAfterComponent?

TheShuEd (approved):
Except for the problem below everything looks good and works as designed. The problem with DoAfterArgs can also be solved by a separate PR

slarticodefast (dismissed):
Just marking my previous comment as a blocker:
`DoAfterComponent` is for keeping track of active doafters.
Giving actions a doafter should be moved into a separate component, otherwise all those new datafields will be redundant for any non-action entity.

Content.Shared/DoAfter/DoAfterArgsComponent.cs
```diff
@@ -0,0 +1,116 @@
...
+namespace Content.Shared.DoAfter;
+
+/// <summary>
+/// For setting DoAfterArgs on an entity level
+/// Would require some setup, will require a rework eventually
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+[Access(typeof(SharedDoAfterSystem))]
+public sealed partial class DoAfterArgsComponent : Component
+{
```
ScarKy0, line 13: The doAfter works by having this component directly on an action entity yes?
Why not just name this DoAfterAction then?

keronshb: Slart and I talked about it on discord, this isn't going to be exclusive to actions in the future. It's not necessarily locked to actions now, would require some setup to get it working on items too.

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -77,6 +85,50 @@ public override void Initialize()
         SubscribeAllEvent<RequestPerformActionEvent>(OnActionRequest);
     }
 
+    private void OnActionDoAfterAttempt(Entity<ActionComponent> ent, ref ActionDoAfterEvent args)
+    {
+        if (!TryComp<DoAfterComponent>(ent, out var actionDoAfter) || !TryComp<DoAfterArgsComponent>(ent, out var actionDoAfterArgsComp))
```
Princess-Cheeseballs: Unused.

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -77,6 +85,50 @@ public override void Initialize()
...
+
+        Entity<ActionComponent?>? action = (ent, ent);
+
+        // If this doafter is on repeat and was cancelled, start use delay as expected
+        if (args.Cancelled && actionDoAfterArgsComp.Repeat)
+        {
+            SetUseDelay(action, args.OriginalUseDelay);
+            RemoveCooldown(action);
+            StartUseDelay(action);
+            UpdateAction(ent);
```
Princess-Cheeseballs: I think you should just use (ent, ent) here instead of what you're currently doing.

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -77,6 +85,50 @@ public override void Initialize()
...
+            return;
+        }
+
+        args.Repeat = actionDoAfterArgsComp.Repeat;
+
+        // Set the use delay to 0 so this can repeat properly
+        if (actionDoAfterArgsComp.Repeat)
+        {
+            SetUseDelay(action, TimeSpan.FromSeconds(DoAfterRepeatUseDelay));
```
Princess-Cheeseballs: We have TimeSpan.Zero for this?

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -77,6 +85,50 @@ public override void Initialize()
...
+        if (TryPerformAction(args.Input, args.Performer, skipDoActionRequest: true))
+            return;
+
+        // This fixes the mispredicting but it doesn't fix the extra repeat
+        if (_netMan.IsClient)
+            return;
+
+        // Cancel this doafter if we can't validate the action
+        _doAfter.Cancel(args.DoAfter.Id, force: true);
+    }
```
Princess-Cheeseballs: DoAfters should be able to cancel properly with prediction in shared. You just need to track the DoAfterId ushort in a component.

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -309,10 +369,20 @@ private void OnActionRequest(RequestPerformActionEvent ev, EntitySessionEventArg
         };
         RaiseLocalEvent(action, ref validateEv);
         if (validateEv.Invalid)
-            return;
+            return false;
+
+        Entity<ActionsComponent?> performer = (user, component);
+
+        if (HasComp<DoAfterComponent>(action) && !skipDoActionRequest)
```
Princess-Cheeseballs: Honestly it might be better to have this before the IsCooldownActive check and manually handle the stuff you need? That way you don't need to bulldoze the cooldown. Make a generic ValidateAction method that raises the attempt event and Validate event.

Content.Shared/DoAfter/SharedDoAfterSystem.cs
```diff
@@ -29,12 +32,50 @@ public abstract partial class SharedDoAfterSystem : EntitySystem
...
+            DistanceThreshold = doAfterArgsComp.DistanceThreshold,
+            BreakOnDamage = doAfterArgsComp.BreakOnDamage,
+            DamageThreshold = doAfterArgsComp.DamageThreshold,
+            RequireCanInteract = doAfterArgsComp.RequireCanInteract
+        };
+
+        TryStartDoAfter(doAfterArgs, userDoAfter);
+    }
+
     private void OnUnpaused(EntityUid uid, DoAfterComponent component, ref EntityUnpausedEvent args)
```
Princess-Cheeseballs: This should be in actions system and not DoAfterSystem.

Content.Shared/DoAfter/SharedDoAfterSystem.cs
```diff
@@ -1,8 +1,11 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using Content.Shared.ActionBlocker;
+using Content.Shared.Actions.Components;
+using Content.Shared.Actions.Events;
 using Content.Shared.Damage;
 using Content.Shared.Hands.Components;
+using Content.Shared.Strip.Components;
```
Princess-Cheeseballs: Unused

Content.Shared/DoAfter/SharedDoAfterSystem.cs
```diff
@@ -1,8 +1,11 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using Content.Shared.ActionBlocker;
+using Content.Shared.Actions.Components;
```
Princess-Cheeseballs: Unused

Princess-Cheeseballs (changes requested):
Big review, didn't test in game but noticed some things.

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: What's unused about it? I need more of an explanation here.

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: Why would that be?

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: I'm not sure what you mean by this.

Content.Shared/Actions/SharedActionsSystem.cs
Princess-Cheeseballs: Specifically the !TryComp<DoAfterComponent>. Should've been more specific sorry.

Content.Shared/Actions/SharedActionsSystem.cs
Princess-Cheeseballs: That's usually how I see others do it but I guess it's fine?

Content.Shared/Actions/SharedActionsSystem.cs
Princess-Cheeseballs: Rather than overriding and saving the action cooldown, move the HasComp to before the cooldown check and then do whatever logic you need to and return. The whole thing of reducing the cooldown to 0 and then reapplying the proper cooldown later doesn't sit right with me.

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: OH I see what you mean. I've seen both ways. What I'm doing makes it way more clearly labeled on what has the nullable action comp and what just uses the ent.

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: So the action attempt doafter event isn't just for that. It's for setting all of the required values and attempting a doafter. IIRC we shouldn't even attempt to run the attempt action event or the validations if the cooldowns are active.

You need to set the cooldown to 0 and reapply it later for specific things to work, like repeating doafters. If you check if the cooldown is active after the fact you can't really achieve this.

Content.Shared/DoAfter/SharedDoAfterSystem.cs
keronshb: This requires the DoAfterComponent, so shouldn't it be here?

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: This is already tracked in the DoAfterComponent in the NextId and the Dictionary and it still does this, so I'm not sure what you mean.

Princess-Cheeseballs: Lemme debug this a bit but I ran into a similar prediction issue and the solution was the store the doAfterId in another component to track separately.

Content.Shared/Actions/SharedActionsSystem.cs
Princess-Cheeseballs: Ah that makes more sense then.

Content.Shared/DoAfter/SharedDoAfterSystem.cs
Princess-Cheeseballs: This is specifically for Action DoAfters so it should go in SharedActionsSystem. Also did you mean to change Entity\<DoAfterComponent> to Entity\<DoAfterArgsComponent> (?)

keronshb: Nah that still needs to stay as DoAfterComp because the action needs to have the DoAfterComp or it shouldn't be running at all, this is to relay it to the user.

Content.Shared/Charges/Systems/SharedChargesSystem.cs
```diff
@@ -20,11 +21,20 @@ public override void Initialize()
...
     }
 
+    private void OnActionDoAfterAttempt(Entity<LimitedChargesComponent> ent, ref DoAfterAttemptEvent<ActionDoAfterEvent> args)
+    {
+        var charges = GetCurrentCharges((ent.Owner, ent.Comp, null));
+
+        if (charges <= 0)
+            args.Cancel();
+    }
+
```
TheShuEd: This duplicates the check logic of SubscribeLocalEvent<LimitedChargesComponent, ActionAttemptEvent>(OnChargesAttempt);

I see a problem—that if the number of checks for action usage increases, all these checks will have to be duplicated.

keronshb: What would your solution be then?

TheShuEd: https://discord.com/channels/310555209753690112/1409845717328990279/1409847522155040951

Content.Shared/DoAfter/SharedDoAfterSystem.cs
ScarKy0: What I'd propose is to either keep it in ActionSystem, or create a new ActionSystem.DoAfters partial and put it there
having action code in doAfters is a bit meh, and partials are usually used to prevent things like this

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -77,6 +84,50 @@ public override void Initialize()
...
+        if (TryPerformAction(args.Input, args.Performer, skipDoActionRequest: true))
+            return;
+
+        // This fixes the mispredicting but it doesn't fix the extra repeat
+        if (_netMan.IsClient)
+            return;
+
+        // Cancel this doafter if we can't validate the action
+        _doAfter.Cancel(args.DoAfter.Id, force: true);
+    }
```
Princess-Cheeseballs: Move this to SharedActions.DoAfter

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -309,10 +368,20 @@ private void OnActionRequest(RequestPerformActionEvent ev, EntitySessionEventArg
...
+            return false;
+
+        Entity<ActionsComponent?> performer = (user, component);
+
+        if (HasComp<DoAfterComponent>(action) && !skipDoActionRequest)
+        {
+            var attemptDoAfterEv = new ActionAttemptDoAfterEvent(performer, action.Comp.UseDelay, ev);
+            RaiseLocalEvent(action, ref attemptDoAfterEv);
+            return false;
+        }
```
Princess-Cheeseballs: Is there any reason we need to raise an event here instead of just calling the one method listening to it directly? What else would be listening for this potentially.

Content.Shared/Charges/Systems/SharedChargesSystem.cs
Princess-Cheeseballs: I can see the reasoning why but I'm not sure if I agree with it. Is there a good reason why we need to be checking this every single tick and not just before and and after the doafter?

keronshb: What do you mean that it's checking every single tick?

Princess-Cheeseballs: DoAfterAttemptEvent is called by TryAttemptEvent in the DoAfterSystem which is called during the DoAfterSystem's Update method which is called every tick on active DoAfters.

Content.Shared/Actions/SharedActionsSystem.cs
keronshb: Since it's no longer in the doafter class this probably isn't needed any longer.

Content.Shared/Charges/Systems/SharedChargesSystem.cs
keronshb: So charges is technically checked before the doafter because of the charge attempt event that already exists. After I could try to see if it changes anything. With the removal of the extra event it might actually help with the prediction issue. If not, then it'd have to stay as is.

Princess-Cheeseballs: I fixed some prediction issues with an update I pushed.

keronshb:
That should address the remaining comments. There's no longer a 3rd event to check for cancelling actions and the duplicate check for charges is no longer needed since it's being checked anyway with the actionattemptevent.

@Princess-Cheeseballs  - can you also go into detail about what you did to fix the prediction issues? It would help me and anyone else reading this PR in the future to understand how to fix this.

Princess-Cheeseballs:
> That should address the remaining comments. There's no longer a 3rd event to check for cancelling actions and the duplicate check for charges is no longer needed since it's being checked anyway with the actionattemptevent.
> 
> @Princess-Cheeseballs - can you also go into detail about what you did to fix the prediction issues? It would help me and anyone else reading this PR in the future to understand how to fix this.

One of your networked events was passing an unserializable EntityUid, I converted it to a NetEntity so the client could properly handle it and run the same code as the server.

Content.Shared/Actions/SharedActionsSystem.cs
```diff
@@ -256,20 +259,29 @@ public void SetEnabled(Entity<ActionComponent?>? action, bool enabled)
...
     private void OnActionRequest(RequestPerformActionEvent ev, EntitySessionEventArgs args)
     {
         if (args.SenderSession.AttachedEntity is not { } user)
             return;
 
+        TryPerformAction(ev, user);
+    }
+
+    /// <summary>
+    /// <see cref="OnActionRequest"/>
```
Princess-Cheeseballs, line 273: Remember to fill out the parameters to avoid warnings.

Princess-Cheeseballs (changes requested):
One last thing

keronshb:
@Princess-Cheeseballs Done but can you also please leave a review to make changes instead of pushing them yourself? I had some things done in a specific way for readability such as setting a performer variable instead of just slapping it into a parameter, and it would be good to have documented why we're using a query in this case and swapping the method type of OnActionDoAfter

Princess-Cheeseballs:
> @Princess-Cheeseballs Done but can you also please leave a review to make changes instead of pushing them yourself? I had some things done in a specific way for readability such as setting a performer variable instead of just slapping it into a parameter, and it would be good to have documented why we're using a query in this case and swapping the method type of OnActionDoAfter

Query is significantly faster compared to a TryComp, and we're only using the event if an Action has DoAfterArgs so it's an optimization. 

I can request changes in the future next time but this PR is old enough I just wanted to see it complete.

On my phone so I can review later today.

keronshb:
> Query is significantly faster compared to a TryComp, and we're only using the event if an Action has DoAfterArgs so it's an optimization.

Where should we be using querys vs Trycomps? I know one instance could be in an iteration, but here it seems relatively separate.

> I can request changes in the future next time but this PR is old enough I just wanted to see it complete.

Yep that's fair, I'm just huge on documentation and understanding mindsets so I can help others with the same issues.

Princess-Cheeseballs:
> Where should we be using querys vs Trycomps? I know one instance could be in an iteration, but here it seems relatively separate.

Query's should be used in a system where we are commonly doing TryComps. ActionSystem is commonly TryComping for ActionComponent and ActionsComponent so it uses a query, in these cases it's about twice as fast compared to a TryComp. 

> Yep that's fair, I'm just huge on documentation and understanding mindsets so I can help others with the same issues.

Fair enough, I usually try to do this but if a PR starts to get old or if a change feels small I'll just push. If you have questions though don't be afraid to ask, it also helps remind me to review the PR.

Princess-Cheeseballs (approved):
New wiz content just a week (measured in Venusian weeks)

Princess-Cheeseballs:
Gonna wait for Ed to re-review before we should merge. I think you did their requested changes but I'd like them to take a look first.

keronshb:
@TheShuEd Yer up

TheShuEd (approved):
Yarr

PR #38263, opened 2025-06-11
Cleanup 2 `SpriteComponent.GetPrototypeTextures` uses

Tayrtahn:
## About the PR

Fixes 2 warnings from uses of the obsolete `SpriteComponent.GetPrototypeTextures` method.

## Why / Balance

## Technical details

Now that https://github.com/space-wizards/RobustToolbox/pull/6001 has been merged and reached content, we have the replacement system method for `GetPrototypeTextures`. This PR fixes 2 of the 3 uses of it.
- In `MappingState`, we already have a reference to `SpriteSystem` in `_sprite`, so we just use that.
- In `DummyIconTest`, we use the `System` method of the client test instance to get `SpriteSystem` (the standard way of getting entsystem references in tests).

The third use is in `ConstructionPlacementHijack` and I have a fix for it, but it needs `PlacementManager.EntityManager` to be changed from `internal` to `public` (which https://github.com/space-wizards/RobustToolbox/pull/5994 happens to do for other reasons).

## Requirements

### DIFF
--- a/Content.Client/Mapping/MappingState.cs
+++ b/Content.Client/Mapping/MappingState.cs
@@ -454,7 +454,7 @@ public sealed class MappingState : GameplayStateBase
         switch (prototype)
         {
             case EntityPrototype entity:
-                textures.AddRange(SpriteComponent.GetPrototypeTextures(entity, _resources).Select(t => t.Default));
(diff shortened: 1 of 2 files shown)


beck-thompson (approved):
2 more down!

PR #38264, opened 2025-06-11
Different health examine icon depending on health

Orsoniks:
## About the PR
Adds differing states for the Heart Examine icon, depending on the health of the mob you're examining. Different states for Alive, Crit and Dead.

## Why / Balance
To aid players in figuring out how much someone is hurt quickly, and looks nice

## Technical details
Added new icons, changed HealthExaminableSystem.cs

## Media
(Media outdated, living mob health is not shown as a fill anymore, only alive-crit-dead states are shown)

https://github.com/user-attachments/assets/091398a0-8469-4067-9a6d-70826e7a8429

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: The heart examine icon now changes depending on damage


Blackern5000:
Does the icon still change depending on poison/genetic damage? Those don't have an examination text afaik and it could give info you wouldn't otherwise know.
Not entirely against this change, but I feel like this does somewhat invalidate medhuds since it means you can instantly just run up to someone and examine them to see their status. Medhuds make it easier but it's still worth addressing if everyone needs to be able to instantly tell someone's health status with a single shift click

beck-thompson:
I also think it's giving away a little too much :thinking: 

I personally would prefer like a full health, damaged, and dead version so you actually need to use the medi scanner

Mixelz:
> I also think it's giving away a little too much 🤔

I'm not sure if its giving much more than what actual sprite is showing with how damaged they are. Based on the media the sprite of the urist seems to update at the same time the icon does and its already usually difficult to see those changes due to how muddled the pixel art gets when wearing various clothes.

I'd also put forward this being a generally good accessibility feature, more so with the clearer sprites.

Hitlinemoss:
I like this in theory, but not all damage types are actually visible on health examine itself so this shouldn't be an overall health indicator. (One specific edge cases I can think of is people with that one trait that prevents you from seeing your health indicator examining themselves as an easy workaround.)

Would probably be better for the heart icon to change based on _how severe the examine text is_ instead of just using the entity's raw health.

Orsoniks:
alternatively i could just make it change depending on alive/crit/dead

Hitlinemoss:
> alternatively i could just make it change depending on alive/crit/dead

That would also work, but I think there's benefit to being able to shift-click someone to tell "is this person heavily injured or are they just cold or thirsty or something"

Orsoniks:
> > alternatively i could just make it change depending on alive/crit/dead
> 
> That would also work, but I think there's benefit to being able to shift-click someone to tell "is this person heavily injured or are they just cold or thirsty or something"

yes, and ideally i'd make the icon only based off of damage types you can see, but i dont think theres an easy way to do that...

Hitlinemoss:
> yes, and ideally i'd make the icon only based off of damage types you can see, but i dont think theres an easy way to do that...

I don't know much about C#, but I assume you can probably just do some kind of check like "If damage type X is at value Y, show icon Z, unless a different valid damage type is even higher".

Orsoniks:
now only the state of the mob is shown (alive-crit-dead)

Issue #38267, opened 2025-06-11
StomachSystem is causing random debug asserts


slarticodefast:
I could not directly reproduce it yet, but I assume it is a similar bug we had with the UdderSystem that was fixed in https://github.com/space-wizards/space-station-14/pull/35314

Likely caused by https://github.com/space-wizards/space-station-14/pull/38164

@DrSmugleaf 
@Tayrtahn 

```
Unhandled exception. Robust.Shared.Utility.DebugAssertException: Exception of type 'Robust.Shared.Utility.DebugAssertException' was thrown.
   at Robust.Shared.Utility.DebugTools.Assert(Boolean condition) in D:\SS14Dev\space-station-14\RobustToolbox\Robust.Shared\Utility\DebugTools.cs:line 37
   at Content.Shared.Body.Systems.StomachSystem.Update(Single frameTime) in D:\SS14Dev\space-station-14\Content.Shared\Body\Systems\StomachSystem.cs:line 47       
   at Robust.Shared.GameObjects.EntitySystemManager.TickUpdate(Single frameTime, Boolean noPredictions) in D:\SS14Dev\space-station-14\RobustToolbox\Robust.Shared\GameObjects\EntitySystemManager.cs:line 318
   at Robust.Client.GameStates.ClientGameStateManager.PredictTicks(GameTick predictionTarget) in D:\SS14Dev\space-station-14\RobustToolbox\Robust.Client\GameStates\ClientGameStateManager.cs:line 537
   at Robust.Client.GameStates.ClientGameStateManager.ApplyGameState() in D:\SS14Dev\space-station-14\RobustToolbox\Robust.Client\GameStates\ClientGameStateManager.cs:line 472
   at Robust.Client.GameController.Tick(FrameEventArgs frameEventArgs) in D:\SS14Dev\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.cs:line 557
   at Robust.Client.GameController.<StartupContinue>b__65_0(Object sender, FrameEventArgs args) in D:\SS14Dev\space-station-14\RobustToolbox\Robust.Client\GameController\GameController.cs:line 248
   at Robust.Shared.Timing.GameLoop.Run() in D:\SS14Dev\space-station-14\RobustToolbox\Robust.Shared\Timing\GameLoop.cs:line 240
[...]

PR #38268, opened 2025-06-11
Fix `StomachSystem` PVS error

Tayrtahn:
## About the PR

Fixes a bug causing debug assert failures in `StomachSystem`.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/38267.

## Technical details

Exactly the same bug and exactly the same fix as https://github.com/space-wizards/space-station-14/pull/35314.
I'm not wild about using the code repetition here, but I'm not sure how to avoid it.

## Requirements


slarticodefast:
Bugfix speedrun 🫡

slarticodefast (approved):
I'm still not a fan of having to do this for every single shared system that handles solutions, but it works for now.

PR #38270, opened 2025-06-12
negative round start delays

IProduceWidgets:
## About the PR

removes u from the game so it can start already.
(Changes the uint requirement in the command to a signed int.)

## Why / Balance

I am not a patient person. negative good.

## Technical details

it just works. trust.

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- tweak: The `delaystart` command now accepts negative values, allowing to speed up the start of the round.


beck-thompson (changes requested):
is it funny? Yes but its kind of confusing and honestly a very handy command :laughing:

I think it would be better to just change it to modify start time or something and update the loc and stuff
image

slarticodefast:
So what exactly happens when you make it negative?

IProduceWidgets:
> So what exactly happens when you make it negative?

Currently it just starts the round if the timespan is less than the "round time" (quotes because it's pre-round). So if you add -1000 seconds it will start instantly.

IProduceWidgets:
> I think it would be better to just change it to modify start time or something and update the loc.

That's pretty doable. Id probably make a new command altogether though because thats setting (absolute) not delaying (relative).

slarticodefast:
> Currently it just starts the round if the timespan is less than the "round time" (quotes because it's pre-round). So if you add -1000 seconds it will start instantly.

But what does this PR do then? Can't you just set it to 0? I feel like this will just invite bugs to happen.
Please make the PR description a little more clear.

IProduceWidgets:
> > Currently it just starts the round if the timespan is less than the "round time" (quotes because it's pre-round). So if you add -1000 seconds it will start instantly.
> 
> But what does this PR do then? Can't you just set it to 0? I feel like this will just invite bugs to happen. Please make the PR description a little more clear.

The PR lets you set a negative delay by changing the uint requirement to only int.
The logic of the timer countdown is not effected.
Its literally a one character change.
Have you pulled it locally and identified a bug? Or is this the spectre of a hypothetical bug?

Content.Server/GameTicking/Commands/DelayStartCommand.cs
```diff
@@ -35,7 +35,7 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
                 return;
             }
 
-            if (!uint.TryParse(args[0], out var seconds) || seconds == 0)
+            if (!int.TryParse(args[0], out var seconds) || seconds == 0)
```
IProduceWidgets: -u

beck-thompson:
> But what does this PR do then? Can't you just set it to 0? I feel like this will just invite bugs to happen.
Please make the PR description a little more clear.

Yeah its confusing but the command just adds the input from the round time. So if there round starts in 100 seconds and you delaystart 75 it would start in 175 seconds! So putting in -75 changes the round start to 25 seconds.

It's useful but yeah its very unclear and should be its own command / renamed because people will 100% want to use this.

beck-thompson:
> The PR lets you set a negative delay by changing the uint requirement to only int.
The logic of the timer countdown is not effected.
Its literally a one character change.
Have you pulled it locally and identified a bug? Or is this the spectre of a hypothetical bug?

also make sure to remember the code of conduct! Its weird to put in negative numbers and slart is just trying to figure it out 😆

slarticodefast:
> The PR lets you set a negative delay by changing the uint requirement to only int.

Oh it adds an offset to the current time, thats what confused me. I thought it set the timer itself to a negative number.
Wouldn't it be better to have a command for directly setting the time?

> Or is this the spectre of a hypothetical bug?

If I remember correctly it selects the map at around 15 seconds, so we have to make sure that still happens in the correct order when we completely skip that. And maybe clamp the time at 0 to be safe? Not sure if it's needed though, would need to have a closer look at the code for that.

> removes u from the game so it can start already.

I read this as "removes *you* from the game"

IProduceWidgets:
> > The PR lets you set a negative delay by changing the uint requirement to only int.
> 
> Oh it adds an offset to the current time, thats what confused me. I thought it set the timer itself to a negative number. Wouldn't it be better to have a command for directly setting the time?

It does make sense to have that, but it would be a different command. You can already set the time via cvar, but you typically only want to do this ad-hoc in a single lobby and not overall.

> > Or is this the spectre of a hypothetical bug?
> 
> If I remember correctly it selects the map at around 15 seconds, so we have to make sure that still happens in the correct order when we completely skip that. And maybe clamp the time at 0 to be safe? Not sure if it's needed though, would need to have a closer look at the code for that.

It starts normally. There's a small hang the same as you see when using the `startround` command because the round start logic has to do all the map creation etc as you mention and that takes some time. fwiw, the timer logic doesn't account for this elsewise and you can observe the live servers messing this up every round start. I do not know of a way to account for this that wouldn't be misleading though. Its the halting problem. 

> > removes u from the game so it can start already.
>
> I read this as "removes _you_ from the game"

My apologies for the confusion.

little-meow-meow:
If there is any confusion: there is currently no way to reduce the round start timer. When you have a lobby full of players and you fuck up and accidentally delay the round for 30 minutes instead of 3 minutes, you will have to `golobby` and kick everyone's ready status out and potentially reset the selected map and gamemode.

IProduceWidgets:
I also don't intend to create an absolute offset command. I think this is better as a relative offset because there's less mental overhead and in 99% of cases you want a relative offset anyway.

Resources/Locale/en-US/game-ticking/game-ticker.ftl
```diff
@@ -4,7 +4,7 @@ game-ticker-start-round-cannot-start-game-mode-fallback = Failed to start {$fail
 game-ticker-start-round-cannot-start-game-mode-restart = Failed to start {$failedGameMode} mode! Restarting round...
 game-ticker-start-round-invalid-map = Selected map {$map} is inelligible for gamemode {$mode}. Gamemode may not function as intended...
 game-ticker-unknown-role = Unknown
-game-ticker-delay-start = Round start has been delayed for {$seconds} seconds.
+game-ticker-delay-start = Round start has been adjusted {$seconds} seconds.
```
beck-thompson: Could you change this to:
`Round start has been adjusted by {$seconds} seconds.`

beck-thompson (changes requested):

Resources/Locale/en-US/game-ticking/game-ticker.ftl
IProduceWidgets: done!

beck-thompson (dismissed):
Looks good! Could you make this a `LocalizedCommands`? That way you can get rid of the loc strings for description and help

Could you also just rename the command to `adjuststart` or something? Might as well fix it now 😆

beck-thompson:
See my messages in maintainers, I'm taking an extended break. I just pushed some stuff to make it decent - my only other concern is the `DelayStart` function not having any documentation (and its name is just confusing) that it can take negative numbers.

Feel free to revert I just thought I might as well get this in there!

IProduceWidgets:
I would not change the command name.
Its just going to break everyone's muscle memory for something that's primary purpose is allowing you to undo a fuck up.
If we were making this command originally I agree that name would be better, but at this stage we're just making the perfect the enemy of the good.

IProduceWidgets:
I dunno man, next time a server gets delayed for a thousand years, I guess its just going to stay that way.

ScarKy0:
Feeling epic so i wanna reopen this PR and finish it :godo:

ScarKy0 (approved):

ScarKy0:
relevant to the dismissal
image

Will toolshedify this once toolshed can be executed clientside, was also discussed with Simyon

PR #38273, opened 2025-06-12
fix covered pipe interaction popup

lzk228:
## About the PR

now instead of only checking if pipe is covered, it checks if you're using the right tool
previously it was like this (i am clicking with a bottle on a covered vent)
Screenshot_20250612_113123

## Why / Balance

bug

## Technical details

no

## Media

no

## Requirements

## Breaking changes

no

**Changelog**

very small so nod


lzk228:
oh damn, it seems like the fix is already on master, i am too late

perryprog:
Get speedrun, nerd. 🫡

PR #38276, opened 2025-06-12
Add wall-based ambient occlusion

metalgearsloth:
## Why / Balance
Most 13 servers have some form of AO.

Want to add some more CVARs for tweaking it. At the moment it draws to the entity render target rather than lighting as well, plus it also shows in space which may not be ideal and may need stencilling?

Also want to add it to graphics too.

ao-off

ao-on

**Changelog**
:cl:
- add: Added ambient occlusion to walls so they have shadows around them.


Hanzdegloker:
I absolutely love AO, great work!

This + some dropshadows would make ss14 look better and give allot of depth to everything.

deltanedas:
1 step closer to total space station victory its so peak

metalgearsloth:
Added cvars for functionality and graphics menu option.

Emisse:
looks great

Emisse (approved):

Chedd-Error:
Is there a way you can give an entity ambient occlusion without being a vision-blocking occluder?

Issue #31137, opened 2024-08-17
Puncturase heals bleeding


slarticodefast:
## Description
Found on discord by Binbag
https://discord.com/channels/310555209753690112/1274462223787364404

**Reproduction**
- hit yourself with a spear a few times
- inject yourself with puncturase using a hypospray
- hear a cauterization noise and get a "you feel your wounds painfully close" popup, which should only happen for heat damage

**Additional context**
Somehow in the OnDamageChanged function in the BloodstreamSystem the healing effect of the Puncturase in combination with the damaging effect of the bloodloss seems to be interpreted the same as positive heat damage, which cauterizes.

More details in the discord thread.

PR #38278, opened 2025-06-12
Cardiopulmonary Resuscitation

Orsoniks:
## About the PR
Adds CPR, which can be done on breathing mobs which are in the critical state.
CPR pumps every 0.7 seconds, and every pump restores 2 asphyxiation, 0.5 bloodloss but gives 0.2 blunt
The final pump (when the patient stops being a critical state) gives a small bonus boost of asphyx/bloodloss healing, as to make sure the person doesn't immediately collapse back into crit

Needs #31137 to be solved first, as otherwise CPR causes bleeding to get cauterized due to both blunt and healing being applied in the same specificer (and workarounds were discouraged)

## Why / Balance
CPR is a basic first-aid treatment in real life and there's no reason for it to not be in the game; it's cool as hell!
It also lets players do basic treatment for critically hurt people without access to medbay, which is a nice change. It's also not a straight buff to medical, as it does deal some blunt and it generally restores asphyx/bloodloss quite slowly, plus it requires the patient to be crit and the caretaker to stay still.

## Technical details
Adds CPRComponent, SharedCPRSystem, ClientCPRSystem, ServerCPRSystem and CPRLungeEvent. Also adds a new locale file and a new audio file (licenses done!)

## Media
(Media slightly outdated - icon is no longer an examine verb)

https://github.com/user-attachments/assets/f1fbde31-6a25-46c4-9b99-80718b591bf3

https://github.com/user-attachments/assets/0284110c-2acb-4249-af0b-41439b969bd7

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: You can now do CPR on mobs which are in a critical state, and prevent them from suffocating


Mixelz:
Is this an action that requires hands or would borgs be able to perform CPR as well?

SpaceLizard24:
I would like to note that this could be used offensively, by RRing someone with poison and CPR.

LaCumbiaDelCoronavirus:
> I would like to note that this could be used offensively, by RRing someone with poison and CPR.

who would do that if you can just use dex ((or even epi)) (or the million other ways to rr someone that are less convoluted than using cpr and poison)

Errant-4:
> > I would like to note that this could be used offensively, by RRing someone with poison and CPR.
> 
> who would do that if you can just use dex ((or even epi)) (or the million other ways to rr someone that are less convoluted than using cpr and poison)

Anyone who does not have dex at hand

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+using Robust.Shared.Serialization;
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Network;
+using Robust.Client.GameObjects;
+using Robust.Client.Animations;
+using System.Numerics;
+using Robust.Shared.Animations;
+using Content.Client.DoAfter;
+using Content.Shared.CPR;
+using Robust.Shared.Timing;
```
slarticodefast: sort alphabetically

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Content.Client.CPR;
+
+public sealed partial class ClientCPRSystem : SharedCPRSystem
+{
+    [Dependency] private readonly DoAfterSystem _doAfter = default!;
```
slarticodefast: not needed?

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+    [Dependency] private readonly AnimationPlayerSystem _animation = default!;
+    private const string LungeKey = "cpr-lunge";
```
slarticodefast:
```suggestion
    [Dependency] private readonly AnimationPlayerSystem _animation = default!;

    private const string LungeKey = "cpr-lunge";
```

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeNetworkEvent<CPRLungeEvent>(OnLunge);
```
slarticodefast:
```suggestion
        base.Initialize();

        SubscribeNetworkEvent<CPRLungeEvent>(OnLunge);
```

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+    }
+    /// <summary>
+    /// Used for playing the animation of other entities doing CPR
+    /// </summary>
+    /// <param name="evn">The lunge event</param>
+    public void OnLunge(CPRLungeEvent evn)
```
slarticodefast:
```suggestion
    /// <param name="args">The lunge event</param>
    public void OnCPRLunge(CPRLungeEvent args)
```
naming conventions

deltanedas:
its much faster to toolbox someone than get the 0.2 blunt per cpr attempt

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,80 @@
...
+        _animation.Play(user, lunge, LungeKey);
+    }
+
+    private Animation GetLungeAnimation(Vector2 direction)
+    {
+        const float endLength = 0.2f;
+
+        return new Animation
+        {
+            Length = TimeSpan.FromSeconds(endLength * 5f), // has to be long, otherwise cuts off in the middle for some reason
```
slarticodefast: Where is that magic number coming from? Can't you just combine it with the constant?

Content.Server/CPR/ServerCPRSystem.cs
```diff
@@ -0,0 +1,19 @@
+using Robust.Client.GameObjects;
+using Robust.Client.Animations;
+using System.Numerics;
+using Robust.Shared.Animations;
+using Content.Shared.CPR;
+using Robust.Shared.Player;
```
slarticodefast: sort

Orsoniks:
> Is this an action that requires hands or would borgs be able to perform CPR as well?

doesn't require hands, though i can change that if needed

Errant-4:
> its much faster to toolbox someone than get the 0.2 blunt per cpr attempt

~~No I think the idea was that you poison them, and then use CPR to remove the asphyxiation damage to make the poison build up to 200+ which makes them unrevivable since chems won't work on the dead (unless that already changed, there were some options in the work). I would argue that this could be considered an interesting alternative to blunting them until they gib, being a more quiet and inconspicuous, but also more complicated method. 
I don't think it's a big concern tbh~~

edit
Oh, yeah I guess even a tiny amount of blunt would make this more complicated to actually use

Content.Shared/CPR/CPRComponent.cs
```diff
@@ -0,0 +1,49 @@
...
+
+/// <summary>
+/// Stores info about the mob's CPR mechanics
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class CPRComponent : Component
+{
+    // this has to be initialized in the component because otherwise theres a really weird edge case bug where cpr causes a cauterize sound when bleeding
+    [DataField("heal")]
+    [ViewVariables(VVAccess.ReadWrite)]
```
slarticodefast: The name and VV are redundant here, read the warning your IDE gives you.
Also add a proper `<summary>` string for documentation.
Same below.

Content.Shared/CPR/CPRComponent.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Content.Shared.CPR;
+
+/// <summary>
+/// Stores info about the mob's CPR mechanics
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class CPRComponent : Component
+{
+    // this has to be initialized in the component because otherwise theres a really weird edge case bug where cpr causes a cauterize sound when bleeding
```
slarticodefast: that is a separate bug and should not be affected by whether or not you initialize it in the component or in yaml

Orsoniks:
> I would like to note that this could be used offensively, by RRing someone with poison and CPR.

not really, considering cpr gives blunt damage, it would be exceedingly unlikely to do that
you'd have to constantly use a bruise kit to heal the blunt damage (which wont really work anyway since the blunt gets applied way more often, meaning when the person dies they'll very very likely still have blunt on them)
and if you are to use a chem to remove the blunt damage then you might aswell just use dex to ignore cpr completely
not to mention to do the cpr you have to stay still for however long itll take them to die, which is extremely risky and there are much much more convenient ways to RR someone

Content.Shared/CPR/CPRComponent.cs
```diff
@@ -0,0 +1,49 @@
+using Content.Shared.Damage;
+using Robust.Shared.Audio;
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.CPR;
+
+/// <summary>
+/// Stores info about the mob's CPR mechanics
```
slarticodefast: Add better documentation.
What does this do, and is it added to the mob receiving CPR or the one doing it?

Content.Shared/CPR/CPRLungeEvent.cs
```diff
@@ -0,0 +1,18 @@
+using System.Numerics;
+using Robust.Shared.Serialization;
+
+/// <summary>
+/// Data for CPR animations
+/// </summary>
+namespace Content.Shared.CPR;
+
+[Serializable, NetSerializable]
+public sealed partial class CPRLungeEvent : EntityEventArgs
```
slarticodefast: should be a record struct
and use a primary constructor

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+using Content.Shared.Verbs;
+using Robust.Shared.Utility;
+using Content.Shared.DoAfter;
+using Content.Shared.Interaction;
+using Robust.Shared.Serialization;
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Network;
+using Robust.Shared.Timing;
+using Content.Shared.Popups;
+using Content.Shared.IdentityManagement;
```
slarticodefast: sort

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+    [Dependency] private readonly SharedPopupSystem _popup = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<CPRComponent, GetVerbsEvent<ExamineVerb>>(OnGetExamineVerbs);
+        SubscribeLocalEvent<CPRComponent, CPRDoAfterEvent>(OnDoCPREvent);
+    }
+    public void OnDoCPREvent(Entity<CPRComponent> ent, ref CPRDoAfterEvent args)
```
slarticodefast:
```suggestion
    }

    public void OnCPRDoAfter(Entity<CPRComponent> ent, ref CPRDoAfterEvent args)
```
fix spacing and naming conventions

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (args.Handled || args.Cancelled)
+            return;
+        if (!TryComp<DamageableComponent>(ent, out var damage) || !TryComp<CPRComponent>(ent, out var cpr) || !TryComp<MobThresholdsComponent>(ent, out var thresholds))
+            return;
+        DoLunge(args.User);
```
slarticodefast:
```suggestion
        if (args.Handled || args.Cancelled)
            return;

        if (!TryComp<DamageableComponent>(ent, out var damage) || !TryComp<CPRComponent>(ent, out var cpr) || !TryComp<MobThresholdsComponent>(ent, out var thresholds))
            return;

        DoLunge(args.User);
```

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+        if (_net.IsServer)
+        {
+            _audio.PlayPvs(cpr.Sound, ent.Owner);
```
slarticodefast: use PlayPredicted instead of IsServer

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+        _damage.TryChangeDamage(ent, cpr.Heal, interruptsDoAfters: false, damageable: damage, ignoreResistances: true);
+        _damage.TryChangeDamage(ent, cpr.Damage, interruptsDoAfters: false, damageable: damage, ignoreResistances: true);
```
slarticodefast: can't these be combined?

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+        if (thresholds.CurrentThresholdState != MobState.Critical)
+        {
+            _damage.TryChangeDamage(ent, cpr.BonusHeal, interruptsDoAfters: false, damageable: damage, ignoreResistances: true);
+        }
+
+        args.Repeat = thresholds.CurrentThresholdState == MobState.Critical;
+        args.Handled = true;
+    }
+
+    public abstract void DoLunge(EntityUid user);
```
slarticodefast: add documentation to public methods

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (!TryComp<DamageableComponent>(uid, out var damage) || !TryComp<CPRComponent>(uid, out var cpr) || !TryComp<MobThresholdsComponent>(uid, out var thresholds) || !TryComp<MobThresholdsComponent>(args.User, out var myThresholds))
+            return;
+
+        if (thresholds.CurrentThresholdState != MobState.Critical)
+            return;
+
+
+        // low interaction range
+        var inRange = _interactionSystem.InRangeUnobstructed(args.User, args.Target, SharedInteractionSystem.InteractionRange * 0.25f);
```
slarticodefast: no magic numbers, this should be a datafield

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        var doAfterEventArgs = new DoAfterArgs(
+            EntityManager,
+            args.User,
+            TimeSpan.FromSeconds(0.7f),
```
slarticodefast: same

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+            DamageThreshold = 5,
+            BlockDuplicate = true,
+            RequireCanInteract = true
+        };
+
+        var verb = new ExamineVerb()
+        {
+            Act = () =>
+            {
+                if (_doAfter.TryStartDoAfter(doAfterEventArgs) && _net.IsServer)
```
slarticodefast: Why only do this on the server?
All new code has to be predicted.

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    _popup.PopupEntity(Loc.GetString("cpr-start", ("person", Identity.Entity(args.User, EntityManager)), ("target", Identity.Entity(uid, EntityManager))), uid, PopupType.Medium);
```
slarticodefast: PopupPredicted
You should also have a separate loc string for the local player vs others observing you.
You start...
Urist McHands starts...

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,115 @@
...
+            Message = inRange ? null : Loc.GetString("cpr-verb-text-disabled"),
+            Icon = new SpriteSpecifier.Texture(new("/Textures/Interface/VerbIcons/cpr.svg.192dpi.png"))
+        };
+
+        args.Verbs.Add(verb);
+    }
+}
+
+[Serializable, NetSerializable]
+public sealed partial class CPRDoAfterEvent : DoAfterEvent
```
slarticodefast: documentation

Resources/Audio/Effects/attributions.yml
```diff
@@ -246,3 +246,8 @@
   license: "CC-BY-SA-3.0"
   copyright: "Created by mattroks101 for Bee Station. cig_snuff converted to mono"
   source: "https://github.com/BeeStation/BeeStation-Hornet/pull/29"
+
+- files: ["cprpulse.ogg"]
+  license: "CC-BY-SA-3.0"
+  copyright: "Created by mixing some Effects sounds together by Orsoniks"
```
slarticodefast: Are the effects you used for this under any licence?

slarticodefast (changes requested):
This should use a verb instead of being triggered by an icon in the examination window in my opinion. Otherwise it would be inconsistent with other controls.

Resources/Audio/Effects/attributions.yml
Orsoniks: the effects are from the effects folder the file itself is in, the license is their license

Content.Client/CPR/ClientCPRSystem.cs
Orsoniks: what exactly are you referring to?

Resources/Audio/Effects/attributions.yml
slarticodefast: Can you add a list of the ones that were used then?
Just to make sure we attribute everything correctly.

Content.Client/CPR/ClientCPRSystem.cs
slarticodefast: The 5f

Orsoniks:
i'm working on fixing all the requested changes!

Resources/Audio/Effects/attributions.yml
Orsoniks: yes

Content.Client/CPR/ClientCPRSystem.cs
Orsoniks: first number i found that is high enough for the animation to not end early for some reason

Content.Shared/CPR/CPRComponent.cs
Orsoniks: no, i tested it before with it initialized in script/yaml and it did the sound when done in yaml, and didnt when done in script - with the exact same values

Content.Shared/CPR/CPRLungeEvent.cs
Orsoniks: not sure what this means exactly-
this code is mostly a copy of the melee lunge event, which is why i assumed it was fine .. ?

Content.Shared/CPR/SharedCPRSystem.cs
Orsoniks: playpredicted plays multiple times over no matter what i tried, using IsServer is the only thing that worked

Content.Shared/CPR/SharedCPRSystem.cs
Orsoniks: no, because it otherwise triggers a bug present in the bloodstream system, related to bleed speed (this has been in the game for extremely long for some reason)

Content.Shared/CPR/CPRComponent.cs
slarticodefast: That would be a major bug with yaml serialization. It should not matter if you define it here or in yaml.

Orsoniks: correct! it shouldnt! but it does because there is a major bug with yaml serialization apparently - i can link the discovery on discord if needed

Content.Shared/CPR/CPRLungeEvent.cs
slarticodefast: Melee weapon code is very old, some C# stuff did not exist yet then. Structs are better dor performance.

Content.Shared/CPR/SharedCPRSystem.cs
slarticodefast: Playpredicted should be exactly preventing that from happening. Check if you used it correctly. Making the sound not predicted is a bandaid you should not be doing.

Content.Shared/CPR/SharedCPRSystem.cs
slarticodefast: Do we have an open issue for that? I would rather fix it first than introduce bandaid workarounds.

Content.Shared/CPR/CPRComponent.cs
slarticodefast: Would be great if you can open an issue for it then.
Best with reproduction steps.

Content.Shared/CPR/CPRLungeEvent.cs
Orsoniks: it needs to be a class, as otherwise RaiseNetworkEvent() doesn't work with it

Content.Shared/CPR/SharedCPRSystem.cs
Orsoniks: i do not know what else to do in this case

Content.Shared/CPR/SharedCPRSystem.cs
Orsoniks: i was told someone will open up an issue for this soon, though i dont know when

Dragonjspider:
So what would happen if multiple people started doing CPR on the same person? I assume it wouldn't work, but I ask to make sure if it was accounted for

Orsoniks:
> So what would happen if multiple people started doing CPR on the same person? I assume it wouldn't work, but I ask to make sure if it was accounted for

Ah, that would multiply the effect... I'm not sure how I'd account for that to be honest

Pumkin69:
This should probably require hands although i dont know the negative side effects that would cause if any.

Entvari:
As much sense as it would make for it to require hands, a mouse or revenant performing CPR is a genuinely hilarious concept

K-Dynamic:
CPR should only remove asphyx when the mob could be breathing normally; patients should either be in a room with air or have internals on

Orsoniks:
> CPR should only remove asphyx when the mob could be breathing normally; patients should either be in a room with air or have internals on

Not sure how can I detect that

Orsoniks:
Made only one person able to do CPR on someone at a time, added a hands requirement, added alt-click support and cleaned up the code overall

Hrosts:
Would be cool to make the DoAfter delay around 0.6-0.55 seconds to fit the IRL CPR speed. I'm a bit afraid going faster would make it look weird though.

Orsoniks:
> Would be cool to make the DoAfter delay around 0.6-0.55 seconds to fit the IRL CPR speed. I'm a bit afraid going faster would make it look weird though.

I am aware that CPR is usually done at 120 pumps per minute, but the slower speed feels nicer ingame

Errant-4:
> > CPR should only remove asphyx when the mob could be breathing normally; patients should either be in a room with air or have internals on
> 
> Not sure how can I detect that

It would have to be a rewrite of a fairly integral part of this: instead of removing asphyx damage, it would need to make it so that breathing is not blocked for the patient even in crit, if CPR is being performed on them. Breathing on its own will gradually remove asphyx damage as normal, if the gasmix being breathed is good. Or keep hurting if the gas is bad.

Making it work like that would be better, but is imo not a requirement for this PR, it can be a future improvement by either you or someone else.

Errant-4:
Two questions
The default CPR rate is mechanically unoptimal. It removes airloss faster than it builds up, so some of the CPR actions are "wasted", but still cause their blunt damage. Someone who knows what they are doing could start and stop CPR in "bursts" to make the whole process cause less blunt damage overall. This can be seen either as a noobtrap or as rewarding player knowhow, do we want to keep it like this?

It restores bloodloss, is that just because of the linked issue? It seems to be resolved now, is this still necessary? It feels a bit weird that you can restore blood loss just by CPR, but I guess it's not like it couldn't use the buff

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,66 @@
+using Content.Shared.CPR;
```
Errant-4: This file and class should be renamed to just CPRsystem, we do not put Server or Client in the name of the file, that's only for Shared systems

Content.Server/CPR/ServerCPRSystem.cs
```diff
@@ -0,0 +1,15 @@
+using Content.Shared.CPR;
```
Errant-4: This file and class should be renamed to just CPRsystem, we do not put Server or Client in the name of the file, that's only for Shared systems

Content.Client/CPR/ClientCPRSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+                    KeyFrames =
+                    {
+                        new AnimationTrackProperty.KeyFrame(new Vector2(0f,0f), 0f),
+                        new AnimationTrackProperty.KeyFrame(direction.Normalized() * 0.12f, endLength * 0.2f),
+                        new AnimationTrackProperty.KeyFrame(direction.Normalized() * 0.16f, endLength * 0.4f),
+                        new AnimationTrackProperty.KeyFrame(new Vector2(0f,0f), endLength)
+                    }
+                }
+            }
+        };
```
Errant-4: This should be unnested to make it easier to read or edit later. Can't we just use middlemen variables?

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// </summary>
+    /// <param name="recipient">The one receiving CPR</param>
+    /// <param name="giver">The one giving CPR</param>
+    public bool CanDoCPR(EntityUid recipient, EntityUid giver)
+    {
+        if (!TryComp<CPRComponent>(recipient, out var cpr) || !TryComp<MobThresholdsComponent>(recipient, out var thresholds) || !TryComp<MobThresholdsComponent>(giver, out var myThresholds))
+            return false;
+
+        if (thresholds.CurrentThresholdState != MobState.Critical || myThresholds.CurrentThresholdState == MobState.Critical || myThresholds.CurrentThresholdState == MobState.Dead)
+            return false;
```
Errant-4: Break these into lines for easier reading, our max line length is 120 which is in some places exceeded, but here it's very simple to resolve

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// <param name="args">DoAfter arguments</param>
+    public void OnCPRDoAfter(Entity<CPRComponent> ent, ref CPRDoAfterEvent args)
+    {
+        if (args.Handled || args.Cancelled)
+            return;
+
+        if (!CanDoCPR(ent, args.User))
+            return;
+
+        if (!TryComp<DamageableComponent>(ent, out var damage) || !TryComp<CPRComponent>(ent, out var cpr) || !TryComp<MobThresholdsComponent>(ent, out var thresholds))
```
Errant-4: Line breaks

Content.Shared/CPR/SharedCPRSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+            BreakOnMove = true,
+            BlockDuplicate = true,
+            RequireCanInteract = true,
+            NeedHand = true
+        };
+
+        // try starting CPR
+        if (CanDoCPR(recipient, giver) && InRangeForCPR(recipient, giver) && _doAfter.TryStartDoAfter(doAfterEventArgs))
+        {
+            _popup.PopupPredicted(Loc.GetString("cpr-start-you", ("target", Identity.Entity(recipient, EntityManager))), Loc.GetString("cpr-start", ("person", Identity.Entity(giver, EntityManager)), ("target", Identity.Entity(recipient, EntityManager))), giver, giver, PopupType.Medium);
```
Errant-4: Please put the two getstrings in variables and call those, so it's not all one massive line

Errant-4 (changes requested):

Content.Shared/CPR/SharedCPRSystem.cs
Orsoniks: i find line breaks to be very ugly and hard to read for if statements

Orsoniks:
> Two questions The default CPR rate is mechanically unoptimal. It removes airloss faster than it builds up, so some of the CPR actions are "wasted", but still cause their blunt damage. Someone who knows what they are doing could start and stop CPR in "bursts" to make the whole process cause less blunt damage overall. This can be seen either as a noobtrap or as rewarding player knowhow, do we want to keep it like this?
> 
> It restores bloodloss, is that just because of the linked issue? It seems to be resolved now, is this still necessary? It feels a bit weird that you can restore blood loss just by CPR, but I guess it's not like it couldn't use the buff

CPR is meant to bring down asphyx faster than it goes up - that's the point
It restores bloodloss as it is still in the airloss category and without it CPR would be a bit useless (realistically, doing CPR on someone with bloodloss would deliver some more oxygen to their limbs and organs, thus it does make sense in my mind)

Content.Client/CPR/ClientCPRSystem.cs
Orsoniks: how? also, the melee animation system does it the same exact way and it's fine...

Errant-4:
> CPR is meant to bring down asphyx faster than it goes up - that's the point 

Okay that's fair, but allow me to rephrase what I meant. CPR removes asphyx so fast, that you can meaningfully reduce the amount of blunt damage inflicted while keeping asphyx from building up, by only doing CPR for about 5 seconds, then waiting for asphyx to build up for 30 seconds. Rinse repeat. I'm sure the Optimal Timing will be worked out by medical in a day. So that leaves people who do "automatic CPR" causing more damage to the patient than necessary. But maybe this is all actually appropriate and kinda works out. Idunno

Orsoniks:
i think its fine

Errant-4 (dismissed):
As per the maintainer poll in the discussion thread, we are requesting that the direct asphyxiation healing be removed, and instead CPR should just make it possible for a crit person to breathe through the regular system (which will prevent further asphyx damage and gradually remove existing one)

Do you want to look into SharedRespiratorSystem and make this change yourself? If you don't want to, I volunteer to make a PR to your working branch which makes this change

(We might also end up asking that the keypresses be manual and require (somewhat) proper "timing", but that part had less consensus so I'm not sure how much it is a "requirement")

Orsoniks:
> Do you want to look into SharedRespiratorSystem and make this change yourself? If you don't want to, I volunteer to make a PR to your working branch which makes this change

Sure, go ahead and make a PR then

Errant-4:
I remembered I can just... push it. So respirationsystem is now connected. This means that removing the asphyx damage is slower than before. Bloodloss is still removed directly, same as before.

Slightly puzzling that small amounts of airloss damage still somehow get applied to a crit mob even while it's breathing via CPR. Not sure where that's coming from, but it's getting removed by the airloss heal of normal breathing faster than it's added

Orsoniks:
changes look nice

RedSpyy:
Very good PR!

Don't forget to remove defibrillators from maps outside the medical bay. But this is most likely in a separate PR.

Errant-4:
(Forum thread: https://forum.spacestation14.com/t/cardiopulmonary-resuscitation-38278/21481 )

Errant-4:
I still face some issues with the popup messages, but apart from that the design is done for testing purposes

superjj18:
I got some additions that may make for interesting gameplay:

1. CPR on a **bleeding** patient loses blood faster, and causes blood loss damage.  Your accelerating bloodflow so CPR before stopping bleeding is medical malpractice that very likely could lead to the death of the patient.(maybe in the future there could be an additional "hold the wound and stop the bleeding" option so that the patient isn't just screwed unless the rescuer has cloth, bandage, or emergency syringe)

2. CPR on a **dead** patient slows or stalls their rot/decomposition timer as you're giving the brain blood and delaying brain death.(maybe change rot to brain death?  Could become relevant to the borging process)

Princess-Cheeseballs:
Sorry for the wait, are you working on this still?

Errant-4:
It was done, as far as I can remember. Or do you just want the merge conflict resolved?
(I never actually got much of a response on the game design of the current proposal without auto-repeat)

Princess-Cheeseballs:
> It was done, as far as I can remember. Or do you just want the merge conflict resolved? (I never actually got much of a response on the game design of the current proposal without auto-repeat)

There are some changes I would like to request from review, but if they're not working on this I'll just add it to the backburner and push them myself some day.

Errant-4:
Last I heard, they are not working on this anymore. But I'm willing, if there is anything to be done still. What did you have in mind?

Princess-Cheeseballs:
> Last I heard, they are not working on this anymore. But I'm willing, if there is anything to be done still. What did you have in mind?

I can leave a bigger review tomorrow but having a TryComp for assisted breathing hardcoded into the Update loop feels bad, especially when the Inhale and Exhale methods are public and raise events. 

Aside from that the code looks good but I haven't tested in game or scrutinized it that bad.

Basically I think CPR should just call Inhale and Exhale events and the Update loop should check a cached value for "are we breathing" that updates based on mobstate/CPR or w/e. 

I'm eepy so I might have better thoughts tomorrow.

LanceSmites328:
As a player I would absolutely love to see this in. It feels almost nonsensical that we're a space-faring people and the basics of First-Aid/CPR is lost on us.

FairlySadPanda:
This was given conceptual approval via maintainer discussion, 30/10/25.

Errant-4:
Will move the assisted breathing code out of RespiratorySystem as discussed

Content.Shared/CCVar/CCVars.Misc.cs
```diff
@@ -125,4 +125,11 @@ public static readonly CVarDef<float>
     /// </summary>
     public static readonly CVarDef<float> PlaytimeMinutesToday =
         CVarDef.Create("playtime.minutes_today", 0f, CVar.CLIENTONLY | CVar.ARCHIVE);
+
+    /// <summary>
+    ///     Enabled: CPR action will auto-repeat as long as the target is still crit
+    ///     Disabled: CPR has to be performed by repeated manual interaction
+    /// </summary>
+    public static readonly CVarDef<bool> CprRepeat =
```
Samuka-C, line 133:
```suggestion
    [CVarControl(AdminFlags.VarEdit)]
    public static readonly CVarDef<bool> CprRepeat =
```

I would suggest letting admins change this cvar.

Content.Shared/Cpr/AssistedRespirationComponent.cs
```diff
@@ -0,0 +1,17 @@
...
+/// The entity has their breathing assisted by external means, and can breathe even in crit
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class AssistedRespirationComponent : Component
+{
+    /// <summary>
+    /// The time when the assist will expire
+    /// It will be removed on the next respiration attempt
+    /// </summary>
+    [DataField, AutoNetworkedField]
```
Samuka-C, line 15:
```suggestion
    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer)), AutoPausedField]
```

the standard for this kind of datafield

Content.Shared/Cpr/AssistedRespirationComponent.cs
```diff
@@ -0,0 +1,17 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Cpr;
+
+/// <summary>
+/// The entity has their breathing assisted by external means, and can breathe even in crit
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
```
Samuka-C, line 8:
```suggestion
[RegisterComponent, NetworkedComponent, AutoGenerateComponentPause]
```

because of the AutoPausedField

Content.Shared/Cpr/CprComponent.cs
```diff
@@ -0,0 +1,35 @@
...
+    public DamageSpecifier Change;
+
+    /// <summary>
+    /// Healing is applied when the CPR finishes with the mob returning from a critical state - these values should be negative
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public DamageSpecifier BonusHeal;
+
+    [DataField, AutoNetworkedField]
+    public SoundSpecifier? Sound;
```
Samuka-C, line 28: add summary

Content.Shared/Cpr/CprComponent.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    [DataField, AutoNetworkedField]
+    public EntityUid? LastCaretaker;
```
Samuka-C, line 31: add summary

Content.Shared/Cpr/CprComponent.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    [DataField, AutoNetworkedField]
+    public TimeSpan LastTimeGivenCare = TimeSpan.Zero;
```
Samuka-C, line 34: add summary

Content.Shared/Cpr/CprComponent.cs
```diff
@@ -0,0 +1,35 @@
...
+    [DataField, AutoNetworkedField]
+    public DamageSpecifier BonusHeal;
+
+    [DataField, AutoNetworkedField]
+    public SoundSpecifier? Sound;
+
+    [DataField, AutoNetworkedField]
+    public EntityUid? LastCaretaker;
+
+    [DataField, AutoNetworkedField]
```
Samuka-C, line 33:
```suggestion
    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer)), AutoPausedField]
```

Content.Shared/Cpr/CprComponent.cs
```diff
@@ -0,0 +1,35 @@
...
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Cpr;
+
+/// <summary>
+/// Stores info about the mob's CPR mechanics
+/// An entity with this Component can have CPR done on them
+/// Has fields that specify how much this mob should be healed when CPR is done on them and the pump sound
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
```
Samuka-C, line 12:
```suggestion
[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, AutoGenerateComponentPause]
```

Content.Shared/Cpr/SharedCprSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+        cpr.LastTimeGivenCare = Timing.CurTime;
+
+        args.Repeat = thresholds.CurrentThresholdState == MobState.Critical && _cprRepeat;
+        args.Handled = true;
+    }
+    /// <summary>
+    /// Makes a user do the CPR Lunge animation
+    /// </summary>
+    /// <param name="user">The entity to animate</param>
+    public abstract void DoLunge(EntityUid user);
```
Samuka-C, line 159:
```suggestion
    public virtual void DoLunge(EntityUid user) { } 
```

use virtual instead of abstract, then add `override` in the implementation of DoLunge

Content.Shared/Cpr/SharedCprSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+    public const float CprDoAfterDelay = 0.7f;
+    public const float CprAnimationLength = 0.2f;
+    public const float CprAnimationEndTime = 1f; // This is set to much higher than the actual animation length to avoid it stopping prematurely, as it did in testing. Shouldnt affect anything
+
+    // This determines how long the effects of a CPR interaction last, and so how often it needs to be repeated.
+    // Damage will automatically scale to the less frequent "hits", to retain the overall dps of resuscitation
+    public const float CprManualEffectDuration = 5f;
+
+    // The total time window for a "correctly timed" CPR interaction. Outside this, the player will be told they are too fast/slow.
+    public const float CprManualThreshold = 1.5f;
```
Samuka-C, line 44: Why aren't those values in the CPRComponent instead of being magical numbers here?

Content.Shared/Cpr/SharedCprSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+    public bool CanDoCpr(EntityUid recipient, EntityUid giver)
+    {
+        if (!TryComp<CprComponent>(recipient, out var cpr) ||
+            !TryComp<MobThresholdsComponent>(recipient, out var thresholds) ||
+            !TryComp<MobThresholdsComponent>(giver, out var myThresholds))
+            return false;
+
+        if (thresholds.CurrentThresholdState != MobState.Critical ||
+            myThresholds.CurrentThresholdState == MobState.Critical ||
+            myThresholds.CurrentThresholdState == MobState.Dead)
```
Samuka-C, line 80: could just check if the myThresholds.CurrentThresholdState is different from Alive

Content.Shared/Cpr/SharedCprSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+            TimeSpan.FromSeconds(CprDoAfterDelay),
+            new CprDoAfterEvent(),
+            recipient,
+            giver
+            )
+        {
+            BreakOnMove = true,
+            BlockDuplicate = true,
+            RequireCanInteract = true,
+            NeedHand = true
```
Samuka-C, line 180: once https://github.com/space-wizards/space-station-14/pull/44512 is merged. perhaps add
`NeedFreeHand`

since i assume is kinda impossible to do CPR while holding a wrench or some other item

Samuka-C (changes requested):
Not a full review, but some things I noticed

JiyuSoftware:
I wouldn't make the player giving CPR move up and down. When I see that all I can think of is t-bagging on a corpse in back in the day on Halo CE.

I think just a static do-after is sufficient, like most other actions in the game.

Issue #34285, opened 2025-01-08
Some handheld devices need to activate again after closing or have redundant activation feature


K-Dynamic:
## Description

See title. This makes some devices feel clunky and affects user experience.

**Reproduction**

***Type 1 - Devices that need to activate again after closing***
- Gas analyzers
- Anomaly scanners

1. Obtain a device (gas analyzer or anomaly scanner).
2. Press z or in-hand action key to toggle UI.
3. Close UI manually with the 'x' button on the top-right.
4. Press z or in-hand action key again.
5. Notice UI doesn't open unless you repeat z or in-hand action once more.

***Type 2 - Devices with activation that don't affect the device***
- Health analyzers
- Station maps (portable)
- Handheld crew monitor

1. Obtain a device (health analyzer or portable station map)
2. Use drop-down menu on device and press 'activate.'
3. Notice lack of changes to device function. Press z or in-hand action key to toggle UI.
4. Use drop-down menu on device and press 'deactivate.'
5. UI closes as expected.

However, 'activate' doesn't do anything, hence it should be classified as an issue. Toggling menu meanwhile will always enable 'deactivate,' which could be considered redundant due to the 'toggle UI' option in the drop-down menu. What activation should do is turn the device on or off to save power, such as with defibrillators.

**Additional context**

Notably Type 2 affects devices with power cells, while Type 1 affects devices without them.

murolem:
Bump.

Gas analyzer sometimes opening and closing immediately is really annoying.

PR #38280, opened 2025-06-12
Remove redundant activation verbs

themias:
## About the PR

This PR removes a redundant "activate/deactivate" verb from health analyzers, handheld station maps, handheld mass scanners, and handheld crew monitors.

This resolves part 2 of #34285 

(The issue is distinct from part 1, so I am tackling that as part of a separate PR.)

## Why / Balance

These items all have a "toggle UI" right-click option that performs as expected, so having a second verb that means to do the same thing is clutter at best. Most users would just left-click or use E/Z buttons to toggle these, but for those that might right-click, we don't need two options.

## Technical details

These four handheld devices' ItemToggleComponent are already tightly-coupled with the ActivatableUISystem.
This change sets their parent BaseHandheldComputer "onActivate: false" which will not display the verb and leaves the handling to the above system. This also parents the powered health analyzer to BaseHandheldComputer instead of it duplicating this functionality itself.

## Media

(I realized during testing that this affects handheld mass scanner too, but it works the same, trust.)

Behavior before:

https://github.com/user-attachments/assets/13dd6725-b515-4c2e-a101-6673c3af1461

Behavior after:

https://github.com/user-attachments/assets/27681fd9-9c2e-45a5-a5e6-e8e842d5d6bd

## Requirements

## Breaking changes

**Changelog**

probably don't need a cl?

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/base_handheld.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/base_handheld.yml
@@ -6,6 +6,7 @@
   - type: ActivatableUIRequiresPowerCell
   - type: ItemToggle
     onUse: false # above component does the toggling
+    onActivate: false # above component does the toggling
   - type: PowerCellDraw
     drawRate: 0
     useRate: 20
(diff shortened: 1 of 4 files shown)


themias:
merge conflicted. I might fix and reopen later

PR #38284, opened 2025-06-12
Added directional beacons

Nox38:
## About the PR

Has this ever happened to you?
image
It has, this happens all the time and its awful. Introducing directional beacons for solars and escape pods!

## Why / Balance

These new map beacons can be added to all maps to make things comprehensible without breaking localization by manually setting the beacon names.

## Media

image

## Requirements

## Breaking changes

**Changelog**


Archisteks:
Had the nuke get armed on solars once. Not a good time.

Resources/Locale/en-US/navmap-beacons/station-beacons.ftl
```diff
@@ -47,6 +47,14 @@ station-beacon-engineering = Engineering
 station-beacon-ce = CE
 station-beacon-ame = AME
 station-beacon-solars = Solars
+station-beacon-solars-N = North Solars
+station-beacon-solars-NE = NE Solars
```
slarticodefast: It's a little inconsistent having some directions abbreviated and some written out.
I would use the full name for Northeast etc, or do they become too long in that case?

sudobeans:
I prefer "Solars (North)" Over "North Solars", so they all show up next to each other on the map beacon list. Not a dealbreaker but it's how I would do it! Thanks for doing this!

Resources/Locale/en-US/navmap-beacons/station-beacons.ftl
Nox38: They would get too long on the map visually. I could have them all be type direction like Solars N and Escape Pod SW?

slarticodefast: Yeah, that would work. And I like sudobean's suggestion of having the cardinal direction after the name so that they get sorted correctly

ArtisticRoomba (changes requested):
rename all to [Type] [Direction], ex. "Solars N" or "Escape Pod SE".

PR #38290, opened 2025-06-12
Optimizations from server profile

PJB3005:
See commits

Draft PR cuz I might add more to this in a second.


slarticodefast:
Out of interest, how did the regex stuff look like in the profile?

Content.Server/Chat/Managers/ChatSanitizationManager.cs
```diff
@@ -12,86 +12,86 @@ namespace Content.Server.Chat.Managers;
...
+        Entry(":)", "chatsan-smiles"),
+        Entry(":]", "chatsan-smiles"),
+        Entry("=)", "chatsan-smiles"),
+        Entry("=]", "chatsan-smiles"),
+        Entry("(:", "chatsan-smiles"),
+        Entry("[:", "chatsan-smiles"),
+        Entry("(=", "chatsan-smiles"),
+        Entry("[=", "chatsan-smiles"),
+        Entry("^^", "chatsan-smiles"),
+        Entry("^-^", "chatsan-smiles"),
```
Tayrtahn, line 26: So I have no idea if this would be more or less performant, but it would be possible to combine these into a single regex.

Could use `params` on `Entry` and use `string.Join('|', args.Select(Regex.Escape))` to make a single regex that matches all the possible input patterns. But again, I don't know if that would be more or less efficient than just having multiple simpler regexes.

Tayrtahn (approved):

Content.Server/Chat/Managers/ChatSanitizationManager.cs
PJB3005: Doing that, while probably more efficient, would be more involved. I don't wanna risk breaking anything.

PJB3005:
> Out of interest, how did the regex stuff look like in the profile?

Realistically none of this is much compared to the load of physics + PVS but it absolutely hurts my soul

Content.Shared/CCVar/CCVars.cs
```diff
@@ -36,5 +36,5 @@ public sealed partial class CCVars : CVars
     /// Set to true to disable parallel processing in the pow3r solver.
     /// </summary>
     public static readonly CVarDef<bool> DebugPow3rDisableParallel =
-        CVarDef.Create("debug.pow3r_disable_parallel", true, CVar.SERVERONLY);
```
Tayrtahn, line 39: I don't remember why we had this disabled, but was it fixed?

Tayrtahn (approved):

Content.Shared/CCVar/CCVars.cs
PJB3005: image

Tayrtahn: Ah, okay.

PR #38299, opened 2025-06-13
Fix crafting steps with modular machine parts

hoshizora-sayo:
## About the PR

Fixed crafting menu steps that included modular machine parts to have the proper name and icon
## Why / Balance

what the heck is a micro manipulator
## Technical details

There was a missed .ftl change in #37725 that caused recipes containing modular machine parts to say 'micro manipulator' in the crafting menu, that also left some remnants of the matter bin/capacitor. On top of that it was lacking an icon, as some other crafting menu steps do which I'll see if I can fix in another PR.
## Media

(before) what the heck
image
(after) much better
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed crafting menu entries that included modular machine parts


beck-thompson (approved):
looks good! Thanks for the fix

PR #38303, opened 2025-06-13
Allows the cryotube beaker container to be inserted to by Dragging and Dropping(Needs Review)

Skybailey-dev:
## About the PR
This is a very simple change that allows for items that can be inserted into the cryotube's Chemical Beaker slot by clicking on the cryotube itself while holding the item in your hand to also be insertable through dragging and dropping the item entity onto the cryotube

## Why / Balance
My main reason for doing this is to allow borgs to use cryotubes as intended, as they are otherwise unable to insert anything into a Chemical Beaker slot. This would make cryotubes more usable for any sillicon players, which would (hopefully) make cryotubes used more in general.

## Technical details
Only a couple lines of change were made. Reading the code will be faster than reading this explanation:
Added an or statement to the `CanDrop` argument in `OnCryoPodCanDropOn` in `SharedCryoPodSystems,` which makes `CanDrop` true whenever the dragged entity has the `BodyComponent` (default behavior) or the `FitsInDispenserComponent.` I also had to add using `Content.Shared.Chemistry.Components` to allow this check.

Second change is to `HandleDragDropOn` event in `CryoPodSystems,` also simple, it simply tries to insert the item into the Chemical Beaker slot using `TryInsert,` if it fails the argument Handled is set to false, if it succeeds the argument `Handled` is set to true. One more line is changed right after to cause the `HandleOnDragDropOn` to return if `Handled` has been set to true.

This is my first PR, and I am still new to C#, so I expect there to be at least something wrong with how I went about this. Please let me know so I can improve!

## Media

https://github.com/user-attachments/assets/dd6ec930-bd4a-4db4-bca5-738bfabd7954

## Requirements

## Breaking changes
None

**Changelog**
:cl:
- add: Added the ability for the Cryotube's beaker container to have beakers inserted through Dragging and Dropping.

### DIFF
--- a/Content.Server/Medical/CryoPodSystem.cs
+++ b/Content.Server/Medical/CryoPodSystem.cs
@@ -137,7 +137,11 @@ public sealed partial class CryoPodSystem : SharedCryoPodSystem
 
     private void HandleDragDropOn(Entity<CryoPodComponent> entity, ref DragDropTargetEvent args)
     {
-        if (entity.Comp.BodyContainer.ContainedEntity != null)
+        if (_itemSlotsSystem.TryInsert(entity, entity.Comp.SolutionContainerName, args.Dragged, args.User) == false)
+            args.Handled = false;
+        else args.Handled = true;
+
+        if (entity.Comp.BodyContainer.ContainedEntity != null || args.Handled == true)
             return;
 
         var doAfterArgs = new DoAfterArgs(EntityManager, args.User, entity.Comp.EntryDelay, new CryoPodDragFinished(), entity, target: args.Dragged, used: entity)
--- a/Content.Server/Medical/CryoPodSystem.cs
+++ b/Content.Server/Medical/CryoPodSystem.cs
@@ -137,9 +137,10 @@ public sealed partial class CryoPodSystem : SharedCryoPodSystem
 
     private void HandleDragDropOn(Entity<CryoPodComponent> entity, ref DragDropTargetEvent args)
     {
-        if (_itemSlotsSystem.TryInsert(entity, entity.Comp.SolutionContainerName, args.Dragged, args.User) == false)
-            args.Handled = false;
-        else args.Handled = true;
+        if (_itemSlotsSystem.TryInsert(entity, entity.Comp.SolutionContainerName, args.Dragged, args.User) == true) {
+            args.Handled = true;
+            return;
+        }
 
         if (entity.Comp.BodyContainer.ContainedEntity != null || args.Handled == true)
             return;
--- a/Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
+++ b/Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
@@ -72,7 +72,6 @@ namespace Content.Shared.Containers.ItemSlots
                     continue;
 
                 var item = Spawn(slot.StartingItem, Transform(uid).Coordinates);
-                    
                 if (slot.ContainerSlot != null)
                     _containers.Insert(item, slot.ContainerSlot);
             }
--- a/Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
+++ b/Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
@@ -72,6 +72,7 @@ namespace Content.Shared.Containers.ItemSlots
                     continue;
 
                 var item = Spawn(slot.StartingItem, Transform(uid).Coordinates);
+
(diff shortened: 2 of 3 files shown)


Content.Shared/Medical/Cryogenics/SharedCryoPodSystem.cs
```diff
@@ -38,8 +39,8 @@ private void OnCryoPodCanDropOn(EntityUid uid, CryoPodComponent component, ref C
     {
         if (args.Handled)
             return;
+        args.CanDrop = true ? HasComp<BodyComponent>(args.Dragged) | HasComp<FitsInDispenserComponent>(args.Dragged) : true;
```
slarticodefast: This tertianary operator makes no sense, it's always true, so it always returns the first option.

Content.Shared/Medical/Cryogenics/SharedCryoPodSystem.cs
```diff
@@ -12,6 +12,7 @@
 using Content.Shared.Verbs;
 using Robust.Shared.Containers;
 using Robust.Shared.Serialization;
+using Content.Shared.Chemistry.Components;
```
slarticodefast: sort alphabetically

Content.Server/Medical/CryoPodSystem.cs
```diff
@@ -137,7 +137,11 @@ public override void Update(float frameTime)
 
     private void HandleDragDropOn(Entity<CryoPodComponent> entity, ref DragDropTargetEvent args)
     {
-        if (entity.Comp.BodyContainer.ContainedEntity != null)
+        if (_itemSlotsSystem.TryInsert(entity, entity.Comp.SolutionContainerName, args.Dragged, args.User) == false)
+            args.Handled = false;
+        else args.Handled = true;
```
slarticodefast: This event subscription should only start the doafter.
`TryInsert` instantly inserts the beaker.
Also pretty sure this will conflict with dragging a person into the cryopod, as it now puts them into the beaker slot instead.

slarticodefast (changes requested):
I left a code review.
Conceptually I don't think borgs should be able to do this. Not being able to do any hand interactions like inserting beakers in part of their design. And coding this exceptions into the cryo pods makes this inconsistent with all other item insertion logic.

Skybailey-dev:
> I left a code review. Conceptually I don't think borgs should be able to do this. Not being able to do any hand interactions like inserting beakers in part of their design. And coding this exceptions into the cryo pods makes this inconsistent with all other item insertion logic.

Fixing the code in a sec when I have the chance. 
For Borg interactions: completely understand.
For the exception: a couple other things also have this exception (i.e. disposal system), but there are better ways for me to implement this in a less isolated manner.

Content.Server/Medical/CryoPodSystem.cs
Skybailey-dev: > Also pretty sure this will conflict with dragging a person into the cryopod, as it now puts them into the beaker slot instead.

This is using the ItemSlot variation of TryInsert which does check CanInsert, so It shouldn't do this? I will test this later to make sure.

Content.Shared/Medical/Cryogenics/SharedCryoPodSystem.cs
Skybailey-dev: yea idk what I was smoking with that one

Content.Server/Medical/CryoPodSystem.cs
Skybailey-dev: Changed it up a bit; re-review this part to make sure it's valid. Instantly inserting the beaker is intended, but that's easy to change.

Content.Shared/Containers/ItemSlot/ItemSlotsSystem.cs
```diff
@@ -72,7 +72,6 @@ private void OnMapInit(EntityUid uid, ItemSlotsComponent itemSlots, MapInitEvent
                     continue;
 
                 var item = Spawn(slot.StartingItem, Transform(uid).Coordinates);
-
```
Skybailey-dev, line 75: This is just a single whitespace change that annoyed me, no actual changes were actually made to the file 🌾

ArtisticRoomba:
Yeah, I agree with Slarti here. In the future, borg hands whitelisting should be real, and then we can easily define what borgs are allowed to interact with in that regard. But this PR is not the right solution to this problem.

Feel free to contribute in the future.

PR #38307, opened 2025-06-13
Add UI feedback when there's no followed players to warp to as ghost

qwerltaz:
## About the PR

~~Add popup~~ grey out the "warp to most followed" button when there's no followed players as ghost.

## Why / Balance

zero feedback on button press no good.


slarticodefast:
Hmm, usually UIs don't generate popups so this is a little inconsistent in my opinion.
Maybe just grey out the button if there is no target? Or change the text of the button itself?

IProduceWidgets:
Or have it select a random alive player to follow

iaada:
Disabling the button and changing the text feels right. I would like to see a "warp to random player" button on its own

qwerltaz:
Made button gray when nobody is being followed. Seems like the best solution.

https://github.com/user-attachments/assets/7338bb7c-af3d-46a9-b8ff-acbd45e85622

Content.Client/UserInterface/Systems/Ghost/GhostUIController.cs
```diff
@@ -26,8 +31,25 @@ public override void Initialize()
...
+            || Gui?.Visible == false)
+            return;
+
+        if (_lastUpdateTime + UpdateInterval > _gameTiming.CurTime
+            || !Gui!.TargetWindow.IsOpen)
+            return;
+
+        _lastUpdateTime = _gameTiming.CurTime;
+        OnGhostnado(false);
+    }
```
sowelipililimute: don't send a new networked message every frame to request if there's anything to follow, instead, have the server push changes when the most followed changes

sowelipililimute (changes requested):

PR #38309, opened 2025-06-13
Made Kobolds have like health to their genetically barely similar cousin

TheFlyingSentry:
## About the PR
For those unaware the base Kobold has less health than monkeys, taking only 40 less damage to go into crit while monkeys took 100 damage to go into crit
Now for some reason syndicate kobolds have as much health as a monkey which is inconsistent and mildly annoyed me

## Why / Balance
This will hopefully make people more willing to get kobold reinforcements because it's kinda well known that normal kobolds only take 60 damage to go into crit... but it's not well known that syndicate kobolds actually have the normal 100 health threshold of the monkey.

## Technical details
All YAML

## Media
Unless you want me to shoot a kobold an prove it with a med analyzer not much point

## Requirements
## Breaking changes

**Changelog**
:cl:
- tweak: Non-Syndicate Kobolds now have as much health as their genetic counterpart

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
@@ -1551,11 +1551,6 @@
       path: /Audio/Animals/lizard_happy.ogg
     interactFailureSound:
       path: /Audio/Items/wirecutter.ogg
-  - type: MobThresholds
-    thresholds:
-      0: Alive
-      60: Critical
-      125: Dead
   - type: MovementSpeedModifier
     baseWalkSpeed: 3.5
     baseSprintSpeed: 5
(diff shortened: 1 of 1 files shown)


slarticodefast:
Didn't they have other stats that were better as a compensation?

Edit: I just checked and they are 25% faster and have a stronger melee attack.
So if you make their health the same then you now have the same problem, but in the other direction.
grafik

TheFlyingSentry:
i'd agree if it weren't for the syndicate version getting their health back

TheFlyingSentry:
syndicate kobolds are currently just better... but people keep thinking they only have 60hp because of base kobolds

TheFlyingSentry:
image
it dosen't make sense to balance it for syndicates when it's always just been better

TheFlyingSentry:
And it's not like it's unbalanced for chefs either. you can kill them with the freezer temps alone or just by putting them in a crate/locker and killing them from asphyxiation

Morp9x:
I don't understand why they even decided to increase the health of the syndicate's kobolds.

RedBookcase:
Why is one faster in the first place? Imo just remove the speed boost from kobolds and make the health the same as monkeys.

southbridge-fur:
Any updates on this?

PR #38310, opened 2025-06-13
Cleanup SharedRoleCodewordSystem

slarticodefast:
## About the PR
Remove some code that did not do anything.

## Why / Balance
cleanup

## Technical details
The component was made session specific to prevent cheat clients from seeing other player's codewords.
However, this was redundant.
`RoleCodewordComponent` is stored on the player's mind. Mind entities are stored in nullspace and the player it belongs to has a PVS override, so it gets networked to only them, which means other clients cannot see that component in the first place.

I also added a missing `Dirty` call for the `SetRoleCodewords` method. It was only working because it was called in the same tick where the component is added.

## Media
Proof that everything still works and other clients cannot see the component.

https://github.com/user-attachments/assets/76ec887e-94fb-45e3-8de4-6b6ad548ea45

## Requirements

## Breaking changes
`SetRoleCodewords` in the `SharedRoleCodewordSystem` now requires an `Entity<RoleCodewordComponent>`

**Changelog**
none

### DIFF
--- a/Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
@@ -136,8 +136,8 @@ public sealed class TraitorRuleSystem : GameRuleSystem<TraitorRuleComponent>
         // Send codewords to only the traitor client
         var color = TraitorCodewordColor; // Fall back to a dark red Syndicate color if a prototype is not found
 
-        RoleCodewordComponent codewordComp = EnsureComp<RoleCodewordComponent>(mindId);
-        _roleCodewordSystem.SetRoleCodewords(codewordComp, "traitor", factionCodewords.ToList(), color);
+        var codewordComp = EnsureComp<RoleCodewordComponent>(mindId);
+        _roleCodewordSystem.SetRoleCodewords((mindId, codewordComp), "traitor", factionCodewords.ToList(), color);
 
         // Change the faction
         Log.Debug($"MakeTraitor {ToPrettyString(traitor)} - Change faction");
--- a/Content.Shared/Roles/RoleCodeword/SharedRoleCodewordSystem.cs
+++ b/Content.Shared/Roles/RoleCodeword/SharedRoleCodewordSystem.cs
@@ -1,49 +1,11 @@
-using Content.Shared.Mind;
-using Robust.Shared.GameStates;
-using Robust.Shared.Player;
-
 namespace Content.Shared.Roles.RoleCodeword;
 
 public abstract class SharedRoleCodewordSystem : EntitySystem
 {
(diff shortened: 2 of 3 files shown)


Tayrtahn (approved):

VerinSenpai:
that is a diff of beauty.

Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
```diff
@@ -136,8 +136,8 @@ public bool MakeTraitor(EntityUid traitor, TraitorRuleComponent component)
         // Send codewords to only the traitor client
```
ArtisticRoomba: Might be good to clarify that `SendOnlyToOwner` isn't necessary as the comp is stored on the mind in nullspace with a PVS override.

Content.Shared/Roles/RoleCodeword/SharedRoleCodewordSystem.cs
```diff
@@ -1,49 +1,11 @@
...
-        return true;
-    }
-
-    public void SetRoleCodewords(RoleCodewordComponent comp, string key, List<string> codewords, Color color)
+    public void SetRoleCodewords(Entity<RoleCodewordComponent> ent, string key, List<string> codewords, Color color)
     {
         var data = new CodewordsData(color, codewords);
-        comp.RoleCodewords[key] = data;
+        ent.Comp.RoleCodewords[key] = data;
+        Dirty(ent, ent.Comp);
```
ArtisticRoomba: Looks better, no?
```suggestion
        Dirty(ent);
```

ArtisticRoomba (approved):
Good to selfmerge, left suggestions.

Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
slarticodefast:
```suggestion
```

Content.Server/GameTicking/Rules/TraitorRuleSystem.cs
```diff
@@ -136,8 +136,8 @@ public bool MakeTraitor(EntityUid traitor, TraitorRuleComponent component)
...
         var color = TraitorCodewordColor; // Fall back to a dark red Syndicate color if a prototype is not found
 
-        RoleCodewordComponent codewordComp = EnsureComp<RoleCodewordComponent>(mindId);
-        _roleCodewordSystem.SetRoleCodewords(codewordComp, "traitor", factionCodewords.ToList(), color);
+        var codewordComp = EnsureComp<RoleCodewordComponent>(mindId);
```
slarticodefast, line 139:
```suggestion
        // The mind entity is stored in nullspace with a PVS override for the owner, so only they can see the codewords.
        var codewordComp = EnsureComp<RoleCodewordComponent>(mindId);
```

PR #38311, opened 2025-06-13
Power cell examine is now predicted

SpeltIncorrectyl:
## About the PR
Examining a power cell or an entity with a power cell inside will now instantly show you the percentage charge instead of waiting for the server.

## Technical details
PowerCellComponent now has PercentageCharge field that is set server side.
The examine messages are moved to shared and they work off this field.

## Media

It's the same as before it just happens instantly now.
Screenshot from 2025-06-13 19-30-37
Screenshot from 2025-06-13 19-30-56
Screenshot from 2025-06-13 19-31-10

## Requirements

- [X ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

No changelog

### DIFF
--- a/Content.Shared/PowerCell/Components/PowerCellComponent.cs
+++ b/Content.Shared/PowerCell/Components/PowerCellComponent.cs
@@ -9,9 +9,16 @@ namespace Content.Shared.PowerCell;
 /// </summary>
 [NetworkedComponent]
 [RegisterComponent]
+[AutoGenerateComponentState]
 public sealed partial class PowerCellComponent : Component
 {
     public const int PowerCellVisualsLevels = 2;
+
+    /// <summary>
+    /// The percentage charge of this power cell.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float PercentCharge;
 }
 
 [Serializable, NetSerializable]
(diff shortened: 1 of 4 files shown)


Content.Shared/PowerCell/Components/PowerCellComponent.cs
```diff
@@ -9,9 +9,16 @@ namespace Content.Shared.PowerCell;
...
+[AutoGenerateComponentState]
 public sealed partial class PowerCellComponent : Component
 {
     public const int PowerCellVisualsLevels = 2;
+
+    /// <summary>
+    /// The percentage charge of this power cell.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float PercentCharge;
```
slarticodefast, line 21: Instead of adding a datafield for the percentage and networking that the charge and max charge should be networked directly in the `BatteryComponent`.

SpeltIncorrectyl: So I should but the BatteryComponent onto shared?

SpeltIncorrectyl:
Eh I think it would need a full refactor of the power cell system cause some things use PowerCellDrawComponent and other things (like the radio jammer) just rawdog the battery component and do their own query every tick on it.
Otherwise you wouldn't be able to know the the of power cell decay.

SpeltIncorrectyl:
I will close this and maybe one day I will try to do power cell refactor.

PR #38286, opened 2025-06-12
Use correct on-head rainbow lizard plushie sprite

perryprog:
## About the PR

Rainbow lizard plushie had the wrong state for the head sprite, causing it to display everyone's favorite ERROR sprite instead. This adds a grayscale version of the standard lizard plushie head sprite and uses that instead.

Don't believe attribution update is needed in this case but let me know if so.

## Why / Balance

ERROR gaming.

## Media

image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

Issue #38293, opened 2025-06-12
SharedActionSystem causing major server crashes


ArtisticRoomba:
CC @keronshb @deltanedas 

## Description
Recently poor Vulture cried itself to sleep, in its dying breaths we got some logs out of it.

Stacktraces seem to be infinitely failing to resolve metadata on `OnGetState` calls.

Small stacktrace below:
```
1749763809389	Can't resolve "Robust.Shared.GameObjects.MetaDataComponent" on entity 3246480!
   at System.Environment.get_StackTrace()
   at Robust.Shared.GameObjects.EntityManager.GetNetEntity(EntityUid uid, MetaDataComponent metadata) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Network.cs:line 186
   at Robust.Shared.GameObjects.EntityManager.GetNetEntitySet(HashSet`1 entities) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Network.cs:line 512
   at Content.Shared.Actions.SharedActionsSystem.OnGetState(Entity`1 ent, ComponentGetState& args) in /home/runner/work/space-station-14/space-station-14/Content.Shared/Actions/SharedActionsSystem.cs:line 103
   at Robust.Shared.GameObjects.EntityEventBus.<>c__DisplayClass57_0`2.<SubscribeLocalEvent>g__EventHandler|0(EntityUid uid, IComponent comp, TEvent& args) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityEventBus.Directed.cs:line 324
   at Robust.Shared.GameObjects.EntityManager.GetComponentState(IEventBus eventBus, IComponent component, ICommonSession session, GameTick fromTick) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Shared/GameObjects/EntityManager.Components.cs:line 1704
   at Robust.Server.GameStates.PvsSystem.GetEntityState(ICommonSession player, EntityUid entityUid, GameTick fromTick, MetaDataComponent meta) in /home/runner/work/space-station-14/space-station-14/RobustToolbox/Robust.Server/GameStates/PvsSystem.GetStates.cs:line 52
[...]

PR #38312, opened 2025-06-13
Add integration test for equipping/unequipping clothing entities

### LINKED ISSUE #38194: Need integration test for (un)equipping clothing items
We got quite a few clothing items that add actions, spawn entities, add overlays etc. when equipped.
So in order to catch errors we could use a simple test that

- spawns every item with the `ClothingComponent`
- spawns an Urist
- equips the clothing items into the corresponding inventory slot (some items might fit into multiple slots)
- simulates the game for a few ticks
- unequips the item
- simulates again for a few ticks

Note that some inventory slots depend on others, so that Urist might need to wear a jumpsuit so it can equip pdas or similar.
Also there are quite a lot of clothing items, so maybe the test could use some optimization to skip spawning those that do absolutely nothing.
And it is important to test this with a client, not only on the server, so that all overlays are being added.

perryprog:
## About the PR

This PR requires #38274, and will be marked as a draft until it's merged.

This introduces an integration test that, at the broad level:
1. Spawns in every item that has a ClothingComponent and then
2. Equips each one (plus jumpsuit/hardsuit if needed) into each slot it can fit in
3. Triggers any InstantActions it gains on each equip
4. Unequips it (and jumpsuit/hardsuit)

## Why / Balance

Fixes #38194. Would have caught part of #38274 (equip mask, toggle mask, unequip mask), as well as #38286.

## Technical details

- This is my first time writing an integration test for SS14—any nitpicks are welcome. I'll do a self review on a few bits I have questions about.
- Running the test on a downstream passed without issue.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
+++ b/Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
@@ -11,6 +11,7 @@ using Content.Shared.Inventory;
 using Content.Shared.Mobs.Components;
 using Content.Shared.Prototypes;
 using Content.Shared.Storage.Components;
+using Content.Shared.Tag;
 using Robust.Shared.GameObjects;
 using Robust.Shared.Prototypes;
 
@@ -35,6 +36,11 @@ public sealed class ClothingEquipTest : InteractionTest
         typeof(RandomSpawnerComponent),
     ];
 
+    private readonly List<ProtoId<TagPrototype>> _ignoredTags =
+    [
+        "PetOnly",
+    ];
+
     private readonly List<Type> _removedComponents =
     [
         // Material bag can pick up other clothing items.
@@ -51,6 +57,8 @@ public sealed class ClothingEquipTest : InteractionTest
                         && !x.HideSpawnMenu
                         && !x.Abstract
                         && !Pair.IsTestPrototype(x)
+                        && !(x.TryGetComponent<TagComponent>(out var tags, Factory)
+                             && Tags.HasAnyTag(tags, _ignoredTags))
                         && !_ignoredComponents.Any(ignore => x.HasComponent(ignore)));
 
         List<EntityUid> clothings = [];
--- a/Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
+++ b/Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
 using Content.IntegrationTests.Tests.Interaction;
 using Content.Server.Actions;
 using Content.Server.Inventory;
@@ -11,7 +12,6 @@ using Content.Shared.Inventory;
 using Content.Shared.Mobs.Components;
 using Content.Shared.Prototypes;
 using Content.Shared.Storage.Components;
-using Content.Shared.Tag;
 using Robust.Shared.GameObjects;
 using Robust.Shared.Prototypes;
 
@@ -28,7 +28,7 @@ public sealed class ClothingEquipTest : InteractionTest
     private const string HardsuitProto = "ClothingOuterHardsuitEVA";
     private const string JumpsuitProto = "ClothingUniformJumpsuitColorGrey";
 
-    private readonly List<Type> _ignoredComponents =
+    private readonly HashSet<Type> _ignoredComponents =
     [
         // Mobs are a massive pain in this test
         // (Admeme mouse eating your cak, cancer mouse killing you... mice just dying...)
@@ -36,12 +36,7 @@ public sealed class ClothingEquipTest : InteractionTest
         typeof(RandomSpawnerComponent),
     ];
 
-    private readonly List<ProtoId<TagPrototype>> _ignoredTags =
-    [
-        "PetOnly",
-    ];
-
-    private readonly List<Type> _removedComponents =
+    private readonly HashSet<Type> _removedComponents =
     [
         // Material bag can pick up other clothing items.
         typeof(MagnetPickupComponent),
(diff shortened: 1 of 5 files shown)


Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
```diff
@@ -0,0 +1,176 @@
...
+    {
+        var clothingProtos = ProtoMan.EnumeratePrototypes<EntityPrototype>()
+            .Where(x => x.HasComponent<ClothingComponent>()
+                        && !x.HideSpawnMenu
+                        && !x.Abstract
+                        && !Pair.IsTestPrototype(x)
+                        && !x.HasComponent<RandomSpawnerComponent>()
+                        // Mobs are a massive pain in this test
+                        // (Admeme mouse eating your cak, cancer mouse killing you... mice just dying...)
+                        && !x.HasComponent<MobStateComponent>());
```
perryprog: Previously it was mentioned that it might be good to filter out clothing items that don't do anything, however, it turns out this test also serves as a way to test for the validity of sprites in each possible equip position. This makes me think it might be good to keep this broad-as is, but this is definitely on the slow side of things. It seems to take about 30–40 seconds to run for me, though I think part of that is due to maybe some unneeded RunTicks?

Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
```diff
@@ -0,0 +1,176 @@
...
+        EntityUid hardsuit = default!;
+        await Server.WaitPost(() =>
+        {
+            foreach (var clothing in clothingProtos)
+            {
+                var ent = SEntMan.SpawnAtPosition(clothing.ID, ToServer(PlayerCoords));
+                clothings.Add(ent);
+
+                // Material bag can pick up other clothing items.
+                SEntMan.RemoveComponent<MagnetPickupComponent>(ent);
```
perryprog: Ew icky hardcoded edge cases. Should I move this into some sort of `removedComponents` variable? (And probably also an `ignoredComponents` one for use in the prototype filtering?)

Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
```diff
@@ -0,0 +1,176 @@
...
+
+                    DoEquip(invSystem, clothing, slot.Name, clothingComp);
+                });
+                await RunTicks(1);
+
+                var newActions = actionSys.GetActions(SPlayer)
+                    .Except(baseActions)
+                    // Don't bother with more complicated actions (they can take targets, etc.)
+                    .Where(action => SEntMan.HasComponent<InstantActionComponent>(action.Owner))
+                    .ToList();
```
perryprog, line 121: While this does end up passing on both here and a downstream I tried the test on, this feels like it's asking for trouble. Consider an item that grants an instant action which scrambles you. That would then put the player out of range of other items, killing the test.

Should this be filtered down to just actions that have a ToggleAction component? Or even just ToggleClothing?

Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
```diff
@@ -0,0 +1,176 @@
...
+
+                var unremovableComp = SEntMan.GetComponentOrNull<UnremoveableComponent>(clothing);
+                // Electropack or similar has to be force removed, since our unequip actor is always ourselves
+                var needsForce = SEntMan.HasComponent<SelfUnremovableClothingComponent>(clothing);
+
+                if (unremovableComp != null)
+                {
+                    needsForce = true;
+                    // We want to reuse this entity for other slots. (E.g., CluwnePDA)
+                    unremovableComp.DeleteOnDrop = false;
```
perryprog, line 134: Unrelated but there might be an odd transient bug in handling action serialization and DeleteOnDrop unremovable items—I was only able to reproduce it once, though, when force unequipping a cluwne PDA. Was kinda odd.

Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
```diff
@@ -0,0 +1,176 @@
...
+
+                    // If we should've needed to force for this item, but got it off, that's no good.
+                    Assert.That(isRemoved && needsForce,
+                        Is.False,
+                        $"Unequipped {SEntMan.ToPrettyString(clothing)}, which should have been unremovable.");
+
+                    // Just in case some kind of toggle action put us into a state that this test doesn't expect.
+                    // Currently this doesn't seem to get hit.
+                    if (!isRemoved)
+                        RunActions(newActions, actionSys);
```
perryprog, line 151: Should this be removed? In the original implementation I thought it was needed, but I'm not seeing any cases where this is a behavior that exists.

Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
```diff
@@ -0,0 +1,176 @@
...
+
+/// <summary>
+/// This test spawns in every item that has a ClothingComponent and then simulates:
+/// 1. Equipping it (and jumpsuit/hardsuit if needed)
+/// 2. Triggering any InstantActions it grants.
+/// 3. Unequipping it.
+/// </summary>
+public sealed class ClothingEquipTest : InteractionTest
+{
+    private const string HardsuitProto = "ClothingOuterHardsuitEVA";
```
Tayrtahn:
```suggestion
    private static readonly EntProtoId HardsuitProto = "ClothingOuterHardsuitEVA";
```
Make these static `EntProtoId`s so they get validated.

perryprog:
Also—I don’t believe this covers most/all overlays, as it seems their draw code is only invoked in the full game loop during rendering. (I thiiiiink.) I’m not sure if there’s a way we can configure this test to run with that?

Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
perryprog: Lol it was #38293.

Content.IntegrationTests/Tests/Inventory/ClothingEquipTest.cs
perryprog: Guh, I had that on my mental list to fix, too. Thanks!

PR #38313, opened 2025-06-13
refactor: remade EMP effect for chameleon to be client-side and not spam dirty

Fildrance:
## About the PR

Moved most of code, related to EMP effect on chameleon clothing, to client side. Sacrificed item description and identity data flickering (i guess its okay?). Also EMP now blocks chameleon UI from opening (to prevent changing prototype during effect).

## Why / Balance

Spamming dirty on chamelon clothing component 7 times per second is less then ideal.

## Technical details

Moved code for updating visual in cycle to client code.
Reserved client-side only field for prototype that was used before emp landed on entity. Added field dirty for chameleon component, but have not touched other code (not related to chameleon) so transition to field-dirty is not yet complete.

## Media

Previous behaviour was kept except item description, identity flickering and UI not openable during effect.

## Requirements

## Breaking changes

**Changelog**

nocl

### DIFF
--- a/Content.Shared/Clothing/EntitySystems/SharedChameleonClothingSystem.cs
+++ b/Content.Shared/Clothing/EntitySystems/SharedChameleonClothingSystem.cs
@@ -110,7 +110,6 @@ public abstract class SharedChameleonClothingSystem : EntitySystem
             proto.TryGetComponent(out AppearanceComponent? appearanceOther, Factory))
         {
             _appearance.AppendData(appearanceOther, uid);
-            Dirty(uid, appearance);
         }
 
         // properly mark contraband


slarticodefast:
I'm currently predicting EMPs and this PR will make it a little easier, so I fixed the merge conflict and pushed some review changes:
- Moved `GetRandomValidPrototype` to shared.
- Made the `NextEmpChange` time purely client-side, there is no need to network it.
- Restored the original behaviour: If `EmpContinuous` is false then the prototype will be changed once and stay that way until the player changes it. If true then it will be changed and causes the visual effect while disabled.
- Removed `BeforeEmpPrototype` because it originally did not reset to the original sprite, instead we just force a reset to the current one.
- Reduced the rate a little because it was quite hard to look at.

One last bug I found is that the Chameleon PDA is not updating correctly:
ss14

This is probably because it uses Appearance data. So we have to figure something out to fix that.

Fildrance:
So the behaviour with PDA is closely related to resetting of appearance upon prediction. We have cycle that changes appearance only working on client side, but AppearanceSystem expected any change to be dirtied and provided by server. Which is not happening. We can avoid it by having overload of SharedAppearanceSystem.AppendData(AppearanceComponent srcComp, Entity<AppearanceComponent?> dest) with ```bool dirty = true``` additional argument, @slarticodefast do you have any idea, who can tell if thats actually a good idea? Or do i just PR it to RT and time will tell... :|

slarticodefast:
Feels a little iffy as a solution to be honest, but I can't come up with anything else either 😔

Fildrance:
Well, it looks like we should not have un-synced client only appearance data by desing (i have no idea how to make hallucinations without it btw) so that whole part of optimization wont see the light.
I should revert it to be server-side after https://github.com/space-wizards/space-station-14/pull/39802 will be merged. Or Slarti can port all of changes into it and then we will close this one.

slarticodefast:
What changes from this PR could be kept if we have to keep the EMP server side?

Fildrance:
I think the only one you did not mention in previous comment is closing UI on emp hit (i dont think there is any reason to keep it open and enjoy glitching)
https://github.com/space-wizards/space-station-14/pull/38313/files#diff-863264b9b0d1165c46fa101b79ad5b57ad7f9a5a9aefe60a5bd7fc0987adc140R43

PR #35992, opened 2025-03-21
Unborgable trait

youtissoum:
## About the PR

This PR adds a trait that prevents the player's brain from being inserted into a man-machine interface.

## Why / Balance

I believe it can lead to interesting roleplay to have the ability to make characters that *cannot* be brought back after death.

## Technical details

This trait adds an unborgable component to the player which adds a new MMIIncompatible component to the player's brain. Entities with a MMIIncompatible component cancel all attempts to be inserted into an MMI.

## Media

unborgable:

https://github.com/user-attachments/assets/97160c33-d45f-4b70-a879-4fc179ec01ad

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- add: An unborgable trait is now available in the traits menu. This trait will prevent other players from inserting your brain into an MMI.

PR #38314, opened 2025-06-13
Borging confirmation window

youtissoum:
## About the PR

I added a confirmation window when a player's brain is inserted into a man-machine interface.

This PR was split off from #35992 

## Why / Balance

Players often complain about being inserted into a man-machine interface when they do not enjoy playing as a cyborg.
This PR adds an option to not have to become a cyborg that is more intuitive and less annoying that having to type a command into the command line.

## Technical details

Just before actually transferring the player's mind into the MMI and giving them the role, the borg system will now open an eui if the player is not still inside the brain.
This eui will open a new generic confirmation window whose text can be modified depending on the need. This UI is based on the current UIs used for cloning and revival confirmation.

## Media

https://github.com/user-attachments/assets/40742c4d-917f-41e0-8f8e-59f8f4c46f28

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Players will now get a confirmation window when their brain is inserted into a man-machine interface.


VerinSenpai:
I think it looks good. It'd look better IMO if that space at the bottom of the window was removed or the buttons were moved down but otherwise. A little perplexed that this is how a reusable confirmation window gets added.

youtissoum:
> I think it looks good. It'd look better IMO if that space at the bottom of the window was removed or the buttons were moved down but otherwise.

Good idea, the confirmation window should now also handle being resized much better too.

> A little perplexed that this is how a reusable confirmation window gets added.

Do you prefer if I make yet another hardcoded revival confirmation window :trollface:

Shooty8:
this looks great but I do wonder if it can be used to spam a ghosted player since the interaction has no cooldown unlike how reviving a player has a cooldown in that it takes time to heal and defib a dead character

VerinSenpai:
> Do you prefer if I make yet another hardcoded revival confirmation window :trollface:

Certainly not 😅 
After this the other EUI can be updated. Yay cleanup and deduplication.

ilovehans10:
> this looks great but I do wonder if it can be used to spam a ghosted player since the interaction has no cooldown unlike how reviving a player has a cooldown in that it takes time to heal and defib a dead character

I agree with this; while it is a bit less intrusive than being shunted into a new body, you still deal with a popup in the middle of your screen. One way around this is a "Don't show this popup again" checkbox  in the popup that disables the popup for the round when checked.

Edit: Overall I'm still a massive fan of this change and think that implementing this now and adding new features later is the way to go.

Absotively:
I think it might be worth making the buttons more specific than "Yes" and "No", just because people do tend to skim popups quickly. Maybe "Enter MMI" and "Stay dead"?

youtissoum:
> I think it might be worth making the buttons more specific than "Yes" and "No", just because people do tend to skim popups quickly. Maybe "Enter MMI" and "Stay dead"?

Looks good. Thanks for the idea

VerinSenpai:
drats

PR #38315, opened 2025-06-14
Vox and Reptilian can use their claws and talons as knives.

BigfootBravo:
## About the PR
Vox and Reptilians can now use their talons or claws to butcher clothes for cloth. Slicing does not work, for some unknown reason, which is why this is a draft. This may also change or break if #36895 gets merged. Also the message upon butchering something is a bit odd. I could see this being a Vox thing only, not a reptilian, possibly, due to species flavor/lore, with Vox having large-ish talons and Reptilians having small claws.

## Why / Balance
They have sharp, slash dealing melee attacks, so why couldn't they use their noticeably sharp appendages as mini knives! Their unarmed weapons are already essentially very slightly worse shivs, losing 0.5 Slash damage, but are also slightly better in that they don't take a inventory slot and don't need to be held. If their appendages are sharp enough to nearly be the same as a shiv, which is a knife, why can't they use said appendages for similar things as knives, especially with their ability to finely control their appendages, due to them being a part of them instead of a entirely separate tool?

## Technical details
Addition of `Knife` tags, `Sharp` component, utensil `Knife` component, and the `Slicing` tool quality.

## Media

https://github.com/user-attachments/assets/677a86d1-5bb0-441a-914d-7cfef6462bd1

## Requirements

## Breaking changes
None

**Changelog**

:cl:
- add: Vox and Reptilians can now use their talons or claws as knives for some processes.


beck-thompson:
vulp players crying rn

BigfootBravo:
> vulp players crying rn

If Vulps do slash damage or have a reasonable backing for why they could do this, don't see why they couldn't get this functionality as well!

Though still not gonna be making this a full PR yet for any of the species, unless the slicing issue gets figured out, due to how inconsistent it would be if they could use their sharp appendages to cut clothes into cloth, but not cutting open a letter. No idea why the changes made so far won't let the latter occur though, so for the moment this PR is in a bit of limbo, especially while a refactor of the related slicing component is in progress.

lzk228:
i may agree with cutting (e.g. food), but don't agree with ability to butcher things
maybe butchering clothing is fine, but surely not mobs

LevitatingTree:
agreed, not sure about the butchering part, or at least it feels like it should be a lot slower than an actual sharp blade

FungiFellow:
This feels like powercreep and would be a bad change as a prisoner could incredibly easily make a baseball bat because of inbuilt knife fingers.

Hitlinemoss:
In my eyes, claws/talons aren't sharp enough to really cut up clothing/bread/whatever in a practical or precise manner. There's a reason that their damage is so low compared to an actual knife, after all.

EmoGarbage404:
Needs actual justification because presently I'm unconvinced that giving characters unremovable tools is remotely beneficial.

BigfootBravo:
Very reasonable comments all around. I just initially thought it was a bit odd that claws or talons couldn't be used to atleast tear apart clothes into cloth, and then it escalated into them being usable elsewhere.  

This is also part of why I made this a draft, afterall. In one, because it doesn't fully work, and two because it could become a fairly significant change.

BigfootBravo:
> In my eyes, claws/talons aren't sharp enough to really cut up clothing/bread/whatever in a practical or precise manner. There's a reason that their damage is so low compared to an actual knife, after all.

Vox and Reptilian slash damage is only 0.5 less than a shiv, which can do all of what this change would add to the species in question.

BigfootBravo:
Oh, did not mean to close this. Oops! Just a misclick.
Still probably good to get some more discussion on this.

slarticodefast:
Yeah, I agree with Emo and the others here. Adding inherent tools to a species bypasses a lot of interaction and tool mechanics, in particular for crafting.

PR #38317, opened 2025-06-14
Commands for adding and removing actions

ScarKy0:
## About the PR

Added commands for adding and removing actions from entities.
## Why / Balance

Ease of testing. This sorta thing very much lacked when working on the armblade.

## Technical details

Not much to say. Done it similiarly to how addobjective/rmobjective works.

## Media

https://github.com/user-attachments/assets/3a4266f1-acd6-47c0-bdb8-22deb1fed0c3

## Requirements

## Breaking changes

**Changelog**

:cl:
ADMIN:
- add: addaction/rmaction commands. NOTE: Addaction will not work correctly if the action requires the entity to have a component (such as Dragon's Devour).

### DIFF
--- a/Content.Server/Actions/Commands/AddActionCommand.cs
+++ b/Content.Server/Actions/Commands/AddActionCommand.cs
@@ -1,0 +1,75 @@
+﻿using System.Linq;
+using Content.Server.Administration;
+using Content.Shared.Actions;
+using Content.Shared.Actions.Components;
+using Content.Shared.Administration;
+using Content.Shared.Prototypes;
+using Robust.Shared.Console;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.Actions.Commands;
+
+[AdminCommand(AdminFlags.Fun)]
+public sealed class AddActionCommand : LocalizedEntityCommands
+{
+    [Dependency] private readonly IPrototypeManager _prototypes = default!;
+    [Dependency] private readonly SharedActionsSystem _actions = default!;
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+
+    public override string Command => "addaction";
+
+    public override void Execute(IConsoleShell shell, string argStr, string[] args)
+    {
+        if (args.Length != 2)
+        {
+            shell.WriteError(Loc.GetString(Loc.GetString("cmd-addaction-invalid-args")));
+            return;
+        }
+
+        if (!NetEntity.TryParse(args[0], out var targetUidNet) || !EntityManager.TryGetEntity(targetUidNet, out var targetEntity))
+        {
+            shell.WriteLine(Loc.GetString("shell-entity-uid-must-be-number"));
+            return;
+        }
+
+        if (!EntityManager.HasComponent<ActionsComponent>(targetEntity))
+        {
+            shell.WriteError(Loc.GetString("cmd-addaction-actions-not-found"));
+            return;
+        }
+
+        if (!_prototypes.TryIndex<EntityPrototype>(args[1], out var proto) ||
+            !proto.HasComponent<ActionComponent>())
+        {
+            shell.WriteError(Loc.GetString("cmd-addaction-action-not-found", ("action", args[1])));
+            return;
+        }
+
+        if (_actions.AddAction(targetEntity.Value, args[1]) == null)
+        {
+            shell.WriteError(Loc.GetString("cmd-addaction-adding-failed"));
+        }
+    }
+
+    public override CompletionResult GetCompletion(IConsoleShell shell, string[] args)
+    {
+        if (args.Length == 1)
+        {
+            return CompletionResult.FromHintOptions(
+                CompletionHelper.Components<ActionsComponent>(args[0]),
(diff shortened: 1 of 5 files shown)


Content.Server/Actions/Commands/AddActionCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+using Content.Shared.Actions;
+using Content.Shared.Actions.Components;
+using Content.Shared.Administration;
+using Content.Shared.Prototypes;
+using Robust.Shared.Console;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.Actions.Commands;
+
+[AdminCommand(AdminFlags.Fun)]
```
slarticodefast: Hmm, is this the category where this should be?
While it can be used for memes I think it might also make sense to have this in Debug, same as adding and removing components, which normally handles actions. And it would be useful for testing stuff. But maybe ask a headmin.

Content.Server/Actions/Commands/RemoveActionCommand.cs
```diff
@@ -0,0 +1,84 @@
+﻿using Content.Server.Administration;
+using Content.Shared.Actions;
+using Content.Shared.Actions.Components;
+using Content.Shared.Administration;
+using Robust.Shared.Console;
+
+namespace Content.Server.Actions.Commands;
+
+[AdminCommand(AdminFlags.Fun)]
```
slarticodefast: same

Content.Server/Actions/Commands/RemoveActionCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+public sealed class RmActionCommand : LocalizedEntityCommands
+{
+    [Dependency] private readonly SharedActionsSystem _actions = default!;
+
+    public override string Command => "rmaction";
```
slarticodefast, line 14: Did we decide on a namingconvention for these yet? We had some discussion for "list" vs "ls". so maybe this should be "removeaction"? Which one is more consistent with other commands?
Also the file should be named the same as the command

Content.Server/Actions/Commands/RemoveActionCommand.cs
```diff
@@ -0,0 +1,84 @@
...
+            return;
+        }
+
+        if (_actions.GetAction(targetActionEntity) is not { } ent)
+        {
+            shell.WriteError(Loc.GetString("cmd-rmaction-not-an-action"));
+            return;
+        }
+
+        _actions.SetTemporary(ent.Owner, true);
```
slarticodefast, line 48: Should this always be done?
For example if you have an action from holding a flashlight, then the flashlight would permanently break.

Content.Server/Actions/Commands/RemoveActionCommand.cs
ScarKy0: RemoveObjectivesCommand is for rmobjective so i based that off of it

Content.Server/Actions/Commands/RemoveActionCommand.cs
ScarKy0: i mean if you plan to remove it anyways i dont think it would ever work again afterwards

Content.Server/Actions/Commands/RemoveActionCommand.cs
Fildrance: better have them all with one so we can rename them in batch later if we decide on something new :godo:

Content.Server/Actions/Commands/RemoveActionCommand.cs
```diff
@@ -0,0 +1,84 @@
+﻿using Content.Server.Administration;
+using Content.Shared.Actions;
+using Content.Shared.Actions.Components;
+using Content.Shared.Administration;
+using Robust.Shared.Console;
+
+namespace Content.Server.Actions.Commands;
+
+[AdminCommand(AdminFlags.Debug)]
+public sealed class RmActionCommand : LocalizedEntityCommands
```
Fildrance: as u remarked, command class was named Remove - i'd be better of with full name here

Content.Server/Actions/Commands/RemoveActionCommand.cs
Fildrance: its and admin action, its fine if it will break stuff (at least only for certain item or for certain player) - its not a bug! its a feature :3

Content.Server/Actions/Commands/RemoveActionCommand.cs
Fildrance: Well, it actually is not breaking light, you just need to re-pick it and it re-adds actions. Boring. But oh well, its 100% fine then

Fildrance (approved):
Looks okay to me, except maybe remove action naming - it is different from how file is named, please fix that. All good after that.

ScarKy0:
autocomplete deemed as ["it's finnne i guess"](https://discord.com/channels/310555209753690112/770682801607278632/1383549974897233941) by PJB so im keeping it in for now

ScarKy0:
Test fail unrelated

PR #38319, opened 2025-06-14
NPC spiders sometimes spin webs 🕷️🕸️

qwerltaz:
## About the PR

NPC spiders now spin webs sometimes!

Shiva is exempt in order not to web security constantly.
Salvage wreck space spiders are exempt.

## Why / Balance

- Pretty cool innovation and good balancing for the spider when there are no ghosts to take over. This buffs spiders, but the buff is only powerful when nobody is dead yet!
- Gives a nice _spin_ to the spider, making it more fleshed out when controlled by game AI. When approached, it will sometimes have already spun some thick webs around it, and for the clown spider, some slippery webs.

## Media

image

**Changelog**

🆑 
- add: NPC spiders now spin webs sometimes!

### DIFF
--- a/Content.Server/Spider/SpiderSystem.cs
+++ b/Content.Server/Spider/SpiderSystem.cs
@@ -2,19 +2,46 @@ using System.Linq;
 using Content.Server.Popups;
 using Content.Shared.Spider;
 using Content.Shared.Maps;
+using Content.Shared.Mobs.Systems;
 using Robust.Server.GameObjects;
 using Robust.Shared.Map;
+using Robust.Shared.Player;
+using Robust.Shared.Timing;
 
 namespace Content.Server.Spider;
 
 public sealed class SpiderSystem : SharedSpiderSystem
 {
     [Dependency] private readonly PopupSystem _popup = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly MobStateSystem _mobState = default!;
+
 
     public override void Initialize()
     {
         base.Initialize();
         SubscribeLocalEvent<SpiderComponent, SpiderWebActionEvent>(OnSpawnNet);
+    }
+
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        var query = EntityQueryEnumerator<SpiderComponent>();
+        while (query.MoveNext(out var uid, out var spider))
+        {
+            if (HasComp<ActorComponent>(uid)
+                || !spider.SpawnsWebsAsNonPlayer
+                ||_timing.CurTime < spider.NextWebSpawn)
+                continue;
+
+            spider.NextWebSpawn += TimeSpan.FromSeconds(spider.WebSpawnCooldownSeconds);
+
+            if (_mobState.IsDead(uid))
+                continue;
+
+            OnSpawnNet(uid, spider, new SpiderWebActionEvent());
+        }
     }
 
     private void OnSpawnNet(EntityUid uid, SpiderComponent component, SpiderWebActionEvent args)
--- a/Content.Server/Spider/SpiderSystem.cs
+++ b/Content.Server/Spider/SpiderSystem.cs
@@ -30,14 +30,14 @@ public sealed class SpiderSystem : SharedSpiderSystem
         var query = EntityQueryEnumerator<SpiderComponent>();
         while (query.MoveNext(out var uid, out var spider))
         {
-            if (HasComp<ActorComponent>(uid)
-                || !spider.SpawnsWebsAsNonPlayer
-                ||_timing.CurTime < spider.NextWebSpawn)
+            if (_timing.CurTime < spider.NextWebSpawn)
                 continue;
 
             spider.NextWebSpawn += TimeSpan.FromSeconds(spider.WebSpawnCooldownSeconds);
 
-            if (_mobState.IsDead(uid))
+            if (HasComp<ActorComponent>(uid)
+                || _mobState.IsDead(uid)
+                || !spider.SpawnsWebsAsNonPlayer)
                 continue;
 
             OnSpawnNet(uid, spider, new SpiderWebActionEvent());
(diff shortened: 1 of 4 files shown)


Content.Server/Spider/SpiderSystem.cs
```diff
@@ -2,21 +2,48 @@
...
+                continue;
+
+            spider.NextWebSpawn += TimeSpan.FromSeconds(spider.WebSpawnCooldownSeconds);
+
+            if (HasComp<ActorComponent>(uid)
+                || _mobState.IsDead(uid)
+                || !spider.SpawnsWebsAsNonPlayer)
+                continue;
+
+            OnSpawnNet(uid, spider, new SpiderWebActionEvent());
```
beck-thompson: omega sus just make a function called like `SpawnNet` that you call here and the event

Content.Shared/Spider/SpiderComponent.cs
```diff
@@ -18,6 +18,24 @@ public sealed partial class SpiderComponent : Component
...
+    /// Whether the spider will spawn webs when not controlled by a player.
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public bool SpawnsWebsAsNonPlayer = true;
+
+    /// <summary>
+    /// The cooldown in seconds between web spawns when not controlled by a player.
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public float WebSpawnCooldownSeconds = 120f;
```
beck-thompson: Use a `TimeSpan`!

Content.Shared/Spider/SpiderComponent.cs
```diff
@@ -18,6 +18,24 @@ public sealed partial class SpiderComponent : Component
...
+    /// <summary>
+    /// Whether the spider will spawn webs when not controlled by a player.
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public bool SpawnsWebsAsNonPlayer = true;
+
+    /// <summary>
+    /// The cooldown in seconds between web spawns when not controlled by a player.
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
```
beck-thompson: You can get rid of the VV stuff and just do `[DataField]`  (do this for both)

beck-thompson (changes requested):

beck-thompson:
oh also I think technically this should be like an HTN thing but I think this is fine

qwerltaz:
the failed test class has in its remarks "Its a pretty crude test to try catch issues like this, and possibly should just be disabled"  😭 it fails because spiders spawn webs immediately after spawn and then start a new cooldown, which is exactly what I want to happen.

qwerltaz:
I guess it's better not to spawn webs immediately and wait one cooldown first. Fixed.

Content.Shared/Spider/SharedSpiderSystem.cs
```diff
@@ -16,5 +18,6 @@ public override void Initialize()
     private void OnInit(EntityUid uid, SpiderComponent component, MapInitEvent args)
     {
         _action.AddAction(uid, ref component.Action, component.WebAction, uid);
+        component.NextWebSpawn = _timing.CurTime + component.WebSpawnCooldown;
```
beck-thompson: Make `NextWebSpawn` null by default and if its null do this but in the update loop - having it here is sus

beck-thompson (changes requested):

Content.Shared/Spider/SharedSpiderSystem.cs
```diff
@@ -1,10 +1,12 @@
 using Content.Shared.Actions;
+using Robust.Shared.Timing;
 
 namespace Content.Shared.Spider;
 
 public abstract class SharedSpiderSystem : EntitySystem
 {
     [Dependency] private readonly SharedActionsSystem _action = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
```
beck-thompson: Unused

Content.Shared/Spider/SharedSpiderSystem.cs
```diff
@@ -1,10 +1,12 @@
 using Content.Shared.Actions;
+using Robust.Shared.Timing;
```
beck-thompson: Unused

Content.Shared/Spider/SpiderComponent.cs
```diff
@@ -18,6 +18,24 @@ public sealed partial class SpiderComponent : Component
...
+    /// Whether the spider will spawn webs when not controlled by a player.
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
+    public bool SpawnsWebsAsNonPlayer = true;
+
+    /// <summary>
+    /// The cooldown in seconds between web spawns when not controlled by a player.
+    /// </summary>
+    [DataField]
+    public TimeSpan WebSpawnCooldown = TimeSpan.FromSeconds(120f);
```
beck-thompson: Mind decreasing this time just a little bit? Two minuets is pretty long! Maybe a minute (or even 45 seconds!)? Up to you

Content.Shared/Spider/SpiderComponent.cs
```diff
@@ -18,6 +18,24 @@ public sealed partial class SpiderComponent : Component
     public string WebAction = "ActionSpiderWeb";
 
     [DataField] public EntityUid? Action;
+
+    /// <summary>
+    /// Whether the spider will spawn webs when not controlled by a player.
+    /// </summary>
+    [DataField, ViewVariables(VVAccess.ReadWrite)]
```
beck-thompson: Change `[DataField, ViewVariables(VVAccess.ReadWrite)]` -> `[DataField]` (It defaults to read write)

beck-thompson (approved):
Looks good! A few minor things but it should be fast cleanup

Content.Shared/Spider/SpiderComponent.cs
qwerltaz: True, 2min does seem long. After a short test, I think 45s is good.

southbridge-fur (approved):
Looks good to me.

PR #38331, opened 2025-06-15
fix: no longer accidentally rotate held items

little-meow-meow:
## About the PR

Require the user to hover their cursor over a storage window if they want to rotate an item. This makes it much easier to avoid accidentally rotating an item while you're doing something unrelated.

## Why / Balance

It is extremely easy to accidentally rotate an item by right-clicking something while holding that item, which leads to frequent frustration when you go to put that item back into your bag.

With this change, you must be hovering over a storage window to actually rotate the held item. Unless you're dragging it, because you're probably not going to be interacting with anything else in that case.

## Technical details

`Content.Client/UserInterface/Systems/Storage/StorageUIController.cs` has a method `OnMiddleMouse` (legacy naming) that handles the action for rotating a storage item. This change checks for either item dragging or hovering over a StorageWindow control before rotating it. Very straightforward.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Client/UserInterface/Systems/Storage/StorageUIController.cs
+++ b/Content.Client/UserInterface/Systems/Storage/StorageUIController.cs
@@ -225,6 +225,10 @@ public sealed class StorageUIController : UIController, IOnSystemChanged<Storage
         if (!IsDragging && EntityManager.System<HandsSystem>().GetActiveHandEntity() == null)
             return;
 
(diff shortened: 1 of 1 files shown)


TemporalOroboros:
As someone who has 'swap hands' bound to middle mouse, _thank you_.

beck-thompson (approved):

PR #38336, opened 2025-06-15
end round in 20 seconds after nuke explosion

lzk228:
## About the PR

if a nuke explodes in nuke gamemode or with loneops, the time to the end round (to the lobby) will be 20 seconds

## Why / Balance

nobody wants to float 2 mins in space

## Technical details

add a new ccvar and use it on calling EndRound from NukeOpsRuleSystem

## Media

https://youtu.be/Yc8bakG4Jlw

## Requirements

## Breaking changes

no

**Changelog**

:cl:
- tweak: The round will end in 20 secods after nuke explosion in a nuke round.

### DIFF
--- a/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
+++ b/Content.Server/GameTicking/Rules/NukeopsRuleSystem.cs
@@ -10,6 +10,7 @@ using Content.Server.Shuttles.Events;
 using Content.Server.Shuttles.Systems;
 using Content.Server.Station.Components;
 using Content.Server.Store.Systems;
+using Content.Shared.CCVar;
 using Content.Shared.GameTicking.Components;
 using Content.Shared.Mobs;
 using Content.Shared.Mobs.Components;
@@ -20,6 +21,7 @@ using Content.Shared.NukeOps;
 using Content.Shared.Store;
 using Content.Shared.Tag;
 using Content.Shared.Zombies;
+using Robust.Shared.Configuration;
 using Robust.Shared.Map;
 using Robust.Shared.Random;
 using Robust.Shared.Utility;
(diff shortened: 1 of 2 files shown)


iaada:
I agree with delaying the round end. The score screen has always felt anticlimactic and you don't get to watch the explosion.

lzk228:
> I agree with delaying the round end. The score screen has always felt anticlimactic and you don't get to watch the explosion.

perhaps, you didn't get the point
instead of 2 minutes waiting, it's now 20 seconds between the explosion and lobby

VerinSenpai:
Not a huge deal but I'm not a big fan of such a massive reduction. Theres some fun rp to be had post nuke, similar to zombies winning.

Hitlinemoss:
Conceptually I'm a fan but I think that this kind of prevents cases like end-of-round RP as the last survivors on the cargo shuttle (or end-of-round RP as the nukies heading back to base after a successful mission). I think it might be better to only have the time limit reduced if all players are dead, rather than just always reducing it.

SeamLesss:
I agree that this removes potential for end of the round RP, specifically when crew survives on the cargo shuttle or nukies make it onto their ship. It's probably more complicated but the best solution would be to check if everyone (or mostly everyone) is dead before reducing the round end time to 20 seconds.

walksanatora:
this would remove the funniest post-nuke RP I had. the nukies escaped on the cargo shuttle and recruited them to the syndicate.

SuperGDPWYL:
what if it was conditional and the reduction happens if none of the crew survived?

Cojoke-dot:
In every situations some crew survive the blast, be that from running on a ship or the nuke bugging and not blowing up the entire station. The question is to either focus on the majority of players who are dead or the minority of people alive. 

I would say not to add a condition where this does not trigger, at least in LRP. EOR rp is fun and all but with most people dead it just ends up being ghosts flying around the debris field while the people who ran chat in a ship. With zombies you at least have a swarm of people to watch wreck centcom or admeme cburn but nuke going off is just a pile of trash

Aearo-Deepwater:
Adding my voice to the "post-nuke RP is quite good actually" pile.
I do appreciate that it's boring for the dead, though. Maybe instead of this, there's some way to let the dead players participate a bit more actively?

Something something "Nuke Wraith" ghost roles. I don't know what a Nuke Wraith would be, but it sounds pretty sick, no?

notquitehadouken:
:+1:

i don't really wanna wait 2 minutes so that 75 dead people can watch 5 living crew roleplay

if you wanted to talk about the round we have OOC for that

EthanQix:
> 👍
> 
> i don't really wanna wait 2 minutes so that 75 dead people can watch 5 living crew roleplay
> 
> if you wanted to talk about the round we have OOC for that

OOC is by definition not meant for roleplay.

PR #38344, opened 2025-06-16
Added functionality for gasses to be considered "Internal" and ignore insulation for heating/cooling

Skybailey-dev:
## About the PR

Added an `IsInternal` bool to `GasMixture`, allowing any gas to be designated as "inside" someone. Currently only matters for temperature.
Added `ignoreHeatResistance` to this new `IsInternal` bool.
Added `IsInternal` to Cryotube's gas, allowing a Cryotube to cool off people wearing insulative clothing

## Why / Balance

Adding `IsInternal` to `GasMixture` and giving it `ignoreHeatResistance` is intended to enable my yet-to-start-on lung simulation breathing system, and allowing said system to warm/cool any potential lung-havers. You can see my ideaguys-shizo-post on that later part here. When I have the time to start developing it properly, I will be making regular feedback requests as I figure it out. https://discord.com/channels/310555209753690112/311537926376783886/1383933196680564926

The reason for the Cryopod change is to make cryo more user friendly. It makes sense that cryo would ignore insulation when cooling people, as they are breathing the super cold gas, not just suspended in it. This change makes it so people in hardsuits/jackets don't need to be stripped before being placed into cryotubes, which is always bulky to do and impossible for borgs to do.

## Technical details

Incredibly minor changes.
Added a new `DataField` to `GasMixture` and an `AggressiveInlining` `MarkInternal` void (Not needed but there for consistency, as `Immutable` has an equivalent)
Made `IsInternal` = `ignoreHeatResistance` for `OnAtmosExposedUpdate`'s `ChangeHeat` function.
Made `CryoPodAirComponent` have `IsInternal`

## Media

https://github.com/user-attachments/assets/97144a86-9ffd-4478-83b0-2b8b51c863b9
First section of video is with my change, second section is without it. Demonstrates it working

## Requirements

## Breaking changes

I do not think anything here qualifies. `IsInternal` is by default false unless specified in the `GasMixture`, so shouldn't have any impact outside of specified changes.
**Changelog**

:cl:
- tweak: Cryopods will now ignore insulating clothing when cooling down patients.

### DIFF
--- a/Content.Shared/Atmos/GasMixture.cs
+++ b/Content.Shared/Atmos/GasMixture.cs
@@ -31,6 +31,8 @@ namespace Content.Shared.Atmos
 
         [DataField("immutable")]
         public bool Immutable { get; private set; }
+        [DataField("isInternal")]
+        public bool IsInternal { get; private set; }
 
         [ViewVariables]
         public readonly float[] ReactionResults =
@@ -75,14 +77,15 @@ namespace Content.Shared.Atmos
         {
         }
 
-        public GasMixture(float volume = 0f)
+        public GasMixture(float volume = 0f, bool isInternal = false)
         {
             if (volume < 0)
                 volume = 0;
             Volume = volume;
+            IsInternal = isInternal;
         }
 
-        public GasMixture(float[] moles, float temp, float volume = Atmospherics.CellVolume)
+        public GasMixture(float[] moles, float temp, float volume = Atmospherics.CellVolume, bool isInternal = false)
         {
             if (moles.Length != Atmospherics.AdjustedNumberOfGases)
                 throw new InvalidOperationException($"Invalid mole array length");
@@ -94,6 +97,7 @@ namespace Content.Shared.Atmos
             _temperature = temp;
             Moles = moles;
             Volume = volume;
+            IsInternal = isInternal;
         }
 
         public GasMixture(GasMixture toClone)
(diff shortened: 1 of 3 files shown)


Content.Shared/Atmos/GasMixture.cs
```diff
@@ -106,6 +110,12 @@ public void MarkImmutable()
         {
             Immutable = true;
         }
+        // PR reviewer(s): This is definitley not needed right now, but might be required by future uses of IsInternal? Remove it if you don't think so.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void MarkInternal()
+        {
+            IsInternal = true;
```
LaCumbiaDelCoronavirus: why not just make this, for example, `SetInternal(bool $internal) {IsInternal = $internal}` or something like that?
I dont think this implementation is necessary eitherway

Skybailey-dev: Yea looking at why immutable has it I don't see any reason to have it for internal. Fixing it along with changing IsInternal's private set to a normal set

Skybailey-dev: Fixed

slarticodefast:
This probably should not be part of the GasMixture itself.
I'm currently finishing up this PR https://github.com/space-wizards/space-station-14/pull/36709 which unifies all internal air components into one, including lungs and cryo pods. With that you could just check the component.
But I'm also not sure if I like breaking energy conservation like this.

Skybailey-dev:
> This probably should not be part of the GasMixture itself. I'm currently finishing up this PR #36709 which unifies all internal air components into one, including lungs and cryo pods. With that you could just check the component.

Will do, and it looks simple to do as well. I'll assume GasCanisterSystem on #36709 will be mostly the same as your WIP PR, so I'll code with respect to that. Let me know if that's not the case.

> But I'm also not sure if I like breaking energy conservation like this.

Assuming the physics system works as it should, this should still conserve energy, as the total thermal energy of an interacting system with insulation and without insulation would be the same. Still should care about specific heat etc. Unless there is something weird I don't know with how insulation in SS14 works (possible), it _should_ only impact the speed that two systems can achieve thermal equilibrium.
This is easily testable however, so I shall.

Skybailey-dev:
Changing to a draft until Slart finishes up #36709 :)

PR #38345, opened 2025-06-16
Revert "Randomized Markings"

beck-thompson:
Reverts space-wizards/space-station-14#34765

@widgetbeck

Has test semi random test fails. Seems to happen ~50% of the time in the `AddAndGetSingleLog` test. My computer is really struggling to run the test locally but it looks like the error is something to do with human hair or something? Here is the stack trace:

image

I'd try to debug this further but its been a while now and I cannot figure it out (thanks computer :sob:)


ArtisticRoomba:
Yeah, sorry. This was a very cool addition but unfortunately it was not compatible with latest master. It can be easily re-introduced when the issues with it are fixed.

~Merge queue one day......~

ArtisticRoomba (approved):

PR #36951, opened 2025-04-26
Add group for loadouts

qrwas:
## About the PR

This PR is a continuation of https://github.com/space-wizards/space-station-14/pull/36849 and provides grouping to Loadouts.

## Why / Balance

Adds a bit of structure to the loadouts menu

## Technical details
Changed the logic of displaying loadouts, and added SubLoadoutsContainer.
When calling RefreshLoadouts, all available loadouts are grouped by GroupByTag, which I added to LoadoutPrototype if there is one, or elements are added as a unique group.
I iterate through the groups, if there is more than 1 element in the group, a SubLoadoutsContainer is created, and other elements of the group are written to it.

Additionally, each row is a separate GridContainer, so that the expanded SubLoadoutsContainer is under a specific row

For tests, I added groups to pins, towels and jumpsuit for HoS

After adding a new field [DataField("groupBy")]
public string? GroupByTag { get; private set; } = string.Empty; to LoadoutPrototype, I get the error [ERRO] proto: Attempted to resolve invalid ProtoId<LoadoutPrototype>, which does not affect the script's operation, but I would like to fix it.

## Media

image
image

https://github.com/user-attachments/assets/44f2c84f-ab9a-4d6f-a275-a0e9081c00dd

## Requirements

- [x ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ x] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

PR #38347, opened 2025-06-16
add group for many items and clothing in loadout menu

qrwas:
## About the PR

add group for many items and clothing in loadout menu.

## Why / Balance

Grouping items makes the loadouts menu interface a bit cleaner and more consistent. It also sets a general grouping concept for future pull requests.

## Technical details

Clothing divided by jumpsuit/jumpskirt
musical instruments grouped according to cargo order boxes
some berets and caps are also divided accordingly. (medical doctor as an example)
toys and flowers also in a separate group

## Media

image

image

image

## Requirements

needs #36951 

**Changelog**

:cl: qrwas
- tweak: revised and grouped the catalog of items in the loadouts menu.


TheShuEd:
It's strange to me that “category” is also a loadout option button that can be selected

qrwas:
> It's strange to me that “category” is also a loadout option button that can be selected

actually this is not a category name, but a subject that can also be selected

TheShuEd:
yep, and this is strange

qrwas:
> yep, and this is strange

In my opinion, this is the best solution than making a special button that will be responsible only for opening the group, just to display the name.
This allows you to immediately show the items that have been selected and brings them to the top. This means that if you select an item from a group, you don't have to open that group to see what item you're wearing.

as an option: add a caption above the group - but this will only clutter the interface

PJB3005:
image

I don't see the point of this PR. The groups make sense for items like the LGBT pins, but not this. There's barely anything in these menus as-is, why group them like this?

qrwas:
> I don't see the point of this PR. The groups make sense for items like the LGBT pins, but not this. There's barely anything in these menus as-is, why group them like this?

The idea was to divide the overalls into men's and women's, similar to the colors of a lawyer's suits.

dividing them by type will not change the situation with "overflow" of items much if you divide them by color/purpose

qrwas:
although if think about it, dividing it specifically by color would be more convenient for players. you just look at the color, and then choose another type 🤔🤔🤔🤔

LaCumbiaDelCoronavirus:
> although if think about it, dividing it specifically by color would be more convenient for players. you just look at the color, and then choose another type 🤔🤔🤔🤔

entire idea regarding jumpsuit groups is weird -- it should remain as it was, jumpsuits in specific are not really cluttering anyway

qrwas:
but now I've thought about it too. It's more convenient to immediately look at the "color" assortment, and then choose a skirt or jumpsuit.
I will change this.

qrwas:
Clothes are now grouped by type or color

PJB3005:
> Clothes are now grouped by type or color

This doesn't address my concern. There aren't enough uniforms to warrant shoving most of them in separate groups, which I should remind you causes them to be hidden by default.

If you want to provide some logical visual grouping for these, a simple brace visual around them would be a way better idea.

qrwas:
> > Clothes are now grouped by type or color
> 
> This doesn't address my concern. There aren't enough uniforms to warrant shoving most of them in separate groups, which I should remind you causes them to be hidden by default.
> 
> If you want to provide some logical visual grouping for these, a simple brace visual around them would be a way better idea.

globally, I didn't group all the clothes into groups, where there were only up to 4 pieces (or something like that)

PR #38350, opened 2025-06-16
Changed SpriteComponent layer keys to be objects

Citrea-Lingua:
## About the PR

Changes SpriteComponent LayerKey to be an object, rather than a string

## Why / Balance

This appears to be a bit of unfinished work, left over from moving sprite layer keys to Objects for using the HumanoidVisualLayer enum on body parts. The transformation to a string here is unnecessary (combined with an adjustment to the toolbox I am also PRing) and actively causes issues when one tries to actually use this method, which allows objects, with anything but string layer keys.

## Technical details

Removed ToString() so the object is passed on in whole.

## Requirements

(Note that while I did test it, there is a possibility I missed something. I can be extremely oblivious at times .-. )

## Breaking changes
This does not change anything about how anything works, as far as I can tell, beyond fixing that one specific issue.

Relates to: https://github.com/space-wizards/RobustToolbox/pull/6024

### DIFF
--- a/Content.Client/DisplacementMap/DisplacementMapSystem.cs
+++ b/Content.Client/DisplacementMap/DisplacementMapSystem.cs
@@ -70,7 +70,7 @@ public sealed class DisplacementMapSystem : EntitySystem
         }
 
         var displacementLayer = _serialization.CreateCopy(displacementDataLayer, notNullableOverride: true);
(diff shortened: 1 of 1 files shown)


Citrea-Lingua:
Aaand of course, I missed a thing, it seems, going by those submodule checker warnings.

Citrea-Lingua:
_sigh_
Something else seems to be broken now, *somehow*, I'll be back when I fix this
_grumble grumble_

PR #38351, opened 2025-06-16
changes SpriteComponent to use object as LayerKey

Citrea-Lingua:
## About the PR

Changes SpriteComponent LayerKey to be an object, rather than a string

## Why / Balance

This appears to be a bit of unfinished work, left over from moving sprite layer keys to Objects for using the HumanoidVisualLayer enum on body parts. The transformation to a string here is unnecessary (combined with an adjustment to the toolbox I am also PRing) and actively causes issues when one tries to actually use this method, which allows objects, with anything but string layer keys.

## Technical details

Removed ToString() so the object is passed on in whole.

## Requirements

(Note that while I did test it, there is a possibility I missed something. I can be extremely oblivious at times .-. )

## Breaking changes
This does not change anything about how anything works, as far as I can tell, beyond fixing that one specific issue.

Requires https://github.com/space-wizards/RobustToolbox/pull/6025


Citrea-Lingua:
Maybe it wasn't my push which was messed up?
_sigh_
No idea why this is failing the build/test, it literally works fine on my end. Seems to not be taking the changes from the PR to engine into account, which would definitely cause those errors.

Tayrtahn:
Please see the documentation on creating PRs with engine changes: https://docs.spacestation14.com/en/general-development/tips/prs-with-engine-changes.html#making-the-pr

(You need to revert the submodule changes and add `Requires https://github.com/space-wizards/RobustToolbox/pull/6025` to your PR description)

Citrea-Lingua:
Aha! I completely missed that page somehow .-.
Thanks, will do.

Citrea-Lingua:
That... that conflict is for a patchwork bugfix which doesn't fix the underlying issue of *why* it happened in the first place. Guess I'll be overwriting that particular change.

Citrea-Lingua:
closing, as it's become fairly obvious noone is willing to touch that particular part of the toolbox, and I frankly do not have the time or spoons to do anything more with it.
May whoever runs into an issue with this next find my code at least somewhat useful, if they want to pick up the torch :sweat_smile:

PR #38352, opened 2025-06-16
mantis blade

Golubgik:
## About the PR

Added a new weapon to the Traitor and Nuclear Operatives uplink - the Mantis Blade. It replicates the Changeling's Arm Blade and costs 10 Telecrystals, with a maximum discount of 5

## Why / Balance
This weapon allows the traitor to remain more undetected and mobile, but it will also cost more.

## Media

https://github.com/user-attachments/assets/bd70a3c4-1da9-4095-b9ca-e90245c74d46

## Requirements

**Changelog**
:cl: Golub
- add: Added Mantis Blade to Traitor and Nuclear Ops uplink


K-Dynamic:
Interesting

ilovehans10:
Good work choom. Now all we need is a ripperdoc to install it :godo:

On a more serious note, high slash + pierce seems like it will apply a ton of bleed...

Golubgik:
> On a more serious note, high slash + pierce seems like it will apply a ton of bleed...

Anyway, I don't have to worry about balance because the maintainers do it.

keronshb:
I don't think this should be added to traitors at all. It removes identity from the upcoming Changeling antagonist.
It's also the reason we don't have Cult or Wizard items in the uplink.

muburu:
a "holo-sword" implanter would make alot more sense thematically i think, giving them a wacky biological changeling-esque weapon seems super unfitting

Cojoke-dot:
Video is not working, may just be a me thing though.

I'm not the biggest fan of weapons getting cloned between 2 major antags... I guess this is weaker than the ling blade but still.

Maybe increase the price to 12 or drop it to 8 so it fits with the 4tc price intervals. It feels like a less versatile esword though. Once someone calls out you have it sec can easily check you for an implant and there isnt much you can do to hide it. I would say due to this it should be the same price as the esword. 

Esword:
- makes constant sound and light WHEN ON, easy to hide when off
- 30 damage no bleeding
- 25% reflect
- 2 by 1 easy to hide

Mantis blade:
- unique sounds when swung
- 30 damage and bleed
- Cant be dropped(nobody is going to steal it)
- easy to hide in person but can be found easily if known you had one at any point

Mantis is better damage but esword has reflect. Esword can be removed but mantis can't be found in a normal search. Esword makes more sound but there are toys everywhere with said sound, mantis makes a unique swing sound but is overall quieter.

I would say drop it to 8tc but I may be underestimating this weapon.

SlimmSlamm:
> I don't think this should be added to traitors at all. It removes identity from the upcoming Changeling antagonist. It's also the reason we don't have Cult or Wizard items in the uplink.

We aren’t going to be seeing Changelings anytime soon I don’t even expect them until later next year so I don’t see why we can’t have this as a fun item until then. I don’t really think having one or two items from another antagonist is bad either way. Syndicates can be the jack of all trades and whatever lore there is it can just be the faction wanting to mimic as much as they can from other space creatures or factions. 

The Wizard also has some pretty similar stuff to traitors so it’s not like this is a first.

ScarKy0:
Yeah I'm gonna be the 2nd concern and closing this PR.

While I honestly love the idea, this probably will not work for something like traitor. Even if, in the current state this is just a better energy sword while at the same time taking away from the ling.
Now "better" because for few key reasons. It cannot be disarmed, is perfectly concealable and cannot be handcuffed away when equipped(though this might be more of a bug)
15/15 slash/pierce is also huge damage for this amount of versitility, the bleed is would do is NOT something to laugh about.

Something like this works very well for the changeling however consider that is their ONLY proper weapon (at least for now, theyre unreleased, you get the idea) and will probably be balanced afterwards anyways as testing with JUST the armblade do not give very good results as how it would work paired with a transform ability.

It could arguably be used for wizard's spectral blade but I honestly have no idea how it works so cannot confirm that either.

ScarKy0:
> We aren’t going to be seeing Changelings anytime soon

You might be surprised. But probably as admeme only at the start.

TheExetron:
github broke

ScarKy0:
??

SlimmSlamm:
> > We aren’t going to be seeing Changelings anytime soon
> 
> You might be surprised. But probably as admeme only at the start.

I’m open to surprises but these take time to development especially with an open source game and people constantly losing interest or important things coming up. I’m taking it the same as Wizard and we all know that took a pretty long time to develop.

PR #38357, opened 2025-06-16
MapRenderer code fixes

PJB3005:
I'm gonna die dude

See commits

Requires https://github.com/space-wizards/RobustToolbox/pull/6029


Content.MapRenderer/MapViewerData.cs
```diff
@@ -43,16 +45,19 @@ public sealed class LayerGroup
...
             Source = new GroupSource
             {
-                Url = "https://i.imgur.com/3YO8KRd.png",
-                Extent = new Extent(6000, 4000)
+                Url = output.ReferenceResourceFile(resourceManager, new ResPath("/Textures/Parallaxes/layer1.png")),
+                Extent = new Extent(6000, 4000),
             },
+            // REMOVED because using Imgur as a CDN is utterly cringe,
+            // and I can't figure out the source files for these.
+            /*
```
juliangiebel: The source dmi is inside the attribution for that png.
https://github.com/space-wizards/space-station-14/blob/master/Resources/Textures/Parallaxes/meta.json
https://github.com/tgstation/tgstation/blob/master/icons/effects/parallax.dmi.

Please add the other layers

Content.MapRenderer/Program.cs
```diff
@@ -104,45 +102,118 @@ internal static async Task Main(string[] args)
...
+                // so it would actually go through the list of prototypes and match file name arguments
+                // via a *very* coarse check.
+                //
+                // So if we have any input filenames that don't exist... we run the old behavior.
+                // Yes by the way this means a typo means spinning up an entire integration pool pair
+                // before the map renderer can report a proper failure.
+                //
+                // Note that this legacy processing is very important! The map server currently relies on it,
+                // because it wants to work with file names, but we *need* to resolve the input to a prototype
+                // to properly export viewer JSON data.
```
juliangiebel, line 126: This is not true. I adapted it to provide relative paths. Maybe we want to support instances running an older version of the map server but we might just as well remove the legacy behaviour

juliangiebel (changes requested):
Code looks good except for two things. 
I probably missed something. I don't know a lot about how our integration test code works

Content.MapRenderer/Program.cs
PJB3005: > This is not true. I adapted it to provide relative paths.

Yes, but, as I explained on Discord, we *need* to keep using the old behavior of doing path -> prototype resolution, otherwise we cannot properly export correct map names.

Content.MapRenderer/MapViewerData.cs
PJB3005: That `meta.json` isn't even correct since that isn't an RSI. FFS.

juliangiebel (approved):

PR #38358, opened 2025-06-16
Fix SharedActionSystem spamming errors

slarticodefast:
## About the PR
This should fix https://github.com/space-wizards/space-station-14/issues/38293 but I only did limited testing so far.
So before merging it this should be thoroughly tested to make sure this does not introduce other problems.

## Why / Balance
server crashes bad

## Technical details
Deleted entities granting actions were not properly removed from the ActionsComponent.
To reproduce, spawn a flashlight, pick it up, and delete it while it is in your hand.
Then every time your actions change, for example when (un)equipping your PDA the server will try to network the action entity and fails.

This was caused because GetAction returns null when the entity granting the action is terminating.
However we still need to get it at this point as part of `RemoveProvidedActions`
grafik

which in turn should be called when the flashlight is removed from your hand when deleting it
grafik

## Requirements

## Breaking changes
none

**Changelog**
:cl:
- fix: Fixed actions sometimes not disappearing and causing server errors.

### DIFF
--- a/Content.Shared/Actions/SharedActionsSystem.cs
+++ b/Content.Shared/Actions/SharedActionsSystem.cs
@@ -108,7 +108,7 @@ public abstract class SharedActionsSystem : EntitySystem
     /// </summary>
     public Entity<ActionComponent>? GetAction(Entity<ActionComponent?>? action, bool logError = true)
     {
-        if (action is not {} ent || TerminatingOrDeleted(ent))
+        if (action is not {} ent || Deleted(ent))
             return null;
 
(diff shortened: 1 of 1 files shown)


Tayrtahn (approved):
This fixes the bug and I haven't been able to find any side effects!

ScarKy0:
Should this not be hotfixed?

Fildrance:
Worth a try as a hotfix imo. We are in a bit of a desperate situation.

Fildrance (approved):

slarticodefast:
Rebased to stable.

PR #38370, opened 2025-06-17
Generic Numeric Alerts

ScarKy0:
## About the PR

Added a system for displaying generic numeric alerts (such as revenant's essence display)
Code taken from @EmoGarbage404's blob PR.

## Why / Balance

Less hardcoded stuff. Will probably be useful for things such as ling's biomass/chemicals to make the visuals more responsive.

## Technical details

Adds a new clientside system for creating a generic counter alert.
The system uses an event to learn what amount it should display. The event is subscribed to in it's respective system (revenant, in this case) and updates the sprite layers on the alert accordingly.

## Media

Nothing to show in-game.

## Requirements

## Breaking changes

**Changelog**

Not player facing.

### DIFF
--- a/Content.Client/Alerts/GenericCounterAlertSystem.cs
+++ b/Content.Client/Alerts/GenericCounterAlertSystem.cs
@@ -28,11 +28,15 @@ public sealed class GenericCounterAlertSystem : EntitySystem
         if (!ev.Handled)
             return;
 
+        // It cannot be null if its handled, but good to check to avoid ugly null ignores.
+        if (ev.Amount == null)
+            return;
+
         // How many digits can we display
         var maxDigitCount = GetMaxDigitCount((ent, ent, sprite));
 
         // Clamp it to a positive number that we can actually display in full (no rollover to 0)
-        var amount = (int) Math.Clamp(ev.Amount!.Value, 0, Math.Pow(10, maxDigitCount) - 1);
+        var amount = (int) Math.Clamp(ev.Amount.Value, 0, Math.Pow(10, maxDigitCount) - 1);
 
         // This is super wack but ig it works?
         var digitCount = ent.Comp.HideLeadingZeroes
--- a/Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
+++ b/Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
@@ -37,7 +37,8 @@ public sealed partial class GenericCounterAlertComponent : Component
 
     /// <summary>
     /// Digits that can be displayed by the alert, represented by their sprite layer.
-    /// Ones would be on layer "1", tens on layer "10" etc.
+    /// Order defined corresponds to the digit it affects. 1st defined will affect 1st digit, 2nd affect 2nd digit and so on.
+    /// In this case ones would be on layer "1", tens on layer "10" etc.
     /// </summary>
     [DataField]
     public List<string> DigitKeys = new()


ScarKy0:
Cleaned up and all that. Should be review ready.

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// If true, "005" would display as "5".
+    /// </summary>
+    [DataField]
+    public bool HideLeadingZeroes = true;
+
+    [DataField]
+    public Vector2i AlertSize = new(32, 32);
+
+    [DataField]
+    public List<string> DigitKeys = new()
```
slarticodefast: add documentation

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    [DataField]
+    public Vector2i AlertSize = new(32, 32);
+
+    /// <summary>
+    /// Digits that can be displayed by the alert.
+    /// Ones, Tens, Hundreds, etc.
+    /// </summary>
+    [DataField]
+    public List<string> DigitKeys = new()
```
LaCumbiaDelCoronavirus, line 44: An `int` of how many digits can be displayed seems more intuitive than this approach.

Content.Client/Alerts/GenericCounterAlertSystem.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// </summary>
+    /// <returns>The number of digits.</returns>
+    private int GetMaxDigitCount(Entity<GenericCounterAlertComponent, SpriteComponent> ent)
+    {
+        var comp = ent.Comp1;
+        var sprite = ent.Comp2;
+
+        for (var i = comp.DigitKeys.Count - 1; i >= 0; i--)
+        {
+            if (_sprite.LayerExists((ent.Owner, sprite), comp.DigitKeys[i]))
```
LaCumbiaDelCoronavirus: You can just directly write `ent.Comp2` and remove the `sprite` definition.

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
ScarKy0: It's like that cause that is how the sprite layers that are edited are called.
If you want the 1s to be modified then it edits the layer "1", 10s modifies the layer "10" etc

Content.Client/Alerts/GenericCounterAlertSystem.cs
```diff
@@ -0,0 +1,86 @@
...
+    private int GetMaxDigitCount(Entity<GenericCounterAlertComponent, SpriteComponent> ent)
+    {
+        for (var i = ent.Comp1.DigitKeys.Count - 1; i >= 0; i--)
+        {
+            if (_sprite.LayerExists((ent.Owner, ent.Comp2), ent.Comp1.DigitKeys[i]))
+                return i + 1;
+        }
+
+        return 0;
+    }
```
LaCumbiaDelCoronavirus, line 89: example with a hypothetical `Digits` integer that defines how many digits there is on the counter:
```suggestion
    private int GetMaxDigitCount(Entity<GenericCounterAlertComponent, SpriteComponent> ent)
    {
        for (var i = ent.Comp1.Digits; i >= 0; i--)
        {
            if (_sprite.LayerExists((ent.Owner, ent.Comp2), Math.Pow(1, i)))
                return i;
        }

        return 0;
    }
```

ScarKy0: I'm not entirely if its great cause this assumes that the layers will always be 1 10 100, while the previous approach lets you specifiy the layer names if you need them to be different. I'll wait for a 2nd opinion.

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Content.Shared.Alert.Components;
+
+/// <summary>
+/// This is used for an alert which simply displays a generic number
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class GenericCounterAlertComponent : Component
+{
+    /// <summary>
+    /// The width, in pixels, of an individual glyph, accounting for the space between glyphs.
```
Fildrance, line 12: i'd say better specify if this between glyph and NEXT glyph (so u need to account for spacing just from 1 side).

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
```diff
@@ -0,0 +1,62 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Alert.Components;
+
+/// <summary>
+/// This is used for an alert which simply displays a generic number
```
Fildrance: missing dot at the end of comment :3
also maybe we should mention that it can actually display some number over other sprite? it can, right?

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
```diff
@@ -0,0 +1,62 @@
...
+
+/// <summary>
+/// Event raised to gather the amount the alert will display.
+/// </summary>
+/// <param name="Alert">The alert which is currently requesting an update.</param>
+/// <param name="Amount">The number to display on the alert.</param>
+[ByRefEvent]
+public record struct GetGenericAlertCounterAmountEvent(AlertPrototype Alert, int? Amount = null)
+{
+    public bool Handled => Amount != null;
```
Fildrance: Amount.HasValue is more readable?

Content.Client/Alerts/GenericCounterAlertSystem.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    private void OnUpdateAlertSprite(Entity<GenericCounterAlertComponent> ent, ref UpdateAlertSpriteEvent args)
+    {
+        var sprite = args.SpriteViewEnt.Comp;
+
+        var ev = new GetGenericAlertCounterAmountEvent(args.Alert);
+        RaiseLocalEvent(args.ViewerEnt, ref ev);
+
+        if (!ev.Handled)
```
Fildrance, line 28: To be honest, i am not sold on this one - first you call Handled and then u use ignore on null check...
You might want to use something like  MemberNotNullWhen but thats gonna cause more pain then pleasure i feel (as it works only on props and methods i think?)

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    [DataField]
+    public Vector2i AlertSize = new(32, 32);
+
+    /// <summary>
+    /// Digits that can be displayed by the alert, represented by their sprite layer.
+    /// Ones would be on layer "1", tens on layer "10" etc.
+    /// </summary>
+    [DataField]
+    public List<string> DigitKeys = new()
```
Fildrance, line 44: i suppose the code (GetMaxDigitCount) is expecting that this thing should be ordered.
the uff thing here is that it COULD be not sorted on input, as it is a dataField, and u kinda have to either 
1. make it property and sort it in backing field and use getter setter here
2. explicitly say that it SHOULD BE SORTED AND IN WHICH ORDER.

Content.Client/Alerts/GenericCounterAlertSystem.cs
```diff
@@ -0,0 +1,86 @@
...
+        // ReSharper disable once PossibleLossOfFraction
+        var baseOffset = (ent.Comp.AlertSize.X - digitCount * ent.Comp.GlyphWidth) / 2 * (1f / EyeManager.PixelsPerMeter);
+
+        for (var i = 0; i < ent.Comp.DigitKeys.Count; i++)
+        {
+            if (!_sprite.LayerMapTryGet(ent.Owner, ent.Comp.DigitKeys[i], out var layer, false))
+                continue;
+
+            var result = amount / (int) Math.Pow(10, i) % 10;
+            _sprite.LayerSetRsiState(ent.Owner, layer, $"{result}");
```
Fildrance: interpolation is not a shortcut to ToString(), use ToString here - or u will allocate x2 here
image

Content.Client/Alerts/GenericCounterAlertSystem.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                if (!_sprite.LayerMapTryGet(ent.Owner, ent.Comp.DigitKeys[i], out var layer, false))
+                    continue;
+
+                _sprite.LayerSetVisible(ent.Owner, layer, i <= digitCount - 1);
+            }
+        }
+
+        // ReSharper disable once PossibleLossOfFraction
+        var baseOffset = (ent.Comp.AlertSize.X - digitCount * ent.Comp.GlyphWidth) / 2 * (1f / EyeManager.PixelsPerMeter);
```
Fildrance, line 58: you calculate the thing even if u dont use... probably could condition it out? it is executed for each frame >_>

Content.Client/Alerts/GenericCounterAlertSystem.cs
Fildrance: i do agree that its better to have it as string and not calc using pow here...

Fildrance:
And btw it looks like its not related to this PR but rev have wrong sprite icon mode on his abilities actions 
image

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
ScarKy0: It's designed for alerts. It technically could if you tweak it a bit but its not the point

ScarKy0: Unless you mean over another sprite of the alert, then yes, it can

Content.Client/Alerts/GenericCounterAlertSystem.cs
ScarKy0: I mean systems can just mark it as handled manually but checking for a value is just easier
Ignoring it being null makes sense cause it CANNOT be null at that point, as it was handled

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
ScarKy0: It doesn't expect it to be ordered. These simply are the layers it will modify, index being the number it will correspond to to.
It might as well be "whiskey" "echo" "hotel" and these will be for the 1st number, 2nd number and 3rd number

Content.Client/Alerts/GenericCounterAlertSystem.cs
ScarKy0: I mean I could put it inside of the for loop, but is that much better? Otherwise the code looks ugly

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
ScarKy0: > A 3 pixel wide glyph with one pixel of space between it and the next would be a width of 4.

It does specify it though. It accounts for the next glyph, so it is just from one side

Fildrance: i am too lazy to argue

Content.Client/Alerts/GenericCounterAlertSystem.cs
Fildrance: well fine, i agree that ugly code makes hands itchy... let it stay... but thats suboptimal and now u need to sleep with that thing in mind

Content.Shared/Alert/Components/GenericCounterAlertComponent.cs
Fildrance: at least add this info into comment

Fildrance (approved):

Issue #38355, opened 2025-06-16
Lobby Music Sound Collection should be a CVAR


EmoGarbage404:
presently, it's just hardcoded as a const string in `ContentAudioSystem`. Not being able to easily change it means oftentimes conflicting with the upstream lobby selection which is generally a pain. Pulling it out is a pretty simple in-and-out 15 minute adventure.

PR #38375, opened 2025-06-17
Lobby Audio from static to CVar

beesterman:
## About the PR

Changed Lobby Music Sound Collection from static string to Cvar editable by admin cli
## Why / Balance
Requested in issue #38355 
## Technical details
Added the LobbyMusicCollection Cvar in CCVars.Audio and refactored the code in the ContentAusioSystem to use the Cvar. Also added code to update the music collection and reshuffle the lobby music from the admin cli.

## Requirements

**Changelog**

### DIFF
--- a/Content.Server/Audio/ContentAudioSystem.cs
+++ b/Content.Server/Audio/ContentAudioSystem.cs
@@ -3,13 +3,13 @@ using Content.Server.GameTicking;
 using Content.Server.GameTicking.Events;
 using Content.Shared.Audio;
 using Content.Shared.Audio.Events;
+using Content.Shared.CCVar;
 using Content.Shared.GameTicking;
 using Robust.Server.Audio;
 using Robust.Shared.Audio;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
 using Robust.Shared.Configuration;
-using Content.Shared.CCVar;
 
 namespace Content.Server.Audio;
 
@@ -27,10 +27,12 @@ public sealed class ContentAudioSystem : SharedContentAudioSystem
     {
         base.Initialize();
 
-        //Onvalue chage changes the music collection and reshuffles the playlist to update the lobby music
-        _cfg.OnValueChanged(
+        //changes the music collection and reshuffles the playlist to update the lobby music
+        Subs.CVar(
+            _cfg,
             CCVars.LobbyMusicCollection,
-            x => {
+            x =>
+            {
                 _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(x);
                 _lobbyPlaylist = ShuffleLobbyPlaylist();
             },
--- a/Content.Server/Audio/ContentAudioSystem.cs
+++ b/Content.Server/Audio/ContentAudioSystem.cs
@@ -30,6 +30,8 @@ public sealed class ContentAudioSystem : SharedContentAudioSystem
         _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(_cfg.GetCVar(CCVars.LobbyMusicCollection));
         _lobbyPlaylist = ShuffleLobbyPlaylist();
 
+        _cfg.OnValueChanged(CCVars.LobbyMusicCollection, x => _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(x), true);
+
         SubscribeLocalEvent<RoundEndMessageEvent>(OnRoundEnd);
         SubscribeLocalEvent<PlayerJoinedLobbyEvent>(OnPlayerJoinedLobby);
         SubscribeLocalEvent<RoundRestartCleanupEvent>(OnRoundCleanup);
--- a/Content.Server/Audio/ContentAudioSystem.cs
+++ b/Content.Server/Audio/ContentAudioSystem.cs
@@ -7,9 +7,10 @@ using Content.Shared.CCVar;
 using Content.Shared.GameTicking;
 using Robust.Server.Audio;
 using Robust.Shared.Audio;
+using Robust.Shared.Configuration;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
-using Robust.Shared.Configuration;
+
 
 namespace Content.Server.Audio;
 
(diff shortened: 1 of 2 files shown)


Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -8,17 +8,17 @@
 using Robust.Shared.Audio;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
+using Robust.Shared.Configuration;
+using Content.Shared.CCVar;
```
slarticodefast: sort alphabetically

Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -27,8 +27,14 @@ public override void Initialize()
...
-        _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(LobbyMusicCollection);
-        _lobbyPlaylist = ShuffleLobbyPlaylist();
+        //Onvalue chage changes the music collection and reshuffles the playlist to update the lobby music
+        _cfg.OnValueChanged(
+            CCVars.LobbyMusicCollection,
+            x => {
+                _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(x);
+                _lobbyPlaylist = ShuffleLobbyPlaylist();
+            },
+            true);
```
slarticodefast: just use `Subs.CVar` for this

slarticodefast (changes requested):

Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -3,22 +3,22 @@
 using Content.Server.GameTicking.Events;
 using Content.Shared.Audio;
 using Content.Shared.Audio.Events;
+using Content.Shared.CCVar;
 using Content.Shared.GameTicking;
 using Robust.Server.Audio;
 using Robust.Shared.Audio;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
+using Robust.Shared.Configuration;
```
slarticodefast:
```suggestion
using Robust.Shared.Configuration;
using Robust.Shared.Prototypes;
using Robust.Shared.Random;
```

Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -27,8 +27,16 @@ public override void Initialize()
...
 
-        _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(LobbyMusicCollection);
-        _lobbyPlaylist = ShuffleLobbyPlaylist();
+        //changes the music collection and reshuffles the playlist to update the lobby music
+        Subs.CVar(
+            _cfg,
+            CCVars.LobbyMusicCollection,
+            x =>
+            {
+                _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(x);
```
slarticodefast: might need some error handling in case the string is not a valid proto Id

beesterman: Potentially fixed this. Im a little worried that this could cause a infinite loop if the default is ever changed but from what I have seen there is no way to pull the default value from the Cvar only set it. I put notes on the Cvar. Marking it resolved for now.

Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -27,8 +28,28 @@ public override void Initialize()
...
-        _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(LobbyMusicCollection);
-        _lobbyPlaylist = ShuffleLobbyPlaylist();
+        //changes the music collection and reshuffles the playlist to update the lobby music
+        Subs.CVar(
+            _cfg,
+            CCVars.LobbyMusicCollection,
+            x =>
+            {
+
+                try
```
slarticodefast: just use `TryIndex` instead of catching the error

Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -27,8 +28,28 @@ public override void Initialize()
...
+                {
+                    _lobbyMusicCollection = _prototypeManager.Index<SoundCollectionPrototype>(x);
+                }
+                catch
+                {
+                    Log.Error("No matching sound collection found. Defaulting to LobbyMusic collection");
+                    //This realy should default to the default set inside of the Cvar to avoid a potential loop
+                    //  if the default is ever changed.
+                    _cfg.SetCVar("ambience.lobby_music_collection", "LobbyMusic", true);
```
slarticodefast: This is problematic because `LobbyMusic` is not verified for existance and now it is just as hardcoded as it was before.
Instead we should make `_lobbyMusicCollection` nullable and just not let it play any music in that case

Content.Shared/CCVar/CCVars.Audio.cs
```diff
@@ -58,4 +60,10 @@ public sealed partial class CCVars
         public static readonly CVarDef<float> InterfaceVolume =
             CVarDef.Create("audio.interface_volume", 0.50f, CVar.ARCHIVE | CVar.CLIENTONLY);
 
+    /// <summary>
+    ///     Lobby music collection string
+    /// </summary>
+    [CVarControl(AdminFlags.VarEdit)]
+    public static readonly CVarDef<string> LobbyMusicCollection =
+        CVarDef.Create("ambience.lobby_music_collection", "LobbyMusic", CVar.REPLICATED | CVar.SERVER);
```
PJB3005: `ambience` does not seem like the right cvar category for this?

Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -27,8 +28,26 @@ public override void Initialize()
...
+            {
+                //Checks to see if the sound collection exists. If it does change it if not set a empty prototype
+                // as the new _lobbyMusicCollection meaning it wont play anything in the lobby.
+                if(_prototypeManager.TryIndex<SoundCollectionPrototype>(x, out var outputSoundCollection))
+                {
+                    _lobbyMusicCollection = outputSoundCollection;
+                }
+                else
+                {
+                    _lobbyMusicCollection = new SoundCollectionPrototype();
```
PJB3005: Dynamically instantiating prototypes creates invalid objects and is not a good idea. Set the field to null if it's invalid.

PJB3005: Also, this should report an error.

PJB3005 (changes requested):

Content.Server/Audio/ContentAudioSystem.cs
```diff
@@ -3,32 +3,52 @@
...
+            {
+                //Checks to see if the sound collection exists. If it does change it if not defaults to null
+                // as the new _lobbyMusicCollection meaning it wont play anything in the lobby.
+                if(_prototypeManager.TryIndex<SoundCollectionPrototype>(x, out var outputSoundCollection))
+                {
+                    _lobbyMusicCollection = outputSoundCollection;
+                }
+                else
+                {
+                    Log.Error("Attempted to change Lobby Music Sound collection. No matching collection was found. Defaulting to null");
```
PJB3005: Repeat the actual value entered in the log entry.
```suggestion
                    Log.Error($"Invalid Lobby Music sound collection specified: {x}");
```

PJB3005 (approved):

slarticodefast (approved):

PR #38376, opened 2025-06-17
fix: spellbooks can have infinite charges

### LINKED ISSUE #38343: Non-wizard spellbooks are broken
## Description

Spellbook prototypes like the blink prototype were meant to have infinite charges as specified by their value of -1 in the SpellActions field of the SpellbookComp, but the action charges refactor (#33993) removed this meaning. This means we no longer have spellbooks with infinite charges, which was most prototype'd ones.

**Reproduction**

1. Use blink spellbook.
2. Get blink action with zero chargers.
3. Cool.

**Screenshots**

**Additional context**

CC @metalgearsloth, do you have a particular desire to look into this or should I give it a go.

perryprog:
## About the PR

Restore functionality to spellbooks with infinite charges. (Currently all spawnable entities with "spellbook" in their name.)

## Why / Balance

Fixes #38343.

## Technical details

I kinda hate the -1 having special meaning thing and would rather it be represented differently, but that'd end up being a breaking change. Happy to adjust if it's desired though.

## Requirements

## Breaking changes

- SpellbookComponent.SpellActions now has its values as optional ints. You should set these to null for a given spell for the spell to have infinite charges, instead of setting -1.

**Changelog**

:cl:
- fix: Spellbooks with infinite charges work again.

### DIFF
--- a/Content.Shared/Magic/Components/SpellbookComponent.cs
+++ b/Content.Shared/Magic/Components/SpellbookComponent.cs
@@ -16,8 +16,10 @@ public sealed partial class SpellbookComponent : Component
     [ViewVariables]
     public readonly List<EntityUid> Spells = new();
 
+    // The three fields below are just used for initialization.
     /// <summary>
-    /// The three fields below is just used for initialization.
+    /// Dictionary of spell prototypes to charge counts.
+    /// If the charge count if negative one, it means the spell has infinite charges.
     /// </summary>
     [DataField]
     [ViewVariables(VVAccess.ReadWrite)]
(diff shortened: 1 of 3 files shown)


beck-thompson (changes requested):
I personally think your gut is right and having -1 be infinite charge is sus. Make it another field or something!

perryprog:
Hmm, let's try making the charge count optional I guess? I suppose null as a magic value is slightly better.

The only other option I can think of is making a SpellbookSpell struct that has a charges and isInfinite field, but that feels like it'd get messy on the yaml side of things.

I also don't want to make it a component field itself since then you can't have a spellbook that grants one spell with limited charges and another with infinite charges.

beck-thompson:
> The only other option I can think of is making a SpellbookSpell struct that has a charges and isInfinite field, but that feels like it'd get messy on the yaml side of things.

Didn't realize how sus it was yeah null is fine imo :thinking:

beck-thompson (approved):
I hate the `charges is { } count` c# evil thing but because your such a pro perry I merge

perryprog:
Pattern matching my beloved…

I’ve started to like it over `if (awa.thing != null) florp(awa.thing.Value)` to be honest

Issue #36339, opened 2025-04-06
SMES connect to cable terminals that aren't facing them


PJB3005:
![Image](https://github.com/user-attachments/assets/0f68d4b2-b1d6-47ed-9c88-5c1716eeaab4)

Depends on construction order.

deltanedas:
maybe related to cable terminals sometimes blocking connections to the wrong side of hv wires depending on construction/wiring order

ArtisticRoomba:
They connect but don't share power. So not sure what to do here.

PJB3005:
> They connect but don't share power. So not sure what to do here.

It is literally impossible for them to not share the power connection given my screenshot.

ArtisticRoomba:
Yeah, I re-tested [#38377](https://github.com/space-wizards/space-station-14/pull/38377) and see it now. 3 AM me.

whatston3:
CableTerminalPortNode doesn't care about the direction of the cable terminal in GetReachableNodes.  It should.

whatston3:
Should be fixed at the link above.

![Image](https://github.com/user-attachments/assets/0fff440f-a129-44df-bfc4-994c9e7395d6)

PR #38377, opened 2025-06-17
Fix adjacent north/south cable terminals breaking east/west HV cable

whatston3:
## About the PR

Fixes an issue with adjacent cable terminals breaking connections they shouldn't.

## Why / Balance

Fixes (not) https://github.com/space-wizards/space-station-14/issues/36339#issuecomment-2781189884 (as far as I can tell - _only_ the comment, not the larger issue)

## Technical details

C# one-liner.  Having a cable terminal across from you does not mean you have no more nodes to connect to.

## Media

This setup is impossible to create without this PR - the bottom row of cables would be split in two between the SMESes.
image

## Requirements

## Breaking changes

**Changelog**

No CL, really only of use for mappers.

### DIFF
--- a/Content.Server/Power/Nodes/CableNode.cs
+++ b/Content.Server/Power/Nodes/CableNode.cs
@@ -52,7 +52,6 @@ namespace Content.Server.Power.Nodes
                         {
                             // Target tile has a terminal towards us, block the direction.
                             terminalDirs |= 1 << (int) dir;
-                            break;
                         }
                     }
                 }


K-Dynamic:
I raised this problem to the mapper of Core a while ago, good to see it fixed via code

ArtisticRoomba:
Works, thanks for the fix, but the PR doesn't fix the comment, though the connected nodes don't share power. So not sure what to do.

PR #38338, opened 2025-06-15
Command resolve killing, LEC conversions, and general cleanup.

VerinSenpai:
## About the PR
This command works on all client side commands. All resolves have been removed with commands being converted to LEC/LC as needed. Also removed any unused using statements and localized the commands (of course). I know Slarti suggested smaller PR's for the sake of review speed, but with as many commands as there are, I don't wanna flood the PR with a hundred micro prs. 

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

PR #38378, opened 2025-06-17
Command resolve and LEC conversion batch 3

VerinSenpai:
## About the PR
~~The mythical negative diff PR Beck was talking about. Surely this makes up for all the other positive diff cleanup PRs~~
Same song and dance. Removes resolves from commands per #13099. Converts commands to LocalizedEntityCommands or LocalizedCommands where relevant and localizes them. I moved around some locale files to put them where ~~I personally believe~~ they belong. I know I've probably removed some functional messages from a few commands in this but if you'll hear me out, I will be making a second pass of all commands to standardize messages and make them consistent.

Everything here is tested and working to the best of my ability.

See also #38367 #38338

## Requirements

### DIFF
--- a/Content.Server/GameTicking/Commands/ForceMapCommand.cs
+++ b/Content.Server/GameTicking/Commands/ForceMapCommand.cs
@@ -10,50 +10,49 @@ using Robust.Shared.Prototypes;
 namespace Content.Server.GameTicking.Commands
 {
     [AdminCommand(AdminFlags.Round)]
-    sealed class ForceMapCommand : IConsoleCommand
+    internal sealed class ForceMapCommand : LocalizedCommands
     {
         [Dependency] private readonly IConfigurationManager _configurationManager = default!;
+        [Dependency] private readonly IGameMapManager _gameMapMaanger = default!;
+        [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
 
-        public string Command => "forcemap";
-        public string Description => Loc.GetString("forcemap-command-description");
-        public string Help => Loc.GetString("forcemap-command-help");
+        public override string Command => "forcemap";
 
-        public void Execute(IConsoleShell shell, string argStr, string[] args)
+        public override void Execute(IConsoleShell shell, string argStr, string[] args)
         {
             if (args.Length != 1)
             {
-                shell.WriteLine(Loc.GetString("forcemap-command-need-one-argument"));
+                shell.WriteLine(Loc.GetString(Loc.GetString($"shell-need-exactly-one-argument")));
                 return;
             }
 
-            var gameMap = IoCManager.Resolve<IGameMapManager>();
             var name = args[0];
 
             // An empty string clears the forced map
-            if (!string.IsNullOrEmpty(name) && !gameMap.CheckMapExists(name))
+            if (!string.IsNullOrEmpty(name) && !_gameMapMaanger.CheckMapExists(name))
             {
-                shell.WriteLine(Loc.GetString("forcemap-command-map-not-found", ("map", name)));
+                shell.WriteLine(Loc.GetString("shell-invalid-map-id"));
                 return;
             }
 
             _configurationManager.SetCVar(CCVars.GameMap, name);
 
             if (string.IsNullOrEmpty(name))
-                shell.WriteLine(Loc.GetString("forcemap-command-cleared"));
+                shell.WriteLine(Loc.GetString("cmd-forcemap-cleared"));
             else
-                shell.WriteLine(Loc.GetString("forcemap-command-success", ("map", name)));
+                shell.WriteLine(Loc.GetString("cmd-forcemap-success", ("map", name)));
         }
 
-        public CompletionResult GetCompletion(IConsoleShell shell, string[] args)
+        public override CompletionResult GetCompletion(IConsoleShell shell, string[] args)
         {
             if (args.Length == 1)
             {
-                var options = IoCManager.Resolve<IPrototypeManager>()
+                var options = _prototypeManager
                     .EnumeratePrototypes<GameMapPrototype>()
                     .Select(p => new CompletionOption(p.ID, p.MapName))
                     .OrderBy(p => p.Value);
 
-                return CompletionResult.FromHintOptions(options, Loc.GetString("forcemap-command-arg-map"));
+                return CompletionResult.FromHintOptions(options, Help);
             }
 
             return CompletionResult.Empty;
--- a/Content.Server/GameTicking/Commands/ForceMapCommand.cs
+++ b/Content.Server/GameTicking/Commands/ForceMapCommand.cs
@@ -10,10 +10,10 @@ using Robust.Shared.Prototypes;
 namespace Content.Server.GameTicking.Commands
 {
     [AdminCommand(AdminFlags.Round)]
-    internal sealed class ForceMapCommand : LocalizedCommands
+    public sealed class ForceMapCommand : LocalizedCommands
     {
         [Dependency] private readonly IConfigurationManager _configurationManager = default!;
-        [Dependency] private readonly IGameMapManager _gameMapMaanger = default!;
+        [Dependency] private readonly IGameMapManager _gameMapManager = default!;
         [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
 
         public override string Command => "forcemap";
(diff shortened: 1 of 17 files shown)


Content.Server/GameTicking/Commands/ForceMapCommand.cs
```diff
@@ -10,50 +10,49 @@
 namespace Content.Server.GameTicking.Commands
 {
     [AdminCommand(AdminFlags.Round)]
-    sealed class ForceMapCommand : IConsoleCommand
+    internal sealed class ForceMapCommand : LocalizedCommands
     {
         [Dependency] private readonly IConfigurationManager _configurationManager = default!;
+        [Dependency] private readonly IGameMapManager _gameMapMaanger = default!;
```
slarticodefast:
```suggestion
        [Dependency] private readonly IGameMapManager _gameMapMananger = default!;
```

Content.Server/GameTicking/Commands/ForceMapCommand.cs
```diff
@@ -10,50 +10,49 @@
...
+                shell.WriteLine(Loc.GetString(Loc.GetString($"shell-need-exactly-one-argument")));
                 return;
             }
 
-            var gameMap = IoCManager.Resolve<IGameMapManager>();
             var name = args[0];
 
             // An empty string clears the forced map
-            if (!string.IsNullOrEmpty(name) && !gameMap.CheckMapExists(name))
+            if (!string.IsNullOrEmpty(name) && !_gameMapMaanger.CheckMapExists(name))
```
slarticodefast:
```suggestion
            if (!string.IsNullOrEmpty(name) && !_gameMapMananger.CheckMapExists(name))
```

Content.Server/GameTicking/Commands/ForceMapCommand.cs
```diff
@@ -10,50 +10,49 @@
...
             {
-                shell.WriteLine(Loc.GetString("forcemap-command-map-not-found", ("map", name)));
+                shell.WriteLine(Loc.GetString("shell-invalid-map-id"));
```
slarticodefast: A map id is something different, it's similar to an uid and just a number.
Keep the old loc string.

Content.Server/GameTicking/Commands/ForceMapCommand.cs
```diff
@@ -10,50 +10,49 @@
 namespace Content.Server.GameTicking.Commands
 {
     [AdminCommand(AdminFlags.Round)]
-    sealed class ForceMapCommand : IConsoleCommand
+    internal sealed class ForceMapCommand : LocalizedCommands
```
slarticodefast:
```suggestion
    public sealed class ForceMapCommand : LocalizedCommands
```
not sure what difference this makes, but this is used everywhere else, so let's keep it consistent

Content.Server/GameTicking/Commands/ForcePresetCommand.cs
```diff
@@ -8,51 +8,49 @@
 namespace Content.Server.GameTicking.Commands
 {
     [AdminCommand(AdminFlags.Round)]
-    sealed class ForcePresetCommand : IConsoleCommand
+    internal sealed class ForcePresetCommand : LocalizedEntityCommands
```
slarticodefast:
```suggestion
    public sealed class ForcePresetCommand : LocalizedEntityCommands
```

Content.Server/GameTicking/Commands/ForceMapCommand.cs
```diff
@@ -10,50 +10,49 @@
...
             if (args.Length == 1)
             {
-                var options = IoCManager.Resolve<IPrototypeManager>()
+                var options = _prototypeManager
                     .EnumeratePrototypes<GameMapPrototype>()
                     .Select(p => new CompletionOption(p.ID, p.MapName))
                     .OrderBy(p => p.Value);
 
-                return CompletionResult.FromHintOptions(options, Loc.GetString("forcemap-command-arg-map"));
+                return CompletionResult.FromHintOptions(options, Help);
```
slarticodefast: This is different from the `Help` text.
`Help` shows the full command usage.
This should only show the currently expected datatype for the argument.
Same for all the other commands.

Content.Server/GameTicking/Commands/GoLobbyCommand.cs
```diff
@@ -8,40 +8,36 @@
...
 
             if (preset != null)
-            {
-                ticker.SetGamePreset(preset);
-            }
+                _gameTicker.SetGamePreset(preset);
 
-            shell.WriteLine($"Enabling the lobby and restarting the round.{(preset == null ? "" : $"\nPreset set to {presetName}")}");
+            shell.WriteLine(Loc.GetString($"cmd-golobby-success") +
+                            $"{(preset == null ? "" : Loc.GetString($"cmd-forcepreset-success", ("preset", presetName)))}");
```
slarticodefast: this should use a single loc string
also the forcepreset one seems wrong here

Content.Server/Interaction/TilePryCommand.cs
```diff
@@ -33,39 +32,38 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
...
+                var coordinates = _mapSystem.GridTileToLocal(playerGrid.Value, mapGrid, tile.GridIndices);
+                var tileDef = (ContentTileDefinition)_tileDefinitionManager[tile.Tile.TypeId];
 
-                if (!tileDef.CanCrowbar) continue;
+                if (!tileDef.CanCrowbar)
+                    continue;
 
-                var plating = tileDefinitionManager["Plating"];
-                mapSystem.SetTile(playerGrid.Value, mapGrid, coordinates, new Tile(plating.TileId));
+                var plating = _tileDefinitionManager["Plating"];
```
slarticodefast: unhardcode this string
Define `private readonly ProtoId<ContentTileDefinition> _platingId = "Plating"` at the top of the class

Content.Server/Mapping/MappingCommand.cs
```diff
@@ -13,32 +13,32 @@
 namespace Content.Server.Mapping
 {
     [AdminCommand(AdminFlags.Server | AdminFlags.Mapping)]
-    sealed class MappingCommand : IConsoleCommand
+    internal sealed class MappingCommand : LocalizedEntityCommands
```
slarticodefast:
```suggestion
    public sealed class MappingCommand : LocalizedEntityCommands
```

Content.Server/Interaction/TilePryCommand.cs
VerinSenpai: you got it.

Content.Server/Mapping/MappingCommand.cs
VerinSenpai: the prefix was missing and internal is what rider suggested. I'll fix em up. Think I got a few of those in my other PR's too.

Content.Server/Interaction/TilePryCommand.cs
VerinSenpai: to be clear, am I just pasing that platingId to _tileDefinitionManager and passing that to Tile?

VerinSenpai:
I believe I've completed the requested changes. I did have one question that is posted on the relevant review.

Content.Server/Interaction/TilePryCommand.cs
slarticodefast: Nevermind, it's not that easy after all because this still uses a string internally and tile definitions are weird.
I guess we leave it as it is for now.

Content.Server/Interaction/TilePryCommand.cs
```diff
@@ -5,19 +5,21 @@
...
+public sealed class TilePryCommand : LocalizedEntityCommands
 {
-    [Dependency] private readonly IEntityManager _entities = default!;
+    [Dependency] private readonly ITileDefinitionManager _tileDefinitionManager = default!;
+    [Dependency] private readonly SharedMapSystem _mapSystem = default!;
 
-    public string Command => "tilepry";
-    public string Description => "Pries up all tiles in a radius around the user.";
-    public string Help => $"Usage: {Command} <radius>";
+    private readonly ProtoId<ContentTileDefinition> _platingId = "Plating";
```
slarticodefast:
```suggestion
    private readonly string _platingId = "Plating";
```
not sure if `ContentTileDefinition` is correct here, I mixed that up
we keep it a string for now

PR #37519, opened 2025-05-16
Let them (Vox) eat (more) trash

BigfootBravo:
## About the PR
Vox now have more trash items they can eat! Cigarettes can be eaten but don't gain their smokable reagents, due to balance concerns with double-dipping reagents. Eating cigarettes gives generic trash reagents for the Vox to consume. Empty bottles and other containers that would store reagents `drink` or `food` are excluded, due to being buggy.

## Why / Balance
Its fun to be able have more things Vox can eat, trash-wise. 

## Technical details
Addition of `SolutionContainerManager` with `Food` reagents of `ToxinTrash` to items only Vox are able to eat, namely some objects with the `Trash` tag and that now have `requiresSpecialDigestion: true.`  Only Vox are able to eat these trash items, and will only gain `ToxinTrash,` no other reagents. Eating cigarettes will only give the reagents under the `Food` container, and smoking a cigarette will not give the smoker any of the `Food` container reagents.

## Media
None needed.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
None.

**Changelog**
:cl:
- add: More trash for Vox to eat!

PR #38379, opened 2025-06-17
Vox can eat cloth and fiberous objects/foods inefficently

BigfootBravo:
## About the PR
lets Vox eat everything Moths can, but at a much lower rate, gaining ~0.50 hunger per unit of fiber, compared to the Moth's getting ~3.00. Exact value uncertain.

## Why / Balance
Vox have strong enough stomachs to digest various forms of trash, such as wrappers, which would require a stomach stronger than the norm. If #37519 were to be merged, this would lend more credence to the strength of their stomach, being able to digest things like Popsicle sticks and broken plates, things which I have heard approval of from Discord. Vox can now digest fiber, but they are not specialized to do so and so they can only do so with a low hunger to unit ratio. This also gives more uses for cotton foods, so that they don't only have to be made for only Moths. Also provides more reason to design and implement new cotton foods by providing more individuals who could consume it, not just one species.

## Technical details
Addition of `ClothMade` to the allowed Vox stomach tags. Addition of a new `!type:SatiateHunger` condition for fiber for Vox stomachs, defining it to have a `factor` of 0.50, regulating how much satiation they get from consuming it.

## Media

https://github.com/user-attachments/assets/614155e2-1fc0-43a2-8545-10cac038c421

## Requirements

## Breaking changes
None.

**Changelog**
:cl:
- add: Vox can now eat Moth Foods, at a inefficient rate.

### DIFF
--- a/Resources/Prototypes/Body/Organs/vox.yml
+++ b/Resources/Prototypes/Body/Organs/vox.yml
@@ -21,12 +21,13 @@
   description: "A stomach that smells of ammonia."
   components:
   - type: Metabolizer #Skreeeee!
-    metabolizerTypes: [Vox]
+    metabolizerTypes: [ Vox ]
   - type: Stomach
 #Bird vs tags
     specialDigestible:
       tags:
       - Trash
+      - ClothMade
     isSpecialDigestibleExclusive: false
 
 - type: entity
--- a/Resources/Prototypes/Reagents/fun.yml
+++ b/Resources/Prototypes/Reagents/fun.yml
@@ -41,6 +41,11 @@
         conditions:
         - !type:OrganType
           type: Moth
+      - !type:SatiateHunger
+        conditions:
+        - !type:OrganType
+          type: Vox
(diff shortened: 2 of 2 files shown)


BigfootBravo:
I imagine this one will need some discussion if it were ever to be considered, but that is partly why I made this PR! Also, because it made sense to me and sounded fun.

slarticodefast:
I kinda think this makes both vox and moths less unique, so I would rather keep their food preferences as they are.
But maybe another maintainer can comment as well.

BigfootBravo:
> I kinda think this makes both vox and moths less unique, so I would rather keep their food preferences as they are. But maybe another maintainer can comment as well.

Very reasonable! That is why I made vox get a lot less hunger from fiber compared to moths, but it is very fair to want to keep the diets from overlapping.

MissKay1994:
This is fun and I love it

beck-thompson:
I agree with slarti - I think its better to not blur that line! Thanks for the pr!

PR #37304, opened 2025-05-09
Adds Nukie IDs and PDAs, makes Nukie IDs able to copy accesses.

PicklOH:
## About the PR
Adds custom sprites for Nukie Operative, Corpsman, and Commander PDA and ID. (Sprites done by @archee1)
Gives Nukie ID the ability to copy accesses the way an Agent ID can.

## Why / Balance
Per comments from @ArtisticRoomba, maintainers are looking for the following changes to the Nukie loadouts:

- Give syndicate IDs the ability to copy accesses from other IDs, like the agent ID. But it shouldn't get the agent ID's job or icon setting capabilities: just the access copying function.
- Nukies should get the blood-red magboots when they start on their feet.

#37295 is handling the latter thanks to @ToastEnjoyer.

This PR is handling the former.

Agent ID is the most purchased item for Nukies/LoneOps. This item is definitely a big QOL buy for Nukies. Instead of having to juggle IDs over and over, you just upgrade as you go. Giving Nukies the ability to copy accesses without giving them any inherent stealth bonuses from the chameleon effect is the goal.

## Technical details

- Added unique sprites for Nukie Operative, Corpsman, and Commander PDA.
- Added unique sprites for Nukie Operative, Corpsman, and Commander ID using layers from base black Syndicate ID.
- Modified Syndicate PDA and ID to match the theming, also using layers for the ID.
- Gave Nukie Operative, Corpsman, and Commander ID the ability to copy accesses from other IDs the same way the Agent ID currently does it.
- Nukies now spawn with the appropriate unique PDA and ID.

## Media
Better pictures maybe coming soon.
basic

https://github.com/user-attachments/assets/22f12937-1be7-4581-ae74-90bfba010a88

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
:cl: PicklOH, Archee
- add: Nukie Operative, Corpsman, and Commander now have unique PDAs with unique IDs that can copy accesses.

PR #38381, opened 2025-06-17
Add Syndicate IDs to the uplink for 1 TC

SuperGDPWYL:
## About the PR
Added the Syndicate ID card to the uplink, in the Deception category, for 1 TC. It has Syndicate access.
## Why / Balance
Allows for multitooling door electronics to be more useful. Right now, it's not really useful unless you're a nuclear operative, syndicate refugee or wizard. The ID itself is obviously Syndicate and unless an agent ID is used to copy its access, is a liability to carry around.

This does allow for rooms where no crew-aligned player can open them, which might end up attracting suspicion from the Captain, AI or Cyborgs if they realise they can't open a door, which means they should be placed carefully. They can still be pried open by jaws of life, set to emergency access by the AI (which may be a bug but fixing it is out of this PR's scope) or hacked open if somebody really wants in.

Due to being useless on-station by itself, it only costs 1 TC.

## Technical details
Added new listing to the uplink: UplinkSyndicateIDCard.

## Media
image

## Requirements

**Changelog**

:cl:
- add: Added the Syndicate ID Card to the uplink for 1 TC.

### DIFF
--- a/Resources/Locale/en-US/store/uplink-catalog.ftl
+++ b/Resources/Locale/en-US/store/uplink-catalog.ftl
@@ -133,7 +133,7 @@ uplink-agent-id-card-name = Agent ID Card
 uplink-agent-id-card-desc = A modified ID card that can copy accesses from other cards and change its name and job title at-will.
 
 uplink-syndicate-id-card-name = Syndicate ID Card
-uplink-syndicate-id-card-desc = A Syndicate ID card, with Syndicate access. Unlikely to be useful on its own, but goes great with reconfigured doors.
+uplink-syndicate-id-card-desc = A Syndicate ID card, with Syndicate access. Unlikely to be useful on its own, but goes great with reconfigured doors. Does not come with access-copying functionality.
 
 uplink-black-jetpack-name = Black Jetpack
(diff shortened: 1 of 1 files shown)


slarticodefast:
I think this is fine on it's own, but if we merge https://github.com/space-wizards/space-station-14/pull/37304 then the syndicate ID cards would become inconsistent where the nukie version can copy access and the non-nukie version can't. So I would wait for a decision on that PR first and then maybe improve the names or descriptions to make this more clear to the player.

walksanatora:
i feel like ALL agent IDs should come with syndi access. why add a new ID just for a single access when we can just mix it with the existing ID card item. since syndicate access doesn't show up on the staton id card console they cant use it to check.

SuperGDPWYL:
That's fair. The Syndicate ID card is cheaper than the agent ID though so it could be offered as an option alongside the agent ID for when you don't have all 3 TC to spend?

SuperGDPWYL:
Given the outcome of the #37304 vote was to merge it, I will add a note to the syndie ID that it does not have access-copying abilities.

SuperGDPWYL:
The Agent ID also appears to come with Syndicate Agent access already, so no need for me to PR that. The Syndie ID is still cheaper though so it still has a reason to be bought.


Edit: It doesn't? Weird.

PR #38382, opened 2025-06-17
Command resolves and LEC conversions batch 4

VerinSenpai:
## About the PR
I feel a bit spammy at this point but we're so close I can taste it.

Removes resolves from commands. Localizes and converts to LEC or LC as needed.
A few commands were only spared this batch because they were using a depen of entitymanager to get systems instead. The silly goobers.

## Requirements

### DIFF
--- a/Content.Server/Chat/Commands/AdminChatCommand.cs
+++ b/Content.Server/Chat/Commands/AdminChatCommand.cs
@@ -6,19 +6,19 @@ using Robust.Shared.Console;
 namespace Content.Server.Chat.Commands
 {
     [AdminCommand(AdminFlags.Adminchat)]
-    internal sealed class AdminChatCommand : IConsoleCommand
+    internal sealed class AdminChatCommand : LocalizedCommands
     {
-        public string Command => "asay";
-        public string Description => "Send chat messages to the private admin chat channel.";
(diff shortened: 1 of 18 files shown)


Content.Server/Chat/Commands/AdminChatCommand.cs
```diff
@@ -6,19 +6,19 @@
 namespace Content.Server.Chat.Commands
 {
     [AdminCommand(AdminFlags.Adminchat)]
-    internal sealed class AdminChatCommand : IConsoleCommand
+    public sealed class AdminChatCommand : LocalizedCommands
```
PJB3005: Why change this to public?

VerinSenpai: I was changing em to match the other commands for consistency. I can revert it if thats not wanted

PJB3005: Yeah just keep it as-is.

VerinSenpai: You got it

VerinSenpai: all good 👌

PJB3005 (approved):

PR #38383, opened 2025-06-17
Fixed mis-labeled component filter

Citrea-Lingua:
## About the PR
Buxfix to allow handheld artifact containers to actually store handheld artifacts

## Why / Balance
This is pretty clearly a bug and unintended, given that handheld artifact containers currently cannot store *anything* due to that error in the filters.

## Technical details
Storage whitelist filtered for the Artifact component (which seems to be a material?) rather than the XenoArtifact component, which is what actually denotes an artifact.

## Media
Pre-fix:
grafik

Works now!
grafik

## Requirements

**Changelog**
:cl:
- fix: Handheld artifact containers can now properly store handheld artifacts.

Not sure if this should be a hotfix, but it does not seem overly urgent, as normal artifact containers can still also safely contain handheld artifacts.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Xenoarchaeology/artifact_equipment.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Xenoarchaeology/artifact_equipment.yml
@@ -124,9 +124,9 @@
     - 0,0,1,1
     whitelist:
       components:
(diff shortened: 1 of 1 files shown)


ArtisticRoomba (approved):
Thanks a bunch for the fix.

PR #32353, opened 2024-09-21
Faxes are colored again

AsnDen:
## About the PR
I've noticed that faxes lost their colors (blue for centcomm and black for sindye)

Hope I've done everything correctly

## Why / Balance
They had their colors but somehow lost them.

## Technical details
Changed layer for fax. Changed sprites for inserting and printing (removed fax from them).

It looked like that `color` also changes color for paper (it looked strange)

Haven't changed `FaxVisualsSystem` (cause I don't know how and what I should change, so just decreased `Length`).

I have removed unique animations when fax gets message (probably I will try to restore them soon)

## Media
Before:
image
After (currently bug in spawn menu. Will fix soon):
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- fix: Faxes have their colors again

PR #38387, opened 2025-06-17
Faxes are colored (again)

AsnDen:
## About the PR
This is revive of pr #32353. I've made all of this from scratch so I've decided to open new pr rather than reopen the old one.

So faxes are colored. Yea. They also turn off when there is no power (they don't do it now).

This version of pr is better as fax "eyes" were not removed

## Why / Balance
Faxes used to be colored

## Technical details
I've made additional layers for fax prototype:
1. Fax itself
2. Its eyes
3. Printing layer
4. Inserting layer

Also I've removed fax from sprites. Now only icon sprite has fax in it. This allows us to color only the fax machine. (When faxes were colored, they've also colored papers, hamsters, mouses. I don't like it)

## Media

image

No power - no "eyes"

https://github.com/user-attachments/assets/8f84107e-d9a2-42ce-a925-c723f0e25bc1

## Requirements

**Changelog**

:cl:
- fix: Faxes now are visually turned off when there is no power
- fix: Faxes are colored again


Spangs04:
Could you potentially make the captain's long-range fax machine different from the normal ones? I have seen a couple of thieves not be able to find it because it looks exactly like a normal one.

AsnDen:
> Could you potentially make the captain's long-range fax machine different from the normal ones? I have seen a couple of thieves not be able to find it because it looks exactly like a normal one.

Yea, we just need to add color to state icon. It's not a problem but it should be done in separate PR I guess (cause this one mostly fixes layers in faxes so they work properly)

AJCM-git (approved):

PR #38389, opened 2025-06-17
Command resolve mega pr batch 5

VerinSenpai:
## About the PR
~~Hey Youtube. Its ya boi, back with another command pr. Today we're taking these command resolves out back and givin em the uncle Teddy special.~~

We're in the home stretch for command resolves, at least on this repo.  Converts commands to LEC or LC depending on needs and removes any resolves. Localizes the commands accordingly using generics where possible. I've skipped the setoutfit command for this one. Thats gonna require some tinkering. ~~I'd like to move the actual meat of that command to a system so I can make it not static and remove its resolves, but we'll see what happens.~~

## Requirements

### DIFF
--- a/Content.Server/Administration/Commands/RemoveExtraComponents.cs
+++ b/Content.Server/Administration/Commands/RemoveExtraComponents.cs
@@ -56,7 +56,7 @@ namespace Content.Server.Administration.Commands
                 shell.WriteLine(Loc.GetString($"cmd-removeextracomponents-success-with-id",
                     ("count", components),
                     ("entities", entities),
-                    ("id", $"{id}")));
+                    ("id", id)));
                 return;
             }
 
--- a/Content.Server/Administration/Commands/RemoveExtraComponents.cs
+++ b/Content.Server/Administration/Commands/RemoveExtraComponents.cs
@@ -5,46 +5,43 @@ using Robust.Shared.Prototypes;
 namespace Content.Server.Administration.Commands
 {
     [AdminCommand(AdminFlags.Mapping)]
-    public sealed class RemoveExtraComponents : IConsoleCommand
+    public sealed class RemoveExtraComponents : LocalizedEntityCommands
     {
-        public string Command => "removeextracomponents";
-        public string Description => "Removes all components from all entities of the specified id if that component is not in its prototype.\nIf no id is specified, it matches all entities.";
-        public string Help => $"{Command} <entityId> / {Command}";
-        public void Execute(IConsoleShell shell, string argStr, string[] args)
+        [Dependency] private readonly IComponentFactory _compFactory = default!;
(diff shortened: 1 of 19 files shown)


VerinSenpai:
When I say we're in the home stretch, I mean we've got like maybe one more pr of this scale left. I haven't been counting RT commands as part of this since the issue exists here but I do plan to work on those too.

Content.Server/Administration/Commands/RemoveExtraComponents.cs
```diff
@@ -54,7 +51,18 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
                     entities++;
             }
 
-            shell.WriteLine($"Removed {components} components from {entities} entities{(id == null ? "." : $" with id {id}")}");
+            if (id != null)
```
VerinSenpai, line 54: @slarticodefast could you... suggest something for this? I don't know any magic C# way to do this better and I hate this.

PJB3005: This is fine.

Content.Server/Administration/Commands/RemoveExtraComponents.cs
```diff
@@ -54,7 +51,18 @@ public void Execute(IConsoleShell shell, string argStr, string[] args)
...
+            {
+                shell.WriteLine(Loc.GetString($"cmd-removeextracomponents-success-with-id",
+                    ("count", components),
+                    ("entities", entities),
+                    ("id", $"{id}")));
```
PJB3005:
```suggestion
                    ("id", id)));
```

Content.Server/Administration/Commands/ShuttleCommands.cs
```diff
@@ -6,46 +6,37 @@
 namespace Content.Server.Administration.Commands
 {
     [AdminCommand(AdminFlags.Round)]
-    public sealed class CallShuttleCommand : IConsoleCommand
+    public sealed class CallShuttleCommand : LocalizedEntityCommands
     {
-        [Dependency] private readonly IEntityManager _e = default!;
+        [Dependency] private readonly ILocalizationManager _locManager = default!;
```
PJB3005: You can use the `LocalizationManager` field from the parent type instead.

PJB3005 (changes requested):

Content.Server/Administration/Commands/RemoveExtraComponents.cs
VerinSenpai: Believe i tried this but rider was complaining about something. Ill give it another go

VerinSenpai: Right you are. all done.

PJB3005 (approved):

PR #38390, opened 2025-06-17
Scale down 4k parallax textures

### LINKED ISSUE #38388: Exo map parallax is way too large
`space_map3.png` and `XenoParallaxNeb.png` are both 4k textures, consuming a grand 128 MB of VRAM when loaded. They also aren't tagged as `preload: false`, so they are always loaded.

SlamBamActionman:
## About the PR

This PR scales down some parallax textures that were oversized and sets them to not preload.

## Why / Balance

Fixes #38388

Note: The textures are scaled back up to their original relative scale ingame to maintain size consistency. While the visual impact isn't very noticeable for `space_map3.png` and `XenoParallaxNeb.png` due to being soft gradients, `space_map2.png` has some strong gradient shifts (from the stars) that make the pixelation more apparent. To counteract this, the texture has the filtered property applied. It might be desirable to change the ingame scaling, but that's left up to the map maintainer. `space_map2.png` is currently used for the map Amber.

## Technical details

25% scale'd, Automatic resampling.

## Media

Before:
image
image

After:
image
image

## Requirements

## Breaking changes

**Changelog**

Not sure this needs a CL? If it does, feel free to add one.


slarticodefast:
Is it supposed to be this pixelated?
grafik

slarticodefast:
Also only 13% file size reduction when making it 1/16th of the pixels seems wrong
grafik
or is that the total for the folder?

SlamBamActionman:
> Is it supposed to be this pixelated? grafik

That is indeed the level of pixelation you get at 25% size. Refer to the note under the Why/Balance section; I prioritized keeping the ingame scale the same (to avoid tiling issues) and the map maintainer may scale it differently in yaml if they so wish.
image
image

> Filesize

image
image

slarticodefast:
Ah, github is lying to me then with the percentage.
Also a shame that they are so pixelated now, but it's reasonable that we can't do 4k textures.

Idea guys moment and probably too much effort to be worth it, but maybe we can use something like this in the future:
https://simonschreibt.de/gat/homeworld-2-backgrounds/
It has amazing results for space backgrounds, where large parts of it are low-detail.

PJB3005:
You can set them to be filtered if you don't want them to be pixelated.

SlamBamActionman:
image

Seems like filtering works, at least until someone gets a chance to look closer at the texture

PJB3005 (approved):

NoElkaTheGod:
Ah, yes, "update size: -14Kb"

PR #38391, opened 2025-06-17
Relocate SetOutfit method to ServerClothingSystem and LEC setoutfit command.

VerinSenpai:
## About the PR
The setoutfit command needed a bit more work to properly remove its resolves. Namely, the method setoutfit was static and was being used by outside systems so I was unable to deal with resolves within without relocating the command to a system that could then be used a dependency. I decided it made sense to move it to ServerClothingSystem but I may be wrong.

Not much else had to be done to the command for this to work.

## Requirements


Content.Server/Administration/Commands/SetOutfitCommand.cs
```diff
@@ -1,36 +1,23 @@
...
-using Robust.Shared.Player;
-using Robust.Shared.Prototypes;
 
 namespace Content.Server.Administration.Commands
 {
     [AdminCommand(AdminFlags.Admin)]
-    public sealed class SetOutfitCommand : IConsoleCommand
+    public sealed class SetOutfitCommand : LocalizedEntityCommands
     {
         [Dependency] private readonly IEntityManager _entities = default!;
```
PJB3005: Can remove this dependency.

Content.Server/Clothing/ClothingSystem.cs
PJB3005, line 1: Make a new system for this instead.

PJB3005, line 1: Make a new system for this instead.

PJB3005:
I love when GitHub completely fucks up the review system and accidentally posts them immediately.

Content.Server/Clothing/ClothingSystem.cs
VerinSenpai: I'll get em done

Content.Server/Clothing/Systems/OutfitSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+
+public sealed class OutfitSystem : EntitySystem
+{
+    [Dependency] private readonly IServerPreferencesManager _preferenceManager = default!;
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+    [Dependency] private readonly HandsSystem _handSystem = default!;
+    [Dependency] private readonly InventorySystem _invSystem = default!;
+    [Dependency] private readonly SharedStationSpawningSystem _spawningSystem = default!;
+
+    public bool SetOutfit(EntityUid target, string gear, IEntityManager entityManager, Action<EntityUid, EntityUid>? onEquipped = null)
```
PJB3005: You don't need to take the entity manager as an input here, you always have it available in an entity system.

PJB3005 (changes requested):

Content.Server/Clothing/Systems/OutfitSystem.cs
VerinSenpai: Right you are again.

VerinSenpai: all good I reckon 👌

PJB3005 (approved):

PR #36564, opened 2025-04-14
Changed the storage sizes of different swords.

RedBookcase:
## About the PR

Inspired by #36473 

Swords have had their tile storage sizes tweaked. The Captain's Sabre, Katana, and Energy Katana have become 1x4, the Claymore has become 2x4, and the Machete and Cutlass (and Foam Cutlass!) have become 1x3.

Additionally, Katanas can now be stored on the Belt and Back slots, and the Claymore can now be stored on the SuitStorage slot.

## Why / Balance

Previously all swords (with the exception of the Throngler) shared the same 2x2 storage size. 

This didn't make much sense, as a Machete is smaller than a Katana or Sabre, and a Katana or Sabre are both smaller than a Claymore, and it was just weird to have these long narrow blades as dinky little 2x2's.

With custom sizes, these weapons should feel a bit more unique and will interact with player storage in different ways. Previous unique storage options (The Energy Katana fitting on the belt and back, the Sabre fitting into it's sheath) still work as before.

## Technical details

Changes sword.yml - reorganized the file and added comments to make it easier to read and more organized. New shape and storedsprite lines for everything except the Throngler (and a cheeky description update for the throngler - not like you can craft it...). New sprites for katana belt/back storage. Duplicate of Claymore back storage for suit storage.

New 64x RSI files for the captain's sabre, claymore, energy katana, and katana. These allow for the big, beautiful sprites you see in the media section, and prevent weird and scrunched up sprites for big items. 

(Special thanks to Emo for pushing 3(!) PR's which made these new sprites possible.)

## Media

Before PR, when all swords had the same size:
image

With new and updated sizes:
image

Additionally, new katana belt and back storage sprites, based off the energy katana:
image

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- add: Various changes to edged weapons, including new storage sprites and new storage options.

PR #36473, opened 2025-04-11
Weapon Resizing

Nox38:
## About the PR
Added custom dimensions to every firearm, for better realism. Slightly nerfs the satchel too, by making it less able to carry a full-length assault rifle. This is a redo of my older PR, since I couldn't figure out how else to fully start from scratch.
I also moved around the sprites of the pistols visually to fit inside their grid.
## Why / Balance
This started as simply "why isn't the Enforcer 1x5?" and was quickly followed by other strange questions like "why is the assault rifle a square?". This has led me to change every single gun to have a somewhat more realistic and balanced inventory size.

## Technical details
Custom dimensions were added to the item component of every firearm.

## Media
image
I will make a separate PR to make the Mosin and Musket 1x5 once I have new sprites which fit them.

These ones are ginormous now, and don't fit in bags.
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
:cl: Nox38
- tweak: Resized every firearm.

PR #38392, opened 2025-06-17
Resized baseball bats to be more realistic

Kittygyat:
## About the PR
I made baseball bats longer in the grid inventory, 1x4 instead of 2x2.

## Why / Balance
This was changed due to being slightly more realistic; a 2x2 baseball bat never made much sense, and looked weird in the inventory.

This is highly unlikely to cause balance issues, due to the overall size remaining the same & still being able to fit in all bags.

## Technical details
I added "shape 0,0,0,3" to the baseball bat, aswell as it's incomplete variant.
I replaced the icon sprites for the baseball bat & incomplete variant with new ones, new sprite credit goes to TiniestShark

## Media
baseballbatprfinal

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Made baseball bats 1x4 instead of 2x2
- add: New baseball bat inventory icon sprites


DogZeroX:
This could use a custom sprite/rotation, some items do this.

Kittygyat:
> This could use a custom sprite/rotation, some items do this.

I would love to hear how to rotate it; I didn't know that was possible!

slarticodefast:
Yeah, you can have a custom sprite for the storage icon.
Look at this PR for reference https://github.com/space-wizards/space-station-14/pull/36564
If you need help with making a sprite for this then you can maybe ask in the #sprite-dungeon channel on our discord server.

keronshb:
Adding concern since we're not  really accepting minor tweaks like this at the moment.

https://github.com/space-wizards/space-station-14/issues/8524

lzk228:
i think just rotation will do
an example
https://github.com/space-wizards/space-station-14/blob/4638833ae3defeca6a15d2f9d96a32ae2e3ea32b/Resources/Prototypes/Entities/Clothing/Eyes/base_clothingeyes.yml#L10-L12

you just need to write 45 or -45

TheShuEd:
Also please update media after rotation

PicklOH:
I don't see this PR as any different from what was accepted in #36564. Seems like a simple enough change.

Also fits the global weapon resizing we're seeing with #36473 as well.

ArtisticRoomba:
> Adding concern since we're not really accepting minor tweaks like this at the moment.

The changes in this PR are perfectly fine. There are no other wide groups this can be extended to and unnecessary overextension into unrelated groups will just spark an unnecessary discussion.

The freeze's intent was to primarily curb extreme low-justification microbalancing like:
- Increasing/decreasing a single gun's fire rate/spread
- Marginal increases/decreases on suit stats

These PRs generally accomplish very little, are usually very hard to observe the effects of the changes, and take up maintainer time discussing the changes and going through the whole vote procedure. Changes should generally reflect a larger group and work towards a *goal* or intent.

This PR works towards a larger previously discussed goal of making inventory storages more interesting, as most of them are simply inherited from the tiny/small/medium/large default sizes (1x1, 2x2, 3x3, 4x4). Since the changes can't really be expanded into any other meaningful group, I see them being perfectly fine.

TiniestShark:
baseball bat 1

bat unfinished

Here's two upright sprites you can use for the storage.

Kittygyat:
> Here's two upright sprites you can use for the storage.

thanks!, I'll add them when I can get around to it.

Kittygyat:
Apologies for the delay, I took a break from PRs & SS14 as a whole for a bit, hopefully this works alright!

slarticodefast (changes requested):
Keep the icon sprite as it is, only use the vertical sprite for the `StoredSprite` in the `ItemComponent`

lzk228:
i think there's no reason for adding bew sprites
just use the stored rotation datafield
see pickaxe for an example

slarticodefast:
> i think there's no reason for adding bew sprites
just use the stored rotation datafield
see pickaxe for an example

Like I said, the normal icon can stay the same. That way it will still be diagonal in the hand slot and won't overlap.
grafik

But the inventory sprite should be the new one because simply rotating it will result in diagonal pixels.

slarticodefast (approved):
Looks good to me, thank you for your contribution!

slarticodefast:
I asked Keron on discord and they said it's good to merge with the new storage sprite.

PR #38394, opened 2025-06-17
Add directionality to cable terminal port nodes

### LINKED ISSUE #36339: SMES connect to cable terminals that aren't facing them

whatston3:
## About the PR

Adds directionality to the cable terminal port node ReachableNodes check.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/36339

Roomba looking down its whirring nose at me.  The _gall_.  The _cheek_.

## Technical details

Replicates the terminal check in cable node in the CableTerminalPortNode.

## Media

Simple connection case.
image

Case from the bug.  Note that each terminal port node only connects to one terminal.
image

## Requirements

## Breaking changes

**Changelog**

no

### DIFF
--- a/Content.Server/Power/Nodes/CableTerminalPortNode.cs
+++ b/Content.Server/Power/Nodes/CableTerminalPortNode.cs
@@ -21,9 +21,11 @@ namespace Content.Server.Power.Nodes
             var gridIndex = grid.TileIndicesFor(xform.Coordinates);
 
             var nodes = NodeHelpers.GetCardinalNeighborNodes(nodeQuery, grid, gridIndex, includeSameTile: false);
-            foreach (var (_, node) in nodes)
+            foreach (var (dir, node) in nodes)
             {
-                if (node is CableTerminalNode)
+                if (node is CableTerminalNode
+                    && dir != Direction.Invalid
+                    && xformQuery.GetComponent(node.Owner).LocalRotation.GetCardinalDir().GetOpposite() == dir)
                     yield return node;
             }
         }


Content.Server/Power/Nodes/CableTerminalPortNode.cs
```diff
@@ -21,9 +21,11 @@ public override IEnumerable<Node> GetReachableNodes(TransformComponent xform,
             var gridIndex = grid.TileIndicesFor(xform.Coordinates);
 
             var nodes = NodeHelpers.GetCardinalNeighborNodes(nodeQuery, grid, gridIndex, includeSameTile: false);
-            foreach (var (_, node) in nodes)
+            foreach (var (dir, node) in nodes)
             {
-                if (node is CableTerminalNode)
+                if (node is CableTerminalNode
+                    && dir != Direction.Invalid
+                    && xformQuery.GetComponent(node.Owner).LocalRotation.GetCardinalDir().GetOpposite() == dir)
```
PJB3005, line 28:
```suggestion
                    && xform.LocalRotation.GetCardinalDir().GetOpposite() == dir)
```

PJB3005 (changes requested):

Content.Server/Power/Nodes/CableTerminalPortNode.cs
whatston3: Correct me if I'm wrong - that's the transform for the port.  It needs to be the transform for the terminal itself, which should be at node.Owner.  [This is correct as it exists and should not be changed.]

whatston3: e.g. the orientation for a SMES doesn't matter when it's determining what [terminals] it connects to.  The port itself is largely directionless.

whatston3: image

In this case, xform.LocalRotation would be 0, CardinalDir would be South, and opposite would be North.  dir is West.  This wouldn't connect.

PJB3005: I'm stupid, you're right.

PJB3005 (approved):

Content.Server/Power/Nodes/CableTerminalPortNode.cs
whatston3: > I'm stupid, you're right.

You're good, and you've got more on your plate than I can fathom.

PR #33124, opened 2024-11-03
Throw and embed everything sharp (shards, knives, swords)

K-Dynamic:
## About the PR

Adds embedding to a bunch of sharp objects where it would make sense, including:
- Glass shards, shivs and broken bottles
- Knives
- Scalpels
- Flatcaps
- Swords and katanas
- Hatchets and fireaxes

## Why / Balance

Added embedding property to items missing them (i.e. throwing dmg with no embed) for parity to match with spears and knives, meaning glass shards now embed on throwing. This sort of nerfs any glass cannons but also makes explosions with glass shards more interesting.

Added throwing damage to a lot of knives and swords because it seems to make sense, as well as adding some more depth to improvised combat. For instance, the ninja could throw their katana at an opponent then recall it. It also didn't make sense why only survival/combat knives could do throwing damage but not something like normal kitchen knives.

## Technical details

- `Resources/Prototypes/Entities/Objects/Materials/shards.yml` (glass shards now embed)
- `Resources/Prototypes/Entities/Objects/Misc/broken_bottle.yml` (broken bottle now embeds)
- `Resources/Prototypes/Entities/Objects/Specific/Medical/surgery.yml` (scalpels do throwing damage and embed)
- `Resources/Prototypes/Entities/Objects/Weapons/Melee/knife.yml` (added throwing damage and embed to a bunch of shivs and kitchen knives)
- `Resources/Prototypes/Entities/Objects/Weapons/Melee/mining.yml` (added throwing damage and embed to crusher dagger)
- `Resources/Prototypes/Entities/Objects/Weapons/Melee/sword.yml` (added throwing damage and embed to all swords)

## Media

image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

N/A

**Changelog**

:cl:
- add: items under the category of swords (including katanas, sabres and machetes), knives (including glass shivs, kitchen knives, kukris) now deal throwing damage and embed in mobs.
- add: axes, hatchets, scythes, crusher daggers, and scalpels now deal throwing damage and embed in mobs.
- tweak: glass shards and broken bottles now embed in mobs.

PR #38396, opened 2025-06-17
Kitchen knives can now hit and embed into targets upon being thrown!

Kittygyat:
## About the PR
I made kitchen knifes embed into targets upon being thrown, dealing their usual damage.
The chef has finally learnt how to throw them properly!

## Why / Balance
It's always felt weird that kitchen knives just slide around harmlessly when thrown, instead of functioning as you'd expect and stylishly embedding in whatever they're thrown at.

Minimal balance concerns, due to it only being a 10 slash weapon, this is mostly for RP effect and dealing with mice, very rarely will it actually be helpful in a fight.

## Technical details
Brought EmbedProjectile and the related code over from combat knives to the kitchen knife, aswell as ThrowingAngle 225.

## Media
https://github.com/user-attachments/assets/670d56a2-ab81-4cde-8f63-b0e36801c24c

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Let kitchen knives embed into targets on throw!, the chef has learnt some skills!

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/knife.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/knife.yml
@@ -41,6 +41,16 @@
   - type: Sprite
     sprite: Objects/Weapons/Melee/kitchen_knife.rsi
     state: icon
+  - type: EmbeddableProjectile
+    sound: /Audio/Weapons/star_hit.ogg
+    offset: -0.15,0.0
+  - type: LandAtCursor
+  - type: DamageOtherOnHit
+    damage:
+        types:
+          Slash: 10
+  - type: ThrowingAngle
+    angle: 225
   - type: Item
     sprite: Objects/Weapons/Melee/kitchen_knife.rsi
   - type: GuideHelp


kosticia:
Please, look at the #37498

Kittygyat:
> Please, look at the https://github.com/space-wizards/space-station-14/pull/37498

Huh.
I thought there was a sizeable delay between throwing actions, atleast now?
Unless I'm doing something wrong, there's definitely *something* that's stopping me from spam-throwing items to the point that a 10 slash kitchen knife is an issue.

Cojoke-dot:
> I thought there was a sizeable delay between throwing actions, atleast now?
> 
> Unless I'm doing something wrong, there's definitely *something* that's stopping me from spam-throwing items to the point that a 10 slash kitchen knife is an issue.

There is a delay, I believe it was there a month ago too with the last pr but I'm not 100% sure.

iaada:
It was quite a while ago, in fact #23311

slarticodefast:
For the embedding there was also a previous PR that was declined https://github.com/space-wizards/space-station-14/pull/33124
Personally I think this should only be a thing for throwing knifes or similar throwing weapons, since embedding is what makes them special as an item.
Also the kitchen knifes are already a common item used by raiders since it is so easily obtainable, so not sure if it should get a buff.

Kittygyat:
> For the embedding there was also a previous PR that was declined https://github.com/space-wizards/space-station-14/pull/33124
Personally I think this should only be a thing for throwing knifes or similar throwing weapons, since embedding is what makes them special as an item.
Also the kitchen knifes are already a common item used by raiders since it is so easily obtainable, so not sure if it should get a buff.

What about embedding for kitchen knives is bad?, it actively restricts the weapon from being spammed into a target multiple times, and the target can just..walk off with their weapon.

I don't think throwing away your okay weapon for 1 hit is truly a buff; it still requires 10 hits, and there's only 3 kitchen knives per PlasteelChef, plus, the chef gets the butcher knife for actual combat anyways, which deals roughly _double_ the damage of a kitchen knife.

PubliclyExecutedPig:
thats some pretty good hotkeys dude

SlamBamActionman (changes requested):
I think that not allowing kitchen knives to have throwing/embedding helps characterize them as a primarily non-combat item as opposed to the salvage or combat knife. Plus we're talking a big wood handle kind of knife which wouldn't be too easy to throw to begin with. I could much more see this sort of behavior work with the cleaver. 

I can accept it as an RP thing for knives if it did less damage, and it probably should not embed. Make it a 6 slash non-embeddable throwable item; this should better represent just chucking a non-specialized knife at someone. As a comparison, glass shards do the same but with 2 damage.

For the butcher's cleaver, since it's a bad-ass low-availability weapon that primarily gets its damage from its high attack rate, I could see it work as an embeddable throwable item but with slightly reduced damage, e.g. 10 slash. Keep the spinning when throwing though. 😉

SlamBamActionman:
Since there has been more than 6 weeks of inactivity awaiting changes, this PR has been marked as stale and will be closed.

My comment above still stands if you or anyone else is interested in opening a new PR along those lines.

PR #38403, opened 2025-06-17
Convert `AtmosphereSystem` sound path strings to `SoundSpecifier`s

Tayrtahn:
## About the PR

Converts 2 sound file path strings in `AtmosphereSystem` partials into `SoundSpecifier`s.

## Why / Balance

> String literals for sounds are deprecated, use a SoundSpecifier or ResolvedSoundSpecifier as appropriate instead

Fixes 2 warnings for https://github.com/space-wizards/space-station-14/issues/33279.

## Technical details

- Replaced the `string?` properties with `SoundSpecifier?`s.
- Created `SoundCollection` prototypes for space wind and hotspots.
- Added `static readonly ProtoId<SoundCollectionPrototype` fields to validate the default sound collection IDs.
- Modified the `PlayPvs` calls so they don't bulldoze all the `AudioParams` of the specifiers.
- For space wind, moved the `WithVariation` call into the `SoundCollectionSpecifier` constructor call so it doesn't bulldoze any changes to the specifier. The other `WithWhatever` calls use dynamic values so they had to stay.

## Requirements

### DIFF
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
@@ -7,6 +7,7 @@ using Robust.Shared.Audio;
 using Robust.Shared.Map;
 using Robust.Shared.Physics;
 using Robust.Shared.Physics.Components;
+using Robust.Shared.Prototypes;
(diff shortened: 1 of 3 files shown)


PJB3005 (changes requested):
These need to be made sound collections

PR #38420, opened 2025-06-18
Split Quick- and Area-Pickup into their own Components

Centronias:
## About the PR

Split Quick- and Area-Pickup into their own Components so that they can be reused for ballistic ammo providers.
Specifically also adds quick and area pickup to the foam dart ammo box to make cleaning up foam darts less carpal-tunnel-inducing.

## Why / Balance

I wanted to reuse the functionality on the foam dart ammo box which doesn't use `StorageComponent`, plus there was a TODO to split Area pickup from Storage component.

### Area Pickup
- add `AreaPickupComponent`
- add `AreaPickupSystem`
- add `BeforeAreaPickupEvent`
  - raised by `AreaPickupSystem` to determine what to try to pick up
  - handled by storage-like systems by invoking a function on `AreaPickupSystem` with a predicate
- change `AreaPickupDoAfterEvent` to be handled by storage-like systems
  - invoke a function on `AreaPickupSystem` with a function value that actually does the picking up
- move area pickup stuff from storage component & system
- can now area pickup into `BallisticAmmoProviderComponent`
- small cleanup to `AreaPickupDoAfterEvent`

### Quick Pickup
- add `QuickPickupComponent`
- add `QuickPickupSystem`
- add `QuickPickupEvent` which allows storage-like systems to implement quick pickup behavior
  - handle by calling a function on `QuickPickupSystem` with a function value which actually does the picking up
- move area pickup stuff from storage component and system
- can now quick pickup into `BallisticAmmoProviderComponent`

### Ballistic Ammo changes
- add `SilentInsertUserTag` to match functionality from `StorageComponent`
  - only applies to insertion sounds
- add public methods to `GunSystem` for use by new interactions
  - `IsFullBallistic`
  - `CanInsertBallistic`
  - `TryBallisticInsert`

### Prototype Changes
- migrate preexisting storage quick/area insert to new components
- "box of foam darts" (`BoxDonkSoftBox`) can now area and quick pickup darts

## Media

https://github.com/user-attachments/assets/9ef88137-ae68-45c9-a7a2-6e26d00f6ce6

## Requirements

## Breaking changes

Quick and Area Insert behavior is no longer tied to `StorageComponent` and instead are granted by `QuickPickupComponent` and `AreaPickupComponent` respectively. To migrate existing usages to these new components, make the following changes:
(Quick insert)
```yaml
# Before
components:
- type: Storage
  quickInsert: true

# After
components:
- type: Storage
- type: QuickPickup
```

(Area insert)
```yaml
# Before
components:
- type: Storage
  areaInsert: true
  areaInsertRadius: 10 # optional

# After
components:
- type: Storage
- type: AreaPickup
  radius: 10 # optional, same default as Storage's previous default
```

There are no modifications to existing C# public APIs aside from the removal of fields on `StorageComponent`.

**Changelog**

:cl:
- add: The Box of Foam Darts can now quickly pick up foam darts in the same way that trash bags pick up items.


Centronias:
I realized as I was writing up the descripto for this PR that it's maybe dumb for `QuickPickup` and `AreaPickup` to know about the "storage-like" `Storage` and `BallisticAmmoProvider` stuff, so another solution could be to make the coupling between them be via events. See https://github.com/Centronias/space-station-14/pull/2 for an example of what that would look like if it's preferable (Noting I did it only for Quick Pickup since Area Pickup would take probably like 4x more effort, so I didn't wanna do it unless it's needed)

VerinSenpai:
For context. Double triage, dont shoot me 🙏

Resources/Prototypes/Entities/Objects/Misc/folders.yml
```diff
@@ -379,7 +379,9 @@
   - type: Storage
     grid:
     - 0,0,4,3
-    quickInsert: true
+  - type: QuickPickup
+  - type: StorageFill
+    contents: [] #to override base clipboard fill
```
Centronias, line 384:
```suggestion
  - type: QuickPickup
```
oopers merge defect

PR #38421, opened 2025-06-18
Chloral Hydrate metabolism rate change

Skybailey-dev:
## About the PR

Doubled the duration of chloral hydrate by reducing its metabolism rate, making sure to modify its OD poison damage to compensate (Will deal the same amount of OD per unit as before, just slower.)

## Why / Balance

Chloral is really bad at what it's supposed to do. Making it last longer makes it more viable to put someone to sleep, for antagonist or non antagonist purposes (I.e. injecting to a critical criminal patient so they can't run as easily after treated, CMO/medborg dealing with a rowdy or violent crew member using a hypo, or antagonist spiking someones drink)
## Technical details

Changed 2 lines in toxins.yml
Will do the same amount of poison damage per U administered over the OD mark
## Media

N/A
## Requirements

## Breaking changes

None
**Changelog**

:cl:
- tweak: Chloral Hydrate now lasts longer


Princess-Cheeseballs:
Two hits with a hypopen adds 80 seconds of drowsiness, this doubles that. It can also be made worse by the fact foam can make your drowsiness stack by even more. I don't think this needs to be made even longer.

slarticodefast:
I agree with Princess.
I prefer the other open PR that makes it have a forced sleep effect at a certain threshold (higher than nocturine)

PR #38424, opened 2025-06-18
Adjusted minimumPlayers for Wizard midround events.

MidZik:
## About the PR

~~Minimum number of players required for a ninja has been reduced from 30 to 20, to match Loneop and Dragon, as they feel to be roughly in the same ballpark as a midround event.~~

Wizard, as the most disruptive antag of the bunch, is changing from 10 to 30.

## Why / Balance

Between the Loneop, Dragon, Ninja, and Wizard, the Wizard seems to be the most disruptive overall, and a Ninja the least. Yet the Wizard has the loosest spawn requirements, and the Ninja the most stringent.

As such, wizard now requires more players in the round, ~~and ninja now matches loneop and dragon.~~

## Requirements

**Changelog**

:cl:
- tweak: Midround wizard player count requirement increased from 10 to 30.

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -193,7 +193,7 @@
     duration: null
     earliestStart: 30
     reoccurrenceDelay: 20
-    minimumPlayers: 20
+    minimumPlayers: 30
   - type: SpaceSpawnRule
   - type: AntagLoadProfileRule
   - type: AntagObjectives
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -291,7 +291,7 @@
     duration: 1
     earliestStart: 30
     reoccurrenceDelay: 60
-    minimumPlayers: 20
+    minimumPlayers: 30
   - type: AntagSelection
     agentName: wizard-round-end-name
     definitions:
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -193,7 +193,7 @@
     duration: null
     earliestStart: 30
     reoccurrenceDelay: 20
-    minimumPlayers: 30
+    minimumPlayers: 20
   - type: SpaceSpawnRule
   - type: AntagLoadProfileRule
   - type: AntagObjectives
@@ -291,7 +291,7 @@
     duration: 1
     earliestStart: 30
     reoccurrenceDelay: 60
-    minimumPlayers: 10
+    minimumPlayers: 20
   - type: AntagSelection
     agentName: wizard-round-end-name
     definitions:


ScarKy0:
I think the better option is to make them all 30 and wizard maybe 40 even. They are all extremely disruptive on such low pops and wizard basically stands uncontested.

MidZik:
I play on Sala during lowpop hours, and in my experience the dragon and loneop are not *usually* an unrecoverable problem unless they arrive together with the others (especially in survival rounds). The worst a ninja who isn't murderboning can do is call in a dragon. But a wizard will absolutely end the round if they want and are competent.

30 for all of them is too high, although for wizard I think it can be justified to be higher. I think many rounds need the added excitement, and they serve as excellent distractions for antags.

ArtisticRoomba:
I agree with what Scar said.

ScarKy0:
> I play on Sala during lowpop hours, and in my experience the dragon and loneop are not usually an unrecoverable problem unless they arrive together

This depends entirely on how well the lobby is doing. In 99% of cases evac is called after a dragon, and even if a ninja themselves isnt that bad, they CAN summon a dragon as well.
The most fair arrangement would be 30 for midround wizard, ninja, dragon and loneop and 40 for roundstart wizard.
It's basically just in the range the crew can fight back or is prepared to deal with a threat.

keronshb:
> I think the better option is to make them all 30 and wizard maybe 40 even. They are all extremely disruptive on such low pops and wizard basically stands uncontested.


There's no way Wizard should be 40 pop. 20 is definitely fine.

keronshb (approved):
Approval for 20 pop

ScarKy0:
> There's no way Wizard should be 40 pop. 20 is definitely fine.

Aboslutely not. From adminning rounds wizard ruins anything that isnt at least 30 pop, 40 is a safer number

keronshb:
> > There's no way Wizard should be 40 pop. 20 is definitely fine.
> 
> Aboslutely not. From adminning rounds wizard ruins anything that isnt at least 30 pop, 40 is a safer number

There aint no way it's gonna be 40 pop. 40 is way too high and is not my design.

ScarKy0:
> There aint no way it's gonna be 40 pop. 40 is way too high and is not my design.

Can be 30 then, however otherwise its too low.

keronshb:
> > There aint no way it's gonna be 40 pop. 40 is way too high and is not my design.
> 
> Can be 30 then, however otherwise its too low.

20 is how it was on Citadel on SS13, it worked fine especially because of the rarity.

ilovehans10:
> There aint no way it's gonna be 40 pop. 40 is way too high and is not my design.

Speaking of your design, how is that design document coming? It would be a lot easier to refer to a design doc than waiting for you to come in and make up values for what the intended design is.

I do agree with ScarKy0 that having higher player count numbers for the wizard would be better seeing as they can create a lot of destruction very easily.

keronshb:
> Speaking of your design, how is that design document coming? It would be a lot easier to refer to a design doc than waiting for you to come in and make up values for what the intended design is.

Planned sometime after the next batch of action refactors is done + new targeting system + remaining SS13 spells.

MidZik:
Honestly my main intention was to make ninja more common, since I noticed that ninja was appearing way less often than wizard, and I was longing for ninja rolls. And as I said, I have never experienced a situation where I felt a dragon or loneop shouldn't have appeared due to the player count. That has only ever happened to me with wizard, and now I know why. 10 is very low.

> Aboslutely not. From adminning rounds wizard ruins anything that isnt at least 30 pop, 40 is a safer number

I don't think making events less common because of adminning issues is how it should be balanced. If it's a mechanics issue, change the mechanics. If it's a staffing issue, get more admins onboard.

It seems at a glance that the majority would prefer 30 to be the minimum for all the roles. I'm not interested in contributing such a change, so if that is the consensus, feel free to close this and create a new PR.

Cojoke-dot:
> I don't think making events less common because of adminning issues is how it should be balanced. If it's a mechanics issue, change the mechanics. If it's a staffing issue, get more admins onboard.

I believe they are trying to say that from the rounds they were admining, wizards destroy the station a ton with 30 pop, not something actually involving administering.

ArtisticRoomba:
Wizard usually dumpsters on anything 30 pop and below to the point where the round isn't really fun and no interesting persistent chaos comes out of it. Security easily gets overwhelmed and medical is usually comprised of two people that die equally fast. There is no burst of fun and interesting gameplay with crew interactions; everyone just dies pretty quickly with little anyone else left alive can do.

Given that the Wizard also has no end condition, unnecessarily long rounds are possible where nobody can recall and you're just relying on the wizard being a Good Samaritan and recognizing that ghosts want the round to end. There's no mechanical prevention of round stalling.

This is exacerbated by wizard's inclusion in sub-gamemodes, which contributes to its high "seen rate". In survival, this also commonly introduces the Wizard well into the hour or hour thirty mark where many people have cryod (due to a separate problem with lowpop's chaos ramping) which leaves little people to fight the Wizard, which leads to all of which I've previously described.

I much agree with ScugKy and it should just be 40 pop. Wizard's current power and available mechanics don't contribute to something fun when populations are below that.

MidZik:
Seems most of the hangup is with the wizard requirement, so I raised it to 30. I'd prefer being more conservative with reducing the amount of chaos in the game.

Resources/Prototypes/GameRules/events.yml
```diff
@@ -291,7 +291,7 @@
     duration: 1
     earliestStart: 30
     reoccurrenceDelay: 60
-    minimumPlayers: 10
+    minimumPlayers: 30
```
keronshb, line 294: 20 is fine.

keronshb (changes requested):

keronshb:
30 is such an exorbitant amount for the Wizard. I really do think 20 is more than fine.

We should be looking at the other antags, including midrounds, to see what runs 20 and below if the Wizard is somehow triggering more often on lower pops. Then figure out pop from there.

The wizard itself is quite rare, and having it run at 20 and above is more than fair.

Resources/Prototypes/GameRules/events.yml
ScarKy0: We should put this up to a maintainer vote at this point

ScarKy0:
> The wizard itself is quite rare, and having it run at 20 and above is more than fair.

Yeah and when it does trigger its miserable instead of a fun experience. We should put this up to a vote at this point.

shamblestf:
Yeah wizard should definitely be 40 pop

MidZik:
Will there be a vote for this, then?

slarticodefast:
I opened a maint vote
https://forum.spacestation14.com/t/adjusted-minimumplayers-for-ninja-and-wizard-midround-events/23538

Resources/Prototypes/GameRules/events.yml
```diff
@@ -193,7 +193,7 @@
     duration: null
     earliestStart: 30
     reoccurrenceDelay: 20
-    minimumPlayers: 30
+    minimumPlayers: 20
```
slarticodefast:
```suggestion
    minimumPlayers: 30
```
The maintainer vote decided 30 is a good number for both.
Could you change it to that value?
Normally I would just push the change myself, but you disabled the checkmark allowing that 😄

Resources/Prototypes/GameRules/events.yml
MidZik: Vote ended up on 30 so it's staying at 30.

MidZik:
Done.

PR #38426, opened 2025-06-18
Revert arrow smite to be infinite

Stonylogray:
Reverts https://github.com/space-wizards/space-station-14/pull/37102. If other admins prefer its intentional function, then we can keep it as-is. Even stopping at five arrows it's just not a practical smite for our purposes, so I figure it's all the same keeping it as it was.

No changelog.


Stonylogray:
I FUCKING PUT THIS INTO WIZDEN KMS

TeenSarlacc:
woe, 9,180 commits upon ye

Issue #31955, opened 2024-09-08
Biogenerator does not take biomass in


murolem:
## Description
Cannot put it in - nothing is happening when I do it.

Biogenerator PR https://github.com/space-wizards/space-station-14/pull/30694

**Reproduction**
Try to put biomass cubes into the Biogenerator.

**Screenshots**
https://github.com/user-attachments/assets/daff0597-4d7a-4ece-a289-96b8c12fb9c7

deltanedas:
it takes in plants

murolem:
I'm putting galaxythistle in and nothing is getting added.

https://github.com/user-attachments/assets/a4059abe-d62c-4873-b144-759fe19597b7

deltanedas:
> the biogenerator converts **nutriment** in plants into biomass, which it then uses to create items.

thistle has no nutriment

Pok27:
That has been fixed.

PR #38427, opened 2025-06-18
make biogenerator not accept low-nutrient plants

kosticia:
## About the PR
Now plant can't be deleted by biogenerator if it has extremely low nutrient amount.

## Why / Balance
related to #31955. Issue itself feels not real issue, but this shows how players can be confused with current biogen system.

## Technical details
system not predicted, so no clientpopup(((

## Media
image here popup from the ftl file over biogenerator. I forgot to take a screenshot.

## Requirements

### DIFF
--- a/Content.Server/Materials/ProduceMaterialExtractorSystem.cs
+++ b/Content.Server/Materials/ProduceMaterialExtractorSystem.cs
@@ -4,6 +4,7 @@ using Content.Server.Materials.Components;
 using Content.Server.Power.EntitySystems;
 using Content.Shared.Chemistry.EntitySystems;
 using Content.Shared.Interaction;
+using Content.Shared.Popups;
 using Robust.Server.Audio;
 
 namespace Content.Server.Materials;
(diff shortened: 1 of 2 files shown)


Content.Server/Materials/ProduceMaterialExtractorSystem.cs
```diff
@@ -39,6 +41,13 @@ private void OnInteractUsing(Entity<ProduceMaterialExtractorComponent> ent, ref
         var matAmount = solution.Value.Comp.Solution.Contents
             .Where(r => ent.Comp.ExtractionReagents.Contains(r.Reagent.Prototype))
             .Sum(r => r.Quantity.Float());
+
+        if (matAmount <= 0.1)
```
beck-thompson: sus magic number - the reason its trolling you is because the `int` cast on line 51. Anything under 1 is just truncated to 0. Probably the best course of action would be to:
1.) Do the cast (slightly sus) and store it in a new varable.
2.) Check if that variable is 0, if it's zero, do the popup
3.) Otherwise just add the material like normal

Resources/Locale/en-US/materials/material-extractor.ftl
```diff
@@ -0,0 +1 @@
+material-extractor-comp-wrongreagent = {CAPITALIZE(THE($used))} has no required reagents for extraction.
```
beck-thompson: I'd rename this to like "does not have enough reagents for extraction"

beck-thompson (changes requested):

PR #38428, opened 2025-06-18
Remove `SalvageShuttleCircuitboardStealObjective` from the objective group

VasilisThePikachu:
## Why / Balance

So this was causing vulture to fail to round start because it could not find the prototype... this is being migrated into null anyway so no point in it being here

I am gonna git bisect this now to figure out what broke it, but will quickly test if this fixes round start,

## Media

image

## Requirements


slarticodefast:
I made an issue so we validate these in the future
https://github.com/space-wizards/space-station-14/issues/38429

PR #38431, opened 2025-06-18
Fail if we attempt to run publish on master

VasilisThePikachu:
## Breaking changes

FORKS REMOVE THIS (unless your main branch is not named "master")

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,6 +14,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Fail if we are attempting to run on the master branch
+      if: ${{GITHUB.REF_NAME == 'master'}}
+      run: exit 1
+
     - name: Install dependencies
       run: sudo apt-get install -y python3-paramiko python3-lxml
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
     - name: Fail if we are attempting to run on the master branch
-      if: ${{GITHUB.REF_NAME == 'master'}}
+      if: ${{GITHUB.REF_NAME == 'master'}} && github.repository == 'space-wizards/space-station-14'
       run: exit 1
 
     - name: Install dependencies


Errant-4:
Kid called Breaking Changes

Errant-4:
This is a good change, and thank you, but I am concerned about breaking a lot of downstreams, breaking changes or not. Is there.. anything we could possibly do about that?

.github/workflows/publish.yml
```diff
@@ -14,6 +14,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Fail if we are attempting to run on the master branch
+      if: ${{GITHUB.REF_NAME == 'master'}}
```
FairlySadPanda: This is Wizden-specific, and therefore this check should also check what repo this is being run off.

image

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs

FairlySadPanda (changes requested):

VasilisThePikachu:
> This is a good change, and thank you, but I am concerned about breaking a lot of downstreams, breaking changes or not. Is there.. anything we could possibly do about that?

Will admit, while I did only now think about restricting it to our repo. Downstreams should probably always consider changes in publish actions breaking imo

FairlySadPanda (approved):

FairlySadPanda:
Remember to make sure this works :slugcat_nod:

.github/workflows/publish.yml
```diff
@@ -14,6 +14,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Fail if we are attempting to run on the master branch
+      if: ${{GITHUB.REF_NAME == 'master'}} && github.repository == 'space-wizards/space-station-14'
```
Centronias, line 18:
```suggestion
      if: GITHUB.REF_NAME == 'master' && github.repository == 'space-wizards/space-station-14'
```
Only the left side of the `&&` is an expression and the rest of it is evaluated as a string with some kind of strange truthy value:

image

I'd make this into a real PR so that forks don't suffer, but if the advice is "FORKS REMOVE THIS", eh.

PJB3005: I fucking love YAML programming

VasilisThePikachu: Thanks for bringing this to my attention, I made a fix #38556

You should not have to actually remove it since it should resolve to only run on our fork... but it's an unnecessary step for most forks, so I would suggest so (plus forks really should monitor edits to this action)

PR #38433, opened 2025-06-18
Fix stacks only recycling one instead of the stack amount

### LINKED ISSUE #25516: Recycling whole stacks of materials in reclaimer is bugged
## Description
In case of a single rod and stack of 30 of them both will give you 1 steel sheet if recycled twice, only difference is if there is more than a single rod it will take that much longer to recycle with worst case being full stack of 30 that takes 30 times as long.

**Reproduction**
Insert whole stack of items into reclaimer, then another.
Now insert a single/lesser amounts twice and compare what you get.

beck-thompson:
## About the PR

Resolves #25516 

## Technical details

Not sure if I should use an event or just do a check like this - an event seems a little overkill but if someone can think of another example of where this might be needed I'll update it.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Recycler now properly recycles item stacks.

### DIFF
--- a/Content.Server/Materials/MaterialReclaimerSystem.cs
+++ b/Content.Server/Materials/MaterialReclaimerSystem.cs
@@ -19,13 +19,13 @@ using Content.Shared.Materials;
 using Content.Shared.Mind;
 using Content.Shared.Nutrition.EntitySystems;
 using Content.Shared.Power;
+using Content.Shared.Stacks;
 using Robust.Server.GameObjects;
 using Robust.Shared.Player;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Utility;
 using System.Linq;
 using Content.Shared.Humanoid;
-using Content.Shared.Stacks;
 
 namespace Content.Server.Materials;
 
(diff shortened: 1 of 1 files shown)


Content.Server/Materials/MaterialReclaimerSystem.cs
```diff
@@ -25,6 +25,7 @@
 using Robust.Shared.Utility;
 using System.Linq;
 using Content.Shared.Humanoid;
+using Content.Shared.Stacks;
```
slarticodefast: sort 😢

Content.Server/Materials/MaterialReclaimerSystem.cs
```diff
@@ -183,7 +184,11 @@ public override void Reclaim(EntityUid uid,
 
         var xform = Transform(uid);
 
-        SpawnMaterialsFromComposition(uid, item, completion * component.Efficiency, xform: xform);
+        var modifer = 1.0f;
+        if (TryComp<StackComponent>(item, out var stack))
```
slarticodefast: I would move this into the `SpawnMaterialsFromComposition` method itself because this is a property of the item we are passing into it.
Also now you have two parameters, `efficiency` and `modifier` which do exactly the same thing.

Content.Server/Materials/MaterialReclaimerSystem.cs
beck-thompson: Rider actually trolled me

Content.Server/Materials/MaterialReclaimerSystem.cs
beck-thompson: yeah that's fair :thinking:

Content.Server/Materials/MaterialReclaimerSystem.cs
slarticodefast: they always say that
poor rider 😞

slarticodefast (approved):
Looks good!

PR #38436, opened 2025-06-19
Gave bartender zipties & handcuff legality

Kittygyat:
## About the PR
I gave the bartender legal rights to hold zipties, aswell as other cuffs, and added two zipties to the bardrobe for them to use, I also made a new contraband level to suit this new situation, a hybrid security command & bartender legality.

## Why / Balance
This was done to give the bartender a legal way to restrain problem customers until security arrives, if they've caused enough of an issue to deserve being stunned by the bartenders beanbag rounds, without the bartender having to hold illegal zipties to do so.

It's felt odd to me that the bartender has access to a stunning firearm, yet had no (legal) way to deal with the threat/issue once they're actually stunned, short of critting them whilst they're down, or letting them stand up & continue, potentially causing even bigger issues for the bartender, so I gave them a potential solution without the legality concerns.

Balance wise, antagonists might feel a bit more of an allure to the role, but they get armour & firearms already, so they're already a solid pick for them, another - minor - concern is that some bartenders might get too excited & go validhunt an antagonist, but they can do that even before this PR, with only 15 lv wires for some cuffs of their own.

## Technical details
Added a new contraband level to base_contraband; "BaseSecurityCommandBartenderContraband"

Changed base handcuff contraband level to "BaseSecurityCommandBartenderContraband"

Added two zipties to the Bardrobe's inventory

## Media
bartendercuffupdate
bartendercuffsvender

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Gave bartenders legal rights & access to cuffs, with two zipties now being available in their Bardrobe.

### DIFF
--- a/Resources/Prototypes/Catalog/VendingMachines/Inventories/bardrobe.yml
+++ b/Resources/Prototypes/Catalog/VendingMachines/Inventories/bardrobe.yml
@@ -16,5 +16,6 @@
     ClothingOuterVest: 2
     ClothingBeltBandolier: 2
     ClothingEyesGlassesSunglasses: 2
+    Zipties: 2
   contrabandInventory:
     ToyFigurineBartender: 1
--- a/Resources/Prototypes/Entities/Objects/Misc/handcuffs.yml
+++ b/Resources/Prototypes/Entities/Objects/Misc/handcuffs.yml
@@ -2,7 +2,7 @@
   name: handcuffs
   description: Used to detain criminals and other assholes.
   id: Handcuffs
-  parent: [BaseItem, BaseSecurityCommandContraband]
+  parent: [BaseItem, BaseSecurityCommandBartenderContraband]
   components:
   - type: Item
     size: Small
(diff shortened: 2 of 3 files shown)


YoungThugSS14:
I would rather valid hunting not be encouraged

RedBookcase:
Against this as it does encourage validhunting and also makes bartender even more annoying when raiders take the role. Bartenders should be calling Security if they get to the point that they're considering handcuffing someone.

ArtisticRoomba:
Yeah, not really a fan of this one. I don't really like bartenders just having these and being allowed to carry them - the shotgun is really just to take potshots at the tider breaking all of your glasses or people hopping the counter (after you tell them to get the hell out and they don't listen).

HyperB1:
The point of bartender's self-defense weapons is to keep aggressive drunks away from themselves, guests or the equipment, not to arrest or restrain anybody. There is no sense in making bartenders _technically_ able to do security's work. It also creates a grey area in the rules of what bartenders are and are not allowed to do.

EmoGarbage404:
As far as i see it, this isn't a problem. Nothing is mechanically stopping bartenders from making zipties or having cuffs on them and using them. Security can and should have the right to confiscate those if they see fit, which is currently how it functions. Just because they occasionally deal with rowdy shitters in bar doesn't mean that they need to be empowered as sec-lite.

PR #38437, opened 2025-06-19
Fix PDAs placed on belt slot not identifying players properly

hoshizora-sayo:
## About the PR

Players that place their PDA on their belt slot can now have the job icon of the contained ID, and can be properly identified while covering their face. 
## Why / Balance

I think if you're allowed to put a PDA on your belt slot it probably should work just like as if you had it in your ID slot. This also lets the "free spirits" who don't like jumpsuits and can't use the ID slot be considered crew by silicons and not have security harass them for appearing ID-less. 
## Technical details

Adjusted some conditionals to also check the belt slot, first time doing any c# changes so I may have messed up some sort of coding convention/standard for this codebase. In the case of multiple equipped IDs, the one in the ID slot will take precedence.
## Media

(before) PDA on belt but no identification
image
(after) Fixed
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Players wearing their PDA on their belt slot can now be properly identified

### DIFF
--- a/Content.Shared/Access/Systems/AccessReaderSystem.cs
+++ b/Content.Shared/Access/Systems/AccessReaderSystem.cs
@@ -629,7 +629,7 @@ public sealed class AccessReaderSystem : EntitySystem
         items = new(_handsSystem.EnumerateHeld(uid));
 
         // maybe its inside an inventory slot?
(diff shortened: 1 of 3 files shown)


slarticodefast:
Why do they even fit into the belt slot in the first place?
It's not like the pda slot would be blocked by anything else.

PR #38123, opened 2025-06-07
Thieves can innately pickpocket - thieving gloves rework

ArtisticRoomba:
## About the PR
Thieves can now innately pickpocket. They do not require any gloves or clothing to do this.
They can also toggle this action in order to fake not pickpocketing.

The Thieving gloves have been removed from the Traitor uplink.

## Why / Balance
Thieving gloves are one of the most annoying things out there, both from a gameplay perspective and an admin perspective.

Admins have to deal with the endless peril that is dealing with metagaming rulebreaks, as well as the malds from players around whether (x, y, z) got meta'd or not.

If Thieves get their gloves meta'd and stolen, or just stolen in general, they lose out on a pretty big one-up that defines them. Stealing stuff stealthily is a huge boon. Nab an engi ID and you have engi access. Go get some free insuls or an RCD. Nab a sci ID and you have sci access. Grab yourself stuff from the lathe. That type of stuff. Generally if you lose this ability early into the round (meta'd or not) it dramatically lowers your options and makes the role a little bit unfun.

Thieves, now having intrinsic gloves, no longer have to worry about being metagamed and having their gloves stolen early, as now they themselves are the stealth item, only now they can toggle stealing whenever they want in order to wiggle out of anything. Thieve-checking isn't possible: the only thing you'll be able to bonk em for is restricted contraband.

Thieving gloves were removed from Traitors for two reasons:
- Traitors using thieving gloves to silently steal high-value items is the most boring thing on the planet and does not lead to any interesting roleplay. It only leads to LOOC malding about metagaming and AHelps.
- If we remove thieving gloves from the game entirely, we don't have to worry about metagaming.

Traitors should be shooting and killing people to get to their high value targets. Going back into "stealth mode" after committing a crime of shooting someone is both rewarding for Security and the Traitor:
- Traitor gets an adrenaline rush for making a play
	- If Traitor gets away with it, Traitor gets a wave of enjoyment from getting away with a crime
- Security now has an active threat to respond with
	- Adds actual gameplay to the role instead of Sec/Detective having to put their hands up in the air
	- Gives evidence for the detective to make a move on (yes I know the detective has barely any content, if you argue this in this PR I will be very disappointed). 

## Technical details
Code is now fully and properly predicted. We add a new event and action for toggling the ability.

## Media
rider64_iG7tOdqZxp

## Requirements

[...]

PR #38442, opened 2025-06-19
Ripping gloves

pheenty:
## About the PR

Taking an item (or forcefully giving one) from a person now gives them a popup not only when starting the doafter, but also when it's done. This does not apply to stealthy stealing, that thief and space ninja have.
Added in-between of normal (visible) and fully stealthy stealing - ripping. It will hide the initial popup and doafter bar, but will still show popup at the end.
Added chameleon ripping gloves to uplink.
## Why / Balance

https://github.com/space-wizards/space-station-14/pull/38123#issuecomment-2952717387
## Media

https://github.com/user-attachments/assets/34d3fc5d-92e0-48f5-9713-fbff2d2e0c95

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Successfully stripping someone of an item, or giving an item, will now also give a popup.
- add: Syndicate agents can now purchase chameleon ripping gloves. They hide the stealing process, but the target would still know that you ripped an item from them.

### DIFF
--- a/Content.Shared/Clothing/EntitySystems/ToggleableClothingSystem.cs
+++ b/Content.Shared/Clothing/EntitySystems/ToggleableClothingSystem.cs
@@ -95,7 +95,7 @@ public sealed class ToggleableClothingSystem : EntitySystem
         if (component.StripDelay == null)
             return;
 
-        var (time, stealth) = _strippable.GetStripTimeModifiers(user, wearer, item, component.StripDelay.Value);
+        var (time, stealth, _) = _strippable.GetStripTimeModifiers(user, wearer, item, component.StripDelay.Value);
 
         var args = new DoAfterArgs(EntityManager, user, time, new ToggleClothingDoAfterEvent(), item, wearer, item)
         {
--- a/Content.Shared/Strip/Components/StrippableComponent.cs
+++ b/Content.Shared/Strip/Components/StrippableComponent.cs
@@ -32,14 +32,15 @@ namespace Content.Shared.Strip.Components
     public sealed class StrippingEnsnareButtonPressed : BoundUserInterfaceMessage;
 
     [ByRefEvent]
-    public abstract class BaseBeforeStripEvent(TimeSpan initialTime, bool stealth = false) : EntityEventArgs, IInventoryRelayEvent
+    public abstract class BaseBeforeStripEvent(TimeSpan initialTime, bool stealth = false, bool popup = true) : EntityEventArgs, IInventoryRelayEvent
     {
         public readonly TimeSpan InitialTime = initialTime;
         public float Multiplier = 1f;
         public TimeSpan Additive = TimeSpan.Zero;
         public bool Stealth = stealth;
+        public bool Popup = popup;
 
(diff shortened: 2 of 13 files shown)


qrwas:
maybe it's better to add a delay for this notification if it concerns theft using gloves, and add this to the thief, but with a longer delay? conditionally after 3 seconds the inscription (you feel like something has been taken from you), or something else?

qrwas:
or make two modes: quick removal, which will not be visible, but it will show the notification immediately, and hidden, which will be 1.5-2 times longer without the notification?

pheenty:
> maybe it's better to add a delay for this notification if it concerns theft using gloves, and add this to the thief, but with a longer delay? conditionally after 3 seconds the inscription (you feel like something has been taken from you), or something else?

Thief's whole point is to stay stealthy. They also have more objectives than traitors.
I don't think traitors should have stealth stealing ever.
Ripping already allows to steal something and then scram/quickly run away on hyperzine, which is enough.
Also popup respects your identity, so you can also combine this with hiding your face.
Take ID off, hide your face, come to medical, rip hypo out of cmo's hands, inject yourself with hyperzine and run away.
Thta does not require killing anyone, keeps your real identity clear, but actually brings something into the round instead of just silently stealing hypo and going away like nothing happened without CMO even noticing.

Hitlinemoss:
If it isn't already, the theft notification should be visible to all players, not just the thief and the victim. Would suck for a traitor to just rip the captain's backpack off his body and the HoS standing right next to them to not even notice until the captain types out "that guy stole my bag".

pheenty:
> If it isn't already, the theft notification should be visible to all players, not just the thief and the victim. Would suck for a traitor to just rip the captain's backpack off his body and the HoS standing right next to them to not even notice until the captain types out "that guy stole my bag".

It doesn't really show the popup already. While the doafter is visible, it can be literally anything else. I regularly just steal items from dead bodies as a traitor md or paramedic in the middle of medical and people usually don't care. Your point is fair enough, but I'd like to see what maintainers think

Hitlinemoss:
Mostly I'm just thinking that violently ripping an item off of someone should be obvious to onlookers in the same way as giving a hug or something. But yeah, maintainer feedback on this would be best.

Cojoke-dot:
3 tc is a little cheap. A lot of stuff sits in intervals of 4 for some reason, so maybe bump it up by one at the very least. How much were thiefing gloves before?

The alert someone just took/gave something should be the same size as the strip start message, really big white text and small red text feels weird. Idk about if its a public message everyone can see though if it is a bigger size, that could get annoying.

Yeah I like this, it gives security a target if the person who got stripped actually reads the message

pheenty:
> 3 tc is a little cheap. A lot of stuff sits in intervals of 4 for some reason, so maybe bump it up by one at the very least. How much were thiefing gloves before?

Thieving gloves were 5. Even 4 initially, then they were given chameleon ability for an extra tc. And they weren't really OP, they were just boring. So 3 for a nerfed version should be fine

> The alert someone just took/gave something should be the same size as the strip start message, really big white text and small red text feels weird. Idk about if its a public message everyone can see though if it is a bigger size, that could get annoying.

True, I might change that later

pheenty:
Okay can't be bothered to fix the merge conflict, at least for now, feel free to pick this up

PR #38438, opened 2025-06-19
HandsSystem Refactor

EmoGarbage404:
## About the PR

i just wanted to define hands in yaml, man...

rewrites a lot of Hands code to not do insane things and actually work for serialization. Accidentally ends up touching an insane number of random files. Also sniped the horrific client state handling code that was causing @slarticodefast hell.

Basically no gameplay effects though i did change the aghost to use yaml-defined hands as a proof-of-concept.

## Technical details

Primary technical change is that hands now serialize (yay) and that the `Hand` class has been massively simplified. Instead of containing the location, container id, container reference, and helper methods, it now is just a simple class that contains only the location.

As such, most of the API has been flipped to replace usages of `Hand` with the hand's corresponding ID. This was a part of the old system too but this ID actually got stored twice which is kinda a nightmare so now it's just one spot. Additionally, anything using the Hand class to get data now needs to go through a system method. Which means.... breaking changes.

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

`HandHelpers.cs` has been entirely removed. These can all be replaced with `HandsSystem` method calls. 

The `Hand` class has been changed to only contain semantic data pertaining to hand configurations. This means that it no longer contains a reference to the container ID, the container slot, or helper methods. To access the held item, use `SharedHandsSystem.TryGetHeldEntity` with the hand-entity and corresponding handId passed in. Checking if the hand is empty is similarly done with `SharedHandsSystem.HandIsEmpty`.

`HandsComponent.ActiveHand` has been removed. Instead, it now contains `ActiveHandId`, which is the ID of the corresponding hand in the `Hands` dict. `SharedHandsSystem` has helper methods for accessing things related to the active hand, namely `GetActiveHand`, `GetActiveItem`, `ActiveHandIsEmpty`, and `TryGetActiveItem`.

Various methods in `SharedHandsSystem` now take-in or return a string corresponding to the hand ID rather than the hand class itself. For brevity, i'll just make a list:
- `CanDropHeld`
- `TryDrop`
- `DoDrop`
- `TryPickup`
- `CanPickupToHand`
- `DoPickup`
- `AddHand`
- `TryGetEmptyHand`
- `TryGetActiveItem`
- `GetActiveHand`
- `EnumerateHands`
- `SetActiveHand`
- `IsHolding`
[...]

PR #38452, opened 2025-06-19
Add interaction test for retractable arm blade

Tayrtahn:
## About the PR

Adds an interaction test that makes sure that `RetractableItemActionSystem` is working correctly.

## Why / Balance

https://github.com/space-wizards/space-station-14/pull/38438#discussion_r2157022917 identified an issue with the `HandsSystem` refactor that would've been caught if we had more test coverage. Having this test in place will make sure that we can't break it in the future.

Additionally, running this test against the `HandsSystem` refactor showed that the supposed fix for the issue was insufficient. This test can also be used to help find the source of the problem.

## Technical details

- Make sure the player starts with an empty hand.
- Give the arm blade action to the player.
- Make sure the player still has an empty hand.
- Activate the arm blade action.
- Make sure the player is now "holding" the arm blade.
- Activate the arm blade action again.
- Make sure the player's hand is empty again.

## Requirements

### DIFF
--- a/Content.IntegrationTests/Tests/Actions/RetractableItemActionTest.cs
+++ b/Content.IntegrationTests/Tests/Actions/RetractableItemActionTest.cs
@@ -3,7 +3,6 @@ using Content.IntegrationTests.Tests.Interaction;
 using Content.Shared.Actions;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.RetractableItemAction;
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
Most of this logic should be in a `WaitAssertion` block directly. You are playing with fire by mixing threads like this.

PJB3005 (approved):

PR #38456, opened 2025-06-19
Remove Icon Smoothing for Airlocks, Doors, and Plastic Flaps, and Shutters from Walls

ps3moira:
## About the PR

Remove IconSmoothing from Airlocks, High Security Doors, Regular Doors and Shutters.
## Why / Balance

These were left over from when the cev-eris sprites were used. Since cev-eris sprites are no longer used so it makes no sense to keep using the left over IconSmoothing.

## Media

image
image
image

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Doors/Airlocks/airlocks.yml
+++ b/Resources/Prototypes/Entities/Structures/Doors/Airlocks/airlocks.yml
@@ -342,9 +342,6 @@
   components:
   - type: Sprite
     sprite: Structures/Doors/Airlocks/Standard/xeno.rsi
-  - type: IconSmooth
-    key: chitin
-    mode: NoSprite
 
 - type: entity
   parent: AirlockGlass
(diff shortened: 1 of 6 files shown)


slarticodefast:
For reference: This was how the game looked like when it was added to airlocks 😄 
grafik

slarticodefast:
It should probably be removed from plastic flaps as well
grafik

ps3moira:
> It should probably be removed from plastic flaps as well grafik

true almost forgot!

EmoGarbage404 (approved):
you make me proud, lad

slarticodefast (commented):
`BlastDoorXeno` still has smoothing.
Although those might be different from normal airlocks due to the organic look.

walksanatora:
imma be honest... I dont see the diffrence...

walksanatora:
OH WAIT it wasn't anything with the door sprite it's self. it was that the walls were no longer merging with the door. yeah good change. looks nice.

PR #38458, opened 2025-06-19
fix: set correct layer visibility for power cell sprites

perryprog:
## About the PR

Corrects the logic for power cell sprites' layer visibility.

## Why / Balance

Once a power cell hit the uncharged state where it has no little indicator light, it got permanently stuck there.

## Technical details

I also slightly cleaned up the logic, even though this could also be fixed by changing a false to a true. I can switch to that implementation if wanted.

## Requirements

## Breaking changes

**Changelog**

no

### DIFF
--- a/Content.Client/PowerCell/PowerCellSystem.cs
+++ b/Content.Client/PowerCell/PowerCellSystem.cs
@@ -49,7 +49,7 @@ public sealed class PowerCellSystem : SharedPowerCellSystem
             return;
 
         if (!_appearance.TryGetData<byte>(uid, PowerCellVisuals.ChargeLevel, out var level, args.Component))
-            return;
+            level = 0;
 
         var positiveCharge = level > 0;
         _sprite.LayerSetVisible((uid, args.Sprite), PowerCellVisualLayers.Unshaded, positiveCharge);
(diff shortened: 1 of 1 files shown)


Content.Client/PowerCell/PowerCellSystem.cs
```diff
@@ -48,17 +48,14 @@ private void OnPowerCellVisualsChange(EntityUid uid, PowerCellVisualsComponent c
...
             return;
 
-        if (_appearance.TryGetData<byte>(uid, PowerCellVisuals.ChargeLevel, out var level, args.Component))
-        {
-            if (level == 0)
-            {
-                _sprite.LayerSetVisible((uid, args.Sprite), PowerCellVisualLayers.Unshaded, false);
-                return;
-            }
+        if (!_appearance.TryGetData<byte>(uid, PowerCellVisuals.ChargeLevel, out var level, args.Component))
```
PJB3005, line 51: While you're at it, please fix this check: the visualizer should default this value, not bail entirely.

PJB3005 (changes requested):

Content.Client/PowerCell/PowerCellSystem.cs
perryprog: Good point!

PR #36881, opened 2025-04-24
Crawling Part 1: The Knockdownening

Princess-Cheeseballs:
## About the PR
Adds crawling 
Requires #36847
Requires #37196

Resolves: #34894 
Resolves: #1764

This is part 1 of adding crawling to upstream. Currently included in this PR:

### Crawling!!!
- You can press the C button to crawl (can be rebound and character menu has been moved to U).
- Crawling is the same as the knockdown state
- After a minimum knockdown time you can go through a DoAfter to stand up, this DoAfter happens automatically if you didn't purposefully fall down
- You cannot stand up if it would cause you to get stuck (NO MORE CLIPPING INTO WALLS WHEN SLIPPING!!!)
- Lube no longer chain stuns you, instead it keeps you knocked down (You also can't crawl on lube it's too slippery)
- Lizards can now bask for real

### Force Stand!!!
- You can right click yourself to force yourself to stand before your DoAfter is complete
- Only works if you're not stunned, incapacitated, and if your knockdown timer is up
- Costs 10 Stamina, will fail if it would StamCrit you
- You cannot force stand if you do not have an empty hand

### Tasers!!!
- Tasers have been revamped into being a knockdown weapon
- Tasers have very short range only about 4 tiles and cause 2.5 seconds of knockdown on hit as well as slowing the target
- Added an Admeme Taser which is like the old Taser in case you enjoy the stun meta

### Slipping Changes
- Slipping now does 0.5 Stun by default (configurable)
- Slip time has been turned into KnockdownTime
- Slipping now does stamina damage, default 25 goes up to 50 for Syndie Soap and Omega Soap (THIS CAN STAMCRIT)
- Slipping has a FrictionStatus it can apply, default is 0.5 friction and 0.5 seconds time
- FrictionStatus uses the new status effects system!

### SlidingComp is back!!!
- When crawling, you will inherit the friction from puddles just like if you were StamCrit
- You can now slide across lube puddles and propel yourself by throwing items or shooting a gun just like when weightless (works on all puddles but lube is the most effective)
- It also no longer throws a billion Debug Asserts

### Knocked Down Icon!!!
- Cool icon for being knocked down
- Shows how long until you can start standing
- Can be clicked to try and stand
- Will try force stand if you're in a do-after or if the timer isn't up, otherwise will try to stand you up normally

### Worm Smite
- Doesn't actually turn them into a worm (we don't have worms apparently that's fucked up)
- Makes it so they start crawling and they cannot stand up

### Missing Stuff
[...]

PR #38459, opened 2025-06-20
Down State Friction Fix

Princess-Cheeseballs:
## About the PR
Fixed a bug/oversight involving the knocked down friction modifier not being applied to all states

Before  #36847 was merged, stunned, crit, and dead mobs had 60% less friction than their standing counterparts, but this was tied to being unable to move in the tileFrictionController which lead to bad friction behavior when trying to implement crawling.

My Movement Refactor PR fixed this by syncing the two values as much as possible, but I incorrectly assumed that when a mob was down they had the KnockedDownComponent which was an incorrect assumption. KnockedDown is only applied when you are stunned (or sometimes when you're instantly killed I think?) which meant that for basically every downed mob state except for stunned, friction was incorrectly 2.5x higher than it should have been. 

This PR fixes that.

## Why / Balance
Unintended behavior.

## Technical details
- Revert the const in SharedStunSystem to its original value

- Moved the 0.4 friction modifier that was initially bundled into the SharedStunSystem const to the StandingStateComponent as a datafield

- Added event listeners to StandingStateSystem to apply this friction modifier when downed.

- Added event listeners to MovementModifierSystem to update friction when downed.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Dead and crit bodies are again easier to drag.


Princess-Cheeseballs:
I merged the fixes from this into #36881 so if that PR gets merged I'll close this one and if this one gets merged I'll do absolutely nothing because I don't need to worry about merge conflicts anymore.

Princess-Cheeseballs:
With #36881 merged this has been resolved.

PR #38460, opened 2025-06-20
Paracusia + MassHallucination minor refactor

perryprog:
## About the PR

This refactors the paracusia trait to be a bit more in-line with our current standards, especially with regard to TimeSpan updating, fixing a few bugs in the process. Most notably, paracusia incidents would build up while your player was detached from the character, making reattaching VERY loud.

Also cleans up the component definition to use auto paused fields, removes some unneeded field names, and unmarks fields as required that didn't need to be.

## Why / Balance

Headphone ops.

## Technical details

If this should've been two PRs I'm sorry :(

Also dang, positive diff for a "cleanup" PR...

## Requirements

## Breaking changes

- Server-side methods in ParacusiaSystem have been moved to use the Entity<T> pattern.
- Several fields in ParacusiaComponent and MassHallucinationsRuleComponent have been moved to be TimeSpans instead of float seconds.

**Changelog**

no

### DIFF
--- a/Content.Client/Traits/ParacusiaSystem.cs
+++ b/Content.Client/Traits/ParacusiaSystem.cs
@@ -1,25 +1,28 @@
 using System.Numerics;
+using Content.Shared.ActionBlocker;
 using Content.Shared.Traits.Assorted;
-using Robust.Shared.Random;
 using Robust.Client.Player;
-using Robust.Shared.Player;
-using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
+using Robust.Shared.Player;
+using Robust.Shared.Random;
 using Robust.Shared.Timing;
 
 namespace Content.Client.Traits;
 
 public sealed class ParacusiaSystem : SharedParacusiaSystem
 {
-    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly ActionBlockerSystem _actionBlock = default!;
+    [Dependency] private readonly SharedAudioSystem _audio = default!;
     [Dependency] private readonly IPlayerManager _player = default!;
     [Dependency] private readonly IRobustRandom _random = default!;
-    [Dependency] private readonly SharedAudioSystem _audio = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
 
     public override void Initialize()
     {
         base.Initialize();
-        SubscribeLocalEvent<ParacusiaComponent, ComponentStartup>(OnComponentStartup);
+
+        SubscribeLocalEvent<ParacusiaComponent, MapInitEvent>(OnMapInit);
+        SubscribeLocalEvent<ParacusiaComponent, LocalPlayerAttachedEvent>(OnPlayerAttached);
         SubscribeLocalEvent<ParacusiaComponent, LocalPlayerDetachedEvent>(OnPlayerDetach);
     }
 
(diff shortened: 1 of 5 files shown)


Content.Shared/Traits/Assorted/ParacusiaComponent.cs
```diff
@@ -9,39 +9,41 @@ namespace Content.Shared.Traits.Assorted;
...
+    [AutoNetworkedField]
+    public float MaxSoundDistance;
+
     /// <summary>
     /// The maximum time between incidents in seconds
     /// </summary>
-    [DataField("maxTimeBetweenIncidents", required: true), ViewVariables(VVAccess.ReadWrite)]
+    [DataField]
     [AutoNetworkedField]
     public float MaxTimeBetweenIncidents = 60f;
```
slarticodefast: this should be a timespan as well

Content.Shared/Traits/Assorted/ParacusiaComponent.cs
```diff
@@ -9,39 +9,41 @@ namespace Content.Shared.Traits.Assorted;
...
 
     /// <summary>
     /// The minimum time between incidents in seconds
     /// </summary>
-    [DataField("minTimeBetweenIncidents", required: true), ViewVariables(VVAccess.ReadWrite)]
+    [DataField]
     [AutoNetworkedField]
     public float MinTimeBetweenIncidents = 30f;
```
slarticodefast: same here

Content.Shared/Traits/Assorted/ParacusiaComponent.cs
perryprog: Augh, good point. This has a chain of use that goes a bit further, so unless you have any objection I'll clean those up too. (Converting server-side ParacusiaSystem to Entity<T>, cleaning up MassHallucinationsRuleComponent.)

perryprog:
On second thought, I wonder if I should have bothered with changing the server-side ParacusiaSystem methods to be Entity<T> style, since that's a breaking change (I'm sure this is a super heavily used system downstream...). Should I revert that part, or alternatively, should I add obsolete EntityUid overloads?

Content.Client/Traits/ParacusiaSystem.cs
```diff
@@ -30,46 +33,63 @@ public override void Update(float frameTime)
...
             return;
 
         PlayParacusiaSounds(localPlayer);
     }
 
-    private void OnComponentStartup(EntityUid uid, ParacusiaComponent component, ComponentStartup args)
+
+    private void OnMapInit(Entity<ParacusiaComponent> ent, ref MapInitEvent args)
+    {
+        ent.Comp.NextIncidentTime = _timing.CurTime;
```
slarticodefast, line 46: MapInit is only raised on the server so this does nothing.
Also the datafield is autonetworked, so even if we would call this on the client it would get bulldozed again by the server state.

Content.Server/StationEvents/Components/MassHallucinationsRuleComponent.cs
```diff
@@ -1,30 +1,29 @@
...
     public float MaxSoundDistance;
 
-    [DataField("sounds", required: true)]
-    public SoundSpecifier Sounds = default!;
+    [DataField(required: true)]
+    public SoundSpecifier Sounds;
 
-    [DataField, ViewVariables(VVAccess.ReadOnly)]
-    public List<EntityUid> AffectedEntities = new();
+    [ViewVariables(VVAccess.ReadOnly)]
```
slarticodefast, line 27: why did you remove the datafield attribute?
This will break persistance if the game is saved and loaded while the rule is active.

Content.Shared/Traits/Assorted/ParacusiaComponent.cs
```diff
@@ -9,39 +9,41 @@ namespace Content.Shared.Traits.Assorted;
...
+    /// The minimum time between incidents.
     /// </summary>
-    [DataField("maxSoundDistance", required: true), ViewVariables(VVAccess.ReadWrite)]
+    [DataField]
     [AutoNetworkedField]
-    public float MaxSoundDistance;
+    public TimeSpan MinTimeBetweenIncidents = TimeSpan.FromSeconds(30);
+
+    [AutoPausedField]
+    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
```
slarticodefast, line 38: If this is played purely client side then it does not really make sense to have this by a datafield or autopaused

slarticodefast (changes requested):

TheShuEd:
This PR has been stuck in the “Awaiting Changes” status for over two months with no visible activity, so I'm closing it as abandoned.

If you still plan to continue working on this, please let us know.

PR #38462, opened 2025-06-20
Rearranged layout of main ATS room

Hitlinemoss:
## About the PR

The layout of the ATS's main room has been adjusted slightly, with the main change being that the vertical wall in the center of the room has been rotated horizontally, one side being a single large buying pallet and one side being a single large selling pallet (alongside the sale computer).

The ATS's exterior walls and the layouts of other rooms remain unchanged.

## Why / Balance

The ATS's current layout is slightly inconvenient in a frustrating way; crates spawn on random pallets, so there's a random chance that you will need to walk around to the far side of the wall and drag a crate all the way across the room. In extreme cases, the vision-obstruction of the wall can cause players to miss a crate entirely. Rotating the wall improves Cargo's ability to see and move crates, which makes ATS trips slightly faster to accomplish.

One side effect of the new pallet shapes is that each type of pallet has slightly more area (a 7x4 area instead of two 3x4 areas), meaning that the ATS can hold slightly more purchases at once (4 more crates). It is extremely rare for the ATS to actually fill to capacity (and in situations like that it's usually because nobody is capable of actually making ATS trips anyway), so this likely isn't a balance concern.

## Technical details

Mapping changes to Resources\Maps\Shuttles\trading_outpost.yml

## Media

moss-betterats 1
moss-betterats 2

## Requirements

## Breaking changes

N/A

**Changelog**

:cl:
- tweak: The ATS's layout has been adjusted slightly to expedite the process of loading and unloading shipments.

### DIFF
--- a/Resources/Maps/Shuttles/trading_outpost.yml
+++ b/Resources/Maps/Shuttles/trading_outpost.yml
@@ -1,11 +1,11 @@
 meta:
   format: 7
   category: Grid
-  engineVersion: 260.2.0
+  engineVersion: 262.0.0
   forkId: ""
   forkVersion: ""
-  time: 05/28/2025 08:16:27
-  entityCount: 1096
+  time: 06/20/2025 05:09:03
+  entityCount: 1068
 maps: []
 grids:
 - 2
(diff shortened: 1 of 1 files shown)


ArtisticRoomba:
While it looks cool, I did like the separation as ATS fights would have some interesting ring-around-the-rosie gameplay, protecting people from fire while they got into their shuttle. I'm pretty neutral on the removal overall though. I'll see what @spanky-spanky thinks about it.

Hitlinemoss:
Removed those interior windows between the two halves of the room to make it faster to just walk between the buy and sale pallets (which is the main point of this PR in the first place), replacing them with firelocks like there already were at the edges of the wall. This further complicates the ring-around-the-rosie gameplay a few people have talked about, though the sections of reinforced wall remain so those are still usable as cover.

Also added a second request computer since it's been pointed out to me that having only one computer makes it possible for a really unlucky meteor to just kill the ATS until the QM grabs their spare computer board and replaces it.

TiniestShark:
I personally never really found this frustrating enough to warrant a change and it has an artificial appearance to it compared to the old layout. Not to mention it's going to be much harder to outmaneuver any vent critters that spawn on the ATS if you're going solo. I'd rather have crates only deliver to one palette until it's full then switch to the other but that's a far more difficult solution I'd imagine.

TrixxedHeart:
as a cargo main I approve please god the ATS endlessly made me upset lmao

I think the artificial look is fine, it ironically feels more natural to me given how unintuitive the old one is.

southbridge-fur:
darn it, now I need to remake the ATS portion of Amber!

Hitlinemoss:
> darn it, now I need to remake the ATS portion of Amber!

Maps directly load the trade_outpost.yml file when spawning the ATS, so there shouldn't need to be any map-specific updates. Unless there's some specific case with Amber that I'm not aware of.

MidZik:
> > darn it, now I need to remake the ATS portion of Amber!
> 
> Maps directly load the trade_outpost.yml file when spawning the ATS, so there shouldn't need to be any map-specific updates. Unless there's some specific case with Amber that I'm not aware of.

Amber's cargo area looks like the ATS, it's just part of its design. So it has a figurative design dependency, not a literal code dependency.

Hitlinemoss:
> Amber's cargo area looks like the ATS, it's just part of its design. So it has a figurative design dependency, not a literal code dependency.

Ah, okay.

Admiral-Obvious-001:
How does this address cargo or whoever sawing off the ATS and moving it to the station? The current version we have got an exterior hull plate reinforced tile spine to prevent this.

EthanQix:
> How does this address cargo or whoever sawing off the ATS and moving it to the station? The current version we have got an exterior hull plate reinforced tile spine to prevent this.

Is it something that actually happens on the regular, or could it be fixed with a couple rolebans ?

Admiral-Obvious-001:
> > How does this address cargo or whoever sawing off the ATS and moving it to the station? The current version we have got an exterior hull plate reinforced tile spine to prevent this.
> 
> Is it something that actually happens on the regular, or could it be fixed with a couple rolebans ?

When people figured out how to do it on Lizard, people consistently sawed the TS off from the anchoring room, took the cargo shuttle, moved the TS to cargo, moved the console and generator over, and then hard docked the ATS to cargo, giving salvage the cargo shuttle. Granted you literally only needed a wirecutter to do so.

You can do it with the current ATS using some creative thinking and/or exploits, but the proposed design will be sawn off using some RCD charges and tools.

K-Dynamic:
> darn it, now I need to remake the ATS portion of Amber!

Obviously Amber is just taking the old ATS

Hitlinemoss:
> How does this address cargo or whoever sawing off the ATS and moving it to the station? The current version we have got an exterior hull plate reinforced tile spine to prevent this.

This is just a change to the layout of the main room, so as far as I'm aware there shouldn't be any major problems that could pop up. That being said, this is my first time actually doing a mapping PR, so if there's some subtle thing I might've missed that could cause a problem please let me know so I can take a look at it.

The base floor of the room (underneath the tiles) *should* still all be exterior hull plate, but I'll go into a dev environment in a bit and tear up all the floor tiles with a crowbar to double-check.

Hitlinemoss:
> The base floor of the room (underneath the tiles) _should_ still all be exterior hull plate, but I'll go into a dev environment in a bit and tear up all the floor tiles with a crowbar to double-check.

Okay, I see what the problem is now. The exterior walls of the engineering / plastitanium rooms are all on top of reinforced exterior hull plate, and the reinforced exterior hull plate floors vertically crossing the main room on the current ATS connect to those, which makes it impossible to separate the actual buy/sell pallets from the same grid as the anchor. The proposed design doesn't have the pallets connected to the anchor like this.

Easy solution here would be to just keep the vertical line of hull plating, but that kind of looks... ugly. And I imagine that it's not really ideal to extend the plating to the exterior walls of the main room, either, because that has its own balance concerns.

I'll spend the day today trying to make the vertical line look less ugly, I guess.

Hitlinemoss:
Closing this PR for now since it's gotten stale + this new layout could probably be improved on anyway,

PR #38473, opened 2025-06-20
Cleanup duplicate dependency in `AdminVerbSystem.Smites`

Tayrtahn:
## About the PR

Removes a duplicate dependency in `AdminVerbSystem.Smites.cs`.

## Why / Balance

Fighting the good fight for [PZW](https://github.com/space-wizards/space-station-14/issues/33279).

## Technical details

Deleted the extra dependency and changed the single use of it to reference the other dependency of the same type.

## Requirements

### DIFF
--- a/Content.Server/Administration/Systems/AdminVerbSystem.Smites.cs
+++ b/Content.Server/Administration/Systems/AdminVerbSystem.Smites.cs
@@ -79,7 +79,6 @@ public sealed partial class AdminVerbSystem
     [Dependency] private readonly SharedTransformSystem _transformSystem = default!;
     [Dependency] private readonly SuperBonkSystem _superBonkSystem = default!;
     [Dependency] private readonly SlipperySystem _slipperySystem = default!;
-    [Dependency] private readonly OutfitSystem _outfitSystem = default!;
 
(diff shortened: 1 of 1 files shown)


slarticodefast:
How do these keep coming back? 😄 
Can we maybe have this fail an integration test or promote them to an error?

slarticodefast (approved):

EmoGarbage404:
partial systems are a menace. you'll have like 15 dependencies declares across 6 files with 3 different naming schemes.

Tayrtahn:
Promoting it to an error would do the trick!

Tayrtahn:
~I can't do it myself right now, but it'd be a super easy engine PR - just change the enum value here: https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Analyzers%2FDuplicateDependencyAnalyzer.cs#L25~

PJB3005:
> I can't do it myself right now, but it'd be a super easy engine PR - just change the enum value here: https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Analyzers%2FDuplicateDependencyAnalyzer.cs#L25

No, use `WarnAsError` instead.

PR #38474, opened 2025-06-21
SignalSwitchSystem: Check button is locked before toggling

whatston3:
## About the PR

Access restricted locked switches can currently be toggled without having the correct access when locked.  This fixes that by first checking if they're locked.

## Why / Balance

This fixes a bug.  A locked button should only be pressable when the lock is open - this effectively removes the point of having a lock in the first place if anyone can activate it.

## Technical details

New dependency in SignalSwitchSystem, one call in the ActivateInWorld event handler.

## Media

Before:

There are two switches.  The left is a signal button, the right is a locked button behind some access.  The user, without access, uses both buttons.  Both trigger and turn on a linked light.  The right should not.

https://github.com/user-attachments/assets/286182d1-cd7b-4b4a-ba2a-ff9c3f326d9b

After:

The same setup is repeated.  The user, without access, uses both buttons.  The left triggers, the right does not.  The user picks up their PDA from the floor, opens the right button, and uses it.  It triggers.  They drop the PDA and use it, it triggers.  This shows the desired behaviour.

https://github.com/user-attachments/assets/2f577acf-9f20-4ea1-9248-a09a4e5a98f1

## Requirements

## Breaking changes

**Changelog**

honestly, pretty negligible
:cl:
- fix: Locked buttons must now be unlocked to trigger.

### DIFF
--- a/Content.Server/DeviceLinking/Systems/SignalSwitchSystem.cs
+++ b/Content.Server/DeviceLinking/Systems/SignalSwitchSystem.cs
@@ -1,6 +1,7 @@
 using Content.Server.DeviceLinking.Components;
 using Content.Server.DeviceNetwork;
 using Content.Shared.Interaction;
+using Content.Shared.Lock;
 using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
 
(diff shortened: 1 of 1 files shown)


PJB3005:
Genuinely struggling to understand how this could have happened but well I can reproduce it so...

PJB3005 (approved):

PR #38360, opened 2025-06-16
Cleanup duplicate dependency in `AddActionCommand`

Tayrtahn:
## About the PR

There were two `IPrototypeManager` dependencies in `AddActionCommand`. Now there's just one.

## Why / Balance

https://github.com/space-wizards/space-station-14/issues/33279

@ScarKy0 you so silly.

## Technical details

Deleted one of the dependency fields and changed the code using it to use the other.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

PR #38480, opened 2025-06-21
Treat duplicate dependency warnings as errors

Tayrtahn:
## About the PR

Adjusts build settings so that `RA0032 (Duplicate [Dependency])` warnings are considered errors.

## Why / Balance

These warnings keep sneaking back in (https://github.com/space-wizards/space-station-14/pull/38099, https://github.com/space-wizards/space-station-14/pull/38360, https://github.com/space-wizards/space-station-14/pull/38473) and it's tedious having to keep doing cleanup. Making it an error will cause builds with violations to fail, which will make it much more obvious and stop these getting merged.

## Technical details

Added `RA0032` to the `WarningsAsErrors` properties of the project files for Server, Client, Shared, and Replay.
Removed the command line parameters in the GitHub workflow files that override the project files.

## Requirements

## Breaking changes
Having two `[Dependency]` members of the same Type in a class is now treated as an error rather than just a warning. If you have violations in your code, clean them up by removing the redundant members.

### DIFF
--- a/.github/workflows/build-test-debug.yml
+++ b/.github/workflows/build-test-debug.yml
@@ -42,7 +42,7 @@ jobs:
       run: dotnet restore
 
     - name: Build Project
-      run: dotnet build --configuration DebugOpt --no-restore /p:WarningsAsErrors=nullable /m
+      run: dotnet build --configuration DebugOpt --no-restore /m
 
     - name: Run Content.Tests
       run: dotnet test --no-build --configuration DebugOpt Content.Tests/Content.Tests.csproj -- NUnit.ConsoleOut=0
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test-debug.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       run: dotnet restore
 
     - name: Build Project
-      run: dotnet build --configuration DebugOpt --no-restore /p:WarningsAsErrors=nullable /m
```
Tayrtahn, line 45: Setting these properties by command line args overrides the values set in the project files. Does GitHub not use the project files somehow, or are these redundant?

beck-thompson:
make sure to add this to breaking changes!

.github/workflows/build-test-debug.yml
PJB3005: Probably outdated.

PJB3005 (approved):

PR #38487, opened 2025-06-21
make scurret crate the same price as big lotto (10,000 spesos)

TrixxedHeart:
## About the PR
ʕ●.●ʔ <( wawa

made scurret crates 10,000 spesos instead of 25,000

## Why / Balance
I feel like having scurret crates be essentially the equivalent of 2.5 big lottos is a bit nuts. Personally I think it could end up in causing cargo to neglect more legitimate duties in the pursuits of getting scurrets, which could cause other parts of the station to suffer.

Having it be the same price as the big lotto would at the very least make it a choice for cargo between a big lotto or a scurret, which I think is much more reasonable.

## Technical details
changed cargo_fun.yml scurrets from 25000 to 10000

## Media
N/A

## Requirements

## Breaking changes
changes scurret price

**Changelog**
:cl:
- tweak: Adjusted Scurret Crate from 25,000 to 10,000 spesos.

### DIFF
--- a/Resources/Prototypes/Catalog/Cargo/cargo_fun.yml
+++ b/Resources/Prototypes/Catalog/Cargo/cargo_fun.yml
@@ -224,7 +224,7 @@
     sprite: Structures/Wallmounts/posters.rsi
     state: poster55_legit
   product: CrateFunScurret
-  cost: 25000 # You will surely not regret buying this
+  cost: 10000 # You will surely not regret buying this
   category: cargoproduct-category-name-fun
   group: market
 


EmoGarbage404:
can we wait more than a day before making balance changes. why are we guessing at the effects of the price could be instead of just waiting slightly longer to actually see how it shakes out.

TrixxedHeart:
> can we wait more than a day before making balance changes. why are we guessing at the effects of the price could be instead of just waiting slightly longer to actually see how it shakes out.

source: I foresee it.

Non-joking though I don't see any harm in making a PR for this now if the next few days do shake out that it should be changed. Also generally invites better discussion on this since I saw multiple other people concerned about the price as well, I wouldn't have bothered if other people didn't seem a least a little concerned.

slarticodefast:
Yeah, let's wait a little with this. Players will want to buy the crate more right now since it's new, but once this has stabilized we can look at balance. Also there are plans to use the scurrets in other ways (pizza delivery critter or bartender assistant) and the crate is just there to make them somehow available early.

FairlySadPanda:
Agreed. Long term I think 20k might be the sweet spot. And if players like expensive crates to target as goals for Cargo each round maybe we can come up with some more fun ideas as spesos sinks.

TrixxedHeart:
> Agreed. Long term I think 20k might be the sweet spot. And if players like expensive crates to target as goals for Cargo each round maybe we can come up with some more fun ideas as spesos sinks.

I don't think its a good idea for a speso sink to ever be more expensive than a big lotto honestly.

_Is it funny?_ Yes.
_Is it healthy for cargo's economy?_ I don't think so.

There's gotta be some rhyme and reason to it not to make it too expensive to the point Cargo is tearing up floorboards to get the money for it while neglecting station needs but also not too cheap where you can just buy 500 mothroaches on a whim. I get making scurrets expensive, I don't see it as a good reason to be more expensive than the big lotto. ~~Unless you want to make an excuse that they're exotic space pets and you're buying a permit to own one. (Although I don't think "lore reasons" is a good excuse to change something for gameplay if it could make an experience worse)~~

PR #38490, opened 2025-06-21
fix: use QueueDel for DeleteEntity construction graph action

perryprog:
## About the PR

The DeleteEntity construction graph action was using an unqueued entity delete, meaning systems that modified damage but still need the entity for further processing would explode.

## Why / Balance

The specific bug I ran into is that destroying an unfinished machine frame via melee ends up attempting to grab the transform of the deleted machine frame, as the threshold behavior triggered a node change triggering the EntityDelete construction action. This then causes the rest of the melee handling code catches fire.

## Technical details

Okay, so, the more I think about it the less sure I am this is the way to go about fixing this issue: this has surprisingly far reaching ramifications, though I _think_ it's mostly limited to anything that changes construction nodes and is used to checking for deletion rather than queued deletion.

I wanted to put out this version first since I think it does make more sense for the node action to queue deletion (this is what the Destroy node graph action, via DestructibleSystem, as well as DeleteEntitiesInContainer), but I'm not sure if it's worth the possible pain.

Arguably the real fix here is to just fix the machine frame construction graph to not have a delete action on its starting node, and if that's thought to be best I can just switch to that.

That being said, things seem to test fine™.

## Requirements

## Breaking changes

- The DeleteEntity construction graph action now queues the entity deletion rather than deleting the entity immediately.

**Changelog**

No user facing changes. Probably.

### DIFF
--- a/Content.Server/Construction/Completions/DeleteEntity.cs
+++ b/Content.Server/Construction/Completions/DeleteEntity.cs
@@ -23,7 +23,7 @@ namespace Content.Server.Construction.Completions
             entityManager.EventBus.RaiseLocalEvent(uid, ev);
 
             if (!ev.Cancelled)
-                entityManager.DeleteEntity(uid);
+                entityManager.QueueDeleteEntity(uid);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


AJCM-git:
@perryprog  I cant reproduce your issues with deleting an unfinished frame, got specific steps to reproduce?

PR #38496, opened 2025-06-22
fix: wide swings with resistanceBypass now bypass resists

### LINKED ISSUE #38478: resistanceBypass does not in fact bypass resistances
## Description
Melee weapons with resistanceBypass: true don't bypass target's resistances

**Reproduction**
- Get a melee weapon with resistanceBypass: true
- Hit armored target
- Watch damage be reduced by said armor, even though it should ignore it

**Screenshots**
Nope

**Additional context**
Nope

perryprog:
## About the PR

Wide swings weren't ignore damage resists regardless of if resistanceBypass is set. Now that doesn't happen.

## Why / Balance

Fixes #38478.

## Technical details

This is the easy fix but this system really needs a refactor. For example, resistance bypass (which is set by the comp but also an event) in particular is getting retrieved twice per weapon swing—in both GetDamage and in GetResistanceBypass.

Also lots of code duplication.

## Requirements

## Breaking changes

**Changelog**

Not user facing; only affects downstreams from what I can tell.

### DIFF
--- a/Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
+++ b/Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
@@ -577,6 +577,7 @@ public abstract class SharedMeleeWeaponSystem : EntitySystem
         var distance = Math.Min(component.Range, direction.Length());
 
         var damage = GetDamage(meleeUid, user, component);
+        var resistanceBypass = GetResistanceBypass(meleeUid, user, component);
         var entities = GetEntityList(ev.Entities);
 
         if (entities.Count == 0)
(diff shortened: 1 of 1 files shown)


slarticodefast (approved):
Makes sense to me. I'll let someone else look at this in case this was originally done on purpose for balance reasons. But I would have expected a code comment if that was the case.

perryprog:
IIRC resistance bypass was just added for the execution system which only ran through the light attack code path. When I looked, we don’t have anything that actually has ignore resistance set (execution system just modifies the flag).

beck-thompson:
I can't see how this would be intentional - I looked at some older prs and it didnt really mention anything :thinking:

I agree with perry the system needs refactor so this stuff is clearer :thinking:

PR #38497, opened 2025-06-22
Add a safety net for making sure you don't breathe from the wrong tank

notquitehadouken:
## About the PR

Gas tanks will now warn you if you try to use an improper one

This is not based on what is actually in the tank, only what the tank looks like

## Why / Balance

New players may pick a non-oxygen breathing character and try to breathe from an oxygen tank, then get confused when it says they're not getting any air. This communicates the fact that not every role breathes oxygen. also i was bored

## Media

https://github.com/user-attachments/assets/186cb66c-27c6-4909-981a-aebf37211d16

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Tanks that might have gas you can't breathe will now warn you if you try to use them. Use them again to breathe from them anyway.

### DIFF
--- a/Content.Shared/AirTank/AirTankLooksLikeComponent.cs
+++ b/Content.Shared/AirTank/AirTankLooksLikeComponent.cs
@@ -1,19 +1,22 @@
-﻿namespace Content.Shared.AirTank;
+﻿using Robust.Shared.GameStates;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared.AirTank;
 
 /// <summary>
 ///     Specifies what a certain tank APPEARS to contain. Is not actually what it contains.
 /// </summary>
-[RegisterComponent]
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
 public sealed partial class AirTankLooksLikeComponent : Component
 {
     /// <summary>
     ///     What this air tank would normally contain
     /// </summary>
-    /// <returns></returns>
-    [DataField]
+    [DataField, AutoNetworkedField]
     public AirTankLooksLike Contains = AirTankLooksLike.NotAir;
 }
 
+[Serializable, NetSerializable]
 public enum AirTankLooksLike : byte
 {
     Invalid,
--- a/Content.Shared/AirTank/AirTankLooksLikeComponent.cs
+++ b/Content.Shared/AirTank/AirTankLooksLikeComponent.cs
@@ -1,0 +1,25 @@
+﻿namespace Content.Shared.AirTank;
+
+/// <summary>
+///     Specifies what a certain tank APPEARS to contain. Is not actually what it contains.
+/// </summary>
+[RegisterComponent]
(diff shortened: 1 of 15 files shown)


Dragonjspider:
Nevermind, read it again, it does clarify, my bad

Cojoke-dot:
This seems nice for new players I guess, but this also seems a little uh… too much? There is already a warning if you're not getting the gas you need, a double check based on the sprite of the tank feels excessive

K-Dynamic:
Message is rather long, I'd shorten it to: "Breath from the X tank that may asphyxiate you?" or "poison you?"

beck-thompson:
> This seems nice for new players I guess, but this also seems a little uh… too much? There is already a warning if you're not getting the gas you need, a double check based on the sprite of the tank feels excessive

When I was new it took me hours to realize that as a slime I couldn't breath from oxygen tanks - I don't think the current messaging is enough

notquitehadouken:
don't merge yet i'm doing thing...

perryprog:
Spitballing, but what about a message on the examine text of the tank itself?

beck-thompson:
> Spitballing, but what about a message on the examine text of the tank itself?

I think that would be nice, but I worry that new players will just never see it. They are very clueless :laughing:

notquitehadouken:
ok there we go it still works and is less ass ayayayayayaya

Errant-4:
I think it might make more sense for this to just be on the Examine?

archee1:
A feature like this feels like it would only be useful for vox getting given O2 poisoned N2 tanks, but since its only going off the style of canister it just feels a little bloaty (the prior also being very powercreep'y, gas analyser exists), as others said it would fit better in the examine text similar how we have contraband tags that tell you if something is legal for you, if a gas canister is typical for you.

The message it warns is also really long, unless you're ready to read it most people are gonna miss what it says in its entirety especially beginner players that a feature like this is intended for, it may be better as something more brief like "this isn't typically what I breath from". I also cant think of once when i've actually used an N2 canister by accident, though I can think of times where I've intentionally used one solely to prevent breathing frezon/nitrous oxide.

Linkbro1:
if i may comment, i do think that something more lowkey, exactly like the contraband tooltip, would be good.
big red text in the examine of "this style of tank is usually used for plasma gas, which you cannot breathe." or somesuch?
no actual insight into what actually is in the tank, to allow fun poisoning shenanigans, but giving new players a little tip of "oh, apparently my character knows he breathes red not blue"

notquitehadouken:
i'll add the text to examine

notquitehadouken:
done : )

notquitehadouken:
also for reference this is supposed to be for new players

notquitehadouken:
...

beck-thompson:
I still think this PR was a good idea - I don't think the arguments against it are worth hurting new players

notquitehadouken:
i'd revive but lazieye lol

PR #38499, opened 2025-06-22
Add subscription event for ion storms (minor refactor)

mqole:
## About the PR
Previously ion storms were handled with an `EntityQueryEnumerator<SiliconLawBoundComponent, TransformComponent, IonStormTargetComponent>` which called the relevant method in `IonStormSystem` to give borgs a corrupted law. This is a minor refactor to swap that method over to an event-based subscription.

## Why / Balance
Mostly just futureproofing in case anyone wants ion storms to affect non-silicon entities- for example, computers, artifacts, whatever else.

## Technical details
- Creates a new ByRef event, `IonStormEvent`, that is raised on entities with `IonStormTargetComponent` attached to a station that experiences an ion storm.
- Modifies methods in 'IonStormSystem' and 'StartIonStormedSystem' to use events.
- Cleaned up some loose `using`s.

## Requirements

## Breaking changes
Ion storms now function by using `IonStormEvent` subscriptions rather than calling specific methods.

**Changelog**
no cl, not player-facing. yippee!

### DIFF
--- a/Content.Server/Silicons/Laws/StartIonStormedSystem.cs
+++ b/Content.Server/Silicons/Laws/StartIonStormedSystem.cs
@@ -1,3 +1,4 @@
+using Content.Server.StationEvents.Events;
 using Content.Shared.Silicons.Laws.Components;
 using Content.Shared.Administration.Logs;
 using Content.Shared.Database;
@@ -9,7 +10,6 @@ namespace Content.Server.Silicons.Laws;
 /// </summary>
 public sealed class StartIonStormedSystem : EntitySystem
 {
-    [Dependency] private readonly IonStormSystem _ionStorm = default!;
     [Dependency] private readonly ISharedAdminLogManager _adminLogger = default!;
     [Dependency] private readonly SiliconLawSystem _siliconLaw = default!;
 
@@ -28,7 +28,8 @@ public sealed class StartIonStormedSystem : EntitySystem
 
         for (int currentIonStorm = 0; currentIonStorm < ent.Comp.IonStormAmount; currentIonStorm++)
         {
-            _ionStorm.IonStormTarget((ent.Owner, lawBound, target), false);
+            var ev = new IonStormEvent(false);
+            RaiseLocalEvent(ent, ref ev);
         }
 
         var laws = _siliconLaw.GetLaws(ent.Owner, lawBound);
--- a/Content.Server/StationEvents/Events/IonStormRule.cs
+++ b/Content.Server/StationEvents/Events/IonStormRule.cs
@@ -1,4 +1,3 @@
-using Content.Server.Silicons.Laws;
 using Content.Server.StationEvents.Components;
 using Content.Shared.GameTicking.Components;
 using Content.Shared.Silicons.Laws.Components;
@@ -8,8 +7,6 @@ namespace Content.Server.StationEvents.Events;
 
 public sealed class IonStormRule : StationEventSystem<IonStormRuleComponent>
 {
-    [Dependency] private readonly IonStormSystem _ionStorm = default!;
-
     protected override void Started(EntityUid uid, IonStormRuleComponent comp, GameRuleComponent gameRule, GameRuleStartedEvent args)
     {
         base.Started(uid, comp, gameRule, args);
@@ -17,14 +14,20 @@ public sealed class IonStormRule : StationEventSystem<IonStormRuleComponent>
         if (!TryGetRandomStation(out var chosenStation))
             return;
 
-        var query = EntityQueryEnumerator<SiliconLawBoundComponent, TransformComponent, IonStormTargetComponent>();
-        while (query.MoveNext(out var ent, out var lawBound, out var xform, out var target))
+        var query = EntityQueryEnumerator<TransformComponent, IonStormTargetComponent>();
+        while (query.MoveNext(out var ent, out var xform, out _))
         {
             // only affect law holders on the station
             if (CompOrNull<StationMemberComponent>(xform.GridUid)?.Station != chosenStation)
                 continue;
-
-            _ionStorm.IonStormTarget((ent, lawBound, target));
+            var ev = new IonStormEvent();
+            RaiseLocalEvent(ent, ref ev);
         }
     }
 }
+
+/// <summary>
+/// Event raised on an entity with <see cref="IonStormTargetComponent"/> when an ion storm occurs on the attached station.
+/// </summary>
+[ByRefEvent]
+public record struct IonStormEvent(bool Adminlog = true);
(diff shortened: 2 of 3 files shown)


Content.Server/StationEvents/Events/IonStormRule.cs
```diff
@@ -8,23 +7,27 @@ namespace Content.Server.StationEvents.Events;
...
     protected override void Started(EntityUid uid, IonStormRuleComponent comp, GameRuleComponent gameRule, GameRuleStartedEvent args)
     {
         base.Started(uid, comp, gameRule, args);
 
         if (!TryGetRandomStation(out var chosenStation))
             return;
 
-        var query = EntityQueryEnumerator<SiliconLawBoundComponent, TransformComponent, IonStormTargetComponent>();
-        while (query.MoveNext(out var ent, out var lawBound, out var xform, out var target))
+        var query = EntityQueryEnumerator<TransformComponent, IonStormTargetComponent>();
```
perryprog: This should have IonStormTargetComponent first in the query; otherwise this iterates over each entity checking every one for IonStormTarget, rather than just iterating over all IonStormTargets.

mqole:
noticed that i missed migrating over the 0.8% check, fixed that

PJB3005 (changes requested):
This doesn't make sense. `IonStormTargetComponent` is already specific to silicons and your PR doesn't fix this:
image

mqole:
> This doesn't make sense. `IonStormTargetComponent` is already specific to silicons and your PR doesn't fix this: image

My original goal was to allow `IonStormTargetComponent` to function as a generic component that would allow any entity to catch `IonStormEvent`, I didn't catch that there were some datafields in there exclusive to silicons (oops)

Would moving those datafields over to `SiliconLawBoundComponent` be an appropriate fix?

PJB3005:
Honestly, I'm wary of making any breaking changes like this without a clear goal in mind.

mqole:
To give additional context I made changes to this system for a [downstream fork](https://github.com/impstation/imp-station-14/pull/2850) with a non-silicon species that had specific ion storm interactions. Ion storms are a cool mechanic that feel a bit underutilized (especially with all the recent development on AI) and I mostly just didn't want anyone else who wanted to use them to have to work around the existing code. If it's not what upstream wants though I totally understand.

PJB3005:
> To give additional context I made changes to this system for a [downstream fork](https://github.com/impstation/imp-station-14/pull/2850) with a non-silicon species that had specific ion storm interactions. Ion storms are a cool mechanic that feel a bit underutilized (especially with all the recent development on AI) and I mostly just didn't want anyone else who wanted to use them to have to work around the existing code. If it's not what upstream wants though I totally understand.

That's fine, then. Go ahead and split it off properly.

slarticodefast:
Hmm, having these datafield in the `SiliconLawBoundComponent` does not seem ideal to me either.
Because without additionally having the `IonStormTargetComponent` they would be redundant and do nothing. Not all lawbound silicons are ion storm targets.

mqole:
yeah, im not 100% sure what the solution is there. in an ideal world you'd probably want to account for
- silicons with laws who can be ion stormed
- silicons with laws who cannot be ion stormed
- unrelated entities who can be ion stormed

it might be redundant, but i'm hesitant to make an entirely new component just for lawbound ion storm silicons. from a yaml warrior perspective i feel like it would be more confusing to have two components that you need to add to make a silicon with laws susceptible to ion storms. i cant think of a good solution to that, unfortunately.

fully prepared to be wrong here though, lmk if youd like those datafields moved to a new component- super easy to do, either way o7

PJB3005:
> it might be redundant, but i'm hesitant to make an entirely new component just for lawbound ion storm silicons.

This is the correct solution.

Content.Server/Silicons/Laws/StartIonStormedSystem.cs
```diff
@@ -28,7 +27,8 @@ private void OnMapInit(Entity<StartIonStormedComponent> ent, ref MapInitEvent ar
 
         for (int currentIonStorm = 0; currentIonStorm < ent.Comp.IonStormAmount; currentIonStorm++)
         {
-            _ionStorm.IonStormTarget((ent.Owner, lawBound, target), false);
+            var ev = new IonStormEvent(false);
```
slarticodefast, line 30: This should still be an API function rather than a method event. And please keep the `IonStormTarget` API to prevent breaking changes. So don't turn it into a subscription method (those should be private).
See our code conventions
https://docs.spacestation14.com/en/general-development/codebase-info/conventions.html#method-events-vs-entity-system-methods

Content.Server/StationEvents/Events/IonStormRule.cs
```diff
@@ -17,14 +17,15 @@ protected override void Started(EntityUid uid, IonStormRuleComponent comp, GameR
...
         {
-            // only affect law holders on the station
-            if (CompOrNull<StationMemberComponent>(xform.GridUid)?.Station != chosenStation)
+            // only affect entities on the station, and check random chance
+            if (CompOrNull<StationMemberComponent>(xform.GridUid)?.Station != chosenStation ||
+                !_random.Prob(target.Chance))
                 continue;
-
-            _ionStorm.IonStormTarget((ent, lawBound, target));
+            var ev = new IonStormEvent();
```
slarticodefast, line 27: same

slarticodefast (changes requested):

widgetbeck:
> Honestly, I'm wary of making any breaking changes like this without a clear goal in mind.

I reckon anything that makes old, hardcoded systems easier to work with and add new functionality to is a good thing even if it's not immediately implemented. I run into a lot of stuff that I find myself wishing was just a *liiiiittle* more adaptable. 
The immediate thing that springs to mind with this PR in particular is that it would make it possible to easily have machines like vendors and consoles respond to ion storms, which is a really neat gameplay space I've wanted to explore.

mqole:
@slarticodefast youll have to forgive me for needing a bit of an assist here, i'm not super proficient in c#:

- will keep existing API, thanks for pointing it out. should the deprecated tag be added to it in this case?
- the purpose of making the lines you pointed out ByRefEvents is to allow for flexibility in which method is called, eg, perhaps you want a silicon to run a method called `ChangeLaws` but you'd want a computer console to run `MakeIonAnnouncement`. i'm not sure that theres another way to do this without using ByRefEvent.

if im missing or misunderstanding something about the code conventions please let me know. i'll hold off on making any edits until i know exactly what i should be doing here ^^;

slarticodefast:
> will keep existing API, thanks for pointing it out. should the deprecated tag be added to it in this case?

No, the API method should stay. According to our code conventions method events should only be used when wrapped by an API method.

> the purpose of making the lines you pointed out ByRefEvents is to allow for flexibility in which method is called, eg, perhaps you want a silicon to run a method called ChangeLaws but you'd want a computer console to run MakeIonAnnouncement. i'm not sure that theres another way to do this without using ByRefEvent.

Look at the example in the code convention. You should not be raising a method event from another system.

slarticodefast:
Leave a comment when you come back to this and I will re-open the PR.

ZigTag:
After a discussion with mqole, PR is being continued in #43302

PR #38511, opened 2025-06-22
Handle inventory template updating

ScarKy0:
## About the PR

Handles updating inventory templates when the template ID changes.

## Why / Balance

Needed for #34002 cause it updates inventories.

## Technical details

Manually networking InventoryComponent.
InventorySlotsComponent now get properly updated when the inventory template gets changed.

## Media

https://github.com/user-attachments/assets/8e7c7257-8c3c-4123-8f71-87a93abf8614

## Requirements

## Breaking changes

InventoryComponent is now networked manually. Anything using AfterAutoHandleStateEvent with the InventoryComponent will need to be adjusted accordingly.

**Changelog**

I wish it was player facing, yet here we are.


Content.Shared/Inventory/InventoryComponent.cs
```diff
@@ -1,22 +1,26 @@
 ﻿using Content.Shared.DisplacementMap;
+using Content.Shared.Humanoid.Prototypes;
```
Fildrance: unused using

Content.Shared/Inventory/InventoryComponent.cs
```diff
@@ -35,6 +39,19 @@ public sealed partial class InventoryComponent : Component
     public Dictionary<string, DisplacementData> MaleDisplacements = new();
 }
 
+[Serializable, NetSerializable]
+public sealed class InventoryComponentState : ComponentState
+{
+    public ProtoId<InventoryTemplatePrototype> Template;
+    public string? Species;
+
+    public InventoryComponentState(ProtoId<InventoryTemplatePrototype> template, string? species)
```
Fildrance: would probably be better if you respected original variable names :< even if template have protoId in signature so its obvious its an ID, species does not.

Content.Shared/Inventory/InventorySystem.Slots.cs
```diff
@@ -71,22 +79,35 @@ protected virtual void OnInit(EntityUid uid, InventoryComponent component, Compo
...
             return;
 
-        if (!_prototypeManager.TryIndex(ent.Comp.TemplateId, out InventoryTemplatePrototype? invTemplate))
+        if (!_prototypeManager.TryIndex(ent.Comp.TemplateId, out var invTemplate))
             return;
 
-        DebugTools.Assert(ent.Comp.Slots.Length == invTemplate.Slots.Length);
+        foreach (var container in ent.Comp.Containers)
+        {
+            if (invTemplate.Slots.Any(s => s.Name == container.ID))
```
Fildrance, line 94: To pass 'container' variable it allocates using 'closure' mechanism, and also allocates for the lambda itself... you could unwrap this into manual loop, or could use some HashSet with pre-set comparer, which is much more complex, as this thing created as prototype field...
or you could create custom IEqualityComparer (store static readonly instance of it in class) and use it in Contains overload that takes it. 
Overall not too critical - as this is not high traffic area

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -41,13 +46,73 @@ public override void Initialize()
...
+            if (!_proto.TryIndex(ent.Comp.TemplateId, out var index))
+                return;
+
+            ent.Comp.Slots = index.Slots;
+            ent.Comp.Containers = new ContainerSlot[ent.Comp.Slots.Length];
+
+            // Remove any slots from the inventory slots that arent in the new updated template.
+            foreach (var slot in slots.SlotData)
+            {
+                if (ent.Comp.Slots.Any(s => s.Name == slot.Key))
```
Fildrance: same for linq stuff, same for rare usage /shrug

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -41,13 +46,73 @@ public override void Initialize()
...
+                TryRemoveSlotDef(ent, slots, slot.Value);
+            }
+
+            // Add new slots that are missing from the new template
+            foreach (var slot in ent.Comp.Slots)
+            {
+                if (slots.SlotData.Any(s => s.Key == slot.Name))
+                    continue;
+
+                TryAddSlotDef(ent, slots, slot);
```
Fildrance: Slot info is stored in dictionary and there is no detailed 'position' data there, so stuff like arachnid 'extra pockets' just gets placed at the end of the list. But not because they are stored in dictionary, more like because adding slot to UI is done via events that does not have any 'position' too - it have only add/remove events and is doin remove by name and adds in the end of block where slot button supposed to go in. SlotGroup does not contain details, just container where button should go. I suppose its important to do some design before diggin into THAT hole - so maybe we should just let it be?

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -41,13 +46,73 @@ public override void Initialize()
...
+            if (!TryComp<InventorySlotsComponent>(ent, out var slots))
+                return;
+
+            if (!_proto.TryIndex(ent.Comp.TemplateId, out var index))
+                return;
+
+            ent.Comp.Slots = index.Slots;
+            ent.Comp.Containers = new ContainerSlot[ent.Comp.Slots.Length];
+
+            // Remove any slots from the inventory slots that arent in the new updated template.
```
Fildrance: ```Remove any slots that are not in new template```

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -41,13 +46,73 @@ public override void Initialize()
...
+
+            UpdateInventoryTemplate(ent);
+        }
+
+        private void UpdateInventoryTemplate(Entity<InventoryComponent> ent)
+        {
+            if (!TryComp<InventorySlotsComponent>(ent, out var slots))
+                return;
+
+            if (!_proto.TryIndex(ent.Comp.TemplateId, out var index))
```
Fildrance: its not index, its inventoryTemplate

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -41,13 +46,73 @@ public override void Initialize()
...
+            // Add new slots that are missing from the new template
+            foreach (var slot in ent.Comp.Slots)
+            {
+                if (slots.SlotData.Any(s => s.Key == slot.Name))
+                    continue;
+
+                TryAddSlotDef(ent, slots, slot);
+            }
+
+            // As the contaienrs got cleared, we need to reassign them to ensure they're all in the same order as the slots.
```
Fildrance: ``` As cotnai*ne*rs```

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -189,9 +250,47 @@ public bool TryAddSlotDef(EntityUid owner, InventorySlotsComponent component, Sl
...
+        public bool TryRemoveSlotDef(EntityUid owner, InventorySlotsComponent component, SlotDefinition newSlotDef)
+        {
+            SlotData newSlotData = newSlotDef; //convert to slotData
+            if (!component.SlotData.Remove(newSlotDef.Name))
+            {
+                return false;
+            }
+
+            if (owner == _playerManager.LocalEntity)
+                OnSlotRemoved?.Invoke(newSlotData);
```
Fildrance, line 241: empty line after if block

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -189,9 +250,47 @@ public bool TryAddSlotDef(EntityUid owner, InventorySlotsComponent component, Sl
...
                 OnSlotAdded?.Invoke(newSlotData);
+
             return true;
         }
 
+        public bool TryRemoveSlotDef(EntityUid owner, InventorySlotsComponent component, SlotDefinition newSlotDef)
+        {
+            SlotData newSlotData = newSlotDef; //convert to slotData
+            if (!component.SlotData.Remove(newSlotDef.Name))
+            {
```
Fildrance: braces are redundant, make their usage consistent :>

Content.Client/Inventory/ClientInventorySystem.cs
```diff
@@ -189,9 +250,47 @@ public bool TryAddSlotDef(EntityUid owner, InventorySlotsComponent component, Sl
...
+            return true;
+        }
+
+        /// <summary>
+        /// Updates the containers in InventorySlotsComponent according to their slots.
+        /// </summary>
+        /// <param name="ent">Entity to update.</param>
+        public void UpdateContainerData(Entity<InventorySlotsComponent> ent)
+        {
+            if (TryGetSlots(ent, out var definitions))
```
Fildrance: ehh, kinda up to you but this might be early return :3

Fildrance (changes requested):
Considering button re-creation on inv template change - making this work properly (not re-create whole thing), the best bet would probably be to revamp inventory group controls and item slots data to contain 'preferred index' or 'priority'. I am not sure how critical that thing is currently, and if we should change it in this PR, but that will make PR harder to look at for sure.

ScarKy0:
Superceded by #39246

PR #38516, opened 2025-06-22
Fixup Sericulture to be clonable

poklj:
## About the PR

Required for #34002 for transforming into an arachnid, Add a default entProtoId and add a OnClone event to ensure that any changes done to the prototype are cloned correctly.
## Why / Balance

Does not affect gamebalance
## Technical details

Added an EntProtoId default, Added a Subscription to Clone Event as Sericulture is an ActionComponent and thus cannot be Deepcopied via CopyComps (there is an issue with copycomps on components with referenced entityUid's being copied, Like Screaming and Wagging, thus requiring facilitation by component itself to be copied to a given target)
## Media

No media, no ingame effects yet
## Requirements

## Breaking changes

**Changelog**

No CL No Fun

### DIFF
--- a/Content.Shared/Sericulture/SericultureComponent.cs
+++ b/Content.Shared/Sericulture/SericultureComponent.cs
@@ -34,7 +34,7 @@ public sealed partial class SericultureComponent : Component
     [DataField(required: true)]
     [ViewVariables(VVAccess.ReadWrite)]
     [AutoNetworkedField]
-    public EntProtoId Action;
+    public EntProtoId Action = "ActionSericulture";
 
     [AutoNetworkedField]
     [DataField("actionEntity")]
--- a/Content.Shared/Sericulture/SericultureComponent.cs
+++ b/Content.Shared/Sericulture/SericultureComponent.cs
@@ -31,7 +31,7 @@ public sealed partial class SericultureComponent : Component
     /// <summary>
     /// The entity needed to actually preform sericulture. This will be granted (and removed) upon the entity's creation.
     /// </summary>
-    [DataField(required: true)]
+    [DataField]
     [ViewVariables(VVAccess.ReadWrite)]
     [AutoNetworkedField]
     public EntProtoId Action = "ActionSericulture";
--- a/Content.Shared/Sericulture/SericultureSystem.cs
+++ b/Content.Shared/Sericulture/SericultureSystem.cs
@@ -38,12 +38,16 @@ public abstract partial class SharedSericultureSystem : EntitySystem
 
     private void OnClone(Entity<SericultureComponent> ent, ref CloningEvent args)
     {
+        if(!args.Settings.EventComponents.Contains("Sericulture"))
+            return;
+
         var comp = EnsureComp<SericultureComponent>(args.CloneUid);
         comp.PopupText = ent.Comp.PopupText;
         comp.ProductionLength = ent.Comp.ProductionLength;
         comp.HungerCost = ent.Comp.HungerCost;
         comp.EntityProduced = ent.Comp.EntityProduced;
         comp.MinHungerThreshold = ent.Comp.MinHungerThreshold;
+        Dirty<SericultureComponent>((args.CloneUid, comp));
     }
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


Content.Shared/Sericulture/SericultureComponent.cs
```diff
@@ -34,7 +34,7 @@ public sealed partial class SericultureComponent : Component
     [DataField(required: true)]
     [ViewVariables(VVAccess.ReadWrite)]
     [AutoNetworkedField]
-    public EntProtoId Action;
+    public EntProtoId Action = "ActionSericulture";
```
slarticodefast, line 37: remove the `Required: true`

Content.Shared/Sericulture/SericultureSystem.cs
```diff
@@ -32,6 +33,17 @@ public override void Initialize()
         SubscribeLocalEvent<SericultureComponent, ComponentShutdown>(OnCompRemove);
         SubscribeLocalEvent<SericultureComponent, SericultureActionEvent>(OnSericultureStart);
         SubscribeLocalEvent<SericultureComponent, SericultureDoAfterEvent>(OnSericultureDoAfter);
+        SubscribeLocalEvent<SericultureComponent, CloningEvent>(OnClone);
+    }
+
+    private void OnClone(Entity<SericultureComponent> ent, ref CloningEvent args)
```
slarticodefast, line 39: we only want to clone this component if it is whitelisted in the `CloningSettingsPrototype given in the event`

Content.Shared/Sericulture/SericultureSystem.cs
```diff
@@ -32,6 +33,17 @@ public override void Initialize()
...
+    {
+        var comp = EnsureComp<SericultureComponent>(args.CloneUid);
+        comp.PopupText = ent.Comp.PopupText;
+        comp.ProductionLength = ent.Comp.ProductionLength;
+        comp.HungerCost = ent.Comp.HungerCost;
+        comp.EntityProduced = ent.Comp.EntityProduced;
+        comp.MinHungerThreshold = ent.Comp.MinHungerThreshold;
```
slarticodefast, line 49: need to be dirtied

Content.Shared/Sericulture/SericultureSystem.cs
```diff
@@ -32,6 +33,21 @@ public override void Initialize()
         SubscribeLocalEvent<SericultureComponent, ComponentShutdown>(OnCompRemove);
         SubscribeLocalEvent<SericultureComponent, SericultureActionEvent>(OnSericultureStart);
         SubscribeLocalEvent<SericultureComponent, SericultureDoAfterEvent>(OnSericultureDoAfter);
+        SubscribeLocalEvent<SericultureComponent, CloningEvent>(OnClone);
+    }
+
+    private void OnClone(Entity<SericultureComponent> ent, ref CloningEvent args)
+    {
+        if(!args.Settings.EventComponents.Contains("Sericulture"))
```
slarticodefast: Hmm, can you use something like `nameof` here rather than the component name as a string? That should be more future proof.

Content.Shared/Sericulture/SericultureSystem.cs
```diff
@@ -32,6 +33,21 @@ public override void Initialize()
...
+            return;
+
+        var comp = EnsureComp<SericultureComponent>(args.CloneUid);
+        comp.PopupText = ent.Comp.PopupText;
+        comp.ProductionLength = ent.Comp.ProductionLength;
+        comp.HungerCost = ent.Comp.HungerCost;
+        comp.EntityProduced = ent.Comp.EntityProduced;
+        comp.MinHungerThreshold = ent.Comp.MinHungerThreshold;
+        Dirty<SericultureComponent>((args.CloneUid, comp));
```
slarticodefast:
```suggestion
        Dirty(args.CloneUid, comp);
```

Content.Shared/Sericulture/SericultureSystem.cs
poklj: I worked this out last night the method to get the YAML name is via:
`Factory.GetRegistration(ent.Comp.GetType()).Name)`,

issue with `nameof()` is that it return the FullName which includes the "Component" bit, which is not contained in the list

slarticodefast (approved):

PR #38525, opened 2025-06-23
[ADMIN] Added admin log to whenever station's alert level is changed.

Velken:
## About the PR

Added admin log to the SetLevel function.

## Why / Balance

~~If an entity had the component AlertLevelChangeOnTrigger, someone could use it to change the station's alert and admins would have a hard time to find the entity. Now there is a log with the entity uid and in which station, for servres running more than one station at once.~~
Whenever something/someone changed the alert level of a station, there was no admin logs for such, now there are.

## Technical details

Added new optional parameter to the AlertLevelSytem.SetLevel() function.
Returns UNKNOWN if not given an uid (i.e.: console command)

## Media

N/A 

## Requirements

## Breaking changes

none

**Changelog**

:cl:
ADMIN
- add: Changes to a station's alert level are now logged, find via "Non-Players" with "Medium" impact. Shows the UID of the entity that changed the level if one is available.


slarticodefast:
Wouldn't it be better to have this be part of SetAlertLevel instead?

Velken:
> Wouldn't it be better to have this be part of SetAlertLevel instead?

sounds good, will update the PR to that

Velken:
added toprettystring
UNKNOWN is from there is no entity (i.e.: console command)
image

Content.Server/AlertLevel/AlertLevelSystem.cs
```diff
@@ -207,12 +210,23 @@ public void SetLevel(EntityUid station, string level, bool playSound, bool annou
...
+        string callerString = caller.ToString() ?? "NULL";
+        if (caller.HasValue == false)
+        {
+            callerString = "UNKNOWN";
+        }
+        else
+        {
+            callerString = ToPrettyString(caller);
+        }
+        _adminLog.Add(Shared.Database.LogType.Trigger, Shared.Database.LogImpact.Medium, $"{callerString} changed alert level to {level} on station {station}.");
```
slarticodefast, line 224: Just do two separate logs for the case that the caller is null instead of using `UNKNOWN`.
For example
`$"{ToPrettyString(user):player} changed alert level to {level} on station {station}.");`
`$"Changed alert level to {level} on station {station}.");`
Also the this should be another log type than trigger.

Content.Server/AlertLevel/AlertLevelSystem.cs
```diff
@@ -139,8 +141,9 @@ public string GetDefaultLevel(Entity<AlertLevelComponent?> station)
     /// <param name="announce">Say the alert level's announcement.</param>
     /// <param name="force">Force the alert change. This applies if the alert level is not selectable or not.</param>
     /// <param name="locked">Will it be possible to change level by crew.</param>
+    /// <param name="caller">Entity that called this method, for logging purposes.</param>
     public void SetLevel(EntityUid station, string level, bool playSound, bool announce, bool force = false,
-        bool locked = false, MetaDataComponent? dataComponent = null, AlertLevelComponent? component = null)
+        bool locked = false, EntityUid? caller = null, MetaDataComponent? dataComponent = null, AlertLevelComponent? component = null)
```
slarticodefast, line 146:
```suggestion
        bool locked = false, EntityUid? user= null, MetaDataComponent? dataComponent = null, AlertLevelComponent? component = null)
```
this should be the player changing the alert level, not the device

Velken: > this should be the player changing the alert level, not the device

how would I even get the player from something, like, an entity that has AlertLevelChangeOnTrigger that got triggered by a TriggeOnSignal?

i think it would be easier to list what entity caused the alert level to change, so an admin can find it if needed.

PR #38528, opened 2025-06-23
Randomized Markings (take two)

mqole:
### REQUIRES https://github.com/space-wizards/RobustToolbox/pull/6720

Second attempt at #34765 / #38345 with @widgetbeck's permission and some additional tweaks.

Ensures that randomly generated characters will have markings, and prettier-looking colours.

On the dev side, this also allows markings/species authors to fine-tune random generation of markings on a species, determining the chance of rolling markings on individual categories, or even making some markings rarer than others.

I've given humans, voxes, slimes, vulps, and species that use `TintedHues` some parameters to generate what i consider nicer-looking characters by default. These can all be altered as desired.

I've also changed some of the default color parameters in various species markings. These changes are very minorly player-facing as they just change the default colour of the new marking when it's added.

Also ALSO added shader support for character customizable markings. This means that you have the option to make glow in the dark markings, among other things.

ports of:
- https://github.com/impstation/imp-station-14/pull/1600
- https://github.com/impstation/imp-station-14/pull/2148
- https://github.com/impstation/imp-station-14/pull/2149
- https://github.com/impstation/imp-station-14/pull/842
- https://github.com/impstation/imp-station-14/pull/2855

**NUBODY UPDATE**:  _because of the body system overhaul, this was stuck in dev hell for a while as i iterated on it downstream. it's now better described as a port of:_
- https://github.com/syndicate-ss14/macrocosm/pull/2
- https://github.com/syndicate-ss14/macrocosm/pull/23

## Why / Balance
From the original PR:

> Randomized characters have always felt very lacking, because they were - All species without hair would generate looking identical and flat. This fixes that.

Most of the additions to this PR that werent in #34765 are QoL tweaks implemented after a period of playtesting the initial PR downstream, and have been pretty popular.

Marking shaders are another addition that have been very popular on our fork and have seen a lot of use on various downstreams.

MARKING WEIGHTS: okay, bear with me here. These were initially added downstream because the authors of species wanted to include rare markings that had a small chance to roll on ghostrole species (as an example, the cat ears). But they also found another use in guaranteeing that certain species don't have a 60% chance to wind up with the default markings. https://github.com/impstation/imp-station-14/pull/2855 has an example of this for the downstream Kodepiia species. check it out

### Shaders:

`MarkingPrototype` now has an additional 'shaders' datafield, which takes a dictionary of two strings: state, and shader name.
If this field is undefined, markings will use the default shader.

Shaders are modified in `ApplyMarkings`. Unfortunately shader layer methods arent in the system, so we had to grab `SpriteComponent`. This isnt ideal but I'm not fucking with robust toolbox. In a beautiful world shaders would be attached to sprites directly anyway

### Markings:

In addition to modifying `HumanoidCharacterAppearance`, Ive added a new partial class `HumanoidCharacterAppearance.Random` which contains the bulk of helper methods involved in handing random markings for a layer, random color generation, etc. All methods have been commented, but please let me know if more documentation is desired as i have removed a lot of beck's comments in the process of reworking this.

Anyway, there's lots going on under the hood here. At the moment when we generate a character, the palette will either be a split complimentary, a triadic complimentary, or a plain complimentary- either way, we get 3 colours, where the base colour used is assigned to skintone and the other 2 become eye and hair colours. When we add a marking to our doll, we will colour each layer of a marking by first checking to see if that layer has a defined `MarkingColoring`. If it does, we'll colour it using those rules. Otherwise, we'll pick either the hair or eye color at random.

Randomizer lock checks have been moved inside these palette methods.

There are also 2 new `SkinColorationPrototype` parameters. I elaborate on these more below in the breaking changes, but these parameters serve as replacements for upstream behavior that I commented out in the editing of this system.

We cycle through organs -> `HumanoidVisualLayers` in that organ, then add markings to the organ. Each `HumanoidVisualLayers` layer enters a loop where it first checks the layer weight to see if a marking should roll at all (would look pretty chaotic if we always rolled max markings), then gets a random marking from that category. This loop is repeated a number of times equal to the value that layer has in the species `MarkingLimits.Limit`. These markings are all eventually collected in a dictionary which is used to construct our final `HumanoidCharacterAppearance`.

> Only real quirks are:

> - Slime People have a tendency to look bad because their arm and leg gradients can roll without the corresponding hand/foot markings. I looked into fixing that, but determined it was out of scope for this PR.

Currently by default all marking categories will have a 60% chance of rolling, and all the individual markings are evenly weighted. Notable exception is that I've guaranteed humans, dwarves, voxes and slimes will always roll hair using their species yml files. Obviously this can easily be adjusted however.

### Why did you do it like this?

I have too many thoughts and feelings about randomization. apologies in advance

Hair and facial hair supposedly now function more similarly to other markings than they did pre-nubody, which is great- theyve always been a weird exception. (Apparently they are however still different enough to require their own generation rules. That's fine.) Fundamentally, not every species will have hair, the same way not every species will have, say, a tail. So it makes sense for nubody to update `HumanoidCharacterAppearance` and remove the requirement that every species generates a hair colour, since they aren't all using it!

The best way to assign markings colours is by using the archaic `MarkingColoring` system, as it has a lot of validation rules which come in handy when importing characters: if the file you import is slightly off, then instead of your character being imported with bright white markings, the markings will use a 'default' colour- your skin tone, the color of other markings, or even a darker variation of your skin tone (this is the tattoo default coloring!). All the values are taken from the colors in `HumanoidCharacterAppearance`.

You might be able to predict the trouble I ran into here by combining these two facts. The issue that comes with this is that we are (using the pre-nubody random system, which i did spend a long time iterating on) generating a palette with 3 tones: skin, hair, and eyes.

*However, since `HumanoidCharacterAppearance` no longer stores a hair colour, we are unable to use `MarkingColoring` to assign markings this colour, which means we are limited to using two colours in our randomly generated palette.*

THE SOLUTION: Fallback colours. `MarkingColoring` has three datafields - a nullable `Type`, a list `FallbackTypes`, and a color `FallbackColor`. What's supposed to happen is that those three types will get used in descending order, depending on availability- if a type is not provided, you use the fallback type, and if thats not there you use the fallback color. In upstream, however, the nullable `Type` is always set to `SkinColoring`, so these fallbacks rarely if ever get used.

What I've done is removed the default value from `Type`, and moved `SkinColoring` to `FallbackTypes` if that list is undefined. If the system cannot find a `Type`, it will randomly use either hair or eye color. If it does find a type, it will use that instead!

This solution is a bit unorthodox, but its making edits to a 3-year old system which I suspect has outlived its original purpose for marking coloration. It allows users to use yaml to customize what randomization looks like on a per-marking basis, which is my main goal with this PR!

### Shaders:
image

### Markings:

current randomization:

https://github.com/user-attachments/assets/1834ca04-d39e-4d45-9fa1-693144fb3649

new randomization:

https://github.com/user-attachments/assets/f824bd66-0dac-441f-8e47-c53108932fe0

https://github.com/user-attachments/assets/9627e40e-1552-4ac0-a5d6-a59b5c06d9ef

## Requirements

## Breaking changes
Markings now support shaders on individual sprite layers. Please see the associated PR for an example on how to implement this.

In addition, randomly generated characters will now roll markings! The rarity of markings can be altered per species. To modify the chance that a marking will randomly roll for an entire category, use `MarkingsGroupPrototype.Weight`. To modify the weight of an individual marking rolling over other markings in that category, use `MarkingPrototype.RandomWeight`.

Parameters in  `HumanoidCharacterAppearance.Random` have been modified from `(RandomizeCfg, HumanoidCharacterAppearance, ProtoId<SpeciesPrototype>, Sex)` to `(SpeciesPrototype, Sex, RandomizeCfg? = null, HumanoidCharacterAppearance? = null)`. The last two values are now optional, and if not passed in the method will apply randomization as if all `RandomizeCfg` parameters are unlocked.

Species `SkinColorationPrototypes` now have additional options `RealisticColors` and `SquashEyeHairColors`, which can be toggled on per prototype to generate more realistic colours. As an example, `RealisticColors` can be used for humans to generate only real-world hair and eye colours, while `SquashEyeHairColors` can be used to make vulpkanin fur tones more realistic (by making 'hair' fit the same coloration as 'skin').

`MarkingColoration` now has its default value for `Type` set to null, with `SkinColoring` being moved to `FallbackTypes`.

Some markings have also had their default coloring altered.

:cl: widgetbeck, mqole
- add: Randomized characters will now roll markings.
- add: Added a 'Lock Markings' button to the character randomizer settings.


hivehum:
https://github.com/user-attachments/assets/1380fb6c-581d-4806-bc7f-6d30002d39bb

demonstration i made for delta-v of various markings using the shader code. none of these markings are included in this PR, but they are _very_ popular in downstreams, and crocodilecarousel and i are interested in bringing them upstream if this is merged

mqole:
https://github.com/user-attachments/assets/a72e169b-bbad-48eb-96ba-40b8a44f9da0

media confirming this now works with skincoloration stuff :>
apologies for the delay, took a bit to figure this out lol

mqole:
heisentest?

sowelipililimute:
this is a PR i have interest in now that the humanoid appearance refactor is done if you're willing to pick up work on it

mqole:
i believe bodysystem is currently being overhauled with the intent to provide documentation on the new system when everything's fully done, so im more than happy to resolve conflicts when docs are done

sowelipililimute:
yeah it's done; freeze on these PRs is lifted

mqole:
waiting until #42840 is resolved to work on this

mqole:
https://github.com/user-attachments/assets/e6b7cc32-2d61-4c41-89f2-090c800369fb

FINALLY managed to get this working with nubody (phew!!!). still ironing out a few small things that i missed, but i have not abandoned this. very excited to be able to open up this pr again!

mqole:
if that isnt heisentest i'll eat my hat

mqole:
changes applied from https://github.com/Space-Wizards-Federation/space-station-14/pull/87

Content.Shared/Humanoid/SkinColorationPrototype.cs
```diff
@@ -20,6 +20,20 @@ public sealed partial class SkinColorationPrototype : IPrototype
...
+    ///     so markings don't appear too bright on darker skin.
+    /// </summary>
+    [DataField]
+    public bool RealisticColors;
+
+    /// <summary>
+    ///     If true, will also squash hair and eye colors to the coloration strategy.
+    /// </summary>
+    [DataField]
+    public bool SquashAllColors;
```
AJCM-git: I feel like the names of these fields should be more specific, as in mentioning the body parts setting these would affect/depend on somewhere in the name, as long as it isnt excruciatingly long of course

Content.Client/Body/VisualBodySystem.cs
```diff
@@ -207,6 +207,14 @@ private void ApplyMarkings(Entity<VisualOrganMarkingsComponent> ent, Entity<Spri
 
                 if (displacement != null && proto.CanBeDisplaced)
                     _displacement.TryAddDisplacement(displacement, (target, target.Comp), index + i + 1, layerId, out _);
+
+                if (proto.Shaders is not null &&
+                    proto.Shaders.TryGetValue(rsi.RsiState, out var shader))
+                {
+                    // TODO: fix this when LayerSetShader is moved out of component
+                    EnsureComp<SpriteComponent>(target, out var spriteComp);
```
AJCM-git: the resolve at the start already ensures this component is present in the target, just use the component

Content.Shared/Humanoid/Markings/MarkingPrototype.cs
```diff
@@ -36,9 +36,25 @@ public sealed partial class MarkingPrototype : IPrototype
         [DataField("sprites", required: true)]
         public List<SpriteSpecifier> Sprites { get; private set; } = default!;
 
+        /// <summary>
+        ///     Optional dictionary allowing assignment of shaders to sprite layers in a marking.
+        ///     This implementation is very messy but unfortunately Robust doesn't like shaders in SpriteSpecifiers.
+        /// </summary>
+        [DataField]
+        public Dictionary<string, string>? Shaders { get; private set; }
```
AJCM-git, line 44: could be wrong but i dont think any our markings use this? should at least add 1 to demostrate

Content.Shared/Humanoid/HumanoidCharacterAppearance.Markings.cs
```diff
@@ -0,0 +1,280 @@
...
+        if (markings.Count == 0)
+            return null;
+
+        var weights = markings.ToDictionary(m => m.Key, m => m.Value.RandomWeight);
+
+        return random.Pick(weights).Key;
+    }
+
+    #region Color Helpers
+    // TODO: These are probably better off in Robust.Shared.Maths.Color
```
AJCM-git: Yeah these are way too generic for content, cant really work around this. Ill try to get the engine PR in if you make it

AJCM-git (changes requested):
Be aware that ill merge https://github.com/space-wizards/space-station-14/pull/44363 first. Should add the locks to markings too if possible

Content.Shared/Humanoid/Markings/MarkingPrototype.cs
mqole: none do, see example media in the pr header. i dont really want to include new markings in this pr because that feels out of scope

AJCM-git: We try to not have these kinds of unused features with no in-game access to avoid bit-rotting (bar some edge cases).
How about separating the shader stuff to another PR? It can be something simple like only adding 1 marking with a shader and the shader-related changes in this PR, so like 4-5 files changed

mqole: see hivehum's comment above, there are active plans to PR some glow-in-the-dark markings provided this feature is approved. to be blunt this pr has been in review limbo for over a year and i really dont want it to get stalled by the additional caveat of marking art approval, i would rather just have all the technical stuff bundled together if possible

AJCM-git: From my point of view the 10ish lines of code you would have to move have no real purpose until there's an actual use-case so it really makes way more sense to put them on a new PR with some shader markings if its already planned, i would even approve that code in the new PR granted my review is addressed.

But then again its just a couple lines that have a working functionality that someone *may* use, so if we are at an impasse I'm perfectly willing to let this go on this case. So I'm marking this as a non-merge blocking suggestion rather than a review

mqole: forgot to mention as well that shader markings is tech that is currently used by a lot of downstreams, so having the support for it upstream would definitely not be purposeless. if nothing else consider it unifying the implementation

Content.Shared/Humanoid/SkinColorationPrototype.cs
mqole: i have no idea what alternative names to give them, open to suggestions. ive included some documentation in `skin_colorations.yml` which will hopefully make it clearer for our yaml warriors at least

mqole:
https://github.com/user-attachments/assets/3592e212-3d05-4c4b-8cfc-cd4b7b093f9d

works with locks

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
```diff
@@ -117,42 +117,71 @@ public static Color RandomSkin(ProtoId<SpeciesPrototype> species)
...
     /// <param name="baseAppearance">Appearance to base the new appearance on. Values that are not randomized will be taken from this appearance.</param>
     /// <param name="species">Species prototype ID.</param>
     /// <param name="sex">Sex.</param>
-    /// <returns>A new character appearance with selected values randomized</returns>
-    public static HumanoidCharacterAppearance Random(RandomizeCfg charEditorRandomizeConfig, HumanoidCharacterAppearance baseAppearance, ProtoId<SpeciesPrototype> species, Sex sex)
+    public static HumanoidCharacterAppearance Random(SpeciesPrototype species, Sex sex, RandomizeCfg? charEditorRandomizeConfig, HumanoidCharacterAppearance? baseAppearance)
     {
-        var appearance = new HumanoidCharacterAppearance { Markings = new () };
-        appearance.EyeColor = (charEditorRandomizeConfig & RandomizeCfg.Eyes) != 0 ? RandomEyes() : baseAppearance.EyeColor;
-        appearance.SkinColor = (charEditorRandomizeConfig & RandomizeCfg.Skin) != 0 ? RandomSkin(species) : baseAppearance.SkinColor;
```
AJCM-git, line 148: RandomEyes and RandomSkin are unused now, i recommend restructuring your code so it calls them in this method like before

Content.Shared/Humanoid/SkinColorationPrototype.cs
AJCM-git: Hmm how about LimitEyeHairColoring and LimitEyeHairStrategy or something like that, gives a better immediate idea of their purpose when you read the prototype

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
```diff
@@ -117,42 +117,71 @@ public static Color RandomSkin(ProtoId<SpeciesPrototype> species)
...
+
+        var baseColor = new Color(random.NextFloat(1), random.NextFloat(1), random.NextFloat(1), 1);
+        var colorPalette = GetPaletteFromBase(baseColor, random.Next(3));
+
+        var colorDict = ClampPaletteToStrategy(colorPalette, protoMan.Index(skinType), random, charEditorRandomizeConfig, baseAppearance);
+
+        var markingData = markingManager.GetMarkingData(species);
+        Dictionary<ProtoId<OrganCategoryPrototype>, Dictionary<HumanoidVisualLayers, List<Marking>>> newMarkings = [];
+
+        if ((charEditorRandomizeConfig & RandomizeCfg.Markings) != 0 || baseAppearance is null)
```
AJCM-git: move body to a RandomizeMarkings method

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
```diff
@@ -117,42 +117,71 @@ public static Color RandomSkin(ProtoId<SpeciesPrototype> species)
...
+    ///     When <see cref="RandomizeCfg"/> and an existing <see cref="HumanoidCharacterAppearance"> are passed in,
+    ///     values will be selectively randomized with the option to maintain existing values.
+    /// </remarks>
     /// <param name="charEditorRandomizeConfig">Which values to randomize.</param>
     /// <param name="baseAppearance">Appearance to base the new appearance on. Values that are not randomized will be taken from this appearance.</param>
     /// <param name="species">Species prototype ID.</param>
     /// <param name="sex">Sex.</param>
-    /// <returns>A new character appearance with selected values randomized</returns>
-    public static HumanoidCharacterAppearance Random(RandomizeCfg charEditorRandomizeConfig, HumanoidCharacterAppearance baseAppearance, ProtoId<SpeciesPrototype> species, Sex sex)
+    public static HumanoidCharacterAppearance Random(SpeciesPrototype species, Sex sex, RandomizeCfg? charEditorRandomizeConfig, HumanoidCharacterAppearance? baseAppearance)
```
AJCM-git: Param order should be the same as before

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
```diff
@@ -117,42 +117,71 @@ public static Color RandomSkin(ProtoId<SpeciesPrototype> species)
...
     {
-        var appearance = new HumanoidCharacterAppearance { Markings = new () };
-        appearance.EyeColor = (charEditorRandomizeConfig & RandomizeCfg.Eyes) != 0 ? RandomEyes() : baseAppearance.EyeColor;
-        appearance.SkinColor = (charEditorRandomizeConfig & RandomizeCfg.Skin) != 0 ? RandomSkin(species) : baseAppearance.SkinColor;
+        var random = IoCManager.Resolve<IRobustRandom>();
+        var markingManager = IoCManager.Resolve<MarkingManager>();
+        var protoMan = IoCManager.Resolve<IPrototypeManager>();
+
+        var skinType = protoMan.Index<SpeciesPrototype>(species).SkinColoration;
```
AJCM-git: you are passing the prototype directly now so this index in unnecessary,

Content.Shared/Humanoid/HumanoidCharacterAppearance.Markings.cs
```diff
@@ -0,0 +1,215 @@
...
+            newHairColor = skinType.Strategy.ClosestSkinColor(newHairColor);
+            newEyeColor = skinType.Strategy.ClosestSkinColor(newEyeColor);
+        }
+
+        return new Dictionary<string, Color>
+        {
+            { SkinColorKey, newSkinColor },
+            { HairColorKey, newHairColor },
+            { EyeColorKey, newEyeColor }
+        };
```
AJCM-git: It would be more preferable to split this into 3 functions getting each of these

Content.Shared/Humanoid/HumanoidCharacterAppearance.Markings.cs
```diff
@@ -0,0 +1,215 @@
...
+    private static Dictionary<string, Color> ClampPaletteToStrategy(Color[] colorPalette, SkinColorationPrototype skinType, IRobustRandom random, RandomizeCfg? charEditorRandomizeConfig, HumanoidCharacterAppearance? baseAppearance)
+    {
+        if (colorPalette.Length != 3)
+            throw new ArgumentException($"Palettes must have exactly 3 colours, palette contains {colorPalette.Length} colours");
+
+        var newSkinColor = (charEditorRandomizeConfig & RandomizeCfg.Skin) != 0 || baseAppearance is null
+            ? colorPalette[0] : baseAppearance.SkinColor;
+        var newHairColor = colorPalette[1];
+        var newEyeColor = (charEditorRandomizeConfig & RandomizeCfg.Eyes) != 0 || baseAppearance is null
+            ? colorPalette[2] : baseAppearance.EyeColor;
```
AJCM-git: ideally these would be in the main random method

Content.Shared/Humanoid/HumanoidCharacterAppearance.Markings.cs
```diff
@@ -0,0 +1,215 @@
...
+    ///     Optionally accepts <see cref="RandomizeCfg"/> and a base <see cref="HumanoidCharacterAppearance"/>
+    ///     to retain values of an existing appearance.
+    /// </remarks>
+    /// <returns>
+    ///     A 3-toned color palette with keys skinColor, hairColor, and eyeColor.
+    /// </returns>
+    private static Dictionary<string, Color> ClampPaletteToStrategy(Color[] colorPalette, SkinColorationPrototype skinType, IRobustRandom random, RandomizeCfg? charEditorRandomizeConfig, HumanoidCharacterAppearance? baseAppearance)
+    {
+        if (colorPalette.Length != 3)
+            throw new ArgumentException($"Palettes must have exactly 3 colours, palette contains {colorPalette.Length} colours");
```
AJCM-git: this wont be necessary if you split the function and take a single pallete per appearance part

AJCM-git (changes requested):
Everything but the HumanoidCharacterAppearance changes I approve of, there needs to be some code restructuring done in there

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
mqole: cannot do this as those methods are functionally incompatible with the procedure used to generate color palettes. i initially removed `RandomEyes` and `RandomSkin` entirely but left them in just in case to avoid breaking changes, if someone else wants to use them later down the track

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
mqole: since `charEditorRandomizeConfig` and `baseAppearance` are optional values i believe they have to be at the end no? i can include the reorder as a breaking change in the header if needed

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
mqole:
```cs
private static Dictionary<HumanoidVisualLayers, List<Marking>> RandomizeMarkings(OrganMarkingData organData, MarkingManager markingManager, MarkingsGroupPrototype groupProto, Sex sex, Dictionary<string, Color> colorDict, IRobustRandom random)
```

are you sure?? i agree that its good to keep code as neat as possible, but this is a shit ton of args to pass into what is essentially just a helper method

Content.Shared/Humanoid/HumanoidCharacterAppearance.Markings.cs
mqole: see above comment - colors for both instances of the lock (new random color and old color) will both need to be clamped to species color strategy. if i put them in the main random method it means clamping code will need to be run twice (once for generated palette, and once AGAIN for color replacements), which is redundant

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
AJCM-git: Actually that doesnt look too bad, just put the manager and random at the end of the call

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
AJCM-git: pretty sure they only have to go last if you assign them a default value so it isnt really necessary in this case, also is there any benefit to making RandomizeCfg nullable? had a quick look around and it didnt look like it

Content.Shared/Humanoid/HumanoidCharacterAppearance.Markings.cs
AJCM-git: I say its worth restructuring and doing that twice for now (even if it isnt ideal). Later we can see if we can optimize some things. If we cant  we can always revert to the old method but if the code ends up being way more straightforward as a result, the small perf hit would be worth it. For something like this it isnt that critical to over-optimize so we can trade-off if necessary

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
AJCM-git: hmmm having the old behavior around could be useful I guess,  ill consider it  as the PR develops

Content.Shared/Humanoid/HumanoidCharacterAppearance.cs
mqole: forgot to give them defaults oops
image

benefit for making `RandomizeCfg` nullable is to allow a humanoidappearance to be completely randomized without needing cfg settings, i was thinking this could be a nice method to use for something like DNA scrambler or changeling

AJCM-git (approved):

AJCM-git:
test fail real

portfiend:
"adding portfiend cologne"?

PR #38537, opened 2025-06-23
Adds BallisticAmmoSelfRefillerComponent

Centronias:
## About the PR

Adds BallisticAmmoSelfRefillerComponent, a component analogous to `BatterySelfRechargerComponent`, but which refills `BallisticAmmoProviderComponent`s instead of `BatteryComponent`s.
This PR also swaps Syndicate L6 and Viper modules to use this new component instead of batteries.

## Why / Balance

Personally, I found it weird that a battery just makes ammo; but the real reason is I need this to make a borg module which can refill a foam force rifle with foam darts in a way that can also be manually reloaded.

There is one balance change in that presently an EMP will delete charge (ie shots) and pause recharge on the battery-guns; while this change does only the latter with EMPs not affecting ammunition already in the gun at all.

### Automagic Ballistic Ammo Refilling
- Add `BallisticAmmoSelfRefillerComponent`
- Handle `EmpPulseEvent` to pause refilling behavior for EMP's duration

### Supporting Changes
- Change `Content.Server.Weapons.Ranged.Systems.Update` override in `GunSystem.AutoFire.cs` to `UpdateAutoFire`
- Add `Content.Server.Weapons.Ranged.Systems.Update` to `GunSystem.cs` so that it can call `UpdateAutoFire` and `UpdateBallistic`
- Add public methods to GunSystem for use by refilling implementation
  - PauseSelfRefill
  - IsFullBallistic (same as #38420)
  - CanInsertBallistic (same as #38420)
  - TryBallisticInsert (same as #38420)

## Media

https://github.com/user-attachments/assets/84956d00-d7e1-48ec-8d48-0fef4fdb35e6

## Requirements

## Breaking changes

No breakages, I think it's all additive.

**Changelog**

IMO no changelog since it's more of cosmetic than functional of a change.

### DIFF
--- a/Content.Server/Weapons/Ranged/Systems/GunSystem.Ballistic.cs
+++ b/Content.Server/Weapons/Ranged/Systems/GunSystem.Ballistic.cs
@@ -1,11 +1,22 @@
+using Content.Server.Emp;
 using Content.Shared.Weapons.Ranged.Components;
 using Content.Shared.Weapons.Ranged.Events;
 using Robust.Shared.Map;
+using Robust.Shared.Timing;
 
 namespace Content.Server.Weapons.Ranged.Systems;
 
 public sealed partial class GunSystem
 {
+    [Dependency] private readonly IGameTiming _timing = default!;
+
+    protected override void InitializeBallistic()
+    {
+        base.InitializeBallistic();
+
+        SubscribeLocalEvent<BallisticAmmoSelfRefillerComponent, EmpPulseEvent>(OnRefillerEmpPulsed);
+    }
+
     protected override void Cycle(EntityUid uid, BallisticAmmoProviderComponent component, MapCoordinates coordinates)
     {
         EntityUid? ent = null;
@@ -33,5 +44,37 @@ public sealed partial class GunSystem
 
         var cycledEvent = new GunCycledEvent();
         RaiseLocalEvent(uid, ref cycledEvent);
+    }
+
+    private void OnRefillerEmpPulsed(Entity<BallisticAmmoSelfRefillerComponent> entity, ref EmpPulseEvent args)
+    {
+        PauseSelfRefill(entity, args.Duration);
+    }
+
+    private void UpdateBallistic()
+    {
+        // Handle `BallisticAmmoSelfRefillerComponent` refills.
+        var query = EntityQueryEnumerator<BallisticAmmoSelfRefillerComponent, BallisticAmmoProviderComponent>();
+        while (query.MoveNext(out var uid, out var refiller, out var ammo))
+        {
+            var entity = new Entity<BallisticAmmoProviderComponent>(uid, ammo);
+            if (!refiller.AutoRefill ||
+                IsFull(entity) ||
+                _timing.CurTime < refiller.NextAutoRefill)
+                continue;
+
+            var ammoEntity = Spawn(refiller.AmmoProto);
+            var insertSucceeded = TryBallisticInsert(entity, ammoEntity, null, suppressInsertionSound: true);
+            if (!insertSucceeded)
+            {
+                QueueDel(ammoEntity);
+                Log.Error(
+                    $"Failed to insert ammo {ammoEntity} into non-full {entity}. Is the {nameof(BallisticAmmoSelfRefillerComponent)}'s {nameof(BallisticAmmoSelfRefillerComponent.AmmoProto)} incorrect for the {nameof(BallisticAmmoProviderComponent)}'s {nameof(BallisticAmmoProviderComponent.Whitelist)}?");
+                continue;
+            }
+
+            refiller.NextAutoRefill = _timing.CurTime + refiller.AutoRefillRate;
+            Dirty(uid, refiller);
+        }
     }
 }
--- a/Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
+++ b/Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
@@ -1,0 +1,50 @@
+﻿using Content.Shared.Weapons.Ranged.Systems;
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Weapons.Ranged.Components;
+
+/// <summary>
+/// This component, analogous to <c>BatterySelfRechargerComponent</c>, will attempt insert ballistic ammunition into
+/// its owner's <see cref="BallisticAmmoProviderComponent"/>.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, AutoGenerateComponentPause, Access(typeof(SharedGunSystem))]
+public sealed partial class BallisticAmmoSelfRefillerComponent : Component
+{
+    /// <summary>
+    /// Whether or not the refilling behavior is active.
+    /// </summary>
+    [DataField, ViewVariables, AutoNetworkedField]
+    public bool AutoRefill = true;
+
+    /// <summary>
+    /// How often a new piece of ammunition is inserted into the owner's <see cref="BallisticAmmoProviderComponent"/>.
+    /// </summary>
+    [DataField, ViewVariables, AutoNetworkedField]
+    public TimeSpan AutoRefillRate;
+
+    /// <summary>
+    /// If true, causes the refilling behavior to be delayed by at least <see cref="AutoRefillPauseDuration"/> after
+    /// the owner is fired.
+    /// </summary>
+    [DataField, ViewVariables, AutoNetworkedField]
+    public bool FiringPausesAutoRefill = false;
+
+    /// <summary>
+    /// How long to pause for if <see cref="FiringPausesAutoRefill"/> is true.
+    /// </summary>
+    [DataField, ViewVariables, AutoNetworkedField]
+    public TimeSpan AutoRefillPauseDuration = TimeSpan.Zero;
+
+    /// <summary>
+    /// What entity to spawn and attempt to insert into the owner.
+    /// </summary>
+    [DataField(required: true), ViewVariables, AutoNetworkedField]
+    public EntProtoId AmmoProto;
+
+    /// <summary>
+    /// When the next auto refill should occur. This is just implementation state.
+    /// </summary>
+    [ViewVariables, AutoNetworkedField, AutoPausedField]
+    public TimeSpan NextAutoRefill = TimeSpan.Zero;
+}
--- a/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
+++ b/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
@@ -8,6 +8,7 @@ using Content.Shared.Weapons.Ranged.Events;
 using Robust.Shared.Containers;
 using Robust.Shared.Map;
 using Robust.Shared.Serialization;
+using Robust.Shared.Timing;
 
 namespace Content.Shared.Weapons.Ranged.Systems;
 
@@ -15,8 +16,9 @@ public abstract partial class SharedGunSystem
 {
     [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
     [Dependency] private readonly SharedInteractionSystem _interaction = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
 
-
+    [MustCallBase]
     protected virtual void InitializeBallistic()
     {
         SubscribeLocalEvent<BallisticAmmoProviderComponent, ComponentInit>(OnBallisticInit);
@@ -46,19 +48,8 @@ public abstract partial class SharedGunSystem
         if (args.Handled)
             return;
 
-        if (_whitelistSystem.IsWhitelistFailOrNull(component.Whitelist, args.Used))
-            return;
-
-        if (GetBallisticShots(component) >= component.Capacity)
-            return;
-
-        component.Entities.Add(args.Used);
-        Containers.Insert(args.Used, component.Container);
-        // Not predicted so
-        Audio.PlayPredicted(component.SoundInsert, uid, args.User);
-        args.Handled = true;
-        UpdateBallisticAppearance(uid, component);
-        DirtyField(uid, component, nameof(BallisticAmmoProviderComponent.Entities));
+        if (TryBallisticInsert((uid, component), args.Used, args.User))
+            args.Handled = true;
     }
 
     private void OnBallisticAfterInteract(EntityUid uid, BallisticAmmoProviderComponent component, AfterInteractEvent args)
@@ -251,6 +242,11 @@ public abstract partial class SharedGunSystem
                 component.Entities.RemoveAt(component.Entities.Count - 1);
                 DirtyField(uid, component, nameof(BallisticAmmoProviderComponent.Entities));
                 Containers.Remove(entity, component.Container);
+
+                if (TryComp<BallisticAmmoSelfRefillerComponent>(uid, out var refiller))
+                {
+                    PauseSelfRefill((uid, refiller));
+                }
             }
             else if (component.UnspawnedCount > 0)
             {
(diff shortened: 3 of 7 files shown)


IProduceWidgets:
Oh neat, I was just lamenting my need for this when making a gag item.

Content.Server/Weapons/Ranged/Systems/GunSystem.Ballistic.cs
```diff
@@ -34,4 +43,36 @@ protected override void Cycle(EntityUid uid, BallisticAmmoProviderComponent comp
         var cycledEvent = new GunCycledEvent();
         RaiseLocalEvent(uid, ref cycledEvent);
     }
+
+    private void OnRefillerEmpPulsed(Entity<BallisticAmmoSelfRefillerComponent> entity, ref EmpPulseEvent args)
+    {
+        PauseSelfRefill(entity, args.Duration);
```
IProduceWidgets: I would add a bool to the comp for this for items that shouldn't be effected by emp's since these aren't necessarily "electric."

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// How long to pause for if <see cref="FiringPausesAutoRefill"/> is true.
+    /// </summary>
+    [DataField, ViewVariables, AutoNetworkedField]
+    public TimeSpan AutoRefillPauseDuration = TimeSpan.Zero;
+
+    /// <summary>
+    /// What entity to spawn and attempt to insert into the owner.
+    /// </summary>
+    [DataField(required: true), ViewVariables, AutoNetworkedField]
+    public EntProtoId AmmoProto;
```
IProduceWidgets: should probably use the proto in the BallisticAmmoProvider?

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -270,6 +265,72 @@ private void OnBallisticAmmoCount(EntityUid uid, BallisticAmmoProviderComponent
...
+        bool suppressInsertionSound = false
+    )
+    {
+        if (!CanInsertBallistic(entity, inserted))
+            return false;
+
+        entity.Comp.Entities.Add(inserted);
+        Containers.Insert(inserted, entity.Comp.Container);
+        if (!suppressInsertionSound)
+        {
```
IProduceWidgets, line 333: Why not just increase the count of the unspawned ammo in the BallisticAmmoProvider?

Content.Server/Weapons/Ranged/Systems/GunSystem.Ballistic.cs
```diff
@@ -34,4 +43,36 @@ protected override void Cycle(EntityUid uid, BallisticAmmoProviderComponent comp
...
+                Log.Error(
+                    $"Failed to insert ammo {ammoEntity} into non-full {entity}. Is the {nameof(BallisticAmmoSelfRefillerComponent)}'s {nameof(BallisticAmmoSelfRefillerComponent.AmmoProto)} incorrect for the {nameof(BallisticAmmoProviderComponent)}'s {nameof(BallisticAmmoProviderComponent.Whitelist)}?");
+                continue;
+            }
+
+            refiller.NextAutoRefill = Timing.CurTime + refiller.AutoRefillRate;
+            Dirty(uid, refiller);
+        }
+    }
 }
```
metalgearsloth: If we're adding a specific thing for ballistic and taking the performance hit of it at the very least don't make it spawn entities and make it actually use the unspawned counter.

pheenty:
Unless you want a self-refilling magazine, I think RechargeBasicEntityAmmo would work just fine?..

Centronias:
> Unless you want a self-refilling magazine, I think RechargeBasicEntityAmmo would work just fine?..

This'd perhaps work for the Borg Viper/L6 cases, but I do specifically want a Foam Force borg module which can reuse existing foam darts on the ground as well as have a very slow trickle of new darts coming in to combat curmudgeons who trash existing darts.

Centronias:
Changes responding to Unspawned Count and Ammo Provider ammo proto made.

I've left in the capability for the refiller to use a proto different from the provider's proto in case somebody wants to do something crazy with that. If that's undesireable, lemme know and I can rip it out.

For now, this means that the way the refilling works is:
- Provider + Refiller protos both null -> error
- Provider proto specified + Refiller proto null -> use provider proto, adding to unspanwed
- Provider proto null + Refiller proto specified -> use refiller proto, spawning entity and inserting into container (respects provider whitelist)
- Neither null, but don't match -> use refiller proto, spawning entity and inserting into container (respects provider whitelist)
- Neither null, match -> use provider proto, adding to unspanwed

The logic to do this is a bit hairy, but I don't think it's spaghetti levels, so lemme know what you all think.

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -252,6 +243,11 @@ private void OnBallisticTakeAmmo(EntityUid uid, BallisticAmmoProviderComponent c
                 component.Entities.RemoveAt(component.Entities.Count - 1);
                 DirtyField(uid, component, nameof(BallisticAmmoProviderComponent.Entities));
                 Containers.Remove(entity, component.Container);
+
+                if (TryComp<BallisticAmmoSelfRefillerComponent>(uid, out var refiller))
+                {
+                    PauseSelfRefill((uid, refiller));
+                }
```
SlamBamActionman: Shouldn't this also occur in the `else if (component.UnspawnedCount > 0)` right below as well?

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
```diff
@@ -0,0 +1,58 @@
...
+/// This component, analogous to <c>BatterySelfRechargerComponent</c>, will attempt insert ballistic ammunition into
+/// its owner's <see cref="BallisticAmmoProviderComponent"/>.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, AutoGenerateComponentPause, Access(typeof(SharedGunSystem))]
+public sealed partial class BallisticAmmoSelfRefillerComponent : Component
+{
+    /// <summary>
+    /// Whether or not the refilling behavior is active.
+    /// </summary>
+    [DataField, ViewVariables, AutoNetworkedField]
```
SlamBamActionman: There is no need to include ViewVariables, as that is automatically included in DataField. They're only useful if we explicitly want to restrict it, so just remove them in this case.

SlamBamActionman (approved):

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -290,6 +354,70 @@ public void SetBallisticUnspawned(Entity<BallisticAmmoProviderComponent> entity,
...
+            if (!insertSucceeded)
+            {
+                PredictedQueueDel(ammoEntity);
+                Log.Error(
+                    $"Failed to insert ammo {ammoEntity} into non-full {entity}. This is a configuration error. Is the {nameof(BallisticAmmoSelfRefillerComponent)}'s {nameof(BallisticAmmoSelfRefillerComponent.AmmoProto)} incorrect for the {nameof(BallisticAmmoProviderComponent)}'s {nameof(BallisticAmmoProviderComponent.Whitelist)}?");
+                return;
+            }
+        }
+
+        refiller.NextAutoRefill = Timing.CurTime + refiller.AutoRefillRate;
```
slarticodefast: move this up so that the next refill gets increased and dirtied even if we return early, otherwise it will get spammed

Boaz1111:
Your changelog is missing the 🆑 icon

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Whether or not the refilling behavior is active.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool AutoRefill = true;
+
+    /// <summary>
+    /// How often a new piece of ammunition is inserted into the owner's <see cref="BallisticAmmoProviderComponent"/>.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public TimeSpan AutoRefillRate;
```
ScarKy0: Give this some default value via TimeSpan.FromSeconds(X)

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
```diff
@@ -0,0 +1,58 @@
+﻿using Content.Shared.Weapons.Ranged.Systems;
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Weapons.Ranged.Components;
+
+/// <summary>
+/// This component, analogous to <c>BatterySelfRechargerComponent</c>, will attempt insert ballistic ammunition into
```
ScarKy0:
```suggestion
/// This component, analogous to <see cref="BatterySelfRechargerComponent"/>, will attempt insert ballistic ammunition into
```
I'm not sure what `<c>` does, did you mean to use `<see cref>`?

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// If true, EMPs will pause this component's behavior.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool AffectedByEmp = false;
+
+    /// <summary>
+    /// When the next auto refill should occur. This is just implementation state.
+    /// </summary>
+    [ViewVariables, AutoNetworkedField, AutoPausedField]
+    public TimeSpan NextAutoRefill = TimeSpan.Zero;
```
ScarKy0, line 60:
```suggestion
    /// </summary>
    [DataField, AutoNetworkedField, AutoPausedField]
    public TimeSpan NextAutoRefill = TimeSpan.Zero;
```

Also needs to be a TimeOffsetSerializer

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
```diff
@@ -0,0 +1,58 @@
...
+/// This component, analogous to <c>BatterySelfRechargerComponent</c>, will attempt insert ballistic ammunition into
+/// its owner's <see cref="BallisticAmmoProviderComponent"/>.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState, AutoGenerateComponentPause, Access(typeof(SharedGunSystem))]
+public sealed partial class BallisticAmmoSelfRefillerComponent : Component
+{
+    /// <summary>
+    /// Whether or not the refilling behavior is active.
+    /// </summary>
+    [DataField, AutoNetworkedField]
```
ScarKy0, line 20:
```suggestion
    /// <summary>
    /// Whether the refilling behavior is active.
    /// </summary>
    [DataField, AutoNetworkedField]
```

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -290,6 +354,71 @@ public void SetBallisticUnspawned(Entity<BallisticAmmoProviderComponent> entity,
...
+    )
+    {
+        var ammo = entity.Comp1;
+        var refiller = entity.Comp2;
+        if (!refiller.AutoRefill ||
+            IsFull(entity) ||
+            Timing.CurTime < refiller.NextAutoRefill)
+            return;
+
+        refiller.NextAutoRefill = Timing.CurTime + refiller.AutoRefillRate;
```
ScarKy0:
```suggestion
        refiller.NextAutoRefill += refiller.AutoRefillRate;
```

We set NextAutoRefill in MapInit. This is because the prior method introduces some imprecision that will add up over time, while setting it on MapInit and only adding the update interval makes sure it stays consistent.

ScarKy0:
```suggestion
        var ammo = entity.Comp1;
        var refiller = entity.Comp2;
        
        if (Timing.CurTime < refiller.NextAutoRefill)
        	return;
        	
        refiller.NextAutoRefill = Timing.CurTime + refiller.AutoRefillRate;
        
        if (!refiller.AutoRefill || IsFull(entity))
            return;

        refiller.NextAutoRefill = Timing.CurTime + refiller.AutoRefillRate;
```
Since we're moving this to MapInit, this will need to be changed a bit as to make sure we always get the time updated.
Might also wanna DirtyField the NextAutoRefill here but I'm not entirely sure.

ScarKy0: Ah apparently one of my review comments got eaten.

Yeah you'll need to set `NextAutoRefill` to `CurTime + AutoRefillRate` in a `MapInitEvent`.
We do this because using CurTime every time introduces some inconsistencies, while setting it once OnMapInit and only adding the interval makes sure the timing is consistent.

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
Centronias: `<c>` makes it "code formatted" and I would've used `see`, but this was way back when that component was in Server, and thus not able to be referenced. I can check if I can reference it now.

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
Centronias: I mean this as a rebuke to Jetbrains's garbage grammar checks and not you, Scar, but I'm a native english speaker and this change is purely prescriptive and, in opinion, incorrect.

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
Centronias: So this is something I'm not familiar with -- it'd be wrong to include a value for this field in a proto which uses this component, imo, but is `DataField` needed to be able to like... serialize the state of this component if a grid is saved with it or something?

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
Centronias: Huh really? I'd assumed it was accruing of floating point imprecision that was meant to be avoided -- but I guess that's maybe what `CurTime` is doing behind the scenes?

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
ScarKy0: I believe both forms are correct, i personally use this one more hence why I pointed it out lmao

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
ScarKy0: Im fairly certain ViewVariables are not serialized but DataFields are. Im pretty sure DataFields are used everywhere in update time fields like this too(?)

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
ScarKy0: Im pretty sure CurTime is what causes the imprecision, apparently it once caused a bug that allowed singularities to very rarely eat through containment

Centronias: Sounds like a feature to me :godo:

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
Centronias: I just did a search for `TimeSpan Next` and you're right that all of them are `DataField`s, except in cases where they're specifically client-sided UI limits.

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
Centronias: Does `DirtyField` work without field deltas? Or perhaps you can direct me on how to set up field deltas for this since I wasn't able to figure it out myself.

ScarKy0: You gotta do field deltas.
In the `AutoGenerateComponentState` of the component you do `AutoGenerateComponentState(fieldDeltas: true)`
Then instead of doing Dirty(), you have to do DirtyField() for every modified field (Basically in every place where you Dirty, you will have to DirtyField every modified field seperately for the refiller comp)

Example from SharedDeliverySystem: `DirtyField(ent, ent.Comp, nameof(DeliveryComponent.IsLocked));`, this Dirties the `IsLocked` field of `ent.Comp` for the entity `ent`, where `ent.Comp` is the `DeliveryComponent`.

In this case youd do something like `DirtyField(ent, refiller, nameof(BallisticAmmoSelfRefillerComponent.NextAutoRefill));`

**NOTE:** Your IDE will complain that this method is incorrect. It is. You should be able to build normally.

Content.Shared/Weapons/Ranged/Components/BallisticAmmoSelfRefillerComponent.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// If true, EMPs will pause this component's behavior.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool AffectedByEmp = false;
+
+    /// <summary>
+    /// When the next auto refill should occur. This is just implementation state.
+    /// </summary>
+    [DataField(customTypeSerializer: typeof(TimespanSerializer)), AutoNetworkedField, AutoPausedField]
+    public TimeSpan NextAutoRefill = TimeSpan.Zero;
```
ScarKy0, line 60:
```suggestion
    /// <summary>
    /// When the next auto refill should occur. This is just implementation state.
    /// </summary>
    [DataField(customTypeSerializer: typeof(TimeOffsetSerializer)), AutoNetworkedField, AutoPausedField]
    public TimeSpan NextAutoRefill = TimeSpan.Zero;
```

ScarKy0 (changes requested):

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
Centronias: Ahh okay, I think it missing the generated code is why I didn't get it working before. Thanks 🙇

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -290,6 +360,71 @@ public void SetBallisticUnspawned(Entity<BallisticAmmoProviderComponent> entity,
...
+        if (!refiller.AutoRefill ||
+            Timing.CurTime < refiller.NextAutoRefill)
+            return;
+
+        refiller.NextAutoRefill += refiller.AutoRefillRate;
+        DirtyField(entity, refiller, nameof(BallisticAmmoSelfRefillerComponent.NextAutoRefill));
+
+        if (IsFull(entity))
+            return;
+
```
ScarKy0, line 395: We cannot have any checks except the time check before we update the time.
In this case, if we toggle AutoRefill off for whatever reason, when it gets enabled again will cause the server to quickly catch up and refill a bunch of bullets at once.
It should be together with the IsFull check like I showed in the previous review

Centronias: I don't think that's right, considering the update here only ever inserts one piece of ammunition. If autorefill is false for a while and then swapped to true, this update will run once, inserting one ammo, and then set the next refill time to whatever the normal update rate is.
Because the update behavior here isn't scaled with the update time delta and not with the time since last update to this component's ammo, I don't think it would be an issue.

Not to say your change is wrong functionally; it just, I think, needlessly pushes the update time to clients when that could be avoided.

ScarKy0: The issue is we no longer use CurTime here.
Lets say I insert 1 ammo every 1 second.
At 1s i pause, so i have 1 ammo in my mag
Real time: 2s | Update Loop Time: 1s - 1 ammo
Real time: 3s | Update Loop Time: 1s - 1 ammo
Real time: 4s | Update Loop Time: 1s - 1 ammo

Now, i unpause the autorefill. The update loop still thinks we sit at 1s, and quickly tries to catch up to 4s, giving me 3 ammo immediately.

Centronias: OH! You're 100% right.
Imagine me internalizing anything that ever changes.
garfieldyeah

PR #10897, opened 2022-08-28
Melee refactor

metalgearsloth:
Re-opening because it's my main focus again. See changelog for the highlights. Even if we end up reverting windups it's still good to have the update code there because it makes it much easier for forks.

Want to make melee feel as good as guns. I also want to get right of the: open context menu > left-click mob meta.

Requires https://github.com/space-wizards/RobustToolbox/pull/3253 for the new keybind

Edit:
Alright so I've made like 6 prototypes at this point and the current one I'm settled on:

- Light attacks happen on click. These attack whatever is under your cursor.
- Heavy attacks happen after a wind-up on right-click. This also means context menu is disabled in combat mode. These are just the current wide attacks but they do additional damage. Both of these will use stamina in the near future (after I predict it) so heavy use more stamina proportionally so it becomes a trade-off.

:cl:
- fix: Melee combat swings are now predicted. The damage effect still only shows when the server confirms the hit.
- fix: Melee combat sprite lunge direction has been fixed.
- fix: You won't pick up items (e.g. rats, spears) when trying to melee them anymore.
- add: Melee combat now has light and heavy attacks. Light attacks under your cursor and heavy attack in your cursor's direction. Heavy attacks also do more DPS and can cleave but require a windup.
- tweak: Disarm is now on right-click when you have a free hand.
- tweak: The context menu won't show up when in combat mode anymore as right-click has been repurposed.
- tweak: Most melee values have been tweaked (mainly attack rates and damage values) to make weapons more distinct. DPS values are similar (excluding the change in frequency on bleeding / chemical injection procs).

PR #7811, opened 2022-04-27
Make Disarm opt-in

Mirino97:
## About the PR 
Adds a bool to CombatMode to switch Disarm opt-out to opt-in.
It's a saner way of dealing with instead of adding new mobs/species and getting an issue on github saying "Bee's should disarm" then having to retroactively remove the function.

That being said, I'm unsure if this is a sane way of implementing it. I'm still very much in the "if it works, it works" stages of coding, so feel free to scrutinize, thank you!

PR #38542, opened 2025-06-24
Lets monkeys & kobolds shove/disarm!

Kittygyat:
## About the PR
I let monkeys & kobolds shove/disarm, with combat mode RMB whilst unarmed.

## Why / Balance
Oftentimes, a natural monkey/kobold ghost role has to escape and/or fight the chef for survival, and usually they won't have _any_ tools or weapons at their disposal, this PR lets them shove for a chance at survival; disarming the chef & providing the monkey/kobold with hope.

## Technical details
I just added "canDisarm: true" to "MobBaseAncestor"

## Media

https://github.com/user-attachments/assets/5426ff9e-4580-418f-a509-640f9a285617

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Monkeys & Kobolds can now shove/disarm like regular species!

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/animals.yml
@@ -1325,6 +1325,7 @@
   abstract: true
   components:
   - type: CombatMode
+    canDisarm: true
   - type: Inventory
     templateId: monkey
     speciesId: monkey


Velken:
lone op with kobold army disarming all of sec.

slarticodefast:
Ok, according to git blame and this PR https://github.com/space-wizards/space-station-14/pull/7811 monkey used to be able to do this, which means it was removed on purpose at some point. Since this means this PR reverts that, could you try finding the PR that removed that ability from them so we can look at the balance reasons given for that?

Kittygyat:
Okay, it seems like back in PR #10897 the canDisarm for monkeys was changed from ``canDisarm: True`` to ``disarm: null``, and I have no idea why, or if it was even intentional.
@slarticodefast

slarticodefast:
Ok, thanks for finding it. Looks like it was just silently removed there without any reasoning.
This PR should probably also adjust Scurrets to keep them consistent with kobolts and monkeys.
I'm also doing a combat workgroup ping for combat balance feedback.

keronshb:
On the basis it was silently removed it should be readded.
Monkeys should definitely be able to shove/disarm, it makes them that much more threatening.
Since then Scurrets were also added so add it to them and this is good to go.

SlamBamActionman:
My base instinct is that shoving is something that someone with a hand slot should be able to do. 👍

Kittygyat:
> Since then Scurrets were also added so add it to them and this is good to go.

They share the same parent as kobolds & monkeys, so this already affects them!

slarticodefast:
Should be good to merge then!

PR #38551, opened 2025-06-24
Don't add ImplicitRoof to grids with roof component

### LINKED ISSUE #38401: planet stations have roof
yes we still don't have it on upstream, but if we are going to have it, there is a problem

all grids, which planet stations are, get ImplicitRoofComponent on initialize so MapLightComponent can't work properly

related code
https://github.com/space-wizards/space-station-14/blob/5a69b7e2fbca585e479cd6c5bc2c9965ab5db4f1/Content.Server/Shuttles/Systems/ShuttleSystem.cs#L100-L107

mtrs163:
## About the PR

Fixes #38401

ImplicitRoof component will not be forced upon the grids that already have RoofComponent. ImplicitRoof will also be removed if it loads with a grid that has RoofComponent.

## Why / Balance

This could potentially cause issues for the planet/island stations (if there will ever be such stations) that do have roofs and MapLight components and for forks that already have such stations.

## Technical details

ShuttleSystem now checks if the grid has RoofComponent. If it does, system removes the ImplicitRoof component, else it ensures that ImplicitRoof is present.

## Media

Testing demonstration (?)

https://github.com/user-attachments/assets/560514e8-59d9-4894-8259-9d27e3149453

## Requirements

## Breaking changes

None

**Changelog**

I guess it's not needed since there are no player-facing changes

### DIFF
--- a/Content.Server/Shuttles/Systems/ShuttleSystem.cs
+++ b/Content.Server/Shuttles/Systems/ShuttleSystem.cs
@@ -105,13 +105,11 @@ public sealed partial class ShuttleSystem : SharedShuttleSystem
 
         EnsureComp<ShuttleComponent>(ev.EntityUid);
 
+        // This and RoofComponent should be mutually exclusive, so ImplicitRoof should be removed if the grid has RoofComponent
         if (HasComp<RoofComponent>(ev.EntityUid))
-        {
             RemComp<ImplicitRoofComponent>(ev.EntityUid);
-            return;
-        }
-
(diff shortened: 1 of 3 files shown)


iaada (commented):
Since the idea is to make the two components mutually exclusive, I think is would read better as an if{}else instead of an early return.

mtrs163:
> Since the idea is to make the two components mutually exclusive, I think is would read better as an if{}else instead of an early return.
> 
> It would also be nice if those components had a comment saying they were mutually exclusive.

Is it OK now?

PR #38552, opened 2025-06-24
fix: correct wall AO having weird sharp edges (macOS only?)

perryprog:
## About the PR

Wall AO had very weird edges and odd looking blur for me on macOS Apple Silicon, not matching the screenshots in #38276, so this is probably macOS only? Anyway see media.

## Technical details

Seems like BlurRenderTarget isn't meant to have its buffer and render target be the same render target, and making a new render target fixes it for me... but if this _is_ macOS specific then that's pretty bizarre.

Please ensure this gets tested on several devices before merging (at the lesat, on Windows).

## Media

<details>
<summary>Screenshots</summary>

Before close-up with default color:
image

Before with red AO color:
image

After:
image

After with default color:
image

</details>

## Requirements

## Breaking changes

**Changelog**

no

### DIFF
--- a/Content.Client/Light/AmbientOcclusionOverlay.cs
+++ b/Content.Client/Light/AmbientOcclusionOverlay.cs
@@ -24,6 +24,7 @@ public sealed class AmbientOcclusionOverlay : Overlay
     public override OverlaySpace Space => OverlaySpace.WorldSpaceBelowEntities;
 
     private IRenderTexture? _aoTarget;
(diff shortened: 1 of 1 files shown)


PJB3005:
P0 as I'm pretty sure this is invalid graphics API usage and could result in all kinds of nonsensical visual corruption, which could pose actual risk to users.

perryprog:
In that case I missed my chance to changelog something like
> - fix: remove memetic hazard

:P

PR #38556, opened 2025-06-24
Properly evaluate publish condition for master branch

VasilisThePikachu:
Was made aware of this by @Centronias in https://github.com/space-wizards/space-station-14/pull/38431#discussion_r2164747248

This should only run on wizden

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
     - name: Fail if we are attempting to run on the master branch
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     steps:
     - name: Fail if we are attempting to run on the master branch
-      if: ${{GITHUB.REF_NAME == 'master'}} && github.repository == 'space-wizards/space-station-14'
+      if: ${{GITHUB.REF_NAME == 'master' && github.repository == 'space-wizards/space-station-14'}}
```
Centronias, line 18: You can omit the `${{...}}` around it entirely, fwiw

PR #38569, opened 2025-06-25
Convert DumpReagentGuideText command to LEC.

VerinSenpai:
## About the PR
Localized the command (I ain't writing that damn name out again, its like a word twister for my fingers).

I am as of present unsure of how I wish to use the argument shell strings. Whether to show as it is implied it should be used the argument from the help string or (as I think is probably better) show the users input. If the latter the string should probably be named better but I digress. Work to be done another day.

## Requirements

### DIFF
--- a/Content.Server/Chemistry/Commands/DumpReagentGuideText.cs
+++ b/Content.Server/Chemistry/Commands/DumpReagentGuideText.cs
@@ -1,5 +1,4 @@
-﻿using System.Linq;
-using Content.Server.Administration;
+﻿using Content.Server.Administration;
 using Content.Shared.Administration;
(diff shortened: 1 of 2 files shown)


VerinSenpai:
I opted to convert the dual for loop to a linq expression per riders suggestion but I'm not familiar with the performance impact of linq. Is linq bad to use in this instance? it is only a command and is not used in an update loop of any sort.

slarticodefast:
Keep the loop, linq can be slow sometimes.
It probably does not matter in this case and the command won't be used a lot, but it's good practice in general.

VerinSenpai:
you got it.

VerinSenpai:
all good 👌

PR #38572, opened 2025-06-25
RespiratorSystem Cleanup

Princess-Cheeseballs:
## About the PR
Eventified a decent amount of the RespiratorSystem so it can be used without a BodySystem.

Made the entityQuery only for Respirators.

## Why / Balance

Useful for downstream and necessary for decoupling BodySystem from RespiratorSystem.

Needed for downstream + general cleanup

## Technical details

- Cleaned up some tests to not copypaste code
- Added two new events that raise the gas inhaled and gas exhaled that are handleable for downstream use and also for Body System to do its thing if it exists.
- Moved one const to be used in the component instead
- Breath volume is now able to be changed on a per respirator basis.

## Requirements

## Breaking changes

**Changelog**

no cl no fun, no breaking changes

### DIFF
--- a/Content.Server/Body/Components/RespiratorComponent.cs
+++ b/Content.Server/Body/Components/RespiratorComponent.cs
@@ -22,18 +22,6 @@ namespace Content.Server.Body.Components
         };
 
         /// <summary>
-        ///     The gas our entity breathes
-        /// </summary>
-        [DataField]
-        public Gas MetabolizedGas = Gas.Oxygen;
-
-        /// <summary>
-        ///     The gas our entity exhales
-        /// </summary>
-        [DataField]
-        public Gas ByproductGas = Gas.CarbonDioxide;
-
-        /// <summary>
         ///     Volume of our breath in liters
         /// </summary>
         [DataField]
@@ -44,12 +32,6 @@ namespace Content.Server.Body.Components
         /// </summary>
         [DataField]
         public float Ratio = 1.0f;
-
-        /// <summary>
-        ///     Saturation per mol of gas
-        /// </summary>
-        [DataField]
-        public float SaturationMultiplier = Atmospherics.BreathMolesToReagentMultiplier;
 
         /// <summary>
         ///     The next time that this body will inhale or exhale.
@@ -104,12 +86,6 @@ namespace Content.Server.Body.Components
         /// </summary>
         [DataField]
         public ProtoId<EmotePrototype> GaspEmote = "Gasp";
-
-        /// <summary>
-        /// The type of gas this respirator needs. Used only for the breathing alerts, not actual metabolism.
-        /// </summary>
-        [DataField]
-        public ProtoId<AlertPrototype> Alert = "LowOxygen";
 
         /// <summary>
         ///     How many cycles in a row has the mob been under-saturated?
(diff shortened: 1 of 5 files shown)


Princess-Cheeseballs:
Test fail incoming, and patch incoming.

Princess-Cheeseballs:
Surprise test fail where Nukies have poison in their lungs apparently? Only nukies too.

Princess-Cheeseballs:
While this is up should I move either of the two consts into the component? Specifically the molesToReagent and BreathVolume consts?

Princess-Cheeseballs:
> While this is up should I move either of the two consts into the component? Specifically the molesToReagent and BreathVolume consts?

This wasn't hard to implement and was so I went ahead and did it, default values are equal to the atmos consts so nothing has changed but they are now manipulable if one so chooses to manipulate them.

FairlySadPanda:
I'm helicoptering here, apologies: I'm pretty sure in my analysis of Respirator my analysis is that it's Mob code that's been bolted on to Body, not Body code itself.

_checks code_

Y-es I think that's still my take. You can tell with the way that the component itself can freely ignore Body exists but you've got this "respirator->lungs" and "lungs->respirator" patch.

So it's probably incomplete in terms of how far I'd take this refactor but there's no sense making the PR ever-larger to try and define a Mob namespace etc

Content.Server/Body/Systems/InternalsSystem.cs
```diff
@@ -9,6 +10,7 @@
 using Content.Shared.Internals;
 using Content.Shared.Inventory;
 using Content.Shared.Roles;
+using NetCord.Rest;
```
sowelipililimute: stray change?

Content.Server/Body/Components/RespiratorComponent.cs
```diff
@@ -9,6 +11,46 @@ namespace Content.Server.Body.Components
...
+        ///     The gas our entity exhales
+        /// </summary>
+        [DataField]
+        public Gas ByproductGas = Gas.CarbonDioxide;
+
+        /// <summary>
+        ///     Volume of our breath in liters
+        /// </summary>
+        [DataField]
+        public float BreathVolume = Atmospherics.BreathVolume;
```
sowelipililimute: as said on discord, i'm not keen on having explicitly incomplete fallback behaviour that will get obsoleted to better simulate both simplemobs and psychomed mobs. i'd either like to see this have more fleshed out fallback behaviour or i'd like to see it get gutted so that this PR is laser focused on "decouple bodysystem from respiratorsystem"

Content.Server/Body/Systems/RespiratorSystem.cs
```diff
@@ -296,11 +387,37 @@ bool CanMetabolize(EntityEffect effect)
...
         if (ent.Comp.SuffocationCycles == 2)
             _adminLogger.Add(LogType.Asphyxiation, $"{ToPrettyString(ent):entity} started suffocating");
 
+        _damageableSys.TryChangeDamage(ent, ent.Comp.Damage, interruptsDoAfters: false);
+
+        if (!HasComp<BodyComponent>(ent))
+        {
+            _alertsSystem.ShowAlert(ent, ent.Comp.Alert);
+            return;
+        }
```
sowelipililimute: this feels like it should be based on an event being handled or not, but also might be included in my concerns about incomplete fallback

sowelipililimute (changes requested):

sowelipililimute (approved):
looks fine to me now

PR #38573, opened 2025-06-25
allow combat mode toggling when unable to interact

qwerltaz:
## About the PR

allow combat mode toggling when cuffed or crit, or when unable to perform other normal actions. fix https://discord.com/channels/310555209753690112/1385281778486345870

## Why / Balance

This prevents being stuck in combat mode while in crit, and while cuffed, so you don't stare awkwardly in one direction and can still examine things.


VerinSenpai:
Incredible.

PR #38579, opened 2025-06-25
New Status Effects system: Relay events

TheShuEd:
## About the PR

As I started porting other status effects, I realized that we needed a full effects event relay, similar to inventory relays
This pr adds this relay, updates some internal logic to support working with Relay, no breaking changes.

Also, I made a separate status effect for SSD sleep, and removed unnecessary logic that is no longer needed.

## Why / Balance

Must have

## Requirements


Fildrance (approved):
Looks good

Issue #38567, opened 2025-06-25
The lathe build queue is inefficient


slarticodefast:
## Description
Related to https://github.com/space-wizards/space-station-14/issues/38564
~~This is very inefficient. Queueing a few thousand products to be printed brings the server to a complete stop.
It should just store the ProtoId, since we don't need any information from the prototype until we actually print it.~~

Edit: Like delta pointed out this should just be a pointer saved in the list, but for some reason this still makes the game lag like hell if you add 1000+ of them.

Also successive entries of the same recipe could be combined into one by adding the amount to be printed. That way the list would not become gigantic if you queue a lot of items.

![Image](https://github.com/user-attachments/assets/b61fcce7-8dfd-4d7e-baa4-33df8844962f)

deltanedas:
isnt it just a (shared gc) pointer though

slarticodefast:
Oh, good point. Somehow the game is still shitting bricks when you add 1000+ of them though. A list of pointers should not cause such a problem.

deltanedas:
would still be good to be ids for persistence, but something else must be causing the extreme slowdown

Issue #38564, opened 2025-06-25
Lathe build queue has no limit


slarticodefast:
## Description
https://github.com/space-wizards/space-station-14/pull/38557 fixed a bug where tiles were free, so you could just queue several thousand of them, instantly crashing the server.
This is still possible for other very cheap items or if you have a lot of materials.
The print queue should have a maximum to prevent this from happening.

**Reproduction**
- Spawn a Cutter Machine
- Fill it with 1000 Steel
- Print 4000 steel tiles
- Lag

**Screenshots**

![Image](https://github.com/user-attachments/assets/5aa6449e-2844-4894-8077-65b05810431c)

iaada:
This appears to have been resolved from my (not very robust) testing

https://github.com/user-attachments/assets/5487613f-b073-40ed-8f54-c025a97f98e5

PR #38583, opened 2025-06-25
Improve lathe queue performance

Tayrtahn:
## About the PR

Makes lathe queue logic more efficient, especially with large queues.

## Why / Balance

Resolves https://github.com/space-wizards/space-station-14/issues/38567.

## Technical details

Two main changes:
- The queue now stores `ProtoId<LatheRecipePrototype>`s instead of `LatheRecipePrototype`s. This significantly reduces the size of the BUI state, reducing network use.
- The queue is now actually a Queue. It was a List before this change, despite being used as a queue. Notably, this meant that removing the first element from the front of the queue with `RemoveAt(0)` was quite inefficient when the queue was long: O(n), as each element needed to be shifted forward. Now that the queue is a proper Queue, this is a simple `Dequeue`, which is O(1).
 
## Requirements

## Breaking changes

`LatheComponent.Queue` is now a `Queue<ProtoId<LatheRecipePrototype>>` instead of a `List<LatheRecipePrototype>`.

### DIFF
--- a/Content.Shared/Lathe/LatheMessages.cs
+++ b/Content.Shared/Lathe/LatheMessages.cs
@@ -10,11 +10,11 @@ public sealed class LatheUpdateState : BoundUserInterfaceState
 {
     public List<ProtoId<LatheRecipePrototype>> Recipes;
 
-    public NetListAsArray<ProtoId<LatheRecipePrototype>> Queue;
+    public ProtoId<LatheRecipePrototype>[] Queue;
 
     public ProtoId<LatheRecipePrototype>? CurrentlyProducing;
 
-    public LatheUpdateState(List<ProtoId<LatheRecipePrototype>> recipes, NetListAsArray<ProtoId<LatheRecipePrototype>> queue, ProtoId<LatheRecipePrototype>? currentlyProducing = null)
+    public LatheUpdateState(List<ProtoId<LatheRecipePrototype>> recipes, ProtoId<LatheRecipePrototype>[] queue, ProtoId<LatheRecipePrototype>? currentlyProducing = null)
     {
         Recipes = recipes;
         Queue = queue;
--- a/Content.Shared/Lathe/LatheMessages.cs
+++ b/Content.Shared/Lathe/LatheMessages.cs
@@ -1,4 +1,5 @@
 using Content.Shared.Research.Prototypes;
+using NetSerializer;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Serialization;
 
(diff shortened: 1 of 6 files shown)


Content.Shared/Lathe/LatheMessages.cs
```diff
@@ -9,11 +9,11 @@ public sealed class LatheUpdateState : BoundUserInterfaceState
 {
     public List<ProtoId<LatheRecipePrototype>> Recipes;
 
-    public List<LatheRecipePrototype> Queue;
+    public Queue<ProtoId<LatheRecipePrototype>> Queue;
```
PJB3005: `Queue<T>` should not be sent over the network. Use `NetListAsArray<T>` instead.

PJB3005 (changes requested):

PJB3005:
Also, the lathe *really* should not be allowing people to queue 1,000 crafts at once.

slarticodefast:
> Also, the lathe really should not be allowing people to queue 1,000 crafts at once.

Yeah, I already made an issue for this. It's why we even noticed the List problem.
https://github.com/space-wizards/space-station-14/issues/38564

whatston3:
> Notably, this meant that removing the first element from the front of the queue with RemoveAt(0) was quite inefficient when the queue was long

Couldn't you use RemoveSwap?

If it's kept as a list, you maintain the ability to easily adjust the relative position of jobs in the queue, no?  If there's any interest in batching items and swapping their order (https://github.com/new-frontiers-14/frontier-station-14/pull/3205), I'd be happy to port it.

deltanedas:
swap remove would then put the newest item at the start of the "queue" instead of being FIFO

whatston3:
> swap remove would then put the newest item at the start of the "queue" instead of being FIFO

~~This would be on a List and not the Queue - the first index in the list is the item being produced at the moment, yes?  [If so, the last element would be at the end of the list - orphaning it and reducing the count should be fine.]  Also, RemoveSwap works using IList?~~ I'm a dingus.

PJB3005 (changes requested):
Remove `[Serializable]` from `LatheRecipePrototype` too while you're at it.

Tayrtahn:
Yeah, RemoveSwap won't work for this, but you make a good point about being able to edit the queue. Maybe a linked list would be better. 🤔

Content.Shared/Lathe/LatheMessages.cs
```diff
@@ -9,11 +10,11 @@ public sealed class LatheUpdateState : BoundUserInterfaceState
 {
     public List<ProtoId<LatheRecipePrototype>> Recipes;
 
-    public List<LatheRecipePrototype> Queue;
+    public NetListAsArray<ProtoId<LatheRecipePrototype>> Queue;
```
metalgearsloth: If you want to be sweaty make a deltastate for queue vs recipes and that will probably cut bandwidth by a large amount. Then just move all of this logic over to the compstate.

PJB3005: This isn't worth the complexity, especially if we do the sane thing and limit the amount of queue entries.

PJB3005 (approved):

PR #38585, opened 2025-06-25
Power stat and nuke codes commands get some LEC love.

VerinSenpai:
## About the PR

Not much to say, just localized these two commands.


Resources/Locale/en-US/commands/power-stat.ftl
```diff
@@ -0,0 +1,6 @@
+﻿cmd-powerstat-desc = Shows statistics for pow3r.
```
slarticodefast: for consistency rename the file to `powerstat-command.ftl`

Resources/Locale/en-US/commands/nuke-codes-command.ftl
```diff
@@ -0,0 +1,2 @@
+﻿cmd-nukecodes-desc = Send nuke codes to a station's communication consoles.
```
slarticodefast: same here with `nukecodes-command.ftl`

slarticodefast (changes requested):

Resources/Locale/en-US/commands/power-stat.ftl
VerinSenpai: I got nothing for this. Not sure how I messed that up.

Resources/Locale/en-US/commands/nuke-codes-command.ftl
VerinSenpai: Mhmm Ill make the requested change, though Ive been adding this space between words on my other prs.

slarticodefast: Hmm, ok maybe keep it consistent with the other files then and don't change it.
Having the file be the same name as the command would have made it easier to find, but I guess it does not matter that much.

VerinSenpai: Your logic sways me. Incoming pr to change 500 localization file names.

VerinSenpai:
@slarticodefast Changes done. I decided to follow your advice and rename it to nukecodes-command.ftl

The files are still pretty inconsistent through out so this is something I can work to improve over time.

slarticodefast (approved):

PR #38588, opened 2025-06-25
ShowHealthBarsCommand to LEC.

VerinSenpai:
## About the PR
Used generics where possible, performed some cleanup and converted the command to LEC.

### DIFF
--- a/Content.Client/Commands/ShowHealthBarsCommand.cs
+++ b/Content.Client/Commands/ShowHealthBarsCommand.cs
@@ -19,7 +19,7 @@ public sealed class ShowHealthBarsCommand : LocalizedEntityCommands
             return;
         }
 
-        if (player.AttachedEntity is not {} playerEntity)
+        if (player.AttachedEntity is not { } playerEntity)
         {
             shell.WriteError(Loc.GetString("shell-must-be-attached-to-entity"));
             return;


Content.Client/Commands/ShowHealthBarsCommand.cs
```diff
@@ -1,57 +1,46 @@
...
         if (player == null)
         {
-            shell.WriteError(LocalizationManager.GetString($"cmd-{Command}-error-not-player"));
+            shell.WriteError(Loc.GetString("shell-only-players-can-run-this-command"));
             return;
         }
 
-        var playerEntity = player?.AttachedEntity;
-        if (!playerEntity.HasValue)
+        if (player.AttachedEntity is not {} playerEntity)
```
slarticodefast:
```suggestion
        if (player.AttachedEntity is not { } playerEntity)
```
fix formatting

slarticodefast:
Thanks for all the cleanup effort with these PRs!

Content.Client/Commands/ShowHealthBarsCommand.cs
VerinSenpai: The space is preferred. Noted 🙂

PR #38589, opened 2025-06-26
Convert a few debugging commands and the mapping setup command to LEC.

VerinSenpai:
## About the PR
Performs cleanup as I go. Don't think any localization was needed here as it was taken care of.


Content.Client/Commands/MappingClientSideSetupCommand.cs
```diff
@@ -1,32 +1,29 @@
...
-            _lightManager.Enabled = false;
-            shell.ExecuteCommand("showsubfloor");
-            _entitySystemManager.GetEntitySystem<ActionsSystem>().LoadActionAssignments("/mapping_actions.yml", false);
-        }
+        if (_lightManager.LockConsoleAccess)
+            return;
+
+        _markerSystem.MarkersVisible = true;
+        _lightManager.Enabled = false;
+        shell.ExecuteCommand("showsubfloor"); // boop
```
VerinSenpai: Is this really ideal? I feel like it'd be more clear to just have the subfloorhidesystem brought in and set showall to true explicitly here.

VerinSenpai: actually I'm fairly certain this would hide the subfloor if you already had it visible.

VerinSenpai: I just went ahead and sent a change for this.

PR #38591, opened 2025-06-26
Fix pinpointer screen rotation solution 1

Quantum-cross:
## About the PR

The rotation of the screen layer of the pinpointer is off by half a pixel. This fixes that.

Note: this PR is mutually exclusive with #38657, leaving it up to maints to choose the solution.

## Why / Balance

bug fix

## Technical details

Just need to use a transformation matrix to rotate the layer about a different point.

## Media

The screen layer needs to be rotated around the red dot, it's actually rotated around the intersection of the lines.

image

### Before

image image image image

### After

image image image image

## Requirements

## Breaking changes

none

**Changelog**
:cl:
- fix: pinpointer will no longer trigger sufferers of OCD

### DIFF
--- a/Content.Client/Pinpointer/PinpointerSystem.cs
+++ b/Content.Client/Pinpointer/PinpointerSystem.cs
@@ -10,6 +10,20 @@ public sealed class PinpointerSystem : SharedPinpointerSystem
     [Dependency] private readonly IEyeManager _eyeManager = default!;
     [Dependency] private readonly SpriteSystem _sprite = default!;
 
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<PinpointerComponent, AppearanceChangeEvent>(OnAppearanceChange);
+    }
+
+    private void OnAppearanceChange(Entity<PinpointerComponent> ent, ref AppearanceChangeEvent args)
+    {
+        if (args.Sprite is null)
+            return;
+
+        UpdateArrow((ent, ent.Comp, args.Sprite), force: true);
+    }
+
     public override void Update(float frameTime)
     {
         base.Update(frameTime);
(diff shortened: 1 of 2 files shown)


Quantum-cross:
As far as I can tell it's not possible to rotate a layer not around the center without the transformations. Maybe later I can math out a closed solution to avoid the multiplications.

whatston3:
Why not make it a wide boy so it's also of even width?
image

Quantum-cross:
The size of the screen of the pin pointer sprite would have to be changed and centered

Quantum-cross:
Realistically the angle should be saved and the sprite matrix transformation should only be updated of the angle is changed.

whatston3:
> The size of the screen of the pin pointer sprite would have to be changed and centered

Not here, it's still 4 pixels away on one side.

[never mind, I'm speaking out of my ass - it would be smaller, an (8x8?) vs. a (9x9?), should be fine? in the pictures you posted, it'd still be within bounds, yes?]

Quantum-cross:
> > The size of the screen of the pin pointer sprite would have to be changed and centered
> 
> Not here, it's still 4 pixels away on one side.
> 
> [never mind, I'm speaking out of my ass - it would be smaller, an (8x8?) vs. a (9x9?), should be fine? in the pictures you posted, it'd still be within bounds, yes?]

yeah but then it wouldn't be centered on the screen

whatston3:
image

Seems alright to me.

[pinpointer.rsi.zip](https://github.com/user-attachments/files/20916836/pinpointer.rsi.zip)

Quantum-cross:
what I mean is that it's less to do with the arrow sprite and more to do with the pinpointer sprite.

image

Whatever the arrow is, it needs to rotate around the red point here.

Your arrow still rotates incorrectly:

image  image image image

Notice how it is always to one side of the center line.

Quantum-cross:
that being said, I've pushed a change that will only update the rotation matrix when the arrow angle has changed or the appearance has changed (like doing from "reached" to "close")

whatston3:
> snip

No, the sprites I showed you have an 8x8 screen.  Everything that involves that rotation is of an even width.

whatston3:
To be unambiguous - the zip file included is a complete set of sprites for the pinpointer with an 8x8 screen and 2px thick arrows.  This should solve your issue without oddball rotation.

[A fix for "this sprite should be rotated around such and such pixel", seems like it's something that should be most useful in the engine, and would be hugely useful for odd-width/height sprites more generally.  But I'm just some yahoo.]

Quantum-cross:
> To be unambiguous - the zip file included is a complete set of sprites for the pinpointer with an 8x8 screen and 2px thick arrows. This should solve your issue without oddball rotation.

Understood. then that is a valid solution if every sprite has been enlarged to be accurately centered in the sprite.

I can open a PR with your solution crediting you @whatston3 if you want, then maints can decide which solution to merge.

Personally I don't like the chonky arrow as much, but it certainly simplifies the code.

Quantum-cross:
superceded to #38657

PR #38602, opened 2025-06-26
Stop network serializing prototypes

PJB3005:
Send the damn proto ID instead.

# Breaking Changes

Various prototypes have been made no longer `[NetSerializable, Serializable]`, and places these were directly serialized over the network have been changed to use `ProtoId<T>` instead.


slarticodefast:
Test fail is real.
Looks like a sandbox violation.

Content.Shared/Administration/PlayerInfo.cs
```diff
@@ -11,7 +12,7 @@ public sealed record PlayerInfo(
     string IdentityName,
     string StartingJob,
     bool Antag,
-    RoleTypePrototype RoleProto,
+    ProtoId<RoleTypePrototype>? RoleProto,
```
Tayrtahn, line 15: Breaking change should be documented.

PR #38609, opened 2025-06-26
Make vending machine restocks predicted (and its sound not spammable)

perryprog:
## About the PR

Make restocking stuff predicted and also stop the restock sound if the restock doafter is canceled since it's mega spammable.

## Technical details

Was basically already all predicted. 
- Switched two separate popup calls to the combined method.
- Added a stream for tracking the played audio for stopping it.
  Had to throw in some IsFirstTimePredicteds in there to deal with making sure the sound gets stopped properly on the initiator's client if they spamclick the restock.
- Standard field cleanup to the restock component.

## Requirements

## Breaking changes

**Changelog**

nah

### DIFF
--- a/Content.Shared/VendingMachines/VendingMachineComponent.cs
+++ b/Content.Shared/VendingMachines/VendingMachineComponent.cs
@@ -139,6 +139,11 @@ namespace Content.Shared.VendingMachines
         [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
         public TimeSpan NextEmpEject = TimeSpan.Zero;
 
+        /// <summary>
+        /// Audio entity used during restock in case the doafter gets canceled.
+        /// </summary>
+        public EntityUid? RestockStream;
+
         #region Client Visuals
         /// <summary>
         /// RSI state for when the vending machine is unpowered.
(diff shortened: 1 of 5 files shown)


slarticodefast:
Does stopping the audio on the client work properly now?
grafik

perryprog:
Yeah, for this one it was fine, I think because of a dirty it had somewhere. You can test by starting the restock at the edge of the interaction range and then walking away as soon as you click.

Content.Shared/VendingMachines/VendingMachineComponent.cs
```diff
@@ -139,6 +139,11 @@ public sealed partial class VendingMachineComponent : Component
         [DataField(customTypeSerializer: typeof(TimeOffsetSerializer))]
         public TimeSpan NextEmpEject = TimeSpan.Zero;
 
+        /// <summary>
+        /// Audio entity used during restock in case the doafter gets canceled.
+        /// </summary>
+        public EntityUid? RestockStream;
```
slarticodefast, line 146:
```suggestion
        [DataField]
        public EntityUid? RestockStream;
```

slarticodefast:
Good stuff 👍

Issue #38575, opened 2025-06-25
Old status effects need to be converted to the new system


slarticodefast:
https://github.com/space-wizards/space-station-14/pull/37238 Refactored the status effect system, but to make the transition easier we currently have both in use until all old status effects have been moved over.

If you want to help out best split the PRs up into multiple batches for easier review instead of making one giant PR with all of them.

- [ ] Stun
- [ ] KnockedDown
- [ ] SlowedDown
- [ ] Jitter
- [ ] Stutter
- [ ] AllCaps
- [ ] SeeingRainbows
- [ ] Electrocution
- [ ] Drunk
- [ ] SlurredSpeech
- [ ] PressureImmunity
- [ ] Muted
- [ ] Corporeal
- [ ] ForcedSleep
- [ ] TemporaryBlindness
- [ ] Pacified
- [ ] RatvarianLanguage
- [ ] StaminaModifier
- [ ] Flashed
- [ ] RadiationProtection
- [ ] Adrenaline

TheShuEd:
KnockedDown is done iirc?

TheShuEd:
Stutter, Slurred speach and Drunk ported in #38678

Pok27:
At this point, every item on this list has either already been completed or has a PR. Just a note for contributors

PR #38620, opened 2025-06-27
Refactor SeeingRainbows to new status effect system

TheShuEd:
## About the PR
Helps with: #38575 

Project Zero warnings PR basically.

## Requirements

## Breaking changes

Migrated the SeeingRainbows status effect to use the new status effect system with SeeingRainbowsStatusEffectComponent. Updated all relevant code, overlays, and reagent prototypes to reference the new effect and component. Removed legacy status effect definitions and updated effect application to use ModifyStatusEffect with effectProto: StatusEffectSeeingRainbow.

### DIFF
--- a/Content.Client/Drugs/RainbowOverlay.cs
+++ b/Content.Client/Drugs/RainbowOverlay.cs
@@ -1,6 +1,5 @@
 using Content.Shared.CCVar;
 using Content.Shared.Drugs;
-using Content.Shared.StatusEffect;
 using Content.Shared.StatusEffectNew;
 using Robust.Client.Graphics;
 using Robust.Client.Player;
@@ -39,6 +38,7 @@ public sealed class RainbowOverlay : Overlay
     public RainbowOverlay()
     {
         IoCManager.InjectDependencies(this);
+
         _statusEffects = _sysMan.GetEntitySystem<SharedStatusEffectsSystem>();
 
         _rainbowShader = _prototypeManager.Index<ShaderPrototype>("Rainbow").InstanceUnique();
--- a/Content.Shared/StatusEffectNew/StatusEffectSystem.API.cs
+++ b/Content.Shared/StatusEffectNew/StatusEffectSystem.API.cs
@@ -260,7 +260,7 @@ public abstract partial class SharedStatusEffectsSystem
             }
         }
 
-        return effects != null;
+        return effects is not null;
     }
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


Content.Client/Drugs/RainbowOverlay.cs
```diff
@@ -54,18 +58,23 @@ protected override void FrameUpdate(FrameEventArgs args)
...
+            if (statusEffectComp.EndEffectTime > remainingTime)
+                remainingTime = statusEffectComp.EndEffectTime;
+        }
+
+        if (remainingTime is null)
             return;
 
-        var timeLeft = (float)(time.Value.Item2 - time.Value.Item1).TotalSeconds;
+        var curTime = _timing.CurTime;
+        var timeLeft = (float)(remainingTime - curTime).Value.TotalSeconds;
```
Tayrtahn: It seems like this is going to be a pretty common pattern for effect overlays and probably other stuff. We should make a helper method in `StatusEffectSystem` that does this so we don't wind up with this code copy-pasted all over the place.

TheShuEd: Make sense

Content.Shared/StatusEffectNew/StatusEffectSystem.API.cs
```diff
@@ -262,4 +262,40 @@ public bool TryEffectsWithComp<T>(EntityUid? target, [NotNullWhen(true)] out Has
...
+            }
+
+            if (statusComp.EndEffectTime > endTime)
+                endTime = statusComp.EndEffectTime;
+        }
+
+        if (endTime is null)
+            return false;
+
+        return true;
```
Tayrtahn:
```suggestion
        return endTime is not null;
```

Content.Shared/StatusEffectNew/StatusEffectSystem.API.cs
```diff
@@ -262,4 +262,40 @@ public bool TryEffectsWithComp<T>(EntityUid? target, [NotNullWhen(true)] out Has
 
         return effects != null;
     }
+
+    /// <summary>
+    /// Helper function for calculating how long it takes for all effects with a particular component to disappear. Useful for overlays.
+    /// </summary>
+    /// <param name="target">An entity from which status effects are checked.</param>
+    /// <param name="endTime">The farthest end time of effects with this component is returned. Can be null if one of the effects is infinite.</param>
+    /// <returns>Helper function. Returns True if effects with the specified component were found, or False if there are no such effects.</returns>
```
Tayrtahn:
```suggestion
    /// <returns>True if effects with the specified component were found, or False if there are no such effects.</returns>
```

PR #38638, opened 2025-06-28
Changes for displaying weather on tiles

mtrs163:
## About the PR

Requires #38551 I guess.
Weather now can be displayed on all tiles. ImplicitRoof now affects weather on tiles, not only lighting.

## Why / Balance

We already have roofs that can limit where weather can be, so I don't think tiles should determine whether or not they will show weather.

## Technical details

- Check for ImplicitRoof has been added to CanWeatherAffect function
- Removed ContentTileDefenition's "Weather" property

## Media

https://github.com/user-attachments/assets/62ce111f-8a2d-45c6-a374-2574442d28e3

Testing demonstration.

## Requirements

## Breaking changes

Weather property was removed from ContentTileDefinition. Tiles no more determine whether or not weather will be displayed on them. This should now be determined using RoofComponent or ImplicitRoofComponent.

**Changelog**

No player facing changes.

### DIFF
--- a/Content.Shared/Weather/SharedWeatherSystem.cs
+++ b/Content.Shared/Weather/SharedWeatherSystem.cs
@@ -48,9 +48,7 @@ public abstract class SharedWeatherSystem : EntitySystem
         if (Resolve(uid, ref roofComp, false) && _roof.IsRooved((uid, grid, roofComp), tileRef.GridIndices))
             return false;
 
-        var tileDef = (ContentTileDefinition) _tileDefManager[tileRef.Tile.TypeId];
-
-        if (!tileDef.Weather)
+        if (HasComp<ImplicitRoofComponent>(uid))
             return false;
 
         var anchoredEntities = _mapSystem.GetAnchoredEntitiesEnumerator(uid, grid, tileRef.GridIndices);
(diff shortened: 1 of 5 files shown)


little-meow-meow:
This is very nice. My only comment is that, because ambient lighting doesn't simulate 'bouncing', I like to remove the roof on tiles adjacent to exterior windows so that it looks more convincing. This change interferes with that technique. Perhaps `RoofComponent` could be modified to track weather rendering in a backwards-compatible way? Then adding a weather toggle to `SetRoofComponent` and changing `SharedRoofSystem` accordingly. I can make a follow-up PR if that's too much scope creep.

mtrs163:
I'm not sure how can I do this without just copy-pasting the roof code to the weather systems.

Content.Shared/Weather/SharedWeatherSystem.cs
little-meow-meow, line 1: `_tileDefManager` is no longer used after these changes

mtrs163: Done

little-meow-meow:
I back-linked a pull request that has a quick and sloppy implementation. `teamstarcup/starcup` is GPL-licensed, but I'll explicitly dual-license my code in that pull request to MIT.

mtrs163:
oh shit looks like i broke everything

Princess-Cheeseballs:
Actually no that test fail was unrelated and fixed by: #39116

PR #38644, opened 2025-06-28
Suppress prototype instantiation warning in `BodyPrototypeSerializer`

Tayrtahn:
## About the PR

Suppresses a warning from manually instantiating a prototype in some code that needs to be redone anyway.

## Why / Balance

## Technical details

The entire purpose of this class is to consume a `MappingDataNode` and return a new `BodyPrototype`, which means that it has to make a new prototype instance. This code needs to be redone so it doesn't need a custom serializer to read the prototype, and body code in general needs to be rethought, so it's not really worth the effort to redo it now.

## Media

Screenshot 2025-06-28 at 4 18 12 PM

## Requirements

### DIFF
--- a/Content.Shared/Body/Prototypes/BodyPrototypeSerializer.cs
+++ b/Content.Shared/Body/Prototypes/BodyPrototypeSerializer.cs
@@ -173,6 +173,9 @@ public sealed class BodyPrototypeSerializer : ITypeReader<BodyPrototype, Mapping
             slots.Add(slotId, slot);
         }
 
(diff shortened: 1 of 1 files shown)


slarticodefast:
I guess the code needs to be rewritten anyways, but shouldn't we leave the warning in until that has happened?
I don't see any advantage in suppressing it other than sweeping it under the rug.
And fulfilling PZW like this feels like cheating 😄

Tayrtahn:
Yeah, you're right - we've gotta earn this one!

The warning can be a reminder/motivation to fix it properly.

PR #38645, opened 2025-06-29
Allow backwards travel through disposal routers

perryprog:
## About the PR

This adds a toggle to the UI for disposal routers that will  allow items to move "backwards" (opposite the sprite's arrow) through it, as opposed to making the item do a 180° if it doesn't match the filter. This effectively turns it into a Y-junction.

## Why / Balance

- Currently it's very difficult to build large routed disposals network if they don't conform to a sort of circular shape. This makes it _much_ easier to make long router runs that don't double back on themselves.
- This is a UI toggle and not just a change to its behavior for map backwards compatibility (mapwards compatibility?). If this is needed, I would vastly prefer this just being the default behavior, and is what my first commit does.
- I'm open to changing this to be a new disposal pipe that is just a filtered Y junction.
- Disposal junctions have a similar issue, except it will roll 50/50 on which direction it throws your item if it's entering backwards. This adds the flag to junctions as well, but there's no UI to toggle it (yet, see technical details). 

## Technical details

- Disposal code is currently really outdated, and for the sake of scope I'm not going to update it to newer standards in this PR. Notably, this includes the fact that the router UI is not predicted which makes using it feel just... not great.
- I plan on making a second PR that brings disposal code up to date, makes its UIs predicted, and so on without having to include any user-facing changes in it.
  - As part of this I'm hoping to merge routers and junctions, which would give junctions a UI with this toggle—assuming this stays as a toggle.

## Media

https://github.com/user-attachments/assets/996a4521-4e02-4c63-90cb-0bcedaa35025

If someone has a better label for the checkbox please say.

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Disposal routers now have an option that allows items to travel backwards through them.

### DIFF
--- a/Content.Client/Disposal/Tube/DisposalRouterBoundUserInterface.cs
+++ b/Content.Client/Disposal/Tube/DisposalRouterBoundUserInterface.cs
@@ -23,14 +23,19 @@ namespace Content.Client.Disposal.Tube
 
             _window = this.CreateWindow<DisposalRouterWindow>();
 
-            _window.Confirm.OnPressed += _ => ButtonPressed(UiAction.Ok, _window.TagInput.Text);
-            _window.TagInput.OnTextEntered += args => ButtonPressed(UiAction.Ok, args.Text);
+            _window.Confirm.OnPressed += _ =>
+                ButtonPressed(UiAction.Ok, _window.TagInput.Text, _window.BackwardsAllowed.Pressed);
+            _window.TagInput.OnTextEntered +=
+                args => ButtonPressed(UiAction.Ok, args.Text, _window.BackwardsAllowed.Pressed);
+            _window.BackwardsAllowed.OnPressed +=
+                args => ButtonPressed(UiAction.Ok, _window.TagInput.Text, args.Button.Pressed, false);
         }
 
-        private void ButtonPressed(UiAction action, string tag)
+        private void ButtonPressed(UiAction action, string tag, bool backwardsAllowed, bool close = true)
         {
-            SendMessage(new UiActionMessage(action, tag));
-            Close();
+            SendMessage(new UiActionMessage(action, tag, backwardsAllowed));
+            if (close)
+                Close();
         }
 
         protected override void UpdateState(BoundUserInterfaceState state)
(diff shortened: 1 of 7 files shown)


perryprog:
Oh also @Emisse @ArtisticRoomba since this (potentially) affects mapping. I'd really love to do the version of this without the GUI change but I'm not sure how impactful that might be on maps with more complex disposal networks. (I don't know wizden maps...) (Also downstreams exist.)

ArtisticRoomba:
If the default behavior of the router is the current behavior (with the toggle being the optional inversed behavior) then it won't affect mapping. I will note though that while this does make certain disposals routing setups compact, I think it will cause a lot of confusion, as if I recall correctly, backwards travel with combiners is non-deterministic.

Maybe someone who knows a bit more about disposals can comment on it better than I can.

perryprog:
Yeah, default behavior is current behavior, but design-wise I'm not sure it'd be clear why such a toggle exists. Since I know the majority of maps definitely default to the giant "loop" technique I'm not sure if it'd actually be that impactful to use the breaking change version of this PR, since that type of setup is fully compatible with either option. (To see the change in behavior you'd need to actively have a disposal/mailing unit on the wrong side of the arrow.)

> if I recall correctly, backwards travel with combiners is non-deterministic.
If by combiners you mean junctions then yes, it currently is a 50/50 on which exit you take if you enter through the exit side of a junction. For router junctions you get routed as if you came in through the normal entrance. 

I'm also considering to (though not in this PR) give this same option to disposal junctions to allow deterministic backwards travel through them, since if you can think of a clever way to use that, I think that's neat. I'm less sure about that one though because giving _disposal junctions_ a UI (or verb, interaction, whatever) seems a little bloaty compared to extending the already existing UI of junction routers.

ArtisticRoomba:
Disposals works on the concept of pressure - not that in SS14 we have any major problems with butting heads against realism. So in this instance, the item *not* magically reversing and instead choosing to continue along its direction, makes some sense to me.

As for the mapping breaking changes, the default behavior stays the same. So no issues there from me.

perryprog:
Sounds good to me!

But I mean, I dunno, I can't think of anywhere we have a blurry line with realism. Nowhere at all.

ArtisticRoomba:
Thanks for your work contributing as always.

chromiumboy:
Frozen due to the disposals refactor (#40540)

ArtisticRoomba:
Heya perry. The disposals refactor has been merged now so this is technically unfrozen. However this is so out of date that I don't think this is mergable anymore and the behavior might have been fixed in the refactor. You're always welcome to reopen, sorry it has to be like this.

perryprog:
Oh no worries at all! Cheers! 🫡

Issue #38628, opened 2025-06-27
Some vending machines at the terminal will say they are unpowered when they are infact powered


Moomoobeef:
## Description
Some vending machines at the spawn terminal will say that they are unpowered when you interact, but are infact powered and can be interacted with. My best guess is that this is a mis-prediction, but I haven't actually looked at the code.

**Reproduction**
1. spawn at the terminal
2. go to a vending machine
3. some machines behave normally, some do not

**Screenshots**
![Image](https://github.com/user-attachments/assets/6d4540aa-1a79-4347-82b8-df94e1c6f59b)
![Image](https://github.com/user-attachments/assets/fa7577df-21d7-4a1b-880c-729013558e00)

Moomoobeef:
![Image](https://github.com/user-attachments/assets/2eced038-ade5-4526-a89b-e2f238dde8c8)
the arcades also do this

Tayrtahn:
Thanks for making an issue for this - I've been meaning to investigate it for a while and this reminded me to actually do it.

I found the problem and it's quite silly and easy to fix. I'm going out for a bit, but I'll PR a fix in a few hours.

PR #38647, opened 2025-06-29
Fix devices in terminal mispredicting power state

Tayrtahn:
## About the PR

Devices like vending machines and arcade games in the terminal no longer incorrectly state that they do not have power before opening.

## Why / Balance

Fixes https://github.com/space-wizards/space-station-14/issues/38628.

## Technical details

Spot the difference:
Screenshot 2025-06-28 at 4 41 00 PM

Screenshot 2025-06-28 at 4 41 20 PM

This PR changes the client version to also return true from that conditional.

Yes, I also think it's silly/bad to have two seemingly-identical static classes that need to be kept in sync - there really should just be a single class in Shared. Unfortunately, the client and server have different `ApcPowerReceiverComponent` definitions and `TryGetComponent` doesn't work with inheritance (so it can't be used with `SharedApcPowerReceiverComponent`). Try building and running with [the reverted commit](https://github.com/space-wizards/space-station-14/commit/611e56e031636734abab1ad3e77bf88b69a4de13) in this PR if you want to see the problem yourself. This should be dealt with in the future, but for now, this PR just puts the client and server classes back in sync.

## Requirements


perryprog:
Technically you can do this very normal and very okay way :^)

```c#
public static bool IsPowered(this EntitySystem system, EntityUid uid, IEntityManager entManager, SharedApcPowerReceiverComponent? receiver = null)
{
    var comp = entManager.ComponentFactory.GetRegistration("ApcPowerReceiver");
    IComponent? teehee = null;
    if (receiver == null && !entManager.TryGetComponent(uid, comp, out teehee))
        return true;
    receiver = teehee as SharedApcPowerReceiverComponent;
    return receiver?.Powered ?? true;
}
```

Tayrtahn:
> Technically you can do this very normal and very okay way :^)
> 
> ```cs
> public static bool IsPowered(this EntitySystem system, EntityUid uid, IEntityManager entManager, SharedApcPowerReceiverComponent? receiver = null)
> {
>     var comp = entManager.ComponentFactory.GetRegistration("ApcPowerReceiver");
>     IComponent? teehee = null;
>     if (receiver == null && !entManager.TryGetComponent(uid, comp, out teehee))
>         return true;
>     receiver = teehee as SharedApcPowerReceiverComponent;
>     return receiver?.Powered ?? true;
> }
> ```

That's very clever and I kinda hate it!

beck-thompson (approved):
Yeah agree with what you said they should be combined but this is clearly a mistake!

PR #38650, opened 2025-06-29
Minor escape menu update

beck-thompson:
## Why / Balance

It looks better! Not 100% sure on the red quit button

## Media

image

## Requirements

**Changelog**

:cl:
- tweak: The escape menu ui has been updated slightly.

### DIFF
--- a/Content.Client/Options/UI/EscapeMenu.xaml
+++ b/Content.Client/Options/UI/EscapeMenu.xaml
@@ -6,13 +6,15 @@
             Resizable="False">
 
     <BoxContainer Orientation="Vertical" SeparationOverride="4" MinWidth="150">
-        <changelog:ChangelogButton Access="Public" Name="ChangelogButton"/>
         <ui:VoteCallMenuButton />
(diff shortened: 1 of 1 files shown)


Kowlin:
I think the red quit button is fitting. Its a button that expresses "danger". Something that impacts more when just a normal button.

Errant-4:
Can you please also split off `Call Vote` into its own group, since it does not really belong with those other 3 buttons conceptually (but could be accompanied by the Bug Report/Feedback feature's button, whenever that happens)

lzk228:
maybe it's not within the scope of this pr, but i'd like the quit button to require two clicks 
like it's done with choosing type of borg chassis

beck-thompson:
> maybe it's not within the scope of this pr, but i'd like the quit button to require two clicks
like it's done with choosing type of borg chassis

It wasn't super clear cut what people thought of it when I brought it up so I didn't want to add it to this pr specifically! Not a bad idea though

PR #38653, opened 2025-06-29
Delete label workflows in favor for SS14.Labeller

Simyon264:
## About the PR

Deletes our current labeling workflows in favor of https://github.com/space-wizards/SS14.Labeller

## Why / Balance

Currently, our label workflows are kind of bad. They race condition each other and the review label workflow does not work at all.
As a solution I have made SS14.Labeller, a NativeAOT ASP NET Core application that labels PRs and issues using a webhook and GitHub PAT Token. Because it uses NativeAOT compilation, the application only uses around 20mb of RAM.

## Requirements

## Breaking changes

Label workflows have been removed in favor of https://github.com/space-wizards/SS14.Labeller


ArtisticRoomba:
my hero

Simyon264:
DNM while SS14.Labeller isn't setup.

PJB3005 (changes requested):
Do we want to just disable these workflows instead? So downstreams can use them.

Simyon264:
If we just disable the checks it (probably, not sure) show up under every PR as a "skipped" check which is annoying

PJB3005:
> If we just disable the checks it (probably, not sure) show up under every PR as a "skipped" check which is annoying

This has been shown to be incorrect.

PR #38660, opened 2025-06-29
New Status API Additions

Princess-Cheeseballs:
## About the PR

Added a couple new methods to SharedStatusEffectSystem for API purposes and bugfixing purposes.

Renamed some methods as well to make it more readable after significant method shuffling

Will need input on the names because they are currently ass, we might need to rename a bunch of the methods for it to all be readable.

## Why / Balance

You should really be able to update a status effect to be infinite with a TryAdd.
Also you should really be able to not update the time if it would last longer than the new time.

## Technical details

- Let SetStatusEffectTime allow a null TimeSpan
- SetStatusEffectTime can set EndEffectTime to null
- Added SetGreaterStatusEffectTime which sets the greater of current EndEffectTime and the passed duration
- Added GetStatusEffectTime as an internal TimeSpan
### TryAddStatusEffect has been renamed to EnsureOrUpdateStatusEffect
- Added two new Methods, EnsureStatusEffect and EnsureLongestStatusEffect
- EnsureStatusEffect will update an existing status effect to the inputted duration or create a new effect
- EnsureLongestStatusEffect will update an existing status effect's duration only if the new duration would last longer, or create a new effect
- EnsureOrUpdateStatusEffect will update an existing status effect's duration only if the old duration wasn't infinite
- New Method TryAddStatusEffect, tries to add a status effect. If the other EnsureMethods can't find an existing status effect entity they will call this method.
- Added TryGetMaxTime method from a different status PR of mine.
- Methods that add a status effect entity now have a [NotNullIf(true)] for the EntityUid they return.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/EntityEffects/Effects/StatusEffects/ModifyStatusEffect.cs
+++ b/Content.Shared/EntityEffects/Effects/StatusEffects/ModifyStatusEffect.cs
@@ -20,7 +20,7 @@ public sealed partial class ModifyStatusEffect : EntityEffect
     public float Time = 2.0f;
 
     /// <remarks>
-    /// true - refresh status effect time, false - accumulate status effect time.
+    /// true - refresh status effect time (update to greater value), false - accumulate status effect time.
     /// </remarks>
     [DataField]
     public bool Refresh = true;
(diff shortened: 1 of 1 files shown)


Princess-Cheeseballs:
PR description should be accurate now

Fildrance:
i'll try to look at it today. Ed proposed following names
AddStatusEffect - add timer
UpdateStatusEffect - update if greater
SetStatusEffect - set timer without any regards to previous values

need to look closely how this will look :>

TheShuEd (approved):
Ok, thats seems good for me 🤔

Fildrance:
I had to look at GenericStatusEffect to ensure that we ported implementation faithfully and realized that Refresh on it actually was standing for setting greater value between existing one and provided. The fact that its called 'Refresh' annoyed me greatly but i feel like chaning this could lead to poor experience with porting old status effect-related stuff onto new system, so i respected old API (for now). Migration that will maybe change it - should follow after we remove old system (thus ensuring that all downstreams who were interested ported their related yaml so we can provide some simple migration recommendation).

PR #38661, opened 2025-06-29
Resolving Wizard casting recall on nuke disk making it impossible to disarm

### LINKED ISSUE #38501: Wizard recall-item spell casted on nuke-disk while nuke is armed prevents disarm entirely
## Description
Using the Recall item spell and linking to the nuclear authentication disk, then using recall item to pull the disk to your hand while the nuke is armed, prevents you placing the disk back in, and prevents disarming the nuke.

## Reproduction
1. Learn the recall-item wizard spell
2. Link the spell to the nuclear authentication disk
3. Place the disk into the regular station nuke (nukeops nuke acts the same)
4. Enter the code and arm the nuke
5. Use recall-item to pull the disk to your hand
6. Note that you can no longer disarm the nuke via the UI with the button being grayed out

maland1:
## About the PR

Updating Nuke code to properly handle cases where the disk is removed "magically" without disarming and ejecting properly.

## Why / Balance

Fixes #38501 

## Technical details

- Adding a DisarmBomb case to nuke status update loop
- Changing a few methods and parameters to properly follow formatting standards
- Updating some names to follow camelCase

## Media

https://github.com/user-attachments/assets/0c412ac1-c7dd-4af4-9e72-91ebf233fb81

image

## Requirements

## Breaking changes

Renamed:
DisarmDoafterLength to DisarmDoAfterLength
DisarmBombDoafter to DisarmBombDoAfter

**Changelog**

:cl:
- fix: Fixed Wizard recalling nuke disk from inside an armed nuke making it impossible to defuse


slarticodefast:
What happens if the nuke is either in the `AWAIT_ARM` or the `AWAIT_CODE` state and the disk is recalled? Wouldn't we have to switch states in that case as well?

maland1:
> What happens if the nuke is either in the `AWAIT_ARM` or the `AWAIT_CODE` state and the disk is recalled? Wouldn't we have to switch states in that case as well?

As shown below, both those cases are already covered by the original code so no changes necessary for that!


https://github.com/user-attachments/assets/26ab1053-9c2e-40b2-b492-144882724d9f

RangerXVII:
Off topic but did this actually happen during a round?

SlimmSlamm:
I think this is something wizards would actually do though. They shouldn't be able to take it out while its armed though.

maland1:
> Off topic but did this actually happen during a round?

Not sure, I would assume so given the issue was raised.


> I think this is something wizards would actually do though. They shouldn't be able to take it out while its armed though.

I don't have any strong opinions - but I think if the Wizard does somehow manage to get recall on the disk then while annoying for Nukies it rewards the wizard and by extent the station because they then have to kill him to get it back. 

This was just the easiest solution but it's welcome to be discussed in this PR anyways and we can see what some others think regarding blocking the disk being removed via the recall.

archee1:
> Off topic but did this actually happen during a round?

 didn't happen in a round, i was experimenting with ways to RR the nuke disk and found this along the way

maland1:
Ah good catch in that case!

Content.Server/Nuke/NukeComponent.cs
```diff
@@ -55,8 +55,8 @@ public sealed partial class NukeComponent : SharedNukeComponent
         /// <summary>
         ///     How long a user must wait to disarm the bomb.
         /// </summary>
-        [DataField("disarmDoafterLength")]
-        public float DisarmDoafterLength = 30.0f;
+        [DataField("disarmDoAfterLength")]
```
slarticodefast:
```suggestion
        [DataField]
```
Datafield names are automatically converted from PascalCase to camelCase in yaml, so this is redundant
But I would probably just keep it as it is to prevent a breaking change for forks. Could you undo renaming this?

slarticodefast (approved):
Thanks for the bugfix!

PR #38663, opened 2025-06-29
Add Breach of Permanent Confinement to Space Law

SlamBamActionman:
## About the PR

Based on a discussion with AdmiralOblivious.

This PR adds a new Space Law charge under the Capital category: **Breach of Permanent Confinement**

This is very similar to the existing _Breach of Custody_ charge, with the main difference being that the escapee of the breach was sentenced to permanent confinement before escaping.

_Refusal of Mental Shielding_ has been moved from the (E)/00 category to (R)/07 to make space for the new charge. 

## Why / Balance

The previous description of Breach of Custody had the following line:
> While rare, this charge can be bumped to an execution if the suspect has repeatedly attempted to break out of the permanent brig.
which was vague for when it could apply. This, combined with the out of game consequences for incorrectly applying this exception, makes it rare to apply. 

Breach of Permanent Confinement adds proper definitions for what is required for Capital punishment to be issued by clearly stating it requires the escapee to have been sentenced to perma. It also clearly lays out that assisting with a perma escape is similarly punishable. The latter has been made more common with the introduction of genpop, where a lower-level detainee decides to help out a perma-prisoner simply by virtue of being able to. 

This new charge does give Security clearer guidelines for when an execution may be issued for breaching custody, which may cause an uptick in them. If that's the case, it might be worth looking over the 15 minute timing requirement for when someone may be perma'd. 

Breach of Custody had its description edited to follow Breach of Permanent Confinement's description. Refusal of Mental Shielding was previously in the detainment category and therefore moved to make space, but it probably fits better in the rioting category anyhow.

## Technical details

Text editing

## Media

image

image

## Requirements

## Breaking changes

**Changelog**

No CL, it's a Space Law change.

### DIFF
--- a/Resources/ServerInfo/Guidebook/ServerRules/SpaceLaw/SLCrimeList.xml
+++ b/Resources/ServerInfo/Guidebook/ServerRules/SpaceLaw/SLCrimeList.xml
@@ -253,7 +253,7 @@
     </ColorBox>
     <ColorBox Color="#3b4354">
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
-        Refusal of Mental Shielding (E)
+        Refusal of Mental Shielding (R)
       </Box>
     </ColorBox>
     <ColorBox>
--- a/Resources/ServerInfo/Guidebook/ServerRules/SpaceLaw/SLCrimeList.xml
+++ b/Resources/ServerInfo/Guidebook/ServerRules/SpaceLaw/SLCrimeList.xml
@@ -56,7 +56,7 @@
     </ColorBox>
     <ColorBox Color="#593a4e">
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
-        Refusal of Mental Shielding (E)
+        Breach of Permanent Confinement (E)
       </Box>
     </ColorBox>
     <ColorBox>
@@ -251,8 +251,9 @@
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
       </Box>
     </ColorBox>
-    <ColorBox>
+    <ColorBox Color="#3b4354">
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
+        Refusal of Mental Shielding (E)
       </Box>
     </ColorBox>
     <ColorBox>
@@ -752,7 +753,7 @@
     </ColorBox>
     <ColorBox>
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
-        While rare, this charge can be bumped to an execution if the suspect has repeatedly attempted to break out of the permanent brig. Includes people breaking others out.
+        Includes breaking out other people.
       </Box>
     </ColorBox>
     <ColorBox>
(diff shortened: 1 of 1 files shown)


Coolsurf6:
I think it probably would be nice having a changelog to let people know the definition is much better now. Might make it easier for security to realise the change for the goobers who read the changelogs :D

Admiral-Obvious-001:
Unanimously accepted by the admin team (who voted).

Resources/ServerInfo/Guidebook/ServerRules/SpaceLaw/SLCrimeList.xml
```diff
@@ -847,17 +848,17 @@
...
     <ColorBox>
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
-        Refusal of Mental Shielding
+        Breach of Permanent Confinement
       </Box>
     </ColorBox>
     <ColorBox>
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
-        To refuse to comply with a reasonable Mind Shielding procedure.
+        To break someone sentenced to Permanent Confinement out of a cell or custody with the intention of escaping.
```
LaCumbiaDelCoronavirus, line 856:
```suggestion
        To enable or cause escape from custody serving as permanent confinement.
```
is clearer, and 'custody' refers to any form of imprisonment anyway (can rewrite it to "any custody" otherwise but seems redundant to me). Implicitly includes both self escape and helping others to escape so you can omit the remarks section for this

SlamBamActionman: I do not think this is clearer, due to the following reason:
"enable escape from custody serving as permanent confinement" allows for double-reading it as treating the *location* itself as "permanent confinement". The original version is explicit in that a person sentenced to Permanent Confinement must be involved.

"Well they broke the wall to the place where we keep perma prisoners. Execute them."
"Was there anyone sentenced to perma that they broke out?"
"No, but genpop is serving as perma too, and they enabled escaping from it. Off with their heads."

Now I do not think that most players will read it that way, but do not underestimate the impressive rules-lawyering Salamander players can go through.

SlamBamActionman:
PR has been rebased onto stable.

Resources/ServerInfo/Guidebook/ServerRules/SpaceLaw/SLCrimeList.xml
LaCumbiaDelCoronavirus: >  for double-reading it as treating the location itself as "permanent confinement".

This was intended but I didnt actually foresee it getting abused to this extent, thanks for pointing it out. I think you could do
```suggestion
        To cause escape, or to have intentionally enabled conditions or opportunities that acted as a substantial and direct cause of escape, of any individual expressly imposed with Permanent Confinement 
```
so it only applies if you just broke someone out, or intentionally caused a chain of events that directly led to perma prisoner's escape, which wasnt well covered in the original

its probably resistant enough against being abused in the way that you specified, though I intentionally allowed being able to waive this punishment due to intent: e.x., a secoff who died in perma to skill issue and allowed the prisoner to escape isnt *necessarily* liable for the escape unless someone makes the executive judgement that it is infact the secoff's fault

For example: something like killing cameras in genpop and the person breaking out successfully would land you this charge, and so would breaking open genpop with a c4, but this won't actually apply if the prisoner who broke out wasnt serving a perma sentence,  or the cause of escape was coincidental

It also intentionally doesn't cover failed escapes, but what an escape is, is up to reasonable imagination of sec (some people interpret it as breaking out of your confinement, others as having totally and successfully avoided sec after getting out)

(Reason for this: I wish to be 1984)

NoahCHeaverin:
Why does this have the Changes: Audio tag?

PicklOH:
Space Law is very player facing for a whole department. It should get a minor CL entry.

Admiral-Obvious-001:
Vote finalized 100% yes.

ArtisticRoomba (changes requested):
Add detailed CL please. I know that headmins make announcements when changing space law but a CL entry would be good as well.

Resources/ServerInfo/Guidebook/ServerRules/SpaceLaw/SLCrimeList.xml
```diff
@@ -251,8 +251,9 @@
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
       </Box>
     </ColorBox>
-    <ColorBox>
+    <ColorBox Color="#3b4354">
       <Box HorizontalAlignment="Left" VerticalAlignment="Stretch" Margin="2">
+        Refusal of Mental Shielding (E)
```
crazybrain23: Shouldn't this be (R) since you moved it down to the (R) row?

crazybrain23: The reason the colour and letter are important is because they indicate what crimes are linked are cannot be stacked.

PR #38666, opened 2025-06-30
Dsay Dirty and Follow commands converted to LEC and localized.

VerinSenpai:
## About the PR
Converts the aforementioned commands to LocalizedEntityCommands and localizes where necessary. Performs minor cleanup where able. Moves locale to local/en-us/commands/ to be with the rest of em.

Question: is the player status check in the follow command necessary? I'm unfamiliar with how this check is supposed to work be it as it returns true whether you're in game or in lobby.

## Requirements

### DIFF
--- a/Content.Server/Administration/Commands/DSay.cs
+++ b/Content.Server/Administration/Commands/DSay.cs
@@ -2,39 +2,36 @@ using Content.Server.Chat.Systems;
 using Content.Shared.Administration;
 using Robust.Shared.Console;
 
-namespace Content.Server.Administration.Commands
+namespace Content.Server.Administration.Commands;
+
+[AdminCommand(AdminFlags.Moderator)]
+public sealed class DsayCommand : LocalizedEntityCommands
 {
-    [AdminCommand(AdminFlags.Moderator)]
-    sealed class DSay : IConsoleCommand
+    [Dependency] private readonly ChatSystem _chatSystem = default!;
+
+    public override string Command => "dsay";
+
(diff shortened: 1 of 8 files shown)


Content.Server/Administration/Commands/DSay.cs
```diff
@@ -2,39 +2,36 @@
...
-            {
-                shell.WriteError(Loc.GetString("shell-cannot-run-command-from-server"));
-                return;
-            }
+            shell.WriteError(Loc.GetString("shell-cannot-run-command-from-server"));
+            return;
+        }
 
-            if (player.AttachedEntity is not { Valid: true } entity)
-                return;
```
slarticodefast, line 27: why remove the error message?

slarticodefast:
I assume the status check is there because you can run commands while you are in the lobby. But in that case you have no body to follow anyone with.

Content.Server/Administration/Commands/DSay.cs
VerinSenpai: I believe im adding an error message there, unless thats what you meant

VerinSenpai:
Thats the thing, it seemingly returns true whether youre actively in the round or sitting in the lobby. Ill have to check that again

Content.Server/Administration/Commands/DSay.cs
slarticodefast: Wait, I'm stupid 😆

slarticodefast (approved):

PR #38674, opened 2025-06-30
Add a safety net so you don't accidentally throw important items away

notquitehadouken:
## About the PR

If you throw away what is considered an "Important Item" (right now basically just swords, cap laser, and syndi contra), a popup will appear and after a short delay you will be able to throw away that item.

## Media

https://github.com/user-attachments/assets/b68d3413-75a8-4266-a577-96cda8a12fd2

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: It is now slightly harder to accidentally throw away important things.

### DIFF
--- a/Content.Client/Disposal/Unit/DisposalUnitSystem.cs
+++ b/Content.Client/Disposal/Unit/DisposalUnitSystem.cs
@@ -1,5 +1,7 @@
+using Content.Client.Popups;
 using Content.Shared.Disposal.Components;
 using Content.Shared.Disposal.Unit;
+using Content.Shared.Popups;
 using Robust.Client.Animations;
 using Robust.Client.GameObjects;
 using Robust.Client.Graphics;
@@ -14,6 +16,7 @@ public sealed class DisposalUnitSystem : SharedDisposalUnitSystem
     [Dependency] private readonly SharedAudioSystem _audioSystem = default!;
     [Dependency] private readonly SharedUserInterfaceSystem _uiSystem = default!;
     [Dependency] private readonly SpriteSystem _sprite = default!;
+    [Dependency] private readonly PopupSystem _popupSystem = default!;
 
     private const string AnimationKey = "disposal_unit_animation";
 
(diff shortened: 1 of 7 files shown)


MidZik:
Not a fan. Will get in the way of antags trying to quickly hide/stash items, both their own contra or items they have stolen. And throwing away such an item is a reversible mistake, I don't think it warrants a warning that prevents the item from being tossed.

I'd be more OK with a warning popup appearing but the action still being allowed. Something like "Your sabre felt important...".

Winkarst-cpu:
Maybe make it require doubleclick instead?

First click - popup
Second - throw

notquitehadouken:
> Maybe make it require doubleclick instead?
> 
> First click - popup Second - throw

it is specifically not designed like this because people tend to spam click

MidZik:
Either way, it still changes how the interaction works. Some items would require one click and others would need two. Having two different incompatible workflows for the same action isn't great for the user experience.

Winkarst-cpu:
> Either way, it still changes how the interaction works. Some items would require one click and others would need two. Having two different incompatible workflows for the same action isn't great for the user experience.

Also I can already predict that somebody will die or fail at their tider activities because of that.

notquitehadouken:
oh yeah that reminds me i was supposed to add a new accessibility option for this 😂

slarticodefast:
I think this is just bad UX and unintuitive as you will need to know which items are considered important to be able to know how the game will react when trying to throw something.
How often does it even happen that you accidentally throw something away? The default keybind does not have any overlap with others, so I don't see this as a problem that needs fixing.

ArtisticRoomba:
Brings back feelings of Hypixel Skyblock where it would scream at me for dropping a Rookie Axe.

Jokes aside though, I do agree with Slarti. Players should be allowed to slip up and do silly things.

Feel free to contribute in the future.

notquitehadouken:
> Feel free to contribute in the future.

🫃

PR #38678, opened 2025-06-30
[NEW STATUS SYSTEM] Drunkenness, Stuttering, and Bloodloss

Princess-Cheeseballs:
## About the PR

Moved drunkenness, stuttering to the new status system.
Added Bloodloss as its own status effect.

## Why / Balance

New status good

## Technical details

- Moved DrunkSystem and StutteringSystem to NewStatusSystem
- Renamed DrunkSystem to SharedDrunkSystem
- Fixed a typo in DrunkSystem
- Made DrunkOverlay compatible with the new status system

## Requirements

## Breaking changes

Bloodloss no longer uses Drunk and Stuttering statuses
TryRemoveDrunkenessTime has been renamed to TryRemoveDrunkennessTime to fix the typo.

**Changelog**

no cl no fun


Content.Shared/Drunk/DrunkStatusEffectComponent.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Content.Shared.Drunk;
+
+/// <summary>
+/// This is used by a status effect entity to apply the <see cref="DrunkComponent"/> to an entity.
+/// </summary>
+[RegisterComponent]
+public sealed partial class DrunkStatusEffectComponent : Component
+{
+}
```
EmoGarbage404: needs `NetworkedComponent`. also i think you can remove the class body (the `{}`)

Content.Server/Speech/EntitySystems/SlurredSystem.cs
```diff
@@ -40,12 +40,19 @@ public override void DoSlur(EntityUid uid, TimeSpan time, StatusEffectsComponent
...
+        // This is a magic number. Why this value? No clue it was made 3 years before I refactored this.
+        var magic = SharedDrunkSystem.MagicNumber;
+
+        if (time.Item2 != null)
+        {
+            var curTime = _timing.CurTime;
+            magic = (float) (time.Item2 - curTime).Value.TotalSeconds - 80f;
+        }
+
+        return Math.Clamp(magic / 1100, 0f, 1f);
```
EmoGarbage404:
```suggestion
        return Math.Clamp(magic / SharedDrunkSystem.MagicNumber, 0f, 1f);
```

at least reuse your magic number

Content.Shared/Drunk/SharedDrunkSystem.cs
```diff
@@ -0,0 +1,60 @@
...
+        SubscribeLocalEvent<DrunkStatusEffectComponent, StatusEffectRemovedEvent>(OnStatusRemoved);
+    }
+
+    public void TryApplyDrunkenness(EntityUid uid,
+        TimeSpan boozePower,
+        bool applySlur = true)
+    {
+        // TODO: this should probably raise an event.
+        if (TryComp<LightweightDrunkComponent>(uid, out var trait))
+            boozePower *= trait.BoozeStrengthMultiplier;
```
EmoGarbage404: :godo:

it's only one event

Content.Shared/Drunk/SharedDrunkSystem.cs
```diff
@@ -0,0 +1,60 @@
+using Content.Shared.Speech.EntitySystems;
+using Content.Shared.StatusEffectNew;
+using Content.Shared.Traits.Assorted;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.Drunk;
+
+public abstract class SharedDrunkSystem : EntitySystem
```
EmoGarbage404, line 8: i don't really understand why this system exists. All it seems to do is call StatusEffectsSystem functions. At least with the old API, managing a singleton kinda makes sense for a system: i don't get why you'd still have it now.

Princess-Cheeseballs: I kind of wish that the new status effects had a way to just define a comp to add on creation and remove on deletion in yaml since a lot of statuses are just: On init, EnsureComp, On die, RemComp.

EmoGarbage404: that's largely vestigial since the old status effect system literally could only apply components. The issue with that however is that it breaks hard on stacking and thus is not an acceptable solution.

Princess-Cheeseballs: I can probably just kill the DrunkComponent honestly, not really much stopping me from having the system relay to StatusEffects. 

TryApplyDrunknness is still necessary for lightweight drinking to work unless I put a TryComp in the entity effect and call it a day which I'm considering doing.

Princess-Cheeseballs: I replaced the EnsureComps with relays for all the statuses to take advantage of the new system, Drunkenness needs some input. I could either hardcode the LightweightDrinker check into the Drunk.cs and kill the DrunkSystem basically as a whole, or we could leave it as is and accept that it sucks.

Princess-Cheeseballs:
Just noticed bot is being funny and has requested changes tag still on even though changes were made.

PR #38679, opened 2025-06-30
Fix spawnpoints being ignored if arrivals is disabled; fix cryosleep spawn preference always being ignored

### LINKED ISSUE #31753: New joins are not spawning in on cryosleep.
## Description
New joins are not spawning in on cryosleep. Always goes to terminal.

**Reproduction**
- set cyrosleep as priority
- join game

**Screenshots**
NA

**Additional context**
NA

### LINKED ISSUE #38418: Disabling shuttle arrivals breaks late-join spawn points
## Description

It seems that if arrivals is disabled via cvar and a map has no late join spawn points (that is, only has CryogenicSleepUnitSpawnerLateJoin), then ContainerSpawnPointSystem never tries to find a spawn point handing it off to SpawnPointSystem which will fallback to picking a random-ish spawn point. (It's not actually random, it's the first spawn point it has—that's a separate bug, though.)

**Reproduction**

1. Run in any release configuration.
2. `> cvar shuttle.arrivals false`
3. `golobby` if needed.
4. `forcemap Amber`, `startround`
5. Set your spawn priority to "none" or "arrivals". Late join as anything but atmos.
6. Get spawned in atmos locker room every time.

I was able to get a second person to reproduce this.

**Screenshots**

**Additional context**

ContainerSpawnPointSystem is bailing because it has a check that 1) if your spawn priority isn't set to cryo, and 2) your JobEntity is unset, then bail. This doesn't make sense to me though—JobEntity is meant to be used for things like AI and borgs, so in most cases it is unset. 

https://github.com/space-wizards/space-station-14/blob/93c7bdc1345dd3b76727a386fe7ce30e6edd28b2/Content.Server/Spawners/EntitySystems/ContainerSpawnPointSystem.cs#L33-L38

This seems to have been introduced in #32058, so CC @metalgearsloth.

perryprog:
## About the PR

This reorders spawning logic slightly and tweaks a few of its checks to properly handle arrivals being disabled.

The behavior is now like this:

<table><thead>
  <tr>
    <th rowspan="2">Priority</th>
    <th colspan="2">Arrivals enabled</th>
    <th colspan="2">Arrivals disabled</th>
  </tr>
  <tr>
    <th>Round start</th>
    <th>Late join</th>
    <th>Round start</th>
    <th>Late join</th>
  </tr></thead>
<tbody>
  <tr>
    <td>None</td>
    <td>Spawn point</td>
    <td>Arrivals</td>
    <td>Spawn point</td>
    <td>Cryosleep</td>
  </tr>
  <tr>
    <td>Arrivals</td>
    <td>Spawn point</td>
    <td>Arrivals</td>
    <td>Spawn point</td>
    <td>Cryosleep</td>
  </tr>
  <tr>
    <td>Cryosleep</td>
    <td>Spawn point</td>
    <td>Cryosleep</td>
    <td>Spawn point</td>
    <td>Cryosleep</td>
  </tr>
</tbody>
</table>

## Why / Balance

Fixes #38418, fixes #31753.

## Technical details

I've tested this as thoroughly as I can but this whole system super duper badly needs integration tests, especially if GameTicker is ever going to be refactored.

I've tried every combination listed in the table above, all with nukeops going, and didn't run into any issues.

I also introduced a CanSpawnInArrivals prototype flag which is exclusively false on AI. The previous setup had arrivals logic run after container spawner logic, with container spawners attempting to spawn anything with a set JobEntity regardless of job preference. (This is why AI always spawned correctly.) Borgs, on the other hand, could latejoin  and fail to spawn in any cryotubes, before then being handled by arrivals. I chose to keep this behavior, so this required a new flag. (This makes more sense than having the flag be implied by JobEntity being set anyway.)

Cryosleep as a spawn preference was just being totally ignored, and the logic change at the top of container spawner system fixes that.

This also slightly changes the final fallback logic of picking a random spawn point if nothing else worked—it was clearly meant to be implemented as a random pick between all spawn points it could find, but it was actually stopping after it found a single spawn point. I also added a check that the spawn point is actually part of the target station, just in caseies™.

Anyway; this area of code is absurdly fragile, please merge this cautiously!

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Cryosleep as a spawn preference is now respected.

### DIFF
--- a/Content.Server/Shuttles/Systems/ArrivalsSystem.cs
+++ b/Content.Server/Shuttles/Systems/ArrivalsSystem.cs
@@ -23,6 +23,7 @@ using Content.Shared.GameTicking;
 using Content.Shared.Mobs.Components;
 using Content.Shared.Movement.Components;
 using Content.Shared.Parallax.Biomes;
+using Content.Shared.Preferences;
 using Content.Shared.Salvage;
 using Content.Shared.Shuttles.Components;
 using Content.Shared.Tiles;
@@ -93,7 +94,8 @@ public sealed class ArrivalsSystem : EntitySystem
     {
         base.Initialize();
 
-        SubscribeLocalEvent<PlayerSpawningEvent>(HandlePlayerSpawning, before: new []{ typeof(SpawnPointSystem)}, after: new [] { typeof(ContainerSpawnPointSystem)});
+        // Order is arrivals → container spawn (cryosleep) → spawn points (includes fallback)
+        SubscribeLocalEvent<PlayerSpawningEvent>(HandlePlayerSpawning, before: [typeof(SpawnPointSystem)]);
 
         SubscribeLocalEvent<StationArrivalsComponent, StationPostInitEvent>(OnStationPostInit);
 
@@ -333,16 +335,23 @@ public sealed class ArrivalsSystem : EntitySystem
 
     public void HandlePlayerSpawning(PlayerSpawningEvent ev)
     {
+        // Someone already set a spawn result
         if (ev.SpawnResult != null)
             return;
 
-        // We use arrivals as the default spawn so don't check for job prio.
+        // Arrivals is checked as long as the spawn preference isn't explicitly set to cryosleep
+        if (ev.HumanoidCharacterProfile?.SpawnPriority == SpawnPriorityPreference.Cryosleep)
+            return;
 
         // Only works on latejoin even if enabled.
         if (!Enabled || _ticker.RunLevel != GameRunLevel.InRound)
             return;
 
         if (!HasComp<StationArrivalsComponent>(ev.Station))
+            return;
+
+        // If you aren't allowed in arrivals, begone.
+        if (!_protoManager.TryIndex(ev.Job, out var jobProto) || !jobProto.CanSpawnInArrivals)
             return;
 
         TryGetArrivals(out var arrivals);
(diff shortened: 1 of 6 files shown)


perryprog:
Lol, test fail super real. I just realized when I ran tests locally I did it on the wrong branch.

perryprog:
It's because empty.yml's spawn is technically off grid... 🤦 
image

UpAndLeaves:
Could you clarify, that if you round-start join with preference set to Cryosleep, which of the following will happen:
1. You will spawn at cryosleep, no matter which type.
2. If Job Spawn Cryosleep pods are mapped, you will spawn at them. Otherwise, you will spawn at the job spawn point.

Also could the None preference be removed as it is, according to this, identical to the Arrivals preference?

perryprog:
Ugh, that's super annoying—I don't know how I got this wrong?? To correct myself: At round-start with cryosleep set as your spawn preference, you will always spawn at a job spawn point (not a cryosleep pod (unless the spawn point overlaps with the cryopod :P)). I'll update the PR description in a second, just not at all sure how I wrote that down incorrectly. Regardless—thanks for clarifying and helping me catch that.

Since spawn preferences are stored in the database, I'd rather not touch that, at least not in this PR. I feel like even a "safe" removal of just not displaying that option to users has a lot of edge cases that I don't want to worry about here.

AJCM-git:
@perryprog are you coming back to this? is it still relevant?

perryprog:
It’s prolly still relevant but I’m not planning to come back to SS14 stuff at the moment (sorry!).

PR #38681, opened 2025-07-01
fix water coolers

slarticodefast:
## About the PR
You could not grab any water cups from them sometimes.

## Why / Balance
bugfix

## Technical details
It was adding all entities inserted into any container into the queue of entities that can be grabbed via interaction.
However the water reagent inside is a solution entity, which gets spawned on map init by SharedSolutionContainerSystem and then inserted into a container and thus got added to the queue as well. So you were trying to pick up the water entity in the tank with your hands which was failing. To fix it we check that the entities are inserted into the correct container.

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: Fixed water cups not being able to be picked up from water coolers.

### DIFF
--- a/Content.Shared/Storage/EntitySystems/BinSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/BinSystem.cs
@@ -67,12 +67,18 @@ public sealed class BinSystem : EntitySystem
 
     private void OnEntInserted(Entity<BinComponent> ent, ref EntInsertedIntoContainerMessage args)
     {
+        if (args.Container.ID != ent.Comp.ContainerId)
+            return;
+
         ent.Comp.Items.Add(args.Entity);
     }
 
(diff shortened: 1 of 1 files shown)


Hanzdegloker:
so, not too well known but you can put water cups back on the cooler after you finish drinking from them. Besdies being insanely unsanitary, is that intended to be able to be done?

slarticodefast:
Yeah, that's intended so you can restock it
grafik

Hanzdegloker:
> Yeah, that's intended so you can restock it
> grafik
> 

Fantastic, twas just curious 

I will then continue drinking from the cups and putting them right back in the stack for the next crewmember

ArtisticRoomba:
Probably nobody knows this but secretly my character puts the little water cones back after drinking from them. Smile.

PR #37702, opened 2025-05-21
Catchable items, playable basketball

slarticodefast:
## About the PR
Adds a component that makes it possible to catch items that are thrown at you and adds it to balls.
Adds a clumsiness interaction making you take a small amount of damage from failing to catch something.

## Why / Balance
Now you can actually play basketball on the basketball shuttle.

## Technical details
Added a simple `CatchableComponent` and corresponding system.

On a thrown item collision check if the target has free hands, is in combat mode (if required), passes an optional whitelist, and decide if you catch the item with a certain chance. To make the randomness not mispredict we set the seed depending on the current game tick and the entity's net id. Ideally this will be later replaced by https://github.com/space-wizards/RobustToolbox/pull/5849

I applied the same randomness prediction trick to `ClumsySystem`, which fixes https://github.com/space-wizards/space-station-14/issues/38538 and resolves https://github.com/space-wizards/space-station-14/issues/38369

Added the `CatchableComponent` to the basketball, football and beach ball and adjust their fixtures to be better.

## Media
example

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes
Moved and renamed clumsiness related Loc strings.

**Changelog**
:cl:
- add: You can now catch basketballs, footballs and beach balls that are thrown at you!

PR #38683, opened 2025-07-01
Bloodstream to Shared

Princess-Cheeseballs:
## About the PR

Title, moved BloodstreamSystem and BloodstreamComponent to shared

## Why / Balance

Useful for making medical better

## Technical details

Moved BloodstreamSystem and BloodstreamComponent to shared, renamed BloodstreamSystem to SharedBloodstreamSystem

## Requirements

## Breaking changes

BloodstreamSystem and BloodstreamComponent have been moved to shared
BloodstreamSystem has been renamed to SharedBloodstreamSystem

**Changelog**


ScarKy0:
Slarti was working on predicting bloodstream maybe you should consult them

Princess-Cheeseballs:
> Slarti was working on predicting bloodstream maybe you should consult them

I'll give em a ping on discord then

slarticodefast:
Yeah, I got a branch where I did the same but I ran into major prediction problems, in particular with the bleeding alert. I'll check out your solution later.

Princess-Cheeseballs:
> Yeah, I got a branch where I did the same but I ran into major prediction problems, in particular with the bleeding alert. I'll check out your solution later.

I'll test the alert a bit to see if I can replicate the prediction issues.

Princess-Cheeseballs:
> Yeah, I got a branch where I did the same but I ran into major prediction problems, in particular with the bleeding alert. I'll check out your solution later.

Yeah I'm getting issues with the bleeding alert not predicting properly. I'm gonna look into it tomorrow. If you have any tips please share them.

Edit: Didn't sleep, on client it appears to be receiving the DamageChangedEvent twice (excluding prediction) which is strange.

Edit2: One frame after the initial melee hit, it receives it again and takes the damage again but then it corrects itself.

Princess-Cheeseballs:
Fixed the prediction issues, although not sure if my solution was the best.

I just had the OnDamageChanged return if the game is applying a state from the server to avoid double adding damage, and moved the update loop to shared since the client and server couldn't agree on what time that bleed was to be reduced. 

I also probably fucked something up in the meantime because I should've been asleep 2 hours ago.

Content.Server/Body/Systems/BloodstreamSystem.cs
```diff
@@ -1,108 +1,25 @@
...
-            && args.Name != entity.Comp.ChemicalSolutionName
-            && args.Name != entity.Comp.BloodTemporarySolutionName)
-        {
-            return;
-        }
-
-        OnReactionAttempt(entity, ref args.Event);
     }
 
     public override void Update(float frameTime)
```
slarticodefast, line 25: the update loop should be in shared or this will be networked every single update for every player

Content.Shared/Body/Components/BloodstreamComponent.cs
```diff
@@ -41,52 +41,52 @@ public sealed partial class BloodstreamComponent : Component
         /// <remarks>
         ///     This generally corresponds to an amount of damage and can't go above 100.
         /// </remarks>
-        [ViewVariables(VVAccess.ReadWrite)]
-        public float BleedAmount;
+        [DataField, AutoNetworkedField]
+        public FixedPoint2 BleedAmount;
```
slarticodefast, line 45: the bleed reduction is still a float

Content.Shared/Body/Systems/SharedBloodstreamSystem.cs
```diff
@@ -0,0 +1,441 @@
...
+        var totalFloat = total.Float();
+        TryModifyBleedAmount(ent!, totalFloat);
+
+        /*
+        ///     Critical hit. Causes target to lose blood, using the bleed rate modifier of the weapon, currently divided by 5
+        ///     The crit chance is currently the bleed rate modifier divided by 25.
+        ///     Higher damage weapons have a higher chance to crit!
+        */
+        var prob = Math.Clamp(totalFloat / 25, 0, 1);
+        if (totalFloat > 0 && _robustRandom.Prob(prob))
```
slarticodefast, line 145: This will mispredict because server and client are rolling different random numbers.
Ideally we would use the RandomPredicted engine PR, but we can use the workaround I used in this PR in the meantime
https://github.com/space-wizards/space-station-14/pull/37702

slarticodefast (changes requested):
In the future actually test your changes in-game before making a PR according to our PR guidelines. This is why we require media to be added.
I assume you moved the update loop back to the server because of the solution resolve  debug asserts it was causing, but it should be in shared so it's properly predicted and the correct fix is the same as in this PR https://github.com/space-wizards/space-station-14/pull/35314
Also `DamageChangedEvent` is currently being raised with the wrong sign for the damage value on the client, I got a fix for that in my branch where I tired predicting BloodstreamSystem as well.
https://github.com/space-wizards/space-station-14/compare/master...slarticodefast:space-station-14:predict-healing
I would prefer if we use that one as a basis because it already has my review applied and also predicts topicals. I temporarily commented out critical hits for more consistent debugging.
And it turns out I just missed a dirty call, so now it seems to work properly in my branch. Maybe you can give it a test as well.

Princess-Cheeseballs:
> In the future please actually test your changes in-game before making a PR according to our PR guidelines. This is why we require media to be added. I assume you moved the update loop back to the server because of the solution resolve debug asserts it was causing, but it should be in shared so it's properly predicted and the correct fix is the same as in this PR #35314 Also `DamageChangedEvent` is currently being raised with the wrong sign for the damage value on the client, I got a fix for that in my branch where I tired predicting BloodstreamSystem as well. [master...slarticodefast:space-station-14:predict-healing](https://github.com/space-wizards/space-station-14/compare/master...slarticodefast:space-station-14:predict-healing) I would prefer if we use that one as a basis because it already has my review applied and also predicts topicals. I temporarily commented out critical hits for more consistent debugging. And it turns out I just missed a dirty call, so now it seems to work properly in my branch. Maybe you can give it a test as well.

I actually moved it to server because client wasn't able to predict properly when bleed reduction would happen during the update loop when it was in shared. I tested your branch and it's an issue there as well although your approach to basically everything is better than mine.  

I can either merge your branch into mine and try and solve update loop prediction in shared, move bleed to fixed point 2 and change the RobustRandom to take the current time as the seed, or just close this PR and leave it to you. Whichever you prefer.

Princess-Cheeseballs:
Superseded by #38690

PR #38684, opened 2025-07-01
Convert `ProjectileComponent` to use `WeakEntityReference`

Tayrtahn:
## About the PR

The `Shooter` and `Weapon` fields of `ProjectileComponent` (used to track the mob and gun that fired them) are now `WeakEntityReference`s instead of `EntityUid?`s.

Requires https://github.com/space-wizards/RobustToolbox/pull/5577

## Why / Balance / Technical details

This is intended as a test run and demo of `WeakEntityReference`. The intention is to find anything that might be lacking in the API and to show how it's useful for this type of situation. It's also meant to make sure that I'm using the API in the correct way!

This PR includes a [test](https://github.com/space-wizards/space-station-14/commit/4f350df8cc1d42b4b1f7775656c154fd6980af43) that demonstrates a bug currently in master - if a projectile is fired and the entity that fired it is deleted, the component state will become invalid and log errors whenever it is serialized. The test spawns entities to use as a gun and a bullet, then fires the bullet from the gun and deletes the gun. Forcing a PVS update makes the server attempt to serialize the bullet entity, which is now contains a reference to a deleted entity in its `Weapon` field. This causes an error and fails the test.

In some similar cases, this situation has been resolved by adding two-way tracking - for example, followers have a reference to the entity they are following, and followed entities keep track of everything currently following them, so that if either is deleted, they can inform the other to stop tracking. This works well enough for that case, but in this instance would require guns to manage references to every projectile they fire, which is clunky and a lot of overhead. `WeakEntityReference` makes this much easier to deal with, as it allows the projectile to store a reference to the gun while telling the serializer that it's fine if it has been deleted.

The rest of the PR converts the `Shooter` and `Weapon` fields to `WeakEntityReference`s and makes the needed changes to other systems to support this - either wrapping assignments to these fields in `WeakEntityReference`s or using `Resolve` to convert them into usable `EntityUid`s.

## Requirements

## Breaking changes

Remind me to fill this out when undrafting this.

**Changelog**

Not player-facing.

### DIFF
--- a/Content.Client/Weapons/Ranged/Systems/FlyBySoundSystem.cs
+++ b/Content.Client/Weapons/Ranged/Systems/FlyBySoundSystem.cs
@@ -30,7 +30,7 @@ public sealed class FlyBySoundSystem : SharedFlyBySoundSystem
         if (attachedEnt == null ||
             args.OtherEntity != attachedEnt ||
             TryComp<ProjectileComponent>(uid, out var projectile) &&
-            projectile.Shooter == attachedEnt)
+            Resolve(ref projectile.Shooter) == attachedEnt)
         {
             return;
         }
--- a/Content.IntegrationTests/Tests/Weapons/Guns/GunNetworkingTest.cs
+++ b/Content.IntegrationTests/Tests/Weapons/Guns/GunNetworkingTest.cs
@@ -1,0 +1,46 @@
+using System.Numerics;
+using Content.Shared.Weapons.Ranged.Systems;
+using Robust.Server.Player;
+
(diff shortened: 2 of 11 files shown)


slarticodefast:
I have watched grafana regularly recently to see what errors are popping up and most of them are from a multitude of systems trying to network uids from deleted entities. So we will have to do this on a large scale somehow. Either go through all components one by one or come up with an integration test that randomly deletes entities in different scenarios. Or just write an analyzer that warns you about networked EntityUids that are not weak references, even though some systems are handling them correctly.

slarticodefast:
I'm closing this for now until we have a new engine PR for this.

PR #13120, opened 2022-12-20
Have crew monitor display entity coordinates instead

keronshb:
## About the PR 

closes #13042 

Crew monitors now display entity coordinates instead of map coordinates

**Media**

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl:

PR #38685, opened 2025-07-01
Make GPS able to show both Map and Entity coords

Centronias:
## About the PR

Handheld GPS can toggle between "absolute" and "relative" mods, which are MapCoords and EntityCoords (relative to parent grid) respectively.
Oh, and there's a verb to toggle mode for use on PDAs with astronav.

Also added localization to GPS ~~and Crewmon~~.

## Why / Balance

Removes confusion; I was in a round recently where the warden and AI were feeding crewmon coords to some secoffs who were offstation looking for some hostages. When the secoffs got to the coords, they found nothing because -- oops! -- the coordinates are not the same between the two mechanics!

## Technical details

- Add toggleable mode to GPS, based kinda on the Multitool's toggleable mode
- Changed the code which provides the markup for GPS examine and GPS in-hand-title thingy to:
  - use localization
  - show the mode
  - change how the coordinate values are calculated based on mode (code copied and then massaged from Crewmon code for "relative" mode)
  - to actually use the same code

## Media

https://github.com/user-attachments/assets/1c592f7d-e5e0-4a53-8a78-7134a3e31bb8

image
verb

## Requirements

## Breaking changes

Shouldn't be any, purely addative changes to public interfaces.

**Changelog**

:cl:
- add: Handheld GPS and PDAs with Astronav are now able to toggle between absolute and relative coordinate modes. Relative coordinates are the same as what's displayed on the crew monitor, absolute coordinates are the previous behavior.


Content.Client/Medical/CrewMonitoring/CrewMonitoringNavMapControl.cs
```diff
@@ -64,7 +68,8 @@ protected override void FrameUpdate(FrameEventArgs args)
             if (!LocalizedNames.TryGetValue(netEntity, out var name))
                 name = "Unknown";
 
-            var message = name + "\nLocation: [x = " + MathF.Round(blip.Coordinates.X) + ", y = " + MathF.Round(blip.Coordinates.Y) + "]";
+            var mapCoords = _transform.ToMapCoordinates(blip.Coordinates);
+            var message = name + "\nLocation: [x = " + MathF.Floor(mapCoords.X) + ", y = " + MathF.Floor(mapCoords.Y) + "]";
```
slarticodefast: I know this was like this before, but this really should be localized

Centronias:
@slarticodefast , lemme know if this needs anything else. I do hope this can be a quick merge and not get lost among the great sea of PRs 🫡

slarticodefast:
The code looks good, but conceptually I'm not sure if this is the right approach. Having the coordinates be the same as the gps is useful for searching for someone, but it was converted to entity coordinates in https://github.com/space-wizards/space-station-14/pull/13120 for a reason. If the station is rotated by 45° for example it makes them really unintuitive as the unit axis are now diagonal on the monitor, whereas entity coordinates relative to the grid directly correspond to what you see. Maybe the gps should get a mode for local coords instead?

Centronias:
Ah, very good point. Any opposition to both having both?

Centronias:
@slarticodefast , changed to have dual-coord ability on GPS, removed the crewmon functional changes.
I kept the cleanup and localization for crewmon. Lemme know if you want me to just revert all of that, or just revert the cleanup or whatever, I don't mind.

PR #38686, opened 2025-07-01
Monochromacy typo fix

imatsoup:
## About the PR
Fixes a typo with the Monochromacy trait.

## Why / Balance
I saw this while working on a different problem and it was going to bother me forever if I didn't pr this. Incidentally, this should allow for clones to properly copy over the trait (it was causing an error looking for Monochromacy when the id before was 'Monochromancy').

## Technical details
Rename the trait's id in disabilities.yml to Monochromacy, update the trait name and description vars in traits.ftl. clone.yml now tries to copy the BlackAndWhiteOverlay component instead of a trait.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Player/clone.yml
+++ b/Resources/Prototypes/Entities/Mobs/Player/clone.yml
@@ -16,8 +16,8 @@
   - NpcFactionMember
   # traits
   # - LegsParalyzed (you get healed)
-  - LightweightDrunk
   - BlackAndWhiteOverlay
+  - LightweightDrunk
   - Muted
   - Narcolepsy
   - Pacified


slarticodefast:
Good catch with the typo, but cloning does not check for traits, it checks for components. So this part in `clone.yml` should really be `BlackAndWhiteOverlay` instead. Could you fix that as well?
grafik

imatsoup:
> Good catch with the typo, but cloning does not check for traits, it checks for components. So this part in `clone.yml` should really be `BlackAndWhiteOverlay` instead. Could you fix that as well? grafik

Yep, just pushed this change. Thanks.

Resources/Prototypes/Entities/Mobs/Player/clone.yml
```diff
@@ -17,7 +17,7 @@
   # traits
   # - LegsParalyzed (you get healed)
   - LightweightDrunk
```
slarticodefast:
```suggestion
  - BlackAndWhiteOverlay
  - LightweightDrunk
```
let's keep them sorted alphabetically

Resources/Prototypes/Entities/Mobs/Player/clone.yml
```diff
@@ -17,7 +17,7 @@
...
-  - Monochromacy
+  - BlackAndWhiteOverlay
```
slarticodefast, line 19:
```suggestion
```

slarticodefast (approved):
Thanks!

PR #38693, opened 2025-07-01
Security can now see how long someone has been handcuffed for

SpeltIncorrectyl:
## About the PR
If you are wearing secglasses or a sechud, you can examine a cuffed person to see how long they have been cuffed for.
Only works for handcuffs and zipties (stuff sec will use), not makeshift cuffs.

## Why / Balance
When sec arrests someone and gives them brig time they are meant to factor in how long someone has already been detained for.
But no one has ever bothered to check this ever and even if they do factor it in they go off vibes.
Being able to examine someone you have in custody and see exactly how long they have spent in custody will be a great QoL feature for sec!

## Technical details
* HandcuffComponent now has field for ShowCuffedTime, set to true on handcuffs and zipties.
* CuffableComponent now has a nullable field called CuffedTime which is set to the current time when the first pair of of cuffs are applied and set to null when the final pair of cuffs is removed.

> (the Cuffable system has legacy code from when arachnids had 4 arms so it has to deal with tons of edge cases which will never happen, like more than one pair of cuffs. I think Slarti or Emo said that if they give arachnids back their 4 arms then they will refactor it so one pair of cuffs will cuff any number of arms)

* there is ShowCuffedTimeComponent on secglasses and sec huds, if you have it on your entity or you are wearing something with it on your eyes then it lets you see the cuff time
* as long as you have the right hud and atleast one of the cuffs has ShowCuffedTime set to true then you will see the examine text which calculates cuffed time by comparing the CuffedTime field with the current time (has to factor for multiple handcuffs despite this not being possible and never gonna happen :pensive:)

## Media

Screenshot from 2025-07-01 23-24-52
Screenshot from 2025-07-01 22-50-58

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Security can now see how long someone has been handcuffed for.

### DIFF
--- a/Content.Client/Cuffs/CuffableSystem.cs
+++ b/Content.Client/Cuffs/CuffableSystem.cs
@@ -33,6 +33,7 @@ public sealed class CuffableSystem : SharedCuffableSystem
             return;
 
         component.CanStillInteract = cuffState.CanStillInteract;
+        component.CuffedTime = cuffState.CuffedTime;
         _actionBlocker.UpdateCanMove(uid);
 
         var ev = new CuffedStateChangeEvent();
--- a/Content.Server/Cuffs/CuffableSystem.cs
+++ b/Content.Server/Cuffs/CuffableSystem.cs
@@ -29,7 +29,8 @@ namespace Content.Server.Cuffs
                 component.CanStillInteract,
                 cuffs?.CuffedRSI,
                 $"{cuffs?.BodyIconState}-{component.CuffedHandCount}",
-                cuffs?.Color);
+                cuffs?.Color,
+                component.CuffedTime);
             // the iconstate is formatted as blah-2, blah-4, blah-6, etc.
             // the number corresponds to how many hands are cuffed.
         }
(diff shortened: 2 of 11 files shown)


retequizzle:
i'm gonna be blunt i've given up on actually weighing in on PRs because in a lot of situations it feels like wasted effort, but i got told to voice my point on the PR

so against my better judgment i'm going to just copypaste my points from discord into this, you all can do what you want with it:

- i get why that's a logical thing to add but i still hate how reliant the game's becoming on hud elements to tell people stuff (this is nothing against the PR author, because i can see a use case for this in other scenarios)
- a lot of it just feels like extra fluff because people don't like abstract approximations of a situation and want to see exact numbers, exact details, exact everything
- details like this should be discovered via interacting with other players, not just shiftclicking them and getting an exact summary of everything they've been through int he last 10 minutes (example: like you should be able to lie to sec as a detainee about how long you've been cuffed for)
- space station as a concept was designed around social engineering and paranoia about who is who and who isn't what they claim to be, introducing all these extra elements pretty much eliminates the uncertainty that comes with 90% of it

EmoGarbage404:
i agree with ret: well put.

K-Dynamic:
I feel this would've been more useful prior to genpop

slarticodefast:
Yeah, I agree with Ret as well.
Usually small QoL additions like this are a good thing, but this can also lead to visual bloat in the examination window.
And while a high-tech handcuff could in theory have a build-in timer this still feels a little off to me mechanically.

PR #38696, opened 2025-07-02
LOOC Modifications

EmoGarbage404:
## About the PR

Makes LOOC a proper out-of-character channel, making it no longer appear above player's heads or use their in-game names.

## Why / Balance

LOOC is notoriously an IC malding channel to the point where it is painfully detrimental to any attempt to meaningfully RP in game. Rather than being used for actual light OOC chatting (i had some good dinner today, the weather has been nice, my gf just left me, etc.), it instead is almost exclusively used for either complaining about rule breaks, in-game changes, or other players.

This plainly is just frustrating as a player. Its existence basically just perpetuates toxicity which makes it extremely annoying to deal with.

This is due to the fact that it largely just resembles local chat. Rather than normal OOC chat, which uses your username and doesn't appear tied to a character, LOOC both uses your IC name (which doesn't make sense) but also appears above the player's head, further cementing it as an IC char--something it explicitly is not. This is to a point where i've heard people say that it actually refers to "light" OOC chat.

My goal is that by removing these more IC features, it'll push players to not use it for that type of discussion and ideally resolve things IC more.

## Technical details

basically just straight removal and very minor changing in some of the chat stuff.

## Media

image

## Requirements

## Breaking changes

Removed `SpeechBubble.SpeechType.Looc` and `LoocAboveHeadShow` CVAR.

**Changelog**
:cl:
- tweak: LOOC chat now uses your username rather than your character name. Additionally, it no longer appears above a character's head when they speak.

### DIFF
--- a/Content.Client/Chat/UI/SpeechBubble.cs
+++ b/Content.Client/Chat/UI/SpeechBubble.cs
@@ -25,7 +25,6 @@ namespace Content.Client.Chat.UI
             Emote,
             Say,
             Whisper,
-            Looc
         }
 
         /// <summary>
@@ -76,9 +75,6 @@ namespace Content.Client.Chat.UI
 
                 case SpeechType.Whisper:
                     return new FancyTextSpeechBubble(message, senderEntity, "whisperBox");
-
-                case SpeechType.Looc:
-                    return new TextSpeechBubble(message, senderEntity, "emoteBox", Color.FromHex("#48d1cc"));
 
                 default:
                     throw new ArgumentOutOfRangeException();
--- a/Content.Client/Options/UI/Tabs/MiscTab.xaml.cs
+++ b/Content.Client/Options/UI/Tabs/MiscTab.xaml.cs
@@ -45,7 +45,6 @@ public sealed partial class MiscTab : Control
 
         Control.AddOptionCheckBox(CVars.DiscordEnabled, DiscordRich);
         Control.AddOptionCheckBox(CCVars.ShowOocPatronColor, ShowOocPatronColor);
-        Control.AddOptionCheckBox(CCVars.LoocAboveHeadShow, ShowLoocAboveHeadCheckBox);
         Control.AddOptionCheckBox(CCVars.HudHeldItemShow, ShowHeldItemCheckBox);
         Control.AddOptionCheckBox(CCVars.CombatModeIndicatorsPointShow, ShowCombatModeIndicatorsCheckBox);
         Control.AddOptionCheckBox(CCVars.OpaqueStorageWindow, OpaqueStorageWindowCheckBox);
--- a/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs
+++ b/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs
@@ -890,11 +890,6 @@ public sealed partial class ChatUIController : UIController
             case ChatChannel.Emotes:
                 AddSpeechBubble(msg, SpeechBubble.SpeechType.Emote);
                 break;
-
-            case ChatChannel.LOOC:
-                if (_config.GetCVar(CCVars.LoocAboveHeadShow))
-                    AddSpeechBubble(msg, SpeechBubble.SpeechType.Looc);
-                break;
         }
     }
 
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -619,7 +619,7 @@ public sealed partial class ChatSystem : SharedChatSystem
             return;
 
         var wrappedMessage = Loc.GetString("chat-manager-entity-looc-wrap-message",
-            ("entityName", name),
+            ("entityName", player.Name),
             ("message", FormattedMessage.EscapeText(message)));
 
         SendInVoiceRange(ChatChannel.LOOC, message, wrappedMessage, source, hideChat ? ChatTransmitRange.HideChat : ChatTransmitRange.Normal, player.UserId);
--- a/Content.Shared/CCVar/CCVars.Hud.cs
+++ b/Content.Shared/CCVar/CCVars.Hud.cs
@@ -13,9 +13,6 @@ public sealed partial class CCVars
     public static readonly CVarDef<bool> CombatModeIndicatorsPointShow =
         CVarDef.Create("hud.combat_mode_indicators_point_show", true, CVar.ARCHIVE | CVar.CLIENTONLY);
 
-    public static readonly CVarDef<bool> LoocAboveHeadShow =
-        CVarDef.Create("hud.show_looc_above_head", true, CVar.ARCHIVE | CVar.CLIENTONLY);
-
     public static readonly CVarDef<float> HudHeldItemOffset =
         CVarDef.Create("hud.held_item_offset", 28f, CVar.ARCHIVE | CVar.CLIENTONLY);
 


K-Dynamic:
Rip !edibles, good change otherwise

deathride58:
It'd probably be a good idea to make this behavior a cvar (with a server-side cvar taking priority over the client-side cvar for whether overhead LOOC should be disabled) rather than removing the capability outright from the codebase. 

While it's absolutely correct that LOOC rarely sees good-faith use in lower-RP environments (in LRP, it's almost exclusively used for salt), LOOC *does* see a higher rate of good-faith use (ie: LOOC being used to teach controls or give mechanical explanations to newer players) in higher-roleplay environments, which are more likely to want to keep this feature as-is.

ArcaneOcto:
How do I tell who is speaking with these changes? This also might make things like training new players a little bit harder. Personally, I definately I prefer the way it is. The benefits outweigh the salt imo. I think salty players would just mald in regular chat

Djungelskog2:
> How do I tell who is speaking with these changes? This also might make things like training new players a little bit harder. Personally, I definately I prefer the way it is. The benefits outweigh the salt imo. I think salty players would just mald in regular chat

they just have to read chat

Pumkin69:
I really don't know how i feel about this. I enjoy having conversations with friends in looc or simply saying something that i dont know how to say IN CHARACTER, OUT OF CHARACTER. It would make these extremely hard. Not to mention making teaching new players in game really difficult i genuinely think this is a massive net negative over positive. I get that mald in looc is bad but the amount of good things looc provides is very helpful especially with my point of teaching new players.

ArtisticRoomba:
> How do I tell who is speaking with these changes?

You don't know the character, same case with OOC channels. Except LOOC doesn't go to everyone (hence *Local* Out Of Character).

> This also might make things like training new players a little bit harder.

In my opinion, the learning curve is the same. Little Passenger will have the same difficulty in knowing what LOOC is whether the message shows above the head of the user or not.

> The benefits outweigh the salt imo. I think salty players would just mald in regular chat

LOOC has become the de-facto """acceptable""" chat to say literally anything in right now. This is coming from a long time Mander player, 95% of players do not use this channel properly and it largely leads to the instances described in the PR's  justification.

In general SS14 doesn't have very good roleplay standards compared to SS13, to the point where OOC as a channel is disabled until round end (though notably we do not have OOC/LOOC channel bans right now). In a good world, people abusing LOOC/OOC would just be automatically shunned for fail RP - right now it is casually used to freely disagree with a prison sentence and insult the officer tending to you.

> I enjoy having conversations with friends in looc

This won't really be impacted by the changes.

> simply saying something that i dont know how to say IN CHARACTER, OUT OF CHARACTER.

You are saying IC things OOC, which is just not good roleplay and against server rules for nearly any RP server. If you don't know how to say something IC... maybe your character should act stunned or act like they're at a *loss for words*.

Djungelskog2:
> I really don't know how i feel about this. I enjoy having conversations with friends in looc or simply saying something that i dont know how to say IN CHARACTER, OUT OF CHARACTER. It would make these extremely hard. Not to mention making teaching new players in game really difficult i genuinely think this is a massive net negative over positive. I get that mald in looc is bad but the amount of good things looc provides is very helpful especially with my point of teaching new players.

you hold those conversations elsewhere or pre/post-round, teaching new players would essentially be the same

Hitlinemoss:
I don't think this is a good change since it just makes "intended" usage of LOOC (teaching new players, remarking on weird bugs, saying you're going AFK, etc) more confusing, and it doesn't actually prevent people from using LOOC to complain anyway (if anything it might make people complain *more* because of the added layer of pseudo-anonymity).

MidZik:
I don't like losing the context of who is saying the LOOC chat. Everyone should understand that it's not your character saying it. But sometimes you want to give some OOC info or have an OOC question, and knowing which character it's related to is _useful_.

Someone is teaching me something and needs to tell me controls? I'm usually looking at their character, not the side bar. If they LOOC me the controls and it doesn't appear above their character, I'm going to miss it. Even if I _do_ catch the LOOC chat, the person saying it has changed and I have to jump through a pointless mental hoop to connect that the LOOC is related to what I'm being taught.

Spessman123 LOOC says they have to go AFK for a bit? Ok... who is that? It's way more useful to know which character said it than which player said it.

This change feels like it's cutting of the nose to spite the face. It adds more friction to using LOOC in all cases and I hardly think LOOC malding is going to decrease. I don't think they care if it's a "proper out-of-character" channel or not in the first place, to them it's probably just a place to vent about the injustices they are facing either way.

You can add the username to it if you want to make it "feel" more OOC, but I think taking away all the contextual information it currently gives would be a mistake.

whatston3:
Removing a way to express toxicity will not remove the toxicity.  Players will mald regardless, be it in IC or OOC chat - players' behaviour will still be an administrative issue, even if it isn't in moderating LOOC chat.
There's value in being able to follow a conversation, and showing who is sending the message is important to that.  That's not important for global OOC, but for local OOC, it is - if anything, show _both_ the username and the in-character name.

Suppose twenty people are in a room.  Somebody says something in LOOC.  With these changes, how do I even know who said it?

Admiral-Obvious-001:
The point of LOOC is that it's literally not the character saying it.

It's OOC, but Local. It shouldn't matter which character is saying jt, LOOC is not associated with the character, but the player (usually malding that sec searched them on green after they mag dumped the janitor with a Viper).

Chat appearing above the character, while convenient is also in my opinion part of why people are so incredibly oblivious to radio messages as well, but that's a seperate issue.

PicklOH:
As everyone else already said, this doesn't really solve any problems, it only makes things less intuitive for the _good_ uses of LOOC.

People who want to LOOC mald will continue to do so. This doesn't stop them, nor does it make it easier to ignore it. The relevant context behind "man this is such bullshit I only welderbombed why shitsec" is not hidden behind an IC name. If I see that message appear in my _local_ chat I probably know who it's from.

A much more robust solution to this would be LOOC muting. If someone is malding at me, let me mute them. This lets people shun the LOOC malders, as they should, while still enabling wanted discussions.

BigfootBravo:
Yeah, I could see this making it just a lot harder to tell who is saying what, often with important info. Others have also mentioned this, but informing others of controls or how to use different mechanics would be a lot harder if this info was shoved to the chat bar, in an easily missable location where it may be difficult to tell who they are referring to. 
Also as others have mentioned, it would be difficult to relay if you are experiencing a bug, have to go afk, or are asking a question, and it being coming from your character. It not saying who is saying the message in character, and it being only showed in the text log would make it difficult to relay potentially critical info in a timely manner, and would obscure who is saying it. I could even imagine some individuals using this semi-anonymity to OOC impersonate others, like say a prisoner/tider/whoever pretending to speak from a member of security's perspective to paint them in a negative light, which would only be able to be countered if someone is familiar with that player's OOC and IC name, so they can tell who is who.

I most often see LOOC being used to remark on bugs, clarify things, or explain controls or non-diegetic systems; I very rarely (compared to positive interactions) ever saw LOOC being used in negative or obstructive ways.

LOOC's current implementation allows for much good, and while it can allow some negative factors to propagate, its near constant benefits outweigh its occasional misuse. It is easy to use, easy to follow, and easy to comprehend for new and experienced players alike when relaying critical matters, be out-of-game circumstances, rule clarifications when someone is unsure on a matter, general guidance, and more.

Zalycon:
I share the sentiments of what others have said regarding this and how it would make teaching and most "good" uses of LOOC difficult or inconvenient. That is to say I agree with the said sentiments.

retequizzle:
i do think the username replacing the character name is a general step in the right direction towards anonymizing chat messages that may pertain to the round directly (a player asking, "How do I access my uplink?" should not have their IC name attached to it just as a degree of separation from being outed as a not-good-guy by merits of not knowing game mechanics) but i'm admittedly torn on the overhead message above the character issuing the LOOC message, because i can see how that _could_ be a learning aid for teaching new players. maybe a toggle in client settings would be a decent middleground on this - at least this way if a player wants to use LOOC as a teaching tool, they're not outright shafted by doing so.

now that i've spoken to the PR contents itself, i do want to bring up a separate point that may very well undermine this PR's entire premise: OOC's existence and why it's not used within rounds.

we've had to disable OOC from being used during rounds in general (and have had this set for at least the last two years barring an admin enabling it temporarily) **specifically because people abused it in the exact same way that emo outlined in the PR submission**, and the behavior has slowly been cropping up in LOOC with little being done about it because it's not logged and rarely gets ahelped about unless it's something egregious. i don't see how making these changes to LOOC addresses the underlying issue of the community not knowing what is and is not LOOC worthy (spoiler alert: bitching about the round or how someone is playing is never a reason to use LOOC and you're a tool if you do use it for that)

there's probably one of two ways to go about this and neither of them are really all that great - either we log LOOC chats as well and start issuing bans for LOOC abuse (which essentially puts us back where we were two years ago with OOC abuse) or we follow through with the spirit of the PR, and try to make some sort of attempt to separate LOOC chat from IC chat, whether it be a simple change like converting from IC names to usernames, disabling overhead LOOC message popups, or nuclear option of disabling LOOC entirely as well. 

there's also the option where we implement something akin to a whitelist for LOOC usage but frankly if we have to go that far to justify keeping LOOC intact then we lost the plot and should get rid of it anyway.

TrixxedHeart:
I agree 100% with making it use the player's username but also I think having LOOC over the head is important for the many reasons other people have already explained and I don't feel like bears repeating.

I think _just_ having the player's username alone being the one attached more directly gives a sense of more responsibility behind it. People will feel less like it's representing their mask and moreso themselves as a person and be a bit more accountable at the very least, maybe think twice hopefully. If not, people will know who to point and laugh at for being a malder without having to figure it out later.

ArcaneOcto:
I agree with what @TrixxedHeart is saying here. Switching to use the player username is a good feature I think. But removing the chat bubble is not great. The only way to solve the issue of malding is a robust and automatic reporting system. If someone is being a jerk in looc than they should get reported and be timed out from that chat. This PR feels too much like a band-aid to justify the drawbacks people have mentioned in this discussion.

Djungelskog2:
> I agree with what @TrixxedHeart is saying here. Switching to use the player username is a good feature I think. But removing the chat bubble is not great. The only way to solve the issue of malding is a robust and automatic reporting system. If someone is being a jerk in looc than they should get reported and be timed out from that chat. This PR feels too much like a band-aid to justify the drawbacks people have mentioned in this discussion.

some sort of admin help... or... Ahelp for short

BigfootBravo:
> I agree 100% with making it use the player's username but also I think having LOOC over the head is important for the many reasons other people have already explained and I don't feel like bears repeating.

This feels reasonable to me! Detaching it from the char name will make it feel less like an IC method of communication,  while still allowing for ease of use and understanding

Velken:
Just putting my opinion and what i see on Vulture regating how LOOC  is, for the most part, used for:
-Teach players about mechanics they don't know
-Talk about new updates (since not everyone reads the patch notes/follows the github PRs)
-Quick catch up with friends
-Explain to someone why they are/aren't allowed to do something
-Minor complaint about previous round stuff
-Comment about funny moment on previous round that other person has missed.
I'd like to add that, i see all thoses cases are fine for LRP use, since none of these conversations last longer than a few seconds.

LaCumbiaDelCoronavirus:
I like that it uses ooc usernames now but the justification isnt good since you should:

Make an attempt to change rules to solve a rules problem, not an attempt to do anything else unless its proven that previously changing rules did not work. 
And for admins, LOOC chat is already a distinct enough colour from other messages that its easy enough for the messages to not get drowned out, unless multiple people are abusing it, in which case _punish them_

<hr/>

Or even simply make the chat filter (the one in the chatbox where you can select what kinds of messages you want or dont want to hear) have an option for hiding overhead messages too

TrixxedHeart:
> Or even simply make the chat filter (the one in the chatbox where you can select what kinds of messages you want or dont want to hear) have an option for hiding overhead messages too

iirc the "hide LOOC chat bubbles" thing is something that currently exists in-game right now as an option.

slarticodefast:
Hmm, I think this change will be a little problematic. A common usecase of LOOC for me is teaching newer players. This is mostly done IC, but sometimes when they have trouble doing something I switch to LOOC saying something like "press that key to do that". At that point I don't expect them to also keep an eye on the chat window while actively trying out things and looking at my character to read what I'm explaining. The game has a steep learning curve and can be visually hard to parse sometimes. So I would prioritize the new player experience over potential abuse cases of LOOC.

Simyon264:
I have seen LOOC get abused for malding and/or IC content many times and it honestly breaks immersion.

LOOC chat is reserved for Out Of Character and seeing someone mald when getting arrested is not that. 

Using your username makes it more clear that it is not IC. 

As for removing the char bubbles: Chat bubbles will still connnect it with your character, which is what we want to avoid. More so, seeing 3 people converse in LOOC looks bad and encourages abuse.

Simyon264:
For teaching I would rather have mentorhelp implemented and a tutorial showing where LOOC chat shows up would also be preferable.

SlamBamActionman:
1. Make it a server-controlled cvar, it's easy and you'll mess with downstreams who benefit from the current system otherwise.
2. I am willing to try this, but I do not think it will solve LOOC malding. I am probably one of the few who strongly berates anyone for arguing Space Law/rules semantics in LOOC ("STOP TALKING IN LOOC, AHELP IT IF IT IS AN ISSUE"), and I expect those sorts of players are still gonna mald in LOOC with these changes. As long as the person they feel offended by sees it, the delivery mechanism won't matter much.

I have also used LOOC to great effect when teaching new players and I think this PR would be a step back in that regard. Sure, use IC when possible (not that new players really can tell the difference) but sometimes you just gotta explain in LOOC and the current version is really helpful in that regard. Should the new player experience be good enough to not require LOOC? Certainly. But that's a way bigger hill to climb.

I don't have any good solutions for the malding player LOOC issue off-the-cuff, and I do not think this is it. But we can always try it, it may have unexpected results.

SlamBamActionman (changes requested):
Make it a server cvar; doing a Request Changes so this doesn't get merged without it/someone explicitly overrules this request.

Hitlinemoss:
I'm seeing some people here mention the idea of keeping the name-over-head system and just changing it to use your username in chat instead of character name, but frankly I don't think that's a good idea either; usernames should probably be entirely hidden mid-round as a measure against metafriending/metagrudging.

Kowlin:
> 1. Make it a server-controlled cvar, it's easy and you'll mess with downstreams who benefit from the current system otherwise.
> 
>     2. I am willing to try this, but I do not think it will solve LOOC malding. I am probably one of the few who strongly berates anyone for arguing Space Law/rules semantics in LOOC ("STOP TALKING IN LOOC, AHELP IT IF IT IS AN ISSUE"), and I expect those sorts of players are still gonna mald in LOOC with these changes. As long as the person they feel offended by sees it, the delivery mechanism won't matter much.
> 
> 
> I have also used LOOC to great effect when teaching new players and I think this PR would be a step back in that regard. Sure, use IC when possible (not that new players really can tell the difference) but sometimes you just gotta explain in LOOC and the current version is really helpful in that regard. Should the new player experience be good enough to not require LOOC? Certainly. But that's a way bigger hill to climb.
> 
> I don't have any good solutions for the malding player LOOC issue off-the-cuff, and I do not think this is it. But we can always try it, it may have unexpected results.

I am with you on this one. In my **personal** opinion as a game admin, this will not solve the problem of LOOC arguments. I don't think this problem can be mechanically enforced.

superjj18:
This will force people to use IC chat to refer to the chatbox if a new player seems to be confused AND doesn't notice the chat.

"Check the skybox" 
either that or using /me


It's an issue I saw in GTARP a lot when dealing with new players.

PJB3005:
https://forum.spacestation14.com/t/looc-modifications/22004

FairlySadPanda:
Keeping character name as the LOOC name is preferable. Means someone can leak their user if appropriate (e.g. explaining they worked on a PR someone likes - happened to me chatting about scurrets and explaining them on Vulture) but they don't need to otherwise.

Agreed with PJB on not having LOOC as an aside. LOOC and LIC (local) are pretty much the same chat channel and should work similarly except for where LOOC should be accessible for someone when they obviously should not be able to speak in Local. That was IIRC how I did things in my attempt at chatfactor, certainly.

Having username leak into in-simulation logic at all is probably also a domain boundary breach. Usernames are things servers care about, not the game state.

Entvari:
Unsure if this has been stated before as this is a 2 gorrillion comment thread but the act of using usernames instead of IC names allows for metagrudging.
If I hate JohnSpessman123 for doing something I really don't like and I see his default character I am naturally inclined to treat him worse. But now if JohnSpessman123 uses a fresh character to prevent being tied to his reputation and ever chooses to use LOOC, even out of necessity (such as sleep bug), he outs who he is and is now vulnerable to metagrudge once again.
This strips a large chunk of the anonimity of your actions and characters, if someone hates you they don't need to wait until endround summary to complain, they can start the grudging during the match if you ever opt to use LOOC.

This is admittedly a more niche issue but it's a real one nonetheless.

PR #38697, opened 2025-07-02
Centcomm update

Emisse:
yoinky sploinky add new headmin shrines, update niks shrine for retirement


retequizzle:
skub

LaCumbiaDelCoronavirus:
Is this supposed to go somewhere

Emisse:
thats from the last centcomm update
ill fix in a bit

ArtisticRoomba:
turrets in armory or no

Emisse:
no itll make setting up admin events harder

qrwas:
Is it possible to spawn an FTL disk somewhere with the coordinates of the CC and shuttle, so that it would be easier and faster to issue roles to players who will work on the CC?

Emisse:
someone would just need to make a prototype for it and i can map it in

PR #35536, opened 2025-02-26
Allow the Command & Super door remotes to use the access of their user. (Re-creation of PR due to changes to game balance)

BramvanZijp:
Re-doing #26996 because the balance surrounding it has changed with AIs being a thing now.

## About the PR
Added the Custom Door Remote, this version of the command door remote has no access built into the remote, but can effect any door that the user has access to. This functionality is also copied over to the Super Door Remote.

## Why / Balance
This was removed because it made remotes too OP, however now the captains door remote cannot edit most doors while it should be able to. So then somebody made a PR to give caps remote AA, but this was rejected due to making the captain have 2 AA items to loot. This solution takes the best from each, allowing the captain to use his remote alongside his access, meaning he can bolt or emergency access any door so long as he also has his AA ID.

That previous PR got closed primarily due to concerns about a lack of counterplay, but since there is now an AI that can bolt/unbolt everything, I feel like this now only ADDS to counterplay, since the AI counters this, and this counters the AI.

## Technical details
Added a datafield to the DoorRemote component which is false by default, but true for the Custom & Super Remotes.
The code now checks for the access of Args.Used (the remote itself) if the field is false, like it does right now, while checking the access of Args.User (the person using the remote, who has the access of the remote due to holding it but also the access of any potential IDs.) if the field is true, which it only is for the command & super remotes (unless the component is changed in VV.)

## Media
Not much to show.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**
:cl:
- add: Added the custom door remote, it does not have any built-in access, instead inheriting the ID access that the user has.
- tweak: The captain's locker now comes with a custom door remote instead of a command door remote.
ADMIN:
- tweak: The Super Door Remote can now effect doors that the user's ID card has access to, even if the remote itself somehow cannot access them.

PR #38702, opened 2025-07-03
Made all door remotes use the access of the user, instead of having their own access.

RangerXVII:
## About the PR
Adds to #35536 

Removes all door remotes with their own access (except super and xenoborg because I didn't want to mess with that), and replaces them with door remotes that get the access of only the user.

Also renames them to [head]'s door remote

## Why / Balance
It doesn't make sense for heads to not be able to bolt or emergency access doors that they can access. (Big discussion in #35536)

## Technical details
I have _**NO IDEA**_ why there was a security door remote mapped in marathon, but there was, and I didn't want any errors or anything, so I just removed it.

Other than that is just YAML

## Media
Content Client_ZLAdwMqtQ1
Content Client_QDtHr01nJz

## Requirements

## Breaking changes
I hope not

**Changelog**
:cl:

- tweak: Door remotes now use the access of the owner, instead of having their own accesses.
- tweak: Door remotes are now named after their owner, not department.

### DIFF
--- a/Resources/Maps/marathon.yml
+++ b/Resources/Maps/marathon.yml
@@ -62244,19 +62244,6 @@ entities:
     - type: Transform
       pos: 3.6301537,20.758205
       parent: 30
-- proto: DoorRemoteSecurity
-  entities:
-  - uid: 2429
-    components:
-    - type: Transform
-      pos: -23.516384,50.539074
-      parent: 30
-    - type: Access
-      tags:
-      - HeadOfSecurity
-      - Security
-      - Armory
-      - Brig
 - proto: DoubleEmergencyOxygenTankFilled
   entities:
   - uid: 792
--- a/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
+++ b/Resources/Prototypes/Catalog/Fills/Lockers/heads.yml
@@ -13,7 +13,7 @@
     - id: CigPackGreen
       prob: 0.50
     - id: ClothingHeadsetAltCargo
-    - id: DoorRemoteCargo
+    - id: DoorRemoteQm
     - id: RubberStampApproved
     - id: RubberStampDenied
     - id: RubberStampQm
@@ -43,7 +43,7 @@
     - id: ClothingHeadsetAltCommand
     - id: ClothingOuterArmorCaptainCarapace
     - id: CommsComputerCircuitboard
-    - id: DoorRemoteCustom
+    - id: DoorRemoteCaptain
     - id: MedalCase
     - id: NukeDisk
     - id: PinpointerNuclear
@@ -132,7 +132,7 @@
       prob: 0.5
     - id: ClothingHeadsetCommand
     - id: ClothingNeckGoldmedal
-    - id: DoorRemoteService
+    - id: DoorRemoteHop
     - id: HoPIDCard
     - id: IDComputerCircuitboard
     - id: FundingAllocationComputerCircuitboard
@@ -166,7 +166,7 @@
     - id: ClothingEyesGlassesMeson
     - id: ClothingHandsGlovesColorYellow
     - id: ClothingHeadsetAltEngineering
-    - id: DoorRemoteEngineering
+    - id: DoorRemoteCe
     - id: CargoRequestEngineeringComputerCircuitboard
     - id: RCD
     - id: RCDAmmo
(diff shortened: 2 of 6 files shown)


keronshb:
Definitely have a lot of concerns with this.
It makes the remotes 0 risk to lose, which is a terrible balancing for it.
If this is the direction we're going to go, there's 0 reason to have them split up and unique you can just have 1 custom door remote and that's it. It doesn't make sense to have all these custom ones.

How does this effect airlocks and firelocks? People shouldn't be able to bolt those with remotes, it should require hacking or the AI.

VerinSenpai:
I was worried I was reading that incorrectly but I agree, not a fan of this one.

EmoGarbage404:
Agree with keron + very little justification on this PR.

superjj18:
Instead of using access, can we just add maints access and command access to all these remotes?(assuming they don't already have command access) Biggest issue is the QM for instance can't manipulate cargo maints doors because their cargo remote only work on "cargo" doors which is arbitrary.

No reason for the lower command members to not be able to manipulate maints access doors or command access doors given that any other command member/remote holder would also be able to override it.

Only issue with this is it would give all command members access to other departments via maints, so maybe the remotes need to meet ALL access requirements for a particular airlock.

A cargo maints door would require a remote that has both cargo AND maint door access, but any command member can manipulate a door that only has maints access requirements.

RangerXVII:
@keronshb 
> It makes the remotes 0 risk to lose, which is a terrible balancing for it. 

If only a passenger gets their hands on one, they can bolt off common and maints doors, which can help them escape if they are an antag or something. It doesn't let them open anything they don't have access to, though. Since security officers don't have door remotes they would have to wait for the AI or someone who does.

Also if you are in a position to steal someone's door remote, you are in a position to steal their ID, which would let you rival the AI as bolter of doors, but, the AI and other heads are still there, which does not make you too OP.

>If this is the direction we're going to go, there's 0 reason to have them split up and unique you can just have 1 custom door remote and that's it. It doesn't make sense to have all these custom ones.

This is the same as replacing every heads clothes with the command ones. Why would that get denied? It's boring. All the heads equipment should be unique. I guess I could change it but it would be weird for all the heads to use the same remote.

> How does this effect airlocks and firelocks? People shouldn't be able to bolt those with remotes, it should require hacking or the AI.

No one can bolt/emergency access firelocks. Only open/close them.

> Instead of using access, can we just add maints access and command access to all these remotes?(assuming they don't already have command access) Biggest issue is the QM for instance can't manipulate cargo maints doors because their cargo remote only work on "cargo" doors which is arbitrary.
>
> No reason for the lower command members to not be able to manipulate maints access doors or command access doors given that any other command member/remote holder would also be able to override it.

Command members should all be able to manipulate public doors like maints. Why not command too, because they ARE part of that department.

> Only issue with this is it would give all command members access to other departments via maints, so maybe the remotes need to meet ALL access requirements for a particular airlock.
>
> A cargo maints door would require a remote that has both cargo AND maint door access, but any command member can manipulate a door that only has maints access requirements.

That's not how access works, If a door has cargo and maints, it lets someone in if they have cargo **OR** maints access, making cargo useless, since everyone has maintenance.

BramvanZijp:
> Instead of using access, can we just add maints access and command access to all these remotes?(assuming they don't already have command access) Biggest issue is the QM for instance can't manipulate cargo maints doors because their cargo remote only work on "cargo" doors which is arbitrary.
> 
> No reason for the lower command members to not be able to manipulate maints access doors or command access doors given that any other command member/remote holder would also be able to override it.
> 
> Only issue with this is it would give all command members access to other departments via maints, so maybe the remotes need to meet ALL access requirements for a particular airlock.
> 
> A cargo maints door would require a remote that has both cargo AND maint door access, but any command member can manipulate a door that only has maints access requirements.

If this is bothering you in a round, ask the HoP or the CE if you can borrow their access configurator, and change the door from maint to maint/cargo, which then means the remote can access it.

Making it so remotes need ALL accesses does not make sense, so I would discourage it.

BramvanZijp:
Or if you mean the doors going FROM maints INTO cargo, then these should already be only cargo access and thus editable by your remote.

BramvanZijp:
Furthermore.
IF this is to be implemented, the remotes should keep their access, this behaviour set to DEFAULT to true on the component, and the captain should get their command remote back as opposed to the custom one.

RangerXVII:
¯\_(ツ)_/¯

Can't combat the speedclose.

I'm done with door remotes.

PR #38705, opened 2025-07-03
HOTFIX Fix pickup effects occurring with verb creation

### LINKED ISSUE #38248: MultiHandedItems will show popups from hovering in context menus
## Description

If you have no free hands and hover over a MultiHandedItem, you'll get a popup saying you need a free hand to pickup the item... except you didn't try to pick it up.

**Reproduction**

1. Spawn multi-handed item (e.g., salvage scrap)
2. Hold soda in _non-active hand_
3. Right click and mouse over the item
4. See popup saying you can't pick it up

**Screenshots**

Image

**Additional context**

This is happening because the MultiHandedItemSystem is throwing up this popup in response to GettingPickedUpAttemptEvent, which it shouldn't do—it should just silently cancel the event as applicable. The issue is I'm not seeing a good place for where this popup /should/ be thrown. My best guess is that the GettingPickedUpAttempt event should have a reason field that that gets put into, which then I guess gets displayed in `CanPickupToHand` if it gets passed a new `silent = false` argument? This feels a little messy though, and I'd like to hear if that's the way to go before I try PRing something.

perryprog:
## About the PR

1. Fixes multi handed items having a popup saying you can't pick them up when you hover over their entry in the item context list thing.
2. Fixes lubed items yeeting themselves away when you do the same.

## Why / Balance

Fixes #38248 and probably some other issues to be honest (see below).

## Technical details

Add an event that's between PickupAttempt and the actual pickup where you're put into a container. This kinda makes me wish we called "attempt" events "can" events since I think it's more clear that the former could have side effects (and would be called after the "can" event) while the latter wouldn't have side effects and is fully deterministic, but oh well.

## Requirements

## Breaking changes
`CanPickupToHand` and `CanPickupAnyHand` in `SharedHandsSystem` have a new `showPopup` parameter.

**Changelog**
:cl:
- fix: Fixed lubed items being thrown when looking at the pickup verb.
- fix: Fixed multihanded items showing a popup when looking at the pickup verb.
- fix: Fixed a bug with lubed handcuffs.

### DIFF
--- a/Content.Server/Lube/LubedSystem.cs
+++ b/Content.Server/Lube/LubedSystem.cs
@@ -23,7 +23,7 @@ public sealed class LubedSystem : EntitySystem
         base.Initialize();
 
         SubscribeLocalEvent<LubedComponent, ComponentInit>(OnInit);
-        SubscribeLocalEvent<LubedComponent, BeforeGettingEquippedHandEvent>(OnHandPickUp, after: [typeof(MultiHandedItemSystem)]);
+        SubscribeLocalEvent<LubedComponent, BeforeGettingEquippedHandEvent>(OnHandPickUp);
         SubscribeLocalEvent<LubedComponent, RefreshNameModifiersEvent>(OnRefreshNameModifiers);
     }
 


Content.Shared/ActionBlocker/ActionBlockerSystem.cs
```diff
@@ -167,16 +167,30 @@ public bool CanDrop(EntityUid uid)
...
+        /// If the item cannot be picked up, a localized string describing why, suitable for display.
+        /// </param>
+        /// <param name="physicalAttempt">
+        /// Whether or not this is from a physical interaction, and therefore may lead to side effects occuring. (Such
+        /// as items slipping out of hands.)
+        /// </param>
+        public bool CanPickup(EntityUid user, EntityUid item, out string? reason, bool physicalAttempt = false)
         {
-            var userEv = new PickupAttemptEvent(user, item);
+            var userEv = new PickupAttemptEvent(user, item, physicalAttempt);
```
perryprog: Huh this should just be a separate event; past me is so silly. Just noting this so I don't forget to fix later.

Princess-Cheeseballs: Perry forgor

perryprog: FRICK

slarticodefast:
I pushed some review changes.
- I reverted MultihandedItemSystem back to the attempt event, because that should be cancelled if you cannot pick the item up. Otherwise the action blocker will be wrong. To handle the popup when trying to pickup, but not when looking at a verb I added a new parameter.
- Added an equivalent `BeforeEquippingHandEvent` raised on the user for future use.
- Removed raising the pickup event when only switching hands.
- Added some code comments.

Looks good to me now, thanks for the bugfix!

VasilisThePikachu (approved):

perryprog:
Lol, what happened that made this become a hotfix?

Princess-Cheeseballs:
> Lol, what happened that made this become a hotfix?

war...

Errant-4:
> Lol, what happened that made this become a hotfix?

Nothing new happened, but if I simply merged it into master, it would not have been included in the upcoming release

